PR #43, opened 2017-02-03
Navigation UI

frederoni:
First draft of the Nav UI

simulator screen shot 8 feb 2017 14 12 46

- [x] Customize UI via appearance proxy or similar
- [ ] UI Tests
- [x] Examples
- [x] Storyboard support
- [x] Write documentation

PR #44, opened 2017-02-21
MapboxDirections.swift, iOS SDK v3.4.2

1ec5:
Upgraded MapboxDirections.swift to a Carthage-powered commit on master. Upgraded Mapbox iOS SDK to v3.4.1.

/cc @frederoni

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-MapboxNavigation.xcworkspace
+Example/MapboxNavigation.xcworkspace
 
 .DS_Store
 
@@ -8,7 +8,7 @@ xcuserdata
 *.xcuserstate
 *.xcscmblueprint
 
-Pods/
+Example/Pods/
 
 Carthage/Build
 Carthage/Checkouts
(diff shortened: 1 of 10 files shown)


frederoni:
This is already happening in https://github.com/mapbox/MapboxNavigation.swift/pull/43

1ec5:
Builds are failing because `SWIFT_VERSION` is unspecified for the Polyline target in the example workspace’s Pods project. CocoaPods is pulling in Polyline via the dependency declared in MapboxDirections.swift.podspec. I don’t know why CocoaPods is failing to notice the .swift-version file there that specifies Swift 3. In mapbox/MapboxDirections.swift#102, we had no problem with a similar setup: a CocoaPods Podfile (for the DirectionsExample project) that depended on a podspec in an otherwise Carthage-managed repository (MapboxDirections.swift), which in turn depended on a podspec in an otherwise Carthage-managed repository (Polyline). So I’m stumped as to why builds are failing here.

1ec5:
I suspect we got lucky in mapbox/MapboxDirections.swift#102. Over here, there’s actually one additional level: a Podfile that depends on three levels of Carthage projects’ podspecs (MapboxNavigation, MapboxDirections, Polyline). Since this setup doesn’t seem to have a hope of passing, we’re going to move all the examples over to Carthage in #43.

PR #48, opened 2017-02-27
Use Carthage submodules

frederoni:
Switched to Carthage submodules.
Note that you should  `--use-submodules` along with every checkout or update command.

Added a workspace that links to all dependencies we're actively developing for convenience.

@bsudekum 👀


frederoni:
There are two viable solutions for working with development libs. One is using the `--use-submodules` flag and the other one is using a workspace together with the `--no-build` flag but none of them work in conjunction with a binary dependency which we have in this repo. Let's switch back to CocoaPods for development until we've figured out something better.

PR #51, opened 2017-02-28
Add AWS Polly voice support

bsudekum:
Third try's a charm. Adds optional support for AWS Polly voice. 

@frederoni  this is pinned to https://github.com/Project-OSRM/osrm-text-instructions.swift/pull/8 which should be ready to go.

/cc @1ec5


MapboxNavigation/RouteStepFormatter.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
```
frederoni: We have a RouteStepFormatter in NavUI as well, let's get rid of the one in NavUI since this is already public.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -0,0 +1,321 @@
...
+    deinit {
+        if let currentItem = audioPlayer.currentItem {
+            currentItem.removeObserver(self, forKeyPath: "status")
+        }
+        suspendNotifications()
+        speechSynth.stopSpeaking(at: .word)
+    }
+    
+    func resumeNotifications() {
+        NotificationCenter.default.addObserver(self, selector: #selector(self.alertLevelDidChange(notification:)), name: RouteControllerAlertLevelDidChange, object: nil)
```
frederoni: Nit: Drop self

Examples/Swift/ViewController.swift
```diff
@@ -1,26 +1,29 @@
+
```
frederoni: Drop this line

MapboxNavigation/RouteStepFormatter.swift
bsudekum: Removed this file from NavUI since NavUI depends on Nav.

Podfile
```diff
@@ -10,8 +10,9 @@ target 'MapboxNavigationUI' do
   pod 'MapboxDirections.swift', '~> 0.8'
   pod 'MapboxGeocoder.swift', '~> 0.6'
   pod 'SDWebImage', '~> 4.0.0-beta2'
-  pod 'OSRMTextInstructions', :git => "git@github.com:Project-OSRM/osrm-text-instructions.swift.git", :branch => "master"
+  pod 'OSRMTextInstructions', :git => 'https://github.com/Project-OSRM/osrm-text-instructions.swift.git', :commit => 'bdd962bf5593f951bbbbd26435f732a829dacda7'
```
bsudekum: Let's move off master

frederoni:
We need to add AWSPolly as a dependency in the Cartfile as well

1ec5:
> Adds optional support for AWS Polly voice.

To be clear, this support isn’t optional. Due to embedding the AWS Polly SDK, the application will experience a size increase, potential memory increase, and startup time increase, regardless of whether Polly ends up being used. To truly make Polly optional, we’d need to [weakly link that framework](https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html#//apple_ref/doc/uid/20002378-107026), which would be feasible but nontrivial.

PR #57, opened 2017-03-04
Snap to route

frederoni:
WIP
Outlined how snap to route could work.

@1ec5 👀


MapboxNavigationUI/MGLMapView+MGLNavigationAdditions.h
```diff
@@ -0,0 +1,7 @@
+#import <Mapbox/Mapbox.h>
+
+@interface MGLMapView (MGLNavigationAdditions)
+
+- (void)locationManager:(CLLocationManager *)manager didUpdateLocations:(NSArray *)locations;
```
frederoni: This should be exposed in [mapbox-gl-native](https://github.com/mapbox/mapbox-gl-native)

1ec5: As I mentioned in mapbox/mapbox-gl-native#6867, I think the map SDK should instead expose more specific methods like `-didUpdateLocationWithUserTrackingAnimated:`.

bsudekum: Whichever method we choose to override, I'm okay with doing this logic here for the time being while we experiment.

MapboxNavigationUI/MGLMapView.swift
```diff
@@ -174,3 +174,22 @@ extension MGLMapView {
         }
     }
 }
+
+protocol NavigationMapViewDelegate {
+    func navigationMapView(_ mapView: NavigationMapView, shouldUpdateTo location: CLLocation) -> CLLocation?
+}
+
+public class NavigationMapView: MGLMapView {
```
1ec5: This code should go in a separate NavigationMapView.swift file.

bsudekum:
Should be make snapping an optional param?

bsudekum:
I think we should also look into snapping bearing/heading.

bsudekum:
@1ec5 this is ready for review again. Made snapping optional with https://github.com/mapbox/MapboxNavigation.swift/pull/57/commits/15da62ddb7bd8720283dfe100c3e4a2e80b2f5b7

MapboxNavigation/RouteController.swift
```diff
@@ -24,6 +24,12 @@ open class RouteController: NSObject {
     
     
     /*
+     If true, the user puck is snapped to closest location on the route.
+    */
+    public var snapUserToRoute = false
```
1ec5: “Snap user to route” sounds like a command, so this sounds like an action method instead of a property that can be set. (In Objective-C, an action method and a property getter call can look identical.) Also, it sounds funny to say that we’re moving the user themselves; we’re actually moving the user location annotation.

Let’s call this property `snapsUserLocationAnnotationToRoute`. It’s a mouthful, but the developer won’t have to call it often anyways.

MapboxNavigation/Constants.swift
```diff
@@ -20,6 +20,10 @@ public let RouteControllerShouldReroute = Notification.Name(MBRouteControllerSho
 */
 public var RouteControllerMaximumDistanceBeforeRecalculating: CLLocationDistance = 50
 
+/*
+ Accepted deviation excluding horizontal accuracy before the user is considered to be off route.
+ */
+public var RouteControllerHorizontalAccuracyAnomaly: CLLocationDistance = 10
```
1ec5: This global variable doesn’t represent an anomaly; it represents the maximum deviation distance that can be considered an anomaly. How about calling this global `RouteControllerUserLocationSnappingDistance`, to describe the desired output instead of one specific input (which may not be the only output in the future)?

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -218,6 +219,30 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
     }
 }
 
+// MARK: NavigationMapViewDelegate
+
+extension RouteMapViewController: NavigationMapViewDelegate {
+    
+    func navigationMapView(_ mapView: NavigationMapView, shouldUpdateTo location: CLLocation) -> CLLocation? {
```
1ec5: This is a weird name for a method that returns a CLLocation. The name implies that it returns only a Boolean indicating whether the map view should… update.

1ec5: This will have to do for now, although it’s going to look _really_ weird in Objective-C.

Eventually, as part of mapbox/mapbox-gl-native#6867, we’ll want the MGLMapViewDelegate method upstream to accept an already positioned MGLUserLocation, rather than the raw CLLocationCoordinate2D, so that the heading can also be manipulated.

MapboxNavigationUI/MGLMapView+MGLNavigationAdditions.h
```diff
@@ -0,0 +1,7 @@
+#import <Mapbox/Mapbox.h>
+
+@interface MGLMapView (MGLNavigationAdditions)
```
1ec5: Rather than redeclare the method below, let’s have this class extension conform to the CLLocationManagerDelegate protocol.

1ec5: Also add a `FIXME:` comment noting that this class extension will be reworked following mapbox/mapbox-gl-native#6867.

MapboxNavigationUI/RouteMapViewController.swift
1ec5: Maybe add `@objc(navigationMapView:shouldUpdateToLocation:)` to help with Objective-C.

PR #58, opened 2017-03-04
Improve off route calculation

frederoni:
`userIsOnRoute(_:)` previously used a constant of 50 meters no matter what the horizontal accuracy was. This PR tweaks that by using a lower value if we know that the accuracy is lower than `RouteControllerMaximumDistanceBeforeRecalculating`.

This greatly improves off route detection for cycling and driving. It also improves the snap-to-route behavior.

@1ec5 @bsudekum 👀

### DIFF
--- a/MapboxNavigation/RouteController.swift
+++ b/MapboxNavigation/RouteController.swift
@@ -112,7 +112,13 @@ extension RouteController: CLLocationManagerDelegate {
         let metersInFrontOfUser = location.speed * RouteControllerDeadReckoningTimeInterval
         let locationInfrontOfUser = location.coordinate.coordinate(at: metersInFrontOfUser, facing: location.course)
         let newLocation = CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
-        return newLocation.isWithin(RouteControllerMaximumDistanceBeforeRecalculating, of: routeProgress.currentLegProgress.currentStep)
+        var radius = RouteControllerMaximumDistanceBeforeRecalculating
+        
+        if 0...RouteControllerMaximumDistanceBeforeRecalculating ~= location.horizontalAccuracy {
+            radius = location.horizontalAccuracy + RouteControllerHorizontalAccuracyAnomaly
+        }
+        
+        return newLocation.isWithin(radius, of: routeProgress.currentLegProgress.currentStep)
     }
     
     func monitorStepProgress(_ location: CLLocation) {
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteController.swift
```diff
@@ -112,7 +112,13 @@ extension RouteController: CLLocationManagerDelegate {
         let metersInFrontOfUser = location.speed * RouteControllerDeadReckoningTimeInterval
         let locationInfrontOfUser = location.coordinate.coordinate(at: metersInFrontOfUser, facing: location.course)
         let newLocation = CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
-        return newLocation.isWithin(RouteControllerMaximumDistanceBeforeRecalculating, of: routeProgress.currentLegProgress.currentStep)
+        var radius = RouteControllerMaximumDistanceBeforeRecalculating
+        
+        if 0...RouteControllerMaximumDistanceBeforeRecalculating ~= location.horizontalAccuracy {
+            radius = location.horizontalAccuracy + RouteControllerHorizontalAccuracyAnomaly
```
1ec5: More succinctly, we could set `RouteControllerMaximumDistanceBeforeRecalculating` to 45 meters, then set `radius` to `min(location.horizontalAccuracy, RouteControllerMaximumDistanceBeforeRecalculating) + RouteControllerHorizontalAccuracyAnomaly`.

frederoni:
Before we can merge this PR we need to adapt the tests and figure out how to prevent continuous rerouting when being stationary off route.

bsudekum:
I think this is superseded by https://github.com/mapbox/MapboxNavigation.swift/pull/57?

1ec5:
I’m not so sure about that: the point of this PR was to avoid hard-coding `RouteControllerUserLocationSnappingDistance`, or at least to make it dependent on the user location’s `horizontalAccuracy`.

frederoni:
It should've been superseded by #57 including @1ec5's [improvements](https://github.com/mapbox/MapboxNavigation.swift/pull/57/files#diff-bbbf10e7acabbe34ecfd2d578a458e67R121) suggested here https://github.com/mapbox/MapboxNavigation.swift/pull/58#discussion_r104280767 but snap to route seems to be broken on latest#master.

1ec5:
> snap to route seems to be broken on latest#master

Is this still the case?

frederoni:
This branch is stale and superseded by a couple of other branches.

PR #70, opened 2017-03-09
Include voice in NavUI

### LINKED ISSUE #69: Refactor RouteVoiceController
If the developer buys into using UI, they should not have to also worry about voice. Let's make initialization and management controlled within NavigationUI.

/cc @frederoni

frederoni:
Fixes #69 

RouteViewController now keeps a ref to a RouteVoiceController.
An optional identityPoolId can be set if Polly is preferred over default speech synthesizer.
Also added a boolean to enable/disable voice.

Tail work for this PR would be to make the AWSPolly dependency optional but we might want to wait with that until this accepted proposal has been released: https://github.com/apple/swift-evolution/blob/master/proposals/0075-import-test.md

@bsudekum 👀

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -19,15 +19,11 @@ class ViewController: UIViewController, MGLMapViewDelegate {
     @IBOutlet weak var toggleNavigationButton: UIButton!
     @IBOutlet weak var howToBeginLabel: UILabel!
     
-    var routeVoiceController = RouteVoiceController()
     
     override func viewDidLoad() {
         super.viewDidLoad()
         
         mapView.delegate = self
-        
-        // If you want to use Polly instead of the built-in speech synthesizer, just set the identityPoolId
-        //routeVoiceController.identityPoolId = "<#Your AWS IdentityPoolId.#>"
         
         mapView.userTrackingMode = .follow
         resumeNotifications()
@@ -35,7 +31,6 @@ class ViewController: UIViewController, MGLMapViewDelegate {
     
     deinit {
         suspendNotifications()
-        routeVoiceController.suspendNotifications()
         navigation?.suspendLocationUpdates()
     }
     
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -151,6 +151,12 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         // 1. the route the user will take
         // 2. A `Directions` class, used for rerouting.
         let viewController = NavigationUI.routeViewController(for: route)
+        
+        // If you'd like to use AWS Polly, provide your IdentityPoolId below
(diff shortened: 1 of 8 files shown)


Examples/Swift/ViewController.swift
```diff
@@ -19,15 +19,16 @@ class ViewController: UIViewController, MGLMapViewDelegate {
     @IBOutlet weak var toggleNavigationButton: UIButton!
     @IBOutlet weak var howToBeginLabel: UILabel!
     
-    // `identityPoolId` is a required value for using AWS Polly voice instead of iOS's built in AVSpeechSynthesizer
-    // You can get a token here: http://docs.aws.amazon.com/mobile/sdkforios/developerguide/cognito-auth-aws-identity-for-ios.html
-    var routeVoiceController = RouteVoiceController(identityPoolId: "<#Your AWS IdentityPoolId. Remove Argument if you do not want to use AWS Polly#>")
+    var routeVoiceController = RouteVoiceController()
```
frederoni: Oh, this can be removed. It's just that the example tries to show how to do this w/ _and_ w/o NavigationUI.

bsudekum: Ah I see. Maybe we can make a comment on this.

frederoni: Removed the voice controller from the example.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -45,15 +51,27 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
+    /**
+     `identityPoolId` is a required value for using AWS Polly voice instead of iOS's built in AVSpeechSynthesizer.
+     You can get a token here: http://docs.aws.amazon.com/mobile/sdkforios/developerguide/cognito-auth-aws-identity-for-ios.html
+     */
+    public var identityPoolId: String? {
+        didSet {
+            if let poolId = identityPoolId {
+                let credentialsProvider = AWSCognitoCredentialsProvider(regionType:regionType, identityPoolId: poolId)
+                let configuration = AWSServiceConfiguration(region:regionType, credentialsProvider:credentialsProvider)
+                AWSServiceManager.default().defaultServiceConfiguration = configuration
```
bsudekum: Just tested this out. I added 

```
var routeVoiceController = RouteVoiceController()

routeVoiceController.identityPoolId = "pool-id"
```

Both voices occur are spoken. I think we need to set `useDefaultVoice ` here as well.

bsudekum: Nvm, it looks like this is being inited twice. `useDefaultVoice` code looks good

frederoni: Oh, I see. Yes, that would add another voice controller. Set the identityPoolId on the existing voiceController `routeViewController.voiceController.identityPoolId = ""`

PR #74, opened 2017-03-13
Minor UI adjustments

bsudekum:
before

simulator screen shot mar 13 2017 12 03 31 pm

after
simulator screen shot mar 13 2017 11 58 11 am

* Increase way name font size
* improves padding
* Now we fallback to the instruction if there is no way name or destination

/cc @1ec5

### DIFF
--- a/MapboxNavigationUI/RouteTableViewController.swift
+++ b/MapboxNavigationUI/RouteTableViewController.swift
@@ -43,14 +43,10 @@ class RouteTableViewController: UIViewController {
         
         if routeProgress.durationRemaining < 5 {
             headerView.distanceRemaining.text = nil
-        } else {
-            headerView.distanceRemaining.text = distanceFormatter.string(from: routeProgress.distanceRemaining)
-        }
-        
-        if routeProgress.durationRemaining < 60 {
+        } else if routeProgress.durationRemaining < 60 {
             headerView.timeRemaining.text = String.localizedStringWithFormat(NSLocalizedString("LESS_THAN", value: "<%@", comment: "Format string for less than; 1 = duration remaining"), dateComponentsFormatter.string(from: 61)!)
         } else {
-            headerView.timeRemaining.text = dateComponentsFormatter.string(from: routeProgress.durationRemaining)
+            headerView.distanceRemaining.text = distanceFormatter.string(from: routeProgress.distanceRemaining)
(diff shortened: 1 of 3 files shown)


1ec5:
> Now we fallback to the instruction if there is no way name or destination

With an instruction like “Turn left” being so much more prominent than the distance, is there a risk that the user would misinterpret the “Turn left” to mean “Turn left now” (and misinterpret “2 mi” to mean “then go 2 miles”)? Instead, what if we hide the normal distance label and put the distance in place of the road name?

bsudekum:
> (and misinterpret “2 mi” to mean “then go 2 miles”)?

I mean, this mistake could be made even if there is a way name. 

simulator screen shot mar 13 2017 1 26 37 pm

I think shuffling things around in the UI is confusing for the user since they expect the distance to be in the lower left. In my opinion, this feels more helpful than putting nothing.

1ec5:
> I think shuffling things around in the UI is confusing

I agree. Let’s go with this for now and revisit the UI in #75.

MapboxNavigationUI/RouteTableViewController.swift
```diff
@@ -43,14 +43,10 @@ class RouteTableViewController: UIViewController {
...
-        } else {
-            headerView.distanceRemaining.text = distanceFormatter.string(from: routeProgress.distanceRemaining)
-        }
-        
-        if routeProgress.durationRemaining < 60 {
+        } else if routeProgress.durationRemaining < 60 {
             headerView.timeRemaining.text = String.localizedStringWithFormat(NSLocalizedString("LESS_THAN", value: "<%@", comment: "Format string for less than; 1 = duration remaining"), dateComponentsFormatter.string(from: 61)!)
         } else {
-            headerView.timeRemaining.text = dateComponentsFormatter.string(from: routeProgress.durationRemaining)
+            headerView.distanceRemaining.text = distanceFormatter.string(from: routeProgress.distanceRemaining)
```
1ec5: The new logic seems to ignore half of the UI on each update:

* If there’s very little time left, hide the distance but don’t update the duration.
* Otherwise, if there isn’t a lot of time left, show a “less than” duration but don’t update the distance.
* Otherwise, if there’s plenty of time left, update the distance but don’t update the duration.

PR #80, opened 2017-03-17
clear default values from the storyboard labels

joneswah:
Because the storyboard has default values, when the view first loads there is flash from the default values in the storyboard until the real values are shown. This is quite obvious when the users locale is metric because you see "8.3 mi" before the actual metre/km value is shown.

there are a number of ways to address this minor issue. you could update the labels in the storyboard or do it in this view or in the view controller. I thought it might be cleaner in the view awakeFromNib()?


bsudekum:
@joneswah thanks for the PR. @1ec5 do you know of any less hackish ways to accomplish this? Is there a way to only have values show up in the storyboard? Maybe we can set these values sooner.

PR #84, opened 2017-03-22
Don't continual reroute user if standing still, allow user time to get to route start

bsudekum:
This addresses two issues:

1. A user is standing still far from the route. Consider if someone starts a route in a large parking lot. Yes, they are from the route, but we don't need to reroute them. 
1. The user is moving to/from the start. Again consider a large parking lot. If the user is traveling towards the beginning of the route, we do not need to reroute them even if they are greater than the max reroute distance.

/cc @frederoni @1ec5

### DIFF
--- a/MapboxNavigation/Constants.swift
+++ b/MapboxNavigation/Constants.swift
@@ -50,6 +50,10 @@ public var RouteControllerHighAlertInterval: TimeInterval = 15
 */
 public var RouteControllerManeuverZoneRadius: CLLocationDistance = 40
 
+
+/*
+ Maximum number of seconds the user can travel away from the start of the route before rerouting occurs
+*/
 public var MaxSecondsSpentTravelingAwayFromStartOfRoute: TimeInterval = 3
 
 
--- a/MapboxNavigation/Constants.swift
+++ b/MapboxNavigation/Constants.swift
@@ -50,6 +50,8 @@ public var RouteControllerHighAlertInterval: TimeInterval = 15
 */
 public var RouteControllerManeuverZoneRadius: CLLocationDistance = 40
 
+public var MaxSecondsSpentTravelingAwayFromStartOfRoute: TimeInterval = 3
+
 
 /*
  Distance in meters for the minimum length of a step for giving a `medium` alert.
--- a/MapboxNavigation/RouteController.swift
+++ b/MapboxNavigation/RouteController.swift
@@ -11,7 +11,7 @@ open class RouteController: NSObject {
     
     var lastUserDistanceToStartOfRoute = Double.infinity
     
-    var userIsMovingAwayFromStartCounter: Int = 0
+    var lastTimeStampSpentMovingAwayFromStart = Date()
     
     /*
      Monitor users location along route.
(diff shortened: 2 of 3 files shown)


MapboxNavigation/RouteController.swift
```diff
@@ -110,9 +107,46 @@ extension RouteController: CLLocationManagerDelegate {
             }
         }
         
+        if routeProgress.currentLegProgress.stepIndex == 0 {
+            let step = routeProgress.currentLegProgress.currentStepProgress.step
+            let stepLocation = CLLocation(latitude: step.coordinates!.first!.latitude, longitude: step.coordinates!.first!.longitude)
+            
+            var currentUserToStartOfRoute = location.distance(from: stepLocation)
+            if currentUserToStartOfRoute > lastUserDistanceToStartOfRoute {
```
1ec5: This seems to calculate the user’s course by storing and comparing the last _n_ distances. Could we achieve the same effect or better by looking at whether `location.course` points toward the route? You could do that by comparing `location.course` to the direction of the closest path to the route line (using `CLLocationCoordinate2D.direction(to:)`).

MapboxNavigation/RouteController.swift
```diff
@@ -110,9 +107,46 @@ extension RouteController: CLLocationManagerDelegate {
             }
         }
         
+        if routeProgress.currentLegProgress.stepIndex == 0 {
+            let step = routeProgress.currentLegProgress.currentStepProgress.step
+            let stepLocation = CLLocation(latitude: step.coordinates!.first!.latitude, longitude: step.coordinates!.first!.longitude)
```
bsudekum: Oh yeah, I guess the first steps maneuver location is the beginning of the route

MapboxNavigation/RouteController.swift
```diff
@@ -110,9 +107,48 @@ extension RouteController: CLLocationManagerDelegate {
             }
         }
         
+        if routeProgress.currentLegProgress.currentStepProgress.step.maneuverType == .depart && !userIsOnRoute(location) {
+            let step = routeProgress.currentLegProgress.currentStepProgress.step
```
1ec5: Move this line up so you don't have to repeat yourself.

bsudekum:
@1ec5 I think this is a good start. We should look at using course in a seperate PR.

MapboxNavigation/Constants.swift
```diff
@@ -50,6 +50,8 @@ public var RouteControllerHighAlertInterval: TimeInterval = 15
 */
 public var RouteControllerManeuverZoneRadius: CLLocationDistance = 40
 
+public var MaxSecondsSpentTravelingAwayFromStartOfRoute: TimeInterval = 3
```
1ec5: This public global should have documentation.

MapboxNavigation/RouteController.swift
```diff
@@ -110,9 +107,48 @@ extension RouteController: CLLocationManagerDelegate {
...
+            
+            // Don't check `userIsOnRoute` if the user has not moved
+            guard userSnappedDistanceToClosestCoordinate != lastUserDistanceToStartOfRoute else {
+                lastUserDistanceToStartOfRoute = userSnappedDistanceToClosestCoordinate
+                return
+            }
+            
+            lastUserDistanceToStartOfRoute = userSnappedDistanceToClosestCoordinate
+            
+            if userSnappedDistanceToClosestCoordinate > lastUserDistanceToStartOfRoute {
```
1ec5: `userSnappedDistanceToClosestCoordinate` is never greater than `lastUserDistanceToStartOfRoute`, because the line right above sets `lastUserDistanceToStartOfRoute` to be equal to `userSnappedDistanceToClosestCoordinate`.

MapboxNavigation/Constants.swift
```diff
@@ -52,6 +52,12 @@ public var RouteControllerManeuverZoneRadius: CLLocationDistance = 40
 
 
 /*
+ Maximum number of seconds the user can travel away from the start of the route before rerouting occurs
+*/
+public var MaxSecondsSpentTravelingAwayFromStartOfRoute: TimeInterval = 3
```
1ec5: We don’t consider the start of the route at all anymore, do we? https://github.com/mapbox/MapboxNavigation.swift/pull/84#discussion_r107560286

bsudekum: True. Maximum time spent moving away from first step

frederoni:
This issue was introduced by snap to route in #57 but also fixed in the same PR https://github.com/mapbox/MapboxNavigation.swift/pull/57/files#diff-36ca7eeae4537f27ed429d4f64eaa209R157. Does this leave the previous logic obsolete or will it work well together?

PR #86, opened 2017-03-24
Posting a notification when new route is obtained

azharaali:
We are building a ride sharing App and need to share one users route with another user, the primary user is using the Mapbox Turn-by-Turn Navigation UI. To do this we need to have access to the polyline as the route changes. So we are proposing adding in a new notification that users can register to receive when the route changes. The existing notifications are triggered before a reroute occurs and we need to know the actual route that was determined after the rerouting process.


bsudekum:
This makes sense to me. @1ec5 good with this?

MapboxNavigation/MBRouteController.m
```diff
@@ -13,3 +13,6 @@
 NSString *const MBRouteControllerNotificationProgressDidChange  = @"RouteControllerProgressDidChange";
 NSString *const MBRouteControllerAlertLevelDidChange            = @"RouteControllerAlertLevelDidChange";
 NSString *const MBRouteControllerShouldReroute                  = @"RouteControllerShouldReroute";
+
+NSString *const MBRouteControllerDidRerouteKey = @"RouteControllerDidRerouteKey";
+NSString *const MBRouteControllerDidRerouteNotificationRouteProgressKey = @"RouteControllerDidRerouteNotificationRouteProgressKey";
```
1ec5: No need to include “Key” in the values of the constants.

MapboxNavigation/Constants.swift
```diff
@@ -15,6 +15,9 @@ public let RouteControllerProgressDidChange = Notification.Name(MBRouteControlle
 public let RouteControllerAlertLevelDidChange = Notification.Name(MBRouteControllerAlertLevelDidChange)
 public let RouteControllerShouldReroute = Notification.Name(MBRouteControllerShouldReroute)
 
+public let RouteControllerDidReroute = Notification.Name(MBRouteControllerDidRerouteKey)
+public let RouteControllerDidRerouteNotifcationRouteProgressKey = Notification.Name(MBRouteControllerDidRerouteNotificationRouteProgressKey)
```
1ec5: Maybe we should rename `MBRouteControllerAlertLevelDidChangeNotificationRouteProgressKey` to be more generic and reuse the consolidated key in the new notification.

azharaali:
@1ec5 thanks for the feedback, i have made the changes you suggested and submitted another commit.

MapboxNavigation/Constants.swift
```diff
@@ -6,7 +6,7 @@ public let RouteControllerProgressDidChangeNotificationProgressKey = MBRouteCont
 public let RouteControllerProgressDidChangeNotificationLocationKey = MBRouteControllerProgressDidChangeNotificationLocationKey
 public let RouteControllerProgressDidChangeNotificationSecondsRemainingOnStepKey = MBRouteControllerProgressDidChangeNotificationSecondsRemainingOnStepKey
 
-public let RouteControllerAlertLevelDidChangeNotificationRouteProgressKey = MBRouteControllerAlertLevelDidChangeNotificationRouteProgressKey
+public let RouteControllerDidChangeNotificationRouteProgress = MBRouteControllerDidChangeNotificationRouteProgress
```
1ec5: Nit: “Key” should be part of the _name_ of the constant, just not its string value.

bsudekum: @azharaali have a sec to fix this?

azharaali: just made the suggested fix

bsudekum:
Thanks @azharaali

frederoni:
> The build is failing because of #87, but it’s safe to merge despite that failure, because the merge commit won’t be affected by that bug.

It also fails because the constant string in Obj-C `MBRouteControllerDidReroute` doesn't match in Constants.swift `Notification.Name(MBRouteControllerDidRerouteKey)
`

It also struck me that Navigation.swift is not responsible for the actual rerouting, it only informs that we should reroute, so the notification should live inside NavUI instead of Navigation.swift. Perhaps also use the delegate pattern we use there instead of the notification pattern?

MapboxNavigationUI/RouteViewController.swift
```diff
@@ -182,12 +182,17 @@ public class RouteViewController: NavigationPulleyViewController {
                 
                 strongSelf.mapViewController?.notifyDidReroute(route: route)
                 strongSelf.tableViewController?.notifyDidReroute()
+                
+                // posting a notification that could be listened to get new route info externally
+                NotificationCenter.default.post(name: RouteControllerDidReroute, object: strongSelf.routeController, userInfo: [
+                    RouteControllerDidRerouteNotifcationRouteProgressKey: strongSelf.routeController.routeProgress
```
bsudekum: The string `RouteControllerDidRerouteNotifcationRouteProgressKey` does not exist.

1ec5: There's a typo: "Notifcation" should be "Notification".

PR #88, opened 2017-03-24
Add course snapping [wip]

bsudekum:
Adds course snapping to the `snapsUserLocationAnnotationToRoute` option.

More ideas on implemenation should be explored.

/cc @1ec5

### DIFF
--- a/MapboxNavigationUI/RouteMapViewController.swift
+++ b/MapboxNavigationUI/RouteMapViewController.swift
@@ -243,12 +243,15 @@ extension RouteMapViewController: NavigationMapViewDelegate {
         let closest = closestCoordinate(on: coords, to: location.coordinate)!
         let slicedLine = polyline(along: coords, from: closest.coordinate, to: coords.last)
         
+        
+        let userDistanceBuffer = location.speed * RouteControllerDeadReckoningTimeInterval
+        
         // Get closest point infront of user
-        let infrontPointOne = coordinate(at: 2, fromStartOf: slicedLine)!
+        let infrontPointOne = coordinate(at: userDistanceBuffer, fromStartOf: slicedLine)!
         let infrontClosest = closestCoordinate(on: coords, to: infrontPointOne)!
         
         // Get closest point behind in front of user
-        let infrontPointTwo = coordinate(at: 4, fromStartOf: slicedLine)!
+        let infrontPointTwo = coordinate(at: userDistanceBuffer * 2, fromStartOf: slicedLine)!
         let behindClosest = closestCoordinate(on: coords, to: infrontPointTwo)!
         
         // Get direction of these points
@@ -258,9 +261,13 @@ extension RouteMapViewController: NavigationMapViewDelegate {
         let normalizedInfrontDirection = wrap(infrontDirection, min: 0, max: 360)
         let normalizedBehindDirection = wrap(behindDirection, min: 0, max: 360)
         
-        let averagedCourse = (normalizedInfrontDirection + normalizedBehindDirection) / 2
-        let course = differenceBetweenAngles(location.course, averagedCourse) <= RouteControllerMaximumAllowedDegreeOffsetForTurnCompletion ? averagedCourse : location.course
+        let diffBetweenAngleInFront = differenceBetweenAngles(location.course, normalizedInfrontDirection)
+        let diffBetweenAngleBehind = differenceBetweenAngles(location.course, normalizedBehindDirection)
         
+        let averageDifference = (diffBetweenAngleInFront + diffBetweenAngleBehind) / 2
+        let averagedCourse = (normalizedBehindDirection + normalizedInfrontDirection) / 2
+        
+        let course = averageDifference <= RouteControllerMaximumAllowedDegreeOffsetForTurnCompletion ? averagedCourse : location.course
         
         return CLLocation(coordinate: newCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: course, speed: location.speed, timestamp: location.timestamp)
     }
--- a/MapboxNavigationUI/RouteMapViewController.swift
+++ b/MapboxNavigationUI/RouteMapViewController.swift
@@ -246,24 +246,23 @@ extension RouteMapViewController: NavigationMapViewDelegate {
         let userDistanceBuffer = location.speed * RouteControllerDeadReckoningTimeInterval
         
         // Get closest point infront of user
-        let infrontPointOne = coordinate(at: userDistanceBuffer, fromStartOf: slicedLine)!
-        let infrontClosest = closestCoordinate(on: coords, to: infrontPointOne)!
+        let pointOneSliced = coordinate(at: userDistanceBuffer, fromStartOf: slicedLine)!
+        let pointOneClosest = closestCoordinate(on: coords, to: pointOneSliced)!
         
-        // Get closest point behind in front of user
-        let infrontPointTwo = coordinate(at: userDistanceBuffer * 1.5, fromStartOf: slicedLine)!
-        let behindClosest = closestCoordinate(on: coords, to: infrontPointTwo)!
+        let pointTwoSliced = coordinate(at: userDistanceBuffer * 1.5, fromStartOf: slicedLine)!
+        let pointTwoClosest = closestCoordinate(on: coords, to: pointTwoSliced)!
         
         // Get direction of these points
-        let infrontDirection = closest.coordinate.direction(to: infrontClosest.coordinate)
-        let behindDirection = closest.coordinate.direction(to: behindClosest.coordinate)
+        let pointOneDirection = closest.coordinate.direction(to: pointOneClosest.coordinate)
+        let pointTwoDirection = closest.coordinate.direction(to: pointTwoClosest.coordinate)
         
-        let wrappedInFront = wrap(infrontDirection, min: -180, max: 180)
-        let wrappedBehind = wrap(behindDirection, min: -180, max: 180)
+        let wrappedPointOne = wrap(pointOneDirection, min: -180, max: 180)
+        let wrappedPointTwo = wrap(pointTwoDirection, min: -180, max: 180)
         
-        let relativeAngleInFront = differenceBetweenAngles(location.course, wrappedInFront)
-        let relativeAngleBehind = differenceBetweenAngles(location.course, wrappedBehind)
+        let relativeAnglepointOne = differenceBetweenAngles(location.course, wrappedPointOne)
+        let relativeAnglepointTwo = differenceBetweenAngles(location.course, wrappedPointTwo)
         
-        let averageRelativeAngle = (relativeAngleInFront + relativeAngleBehind) / 2
+        let averageRelativeAngle = (relativeAnglepointOne + relativeAnglepointTwo) / 2
 
         let absoluteDirection = wrap(location.course + averageRelativeAngle, min: 0 , max: 360)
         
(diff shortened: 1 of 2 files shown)


bsudekum:
Also tried implementing this with 

```swift
let coords = routeController.routeProgress.route.coordinates!
        guard let closest = closestCoordinate(on: coords, to: location.coordinate) else {
            return CLLocation(coordinate: newCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: location.course, speed: location.speed, timestamp: location.timestamp)
        }
        let pointInfrontOfClosest = polyline(along: coords, within: 2, of: closest.coordinate).last!
        let pointBehindOfClosest = polyline(along: coords, within: -2, of: closest.coordinate).first!
        
        let infrontDirection = closest.coordinate.direction(to: pointInfrontOfClosest)
        let behindDirection = closest.coordinate.direction(to: pointBehindOfClosest)
        
        print(behindDirection,location.course,infrontDirection)
        
        let normalizedCourse = wrap((infrontDirection + behindDirection) / 2, min: 0, max: 360)
        let course = differenceBetweenAngles(wrap(location.course, min: 0, max: 360), normalizedCourse) <= RouteControllerMaximumAllowedDegreeOffsetForTurnCompletion ? normalizedCourse : location.course        
```

However, I'm seeing not great results; the calculated course is usually the `location.course`

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,46 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
         }
         
-        return CLLocation(coordinate: newCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: location.course, speed: location.speed, timestamp: location.timestamp)
+        let defaultReturn = CLLocation(coordinate: newCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: location.course, speed: location.speed, timestamp: location.timestamp)
+        
+        guard location.course == -1 else {
+            return defaultReturn
+        }
+        
+        guard let coords = routeController.routeProgress.route.coordinates else {
```
1ec5: `coordinates` is only `nil` if `RouteOptions.routeShapeResolution` was set to `.none`, in which case we’ve got much bigger problems. Force-unwrap (`!`) `coordinates` instead of silently failing here. We should also add an assertion as soon as possible that `coordinates` is non-nil, probably in `RouteController(route:)`.

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,46 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            return defaultReturn
+        }
+        
+        guard let closest = closestCoordinate(on: coords, to: location.coordinate) else {
```
1ec5: `closestCoordinate(on:to:)` only returns `nil` if `coords` is empty. Assert that `coords` isn’t empty beforehand, then force-unwrap this result too.

1ec5: For a long route, a linear search through the entire route geometry is bound to be slow and pointless, and doing so once a second is could hasten the battery’s demise. On other hand, if we only look at the current step’s geometry, we may miss a closer coordinate along the previous or next step.

So figure out the closest coordinate on the current step, the previous step, and the next step, then take the one nearest to the user’s location. This way we only consider relevant steps, and we also avoid snapping to the wrong portion of the route when the route doubles back on itself, as seen in mapbox/navigation.js#12.

(×3)

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,46 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            return defaultReturn
+        }
+        let slicedLine = polyline(along: coords, from: closest.coordinate, to: coords.last)
+        
+        // Get closest point infront of user
+        guard let infrontPointOne = coordinate(at: 2, fromStartOf: slicedLine) else {
```
1ec5: Likewise, this method only returns `nil` if `polyline` is empty. We could make that clear in the function’s documentation.

(×2)

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,46 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            return defaultReturn
+        }
+        guard let infrontCloest = closestCoordinate(on: coords, to: infrontPointOne) else {
```
1ec5: s/Cloest/Closest/

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,46 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            return defaultReturn
+        }
+        
+        // Get closest point begind infront of user
```
1ec5: s/begind/behind/
s/infront/in front/

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,46 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+        guard let infrontPointTwo = coordinate(at: 4, fromStartOf: slicedLine) else {
+            return defaultReturn
+        }
+        guard let behindCloset = closestCoordinate(on: coords, to: infrontPointTwo) else {
```
1ec5: s/behindCloset/behindClosest/

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,46 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            return defaultReturn
+        }
+        
+        // Get direction of these points
+        let infrontDirection = closest.coordinate.direction(to: infrontCloest.coordinate)
+        let behindDirection = closest.coordinate.direction(to: behindCloset.coordinate)
+        
+        let normalizedCourse = wrap((infrontDirection + behindDirection) / 2, min: 0, max: 360)
```
1ec5: The whole point of `differenceBetweenAngles(_:_:)` is to avoid averaging angles this way. What happens if `infrontDirection` is 1° and `behindDirection` is 359°?

bsudekum:
@1ec5 addressed the issues above

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -242,41 +242,33 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
         let slicedLine = polyline(along: coords, from: closest.coordinate, to: coords.last)
         
         // Get closest point infront of user
-        guard let infrontPointOne = coordinate(at: 2, fromStartOf: slicedLine) else {
-            return defaultReturn
-        }
-        guard let infrontCloest = closestCoordinate(on: coords, to: infrontPointOne) else {
-            return defaultReturn
-        }
+        let infrontPointOne = coordinate(at: 2, fromStartOf: slicedLine)!
```
1ec5: This code slices the line in order to get the coordinate two meters ahead of the snapped coordinate along the line. There is no difference between this code and `closest.coordinate.direction(to: coords[closest.coordinate.index + 1])`, except that the code here has to iterate over a potentially long step geometry.

Furthermore, you’d be hard-pressed to find a public road with two nodes spaced only two meters apart. A windy residential road may have two nodes spaced around ten meters apart, while a freeway would have two nodes spaced at least 20 meters apart. Anything closer than that is redundant and useless to us. (It’s called “overnoding” among OpenStreetMap mappers.)

What really matters is not an absolute distance but how far the user will travel in a fixed amount of time when going the speed limit. Roads with higher speed limits are built with gentler, broader curves. mapbox/MapboxDirections.swift#118 will give us something similar to speed limits. In the meantime, let’s decide on a time threshold like one second. Then divide the step’s distance by its expected travel time and multiply by the time threshold. Finally, pass that distance into `coordinate(at:fromStartOf:)`.

If that sounds familiar, it’s because we’ve done exactly this before, in #190.

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -242,41 +242,33 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
-        let behindDirection = closest.coordinate.direction(to: behindCloset.coordinate)
+        let infrontDirection = closest.coordinate.direction(to: infrontClosest.coordinate)
+        let behindDirection = closest.coordinate.direction(to: behindClosest.coordinate)
+        
+        let normalizedInfrontDirection = wrap(infrontDirection, min: 0, max: 360)
+        let normalizedBehindDirection = wrap(behindDirection, min: 0, max: 360)
         
-        let normalizedCourse = wrap((infrontDirection + behindDirection) / 2, min: 0, max: 360)
-        let course = differenceBetweenAngles(location.course, normalizedCourse) <= RouteControllerMaximumAllowedDegreeOffsetForTurnCompletion ? normalizedCourse : location.course
+        let averagedCourse = (normalizedInfrontDirection + normalizedBehindDirection) / 2
```
1ec5: Again, if `normalizedInfrontDirection` is 1° (basically north) and `normalizedBehindDirection` is 359° (basically north), what is the average? Does it still point north?

This is why `differenceBetweenAngles(_:_:)` exists: to correctly determine the difference between two angles without averaging the angles.

bsudekum:
Updated with feedback

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,45 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
         
-        return CLLocation(coordinate: newCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: location.course, speed: location.speed, timestamp: location.timestamp)
+        let defaultReturn = CLLocation(coordinate: newCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: location.course, speed: location.speed, timestamp: location.timestamp)
+        
+        guard location.course != -1 else {
+            return defaultReturn
+        }
+        
+        let coords = routeController.routeProgress.currentLegProgress.nearbyCoordinates
+        assert(!coords.isEmpty)
```
1ec5: https://github.com/mapbox/MapboxNavigation.swift/pull/88#discussion_r108023025

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -240,7 +240,45 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+        let behindDirection = closest.coordinate.direction(to: behindClosest.coordinate)
+        
+        let normalizedInfrontDirection = wrap(infrontDirection, min: 0, max: 360)
+        let normalizedBehindDirection = wrap(behindDirection, min: 0, max: 360)
+        
+        let diffBetweenAngleInFront = differenceBetweenAngles(location.course, normalizedInfrontDirection)
+        let diffBetweenAngleBehind = differenceBetweenAngles(location.course, normalizedBehindDirection)
+        
+        let averageDifference = (diffBetweenAngleInFront + diffBetweenAngleBehind) / 2
+        let averagedCourse = (normalizedBehindDirection + normalizedInfrontDirection) / 2
```
1ec5: If `location.course` is 0°, `normalizedInfrontDirection` is 1°, and `normalizedBehindDirection` is 359°, then `averagedDifference` is 0°, but `averagedCourse` is 179° – which is almost opposite of what you want.

bsudekum:
foo

This is working well

frederoni:
The location manager in MGLMapView and Navigation.swift works independently even though they share the same location services. We tweak and snap the location in NavUI but Nav.swift can trigger a reroute based on a signal jump which isn't reflected in NavUI. We need to figure out a way to use a single location manager in both libs or at least process the location updates through the same pipeline.

PR #91, opened 2017-03-24
Insert route below text on reroute

bsudekum:
* Cleans up the arguments for `annotate`
* Removes the arrow sooner when rerouting
* ^ This actually solves inserting the route line below labels. Before, we were inserting below the layer `arrow` instead.
* Only redraw the arrow when it actually changes.

closes: https://github.com/mapbox/MapboxNavigation.swift/issues/90 https://github.com/mapbox/MapboxNavigation.swift/issues/89

/cc @1ec5

### DIFF
--- a/MapboxNavigationUI/MGLMapView.swift
+++ b/MapboxNavigationUI/MGLMapView.swift
@@ -89,6 +89,10 @@ extension MGLMapView {
         let maneuverCoordinate = routeProgress.currentLegProgress.upComingStep?.maneuverLocation
         let polylineCoordinates = routeProgress.route.coordinates
         
+        guard let style = style else {
+            return
+        }
+        
         let shaftLength = max(min(50 * metersPerPoint(atLatitude: maneuverCoordinate!.latitude), 50), 10)
         let shaftCoordinates = polyline(along: polylineCoordinates!, within: -shaftLength / 2, of: maneuverCoordinate!)
             + polyline(along: polylineCoordinates!, within: shaftLength, of: maneuverCoordinate!)
@@ -145,11 +149,21 @@ extension MGLMapView {
             arrowStroke.lineWidth = MGLStyleValue(rawValue: 8)
             arrowStroke.lineColor = MGLStyleValue(rawValue: NavigationUI.shared.tintColor)
             
-            style?.addSource(arrowSourceStroke)
-            style?.addSource(arrowSource)
+            if let source = style.source(withIdentifier: "arrowSource") {
+                let s = source as! MGLShapeSource
+                s.shape = MGLShapeCollection(shapes: maneuverArrowPolylines)
+            } else {
+                style.addSource(arrowSource)
+                style.addLayer(arrow)
+            }
             
-            style?.addLayer(arrow)
-            style?.insertLayer(arrowStroke, below: arrow)
+            if let source = style.source(withIdentifier: "arrowSourceStroke") {
+                let s = source as! MGLShapeSource
+                s.shape = MGLShapeCollection(shapes: maneuverArrowStrokePolylines)
+            } else {
+                style.addSource(arrowSourceStroke)
+                style.insertLayer(arrowStroke, below: arrow)
+            }
         }
     }
     
--- a/MapboxNavigationUI/MGLMapView.swift
+++ b/MapboxNavigationUI/MGLMapView.swift
@@ -166,26 +166,6 @@ extension MGLMapView {
             }
         }
     }
-    
-    func removeArrow() {
-        guard let style = style else { return }
-        
-        if let arrow = style.layer(withIdentifier: "arrow") {
-            style.removeLayer(arrow)
-        }
-        
-        if let arrow = style.layer(withIdentifier: "arrowStroke") {
-            style.removeLayer(arrow)
-        }
-        
-        if let arrowStrokeSourceCheck = style.source(withIdentifier: "arrowSourceStroke") {
-            style.removeSource(arrowStrokeSourceCheck)
-        }
-        
-        if let arrowSourceCheck = style.source(withIdentifier: "arrowSource") {
-            style.removeSource(arrowSourceCheck)
-        }
-    }
 }
 
 protocol NavigationMapViewDelegate {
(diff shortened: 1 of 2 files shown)


MapboxNavigationUI/MGLMapView.swift
```diff
@@ -9,8 +9,8 @@ let routeLayerCasingIdentifier = "routeLayerCasing"
 
 extension MGLMapView {
     
-    public func annotate(_ routes: [Route], clearMap: Bool) {
-        guard let route = routes.first, var coordinates = route.coordinates else {
+    public func annotate(_ route: Route) {
```
1ec5: I think this method originally took an array in anticipation of alternative route support, but the next line brushed that idea away anyways. 😄

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -33,6 +33,7 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
     let webImageManager = SDWebImageManager.shared()
     var shieldAPIDataTask: URLSessionDataTask?
     var shieldImageDownloadToken: SDWebImageDownloadToken?
+    var arrowCurrentStep: RouteStep?
```
1ec5: (sigh) If we were using annotations instead of runtime styling, we could just make RouteStep conform to MGLAnnotation and ask the map which step is being annotated…

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -210,12 +212,16 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
...
-        guard routeProgress.currentLegProgress.upComingStep != nil else {
+        guard let step = routeProgress.currentLegProgress.upComingStep else {
             return
         }
         
-        mapView.removeArrow()
-        mapView.addArrow(routeProgress)
+        if step != arrowCurrentStep {
+            mapView.removeArrow()
+            mapView.addArrow(routeProgress)
```
1ec5: Instead of tearing down all the style layers and sources then rebuilding them from scratch, we should simply update the source’s `shape` property to reflect the current step. The check above becomes unnecessary, because mbgl will automatically redraw the style layers without any flashing.

MapboxNavigationUI/RouteMapViewController.swift
bsudekum: We're using runtime styling

1ec5: [Someday](https://github.com/mapbox/mapbox-gl-native/issues/1734#issuecomment-242802281).

bsudekum:
Updated with feedback

MapboxNavigationUI/MGLMapView.swift
```diff
@@ -145,31 +153,21 @@ extension MGLMapView {
             arrowStroke.lineWidth = MGLStyleValue(rawValue: 8)
             arrowStroke.lineColor = MGLStyleValue(rawValue: NavigationUI.shared.tintColor)
             
-            style?.addSource(arrowSourceStroke)
-            style?.addSource(arrowSource)
+            if let source = style.source(withIdentifier: arrowSourceIdentifier) {
+                let s = source as! MGLShapeSource
```
1ec5: Combine these two lines:

```swift
if let source = style.source(withIdentifier: arrowSourceIdentifier) as? MGLShapeSource
```

MapboxNavigationUI/MGLMapView.swift
```diff
@@ -145,31 +153,21 @@ extension MGLMapView {
...
+                s.shape = MGLShapeCollection(shapes: maneuverArrowPolylines)
+            } else {
+                style.addSource(arrowSource)
+                style.addLayer(arrow)
```
1ec5: Move all the code that sets up `arrowSource` and `arrow` into this `else` block. That way it’s clear that, on the second time around, the only thing that happens is the existing source’s shape getting modified.

MapboxNavigationUI/MGLMapView.swift
```diff
@@ -145,31 +153,21 @@ extension MGLMapView {
...
-        }
-        
-        if let arrowSourceCheck = style.source(withIdentifier: "arrowSource") {
-            style.removeSource(arrowSourceCheck)
+            if let source = style.source(withIdentifier: arrowSourceStrokeIdentifier) {
+                let s = source as! MGLShapeSource
+                s.shape = MGLShapeCollection(shapes: maneuverArrowStrokePolylines)
+            } else {
+                style.addSource(arrowSourceStroke)
+                style.insertLayer(arrowStroke, below: arrow)
```
1ec5: Likewise, move all the code that sets up `arrowSourceStroke` and `arrowStroke` into this `else` block.

MapboxNavigationUI/RouteMapViewController.swift
```diff
@@ -122,15 +123,14 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
     
     func notifyDidReroute(route: Route) {
         routePageViewController.notifyDidReRoute()
-        mapView.annotate([route], clearMap: true)
+        mapView.addArrow(routeController.routeProgress)
+        mapView.annotate(route)
```
1ec5: Let’s fix `annotate(_:)` the same way you’ve fixed `addArrow(_:)`: instead of blowing away the source and layers and adding them back in, just change the source’s shape. That’ll fix any flashing that occurs during rerouting and also reduce the amount of work MGLMapView has to do every time the route changes.

MapboxNavigationUI/MGLMapView.swift
```diff
@@ -124,52 +136,42 @@ extension MGLMapView {
...
             let join = NSValue(mglLineJoin: .round)
             
-            arrowStroke.lineCap = MGLStyleValue(rawValue: cap)
-            arrowStroke.lineJoin = MGLStyleValue(rawValue: join)
-            arrow.lineCap = MGLStyleValue(rawValue: cap)
-            arrow.lineJoin = MGLStyleValue(rawValue: join)
+            let arrowSourceStroke = MGLShapeSource(identifier: arrowSourceStrokeIdentifier, shape: arrowStrokeShape, options: nil)
+            let arrowStroke = MGLLineStyleLayer(identifier: arrowSourceIdentifier, source: arrowSourceStroke)
+            let arrowSource = MGLShapeSource(identifier: arrowSourceIdentifier, shape: arrowShape, options: nil)
+            let arrow = MGLLineStyleLayer(identifier: arrowLayerIdentifier, source: arrowSource)
```
1ec5: arrow and arrowStroke should ideally be declared down in the else block instead of up here, since they aren't used in the if let block.

PR #93, opened 2017-03-25
Animate progress bar

bsudekum:
Closes: https://github.com/mapbox/MapboxNavigation.swift/issues/92

/cc @1ec5


MapboxNavigationUI/RouteTableViewHeaderView.swift
```diff
@@ -40,7 +40,9 @@ class RouteTableViewHeaderView: UIView {
             if (progressBarWidthConstraint != nil) {
                 progressBar.backgroundColor = NavigationUI.shared.tintColor
                 progressBarWidthConstraint.constant = bounds.width * progress
-                setNeedsUpdateConstraints()
+                UIView.animate(withDuration: 0.5) {
```
1ec5: Should probably add `[weak self]` to this closure to avoid a retain cycle.

PR #96, opened 2017-03-28
Adopt Swift 3.1

frederoni:
Silenced warnings by updating deprecated methods and constants.
Also updated to Mapbox iOS SDK 3.5.0

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
-binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "3.4.2"
+binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "3.5.0"
 github "52inc/Pulley" "1.3.1"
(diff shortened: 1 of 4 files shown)


Cartfile.resolved
```diff
@@ -1,4 +1,4 @@
-binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "3.4.2"
+binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "3.5.0"
```
bsudekum: Can you also update https://github.com/mapbox/MapboxNavigation.swift/blob/master/Cartfile#L1

frederoni: Amended to the previous commit.

friedbunny:
[`.swift_version`](https://github.com/mapbox/mapbox-navigation-ios/blob/b765b4225b727a85759914bc683a57b7f24fa3cb/.swift-version) is still at `3.0` — is that accurate in light of this PR?

PR #101, opened 2017-03-28
Don't manipulate user puck course past certain point

bsudekum:
Closes: https://github.com/mapbox/MapboxNavigation.swift/issues/99

If the difference between the calculated course and the actual course is greater than a certain threshold, we should probably not manipulate the puck.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/Constants.swift
+++ b/MapboxNavigation/Constants.swift
@@ -111,3 +111,9 @@ public func RouteControllerMinimumDistanceForHighAlert(identifier: MBDirectionsP
  When calculating whether or not the user is on the route, we look where the user will be given their speed and this variable.
 */
 public var RouteControllerDeadReckoningTimeInterval:TimeInterval = 1.0
+
+
+/*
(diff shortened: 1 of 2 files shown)


MapboxNavigation/Constants.swift
```diff
@@ -111,3 +111,9 @@ public func RouteControllerMinimumDistanceForHighAlert(identifier: MBDirectionsP
  When calculating whether or not the user is on the route, we look where the user will be given their speed and this variable.
 */
 public var RouteControllerDeadReckoningTimeInterval:TimeInterval = 1.0
+
+
+/*
+ Maximum angle the user puck will be rotated when snapping the user's course to the route line.
+*/
+public var RouteControllerMaxManipulatedCourseAngle:CLLocationDirection = 25
```
1ec5: CLLocationDirection is used for directions, not direction _deltas_. A CGFloat or Double would’ve sufficed here, along with replacing “angle” with “degreeOffset” and replacing “max” with “maximum”.

PR #102, opened 2017-03-28
Fix polly continue way name

bsudekum:
Closes: https://github.com/mapbox/MapboxNavigation.swift/issues/98

Missing some SSML markup here.

/cc @1ec5

### DIFF
--- a/MapboxNavigationUI/RouteVoiceController.swift
+++ b/MapboxNavigationUI/RouteVoiceController.swift
@@ -227,10 +227,10 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
             if userDistance < minimumDistanceForHighAlert {
                 text = String.localizedStringWithFormat(NSLocalizedString("LINKED_WITH_DISTANCE_UTTERANCE_FORMAT", value: "%@, then in %@, %@", comment: "Format for speech string; 1 = current instruction; 2 = formatted distance to the following linked instruction; 3 = that linked instruction"), currentInstruction!, escapeIfNecessary(maneuverVoiceDistanceFormatter.string(from: userDistance)), upComingInstruction)
             } else {
-                text = String.localizedStringWithFormat(NSLocalizedString("CONTINUE", value: "Continue on %@ for %@", comment: "Format for speech string; 1 = way name; 2 = distance"), escapeIfNecessary(localizeRoadDescription(step)), escapeIfNecessary(maneuverVoiceDistanceFormatter.string(from: userDistance)))
+                text = String.localizedStringWithFormat(NSLocalizedString("CONTINUE", value: "Continue on %@ for %@", comment: "Format for speech string; 1 = way name; 2 = distance"), localizeRoadDescription(step, markUpWithSSML: markUpWithSSML), escapeIfNecessary(maneuverVoiceDistanceFormatter.string(from: userDistance)))
             }
         } else if routeProgress.currentLegProgress.currentStep.distance > 2_000 {
-            text = String.localizedStringWithFormat(NSLocalizedString("CONTINUE", value: "Continue on %@ for %@", comment: "Format for speech string; 1 = way name; 2 = distance"), escapeIfNecessary(localizeRoadDescription(step)), escapeIfNecessary(maneuverVoiceDistanceFormatter.string(from: userDistance)))
+            text = String.localizedStringWithFormat(NSLocalizedString("CONTINUE", value: "Continue on %@ for %@", comment: "Format for speech string; 1 = way name; 2 = distance"), escapeIfNecessary(localizeRoadDescription(step, markUpWithSSML: markUpWithSSML)), escapeIfNecessary(maneuverVoiceDistanceFormatter.string(from: userDistance)))
         } else if alertLevel == .high && stepDistance < minimumDistanceForHighAlert {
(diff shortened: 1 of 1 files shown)


MapboxNavigationUI/RouteVoiceController.swift
```diff
@@ -242,16 +242,23 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
+    func localizeRoadDescription(_ step: RouteStep, markUpWithSSML: Bool) -> String {
         var road = ""
+        let escapeIfNecessary = {(distance: String) -> String in
+            return markUpWithSSML ? distance.addingXMLEscapes : distance
+        }
         if let name = step.names?.first {
             if let code = step.codes?.first {
-                road = String.localizedStringWithFormat(NSLocalizedString("NAME_AND_REF", value: "%@ (%@)", comment: "Format for speech string; 1 = way name; 2 = way route number"), name, code)
+                if markUpWithSSML {
+                    road = String.localizedStringWithFormat(NSLocalizedString("NAME_AND_REF", value: "<say-as interpret-as=\"address\">%@</say-as> (<say-as interpret-as=\"address\">%@</say-as>)", comment: "Format for speech string; 1 = way name; 2 = way route number"), name.addingXMLEscapes, code.addingXMLEscapes)
```
1ec5: Any string provided in the .strings files for a particular localization will override this value.

1. You need to provide distinct string IDs for distinct strings.
1. Maybe this isn’t the right place for this markup. Are you sure you don’t want to surround `name.addingXMLEscapes` in `<say-as>` _before_ inserting it into the format string?

bsudekum: If it's around `name.addingXMLEscapes ` you get something that sounds like: 

```
Continue on Junipero Freeway opening parenthesis eye two eighty closing parenthesis.
```

bsudekum: This line cannot have the parens inside the markup https://github.com/mapbox/MapboxNavigation.swift/blob/7aeaff76c175183e3b5b6b75a22d59a811c7b31e/MapboxNavigationUI/RouteVoiceController.swift#L253

1ec5: This doesn’t work?

```swift
road = String.localizedStringWithFormat(NSLocalizedString("NAME_AND_REF", value: "%@ (%@)", comment: "Format for speech string; 1 = way name; 2 = way route number"), "<say-as interpret-as=\"address\">\(name.addingXMLEscapes)</say-as>", "<say-as interpret-as=\"address\">\(code.addingXMLEscapes)</say-as>")
```

(Split into multiple lines of course.)

It should have the same effect as the code you’ve written, except that now you’d be able to use the same localized format string for both cases and only vary the strings you end up inserting into the format string.

MapboxNavigationUI/RouteVoiceController.swift
```diff
@@ -242,16 +242,23 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
+        let escapeIfNecessary = {(distance: String) -> String in
+            return markUpWithSSML ? distance.addingXMLEscapes : distance
+        }
         if let name = step.names?.first {
             if let code = step.codes?.first {
-                road = String.localizedStringWithFormat(NSLocalizedString("NAME_AND_REF", value: "%@ (%@)", comment: "Format for speech string; 1 = way name; 2 = way route number"), name, code)
+                if markUpWithSSML {
+                    road = String.localizedStringWithFormat(NSLocalizedString("NAME_AND_REF", value: "%@ (%@)", comment: "Format for speech string; 1 = way name; 2 = way route number"), "<say-as interpret-as=\"address\">\(name.addingXMLEscapes)</say-as>", "<say-as interpret-as=\"address\">\(code.addingXMLEscapes)</say-as>")
+                } else {
+                    road = String.localizedStringWithFormat(NSLocalizedString("NAME_AND_REF", value: "%@ (%@)", comment: "Format for speech string; 1 = way name; 2 = way route number"), name, code)
```
1ec5: Factor out two local variables, `markedUpName` and `markedUpCode`, that contain the two strings you intend to insert into the `NAME_AND_REF` format string, which will have different values depending on `markUpWithSSML`. Then call `String.localizedStringWithFormat()` just once, passing in `markedUpName` and `markedUpCode`.

MapboxNavigationUI/RouteVoiceController.swift
```diff
@@ -242,16 +242,21 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
-    func localizeRoadDescription(_ step: RouteStep) -> String {
+    func localizeRoadDescription(_ step: RouteStep, markUpWithSSML: Bool) -> String {
         var road = ""
+        let escapeIfNecessary = {(distance: String) -> String in
+            return markUpWithSSML ? distance.addingXMLEscapes : distance
+        }
         if let name = step.names?.first {
             if let code = step.codes?.first {
-                road = String.localizedStringWithFormat(NSLocalizedString("NAME_AND_REF", value: "%@ (%@)", comment: "Format for speech string; 1 = way name; 2 = way route number"), name, code)
+                let markedUpName = markUpWithSSML ? "<say-as interpret-as=\"address\">\(name.addingXMLEscapes)</say-as>" : name
```
1ec5: You’ve probably noticed some parallel between this line and `escapeIfNecessary` above. 😉

Issue #103, opened 2017-03-29
Carthage hangs on Checking out osrm-text-instructions.swift


bennyty:
I have the same problem as [this user](https://github.com/mapbox/MapboxNavigation.swift/issues/77#issuecomment-286566146) building the project using Carthage. They used Cocoapods as a workaround to solve their issue. Unfortunately, Cocoapods is not available to our team.

UPDATE: Tried a workaround that did not work previously, (last week).

If the Cartfile reads:
```
github "SwiftyBeaver/SwiftyBeaver"
github "aws/aws-sdk-ios"
github "mapbox/MapboxNavigation.swift"
```
Then the error is
```
$ carthage update --platform ios MapboxNavigation.swift
*** Fetching aws-sdk-ios
*** Fetching SwiftyBeaver
*** Fetching MapboxNavigation.swift
*** Checking out MapboxNavigation.swift at "v0.0.4"
*** xcodebuild output can be found in /var/folders/b7/pzkvpztn745_zj_bgsrtzpqw0000gq/T/carthage-xcodebuild.SOgltt.log
*** Building scheme "MapboxNavigation" in MapboxNavigation.xcodeproj
Build Failed
        Task failed with exit code 65:
        /usr/bin/xcrun xcodebuild -project /Users/benjaminespey/Documents/gForce/Carthage/Checkouts/MapboxNavigation.swift/MapboxNavigation.xcodeproj -scheme MapboxNavigation -configuration Release -derivedDataPath /Users/benjaminespey/Library/Caches/org.carthage.CarthageKit/DerivedData/MapboxNavigation.swift/v0.0.4 -sdk iphoneos ONLY_ACTIVE_ARCH=NO BITCODE_GENERATION_MODE=bitcode CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean build

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/b7/pzkvpztn745_zj_bgsrtzpqw0000gq/T/carthage-xcodebuild.SOgltt.log
```

If instead "master" is appended as the target branch
```
github "SwiftyBeaver/SwiftyBeaver"
github "aws/aws-sdk-ios"
github "mapbox/MapboxNavigation.swift" "master"
```
Then carthage builds successfully
```
$ carthage update --platform ios MapboxNavigation.swift
*** Fetching MapboxNavigation.swift
*** Fetching aws-sdk-ios
[...]

PR #105, opened 2017-03-29
Add "master" pin to instructions for Cartfile

bennyty:
Proposed change to address #103. Is this necessary? Is another release due to replace (v0.0.4) soon enough?


bsudekum:
I'd like to steer clear of pinning to master since things in master might not be prime time until we cut a release. We just released v0.1.0, give that a go.

PR #109, opened 2017-03-31
Posting a notification when new route is obtained

azharaali:
Another pull request for this, with the fixes that was breaking the code before which were strings or keys that were missing

@bsudekum took another stab at this


Examples/Objective-C/ViewController.m
```diff
@@ -89,7 +89,7 @@ - (void)progressDidChange:(NSNotification *)notification {
     // If you are not using MapboxNavigationUI,
     // this would be a good time to update UI elements.
     // You can grab the current routeProgress like:
-    // let routeProgress = notification.userInfo![RouteControllerAlertLevelDidChangeNotificationRouteProgressKey] as! RouteProgress
```
azharaali: it's not being removed, its being updated to the correct key since there is a slight name change in Constants.swift file

MapboxNavigation/Constants.swift
```diff
@@ -15,6 +15,8 @@ public let RouteControllerProgressDidChange = Notification.Name(MBRouteControlle
 public let RouteControllerAlertLevelDidChange = Notification.Name(MBRouteControllerAlertLevelDidChange)
 public let RouteControllerShouldReroute = Notification.Name(MBRouteControllerShouldReroute)
 
+public let RouteControllerDidReroute = Notification.Name(MBRouteControllerDidReroute)
```
frederoni: RouteController or MapboxNavigation.swift never use this notification. Should we add a Constants.swift to MapboxNavigationUI.swift which is responsible for firing the notification and put it there instead?

azharaali: Yeah since the RouteController calls RouteControllerShouldReroute notification and RouteViewController handles it by making another call to get directions, so this notification is used in the success block there instead of the RouteController, so placed the notification key constant there just to follow how it was being done for the other notifications

bsudekum:
This is done in https://github.com/mapbox/mapbox-navigation-ios/pull/251

PR #111, opened 2017-04-02
Route simulation

frederoni:
This PR adds support for simulating location updates.
Just set `RouteViewController.simulate` to true and the given route's coordinates will be used for simulation.

The route can be archived for offline use cases.

@bsudekum @1ec5 👀

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -136,10 +136,14 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         // You can get a token here: http://docs.aws.amazon.com/mobile/sdkforios/developerguide/cognito-auth-aws-identity-for-ios.html
         // viewController.voiceController?.identityPoolId = "<#Your AWS IdentityPoolId. Remove Argument if you do not want to use AWS Polly#>"
         
-        navigationViewController.routeController.snapsUserLocationAnnotationToRoute = true
-        navigationViewController.voiceController?.volume = 0.5
-        navigationViewController.navigationDelegate = self
-        navigationViewController.pendingCamera = mapView.camera
+        let camera = mapView.camera
+        camera.pitch = 45
+        camera.altitude = 1_000
+        
+        viewController.pendingCamera = camera
+        viewController.simulatesLocationUpdates = true
+        viewController.routeController.snapsUserLocationAnnotationToRoute = true
+        viewController.voiceController?.volume = 0.5
         
         present(navigationViewController, animated: true, completion: nil)
     }
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -53,11 +53,11 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
     }
     
     @IBAction func didTapStartNavigation(_ sender: Any) {
-        startNavigation(userRoute!)
+        startNavigation(along: userRoute!)
     }
     
     @IBAction func didTapSimulateNavigation(_ sender: Any) {
-        startNavigation(userRoute!, true)
+        startNavigation(along: userRoute!, simulatesLocationUpdates: true)
     }
     
     func resumeNotifications() {
@@ -132,7 +132,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         }
     }
     
-    func startNavigation(_ route: Route,_ simulatesLocationUpdates: Bool = false) {
+    func startNavigation(along route: Route, simulatesLocationUpdates: Bool = false) {
         // Pass through a
         // 1. the route the user will take
         // 2. A `Directions` class, used for rerouting.
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -14,13 +14,16 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
     var userRoute: Route?
     
     @IBOutlet weak var mapView: NavigationMapView!
-    @IBOutlet weak var toggleNavigationButton: UIButton!
+    @IBOutlet weak var startNavigationButton: UIButton!
+    @IBOutlet weak var simulateNavigationButton: UIButton!
     @IBOutlet weak var howToBeginLabel: UILabel!
     
     
     override func viewDidLoad() {
         super.viewDidLoad()
         
+        automaticallyAdjustsScrollViewInsets = false
+        mapView.contentInset = UIEdgeInsets(top: 0, left: 0, bottom: 44, right: 0)
         mapView.delegate = self
         mapView.navigationMapDelegate = self
         
@@ -49,8 +52,12 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         getRoute()
     }
     
-    @IBAction func didToggleNavigation(_ sender: Any) {
+    @IBAction func didTapStartNavigation(_ sender: Any) {
         startNavigation(userRoute!)
+    }
+    
+    @IBAction func didTapSimulateNavigation(_ sender: Any) {
+        startNavigation(userRoute!, true)
     }
     
     func resumeNotifications() {
@@ -114,9 +121,9 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
             }
             
             self?.userRoute = route
-            self?.toggleNavigationButton.isHidden = false
+            self?.startNavigationButton.isHidden = false
+            self?.simulateNavigationButton.isHidden = false
             self?.howToBeginLabel.isHidden = true
-            
             
             // Open method for adding and updating the route line
             self?.mapView.showRoute(route)
@@ -125,7 +132,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         }
     }
     
-    func startNavigation(_ route: Route) {
+    func startNavigation(_ route: Route,_ simulatesLocationUpdates: Bool = false) {
         // Pass through a
         // 1. the route the user will take
         // 2. A `Directions` class, used for rerouting.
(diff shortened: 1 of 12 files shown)


RouteSimulator/ViewController.swift
```diff
@@ -0,0 +1,103 @@
...
+    }
+    
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: false)
+        
+        let fileName = files[indexPath.row]
+        let filePath = URL(fileURLWithPath: FREIO.storagePath().appending("/\(fileName)"))
+        guideDog.filePath = filePath
+
+        guard let firstLocation = guideDog.timeline.locations.firstObject as? CLLocation, let lastLocation = guideDog.timeline.locations.lastObject as? CLLocation else {
```
1ec5: Nit: use `first` instead of `firstObject` and `last` instead of `lastObject`.

RouteSimulator/ViewController.swift
```diff
@@ -0,0 +1,103 @@
...
+    
+    func navigate(_ route: Route) {
+        let viewController = NavigationUI.routeViewController(for: route)
+        viewController.navigationDelegate = self
+        viewController.routeController.snapsUserLocationAnnotationToRoute = true
+        viewController.voiceController?.volume = 0.5
+        
+        present(viewController, animated: true) { [weak self] in
+            guard let strongSelf = self else { return }
+            strongSelf.guideDog.attach(viewController.routeController)
```
1ec5: “Attaching” and “detaching” is an antipattern, and this particular interface doesn’t make the ownership clear: is the controller is being attached to FRELocalGuideDog or vice versa?

Let’s think about more ergonomic ways to associate the two with each other. Perhaps GuideDog could vend a protocol that requires the class (RouteController in this case) to provide a weak, FRELocalGuideDog-typed property that you can then set to `strongSelf.guideDog` here.

1ec5:
There’s a lot going on in the Xcode project file – can you summarize the changes you made? Some of the changes appear to be related to an Xcode upgrade; perhaps you could split that out into a separate commit to make reviewing easier.

I take it the separate RouteSimulation application is a temporary sandbox for functionality we’d want to fold into Voyage proper?

frederoni:
I wanted to avoid making changes to existing examples but I agree with folding this into MapboxNavigationUI. Not having to maintain yet another target as long as we can keep the examples clear.

RouteSimulator/ViewController.swift
frederoni: Using a protocol would require more hands-on when using the SDK.
Would renaming it to `guideDog.enableLocationReporting(for: mapView)` make it more clear along with `disableLocationReporting(for:)`?

1ec5: That would make the relationship a little clearer, but it still feels backwards. If RouteController is a controller (in MVC terminology), it should do things to itself or to other objects. If FRELocalGuideDog is another controller, it could reasonably manipulate another controller such as RouteController, but only if it _owns_ RouteController.

What if we actually reverse the relationship: GuideDog could implement a category on RouteController that implements `reportsLocation` as a property. I think this would only work if FRELocalGuideDog becomes a singleton, though.

bsudekum:
@frederoni what is left to do here?

frederoni:
@bsudekum Generate a route between the first and the last point in the simulated route and perhaps https://github.com/mapbox/mapbox-navigation-ios/issues/118 
Should we bundle the route along with the simulated locations or request them using the Directions API each time? AFAIK, there's no way to guarantee that the same route will be used next time. Traffic might take you on a different route. Keeping the simulated routes short could make sure that doesn't happen.

1ec5:
> Should we bundle the route along with the simulated locations or request them using the Directions API each time? AFAIK, there's no way to guarantee that the same route will be used next time.

A simulation feature is also useful for circumstances where the Directions API is unreachable. For instance, if the developer is demonstrating their application to in front of an audience, Murphy's Law would dictate that the network connection would cut out just before the application attempts to request a route.

Bundling the route with the simulated trace would make sense. If the trace is saved using a keyed archiver, Route would be compatible because it conforms to NSSecureCoding. If the trace is saved in a JSON format of some kind, we might want to keep the Directions API response around.

frederoni:
Removed the external dependency and added `Simulation.swift` instead.

@bsudekum the issue you mentioned a while ago with duplicated notifications (`RouteControllerProgressDidChange`) is still occurring. Though they're not exactly duplicated, one of them is reporting the correct `fractionsTraveled` but the other one is reporting a negative value.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -32,19 +32,23 @@ open class RouteController: NSObject {
...
     
+    var simulate: Bool = false
     
     /*
      Intializes a new `RouteController`.
      
      - parameter Route: A `Route` object representing the users route it will follow.
      */
-    public init(route: Route) {
+    public init(route: Route, simulate: Bool = false) {
```
1ec5: Argument labels shouldn’t be verbs either. How about `simulatingLocationUpdates` as the external label and `simulatesLocationUpdates` as the internal variable name?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -32,19 +32,23 @@ open class RouteController: NSObject {
     */
     public var snapsUserLocationAnnotationToRoute = false
     
+    var simulate: Bool = false
```
1ec5: Properties shouldn’t be verbs, because that causes them to look like methods. How about `simulatesLocationUpdates`?

MapboxNavigation/RouteViewController.swift
```diff
@@ -9,6 +9,8 @@ public class NavigationPulleyViewController: PulleyViewController {}
 
 public protocol RouteViewControllerDelegate {
     func routeViewControllerDidCancelNavigation(_:RouteViewController)
+    func routeViewController(_ routeViewController: RouteViewController, willChange route: Route)
```
1ec5: Public methods should be documented.

MapboxNavigation/Simulation.swift
```diff
@@ -0,0 +1,89 @@
...
+
+class Simulation : NSObject {
+    
+    var speed: CLLocationSpeed = 10
+    
+    var distanceFilter: CLLocationDistance = 10
+    
+    var verticalAccuracy: CLLocationAccuracy = 40
+    
+    var horizontalAccuracy: CLLocationAccuracy = 40
```
1ec5: Should it be possible for the horizontal accuracy to vary among location updates, the way it does in the field?

MapboxNavigation/Simulation.swift
```diff
@@ -0,0 +1,89 @@
+import Foundation
+import CoreLocation
+
+protocol SimulationDelegate : class {
+    func simulation(_ locationManager: CLLocationManager, didUpdateLocations locations: [CLLocation])
+}
+
+class Simulation : NSObject {
```
1ec5: I think “SimulatedRoute” would make the nature of this class clearer. “Simulation” could be a system for simulating routes.

MapboxNavigation/Simulation.swift
```diff
@@ -0,0 +1,89 @@
...
+        locations.remove(at: 0)
+    }
+    
+    func stop() {
+        NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(tick), object: nil)
+    }
+}
+
+extension CLLocation {
+    func with(timestamp: Date) -> CLLocation {
```
1ec5: How about `shifted(to:)`, as in “time-shifted to a date”, to clarify the relationship between the receiver and the new location?

MapboxNavigation/Simulation.swift
frederoni: Yes, I think adding some noise both to the coordinates and the accuracy will be great for testing snapping functionality but let's make sure this works properly first.

MapboxNavigation/RouteViewController.swift
frederoni: Actually, we don't need these now when the simulated route feature resides in the same framework.

frederoni:
This PR is ready for another round.
I've tested using the simulator and device w/ and w/o simulation and using the built-in routes.

frederoni:
@bsudekum It's already in the Swift example. I see this as an internal tool even though we have to expose it publicly.

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -0,0 +1,89 @@
...
+        locations.remove(at: 0)
+    }
+    
+    func stop() {
+        NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(tick), object: nil)
+    }
+}
+
+extension CLLocation {
+    func shifted(_ to: Date) -> CLLocation {
```
1ec5: `to` should be the argument’s external label (to be used at each call site). `timestamp` should be the argument’s local variable name (to be used within the method):

```swift
func shifted(to timestamp: Date) -> CLLocation
```

So each call site will look like this:

```swift
location.shifted(to: timestamp)
```

Examples/Swift/ViewController.swift
```diff
@@ -114,18 +121,18 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
...
-            
             // Open method for adding and updating the route line
             self?.mapView.showRoute(route)
             
             didFinish?()
         }
     }
     
-    func startNavigation(_ route: Route) {
+    func startNavigation(_ route: Route,_ simulatesLocationUpdates: Bool = false) {
```
1ec5: Let’s follow [Swift argument naming guidelines](https://swift.org/documentation/api-design-guidelines/#argument-labels). Only use `_` for the first argument, and only when the name of the method outside the parentheses already describes the first argument. (In general, we want the call site to include labels like `foo: `.) In this case, “navigation” doesn’t describe the first parameter, and the second argument shouldn’t have `_` either:

```swift
func startNavigation(along route: Route, simulatingLocationUpdates simulatesLocationUpdates: Bool = false)
```

This way, the call site looks like this:

```swift
startNavigation(along: route, simulatingLocationUpdates: true)
```

Someone reading this call site knows what the method will do with `route`, without having to read documentation.

MapboxNavigation/SimulatedRoute.swift
frederoni: Wanted to avoid timestamp because it shadows the internal CLLocation.timestamp but you're right.

1ec5: Ah, that makes sense. `shiftedTimestamp` is another possibility.

PR #112, opened 2017-04-04
Specify versions in pod dependencies

1ec5:
This PR specifies minimum versions for each of the third-party library dependencies in the MapboxNavigationUI podspec. This is especially important because previous versions may be written in Swift 2.3. I left out a minimum version for MapboxNavigation.swift in this podspec, because MapboxNavigation.swift has yet to be pushed to CocoaPods trunk.

/cc @frederoni @bsudekum

### DIFF
--- a/MapboxNavigationUI.swift.podspec
+++ b/MapboxNavigationUI.swift.podspec
@@ -43,11 +43,11 @@ Pod::Spec.new do |s|
   s.module_name = "MapboxNavigationUI"
 
   s.dependency "MapboxNavigation.swift"
-  s.dependency "MapboxDirections.swift"
-  s.dependency "MapboxGeocoder.swift"
-  s.dependency "Mapbox-iOS-SDK"
-  s.dependency "OSRMTextInstructions"
-  s.dependency "Pulley"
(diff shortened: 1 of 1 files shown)


MapboxNavigationUI.swift.podspec
```diff
@@ -43,11 +43,11 @@ Pod::Spec.new do |s|
   s.module_name = "MapboxNavigationUI"
 
   s.dependency "MapboxNavigation.swift"
-  s.dependency "MapboxDirections.swift"
-  s.dependency "MapboxGeocoder.swift"
-  s.dependency "Mapbox-iOS-SDK"
-  s.dependency "OSRMTextInstructions"
-  s.dependency "Pulley"
+  s.dependency "MapboxDirections.swift", "~> 0.8"
+  s.dependency "MapboxGeocoder.swift", "~> 0.6"
```
bsudekum: Can this one be removed?

1ec5: Yes, I believe so. We aren’t importing MapboxGeocoder anywhere.

PR #116, opened 2017-04-05
Make route line more configurable

bsudekum:
This makes the annotate not only public, but also overrideable. This way, developers can specify a custom source for route line and also styling for it too.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigationUI/NavigationMapView.swift
+++ b/MapboxNavigationUI/NavigationMapView.swift
@@ -1,10 +1,14 @@
 import Foundation
+import MapboxDirections
 
-public class NavigationMapView: MGLMapView {
+open class NavigationMapView: MGLMapView {
     
     var navigationMapDelegate: NavigationMapViewDelegate?
     
-    public override func locationManager(_ manager: CLLocationManager!, didUpdateLocations locations: [Any]!) {
+    let cap = NSValue(mglLineCap: .round)
+    let join = NSValue(mglLineJoin: .round)
+    
+    open override func locationManager(_ manager: CLLocationManager!, didUpdateLocations locations: [Any]!) {
         guard let location = locations.first as? CLLocation else { return }
         
         if let modifiedLocation = navigationMapDelegate?.navigationMapView(self, shouldUpdateTo: location) {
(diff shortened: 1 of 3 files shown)


bsudekum:
Closes: https://github.com/mapbox/MapboxNavigation.swift/issues/115

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -1,10 +1,14 @@
...
-public class NavigationMapView: MGLMapView {
+open class NavigationMapView: MGLMapView {
     
     var navigationMapDelegate: NavigationMapViewDelegate?
     
-    public override func locationManager(_ manager: CLLocationManager!, didUpdateLocations locations: [Any]!) {
+    let cap = NSValue(mglLineCap: .round)
+    let join = NSValue(mglLineJoin: .round)
+    
+    open override func locationManager(_ manager: CLLocationManager!, didUpdateLocations locations: [Any]!) {
```
1ec5: This change appears to be unrelated. What does it address?

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -1,10 +1,14 @@
...
+import MapboxDirections
 
-public class NavigationMapView: MGLMapView {
+open class NavigationMapView: MGLMapView {
     
     var navigationMapDelegate: NavigationMapViewDelegate?
     
-    public override func locationManager(_ manager: CLLocationManager!, didUpdateLocations locations: [Any]!) {
+    let cap = NSValue(mglLineCap: .round)
+    let join = NSValue(mglLineJoin: .round)
```
1ec5: A boxed line join generally isn’t interesting enough to split out into a separate variable.

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -13,4 +17,77 @@ public class NavigationMapView: MGLMapView {
...
+                    layer.identifier != arrowLayerIdentifier && layer.identifier != arrowSourceIdentifier {
+                    style.insertLayer(line, above: layer)
+                    style.insertLayer(lineCasing, below: line)
+                    return
+                }
+            }
+        }
+    }
+    
+    public func shape(for route: Route) -> MGLShape? {
```
1ec5: s/for/describing/ to make the relationship clearer.

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -13,4 +17,77 @@ public class NavigationMapView: MGLMapView {
             super.locationManager(manager, didUpdateLocations: locations)
         }
     }
+    
+    public func annotate(_ route: Route) {
```
1ec5: Public APIs need public documentation.

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -13,4 +17,77 @@ public class NavigationMapView: MGLMapView {
...
+    
+    public func shape(for route: Route) -> MGLShape? {
+        guard var coordinates = route.coordinates else {
+            return nil
+        }
+        
+        return MGLPolylineFeature(coordinates: &coordinates, count: route.coordinateCount)
+    }
+    
+    public func lineStyle(_ line: MGLLineStyleLayer) -> MGLLineStyleLayer {
```
1ec5: This method needs a more descriptive name. Imagine reading aloud this code:

```swift
let line = lineStyle(lineLayer)
```

“Let line be line style line layer”

The reason it’s difficult to come up with a descriptive name right now is that it needs to be scoped better. This method should be responsible for creating the layer too:

```swift
public func routeStyleLayer(identifier: String, source: MGLSource) -> MGLStyleLayer
```

Now the call site could look similar to this:

```swift
let line = routeStyleLayer(identifier: "routeLayerIdentifier", source: source)
```

“Let line be the route style layer with identifier …”

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -13,4 +17,77 @@ public class NavigationMapView: MGLMapView {
...
+        line.lineColor = MGLStyleValue(rawValue: NavigationUI.shared.tintStrokeColor.withAlphaComponent(0.6))
+        line.lineWidth = MGLStyleValue(rawValue: 5)
+        
+        line.lineCap = MGLStyleValue(rawValue: cap)
+        line.lineJoin = MGLStyleValue(rawValue: join)
+        
+        return line
+    }
+    
+    public func lineCasingStyle(_ lineCasing: MGLLineStyleLayer) -> MGLLineStyleLayer {
```
1ec5: Similarly, this method would create the MGLLineStyleLayer: `routeCasingStyleLayer(identifier:source:)`.

MapboxNavigationUI/NavigationMapView.swift
bsudekum: @1ec5 I think this needs to happen because of https://github.com/mapbox/mapbox-navigation-ios/pull/116/files#diff-f7f2309384eb86844377711b6c8e12caR4

1ec5: 👍

bsudekum:
@1ec5 Updated, just needs documentation. Will do this once everything else is in place.

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -35,16 +32,16 @@ open class NavigationMapView: MGLMapView {
...
             source.shape = polyline
         } else {
-            let geoJSONSource = MGLShapeSource(identifier: sourceIdentifier, shape: polyline, options: nil)
-            style.addSource(geoJSONSource)
+            let lineSource = MGLShapeSource(identifier: sourceIdentifier, shape: polyline, options: nil)
+            style.addSource(lineSource)
             
-            let line = lineStyle(MGLLineStyleLayer(identifier: routeLayerIdentifier, source: geoJSONSource))
-            let lineCasing = lineCasingStyle(MGLLineStyleLayer(identifier: routeLayerCasingIdentifier, source: geoJSONSource))
+            let line = routeStyleLayer(identifier: routeLayerIdentifier, source: lineSource)
```
1ec5: Now that these method names are more descriptive, it might be worth adding a NavigationMapViewDelegate protocol with optional methods like `mapView(_:routeStyleLayerWithIdentifier:source:)`, and having this code call those methods. We would fall back to the `routeStyleLayer(identifier:source:)` method implemented here if those methods aren’t implemented.

This PR can land without the delegate protocol, but if subclassing turns out to be problematic, this is the solution.

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -54,15 +57,18 @@ open class NavigationMapView: MGLMapView {
...
         guard var coordinates = route.coordinates else {
             return nil
         }
         
         return MGLPolylineFeature(coordinates: &coordinates, count: route.coordinateCount)
     }
     
-    public func routeStyleLayer(identifier: String, source: MGLSource) -> MGLStyleLayer {
+    /**
+     Function for overriding the default route line style.
```
1ec5: These are methods, not functions. (A function stands alone, apart from any class.)

MapboxNavigationUI/NavigationMapView.swift
```diff
@@ -54,15 +57,18 @@ open class NavigationMapView: MGLMapView {
         }
     }
     
-    public func shape(describing route: Route) -> MGLShape? {
+    open func shape(describing route: Route) -> MGLShape? {
```
1ec5: This is publicly visible too, so it deserves some documentation.

MapboxNavigationUI/NavigationMapView.swift
1ec5: Heh, I completely forgot that we already have a NavigationMapViewDelegate protocol that would make a nice home for these methods.

PR #120, opened 2017-04-07
Fix memory leaks

DenHeadless:
Hey!

This PR fixes various memory leaks, that prevents several MapboxNavigationUI components from being deallocated.

Three of them are usual non-weak delegate suspects. 

Two more are NSTimer instances retaining their target until invalidated. I chose to invalidate them in RouteViewController deinit, since after main controller deinit probably there's no use in them

### DIFF
--- a/MapboxNavigationUI/RouteViewController.swift
+++ b/MapboxNavigationUI/RouteViewController.swift
@@ -110,6 +110,8 @@ public class RouteViewController: NavigationPulleyViewController {
     
     deinit {
         suspendNotifications()
+        mapViewController?.resetTrackingModeTimer?.invalidate()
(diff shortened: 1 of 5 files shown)


MapboxNavigationUI/RouteViewController.swift
```diff
@@ -262,7 +264,6 @@ public class RouteViewController: NavigationPulleyViewController {
 
 extension RouteViewController: RouteTableViewHeaderViewDelegate {
     func didTapCancel() {
-        voiceController = nil
```
bsudekum: 🎉

bsudekum:
@DenHeadless, looks good to me. @1ec5 have a second to give this a once over?

Issue #123, opened 2017-04-07
Rename "RouteViewController"


ericrwolfe:
It's unclear that RouteViewController represents turn-by-turn navigation rather than, say, a route overview screen. We should rename this to be more specific and allow for the possibility of other UIs in the future.

Possible ideas:
* NavigatorViewController 
* GuidanceViewController

1ec5:
> NavigatorViewController

This is quite close to NavigationViewController, which we considered but ultimately decided against due to potential confusion with UINavigationController.

> GuidanceViewController

If I’m not mistaken, we tend to use the word “guidance” to refer to instructions and the logic to present them to the user at the appropriate time. This view controller is responsible for not only that but also the map and route display. Does this match your understanding of the term?

For inspiration, I took a look at other iOS navigation SDKs but didn’t find much:

* [HERE iOS SDK](https://developer.here.com/mobile-sdks/documentation/ios-premium/topics/map-guidance.html) provides a NMANavigationManager (analogous to our RouteController) but no view controller.
* [Skobbler iOS SDK](https://developer.skobbler.com/getting-started/ios#sec20) vends an SKRoutingService and SKRoutingDelegate for non-UI logic and an SKNavigationDelegate for UI events. Their examples implement a view controller named RoutingViewController.

frederoni:
Now when revisiting this, I think _not_ using NavigationViewController is even more prone to confusion considering the name of the library. It's namespaced or prefixed for a reason. Moving away from Route now makes even more sense since the route itself is becoming optional in #12, only a destination will be needed and we even discussed making the destination optional.

Other possible ideas:
- TurnByTurnViewController
- DestinationViewController

However, I'm 👍  on NavigationViewController / MBNavigationViewController.

1ec5:
NavigationViewController is quite elegant, although I'm still concerned because we have a RouteController (which handles the non-UI logic), which shouldn't be renamed NavigationController. On the other hand, renaming RouteMapController would bring it in line with the NavigationMapView that it contains.

frederoni:
This is happening in https://github.com/mapbox/mapbox-navigation-ios/pull/133

PR #133, opened 2017-04-12
Idiomatic initializers

### LINKED ISSUE #124: RouteViewController should use an initializer rather than a static function on NavigationUI
Is there any reason why we prefer to instantiate a RouteViewController this way

```swift
let viewController = NavigationUI.routeViewController(for: route)
```

vs this way?

```swift
let viewController = RouteViewController(for: route)
```

/cc @frederoni

frederoni:
Fixes #124, #123 

This PR renames Route->Navigation, clarifies the initializers and rearranges the table view setup so we don't try to access the route controller before it's initialized in the parent NavigationViewController.

@ericrwolfe @bsudekum @1ec5 👀

### DIFF
--- a/Examples/Objective-C/ViewController.m
+++ b/Examples/Objective-C/ViewController.m
@@ -135,7 +135,7 @@
 
 -(void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender {
     if ([segue.identifier isEqualToString:@"StartNavigation"]) {
-        MBRouteViewController *controller = (MBRouteViewController *)[segue destinationViewController];
+        MBNavigationViewController *controller = (MBNavigationViewController *)[segue destinationViewController];
         controller.route = self.route;
         controller.directions = [MBDirections sharedDirections];
     }
@@ -147,7 +147,7 @@
     [self performSegueWithIdentifier:@"StartNavigation" sender:self];
     
     // Using code
-    MBRouteViewController *controller = [[MBRouteViewController alloc] initWithRoute:route directions:[MBDirections sharedDirections]];
+    MBNavigationViewController *controller = [[MBNavigationViewController alloc] initWithRoute:route directions:[MBDirections sharedDirections]];
     controller.pendingCamera = self.mapView.camera;
     //[self presentViewController:controller animated:YES completion:nil];
 }
--- a/Examples/Objective-C/ViewController.m
+++ b/Examples/Objective-C/ViewController.m
@@ -142,8 +142,14 @@
 }
 
 - (void)startNavigation:(MBRoute *)route {
+    // Using storyboard
     [[MBNavigationUI shared] setTintColor:[UIColor redColor]];
     [self performSegueWithIdentifier:@"StartNavigation" sender:self];
+    
+    // Using code
+    MBRouteViewController *controller = [[MBRouteViewController alloc] initWithRoute:route directions:[MBDirections sharedDirections]];
+    controller.pendingCamera = self.mapView.camera;
+    //[self presentViewController:controller animated:YES completion:nil];
 }
 
 @end
--- a/MapboxNavigation/RouteViewController.swift
+++ b/MapboxNavigation/RouteViewController.swift
@@ -105,14 +105,64 @@ public class RouteViewController: NavigationPulleyViewController {
     
     required public init(contentViewController: UIViewController, drawerViewController: UIViewController) {
         fatalError("init(contentViewController:drawerViewController:) has not been implemented. " +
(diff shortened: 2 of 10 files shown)


MapboxNavigation/RouteViewController.swift
```diff
@@ -109,12 +109,65 @@ public class RouteViewController: NavigationPulleyViewController {
                    "or a storyboard reference to Navigation if you are using storyboards.")
     }
     
+    /**
+     Initializes a `RouteViewController` that provides turn by turn navigation
+     for the given route. A optional `direction` object is needed for  potential
+     rerouting.
+
+     See [MapboxDirections.swift](https://github.com/mapbox/MapboxDirections.swift)
```
bsudekum: renamed to `mapbox-navigation-ios`

MapboxNavigation/RouteViewController.swift
```diff
@@ -109,12 +109,65 @@ public class RouteViewController: NavigationPulleyViewController {
...
+     */
+    @objc(initWithRoute:directions:)
+    required public init(for route: Route,  directions: Directions = Directions.shared) {
+        let storyboard = UIStoryboard(name: "Navigation", bundle: Bundle.navigationUI)
+        let mapViewController = storyboard.instantiateViewController(withIdentifier: "RouteMapViewController") as! RouteMapViewController
+        let tableViewController = storyboard.instantiateViewController(withIdentifier: "RouteTableViewController") as! RouteTableViewController
+        
+        super.init(contentViewController: mapViewController, drawerViewController: tableViewController)
+        
+        ({
```
bsudekum: At first glance this syntax looks a little funny. Would 

```swift
super.init(contentViewController: mapViewController, drawerViewController: tableViewController) ({
....
```
work?

frederoni: I agree that it looks funny but it has to be on a separate line or it will trigger a compile-time warning.
It's not tied to super init call but it ensures that the code in the block is executed before the instance is returned.

This line https://github.com/mapbox/mapbox-navigation-ios/blob/fred-init/Examples/Swift/ViewController.swift#L160 crashes trying to access the implicitly unwrapped optional `routeController` as it's not initialized w/o this solution^

MapboxNavigation/RouteViewController.swift
frederoni: This^ refers to MapboxDirections, not navigation. 😄

MapboxNavigation/RouteViewController.swift
frederoni: > The willSet and didSet observers of superclass properties are called when a property is set in a subclass initializer, after the superclass initializer has been called. They are not called  while a class is setting its own properties, before the superclass initializer has been called.

By using the synchronous closure, willSet and didSet observers are getting called properly.
Setting `self.route = route` here is not triggering the didSet observer and therefor not calling `setupRouteController()` w/o the closure. However, we can inline this and call `self.setupRouteController()` manually.

Examples/Objective-C/ViewController.m
```diff
@@ -142,8 +142,14 @@ -(void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender {
 }
 
 - (void)startNavigation:(MBRoute *)route {
+    // Using storyboard
     [[MBNavigationUI shared] setTintColor:[UIColor redColor]];
```
frederoni: Good idea but I think that should be part of a bigger overhaul. Expose a style property on RouteViewController that grabs the tint of the parent instead of this shared instance.

1ec5: → #138

MapboxNavigation/NavigationTableViewCell.swift
```diff
@@ -2,7 +2,7 @@ import UIKit
 import MapboxDirections
 import MapboxCoreNavigation
 
-class RouteTableViewCell: UITableViewCell {
+class NavigationTableViewCell: UITableViewCell {
```
1ec5: It may not be necessary to replace all instances of “route” with “navigation”. This table is a tabular representation of a single route. If in the future this SDK includes additional navigation UI, like view controllers for roadside search or destination details, this table will still be a tabular representation of a route.

frederoni:
> what becomes of RouteController, which is just as much about navigation as NavigationViewController?

I think of it the other way around, you should be able to use the NavigationViewController w/o a route but the RouteController depends on a Route so the name stays intact. However, I agree that I went too far with the renaming. The page view and the table view depends on a route so those classes should not be renamed.

frederoni:
Stealth updated here^

Issue #134, opened 2017-04-12
Expose route line configuration to delegates


1ec5:
Per https://github.com/mapbox/mapbox-navigation-ios/pull/116#discussion_r110241567, methods like `NavigationMapView.routeStyleLayer(identifier:source:)` need to be exposed as delegate methods. An application can subclass NavigationMapView to override these methods, but it can’t force the map view embedded in RouteViewController to be an instance of that subclass.

* [ ] Add corresponding methods like `mapView(_:routeStyleLayerWithIdentifier:source:)` to the existing private NavigationMapViewDelegate protocol
* [ ] Create a new RouteMapViewControllerDelegate protocol and have RouteViewController be the RouteMapViewController’s `delegate`
* [ ] Implement `mapView(_:routeStyleLayerWithIdentifier:source:)` etc. on RouteMapViewController (which already conforms to the NavigationMapViewDelegate protocol) by calling the corresponding RouteMapViewControllerDelegate methods
* [ ] Add `mapView(_:routeStyleLayerWithIdentifier:source:)` etc. to the existing RouteViewControllerDelegate protocol
* [ ] Implement `mapView(_:routeStyleLayerWithIdentifier:source:)` etc. on RouteViewController by calling the corresponding RouteViewControllerDelegate methods

/cc @frederoni @bsudekum

PR #135, opened 2017-04-12
Expose route line configuration to delegates

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/134

Exposes 3 delegate methods:

```swift
func navigationMapView(_ mapView: NavigationMapView, routeStyleLayerWithIdenitier identifier: String, source: MGLSource) -> MGLStyleLayer?
func navigationMapView(_ mapView: NavigationMapView, routeCasingStyleLayerWithIdenitier identifier: String, source: MGLSource) -> MGLStyleLayer?
func navigationMapView(_ mapView: NavigationMapView, shapeDescribing route: Route) -> MGLShape?
```

/cc @frederoni @1ec5

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -144,17 +144,6 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         present(navigationViewController, animated: true, completion: nil)
     }
     
-    func removeRoutesFromMap() {
-        guard let style = mapView.style else {
-            return
-        }
-        if let line = style.layer(withIdentifier: layerIdentifier) {
-            style.removeLayer(line)
-        }
-        if let source = style.source(withIdentifier: sourceIdentifier) {
-            style.removeSource(source)
-        }
-    }
     
     func roundToTens(_ x: CLLocationDistance) -> Int {
         return 10 * Int(round(x / 10.0))
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -8,11 +8,10 @@ import CoreLocation
 let sourceIdentifier = "sourceIdentifier"
 let layerIdentifier = "layerIdentifier"
 
-class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewControllerDelegate {
+class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewControllerDelegate, NavigationMapViewDelegate {
     
     var destination: MGLPointAnnotation?
     var navigation: RouteController?
-    var navigationViewController: NavigationViewController?
     var userRoute: Route?
     
     @IBOutlet weak var mapView: NavigationMapView!
@@ -24,6 +23,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         super.viewDidLoad()
         
         mapView.delegate = self
+        mapView.navigationMapDelegate = self
         
         mapView.userTrackingMode = .follow
         resumeNotifications()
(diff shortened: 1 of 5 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -409,3 +422,9 @@ extension RouteMapViewController: RoutePageViewControllerDelegate {
         return routeController.routeProgress.currentLegProgress.stepAfter(step)
     }
 }
+
+protocol RouteMapViewControllerDelegate: class {
+    func navigationMapView(_ mapView: NavigationMapView, routeStyleLayerWithIdenitier identifier: String, source: MGLSource) -> MGLStyleLayer?
```
1ec5: Typo: s/Idenitier/Identifier/

1ec5: Normally you’d expect the first argument to be the RouteMapViewController itself: `routeMapViewController(_:navigationMapView:routeStyleLayerWithIdentifier:source:)`. That would let the developer vary the behavior from route map view controller to route map view controller. But that would be really unwieldy, and it’s relatively unlikely that the developer would be dealing with more than one route view controller at a time.

MapboxNavigation/RouteViewController.swift
```diff
@@ -9,7 +9,10 @@ public class NavigationPulleyViewController: PulleyViewController {}
 
 @objc(MBRouteViewControllerDelegate)
 public protocol RouteViewControllerDelegate {
-    func routeViewControllerDidCancelNavigation(_:RouteViewController)
+    @objc optional func routeViewControllerDidCancelNavigation(_:RouteViewController)
+    func navigationMapView(_ mapView: NavigationMapView, routeStyleLayerWithIdenitier identifier: String, source: MGLSource) -> MGLStyleLayer?
+    func navigationMapView(_ mapView: NavigationMapView, routeCasingStyleLayerWithIdenitier identifier: String, source: MGLSource) -> MGLStyleLayer?
+    func navigationMapView(_ mapView: NavigationMapView, shapeDescribing route: Route) -> MGLShape?
```
1ec5: I know this protocol has always been undocumented, but let’s start by adding documentation for any new methods.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -24,28 +24,16 @@ open class NavigationMapView: MGLMapView {
...
-        if let lineCasing = style.layer(withIdentifier: routeLayerCasingIdentifier) {
-            style.removeLayer(lineCasing)
-        }
-        
-        if let source = style.source(withIdentifier: sourceIdentifier) {
-            style.removeSource(source)
-        }
-        
-        let polyline = shape(describing: route)
+        let polyline = navigationMapDelegate?.navigationMapView(self, shapeDescribing: route) ?? shape(describing: route)
```
1ec5: `NavigationMapView.shape(describing:)` etc. no longer need to be public, because the developer is expected to use the delegate methods instead.

MapboxNavigation/MGLMapView.swift
```diff
@@ -121,6 +121,10 @@ extension MGLMapView {
     }
 }
 
+@objc
```
frederoni: Prefix the protocol

Examples/Swift/ViewController.swift
```diff
@@ -159,6 +159,7 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         
         viewController.routeController.snapsUserLocationAnnotationToRoute = true
         viewController.voiceController?.volume = 0.5
+        viewController.navigationDelegate = self
```
bsudekum: @1ec5 is this coming in too late? I'm not hitting the delegate methods down below.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -54,9 +61,14 @@ open class NavigationMapView: MGLMapView {
         return MGLPolylineFeature(coordinates: &coordinates, count: route.coordinateCount)
     }
     
-    /**
-     Function for overriding the default route line style.
```
1ec5: FYI, it’s OK (and a good idea) to document internal functions.

MapboxNavigation/MGLMapView.swift
```diff
@@ -121,6 +118,11 @@ extension MGLMapView {
...
 protocol NavigationMapViewDelegate: class  {
```
1ec5: This protocol should probably go in NavigationMapView.swift alongside NavigationMapView.

Examples/Swift/ViewController.swift
1ec5: At the point where `NavigationViewController.navigationMapView(_:shapeDescribing:)` attempts to call `ViewController.navigationMapView(_:shapeDescribing:)`, is NavigationViewController’s `navigationController` property still set?

PR #140, opened 2017-04-13
Merge both frameworks into a single podspec

### LINKED ISSUE #108: MapboxNavigation and MapboxCoreNavigation should be subspecs
There should only be one podspec in this repository, not two, and it should contain two [subspecs](https://guides.cocoapods.org/syntax/podspec.html#subspec) – one for the UI stuff and one for the non-UI stuff. That’ll makes it more efficient to install the entire navigation SDK via CocoaPods. Not clear whether both subspecs should be included by default.

I’d consider this a blocker for publishing the podspecs on CocoaPods.org, because there isn’t a real way to remove a pod once it’s up there.

/ref Project-OSRM/osrm-text-instructions.swift#15
/cc @bsudekum @frederoni

1ec5:
MapboxCoreNavigation is now a subspec of MapboxNavigation.

Fixes #108.

/cc @frederoni @bsudekum


1ec5:
This isn’t working correctly yet. If you use `pod 'MapboxNavigation'`, CocoaPods winds up creating a MapboxNavigation.framework target but not a MapboxCoreNavigation.framework target. As a result, any `import MapboxCoreNavigation` statement in MapboxNavigation causes a build failure:

```
/path/to/mapbox-navigation-ios/MapboxNavigation/MGLMapView.swift:4:8: No such module 'MapboxCoreNavigation'
```

1ec5:
If you use `pod 'MapboxNavigation/MapboxCoreNavigation'`, CocoaPods creates a MapboxNavigation.framework target, which means you have to `import MapboxNavigation` rather than `MapboxCoreNavigation`.

1ec5:
We may have to create a custom modulemap for the outer `MapboxNavigation` pod and specify it in the podspec using the `module_map` directive. But I’m not sure whether the subspec will be able to override the outer pod’s `module_map` directive to include its own classes.

1ec5:
I don’t see a good way forward on this issue: using a subspec for Mapbox Core Navigation is tantamount to merging it and Mapbox Navigation into the same product, which we need to avoid.

Issue #141, opened 2017-04-13
Add spanish and russian languages to polly


bsudekum:
Once https://github.com/Project-OSRM/osrm-text-instructions.swift/releases is released, we can add these languages.

1ec5:
https://github.com/Project-OSRM/osrm-text-instructions.swift/releases/tag/v0.1.1

PR #153, opened 2017-04-17
Add russian and spanish polly languages

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/141

/cc @frederoni @1ec5


bsudekum:
Added swedish

Issue #142, opened 2017-04-14
show current street name below user puck


bsudekum:
I think instead of querying the map tiles for the current street, we could:

* snapping is enabled
* if we're snapping the user to the route, we have a good idea where the user is not only on the route but in the world
* Grab the steps way name, put below the puck

/cc @frederoni @1ec5

1ec5:
One hiccup is that, if I'm not mistaken, the Directions API coalesces similar-looking names, such as "North 1st Street" and "South 1st Street", as well as temporary name changes (Project-OSRM/osrm-backend#2744).

1ec5:
If we do rely on querying features in the map, we can rely on the fact that a route with full resolution matches the vector tiles exactly (modulo any differences in precision), regardless of whether the developer has enabled snapping to the route.

We can query the map as soon as the user passes each maneuver point or intersection (to avoid the coalescing issue above). Roundabouts pose an interesting challenge, but maybe we can live with a missing or slightly inaccurate street name label within roundabouts for now.

emiltin:
if using a custom osrm server, i guess there's no guarantee that tiles and route really match

1ec5:
That’s true. However, is there a use case where having the tiles and route line up (even with minor discrepancies) is unimportant for the user experience?

Along these lines, feature querying would break down if the developer uses a style that doesn’t show the road that the route traverses (such as Mapbox Satellite), but I’m inclined to think the developer should make sure to choose a style that does include roads (such as Mapbox Satellite Streets).

frederoni:
I would also prefer to get the way name from the step.
I noticed a few caveats when working with `queryRenderedFeatures/visibleFeatures`.
A tile may not be loaded or fully rendered when we try to grab the features either due to slow connection or a significant location update. An animated user puck adds another layer of complexity when converting coordinate space.

The route controller approach seems rock solid as it doesn't rely on visual elements being aligned and in sync.

1ec5:
> A tile may not be loaded or fully rendered when we try to grab the features either due to slow connection or a significant location update.

I agree that this could be problematic. This would be a good argument for falling back to the way name in the route step.

> An animated user puck adds another layer of complexity when converting coordinate space.

The user puck isn’t animated; the camera is, using mbgl transforms. So the conversion reflects the instantaneous state of the camera.

[...]

PR #155, opened 2017-04-17
Show current way name

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/142

simulator screen shot apr 17 2017 4 31 51 pm

I'm still not convinced this any better than using the way name returned in the route name. Perhaps in cases where we cannot find a feature, we can fallback to the way name in the route?

/c @frederoni @1ec5


1ec5:
> I'm still not convinced this any better than using the way name returned in the route name.

What do you think of the consideration in https://github.com/mapbox/mapbox-navigation-ios/issues/142#issuecomment-294176614? OSRM is suppressing name changes that are unworthy of a voice notification but worthy of an UI update. That tradeoff still makes some sense to me, but it does complicate this feature.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -258,6 +260,34 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            //let streets = MGLSource(identifier: "mapbox://mapbox-streets-v7")
+            //guard mapView.style!.sources.contains(streets) else {
+            //  return
+            //}
+            
+            let userPuck = mapView.view(for: userLocation)
+            if let userPuck = userPuck {
+                let features = mapView.visibleFeatures(in: userPuck.frame)
+                
+                for layer in features {
```
1ec5: `visibleFeatures(in:)` returns an array of features, not an array of layers. The iOS SDK currently lacks a way to tell what style layer each visible feature belongs to: mapbox/mapbox-gl-native#7162. As a workaround, you can call `visibleFeatures(in:styleLayerIdentifiers:)` multiple times, once per relevant layer.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -258,6 +260,34 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                    
+                    // TODO: check for acceptable classes
+                    // if let feature = layer as? MGLPolylineFeature,
```
1ec5: MGLPolylineFeature is a kind of feature; a feature is part of a shape or vector source. But note for example that a single vector source might include features of many types at the same time.

You want to find out which style layers would display roads, which are linear. A style layer that renders linear shapes is an MGLLineStyleLayer. After you filter the style’s list of layers to include only MGLLineStyleLayers, you can further filter the list to layers whose source is Mapbox Streets and whose source layer identifier is [`road`](https://www.mapbox.com/vector-tiles/mapbox-streets-v7/#road).

You could also filter the layers by their predicates, but note that predicates can be arbitrarily complex, so this might be more trouble than it’s worth. For example, if you wanted to know whether a layer’s predicate includes only motorways, you’d have to account for predicates such as `(class IN {'motorway_link', 'trunk_link'} AND class != 'motorway' AND oneway = FALSE) OR layer = -2` etc.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -258,6 +260,34 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                    
+                    // TODO: Localize
+                    if let name = layer.attribute(forKey: "name") as? String {
```
1ec5: Should this match the map’s language or the UI language? Should we [automatically localize the map](https://github.com/mapbox/mapbox-gl-native/blob/8eb23cbac8a6c5821dc5935e27689700216c3f1a/platform/macos/app/MapDocument.m#L340-L387) to match the UI language (mapbox/ios-sdk-examples#36)?

bsudekum:
Updated the logic a bit in https://github.com/mapbox/mapbox-navigation-ios/pull/155/commits/06fac74a3e928dfb338c4d3594f5e71d43dcbd11 but now I'm not getting any features

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -268,6 +270,56 @@ extension RouteMapViewController: NavigationMapViewDelegate {
             return defaultReturn
         }
         
+        if let userLocation = mapView.userLocation,
+            let style = mapView.style,
+            routeController.showCurrentWayNameLabel == true {
```
1ec5: If `showCurrentWayNameLabel` is a Boolean, is it necessary to compare it to `true`?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -268,6 +270,56 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            
+            let validSources = style.sources.filter { (source: MGLSource) -> Bool in
+                if let vectorSource = source as? MGLVectorSource {
+                    if vectorSource.configurationURL?.absoluteString.range(of: "mapbox.mapbox-streets-v7") != nil {
+                        return true
+                    } else {
+                        return false
+                    }
+                } else {
+                    return false
```
1ec5: This `filter` call is more complex than it needs to be:

```swift
let streetsSourceIdentifiers = style.sources.flatMap {
    $0 as? MGLVectorSource
}.filter {
    $0.configurationURL?.absoluteString.range(of: "mapbox.mapbox-streets-v7") != nil
}.map {
    $0.identifier
}
```

1ec5: I hope we never release a `mapbox.mapbox-streets-v7.1`. 😉

Implement this extension:

```swift
extension MGLVectorSource {
    var isMapboxStreets: Bool {
        guard let configurationURL = configurationURL else {
            return false
        }
        return configurationURL.scheme = "mapbox" && configurationURL.host.components(separatedBy: ",").contains("mapbox.mapbox-streets-v7")
    }
}
```

then filter on it:

```swift
let streetsSourceIdentifiers = style.sources.flatMap {
    $0 as? MGLVectorSource
}.filter {
    $0.isMapboxStreets
}.map {
    $0.identifier
}
```

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -268,6 +270,56 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            }
+            assert(!validSources.isEmpty, "The option `showCurrentWayNameLabel` must contain the source `mapbox.mapbox-streets-v7`")
+            
+            if let userPuck = mapView.view(for: userLocation) {
+                
+                let lineLayers = style.layers.flatMap {
+                    $0 as? MGLLineStyleLayer
+                }
+                
+                let sourceIdenitifers = validSources.map {
```
1ec5: Typo: `Idenitifers`

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -268,6 +270,56 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                    $0.identifier
+                }
+                
+                let layerIdentifiers = lineLayers.filter {
+                    if let identifier = $0.sourceIdentifier {
+                        return sourceIdenitifers.contains(identifier) && $0.sourceLayerIdentifier == "road"
+                    } else {
+                        return false
```
1ec5: This `filter` could be written more succinctly as:

```swift
lineLayers.filter { streetsSourceIdentifiers.contains($0.sourceIdentifier ?? "") && $0.sourceLayerIdentifier == "road" }
```

1ec5:
Playing with macosapp a bit, I’m discovering that if you perform a query at the location marked by a `•`:

```
===•=== Main St ========
```

you get a feature with only the attributes that mbgl considers for rendering line layers (`class`, `oneway`, etc.). Whereas if you query here:

```
======= Main•St ========
```

you get the same, plus a feature from the symbol layer that includes the `name`. In both cases, the returned feature comes from the same feature in the vector tile, but unless I’m missing something, mbgl has optimized out any vector tile properties that are irrelevant to the style layer that returned the feature.

This wouldn’t be a problem with source querying, but source querying doesn’t allow you to specify a bounding box or point at which to query – it returns all loaded features.

Unless I’m missing something, unfortunately this could rule out feature querying as a source for this street name label.

1ec5:
> plus a feature from the symbol layer that includes the `name`.

Ah, that’s the [`road_label`](https://www.mapbox.com/vector-tiles/mapbox-streets-v7/#road_label) layer, handled in 98d53dc559331fa594af043796f46c4294da54cb. 👍

bsudekum:
@1ec5 ready for review again. Now, we're adding the `road_label` layer as a line instead of as a symbol layer.

bsudekum:
Updated screenshot: simulator screen shot apr 19 2017 12 15 23 pm

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -253,6 +262,51 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            
+            let streetsSources = style.sources.flatMap {
+                $0 as? MGLVectorSource
+                }.filter {
+                    $0.isMapboxStreets
+            }
+            let streetsSourceIdentifiers = streetsSources.map {
+                $0.identifier
+            }
+            assert(!streetsSourceIdentifiers.isEmpty, "The option `showCurrentWayNameLabel` must contain the source `mapbox.mapbox-streets-v7`")
```
1ec5: If the style doesn’t contain Mapbox Streets source, we could add it.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -253,6 +262,51 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                style.addLayer(streetLabelLayer)
+            }
+            
+            if let userPuck = mapView.view(for: userLocation) {
+                let features = mapView.visibleFeatures(in: userPuck.frame, styleLayerIdentifiers: Set([RoadLabelLayerIdentifier]))
+                
+                for feature in features {
+                    
+                    var key = "name"
+                    if let language = Locale.preferredLanguages.first!.components(separatedBy: "-").first,
```
1ec5: It’d be great to adapt this UI to the user’s preferred language, but are we doing that in the turn banner and step table and on the map? I think it’d be more important to be consistent throughout the UI. We could use the runtime styling API to localize the map, with code very similar to this code. However, the Directions API doesn’t yet return localized way names, so the turn banner and step table are stuck in the local language.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -40,6 +41,9 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
     var shieldImageDownloadToken: SDWebImageDownloadToken?
     var arrowCurrentStep: RouteStep?
     
+    let streetsLanguages = ["zh", "ru", "fr", "es", "en"]
+    let RoadLabelLayerIdentifier = "roadLabelLayer"
```
1ec5: This property is only used in one method, and there won’t ever be a need for another class to know about it. Make this a local variable instead.

bsudekum:
Ok, bridges, overpasses, etc,  seem to be causing trouble: 
simulator screen shot apr 19 2017 4 16 35 pm

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -253,6 +260,61 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+            }
+            
+            if let userPuck = mapView.view(for: userLocation) {
+                let features = mapView.visibleFeatures(in: userPuck.frame, styleLayerIdentifiers: Set([roadLabelLayerIdentifier]))
+                
+                for feature in features {
+                    
+                    var key = "name"
+                    if let language = Locale.preferredLanguages.first!.components(separatedBy: "-").first,
+                        streetsLanguages.contains(language) || language == "zh-Hans" {
```
1ec5: If the first item in `Locale.preferredLanguages` is `zh-Hans`, `language` is `zh`, not `zh-Hans`, and this special case won’t work.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -253,6 +260,61 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                        key += "_\(language)"
+                    }
+                    
+                    if let name = feature.attribute(forKey: key) as? String {
+                        if name != wayNameLabel.text {
+                            currentWayNameCounter += 1
+                        }
+                        print(currentWayNameCounter)
+                        if currentWayNameCounter >= 10 {
```
1ec5: It takes ten location updates (at least ten seconds) to update the label? What if the user has already turned onto another road? Wouldn’t they get confused when they glance at the label in the meantime and find that it’s still the old road name?

Instead of delaying the update, let’s check whether the feature matches the route line. If more than one consecutive coordinates in the feature’s `coordinates` are also in the route step’s `coordinates`, then use the feature for the label. Otherwise, ignore the feature.

bsudekum: Added Levenshtein comparison in https://github.com/mapbox/mapbox-navigation-ios/pull/155/commits/04cba873523eb6dd4f4f572f0dde38db895f3d7b

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -253,6 +259,66 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                        key += "_\(language)"
+                    }
+                    
+                    if let name = feature.attribute(forKey: key) as? String {
+                        
+                        let filteredStreetNames = routeWayNames.filter {
+                            return name.distanceFrom(string: $0) <= 6
+                        }
+                        
+                        if !filteredStreetNames.isEmpty {
```
1ec5: OK, I misread the code: a “filtered” street name must be w/i 6 of _any_ of the streets along the route. Presumably there would be an edit distance of 0 between Broadway Avenue and the Broadway Avenue in the route data. However, there’s also the problem of false positives: if the user drives along State Route 185 where it passes over State Route 85, the label will happily switch to “State Route 85” because of an edit distance of one.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -253,6 +259,66 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                streetLabelLayer.sourceLayerIdentifier = "road_label"
+                
+                // If the opacity is set to 0, the feature will be ignored in `mapView.visibleFeatures()`
+                streetLabelLayer.lineOpacity = MGLStyleValue(rawValue: 0.001)
+                streetLabelLayer.lineWidth = MGLStyleValue(rawValue: 10)
+                streetLabelLayer.lineColor = MGLStyleValue(rawValue: .white)
+                style.addLayer(streetLabelLayer)
+            }
+            
+            if let userPuck = mapView.view(for: userLocation),
```
1ec5: All we need for querying is the coordinate (snapped to the route line), not the view itself. Use [`MGLMapView.convert(_:toCoordinateFrom:)`](https://www.mapbox.com/ios-sdk/api/3.5.2/Classes/MGLMapView.html#/c:objc%28cs%29MGLMapView%28im%29convertPoint:toCoordinateFromView:) to get the point corresponding to the snapped location.

bsudekum:
simulator screen shot apr 27 2017 12 05 47 pm

Getting there. For some reason, a lot of ways have a `distanceBetwenAheadOfUserAndFeature` that is often 10m. Oddly enough, this is double the distance we're looking ahead. If I change this value, it is always the sum of `tenMetersAheadOfUser` and `tenMetersAheadofFeature`.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -254,55 +259,124 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                }
+                
+                for line in allLines {
+                    let featureCoordinates =  Array(UnsafeBufferPointer(start: line.coordinates, count: Int(line.pointCount)))
+                    let slicedLine = polyline(along: coordinates, from: newCoordinate)
+                    let featureSlice = polyline(along: featureCoordinates, from: newCoordinate)
+                    // todo: account for -10
+                    let tenMetersAheadofFeature = coordinate(at: 5, fromStartOf: featureSlice)!
+                    let tenMetersAheadOfUser = coordinate(at: 5, fromStartOf: slicedLine)!
+                    let distanceBetwenAheadOfUserAndFeature = tenMetersAheadofFeature - tenMetersAheadOfUser
```
bsudekum: @1ec5 I feel like in some cases we're looking at either end of the the query point:

```
This is 10m
<-----5m----*-----5m---->
```
Instead of 

```
On the money
* -----5m---->
*-----5m---->
```

1ec5: > Oddly enough, this is double the distance we're looking ahead. If I change this value, it is always the sum of tenMetersAheadOfUser and tenMetersAheadofFeature.

This is because of:

> ```// todo: account for -10```

On a two-way street, the route line may travel in the direction opposite the polyline feature.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -254,55 +259,133 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                    let slicedLine = polyline(along: coordinates, from: newCoordinate)
+                    let featureSlice = polyline(along: featureCoordinates, from: newCoordinate)
+                    // todo: account for -10
+                    
+                    let lookAheadDistance:CLLocationDistance = 10
+                    
+                    let pointAheadFeature = coordinate(at: lookAheadDistance, fromStartOf: featureSlice)!
+                    let pointAheadUser = coordinate(at: lookAheadDistance, fromStartOf: slicedLine)!
+                    let pointBehindFeature = coordinate(at: -lookAheadDistance, fromStartOf: featureSlice)!
+                    let pointBehindUser = coordinate(at: -lookAheadDistance, fromStartOf: slicedLine)!
```
bsudekum: @1ec5 how can i clean this up? Even when i guard here, it crashes. I'm not sure if coordinate(at:) supports negative values.

1ec5: Correct, `coordinate(at:fromStartOf:)` doesn’t support negative distances. We could either create a `coordinate(at:fromEndOf:)` method or pass in the `reversed()` coordinate array.

bsudekum:
Working pretty well now
untitled

bsudekum:
@frederoni when you get in, could you take a look at adding style support to this PR? Ideally, the label could be stylized as needed.

frederoni:
I added `MBHighlightedLabel` that defaults to the tint color. We can add more customizability like as the design solidifies.

bsudekum:
@1ec5 @frederoni this is ready for review.

MapboxNavigation/MGLVectorSource.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  MGLVectorSource.swift
```
frederoni: Nit: Drop default header comments

MapboxNavigation/Resources/Navigation.storyboard
```diff
@@ -68,15 +68,46 @@
...
+                                        <subviews>
+                                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Street Label" textAlignment="center" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="bcm-ca-nGv" customClass="MBStyleLabel">
+                                                <rect key="frame" x="8" y="2" width="69" height="21"/>
+                                                <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                                <fontDescription key="fontDescription" type="system" pointSize="12"/>
+                                                <nil key="textColor"/>
+                                                <nil key="highlightedColor"/>
+                                                <userDefinedRuntimeAttributes>
+                                                    <userDefinedRuntimeAttribute type="number" keyPath="inspectableTextStyle">
```
frederoni: We don't have these `inspectableTextStyle` anymore. Should be able to just drop the key.

MapboxNavigation/Resources/Navigation.storyboard
```diff
@@ -68,15 +68,46 @@
...
+                                    <view hidden="YES" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="drb-do-FoT" userLabel="Way Name View">
+                                        <rect key="frame" x="145" y="208.5" width="85" height="25"/>
+                                        <subviews>
+                                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Street Label" textAlignment="center" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="bcm-ca-nGv" customClass="MBStyleLabel">
+                                                <rect key="frame" x="8" y="2" width="69" height="21"/>
+                                                <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                                <fontDescription key="fontDescription" type="system" pointSize="12"/>
+                                                <nil key="textColor"/>
```
frederoni: Hm, is this just an empty key? Does it log any errors? Can we drop the key?

bsudekum:
Going to merge this. Tests failures are out of our control right now.

Examples/Swift/ViewController.swift
```diff
@@ -145,7 +145,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         
         navigationViewController.simulatesLocationUpdates = simulatesLocationUpdates
         navigationViewController.routeController.snapsUserLocationAnnotationToRoute = true
-        navigationViewController.voiceController?.volume = 0.5
+        navigationViewController.voiceController?.volume = 0
```
1ec5: Did you mean to check this in?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -252,47 +259,122 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                    
+                    let distanceBetweenPointsAhead = pointAheadFeature - pointAheadUser
+                    let distanceBetweenReversedPoint = reversedPoint - pointAheadUser
+                    let minDistanceBetweenPoints = min(distanceBetweenPointsAhead, distanceBetweenReversedPoint)
+                    
+                    if minDistanceBetweenPoints < smallestLabelDistance {
+                        smallestLabelDistance = minDistanceBetweenPoints
+                        
+                        var key = "name"
+                        if let languages = Locale.preferredLanguages.first,
```
1ec5: The text in the label should match the map, turn banner, or voice instructions, even if we happen to have a translated name in the vector tiles that is closer to the user’s preferred language. After all, it’s important to be able to match what’s on screen to real-world signage.

Since the Directions API always uses OpenStreetMap’s `name` tag, we should always use the `name` attribute for consistency with the turn banner and voice instructions.

PR #157, opened 2017-04-18
Set maximum progress upon arrival

### LINKED ISSUE #97: Progress bar never reaches the end

frederoni:
Fixes #97 

The `.arrive` notification is fired within a certain radius of the destination when `fractionsTraveled` is below 1.0. This PR avoids an incomplete progress bar by setting the progress to 1.0 if `AlertLevel` is `.arrive`.

@bsudekum 👀 

arrive100

### DIFF
--- a/MapboxNavigation/RouteTableViewController.swift
+++ b/MapboxNavigation/RouteTableViewController.swift
@@ -49,7 +49,7 @@ class RouteTableViewController: UIViewController {
     }
     
     func notifyDidChange(routeProgress: RouteProgress) {
-        headerView.progress = CGFloat(routeProgress.fractionTraveled)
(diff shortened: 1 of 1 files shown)


bsudekum:
I wonder if we can get `routeProgress.fractionTraveled` closer to 1 instead of faking it. The fact we never really get to 1 might actually be a bug.

frederoni:
I think it'll be spot on 1.0 if you hit the exact same coordinate as the destination.
We need some threshold around the destination one way or another but I will dig a bit further.

bsudekum:
I think this is the issue: 

1. when we're x distance units away from the end of the route, we enter `arrive`
1. We stop progress updates once the alert is of type `arrive` https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/RouteController.swift#L88-L97

PR #162, opened 2017-04-20
Custom styles

### LINKED ISSUE #127: Add support for custom fonts in NavigationUI
UIFont has an init function
```swift
init?(name fontName: String, size fontSize: CGFloat)
```
so we could accept the font name as a string and use our default font sizes internally.

As for the font string, we could either accept a single "font family" string and try to match up specific font styles under the hood (bold/heavy/italic/oblique/thin/etc), or have multiple font name properties for specific font styles in the UI (`fontName`, `boldFontName`, `italicFontName`).

/cc @frederoni

frederoni:
Fixes #127

TODO:
- [x] Remove NavigationUI.swift
- [x] Add the remaining customizable components (arrows, route line, ~~traffic?~~)

This PR improves custom styling by utilizing UIAppearance and UITraitCollection.

Styles can now be applied to a specific device (iPhone, iPad, CarPlay etc), a specific size class, capabilities like force touch, screen density and more.

Instead of specifying a layout for landscape or portrait, the best practice is to design for a specific size class:
screenshot 2017-04-20 13 09 00

By using these size classes, we can get an appropriate layout when using slide-over-multitasking on an iPad:

Different styles are applied through the appearance proxy depending on the current system traits:
style

**Fonts**
We should use the built-in text styles in cases where the size is picked mostly for aesthetic reasons. A good case for this would be the labels in the table view cells where there's space for the labels to grow. It makes less sense to use dynamic sizing on the more prominent labels at the top.
The available text styles are `title1`, `title2`, `title3`, `headline`, `subheadline`, `body`, `callout`, `footnote`, `caption1`, and `caption2`.

By using these predefined styles, a user can tweak the font size, weight and more by going to iOS->⚙️ -> Accessibility -> Larger Text.

**How do we develop with customizable styles?**
- Add all customizable variables to `Style.swift` as an extension to the corresponding type.
- Define and hook up the style in `Style.defaultStyle` and `Style.apply`
- The default style is now up-to-date and a developer can override it w/ or w/o a specific system trait.

**Customize the default style?**
```swift
// Change height of the maneuver view in compact mode in order to give the map view more space.
let style = Style(traitCollection: UITraitCollection(verticalSizeClass: .compact))
style.maneuverViewHeight = 80

// Change colors
style.maneuverViewBackgroundColor = .blue
style.headerBackgroundColor = .blue

style.apply()
```

@bsudekum @1ec5  👀


Examples/Swift/ViewController.swift
```diff
@@ -148,6 +148,12 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         navigationViewController.voiceController?.volume = 0.5
         navigationViewController.navigationDelegate = self
         
+        // Uncomment to apply custom styles
+//        styleForRegular().apply()
+//        styleForCompact().apply()
+//        styleForiPad().apply()
+//        styleForCarPlay().apply()
```
1ec5: This approach accomplishes the goal of allowing developers to theme the navigation UI and even vary it by size class. However, I wonder if we can make more direct use of the UIAppearance system.

Since RouteViewController is the entry point for the entire SDK’s UI, would it make sense to encapsulate all these style properties within that class? RouteViewController would expose properties like `maneuverViewHeight` by passing those values along to the maneuver view. The developer would call `appearance(for:)` directly on RouteViewController in order to set these properties. This would allow us to avoid making classes like SeparatorView public and also make it easier to use a single appearance proxy for the entire themeable interface.

Down the line, if we decide to make ManeuverView public and more customizable, for example, we could expose a `maneuverView` property on RouteViewController; since UIView already conforms to UIAppearanceContainer, we’d get themeability “for free”.

MapboxNavigation/Style.swift
```diff
@@ -0,0 +1,311 @@
...
+    
+    dynamic var height: CGFloat = .defaultManeuverViewHeight {
+        didSet {
+            heightConstraint.constant = height
+            setNeedsUpdateConstraints()
+        }
+    }
+}
+
+class RouteStyleLayer: MGLLineStyleLayer {
```
1ec5: This class should be public. #135 added delegate methods like `routeStyleLayer(identifier:source:)` in order to allow the developer to provide their own route layer. Since NavigationMapView’s default behavior creates an instance of RouteStyleLayer, it should be possible for the developer’s implementation to create a RouteStyleLayer as well. Otherwise, any appearance settings would be ignored as soon as the developer customizes the route display via the delegate methods.

1ec5: As part of the refactoring suggested above, this class would conform to UIAppearanceContainer. That would obviate the `appearanceColor` property, because the developer could set `RouteStyleLayer.appearance().lineColor` directly.

Examples/Swift/ViewController.swift
frederoni: I like the idea of a more direct use when we're talking specifically about UIViews and it crossed my mind a few times. However, a property like `maneuverViewHeight` will need some restrictions like MinMax(50, 200) to make sure the subviews fit inside.

An advanced user would probably already accomplish this by exposing a couple of private properties publicly and style anything using the appearance proxy but the intention here was also to keep it clean and simple. Changing the tint color for example updates `NavigationMapView`, `ProgressBar` and `Button`. Styling these w/o these helper methods would quickly become bloated.

```swift
NavigationMapView.appearance(for: traitCollection).tintColor = color
ProgressBar.appearance(for: traitCollection).backgroundColor = color
Button.appearance(for: traitCollection).tintColor = color
```
instead of
```swift
style.tintColor = color
```
excluding the initial style setup ofc.

Also, UIAppearance works on classes but not instances. We could expose ManeuverView but not as a property on NavigationViewController.

MapboxNavigation/Style.swift
frederoni: I overlooked these delegate methods. Not sure if we should keep them or make them stylable by UIAppearance. The problem is that only UIViews have proper support for UIAppearance. You can achieve some of it by using Objective-C's `UI_APPEARANCE_SELECTOR` (or just dynamic in Swift) but it's just a lucky side effect from MGLMapView that makes it update using the appearance when changing size classes.

TBH, I think we should remove this from the style and keep the delegate methods.

frederoni: The container protocol is used for `whenContainedInInstancesOf:`, it doesn't expose the class’s member variables. AFAIK, only subclasses of UIView do that because it conforms to UIAppearance. `RouteStyleLayer cannot conform to UIAppearance because it has requirements that cannot be satisfied`.

1ec5: Right, I was thinking of having these classes conform to UIAppearance as well, but I guess that could be overly complicated due to how style layers are initialized.

1ec5: How common would the use case of customizing only the tint color of the route line be, versus customizing additional options that would warrant the full delegate method?

Examples/Swift/ViewController.swift
frederoni: > What about moving these properties to NavigationViewController as they are, though? 

I guess it makes sense for sizes and dimensions but one drawback is the life cycle of the NavigationViewController. It means that you have to:
1. Initialize the NavigationViewController
2. Present it
3. Now you can modify the constraints, the outlets are hooked up (except for the ones in the page view and other dynamic views that are loaded after)

When using the appearance proxy, you can apply the style whenever, even in the app delegate.

> Is it necessary to separate them into a separate Style class?

A unique class or a unique container view is needed.

> removing NavigationUI moves us slightly away from the idea of separating presentation from functionality.

NavigationUI was kinda separated but the singleton manager approach is rarely recommended. The appearance however, is completely de-coupled. The view doesn't even use a reference to the style, except for custom objects like our style layers.

1ec5: OK, these are all good reasons for a separate Style object, then. 👍

README.md
```diff
@@ -67,27 +67,27 @@ Directions.shared.calculate(options) { (waypoints, routes, error) in
...
+style.maneuverViewHeight = 80
+style.primaryTextColor = .black
+style.headerBackgroundColor = .white
+style.cellTitleLabelFont = .preferredFont(forTextStyle: .headline)
+style.apply()
+```
 
-// Used for subtitles, distances and accessory labels
-NavigationUI.shared.secondaryTextColor = .pink
+Or even for a specific system trait in an interface’s environment.
```
1ec5: Nit: omit “even”, which makes the text feel more like an advertisement.

MapboxNavigation/Style.swift
```diff
@@ -0,0 +1,311 @@
...
+     */
+    public var routeCasingColor: UIColor?
+    
+    /**
+     `routeLayerColor` sets the color of the route line.
+     
+     Note that this color will not be used if you are implementing
+     `routeCasingStyleLayerWithIdentifier(_:identifier:source:)`.
+     */
+    public var routeLayerColor: UIColor?
```
bsudekum: Why do we have both a variable and a delegate method for defining the color?

1ec5: See https://github.com/mapbox/mapbox-navigation-ios/pull/162#discussion_r112711388.

bsudekum: I'm not sure if we should have 2 options for styling the same thing.

1ec5: I think we’re in agreement: https://github.com/mapbox/mapbox-navigation-ios/pull/162#discussion_r112728486.

PR #167, opened 2017-04-24
Remove redundant instructions strings; add format strings

1ec5:
Deleted Instructions.strings, which is already provided by OSRM Text Instructions. Imported Chinese translations of instruction-related format strings. (The base translations are already provided inline in the source code.)

/cc @bsudekum @frederoni


1ec5:
I included Simplified Chinese translations in this PR because we already had them in an internal application. But we should expand our translation capabilities using Transifex: #168.

PR #170, opened 2017-04-24
Improve instructions for running example project

ericrwolfe:
Also: 
* Adds a more obvious access token placeholder to `Info.plist`
* Adds instructions for adding your access token

/cc @bsudekum @1ec5

### DIFF
--- a/Examples/Objective-C/Info.plist
+++ b/Examples/Objective-C/Info.plist
@@ -23,7 +23,7 @@
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>MGLMapboxAccessToken</key>
-	<string>pk.feedCafeDadeDeadBeef-BadeBede.FadeCafeDadeDeed-BadeBede</string>
+	<string>PASTE MAPBOX ACCESS TOKEN HERE</string>
 	<key>NSLocationAlwaysUsageDescription</key>
 	<string>Get user location</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ Mapbox Navigation requires a few additions to your `Info.plist`. Be sure to sign
 
 1. Add a `MGLMapboxAccessToken` key and paste your [Mapbox Access Token](https://www.mapbox.com/studio/account/tokens/)
 1. Add a `NSLocationWhenInUseUsageDescription` key if you haven't already
+1. If you need voice guidance while your app is in the background, you'll also need to add the `audio` and `location` value to the `UIBackgroundModes` array. You can also do this by navigating to the `Capabilities` tab -> `Background Modes` and enabling the following:
+    - `Audio, AirPlay, and Picture in Picture`
+    - `Location updates`
 
 #### Storyboards
 See [this guide](https://github.com/mapbox/mapbox-navigation-ios/blob/master/Docs/Storyboards.md) for usage with storyboards.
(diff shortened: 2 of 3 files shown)


Examples/Objective-C/Info.plist
```diff
@@ -23,7 +23,7 @@
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>MGLMapboxAccessToken</key>
-	<string>pk.feedCafeDadeDeadBeef-BadeBede.FadeCafeDadeDeed-BadeBede</string>
+	<string>PASTE MAPBOX ACCESS TOKEN HERE</string>
```
1ec5: You didn’t like my hexadecimal poem? 😄

README.md
```diff
@@ -66,6 +75,13 @@ Directions.shared.calculate(options) { (waypoints, routes, error) in
 }
 ```
 
+#### Required Info.plist Keys
+Mapbox Navigation requires a few additions to your `Info.plist`. Be sure to sign up or log in to your Mapbox account and grab a [Mapbox Access Token](https://www.mapbox.com/studio/account/tokens/).
+
+1. Add a `MGLMapboxAccessToken` key and paste your [Mapbox Access Token](https://www.mapbox.com/studio/account/tokens/)
+1. Add a `NSLocationWhenInUseUsageDescription` key if you haven't already
```
1ec5: To continue receiving voice announcements and rerouting in background mode, `UIBackgroundModes` also needs to be set to `audio` and `location`.

README.md
```diff
@@ -95,17 +111,6 @@ NavigationUI.shared.lineColor = .yellow
 
 * `routeControllerDidCancelNavigation`: Fired when the user taps `Cancel`. You are responsible for dismissing the UI
 
-## Examples
-
-We provide examples in Swift and Objective-C. Run `carthage update --platform ios` from the root folder and open MapboxNavigation.xcodeproj to try it out.
-
-### Running the example app
-
-1. If running in the simulator, you can simulate the user location by selecting `Debug` -> `Location` -> `City Bicycle Ride`
```
1ec5: The default simulated location preset is either Apple or None, which doesn’t do a whole lot. A nod to this setting would be helpful for developers who might forget about this setting’s existence.

ericrwolfe: @frederoni added a "Simulate Navigation" button which removes the need for the simulator-based location simulation

1ec5: Ah, good point! 👍

README.md
ericrwolfe: Ah hrmm, yes, tricky. Do we encourage developers to add this to their Info.plist themselves or use the `Capabilities` tab -> `Background Modes`?

ericrwolfe:
@bsudekum not sure where such an assertion should live (core, UI, or just in the example), so we should probably ticket this out separately.

README.md
1ec5: Either is fine. 👍

1ec5:
> Want to add an assertion that checks for an access token?

The access token is normally needed for Directions API and Styles API access. However, so far we’ve entertained the possibility of connecting to a self-hosted OSRM server, which wouldn’t need an access token. Meanwhile, the map SDK allows the developer to self-host their style and tiles, in which case no access token is needed.

Even if we wanted to restrict the custom OSRM server case, MapboxDirections.swift doesn’t currently provide any way to know whether the Route object passed into NavigationViewController’s initializer was created from a Directions object with an access token.

PR #179, opened 2017-04-25
Set correct class name for NavigationViewController in storyboard.

frederoni:
Fixes an issue from #133 where we used the wrong custom class for `Route/NavigationViewController` in Navigation.storyboard.

@bsudekum 👀


frederoni:
cc @1ec5

Issue #183, opened 2017-04-26
Fallback voice is always Australian


bsudekum:
This should change to match the users region.

PR #187, opened 2017-04-28
Use Locale for deciding current language code

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/183

`Bundle.main.preferredLocalizations.first`  only would specify `en` vs with `Locale.preferredLanguages` which produces `en-us`.  We use the same lookup for Polly voices.

/cc @incanus

### DIFF
--- a/MapboxNavigation/Locale.swift
+++ b/MapboxNavigation/Locale.swift
@@ -9,8 +9,7 @@ extension Locale {
             return firstBundleLocale
         }
         
-        let countryCode = NSLocale.Key.countryCode
-        let countryString = (NSLocale.current as NSLocale).object(forKey: countryCode) as! String
+        let countryString = (NSLocale.current as NSLocale).object(forKey: .countryCode) as! String
         return "\(bundleLocale.first!)-\(countryString)"
     }
     
--- a/MapboxNavigation/Locale.swift
+++ b/MapboxNavigation/Locale.swift
@@ -1,0 +1,17 @@
+import Foundation
+
+extension Locale {
+    static var preferredLocalLanguageCountryCode: String {
+        let bundleLanguage = Bundle.main.preferredLocalizations.first!
+        let bundleLanguages = bundleLanguage.components(separatedBy: "-")
+        
+        if bundleLanguages.count == 2 {
+            return bundleLanguage
+        }
+        
+        let countryCode = NSLocale.Key.countryCode
+        let locale = NSLocale.current as NSLocale
+        let countryString = locale.object(forKey: countryCode) as! String
+        return "\(bundleLanguages.first!)-\(countryString)"
+    }
+}
--- a/MapboxNavigation/Locale.swift
+++ b/MapboxNavigation/Locale.swift
@@ -2,16 +2,17 @@ import Foundation
 
 extension Locale {
     static var preferredLocalLanguageCountryCode: String {
-        let bundleLanguage = Bundle.main.preferredLocalizations.first!
-        let bundleLanguages = bundleLanguage.components(separatedBy: "-")
+        let firstBundleLocale = Bundle.main.preferredLocalizations.first!
+        let bundleLocale = firstBundleLocale.components(separatedBy: "-")
         
-        if bundleLanguages.count == 2 {
-            return bundleLanguage
+        if bundleLocale.count > 1 {
+            return firstBundleLocale
         }
         
         let countryCode = NSLocale.Key.countryCode
-        let locale = NSLocale.current as NSLocale
-        let countryString = locale.object(forKey: countryCode) as! String
-        return "\(bundleLanguages.first!)-\(countryString)"
+        let countryString = (NSLocale.current as NSLocale).object(forKey: countryCode) as! String
+        return "\(bundleLocale.first!)-\(countryString)"
     }
+    
+    static var nationalizedCurrent = Locale(identifier: preferredLocalLanguageCountryCode)
 }
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -52,7 +52,7 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
         super.viewDidLoad()
         automaticallyAdjustsScrollViewInsets = false
         
-        distanceFormatter.numberFormatter.locale = Locale.nationalizedCurrent
+        distanceFormatter.numberFormatter.locale = .nationalizedCurrent
         
         mapView.delegate = self
         mapView.navigationMapDelegate = self
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -52,7 +52,7 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
         super.viewDidLoad()
         automaticallyAdjustsScrollViewInsets = false
         
-        distanceFormatter.numberFormatter.locale = Locale.init(identifier: Locale.preferredLocalLanguageCountryCode)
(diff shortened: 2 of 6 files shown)


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -369,7 +369,7 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
         }
         
         // Only localized languages will have a proper fallback voice
-        utterance.voice = AVSpeechSynthesisVoice(language: Bundle.main.preferredLocalizations.first)
+        utterance.voice = AVSpeechSynthesisVoice(language: Locale.preferredLanguages.first!)
```
1ec5: [`Bundle.main.preferredLocalizations.first`](https://developer.apple.com/reference/foundation/bundle/1413220-preferredlocalizations) is the localization used by the current application (taking into account the Preferred Languages in Settings), whereas Locale.preferredLanguages are the locales listed under Preferred Languages in Settings (regardless of the available localizations).

For example, if the application is localized into only English and Simplified Chinese, but the user's preferred languages are Spanish and English (in that order), this new code will cause the application to read Vietnamese instructions (because OSRM Text Instructions are available in Spanish) with an English voice, spliced into English format strings. You'd hear, "**In** 3 kilómetros, gire a la izquierda, **then** ir por 10 kilómetros" – all with an English voice. The old code would've given English instructions in English, consistent with an interface already in English. If the application wants to support Spanish, it needs to provide a Spanish localization.

1ec5: Instead of this change, I suspect [this line](https://github.com/mapbox/mapbox-navigation-ios/blob/98f9286234733bbee62f68c8fd899318772da2c8/MapboxNavigation/RouteVoiceController.swift#L273) that determines the Polly voice needs to be switched from Locale to Bundle.

bsudekum: @1ec5 how can we get even more specific localization than just `en`. If I set my region to `en-us`, `Bundle.main.preferredLocalizations.first` will only return `en`.

1ec5: `en` should be good enough for AVSpeechSynthesizer. If it isn’t specific enough for Polly, use `Locale.current.regionCode`. (`regionCode` is only available on iOS 10+, but you can use `Locale.object(forKey:)` with `NSLocale.Key.countryCode` to get the same information.

1ec5:
While we’re at it, we should also be using the Amy or Emma voice for British English.

bsudekum:
@1ec5 I updated this in https://github.com/mapbox/mapbox-navigation-ios/pull/187/commits/328c0fad66518d474f8787743feb231962865797 but I'm still getting meters when the language country code of `en-fr`. See anything that I'm missing?

1ec5:
> I'm still getting meters when the language country code of en-fr. See anything that I'm missing?

That’s the expected behavior, since France uses the metric system.

MapboxNavigation/Locale.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+extension Locale {
+    static var preferredLocalLanguageCountryCode: String {
+        let bundleLanguage = Bundle.main.preferredLocalizations.first!
+        let bundleLanguages = bundleLanguage.components(separatedBy: "-")
```
1ec5: This isn’t a list of languages, it’s a list of components detailing a single locale. See [this article](https://en.wikipedia.org/wiki/IETF_language_tag) for more information.

MapboxNavigation/Locale.swift
```diff
@@ -0,0 +1,17 @@
...
+        
+        if bundleLanguages.count == 2 {
+            return bundleLanguage
+        }
+        
+        let countryCode = NSLocale.Key.countryCode
```
1ec5: No need for a separate variable for this constant.

MapboxNavigation/Locale.swift
```diff
@@ -0,0 +1,17 @@
...
+        let locale = NSLocale.current as NSLocale
```
1ec5: No need for a separate variable for the result of this cast. Surround it with parentheses and inline it below.

MapboxNavigation/Locale.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+extension Locale {
+    static var preferredLocalLanguageCountryCode: String {
+        let bundleLanguage = Bundle.main.preferredLocalizations.first!
+        let bundleLanguages = bundleLanguage.components(separatedBy: "-")
+        
+        if bundleLanguages.count == 2 {
```
1ec5: It’s possible for the host application to have a localization whose locale code contains more than two components, for instance `zh-Hant-HK` for Traditional Chinese in Hong Kong versus `zh-Hant-TW` for the same in Taiwan.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -52,6 +52,8 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
         super.viewDidLoad()
         automaticallyAdjustsScrollViewInsets = false
         
+        distanceFormatter.numberFormatter.locale = Locale.init(identifier: Locale.preferredLocalLanguageCountryCode)
```
1ec5: Nit: drop `.init`.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -83,6 +83,7 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     override public init() {
         super.init()
         maneuverVoiceDistanceFormatter.unitStyle = .long
+        maneuverVoiceDistanceFormatter.numberFormatter.locale = Locale.init(identifier: Locale.preferredLocalLanguageCountryCode)
```
1ec5: Nit: drop `.init`.

MapboxNavigation/RouteTableViewCell.swift
```diff
@@ -20,6 +20,7 @@ class RouteTableViewCell: UITableViewCell {
             
             turnArrowView.step = step
             titleLabel.text = routeStepFormatter.string(for: step)
+            distanceFormatter.numberFormatter.locale = Locale.init(identifier: Locale.preferredLocalLanguageCountryCode)
```
1ec5: Given how often we have to write this, it might be more convenient to add another variable to Locale called `nationalizedCurrent` that returns the same Locale as this line is creating.

bsudekum:
@1ec5 updated

MapboxNavigation/Locale.swift
```diff
@@ -0,0 +1,18 @@
...
+extension Locale {
+    static var preferredLocalLanguageCountryCode: String {
+        let firstBundleLocale = Bundle.main.preferredLocalizations.first!
+        let bundleLocale = firstBundleLocale.components(separatedBy: "-")
+        
+        if bundleLocale.count > 1 {
+            return firstBundleLocale
+        }
+        
+        let countryCode = NSLocale.Key.countryCode
```
1ec5: Nit: inline this expression. It’s a constant, so there’s no advantage to breaking it out. In fact, since `NSLocale.object(forKey:)` takes an `NSLocale.Key`, which is an enumeration, you can pass in just `.countryCode`.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -52,6 +52,8 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
         super.viewDidLoad()
         automaticallyAdjustsScrollViewInsets = false
         
+        distanceFormatter.numberFormatter.locale = Locale.nationalizedCurrent
```
1ec5: You might be able to say just `.nationalizedCurrent` here, but I’m not sure.

PR #189, opened 2017-05-01
Add custom styling to way name label

bsudekum:
Otherwise, it's always black.


Examples/Swift/ViewController.swift
```diff
@@ -172,6 +172,8 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
         style.secondaryTextColor = #colorLiteral(red: 0.9626983484, green: 0.9626983484, blue: 0.9626983484, alpha: 1)
         style.buttonTextColor = #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
         
+        style.wayNameTextColor = #colorLiteral(red: 0.9418798089, green: 0.3469682932, blue: 0.5911870599, alpha: 1)
```
1ec5: Do we intend to keep this color in sync with the tint color? If so, set the property to `style.tintColor` so there are fewer color swatches to update in the future. If not, then this is fine.

bsudekum: I just copied the tint color for this example. The developer should be able to change this color as needed.

Issue #172, opened 2017-04-25
MapboxNavigation podspec fails validation


1ec5:
MapboxNavigation.podspec fails CocoaPods validation due to some warnings. (I have to pass `--allow-warnings` every time I push the podspec to trunk.) Some of these warnings are upstream in AWSCore and Pulley. The warnings in MGLMapView+MGLNavigationAdditions.h are caused by the hack in #57. Until mapbox/mapbox-gl-native#6867 is fixed, we should silence the warning using `#pragma clang` compiler directives.

```
Validating podspec
 -> MapboxNavigation (0.3.0)
    - WARN  | xcodebuild:  AWSCore/AWSCore/Fabric/Fabric.h:43:11: warning: parameter 'kits' not found in the function declaration [-Wdocumentation]
    - NOTE  | xcodebuild:  AWSCore/AWSCore/Fabric/Fabric.h:43:11: note: did you mean 'kitClasses'?
    - WARN  | xcodebuild:  Pulley/PulleyLib/PulleyViewController.swift:747:94: warning: 'FLT_EPSILON' is deprecated: Please use 'Float.ulpOfOne' or '.ulpOfOne'.
    - WARN  | xcodebuild:  Pulley/PulleyLib/PulleyViewController.swift:750:75: warning: 'FLT_EPSILON' is deprecated: Please use 'Float.ulpOfOne' or '.ulpOfOne'.
    - WARN  | xcodebuild:  AWSPolly/AWSPolly/AWSPollySynthesizeSpeechURLBuilder.h:200:9: warning: parameter 'configuration' not found in the function declaration [-Wdocumentation]
    - WARN  | xcodebuild:  AWSPolly/AWSPolly/AWSPollySynthesizeSpeechURLBuilder.h:217:9: warning: parameter 'preSignedURLRequest' not found in the function declaration [-Wdocumentation]
    - NOTE  | xcodebuild:  AWSPolly/AWSPolly/AWSPollySynthesizeSpeechURLBuilder.h:217:9: note: did you mean 'getPreSignedURLRequest'?
    - WARN  | xcodebuild:  MapboxNavigation/MapboxNavigation/MGLMapView+MGLNavigationAdditions.m:3:17: warning: method definition for 'locationManager:didUpdateLocations:' not found [-Wincomplete-implementation]
[...]

PR #190, opened 2017-05-01
Silence locationManager warning

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/172


MapboxNavigation/MGLMapView+MGLNavigationAdditions.h
```diff
@@ -1,3 +1,4 @@
+#pragma clang diagnostic ignored "-Wincomplete-implementation"
```
1ec5: This silences the diagnostic for every file that includes this header and thus every project that includes this SDK. Use `clang diagnostic push` and `clang diagnostic pop` to scope this pragma to just the lines that need it. See [this article](http://nshipster.com/clang-diagnostics/) for more information about diagnostic pragmas.

bsudekum:
@1ec5 good call, fixed

Issue #194, opened 2017-05-02
Current road label is translated into user’s preferred language


1ec5:
In a bilingual country, a road might be tagged with e.g. a Spanish name in `name` and a French name in `name_fr` in OpenStreetMap. The Directions API returns the value of `name`, so the turn banner and voice instructions will both indicate the Spanish name. Presumably the map style would also indicate the Spanish name. On the other hand, the current road label tries to be clever and dig out the French name.

The driver needs to be able to match what they see and hear coming from the SDK with what’s on the ground. We should always display the primary signposted name of a street (`name`); if we ever show a translated name, it should only be a secondary display. (I’m intentionally glossing over transliteration here to simplify the proposed solution.) Once the Directions API starts returning localized street names, then we can consider pulling localized names out of the vector tiles.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/155#discussion_r114227029
/cc @bsudekum @cammace @nickidlugash

PR #195, opened 2017-05-02
Use `name` only for way name label

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/194


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -321,16 +321,9 @@ extension RouteMapViewController: NavigationMapViewDelegate {
                     if minDistanceBetweenPoints < smallestLabelDistance {
                         smallestLabelDistance = minDistanceBetweenPoints
                         
-                        var key = "name"
-                        if let languages = Locale.preferredLanguages.first,
-                            let language = languages.components(separatedBy: "-").first,
-                            streetsLanguages.contains(language) || languages == "zh-Hans" {
```
1ec5: `streetsLanguages` is unused.

Issue #199, opened 2017-05-02
Simple left turn caused a reroute


bsudekum:
![image](https://cloud.githubusercontent.com/assets/1058624/25637153/e43c264a-2f38-11e7-93a9-ac845c85b1b1.png)
> [Link](http://geojson.io/#id=gist:anonymous/909d397a69b67bd5e6cd60b68c67fd59&map=19/37.77346/-122.43642)

This one boggles my mind, but moving a normal speed through this turn caused a reroute soon after exiting the maneuver. I'm assuming we never entered the desired exit course and then moved x units away from the route.

ref https://github.com/mapbox/mapbox-navigation-ios/issues/192 which might of helped here.

bsudekum:
What if:

* The user enters the maneuver radius
* Their course is not within x degrees of the exit bearing
* We check if the user is within x units of the following step.
 * Or we check the user is traveling in the direction of the next step
* If so, the user has completed the step

1ec5:
I agree that the current check is too strict. If I understand https://github.com/mapbox/mapbox-navigation-ios/issues/199#issuecomment-298752199 correctly, we should also consider the maneuver complete if the user is already on the next step and moving in the right direction. But that “right direction” means comparing the `course` reading with the route data’s exit bearing, so I’m not sure how that would differ from the current behavior. Perhaps we just need to loosen our course threshold for considering the maneuver complete?

frederoni:
@bsudekum 
I think this is what's happening;
<img src="https://cloud.githubusercontent.com/assets/764476/25671873/cebe510e-3031-11e7-938f-b8dbb8ae3683.png" width=320>
(Color codes explained in https://github.com/mapbox/mapbox-navigation-ios/pull/205)
Why do we use dead reckoning when we have a good accurate fixed position?
When the horizontal accuracy is good (probably this case) the blue dot won't snap to the green dot (causing a reroute), which it shouldn't, it also shouldn't be there in the first place.

PR #202, opened 2017-05-02
Add second heuristic for calculating user did complete step

bsudekum:
Looking to fix: https://github.com/mapbox/mapbox-navigation-ios/issues/199

The gist of this:
* If we do not calculate the users course is with x degrees of the exit bearing for a step, we should have a fallback to another heuristic.
* Now, we check
  * if the user is not on the current step
  * if they are also on the next step
  * And have entered the maneuver radius on this step
* If so, we can assume they have now on the next step and we increment the step counter.

/cc @mapbox/navigation

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -38,8 +38,6 @@ open class RouteController: NSObject {
         }
     }
     
-    var isInsideManeuverZone = false
-    
     /*
      Intializes a new `RouteController`.
      
@@ -140,8 +138,6 @@ extension RouteController: CLLocationManagerDelegate {
             lastUserDistanceToStartOfRoute = userSnappedDistanceToClosestCoordinate
         }
         
-        monitorStepProgress(location)
-        
         guard userIsOnRoute(location) else {
             resetStartCounter()
             NotificationCenter.default.post(name: RouteControllerShouldReroute, object: self, userInfo: [
@@ -149,6 +145,8 @@ extension RouteController: CLLocationManagerDelegate {
                 ])
             return
         }
+        
+        monitorStepProgress(location)
     }
     
     func resetStartCounter() {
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -216,8 +263,8 @@ extension RouteController: CLLocationManagerDelegate {
             
             if routeProgress.currentLegProgress.upComingStep?.maneuverType == ManeuverType.arrive {
                 alertLevel = .arrive
-            } else if courseMatchesManeuverFinalHeading {
-                routeProgress.currentLegProgress.stepIndex += 1
+            } else if courseMatchesManeuverFinalHeading && false {
```
bsudekum: This is initentially false so we always fallback to the code that is implemented in this PR. Will revert this once we're good with the code.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -38,6 +38,8 @@ open class RouteController: NSObject {
         }
     }
     
+    var isInsideManeuverRadius = false
```
1ec5: It's called a maneuver zone; the zone would have a radius.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -162,7 +164,48 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let isCloseToCurrentStep = newLocation.isWithin(radius, of: routeProgress.currentLegProgress.currentStep)
+        
+        // If the user is moving away from the maneuver location
+        // and they are close to the next step
+        // we can safely say they have completed the maneuver.
+        // This is intended to be a fallback case when we do find
+        // that the users course matches the exit bearing.
+        if let upComingStep = routeProgress.currentLegProgress.upComingStep {
+            let isCloseToUpComingStep = newLocation.isWithin(radius, of: upComingStep)
+            if !isCloseToCurrentStep && isCloseToUpComingStep && isInsideManeuverRadius {
```
1ec5: Why does this behavior depend on being within the maneuver zone? If somehow the user has made it to the next step without a location update coming in as they pass through the maneuver zone, we probably would want to advance to the next step regardless of the maneuver zone. You could imagine this case becoming relevant for roundabouts and large intersections.

bsudekum: @1ec5 I'm current guarding against the case where the current step crosses over the following step.

1ec5: That's great, but as I mentioned above, what if the device is unable to update its location while within the maneuver zone but the user has nonetheless proceeded to the next step? (Maybe they passed through the intersection quickly.) What if the intersection happens to be larger than the maneuver zone? (We should probably have different maneuver zone radii for different types of maneuvers.)

bsudekum: @1ec5 i think there are tradeoffs for both methods here.

PR #106, opened 2017-03-29
Add overview button

bsudekum:
Adds an overview button which fits the user location to the end of the route on all location updates

todo:

- [x] style the button
- [x] allow for recenter button to show while panning in overview mode
- [x] i think the camera animations could be improved a bit

foo

PR #203, opened 2017-05-03
Adopt Transifex for localizations

### LINKED ISSUE #168: Adopt Transifex for localizations
We should use Transifex to manage localizations of the navigation SDK, specifically the messages being added in #167.

/cc @bsudekum @frederoni

frederoni:
Fixes #168 
This PR adopts Transifex for pulling in localizations.
New languages added: sv, vi
Update languages: zh-Hans

@1ec5 👀


MapboxNavigation/Resources/sv.lproj/Localizable.strings
```diff
@@ -1,4 +1,4 @@
-﻿﻿/* Format for speech string; 1 = way name; 2 = distance */
+﻿/* Format for speech string; 1 = way name; 2 = distance */
```
frederoni: @1ec5 Any idea why I have to replace this comment manually with the one in Base-language after pulling in from Transifex and converted to UTF-8?
screenshot 2017-05-03 13 40 22

.tx/config
```diff
@@ -0,0 +1,10 @@
+[main]
+host = https://www.transifex.com
+minimum_perc = 80
+type = STRINGS
+
+[mapbox-navigation-ios.localizablestrings]
+file_filter = MapboxNavigation/Resources/<lang>.lproj/Localizable.strings
+lang_map = pt_BR: pt-BR, pt_PT: pt-PT, zh_CN: zh-Hans, zh_TW: zh-Hant
```
1ec5: Since this project only holds iOS code, let’s move `lang_map` up to the `[main]` section.

1ec5:
Catalan is also ready now.

MapboxNavigation/Resources/sv.lproj/Localizable.strings
1ec5: If you used Xcode’s File inspector to convert the file from UTF-16 to UTF-8, it left a [BOM](https://en.wikipedia.org/wiki/Byte_order_mark) at the beginning of the file. Xcode’s strings file copying action, which converts the file from text plist format to binary plist format, only recognizes the BOM when the file is encoded in UTF-16. The workaround is either to cut and paste the file’s contents back into the file in Xcode or to use another text editor (or a tool like `iconv`).

1ec5:
In the iOS SDK project, we’ve added stub Localizable.strings files to the iosapp target so that we can see what the SDK localizations look like while running iosapp. To keep the translation “cost” down, we opted not to make the entire application localizable, since iosapp is rarely seen by developers let alone users. For this project, however, the sample applications’ storyboards include only a few strings, so we might as well make the storyboards localizable as well.

.tx/config
```diff
@@ -0,0 +1,20 @@
+[main]
+host = https://www.transifex.com
+lang_map = pt_BR: pt-BR, pt_PT: pt-PT, zh_CN: zh-Hans, zh_TW: zh-Hant
+minimum_perc = 80
```
1ec5: If #106 lands first, we’ll want to add Navigation.storyboard to the resources and give it a `minimum_perc` much lower than 80%, since most of the strings are “DO NOT TRANSLATE”.

1ec5: If #106 lands first, we’ll want to add Navigation.storyboard to the resources and give it a `minimum_perc` much lower than 80%, since most of the strings are “DO NOT TRANSLATE”.

bsudekum: Note, it hit first.

.tx/config
```diff
@@ -0,0 +1,26 @@
...
+source_file = Examples/Swift/en.lproj/Main.strings
+source_lang = en
+
+[mapbox-navigation-ios.objcexample]
+file_filter = Examples/Objective-C/<lang>.lproj/Main.strings
+source_file = Examples/Objective-C/en.lproj/Main.strings
+source_lang = en
+
+[mapbox-navigation-ios.navigationstrings]
+file_filter = MapboxNavigation/Resources/<lang>.lproj/Localizable.strings
```
1ec5: We want this file to be called Navigation.strings, not Localizable.strings.

.tx/config
```diff
@@ -0,0 +1,26 @@
...
+source_file = MapboxNavigation/Resources/Base.lproj/Localizable.strings
```
1ec5: We need to pull from en.lproj/Navigation.strings.

Issue #192, opened 2017-05-01
Add debug mode


bsudekum:
One item that would be useful for me while driving would to know why a reroute occurred. I think there is potential for more useful information that would be relevant to a developer or a curious driver.

off the cuff list:

* why reroute event occurred (_you were greater than x feet away_)
* number of seconds left on step

This could activated by shake whilst in `DEBUG` release mode. Or just a simple param option the developer sets to true. 

/cc @mapbox/navigation

planemad:
A sample log of reroute events might be worth investigating for potential data issues on OpenStreetMap.

The existing [direction-feedback](https://github.com/mapbox/directions-feedback/issues) and routing failures is inspected by @srividyacb, and rerouting logs could be a potential source to make more data improvements.

cc @shvrm @oini

ericrwolfe:
Addressed by #304.

PR #205, opened 2017-05-03
Added debug mode

frederoni:
Possible fix for #192 

Trying to find why rerouting occurs where it shouldn't.

❤️  Actual location update
💚  Closest point on route
💙  Dead reckoning modified location
<hr>

Everything worked as intended.
<hr>

This ^ triggered a reroute when it should've snapped to the road. Seems to happen more often when the accuracy is good (around 10 in horizontal accuracy)
<hr>

More to come..
@bsudekum will try to incorporate your changes and rerun.

### DIFF
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -1,0 +1,54 @@
+import Foundation
+import CoreLocation
+import MapboxDirections
+
+public extension CLLocation {
+    /**
+     Returns a `CLLocation` infront of the provided location given the interval
+     with the current speed taken into account.
+     */
+    public func advanced(by interval: TimeInterval) -> CLLocation {
+        let metersAhead = speed * interval
+        let locationInfrontOfUser = coordinate.coordinate(at: metersAhead, facing: course)
+        return CLLocation(coordinate: locationInfrontOfUser.coordinate(at: metersAhead, facing: course),
+                          altitude: altitude,
+                          horizontalAccuracy: horizontalAccuracy,
+                          verticalAccuracy: verticalAccuracy,
+                          course: course,
+                          speed: speed,
+                          timestamp: timestamp.addingTimeInterval(interval))
+    }
+    
+    public func isOnRoute(with routeProgress: RouteProgress) -> Bool {
+        let newLocation = advanced(by: RouteControllerDeadReckoningTimeInterval)
+        let isCloseToCurrentStep = newLocation.isWithin(rerouteRadius, of: routeProgress.currentLegProgress.currentStep)
+        
+        // If the user is moving away from the maneuver location
+        // and they are close to the next step
+        // we can safely say they have completed the maneuver.
+        // This is intended to be a fallback case when we do find
+        // that the users course matches the exit bearing.
+        if let upComingStep = routeProgress.currentLegProgress.upComingStep {
+            let isCloseToUpComingStep = newLocation.isWithin(rerouteRadius, of: upComingStep)
+            if !isCloseToCurrentStep && isCloseToUpComingStep {
+                return !isCloseToCurrentStep && isCloseToUpComingStep
+            }
+        }
+        
+        return isCloseToCurrentStep
+    }
+    
+    /**
+     Returns the maximum reroute radius
+     */
+    public static func radiusForRerouting(with horizontalAccuracy: CLLocationAccuracy) -> CLLocationDistance {
+        return max(RouteControllerMaximumDistanceBeforeRecalculating, horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
+    }
+    
+    /**
+     Returns the maximum reroute radius
+     */
+    public var rerouteRadius: CLLocationDistance {
+        return CLLocation.radiusForRerouting(with: horizontalAccuracy)
+    }
+}
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -154,13 +154,26 @@ extension RouteController: CLLocationManagerDelegate {
         lastUserDistanceToStartOfRoute = Double.infinity
     }
     
-    public func userIsOnRoute(_ location: CLLocation) -> Bool {
-        // Find future location of user
+    
+    /**
+     Returns the calculated reroute distance.
+     */
+    public func getRerouteRadius(_ horizontalAccuracy: CLLocationAccuracy) -> CLLocationDistance {
+        return min(RouteControllerMaximumDistanceBeforeRecalculating, horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
+    }
+    
+    /**
+     Takes a `CLLocation` and returns a `CLLocation` infront of the provided location given the current speed.
+     */
+    public func getDeadReckoningLocation(_ location: CLLocation) -> CLLocation {
         let metersInFrontOfUser = location.speed * RouteControllerDeadReckoningTimeInterval
         let locationInfrontOfUser = location.coordinate.coordinate(at: metersInFrontOfUser, facing: location.course)
-        let newLocation = CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
-        let radius = min(RouteControllerMaximumDistanceBeforeRecalculating,
-                         location.horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
+        return CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
+    }
+    
+    public func userIsOnRoute(_ location: CLLocation) -> Bool {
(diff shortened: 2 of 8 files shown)


bsudekum:
@frederoni what do you think about moving away from an `UIAlertController`? It requires the driver to click `Cancel` which could get a little annoying.

frederoni:
I tried to find another solution but each dot contains different information so we need an overlay or something. We could remove the cancel button and let the user dismiss the alert controller wherever the tap occurs.

bsudekum:
Code looks good. @1ec5 how do you feel about the addition as a whole?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -299,6 +303,43 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
...
+    func addDebugAnnotation(location: CLLocation) {
+        let annotation = DebugAnnotation()
+        annotation.coordinate = location.coordinate
+        annotation.title = "Reroute"
+        annotation.color = .red
+        let metersInFrontOfUser = location.speed * RouteControllerDeadReckoningTimeInterval
+        let locationInfrontOfUser = location.coordinate.coordinate(at: metersInFrontOfUser, facing: location.course)
+        let newLocation = CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
+        let radius = min(RouteControllerMaximumDistanceBeforeRecalculating,
+                         location.horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
```
frederoni: It would be safer to factor out and reuse this logic from here and `RouteController` as @1ec5 suggested.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -587,6 +589,8 @@
 				C58159001EA6D02700FC6C3D /* MGLVectorSource.swift */,
 				351BEBEB1E5BCC63006FE110 /* String.swift */,
 				351BEBF01E5BCC63006FE110 /* UIView.swift */,
+				C58159001EA6D02700FC6C3D /* MGLVectorSource.swift */,
```
frederoni: Looks like a merge issue

bsudekum:
factored reroute logic in https://github.com/mapbox/mapbox-navigation-ios/pull/205/commits/063fdcb203c975cb2b47ced824641ebc23661689

MapboxCoreNavigation/RouteController.swift
```diff
@@ -154,13 +154,26 @@ extension RouteController: CLLocationManagerDelegate {
...
+     Returns the calculated reroute distance.
+     */
+    public func getRerouteRadius(_ horizontalAccuracy: CLLocationAccuracy) -> CLLocationDistance {
+        return min(RouteControllerMaximumDistanceBeforeRecalculating, horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
+    }
+    
+    /**
+     Takes a `CLLocation` and returns a `CLLocation` infront of the provided location given the current speed.
+     */
+    public func getDeadReckoningLocation(_ location: CLLocation) -> CLLocation {
```
1ec5: A method name only begins with `get` if it starts an asynchronous network request or similar. [A normal getter starts with a noun phrase](https://swift.org/documentation/api-design-guidelines/#name-according-to-side-effects).

This method has no dependencies on RouteController, so it would be cleaner to put it in an extension on CLLocation, as a method called `advanced(by:)` that takes a `TimeInterval`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -154,13 +154,26 @@ extension RouteController: CLLocationManagerDelegate {
...
         let metersInFrontOfUser = location.speed * RouteControllerDeadReckoningTimeInterval
         let locationInfrontOfUser = location.coordinate.coordinate(at: metersInFrontOfUser, facing: location.course)
-        let newLocation = CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
-        let radius = min(RouteControllerMaximumDistanceBeforeRecalculating,
-                         location.horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
+        return CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
+    }
+    
+    public func userIsOnRoute(_ location: CLLocation) -> Bool {
```
1ec5: One way to tell that the method’s name needs revision is that the first argument is called `location` internally (and is a CLLocation) but is called a “route” externally; a location isn’t a route. To fix this issue, move this method to an extension on CLLocation and make its only argument a Route. There’s already a similar extension method in Geometry.swift.

frederoni:
In the latest [commit](https://github.com/mapbox/mapbox-navigation-ios/pull/205/commits/d256892904dddbb731bef6a2cabe6e6be04bfaac) I moved `increment step` and `reset alert level`
out from `isOnRoute` (which is a confusing place to increment step and reset alert level) into `locationManager(_:didUpdateLocations:)`. This means it will not get called twice per location update like it did before. (We check for `isOnRoute` twice in `didUpdateLocations`)

This resulted in `testDepart` started failing.

^ @bsudekum @1ec5

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -0,0 +1,57 @@
+import Foundation
+import CoreLocation
+import MapboxDirections
+
+public extension CLLocation {
+    /**
+     Returns a `CLLocation` infront of the provided location given the interval
+     with the current speed taken into account.
+     */
+    public func advance(by interval: TimeInterval) -> CLLocation {
```
1ec5: Replace `advance` with `advanced`: `advance` is an imperative form, which according to the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/) means it has a side effect, whereas this method only makes and returns a copy.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public static func rerouteRadius(_ horizontalAccuracy: CLLocationAccuracy) -> CLLocationDistance {
+        return min(RouteControllerMaximumDistanceBeforeRecalculating, horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
+    }
+    
+    /**
+     Returns the calculated reroute distance.
+     */
+    public var rerouteRadius: CLLocationDistance {
+        get {
```
1ec5: Nit: This line isn’t needed for a read-only computed variable.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+        
+        return isCloseToCurrentStep
+    }
+    
+    /**
+     Returns the calculated reroute distance.
+     */
+    public static func rerouteRadius(_ horizontalAccuracy: CLLocationAccuracy) -> CLLocationDistance {
```
1ec5: Replace `_` with `with` or `for`. `_` would only be appropriate if `horizontalAccuracy` could be called a “radius”.

Also, “reroute” is an imperative form, suggesting that this method actually reroutes something. Call this method `radiusForRerouting(with:)`.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -0,0 +1,57 @@
...
+        return CLLocation(coordinate: locationInfrontOfUser.coordinate(at: metersAhead, facing: course),
+                          altitude: altitude,
+                          horizontalAccuracy: horizontalAccuracy,
+                          verticalAccuracy: verticalAccuracy,
+                          course: course,
+                          speed: speed,
+                          timestamp: timestamp.addingTimeInterval(interval))
+    }
+    
+    public func isOnRoute(_ routeProgress: RouteProgress) -> Bool {
```
1ec5: `_` is only appropriate if `routeProgress` is a “route”. In this case, since `routeProgress` is a “route progress”, call this method `isOnRoute(with:)` or have it take the current and upcoming steps directly: `isOn(_:upComingStep:)`.

frederoni:
`testDepart` started failing because `speed` was currently ignored but it's now taken into account with `advanced(by:)` causing the distance to be 19 instead of 9. This was not within the reroute radius due to horizontal accuracy being set to 1. So changed min to max [here](https://github.com/mapbox/mapbox-navigation-ios/pull/205/files#diff-dfd8aa8fe700af70004116a3a54c9d84R45) which @1ec5 suggested as a fix for continuous rerouting.

1ec5:
> I changed min to max [here](https://github.com/mapbox/mapbox-navigation-ios/pull/205/files#diff-dfd8aa8fe700af70004116a3a54c9d84R45) which @1ec5 suggested as a fix for continuous rerouting.

@bsudekum is also making this change in #228.

bsudekum:
Let's either close this out and label as experimental or get this merged in.

frederoni:
Superseded by other debugging methods.

PR #207, opened 2017-05-04
Update dependencies

bsudekum:
brings in new osrm-text-instruction changes.


MapboxNavigation.podspec
```diff
@@ -45,7 +45,7 @@ Pod::Spec.new do |s|
   s.dependency "MapboxCoreNavigation", "#{s.version.to_s}"
   s.dependency "MapboxDirections.swift", "~> 0.9.0"
   s.dependency "Mapbox-iOS-SDK", "~> 3.5"
-  s.dependency "OSRMTextInstructions", "~> 0.1.1"
+  s.dependency "OSRMTextInstructions", "~> 0.1"
```
1ec5: The use of a patch version here was intentional: #150. v0.2 isn’t guaranteed to be backwards-compatible with v0.1.x.

bsudekum: @1ec5 updated and pinned to 0.1.2

Issue #174, opened 2017-04-25
Could not find storyboard named 'Navigation'


Threepwood9985:
I have an existing Objective-C project and installed MapboxNavigation via cocoapods
` pod 'MapboxNavigation'` and run `pod install`

```
Installing AWSCore (2.5.5)
Installing AWSPolly (2.5.5)
Installing Mapbox-iOS-SDK (3.5.2)
Installing MapboxCoreNavigation (0.3.0)
Installing MapboxDirections.swift (0.9.0)
Installing MapboxNavigation (0.3.0)
Installing OSRMTextInstructions (0.1.1)
Installing Polyline (4.1.1)
Installing Pulley (1.3.1)
Installing SDWebImage (4.0.0)
```

I used [this](https://github.com/mapbox/mapbox-navigation-ios/blob/master/docs/Storyboards.md) guide to set up the turn by turn navigation in a storyboard. 
But when I build and run the app and tap on "start navigation" to display the view controller, i get the following error:

`Could not find a storyboard named 'Navigation' in bundle NSBundle`

Do i have to add the folder "MapboxNavigation" like in the Objective-C Example? Because the named storyboard is in there...

Or do i have to embed the Frameworks manually into the "embedded binaries" section? 
Thanks for your support.

frederoni:
Thanks for reporting this issue.
You only have to embed the framework manually if you're using Carthage, not when using CocoaPods.

Can you try to update your Podfile to use the master branch of this repo?
```
pod 'MapboxCoreNavigation', :git => 'https://github.com/mapbox/MapboxNavigation.swift.git', :branch => 'master'
pod 'MapboxNavigation', :git => 'https://github.com/mapbox/MapboxNavigation.swift.git', :branch => 'master'
```
this will take care of the `Could not find a storyboard named 'Navigation' in bundle NSBundle` error until we release a new version.

1ec5:
It looks like @Threepwood9985 is using v0.3.0, which is the latest version released several hours ago. [This documentation](https://github.com/mapbox/mapbox-navigation-ios/blob/69e69e4aa0f2215718e78f362e174f60294e3b1a/docs/Storyboards.md) would be up-to-date. Do you have a particular fix in mind, @frederoni?

frederoni:
Oh, ignore https://github.com/mapbox/mapbox-navigation-ios/issues/174#issuecomment-297042301

CocoaPods renames the bundle identifier.
[...]

PR #212, opened 2017-05-08
Fix resource filepath

bsudekum:
Addresses issues around https://github.com/mapbox/mapbox-navigation-ios/issues/174

### DIFF
--- a/MapboxNavigation.podspec
+++ b/MapboxNavigation.podspec
@@ -35,7 +35,7 @@ Pod::Spec.new do |s|
 
   # ――― Resources ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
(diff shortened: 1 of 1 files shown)


MapboxNavigation.podspec
```diff
@@ -35,7 +35,7 @@ Pod::Spec.new do |s|
 
   # ――― Resources ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
-  s.resources = ['MapboxNavigation/Resources/**']
+  s.resources = ['MapboxNavigation/Resources/*/*']
```
1ec5: This new pattern excludes [Shields.plist](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/Resources/Shields.plist), which sits directly under Resources. Add another path to this array with just `*` to handle this case.

PR #217, opened 2017-05-10
Make distance label larger

bsudekum:
This changes the instruction header to:

* Puts the distance above the street name
* If no highway is present, the text can be full width
* Now the street name must be one line. Because of this, the font size shrinks. However, if there is no distance, it can be 2 lines (currently buggy)

simulator screen shot may 10 2017 1 29 22 pm

simulator screen shot may 10 2017 1 50 52 pm

To fix: 

simulator screen shot may 10 2017 1 50 44 pm
> Font size looks like it's the right size when there is no distance, however, it's clipped to a single line even though I'm saying it can have 2 lines

@frederoni could you take a look into the 2 line issues above?


MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -198,94 +198,115 @@
...
+                                        </subviews>
                                         <constraints>
-                                            <constraint firstAttribute="height" constant="32" id="0ei-is-yQS"/>
-                                            <constraint firstAttribute="width" constant="32" id="4qd-KG-vi7"/>
+                                            <constraint firstItem="gkj-hG-NxS" firstAttribute="top" secondItem="dhM-Ng-gx5" secondAttribute="top" constant="19" id="AOc-29-8va"/>
+                                            <constraint firstItem="c3B-6D-Ukh" firstAttribute="leading" secondItem="dhM-Ng-gx5" secondAttribute="leading" id="Dcy-eQ-VCk"/>
+                                            <constraint firstAttribute="bottom" secondItem="gkj-hG-NxS" secondAttribute="bottom" constant="19" id="EBu-yb-gc6"/>
+                                            <constraint firstAttribute="bottom" secondItem="c3B-6D-Ukh" secondAttribute="bottom" id="Ofy-g6-EBl"/>
+                                            <constraint firstItem="gkj-hG-NxS" firstAttribute="leading" secondItem="c3B-6D-Ukh" secondAttribute="trailing" id="mWA-J9-gaq"/>
```
1ec5: There needs to be a gap between the right edge of the street name label (and distance label) and the left edge of the shield view. The gap should be the same width as the right margin of the shield view and ideally the same width as the gap between the maneuver icon and the street name label.

frederoni:
Changed word wrap to truncate tail and 50% minimum font scale.

Also added [FBSnapshotTestCase](https://github.com/facebook/ios-snapshot-test-case) and fixtures to verify single and multiline. We should add a few more cases to cover w/ and w/o a visible shield.

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
frederoni: Added 8pts spacing in the horizontal stack view.

bsudekum:
@1ec5 I think this ready for review now

bsudekum:
Also noting, we were only showing lanes when paging through steps, added them back in https://github.com/mapbox/mapbox-navigation-ios/pull/217/commits/51f9e1a22ec6003594d57de9b89ed6f6ebcf02fb

MapboxNavigation/Abbreviations.swift
```diff
@@ -34,17 +34,19 @@ extension String {
     }
     
     /// Returns the string abbreviated only as much as necessary to fit the given width and font.
-    func abbreviated(width: CGFloat, font: UIFont) -> String {
+    func abbreviated(bounds: CGRect, font: UIFont) -> String {
```
1ec5: This method should be called `abbreviated(toFit:font:)`, not `abbreviated(bounds:font:)`. The relationship between the method name and the argument value should be clear when reading the method call.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -517,21 +542,20 @@ extension RouteMapViewController: RoutePageViewControllerDelegate {
...
+            maneuverViewController.streetLabel.text = name.abbreviated(bounds: streetLabelBounds, font: streetLabelFont)
         } else if let destinations = step?.destinations?.joined(separator: "\n") {
-            maneuverViewController.streetLabel.text = destinations.abbreviated(width: streetLabelWidth, font: streetLabelFont)
+            maneuverViewController.streetLabel.text = destinations.abbreviated(bounds: streetLabelBounds, font: streetLabelFont)
         } else if let step = step {
-            maneuverViewController.streetLabel.text = routeStepFormatter.string(for: step)?.abbreviated(width: streetLabelWidth, font: streetLabelFont)
+            maneuverViewController.streetLabel.text = routeStepFormatter.string(for: step)?.abbreviated(bounds: streetLabelBounds, font: streetLabelFont)
         }
-        maneuverViewController.distanceLabel.text = step!.distance > 0 ? distanceFormatter.string(from: step!.distance) : ""
+        maneuverViewController.distanceText = step!.distance > 0 ? distanceFormatter.string(from: step!.distance) : nil
```
1ec5: It would be cleaner to replace the `distanceText` property with a `distance` property whose `didSet` takes care of formatting the numeric distance into a string.

frederoni:
> testManeuverViewSingleLine@2x.png doesn’t show much of a gap between the street name label and the shield view – do you know why that might be?

I recorded the 2x before I added the spacing but I removed the unused reference images ( d6b80903ddb32670e882401aeb71231652a73913 ) since we only test on iPhone 6s Plus 3x on Bitrise. Let's add more devices when we've verified that the snapshot tests are reliable and easy to implement in our workflow.

PR #223, opened 2017-05-11
Prevent users from swiping past current step

bsudekum:
I think it's confusing for users when swiping backwards and you end up past the current step.

### DIFF
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -162,8 +162,8 @@ open class RouteLegProgress: NSObject {
         guard let index = leg.steps.index(of: step) else {
             return nil
         }
-        if index > 0 && index > stepIndex {
-            return leg.steps[index - 1]
+        if index > 0 {
+            return leg.steps[index-1]
         }
         return nil
     }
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -162,8 +162,8 @@ open class RouteLegProgress: NSObject {
         guard let index = leg.steps.index(of: step) else {
             return nil
         }
-        if index > 0 {
-            return leg.steps[index-1]
+        if index > 0 && index > stepIndex {
```
1ec5: The stated purpose of this method is to return the step before the given step. Nothing about this method says it has to be used for swiping between steps in the UI – what if we try to use it in the future to discover something about the previous step?

This check should be moved to code that’s specifically about swiping between steps.

bsudekum:
@1ec5 fixed

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -564,6 +564,10 @@ extension RouteMapViewController: RoutePageViewControllerDelegate {
     }
 
     func stepBefore(_ step: RouteStep) -> RouteStep? {
+        guard let index = routeController.routeProgress.currentLeg.steps.index(of: step),
```
1ec5: Nit: this could also be written `routeController.routeProgress.currentLegProgress.leg.steps.index(of: step)`, which would enable us to factor out `routeController.routeProgress.currentLegProgress` into a local variable. It wouldn’t be a significant performance optimization, just a way to keep things short and sweet.

PR #209, opened 2017-05-05
WIP improve simulated route

frederoni:
WIP improving simulated route
Specifically slowing down for sharp turns.

PR #226, opened 2017-05-14
Improve route simulation

frederoni:
This is a cleaned up version of #209 without the Playground.

The route simulation does not take shortcuts or cruise at maximum speed through sharp turns or roundabouts anymore.

@bsudekum @1ec5 👀

### DIFF
--- a/MapboxNavigation/SimulatedRoute.swift
+++ b/MapboxNavigation/SimulatedRoute.swift
@@ -3,7 +3,6 @@ import CoreLocation
 
 fileprivate let maximumSpeed: CLLocationSpeed = 30 // ~108 kmh
 fileprivate let minimumSpeed: CLLocationSpeed = 5 // ~18 kmh
-fileprivate var speed: CLLocationSpeed = 30
 fileprivate var distanceFilter: CLLocationDistance = 10
 fileprivate var verticalAccuracy: CLLocationAccuracy = 40
 fileprivate var horizontalAccuracy: CLLocationAccuracy = 40
@@ -42,8 +41,8 @@ class SimulatedRoute : NSObject {
         processedLocations = polyline
             .simulatedLocationsWithTurnPenalties()
             .interpolated()
-            .calculatedSpeed()
-            .calculatedTimestamp()
+            .withSpeeds()
+            .withShiftedTimestamps()
     }
     
     func start() {
@@ -142,7 +141,7 @@ fileprivate extension Array where Element == SimulatedLocation {
         return processedLocations.count > count ? processedLocations.interpolated() : processedLocations
     }
     
-    fileprivate func calculatedTimestamp() -> [SimulatedLocation] {
+    fileprivate func withShiftedTimestamps() -> [SimulatedLocation] {
         var processedLocations = [SimulatedLocation]()
         
         for location in self {
@@ -150,9 +149,8 @@ fileprivate extension Array where Element == SimulatedLocation {
                 processedLocations.append(location.shifted(to: Date()))
             } else {
                 let distance = location.distance(from: processedLocations.last!)
-                let time = distance / location.speed
-                let newLocation = location.shifted(to: processedLocations.last!.timestamp.addingTimeInterval(time))
-                processedLocations.append(newLocation)
+                let shiftedLocation = location.shifted(to: processedLocations.last!.timestamp.addingTimeInterval(distance / location.speed))
+                processedLocations.append(shiftedLocation)
             }
         }
         
@@ -160,20 +158,19 @@ fileprivate extension Array where Element == SimulatedLocation {
     }
     
     // Calculate speed based on turn penalty.
-    fileprivate func calculatedSpeed() -> [SimulatedLocation] {
-        var processedLocations = [SimulatedLocation]()
-        
-        for location in self {
-            let reversedTurnPenalty = maximumTurnPenalty - location.turnPenalty
-            let speed = reversedTurnPenalty.scale(minimumIn: minimumTurnPenalty, maximumIn: maximumTurnPenalty, minimumOut: minimumSpeed, maximumOut: maximumSpeed)
-            processedLocations.append(location.withSpeed(speed))
-        }
-        
-        return processedLocations
+    fileprivate func withSpeeds() -> [SimulatedLocation] {
+        return map({$0.withPenalizedSpeed})
     }
 }
 
 fileprivate extension SimulatedLocation {
+    
+    var withPenalizedSpeed: SimulatedLocation {
+        let reversedTurnPenalty = maximumTurnPenalty - turnPenalty
+        let speed = reversedTurnPenalty.scale(minimumIn: minimumTurnPenalty, maximumIn: maximumTurnPenalty, minimumOut: minimumSpeed, maximumOut: maximumSpeed)
+        return withSpeed(speed)
+    }
+    
     func shifted(to shiftedTimestamp: Date) -> SimulatedLocation {
         let location = SimulatedLocation(coordinate: coordinate,
                                          altitude: altitude,
--- a/MapboxNavigation/SimulatedRoute.swift
+++ b/MapboxNavigation/SimulatedRoute.swift
@@ -1,6 +1,17 @@
 import Foundation
 import CoreLocation
 
+fileprivate let maximumSpeed: CLLocationSpeed = 30 // ~108 kmh
+fileprivate let minimumSpeed: CLLocationSpeed = 5 // ~18 kmh
+fileprivate var speed: CLLocationSpeed = 30
+fileprivate var distanceFilter: CLLocationDistance = 10
+fileprivate var verticalAccuracy: CLLocationAccuracy = 40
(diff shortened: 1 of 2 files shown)


MapboxNavigation/SimulatedRoute.swift
```diff
@@ -5,85 +5,198 @@ protocol SimulatedRouteDelegate : class {
...
 }
 
+class SimulatedLocation: CLLocation {
+    var turnPenalty: Double = 0
+}
+
 class SimulatedRoute : NSObject {
-    
+    let maximumSpeed: CLLocationSpeed = 30 // ~108 kmh
+    let minimumSpeed: CLLocationSpeed = 5 // ~18 kmh
```
bsudekum: Should any of these variables be made public to the developer?

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -5,85 +5,198 @@ protocol SimulatedRouteDelegate : class {
...
+    fileprivate let minimumTurnPenalty: CLLocationDirection = 0
     
+    var processedLocations: [CLLocation]!
     let locationManager = CLLocationManager()
-    
     weak var delegate: SimulatedRouteDelegate?
     
-    convenience init(_ polyline: [CLLocationCoordinate2D]) {
+    convenience init?(along polyline: [CLLocationCoordinate2D]) {
+        if polyline.count < 3 {
```
bsudekum: This should probably be a guard.

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -5,85 +5,198 @@ protocol SimulatedRouteDelegate : class {
...
+            if i < locations.count-1 {
+                let speed = (maximumSpeed - minimumSpeed) / 2
+                let nextLocation = locations[i+1]
+                let distance = floor(location.distance(from: nextLocation))
+                let timeToTravel = distance / speed
+                
+                if timeToTravel > 1 {
+                    let newCoordinate = coordinate(at: speed, fromStartOf: [location.coordinate, nextLocation.coordinate])
+                    let newLocation = SimulatedLocation(coordinate: newCoordinate!,
+                                                        altitude: 0,
```
bsudekum: Just in case we decided to do something with altitude, say for overlapping over surface streets and highways, maybe this should be changed to:

```
altitude: location.altitude,
```

MapboxNavigation/SimulatedRoute.swift
1ec5: That would probably require us to make SimulateRoute public as well.

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -5,85 +5,198 @@ protocol SimulatedRouteDelegate : class {
...
+        
+        let turnPenaltyBasedLocations = calculateTurnPenalty(coordinates: polyline)
+        let interpolatedLocations = interpolate(locations: turnPenaltyBasedLocations)
+        let speedBasedLocations = calculateSpeed(for: interpolatedLocations)
+        
+        processedLocations = calculateTimestamp(for: speedBasedLocations)
+    }
+    
+    // Calculate timestamp when speed has been calculated.
+    fileprivate func calculateTimestamp(for locations: [SimulatedLocation]) -> [SimulatedLocation] {
```
1ec5: This method [has no side effects](https://swift.org/documentation/api-design-guidelines/#name-according-to-side-effects), so it should be named with a noun phrase instead of a verb. It doesn’t seem to use `self` in any way, so maybe it makes sense for it to be a standalone function, an extension method like `SimulatedLocation.simulatedLocationsBeginningNow(for:)`, or even a method extending `[SimulatedLocation]`.

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -5,85 +5,198 @@ protocol SimulatedRouteDelegate : class {
...
+            } else {
+                processedLocations.append(location.shifted(to: Date()))
+            }
+        }
+        
+        return processedLocations
+    }
+    
+    // Calculate speed based on turn penalty.
+    fileprivate func calculateSpeed(for locations: [SimulatedLocation]) -> [SimulatedLocation] {
```
1ec5: Does it ever make sense for this method or `calculateTimestamp(for:)` to be called independently of `interpolate(locations:)`? If not, this functionality should be folded into `interpolate(locations:)`. Calculating speed and time are intrinsic parts of interpolating locations. Combining the functions would avoid the method naming conundrums too.

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -5,85 +5,198 @@ protocol SimulatedRouteDelegate : class {
...
+        var processedLocations = [SimulatedLocation]()
+        
+        for (i, location) in locations.enumerated() {
+            processedLocations.append(location)
             
-            if let nextCoordinate = coordinate(at: distance + distanceFilter, fromStartOf: polyline) {
-                course = currentCoordinate.direction(to: nextCoordinate)
+            if i < locations.count-1 {
+                let speed = (maximumSpeed - minimumSpeed) / 2
+                let nextLocation = locations[i+1]
```
1ec5: It looks like `i` is only used to find the subsequent location. A functional approach when you want to look ahead would be to `zip` two subarrays of `locations`, similar to [this example](https://github.com/mapbox/MapboxDirections.swift/blob/615be18ac5f0892fbfc4457b19c7e1cc4fe1a2fd/MapboxDirections/MBRoute.swift#L31-L32) in MapboxDirections.swift.

MapboxNavigation/SimulatedRoute.swift
frederoni: Keeping them separated made more sense in the Playground where I visualized them individually in three style layers.

MapboxNavigation/SimulatedRoute.swift
frederoni: I extended `[SimulatedLocation]` initially but it didn't work in a Playground. Perhaps a unit test would be better.

frederoni: Scratch 👆. I extended `Element : SimulatedLocation` instead of `Element == SimulatedLocation` so type of self couldn't be inferred and the Playground recompiled indefinitely.

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -1,6 +1,17 @@
 import Foundation
 import CoreLocation
 
+fileprivate let maximumSpeed: CLLocationSpeed = 30 // ~108 kmh
+fileprivate let minimumSpeed: CLLocationSpeed = 5 // ~18 kmh
+fileprivate var speed: CLLocationSpeed = 30
```
1ec5: This variable no longer appears to be used. (Some functions shadow it with a computed speed.)

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -9,169 +20,156 @@ class SimulatedLocation: CLLocation {
...
-    func stop() {
-        NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(tick), object: nil)
-    }
-}
-
-fileprivate extension Double {
-    func scale(minimumIn: Double, maximumIn: Double, minimumOut: Double, maximumOut: Double) -> Double {
-        return ((maximumOut - minimumOut) * (self - minimumIn) / (maximumIn - minimumIn)) + minimumOut
+    // Calculate speed based on turn penalty.
+    fileprivate func calculatedSpeed() -> [SimulatedLocation] {
```
1ec5: The name `calculatedSpeed()` implies that the method returns a speed. `withSpeeds()` would be more intuitive here.

MapboxNavigation/SimulatedRoute.swift
```diff
@@ -9,169 +20,156 @@ class SimulatedLocation: CLLocation {
...
+        var processedLocations = [SimulatedLocation]()
+        
+        for location in self {
+            let reversedTurnPenalty = maximumTurnPenalty - location.turnPenalty
+            let speed = reversedTurnPenalty.scale(minimumIn: minimumTurnPenalty, maximumIn: maximumTurnPenalty, minimumOut: minimumSpeed, maximumOut: maximumSpeed)
+            processedLocations.append(location.withSpeed(speed))
```
1ec5: Instead of building a mutable array, call `map(_:)` on `self`.

1ec5: On second thought, maybe we should extend SimulatedLocation with a computed variable `withPenalizedSpeed`:

```swift
var withPenalizedSpeed: SimulatedLocation {
    let reversedTurnPenalty = maximumTurnPenalty - turnPenalty
    let speed = reversedTurnPenalty.scale(minimumIn: minimumTurnPenalty, maximumIn: maximumTurnPenalty, minimumOut: minimumSpeed, maximumOut: maximumSpeed)
    return withSpeed(speed)
}
```

Then the call to `calculatedSpeed()` becomes simply `map(SimulatedLocation.withPenalizedSpeed)`.

Issue #122, opened 2017-04-07
Add "routeControllerShouldReroute" delegate method


ericrwolfe:
In addition to supporting automatic rerouting by default, we should consider offering developers finer control of rerouting behavior with an optional `routeControllerShouldReroute` delegate method that returns a bool.

The default implementation would return true, which indicates that automatic rerouting is enabled.

```swift
func routeViewControllerShouldReroute(viewController: RouteViewController) -> Bool {
    // Possible scenarios:
    // Check if we've rerouted too frequently
    // Check if internet connection is spotty or cellular not allowed
    // Log that a user went off course and needed a reroute
    
    // Handle rerouting manually, perhaps with different routing options
    Directions.shared.calculate(options) { (waypoints, routes, error) in
        // Check if this reroute makes sense?
        
        // Possible hook to make your own re-routing announcement
        
        // manually update the navigation route
        viewController.route = routes.first!
    }
    
    return false // disables automatic rerouting
}
```

wwajerowicz:
That's something that would be quite useful for me. I'm happy to have a go at implementing it and submit a PR.

I'd suggest adding a following method to the delegate:
``` 
@objc optional func navigationViewControllerShouldReroute(_ : NavigationViewController, fromLocation location: CLLocation)
```
and modifying `shouldReroute` method to something like this:
```
func shouldReroute(notification: NSNotification) {
        let location = notification.userInfo![RouteControllerNotificationShouldRerouteKey] as! CLLocation
        if navigationDelegate?.routeViewControllerShouldReroute?(self, fromLocation: location) != nil {
           return;
        }
             
  //continue as normal

}
    
```
That way, if someone implements that method in the delegate, they will be responsible for handling rerouting.

Let me know if that makes sense and I'll have a go at implementing it.

frederoni:
@wwajerowicz A PR would be great.
Should we also pass the current `route` object to the receiver along with the last known location?

wwajerowicz:
@frederoni I think `route` object would be useful, but it's possible to get it directly from `NavigationViewController`, so I'm not sure if it's necessary to pass it directly to the function.

PR #227, opened 2017-05-15
Allow custom rerouting logic.

wwajerowicz:
Resolves https://github.com/mapbox/mapbox-navigation-ios/issues/122

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -28,7 +28,7 @@ public protocol NavigationViewControllerDelegate {
      If this method is unimplemented `NavigationViewController` will automatically reroute. If this method is implemented,
      it's responsible for handling rerouting.
     */
-    @objc optional func navigationViewControllerShouldReroute(_ : NavigationViewController, fromLocation location: CLLocation)
+    @objc(navigationViewController:shouldRerouteFromLocation:) optional func navigationViewController(_ : NavigationViewController, shouldRerouteFrom location: CLLocation)
     
     /**
      Returns an `MGLStyleLayer` that determines the appearance of the route line.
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -280,6 +288,10 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
     
     func shouldReroute(notification: NSNotification) {
         let location = notification.userInfo![RouteControllerNotificationShouldRerouteKey] as! CLLocation
+
+        if navigationDelegate?.navigationViewControllerShouldReroute?(self, fromLocation: location) != nil {
```
bsudekum: Nit: want to make this a `guard` instead?

wwajerowicz: Fixed.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -21,6 +21,14 @@ public protocol NavigationViewControllerDelegate {
...
      */
     @objc optional func navigationViewController(_ navigationViewController : NavigationViewController, didArriveAt destination: MGLAnnotation)
+
+    /**
+     Called when `RouteControllerShouldReroute` is sent from `RouteController`.
+
+     If this method is unimplemented `NavigationViewController` will automatically reroute. If this method is implemented,
+     it's responsible for handling rerouting.
+    */
+    @objc optional func navigationViewControllerShouldReroute(_ : NavigationViewController, fromLocation location: CLLocation)
```
1ec5: The method should be called `navigationViewController(_:shouldRerouteFrom:)`, with an attribute of `@objc(navigationViewController:shouldRerouteFromLocation:)`.

wwajerowicz: @1ec5 yes, that method name makes much more sense.

PR #228, opened 2017-05-15
Change reroute radius threshold calculation

bsudekum:
We've had reports of constant rerouting on highways. Our current thinking that this is a combination of a few factors:

* user has very high gps accuracy
* user is traveling on large highway with many lanes
* user has high horizontal distance from the route geometry (I.E. in first lane)

These combinations could produce a rerouting event since their `horizontalAccuracy ` is < 10m and the `RouteControllerUserLocationSnappingDistance` is `10m. This produces a radius threshold of only 20m. 

This change now looks at the maximum value between `RouteControllerMaximumDistanceBeforeRecalculating` (50m) and the `horizontalAccuracy` + `RouteControllerUserLocationSnappingDistance` (10m).

Here are the situations that will change:
* :+1: fewer rerouting on a highway with good accuracy 
* :-1: possibly longer time to reroute while in urban areas (with worse horizontal accuracy)

/cc @mapbox/navigation @danpat @willwhite

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -159,7 +159,7 @@ extension RouteController: CLLocationManagerDelegate {
         let metersInFrontOfUser = location.speed * RouteControllerDeadReckoningTimeInterval
         let locationInfrontOfUser = location.coordinate.coordinate(at: metersInFrontOfUser, facing: location.course)
         let newLocation = CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
-        let radius = min(RouteControllerMaximumDistanceBeforeRecalculating,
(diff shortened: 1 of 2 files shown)


bsudekum:
In https://github.com/mapbox/mapbox-navigation-ios/pull/228/commits/f2c9e6910a5d0524772f2d86f41e09edc2a975d4, I changed when we snap. Previously we:

* if snapping was enabled, we always returned a snapped location
* even if we errored while calculating the snapping course, we still returned a snapped location
* if a user was not snapped, we rerouted

Now:

* if anything errors while calculating the snapped location or location we return a non snapped location
* We check if the snapped location is greater than `RouteControllerUserLocationSnappingDistance`, we return the users actual location and course
* This allows for more wiggle room and less aggressive snapping and rerouting

bsudekum:
@frederoni @1ec5 after testing this for a bit, I think it's makes for a good addition. 👍 on merging?

PR #230, opened 2017-05-16
When swiping check if upcoming step

bsudekum:
We're checking for the wrong step when swiping steps since for maneuver information we're looking for the upcoming step.

### DIFF
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -228,17 +228,10 @@ open class RouteLegProgress: NSObject {
      Return bool whether step provided is the current `Step` the user is on.
     */
     public func isCurrentStep(_ step: RouteStep) -> Bool {
-        return leg.steps.index(of: step) == stepIndex
+        return step == currentStep
     }
+
     
-    /*
-     Return bool whether step provided is the upcoming `Step`.
-     */
-    public func isUpComingStep(_ step: RouteStep) -> Bool {
-        return step == upComingStep
-    }
-
-
     public var currentStepProgress: RouteStepProgress
 
 
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -230,6 +230,16 @@ open class RouteLegProgress: NSObject {
     public func isCurrentStep(_ step: RouteStep) -> Bool {
         return leg.steps.index(of: step) == stepIndex
     }
+    
+    /*
+     Return bool whether step provided is the upcoming `Step`.
+     */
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -230,6 +230,16 @@ open class RouteLegProgress: NSObject {
...
     }
+    
+    /*
+     Return bool whether step provided is the upcoming `Step`.
+     */
+    public func isUpComingStep(_ step: RouteStep) -> Bool {
+        guard 1 + stepIndex < leg.steps.count else {
+            return false
+        }
+        return leg.steps.index(of: step) == 1 + stepIndex
```
1ec5: `index(of:)` performs a linear search through the array, but I think this method and `isCurrentStep(_:)` are both unnecessary. We could get the same result by checking `step == leg.steps[stepIndex + 1]`. In fact, the call site could just say `step == routeController.routeProgress.currentLegProgress.currentStep`.

bsudekum:
@1ec5 fixed

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -230,6 +230,13 @@ open class RouteLegProgress: NSObject {
     public func isCurrentStep(_ step: RouteStep) -> Bool {
         return leg.steps.index(of: step) == stepIndex
     }
+    
+    /*
+     Return bool whether step provided is the upcoming `Step`.
+     */
+    public func isUpComingStep(_ step: RouteStep) -> Bool {
+        return step == upComingStep
```
1ec5: Is a method really needed to perform this equality comparison?

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -230,6 +230,13 @@ open class RouteLegProgress: NSObject {
     public func isCurrentStep(_ step: RouteStep) -> Bool {
         return leg.steps.index(of: step) == stepIndex
```
1ec5: While we’re at it, let’s compare `currentStep` to `step` here as well.

bsudekum:
@1ec5 fixed

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -228,10 +228,10 @@ open class RouteLegProgress: NSObject {
      Return bool whether step provided is the current `Step` the user is on.
     */
     public func isCurrentStep(_ step: RouteStep) -> Bool {
-        return leg.steps.index(of: step) == stepIndex
+        return step == currentStep
     }
 
-
+
```
frederoni: Nit: remove this line

PR #231, opened 2017-05-17
Abbreviation fix

### LINKED ISSUE #229: Abbreviating multiple lines
simulator screen shot may 16 2017 10 00 14 am

This should probably read:

> Alves Drive

or 

> Alves
> Drive

/cc @frederoni @1ec5

frederoni:
Closes #229 

- [x] Fix abbreviation
- [x] Move maneuver view logic out from `RouteMapViewController` to `RouteManeuverViewController`.
- [x] Remove expanding constraints from the distanceLabel.
- [x] Re-enable animation for overview button.
- [x] Add more abbreviation snapshot tests.

Also added `--configuration Debug` to the Carthage step on Bitrise in order to support [testable import](https://github.com/mapbox/mapbox-navigation-ios/pull/231/files#diff-c50b57e7fdbd91a299d8fb56754399f2R3) for MapboxDirections.

@bsudekum @1ec5 👀


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -202,48 +202,9 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
...
-                controller.streetLabel.numberOfLines = 2
-            }
-            
-            if let name = upComingStep.names?.first {
-                controller.streetLabel.text = name.abbreviated(toFit: streetLabelBounds, font: streetLabelFont)
-            } else if let destinations = upComingStep.destinations?.joined(separator: "\n") {
-                controller.streetLabel.text = destinations.abbreviated(toFit: streetLabelBounds, font: streetLabelFont)
-            } else {
-                controller.streetLabel.text = upComingStep.instructions.abbreviated(toFit: streetLabelBounds, font: streetLabelFont)
-                controller.streetLabel.text = routeStepFormatter.string(for: upComingStep)?.abbreviated(toFit: streetLabelBounds, font: streetLabelFont)
```
frederoni: @bsudekum We set the text twice here. Which one is the correct one?

bsudekum: Oh wow. The second one is.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -1,25 +1,75 @@
...
+            streetLabel.text = destinations.abbreviated(toFit: availableStreetLabelBounds, font: streetLabel.font)
+        } else if let step = step {
+            streetLabel.text = routeStepFormatter.string(for: step)?.abbreviated(toFit: availableStreetLabelBounds, font: streetLabel.font)
+        }
+    }
+    
     var distance: CLLocationDistance? {
         didSet {
             if let distance = distance {
+                distanceLabel.isHidden = false
```
1ec5: This is reasonable, but doesn't the label hide automatically if the text is nil?

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -32,6 +82,26 @@ class RouteManeuverViewController: UIViewController {
         }
     }
     
+    var availableStreetLabelBounds: CGRect {
+        return CGRect(origin: .zero, size: maximumAvailableStreetLabelSize)
```
1ec5: It isn't always easy to predict when the available bounds will change. For example, the screen orientation may change, or the shield image may load asynchronously. Could we make this abbreviation mechanism run later, when we know the width of the view?

* A subclass of UILabel has a property `unabridgedText` that we set to the full, unavbreviated name.
* AbridgedLabel has a `didSet` for `bounds` that sets `text` based on the current `bounds`.

MapboxNavigation/RouteManeuverViewController.swift
frederoni: Usually, but I think it also depends on if it's contained in a stack view and what the distribution is set to. The width will shrink to 0 when the text is nil but height will still be based on the font size.

MapboxNavigation/RouteManeuverViewController.swift
frederoni: AbridgedLabel could inherit from various style labels so that wouldn't be convenient.
Instead I added `Abridgeable` protocol that any label can conform to.

I wanted to use `class AbridgedLabel<T:StyleLabel> : Abridgeable { }` but generics can't be hooked up to an IBOutlet. 😔

MapboxNavigation/UILabel.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+protocol Abridgeable: class {
+    var unabridgedText: String? { get set }
+    typealias AvailableBoundsHandler = () -> (CGRect)
+    var availableBounds: AvailableBoundsHandler! { get set }
+}
+
+extension Abridgeable where Self: UILabel {
+    func set(unabridgedText: String?) {
```
1ec5: Could this be a getter on `unabridgedText` instead of a separate method? If it does have to be a separate method, `setUnabridgedText(_:)` would be more forwards-compatible than `set(unabridgedText:)`. For example, in the future, if we need to add another string-typed setter, Swift won’t let us have both `set(unabridgedText:)` and `set(voiceIdentifier:)` on the same class.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -107,6 +71,41 @@ class RouteManeuverViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         turnArrowView.backgroundColor = .clear
+        streetLabel.availableBounds = { CGRect(origin: .zero, size: self.maximumAvailableStreetLabelSize) }
```
1ec5: Does `self` need to be captured strongly or weakly here? Or is this a nonescaping closure?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -205,6 +205,15 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
         guard let controller = routePageViewController.currentManeuverPage else { return }
         
         controller.notifyDidChange(routeProgress: routeProgress, secondsRemaining: secondsRemaining)
+        updateShield(for: controller)
+        
+        // Move the overview button if the lane views become visible
+        if !controller.isPagingThroughStepList {
+            let initialPaddingForOverviewButton:CGFloat = controller.stackViewContainer.isHidden ? -30 : -20 + controller.laneViews.first!.frame.maxY
```
1ec5: It’s unfortunate that we need to do such manual layout…

MapboxNavigation/UILabel.swift
frederoni: That's a good improvement but I was hoping for an even better ;)
Overriding unabridgedText's didSet or observing changes and update super.text accordingly would be optimal but I couldn't figure out a way to achieve that in an extension, that's why I detoured through generics. However, `setUnabridgedText(_:)` will also do the job. 👍

MapboxNavigation/RouteManeuverViewController.swift
frederoni: Good catch

MapboxNavigation/RouteMapViewController.swift
frederoni: It's just because the lane view is out of bounds. We should be able to remove all this but it requires some tweaks in the storyboard. Let's do that in a separate PR.

MapboxNavigation/UILabel.swift
1ec5: Is StreetLabel the only class we foresee conforming to Abridgeable? If so, maybe we could sidestep these protocol complications by folding `unabridgedText` into StreetLabel. Then we could implement `didSet` on `bounds` to make sure `text` gets updated without ever having to trigger an update explicitly.

Otherwise, if other UILabel subclasses will need to be abridgeable, we could fold `unabridgedText` into StyleLabel, then use a `abridgesText` flag to determine whether the `bounds` `didSet` calls `abbreviated(toFit:)` or sets `text` to `unabridgedText` directly.

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -413,7 +407,7 @@
                                                 </userDefinedRuntimeAttributes>
                                             </view>
                                             <view hidden="YES" contentMode="center" translatesAutoresizingMaskIntoConstraints="NO" id="Q0R-gA-tRA" customClass="MBLaneArrowView">
-                                                <rect key="frame" x="240" y="0.0" width="30" height="30"/>
+                                                <rect key="frame" x="0.0" y="0.0" width="0.0" height="30"/>
```
bsudekum: I still don't know which is actually right here.....

frederoni: I didn't change anything here so it will probably change automatically again. It's not related to this PR but I'm going to revisit the layout of these lane views because their constraints break from time to time.

PR #232, opened 2017-05-17
Fix swiping with overview button

bsudekum:
Follow up to https://github.com/mapbox/mapbox-navigation-ios/pull/230 and https://github.com/mapbox/mapbox-navigation-ios/pull/223. The combination of the two of these created some more bugs when in master together.

Beyond fixing some current/next step bugs, this also prevent sliding while in overview mode.


frederoni:
@bsudekum the overview button moves down to the correct place first but then it moves back up under the lane view.
simulator screen shot may 18 2017 6 16 40 pm

bsudekum:
@frederoni okay, let me take a look into this

PR #239, opened 2017-05-22
Don't announce when merging onto highway

bsudekum:
When there is only a single option and it is to merge onto a highway, a `high` alert becomes redundant and is noise. The removes the high alert when the upcoming maneuver is merge and the current maneuver is `takeOnRamp`.

/cc @ericrwolfe

### DIFF
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -182,7 +182,7 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
         fallbackText = speechString(notification: notification, markUpWithSSML: false)
         
         // If the user is merging onto a highway, an announcement to merge is a bit excessive
-        if let upComingStep = routeProgress.currentLegProgress.upComingStep, (routeProgress.currentLegProgress.currentStep.maneuverType == .takeOnRamp && upComingStep.maneuverType == .merge && routeProgress.currentLegProgress.alertUserLevel == .high) {
+        if let upComingStep = routeProgress.currentLegProgress.upComingStep, routeProgress.currentLegProgress.currentStep.maneuverType == .takeOnRamp && upComingStep.maneuverType == .merge && routeProgress.currentLegProgress.alertUserLevel == .high {
             return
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -181,6 +181,11 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
         
         fallbackText = speechString(notification: notification, markUpWithSSML: false)
         
+        // If the user is merging onto a highway, an announcement to merge is a bit excessive
+        guard let upComingStep = routeProgress.currentLegProgress.upComingStep, !(routeProgress.currentLegProgress.currentStep.maneuverType == .takeOnRamp && upComingStep.maneuverType == .merge && routeProgress.currentLegProgress.alertUserLevel == .high) else {
```
1ec5: If there's no upcoming step, this code also bails. Is that intended?

1ec5:
> When there is only a single option

This PR doesn't check that there's only one possible maneuver. Does it need to?

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -182,7 +182,7 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
         fallbackText = speechString(notification: notification, markUpWithSSML: false)
         
         // If the user is merging onto a highway, an announcement to merge is a bit excessive
-        guard let upComingStep = routeProgress.currentLegProgress.upComingStep, !(routeProgress.currentLegProgress.currentStep.maneuverType == .takeOnRamp && upComingStep.maneuverType == .merge && routeProgress.currentLegProgress.alertUserLevel == .high) else {
+        if let upComingStep = routeProgress.currentLegProgress.upComingStep, (routeProgress.currentLegProgress.currentStep.maneuverType == .takeOnRamp && upComingStep.maneuverType == .merge && routeProgress.currentLegProgress.alertUserLevel == .high) {
```
1ec5: Nit: the parentheses are no longer needed here.

PR #243, opened 2017-05-23
Correctly draw uturns

bsudekum:
We're drawing uturns always turning to the right now. This changes to draw either to the left or right  depending on enter/exit bearings.

### DIFF
--- a/MapboxNavigation/TurnArrowView.swift
+++ b/MapboxNavigation/TurnArrowView.swift
@@ -93,7 +93,7 @@ public class TurnArrowView: UIView {
                 flip = true
             case .uTurn:
                 StyleKitArrows.drawArrow180(scale: scale)
-                flip = [.left, .slightLeft, .sharpLeft].contains(ManeuverDirection(angle: angle))
+                flip = angle < 0
             default:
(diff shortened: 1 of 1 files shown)


MapboxNavigation/TurnArrowView.swift
```diff
@@ -93,6 +93,7 @@ public class TurnArrowView: UIView {
                 flip = true
             case .uTurn:
                 StyleKitArrows.drawArrow180(scale: scale)
+                flip = [.left, .slightLeft, .sharpLeft].contains(ManeuverDirection(angle: angle))
```
1ec5: This method has access to the `step` property in the surrounding class. Check the RouteStep’s `maneuverDirection` instead of creating a new one.

1ec5: @bsudekum pointed out that `maneuverDirection` in this case is just `uTurn`, but we want to know the angle underlying that classification. So it’s appropriate to look at `angle`, but I think this is overkill: as long as `angle` is relative (with 0° meaning straight ahead) as opposed to absolute (0° meaning due north), we would flip the arrow if and only if the angle is negative.

bsudekum:
@1ec5 fixed

PR #244, opened 2017-05-24
Fix route simulation

frederoni:
The last coordinate in the given route was ignored.

screenshot 2017-05-24 20 30 44

@1ec5 @bsudekum


bsudekum:
You think it would possible to rework https://github.com/mapbox/mapbox-navigation-ios/blob/be49efd58504a38f77034429969ad774ff96c925/MapboxNavigation/SimulatedRoute.swift#L93 to include the first and last coordinate instead?

frederoni:
Good point. I think I can include the first coordinate but not the last.

PR #246, opened 2017-05-25
Measure British distances in imperial units

### LINKED ISSUE #242: Special-case GB locales to use imperial units
https://github.com/mapbox/mapbox-navigation-ios/issues/211#issuecomment-303495328 points out that 🇬🇧 uses the metric system officially but still measures road distances in miles (and yards). `Locale.usesMetricSystem` returns `true` when the locale’s region code is `GB`, so we need to special case `GB` to format distances with miles instead of kilometers. In this case, we may also want to disable the workaround in DistanceFormatter that [replaces yards with feet](https://github.com/mapbox/mapbox-navigation-ios/blob/7952bdc88244db90bdbe348ff9920919d2ca6d7f/MapboxNavigation/DistanceFormatter.swift#L41), assuming that Britons are more accustomed to yardages than Americans.

/cc @bsudekum @andrey-str

1ec5:
Force the number formatter underlying the distance formatter to use a U.S. locale (for its customary units) and avoid replacing yards with feet.

Fixes #242.

/cc @bsudekum @andrey-str

### DIFF
--- a/MapboxNavigation/DistanceFormatter.swift
+++ b/MapboxNavigation/DistanceFormatter.swift
@@ -36,7 +36,8 @@ public class DistanceFormatter: LengthFormatter {
         let miles = distance / metersPerMile
         let feet = miles * feetPerMile
         
-        let isBritish = Locale.current.regionCode == "GB"
+        // British roads are measured in miles, yards, and feet. Simulate this idiosyncrasy using the U.S. locale.
+        let isBritish = Locale.current.identifier == "en-GB"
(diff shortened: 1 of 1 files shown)


jeeftor:
How do i get rid of yards and have everything in feet?

frederoni:
@jeeftor What is your region (⚙️ iOS Settings -> Language & Region -> Region) set to when you see distances measured in yards?

jeeftor:
United States - English.  I opened a new issue:  #606

Issue #247, opened 2017-05-25
More lenient arrival zones


bsudekum:
Problem:  A user is navigating to the mall. They arrive at the mall and begin to park their car in the parking lot. However, since the user did not reach the destination marker (due to it being in the middle of the mall), we reroute them.

Ideas on how to fix:
* If the users uses the geocoder to find a destination, use it's bounding box if available to see if the user is inside the bounding box.
* Use run time styling to see if user is less then x meters away from destination and in inside a parking lot.
* Ask the user `Have you arrived at the mall?` before rerouting when near the destination.

/cc @mapbox/navigation

1ec5:
> If the users uses the geocoder to find a destination, use it's bounding box if available to see if the user is inside the bounding box.

Using a geofence of some sort would be wonderful, but a bounding box is the crudest kind of geofence. Consider trying to navigate to D.C., which is rotated 45°, so that its bounding box covers much more area than the district itself. (Of course, we’d probably only apply this heuristic to points of interest, not entire cities, but the same consideration applies to smaller features too.)

> Use run time styling to see if user is less then x meters away from destination and in inside a parking lot.

Maybe we should always avoid rerouting while inside a parking lot. This could improve scenarios where the user has to wend their way out of a parking lot to get onto the route line in first place.

bsudekum:
> Maybe we should always avoid rerouting while inside a parking lot. This could improve scenarios where the user has to wend their way out of a parking lot to get onto the route line in first place.

👍, this sounds like a great first move.

1ec5:
After toying around with #248, I think a better solution would be to continue to reroute while in a parking lot but silence the voice announcements. That addresses the annoyance of hearing instructions when you no longer want them as you find a parking spot. If we’re concerned about omitting meaningful instructions during departure, then we can use a heuristic that only announces the instruction if the first named way along the route has a different name than the old route.

ericrwolfe:
Ticketed out @1ec5's alternative solution to suppress voice announcements after reroutes: #374

1ec5:
#374 is distinct from what I proposed above, which was to suppress the voice announcements only while inside a parking lot. #382, for example, treads lightly by only suppressing the voice announcements when the resulting first step is over a kilometer long. But that isn’t necessarily the case inside a parking lot. I think #248 gets most of the way toward a solution to this problem, except that it suppresses rerouting entirely instead of only the voice announcements.

bsudekum:
These are good now.

PR #248, opened 2017-05-26
Don't reroute while in parking lot

bsudekum:
Ref: https://github.com/mapbox/mapbox-navigation-ios/issues/247

Before rerouting, we check if the user is inside of a parking lot. If so, we do not reroute.

I'm slightly worried about this one as it could produce some negative behavior. I wonder if we should limit this to only when the user is near the end of the route. 

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -293,6 +293,31 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
             return
         }
         
+        // If the user is in a parking lot, don't reroute
+        guard let mapView = mapView else { return }
+        guard let style = mapView.style else { return }
+        
+        let roadLabelLayerIdentifier = "parkingLot"
+        var streetsSources = style.sources.flatMap { $0 as? MGLVectorSource }.filter { $0.isMapboxStreets}
+        
+        if streetsSources.isEmpty {
+            let source = MGLVectorSource(identifier: "mapboxStreetsv7", configurationURL: URL(string: "mapbox://mapbox.mapbox-streets-v7")!)
+            style.addSource(source)
+            streetsSources.append(source)
+        }
+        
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -293,6 +293,31 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
             return
         }
         
+        // If the user is in a parking lot, don't reroute
+        guard let mapView = mapView else { return }
+        guard let style = mapView.style else { return }
+        
+        let roadLabelLayerIdentifier = "parkingLot"
```
1ec5: This variable’s name is misleading.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -293,6 +293,31 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
...
+        var streetsSources = style.sources.flatMap { $0 as? MGLVectorSource }.filter { $0.isMapboxStreets}
+        
+        if streetsSources.isEmpty {
+            let source = MGLVectorSource(identifier: "mapboxStreetsv7", configurationURL: URL(string: "mapbox://mapbox.mapbox-streets-v7")!)
+            style.addSource(source)
+            streetsSources.append(source)
```
1ec5: This logic is duplicated from the street label code. Would it be possible to factor it out into an extension method on MGLStyle?

bsudekum:
This is going to lead to too many edge cases. Closing for now, maybe will revisit.

1ec5:
Which edge cases are you thinking of?

bsudekum:
Going to take another look at this. Ideally, when someone is beginning to get closer to the edge of the parking lot polygon, we'd start to reroute them. The thinking here is that the user should have informative information whilst exiting the parking lot.

ericrwolfe:
An alternative solution to suppress voice announcements after reroutes: #374

bsudekum:
Should be fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/386

PR #254, opened 2017-05-31
Localize abbreviations

### LINKED ISSUE #220: Localize road name abbreviations in Transifex
#215 adds a plist lookup table of abbreviations that should be localized into the user’s language. It would be straightforward to import this file into Transifex; however, that would force a translator to specify abbreviations for each word, even if it isn’t normally abbreviated in their language. Moreover, they would be unable to add an abbreviation for a word that English does not abbreviate.

We can add a script that collapses each of the three subdictionaries (such as `directions`) into a single string solely for the purpose of importing into Transifex; it would also convert back to the original structure to be checked into this repository.

/cc @bsudekum @frederoni

frederoni:
Fixes #220 

Added a script that can export Abbreviations.plist to a strings file that we can translate in Transifex (not hooked up yet). The script can also import the translated strings back into the plist files.

**How to use the script**:
- run `make bootstrap`
- `./main.swift export Base,sv,ca`
- `./main.swift import Base,sv,ca`
@1ec5 👀


bsudekum:
I wonder if we can move any of this to https://github.com/Project-OSRM/osrm-text-instructions

frederoni:
> Does this script convert the plist into a strings file where each abbreviateable word is a separate translatable string?

The script did convert from plist to strings until https://github.com/mapbox/mapbox-navigation-ios/pull/254/commits/75372541cf207d49d8439bbe0b895afd7a9e28d6. Now it converts to a CSV-style dictionary and vice versa. 

I assume we have to check in both versions of `Abbreviations.plist`?

ingalls:
Just going to drop this here for those who haven't seen it as well: https://github.com/mapbox/geocoder-abbreviations

frederoni:
Thanks, @ingalls.
It would probably be more convenient to reuse geocoder-abbreviations. Though, it would require a few tweaks in `Abbreviations.swift` to add support for multiple steps of abbreviations.

@ingalls how do you manage these strings/tokens now? Can we hook it up to Transifex?
@1ec5 What do you think about using geocoder-abbreviations to export to plist-files?

1ec5:
As I mentioned in https://github.com/mapbox/mapbox-navigation-ios/issues/220#issuecomment-300931256, geocoder-abbreviations looks like a fantastic starting point, although it’s considerably more aggressive than we’d want to be in this library. For example, “North Beach Harbor Plaza Road” could turn into “N Bch Hbr Plz Rd” unless we’re careful to only abbreviate prefixes and suffixes.

@ingalls, is geocoder-abbreviations used only for fuzzy matching, or does it also influence a result’s display name? If the latter, how does the geocoder select among multiple abbreviations for the same word?

ingalls:
@1ec5 Yeah geocoder-abbreviations is only used for fuzzy matching, we don't alter display names at all and prefer fully expanded names. I don't necessarily know if it will be a great fit for your project but I would love PRs for any tokens you do add that we don't have as tokens can very significantly improve geocoding behavior and it can be really hard to get a complete list without local knowledge.

scripts/abbreviations/main.swift
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env xcrun swift -F ./Carthage/Build/Mac
```
1ec5: This is a pretty well-designed script. 👍 Do you think we should eventually port json2plist.sh to Swift for consistency? (I like json2plist.sh’s brevity, but I don’t know if that would be achievable for this abbreviation script in Bash – maybe with `PlistBuddy`? 🤷‍♂️)

scripts/abbreviations/LocalizeAbbreviations.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,272 @@
+// !$*UTF8*$!
```
1ec5: Could we fold this project into the main MapboxNavigation project? Keeping everything in one place could be more convenient for making local changes. (The SwiftCLI dependency would go in Cartfile.private to avoid interfering with normal developer workflows.)

Or is the idea to keep the abbreviation project separate so that we can easily split it into its own repository in the future?

scripts/abbreviations/main.swift
frederoni: json2plist is neat, DOTADIW and probably won't need any changes so I wouldn't port it just yet.
I'd prefer to keep the abbreviation script in Swift for the sake of readability. We can probably cut a few lines and simplify but adding a plist and csv dependency.

scripts/abbreviations/LocalizeAbbreviations.xcodeproj/project.pbxproj
frederoni: It's not really reusable as is so just adding a ref to the script in the MapboxNavigation project as you mention makes sense. However, SwiftCLI doesn't provide a project file and relies on SPM. We need to checkout, generate project and build with Carthage separately. It would not be convenient to force [this flow](https://github.com/mapbox/mapbox-navigation-ios/pull/254/commits/ad8e1c9d2238a95cd6824a68fd17bb89f4853490#diff-a51af85c97fc8578019ea23f6f3a0319R4) every time we use Carthage in this project.

frederoni:
Transifex doesn't work well with these plist-files. It only searches for the `<string>` tags and ignores the `<key>` tags. (https://www.transifex.com/mapbox/mapbox-navigation-ios/viewstrings/#en/localizableabbreviations/114199680)
Need to figure out another solution..
/cc @1ec5

1ec5:
That’s expected; the key is only displayed as the string’s context. (You have to go into translation mode to see the context.) I had a go at translating the abbreviations into Vietnamese last night, and it was 👌.

frederoni:
One more round 👀  here, especially https://github.com/mapbox/mapbox-navigation-ios/pull/254/commits/60b494523d002a7253f9f181923d9ae964f193df.

/cc @1ec5 @bsudekum

.tx/config
```diff
@@ -9,6 +9,12 @@ file_filter = MapboxNavigation/Resources/<lang>.lproj/Localizable.strings
 source_file = MapboxNavigation/Resources/Base.lproj/Localizable.strings
 source_lang = en
 
+[mapbox-navigation-ios.localizableabbreviations]
+file_filter = MapboxNavigation/Resources/<lang>.lproj/Abbreviations_csv.plist
+source_file = MapboxNavigation/Resources/Base.lproj/Abbreviations_csv.plist
+source_lang = en
+type = PLIST
```
1ec5: Should we override the minimum percentage for this resource to be 100% instead of 80%?

PR #255, opened 2017-05-31
Fix fluttering distance label

bsudekum:
In cases where a user is approaching a maneuver and their speed decreases, `secondsRemaining` could dip below this threshold then go above it. This effect makes the distance remaining on step disappear then show again.

Now when the user is basically at the maneuver will it disappear. The real goal here is to never show `0ft`.

### DIFF
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -85,7 +85,7 @@ class RouteManeuverViewController: UIViewController {
         let stepProgress = routeProgress.currentLegProgress.currentStepProgress
         let distanceRemaining = stepProgress.distanceRemaining
         
-        distance = secondsRemaining > 5 ? distanceRemaining : nil
+        distance = distanceRemaining > 10 ? distanceRemaining : nil
         
         if routeProgress.currentLegProgress.alertUserLevel == .arrive {
             distance = nil


MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -85,7 +85,7 @@ class RouteManeuverViewController: UIViewController {
         let stepProgress = routeProgress.currentLegProgress.currentStepProgress
         let distanceRemaining = stepProgress.distanceRemaining
         
-        distance = secondsRemaining > 5 ? distanceRemaining : nil
+        distance = distanceRemaining > 10 ? distanceRemaining : nil
```
1ec5: On a highway going 65 mph, this basically guarantees that we won’t hide the distance at all. Is that OK?

1ec5: Confirmed with @bsudekum that it’s OK. This logic exists mainly to avoid ever showing “0 ft”, but location updates are spaced far enough part that we wouldn’t see that in the scenario above.

1ec5: Is it OK that, depending on the locale, we’d be hiding at the 11-yard or 33-foot mark instead of a clean number like the 10-meter mark?

bsudekum: I think so, the goal here is really just to hide 0ft.

PR #258, opened 2017-05-31
Don't move alert level backwards

bsudekum:
simulator screen shot may 31 2017 2 52 57 pm

In simulation mode for the test above, when heading south on north de anza blvd, the alert level could flutter between high and medium. This caused many voice prompts. Really, there is no need to ever degrade an alert level from high to medium. This change makes it's so we do not move back to medium if the alert level is already high.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -236,12 +236,11 @@ extension RouteController: CLLocationManagerDelegate {
             if !isCloseToCurrentStep && isCloseToUpComingStep {
                 
                 // Increment the step
-                routeProgress.currentLegProgress.stepIndex += 1
                 
                 // and reset the alert level since we're on the next step
                 let userSnapToStepDistanceFromManeuver = distance(along: routeProgress.currentLegProgress.currentStep.coordinates!, from: location.coordinate)
                 let secondsToEndOfStep = userSnapToStepDistanceFromManeuver / location.speed
-                incrementRouteProgressAlertLevel(secondsToEndOfStep <= RouteControllerMediumAlertInterval ? .medium : .low, location: location)
+                incrementRouteProgress(secondsToEndOfStep <= RouteControllerMediumAlertInterval ? .medium : .low, location: location, updateStepIndex: true)
                 return true
             }
         }
@@ -249,7 +248,18 @@ extension RouteController: CLLocationManagerDelegate {
         return isCloseToCurrentStep
     }
     
-    func incrementRouteProgressAlertLevel(_ newlyCalculatedAlertLevel: AlertLevel, location: CLLocation) {
+    func incrementRouteProgress(_ newlyCalculatedAlertLevel: AlertLevel, location: CLLocation, updateStepIndex: Bool) {
+        
+        if updateStepIndex {
+            routeProgress.currentLegProgress.stepIndex += 1
+        }
+        
+        // If the step is not being updated, don't accept a lower alert level.
+        // A lower alert level can only occur when the user begins the next step.
+        if newlyCalculatedAlertLevel.rawValue < routeProgress.currentLegProgress.alertUserLevel.rawValue, !updateStepIndex {
+            return
+        }
+        
         if routeProgress.currentLegProgress.alertUserLevel != newlyCalculatedAlertLevel {
             routeProgress.currentLegProgress.alertUserLevel = newlyCalculatedAlertLevel
             // Use fresh user location distance to end of step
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -371,7 +371,8 @@ extension RouteController: CLLocationManagerDelegate {
             // Don't alert if the route segment is shorter than X
             // However, if it's the beginning of the route
             // There needs to be an alert
-            routeProgress.currentLegProgress.currentStep.distance > minimumDistanceForMediumAlert {
+            routeProgress.currentLegProgress.currentStep.distance > minimumDistanceForMediumAlert &&
+            alertLevel != .high {
```
1ec5: There are other cases where we’re setting the alert level, and if we were to address those cases equally, we’d quickly wind up with spaghetti. Fortunately, `incrementRouteProgressAlertLevel(_:location:)` offers us a handy bottleneck to accomplish the stated intent of this PR: simply bail if the new alert level is “backwards from” (i.e., greater than) the current alert level. It would be necessary to account for the special cases `none`, `depart`, and `arrive`, but overall I think that approach would be more self-documenting and less prone to edge cases.

bsudekum:
@1ec5 updated to make this a bit cleaner in https://github.com/mapbox/mapbox-navigation-ios/pull/258/commits/e151f036d952d2df14985b42b3bba217d9b889d9

MapboxCoreNavigation/RouteController.swift
```diff
@@ -236,20 +236,30 @@ extension RouteController: CLLocationManagerDelegate {
...
-    func incrementRouteProgressAlertLevel(_ newlyCalculatedAlertLevel: AlertLevel, location: CLLocation) {
+    func incrementRouteProgress(_ newlyCalculatedAlertLevel: AlertLevel, location: CLLocation, updateStepIndex: Bool) {
+        
+        if updateStepIndex {
+            routeProgress.currentLegProgress.stepIndex += 1
+        }
+        
+        // If the step is not being updated, don't accept a lower alert level.
+        // A lower alert level can only occur when the user begins the next step.
+        if newlyCalculatedAlertLevel.rawValue < routeProgress.currentLegProgress.alertUserLevel.rawValue, !updateStepIndex {
```
bsudekum: I don't think a guard will work here.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -236,20 +236,30 @@ extension RouteController: CLLocationManagerDelegate {
             if !isCloseToCurrentStep && isCloseToUpComingStep {
                 
                 // Increment the step
-                routeProgress.currentLegProgress.stepIndex += 1
```
1ec5: The next line depends on `routeProgress.currentLegProgress.currentStep`. Now it’ll be snapping the distance to a different maneuver than before.

MapboxCoreNavigation/RouteController.swift
1ec5: `if foo { return }` is equivalent to `guard !foo else { return }`. The only catch is that `if let foo = bar { return }` is the opposite of `guard let foo = bar else { return }`, so `if let foo = bar, baz { return }` isn’t equivalent to `guard let foo = bar, !baz else { return }`.

The `,` operator is equivalent to `&&` except with lower precedence than `let`: `if foo, bar` is equivalent to `if foo && bar`, but `if let foo = bar, baz` isn’t equivalent to `if let foo = bar && baz`.

If you want to convert this statement to a `guard` statement, first replace the `,` with `&&`, then apply [De Morgan’s](https://en.wikipedia.org/wiki/De_Morgan's_laws).

MapboxCoreNavigation/RouteController.swift
bsudekum: Whoa, this might actually be a fix.... investigating.

bsudekum: @1ec5 i wonder if this is the root cause of https://github.com/mapbox/mapbox-navigation-ios/pull/255

bsudekum: Actually on second thought, this code makes sense.

1ec5: So did this turn out to be a good change to make? We’re going to snap to the maneuver at the _start_ of the current step before incrementing the step. `userSnapToStepDistanceFromManeuver` could be as long as the step itself, and `secondsToEndOfStep` could be as long as the step’s duration.

bsudekum:
@1ec5 updated with feedback

MapboxCoreNavigation/RouteController.swift
```diff
@@ -234,22 +234,28 @@ extension RouteController: CLLocationManagerDelegate {
         if let upComingStep = routeProgress.currentLegProgress.upComingStep {
             let isCloseToUpComingStep = newLocation.isWithin(radius, of: upComingStep)
             if !isCloseToCurrentStep && isCloseToUpComingStep {
-                
-                // Increment the step
-                routeProgress.currentLegProgress.stepIndex += 1
-                
-                // and reset the alert level since we're on the next step
                 let userSnapToStepDistanceFromManeuver = distance(along: routeProgress.currentLegProgress.currentStep.coordinates!, from: location.coordinate)
```
1ec5: I don’t think it’s correct to calculate this distance from the beginning of the current step, which could be many miles behind the user at this point. It could well be correct to increment `stepIndex` after this point instead of before it, but then this line should use `upcomingStep` instead of `currentStep`.

bsudekum: Great catch

bsudekum:
@1ec5 @frederoni this is ready for review again

PR #266, opened 2017-06-07
Fix maneuver swiping

bsudekum:
Swiping the maneuver list suffers from two issues:

* The current step shows up twice
* Swiping backwards can jump from x index all the way to the current step

@frederoni I'm not really certain why this code was here or why it fixed both the issues above, but it does ¯\_(ツ)_/¯.

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -25,6 +25,7 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
     let MBSecondsBeforeResetTrackingMode: TimeInterval = 25.0
 
     var route: Route { return routeController.routeProgress.route }
+    var previousStep: RouteStep?
 
     var destination: MGLAnnotation!
     var pendingCamera: MGLMapCamera? {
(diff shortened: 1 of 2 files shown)


frederoni:
Will this [guard](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/RouteMapViewController.swift#L207) always trigger so that we never update progress or distance on the current maneuver page?

bsudekum:
@frederoni im seeing it update just fine.

frederoni:
Oh, if that's the case we should remove all refs to currentManeuverPage since it doesn't fill its purpose. Replace the [guard](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/RouteMapViewController.swift#L207) with
```swift
guard let controller = routePageViewController.viewControllers?.first as? RouteManeuverViewController else { return }
```

bsudekum:
@frederoni adding this code introduces new errors. Now I'm seeing the current maneuver show up during swiping.

frederoni:
Yeah, we need to keep track of the currentManeuverPage or change the guard not to update unless the current page.step == currentStep besides guarding controller which should always fall through.

Something along these lines:
```swift
guard let controller = routePageViewController.viewControllers?.first as? RouteManeuverViewController else { return }
guard controller.step == currentStep else { return }
```

should act as a replacement for the `currentManeuverPage` tracking.

bsudekum:
@frederoni why are we looking at the first in `routePageViewController.viewControllers?.first `? This seems to be the crux of the problem in my opinion.

frederoni:
Unlike UITableView, a UIPageViewController doesn't have a concept of an index when implementing the data source. It would if we just set UIPageViewController.viewControllers property with all steps:
```swift
setViewControllers([step1VC, step2VC, ...])
// viewControllers.first is always step1VC
```
but when implementing the data source delegate methods:

```swift
pageViewController(_:viewControllerAfter:)
pageViewController(_:viewControllerBefore:)
// viewControllers.first is the current page
```
viewControllers.first is the current page (except during a transition).

frederoni:
In addition to fixing:
> The current step shows up twice
Swiping backwards can jump from x index all the way to the current step

this change also flips page automatically when a step has been completed.
Say you are viewing page 10/step 10 while you are finishing a maneuver for step 5, the page will flip to step 6 immediately without animation.

PR #268, opened 2017-06-08
Allow custom (destination) annotations

bsudekum:
This addition allows the developer to provide a custom annotation image for all annotations on the map. Note, the only annotations on the map currently is the destination marker.

/cc @frederoni

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -94,6 +94,8 @@ public protocol NavigationViewControllerDelegate {
      If this method is unimplemented, the navigation map view represents the route line’s casing using an `MGLPolylineFeature` identical to the one returned by `navigationMapView(_:shapeDescribing:)`.
      */
     @objc optional func navigationMapView(_ mapView: NavigationMapView, simplifiedShapeDescribing route: Route) -> MGLShape?
+    
+    @objc optional func navigationMapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?
 }
 
 /**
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -94,6 +94,13 @@ public protocol NavigationViewControllerDelegate {
      If this method is unimplemented, the navigation map view represents the route line’s casing using an `MGLPolylineFeature` identical to the one returned by `navigationMapView(_:shapeDescribing:)`.
      */
     @objc optional func navigationMapView(_ mapView: NavigationMapView, simplifiedShapeDescribing route: Route) -> MGLShape?
+    
+    /**
+     Return an `MGLAnnotationImage` that represents the destination marker.
+     
+     If this method is unimplemented, the navigation map view will represent the destination annotation with the default marker.
+     */
+    @objc optional func navigationMapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?
```
bsudekum: Should this instead be:

```swift
@objc optional func navigationMapView(_ mapView: NavigationMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?
```

frederoni:
Image annotations is a good start but we should also expose `mapView(_:viewFor:)` for full customizability, but that could be tail work. It would also require the user to provide a user puck.

MapboxNavigation/NavigationViewController.swift
frederoni: Let's stick with MGLMapView for now. Makes it easier to find the right documentation.

PR #186, opened 2017-04-28
Update table view design

frederoni:
Outlined the structure of the new table view

- [x] Hook up voice, satellite & traffic.
- [ ] Add steps

@bsudekum @1ec5 👀

PR #269, opened 2017-06-08
Add reroute UI and optional reroute sound

bsudekum:
This adds:

* A non optional rerouting UI which only displays while the routing request is in flight
* An optional little ding that plays when the user is rerouted and prior to the new first instruction

simulator screen shot jun 8 2017 12 37 48 pm

/cc @1ec5 @frederoni


bsudekum:
Unsure how to make the ding public to any project. I had to manually add it to the example however developers won't be able to check this box (I think?).

image

frederoni:
The audio file should only be in the MapboxNavigation framework. When you access it from the outside, you'd use Bundle.NavigationUI. I think it would make sense to put this audio in the xcassets catalog. Otherwise, we have to make changes to the podspec and App slicing may not work correctly.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -83,6 +84,19 @@ class RouteManeuverViewController: UIViewController {
         super.viewDidLoad()
         turnArrowView.backgroundColor = .clear
         streetLabel.availableBounds = {[weak self] in CGRect(origin: .zero, size: self != nil ? self!.maximumAvailableStreetLabelSize : .zero) }
+        resumeNotifications()
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        suspendNotifications()
```
frederoni: Nit: call `super.viewWillDisappear(_:)`. See [viewWillDisappear:](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621485-viewwilldisappear)

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -28,10 +28,15 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
+            guard let audioPlayer = audioPlayer else {
+                return 1
+            }
             return audioPlayer.volume
         }
         set {
-            audioPlayer.volume = newValue
+            if let audioPlayer = audioPlayer {
+                audioPlayer.volume = newValue
+            }
```
bsudekum: @frederoni have any ideas how to handle volume here? If `AVAudioPlayer ` is not inited yet, you cannot change the volume.

frederoni: I'm making some changes to the volume settings in #186 
I think we should set the volume in UserDefault.
AVAudioPlayer should be optional so when we set the settings:
```swift
settings.volume = newValue
audioPlayer?.volume = newValue
```
 and when AVAudioPlayer is initialized, we grab the volume from the settings.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -61,6 +54,24 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
     /**
+     If true, a noise indicating the user is going to be rerouted will play prior to rerouting.
+     */
+    public var playRerouteSound = false
+
+    
+    /**
+     Sound to play prior to reroute. Inherits volume level from `volume`.
+     */
+    public var rerouteSound = NSDataAsset(name: "reroute-sound", bundle: Bundle.navigationUI)
```
bsudekum: @frederoni how do you feel about the type here? Is there a way we could make this API easier to use?

1ec5: A developer who wants to customize this sound would add the sound to their asset catalog and refer to it by `NSDataAsset(name: "oh", bundle: .main)`. That seems pretty reasonable, but it does require using an asset catalog. In other words, it precludes the traditional workflow of dragging an audio file into the project (thereby adding it to the resource bundle directly) and creating some other type from a path. NSDataAsset seems more robust, in any case.

bsudekum: @1ec5 I think it's the only way we can have provide an mp3 in the framework that can be accessed in apps that import Navigation. Do you think we should use another api?

frederoni: There's a data property on `NSDataAsset` and `AVAudioPlayer(data:)` initializer. That could work? `rerouteSound` would be of type `Data`.

bsudekum: Changed it to bring your own audioplayer https://github.com/mapbox/mapbox-navigation-ios/pull/269/commits/6a636275305e6f70dfe92246dce4c81c4dce92ad

bsudekum:
Updated sound with https://www.freesound.org/people/pan14/sounds/263133/ which has a better [attribution](https://creativecommons.org/publicdomain/zero/1.0/)

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -277,20 +277,48 @@
                                             <constraint firstItem="c3B-6D-Ukh" firstAttribute="top" secondItem="dhM-Ng-gx5" secondAttribute="top" id="r0L-mL-w9Y"/>
                                         </constraints>
                                     </stackView>
+                                    <view hidden="YES" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="c9E-8M-S3k" userLabel="Reroute View">
+                                        <rect key="frame" x="0.0" y="0.0" width="343" height="115"/>
+                                        <subviews>
+                                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Rerouting" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="bUX-dR-1V6">
```
frederoni: This label should have a custom class set. Probably `MBTitleLabel`.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -61,6 +54,18 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
     /**
+     If true, a noise indicating the user is going to be rerouted will play prior to rerouting.
+     */
+    public var playRerouteSound = false
+
+    
+    /**
+     Sound to play prior to reroute. Inherits volume level from `volume`.
+     */
+    public var rerouteSoundPlayer: AVAudioPlayer = try! AVAudioPlayer(data: NSDataAsset(name: "reroute-sound", bundle: Bundle.navigationUI)!.data, fileTypeHint: AVFileTypeMPEGLayer3)
```
frederoni: I'm fine with exposing an audio player here but curious what the reasoning behind not exposing sound data is? I assume it would be the fileTypeHint.

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
frederoni: This label should have a custom class set. Probably MBTitleLabel.

MapboxNavigation/RouteVoiceController.swift
bsudekum: Per conversation, with an `AVAudioPlayer`, we can have a single option that specifies the audio type and data.

Issue #213, opened 2017-05-09
Can't change user's current location annotation view


Jauzee:
Is it possible to change user's location tracking annotation view (circle with arrow shape inside)?

1ec5:
RouteMapViewController is the NavigationMapView’s delegate (conforming to MGLMapViewDelegate). Since a map view can’t have multiple delegates, we need to plumb some additional MGLMapViewDelegate methods through RouteMapViewControllerDelegate, NavigationViewController, and NavigationViewControllerDelegate.

In particular, `MGLMapViewDelegate.mapView(_:viewFor:)` needs to be exposed as `NavigationViewControllerDelegate.navigationViewController(_:mapView:viewFor:)` so that the developer can implement their annotation view to replace the default. (We have an example of that implementation coming in mapbox/ios-sdk-examples#50.)

Jauzee:
@1ec5 will it also support changing user location annotation in `navigation` mode?

Jauzee:
Can you please guide me where in the code default shape for this annotation was set. Is this part of `mapbox-navigation-ios` or ir is a part of `MapboxDirections`?

1ec5:
> will it also support changing user location annotation in navigation mode?

Yes.

> Can you please guide me where in the code the default shape for this annotation was set.

The navigation SDK is currently using the default annotation image provided by the iOS map SDK (mapbox-gl-native).

1ec5:
#498 would expose the delegate method for customizing the user location annotation view.

bsudekum:
This has been implemented.

Issue #204, opened 2017-05-03
Route ends after each Waypoint


JustasL:
Not sure if this is more of Mapbox Directions or Mapbox Navigation question/problem

I would like to navigation over route with multiple specified points (up to 25), so that the path would be specific to my needs.
The problem is, that Mapbox Navigation thinks that every coordinate specified in `RouteOptions` for `Directions.shared.calculate` is a destination, while in reality I destination should be the last point.
Now, the Mapbox Navigation UI seems to understand, that my destination is final coordinate, as it continues to show route to that coordinate + arrival time is correct. What is incorrect, is that after first point it says, that I have arrived to destination and is stuck.

![image uploaded from ios](https://cloud.githubusercontent.com/assets/14614780/25658842/96078908-300d-11e7-98f0-ddd8851d4454.jpg)

So what should I do, that Navigation would not show 

> You have arrived at your destination

Until user actually arrives at its destination?

bsudekum:
Currently multiple waypoint navigation is not supported.

JustasL:
By not supported you mean, I can create a `RouteOptions` with multiple coordinates (up to 25), but  MapboxNavigation will not be able to provide proper navigation, right?

Then follow up question, will such feature be implemented in the near future?

1ec5:
Also tracking on Android in mapbox/mapbox-navigation-android#27.

bsudekum:
> By not supported you mean, I can create a RouteOptions with multiple coordinates (up to 25), but MapboxNavigation will not be able to provide proper navigation, right?

The Mapbox Directions API and MapboxDirections.swift honor waypoints however, the waypoints are not supported in this library yet.

> Then follow up question, will such feature be implemented in the near future?

Currently, this is not on our roadmap.

1ec5:
This feature is under development in #270.

ericrwolfe:
Basic multiple waypoint support was merged into master with https://github.com/mapbox/mapbox-navigation-ios/pull/270 and will be part of the next release.

Issue #114, opened 2017-04-05
Destination pin should match style


1ec5:
We’re currently using the map SDK’s default annotation icon for the destination pin. The red has good contrast with the default navigation style, but maybe we could be using a different color and different shape that more closely matches the style.

Down the line, we might also want to make it easier for the developer to customize the annotation to match the annotation they’re using to represent the destination on a preview map or some other map. But matching the style would be a good start.

/cc @aparlato @pveugen @bsudekum

aparlato:
@1ec5 does this relate to https://github.com/mapbox/cartography/issues/109 (it seemed the workflow described in this ticket was ruled out)? 
Or, if not, what kind of workflow do you see here for this? Is it helpful for carto to provide a destination pin as an SVG that goes with our navigation styles? Does this pin apply to other maps developers are using with the sdk?

1ec5:
> Is it helpful for carto to provide a destination pin as an SVG that goes with our navigation styles?

An SVG would be fine. We’d convert it into a PDF or PNGs, but an SVG would be more maintainable, so we’d check in both.

> Does this pin apply to other maps developers are using with the sdk?

Do you mean something like the main map or preview map on a navigation-focused application? I’d imagine the developer would want to be fully in control of those maps, including specifying a style other than the navigation style.

For now, we should make sure the destination pin looks good on the navigation style, even if it differs from the pin used on other maps in the application. Eventually, we could expose an option for the developer to take the pin from another map and reuse it in the navigation map.

aparlato:
> Do you mean something like the main map or preview map on a navigation-focused application? I’d imagine the developer would want to be fully in control of those maps, including specifying a style other than the navigation style.

I mean in terms of specifying a style other than the navigation style, as in would this pin be applied to a navigation style but also potentially, say, Streets if the developer wanted to use that?

> For now, we should make sure the destination pin looks good on the navigation style

Ok cool, this makes sense :+1:

aparlato:
cc @mayagao it looked like you already had a pin designed/were designing a pin for the SDK for this style?

PR #270, opened 2017-06-08
Add waypoint handling

bsudekum:
To do:

- [x] Add waypoint handling to example app
- [x] Improve instructions upon approaching leg destination
- ~~[x] Add UI elements for arriving and leaving waypoint~~

/cc @1ec5 @frederoni @cammace @ericrwolfe


MapboxNavigation/RouteTableViewController.swift
```diff
@@ -72,14 +72,14 @@ extension RouteTableViewController: UITableViewDelegate, UITableViewDataSource {
     }
     
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        return routeController.routeProgress.currentLeg.steps.count
+        return routeController.routeProgress.allSteps.count
```
1ec5: Instead of flattening the list of steps, we should have multiple sections (instead of just one), each section listing the steps for a single leg.

bsudekum:
Some notes, 

* I deprecated `destination`. It was never really needed and with waypoints, things get a little more confusing if there is a destination option. 
* `didArriveAt` changed to now return the route step since there really is not a destination annotation anymore. It also will only return when the user arrives at the end of the route. However, we could change this to return at ever waypoint.
* In terms of how this will fit into existing code bases, I think it would just work. There are improvements we could make to how waypoints are handled (adding a UI namely), but as a first step, I think this is good and would unblock any users looking for waypoint support.

/cc @1ec5 @frederoni

MapboxCoreNavigation/RouteController.swift
```diff
@@ -257,7 +258,17 @@ extension RouteController: CLLocationManagerDelegate {
...
         
         if routeProgress.currentLegProgress.alertUserLevel != newlyCalculatedAlertLevel {
+            
             routeProgress.currentLegProgress.alertUserLevel = newlyCalculatedAlertLevel
+            
+            if routeProgress.currentLegProgress.alertUserLevel == .arrive,
+                routeProgress.remainingWaypoints.count > 1 {
+                routeProgress.legIndex += 1
+                routeProgress.currentLegProgress.stepIndex = 0
+                routeProgress.currentLegProgress.alertUserLevel = .depart
```
1ec5: Changing `legIndex` automatically [resets](https://github.com/mapbox/mapbox-navigation-ios/blob/5d78f52e52d2e1e847336890a44bfaae6505f653/MapboxCoreNavigation/RouteProgress.swift#L135) `currentLegProgress`.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -124,8 +124,10 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
      on the destination of your route. The last coordinate of the route will be
      used if no destination is given.
     */
+    @available(*, deprecated, message: "Destination is no longer support nor necessary. A destination annotation will automatically be added to map given the route.")
```
1ec5: Note that this shouldn’t have much of an impact on most developers: it isn’t possible to customize this annotation (#213) other than changing its title, subtitle, and coordinates, and none of those properties are likely to be modified by the developer anyways.

1ec5: s/support/supported/

MapboxNavigation/NavigationViewController.swift
```diff
@@ -368,10 +369,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
             }
         }
         
-        if destination == nil {
-            let annotation = MGLPointAnnotation()
-            annotation.coordinate = route.coordinates!.last!
-            destination = annotation
+        for leg in route.legs {
+            if let last = leg.steps.last {
```
1ec5: This code only places an annotation at the last destination. Should we annotate the intermediate waypoints as well?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -368,10 +369,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
             }
         }
         
-        if destination == nil {
-            let annotation = MGLPointAnnotation()
-            annotation.coordinate = route.coordinates!.last!
-            destination = annotation
+        for leg in route.legs {
```
1ec5: If we’re only looking at the last step, there’s no need to iterate over all the steps:

```swift
if let lastStep = route.legs.last?.steps.last {
```

MapboxNavigation/NavigationViewController.swift
```diff
@@ -368,10 +369,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
...
+            if let last = leg.steps.last {
+                let annotation = MGLPointAnnotation()
+                annotation.coordinate = last.coordinates!.last!
```
1ec5: The step’s `destination` is a Waypoint whose `coordinates` property corresponds to the location we’d want to annotate.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -20,7 +20,7 @@ public protocol NavigationViewControllerDelegate {
     /**
      Called when the user arrives at the destination.
      */
-    @objc optional func navigationViewController(_ navigationViewController : NavigationViewController, didArriveAt destination: MGLAnnotation)
+    @objc optional func navigationViewController(_ navigationViewController : NavigationViewController, didArriveAt destination: RouteStep)
```
1ec5: RouteStep isn’t a destination. Arriving at a step means something different than arriving at a destination. I agree that MGLAnnotation is a poor fit for this method, but then we should pass in a Waypoint, not a RouteStep.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -319,9 +319,10 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
 
         mapViewController?.notifyDidChange(routeProgress: routeProgress, location: location, secondsRemaining: secondsRemaining)
         tableViewController?.notifyDidChange(routeProgress: routeProgress)
-        
-        if routeProgress.currentLegProgress.alertUserLevel == .arrive {
-            navigationDelegate?.navigationViewController?(self, didArriveAt: destination)
+
+        if routeProgress.currentLegProgress.alertUserLevel == .arrive,
+            routeProgress.remainingWaypoints.count == 0 {
```
1ec5: Should the developer be notified when the user reaches an intermediate waypoint? I could see use cases for only responding to the final arrival and for responding to intermediate arrivals as well. Perhaps the delegate method should have an additional argument, `numberOfWaypointsRemaining`, to allow the developer to distinguish between the two cases.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,25 @@ class RouteTableViewController: UIViewController {
 extension RouteTableViewController: UITableViewDelegate, UITableViewDataSource {
 
     func numberOfSections(in tableView: UITableView) -> Int {
-        return 1
+        return routeController.routeProgress.route.legs.count
+    }
+    
+    func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
+        return "Leg \(section + 1)"
```
1ec5: The leg number is unimportant to the user. Instead, we should title the section with either of the following formats, preferably the first one:

* “Biloxi to Mobile” (see `leg.source`, `leg.destination`)
* “I-10” (see `leg.name`)

1ec5: If there is only one leg, there’s no need for a section header.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,25 @@ class RouteTableViewController: UIViewController {
...
     }
     
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cell = tableView.dequeueReusableCell(withIdentifier: RouteTableViewCellIdentifier, for: indexPath) as! RouteTableViewCell
-        let leg = routeController.routeProgress.currentLeg
+        let legs = routeController.routeProgress.route.legs
+        
         
-        cell.step = leg.steps[indexPath.row]
+         cell.step = legs[indexPath.section].steps[indexPath.row]
```
1ec5: Nit: funky indentation.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,25 @@ class RouteTableViewController: UIViewController {
...
         
-        if routeController.routeProgress.currentLegProgress.stepIndex + 1 > indexPath.row {
+        if  routeController.routeProgress.legIndex + 1 > indexPath.section && routeController.routeProgress.currentLegProgress.stepIndex + 1 > indexPath.row  {
```
1ec5: If the user is on step 5 of leg 1, only the first five steps of section 0 are dimmed out.

MapboxNavigation/RouteTableViewController.swift
1ec5: Intermediate waypoints returned by the Directions API generally lack names. In that case, the second format is probably the best we can do, unless we’re OK with adding a dependency on the Geocoding API.

1ec5: Blocked by mapbox/MapboxDirections.swift#135.

MapboxNavigation/NavigationViewController.swift
bsudekum: @1ec5 it was actually added in https://github.com/mapbox/mapbox-navigation-ios/pull/268

MapboxNavigation/NavigationViewController.swift
bsudekum: @1ec5 I was thinking about this but wanted to keep the API the same. I think if we want to notify them when they reach a waypoint, we should either rename the method or add a new method for waypoints.

MapboxNavigation/NavigationViewController.swift
bsudekum: @1ec5 hrm, no this annotates the last step of each leg. Or at least it's supposed to.

MapboxNavigation/RouteTableViewController.swift
bsudekum: @1ec5 I think that's how it should work though. Are you saying the current step should be dimmed?

MapboxNavigation/NavigationViewController.swift
1ec5: Ah, I missed that we’re exposing `navigationMapView(_:imageFor:)`. In any case, that method will continue to work; it just won’t be possible to say something like `if annotation == navigationViewController.destination`.

MapboxNavigation/NavigationViewController.swift
1ec5: The method’s current name implies that it’s called whenever the user reaches a waypoint once we added support for multiple waypoints. Why else would it take a second argument?

How about we add a new method, `navigationViewControllerDidArrive(_:)`, that gets called after `navigationViewController(_:didArriveAt:)` when arriving at the final waypoint?

MapboxNavigation/NavigationViewController.swift
1ec5: You’re right, I misread this line.

MapboxNavigation/RouteTableViewController.swift
1ec5: If leg 0 has 10 steps, leg 1 has 10 steps, and the user is on step 5 of leg 2 which has 10 steps, why would all three steps be halfway dimmed?

1ec5: Actually, this code is incorrect but in a different way than I first thought. This code is equivalent to: `if currentLegIndex >= section && currentLegProgress.stepIndex >= row { dim the cell }`. Now if the user is on step 5 of leg 2, the only dimmed cells are those that meet _both_ the following criteria:

1. the section number is at least 2; and
2. the row number is at least 5

In other words:

* Sections 0 and 1 are completely lit, because they fail criteria (1). 🙅‍♂️
* Section 2 is dimmed up to row 4 and lit beginning with row 5. 👍
* Sections 3, 4, etc. are each dimmed up to row 4 but lit beginning with row 5, because rows 5 and beyond fail criteria (2). 🙅‍♂️

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -72,7 +72,17 @@ extension RouteTableViewController: UITableViewDelegate, UITableViewDataSource {
     }
     
     func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
-        return "Leg \(section + 1)"
+        // Don't display section header if there is only one step
+        guard routeController.routeProgress.route.legs.count > 1 else {
+            return nil
+        }
+        
+        let legToFrom = routeController.routeProgress.route.legs[section].description.components(separatedBy: ", ")
```
1ec5: `RouteLeg.description` is equivalent to `RouteLeg.name`. `name` lists the main streets along the leg, not the leg’s origin and destination. For example, `name` might be “Market Street, I-10, Main Street”. If you want “Biloxi” and “Mobile”, you need to look at the `name` properties of RouteLeg’s `source` and `destination` properties, which won’t be very useful until mapbox/MapboxDirections.swift#135 is fixed.

MapboxNavigation/NavigationViewController.swift
bsudekum: Changed it to a single method that returns the waypoint the user is arriving at for each and every waypoint.

bsudekum:
This is ready for another round of reviews

MapboxNavigation/NavigationViewController.swift
```diff
@@ -20,7 +20,7 @@ public protocol NavigationViewControllerDelegate {
     /**
      Called when the user arrives at the destination.
      */
-    @objc optional func navigationViewController(_ navigationViewController : NavigationViewController, didArriveAt destination: MGLAnnotation)
+    @objc optional func navigationViewController(_ navigationViewController : NavigationViewController, didArriveAt destination: Waypoint)
```
1ec5: Should we also add `navigationViewControllerDidArrive(_:)` (with only one argument of type NavigationViewController)?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -332,6 +328,10 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         mapViewController?.notifyAlertLevelDidChange(routeProgress: routeProgress)
         tableViewController?.notifyAlertLevelDidChange()
         
+        if routeProgress.currentLegProgress.alertUserLevel == .arrive {
+            navigationDelegate?.navigationViewController?(self, didArriveAt: routeProgress.route.routeOptions.waypoints[routeProgress.legIndex + 1])
```
1ec5: We should pass in the current route leg’s destination waypoint. Otherwise, if we pass in one of the waypoints used to request the route, its location won’t necessarily correspond to what the Directions API thinks the destination should be (after snapping or whatever).

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -76,7 +75,14 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
         super.viewWillAppear(animated)
 
         mapView.compassView.isHidden = true
-        mapView.addAnnotation(destination)
+        
+        for leg in route.legs {
+            if let last = leg.steps.last {
+                let annotation = MGLPointAnnotation()
+                annotation.coordinate = last.coordinates!.last!
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r121788876

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,37 @@ class RouteTableViewController: UIViewController {
...
+    }
+    
+    func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
+        // Don't display section header if there is only one step
+        guard routeController.routeProgress.route.legs.count > 1 else {
+            return nil
+        }
+        
+        let sectionName = routeController.routeProgress.route.legs[section].name
+        let legToFrom = sectionName.components(separatedBy: ", ")
```
1ec5: The suggestion in https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r121870552 was to use `source` and `destination`, not `name`.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,37 @@ class RouteTableViewController: UIViewController {
...
         let cell = tableView.dequeueReusableCell(withIdentifier: RouteTableViewCellIdentifier, for: indexPath) as! RouteTableViewCell
-        let leg = routeController.routeProgress.currentLeg
+        let legs = routeController.routeProgress.route.legs
+        
         
-        cell.step = leg.steps[indexPath.row]
+        cell.step = legs[indexPath.section].steps[indexPath.row]
         
-        if routeController.routeProgress.currentLegProgress.stepIndex + 1 > indexPath.row {
+        if routeController.routeProgress.legIndex + 1 > indexPath.section && routeController.routeProgress.currentLegProgress.stepIndex + 1 > indexPath.row  {
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r121869861

MapboxNavigation/NavigationViewController.swift
bsudekum: I think one method is good enough here.

MapboxNavigation/RouteTableViewController.swift
bsudekum: I added using source and destination but I also left this code in place for cases where we do not have source and destination.

MapboxNavigation/RouteTableViewController.swift
bsudekum: @1ec5 having a hard time getting something to work here. Have any ideas how to accomplish this?

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -76,14 +76,20 @@ extension RouteTableViewController: UITableViewDelegate, UITableViewDataSource {
...
+        let sourceName = leg.source.name
+        let destinationName = leg.destination.name
         
+        if let sourceName = sourceName?.nonEmptyString, let destinationName = destinationName?.nonEmptyString {
+            return "\(sourceName) to \(destinationName)"
+        }
+        
+        let legToFrom = leg.name.components(separatedBy: ", ")
         if legToFrom.count == 2 {
             return "\(legToFrom[0]) to \(legToFrom[1])"
```
1ec5: Since these are two road names, “Main Street **and** Market Street” would be clearer than “Main Street **to** Market Street”.

MapboxNavigation/RouteTableViewController.swift
1ec5: We need to dim a cell for either of two reasons:

1. Either it represents a step that’s part of a leg that the user has already completed,
1. Or it represents a completed step in the current leg.

So the correct logic would be:

```swift
if indexPath.section < routeController.routeProgress.legIndex
    || (indexPath.section == routeProgress.legIndex
        && indexPath.row <= routeProgress.currentLegProgress.stepIndex) {
    // dim the cell
}
```

1ec5:
Blocked by Project-OSRM/osrm-text-instructions.swift#24.

Examples/Swift/ViewController.swift
```diff
@@ -51,61 +55,40 @@ class ViewController: UIViewController, MGLMapViewDelegate {
...
     }
     
-    @IBAction func didTapStartNavigation(_ sender: Any) {
-        startNavigation(along: userRoute!)
-    }
-    
-    @IBAction func didTapClearmap(_ sender: Any) {
-        mapView.removeAnnotations(mapView.annotations ?? [])
+    @IBAction func replay(_ sender: Any) {
+        let bundle = Bundle(for: ViewController.self)
```
1ec5: This class is inside the application target, so you can say simply `Bundle.main`. It’s only necessary to use `Bundle(for:)` inside a framework, since the framework isn’t the main bundle.

Examples/Swift/ViewController.swift
```diff
@@ -239,39 +218,38 @@ class ViewController: UIViewController, MGLMapViewDelegate {
...
         
         navigationViewController.present(confirmationController, animated: true, completion: nil)
     }
-    
-    func navigationViewController(_ navigationViewController: NavigationViewController, didArriveAt waypoint: Waypoint) {
-        print("User arrived at \(waypoint.name ?? "waypoint")")
+}
+
+extension ViewController: WaypointConfirmationViewControllerDelegate {
+    func confirmationControllerDidConfirm(controller confirmationController: WaypointConfirmationViewController) {
```
1ec5: The first part of this selector, `confirmationControllerDidConfirm`, already contains “confirmation controller”, so `controller` can be `_`.

This code path would be a lot cleaner with an unwind segue instead of the delegate pattern…

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -107,7 +107,13 @@ open class RouteProgress: NSObject {
     }
     
     public var currentLegProgress: RouteLegProgress!
-    
+
+    /**
+     Intializes a new `RouteProgress`.
+     
+     - parameter route: The route to follow.
+     - parameter legIndex: Zero-based index indicating the current leg the user is on.
```
1ec5: `legIndex` is the index of the leg to start on.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,43 @@ class RouteTableViewController: UIViewController {
...
+        let sourceName = leg.source.name
+        let destinationName = leg.destination.name
+        
+        if let sourceName = sourceName?.nonEmptyString, let destinationName = destinationName?.nonEmptyString {
+            return "\(sourceName) to \(destinationName)"
+        }
+        
+        let legToFrom = leg.name.components(separatedBy: ", ")
+        if legToFrom.count == 2 {
+            return "\(legToFrom[0]) to \(legToFrom[1])"
```
1ec5: Not sure why GitHub thinks https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r123045968 is outdated, but it’s still current.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,43 @@ class RouteTableViewController: UIViewController {
...
+        // Don't display section header if there is only one step
+        guard routeController.routeProgress.route.legs.count > 1 else {
+            return nil
+        }
+        let leg = routeController.routeProgress.route.legs[section]
+        
+        let sourceName = leg.source.name
+        let destinationName = leg.destination.name
+        
+        if let sourceName = sourceName?.nonEmptyString, let destinationName = destinationName?.nonEmptyString {
```
1ec5: If the source is the user’s unnamed current location and the destination is a named location, we should say “Destination via Street Name and Street Name”.

If the source is a named location and the destination is unnamed for some reason, we should say “Street Name and Street Name”, same as below.

1ec5:
Project-OSRM/osrm-text-instructions.swift#24 has landed, by the way.

1ec5:
This PR fixes #204.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,41 @@ class RouteTableViewController: UIViewController {
...
-        let leg = routeController.routeProgress.currentLeg
+        let legs = routeController.routeProgress.route.legs
+        
         
-        cell.step = leg.steps[indexPath.row]
+        cell.step = legs[indexPath.section].steps[indexPath.row]
         
-        if routeController.routeProgress.currentLegProgress.stepIndex + 1 > indexPath.row {
+        if indexPath.section < routeController.routeProgress.legIndex || (indexPath.section == routeController.routeProgress.legIndex && indexPath.row <= routeController.routeProgress.currentLegProgress.stepIndex) {
             cell.contentView.alpha = 0.4
```
frederoni: Upcoming steps may appear faded out if a passed step cell is being reused.
Consider adding an else statement or more succinctly `cell.contentView.alpha = didPassStep ? 0.4 : 1`

Examples/Swift/ViewController.swift
```diff
@@ -254,37 +262,6 @@ extension ViewController: NavigationViewControllerDelegate {
 
 extension ViewController: WaypointConfirmationViewControllerDelegate {
     func confirmationControllerDidConfirm(controller confirmationController: WaypointConfirmationViewController) {
```
1ec5: Per https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r126295971, rename `controller` to `_`.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,41 @@ class RouteTableViewController: UIViewController {
...
+        let leg = routeController.routeProgress.route.legs[section]
+        
+        let sourceName = leg.source.name
+        let destinationName = leg.destination.name
+        let legToFrom = leg.name.components(separatedBy: ", ")
+        
+        if let destinationName = destinationName?.nonEmptyString, legToFrom.count > 1 {
+            return "\(destinationName) via \(legToFrom.joined(separator: " and "))"
+        } else if let sourceName = sourceName?.nonEmptyString, let destinationName = destinationName?.nonEmptyString {
+            return "\(sourceName) anz \(destinationName)"
```
1ec5: Typo: and.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -68,20 +68,41 @@ class RouteTableViewController: UIViewController {
...
+            return nil
+        }
+        let leg = routeController.routeProgress.route.legs[section]
+        
+        let sourceName = leg.source.name
+        let destinationName = leg.destination.name
+        let legToFrom = leg.name.components(separatedBy: ", ")
+        
+        if let destinationName = destinationName?.nonEmptyString, legToFrom.count > 1 {
+            return "\(destinationName) via \(legToFrom.joined(separator: " and "))"
```
1ec5: It would be easier to check for `leg.name` containing `, ` and replace `, ` with ` and `. But in any case, we should use a localizable format string instead of string interpolation.

MapboxNavigation/RouteTableViewController.swift
1ec5: We should use a localizable format string instead of string interpolation.

MapboxNavigation/Resources/Base.lproj/Localizable.strings
```diff
@@ -22,3 +22,8 @@
 /* Format for speech string; 1 = formatted distance; 2 = instruction */
 "WITH_DISTANCE_UTTERANCE_FORMAT" = "In %1$@, %2$@";
 
+/* Format for displaying destination and intermediate waypoints; 1 = source ; 2 = destinations */
+"WAYPOINT_DESTINATION_VIA_WAYPOINTS_FORMAT" = "%1$@, via %2$@";
+
+/* Format for displaying start and endpoint for leg; 1 = source; 2 = destination */
+"WAYPOINT_SOURCE_DESTINATION_FORMAT" = "%1$@, and %2$@";
```
1ec5: No comma needed, since there’s only two items in the list.

1ec5:
It might be worth fixing Project-OSRM/osrm-text-instructions#61 as part of this effort, since the past perfect (“you have arrived”) stands out even more for intermediate waypoints. It would be a trivial change as far as I can tell.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -448,10 +440,52 @@ extension RouteMapViewController: MGLMapViewDelegate {
     }
     
     func showRouteIfNeeded() {
-        guard isViewLoaded && view.window != nil else { return }
```
frederoni: UIAppearance styling of the route line should not work w/o this line and L:113

bsudekum:
Updated this:
* Added custom waypoint markers going from A -> Z
* Added `navigationViewController(navigationViewController:shouldIncrementLegWhenArrivingAtWaypoints:)`. By default, the leg index will be incremented automatically when the user arrives at a waypoint. If the developer wants to show a custom view when arriving at a waypoint, they can set this to false, show their UI and on their own increment leg index.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -447,10 +443,52 @@ extension RouteMapViewController: MGLMapViewDelegate {
...
+            let destination = MGLPointAnnotation()
+            destination.coordinate = lastLeg.destination.coordinate
+            mapView.addAnnotation(destination)
+        }
+        
+        guard let style = map.style else { return }
+        
+        if route.legs.count > 1 {
+            var features = [MGLPointFeature]()
+            let letters = (97...122).map({Character(UnicodeScalar($0))}).map { String(describing:$0).uppercased() }
```
bsudekum: Unsure how this will translate.

/cc @1ec5

Examples/Swift/ViewController.swift
```diff
@@ -121,7 +133,7 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         
         _ = Directions.shared.calculate(options) { [weak self] (waypoints, routes, error) in
             guard error == nil else {
```
frederoni: While you're at it, let's replace this with a `guard let` so we don't have to force unwrap the error.

Examples/Swift/WaypointConfirmationViewController.swift
```diff
@@ -1,14 +1,14 @@
 import UIKit
 
 protocol WaypointConfirmationViewControllerDelegate: NSObjectProtocol {
-    func confirmationControllerDidConfirm(controller: WaypointConfirmationViewController)
+    func confirmationControllerDidConfirm(_: WaypointConfirmationViewController)
```
frederoni: Let's keep the name but avoid the prefix so that the scope variable comes with a predefined name. (`_ controller`)

MapboxNavigation/NavigationMapView.swift
```diff
@@ -125,6 +129,56 @@ open class NavigationMapView: MGLMapView {
...
+            addAnnotation(destination)
+        }
+        
+        var features = [MGLPointFeature]()
+        let letters = (97...122).map({Character(UnicodeScalar($0))}).map { String(describing:$0).uppercased() }
+        
+        for (waypointIndex, waypoint) in routeProgress.remainingWaypoints.enumerated().dropLast() {
+            let feature = MGLPointFeature()
+            feature.coordinate = waypoint.coordinate
+            feature.attributes = [ "name": letters[waypointIndex] ]
```
1ec5:
```swift
feature.attributes = [
    "name": Character(UnicodeScalar(0x41 /* A */ + waypointIndex))
]
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -142,37 +142,17 @@ open class NavigationMapView: MGLMapView {
...
-        var features = [MGLPointFeature]()
-        let letters = (97...122).map({Character(UnicodeScalar($0))}).map { String(describing:$0).uppercased() }
-        
-        for (waypointIndex, waypoint) in routeProgress.remainingWaypoints.enumerated().dropLast() {
-            let feature = MGLPointFeature()
-            feature.coordinate = waypoint.coordinate
-            feature.attributes = [ "name": letters[waypointIndex] ]
-            features.append(feature)
-        }
+        let source = navigationMapDelegate?.navigationMapView?(self, shapeFor: routeProgress.remainingWaypoints) ?? shape(for: Array(routeProgress.remainingWaypoints.dropLast()))
```
1ec5: Let’s give this variable and `s` below more descriptive names.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -75,12 +75,19 @@ public protocol NavigationViewControllerDelegate {
      
      If this method is unimplemented, the navigation map view draws the route line using an `MGLLineStyleLayer`.
      */
-    @objc optional func navigationMapView(_ mapView: NavigationMapView, routeStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
+    @objc optional func navigationMapView(_ mapView: NavigationMapView, routeWaypointCircleStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
```
1ec5: The name of this method implies that the return value must be an MGLCircleStyleLayer, but that isn’t the case. A better name would be `navigationMapView(_:waypointStyleLayerWithIdentifier:source:)`.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -75,12 +75,19 @@ public protocol NavigationViewControllerDelegate {
...
+    
+    /**
+     Returns an `MGLStyleLayer` that determines the appearance of the route line.
+     
+     If this method is unimplemented, the navigation map view draws the route line using an `MGLLineStyleLayer`.
+     */
+    @objc optional func navigationMapView(_ mapView: NavigationMapView, routeWaypointSymbolStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
```
1ec5: What if the developer needs more than just a single circle style layer and symbol style layer to represent the waypoint? It seems like we need a method that returns an array of style layers. NavigationMapView would need to keep track of the identifiers of the returned style layers so it can refer to them later on.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -75,12 +75,19 @@ public protocol NavigationViewControllerDelegate {
...
```
1ec5: What if the waypoint needs to be represented by an annotation, to match the main map in their application? There should be alternative, optional methods that mirror `MGLMapViewDelegate.mapView(_:imageForAnnotation:)` and `MGLMapViewDelegate.mapView(_:viewForAnnotation:)`. When either `navigationMapView(_:imageForAnnotation:waypoint:)` or `navigationMapView(_:viewForAnnotation:waypoint:)` is implemented, NavigationMapView would take care of adding an MGLPointAnnotation instead of a source and style layer.

/ref #114

MapboxNavigation/NavigationViewController.swift
```diff
@@ -34,6 +34,9 @@ public protocol NavigationViewControllerDelegate {
     @objc(navigationViewController:shouldRerouteFromLocation:)
     optional func navigationViewController(_ navigationViewController: NavigationViewController, shouldRerouteFrom location: CLLocation) -> Bool
     
+    @objc(navigationViewController:shouldIncrementLegWhenArrivingAtWaypoints:)
```
1ec5: This public method needs a documentation comment. (Same goes for the corresponding method on RouteController.)

MapboxNavigation/NavigationViewController.swift
```diff
@@ -34,6 +34,9 @@ public protocol NavigationViewControllerDelegate {
...
+    optional func navigationViewController(_ navigationViewController: NavigationViewController, shouldIncrementLegWhenArrivingAtWaypoints waypoint: Waypoint) -> Bool
```
1ec5: I’m confused by the `shouldIncrementLegWhenArrivingAtWaypoints` argument: it makes the method sound like there’s only one setting for all waypoints in the route, yet this method is called once for each waypoint. Did you mean to make the argument name singular, i.e. `shouldIncrementLegWhenArrivingAtWaypoint`? If so, the Swift name should be `navigationViewController(_:shouldIncrementLegWhenArrivingAt:)` while the Objective-C name would be `-navigationViewController:shouldIncrementLegWhenArrivingAtWaypoint:`.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -324,8 +322,12 @@ extension RouteMapViewController: NavigationMapViewDelegate {
         return delegate?.navigationMapView(mapView, routeCasingStyleLayerWithIdentifier: identifier, source: source)
     }
 
-    func navigationMapView(_ mapView: NavigationMapView, routeStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer? {
-        return delegate?.navigationMapView(mapView, routeStyleLayerWithIdentifier: identifier, source: source)
```
1ec5: Why does this PR remove the ability to customize the route line?

bsudekum: Fixed.

MapboxNavigation/NavigationViewController.swift
bsudekum: Do you think we should follow the same logic with the route line since it is also made up of multiple parts?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -22,8 +22,8 @@ public protocol RouteControllerDelegate: class {
     @objc(routeController:shouldRerouteFromLocation:)
     optional func routeController(_ routeController: RouteController, shouldRerouteFrom location: CLLocation) -> Bool
     
-    @objc(routeController:shouldIncrementLegWhenArrivingAtWaypoints:)
-    optional func routeController(_ routeController: RouteController, shouldIncrementLegWhenArrivingAtWaypoints waypoint: Waypoint) -> Bool
+    @objc(routeController:shouldIncrementLegWhenArrivingAtWaypoint:)
+    optional func routeController(_ routeController: RouteController, shouldIncrementLegWhenArrivingAtWaypoint waypoint: Waypoint) -> Bool
```
1ec5: The Swift name should be `routeController(_:shouldIncrementLegWhenArrivingAt:)`. “Waypoint” is inferred from the argument’s type.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -142,20 +149,25 @@ open class NavigationMapView: MGLMapView {
...
             style.addSource(sourceShape)
             
-            let circles = navigationMapDelegate?.navigationMapView?(self, routeWaypointCircleStyleLayerWithIdentifier: waypointCircleIdentifier, source: sourceShape) ?? routeWaypointCircleStyleLayer(identifier: waypointCircleIdentifier, source: sourceShape)
+            let circles = navigationMapDelegate?.navigationMapView?(self, waypointStyleLayerWithIdentifier: waypointCircleIdentifier, source: sourceShape) ?? routeWaypointCircleStyleLayer(identifier: waypointCircleIdentifier, source: sourceShape)
             
-            let symbols = navigationMapDelegate?.navigationMapView?(self, routeWaypointSymbolStyleLayerWithIdentifier: waypointSymbolIdentifier, source: sourceShape) ?? routeWaypointSymbolStyleLayer(identifier: waypointSymbolIdentifier, source: sourceShape)
+            let symbols = navigationMapDelegate?.navigationMapView?(self, waypointSymbolStyleLayerWithIdentifier: waypointSymbolIdentifier, source: sourceShape) ?? routeWaypointSymbolStyleLayer(identifier: waypointSymbolIdentifier, source: sourceShape)
```
1ec5: If the developer wants the waypoint to be marked by a circle with no letter, they’d return `nil` for `navigationMapView(_:waypointSymbolStyleLayerWithIdentifier:source:)`, but then we’d go ahead and add a letter anyways.

MapboxNavigation/NavigationViewController.swift
1ec5: Possibly, but that would limit the developer’s options considerably. For example, what if the developer wants to mark the waypoint with a circle and two symbols – a hotel icon overlaid on the circle and a text label underneath?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -197,7 +209,7 @@ open class NavigationMapView: MGLMapView {
         return MGLShapeCollectionFeature(shapes: [baseLine] + lines)
     }
     
-    func shape(for waypoints: [Waypoint]) -> MGLShape? {
+    func shapes(for waypoints: [Waypoint]) -> MGLShape? {
```
1ec5: The name of this method implies that it returns an array of set of MGLShape objects. However, it returns only a single MGLShape object.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -276,6 +288,129 @@ open class NavigationMapView: MGLMapView {
         
         return lineCasing
     }
+    
+    func hideArrow(_ routeProgress: RouteProgress) {
```
1ec5: This method doesn’t hide the arrow.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -298,5 +433,5 @@ public protocol NavigationMapViewDelegate: class  {
     optional func navigationMapView(_ mapView: NavigationMapView, simplifiedShapeDescribing route: Route) -> MGLShape?
     
     @objc(navigationMapView:shapeDescribingWaypoints:)
-    optional func navigationMapView(_ mapView: NavigationMapView, shapeFor waypoints: [Waypoint]) -> MGLShape?
+    optional func navigationMapView(_ mapView: NavigationMapView, shapesFor waypoints: [Waypoint]) -> MGLShape?
```
1ec5: Similarly, this method name now implies that the method returns an array or set of MGLShape objects, which isn’t the case.

MapboxNavigation/NavigationViewController.swift
1ec5: Then remove runtime styling in favor of annotations, because that’s the usual way people add markers to the map. The main reason to expose this customization hook is to allow the developer to make the destination marker match the annotation on the main map in their application (the map that isn’t NavigationMapView).

1ec5: Then remove runtime styling in favor of annotations, because that’s the usual way people add markers to the map. The main reason to expose this customization hook is to allow the developer to make the destination marker match the annotation on the main map in their application (the map that isn’t NavigationMapView).

MapboxNavigation/NavigationViewController.swift
```diff
@@ -105,7 +98,12 @@ public protocol NavigationViewControllerDelegate {
      */
     @objc optional func navigationMapView(_ mapView: NavigationMapView, simplifiedShapeDescribing route: Route) -> MGLShape?
     
-    @objc optional func navigationMapView(_ mapView: NavigationMapView, shapeFor waypoints: [Waypoint]) -> MGLShape?
+    
+    @objc optional func navigationMapView(_ mapView: NavigationMapView, waypointStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
```
1ec5: Remember to document these new public methods.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -142,20 +149,25 @@ open class NavigationMapView: MGLMapView {
             addAnnotation(destination)
         }
         
-        let source = navigationMapDelegate?.navigationMapView?(self, shapeFor: routeProgress.remainingWaypoints) ?? shape(for: Array(routeProgress.remainingWaypoints.dropLast()))
+        let source = navigationMapDelegate?.navigationMapView?(self, shapesFor: routeProgress.remainingWaypoints) ?? shapes(for: Array(routeProgress.remainingWaypoints.dropLast()))
```
1ec5: This isn’t what I meant by https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r132097275. The line below says `let s`. A single-letter variable should be a red flag. It was previously called `source`, but you changed it to `s` in order for this variable to be called `source`. Instead, call them each `<something>Source`.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -382,10 +392,6 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         return navigationDelegate?.navigationMapView?(mapView, routeCasingStyleLayerWithIdentifier: identifier, source: source)
     }
     
-    func navigationMapView(_ mapView: NavigationMapView, routeStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer? {
-        return navigationDelegate?.navigationMapView?(mapView, routeStyleLayerWithIdentifier: identifier, source: source)
```
1ec5: Per https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r132099394, this PR still removes the ability to customize the route line.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -66,20 +66,41 @@ class RouteTableViewController: UIViewController {
...
+            return nil
+        }
+        let leg = routeController.routeProgress.route.legs[section]
+        
+        let sourceName = leg.source.name
+        let destinationName = leg.destination.name
+        let legToFrom = leg.name.components(separatedBy: ", ")
+        
+        if let destinationName = destinationName?.nonEmptyString, legToFrom.count > 1 {
+            return String.localizedStringWithFormat(NSLocalizedString("WAYPOINT_DESTINATION_VIA_WAYPOINTS_FORMAT", bundle: .mapboxNavigation, value: "%@, via %@", comment: "Format for displaying destination and intermediate waypoints; 1 = source ; 2 = destinations"), destinationName, leg.name.replacingOccurrences(of: ", ", with: " and "))
```
1ec5: The word “and” should be localizable.

MapboxNavigation/NavigationMapView.swift
bsudekum: Wow

bsudekum:
Updated the route line and made the non active leg opaque:

simulator screen shot aug 10 2017 2 16 05 pm

The stacked opacity at geometry intersections is due to https://github.com/mapbox/mapbox-gl-native/issues/1771 😢 

When the user enters the new leg, the route is updated accordingly.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -125,49 +137,141 @@ open class NavigationMapView: MGLMapView {
...
-        let baseLine = MGLPolylineFeature(coordinates: coordinates, count: route.coordinateCount)
-        baseLine.attributes["congestion"] = "unknown"
-        let lines = nontrivialCongestionSegments.map { (congestionSegment: CongestionSegment) -> MGLPolylineFeature in
-            let polyline = MGLPolylineFeature(coordinates: congestionSegment.0, count: UInt(congestionSegment.0.count))
-            polyline.attributes["congestion"] = String(describing: congestionSegment.1)
-            return polyline
+        for (waypointIndex, waypoint) in waypoints.enumerated() {
+            let feature = MGLPointFeature()
+            feature.coordinate = waypoint.coordinate
+            feature.attributes = [ "name": letters[waypointIndex] ]
```
1ec5: If we’re going to label each waypoint with a letter, then the set of letters should be localizable, because each language has its own order. (For example, in Vietnamese it begins “A Ă Â B C D Đ E Ê …”. In Simplified Chinese, it begins with “万千百十…”.)

It would be great if we could use `Locale.exemplarCharacterSet`, but unfortunately there isn’t a way to enumerate a CharacterSet’s members while preserving order. We’ll need a single localizable string that contains a space- or comma-delimited list of exemplar characters.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -171,45 +172,72 @@ open class NavigationMapView: MGLMapView {
...
+                } else {
+                    mergedCongestionSegments.append(seg)
+                }
+            }
+            
+            let lines = mergedCongestionSegments.map { (congestionSegment: CongestionSegment) -> MGLPolylineFeature in
+                let polyline = MGLPolylineFeature(coordinates: congestionSegment.0, count: UInt(congestionSegment.0.count))
+                polyline.attributes["congestion"] = String(describing: congestionSegment.1)
+                if let legIndex = legIndex {
+                    polyline.attributes[isOpaqueIdentifier] = index != legIndex
```
bsudekum: 🤔  I'm wondering if it would get confusing to the user if we added a transparency to both non active legs and completed routes.

bsudekum:
@1ec5 this is ready for another round of reviews.

MapboxNavigation/NavigationViewController.swift
1ec5: Exposing `navigationMapView(_:viewForAnnotation:waypoint:)` would also fix #213.

bsudekum:
More reviews are welcome if anyone is up for it @1ec5 @frederoni.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -154,49 +166,168 @@ open class NavigationMapView: MGLMapView {
...
-        let baseLine = MGLPolylineFeature(coordinates: coordinates, count: route.coordinateCount)
-        baseLine.attributes["congestion"] = "unknown"
-        let lines = nontrivialCongestionSegments.map { (congestionSegment: CongestionSegment) -> MGLPolylineFeature in
-            let polyline = MGLPolylineFeature(coordinates: congestionSegment.0, count: UInt(congestionSegment.0.count))
-            polyline.attributes["congestion"] = String(describing: congestionSegment.1)
-            return polyline
+        for (waypointIndex, waypoint) in waypoints.enumerated() {
+            let feature = MGLPointFeature()
+            feature.coordinate = waypoint.coordinate
+            feature.attributes = [ "name": letters[waypointIndex] ]
```
bsudekum: This seems a bit unreadable IMO:

```swift
for (waypointIndex, waypoint) in waypoints.enumerated() {
  let feature = MGLPointFeature()
  feature.coordinate = waypoint.coordinate
  feature.attributes = [ "name": String.localizedStringWithFormat(NSLocalizedString("ALPHABET", bundle: .mapboxNavigation, value: "ABCDEFGHIJKLMNOPQRSTUVWXYZ", comment: "Format string for alphabet;")).characters.map { String(describing: $0) }[waypointIndex] ]
  features.append(feature)
}
```

bsudekum: What if one waypoint has a name and the other does not? So it'd look like `B` (for picking up bobby) and then a waypoint with no name `B` again? Some sort of order makes more sense to me here.

1ec5: Ah, the code I suggested in https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r132049320 was much more readable, but it no longer makes sense with the localization that we’re doing. How about this:

```swift
let letters = String.localizedStringWithFormat(NSLocalizedString("ALPHABET", bundle: .mapboxNavigation, value: "ABCDEFGHIJKLMNOPQRSTUVWXYZ", comment: "Base letters in the alphabet for labeling waypoints, separated by spaces")).components(separatedBy: " ")

// then in the for loop
feature.attributes = ["name": letters[waypointIndex]]
```

1ec5: If any of the waypoints has a name, I think it would be better to label the named ones with initials and leave the unnamed ones… unlabeled.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -13,6 +14,17 @@ open class NavigationMapView: MGLMapView {
...
+    let waypointCircleIdentifier = "waypointsCircle"
+    let waypointSymbolIdentifier = "waypointsSymbol"
+    let arrowSourceIdentifier = "arrowSource"
+    let arrowSourceStrokeIdentifier = "arrowSourceStroke"
+    let arrowLayerIdentifier = "arrowLayer"
+    let arrowSymbolLayerIdentifier = "arrowSymbolLayer"
+    let arrowLayerStrokeIdentifier = "arrowStrokeLayer"
+    let arrowCasingSymbolLayerIdentifier = "arrowCasingSymbolLayer"
+    let arrowSymbolSourceIdentifier = "arrowSymbolSource"
+    let isCurrentLeg = "isCurrentLeg"
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/270#discussion_r134897473 got collapsed for some reason: the name `isCurrentLeg` implies that it’s a Boolean. Since this string is being used as an attribute name, call it `currentLegAttribute` but give it the value `isCurrentLeg`.

hervouinc:
🍾🥂

johnwallco:
When will multiple waypoints be available?

MapboxNavigation/NavigationViewController.swift
1ec5: #1993

PR #271, opened 2017-06-09
Fix coordinate(distance:.. crash

bsudekum:
Shouldn't always assume that those coordinates are always there.

/cc @frederoni @ericrwolfe

### DIFF
--- a/MapboxNavigation/SimulatedRoute.swift
+++ b/MapboxNavigation/SimulatedRoute.swift
@@ -123,12 +123,12 @@ fileprivate extension Array where Element == SimulatedLocation {
             let timeToTravel: TimeInterval = distance / speed
             
             if timeToTravel > 1 {
-                let newCoordinate = coordinate(at: speed, fromStartOf: [location.coordinate, nextLocation.coordinate])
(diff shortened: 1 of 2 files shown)


MapboxNavigation/SimulatedRoute.swift
```diff
@@ -123,12 +123,12 @@ fileprivate extension Array where Element == SimulatedLocation {
             let timeToTravel: TimeInterval = distance / speed
             
             if timeToTravel > 1 {
-                let newCoordinate = coordinate(at: speed, fromStartOf: [location.coordinate, nextLocation.coordinate])
-                let newLocation = SimulatedLocation(coordinate: newCoordinate!,
+                guard let newCoordinate = coordinate(at: speed, fromStartOf: [location.coordinate, nextLocation.coordinate]) else { return processedLocations }
```
bsudekum: Not really sure if this is the right thing to return early here.

ericrwolfe: Calling return in the `guard` statement here will break out of the for loop prematurely. Seems like it might be better to `continue` instead of returning early.

bsudekum:
@ericrwolfe fixed

PR #275, opened 2017-06-12
Don’t announce if muted; cleanup voice code/comments

friedbunny:
We would previously do all of the setup to speak an announcement, even if the volume was muted. This was unnecessary work and could lead to potential crashes.

This PR also removes conditional code aimed at supporting iOS 8, as our minimum deployment version is currently set to iOS 9.

/cc @bsudekum @frederoni


friedbunny:
Tests are failing because snapshot images appear to be missing.

> failed - Snapshot comparison failed: Optional(Error Domain=FBSnapshotTestControllerErrorDomain Code=1 "Unable to load reference image..."

frederoni:
LGTM but trying to figure out why Bitrise is using the wrong simulator or why the fixtures are missing.

Issue #5, opened 2016-11-18
Documentation


1ec5:
The library should have a documentation comment for every public symbol. Swift uses reStructuredText format for documentation comments.

Additionally, we can look into generating HTML documentation based on these comments. The mobile team is quite happy with [jazzy](https://github.com/realm/jazzy), which relies on the compiler for parsing, so you know it’ll work with whatever syntax you throw at it. It also generates a Dash docset for those who like their documentation that way.

/cc @bsudekum @friedbunny

1ec5:
Xcode- and jazzy-compatible documentation was added in #250, unblocking this issue. Now you can run the following command line invocation:

```sh
jazzy --sdk iphonesimulator --github-file-prefix 'https://github.com/mapbox/mapbox-navigation-ios/tree/master' --module-version 0.4.0 --readme README.md --documentation=docs/*.md --root-url 'https://www.mapbox.com/navigation-sdk/ios/0.4.0/' --umbrella_header MapboxCoreNavigation/MapboxCoreNavigation.h --framework_root MapboxCoreNavigation --output ~/Desktop/docs/
```

to get Core Navigation documentation locally:

![jazzy](https://user-images.githubusercontent.com/1231218/27094815-5492aede-5020-11e7-9611-d7a027a4b25f.png)

Some caveats:

* We have two modules, MapboxCoreNavigation and MapboxNavigation. jazzy can only document one module per docset and lacks the ability to tie them together (or even cross-link symbols). I think this should be OK as long as our setup instructions are sufficiently fleshed out.
* The image in the “Storyboards” guide is broken.

We’ll need to find a good online home for this documentation. We could set up GitHub Pages temporarily, but hosting the docsets on mapbox.com would be preferable.

friedbunny:
The Mapbox theme [lives here](https://github.com/mapbox/mapbox-gl-native/tree/9d276f3b0f0e5f1b5089a6cd727927361e6634ac/platform/darwin/docs/theme).

ericrwolfe:
Closed via #285.

PR #279, opened 2017-06-13
Add jazzy docs

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/5#issuecomment-308185252

Mostly concerned about config in https://github.com/mapbox/mapbox-navigation-ios/blob/a8062d8264164729fed0f61732cc0f68a28603b3/scripts/docs.sh

/cc @1ec5 @frederoni


frederoni:
Will the generated files reside in another repo? We shouldn't check in the files here if that's the case.

scripts/docs.sh
```diff
@@ -0,0 +1,26 @@
...
+    fi
+fi
+
+for module in MapboxNavigation MapboxCoreNavigation
+do
+    jazzy \
+        --config jazzy.yml \
+        --module "$module" \
+        --module-version 0.4.0 \
+        --root-url 'https://www.mapbox.com/navigation-sdk/ios/0.4.0/' \
```
1ec5: This documentation would actually have a root URL of https://mapbox.github.io/mapbox-navigation-ios/{MapboxCoreNavigation,MapboxNavigation}/ (without the version). Therefore all the links and image references would be broken.

We should version the docsets in the URLs for better permalinking.

scripts/docs.sh
```diff
@@ -0,0 +1,26 @@
...
+        --github-file-prefix 'https://github.com/mapbox/mapbox-navigation-ios/tree/master' \
```
1ec5: We should pin the file prefix to the current tag. That would necessitate pulling the module version out into a local variable. [This code](https://github.com/mapbox/mapbox-gl-native/blob/5e0018cdc3844994bf4afdc0a5e6ae90636c9967/platform/ios/scripts/document.sh#L18-L20) automatically detects the current release tag.

scripts/docs.sh
bsudekum: I'm not sure I understand why it would drop the `/0.4.0/` even though it's specified in explicitly in `--root-url`. Can you elaborate @1ec5?

bsudekum:
Closing in favor of https://github.com/mapbox/mapbox-navigation-ios/pull/285

scripts/docs.sh
1ec5: `root-url` tells jazzy where the documentation will wind up. The documentation’s location is determined by where you upload it to. In this case, this PR would have used GitHub Pages, which would have made the documentation accessible from mapbox.github.io, not mapbox.com.

Issue #283, opened 2017-06-15
Maneuver arrow is backward when map is zoomed in


hardsetting:
I noticed that sometimes during navigation, the maneuver arrows showed up backward.

<img src="https://user-images.githubusercontent.com/7707719/27184093-44405ee2-51e1-11e7-9197-5c9e578bf088.PNG" height="720" /> <img src="https://user-images.githubusercontent.com/7707719/27184094-44463ace-51e1-11e7-9bb1-bb2c79c77ca0.PNG" height="720"  />

Initially I thought it was an error of the route itself, but digging more into it I found out a couple of little bugs in the generation of the arrows.

I think I might be able to submit a PR in the next days.

hardsetting:
My apologies,
I think the issue arose from a small edit I made in my code.
I could not reproduce with code from the master branch.
Closing.

PR #39, opened 2017-01-30
Round up

bsudekum:
Rounds up for `closestCoordinate`

/cc @1ec5

PR #284, opened 2017-06-15
Fixed bugs in closestCoordinate and addArrow functions.

hardsetting:
Referring to #283 

In _closestCoordinate_ function:
when the closest point to the polyline segment is the projection, now the index is set to the index of the coordinate closest to the intersection.

In _addArrow_ function:
shafts are now generated by concatenating the points before the maneuver **in the right order** with the points after the maneuver, making sure the **closest point to the maneuver is taken once**.

EDIT:
I now realize that the changes in _closestCoordinate_ might not be strictly necessary, but it's probably clearer and safer now anyways.

### DIFF
--- a/MapboxCoreNavigation/Geometry.swift
+++ b/MapboxCoreNavigation/Geometry.swift
@@ -151,7 +151,7 @@ func closestCoordinate(on polyline: [CLLocationCoordinate2D], to coordinate: CLL
     
     var closestCoordinate: CoordinateAlongPolyline?
     
-    for var index in 0..<polyline.count - 1 {
+    for index in 0..<polyline.count - 1 {
         let segment = (polyline[index], polyline[index + 1])
         let distances = (coordinate - segment.0, coordinate - segment.1)
         
@@ -166,11 +166,10 @@ func closestCoordinate(on polyline: [CLLocationCoordinate2D], to coordinate: CLL
             closestCoordinate = CoordinateAlongPolyline(coordinate: segment.0, index: index, distance: distances.0)
         }
         if distances.1 < closestCoordinate?.distance ?? CLLocationDistanceMax {
-            index = index + 1
-            closestCoordinate = CoordinateAlongPolyline(coordinate: segment.1, index: index, distance: distances.1)
+            closestCoordinate = CoordinateAlongPolyline(coordinate: segment.1, index: index+1, distance: distances.1)
         }
         if intersectionDistance != nil && intersectionDistance! < closestCoordinate?.distance ?? CLLocationDistanceMax {
-            closestCoordinate = CoordinateAlongPolyline(coordinate: intersectionPoint!, index: index, distance: intersectionDistance!)
+            closestCoordinate = CoordinateAlongPolyline(coordinate: intersectionPoint!, index: (distances.0 < distances.1 ? index : index+1), distance: intersectionDistance!)
         }
     }
     
--- a/MapboxNavigation/MGLMapView.swift
+++ b/MapboxNavigation/MGLMapView.swift
@@ -44,13 +44,13 @@ extension MGLMapView {
         }
         
         let shaftLength = max(min(50 * metersPerPoint(atLatitude: maneuverCoordinate!.latitude), 50), 10)
-        let shaftCoordinates = polyline(along: polylineCoordinates!, within: -shaftLength / 2, of: maneuverCoordinate!)
-            + polyline(along: polylineCoordinates!, within: shaftLength, of: maneuverCoordinate!)
+        let shaftCoordinates = Array(polyline(along: polylineCoordinates!, within: -shaftLength / 2, of: maneuverCoordinate!).reversed()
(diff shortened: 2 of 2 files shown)


hardsetting:
I have closed the referenced issue, as it emerged from a change in my code.
I'll keep this open in case the changes I made were still useful, but feel free to close.

bsudekum:
@hardsetting thanks for the PR! Could you distill this down to a simple example and maybe link some screenshots to [geojson.io](http://geojson.io/)? 

My first instinct is that adding `index + 1` will in fact return _the next_ closest coordinate and not the actual closest coordinate. FWIW, I believe at one point this function did accept an optional `look next point?` argument. 

/cc @1ec5

MapboxNavigation/MGLMapView.swift
```diff
@@ -44,13 +44,13 @@ extension MGLMapView {
         }
         
         let shaftLength = max(min(50 * metersPerPoint(atLatitude: maneuverCoordinate!.latitude), 50), 10)
-        let shaftCoordinates = polyline(along: polylineCoordinates!, within: -shaftLength / 2, of: maneuverCoordinate!)
-            + polyline(along: polylineCoordinates!, within: shaftLength, of: maneuverCoordinate!)
+        let shaftCoordinates = Array(polyline(along: polylineCoordinates!, within: -shaftLength / 2, of: maneuverCoordinate!).reversed()
+            + polyline(along: polylineCoordinates!, within: shaftLength, of: maneuverCoordinate!).suffix(from: 1))
```
1ec5: mapbox/mapbox-gl-native#8808 (which is in the v3.6.0 betas) makes the map SDK more resilient to redundant vertices, but it’s good to address the root cause too. 👍

MapboxCoreNavigation/Geometry.swift
```diff
@@ -166,11 +166,10 @@ func closestCoordinate(on polyline: [CLLocationCoordinate2D], to coordinate: CLL
             closestCoordinate = CoordinateAlongPolyline(coordinate: segment.0, index: index, distance: distances.0)
         }
         if distances.1 < closestCoordinate?.distance ?? CLLocationDistanceMax {
-            index = index + 1
-            closestCoordinate = CoordinateAlongPolyline(coordinate: segment.1, index: index, distance: distances.1)
+            closestCoordinate = CoordinateAlongPolyline(coordinate: segment.1, index: index+1, distance: distances.1)
         }
         if intersectionDistance != nil && intersectionDistance! < closestCoordinate?.distance ?? CLLocationDistanceMax {
-            closestCoordinate = CoordinateAlongPolyline(coordinate: intersectionPoint!, index: index, distance: intersectionDistance!)
+            closestCoordinate = CoordinateAlongPolyline(coordinate: intersectionPoint!, index: (distances.0 < distances.1 ? index : index+1), distance: intersectionDistance!)
```
1ec5: > My first instinct is that adding `index + 1` will in fact return _the next_ closest coordinate and not the actual closest coordinate.

@bsudekum, this change can be seen as completing the work begun in #39, making the index more consistent with the returned coordinate. `index` initially points to the vertex at the beginning of the segment being examined, so if the coordinate is closer to the end of the segment, the closest vertex would be at `index + 1`.

> FWIW, I believe at one point this function did accept an optional `look next point?` argument.

You’re probably thinking of `includeDistanceToNextCoordinate`, which was removed in #9. In addition to conditionally incrementing the `index` in this third case, the `includeDistanceToNextCoordinate` argument also [updated the `intersectionDistance`](https://github.com/mapbox/mapbox-navigation-ios/blob/5862a33bf76a0247a144c0a4a5f599bd10bb5464/MapboxNavigation/Geometry.swift#L151-L156) (incorrectly).

bsudekum:
@hardsetting this is close! have a sec to update the PR with @1ec5's feedback?

hardsetting:
I apologize for the delay, have been busy for the last couple weeks.
The suggested changes in _closestCoordinate_ function are probably not that important, just a more linear way of doing the same thing.
When I made the change I mistakenly thought the `index = index+1` would make the algorithm skip the next segment (thus leading to some edge case small bugs), as would happen in a C-like programming language, but I now see that's not the case.

Regarding the reversal of first half of the shaft, the `polyline(along:within:of:)` function indeed reverses the points along the main polyline, so that it starts to add up to `distance` from the coordinate specified as parameter, but as a result the returned collection of coordinates goes from the "center" of the arrow to the tail.
To my intuition, the first half of the shaft should go from tail to center, and the second half from center to tip.
I might not actually be a problem at all though.

MapboxCoreNavigation/Geometry.swift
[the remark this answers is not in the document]
1ec5: It turns out this change does cause `polyline(along:from:to:)` (now `LineString.sliced(from:to:)`) to behave erratically: https://github.com/mapbox/turf-swift/issues/27#issuecomment-447513368.

PR #285, opened 2017-06-16
Navigation SDK documentation

ericrwolfe:
Addresses #5 and supersedes #279 with the following changes: 

* Fixes the 2 module issue by creating another `MapboxNavigation-Documentation.podspec` that combines both modules source files into a single module
    * We could hide this under the `docs` folder, but I couldn't move it without getting build errors
    * Open to more elegant solutions here
* Organizes the docs using `jazzy.yml`
* Excludes the generated docset from this repo

See this commit (https://github.com/mapbox/mapbox-navigation-ios/pull/285/commits/7310e1cb25e331770897ca33c1fffd933c1e2920) for the main changes.

routecontroller class reference

Note that quite a number of these classes and methods are undocumented, so we'll want to circle back and document or hide before we host these somewhere.

### Hosting these docs
These generated docs should live and be versioned independently of this repo, but where? 🤔 

Some options:
* `www.mapbox.com/navigation-sdk/ios/docs/x.x.x/`
    * I'm not sure if we can nest under that URL if `navigation-sdk` is already claimed cc @davidtheclark
* `www.mapbox.com/ios-sdk/api/navigation/x.x.x/`
    * This would live in our iOS Maps SDK documentation repo, but it might be hard to manage independently of maps sdk releases @friedbunny 
* `www.mapbox.com/navigation-sdk-ios/docs/x.x.x/`
    * This would require another repo set up like iOS Maps SDK documentation repo but managed independently

cc @bsudekum @1ec5 @captainbarbosa


.gitignore
```diff
@@ -12,3 +12,5 @@ Pods/
 
 Carthage/Build
 Carthage/Checkouts
+/documentation
```
bsudekum: We'll have to check in the files for these to be pushed to mapbox.com

ericrwolfe: Right, but not in this repo. We should probably check these in on a separate repo similar to how we handle https://github.com/mapbox/ios-sdk

scripts/document.sh
```diff
@@ -0,0 +1,43 @@
...
+    --sdk iphonesimulator \
+    --module-version ${SHORT_VERSION} \
+    --github-file-prefix "https://github.com/mapbox/mapbox-navigation-ios/tree/${BRANCH}" \
+    --documentation=docs/guides/*.md \
+    --root-url "https://www.mapbox.com/navigation-sdk/ios/${RELEASE_VERSION}/" \
+    --theme ${THEME} \
+    --output ${OUTPUT}
+
+find ${OUTPUT} -name *.html -exec \
+    perl -pi -e 's/BRANDLESS_DOCSET_TITLE/Mapbox Navigation SDK for iOS $1/, s/MapboxNavigation\s+(Docs|Reference)/Mapbox Navigation SDK for iOS $1/' {} \;
```
1ec5: The “brandless” docset title should be replaced with just “Navigation SDK for iOS” to avoid duplicating the Mapbox logo that appears directly to the left of it.

docs/theme/assets/img/mapbox.svg
```diff
@@ -0,0 +1,21 @@
...
+<svg width="580px" height="156px" viewBox="0 0 580 156" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 3.8.3 (29802) - http://www.bohemiancoding.com/sketch -->
+    <title>layer1</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="mapbox" transform="translate(-1.000000, -1.000000)" fill="#FFFFFF">
+            <g id="layer1" transform="translate(0.428470, 0.781010)">
+                <g id="g3024" transform="translate(0.542850, 0.401720)">
+                    <path d="M0.028592,0.417258 L0.028592,122.817276 L18.028592,122.817276 L18.028592,22.017258 L54.028592,83.217258 L90.028592,22.017258 L90.028592,122.817276 L108.028592,122.817276 L108.028592,0.417258 L93.628592,0.417258 C86.428592,0.394218 80.77382,2.217258 77.597342,7.617258 L54.028592,47.667258 L30.459842,7.617258 C27.283364,2.217258 21.628592,0.394218 14.428592,0.417258 L0.028592,0.417258 L0.028592,0.417258 Z" id="path4479"></path>
```
1ec5: This is Mapbox’s old corporate branding.

scripts/document.sh
```diff
@@ -0,0 +1,43 @@
...
+THEME=${JAZZY_THEME:-$DEFAULT_THEME}
+
+jazzy \
+    --podspec MapboxNavigation-Documentation.podspec \
+    --config docs/jazzy.yml \
+    --sdk iphonesimulator \
+    --module-version ${SHORT_VERSION} \
+    --github-file-prefix "https://github.com/mapbox/mapbox-navigation-ios/tree/${BRANCH}" \
+    --documentation=docs/guides/*.md \
+    --root-url "https://www.mapbox.com/navigation-sdk/ios/${RELEASE_VERSION}/" \
```
1ec5: If we upload the generated files to this repository, the root URL will actually be https://www.mapbox.com/mapbox-navigation-ios/. Is the intention to upload the generated files manually to the mapbox.com repository instead?

MapboxCoreNavigation.podspec
```diff
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
   Mapbox Core Navigation provides the core spatial and timing logic for turn-by-turn navigation along a route. For a complete turn-by-turn navigation interface, use the Mapbox Navigation SDK for iOS (MapboxNavigation).
                    DESC
 
-  s.homepage = "https://www.mapbox.com/directions/"
+  s.homepage = "https://www.mapbox.com/navigation-sdk/"
```
1ec5: Good catch!

MapboxCoreNavigation.podspec
```diff
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
...
```
1ec5: We should also set `s.documentation_url` to the docset’s expected location.

scripts/document.sh
1ec5: We can apply a similar approach to automatically insert links to MapboxDirections.swift (mapbox/MapboxDirections.swift#140). For example, we can insert `<a>` tags into `<span class="kt">Route</span>` etc.

docs/jazzy.yml
```diff
@@ -0,0 +1,92 @@
...
+      - NavigationViewController
+      - NavigationViewControllerDelegate
+      - RouteVoiceController
+      - RouteTableViewController
+      - NavigationPulleyViewController
+  - name: Maps
+    children:
+      - NavigationMapView
+      - NavigationMapViewDelegate
+      - MGLMapView
```
1ec5: I take it this is the extension on MGLMapView. I wonder if we can add a link to the map SDK documentation [here](https://github.com/mapbox/mapbox-navigation-ios/blob/05130b024c6970178a6c8c3e2b4627dec4a02a6f/MapboxNavigation/MGLMapView.swift#L10) to get it to appear in the MGLMapView class documentation in this docset.

docs/theme/assets/img/mapbox.svg
1ec5: Here’s [an updated asset](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/darwin/docs/theme/assets/img/mapbox.svg).

docs/jazzy.yml
ericrwolfe: Do you mean add a link manually in the documentation comment block or something else? I'd like to do a second pass on documenting the core classes and functions in Mapbox Navigation in the coming days

1ec5: Yeah, a link inside the documentation comment block would help developers understand that MGLMapView isn’t just a couple methods like the traffic toggle. We could just as well move this entry down into one of the generic sections since NavigationMapView is the important class here.

scripts/document.sh
```diff
@@ -0,0 +1,53 @@
...
+    --github-file-prefix "https://github.com/mapbox/mapbox-navigation-ios/tree/${BRANCH}" \
+    --documentation=docs/guides/*.md \
+    --root-url "${BASE_URL}/navigation/${RELEASE_VERSION}/" \
+    --theme ${THEME} \
+    --output ${OUTPUT}
+
+REPLACE_REGEXP='s/MapboxNavigation\s+(Docs|Reference)/Mapbox Navigation SDK for iOS $1/, '
+REPLACE_REGEXP+='s/BRANDLESS_DOCSET_TITLE/Navigation SDK for iOS $1/, '
+REPLACE_REGEXP+="s/<span class=\"kt\">(${DIRECTIONS_SYMBOLS})<\/span>/<span class=\"kt\"><a href=\"${BASE_URL//\//\\/}\/directions\/${DIRECTIONS_VERSION}\/Classes\/\$1.html\">\$1<\/a><\/span>/, "
+
```
1ec5: Come to think of it, we could also link MGLMapView to the map SDK documentation this way.

PR #288, opened 2017-06-20
Fix attribution and logo layout

### LINKED ISSUE #236: Attribution and info button should be visible
image

image

/cc @frederoni

frederoni:
Fixes #236 

Besides moving the ornaments, this PR also changes the partially revealed height from 75% to 60% of the screen height.

@bsudekum 👀

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -10,7 +10,7 @@ class ArrowFillPolyline: MGLPolylineFeature {}
 class ArrowStrokePolyline: ArrowFillPolyline {}
 
 
-class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDelegate {
+class RouteMapViewController: UIViewController {
     @IBOutlet weak var mapView: NavigationMapView!
 
     @IBOutlet weak var overviewButton: Button!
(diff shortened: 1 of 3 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -47,8 +47,6 @@ class RouteMapViewController: UIViewController, PulleyPrimaryContentControllerDe
     var arrowCurrentStep: RouteStep?
     var isInOverviewMode = false
 
-    let overviewContentInset = UIEdgeInsets(top: 65, left: 15, bottom: 55, right: 15)
```
bsudekum: Overview mode required some extra padding on all sides to account for the marker height. How does it look right now?

bsudekum:
Looks good other than in this mode, two things:

* The 17m is clipped
* This image is also in overview mode. When we try and fit the bounds in this view, it zooms all the way out it looks like.

simulator screen shot jun 20 2017 9 53 23 am

frederoni:
@bsudekum good catch. The overview content insets have been reverted.
The overlapping was an iOS 11 beta issue. It's the exact opposite but let's not make an exception for the first beta since it's still very flaky.



Here's another iOS 11 shot for demonstration:

PR #290, opened 2017-06-20
Show shield for swiped to step

bsudekum:
We're attempting to show the shield for the current step instead of the provide step. In cases where we're swiping ahead, we should show the highway shield provided in the function call.

/cc @frederoni

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -516,11 +516,7 @@ extension RouteMapViewController: MGLMapViewDelegate {
     }
 
     func updateShield(for controller: RouteManeuverViewController) {
-        let currentLegProgress = routeController.routeProgress.currentLegProgress
-
-        guard let upComingStep = currentLegProgress?.upComingStep else { return }
-        guard let ref = upComingStep.codes?.first else { return }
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -496,11 +496,7 @@ extension RouteMapViewController: MGLMapViewDelegate {
     }
 
     func updateShield(for controller: RouteManeuverViewController) {
-        let currentLegProgress = routeController.routeProgress.currentLegProgress
-
-        guard let upComingStep = currentLegProgress?.upComingStep else { return }
-        guard let ref = upComingStep.codes?.first else { return }
-        guard controller.shieldImage == nil else { return }
```
frederoni: Keeping this guard would act as a failsafe for caching.

Currently, we also call `updateShield(for:)` from `notifyDidChange` one time per second. Can we try to limit that to one request per `willTransitionTo`?

bsudekum: @frederoni 👍 fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/290/commits/a898ffa1abbefdddaf7f7259763b1d02a93f354f

PR #291, opened 2017-06-20
Don't use ceremonial name when way is interstate

bsudekum:
In cases where the user is coming up to a highway, we should not show the ceremonial highway name.

What should the text say? Right now it says `US-280` and also shows the shield image.

### DIFF
--- a/MapboxNavigation/Highways.plist
+++ b/MapboxNavigation/Highways.plist
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>us-interstate</key>
+	<string>I</string>
+	<key>us-interstate-business</key>
+	<string>I</string>
+	<key>us-interstate-duplex</key>
+	<string>I</string>
+	<key>us-interstate-truck</key>
+	<string>I</string>
+	<key>us-highway</key>
+	<string>US</string>
+	<key>us-highway-duplex</key>
+	<string>US</string>
+	<key>us-highway-alternate</key>
+	<string>US</string>
+	<key>us-highway-business</key>
+	<string>US</string>
+	<key>us-highway-bypass</key>
+	<string>US</string>
+	<key>us-highway-truck</key>
+	<string>US</string>
+</dict>
+</plist>
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -171,7 +171,11 @@ class RouteManeuverViewController: UIViewController {
     
     func updateStreetNameForStep() {
         if let name = step?.names?.first {
-            destinationLabel.unabridgedText = name
+            if let step = step, let ref = step.codes?.first, ["i", "us", "ca"].contains(ref.lowercased().components(separatedBy: " ").first!) {
+                destinationLabel.unabridgedText = ref.components(separatedBy: " ").joined(separator: "-")
+            } else {
+                destinationLabel.unabridgedText = name
+            }
         } else if let destinations = step?.destinations {
             destinationLabel.unabridgedText = destinations.prefix(min(numberOfDestinationLines, destinations.count)).joined(separator: "\n")
         } else if let step = step {
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -173,7 +173,7 @@ class RouteManeuverViewController: UIViewController {
         if let destinations = step?.destinations {
             destinationLabel.unabridgedText = destinations.prefix(min(numberOfDestinationLines, destinations.count)).joined(separator: "\n")
         } else if let name = step?.names?.first {
-            if let ref = step.codes?.first, !["freeway", "expressway", "highway"].contains(name.lowercased()) {
+            if let ref = step.codes?.first, name.isFreeway {
                 destinationLabel.unabridgedText = ref
             } else {
                 destinationLabel.unabridgedText = name
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -170,14 +170,21 @@ class RouteManeuverViewController: UIViewController {
     }
     
     func updateStreetNameForStep() {
-        if let name = step?.names?.first {
-            if let step = step, let ref = step.codes?.first, ["i", "us", "ca"].contains(ref.lowercased().components(separatedBy: " ").first!) {
-                destinationLabel.unabridgedText = ref.components(separatedBy: " ").joined(separator: "-")
+        var stepRef: String?
+        if let step = step, let ref = step.codes?.first, ["i", "us"].contains(ref.lowercased().components(separatedBy: " ").first!) {
+            stepRef = ref
+        }
+        
+        if let destinations = step?.destinations {
+            let dest = destinations.prefix(min(numberOfDestinationLines, destinations.count)).joined(separator: "\n")
+            
+            if let stepRef = stepRef {
+                destinationLabel.unabridgedText = "\(stepRef) / \(dest)"
             } else {
-                destinationLabel.unabridgedText = name
+                destinationLabel.unabridgedText = dest
             }
-        } else if let destinations = step?.destinations {
-            destinationLabel.unabridgedText = destinations.prefix(min(numberOfDestinationLines, destinations.count)).joined(separator: "\n")
+        } else if let name = step?.names?.first {
+            destinationLabel.unabridgedText = stepRef ?? name
         } else if let step = step {
             destinationLabel.unabridgedText = routeStepFormatter.string(for: step)
         }
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -170,21 +170,14 @@ class RouteManeuverViewController: UIViewController {
     }
     
     func updateStreetNameForStep() {
-        var stepRef: String?
-        if let step = step, let ref = step.codes?.first, ["i", "us"].contains(ref.lowercased().components(separatedBy: " ").first!) {
-            stepRef = ref
-        }
-        
         if let destinations = step?.destinations {
-            let dest = destinations.prefix(min(numberOfDestinationLines, destinations.count)).joined(separator: "\n")
-            
-            if let stepRef = stepRef {
-                destinationLabel.unabridgedText = "\(stepRef) / \(dest)"
-            } else {
-                destinationLabel.unabridgedText = dest
-            }
+            destinationLabel.unabridgedText = destinations.prefix(min(numberOfDestinationLines, destinations.count)).joined(separator: "\n")
         } else if let name = step?.names?.first {
-            destinationLabel.unabridgedText = stepRef ?? name
+            if let ref = step.codes?.first, !["freeway", "expressway", "highway"].contains(name.lowercased()) {
+                destinationLabel.unabridgedText = ref
+            } else {
+                destinationLabel.unabridgedText = name
+            }
         } else if let step = step {
             destinationLabel.unabridgedText = routeStepFormatter.string(for: step)
         }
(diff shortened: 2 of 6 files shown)


frederoni:
> What should the text say?

This is unclear to me as well. We should add a snapshot test for this case when we figured out what the correct way is.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -119,7 +119,11 @@ class RouteManeuverViewController: UIViewController {
     
     public func updateStreetNameForStep() {
         if let name = step?.names?.first {
-            streetLabel.unabridgedText = name
+            if let step = step, let ref = step.codes?.first, ["i", "us"].contains(ref.lowercased().components(separatedBy: " ").first!) {
```
1ec5: I think we’re on solid footing with Interstates, but I’m not so sure about doing this for U.S. highways. Sure, a few like U.S. 1 and U.S. 101 are freeways in some places, but not in others, and the vast majority are surface streets where the name matters more. (Think Van Ness in San Francisco or Constitution Ave. NW in D.C.)

1ec5: This can be a bit simpler: `let ref = step.codes?.first where ref.hasPrefix("I ")`.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -119,7 +119,11 @@ class RouteManeuverViewController: UIViewController {
...
+                streetLabel.unabridgedText = ref.components(separatedBy: " ").joined(separator: "-")
```
1ec5: U.S. Routes are written “US 123” in most states. Only a few states like Oklahoma hyphenate them.

1ec5: Can we try to fall back to the step’s `destinations`, since the shield will already be visible to the right?

MapboxNavigation/RouteManeuverViewController.swift
ericrwolfe: screen shot 2017-06-20 at 2 29 34 pm
Could we filter and display the US-101 for segments of type `highway` and use the ceremonial name in other cases? `Redwood Highway` in this screenshot is confusing, but to your point the other street names should show their proper names.

ericrwolfe:
Eventually we should show the highway icon + the direction (<I-580> East), but until we can scoot the highway icon to the left, let's just show text on the left plus the highway icon on the right as is currently implemented.

Ultimately, we want our turn banner to match the highway sign overhead as closely as possible, so the banner text should take the following order of precedence, when the appropriate data is available:

* `I-5 S / Los Angeles`: always show direction + destination when available
* `I-5 South`: always show direction when available
* `I-5`: Fallback if no other data is available 

Voice announcements should take similar levels of precedence:
* `I 5 South towards Los Angeles`
* `I 5 South`
* `I 5`

We should also take care to combine/coalesce maneuvers where the highway ref stays the same but only the ceremonial name changes.

Per https://github.com/mapbox/mapbox-navigation-ios/pull/291#discussion_r123102733, can we check the road class for `highway` or `motorway` before displaying the highway ref and fall back to the ceremonial name for smaller streets?

MapboxNavigation/RouteManeuverViewController.swift
1ec5: Unfortunately, the Directions API doesn’t provide any information about the “type” of a road segment. #297 for example must query the map’s vector tiles for this information, but it can only do so for features currently visible on the map, not roads that may be far out of view.

1ec5:
Given the limited information we get from the Directions API, I think we can only address the Interstate case reliably. If we show both the name and shield for a U.S. Route that’s a freeway, there’s a chance that we’d look pedantic, but at least we wouldn’t be hiding any information that might be required for wayfinding.

1ec5:
It’s a bit less reliable, but the other approach I mentioned in https://github.com/Project-OSRM/osrm-text-instructions/issues/51#issuecomment-260088733 would also be feasible: the step’s expected distance divided by expected duration yields an average expected speed. If the speed is at least 50 mph (or the equivalent in metric), treat the way as a freeway.

ericrwolfe:
I'm game for speed as a proxy/hack to road class until we can get more detailed responses back from the directions API or OSRMTextInstructions.

ericrwolfe:
Having highway shields inline like we used to do would be the preferred approach, except with the shields on the left and with a direction suffix and destination `<I-5> S / Los Angeles` instead of the ceremonial name at highway speeds.


One of the most important points in fixing this issue is adding a direction suffix and destination information when available. It's common to see highways where both directions are served by the same initial offramp, and our current ceremonial name approach does a poor job differentiating those.

Here's some logic we've used in the past to estimate if the current step is a highway:
```swift
let speed = currentStep.distance / currentStep.expectedTravelTime
// Convert 47.5mph to m/s 
if speed > 47.5 * 1609.344 / 3600.0 {
}
```

In the case of the ambiguous `us` prefix, we could default to the following logic (`<>` indicates an inline shield):
* `<US-101> N / Portland`: highway speeds: always show direction + destination when available
* `<US-101> North`: highway speeds: always show direction when available
* `Van Ness Ave <US-101>`: local speeds: use name + show the highway shield inlined to the _right_
* `<US-101> Van Ness Ave`: Fallback: highway shield on the left, ceremonial name on the right

1ec5:
> One of the most important points in fixing this issue is adding a direction suffix and destination information when available.

I think this may rely on one of the more fragile parts of the [exit tagging scheme](https://wiki.openstreetmap.org/wiki/Exit_Info). OSM has specific, structured tags for destination refs, but there's no dedicated tag for the direction. (This information is encoded in route relations, but OSRM lacks the ability to process relations.) So mappers have generally taken to saying "US 101 North" as the ref. Anything that requires string munging on our end will have uneven reliability: consider that the direction is usually given as a suffix in California, whereas it may be given as a prefix in other parts of the U.S. due to differing signage standards.

MapboxNavigation/RouteManeuverViewController.swift
bsudekum: @1ec5 this does not account for `ref.hasPrefix("US ")` though. `contains` allows us to add more cases to this array.

1ec5: We shouldn’t unconditionally remove the name from U.S. Routes. See https://github.com/mapbox/mapbox-navigation-ios/pull/291#discussion_r123060422. But I agree with this approach if we end up needing to remove the name from multiple networks.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -118,14 +118,17 @@ class RouteManeuverViewController: UIViewController {
...
-            streetLabel.unabridgedText = destinations.prefix(min(streetLabelLines, destinations.count)).joined(separator: "\n")
+        
+        var stepRef: String?
+        if let step = step, let ref = step.codes?.first, ["i", "us"].contains(ref.lowercased().components(separatedBy: " ").first!) {
+            stepRef = ref
+        }
+        
+        if let destinations = step?.destinations {
+            let dest = destinations.prefix(min(streetLabelLines, destinations.count)).joined(separator: "\n")
+            streetLabel.unabridgedText = stepRef != nil ? "\(String(describing: stepRef)) - \(dest)" : dest
```
1ec5: This could result in a string like:

```
Optional(I) - Mobile
Biloxi
```

bsudekum:
@1ec5 I think this ready to go.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -108,10 +108,21 @@ class RouteManeuverViewController: UIViewController {
     }
     
     func updateStreetNameForStep() {
-        if let name = step?.names?.first {
-            destinationLabel.unabridgedText = name
-        } else if let destinations = step?.destinations {
-            destinationLabel.unabridgedText = destinations.prefix(min(numberOfDestinationLines, destinations.count)).joined(separator: "\n")
+        var stepRef: String?
+        if let step = step, let ref = step.codes?.first, ["i", "us"].contains(ref.lowercased().components(separatedBy: " ").first!) {
```
1ec5: We shouldn't remove the name from all U.S. routes. That would be a severe usability regression, because the vast majority of U.S. routes are surface streets, and in urban areas they're much more likely to be identified by name than by route number.

A proper solution to this problem appears to be coming soon in Project-OSRM/osrm-text-instructions#51, based on Project-OSRM/osrm-backend#4204. If we can't wait until then to solve the relatively few cases of U.S. routes on freeways, then we should only omit the name of a U.S. route if it contains the word "freeway", "expressway", or "highway".

ericrwolfe: > If we can't wait until then to solve the relatively few cases of U.S. routes on freeways, then we should only omit the name of a U.S. route if it contains the word "freeway", "expressway", or "highway".

I'm game for this approach 😃 @bsudekum.

bsudekum:
@1ec5 @ericrwolfe updated this, now:

* We prioritize destination over name
* If there is no destination but there is a name with one of the keywords, use the ref
* If not, use the name.

bsudekum:
I played around with getting the voice instructions to follow this logic as well. It looks like this will require a change to osrm-text-instructions.

bsudekum:
@1ec5 updated this to attempt to change the bottom label to be consistent with the maneuver text.

I'm finding it difficult to to go from `280` to `I-280` though.

MapboxNavigation/Highways.plist
```diff
@@ -0,0 +1,26 @@
...
+	<key>us-interstate-duplex</key>
+	<string>I</string>
+	<key>us-interstate-truck</key>
+	<string>I</string>
+	<key>us-highway</key>
+	<string>US</string>
+	<key>us-highway-duplex</key>
+	<string>US</string>
+	<key>us-highway-alternate</key>
+	<string>US</string>
```
1ec5: Given:

```
shield: us-highway-alternate
ref: 50
```

the label will incorrectly say “US 50”, whereas it would be more accurate to say “US 50 Alt” or “Alt US 50”. Same for the other banners below.

MapboxNavigation/String.swift
```diff
@@ -28,4 +25,10 @@ extension String {
             ("'", "&apos;")
             ])
     }
+
+    var isFreeway: Bool {
+        return self.components(separatedBy: " ").filter {
+            ["freeway", "expressway", "highway", "fwy", "hwy"].contains($0)
```
1ec5: The Mapbox Streets source abbreviates “expressway” as “Expy”.

MapboxNavigation/String.swift
```diff
@@ -28,4 +25,10 @@ extension String {
             ("'", "&apos;")
             ])
     }
+
+    var isFreeway: Bool {
+        return self.components(separatedBy: " ").filter {
```
1ec5: Nit: drop `self.`.

1ec5:
https://github.com/mapbox/mapbox-navigation-ios/pull/291#discussion_r126055522:

> A proper solution to this problem appears to be coming soon in Project-OSRM/osrm-text-instructions#51, based on Project-OSRM/osrm-backend#4204.

Underlying support for this feature is happening in mapbox/MapboxDirections.swift#154 and will also require changes to OSRM Text Instructions in Swift.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -336,18 +336,34 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
+                            }
                         } else {
                             currentName = nil
                         }
                     }
                 }
             }
             
-            if smallestLabelDistance < 5 && currentName != nil {
+            if smallestLabelDistance < 10 && currentName != nil {
```
1ec5: Is this change necessary?

bsudekum: Heh, nope.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -336,18 +336,34 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
-                        } else if let line = feature as? MGLMultiPolylineFeature, let name = line.attribute(forKey: "name") as? String {
-                            currentName = name
+                        if let line = feature as? MGLPolylineFeature {
+                            if let name = line.attribute(forKey: "name") as? String, !name.isFreeway {
+                                currentName = name
+                            } else if let ref = line.attribute(forKey: "ref") as? String, let shield = line.attribute(forKey: "shield") as? String {
+                                if let highwayName = HighwayNamesByPrefix[shield] {
+                                    currentName = "\(highwayName) \(ref)"
+                                } else {
+                                    currentName = ref
```
1ec5: For any country outside the U.S., this logic would display only the number. Even inside the U.S., something like a county road would get just the number, which is opposite what I would expect. I think it would be preferable in all these cases to either show the name and number or to show nothing at all.

1ec5:
#410 is an alternative to this PR, at least as far as the turn banner is concerned. We can pick this PR back up later to deal with the current road name label, but I don’t think it’s a huge deal compared to the turn banner, since the base map also displays freeway names.

bsudekum:
Let's close this out. A followup PR will most likely go in another direction than the one started here.

PR #294, opened 2017-06-20
Update distance

frederoni:
Keep track of the current maneuver page so we update distance on the correct controller.
Also removed an excessive call to `willTransitionTo`.

@bsudekum 👀

### DIFF
--- a/MapboxNavigation/RoutePageViewController.swift
+++ b/MapboxNavigation/RoutePageViewController.swift
@@ -73,7 +73,7 @@ extension RoutePageViewController: UIPageViewControllerDataSource, UIPageViewCon
     
     func pageViewController(_ pageViewController: UIPageViewController, didFinishAnimating finished: Bool, previousViewControllers: [UIViewController], transitionCompleted completed: Bool) {
         if completed {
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RoutePageViewController.swift
```diff
@@ -73,7 +73,7 @@ extension RoutePageViewController: UIPageViewControllerDataSource, UIPageViewCon
     
     func pageViewController(_ pageViewController: UIPageViewController, didFinishAnimating finished: Bool, previousViewControllers: [UIViewController], transitionCompleted completed: Bool) {
         if completed {
-            if let controller = previousViewControllers.first as? RouteManeuverViewController {
+            if let controller = pageViewController.viewControllers?.last as? RouteManeuverViewController {
```
frederoni: This seems to be the root cause of all our problems with the page view controller.

bsudekum: Thats funny, I thought the count here is always 1.

frederoni: first and last is the same here but the issue was `previousViewControllers`

Issue #286, opened 2017-06-19
Polly pronounces “S” in ref as “South”


freenerd:
In  Saxony, Germany there are roads like `S 81`. [Wikipedia has a list](https://de.wikipedia.org/wiki/Liste_der_Staatsstra%C3%9Fen_in_Sachsen) and here are some on the [map](http://map.project-osrm.org/?z=14&center=51.228899%2C13.558288&loc=51.234192%2C13.540306&loc=51.233279%2C13.557472&hl=en&alt=0
). 

Right now these are extended to `South 81` instead of staying `S 81` (via [this code](https://github.com/mapbox/mapbox-navigation-ios/blob/5d78f52e52d2e1e847336890a44bfaae6505f653/MapboxNavigation/Resources/Base.lproj/Abbreviations.csv.plist#L73
)).

This is definitely a fringe problem (only 5000 km of roads in Germany have this abbreviation) but highlights issues with using abbreviation expansion of one place in another.

/cc @1ec5 @bsudekum

1ec5:
Do you have a screenshot of the problem? Abbreviations.plist is used for _abbreviating_, not for expanding abbreviations, and only on the road name (after any `ref` has been stripped away). The expected behavior is that “South” in a `name` tag becomes “S”, but “S” in a `ref` tag should remain unchanged.

On the other hand, there is a geographical issue with Shields.plist, where [`CA`](https://github.com/mapbox/mapbox-navigation-ios/blob/1677c6479a2a293172a4a4d1e3ea45e5c1e3707b/MapboxNavigation/Resources/Shields.plist#L29-L30) may denote California or Cantabría, or [`NH`](https://github.com/mapbox/mapbox-navigation-ios/blob/1677c6479a2a293172a4a4d1e3ea45e5c1e3707b/MapboxNavigation/Resources/Shields.plist#L83-L84) New Hampshire or National Highway (e.g., in India), or [`CO`](https://github.com/mapbox/mapbox-navigation-ios/blob/1677c6479a2a293172a4a4d1e3ea45e5c1e3707b/MapboxNavigation/Resources/Shields.plist#L31-L32) Colorado or “county”. Unfortunately, there isn’t a better way to distinguish shields internationally until either Project-OSRM/osrm-backend#482 or Project-OSRM/osrm-backend#333 is implemented.

freenerd:
I'm pretty sure the voice instruction said "South 94"

[...]

PR #297, opened 2017-06-20
Store current road details in map view controller

1ec5:
RouteMapViewController now stores the current name and classification of the current road, based on the same process by which the current road name label is populated. Eventually, perhaps RouteVoiceController could somehow gain access to this information so that it could word and time voice announcements differently on freeways versus surface streets.

At first glance, this might also look like a good solution to https://github.com/mapbox/mapbox-navigation-ios/pull/291#discussion_r123060422, Project-OSRM/osrm-text-instructions#51, and https://github.com/mapbox/mapbox-navigation-ios/issues/286#issuecomment-309649781. However, this feature querying mechanism only gives us information about the road the user is currently traveling on, not the one the user needs to turn onto.

/cc @bsudekum @frederoni

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -43,6 +43,17 @@ class RouteMapViewController: UIViewController {
         }
     }
     weak var delegate: RouteMapViewControllerDelegate?
+    
+    var currentRoadName: String? {
+        didSet {
+            if let currentRoadName = currentRoadName {
+                wayNameLabel.text = currentRoadName
+                wayNameView.isHidden = false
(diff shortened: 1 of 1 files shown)


1ec5:
In order to use this mechanism for the upcoming road, we’d need to somehow load the tile around the maneuver point off-screen and query that tile for a matching road. It seems like overkill, but the [Maps API’s tile query method](https://www.mapbox.com/api-documentation/#retrieve-features-from-vector-tiles) would give us what we need here.

ericrwolfe:
Tagged as experimental and closing for now. Will revisit when we explore implementing custom navigation road labels.

PR #300, opened 2017-06-21
Revert https://github.com/mapbox/mapbox-navigation-ios/pull/296/

bsudekum:
Reverts https://github.com/mapbox/mapbox-navigation-ios/pull/296

We need to take a look at this in depth later on. With #296, the user never enters `.followWithCourse` mode.

Also added some camera changes so you're facing the right direction when the view loads.

/cc @frederoni

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -82,15 +82,15 @@ class RouteMapViewController: UIViewController {
         } else {
             setDefaultCamera(animated: false)
         }
-        
-        mapView.setUserLocationVerticalAlignment(.bottom, animated: false)
-        mapView.setUserTrackingMode(.followWithCourse, animated: false)
 
         UIDevice.current.addObserver(self, forKeyPath: "batteryState", options: .initial, context: nil)
     }
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+        
+        mapView.setUserTrackingMode(.followWithCourse, animated: false)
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -82,15 +82,15 @@ class RouteMapViewController: UIViewController {
...
-        mapView.setUserTrackingMode(.followWithCourse, animated: false)
 
         UIDevice.current.addObserver(self, forKeyPath: "batteryState", options: .initial, context: nil)
     }
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+        
+        mapView.setUserTrackingMode(.followWithCourse, animated: false)
+        mapView.setUserLocationVerticalAlignment(.bottom, animated: false)
```
1ec5: So we can’t set the vertical alignment in `viewDidLoad` but the tracking mode in `viewDidAppear`?

1ec5: Is this a map SDK limitation? If so, can you whip up a reduced test case and file an issue in gl-native? Thanks!

ericrwolfe:
Somewhat related, it looks like moving
```swift
if simulatesLocationUpdates {
     mapView.locationManager.stopUpdatingLocation()
     mapView.locationManager.stopUpdatingHeading()
}
```
from `viewDidAppear` to `viewWillAppear` has fixed the main issues with user puck hovering all over the place during route simulation. There are big caveats to logic within the view lifecycle events (see https://github.com/mapbox/mapbox-navigation-ios/issues/303).

PR #301, opened 2017-06-21
Separate custom UI example

bsudekum:
simulator screen shot jun 21 2017 2 59 17 pm

todo:
- [ ] figure out tab bar
  * toggle simulation in both examples
  * icons
  * setting selected


Examples/Swift/CustomNavigationUI.swift
```diff
@@ -0,0 +1,170 @@
+import UIKit
+import MapboxCoreNavigation
+import Mapbox
+import CoreLocation
+import AVFoundation
+import MapboxDirections
+
+class CustomNavigationUI: UIViewController, MGLMapViewDelegate, AVSpeechSynthesizerDelegate {
```
frederoni: This class name should end with `ViewController`, however, `CustomNavigationViewController` sounds like it's a subclass of `NavigationViewController` which is misleading.

ericrwolfe:
@bsudekum I've picked this up and will continue work on this tomorrow.

ericrwolfe:
@bsudekum or @frederoni want to take a 👀 when you have a chance?

Examples/Swift/CustomViewController.swift
```diff
@@ -0,0 +1,195 @@
...
+    func alertLevelDidChange(_ notification: NSNotification) {
+        let routeProgress = notification.userInfo![RouteControllerAlertLevelDidChangeNotificationRouteProgressKey] as! RouteProgress
+        let alertLevel = routeProgress.currentLegProgress.alertUserLevel
+        var text: String
+
+        let distance = routeProgress.currentLegProgress.currentStepProgress.distanceRemaining
+        if let upComingStep = routeProgress.currentLegProgress.upComingStep {
+            // Don't give full instruction with distance if the alert type is high
+            if alertLevel == .high {
+            text = upComingStep.instructions
```
bsudekum: nit, indent off here and line `#74`

Examples/Swift/CustomViewController.swift
```diff
@@ -0,0 +1,195 @@
...
+            completion?()
+            self?.addRouteToMap()
+        }
+    }
+
+    func addRouteToMap() {
+        guard let style = mapView.style else { return }
+        guard let userRoute = userRoute else { return }
+
+        mapView.addAnnotation(destination)
```
bsudekum: We'll also want to clear the map of all annotations before adding a new destination.

Examples/Swift/CustomViewController.swift
```diff
@@ -0,0 +1,195 @@
...
+            for layer in style.layers.reversed() {
+                if !(layer is MGLSymbolStyleLayer) {
+                    style.insertLayer(line, above: layer)
+                    break
+                }
+            }
+        }
+    }
+
+    func speak(_ text: String) {
```
bsudekum: Since this is only used in one place, I think we can remove this function and move it up to where `speak()` is called.

Examples/Swift/ViewController.swift
```diff
@@ -45,220 +60,203 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
...
+            
+        present(navigationViewController, animated: true, completion: nil)
     }
     
-    /// Delegate method for changing the route line style
-    func navigationMapView(_ mapView: NavigationMapView, routeStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer? {
-        let lineCasing = MGLLineStyleLayer(identifier: identifier, source: source)
+    // MARK: - Navigation with multiple waypoints
+
+    func startMultipleWaypoints() {
```
bsudekum: Since waypoints is coming through the pipeline in https://github.com/mapbox/mapbox-navigation-ios/pull/270, do you think we should hold off on showing developers this method for doing multi-leg routes? It's slightly an antipattern.

Examples/Swift/ViewController.swift
```diff
@@ -45,220 +60,203 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
...
-        let camera = mapView.camera
-        camera.pitch = 45
-        camera.altitude = 600
-        if let userLocation = mapView.userLocation {
-            camera.centerCoordinate = userLocation.coordinate
-            if let location = userLocation.location {
-                camera.heading = location.course
-            }
-        }
-        navigationViewController.pendingCamera = camera
```
bsudekum: Setting the initial camera should be added back

ericrwolfe: Is there a reason this isn't handled by the SDK by default? NavigationViewController should automatically set the pending camera to the currentLocation || route origin's location and initial bearing.

bsudekum: Yeah this should be possible.

Examples/Swift/ViewController.swift
ericrwolfe: Let's put this in for now for customers who are asking about multiple destinations. 

This example serves both to show how to handle multiple destinations, but also how to display a custom UI over navigation and how to update the route on the fly.

Once this is in master, can you rebase #270 and update the example with how to implement the same concept with waypoints?

Examples/Swift/ViewController.swift
```diff
@@ -45,220 +60,203 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
...
+        let options = RouteOptions(coordinates: [
+            mapView.userLocation!.coordinate,
+            destination.coordinate,
+        ])
         options.includesSteps = true
         options.routeShapeResolution = .full
         options.profileIdentifier = .automobileAvoidingTraffic
         
-        _ = Directions.shared.calculate(options) { [weak self] (waypoints, routes, error) in
+        _ = Directions.shared.calculate(options) { (waypoints, routes, error) in
```
1ec5: Out of curiosity, why has self gone from weak to strong here?

Examples/Swift/ViewController.swift
```diff
@@ -45,220 +60,203 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
...
     }
     
-    func startNavigation(along route: Route, simulatesLocationUpdates: Bool = false) {
-        // Pass through:
-        // 1. The route the user will take.
-        // 2. A `Directions` object, used for rerouting.
+    // MARK: - Basic Navigation
+    
+    func startBasicNavigation() {
+        guard let route = self.currentRoute else { return }
```
1ec5: Nit: drop `self.`.

Examples/Swift/ViewController.swift
1ec5: `RouteMapViewController.viewWillAppear(_:)` falls back to calling `setDefaultCamera(_:)`, which only sets the altitude and pitch. It would be feasible for this method to also set the center coordinate to the route’s initial location and heading. However, it wouldn’t be feasible to set the camera to the user’s current location right off the bat, because the location manager determines the current location asynchronously.

Before this PR lands, we should modify `setDefaultCamera(_:)` to account for the route’s origin.

Examples/Swift/ViewController.swift
```diff
@@ -45,220 +60,203 @@ class ViewController: UIViewController, MGLMapViewDelegate, NavigationViewContro
...
+        guard let navigationViewController = self.presentedViewController as? NavigationViewController else { return }
+
+        // Calculate directions to the next waypoint
+        let options = RouteOptions(coordinates: [
+            navigationViewController.mapView!.userLocation!.coordinate,
+            nextDestination,
+        ])
+        options.includesSteps = true
+        options.routeShapeResolution = .full
+        options.profileIdentifier = .automobileAvoidingTraffic
```
1ec5: Can we get these details from the previous route’s `routeOptions`? That way, if the profile happens to be `.cycling`, it’ll automatically match when going to the next leg of the trip.

Examples/Swift/ViewController.swift
ericrwolfe: Oh, yeah should [weak self] since Directions.shared is a singleton.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -34,6 +34,14 @@ class RouteMapViewController: UIViewController {
         }
         return parent.pendingCamera
     }
+    var defaultCamera: MGLMapCamera {
```
1ec5: This is more like `tiltedCamera` or `cameraFacingRoad`, since it’s a dynamic value rather than a fixed value.

PR #302, opened 2017-06-21
Change route on the fly

frederoni:
This PR makes it a bit more convenient to change the route on the fly by notifying the `RouteMapViewController` and the `RouteTableViewController` about the update along with initializing a `RouteProgress` for the new route.

@bsudekum @ericrwolfe 👀


MapboxNavigation/NavigationViewController.swift
```diff
@@ -369,6 +371,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
             }
         }
         
+        if route != routeController.routeProgress.route {
+            routeController.routeProgress = RouteProgress(route: route)
+            routeController.routeProgress.currentLegProgress.stepIndex = 0
```
bsudekum: This is set to 0 when reseting the route progress.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -369,6 +371,10 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
             }
         }
         
+        if route != routeController.routeProgress.route {
```
frederoni: Is this check safe when we don't implement `-hash`, `-isEqual:` or an overloaded operator?

bsudekum: Do we actually even need to check if theyre different?

frederoni: We can, but it should then be moved to the route’s didSet to avoid initializing an excessive RouteProgress.

1ec5: If we don't override isEqual, it ends up doing a pointer equality check. That won't be as thorough.

MapboxNavigationTests/Fixtures/Fixture.swift
```diff
@@ -31,4 +33,22 @@ internal class Fixture {
             return [:]
         }
     }
+    
+    class func downloadRouteFixture(coordinates: [CLLocationCoordinate2D], fileName: String, completion: @escaping () -> Void) {
```
1ec5: We could put this behind an #if to prevent ourselves from checking in tests that call this function.

PR #304, opened 2017-06-22
Add telemetry events

bsudekum:
~~Telem events manager needs to be made public first: https://github.com/mapbox/mapbox-gl-native/tree/bs-telem-api~~ -> https://github.com/mapbox/mapbox-telemetry-ios

This PR adds the ground work for telemetry events and also adds the first event, rerouting. On every reroute, an event will be pushed to the event manager containing the last 20 locations prior to the reroute and up to 20 locations after the reroute. The event manager in the iOS SDK will then push all events in a single network request to the server.

The goal here is help improve all aspects of the navigation SKD and also the directions API which this SDK relies on. 


MapboxCoreNavigation/RouteController.swift
```diff
@@ -393,4 +464,20 @@ extension RouteController: CLLocationManagerDelegate {
...
         incrementRouteProgress(alertLevel, location: location, updateStepIndex: updateStepIndex)
     }
+    
+    func convertLocastionsObject(locations: [CLLocation]) -> [[String: Any]] {
+        return locations.map { (location: CLLocation) -> [String : Any] in
+            var locationDictionary:[String: Any] = [:]
+            locationDictionary["lat"] = location.coordinate.latitude
+            locationDictionary["lng"] = location.coordinate.latitude
+            locationDictionary["altitude"] = location.altitude
+            locationDictionary["timestamp"] = location.timestamp
```
bsudekum: For some reason, I'm unable to todo `location.timestamp.ISO8601` here even though `location.timestamp` is of type `Date`

MapboxCoreNavigation/MGLMapboxEvents.swift
```diff
@@ -0,0 +1,39 @@
...
+import Polyline
+
+extension MGLMapboxEvents {
+    class func addDefaultEvents(routeProgress: RouteProgress, sessionIdentifier: UUID) -> [String: Any] {
+        var modifiedEventDictionary: [String: Any] = [:]
+    
+        modifiedEventDictionary["platform"] = UIDevice.current.systemName
+        modifiedEventDictionary["operatingSystemVersion"] = "\(UIDevice.current.systemName)-\(UIDevice.current.systemVersion)"
+        modifiedEventDictionary["sdkIdentifier"] = "mapbox-navigation-ios"
+        modifiedEventDictionary["sdkVersion"] = "0.4.0"
```
bsudekum: There should be a better way to do this.

MapboxCoreNavigation/RouteController.swift
frederoni: Drop static from ISO8601 and return `formatter.string(from: self)` should work.

MapboxCoreNavigation/MGLMapboxEvents.swift
frederoni: Something like: `Bundle.navigationUI.object(forInfoDictionaryKey: "CFBundleShortVersionString")`. There should be a constant for CFBundleShort… IIRC

MapboxCoreNavigation/MGLMapboxEvents.swift
```diff
@@ -0,0 +1,39 @@
+import Mapbox
+import Polyline
+
+extension MGLMapboxEvents {
+    class func addDefaultEvents(routeProgress: RouteProgress, sessionIdentifier: UUID) -> [String: Any] {
+        var modifiedEventDictionary: [String: Any] = [:]
+    
+        modifiedEventDictionary["platform"] = UIDevice.current.systemName
```
frederoni: UIDevice should be part of UIKit which this library doesn't depend on. 😕
We should adapt this for watchOS later.

bsudekum: Are there any alternatives?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -161,6 +184,20 @@ extension RouteController: CLLocationManagerDelegate {
             return
         }
         
+        userLocationForRerouteEvent.append(location)
+        // Keep double the amount needed for reroute event since we keep collecting after the reroute
+        if userLocationForRerouteEvent.count >= rerouteLocationArraySize * 20 {
```
frederoni: Should this be `* 2`?

bsudekum: Oh yep

MapboxCoreNavigation/MGLMapboxEvents.swift
ericrwolfe: We can wrap this in a compiler flag by platform and by default we can set this to `unknown`

E.g.
```swift
#if os(iOS)
        UIApplication.shared.isIdleTimerDisabled = true
#endif
```

MapboxCoreNavigation/MGLMapboxEvents.swift
```diff
@@ -0,0 +1,39 @@
...
+        modifiedEventDictionary["operatingSystemVersion"] = "\(UIDevice.current.systemName)-\(UIDevice.current.systemVersion)"
+        modifiedEventDictionary["sdkIdentifier"] = "mapbox-navigation-ios"
```
ericrwolfe: Would there be an easy way to differentiate this between MapboxCoreNavigation (`mapbox-navigation-ios`) and MapboxNavigation (`mapbox-navigation-ios-ui`)?

MapboxCoreNavigation/MGLMapboxEvents.swift
```diff
@@ -0,0 +1,39 @@
...
+        modifiedEventDictionary["screenBrightness"] = UIScreen.main.brightness
+        
+        return modifiedEventDictionary
+    }
+    
+    class func addEventPrefix(eventDictionary: [String: Any], eventPrefix: String) -> [String: Any] {
+        var newDictionary: [String: Any] = [:]
+        
+        for key in Array(eventDictionary.keys) {
+            newDictionary["\(eventPrefix)"] = eventDictionary[key]
```
ericrwolfe: I'm not sure I follow the purpose or logic of this function? It looks like here the resulting dictionary will always have only a single key:
```swift
[
    "eventPrefix": <random value>
]
```

MapboxCoreNavigation/RouteController.swift
```diff
@@ -118,8 +115,34 @@ open class RouteController: NSObject {
...
+    var routeTask: URLSessionDataTask?
+    
+    public var userLocationForRerouteEvent: [CLLocation] = []
+    var rerouteIndexLocationBefore = 0
+    var shouldPushRerouteEventEventually = false
+    var rerouteLocationArraySize = 20
+    var countdownToPushEvent = 20
+    var distanceCompleted: CLLocationDistance = 0
+    var distanceRemaining: CLLocationDistance?
+    var durationRemaining: TimeInterval?
```
ericrwolfe: Should we rename these to `lastRerouteDistanceRemaining` or something to clarify that these properties are stored state from the last reroute? Or should we put these all into a single struct property for the last reroute state?

Something like
```swift
struct LastRerouteState {
    var distanceRemaining: CLLocationDistance
    var durationRemaining: CLLocationDistance
    ...etc...
}

File: MapboxCoreNavigation/RouteController.swift [hunk 8]
```diff
@@ -216,6 +253,27 @@ extension RouteController: CLLocationManagerDelegate {
         lastUserDistanceToStartOfRoute = Double.infinity
     }
     
+    func pushAndResetLocationCounters() {
+        var eventDictionary = MGLMapboxEvents.addDefaultEvents(routeProgress: routeProgress, sessionIdentifier: sessionIdentifier)
+        eventDictionary["startTimestamp"] = startTimestamp
+        eventDictionary["feedbackType"] = "reroute"
+        eventDictionary["locationsBefore"] = convertLocastionsObject(locations: Array(userLocationForRerouteEvent.prefix(rerouteIndexLocationBefore)))
+        eventDictionary["locationsAfter"] = convertLocastionsObject(locations: Array(userLocationForRerouteEvent.suffix(from: rerouteIndexLocationBefore)))
+        eventDictionary["userLocatioBeforeAfterReroute"] = userLocationForRerouteEvent
```
ericrwolfe: This property isn't in the spec so the event may be discarded by the API. What's the thought behind including this?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -216,6 +253,27 @@ extension RouteController: CLLocationManagerDelegate {
...
+        eventDictionary["distanceCompleted"] = distanceCompleted
+        eventDictionary["distanceRemaining"] = distanceRemaining ?? nil
+        eventDictionary["durationRemaining"] = durationRemaining ?? nil
+        
+        let eventDictionaryWithPrefix = MGLMapboxEvents.addEventPrefix(eventDictionary: eventDictionary, eventPrefix: "navigation.reroute")
```
ericrwolfe: We shouldn't need to prefix the event attribute keys with the event name.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -393,4 +464,20 @@ extension RouteController: CLLocationManagerDelegate {
         
         incrementRouteProgress(alertLevel, location: location, updateStepIndex: updateStepIndex)
     }
+    
+    func convertLocastionsObject(locations: [CLLocation]) -> [[String: Any]] {
```
ericrwolfe: Typo `convertLocastionsObject`

MapboxCoreNavigation/MGLMapboxEvents.swift
1ec5: This library is only for iOS, so `#if os(iOS)` is unnecessary. The issue is actually that Core Navigation avoids depending on UIKit. We should add a hook for MapboxNavigation to add to this dictionary.

MapboxCoreNavigation/RouteController.swift
bsudekum: Oh this is old code prior to spec outline.

MapboxCoreNavigation/MGLMapboxEvents.swift
bsudekum: This gets tricky. We could pass the event object along in the didreroute method and then have UI send off the event. But it's impossible for core to know whether UI is actually used. I think we might have to punt on this for now.

MapboxCoreNavigation/String.swift
```diff
@@ -0,0 +1,20 @@
+import UIKit
```
bsudekum: @frederoni you think this okay? I can't use `UIDevice.current.systemName` without it. You think hardcoding strings would be okay here?

MapboxCoreNavigation/MGLMapboxEvents.swift
```diff
@@ -0,0 +1,30 @@
+import Mapbox
+import Polyline
+
+extension MGLMapboxEvents {
+    public class func addDefaultEvents(routeProgress: RouteProgress, sessionIdentifier: UUID, sessionNumberOfReroutes: Int = 0) -> [String: Any] {
```
bsudekum: Is there any way we can share this function with MapboxNavigation without making it public? We need it for `navigation.cancel`

MapboxCoreNavigation/RouteController.swift
```diff
@@ -118,8 +115,45 @@ open class RouteController: NSObject {
         }
     }
     
+    /**
+     IOS 8601 timestamp of when the `RouteController` was initialized.
+     */
+    public let sessionStartTimestamp: String = Date().ISO8601
```
bsudekum: Same question as in https://github.com/mapbox/mapbox-navigation-ios/pull/304/files#r123724999, can make this sharable with MapboxNavigation but not public?

bsudekum:
Added remaining events, cancel, arrive, depart in https://github.com/mapbox/mapbox-navigation-ios/pull/304/commits/9a65dbf2e2e87a55d76e617a49db4d3f2e385a26

MapboxCoreNavigation/String.swift
frederoni: You should do something like:
```swift
#if os(iOS)
import UIKit
#elseif os(watchOS)
import WatchKit
```

However, that could be tail work or part of https://github.com/mapbox/mapbox-navigation-ios/issues/305

frederoni:
Looks good. It's mostly the private communication between core and the ui that we probably should solve. Perhaps a test case to verify the circular buffer and break out the collection of locations from `locationManager(_:didUpdateLocations:)` into a separate function (`enqueue(location:)`).

MapboxCoreNavigation/MGLMapboxEvents.swift
1ec5: It isn’t impossible: see the ideas in mapbox/MapboxDirections.swift#136, specifically `NSClassFromString(_:)` and `Bundle(for:)`.

1ec5: Even more straightforward would be for Core Navigation to expose a hook for MapboxNavigation to tell Core Navigation that it’s present.

MapboxCoreNavigation/MGLMapboxEvents.swift
1ec5: Not really, because MapboxCoreNavigation and MapboxNavigation are separate modules. If you were to declare this method in an Objective-C header but implement it in Swift, it would be possible to give the header “private” membership in the MapboxCoreNavigation target, which would make it accessible to other targets in the same Xcode project, like MapboxNavigation. At least that’s true in Xcode and Carthage land. I don’t think that would work with CocoaPods.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -444,3 +532,121 @@ extension RouteController: CLLocationManagerDelegate {
...
+        if let index = outstandingFeedbackEvents.index(of: event) {
+            outstandingFeedbackEvents.remove(at: index)
+        }
+        
+        let eventName = event.eventDictionary["event"] as! String
+        
+        NSLog("Sending \(eventName)")
+
+        event.eventDictionary["locationsBefore"] = sessionState.pastLocations.allObjects.filter({$0.timestamp <= event.timestamp }).map({$0.dictionary})
+        event.eventDictionary["locationsAfter"] = sessionState.pastLocations.allObjects.filter({$0.timestamp > event.timestamp }).map({$0.dictionary})
```
1ec5: Nit: Use trailing closure syntax.

MapboxCoreNavigation/MGLMapboxEvents.swift
```diff
@@ -0,0 +1,165 @@
...
+    }
+    
+    public var allObjects: Array<T> {
+        get {
+            return Array(objects)
+        }
+    }
+}
+
+class CoreFeedbackEvent: Hashable {
```
1ec5: If neither this class nor any of its subclasses need to be public (to Objective-C code), we should make them structs. Swift will synthesize initializers for these structs, they’ll take up [less memory](https://developer.apple.com/videos/play/wwdc2016/416/), and they’re just easier to work with than value classes.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -444,3 +532,121 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let eventName = "navigation.reroute"
+
+        let timestamp = Date()
+        
+        var eventDictionary = MGLMapboxEvents.addDefaultEvents(routeController: self)
+        eventDictionary["event"] = eventName
+        
+        eventDictionary["secondsSinceLastReroute"] = sessionState.lastReroute != nil ? round(timestamp.timeIntervalSince(sessionState.lastReroute!)) : -1
+        
+        // These are placeholders until the
```
1ec5: Nit: trailing thought. 💭

MapboxCoreNavigation/RouteController.swift
```diff
@@ -157,6 +192,53 @@ open class RouteController: NSObject {
...
+        enqueueRerouteEvent()
+    }
+    
+    func didReroute(notification: NSNotification) {
+        let route = routeProgress.route
+        if let lastReroute = outstandingFeedbackEvents.filter({$0 is RerouteEvent }).last {
+            if let geometry = route.coordinates {
+                lastReroute.eventDictionary["newGeometry"] = Polyline(coordinates: geometry).encodedPolyline
+                lastReroute.eventDictionary["newDistanceRemaining"] = round(route.distance)
+                lastReroute.eventDictionary["newDurationRemaining"] = round(route.expectedTravelTime)
```
1ec5: A `[String: Any]` using string literals for the keys can be rather fragile, especially for code that lacks unit tests. Let’s replace `eventDictionary` with strongly-typed properties for each of the expected fields, then add a method to CoreFeedbackEvent or RerouteEvent that converts to an attributes dictionary.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -444,3 +532,121 @@ extension RouteController: CLLocationManagerDelegate {
...
+    
+    func sendFeedbackEvent(event: CoreFeedbackEvent) {
+        // remove from outstanding event queue
+        if let index = outstandingFeedbackEvents.index(of: event) {
+            outstandingFeedbackEvents.remove(at: index)
+        }
+        
+        let eventName = event.eventDictionary["event"] as! String
+        
+        NSLog("Sending \(eventName)")
```
1ec5: Nit: `print()` is more idiomatic than `NSLog()` in Swift.

1ec5: Remove any print statements before merging. If these print statements are really needed for debugging, we should make them debug-only, either using a compiler flag or perhaps a user default.

MapboxCoreNavigation/String.swift
```diff
@@ -0,0 +1,24 @@
+#if os(iOS) || os(tvOS)
+    import UIKit
+#elseif os(watchOS)
+    import WatchKit
+#endif
+
+extension String {
+    static var systemName: String {
+        #if os(iOS) || os(tvOS)
+            return UIDevice.current.systemName
```
1ec5: Doesn’t this just return “iOS” or “tvOS”?

MapboxCoreNavigation/MGLMapboxEvents.swift
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+extension UIDevice {
+    var machine: String {
+        get {
+            var systemInfo = utsname()
+            uname(&systemInfo)
+            let machineMirror = Mirror(reflecting: systemInfo.machine)
+            let identifier = machineMirror.children.reduce("") { identifier, element in
```
1ec5: How does this dark magic differ from [the telemetry library implementation](https://github.com/mapbox/mapbox-telemetry-ios/blob/2b905791e44b1eaae025569a5d6f1b73d23715b5/MapboxMobileMetrics/MapboxMobileEvents/MMECommonEventData.m#L14), which is also currently used in the map SDK?

MapboxCoreNavigation/MGLMapboxEvents.swift
ericrwolfe: CoreFeedbackEvent was originally a struct until we split out feedback into two separate event types. Structs don't support inheritance, thus the switch.

MapboxCoreNavigation/MGLMapboxEvents.swift
ericrwolfe: It's basically the same. Many of these duplicated class extensions are just stopgaps until we can leverage the full standalone telemetry framework.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -444,3 +537,119 @@ extension RouteController: CLLocationManagerDelegate {
...
+    func sendFeedbackEvent(event: CoreFeedbackEvent) {
+        // remove from outstanding event queue
+        if let index = outstandingFeedbackEvents.index(of: event) {
+            outstandingFeedbackEvents.remove(at: index)
+        }
+        
+        let eventName = event.eventDictionary["event"] as! String
+        
+        event.eventDictionary["locationsBefore"] = sessionState.pastLocations.allObjects.filter({$0.timestamp <= event.timestamp }).map {$0.dictionary}
+        event.eventDictionary["locationsAfter"] = sessionState.pastLocations.allObjects.filter({$0.timestamp > event.timestamp }).map {$0.dictionary}
```
frederoni: Nit: filter is also a candidate for trailing closure syntax

MapboxCoreNavigation/MGLMapboxEvents.swift
1ec5: Ah, in that case protocols are typically used for code sharing between struct types. However, there’s no structural difference between the two subclasses. If we expect the two types of events to diverge in the future, we can use an [enumeration with associated values](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Enumerations.html#//apple_ref/doc/uid/TP40014097-CH12-ID148) instead of a struct. Otherwise, a simple enumeration-typed property inside the struct would suffice.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -649,4 +646,10 @@ extension RouteController {
             sendFeedbackEvent(event: event)
         }
     }
+    
+    func resetSession() {
+        sessionState = SessionState()
+        self.sessionState.originalRoute = routeProgress.route
+        self.sessionState.currentRoute = routeProgress.route
```
1ec5: Nit: `self.` is unnecessary.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -134,10 +154,30 @@ open class RouteController: NSObject {
...
     
     deinit {
         suspendLocationUpdates()
+        checkAndSendOutstandingFeedbackEvents(forceAll: true)
+        sendCancelEvent()
+        suspendNotifications()
+    }
+    
+    func resumeNotifications() {
+        UIDevice.current.isBatteryMonitoringEnabled = true
```
1ec5: This appears to be triggering #688 somehow.

MapboxCoreNavigation/Feedback.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+public enum FeedbackType: String {
+    case general = "general"
```
bsudekum: Since this enum will not public, is it required to make it bridge-compatible?

1ec5: It is `public` – should it be?

MapboxCoreNavigation/RouteController.swift
boundsj: Would it be possible to add some tests for at least the event generation? The spec may change a lot still so relying on tests may actually make more sense at this point (or always).

MapboxCoreNavigation/RouteController.swift
```diff
@@ -142,10 +166,33 @@ open class RouteController: NSObject {
         super.init()
         
         self.locationManager.delegate = self
+        
+        self.resetSession()
+        
+        self.resumeNotifications()
+        
+        if let accessToken = MGLAccountManager.accessToken() {
```
1ec5: Is this the only reason Core Navigation depends on the map SDK? We should instead look directly at `MGLMapboxAccessToken` in Info.plist. There’s a chance the developer set the map access token using `MGLAccountManager.setAccessToken(_:)`, which means `MGLMapboxAccessToken` is `nil`. Regardless, I think it’s more important for directions-related telemetry to use the same access token as MapboxDirections.swift (`Directions.accessToken`), which isn’t necessarily the same access token that MGLAccountManager sees.

MapboxCoreNavigation/Feedback.swift
1ec5: It’s currently impossible for an Objective-C application to call `RouteController.recordFeedback(type:description:)` et al., because this public type doesn’t bridge to Objective-C.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,164 @@
+import Polyline
+import MapboxDirections
+import AVFoundation
+import MapboxMobileEvents
+
+let SECONDS_FOR_COLLECTION_AFTER_FEEDBACK_EVENT: TimeInterval = 20
```
1ec5: Use CamelCase, not UPPER_SNAKE_CASE, for global constants in Swift.

MapboxCoreNavigation/String.swift
```diff
@@ -0,0 +1,24 @@
+#if os(iOS) || os(tvOS)
+    import UIKit
+#elseif os(watchOS)
+    import WatchKit
+#endif
+
+extension String {
+    static var systemName: String {
```
1ec5: These properties shouldn’t extend String just because they’re of type String. ProcessInfo would be a better type to extend.

MapboxCoreNavigation/RouteController.swift
1ec5: Using stronger typing with `FeedbackEvent` will ensure better test coverage. In essence, we need to centralize the logic that converts from native types like Polyline into JSON types, rather than scattering that logic in multiple places, as is the case right now.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,164 @@
...
+                return identifier + String(UnicodeScalar(UInt8(value)))
+            }
+            
+            return identifier
+        }
+    }
+}
+
+extension CLLocation {
+    var dictionary: [String: Any] {
```
1ec5: This property needs a more specific name: `eventDictionary` would be better.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,164 @@
...
+        modifiedEventDictionary["applicationState"] = UIApplication.shared.applicationState.telemString
+        
+        //modifiedEventDictionary["connectivity"] = ??
+        
+        return modifiedEventDictionary
+    }
+}
+
+extension UIApplicationState {
+    var telemString: String {
```
1ec5: Nit: spell out “telemetry”.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,164 @@
...
+                return "Inactive"
+            case .background:
+                return "Background"
+            }
+        }
+    }
+}
+
+extension UIDevice {
+    var machine: String {
```
1ec5: Per mapbox/MapboxStatic.swift#66, any extension of a built-in type should be `@nonobjc` to avoid conflicts with third-party libraries.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,164 @@
...
+        get {
+            var systemInfo = utsname()
+            uname(&systemInfo)
+            let machineMirror = Mirror(reflecting: systemInfo.machine)
+            let identifier = machineMirror.children.reduce("") { identifier, element in
```
1ec5: `utsname.machine` is a C-style `char[256]`. Is this the preferred way to convert that type into a String?

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,164 @@
...
+            locationDictionary["horizontalAccuracy"] = horizontalAccuracy
+            locationDictionary["verticalAccuracy"] = verticalAccuracy
+            locationDictionary["course"] = course
+            locationDictionary["speed"] = speed
+            return locationDictionary
+        }
+    }
+}
+
+class FixedLengthBuffer<T> {
```
1ec5: This would technically be a “queue”.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,164 @@
...
+    }
+    
+    var hashValue: Int {
+        get {
+            return id.hashValue
+        }
+    }
+    
+    static func ==(lhs: CoreFeedbackEvent, rhs: CoreFeedbackEvent) -> Bool {
+        return lhs.id == rhs.id
```
1ec5: If a FeedbackEvent and a RerouteEvent have the same ID, are they equal?

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+class FeedbackEvent: CoreFeedbackEvent {}
+
+class RerouteEvent: CoreFeedbackEvent {}
```
1ec5: Per https://github.com/mapbox/mapbox-navigation-ios/pull/304/#discussion_r124867003, it would be great to see this cluster of classes become a single struct containing an enumeration-typed property. (Down the line, if the two kinds of events begin to diverge, we can replace the entire struct with an enumeration with associated values.)

MapboxCoreNavigation/RouteController.swift
```diff
@@ -165,6 +212,78 @@ open class RouteController: NSObject {
...
+     this function will automatically flag the road segment/maneuver the user is currently on for 
+     closer inspection by Mapbox's system and team.
+     
+     If you provide a custom feedback UI that lets users elaborate on an 
+     issue, you should call this before you show the custom UI so the 
+     location and timestamp are more accurate. You can then call 
+     `updateLastFeedback()` to attach any additional metadata to the 
+     feedback.
+     */
+    public func recordFeedback(type: FeedbackType, description: String?) {
```
1ec5: Eventually, we may want to change this method to take a fully formed Feedback instance, in order to give the developer more fields they can fill out but avoid proliferating arguments.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -122,30 +138,74 @@ open class RouteController: NSObject {
...
+    public init(along route: Route, directions: Directions = Directions.shared, locationManager: NavigationLocationManager = NavigationLocationManager(), accessToken: String? = nil) {
         self.directions = directions
         self.routeProgress = RouteProgress(route: route)
         self.locationManager = locationManager
         self.locationManager.activityType = route.routeOptions.activityType
         super.init()
         
         self.locationManager.delegate = self
+        
+        self.resetSession()
```
ericrwolfe: We should move this call to reset the telemetry session just after the telemetry event initialization.

MapboxCoreNavigation/MMEEventsManager.swift
ericrwolfe: It would be developer error for that to ever happen.

bsudekum:
Hrm, tests are failing with the error

```
The bundle “MapboxNavigationTests” couldn’t be loaded because it is damaged or missing necessary resources. Try reinstalling the bundle.
```

MapboxCoreNavigation/MMEEventsManager.swift
1ec5: If it’s really developer error, then we should add an assertion to that effect here, because the compiler isn’t going to prevent that error.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -5,67 +5,144 @@ import MapboxMobileEvents
...
+        
+        volumeLevel = Int(AVAudioSession.sharedInstance().outputVolume * 100)
+        screenBrightness = Int(UIScreen.main.brightness * 100)
+        
+        batteryPluggedIn = UIDevice.current.batteryState == .charging || UIDevice.current.batteryState == .full
+        batteryLevel = UIDevice.current.batteryLevel >= 0 ? UIDevice.current.batteryLevel * 100 : -1
+        applicationState = UIApplication.shared.applicationState.telemetryString
+    }
+    
+    func convertToDictionary() -> [String: Any] {
```
frederoni: Nit: The verb implies a mutating function. `convertedToDictionary` or `asDictionary` would be more idiomatic.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
...
+    var requestIdentifier: String?
+    var lat: CLLocationDegrees?
+    var lng: CLLocationDegrees?
+    var originalGeometry: String?
+    var originalEstimatedDistance: CLLocationDistance?
+    var originalEstimatedDuration: TimeInterval?
+    var geometry: String?
+    var estimatedDistance: CLLocationDistance?
+    var estimatedDuration: TimeInterval?
+    var created: String
```
frederoni: Nit: This should be of type `Date` until after `convertToDictionary()` ×2

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
...
+                guard let value = element.value as? Int8, value != 0 else { return identifier }
+                return identifier + String(UnicodeScalar(UInt8(value)))
+            }
+            
+            return identifier
+        }
+    }
+}
+
+extension CLLocation {
```
frederoni: We have an extension `CLLocation.swift` in `MapboxCoreNavigationTests`, can we consolidate them?
CLLocation <-> dictionary would be nice to provide in core.

MapboxCoreNavigation/ProcessInfo.swift
```diff
@@ -0,0 +1,24 @@
...
+#elseif os(watchOS)
+    import WatchKit
+#endif
+
+extension ProcessInfo {
+    static var systemName: String {
+        #if os(iOS) || os(tvOS)
+            return UIDevice.current.systemName
+        #elseif os(watchOS)
+            return WKInterfaceDevice.current.systemVersion
```
1ec5: The iOS/tvOS code uses `UIDevice.systemName`, which returns either “iOS” or “tvOS”. But this line uses `WKInterfaceDevice.systemVersion`, which returns a version number like 2.0. Did you mean `systemName`, which returns “watchOS”, instead?

Maybe this method would be more understandable if we hard-code the values we want it to return.

MapboxCoreNavigation.podspec
```diff
@@ -42,5 +42,6 @@ Pod::Spec.new do |s|
 
   s.dependency "MapboxDirections.swift", "~> 0.10.0"
   s.dependency "OSRMTextInstructions", "~> 0.2.0"
+  s.dependency "MapboxMobileEvents", "~> 0.2"
```
1ec5: Does MapboxNavigation-Documentation.podspec need to be updated?

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
...
+import AVFoundation
+import MapboxMobileEvents
+
+let SecondsForCollectionAfterFeedbackEvent: TimeInterval = 20
+
+struct DefaultEventDictionary {
+    var originalRequestIdentifier: String?
+    var requestIdentifier: String?
+    var lat: CLLocationDegrees?
+    var lng: CLLocationDegrees?
```
1ec5: [Avoid abbreviations](https://swift.org/documentation/api-design-guidelines/#avoid-abbreviations) like `lng` for “longitude”.

1ec5: Combine `lat` and `lng` into a single `coordinate` property of type `CLLocationCoordinate2D` and rely on the conversion method to split it apart.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
...
+    var originalGeometry: String?
```
1ec5: This property should be of type `Polyline`; rely on the conversion method to encode it into a string.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
...
+    var originalEstimatedDistance: CLLocationDistance?
```
1ec5: Only the duration is estimated; the distance is known. (×2)

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
...
+    var batteryPluggedIn: Bool
+    var batteryLevel: Float
+    var applicationState: String
+    
+    init(routeController: RouteController, session: SessionState) {
+        created = Date().ISO8601
+        startTimestamp = session.departureTimestamp!.ISO8601
+        
+        platform = ProcessInfo.systemName
+        operatingSystem = "\(ProcessInfo.systemName) \(ProcessInfo.systemVersion)"
```
1ec5: The operating system name and version never change, so there’s no need to store it; just add them in when we convert the struct into a dictionary. This also applies to any other state that won’t change between the time we collect the data and the time we send it.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
...
+    var rerouteCount: Int
+    var volumeLevel: Int
+    var screenBrightness: Int
+    var batteryPluggedIn: Bool
+    var batteryLevel: Float
+    var applicationState: String
+    
+    init(routeController: RouteController, session: SessionState) {
+        created = Date().ISO8601
+        startTimestamp = session.departureTimestamp!.ISO8601
```
1ec5: This initializer should be very straightforward: hang onto any relevant state before it changes, but keep that state in its original types.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
+import Polyline
+import MapboxDirections
+import AVFoundation
+import MapboxMobileEvents
+
+let SecondsForCollectionAfterFeedbackEvent: TimeInterval = 20
+
+struct DefaultEventDictionary {
```
1ec5: A dictionary is a general-purpose container that has no structure. Since this struct has well-defined properties, we should avoid calling it a dictionary. How about `EventDetails`?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -466,3 +613,122 @@ extension RouteController: CLLocationManagerDelegate {
...
+        
+        sessionState.lastReroute = timestamp
+        sessionState.numberOfReroutes += 1
+        
+        outstandingFeedbackEvents.append(RerouteEvent(timestamp: timestamp, eventDictionary: eventDictionary))
+    }
+    
+    func checkAndSendOutstandingFeedbackEvents(forceAll: Bool = false) {
+        let now = Date()
+        let eventsToPush = forceAll ? outstandingFeedbackEvents : outstandingFeedbackEvents.filter({now.timeIntervalSince($0.timestamp) > SecondsForCollectionAfterFeedbackEvent})
```
1ec5: Nit: use trailing closure syntax.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,241 @@
+import Polyline
+import MapboxDirections
+import AVFoundation
+import MapboxMobileEvents
+
+let SecondsForCollectionAfterFeedbackEvent: TimeInterval = 20
```
1ec5: What does “for collection” mean?

MapboxCoreNavigation/MMEEventsManager.swift
bsudekum: This is the actual name of the key we expect server side though.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -122,30 +138,77 @@ open class RouteController: NSObject {
...
+        
+        if let mapboxAccessToken = mapboxAccessToken {
+            events.isDebugLoggingEnabled = true
+            events.isMetricsEnabledInSimulator = true
+            events.isMetricsEnabledForInUsePermissions = true
+            events.initialize(withAccessToken: mapboxAccessToken, userAgentBase: "MapboxEventsNavigationiOS", hostSDKVersion: String(describing: Bundle(for: RouteController.self).object(forInfoDictionaryKey: "CFBundleShortVersionString")!))
+            events.disableLocationMetrics()
+            events.sendTurnstileEvent()
+        } else {
+            assert(false, "`accessToken` must be set in the Info.plist as `MGLMapboxAccessToken`m passed in as a String to `RouteController()`")
```
frederoni: This introduced the test failure for `testRouteSwitching`

MapboxCoreNavigation/MMEEventsManager.swift
bsudekum: This extension is using a private extension so I'm going to leave it here.

MapboxCoreNavigation/RouteController.swift
bsudekum: fixed

MapboxCoreNavigation.podspec
bsudekum: yep https://github.com/mapbox/mapbox-navigation-ios/pull/304/files#diff-24463e219dc9d9fdb8dfde3c38cef64cR52

bsudekum:
@frederoni @1ec5 this is ready for another round of review.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,275 @@
...
+        }
+        
+        platform = ProcessInfo.systemName
+        operatingSystem = "\(ProcessInfo.systemName) \(ProcessInfo.systemVersion)"
+        device = UIDevice.current.machine
+        
+        sdkIdentifier = routeController.usesDefaultUserInterface ? "mapbox-navigation-ui-ios" : "mapbox-navigation-ios"
+        sdkVersion = String(describing: Bundle(for: RouteController.self).object(forInfoDictionaryKey: "CFBundleShortVersionString")!)
+        
+        eventVersion = 2
```
1ec5: All the events produced by a particular version of this library should have the same version, right? Let’s make sure that’s the case by pulling this value out into a constant. No need for a property on the struct; just use the constant in `convertedToDictionary()`.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,275 @@
...
+    
+    init(routeController: RouteController, session: SessionState) {
+        created = Date()
+        if let start = session.departureTimestamp?.ISO8601 {
+            startTimestamp =  start
+        }
+        
+        platform = ProcessInfo.systemName
+        operatingSystem = "\(ProcessInfo.systemName) \(ProcessInfo.systemVersion)"
+        device = UIDevice.current.machine
```
1ec5: `platform`, `operatingSystem`, `sdkVersion`, etc. can never change throughout the session, so set them in `convertedToDictionary()` rather than increasing the size of the struct. The fewer properties in the struct, the smaller each instance of the struct and the higher likelihood that Swift will use an efficient memory representation for it.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,275 @@
...
+    var originalGeometry: Polyline?
+    var originalDistance: CLLocationDistance?
+    var originalEstimatedDuration: TimeInterval?
+    var originalStepCount: Int?
+    var geometry: Polyline?
+    var distance: CLLocationDistance?
+    var estimatedDuration: TimeInterval?
+    var stepCount: Int?
+    var created: Date
+    var startTimestamp: String?
```
1ec5: This should be a Date, like `created`. Wait until `convertedToDictionary()` to lazily convert it to a string.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,275 @@
...
+        
+        volumeLevel = Int(AVAudioSession.sharedInstance().outputVolume * 100)
+        screenBrightness = Int(UIScreen.main.brightness * 100)
+        
+        batteryPluggedIn = UIDevice.current.batteryState == .charging || UIDevice.current.batteryState == .full
+        batteryLevel = UIDevice.current.batteryLevel >= 0 ? UIDevice.current.batteryLevel * 100 : -1
+        applicationState = UIApplication.shared.applicationState.telemetryString
+    }
+    
+    func convertedToDictionary() -> [String: Any] {
```
1ec5: `var eventDictionary` or `var jsonDictionary` would be more descriptive.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -0,0 +1,275 @@
...
+    var sessionIdentifier: String
+    var distanceCompleted: CLLocationDistance
+    var distanceRemaining: TimeInterval
+    var durationRemaining: TimeInterval
+    var rerouteCount: Int
+    var volumeLevel: Int
+    var screenBrightness: Int
+    var batteryPluggedIn: Bool
+    var batteryLevel: Float
+    var applicationState: String
```
1ec5: This should be a UIApplicationState. Wait until `convertedToDictionary()` to lazily convert it to a string.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -472,3 +618,134 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let eventName = "navigation.reroute"
+
+        let timestamp = Date()
+        
+        var eventDictionary = events.addDefaultEvents(routeController: self)
+        eventDictionary["event"] = eventName
+        
+        eventDictionary["secondsSinceLastReroute"] = sessionState.lastReroute != nil ? round(timestamp.timeIntervalSince(sessionState.lastReroute!)) : -1
+        for (key, value) in routeProgress.currentLegProgress.upcomingManeuverDictionary {
+            eventDictionary["step\(key.uppercaseFirst)"] = value
```
1ec5: As long as there aren’t any spaces in the keys, `String.capitalized` does the same thing.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -472,3 +618,134 @@ extension RouteController: CLLocationManagerDelegate {
...
+        eventDictionary["newGeometry"] = nil
+        eventDictionary["screenshot"] = captureScreen(scaledToFit: 250)?.base64EncodedString()
+        
+        sessionState.lastReroute = timestamp
+        sessionState.numberOfReroutes += 1
+        
+        outstandingFeedbackEvents.append(RerouteEvent(timestamp: timestamp, eventDictionary: eventDictionary))
+    }
+    
+    func checkAndSendOutstandingFeedbackEvents(forceAll: Bool = false) {
```
1ec5: It isn’t necessary to provide a default value here. Make all the call sites explicitly say whether to force all or not.

MapboxCoreNavigation/RouteController.swift
1ec5: This assertion introduces a potentially breaking change: until now, it hasn’t been necessary to set anything in Info.plist if you were using Core Navigation by itself and setting the access token in the `Directions(accessToken:)` initializer. Now you must use Info.plist (and might as well use it for `Directions.shared` too). We should note this change in the release notes.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -165,6 +228,78 @@ open class RouteController: NSObject {
...
+    public func recordFeedback(type: FeedbackType, description: String?) {
+        enqueueFeedbackEvent(type: type, description: description)
+    }
+    
+    /**
+     Update the last recorded feedback event, for example if you have a custom feedback UI that lets a user elaborate on an issue.
+     */
+    public func updateLastFeedback(type: FeedbackType, description: String?) {
+        if let lastFeedback = outstandingFeedbackEvents.filter({$0 is FeedbackEvent}).last {
+            lastFeedback.eventDictionary["feedbackType"] = type.rawValue
```
1ec5: `FeedbackType.rawValue` is an integer (e.g., 0 for `general`). Did you mean to set this item to `type.description` instead?

MapboxCoreNavigation/RouteController.swift
1ec5: There still aren’t any tests of the new feature. Some of the feedback above would no doubt be caught by basic tests.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -472,3 +618,134 @@ extension RouteController: CLLocationManagerDelegate {
...
+    
+    var totalDistanceCompleted: CLLocationDistance = 0
+    
+    var numberOfReroutes = 0
+    var lastReroute: Date?
+    
+    var currentRoute: Route!
+    var currentRequestIdentifier: String?
+    
+    var originalRoute: Route!
```
1ec5: Implicitly unwrapped optionals are unsafe. Make this property and `currentRoute` non-optional and pass them in when initializing the SessionState below.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -472,3 +618,134 @@ extension RouteController: CLLocationManagerDelegate {
         incrementRouteProgress(alertLevel, location: location, updateStepIndex: updateStepIndex)
     }
 }
+
+struct SessionState {
+    let identifier = UUID()
+    var departureTimestamp: Date?
+    var arrivalTimestamp: Date?
```
1ec5: Can we call these dates instead of timestamps? They are Dates, after all.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -472,3 +618,134 @@ extension RouteController: CLLocationManagerDelegate {
...
+    
+    var totalDistanceCompleted: CLLocationDistance = 0
+    
+    var numberOfReroutes = 0
+    var lastReroute: Date?
```
1ec5: `lastRerouteDate` would be clearer; otherwise, one could reasonably see `lastReroute` somewhere in the code and think it refers to a coordinate.

MapboxCoreNavigation/RouteController.swift
bsudekum: @1ec5 is there any way we can reach into `Directions.shared ` and check for an access token there?

MapboxCoreNavigation/MMEEventsManager.swift
boundsj: Also, I don't think this value is actually used currently. `convertedToDictionary()` currently sets the value with a literal value.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -122,30 +138,77 @@ open class RouteController: NSObject {
...
+            mapboxAccessToken = accessToken
+        } else if let path = Bundle.main.path(forResource: "Info", ofType: "plist"),
+            let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject],
+            let token = dict["MGLMapboxAccessToken"] as? String {
+            mapboxAccessToken = token
+        }
+        
+        if let mapboxAccessToken = mapboxAccessToken {
+            events.isDebugLoggingEnabled = true
+            events.isMetricsEnabledInSimulator = true
```
boundsj: I don't think we should ship this with metrics enabled in simulator set to true. Maybe this library can define a new user default key (it would not need to be documented) that could be checked and this value could default to false and and use the value in user defaults if available?

ericrwolfe:
@bsudekum good for a final review then I think we're ready to merge

bsudekum:
This is blocked by putting out a new release of MapboxDirections.swift which includes https://github.com/mapbox/MapboxDirections.swift/pull/155.

Waiting on https://github.com/mapbox/MapboxDirections.swift/pull/154 also.

MapboxCoreNavigation/RouteController.swift
1ec5: To close the loop: this was unblocked upstream in mapbox/MapboxDirections.swift#155.

Cartfile
```diff
@@ -1,5 +1,5 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 3.5
-github "mapbox/MapboxDirections.swift" ~> 0.10.0
+github "mapbox/MapboxDirections.swift" ~> 0.10
```
1ec5: Since MapboxDirections.swift hasn’t yet reached v1.0, anything could technically change at any time. We do try to avoid backwards-compatible changes in patch releases, but we make no such guarantees about minor releases.

We should pin to a specific patch version of the library (with the tadpole operator) to avoid surprises. `~> 0.10.1` will require v0.10.x except v0.10.0.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -137,15 +131,75 @@ struct EventDetails {
...
     }
 }
 
 extension MMEEventsManager {
     func addDefaultEvents(routeController: RouteController) -> [String: Any] {
-        return EventDetails(routeController: routeController, session: routeController.sessionState).convertedToDictionary()
+        return EventDetails(routeController: routeController, session: routeController.sessionState).eventDictionary
+    }
+    
+    func navigationDepartEvent(routeController: RouteController) -> [String: Any] {
```
1ec5: Instead of a series of separate methods that each sets a different `event` value, we should have a single method that takes an `MMEEventType` as a parameter; depending on the passed-in `eventType`, we can add additional items to `eventDictionary` as needed.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -195,6 +195,12 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
      */
     public var sendNotifications: Bool = true
     
+    public var showsReportFeedback: Bool = false {
```
1ec5: This method needs documentation.

PR #313, opened 2017-06-27
Refactor voice controllers

### LINKED ISSUE #280: Refactor voice controller
Factor out AV and Polly from RouteVoiceController. 

OOP - RouteVoiceController, AVVoiceController, PollyVoiceController
or
POP - Speakable protocol.

This will make it easier to make AWSPolly optional when that proposal is implemented.

@1ec5 @bsudekum

frederoni:
Fixes #280 

Split up voice controllers into RouteVoiceController and PollyVoiceController.

Replace AVSpeech with Polly:
```swift
navigationViewController.voiceController = PollyVoiceController(identityPoolId: "…")
```

@1ec5 @bsudekum 👀

### DIFF
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -3,6 +3,9 @@ import AVFoundation
 import MapboxDirections
 import MapboxCoreNavigation
 
+/**
+ The `RouteVoiceController` class provides voice guidance.
+ */
 @objc(MBRouteVoiceController)
 open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -2,10 +2,9 @@ import Foundation
 import AVFoundation
 import MapboxDirections
 import MapboxCoreNavigation
-import AWSPolly
 
 @objc(MBRouteVoiceController)
-public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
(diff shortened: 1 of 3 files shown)


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -162,19 +133,23 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     }
     
     func alertLevelDidChange(notification: NSNotification) {
-        guard isEnabled, volume > 0 else { return }
+        process(notification)
+    }
+    
+    func shouldSpeakFor(_ notification: NSNotification) -> Bool {
```
1ec5: `shouldSpeak(for:)`

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -184,15 +159,16 @@ public class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
         
-        if useDefaultVoice {
-            speakFallBack(fallbackText)
-        } else {
-            speakWithPolly(speechString(notification: notification, markUpWithSSML: true))
-        }
+        return true
+    }
+    
+    func process(_ notification: NSNotification) {
```
1ec5: This is a very vague name. Could we fold this into `alertLevelDidChange(_:)` and allow subclasses to override that method?

frederoni:
Interesting idea about the category but less obvious to grasp and debug.

We should add a new target for PollyVoiceControllerwhen we create the podspec to 
provide modularity for both Carthage and CocoaPods users.

1ec5:
If we add a separate target, we'd have to add a separate podspec. Subspecs can't coexist with separate modules.

PR #330, opened 2017-07-03
Improve style coverage

### LINKED ISSUE #327: Improve custom Style coverage
Developers should be able to customize the font and color of nearly every element in our NavigationViewController. 

Right now we're missing a few of the obvious ones:
- [ ] Turn arrow color
- [ ] Turn banner font
- [ ] Turn banner text color

@frederoni

frederoni:
Fixes #327 

Having one class per label made more sense when we use so many different font sizes instead of a generic `primaryTextFont`.

`NavigationViewController` now takes a `[Style]` when its initialized because we need to hold onto each `Style` which now observes changes to the preferred content size to adjust the font size according to the accessibility settings.

- [x] Support Dynamic type
- [x] Turn arrow color
- [x] Turn banner font
- [x] Turn banner text color
- [x] Conveniently change font family and preserve other attributes on all labels (`style.fontFamily = "Georgia"`)

@ericrwolfe @1ec5 👀


Examples/Swift/ViewController.swift
```diff
@@ -165,34 +165,36 @@ class ViewController: UIViewController, MGLMapViewDelegate {
...
         exampleMode = .styled
-
-        let navigationViewController = NavigationViewController(for: route, locationManager: locationManager())
-        navigationViewController.navigationDelegate = self
         
-        // Set a custom style URL
-        navigationViewController.mapView?.styleURL = URL(string: "mapbox://styles/mapbox/navigation-guidance-day-v1")
-
-        let style = Style()
+        let style = DefaultStyle()
```
1ec5: It seems odd to be creating a style from scratch each time we need to refer to it. If `Style.default` was too much, how about making the singleton a global?

MapboxNavigation/Style.swift
```diff
@@ -61,127 +24,144 @@ public class Style: NSObject {
...
+    
+    /**
+     Sets the text color on the distance remaining label.
+     */
+    public var distanceRemainingLabelTextColor: UIColor?
+    
+    /**
+     Sets the font on the ETA label.
+     */
+    public var etaLabelFont: UIFont?
```
1ec5: Nit: We’ve tended to say `arrivalDate` or `estimatedArrivalTime` instead of `eta`.

MapboxNavigation/Style.swift
```diff
@@ -222,87 +200,161 @@ public class Style: NSObject {
...
-        }
-        
-        if let font = wayNameLabelFont {
-            WayNameLabel.appearance(for: traitCollection).font = font
-        }
     }
 }
 
 /**
+ :nodoc:
```
1ec5: If we’re intentionally undocumenting these classes, should we make them private?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -281,6 +294,7 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         super.viewWillAppear(animated)
         
         UIApplication.shared.isIdleTimerDisabled = true
+        styles.forEach { $0.apply() }
```
1ec5: I think this can also be written `styles.forEach(Style.apply)`. Whenever you find more readable.

Examples/Swift/ViewController.swift
frederoni: I think we will maintain at least two styles in the near future, one for landscape and one for portrait (or the corresponding size classes). The example is a quick and dirty approach. I imagine you would subclass DefaultStyle either just once for a generic style or let's say per profile identifier MyBikeStyle or per size class MyCompactStyle.

MapboxNavigation/Style.swift
frederoni: I hesitated a while regarding this. It's just `UILabel` with no additional functionality but it might be useful if you want to do some custom appearance styling to at least expose the class?

MapboxNavigation/NavigationViewController.swift
frederoni: Doesn't seem to work. I think this is a Swift 3.2 feature, looks more readable though.

PR #333, opened 2017-07-05
Continue tracking after arrival

### LINKED ISSUE #318: Continue tracking location after arrival
The route controller currently stops tracking location after the user "arrives": https://github.com/mapbox/mapbox-navigation-ios/blob/5b79231075eb883dd694390ccab8c1b0b0655b84/MapboxCoreNavigation/RouteController.swift#L213

However there are a number of cases where it'd be useful to continue updating the map and distance stats event after the arrive alert has been triggered, particularly since it might be triggered slightly before the destination.

cc @bsudekum

frederoni:
Fixes #318 

@bsudekum @ericrwolfe 👀

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -209,8 +209,6 @@ extension RouteController: CLLocationManagerDelegate {
         let secondsToEndOfStep = userSnapToStepDistanceFromManeuver / location.speed
         
         guard routeProgress.currentLegProgress.alertUserLevel != .arrive else {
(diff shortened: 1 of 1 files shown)


bsudekum:
@frederoni does this mean we will continually reroute the user after arriving?

frederoni:
It could potentially lead to reroute after arriving at your destination if you move max(50, maxAccuracy) away from the arrival step which I'd say is correct behavior.

ericrwolfe:
We could disable reroute notifications by default after a user's arrived, but ultimately it should be up to the navigation UI implementation to determine whether to handle rerouting events after the user's arrived, not the route controller.

PR #336, opened 2017-07-05
Guard against no closest coord

bsudekum:
Seeing a few crashes around this code. We need to guard against not finding the closest coordinate.

/cc @frederoni @1ec5

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -404,7 +404,7 @@ extension RouteMapViewController: NavigationMapViewDelegate {
             return location
         }
         
-        guard location.course != -1 else {
+        guard location.course != -1, location.speed >= 0 else {
             return location
         }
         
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -409,7 +409,7 @@ extension RouteMapViewController: NavigationMapViewDelegate {
         }
         
         let nearByCoordinates = routeController.routeProgress.currentLegProgress.nearbyCoordinates
-        let closest = closestCoordinate(on: nearByCoordinates, to: location.coordinate)!
+        guard let closest = closestCoordinate(on: nearByCoordinates, to: location.coordinate) else { return location }
```
1ec5: This function would only return `nil` if `nearByCoordinates` is an empty array. Is it legitimate for there to be no `nearbyCoordinates` in the current leg?

bsudekum:
@1ec5 updated to guard against negative speeds. Other places that use this function look safe.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -404,7 +404,7 @@ extension RouteMapViewController: NavigationMapViewDelegate {
             return location
         }
         
-        guard location.course != -1 else {
+        guard location.course != -1, location.speed >= 0 else {
```
1ec5: This is OK, but I wonder if we can do a little better by tweaking `userDistanceBuffer` to fall back to a certain minimum distance if `location.speed` is negative.

Issue #338, opened 2017-07-06
roads are black and icon are blank sometimes


zijiazhai:
@hi, guys, I am using version 0.4.0, this only appears on the device of iPhone 6 sometimes, but it is very annoying, please help. 
![wechat_image_20170630153328](https://user-images.githubusercontent.com/10751092/27920573-a71aabee-623a-11e7-9803-1a449ef4f32e.jpg)

bsudekum:
@zijiazhai can you share your setup?
* simulator
* phone version
* ios version

zijiazhai:
~ none
~ iPhone 6 and 5s
~ both 10.3.2
~ did not custom map style
![snip20170706_1](https://user-images.githubusercontent.com/10751092/27930060-e33cc754-625a-11e7-85fd-02be7a8f8b88.png)

icons are fine, black lines only appear a few times  on device of 5s.
but blank icons and black lines appear lots of times on device of iPhone 6.

1ec5:
#339 should’ve fixed this issue; it’ll be in v0.5.0. Please reopen this issue if you continue to experience the black lines and icons in master.

bsudekum:
@1ec5 do you think #339 fixed the blank UI elements though?

1ec5:
Oh right, that change in itself does not fix the UI elements.

@zijiazhai, is your application's tint color white, by any chance? That would cause the user puck and overview button to appear to have no content. Does this issue persist if you give the NavigationViewController a different tint color?

PR #339, opened 2017-07-06
Wait until style has loaded before doing things to it

bsudekum:
Attempts to fix the black lines in https://github.com/mapbox/mapbox-navigation-ios/issues/338.

I think in some cases, https://github.com/mapbox/mapbox-navigation-ios/blob/eab11c929885b270e8c655495ed4e0b25123a78b/MapboxNavigation/RouteMapViewController.swift#L345-L358 could be causing a race condition if the style is not fully loaded.

/cc @frederoni @1ec5 @zijiazhai

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -26,6 +26,8 @@ class RouteMapViewController: UIViewController {
 
     var route: Route { return routeController.routeProgress.route }
     var previousStep: RouteStep?
+    
+    var hasFinishedLoadingStyle = false
 
     var destination: MGLAnnotation!
     var pendingCamera: MGLMapCamera? {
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -315,9 +317,15 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
     }
+    
+    func mapViewDidFinishLoadingMap(_ mapView: MGLMapView) {
+        hasFinishedLoadingStyle = true
+    }
 
     @objc(navigationMapView:shouldUpdateTo:)
     func navigationMapView(_ mapView: NavigationMapView, shouldUpdateTo location: CLLocation) -> CLLocation? {
+        
+        guard hasFinishedLoadingStyle else { return nil }
```
1ec5: If the style hasn’t finished loading, we can’t manipulate the style in order to query it for street names, but we should still run the rest of this method in order to snap the user’s location to the route.

In fact, that’s what the code already does: it [checks for a non-`nil` `style`](https://github.com/mapbox/mapbox-navigation-ios/blob/eab11c929885b270e8c655495ed4e0b25123a78b/MapboxNavigation/RouteMapViewController.swift#L335) before manipulating the style. `style` would be `nil` until the style has finished loading.

1ec5: So you’re actually right about needing to track `hasFinishedLoadingStyle` separately. `style` is `nil` until the style finishes loading, but sources such as Streets may take a little longer to load. But instead of `guard`, we need to modify the existing giant `if` statement.

bsudekum:
@1ec5 fixed

PR #340, opened 2017-07-06
Add POC telem lib integration

boundsj:
This PR illustrates how the new telem lib can be used in this project.

cc @ericrwolfe @bsudekum


MapboxCoreNavigation/RouteController.swift
```diff
@@ -139,6 +142,11 @@ open class RouteController: NSObject {
         super.init()
         
         self.locationManager.delegate = self
+        
+        // This is just a test!!! (needs a real access token, a user agent base, and the actual nav SDK version
+        events.initialize(withAccessToken: "your token here!", userAgentBase: "MapboxEventsiOSNavigation", hostSDKVersion: "42.42.42")
+        events.isDebugLoggingEnabled = true
+        events.sendTurnstileEvent()
```
ericrwolfe: This seems to send a generic `MMEEventTypeAppUserTurnstile`. Do we want these turnstile events to be specific to maps, navigation, etc?

bsudekum: I think @boundsj might just be showing an example of how you would use the API here?

boundsj: @bsudekum is correct, this was just an example. However, as discussed, we are implementing a universal turnstile event that all host SDKs will use.

bsudekum:
@boundsj we'll also need cocoapods support for this too. Can we simply checkin a podspec into https://github.com/mapbox/mapbox-telemetry-ios?

bsudekum:
Closing in favor of https://github.com/mapbox/mapbox-navigation-ios/pull/304

PR #343, opened 2017-07-07
Stop updating location on map view

### LINKED ISSUE #341: Navigation not working when app going to background and back to foreground.
When I start navigation and after my app going to background and I open app that time navigation not working properly.

frederoni:
Fixes #341 

When an application is sent to the background and brought back to the foreground, `MGLMapView` calls `-validateLocationServices` and starts updating locations again.

This is easy to spot when using `SimulatedLocationManager` or `ReplayLocationManager` but _not_ if you're using the `DefaultLocationManager` since the location will be exactly the same.

The fix in this PR works, however, I think a more appropriate solution would be to expose a `manuallyUpdatesLocation` (or similar) on `MGLMapView` that would leave location updating to the user. At the same time, we would also expose that `MGLMapView` conforms to `CLLocationManagerDelegate` so that you can feed `MGLMapView` with locations without using the workaround in https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/MGLMapView%2BMGLNavigationAdditions.h#L5

@1ec5 @bsudekum @ericrwolfe 👀

### DIFF
--- a/MapboxNavigation/MGLMapView+MGLNavigationAdditions.h
+++ b/MapboxNavigation/MGLMapView+MGLNavigationAdditions.h
@@ -5,6 +5,9 @@
 // FIXME: This will be removed once https://github.com/mapbox/mapbox-gl-native/issues/6867 is implemented
 - (void)locationManager:(CLLocationManager *)manager didUpdateLocations:(NSArray<CLLocation *> *)locations;
 
+// FIXME: This will be removed once https://github.com/mapbox/mapbox-navigation-ios/issues/352 is implemented.
+- (void)validateLocationServices;
+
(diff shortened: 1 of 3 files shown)


bsudekum:
Deferring judgement to @1ec5 on this one.

1ec5:
> The fix in this PR works, however, I think a more appropriate solution would be to expose a `manuallyUpdatesLocation` (or similar) on `MGLMapView` that would leave location updating to the user.

This isn’t unreasonable, but between this change and mapbox/mapbox-gl-native#6867, I wonder how much we’re benefiting from the map SDK’s built-in course tracking functionality in the first place. #352 would move us off that built-in functionality in favor of a custom implementation that we’d have more control over.

PR #346, opened 2017-07-07
Add passive smart rerouting

bsudekum:
This PR adds the ability for the route controller to check for routes with better ETAs while the user is navigating. There are some checks in place to make sure we do this only when necessary:

* There much be an upcoming step.
* There must be at least 70 seconds left on the current step to allow ample maneuvering for new route.
* The current route must have at least 10 minutes left on it.
* The new route must have an ETA that is at least 10% faster

If the new route has the same upcoming maneuver as the current route, the step is not announced since it already has been announced.

There currently is no UI around this. This is okay since the heuristics are extremely stiff; triggering a new route in it's current implementation will be extremely rare IMO.

/cc @willwhite @ericrwolfe @1ec5 @frederoni


MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,54 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let durationRemaining = routeProgress.durationRemaining
+        let currentAlertLevel = routeProgress.currentLegProgress.alertUserLevel
+        
+        routeTask?.cancel()
+        routeTask = directions.calculate(options, completionHandler: { [weak self] (waypoints, routes, error) in
+            guard let strongSelf = self else { return }
+            
+            strongSelf.lastLocationTimestamp = nil
+            
+            if let route = routes?.first {
```
1ec5: A `guard` statement would simplify this closure a bit.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,54 @@ extension RouteController: CLLocationManagerDelegate {
...
+                let percentDifference = ((durationRemaining - route.expectedTravelTime) / route.expectedTravelTime) * 100
```
1ec5: If the user is currently 10 minutes from their destination and the new route is 9 minutes long, that’s a 10% improvement, but `percentDifference` is 11.1 because it puts the new route’s time in the denominator instead of the old route’s time.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,54 @@ extension RouteController: CLLocationManagerDelegate {
...
+                
+                // Only use new route if it's at least 10% faster
+                if percentDifference > 10 {
```
1ec5: The comment says 10% faster is adequate, but this `if` statement disagrees.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,54 @@ extension RouteController: CLLocationManagerDelegate {
...
+        
+        guard let currentUpcomingManeuver = routeProgress.currentLegProgress.upComingStep else { return }
+        
+        guard let lastLocationTimestamp = lastLocationTimestamp else {
+            self.lastLocationTimestamp = location.timestamp
+            return
+        }
+
+        // Only check ever 2 minutes for faster route
+        guard location.timestamp.timeIntervalSince1970 - lastLocationTimestamp.timeIntervalSince1970 > 120 else { return }
```
1ec5: Use `Date.timeIntervalSince(_:)`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,54 @@ extension RouteController: CLLocationManagerDelegate {
...
+        
+        let options = routeProgress.route.routeOptions
+        options.waypoints = [Waypoint(coordinate: location.coordinate)] + routeProgress.remainingWaypoints
+        if let firstWaypoint = options.waypoints.first, location.course >= 0 {
```
1ec5: `firstWaypoint` is guaranteed to be the one we created above using `Waypoint(coordinate:)`. Pull it out into a local variable and reuse it below.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,54 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
     
+    func checkForFasterRoute(from location: CLLocation) {
+        
+        // If the user is on a short route, don't check for faster alternatives
```
bsudekum: I'm not certain I see why distance remaining would be more useful here. If there is not much distance remaining on a route, but the duration remaining is high, we should still be looking for a new route.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,54 @@ extension RouteController: CLLocationManagerDelegate {
...
+        guard routeProgress.durationRemaining > 600 else { return }
+
+        // If the user is approaching a maneuver, don't check for a faster alternatives
+        guard routeProgress.currentLegProgress.currentStepProgress.durationRemaining > 70 else { return }
```
bsudekum: @1ec5 how could be force this? Make the first waypoint the upcoming maneuver?

1ec5: Yes, exactly.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,55 @@ extension RouteController: CLLocationManagerDelegate {
...
+        
+        // If the user does not have much time left on the route, don't check for faster alternatives
+        guard routeProgress.durationRemaining > 600 else { return }
+
+        // If the user is approaching a maneuver, don't check for a faster alternatives
+        guard routeProgress.currentLegProgress.currentStepProgress.durationRemaining > 70 else { return }
+        
+        guard let currentUpcomingManeuver = routeProgress.currentLegProgress.upComingStep else { return }
+        
+        guard let lastLocationTimestamp = lastLocationDate else {
```
1ec5: Rename the local variable to lastLocationDate for consistency.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,55 @@ extension RouteController: CLLocationManagerDelegate {
...
+            lastLocationDate = location.timestamp
+            return
+        }
+
+        // Only check ever 2 minutes for faster route
+        guard location.timestamp.timeIntervalSince(lastLocationTimestamp) > 120 else { return }
```
1ec5: The comment implies that 120 seconds is enough for a new route, but this line disagrees.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,55 @@ extension RouteController: CLLocationManagerDelegate {
...
+            guard let strongSelf = self else { return }
+            
+            strongSelf.lastLocationDate = nil
+            
+            guard let route = routes?.first else { return }
+            
+            let percentDifference = ((durationRemaining - route.expectedTravelTime) / durationRemaining) * 100
+            
+            // Only use new route if it's at least 10% faster
+            if percentDifference >= 10 {
```
1ec5: This can be simplified algebraically:

```swift
if route.expectedTravelTime <= 0.9 * durationRemaining
```

This is also safer, on the off-chance that we get into this code when the user is already at the destination. With the current code, there would be a division by zero.

bsudekum:
@1ec5 updated

MapboxCoreNavigation/RouteController.swift
```diff
@@ -342,43 +341,30 @@ extension RouteController: CLLocationManagerDelegate {
         
         guard let currentUpcomingManeuver = routeProgress.currentLegProgress.upComingStep else { return }
         
-        guard let lastLocationTimestamp = lastLocationDate else {
-            lastLocationDate = location.timestamp
+        guard let lastLocationDate = lastLocationDate else {
+            self.lastLocationDate = location.timestamp
```
1ec5: Huh, I wouldn’t’ve expected that `self.` would be required here. 🤷‍♂️

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +332,41 @@ extension RouteController: CLLocationManagerDelegate {
...
+            return
+        }
+
+        // Only check ever 2 minutes for faster route
+        guard location.timestamp.timeIntervalSince(lastLocationDate) >= 120 else { return }
+        let durationRemaining = routeProgress.durationRemaining
+        let currentAlertLevel = routeProgress.currentLegProgress.alertUserLevel
+        
+        getDirections(from: location) { [weak self] (route, error) in
```
1ec5: We should reroute from the upcoming maneuver’s location, not from the user’s current location: https://github.com/mapbox/mapbox-navigation-ios/pull/346#discussion_r126294608.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +339,50 @@ extension RouteController: CLLocationManagerDelegate {
...
+            if route.legs[0].steps[1].expectedTravelTime <= 70 {
+                let maneuverLocation = CLLocation(coordinate: currentUpcomingManeuver.maneuverLocation, altitude: 0, horizontalAccuracy: 1, verticalAccuracy: 1, course: currentUpcomingManeuver.finalHeading ?? currentUpcomingManeuver.initialHeading ?? location.course, speed: 10, timestamp: Date())
+                
+                strongSelf.getDirections(from: maneuverLocation, completion: { (route, error) in
+                    guard let route = route else { return }
+                    
+                    // Only use new route if it's at least 10% faster
+                    if route.expectedTravelTime <= 0.9 * durationRemaining {
+                        // If the upcoming maneuver in the new route is the same as the current upcoming maneuver, don't announce it
+                        strongSelf.routeProgress = RouteProgress(route: route, legIndex: 0, alertLevel: currentUpcomingManeuver.description == route.legs[0].steps[1].description ? currentAlertLevel : .none)
```
1ec5: `RouteStep.description` is really just an alias for `RouteStep.instructions`, which isn’t guaranteed to be unique among RouteSteps. For example, what if `currentUpcomingManeuver` is a ramp step and the new `route` starts with a different ramp step? In both cases, `instruction` might be:

> Take the ramp on the right

but the new step could be much closer.

For the purposes of this feature, two steps are equivalent if they’re located in the same place and have the same maneuver type and direction. mapbox/MapboxDirections.swift#151 would implement a custom `==` operator for RouteStep, but there’s nothing stopping us from implementing the same logic here for now.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +339,50 @@ extension RouteController: CLLocationManagerDelegate {
...
+            guard let strongSelf = self else { return }
+            
+            strongSelf.lastLocationDate = nil
+            
+            guard let route = route else { return }
+            
+            if route.legs[0].steps[1].expectedTravelTime <= 70 {
+                let maneuverLocation = CLLocation(coordinate: currentUpcomingManeuver.maneuverLocation, altitude: 0, horizontalAccuracy: 1, verticalAccuracy: 1, course: currentUpcomingManeuver.finalHeading ?? currentUpcomingManeuver.initialHeading ?? location.course, speed: 10, timestamp: Date())
+                
+                strongSelf.getDirections(from: maneuverLocation, completion: { (route, error) in
```
1ec5: Nit: use trailing closure syntax.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +339,50 @@ extension RouteController: CLLocationManagerDelegate {
...
+                    // Only use new route if it's at least 10% faster
+                    if route.expectedTravelTime <= 0.9 * durationRemaining {
+                        // If the upcoming maneuver in the new route is the same as the current upcoming maneuver, don't announce it
+                        strongSelf.routeProgress = RouteProgress(route: route, legIndex: 0, alertLevel: currentUpcomingManeuver.description == route.legs[0].steps[1].description ? currentAlertLevel : .none)
+                        strongSelf.delegate?.routeController?(strongSelf, didRerouteAlong: route)
+                    }
+                })
+            } else {
+                // Only use new route if it's at least 10% faster
+                if route.expectedTravelTime <= 0.9 * durationRemaining {
```
1ec5: There’s a fair amount of duplication within this method. This `if` statement represents the case where the new route leaves enough time to maneuver. Since it appears twice in the same method, let’s factor it out into a local closure, then call it in both places.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -330,6 +339,50 @@ extension RouteController: CLLocationManagerDelegate {
...
+    func checkForFasterRoute(from location: CLLocation) {
+        guard let currentUpcomingManeuver = routeProgress.currentLegProgress.upComingStep else { return }
+        
+        guard let lastLocationDate = lastLocationDate else {
+            self.lastLocationDate = location.timestamp
+            return
+        }
+
+        // Only check ever 2 minutes for faster route
+        guard location.timestamp.timeIntervalSince(lastLocationDate) >= 120 else { return }
```
ericrwolfe: Let's change this to once every 5 minutes. Eventually, we could adapt this for longer segments (e.g. >50 mile highway driving) and space this out to once every 10-15 minutes.

bsudekum:
Ready for another round of reviews since tests are passing now.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -363,25 +372,16 @@ extension RouteController: CLLocationManagerDelegate {
...
             
             if route.legs[0].steps[1].expectedTravelTime <= 70 {
-                let maneuverLocation = CLLocation(coordinate: currentUpcomingManeuver.maneuverLocation, altitude: 0, horizontalAccuracy: 1, verticalAccuracy: 1, course: currentUpcomingManeuver.finalHeading ?? currentUpcomingManeuver.initialHeading ?? location.course, speed: 10, timestamp: Date())
+                let course = currentUpcomingManeuver.finalHeading ?? currentUpcomingManeuver.initialHeading ?? location.course
+                let maneuverLocation = CLLocation(coordinate: currentUpcomingManeuver.maneuverLocation, altitude: 1, horizontalAccuracy: 1, verticalAccuracy: 1, course: course, speed: 1, timestamp: Date())
                 
-                strongSelf.getDirections(from: maneuverLocation, completion: { (route, error) in
-                    guard let route = route else { return }
+                strongSelf.getDirections(from: maneuverLocation) { (route, error) in
+                    guard let newRoute = route else { return }
```
1ec5: `newRoute` is just an unwrapped version of `route`, so it’s fine to shadow the original variable and call the new local variable `route` as well.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -351,10 +351,19 @@ extension RouteController: CLLocationManagerDelegate {
         }
 
         // Only check ever 2 minutes for faster route
-        guard location.timestamp.timeIntervalSince(lastLocationDate) >= 120 else { return }
+        guard location.timestamp.timeIntervalSince(lastLocationDate) >= 300 else { return }
         let durationRemaining = routeProgress.durationRemaining
         let currentAlertLevel = routeProgress.currentLegProgress.alertUserLevel
         
+        func compareOldRoute(to route: Route) {
```
1ec5: A “compare” function returns whether something is greater, equal, or less than another thing, whereas this method potentially reroutes the user. A more accurate name would be `rerouteIfFaster(onto:)`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -351,10 +351,19 @@ extension RouteController: CLLocationManagerDelegate {
...
+            // Only use new route if it's at least 10% faster
+            if route.expectedTravelTime <= 0.9 * durationRemaining {
+                // If the upcoming maneuver in the new route is the same as the current upcoming maneuver, don't announce it
+                routeProgress = RouteProgress(route: route, legIndex: 0, alertLevel: currentUpcomingManeuver.maneuverLocation == route.legs[0].steps[1].maneuverLocation ? currentAlertLevel : .none)
```
1ec5: We need to check whether the steps are equivalent, not just whether they occur at the same location. What if the current step turns left at this location, whereas the new route’s first step turns right at the same location? See mapbox/MapboxDirections.swift#151 for a rough sketch of what RouteStep equality would look like.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -268,6 +269,14 @@ extension RouteController: CLLocationManagerDelegate {
         }
         
         monitorStepProgress(location)
+        
+        // Check for faster route given users current location
+        //
+        // Only check for faster alternatives if the user has plenty of time left on the route.
+        guard routeProgress.durationRemaining > 600 else { return }
+        // If the user is approaching a maneuver, don't check for a faster alternatives
+        guard routeProgress.currentLegProgress.currentStepProgress.durationRemaining > 70 else { return }
```
ericrwolfe: Can we use the existing low alert level constant here?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -333,6 +342,50 @@ extension RouteController: CLLocationManagerDelegate {
...
+            }
+        }
+        
+        getDirections(from: location) { [weak self] (route, error) in
+            guard let strongSelf = self else { return }
+            guard let route = route else { return }
+            guard let firstLeg = route.legs.first, firstLeg.steps.count > 2 else { return }
+            strongSelf.lastLocationDate = nil
+            
+            if firstLeg.steps[1].expectedTravelTime <= 70 {
```
ericrwolfe: Can we use the existing low alert level constant here?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -333,6 +342,50 @@ extension RouteController: CLLocationManagerDelegate {
...
+                let course = currentUpcomingManeuver.finalHeading ?? currentUpcomingManeuver.initialHeading ?? location.course
+                let maneuverLocation = CLLocation(coordinate: currentUpcomingManeuver.maneuverLocation, altitude: 1, horizontalAccuracy: 1, verticalAccuracy: 1, course: course, speed: 1, timestamp: Date())
+                
+                strongSelf.getDirections(from: maneuverLocation) { (route, error) in
```
ericrwolfe: If you're not yet at the maneuver location, wouldn't this trigger another reroute?

1ec5: That’s entirely up to the logic that follows for determining whether the new route is distinct from the existing route. Currently it doesn’t even check whether the two routes are distinct; it just goes with the new one as long as it’s faster. Conceptually, we should only use the new route if doing so would keep the contents of the turn banner the same for now.

Moreover, in terms of the alert level, it compares the `currentUpcomingManeuver`’s coordinates with those of `firstLeg.steps[1]`, which I think is incorrect. Consider the following scenario:

* `currentManeuver` has a few miles left.
* `currentUpcomingManeuver` is a right turn at Albuquerque.
* `firstLeg.steps[1]` is a departure at the user’s current location.
* `firstLeg.steps[2]` is a right turn at Albuquerque.

As I pointed out in https://github.com/mapbox/mapbox-navigation-ios/pull/346#discussion_r127333181, we should be comparing the maneuver type, direction, and coordinates of `currentUpcomingManeuver` and `firstLeg.steps[2]`. Otherwise, we are indeed guaranteed a reroute.

willwhite:
The behavior here could be a little funny/unexpected unless we also update the current route - for example using map matching. On longer trips, overall traffic conditions across all possible routes can significantly improve across the board and you'd be doing your 10% comparison against a stale version of your current route, putting it at a big disadvantage even thought it might still be a reasonable way to go.

The opposite can also be true - traffic conditions for all possible routes can get significantly worse causing us to never pass the 10% threshold. In this situation we'd never update the ETA or congestion information on the current route.

Updating the ETA (and congestion information) on the existing route as you drive feels like the higher priority.

ericrwolfe:
> Updating the ETA (and congestion information) on the existing route as you drive feels like the higher priority.

It high priority, but also more complicated to build right now given we'll need to build out support for the map-matching API in MapboxDirections.swift.

> On longer trips, overall traffic conditions across all possible routes can significantly improve across the board and you'd be doing your 10% comparison against a stale version of your current route, putting it at a big disadvantage even thought it might still be a reasonable way to go.

In this case, the new route will exceed the 10% threshold and the user will continue on the fastest route, now with an accurate ETA.

> traffic conditions for all possible routes can get significantly worse causing us to never pass the 10% threshold. In this situation we'd never update the ETA or congestion information on the current route.

In this case, nothing changes from what we have now.

We should ship this before ETAs since it's a quick win that will help overall, and in the 2 cases @willwhite provided we're no worse off than where we're at now.

1ec5:
> more complicated to build right now given we'll need to build out support for the map-matching API in MapboxDirections.swift

After sketching out a design mapbox/MapboxDirections.swift#158, this task doesn’t seem all that daunting. But it would be higher-hanging fruit than what we’ve got here.

PR #348, opened 2017-07-07
MapboxDirections.swift v0.10, OSRM Text Instructions v0.2.0

1ec5:
Upgraded to MapboxDirections.swift v0.10.0 and OSRM Text Instructions v0.2.0. In keeping with Project-OSRM/osrm-text-instructions.swift#26, `RouteStepFormatter.string(for:markUpWithSSML:)` has been renamed `RouteStepFormatter.string(for:legIndex:numberOfLegs:markUpWithSSML:)`, taking two additional, optional arguments.

/cc @bsudekum


WA-MIKE:
Do you have to import OSRM Text Instructions into one of the swift files within MapboxDirections.swift? I have the problem where the navigation ends at the first waypoint and was wondering if anything else has to be done besides just installing the podfile, Thank you!

1ec5:
> Do you have to import OSRM Text Instructions into one of the swift files within MapboxDirections.swift?

No, that isn’t necessary. OSRM Text Instructions depends on MapboxDirections.swift, not the other way around.

> I have the problem where the navigation ends at the first waypoint

We’re working on an implementation of this feature in #270.

WA-MIKE:
I see, so porting OSRM Text Instructions doesn't fix the waypoint issue? That implementation is not done yet, is that correct?

1ec5:
Correct. #270 is underway; please keep track of that issue for updates.

PR #349, opened 2017-07-07
Update deps in podfiles

bsudekum:
/cc @1ec5

### DIFF
--- a/MapboxNavigation-Documentation.podspec
+++ b/MapboxNavigation-Documentation.podspec
@@ -43,9 +43,9 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxNavigation"
 
-  s.dependency "MapboxDirections.swift", "~> 0.9.0"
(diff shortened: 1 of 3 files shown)


MapboxNavigation-Documentation.podspec
```diff
@@ -43,9 +43,9 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxNavigation"
 
-  s.dependency "MapboxDirections.swift", "~> 0.9.0"
-  s.dependency "Mapbox-iOS-SDK", "~> 3.5"
-  s.dependency "OSRMTextInstructions", "~> 0.1.2"
+  s.dependency "MapboxDirections.swift", "~> 0.10.0"
+  s.dependency "Mapbox-iOS-SDK", "~> 3.6"
```
1ec5: If I’m not mistaken, this library is still compatible with v3.5, so leaving this line alone will allow developers to upgrade from v3.5.x to v3.6.0 at their own pace.

PR #353, opened 2017-07-09
Display shields for destinations in turn banner

### LINKED ISSUE #320: Display shields for destinations in turn banner
Due to Project-OSRM/osrm-backend#3304, MapboxDirections.swift often ends up with route numbers in `RouteStep.destinations` instead of `RouteStep.destinationCodes`, preventing a shield from being shown in the turn banner. [This on-ramp](https://www.openstreetmap.org/way/28357255) should have a California State Route shield to the right, but instead “CA 87 South” appears in plain text with no accompanying shield:

guadalupe

As a workaround, we should try to match `destinations` (everything up to the directional suffix) to patterns in Shields.plist, as we already do for `destinationCodes`.

/ref #291
/cc @ericrwolfe @bsudekum @frederoni @freenerd

1ec5:
This PR significantly increases the frequency of route shields in the turn banner. Before, the turn banner only considered the route step’s `code` for displaying a shield, but that would only be relevant when merging onto a freeway and when approaching an off-ramp that happens to be part of a route. Now the turn banner also shows a shield based on `destinationCode`, which is populated when approaching an on-ramp and when approaching an off-ramp that isn’t part of a route.

Additionally, we try to treat the route step’s `destination` as a `destinationCode` in case it begins with a route number. This is a workaround for Project-OSRM/osrm-backend#3304 to fix #320.

RouteManeuverViewController is now responsible for fetching route shield images it needs to display. This makes it easier to avoid redundant requests to Wikimedia Commons: we only issue a request when the textual representation of the shield changes.

Guadalupe Freeway Park Avenue

/cc @bsudekum @frederoni @ericrwolfe @willwhite

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -3,8 +3,6 @@ import Pulley
 import Mapbox
 import MapboxDirections
 import MapboxCoreNavigation
-import SDWebImage
-
 
 class ArrowFillPolyline: MGLPolylineFeature {}
 class ArrowStrokePolyline: ArrowFillPolyline {}
@@ -52,9 +50,6 @@ class RouteMapViewController: UIViewController {
 
     var resetTrackingModeTimer: Timer?
 
-    let webImageManager = SDWebImageManager.shared()
-    var shieldAPIDataTask: URLSessionDataTask?
-    var shieldImageDownloadToken: SDWebImageDownloadToken?
     var arrowCurrentStep: RouteStep?
     var isInOverviewMode = false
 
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -107,7 +102,6 @@ class RouteMapViewController: UIViewController {
 
     override func viewWillDisappear(_ animated: Bool) {
```
frederoni: Nit: Let's drop this override now when it's only calling super.

1ec5: Fixed in af5fe5c82dbae957544a9db9979300583f5a05f8.

PR #355, opened 2017-07-10
Fix localizable bundle

frederoni:
The localized utterance format should be parsed from the navigationUI's bundle.
This bug resulted in partially localized voice instructions (`Sväng höger in på Stadiongatan, then Sväng vänster in på Trelleborgsvägen`) <- Swenglish.

@1ec5 @bsudekum 👀


frederoni:
It was also absent from `LESS_THAN` and `NAME_AND_REF`.

PR #356, opened 2017-07-10
Fix voice guidance in background

### LINKED ISSUE #328: Audio alerts don't work if app is in the background
If the app is moved into the background, audio alerts are no longer announced, even though local notifications appear. 

This is separate from the issue with no audio while the vibrate switch is enabled.

cc @bsudekum

frederoni:
Fixes #328 

This PR fixes voice guidance in background for `PollyVoiceController` and informs if `audio` is missing from `UIBackgroundModes`.

@bsudekum @1ec5 👀

### DIFF
--- a/MapboxCoreNavigation/Bundle.swift
+++ b/MapboxCoreNavigation/Bundle.swift
@@ -2,14 +2,20 @@ import Foundation
 
 extension Bundle {
     
-    var backgroundModeLocationSupported: Bool {
-        get {
-            if let modes = object(forInfoDictionaryKey: "UIBackgroundModes") as? [String] {
-                return modes.contains("location")
+    struct BackgroundModes {
+        var location: Bool {
+            return backgroundModesContains("location")
+        }
+        
+        fileprivate func backgroundModesContains(_ key: String) -> Bool {
+            if let modes = Bundle.main.object(forInfoDictionaryKey: "UIBackgroundModes") as? [String] {
+                return modes.contains(key)
             }
             return false
         }
     }
+    
+    var backgroundModes: BackgroundModes { return BackgroundModes() }
     
     var locationAlwaysAndWhenInUseUsageDescription: String? {
         get {
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -23,7 +23,7 @@ open class NavigationLocationManager: CLLocationManager {
         }
         
         if #available(iOS 9.0, *) {
-            if Bundle.main.backgroundModes.location {
+            if Bundle.main.backgroundModes.contains("location") {
                 allowsBackgroundLocationUpdates = true
             }
         }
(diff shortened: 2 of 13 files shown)


MapboxCoreNavigation/Bundle.swift
```diff
@@ -2,15 +2,21 @@ import Foundation
...
-    var backgroundModeLocationSupported: Bool {
-        get {
-            if let modes = object(forInfoDictionaryKey: "UIBackgroundModes") as? [String] {
-                return modes.contains("location")
+    struct BackgroundModes {
+        var location: Bool {
+            return backgroundModesContains("location")
+        }
+        
+        fileprivate func backgroundModesContains(_ key: String) -> Bool {
```
1ec5: For a more natural representation, `BackgroundModes` could be a string-backed `OptionSet` or a `Set<String>`, and `backgroundModes` could be a lazily variable that reads from the info dictionary.

MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -23,7 +23,7 @@ open class NavigationLocationManager: CLLocationManager {
         }
         
         if #available(iOS 9.0, *) {
-            if Bundle.main.backgroundModeLocationSupported {
+            if Bundle.main.backgroundModes.location {
```
1ec5: Then this line would become `Bundle.main.backgroundModes.contains(.location)` or `Bundle.main.backgroundModes.contains("location")`.

MapboxNavigation/Bundle.swift
```diff
@@ -1,18 +1,29 @@
...
-                return modes.contains("location")
+                return modes.contains(key)
             }
             return false
         }
     }
+    
+    var backgroundModes: BackgroundModes { return BackgroundModes() }
+    
+    class var navigationUI: Bundle {
```
1ec5: The `navigationUI` name probably dates to when MapboxNavigation was MapboxNavigationUI. Could we change this to `mapboxNavigation` while we’re at it?

MapboxNavigation/Bundle.swift
```diff
@@ -1,18 +1,29 @@
...
+    struct BackgroundModes {
+        var location: Bool {
+            return backgroundModesContains("location")
+        }
+        
+        var audio: Bool {
+            return backgroundModesContains("audio")
+        }
+        
+        fileprivate func backgroundModesContains(_ key: String) -> Bool {
```
1ec5: We could avoid this duplication by sharing Bundle.swift between the MapboxCoreNavigation and MapboxNavigation modules. (We’d have to rename this file to avoid a build error.)

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -60,6 +60,12 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     
     override public init() {
         super.init()
+        
+        if !Bundle.main.backgroundModes.audio {
+            print("Voice guidance may not work properly. " +
```
1ec5: Should we throw an error to make this requirement more explicit? If the developer intentionally omitted the `audio` background mode, they can catch the error.

frederoni: The voice controller itself and the background mode is somewhat optional so I'd prefer to keep this subtle debug message.

PR #359, opened 2017-07-11
Add congestion to route line

bsudekum:
Note, holding on this until some server side PRs are fixed.

/cc @1ec5 @frederoni

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -115,6 +115,7 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         options.includesSteps = true
         options.routeShapeResolution = .full
         options.profileIdentifier = .automobileAvoidingTraffic
+        options.attributeOptions = [.openStreetMapNodeIdentifier, .speed, .congestionLevel]
         
         _ = Directions.shared.calculate(options) { [weak self] (waypoints, routes, error) in
             guard error == nil else {
@@ -253,6 +254,7 @@ extension ViewController: WaypointConfirmationViewControllerDelegate {
         options.includesSteps = true
         options.routeShapeResolution = .full
         options.profileIdentifier = navigationViewController.route.routeOptions.profileIdentifier
+        options.attributeOptions = [.openStreetMapNodeIdentifier, .speed, .congestionLevel]
 
         _ = Directions.shared.calculate(options) { [weak self] (waypoints, routes, error) in
             guard error == nil else {
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -115,7 +115,9 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         options.includesSteps = true
         options.routeShapeResolution = .full
         options.profileIdentifier = .automobileAvoidingTraffic
-        options.attributeOptions = [.openStreetMapNodeIdentifier, .speed, .congestionLevel]
+        
+        // Adding the optional attribute `.congestionLevel` ensures the route line will show the congestion along the route line
+        options.attributeOptions = [.congestionLevel]
         
         _ = Directions.shared.calculate(options) { [weak self] (waypoints, routes, error) in
             guard error == nil else {
@@ -254,7 +256,9 @@ extension ViewController: WaypointConfirmationViewControllerDelegate {
         options.includesSteps = true
         options.routeShapeResolution = .full
         options.profileIdentifier = navigationViewController.route.routeOptions.profileIdentifier
-        options.attributeOptions = [.openStreetMapNodeIdentifier, .speed, .congestionLevel]
+        
+        // Adding the optional attribute `.congestionLevel` ensures the route line will show the congestion along the route line
+        options.attributeOptions = [.congestionLevel]
 
         _ = Directions.shared.calculate(options) { [weak self] (waypoints, routes, error) in
             guard error == nil else {
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -195,6 +195,14 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         style.cellTitleLabelTextColor = #colorLiteral(red: 0, green: 0, blue: 0, alpha: 1)
         style.cellSubtitleLabelTextColor = #colorLiteral(red: 0.2549019754, green: 0.2745098174, blue: 0.3019607961, alpha: 1)
         
+        // Traffic
+        style.routeCasingColor = #colorLiteral(red: 0.1960784346, green: 0.3411764801, blue: 0.1019607857, alpha: 1)
+        style.trafficUnknownColor = #colorLiteral(red: 0.721568644, green: 0.8862745166, blue: 0.5921568871, alpha: 1)
+        style.trafficLowColor = #colorLiteral(red: 0.5843137503, green: 0.8235294223, blue: 0.4196078479, alpha: 1)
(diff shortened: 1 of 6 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -12,6 +14,11 @@ open class NavigationMapView: MGLMapView {
     let routeLayerIdentifier = "routeLayer"
     let routeLayerCasingIdentifier = "routeLayerCasing"
     
+    let TrafficSevere = UIColor(red:0.54, green:0.06, blue:0.22, alpha:1.0)
+    let TrafficHeavy = UIColor(red:0.91, green:0.20, blue:0.25, alpha:1.0)
+    let TrafficModerate = UIColor(red:0.95, green:0.65, blue:0.31, alpha:1.0)
+    let TrafficLowUnknown = UIColor(red:0.00, green:0.70, blue:0.99, alpha:1.0)
```
bsudekum: I opted to keep these private. If a developer messes with these too much, it might make the directions API seem like it's at fault. Traffic/congestion colors should be pretty standard.

1ec5: `TrafficLowUnknown` should somehow correspond to the route line color, right? What if the developer wants to change the route line’s color?

1ec5: Ah, I see you addressed that in efc14370214b23ab06ac231dd16684c81bc46959. 👍

MapboxNavigation/NavigationMapView.swift
```diff
@@ -21,6 +23,10 @@ open class NavigationMapView: MGLMapView {
         22: MGLStyleValue(rawValue: 18)
     ]
     
+    let TrafficSevere = UIColor(red:0.54, green:0.06, blue:0.22, alpha:1.0)
+    let TrafficHeavy = UIColor(red:0.91, green:0.20, blue:0.25, alpha:1.0)
+    let TrafficModerate = UIColor(red:0.95, green:0.65, blue:0.31, alpha:1.0)
```
1ec5: Even if we decide that no one should be allowed to deviate from the orange/red/maroon color scheme, a designer may want to deviate from the exact colors `#F2A64F` `#E83340` `#8A0F38`, for example to better match the Traffic Night style.

For what it’s worth, I don’t think it would be very risky to expose these colors in Style like all the others. If the developer chooses a misleading color scheme of their own, that doesn’t reflect poorly on our Directions API any more than an ugly custom style would reflect poorly on the Mapbox Streets source or the map SDK.

MapboxNavigation/DefaultStyle.swift
```diff
@@ -13,7 +13,7 @@ extension UIColor {
 extension UIColor {
     // General styling
     fileprivate class var defaultTint: UIColor { get { return #colorLiteral(red: 0.1411764771, green: 0.3960784376, blue: 0.5647059083, alpha: 1) } }
-    fileprivate class var defaultTintStroke: UIColor { get { return #colorLiteral(red: 0.05882352963, green: 0.180392161, blue: 0.2470588237, alpha: 1) } }
+    fileprivate class var defaultTintStroke: UIColor { get { return #colorLiteral(red:0.18, green:0.49, blue:0.78, alpha:1.0) } }
```
1ec5: Do any other properties need to follow suit, like `defaultTint`? In master, the color scheme `#0F2E3F` `#246590` is fairly subtle compared to `#0F2E3F` `#2E7DC7`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -94,22 +100,66 @@ open class NavigationMapView: MGLMapView {
...
-            return nil
+        guard let coordinates = route.coordinates else { return nil }
+        
+        let congestionPerLeg = route.legs.flatMap {
+            $0.segmentCongestionLevels
+        }
+        
+        let combinedCongestionLevel = Array(congestionPerLeg.joined()) // Flatten all leg nodes
+        let destination = coordinates.suffix(from: 1)
+        let segment = zip(coordinates, destination).map { [$0.0, $0.1] }
```
1ec5: We wind up zipping together a full segment array even if `congestionPerLeg` is the empty array. This probably happens in the Objective-C example, which doesn’t specify the `congestion` attribute when requesting the route.

Examples/Swift/ViewController.swift
```diff
@@ -115,6 +115,7 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         options.includesSteps = true
         options.routeShapeResolution = .full
         options.profileIdentifier = .automobileAvoidingTraffic
+        options.attributeOptions = [.openStreetMapNodeIdentifier, .speed, .congestionLevel]
```
1ec5: The documentation should be updated to suggest setting `attributeOptions`, especially if congestion levels require all three attribute options (which is undiscoverable).

MapboxNavigation/NavigationMapView.swift
bsudekum: Attempted to make the colors public in https://github.com/mapbox/mapbox-navigation-ios/pull/359/commits/682e04181283cc0ec52339b76aa506eae2ffa26f but it's not quite working right. /cc @frederoni for guidance on this.

MapboxNavigation/DefaultStyle.swift
bsudekum: @1ec5 here is how we're looking now: 
simulator screen shot jul 14 2017 8 28 37 am
simulator screen shot jul 14 2017 8 28 48 am

frederoni:
The `testRouteSwitching` test fails with the following message:
```
This app has attempted to access privacy-sensitive data without a usage description.
The app's Info.plist must contain an NSLocationWhenInUseUsageDescription key with a string
value explaining to the user how the app uses this data
```
Is it possible to use an Info.plist when testing a framework?

MapboxNavigation/DefaultStyle.swift
```diff
@@ -3,17 +3,23 @@ import Foundation
 
 extension UIColor {
     class var defaultRouteCasing: UIColor { get { return .defaultTintStroke } }
-    class var defaultRouteLayer: UIColor { get { return UIColor.defaultTintStroke.withAlphaComponent(0.6) } }
+    class var defaultRouteLayer: UIColor { get { return #colorLiteral(red:0.00, green:0.70, blue:0.99, alpha:1.0) } }
```
frederoni: It might be confusing that there is no color for the route layer in the style but `unknown` & `low` is being used there right? However, what's route layer color in case of biking/walking or a style w/o traffic?

bsudekum: no traffic it defaults to the blue unknown/low color:

image

MapboxNavigation/DefaultStyle.swift
1ec5: 🤷‍♂️ So long as @mayagao is OK with the new color scheme, I’m OK with it.

lawrencelm:
How do I display the traffic colors on the route on my objective-C app?

1ec5:
@lawrencelm, the easiest way is to use an instance of [NavigationRouteOptions](https://www.mapbox.com/mapbox-navigation-ios/navigation/0.11.0/Classes/NavigationRouteOptions.html) when creating the NavigationViewController. Here’s [a list of regions where Mapbox traffic data is available](https://www.mapbox.com/api-documentation/pages/traffic-countries.html).

If you need further troubleshooting help, please feel free to [contact our support team](https://www.mapbox.com/contact/support/).

PR #378, opened 2017-07-12
Improve slight turn completion heuristic

bsudekum:
Turn completion boils down to two different heuristics:
* is within x meters of maneuver radius
* user course is with x degrees of the exit heading

For slight turns like on/off ramps, this falls apart quickly since the user's course is easily within x degrees of the exit heading. 

This PR looks at the difference between the initial and exit heading. If their difference is less than x degrees, then don't rely on the exit heading check. Instead, rely on their snapped distance from the maneuver location. Once the user is at or moving away from the maneuver location, their snapped distance will always be zero. If it's zero, increment the step.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -394,7 +394,7 @@ extension RouteController: CLLocationManagerDelegate {
             let initialHeadingNormalized = wrap(initialHeading, min: 0, max: 360)
             let finalHeadingNormalized = wrap(finalHeading, min: 0, max: 360)
             let userHeadingNormalized = wrap(location.course, min: 0, max: 360)
-            let differenceBetweenInitialAndFinalHeading = differenceBetweenAngles(initialHeadingNormalized, finalHeadingNormalized)
+            let expectedTurningAngle = differenceBetweenAngles(initialHeadingNormalized, finalHeadingNormalized)
             
             // If the upcoming maneuver is fairly straight,
             // do not check if the user is within x degrees of the exit heading.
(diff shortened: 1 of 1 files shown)


bsudekum:
Also, if we find we want even more time/distance, we can check for their absolute distance as well. Once this is over x meters and their snapped distance is zero, increment.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -390,10 +390,23 @@ extension RouteController: CLLocationManagerDelegate {
...
+            let differenceBetweenInitialAndFinalHeading = differenceBetweenAngles(initialHeadingNormalized, finalHeadingNormalized)
+            
+            // If the upcoming maneuver is fairly straight,
+            // do not check if the user is within x degrees of the exit heading.
+            // For ramps, their current heading will very close to the exit heading.
+            // We need to wait until their moving away from the maneuver location instead.
+            // We can do this by looking at their snapped distance from the maneuver.
+            // Once this distance is zero, they are at more moving away from the maneuver location
+            if differenceBetweenInitialAndFinalHeading <= RouteControllerMaximumAllowedDegreeOffsetForTurnCompletion {
+                courseMatchesManeuverFinalHeading = userSnapToStepDistanceFromManeuver == 0
```
1ec5: Intersections in OpenStreetMap are somewhat idealized. What if a sharp left is modeled like this:

sharp-left

The user’s location never snaps to the precisely the end of the road. Do we have a provision for preceding to the next step if the user is closer to the next step geometry than the current step geometry?

1ec5: Oh, I missed that this check only occurs for slight turns.

bsudekum:
> In the case of a slight turn, we never consider the user’s change in heading at all.

I'm not certain this is true. My understanding is prior to this PR we always looked at the user's course and compared it to the exit heading. In slight turn cases, the difference in angle was always very small causing turn completion to happen at the first location update when entering the maneuver radius.

Issue #374, opened 2017-07-12
Suppress voice announcements and ping after a reroute


ericrwolfe:
Unless there's an impending new maneuver, an immediate voice announcement and the corresponding ping resulting from a reroute should be suppressed by default. 

cc @bsudekum

PR #382, opened 2017-07-13
Suppress reroute announcement on long steps

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/374

In cases where we reroute the user and the new route has a first step which has very long (1km), there is no need to alert the user about the first step. This is because they have ample time until the upcoming maneuver.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -356,8 +356,6 @@ extension RouteController: CLLocationManagerDelegate {
             firstWaypoint.heading = location.course
             firstWaypoint.headingAccuracy = 90
         }
-        
-        let currentStepDistanceRemaining = routeProgress.currentLegProgress.currentStepProgress.distanceRemaining
 
         routeTask = directions.calculate(options, completionHandler: { [weak self] (waypoints, routes, error) in
             guard let strongSelf = self else {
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -356,14 +356,22 @@ extension RouteController: CLLocationManagerDelegate {
...
                 return
             }
             
             if let route = routes?.first {
-                strongSelf.routeProgress = RouteProgress(route: route)
+                
+                // If the new route has a first step that has a distance greater than 1km,
+                // surpress the first announcement since they have ample time until the next maneuver
+                var newAlertLevel: AlertLevel = .none
+                if let firstLeg = route.legs.first, let firstStep = firstLeg.steps.first, firstStep.distance > 1_000 {
```
1ec5: The comment above says this heuristic is based on duration, whereas this statement checks only for distance. If the user is traveling along a freeway in the U.S. and the new upcoming maneuver is 1 kilometer ahead, then the user has already passed the standard 1-mile advance guide sign and the next guide sign they’ll see is directly above the exit. Either we’d need to switch to a heuristic based on remaining duration, or we’d need separate logic for freeways versus surface streets.

1ec5:
What if we suppress the announcement when the new upcoming maneuver is farther away or later than the current upcoming maneuver, but preserve the announcement if the new upcoming maneuver is closer or earlier?

bsudekum:
> What if we suppress the announcement when the new upcoming maneuver is farther away or later than the current upcoming maneuver, but preserve the announcement if the new upcoming maneuver is closer or earlier? 

Sounds good to me.

ericrwolfe:
> What if we suppress the announcement when the new upcoming maneuver is farther away or later than the current upcoming maneuver

If I've missed the maneuver and triggered a reroute, won't the current upcoming maneuver's distance remaining be zero?

1ec5:
>> What if we suppress the announcement when the new upcoming maneuver is farther away or later than the current upcoming maneuver
> 
> If I've missed the maneuver and triggered a reroute, won't the current upcoming maneuver's distance remaining be zero?

#374 calls for silencing the announcement in that case, correct?

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -124,11 +124,11 @@ open class RouteProgress: NSObject {
      - parameter route: The route to follow.
      - parameter legIndex: Zero-based index indicating the current leg the user is on.
      */
-    public init(route: Route, legIndex: Int = 0) {
+    public init(route: Route, legIndex: Int = 0, alertLevel: AlertLevel = .none) {
```
1ec5: This parameter needs documentation (same for the one on RouteLegProgress).

Issue #367, opened 2017-07-12
Extract text instruction generation from voice controllers


ericrwolfe:
The generation and timing of text instructions for voice announcements should be available in CoreNavigation.

cc @bsudekum @frederoni

1ec5:
#385 got backed out in #367.

1ec5:
If understand correctly, this issue tracks moving everything unrelated to speech synthesis to Core Navigation, as #385 attempted to do. It should be unnecessary to go further than that, because the generation of text instructions themselves largely happens in OSRM Text Instructions via RouteStepFormatter, independent of the MapboxNavigation module.

ericrwolfe:
Closed with https://github.com/mapbox/mapbox-navigation-ios/pull/456.

PR #385, opened 2017-07-13
Move text instruction creation to core

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/367

Moves text instruction creation to core and makes the functions `createSpeakText` and `speak` public.


MapboxNavigation.podspec
```diff
@@ -44,9 +44,7 @@ Pod::Spec.new do |s|
   s.module_name = "MapboxNavigation"
 
   s.dependency "MapboxCoreNavigation", "#{s.version.to_s}"
-  s.dependency "MapboxDirections.swift", "~> 0.10.0"
   s.dependency "Mapbox-iOS-SDK", "~> 3.5"
-  s.dependency "OSRMTextInstructions", "~> 0.2.0"
```
bsudekum: Pretty certain since these are in core, we don't need to import these packages.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -346,8 +346,8 @@
 		35C6EDAA1EBB228000A27EF8 /* ca */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = ca; path = ca.lproj/Main.strings; sourceTree = "<group>"; };
 		35C6EDAD1EBB266300A27EF8 /* en */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = en; path = en.lproj/Main.strings; sourceTree = "<group>"; };
 		35C6EDB21EBB286500A27EF8 /* en */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = en; path = en.lproj/Main.strings; sourceTree = "<group>"; };
-		35C9973E1E732C1B00544D1C /* RouteVoiceController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteVoiceController.swift; sourceTree = "<group>"; };
-		35C997401E732C5400544D1C /* RouteStepFormatter.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteStepFormatter.swift; sourceTree = "<group>"; };
+		35C9973E1E732C1B00544D1C /* RouteVoiceController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = RouteVoiceController.swift; path = ../MapboxNavigation/RouteVoiceController.swift; sourceTree = "<group>"; };
```
1ec5: Move RouteVoiceController.swift to the MapboxCoreNavigation folder.

MapboxNavigation.podspec
1ec5: Make this change to MapboxNavigation-Documentation.podspec as well.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -36,12 +37,22 @@ public class PollyVoiceController: RouteVoiceController {
...
+            assert(false)
+            return
+        }
+        speak(instruction: createSpeakText(from: routeProgress, at: userDistance, markUpWithSSML: true), error: nil)
         startAnnouncementTimer()
     }
     
-    override func speak(_ text: String, error: String?) {
-        assert(!text.isEmpty)
+    override public func speak(instruction: String, error: String?) {
```
1ec5: It’s fine to use `_ instruction` here, because `instruction` is precisely what’s being spoken.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -159,6 +159,11 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
             return false
         }
         
+        guard let userDistance = notification.userInfo![RouteControllerAlertLevelDidChangeNotificationDistanceToEndOfManeuverKey] as? CLLocationDistance else {
+            assert(false)
```
1ec5: Instead of `guard`ing, only to fail hard if the condition evaluates to false, it would be clearer to force-unwrap the `userInfo` entry using `as!`.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -170,19 +175,23 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
         return true
     }
     
-    func speechString(notification: NSNotification, markUpWithSSML: Bool) -> String {
-        let routeProgress = notification.userInfo![RouteControllerAlertLevelDidChangeNotificationRouteProgressKey] as! RouteProgress
-        let userDistance = notification.userInfo![RouteControllerAlertLevelDidChangeNotificationDistanceToEndOfManeuverKey] as! CLLocationDistance
+    /**
+     Creates an audio instruction from a `RouteProgress`, the users distance from the maneuver. Optional `markUpWithSSML` will apply `SSML` to the string. This is useful when using AWS's Polly speech synthesizer.
+    */
+    public func createSpeakText(from routeProgress: RouteProgress, at userDistance: CLLocationDistance, markUpWithSSML: Bool = false) -> String {
```
1ec5: `createSpeakText` is ungrammatical; a better method name would be `speechString(describing:at:markUpWithSSML:)`.

MapboxNavigation.podspec
bsudekum: It doesn't look like we require MapboxCoreNavigation in this podspec though. I think we can add it as a dep or leave as is.

1ec5: Ah, I see, that’s because MapboxNavigation-Documentation.podspec [produces a single combined framework](https://github.com/mapbox/mapbox-navigation-ios/blob/c113e518aa38274cf67d64d95c4cb638e732c0a7/MapboxNavigation-Documentation.podspec#L35).

bsudekum:
This is ready.

PR #386, opened 2017-07-14
Don't reroute constantly when not near road

bsudekum:
If the user finds them self away from a road (in a parking lot or in a field), we currently constantly reroute. We had the workings for this before, we were just missing a couple of pieces to make it all work.

This code looks sort of pointless if you assume the user is on a route. Try and picture it that the user is in the middle of the runway at SFO
simulator_screen_shot_jul_14__2017__8_59_51_am
 
In the above screenshot, previously we'd reroute them on every location update because _they are always greater than the reroute distance on their way to the route_. This is unnecessary. Now, we'll reroute them ~twice on their way to their way to the route.

/cc @willwhite @ericrwolfe @1ec5 @frederoni


ericrwolfe:
With `RouteControllerMaximumDistanceBeforeRecalculating` = 50m, my understanding of this is that there'd be a reroute each time the user moved 50m away from the last reroute location. 👍

However, it seems that if the reroute request failed (e.g. while offline), it'd try rerouting on every location update?

bsudekum:
> However, it seems that if the reroute request failed (e.g. while offline), it'd try rerouting on every location update?

True. We can move the resetting of the variable farther up the stack to prevent ddos'ing while in airplane mode and > reroute distance away from the route.

ericrwolfe:
@bsudekum Also, heads up that there's some code in the telem PR `isRerouting` that checks if the app is currently requesting a reroute, and will return early if so: https://github.com/mapbox/mapbox-navigation-ios/pull/304/files#diff-1a2149692b3d7c58b6a0392120d6011aR476

This was added to prevent duplicate reroute telemetry events, so feel free to take a look and see if there's a more comprehensive approach.

bsudekum:
Need to remove commited gpx.

bsudekum:
This is ready for review

Issue #391, opened 2017-07-15
Reroute audio pauses music playback


bsudekum:
Currently the reroute ding sound pauses external music playback, and the music doesn't resume afterwards.

PR #394, opened 2017-07-17
Unduck rerouteSoundPlayer

bsudekum:
The reroute ding is never unducked after player. If the reroute sound plays without a subsequent announcement, the user's music will never resume.

/cc @1ec5 @frederoni


bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/391

Issue #326, opened 2017-06-30
Replace maneuver arrows along the route line


ericrwolfe:
The turn arrows currently implemented in the SDK are both amateur looking and the code is somewhat complex.

Notes from another thread:
> Let's get rid of the line based arrowhead altogether and replace the draw arrowhead with a symbol layer that's just a triangle image that's rotated based on the end bearing of the maneuver. 

> <img width="318" alt="screen shot 2017-06-26 at 9 31 30 am" src="https://user-images.githubusercontent.com/664805/27549877-6225cee6-5a52-11e7-9c2f-9730fa7b6f5f.png">

> For casing, you can have the same triangle image tinted to the casing color, and scaled up slightly.

> Since the arrow extends beyond the maneuver by some distance, possibly curving, the arrow will need `iconRotation` to be set according to the direction of the road at the tip of the arrow’s shaft, not necessarily the maneuver’s heading.

cc @bsudekum @1ec5

1ec5:
Another benefit of implementing the arrowhead as a point-placed symbol is that it can scale better as the user zooms in and out, since its size is expressed in screen units rather than a physical distance (which GL is currently unable to vary according to zoom level). However, the arrow’s shaft will still appear to contract and expand for the same reason as the current arrowhead.

1ec5:
Retitled to distinguish the maneuver arrows along the route line from the maneuver icon in the turn banner, which also looks like an arrow most of the time.

ericrwolfe:
Fixed with https://github.com/mapbox/mapbox-navigation-ios/issues/397

PR #397, opened 2017-07-17
Replace maneuver arrow

frederoni:
WIP #326 

- [ ] Replace the icon with a good looking one
- [ ] Add styling capabilities to the maneuver arrow and shaft
- [ ] Test

@bsudekum @ericrwolfe 👀


bsudekum:
We should also figure out a better layer order here. How does it look underneath street labels?

ericrwolfe:
Would love to see 2 arrowheads: a white one (or one that matches the main stroke) on top, with one that matches the case below and slightly increased in size to roughly match the case width. Also, the current arrowhead looks a little on the small side.

Also, we should change the line cap to `.butt`.

1ec5:
> We should also figure out a better layer order here. How does it look underneath street labels?

Placing the maneuver arrow below any shields, exit numbers, or street labels would be a simple fix. However, I’m concerned that putting anything atop the arrow would make the arrow more difficult to spot, making it look even more like a street:

liberty 2nd

Is there a good use case for reading the current road name off the base map, especially given that a label displays the same information, unrotated, at a consistent location on the screen?

bsudekum:
Updated this with a new a white triangle:
simulator screen shot aug 1 2017 11 08 56 am

ericrwolfe:
That new arrow looks so much better 😄

ericrwolfe:
> Add styling capabilities to the maneuver arrow and shaft

Let's do this as a separate PR cc @frederoni

PR #400, opened 2017-07-18
Feedback UI

frederoni:
Feedback UI

@bsudekum @ericrwolfe 👀


bsudekum:
Would it be possible to split a non UI class out of this that lives in core? This would allow users who use MapboxCoreNavigation to piggy back off the feedback UI class, set appropriate variables and then create their own UI.

ericrwolfe:
@bsudekum you can use the following 3 methods on RouteController to create your own feedback UI: https://github.com/mapbox/mapbox-navigation-ios/blob/12a51b4dffc96b9ec864e9bd20763cfd348bd326/MapboxCoreNavigation/RouteController.swift#L248-L269. Is that what you mean?

MapboxCoreNavigation/Feedback.swift
```diff
@@ -1,12 +1,29 @@
 import Foundation
 
+
 @objc(MBFeedbackType)
 public enum FeedbackType: Int, CustomStringConvertible {
```
bsudekum: Going to need some docs here.

ericrwolfe:
@frederoni added some docs and made feedback events easier to update + cancel. Also cancelled autodismissal on selection and moved dismissal logic to the parent controller so that it's possible to show a confirmation UI afterwards.

@frederoni can you increase the label size as large as possible without wrapping, change the text color to black or near black for added contrast, and change the following label names (keeping at a max of 11 chars)?

* `Turn not allowed` ➡️ `Not allowed`
* `Road closed` ➡️ `Closure`
* `Routing error` ➡️  `Confusing`
* `Other` ➡️ `Other issue`

@willwhite I'm still not sure I feel the route issue choices list is the right balance between clear and simple though comprehensive. Do you think the list covers the majority of issues you've been seeing? Part of me wants to replace `Hazard` with `Closure` for now to give space for another route issue option such as `Wrong` (instruction or route) to complement `Confusing`. We can bring `Hazard` back later once we have a better idea of the issues drivers are facing.

PR #402, opened 2017-07-19
Reinvent the course tracking wheel

### LINKED ISSUE #323: Allow updating location puck bottom offset
This is an upstream issue in mapbox-gl-native (@1ec5 I can't recall the ticket), but based on the release cycles of mapbox-gl-native, we should consider adding a workaround to the navigation SDK that allows developers to customize the bottom location puck offset while in course tracking. 

This is a big hurdle for implementing custom navigation apps from scratch.

cc @bsudekum

### LINKED ISSUE #352: Bespoke course tracking
This issue tracks implementing turn-by-turn-specific course tracking behavior within NavigationMapView instead of relying on the course tracking built into the map SDK. That would allow us to:

* More cleanly expose user puck customization to the developer (#213)
* Clean up or simulate location updates without the redeclaration hack and without interference from the map SDK (#323, #341)
* More precisely customize the on-screen location of the user puck (#324)
* Continue to display a puck instead of a dot as the user pans away

To be sure, each of those issues is surmountable on its own, but taken together, they add up to code that’s unintuitive and fragile. The map SDK’s built-in course tracking is actually not that complicated. What’s complicated are the transitions to other user tracking modes, but we might want custom behavior for that anyways.

/cc @frederoni @bsudekum @ericrwolfe @boundsj

### LINKED ISSUE #370: Throttle map rendering when device unplugged
Right now, rendering 60fps animations between location updates is taxing on the device + battery. When the device is unplugged, we should not animate between location updates.

cc @bsudekum @1ec5

1ec5:
This PR reimplements MGLMapView’s course tracking mode in MapboxNavigation. Various hacks that we previously employed to force-snap MGLMapView’s user puck have been removed. The puck continues to look like a puck during turn-by-turn navigation after panning away.

straight ahead overhead

The developer can customize the user puck by setting the `NavigationMapView.userCourseView` property, which is easier to work with than a delegate method. The default user puck now looks more similar to what the Android navigation SDK displays: a larger disc that lies flatter against the map (with a halo instead of a shadow).

The developer can also customize the user puck’s location on screen by implementing the `NavigationViewControllerDelegate.navigationViewController(_:mapViewUserAnchorPoint:)` method.

More to come:

* [x] Move off `MGLUserTrackingModeFollowWithCourse` (to fix #352)
* [x] Align the camera with the location view’s bottom-weighted position
* [x] Make the user puck’s position customizable (to fix #323)
* [x] Implement a better-looking user puck (or port the relevant parts of MGLFaux3DUserLocationAnnotationView)
* [x] Avoid automatically tracking the user location when in non-tracking mode
* [x] Avoid bouncing back and forth between overview and front-facing views when in overview mode
* [x] Override pinch and tap gesture recognizers to center on the user puck
* [x] Disable animated camera transition while on battery power (to fix #370)
* [x] Hide Recenter button while tracking user course
* [x] Interpolate between location updates (not as much as 60fps, but more than 1fps)
* [ ] Disable implicit animations of the user course view when not tracking course
* [ ] Update example application to customize the new-style user course view

/cc @frederoni @bsudekum


MapboxNavigation/NavigationMapView.swift
```diff
@@ -49,20 +38,60 @@ open class NavigationMapView: MGLMapView {
...
+        didSet {
+            guard let location = userLocationForCourseTracking, CLLocationCoordinate2DIsValid(location.coordinate) else {
+                return
+            }
+            
+            // didUpdateLocationWithUserTrackingAnimated
+            // didUpdateLocationIncrementallyAnimated
+            // TODO: Call setVisibleCoordinateBounds instead to account for the location view’s bottom-weighted position on-screen.
+            let camera = MGLMapCamera(lookingAtCenter: location.coordinate, fromDistance: 300, pitch: 45, heading: location.course)
+            setCamera(camera, animated: true)
```
1ec5: For #370, we need to turn animation off, at least when the device is on battery power. If we leave animation on, we’d also need to enable a linear timing curve and stretch out the animation for one second to eliminate the jumping that currently occurs.

frederoni:
Note that this PR currently only builds with a custom build of the iOS Maps SDK that exposes `-[MGLMapView setCamera:withDuration:animationTimingFunction:edgePadding:completionHandler:]`

bsudekum:
Wow, this rebase is rough times.

bsudekum:
The rebase was so rough I just started fresh. @frederoni @1ec5 let me know if you want me to force push this branch over this one: https://github.com/mapbox/mapbox-navigation-ios/compare/bs-bespoke-course-352

1ec5:
That branch missed some changes. The rebase doesn’t seem so daunting in any case, although I’m getting some compiler errors in unrelated code.

> Also, do we want to move some of this logic to core? I sort of remember this being part of the conversion around this PR but I'm seeing a lot import UIKit.

No, we shouldn’t have Core Navigation depend on the map SDK nor on UIKit if we can help it.

1ec5:
OK, I’ve rebased the PR. Currently there are two MGLMapView categories containing forward declarations, one in MapboxNavigation and the other in the Swift example application. Hopefully we can eliminate the former by taking advantage of mapbox/mapbox-gl-native#9651 and MGLMapViewDelegate.

ericrwolfe:
@1ec5 what're the next steps to finishing this PR out?

1ec5:
> what're the next steps to finishing this PR out?

I’ve been keeping the original post updated with the remaining work. Currently we have the following to-do items:

* Avoid automatically tracking the user location when in non-tracking mode.
  * You can pan the map, but the viewport goes right back to the front-facing view on every location update. This should be pretty simple to fix.
* Avoid bouncing back and forth between overview and front-facing views when in overview mode.
  * When you tap Overview, the viewport switches back to the front-facing view on every location update, only to immediately zoom back out to the overall route.
* Override pinch and tap gesture recognizers to center on the user puck.
  * You don’t really notice this effect given the two issues above, but zoom and tap gestures are currently centered on the center of the map instead of the bottom-weighted puck. This may require exposing an additional MGLMapView internal method.
* Interpolate between location updates.
  * With a 1fps frame rate, it’s just barely possible to follow where the camera is going. But at turns, the camera’s rotation can change significantly, easily causing the user to get disoriented. We’ll need to use a timer in addition to whatever timer the location manager may be using.

frederoni:
@1ec5 How do you feel about migrating the user puck to runtime styling? Porting the relevant parts from https://github.com/mapbox/mapbox-plugins-android/pull/22?

It would solve the issue we are seeing when starting navigation and going from landscape to portrait where the user puck floats around because the camera and UIView animations are not in sync. I think it could also simplify course tracking. Though, styling it would be slightly more limited.

1ec5:
> It would solve the issue we are seeing when starting navigation and going from landscape to portrait where the user puck floats around because the camera and UIView animations are not in sync.

I don’t think this behavior is inherent to using a UIView. In fact, 049814ba3798c41cb31570124f9b33de1e0d8b1b nearly eliminates the initial jump by making the various cameras more consistent with each other. The remaining minor jump seems to correspond to a view’s height – I wonder if we’re initially forgetting to account for a view in the edge padding.

> I think it could also simplify course tracking. Though, styling it would be slightly more limited.

There are pros and cons to using runtime styling for the user puck:

* Con: In user tracking mode, the user dot or user puck must remain fixed on the screen while the camera moves. Mapbox GL has no concept of UI elements that behave completely independently of the map. If we implement the user puck as style layers, as the Android navigation SDK does, we’d have to work around this mismatch by manually updating the puck’s geometry on every frame. That isn’t unreasonably difficult, but it isn’t simpler than what we’re already doing.<p>Part of the motivation behind this PR is to reduce battery usage by putting less work onto mbgl’s shoulders. But if you’ve ever played with the Add Animated Annotation demo in macosapp, you know how resource-intensive frequently changing a single point feature’s coordinate can be, and that’s without moving the camera.

* Weak pro: Once mapbox/mapbox-gl-native#9479 makes it into a release, we’d get a realistic tilting effect using `iconPitchAlignment`, something I’ve struggled with for view-backed annotations (mapbox/mapbox-gl-native#5245). However, the difference would only be noticeable if we were using a rectangular user puck as opposed to the round one we’re using now.

* Con: Mapbox GL has no concept of 3D models (mapbox/mapbox-gl-js#3996), whereas [a SceneKit model](https://blog.classycode.com/how-to-write-a-pok%C3%A9mon-go-clone-for-ios-edf1cf1cf5ce) can already be used as a user puck. (Think 3D car model.)

/ref https://github.com/mapbox/cartography/issues/109#issuecomment-290270488

MapboxNavigation/MGLMapView+MGLNavigationAdditions.h
```diff
@@ -2,12 +2,19 @@
...
+                 edgePadding:(UIEdgeInsets)insets
+                   zoomLevel:(double)zoomLevel
+                   direction:(CLLocationDirection)direction
+                    duration:(NSTimeInterval)duration
+     animationTimingFunction:(nullable CAMediaTimingFunction *)function
+           completionHandler:(nullable void (^)(void))completion;
 
-// FIXME: This will be removed once https://github.com/mapbox/mapbox-navigation-ios/issues/352 is implemented.
-- (void)validateLocationServices;
+- (void)setCenterCoordinate:(CLLocationCoordinate2D)centerCoordinate
```
bsudekum: Might be confusing this with another method, but don't we have this in native now?

frederoni: We added setCamera that takes edge padding. This was just during development. I think we can delete MGLMapView+MGLNavigationAdditions.h now.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -59,7 +59,25 @@ class RouteMapViewController: UIViewController {
...
+                mapView.logoView.isHidden = true
+            } else {
+                overviewButton.isHidden = false
+                recenterButton.isHidden = true
+                mapView.logoView.isHidden = false
+            }
+            
+            if let controller = routePageViewController.currentManeuverPage {
+                controller.step = currentStep
+                routePageViewController.updateManeuverViewForStep()
```
frederoni: @1ec5 What's the purpose of updating the page here?
It looks like `controller.step = currentStep` will be overriden by `upComingStep ?? currentStep` in `updateManeuverViewForStep()`.

frederoni:
@1ec5 I tried two different approaches regarding:
> Interpolate between location updates (not as much as 60fps, but more than 1fps)

```
Normal: Drop no frames (Target 60 FPS)
15-40% CPU usage

Drop 12 frames per update (Target 5 FPS)
5-15% CPU usage

Interpolate 5 location points per location update (Target 5 FPS)
5-30% CPU usage
```

and came to the conclusion that modifying the display link’s frame interval seems to be the cleanest and most efficient way. The decreased frame rate kicks in after 10 seconds of inactivity while unplugged.

1ec5:
Yikes, I&rsquo;m leery of touching the private CADisplayLink at all. We&rsquo;ve have plenty of nasty behavior (crashes, leaks) in the past caused by CADisplayLink, particularly the fact that it holds a strong reference to MGLMapView.

What was the other approach you had in mind for &ldquo;Interpolate between location updates&rdquo;? I was thinking we&rsquo;d use Geometry.swift to calculate points along the route (similar to what SimulatedLocationManager does) and issue explicit camera updates.

frederoni:
> I was thinking we’d use Geometry.swift to calculate points along the route (similar to what SimulatedLocationManager does) and issue explicit camera updates.

Interpolation using Geometry.swift is what I tried, but the result was not satisfactory, partly because the implementation was convoluted but also because the CPU usage didn't drop noteworthy.

I can give it another try but I'd assume moving off of CADisplayLink would expose another way to tweak the frame interval so it would be a simple change when that lands in the Maps SDK.

MapboxNavigation/UserCourseView.swift
```diff
@@ -113,5 +113,143 @@ class UserPuckCourseView: UIView, UserCourseView {
...
+    
+    var location: CLLocation?
     
+    var pitch: CLLocationDegrees = 0 {
+        didSet {
+            if oldValue != pitch {
+                let newPitch = pitch
+                UIView.animate(withDuration: 1, delay: 0, options: [.beginFromCurrentState, .curveLinear], animations: {
+                    var transform = CATransform3DRotate(CATransform3DIdentity, CGFloat(CLLocationDegrees(newPitch).toRadians()), 1.0, 0, 0)
+                    transform.m34 = -1.0 / 1000 // (-1 / distance to projection plane)
```
frederoni: @1ec5 would be cool to do something similar in `MGLAnnotationView` but it should be based on the actual distance from the camera to each individual annotation view.

1ec5: Yes, that’s something I explored in mapbox/mapbox-gl-native#5245. Though it seems like mbgl should expose the actual transform matrix it uses (to transform features in circle layers, for example), to reduce the guesswork.

MapboxNavigation/UserCourseView.swift
```diff
@@ -113,5 +113,143 @@ class UserPuckCourseView: UIView, UserCourseView {
             puckArrow?.setAffineTransform(CGAffineTransform.identity.rotated(by: -CGFloat(angle.toRadians())))
         }
     }
+}
+
+class NewUserPuckCourseView: UIView, UserCourseView {
```
1ec5: The idea is to eventually replace UserPuckCourseView with NewUserPuckCourseView, right?

frederoni: Yes

1ec5:
The recent CADisplayLink changes lead to a very strange effect when the device is plugged in: the puck marches ahead along the route a few times without the camera moving, then the camera lurches forward at an irregular rate, pulling the puck back with animation. As a result, the puck appears to take two steps forward and one step back.

2-and-fro

The puck shouldn’t appear to move on screen while tracking course. This occurs with both UserPuckCourseView and NewUserPuckCourseView. It didn’t occur back in 2a570e22816af14143f67ec0492baed0078b740e, because the user puck was updated at the same interval with as the camera.

frederoni:
I suspect a merge conflict went wrong or similar. The CADisplayLink’s update rate is not tied to the user puck’s position so it shouldn't affect it at all? Any other changes? What happened to the new puck? 🤔

1ec5:
That animation is with the old puck, but the same effect occurs with both.

frederoni:
@1ec5 I can't reproduce the jittering either before or after https://github.com/mapbox/mapbox-navigation-ios/pull/402/commits/b463c6d1730552c502584cdaa014153f4fdbf43e.

Can you try the latest commit and explain how you reproduce if the jittering is still there?
It looks like the frame rate drops too early in your case. Are you forcing a low frame rate from the start?

ericrwolfe:
@frederoni @1ec5 what to we need to finish this out and merge in? Can we split the frame rate throttling or anything else that doesn't bring us to existing parity into a follow on PR?

frederoni:
@bsudekum @1ec5 This PR is ready for testing and review. As @ericrwolfe mentioned, throttling will be enabled in another PR.

bsudekum:
gif

@frederoni is there anything we can do about the animation issues here?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -504,31 +510,12 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
-            } else {
-                recenterButton.isHidden = true
-                mapView.logoView.isHidden = false
-            }
-        }
-        
-        if isInOverviewMode {
-            overviewButton.isHidden = false
-            recenterButton.isHidden = true
-            mapView.logoView.isHidden = false
```
bsudekum: Logo should be hidden when we leave followWithCourse.

bsudekum:
gif

Zooming sees the same animation issue. Can we allow the zoom level to remain after the user changes it?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -34,75 +35,223 @@ open class NavigationMapView: MGLMapView {
...
+    
+    var userLocationForCourseTracking: CLLocation?
+    var animatesUserLocation: Bool = false
+    
+    fileprivate let inactivityInterval: TimeInterval = 10
+    fileprivate let decreasedFrameInterval: Int = 12
+    
+    func resetInactivityTimer() {
+        NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(inactivityTimerFinished), object: nil)
+        self.perform(#selector(inactivityTimerFinished), with: nil, afterDelay: inactivityInterval)
```
bsudekum: I'm not seeing this fire currently. After moving the map away from followWithCourse, it just remains.

MapboxNavigation/UserCourseView.swift
```diff
@@ -0,0 +1,144 @@
...
+}
+
+class UserPuckStyleKitView: UIView {
+    
+    override func draw(_ rect: CGRect) {
+        super.draw(rect)
+        drawNavigation_puck()
+    }
+    
+    func drawNavigation_puck(fillColor: UIColor = UIColor(red: 1.000, green: 1.000, blue: 1.000, alpha: 1.000), puckColor: UIColor = UIColor(red: 0.149, green: 0.239, blue: 0.341, alpha: 1.000), shadowColor: UIColor = UIColor(red: 0.149, green: 0.239, blue: 0.341, alpha: 0.160), circleColor: UIColor = UIColor(red: 1.000, green: 1.000, blue: 1.000, alpha: 1.000)) {
```
bsudekum: Should we use the global tint color for the puck?

MapboxNavigation/NavigationMapView.swift
frederoni: Any gesture (rotate/pan/pinch) should trigger this. Could it be that fling (pan with velocity) doesn't trigger it?

bsudekum: Oh, I was thinking this was a recenter timer after no activity.

bsudekum:
Nice, this is looking a lot better! 
gif

bsudekum:
As a followup task, we should scale the size of the user puck when zooming out. In overview mode, it's rather large.

bsudekum:
Final task I'm seeing: https://github.com/mapbox/mapbox-navigation-ios/pull/402#discussion_r141098621

frederoni:
@bsudekum That's fixed here https://github.com/mapbox/mapbox-navigation-ios/blob/efd2007ae87c9514bf19facb251f62839dedea2f/MapboxNavigation/RouteMapViewController.swift#L363-L373

bsudekum:
gif

@frederoni Can you checkout why this PR would have an impact on other markers?

frederoni:
@bsudekum 

> Can you checkout why this PR would have an impact on other markers?

fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/402/commits/64bd6d8e984b0c316eb89073f4d3b11223f11f6a

bsudekum:
It looks like this PR removes the current way name label.

bsudekum:
Overview mode is sort of difficult to use: 
img_1421

Previously, it worked since the user location was much smaller when in overview mode. Might be worth looking into if we can get an easy win here.

1ec5:
> It looks like this PR removes the current way name label.

I’m still seeing the label:

reading

Does the road you’re testing out suffer from the same issue on master?

1ec5:
> Previously, it worked since the user location was much smaller when in overview mode. Might be worth looking into if we can get an easy win here.

In general, the user puck provided by the iOS map SDK is a lot smaller than the design we’re copying from the Android navigation SDK. The user dot is smaller still, but it’s nice that the puck design has an arrow to indicate the user’s direction of travel, even in overview mode. 7d4cb829c1c689d5f072e974e3405bb1926f1256 shrinks the puck by half when in overview mode. If the developer provides their own puck implementation, they can key off `tracksUserCourse` for the same effect.

straight ahead overhead

1ec5:
As noted in the PR description above, the Swift example application’s custom user location annotation view (off by default) is broken now that we’ve moved away from the annotation API:

https://github.com/mapbox/mapbox-navigation-ios/blob/3c02a54f623196547574664c7943635210100463/Examples/Swift/ViewController.swift#L265-L293

bsudekum:
@frederoni if quick, do you think you could update the example app to show how to add a custom user puck view?

Examples/Swift/MGLMapView+CustomAdditions.h
```diff
@@ -0,0 +1,10 @@
+#import <Mapbox/Mapbox.h>
+
+@interface MGLMapView (CustomAdditions) <CLLocationManagerDelegate>
+
+// FIXME: This will be removed once https://github.com/mapbox/mapbox-gl-native/issues/6867 is implemented
+- (void)locationManager:(CLLocationManager *)manager didUpdateLocations:(NSArray<CLLocation *> *)locations;
+
+@property (nonatomic, readonly) CLLocationManager *locationManager;
+
+@end
```
bsudekum: @frederoni is this file necessary?

frederoni: Unfortunately, it's necessary until we rewrite the custom example to use something like the new course tracking view. We could reuse this implementation but having the custom example rely on MapboxNavigation contradicts the whole purpose of the custom navigation example.

bsudekum: @frederoni so if we were to merge this now, in what case is it necessary for developers to add this file?

frederoni: You only need it for simulating location updates in the custom example/MapboxCoreNavigation. Same as prior to this PR but we moved it from `MGLMapView+MGLNavigationAdditions.h` to `+CustomAdditions`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -54,56 +51,248 @@ open class NavigationMapView: MGLMapView {
...
+            if let userCourseView = userCourseView {
+                addSubview(userCourseView)
+            }
         }
     }
     
-    override open func validateLocationServices() {
-        if !manuallyUpdatesLocation {
-            super.validateLocationServices()
+    var userCourseViewCenter: CGPoint {
```
1ec5: This variable is unused, because `targetPoint` was added at some point to do the same thing. `targetPoint` is pretty unsophisticated; can we go back to something like this that respects the content insets?

1ec5:
As of 8675bed37f9050195a5844e4552b513b6d9d4d26, the user puck’s position is slightly higher than it was before, but now it respects the map view’s content insets and avoids going too high or too low as the step table pulley varies in height.

default step table

The developer can customize the user puck’s location on screen by implementing the `NavigationViewControllerDelegate.navigationViewController(_:mapViewUserAnchorPoint:)` method. Here’s an implementation that sticks the puck in the upper-left corner of the map view:

top-left

bsudekum:
6bc33392-35bd-40b9-848f-dff0bc73aa82

Looks like the user puck is slightly off center.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -54,55 +51,252 @@ open class NavigationMapView: MGLMapView {
...
+        if keyPath == "batteryState" {
+            let batteryState = UIDevice.current.batteryState
+            isPluggedIn = batteryState == .charging || batteryState == .full
+        }
+    }
+    
+    func updateCourseView(_ sender: UIGestureRecognizer) {
+        
+        // Disable course tracking when any kind of user interaction occurs
+        if tracksUserCourse {
```
hervouinc: Why disable course tracking here ?
Isn't it the purpose of the selector: `disableUserCourseTracking` which is only triggered by rotation and pan gesture recognizers ?

MapboxNavigation/NavigationMapView.swift
hervouinc: I do not see any call to `makeGestureRecognizersResetInactivityTimer` in this PR, is it the right behaviour ?

1ec5:
> Looks like the user puck is slightly off center.

Judging from the screenshot, the puck is dead-center on screen, but the user’s location is somewhat offset from the route line. That shouldn’t be a consequence of this PR, though – are you seeing the same behavior on master?

bsudekum:
@1ec5 from testing this weekend, it doesn't really look like we're snapping at all.

MapboxNavigation/NavigationMapView.swift
frederoni: Yes, throttling is currently not enabled.

bsudekum:
@1ec5 @frederoni @ericrwolfe With https://github.com/mapbox/mapbox-navigation-ios/pull/402/commits/429aac9fa5ccc4efb0b244199ad400c1ed5fb69f, I think this about ready to go.

I'd like to still see the example updated though, ref https://github.com/mapbox/mapbox-navigation-ios/pull/402#issuecomment-334808373

bsudekum:
Per chat yesterday, @frederoni is going to see what can be done to modify the zoom level without leaving followWithCourse.

bsudekum:
gif2

@frederoni would it be possible to disable animation so the puck is more stuck to the route to fix ^?

frederoni:
Fixed https://github.com/mapbox/mapbox-navigation-ios/pull/402#issuecomment-336221025

tackswipe

1ec5:
The merge was redone without squashing as b14107790a7647496e597590846a6c7ec77817cb.

PR #403, opened 2017-07-20
Update ETA text color to match congestion on route

bsudekum:
todo:

- [x] update according to remaining segments on route
- [x] use color in route line
simulator screen shot jul 20 2017 11 19 32 am

/cc @frederoni @1ec5


MapboxNavigation/RouteTableViewController.swift
```diff
@@ -46,6 +47,32 @@ class RouteTableViewController: UIViewController {
...
+            var counts: [CongestionLevel: TimeInterval] = [:]
+
+            for (segmentCongestion, segmentTime) in zip(combinedCongestionLevel, combinedTimes) {
+                counts[segmentCongestion] = (counts[segmentCongestion] ?? 0) + segmentTime
+            }
+
+            if let max = counts.max(by: { a, b in a.value < b.value }) {
+                switch max.key {
+                case .unknown:
+                    return
```
1ec5: Is it possible for an `unknown` to appear in a response that otherwise has known congestion values? Would it be better to try and account for the unknown slot than to bail?

Examples/Swift/ViewController.swift
```diff
@@ -117,7 +117,7 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         options.profileIdentifier = .automobileAvoidingTraffic
         
         // Adding the optional attribute `.congestionLevel` ensures the route line will show the congestion along the route line
```
1ec5: Also add a comment explaining why `.expectedTravelTime` appears here.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -46,6 +47,40 @@ class RouteTableViewController: UIViewController {
...
         } else {
             headerView.timeRemaining.text = dateComponentsFormatter.string(from: routeProgress.durationRemaining)
+            
+            guard let coordinates = routeProgress.route.coordinates else { return }
+            
+            // To keep this cheap, use the fraction traveled along the current route.
+            // From this, we can estimate what coordinates in the route remain.
+            // This color change does not need to be 100% accurate, 
+            // just a rough estimation of remaining route congestion.
+            let estimatedCoordinatesRemaining = Int(floor(Double(coordinates.count) * routeProgress.fractionTraveled))
```
bsudekum: @1ec5 how do you feel about this logic?

1ec5: Hard to say. The number of coordinates in a given step has no relation to its distance. For example, a short, windy stretch of road may have many more coordinates than a longer, straighter stretch of road.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -46,6 +47,40 @@ class RouteTableViewController: UIViewController {
             headerView.timeRemaining.text = String.localizedStringWithFormat(NSLocalizedString("LESS_THAN", bundle: .mapboxNavigation, value: "<%@", comment: "Format string for less than; 1 = duration remaining"), dateComponentsFormatter.string(from: 61)!)
         } else {
             headerView.timeRemaining.text = dateComponentsFormatter.string(from: routeProgress.durationRemaining)
+            
+            guard let coordinates = routeProgress.route.coordinates else { return }
+            
+            // To keep this cheap, use the fraction traveled along the current route.
```
1ec5: It would be very convenient to know which segments correspond to a given step, just as we know the geometry along an individual step. When initializing NavigationViewController with a route or when a new route comes in, let’s build a dictionary mapping steps to their segments, which we can reuse each time this method is called.

MapboxNavigation/RouteTableViewController.swift
bsudekum: (Note, I'm looking at the overall route here but your point is still valid.)

MapboxNavigation/RouteTableViewController.swift
bsudekum: @1ec5 how would this help this PR? We're not really too interested in the given step, rather the entire route.

1ec5: Aren’t we considering the total of expected travel times along all the remaining segments of the route? We could recompute that every time this method is called, which you’ve studiously avoided, but one optimization would be to essentially index the segments by step so that it would be cheaper to get all the remaining segments starting from the current location. So when this method is called, we’d sum the expected travel times on all the remaining segments on the current step plus all the remaining steps.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +135,30 @@ open class RouteProgress: NSObject {
...
+        
+        var coordinateIndex = 0
+        for leg in route.legs {
+            
+            var congestionTravelTimesSegmentsByLeg: [[(CongestionLevel, TimeInterval)]] = []
+            
+            for step in leg.steps {
+                if let lastCoord = step.coordinates?.last, let coordinates = route.coordinates, let segmentCongestionLevels = leg.segmentCongestionLevels, let expectedSegmentTravelTimes = leg.expectedSegmentTravelTimes  {
+                    guard let nextIndex = coordinates.index(where: {
+                        $0.latitude == lastCoord.latitude && $0.longitude == lastCoord.longitude
```
bsudekum: @1ec5 @frederoni It looks like this is only truthy once; on the first step. Could there be a rounding error?

1ec5: Yeah, comparing doubles is bound to fail. Instead, let's take advantage of the fact that `nextIndex` is always `coordinateIndex + step.coordinates.count` to avoid a somewhat expensive call to `index(where:)`.

bsudekum:
@1ec5 @frederoni  this is ready for review

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +135,30 @@ open class RouteProgress: NSObject {
         self.legIndex = legIndex
         super.init()
         currentLegProgress = RouteLegProgress(leg: currentLeg, stepIndex: 0, alertLevel: alertLevel)
+        
+        var coordinateIndex = 0
```
1ec5: `maneuverCoordinateIndex` would be more descriptive. A documentation comment like this would help as well (you can give local variables documentation comments too):

> An index into the route’s `coordinates` and `congestionTravelTimesSegmentsByStep` that corresponds to a step’s maneuver location.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +135,30 @@ open class RouteProgress: NSObject {
...
+            
+            var congestionTravelTimesSegmentsByLeg: [[(CongestionLevel, TimeInterval)]] = []
+            
+            for (stepIndex, step) in leg.steps.enumerated() {
+                if let segmentCongestionLevels = leg.segmentCongestionLevels, let expectedSegmentTravelTimes = leg.expectedSegmentTravelTimes  {
+                    
+                    let stepCoordinatesCount = Int(step.coordinateCount) + coordinateIndex - 1
+                    
+                    let congestionSegment = Array(segmentCongestionLevels[coordinateIndex-stepIndex..<stepCoordinatesCount - stepIndex])
+                    let travelTimeSegment = Array(expectedSegmentTravelTimes[coordinateIndex-stepIndex..<stepCoordinatesCount - stepIndex])
```
1ec5: `stepSegmentCongestionLevels` and `stepSegmentTravelTimes`

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +135,30 @@ open class RouteProgress: NSObject {
...
+                    
+                    let zipped = zip(congestionSegment, travelTimeSegment)
```
1ec5: `stepTimedCongestionLevels`

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +135,30 @@ open class RouteProgress: NSObject {
...
         super.init()
         currentLegProgress = RouteLegProgress(leg: currentLeg, stepIndex: 0, alertLevel: alertLevel)
+        
+        var coordinateIndex = 0
+        for leg in route.legs {
+            
+            var congestionTravelTimesSegmentsByLeg: [[(CongestionLevel, TimeInterval)]] = []
+            
+            for (stepIndex, step) in leg.steps.enumerated() {
+                if let segmentCongestionLevels = leg.segmentCongestionLevels, let expectedSegmentTravelTimes = leg.expectedSegmentTravelTimes  {
```
1ec5: Move this `if let` outside the inner `for` loop, since the variables will have the same values for every step along the leg.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +135,30 @@ open class RouteProgress: NSObject {
...
+                    
+                    let stepCoordinatesCount = Int(step.coordinateCount) + coordinateIndex - 1
+                    
+                    let congestionSegment = Array(segmentCongestionLevels[coordinateIndex-stepIndex..<stepCoordinatesCount - stepIndex])
```
1ec5: If I understand correctly, `coordinateIndex` is an index into a leg’s array of coordinates, while `stepIndex` is an index into a leg’s array of steps. Why is it correct to subtract one from the other, given that a step may have more than one coordinate?

MapboxNavigation/Style.swift
```diff
@@ -196,6 +196,11 @@ public class Style: NSObject {
     public var trafficSevereColor: UIColor?
     
     /**
+     Sets an alternate color used throughout the UI to denote low traffic congestion. Not used to style the route line.
+     */
+    public var trafficAlternateLowColor: UIColor?
```
1ec5: Given that this color is used to style text, not the route line, let’s call this property `lowTrafficTextColor`.

(It would be more grammatical to call the other properties `lowTrafficColor`, `moderateTrafficColor`, etc., but I guess that can be a separate PR.)

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -44,8 +45,45 @@ class RouteTableViewController: UIViewController {
...
+            let coordinatesRemainingOnStep = Int(floor((Double(routeProgress.currentLegProgress.currentStepProgress.step.coordinateCount)) * routeProgress.currentLegProgress.currentStepProgress.fractionTraveled))
+            
+            guard coordinatesRemainingOnStep >= 0 else {
+                headerView.timeRemaining.textColor = TimeRemainingLabel.appearance(for: traitCollection).textColor
+                return
+            }
+            
+            let segementsForStep = routeProgress.congestionTravelTimesSegmentsByStep[routeProgress.legIndex][routeProgress.currentLegProgress.stepIndex]
+            
+            guard coordinatesRemainingOnStep <= segementsForStep.count else { return }
```
1ec5: I wonder if we should assert that this is the case.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -44,8 +45,45 @@ class RouteTableViewController: UIViewController {
...
+            
+            let currentStepCongestionTimes = segementsForStep.suffix(from: coordinatesRemainingOnStep)
```
1ec5: This could be useful functionality to put on RouteStepProgress.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -44,8 +45,45 @@ class RouteTableViewController: UIViewController {
         
         if routeProgress.durationRemaining < 60 {
             headerView.timeRemaining.text = String.localizedStringWithFormat(NSLocalizedString("LESS_THAN", bundle: .mapboxNavigation, value: "<%@", comment: "Format string for less than; 1 = duration remaining"), dateComponentsFormatter.string(from: 61)!)
+            headerView.timeRemaining.textColor = TimeRemainingLabel.appearance(for: traitCollection).textColor
         } else {
             headerView.timeRemaining.text = dateComponentsFormatter.string(from: routeProgress.durationRemaining)
+            
+            let coordinatesRemainingOnStep = Int(floor((Double(routeProgress.currentLegProgress.currentStepProgress.step.coordinateCount)) * routeProgress.currentLegProgress.currentStepProgress.fractionTraveled))
```
1ec5: `countOfCoordinatesLeftOnStep` so it’s clear that the property is a number, not an array of coordinates.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -44,8 +45,45 @@ class RouteTableViewController: UIViewController {
...
+                headerView.timeRemaining.textColor = TimeRemainingLabel.appearance(for: traitCollection).textColor
+                return
+            }
+            
+            let segementsForStep = routeProgress.congestionTravelTimesSegmentsByStep[routeProgress.legIndex][routeProgress.currentLegProgress.stepIndex]
+            
+            guard coordinatesRemainingOnStep <= segementsForStep.count else { return }
+            
+            let currentStepCongestionTimes = segementsForStep.suffix(from: coordinatesRemainingOnStep)
+            let remainingStepSegments = Array(routeProgress.congestionTravelTimesSegmentsByStep[routeProgress.legIndex].suffix(from: routeProgress.currentLegProgress.stepIndex)).joined()
```
1ec5: If I’m not mistaken, this could still be a huge array unless we find a way to flatten or precompute the expected travel time buckets for later steps. Can we add another dictionary to `RouteProgress` that would make it possible to change this code to something like the following sketch?

```swift
let timedCongestionLevelsLeftOnStep: [TimedCongestionLevel] = stepTimedCongestionLevels.suffix(from: countOfCoordinatesLeftOnStep)
// var travelTimesByCongestionLevelByStep: [[[CongestionLevel: TimeInterval]]]
let timesLeftOnLegByCongestionLevel: [[CongestionLevel: TimeInterval]] = routeProgress.travelTimesByCongestionLevelByStep[routeProgress.legIndex].suffix(from: legProgress.stepIndex)

// `sum(_:)` is really a call to `reduce(_:_:)`
// The for loop before is OK too – just wrote it this way for clarity.
let lowTimeLeftOnStep: TimeInterval = sum(timedCongestionLevelsLeftOnStep.filter { $0.0.low }.map { $0.1 })
let lowTimeLeftOnLeg: TimeInterval = sum(timesLeftOnLegByCongestionLevel.map { $0.low })
let remainingLowTime = lowTimeLeftOnStep + lowTimeLeftOnLeg
```

bsudekum:
Updated this, but I'm still off by 1 somewhere. The remaining step congestion times do not add up quite yet.

bsudekum:
Got this working 🎉 . @1ec5 ready for another round of review.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -118,11 +118,16 @@ open class RouteProgress: NSObject {
      */
     public var currentLegProgress: RouteLegProgress!
     
+    public typealias TimedCongestionLevel = (CongestionLevel, TimeInterval)
```
1ec5: Since this typealias is public and the method arguments that use it are also public, the typealias needs documentation.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -61,7 +60,7 @@ class RouteTableViewController: UIViewController {
             guard coordinatesLeftOnStepCount <= congestionTimesForStep.count else { return }
             
             let remainingCongestionTimesForStep = congestionTimesForStep.suffix(from: coordinatesLeftOnStepCount)
-            let remainingCongestionTimesForRoute = routeProgress.congestionTimesPerStep[routeProgress.legIndex][routeProgress.currentLegProgress.stepIndex + 1..<routeProgress.currentLeg.steps.count]
+            let remainingCongestionTimesForRoute = routeProgress.congestionTimesPerStep[routeProgress.legIndex].suffix(from: routeProgress.currentLegProgress.stepIndex + 2)
```
1ec5: This will crash if `congestionTimesPerStep[routeProgress.legIndex]` has only two items and we’re on step 1.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +142,38 @@ open class RouteProgress: NSObject {
...
+            congestionTimesPerStep.append([])
+            
+            /// An index into the route’s coordinates and congestionTravelTimesSegmentsByStep that corresponds to a step’s maneuver location.
+            var congestionTravelTimesSegmentsByLeg: [[TimedCongestionLevel]] = []
+            
+            if let segmentCongestionLevels = leg.segmentCongestionLevels, let expectedSegmentTravelTimes = leg.expectedSegmentTravelTimes  {
+                for (stepIndex, step) in leg.steps.enumerated() {
+                    let stepCoordinatesCount = Int(step.coordinateCount) + maneuverCoordinateIndex - 1
+                    
+                    let congestionSegment = Array(segmentCongestionLevels[maneuverCoordinateIndex-stepIndex..<stepCoordinatesCount - stepIndex])
```
1ec5: Can you answer [this question](https://github.com/mapbox/mapbox-navigation-ios/pull/403#discussion_r129901384)?

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +142,38 @@ open class RouteProgress: NSObject {
...
+                    let travelTimeSegment = Array(expectedSegmentTravelTimes[maneuverCoordinateIndex-stepIndex..<stepCoordinatesCount - stepIndex])
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/403#discussion_r129899911

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -131,6 +142,38 @@ open class RouteProgress: NSObject {
...
+                    
+                    let zippedCongestionTimes = Array(zip(congestionSegment, travelTimeSegment))
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/403#discussion_r129900254

bsudekum:
This is ready for review if @frederoni or @1ec5 want to check this out.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -118,6 +118,20 @@ open class RouteProgress: NSObject {
      */
     public var currentLegProgress: RouteLegProgress!
     
+    /**
+     Tupel containing a `CongestionLevel` and a corresponding `TimeInterval` representing the expected travel time for this segment.
```
1ec5: Typo: “tupel”.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -118,6 +118,20 @@ open class RouteProgress: NSObject {
...
+     */
+    public typealias TimedCongestionLevel = (CongestionLevel, TimeInterval)
+    
+    /**
+     If the route contains both `segmentCongestionLevels` and `expectedSegmentTravelTimes`, this array will contain an array of arrays of tuples of `segmentCongestionLevels` and `expectedSegmentTravelTimes`.
```
1ec5: > this array will contain an array of arrays of tuples of `segmentCongestionLevels` and `expectedSegmentTravelTimes`

This isn’t very helpful; Quick Help could’ve told the developer that. On the other hand, it would be helpful to phrase the comment like:

> …this property is set to a deeply nested array of `TimeCongestionLevel`s per segment per step per leg.

PR #404, opened 2017-07-21
Update osrm-text-instruction, MapboxDirections.swift

bsudekum:
Updates osrm-text-instructions.swift and MapboxDirections.swift

First need to merge https://github.com/Project-OSRM/osrm-text-instructions.swift/pull/28/


MapboxNavigation/RouteStepFormatter.swift
```diff
@@ -25,7 +25,7 @@ public class RouteStepFormatter: Formatter {
             return instructions.string(for: step)
         }
         
-        return instructions.string(for: step, legIndex: legIndex, numberOfLegs: numberOfLegs, modifyValueByKey: { (key, value) -> String in
+        return instructions.string(for: step, legIndex: legIndex, numberOfLegs: numberOfLegs, roadClasses: step.intersections?.first?.outletRoadClasses, modifyValueByKey: { (key, value) -> String in
```
bsudekum: Looking at the first intersection on the step should be enough to know what type it is.

Cartfile
```diff
@@ -2,5 +2,5 @@ binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 3.5
 github "mapbox/MapboxDirections.swift" ~> 0.10.0
 github "52inc/Pulley" ~> 1.3.1
 github "rs/SDWebImage" ~> 4.0
-github "Project-OSRM/osrm-text-instructions.swift" ~> 0.2.0
+github "Project-OSRM/osrm-text-instructions.swift" ~> 0.3
```
1ec5: We should pin any pre-1.0 dependency to a patch version (but still use the tadpole operator to allow for compatible upgrades). Fixed in 25af72698ef0323591aeb0cc99d42604f9ad67f5.

PR #405, opened 2017-07-21
Reroute using access token and host used to create request

bsudekum:
When rerouting, use the access token and host that were used to make the initial directions request.

/cc @1ec5

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -86,7 +86,7 @@ open class RouteController: NSObject {
     /**
      The Directions object used to create the route.
      */
-    public let directions: Directions
+    public var directions: Directions
     
     /**
      The route controller’s associated location manager.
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -360,6 +359,10 @@ extension RouteController: CLLocationManagerDelegate {
         
         self.lastRerouteLocation = location
         
+        if let accessToken = routeProgress.route.accessToken, let apiEndpoint = routeProgress.route.apiEndpoint, let host = apiEndpoint.host {
+            directions = Directions(accessToken: accessToken, host: host)
+        }
+        
         routeTask = directions.calculate(options, completionHandler: { [weak self] (waypoints, routes, error) in
```
1ec5: This is the only thing we were ever using the `directions` property for. I think we could get away with removing the `directions` property and the `directions` argument in the initializer, in favor of the new code above plus a fallback to `Directions.shared`. Do you think there would be any downside to that?

ericrwolfe: Doh, didn't see your comment before I merged. Yeah, this is the only spot we're currently using it. 

I still see an advantage to a developer being able to access the `directions` property on RouteController if they want to use it for their own routing/rerouting logic.

Issue #321, opened 2017-06-30
Migrate classes to core navigation


ericrwolfe:
Some of the following classes and logic would be useful to move over to MapboxCoreNavigation from the UI module, to give developers better building blocks for creating custom navigation apps from scratch:

- [x] Distance formatter
- [x] Location simulation
- [ ] Voice instruction formatting (should be extracted from voice controller) #367 
- [ ] Voice controller
- [ ] Location snapping #373 

cc @bsudekum @frederoni

1ec5:
We performed the most straightforward parts of this task in #309 and #310. Moving to v0.6.0 for the rest.

frederoni:
> Voice instruction formatting (should be extracted from voice controller)

Requires adding OSRMTextInstructions.swift as a dependency to core navigation.

ericrwolfe:
@frederoni seems like OSRMTextInstructions.swift is already a dependency of Core Navigation, at least in Cocoapods land: https://github.com/mapbox/mapbox-navigation-ios/blob/c113e518aa38274cf67d64d95c4cb638e732c0a7/MapboxCoreNavigation.podspec#L44

1ec5:
Heh, that was probably an oversight, but I guess it’s OK to have Core Navigation depend on OSRM Text Instructions. The more important consideration is making Polly an optional dependency somehow, because the AWS dependencies are enormous compared to Core Navigation proper.

bsudekum:
The checklist has been moved to core.

PR #408, opened 2017-07-21
Refactor location snapping

### LINKED ISSUE #373: Extract location snapping logic and move to core
Location snapping is a core building block for navigation and should be available in Core Navigation.

cc @bsudekum @frederoni

### LINKED ISSUE #406: snapsUserLocationAnnotationToRoute has no effect
`RouteController.snapsUserLocationAnnotationToRoute` purports to affect whether a map view’s user location annotation is snapped to the route. However, Core Navigation doesn’t actually do anything in response to this property; it only affects how RouteMapViewController behaves in the main navigation SDK.

In the Swift example application, CustomViewController sets `snapsUserLocationAnnotationToRoute` to `true` with the expectation that it snaps the user puck to the route, but it doesn’t. Instead, CustomViewController implements its own snapping in `progressDidChange(_:)`. So setting `snapsUserLocationAnnotationToRoute` is a bit misleading.

Additionally, the documentation claims that `snapsUserLocationAnnotationToRoute` is disabled by default, but it’s actually enabled by default.

/cc @bsudekum @ericrwolfe

### LINKED ISSUE #407: Current road name label ignores snapped course
`RouteMapViewController.navigationMapView(_:shouldUpdateTo:)` updates the current road name label after snapping the user’s location to the route line but before snapping the user’s course to the route line. This could cause the label to unexpectedly display the name of a cross street in the event that the course reading is inaccurate.

/cc @bsudekum @frederoni

1ec5:
This PR moves the user location snapping code from RouteMapViewController in MapboxNavigation to RouteController in CoreNavigation. Conversely, it moves the `snapsUserLocationAnnotationToRoute` property from RouteController in CoreNavigation to RouteMapViewController in MapboxNavigation, where it’s actually used.

RouteController now has public `location` and `snappedLocation` properties, while NavigationViewController has a `snapsUserLocationAnnotationToRoute` property.

Along the way, the current road name label now respects course snapping, and various NavigationMapViewDelegate methods have proper Objective-C names. (Previously, selector pieces ended in dangling prepositions instead of nouns.)

#402 partly refactors the location snapping code as well. I think this PR should land before that one to keep things simple.

To summarize:

* [x] Move user location snapping to Core Navigation (to fix #373)
* [x] Move `snapsUserLocationAnnotationToRoute` property to NavigationViewController (to fix #406)
* [x] Fix current road name label with respect to course snapping (to fix #407)
* [x] Keep labeling the current road even after the user goes off-course
* [x] Fix Objective-C names of NavigationMapViewDelegate methods

/ref #321
/cc @ericrwolfe @frederoni @bsudekum

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -230,15 +230,15 @@ open class RouteController: NSObject {
      
      This is a raw location received from `locationManager`. To obtain an idealized location, use the `snappedLocation` property.
      */
-    public var location: CLLocation?
+    var rawLocation: CLLocation?
     
     /**
      The most recently received user location, snapped to the route line.
      
-     This property contains a `CLLocation` object located along the route line near the `CLLocation` object in the `location` property. This property is set to `nil` if the route controller is unable to snap `location` to the route line for some reason.
+     This property contains a `CLLocation` object located along the route line near the most recently received user location. This property is set to `nil` if the route controller is unable to snap the user’s location to the route line for some reason.
      */
-    public var snappedLocation: CLLocation? {
-        guard let location = location, userIsOnRoute(location) else { return nil }
+    public var location: CLLocation? {
+        guard let location = rawLocation, userIsOnRoute(location) else { return nil }
         guard let stepCoordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return nil }
         guard let snappedCoordinate = closestCoordinate(on: stepCoordinates, to: location.coordinate) else { return location }
         
@@ -380,7 +380,7 @@ extension RouteController: CLLocationManagerDelegate {
         guard let location = locations.last else {
             return
         }
-        self.location = location
+        self.rawLocation = location
         
         delegate?.routeController?(self, didUpdateLocations: [location])
         
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -232,6 +232,67 @@ open class RouteController: NSObject {
     }
     
     /**
+     The most recently received user location.
+     
+     This is a raw location received from `locationManager`. If `snapsUserLocationAnnotationToRoute` is set to `true`, you can obtain an idealized location from the `snappedLocation` property.
+     */
+    public var location: CLLocation?
+    
+    /**
+     The most recently received user location, snapped to the route line.
+     
+     If `snapsUserLocationAnnotationToRoute` is set to `true`, this property contains a `CLLocation` object located along the route line near the `CLLocation` object in the `location` property. This property is set to `nil` if the route controller is unable to snap `location` to the route line for some reason.
+     */
+    public var snappedLocation: CLLocation? {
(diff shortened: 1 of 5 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -273,131 +280,100 @@ extension RouteMapViewController: NavigationMapViewDelegate {
...
-                    $0.isMapboxStreets
-            }
-            
-            // Add Mapbox Streets if the map does not already have it
-            if streetsSources.isEmpty {
-                let source = MGLVectorSource(identifier: "mapboxStreetsv7", configurationURL: URL(string: "mapbox://mapbox.mapbox-streets-v7")!)
-                style.addSource(source)
-                streetsSources.append(source)
-            }
+        guard let snappedLocation = routeController.snappedLocation else {
```
frederoni: Relying on two location managers, route controller being the main source for driving navigation and MGLMapView asking for `shouldUpdateTo` based on its own location updates feels like it's set up for a race condition. However, #402 should eventually move the data in the other direction and the way CLLocationManager is shared throughout the OS should prevent that but there's no guarantee that the snapped location won't lag 1 location update behind.

1ec5: Ah, so should this method be passing `location` into `snappedLocation` then?

frederoni: I'd say the other way around, the route controller should drive the map view, but that might be out of scope for this PR?

1ec5: Yes, I agree. We should have the route controller drive the map view, and #402 will accomplish that pretty cleanly. In the meantime, do you think what’s happening here is any riskier than what’s in master?

frederoni: It should be the same.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -165,6 +159,62 @@ open class RouteController: NSObject {
         locationManager.stopUpdatingLocation()
         locationManager.stopUpdatingHeading()
     }
+    
+    /**
+     The most recently received user location.
+     
+     This is a raw location received from `locationManager`. To obtain an idealized location, use the `snappedLocation` property.
+     */
+    public var location: CLLocation?
```
bsudekum: Do you think this should be public? If a developer needs the raw user location, shouldn't they just create their own location manager?

1ec5: snappedLocation has to be public so that MapboxNavigation can use it. I figured if we expose snappedLocation, we might as well expose the raw location too. But we could instead expose snappedLocation as location and pretend there isn't a raw location to begin with.

bsudekum: ok, that makes sense.

1ec5: Fixed in 29da08e99177d1f685db285638729d9005485052.

1ec5:
29da08e99177d1f685db285638729d9005485052 causes the current road name to continue to be labeled even if the user diverges from the route.

PR #410, opened 2017-07-21
Suppress freeway names in turn banner

1ec5:
If the upcoming road is a freeway, suppress its name in the turn banner in favor of a textual representation of the route number.

I-75 SR 126

This is a more tightly scoped alternative to #291 based on Project-OSRM/osrm-text-instructions.swift#28. It focuses on the turn banner and leaves the current road name label alone for the time being.

/cc @ericrwolfe @frederoni

### DIFF
--- a/MapboxNavigation/RoutePageViewController.swift
+++ b/MapboxNavigation/RoutePageViewController.swift
@@ -56,13 +56,13 @@ class RoutePageViewController: UIPageViewController {
 extension RoutePageViewController: UIPageViewControllerDataSource, UIPageViewControllerDelegate {
     func pageViewController(_ pageViewController: UIPageViewController, viewControllerAfter viewController: UIViewController) -> UIViewController? {
         let controller = viewController as! RouteManeuverViewController
-        let stepAfter = maneuverDelegate.stepAfter(controller.step)
+        let stepAfter = maneuverDelegate.stepAfter(controller.step!)
         return routeManeuverViewController(with: stepAfter)
     }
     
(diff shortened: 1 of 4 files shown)


MapboxNavigation/RoutePageViewController.swift
```diff
@@ -56,13 +56,13 @@ class RoutePageViewController: UIPageViewController {
 extension RoutePageViewController: UIPageViewControllerDataSource, UIPageViewControllerDelegate {
     func pageViewController(_ pageViewController: UIPageViewController, viewControllerAfter viewController: UIViewController) -> UIViewController? {
         let controller = viewController as! RouteManeuverViewController
-        let stepAfter = maneuverDelegate.stepAfter(controller.step)
+        let stepAfter = maneuverDelegate.stepAfter(controller.step!)
```
ericrwolfe: @1ec5 Is there a reason for changing the step property to an optional, but force unwrapping it here?

1ec5: I realized when running the tests that it is possible for `step` to be `nil` when the tests create a minimal RouteManeuverViewController. The tests were crashing originally when I failed to say `step?` in my changes. But we do continue to assume that another view controller that initializes RouteManeuverViewController would set the step. Open to ideas here.

1ec5:
f257dba97c37c0e452cc69190e5781f2295e9818 fixes an issue where neither #404 nor Project-OSRM/osrm-text-instructions.swift#26 had any effect on non-SSML-formatted instructions (instructions using AVFoundation or instructions in the step table).

PR #417, opened 2017-07-24
Prefer destination info over everything else

bsudekum:
We we're preferring name/ref ceremonial hack over destination information. We should show destination information since we'll already get the shield representing the highway number on the right.

Also, I removed the `numberOfDestinationLines` of math for destinations. This is because we want the destination info in the UI to match real world 1:1. Perhaps there should be some guard in place so this does not get out of hand. However, comparing this to waze, they put even more than just destination info. They also stuff in exit information when available.


bsudekum:
@1ec5 this is ready for review

MapboxNavigation/Resources/Base.lproj/Localizable.strings
```diff
@@ -19,6 +19,9 @@
 /* Delimiter between route numbers in a road concurrency */
 "REF_DELIMITER" = " / ";
 
+/* Delimiter between destinations in a road concurrency */
```
1ec5: A road concurrency is when a stretch of road is part of multiple routes. This delimiter is merely used when there are multiple destinations, so not necessarily a road concurrency.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -170,12 +170,12 @@ class RouteManeuverViewController: UIViewController {
     }
     
     func updateStreetNameForStep() {
-        if let step = step, step.isNumberedMotorway, let codes = step.codes {
+        if let destinations = step?.destinations {
+            destinationLabel.unabridgedText = destinations.joined(separator: NSLocalizedString("DESTINATION_DELIMITER", bundle: .mapboxNavigation, value: " / ", comment: "Delimiter between destinations in a road concurrency"))
```
1ec5: We should still find a way to ensure that a destination doesn’t get cut off in the middle, like “Stevens Creek Blvd / West San ~~Carlos Street~~”.

1ec5: This change could cause the label’s font size to shrink beyond what it normally used to. At some point, would it be more readable to allow a second line of destination text?

bsudekum: Here is how small it gets and how much we can stuff in there:

simulator screen shot jul 25 2017 11 18 19 am

I think this is okay.

bsudekum:
@1ec5 I think the truncation will be okay. We can adjust the look of many destinations when we refactor this view.

1ec5:
Mind adding a snapshot test or two so we get a better sense of how this looks in practice?

bsudekum:
@1ec5 updated and also enabled tests and updated old fixtures.

bsudekum:
@1ec5 alright, this is ready I believe

frederoni:
Nit: The snapshots are only tested on iPhone 6S Plus running iOS 10.3.1 so there's no need to add 2x reference images or for other versions of iOS.

PR #418, opened 2017-07-24
Actively remove parens from voice instructions

bsudekum:
Until we find a [fix upstream](https://github.com/mapbox/mapbox-navigation-ios/issues/286#issuecomment-315584617), we should omit parens completely from a voice instruction.

/cc @1ec5


MapboxNavigation/String.swift
```diff
@@ -25,7 +25,9 @@ extension String {
             ("&", "&amp;"),
             ("<", "&lt;"),
             ("\"", "&quot;"),
-            ("'", "&apos;")
+            ("'", "&apos;"),
+            ("(", ""),
+            (")", "")
```
1ec5: This turns `addingXMLEscapes` into a misnomer. This method may be used in places where we just want to mark up XML and don’t want to remove parentheses. Instead of repurposing this method, let’s modify the call sites to remove parentheses.

1ec5: Along the lines of osmlab/osmlint#234, are there any other characters that we should strip out?

bsudekum:
@1ec5 updated. `(, ) and _` we're the only offenders from https://github.com/osmlab/osmlint/issues/234 that are spoken.

MapboxNavigation/String.swift
```diff
@@ -25,9 +25,15 @@ extension String {
             ("&", "&amp;"),
             ("<", "&lt;"),
             ("\"", "&quot;"),
-            ("'", "&apos;"),
+            ("'", "&apos;")
+            ])
+    }
+    
+    var withoutPuncation: String {
```
1ec5: `removingPunctuation` would be more in keeping with existing String methods.

PR #419, opened 2017-07-24
Display last waypoint name when arriving

bsudekum:
Instead of the upcoming (non-existent) step.

/cc @1ec5


1ec5:
(Also, the previous code was actually trying to format the already formatted instruction and bailing because it was a String instead of a RouteStep.)

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -121,7 +121,7 @@ class RouteManeuverViewController: UIViewController {
         
         if routeProgress.currentLegProgress.alertUserLevel == .arrive {
             distance = nil
-            destinationLabel.unabridgedText = routeStepFormatter.string(for: routeStepFormatter.string(for: routeProgress.currentLegProgress.upComingStep))
+            destinationLabel.unabridgedText = routeProgress.remainingWaypoints.last?.name ?? routeStepFormatter.string(for: routeProgress.currentLegProgress.upComingStep)
```
1ec5: Let’s use `routeProgress.currentLeg.destination.name` instead; that makes things a little clearer and prepares this method for when #270 lands.

1ec5: We also need to repeat this line in `updateStreetNameForStep()`, somewhere before [this case](https://github.com/mapbox/mapbox-navigation-ios/blob/ff6f9e56ac213991e517d7eb3c13f3733bc8b558/MapboxNavigation/RouteManeuverViewController.swift#L175) (since the name is sometimes filled in by the Directions API). Otherwise, the name provided by the Directions API appears the entire time the user is on the destination’s street, and the name provided by the application appears only when nearing the destination itself.

bsudekum:
@1ec5 this is ready for another check

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -169,13 +169,15 @@ class RouteManeuverViewController: UIViewController {
         }
     }
     
-    func updateStreetNameForStep() {
+    func updateStreetNameForStep(currentLeg: RouteLeg? = nil) {
```
1ec5: There aren’t that many call sites, so remove the `= nil` to make it explicit when `nil` is passed in.

1ec5: On second thought, since `step` is a property that [gets set](https://github.com/mapbox/mapbox-navigation-ios/blob/ff6f9e56ac213991e517d7eb3c13f3733bc8b558/MapboxNavigation/RoutePageViewController.swift#L51) when the view controller is set up, maybe `leg` should be a property too? That would eliminate the need to pass it in here.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -169,13 +169,15 @@ class RouteManeuverViewController: UIViewController {
...
         if let step = step, step.isNumberedMotorway, let codes = step.codes {
             destinationLabel.unabridgedText = codes.joined(separator: NSLocalizedString("REF_DELIMITER", bundle: .mapboxNavigation, value: " / ", comment: "Delimiter between route numbers in a road concurrency"))
         } else if let name = step?.names?.first {
             destinationLabel.unabridgedText = name
         } else if let destinations = step?.destinations {
             destinationLabel.unabridgedText = destinations.prefix(min(numberOfDestinationLines, destinations.count)).joined(separator: "\n")
+        } else if let currentLeg = currentLeg, let step = step, step.maneuverType == .arrive {
```
1ec5: This case needs to appear above all the other cases. Otherwise, if the step has a `name` (usually just the street name), it’ll override whatever more specific name the developer assigned to the waypoint.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -20,7 +20,7 @@ class RouteManeuverViewController: UIViewController {
         didSet {
             if isViewLoaded {
                 roadCode = step?.codes?.first ?? step?.destinationCodes?.first ?? step?.destinations?.first
-                updateStreetNameForStep()
+                updateStreetNameForStep(currentLeg: nil)
```
1ec5: If I’m not mistaken, this would set up a race condition in which the label initially displays the Directions API–provided name (e.g., “Race Street”) until the first location update, at which point it would display the developer-provided name (“Starbucks” or “102 Race Street”). https://github.com/mapbox/mapbox-navigation-ios/pull/419#discussion_r129280021 would solve this issue.

bsudekum:
@1ec5 updated this, but I'm sort of not liking the instruction that is shown:

simulator screen shot jul 25 2017 4 52 03 pm

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -25,6 +25,14 @@ class RouteManeuverViewController: UIViewController {
         }
     }
     
+    var leg: RouteLeg? {
+        didSet {
+            if let _ = roadCode {
```
1ec5: If `leg` is set but there’s no `roadCode`, the street name doesn’t get updated. I’m not sure I understand what `roadCode` has to do with the leg.

1ec5:
> I'm sort of not liking the instruction that is shown

The waypoint name provided by the Directions API is clearly suboptimal, but that’s an upstream issue that’s out of this library’s control. What we can do is to ensure that, as long as the developer creates a RouteOptions with a named destination waypoint, that name appears when arriving at the destination.

bsudekum:
@1ec5 fixed https://github.com/mapbox/mapbox-navigation-ios/pull/419#discussion_r129460025. I guess I can live with the comma for now while we figure out an upstream fix.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -25,9 +25,9 @@ class RouteManeuverViewController: UIViewController {
         }
     }
     
-    var leg: RouteLeg? {
+    weak var leg: RouteLeg? {
```
1ec5: Why does this property need to be weak? If `leg` is set using a local variable, it’ll unexpectedly become `nil` right after the local variable goes out of scope. (Same for `step`, for that matter.)

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -170,7 +178,9 @@ class RouteManeuverViewController: UIViewController {
     }
     
     func updateStreetNameForStep() {
-        if let destinations = step?.destinations {
+        if let currentLeg = leg, let step = step, step.maneuverType == .arrive {
+            destinationLabel.unabridgedText = currentLeg.name
```
1ec5: The label should be set to the destination’s name, not the leg’s name. Is this why you were seeing two street names separated by a comma in https://github.com/mapbox/mapbox-navigation-ios/pull/419#issuecomment-317906512?

PR #420, opened 2017-07-25
Add audio type to telemetry events

ericrwolfe:
`speaker`, `bluetooth`, or `headphones`

cc @bsudekum

### DIFF
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -223,16 +223,19 @@ extension UIApplicationState {
 
 extension AVAudioSession {
     var audioType: String {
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -218,6 +221,36 @@ extension UIApplicationState {
...
 
+extension AVAudioSession {
+    var audioType: String {
+        if bluetoothAudioConnected() {
+            return "bluetooth"
+        }
+        if headphonesConnected() {
+            return "headphones"
+        }
+        return "speaker"
```
bsudekum: Is there a specific check we can do for `speaker` and fallback to `other` if nothing else is found?

PR #421, opened 2017-07-25
Resume button cover logo completely

bsudekum:
Simple fix for https://github.com/mapbox/mapbox-navigation-ios/issues/384
Now:
simulator screen shot jul 24 2017 5 41 33 pm

/cc @frederoni @1ec5

PR #424, opened 2017-07-25
Hide logo when resume is enabled

bsudekum:
Redoing https://github.com/mapbox/mapbox-navigation-ios/pull/421

When I toggled the image, it deemphasized the button. Since it's very important, I'm going to keep the location but as @1ec5 recommended, just hide the logo.

/cc @frederoni


1ec5:
> When I toggled the image, it deemphasized the button.

Can you elaborate? Why would changing the overview button’s image deemphasize (dim?) the button?

bsudekum:
@1ec5 the PR toggled the image in the overview button in the upper left corner. When the user changes course tracking mode from followWithCourse to none, they should have a big button to get them back to followWithCourse mode.

1ec5:
Ah, OK, sounds like you were referring to the prominence of the button rather than some sort of unintended dimming effect. 👍

PR #429, opened 2017-07-26
Update readme

frederoni:
API changed since we last updated the readme.

@ericrwolfe @bsudekum  👀

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,8 @@ options.includesSteps = true
 
 Directions.shared.calculate(options) { (waypoints, routes, error) in
     guard let route = routes?.first else { return }
-
+    
     let viewController = NavigationViewController(for: route)
-    viewController.simulatesLocationUpdates = true
     self.present(viewController, animated: true, completion: nil)
 }
 ```


Svantulden:
Should the example also import `MapboxCoreNavigation` as the example now uses `SimulatedLocationManager`?

README.md
```diff
@@ -74,9 +75,9 @@ options.includesSteps = true
 
 Directions.shared.calculate(options) { (waypoints, routes, error) in
     guard let route = routes?.first else { return }
-
-    let viewController = NavigationViewController(for: route)
-    viewController.simulatesLocationUpdates = true
+    
+    let locationManager = SimulatedLocationManager(route: route)
```
bsudekum: @frederoni what's your thinking behind adding `SimulatedLocationManager` here? I'm sort of thinking we want to make the example that gets the most eyes, the most barebones example we can come up with.

frederoni: I agree. Let's remove the simulation.

frederoni: fixed

PR #430, opened 2017-07-26
Consolidate serializing and deserializing CLLocation

frederoni:
Besides isolating serialization and deserialization of CLLocation to `CLLocation.swift`,
this PR also adds support for parsing various dictionary representations of CLLocations to be able to import from sources like feedback events, GuideDog recorded routes, and GPX generated traces.

- Latitude can be `latitude` or `lat`.
- Longitude can be `longitude`, `lng`, or `lon`.
- Timestamp can be represented as a unix timestamp or ISO8601 date string.

@ericrwolfe @bsudekum @1ec5 👀


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -0,0 +1,104 @@
...
+         "horizontalAccuracy": 5,
+         "course": 0.48
+         "timestamp": 1497475447,   // timestamp as unix timestamp or ISO8601Date
+         "altitude": 57.26
+     }]
+     
+     - parameter filePath: The file’s path.
+     - returns: A [CLLocation].
+     */
+    public static func locations(from filePath: String) -> [CLLocation]! {
```
bsudekum: I think if this is going to be public and used outside of this file, it should be moved to it's own `Array.swift` just so we can stay organized/predictable.

Issue #365, opened 2017-07-12
Left-aligned highway shields


ericrwolfe:
We should left align the highway shields in the turn banner making them appear inline with the rest of the instruction text. 

cc @frederoni @1ec5

1ec5:
At the same time, let’s avoid making the shield any smaller than it is now. The numbers inside certain shield images can already be difficult to discern at the current size.

1ec5:
Started work in the [1ec5-inline-shield-365](https://github.com/mapbox/mapbox-navigation-ios/tree/1ec5-inline-shield-365) branch.

1ec5:
We’re also eliding any cardinal direction from the turn banner whenever we choose to display a shield. As part of this feature, we should make sure to display any cardinal direction next to the shield.

michaelsteffen:
🎉 yes pls

bsudekum:
This has been implemented.

1ec5:
Specifically, the shields were inlined in #745.

Issue #334, opened 2017-07-05
Expand shield coverage beyond the United States


1ec5:
The turn banner and step table show graphical route shields for a number of route networks in the United States. Shields.plist maps `ref` prefixes to file name patterns on Wikimedia Commons. Commons has a much larger repertoire of shields, but we’ve run up against the limits of parsing the `ref` tag on ways, and even what we have now is prone to false positives. For example:

* [`CA`](https://github.com/mapbox/mapbox-navigation-ios/blob/1677c6479a2a293172a4a4d1e3ea45e5c1e3707b/MapboxNavigation/Resources/Shields.plist#L29-L30) may denote California or Cantabría, Spain
* [`NH`](https://github.com/mapbox/mapbox-navigation-ios/blob/1677c6479a2a293172a4a4d1e3ea45e5c1e3707b/MapboxNavigation/Resources/Shields.plist#L83-L84) may denote New Hampshire or National Highway (e.g., in India)
* [`CO`](https://github.com/mapbox/mapbox-navigation-ios/blob/1677c6479a2a293172a4a4d1e3ea45e5c1e3707b/MapboxNavigation/Resources/Shields.plist#L31-L32) may denote Colorado or a county road.
* We can’t add support for German shields to the turn banner, because we can’t tell whether to use a German, Austrian, Australian, or Chinese shield for [`A`](https://en.wikipedia.org/wiki/A_road).

Unfortunately, there isn’t a better way to distinguish shields internationally until either Project-OSRM/osrm-backend#482 or Project-OSRM/osrm-backend#333 is implemented. This issue affects routes in the U.S. as well: state trunkline highways in Michigan have the prefix `M`, which is identical to motorways in several other countries.

For more background on graphical route shields and why they matter in navigation, see gravitystorm/openstreetmap-carto#508 and [this OSM diary entry](http://www.openstreetmap.org/user/Minh%20Nguyen/diary/39123), or watch [this talk from State of the Map 2011](https://wiki.openstreetmap.org/wiki/SotM_2011_session:_Shields_Up!).

/ref https://github.com/mapbox/mapbox-navigation-ios/issues/286#issuecomment-309649781
/cc @ericrwolfe @danpat @freenerd

1ec5:
[...]

PR #431, opened 2017-07-26
Avoid matching bannered routes to shields

1ec5:
A destination code such as `I 75 South` should match the Interstate shield, but a code such as `I 75 Alt` or `I 75 Truck` should not. Bannered routes are usually less important, and it’s misleading to show a bannered route shield without the banner.

before after

For #365, we’ll have an opportunity to display the banner as text to the right of the shield.

/cc @frederoni

### DIFF
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -49,7 +49,7 @@ class RouteManeuverViewController: UIViewController {
                 return
             }
             
-            if components.count == 2 || (components.count == 3 && ["North", "South", "East", "West"].contains(components[2])) {
+            if components.count == 2 || (components.count == 3 && ["North", "South", "East", "West", "Nord", "Sud", "Est", "Ouest"].contains(components[2])) {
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -49,7 +49,7 @@ class RouteManeuverViewController: UIViewController {
                 return
             }
             
-            if components.count > 1 {
+            if components.count == 2 || (components.count == 3 && ["North", "South", "East", "West", "Nord", "Sud", "Est", "Ouest"].contains(components[2])) {
```
1ec5: This array contains the cardinal directions in English (for the United States and most of Canada) and French (for Québec). These appear to be the only regions that put cardinal directions in the [`destination:ref`](https://taginfo.openstreetmap.org/keys/destination:ref) tag in OpenStreetMap, and besides, Shields.plist only covers the U.S. at the moment (#334). I refrained from making these strings localizable, because we need to continue to match cardinal directions in road codes regardless of whether the user is using the country’s official language.

Issue #371, opened 2017-07-12
Enable battery monitoring


ericrwolfe:
Currently, we never enable battery monitoring, so features such as [adjusting location accuracy if the device is plugged in](https://github.com/mapbox/mapbox-navigation-ios/blob/be67b60eb0b2b1615bf0a3b1fafc9adaf93fd7ad/MapboxNavigation/DefaultLocationManager.swift#L18) don't actually work as expected.

There were unexpected crashes when battery monitoring was enabled in RouteController.init and multiple RouteController/NavigationViewControllers were initialized in close succession, so we'll need to find a workaround.

cc @bsudekum @frederoni

PR #432, opened 2017-07-26
Enable battery monitoring

bsudekum:
The fatal we were seeing before was most likely fixed with https://github.com/mapbox/mapbox-navigation-ios/pull/399

Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/371

/cc @1ec5 @ericrwolfe

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -164,6 +164,7 @@ open class RouteController: NSObject {
         self.routeProgress = RouteProgress(route: route)
         self.locationManager = locationManager
         self.locationManager.activityType = route.routeOptions.activityType
+        UIDevice.current.isBatteryMonitoringEnabled = true
         super.init()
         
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -198,6 +198,7 @@ open class RouteController: NSObject {
             events.initialize(withAccessToken: mapboxAccessToken, userAgentBase: "MapboxEventsNavigationiOS", hostSDKVersion: String(describing: Bundle(for: RouteController.self).object(forInfoDictionaryKey: "CFBundleShortVersionString")!))
             events.disableLocationMetrics()
             events.sendTurnstileEvent()
+            UIDevice.current.isBatteryMonitoringEnabled = true
```
frederoni: Should we also stop monitoring? Perhaps in init and deinit of `RouteController` or `NavigationLocationManager`?

PR #435, opened 2017-07-27
Fix tracking mode

bsudekum:
This fixes:

- Initial camera view when entering navigation mode
- Shows the route earlier
- Puck float while rotating the screen

untitled

However, toggling `automaticallyAdjustsScrollViewInsets = false` causes the camera to zoom from 0,0 to the current location. This is even though I'm not animating any camera/tracking mode changes. Also turning it off prevents the drawer change in position from changing the inset location for items on the map.

I need to play with this more, but it's getting close.

/cc @frederoni @1ec5

### DIFF
--- a/MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
+++ b/MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
@@ -106,7 +106,6 @@
                                     <constraint firstItem="drb-do-FoT" firstAttribute="centerX" secondItem="nNr-30-cGD" secondAttribute="centerX" id="TCg-jZ-7G3"/>
                                 </constraints>
                                 <userDefinedRuntimeAttributes>
-                                    <userDefinedRuntimeAttribute type="boolean" keyPath="showsUserLocation" value="YES"/>
                                     <userDefinedRuntimeAttribute type="string" keyPath="styleURL__" value="mapbox://styles/mapbox/navigation-guidance-day-v2"/>
                                 </userDefinedRuntimeAttributes>
                             </view>
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -63,13 +63,12 @@ class RouteMapViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        automaticallyAdjustsScrollViewInsets = false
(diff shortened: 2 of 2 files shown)


MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -106,7 +106,6 @@
                                     <constraint firstItem="drb-do-FoT" firstAttribute="centerX" secondItem="nNr-30-cGD" secondAttribute="centerX" id="TCg-jZ-7G3"/>
                                 </constraints>
                                 <userDefinedRuntimeAttributes>
-                                    <userDefinedRuntimeAttribute type="boolean" keyPath="showsUserLocation" value="YES"/>
```
bsudekum: Setting this to true prevents the developer from setting map location settings in viewDidLoad.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -79,34 +78,34 @@ class RouteMapViewController: UIViewController {
...
+                camera.centerCoordinate = userLocation.coordinate
+                if let course = userLocation.location?.course {
+                    camera.heading = course
+                }
             }
+            camera.pitch = 60
+            camera.altitude = 600
             mapView.setCamera(camera, animated: false)
         }
+        mapView.manuallyUpdatesLocation = true
```
bsudekum: Order matters here. This needs to be called after `mapView.setUserTrackingMode(.followWithCourse, animated: false)`

bsudekum:
@frederoni what if we removed `automaticallyAdjustsScrollViewInsets = false` and just adjusted the height of the map?

1ec5:
Is mapbox/mapbox-gl-native#9637 related to this PR somehow?

frederoni:
@bsudekum RouteMapViewController do not have top or bottom layout guides so `automaticallyAdjustsScrollViewInsets` should not have any effect. We set the `contentInsets` manually depending on the state of the drawer.

bsudekum:
> Is mapbox/mapbox-gl-native#9637 related to this PR somehow?

Yes, very much so.

ericrwolfe:
As a workaround, in the NavigationMapView subclass, could we we force a call to `didUpdateLocations` with the last known location manager location after the view gets set up in `viewDidAppear` or in `layoutSubviews`?

PR #441, opened 2017-07-28
Filter inaccurate locations

### LINKED ISSUE #433: Filter inaccurate GPS
GPS locations are not guaranteed to be accurate. We should filter out any locations with poor accuracy (>= 100m horizontal, <= 0m horizontal, >= 30m vertical, <= 0m vertical).

cc @mapbox/navigation-ios

frederoni:
Fixes #433 - Filter inaccurate locations

@ericrwolfe @bsudekum  👀

### DIFF
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -3,6 +3,12 @@ import CoreLocation
 
 extension CLLocation {
     
+    var isQualified: Bool {
+        return
+            0...100 ~= horizontalAccuracy &&
+            0...30 ~= verticalAccuracy
+    }
+    
     /// Returns a dictionary representation of the location.
     public var dictionaryRepresentation: [String: Any] {
         var locationDictionary: [String: Any] = [:]
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -390,9 +390,15 @@ extension RouteController: CLLocationManagerDelegate {
         }
         self.rawLocation = location
         
-        delegate?.routeController?(self, didUpdateLocations: [location])
+        if location.isQualified {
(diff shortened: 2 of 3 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -390,9 +390,15 @@ extension RouteController: CLLocationManagerDelegate {
...
         self.rawLocation = location
         
-        delegate?.routeController?(self, didUpdateLocations: [location])
+        if location.isQualified {
+            delegate?.routeController?(self, didUpdateLocations: [location])
+        }
         
         sessionState.pastLocations.push(location)
+        
+        guard location.isQualified else {
```
bsudekum: Could this guard be moved above addition above?

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -3,6 +3,12 @@ import CoreLocation
 
 extension CLLocation {
     
+    var isQualified: Bool {
+        return
+            0...100 ~= horizontalAccuracy &&
```
bsudekum: If the wifi is off and the user is in an urban environment, will this ever be below 100m? Should we start higher and filter down slowly?

MapboxCoreNavigation/RouteController.swift
frederoni: We still want to collect the location, even though it's not qualified. However, we could move down L:393

MapboxCoreNavigation/CLLocation.swift
frederoni: The two modes that we use, `kCLLocationAccuracyBest` and `kCLLocationAccuracyBestForNavigation` are not using wifi, perhaps for a cold start but wifi doesn't support that level of accuracy. However, we should still run the numbers.

PR #443, opened 2017-07-31
Consolidate location managers

frederoni:
This PR breaks out a small part of #402 - reinvent the course tracking.
More specifically it combines `DefaultLocationManager` and `NavigationLocationManager` so that battery monitoring resides in core and we can reuse it when using `Simulated/ReplayLocationManager`.

@ericrwolfe @bsudekum 👀


MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -10,6 +14,11 @@ open class NavigationLocationManager: CLLocationManager {
     
     var lastKnownLocation: CLLocation?
     
+    /**
+     Indicates whether the device is plugged in or not.
+     */
+    public var isPluggedIn: Bool = false
```
bsudekum: Should this public and mutable? If it needs to be public can we make this a getter somehow?

frederoni: fixed

MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -27,5 +36,23 @@ open class NavigationLocationManager: CLLocationManager {
                 allowsBackgroundLocationUpdates = true
             }
         }
+        
+        #if os(iOS)
+            UIDevice.current.addObserver(self, forKeyPath: "batteryState", options: [.initial, .new], context: nil)
```
bsudekum: Should this observer be removed on deinit?

frederoni: Yep, that's already the case 😄

Issue #434, opened 2017-07-27
Only show lane indicators when a turn is approaching


ericrwolfe:
Right now we show the lane indicators for any step length, which can be confusing if a step is miles long and the users get in the wrong lanes prematurely.

We should wait until the alert level changes to medium/high to show the lane indications.

cc @bsudekum @danesfeder

bsudekum:
I think a safe first move would to hide when `.low`, else show.

PR #444, opened 2017-08-01
Only show lanes for medium/high alerts

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/434

This changes only to show the lane UI when the user is approaching the intersection. This seems like a good/safe first move.

/cc @ericrwolfe @frederoni @1ec5

### DIFF
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -132,7 +132,9 @@ class RouteManeuverViewController: UIViewController {
             destinationLabel.unabridgedText = routeProgress.currentLeg.destination.name ?? routeStepFormatter.string(for: routeStepFormatter.string(for: routeProgress.currentLegProgress.upComingStep, legIndex: routeProgress.legIndex, numberOfLegs: routeProgress.route.legs.count, markUpWithSSML: false))
         } else if let upComingStep = routeProgress.currentLegProgress?.upComingStep {
             updateStreetNameForStep()
-            showLaneView(step: upComingStep)
+            if routeProgress.currentLegProgress.alertUserLevel == .medium || routeProgress.currentLegProgress.alertUserLevel == .high {
+                showLaneView(step: upComingStep)
+            }
         }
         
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -132,7 +132,9 @@ class RouteManeuverViewController: UIViewController {
             destinationLabel.unabridgedText = routeProgress.currentLeg.destination.name ?? routeStepFormatter.string(for: routeStepFormatter.string(for: routeProgress.currentLegProgress.upComingStep, legIndex: routeProgress.legIndex, numberOfLegs: routeProgress.route.legs.count, markUpWithSSML: false))
         } else if let upComingStep = routeProgress.currentLegProgress?.upComingStep {
             updateStreetNameForStep()
-            showLaneView(step: upComingStep)
+            if routeProgress.currentLegProgress.alertUserLevel == .medium || routeProgress.currentLegProgress.alertUserLevel == .high {
```
frederoni: We also call `showLaneView(step:)` in `RoutePageViewControllerDelegate.routePageViewController(_:willTransitionTo:)`. Wouldn't this mean that if you swipe to the current step, you'll see the lane view, but when you get an update from `RouteManeuverViewController.notifyDidChange(routeProgress:secondsRemaining:)`, the lane view disappears (for the same step at a low alert level)?

Can we move the show and hide logic to `showLaneView(step:)` and perhaps take routeProgress or alertLevel as a parameter?

bsudekum: I sort of think this is a good flow. If you're scrolling ahead and looking at future maneuvers, I think you should be able to see all available information.

frederoni: I agree about the scrolling ahead.

However, let's assume we're going from one step to another (w/o swiping) which currently is at low alert level.
The page changes, `(willTransitionTo:)` gets called which in turn shows the lane view unconditionally. I think this is the case and also exactly what this PR is trying to avoid?

PR #445, opened 2017-08-01
Do not filter locations when using simulator

frederoni:
GPX-based locations reports a vertical and horizontal accuracy of -1.

@bsudekum 👀

### DIFF
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -4,9 +4,13 @@ import CoreLocation
 extension CLLocation {
     
     var isQualified: Bool {
-        return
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -4,9 +4,13 @@ import CoreLocation
 extension CLLocation {
     
     var isQualified: Bool {
-        return
-            0...100 ~= horizontalAccuracy &&
-            0...30 ~= verticalAccuracy
+        #if (arch(i386) || arch(x86_64)) && os(iOS)
```
bsudekum: An edge case, but should we include tvos and watchos here too?

PR #446, opened 2017-08-02
Add mute button and persistent settings

### LINKED ISSUE #363: Volume/mute button
Add a volume button that allows users to manually enable/disable voice announcements.

cc @frederoni

frederoni:
Fixes #363 - Mute button

I'm using NSUserDefaults for storing the mute state between sessions, however, UserDefaults are meant for Application targets or specifically for App Groups when using suite name. I wonder if we should look into another persistent store?

@bsudekum @ericrwolfe 👀


bsudekum:
Nice! What do you think about putting all actions down the left side? 

/cc @mayagao

frederoni:
@bsudekum There's not enough space in compact/landscape mode. Can we revisit when we have a separate design for landscape?

bsudekum:
@frederoni ah good call, as is is fine.

Issue #427, opened 2017-07-26
Rationalize freeway alert level thresholds


1ec5:
This is a proposal to overhaul the logic that determines the current alert level to make voice announcements more straightforward and predictable for off-ramp steps, potentially as a first step to overhauling the rest of the alert level logic.

Essentially, the `medium` and `high` alert levels would occur at specific distances ahead of the maneuver, with time thresholds still in place as fallbacks. Hopefully this approach will make the SDK feel more predictable to the user, so that they can better intuit the level of urgency of a particular voice announcement.

## Current behavior

`RouteController.monitorStepProgress(_:)` tracks the expected time until the user reaches the maneuver point and the distance to the maneuver point. Whichever threshold is reached first determines when the alert level changes to medium then high. The time threshold gives us the flexibility to handle diverse speeds and traffic conditions, while the distance threshold serves as a failsafe for the time threshold.

Unfortunately, we arrived at the current thresholds by trial and error, making them difficult to explain in a coherent manner. We’ve had to put lots of band-aids on top of this logic just to make it work. Some of these thresholds actually assume that the user is traveling in a parking lot, even if they’re on a freeway:

* We change to the `high` alert level 15 seconds or 100 meters ahead of a maneuver, regardless of the type of road. However, at a typical Canadian expressway speed of 100 km/h, a driver would cover 417 meters in 15 seconds or 100 meters in 3.6 seconds. In order for the distance threshold to kick in, the driver would need to be traveling no more than 24 km/h (15 mph).
* Similarly, we change to the `medium` alert level 70 seconds or 400 meters ahead of a maneuver. At 100 km/h, a driver would cover 1.9 km in 70 seconds or 400 meters in 14.4 seconds. The distance threshold assumes a speed of 21 km/h (13 mph).

Alert level | Time remaining | Distance remaining | Effective distance remaining<br>(65&nbsp;mph / 100&nbsp;km/h) | Speeds triggering distance threshold
----|----|----|----|----
`low` | — | — | — | — | —
`medium` | 70 s | 400 m | 1¼ mi<br>1.9 km | <13 mph<br><21 km/h
`high` | 15 s | 100 m | ¼ mi<br>417 m | <15 mph<br><24 km/h

## Proposed rationale

[...]

PR #82, opened 2017-03-21
Add custom minimum alert distances for Cycling/Walking/Driving

davidyou3:
- Replace RouteControllerMinimumDistanceForMediumAlert and RouteControllerMinimumDistanceForHighAlert with
RouteControllerMinimumDistanceForHighAlertDriving, RouteControllerMinimumDistanceForHighAlertCycling, RouteControllerMinimumDistanceForHighAlertWalking
RouteControllerMinimumDistanceForMediumAlertDriving, RouteControllerMinimumDistanceForMediumAlertCycling, RouteControllerMinimumDistanceForMediumAlertWalking
to allow for different minimum distance per transportation method.

- Add RouteControllerMinimumDistanceForMediumAlert(identifier:) and RouteControllerMinimumDistanceForHighAlert(identifier:)
to choose which minimum distance is approriate for which MBDirectionsProfileIdentifier type.

- Modifiy MapboxNavigation/RouteController and MapboxNavigationUI/RouteVoiceController to use
new minimum distance values.

@bsudekum @1ec5 @boundsj

PR #448, opened 2017-08-02
Use calculated speed versus user speed [WIP]

bsudekum:
Seconds remaining on a step is currently calculated using the users instantaneous speed. Seconds remaining on step is then used to figure out when to give an alert. If below x give y alert.

This PR moves away from using the users speed and instead calculates seconds remaining by looking at their percent complete and then subtracting this from the step duration. The idea here is that the server can influence a little more about when alerts are give. This also prevents the case where a user speeds up suddenly triggering a premature alert. 

This PR also adds a new `calculateNewAlert(from upcomginStep:)` function which helps in figuring out the alert level on the new step. This will also be useful when improving linked instructions when we move voice logic over to core.

/cc @1ec5 @frederoni @willwhite @ericrwolfe @cammace

### DIFF
--- a/MapboxCoreNavigation/Constants.swift
+++ b/MapboxCoreNavigation/Constants.swift
@@ -94,7 +94,7 @@ public var RouteControllerMediumAlertInterval: TimeInterval = 70
 /**
  Number of seconds left on step when a `high` alert is emitted.
  */
-public var RouteControllerHighAlertInterval: TimeInterval = 15
+public var RouteControllerHighAlertInterval: TimeInterval = 20
 
 
 /**
@@ -110,81 +110,15 @@ public var MaxSecondsSpentTravelingAwayFromStartOfRoute: TimeInterval = 3
 
 
 /**
- Distance in meters for the minimum length of a step for giving a `medium` alert while using `MBDirectionsProfileIdentifierAutomobile` or `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`.
+ Remaing distance on a motorway at which the `.high` `AlertLevel` will be given. This overrides `RouteControllerHighAlertInterval` only when the current step is a motorway.
  */
-public var RouteControllerMinimumDistanceForMediumAlertDriving: CLLocationDistance = 400
+public var RouteControllerMotorwayHighAlertDistance: CLLocationDistance = 804.672
 
 
 /**
- Distance in meters for the minimum length of a step for giving a `medium` alert while using `MBDirectionsProfileIdentifierCycling`.
+ Remaing distance on a motorway at which the `.medium` `AlertLevel` will be given. This overrides `RouteControllerMediumAlertInterval` only when the current step is a motorway.
  */
-public var RouteControllerMinimumDistanceForMediumAlertCycling: CLLocationDistance = 200
-
-
-/**
- Distance in meters for the minimum length of a step for giving a `medium` alert while using `MBDirectionsProfileIdentifierWalking`.
- */
-public var RouteControllerMinimumDistanceForMediumAlertWalking: CLLocationDistance = 100
-
-
-/**
- Returns the appropriate `mediium` `AlertLevel` distance for a given `MBDirectionsProfileIdentifier`.
- */
-public func RouteControllerMinimumDistanceForMediumAlert(identifier: MBDirectionsProfileIdentifier) -> CLLocationDistance {
-    switch identifier {
-    case MBDirectionsProfileIdentifier.automobileAvoidingTraffic:
-        return RouteControllerMinimumDistanceForMediumAlertDriving
-    case MBDirectionsProfileIdentifier.automobile:
-        return RouteControllerMinimumDistanceForMediumAlertDriving
-    case MBDirectionsProfileIdentifier.cycling:
-        return RouteControllerMinimumDistanceForMediumAlertCycling
-    case MBDirectionsProfileIdentifier.walking:
-        return RouteControllerMinimumDistanceForMediumAlertWalking
-    default:
-        break
-    }
-    
-    return RouteControllerMinimumDistanceForMediumAlertDriving
-}
-
-
-/**
- Distance in meters for the minimum length of a step for giving a `high` alert while using `MBDirectionsProfileIdentifierAutomobile` or `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`.
- */
-public var RouteControllerMinimumDistanceForHighAlertDriving: CLLocationDistance = 100
-
-
-/**
- Distance in meters for the minimum length of a step for giving a `high` alert while using `MBDirectionsProfileIdentifierCycling`.
- */
-public var RouteControllerMinimumDistanceForHighAlertCycling: CLLocationDistance = 60
-
-
-/**
- Distance in meters for the minimum length of a step for giving a `high` alert while using `MBDirectionsProfileIdentifierWalking`.
- */
-public var RouteControllerMinimumDistanceForHighAlertWalking: CLLocationDistance = 20
-
-
-/**
- Returns a the appropriate `high` `AlertLevel` distance for a given `MBDirectionsProfileIdentifier`.
- */
-public func RouteControllerMinimumDistanceForHighAlert(identifier: MBDirectionsProfileIdentifier) -> CLLocationDistance {
-    switch identifier {
-    case MBDirectionsProfileIdentifier.automobileAvoidingTraffic:
-        return RouteControllerMinimumDistanceForHighAlertDriving
-    case MBDirectionsProfileIdentifier.automobile:
-        return RouteControllerMinimumDistanceForHighAlertDriving
-    case MBDirectionsProfileIdentifier.cycling:
-        return RouteControllerMinimumDistanceForHighAlertCycling
-    case MBDirectionsProfileIdentifier.walking:
-        return RouteControllerMinimumDistanceForHighAlertWalking
-    default:
-        break
-    }
-    
-    return RouteControllerMinimumDistanceForHighAlertDriving
-}
+public var RouteControllerMotorwayMediumAlertDistance: CLLocationDistance = 3218.69
 
 
 /**
--- a/MapboxCoreNavigation/Constants.swift
+++ b/MapboxCoreNavigation/Constants.swift
@@ -94,7 +94,7 @@ public var RouteControllerMediumAlertInterval: TimeInterval = 70
 /**
  Number of seconds left on step when a `high` alert is emitted.
  */
-public var RouteControllerHighAlertInterval: TimeInterval = 20
+public var RouteControllerHighAlertInterval: TimeInterval = 15
 
 
 /**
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -729,6 +727,18 @@ extension RouteController: CLLocationManagerDelegate {
         
         incrementRouteProgress(alertLevel, location: location, updateStepIndex: updateStepIndex)
     }
+    
+    func calculateNewAlert(from upcomginStep: RouteStep) -> AlertLevel {
+        let expectedTravelTime = upcomginStep.expectedTravelTime
+        switch expectedTravelTime {
+        case let x where x <= RouteControllerHighAlertInterval:
+            return .high
```
bsudekum: This should always trigger a linked instruction

bsudekum:
Added a check for motorways in https://github.com/mapbox/mapbox-navigation-ios/pull/448/commits/e23b33275f113aa59c7748d945054897d1b9e94c:
* At 2 miles, give medium alert if on highway
* At .25 miles, give high alert if on highway

MapboxCoreNavigation/RouteController.swift
```diff
@@ -685,6 +684,12 @@ extension RouteController: CLLocationManagerDelegate {
             if secondsToEndOfStep <= RouteControllerHighAlertInterval {
                 alertLevel = .high
             }
+        } else if let _ = outletRoadClasses?.contains(.motorway),
```
bsudekum: It'd be nice if we could filter out on/off ramps. These should have the normal alert level progression. Perhaps we can look if the current step has a name/dest/ref since I believe trunks do not.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -685,6 +684,12 @@ extension RouteController: CLLocationManagerDelegate {
...
+            routeProgress.currentLegProgress.currentStepProgress.distanceRemaining <= 402.336 {
+            alertLevel = .high
+        } else if let _ = outletRoadClasses?.contains(.motorway),
+            routeProgress.currentLegProgress.currentStepProgress.distanceRemaining <= 3218.69 {
```
bsudekum: Will pull this out to a constance.

bsudekum:
Noting in https://github.com/mapbox/mapbox-navigation-ios/pull/448/commits/a7c296d077391458c5c8920f85d409630bb1b97c I removed some old code that takes a look at a minimum distance heuristic. I believe we can only compared durations now that it's fixed to the step duration.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -719,23 +721,31 @@ extension RouteController: CLLocationManagerDelegate {
                 
                 // Look at the following step to determine what the new alert level should be
                 if let upComingStep = routeProgress.currentLegProgress.upComingStep {
-                    alertLevel = upComingStep.expectedTravelTime <= RouteControllerMediumAlertInterval ? .medium : .low
+                    alertLevel = calculateNewAlert(from: upComingStep)
                 } else {
                     assert(false, "In this case, there should always be an upcoming step")
                 }
             }
-        } else if secondsToEndOfStep <= RouteControllerHighAlertInterval && routeProgress.currentLegProgress.currentStep.distance > minimumDistanceForHighAlert {
```
bsudekum: @1ec5 @frederoni @ericrwolfe I'm removing the `minimumDistanceForMediumAlert` heuristics to keep this logic cleaner and focused only on seconds remaining. Also see [constants diff](https://github.com/mapbox/mapbox-navigation-ios/pull/448/files#diff-1e0aeb18d31c8c1abfce278ab3489151L121).

MapboxCoreNavigation/Constants.swift
```diff
@@ -110,81 +110,15 @@ public var MaxSecondsSpentTravelingAwayFromStartOfRoute: TimeInterval = 3
 
 
 /**
- Distance in meters for the minimum length of a step for giving a `medium` alert while using `MBDirectionsProfileIdentifierAutomobile` or `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`.
+ Remaing distance on a motorway at which the `.high` `AlertLevel` will be given. This overrides `RouteControllerHighAlertInterval` only when the current step is a motorway. Default value is a half mile.
  */
-public var RouteControllerMinimumDistanceForMediumAlertDriving: CLLocationDistance = 400
+public var RouteControllerMotorwayHighAlertDistance: CLLocationDistance = 804.672
```
1ec5: It isn’t apparent from reading `804.672` that this variable is set to the right default value. To make it apparent, set it to `0.5 * milesToMeters` instead. (×2)

MapboxCoreNavigation/RouteController.swift
```diff
@@ -692,6 +688,12 @@ extension RouteController: CLLocationManagerDelegate {
             if secondsToEndOfStep <= RouteControllerHighAlertInterval {
```
1ec5: Can you provide a table similar to the ones in #427, so we can get a sense of the speeds that trigger the distance thresholds versus the speeds that trigger the time thresholds?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -653,14 +651,12 @@ extension RouteController: CLLocationManagerDelegate {
         // Force an announcement when the user begins a route
         var alertLevel: AlertLevel = routeProgress.currentLegProgress.alertUserLevel == .none ? .depart : routeProgress.currentLegProgress.alertUserLevel
         var updateStepIndex = false
-        let profileIdentifier = routeProgress.route.routeOptions.profileIdentifier
         
         let userSnapToStepDistanceFromManeuver = distance(along: routeProgress.currentLegProgress.currentStep.coordinates!, from: location.coordinate)
-        let secondsToEndOfStep = userSnapToStepDistanceFromManeuver / location.speed
+        let secondsToEndOfStep = routeProgress.currentLegProgress.currentStepProgress.durationRemaining
```
1ec5: `durationRemaining` assumes the user travels an even speed across the entire step. Would we achieve better or worse results by having `durationRemaining` look at the remaining segments’ expected travel times?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -719,23 +721,31 @@ extension RouteController: CLLocationManagerDelegate {
...
-            // There needs to be an alert
-            routeProgress.currentLegProgress.currentStep.distance > minimumDistanceForMediumAlert {
+        } else if secondsToEndOfStep <= RouteControllerMediumAlertInterval {
             alertLevel = .medium
         }
         
         incrementRouteProgress(alertLevel, location: location, updateStepIndex: updateStepIndex)
     }
+    
+    func calculateNewAlert(from upcomginStep: RouteStep) -> AlertLevel {
```
1ec5: This method returns an alert level and has no side effects, so its name should begin with a noun, not a verb.

1ec5: Typo: `upcomingStep`.

MapboxCoreNavigation/Constants.swift
```diff
@@ -110,81 +110,15 @@ public var MaxSecondsSpentTravelingAwayFromStartOfRoute: TimeInterval = 3
 
 
 /**
- Distance in meters for the minimum length of a step for giving a `medium` alert while using `MBDirectionsProfileIdentifierAutomobile` or `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`.
+ Remaing distance on a motorway at which the `.high` `AlertLevel` will be given. This overrides `RouteControllerHighAlertInterval` only when the current step is a motorway. Default value is a half mile.
```
1ec5: `AlertLevel.high` will be linked automatically by jazzy. (×2)

MapboxCoreNavigation/RouteController.swift
1ec5: As far as I can tell, the plans are to have the Directions API provide a distance ahead of the maneuver, not a duration. It seems like deemphasizing the time-based heuristics in favor of the distance thresholds would be more future-proof in that case.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -719,23 +721,31 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let expectedTravelTime = upcomginStep.expectedTravelTime
+        switch expectedTravelTime {
+        case let x where x <= RouteControllerHighAlertInterval:
```
1ec5: This is a pretty convoluted way of writing an if/else if/else statement. To fully take advantage of switch/case, use ranges like this:

```swift
switch expectedTravelTime {
case 0..<RouteControllerHighAlertInterval: …
case RouteControllerHighAlertInterval..<RouteControllerMediumAlertInterval: …
default: …
}
```

MapboxCoreNavigation/Constants.swift
```diff
@@ -110,81 +110,15 @@ public var MaxSecondsSpentTravelingAwayFromStartOfRoute: TimeInterval = 3
...
  */
-public var RouteControllerMinimumDistanceForMediumAlertDriving: CLLocationDistance = 400
+public var RouteControllerMotorwayHighAlertDistance: CLLocationDistance = 804.672
 
 
 /**
- Distance in meters for the minimum length of a step for giving a `medium` alert while using `MBDirectionsProfileIdentifierCycling`.
+ Remaing distance on a motorway at which the `.medium` `AlertLevel` will be given. This overrides `RouteControllerMediumAlertInterval` only when the current step is a motorway. Defauly value is 2 miles.
  */
-public var RouteControllerMinimumDistanceForMediumAlertCycling: CLLocationDistance = 200
```
1ec5: This PR effectively reverts #82. Have we concluded that the driving profile’s thresholds are suitable for cycling and walking too?

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -215,16 +214,19 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
         // This should be caught above.
         let upComingInstruction = routeStepFormatter.string(for: routeProgress.currentLegProgress.upComingStep, legIndex: upcomingLegIndex, numberOfLegs: numberOfLegs, markUpWithSSML: markUpWithSSML)!
-        let stepDistance = routeProgress.currentLegProgress.upComingStep!.distance
+        let upcomingStepDuration = routeProgress.currentLegProgress.upComingStep!.expectedTravelTime
         let currentInstruction = routeStepFormatter.string(for: routeProgress.currentLegProgress.currentStep, legIndex: routeProgress.legIndex, numberOfLegs: numberOfLegs, markUpWithSSML: markUpWithSSML)
         let step = routeProgress.currentLegProgress.currentStep
         var text: String
         
+        // Prevent back to back instructions by adding a little more wiggle room
+        let linkedInstructionMultiplier = RouteControllerHighAlertInterval * 1.2
```
1ec5: This variable is of type `TimeInterval`, so it’s an “interval”, not a “multiplier” (which implies no units).

Also factor out 1.2 (which is the real multiplier) into a constant. Provide a documentation comment that explains why 1.2 is the right value for this constant.

MapboxCoreNavigation/RouteController.swift
bsudekum: I'm not sure about this. Distance is way less scalable then duration which accounts better for expected speed, road type and congestion.

/cc @willwhite @ericrwolfe

MapboxCoreNavigation/Constants.swift
bsudekum: If we continue to use step duration, this value should scale with other profiles these profiles account for walking/biking times.

MapboxCoreNavigation/RouteController.swift
bsudekum: 🤔  curious idea. This would require the developer always pass in this annotation option. I guess we could fallback though. Let me check this out. Could be yuge.

bsudekum: Actually, punting on this for now.

PR #452, opened 2017-08-02
Fix buttons dip down

### LINKED ISSUE #451: After rerouting, buttons dip down
simulator screen shot aug 2 2017 2 08 10 pm

/cc @frederoni

frederoni:
Fixes #451 

@bsudekum 👀

### DIFF
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -215,6 +215,7 @@ class RouteManeuverViewController: UIViewController {
                 laneView.isValid = usableLanes.contains(i as Int)
                 laneView.setNeedsDisplay()
             }
+            stackViewContainer.isHidden = false
         } else {
             stackViewContainer.isHidden = true
         }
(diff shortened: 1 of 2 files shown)


frederoni:
Oh, this is probably related to https://github.com/mapbox/mapbox-navigation-ios/commit/60a02ad94bd02f4089429f942ad79537ab8eb935

You're browsing while the alert level is low. It should be ignored if it's not the current step.

bsudekum:
Hrm it should be showing the lanes for the swiped-to-step and using the alert level passed through here https://github.com/mapbox/mapbox-navigation-ios/commit/60a02ad94bd02f4089429f942ad79537ab8eb935#diff-292af689ce88d5b5aa06966456abc0e9R478 which is always .high

frederoni:
It's possible for a step to be the first/current and last at the same time which resulted in:



these incorrect left and right overlays are fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/452/commits/534d9f4ad8fb03d332fd4693eb316333d2574d1e

bsudekum:
Looks good, only thing I'm seeing now is a slight delay when after incrementing the step and showing the lane view:

untitled

PR #453, opened 2017-08-03
Move voice to core [WIP]

bsudekum:
This moves voice and instructions creation to core. This will allow developers to who use core to create instructions to be used in their own speech synthesizer or use the one built into ios.


bsudekum:
https://github.com/mapbox/mapbox-navigation-ios/pull/456

PR #454, opened 2017-08-03
Add script to extract localizable strings

frederoni:
Added a script that extracts strings to make it a bit more convenient to keep the strings on Transifex up-to-date.

@bsudekum @1ec5 👀


frederoni:
Actually, we only need to generate the base language, right? Individual languages are pulled down when the translated rate exceeds `minimum_perc`.

bsudekum:
Deferring this PR to @1ec5

frederoni:
Superseded by #468

PR #455, opened 2017-08-03
Defer presenting UIAlertController

bsudekum:
We've seen slow speeds when entering navigation mode. Turns out, a lot of time is spent adding actions to the UIAlertController which is not thread safe. This PR defers presenting the controller to the next tick. Feels 💯  and much snappier.

/cc @frederoni

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -101,7 +101,9 @@ class ViewController: UIViewController, MGLMapViewDelegate {
             self.startMultipleWaypoints()
         }))
         alertController.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: nil))
-        present(alertController, animated: true, completion: nil)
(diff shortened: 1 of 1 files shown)


Examples/Swift/ViewController.swift
```diff
@@ -101,7 +101,9 @@ class ViewController: UIViewController, MGLMapViewDelegate {
             self.startMultipleWaypoints()
         }))
         alertController.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: nil))
-        present(alertController, animated: true, completion: nil)
+        DispatchQueue.main.async {
```
frederoni: The delay we're trying to address here, is that the delay when tapping on "Start Navigation" or "Default UI"?

bsudekum: The entire screen is locked up actually when choosing between the different views. Even when selecting cancel, there is a long wait.

PR #456, opened 2017-08-03
Move instruction creation to core

bsudekum:
_Taking a 3rd stab at https://github.com/mapbox/mapbox-navigation-ios/pull/453_

This moves voice and instructions creation to core. This will allow developers to who use core to create instructions to be used in their own speech synthesizer or use the one built into ios.

todo:

- [x] move speech strings over to core
- [x] double check what must be made public
- [x] document new public methods


MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -28,4 +28,8 @@ extension Bundle {
             return object(forInfoDictionaryKey: "NSLocationAlwaysUsageDescription") as? String
         }
     }
+    
+    class var mapboxCoreNavigation: Bundle {
+        get { return Bundle(for: self) }
```
bsudekum: Unsure what to make this value.

MapboxCoreNavigation/SpokenInstructionFormatter.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
```
bsudekum: Can this file be merged with any others?

MapboxCoreNavigation/BundleAdditions.swift
frederoni: `mapboxCoreNavigation` makes sense. `Bundle(for: self)` should work even though it takes a class but we're inside an extension. How about the root of core navigation, which is RouteController? `Bundle(for: RouteController.self)`

MapboxCoreNavigation/SpokenInstructionFormatter.swift
frederoni: I'd probably assume that this class extends NSFormatter but it doesn't need any of its functionality. I'd keep it as it is.

bsudekum:
I also added UI instruction creation to this PR.

PR #457, opened 2017-08-04
Various design fixes

frederoni:
- [x] Allow subclassing `Style`
- [x] Add way name background to the style
- [x] Align top banner content
- [x] Replace feedback icon
- [x] Extend way name styling
- [x] Stack floating buttons (overview, mute, feedback) to the left

and various tweaks in the default style

@bsudekum @ericrwolfe @pveugen  👀


MapboxNavigation/RouteTableViewController.swift
```diff
@@ -43,7 +43,7 @@ class RouteTableViewController: UIViewController {
         }
         
         if routeProgress.durationRemaining < 60 {
-            headerView.timeRemaining.text = String.localizedStringWithFormat(NSLocalizedString("LESS_THAN", bundle: .mapboxNavigation, value: "<%@", comment: "Format string for less than; 1 = duration remaining"), dateComponentsFormatter.string(from: 61)!)
+            headerView.timeRemaining.text = String.localizedStringWithFormat(NSLocalizedString("LESS_THAN", bundle: .mapboxNavigation, value: "%@", comment: "Format string for less than; 1 = duration remaining"), dateComponentsFormatter.string(from: 61)!)
```
1ec5: Why was the less-than sign removed? The label was intended to say for instance “<1 s`. If this removal was intentional, then there’s no point to a format string that consists of nothing but a single placeholder.

1ec5: #527 undoes this change.

PR #462, opened 2017-08-07
Prevent Polly from making UI laggy

bsudekum:
`try Data(contentsOf: url as URL)` was really blocking up UI rendering. Removing this and moving to an async method for fetching data allows the UI to continue rendering while the data is fetched and unpacked. The UI during speaking is noticeably smoother.

/cc @willwhite @ericrwolfe @frederoni

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -24,6 +24,8 @@ public class PollyVoiceController: RouteVoiceController {
      */
     public var identityPoolId: String
     
+    var pollyTask: URLSessionDataTask?
+    
     public init(identityPoolId: String) {
         self.identityPoolId = identityPoolId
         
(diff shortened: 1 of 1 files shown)


MapboxNavigation/PollyVoiceController.swift
```diff
@@ -111,18 +117,38 @@ public class PollyVoiceController: RouteVoiceController {
             return
         }
         
-        do {
-            let soundData = try Data(contentsOf: url as URL)
```
1ec5: Specifically, it was this line that was bogging down the UI.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -111,18 +117,38 @@ public class PollyVoiceController: RouteVoiceController {
...
+        pollyTask = URLSession.shared.dataTask(with: url as URL) { [weak self] (data, response, error) in
+            guard let strongSelf = self else { return }
+            
+            // If the task is canceled, don't speak.
+            // But if it's some sort of other error, use fallback voice.
+            if let error = error as NSError?, error.code == NSURLErrorCancelled {
+                return
+            } else if let error = error {
+                // Cannot call super in a closure
+                strongSelf.callSuperSpeak(strongSelf.fallbackText, error: error.localizedDescription)
```
1ec5: Do we need to call super’s `speak(_:error:)` specifically? Why not call `self.speak(_:error:)`?

bsudekum: @1ec5 that would result in the polly voice controller trying to fetch a new polly mp3 if it were to fail. The idea here is that if it fails, it falls back to the default speech synth.

PR #463, opened 2017-08-08
Various UI tweaks

frederoni:


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -45,8 +45,8 @@ class RouteMapViewController: UIViewController {
     var tiltedCamera: MGLMapCamera {
         get {
             let camera = mapView.camera
-            camera.altitude = 600
-            camera.pitch = 45
+            camera.altitude = 150
+            camera.pitch = 70
```
bsudekum: I think 60 might be the max.

bsudekum:
@frederoni mind posting a screenshot?

MapboxNavigation/RouteMapViewController.swift
ericrwolfe: Also, what are the implications of this on battery life @pveugen?

pveugen: Didn't do any pre- and post testing, but anecdotally this does have some impact.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -42,6 +42,8 @@ class RouteTableViewController: UIViewController {
             headerView.distanceRemaining.text = distanceFormatter.string(from: routeProgress.distanceRemaining)
         }
         
+        dateComponentsFormatter.unitsStyle = routeProgress.durationRemaining < 3600 ? .short : .abbreviated
```
1ec5: I think `dateComponentsFormatter` rounds up when the duration is close to an hour. So there might be an edge case where it still uses `short` even though you want `abbreviated`.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -20,8 +20,8 @@ class RouteTableViewController: UIViewController {
         setupTableView()
         dateFormatter.timeStyle = .short
         dateComponentsFormatter.maximumUnitCount = 2
```
1ec5: This is no longer necessary because `day` has been removed below.

frederoni:
Superseded by https://github.com/mapbox/mapbox-navigation-ios/pull/480

PR #466, opened 2017-08-08
Prefetch polly speech strings

bsudekum:
_This is a branched off of https://github.com/mapbox/mapbox-navigation-ios/pull/456 and bigly a work in progress._

This moves to prefetch all Polly mp3s when the PollyVoiceController is inited. 

todo:

- [ ] Calculate distance remaining using the alert level constants
- [ ] Only prefetch when there is down time. At the beginning of a session, there can be many instructions back to back. We should prioritize these since prefetching all instructions is resource intensive
- [ ] profile
- [ ] use a better queue

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -25,9 +25,17 @@ public class PollyVoiceController: RouteVoiceController {
      */
     public var identityPoolId: String
     
+    let alerts: [AlertLevel: TimeInterval] = [
+        .depart: 100,
+        .low: 100,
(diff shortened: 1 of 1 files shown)


ericrwolfe:
Noting that prefetching all files isn't just resource intensive, but $$$ expensive in the case of frequent reroutes or long routes with many maneuvers.

bsudekum:
This is more actionable now that https://github.com/mapbox/mapbox-navigation-ios/pull/456#pullrequestreview-56719558 is merged.

bsudekum:
Going to close this and revisit it later.

PR #464, opened 2017-08-08
UI Tweaks

frederoni:
- [x] Dock progress bar to the bottom
- [x] Rearrange lane views to avoid glitchy swipe in and out animation
- [x] Remove reset tracking mode timer

@bsudekum @ericrwolfe @pveugen 👀

PR #468, opened 2017-08-08
Add status view

frederoni:
The rerouting UI was removed in #464 
This PR adds a new reusable status view that also acts as a rerouting UI.

@bsudekum @ericrwolfe 👀 

_The rerouting ui is presented manually in this gif_

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -18,6 +18,7 @@ class RouteMapViewController: UIViewController {
     @IBOutlet weak var wayNameLabel: WayNameLabel!
     @IBOutlet weak var wayNameView: UIView!
     @IBOutlet weak var maneuverContainerView: ManeuverContainerView!
+    @IBOutlet weak var statusView: StatusView!
     @IBOutlet weak var laneViewsTopConstraint: NSLayoutConstraint!
     @IBOutlet weak var laneViewsContainerView: UIView!
     @IBOutlet var laneViews: [LaneArrowView]!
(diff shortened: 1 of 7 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -204,6 +220,11 @@ class RouteMapViewController: UIViewController {
             wayNameView.isHidden = true
         }
     }
+    
+    func willReroute(notification: NSNotification) {
+        let title = NSLocalizedString("REROUTING", bundle: .mapboxNavigation, value: "Rerouting…", comment: "Indicates that rerouting is in progress")
```
bsudekum: Note, we'll need to update the translation files.

MapboxNavigation/Style.swift
```diff
@@ -512,3 +514,51 @@ class ManeuverContainerView: UIView {
...
+    @IBOutlet weak var activityIndicatorView: UIActivityIndicatorView!
+    @IBOutlet weak var textLabel: UILabel!
+    @IBOutlet weak var topConstraint: NSLayoutConstraint!
+    
+    func show(_ title: String, showSpinner: Bool, duration: TimeInterval) {
+        textLabel.text = title
+        activityIndicatorView.isHidden = !showSpinner
+        activityIndicatorView.startAnimating()
+        
+        show()
```
bsudekum: Maybe call these `show()` function something a little different? hard to know what they each do.

MapboxNavigation/RouteMapViewController.swift
frederoni: Included the `extract_localizable.sh` script. Seems to work great.

MapboxNavigation/Style.swift
frederoni: They're fileprivate so shouldn't affect too much, but if you have a better name 😊 
How about `updateConstraints(showOrHide: Bool)`? 🤔

frederoni: Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/468/commits/b6a51bb351e8d7c5f3848acfe27f4122456934cd

PR #474, opened 2017-08-09
Remove reroute status bar after route is fetched

bsudekum:
We should hide the status bar once the new route is fetched.

/cc @frederoni


bsudekum:
@frederoni What if we hid it when the route request errored as well? I agree that it can stick around for a little longer since it's so brief, maybe we can add another 0.5-1s timeout?

Issue #381, opened 2017-07-13
Discard improbable headings


ericrwolfe:
When course snapping is enabled, we should discard any headings exactly equal to 0 or -1. 

![](https://user-images.githubusercontent.com/1594300/27805910-423a2556-5fec-11e7-819b-42841f885a10.jpg)

cc @bsudekum

ericrwolfe:
Likewise, we might want to prevent any course snapping changes when the user is stopped at a light and the speed is close to zero.

![](https://user-images.githubusercontent.com/1594300/27805905-3d1f09a6-5fec-11e7-8431-f30140d287c8.jpg)

bsudekum:
So in cases where we have a course of -1, 0 or a speed of 0, we should just snap the user to the route line?

Currently, we returning early and using the course from the device https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/RouteMapViewController.swift#L362-L364

@1ec5 this make sense to you?

1ec5:
It’s probably possible for a CLLocation’s course to be −1 at the same time its coordinate is valid. So at this point, we’d create a new CLLocation with all the same information but a course based on the route line, then continue to use that synthesized CLLocation throughout the method.

Only negative values are invalid; 0° is no more improbable than 35°. If we were to discard 0° in favor of a course based on the route line, we’d introduce inconsistent behavior. For example, if the route continues due east but the user turns left, we’d snap the course up to −30°, then accurately reflect `CLLocation.course` after that, until the user gets to 0°. At that point, we swing back to due east before swinging northward again as the user crosses over to 359°.

ericrwolfe:
> 0° is no more improbable than 35°

Noting here that we're talking about the exact value of 0.0000000...°, which is highly improbable with real world, unrounded, double precision coordinates. A simple `course > 0` instead of `course != -1` would suffice.

1ec5:
My point is that negative values are the only ones documented to be invalid. Guarding against `course < 0` is 👌.

bsudekum:
Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/445

ericrwolfe:
I don't think #445 actually fixes this issue.

ericrwolfe:
This was directly addressed in https://github.com/mapbox/mapbox-navigation-ios/pull/449.

[...]

PR #475, opened 2017-08-09
Look at points behind user for snapping

bsudekum:
While investigating using the instantaneous calculated user course for determining off routing, I discovered that our calculated user course at a given point can be improved. Prior to this, we were looking ahead at two points on a line. Instead of looking at two points at a given distance ahead of the user, we should look at the angles at points ahead and behind the user. This will give a smoother  and more accurate representation of the route snapped course.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -256,18 +256,19 @@ open class RouteController: NSObject {
         }
         
         let nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
-        guard let closest = closestCoordinate(on: nearByCoordinates, to: location.coordinate) else { return location }
-        let slicedLine = polyline(along: nearByCoordinates, from: closest.coordinate, to: nearByCoordinates.last)
-        let userDistanceBuffer = location.speed * RouteControllerDeadReckoningTimeInterval
-
-        // Get closest point infront of user
-        guard let pointOneSliced = coordinate(at: userDistanceBuffer, fromStartOf: slicedLine) else { return location }
-        guard let pointOneClosest = closestCoordinate(on: nearByCoordinates, to: pointOneSliced) else { return location }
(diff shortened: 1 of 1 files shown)


frederoni:
> This will give a smoother and more accurate representation of the route snapped course.

To me, this sounds like something we want to achieve with the camera, not the user puck.
Smoothing out the camera position based the average course of upcoming steps.

The user puck should always (when snapped) be aligned with the route line since it's hard to drive sideways along a route.

Also, I think we should remove the `guard location.speed >= 0` to avoid floating around on the route line at a stop sign.

bsudekum:
> To me, this sounds like something we want to achieve with the camera, not the user puck.
Smoothing out the camera position based the average course of upcoming steps.

I think there is merit to this, but until https://github.com/mapbox/mapbox-navigation-ios/pull/402 is merged, I think we need to stick with this.

> The user puck should always (when snapped) be aligned with the route line since it's hard to drive sideways along a route.

The problem with this approach is that there is no in between/grace period between being snapped and not being snapped. The truth is that between the location snapping Apple does internally, GPS noise and the difference between OSM data and real world roads, we need some breathing room. At one point we did have either the user was snapped or they were rerouted and it caused tons of rerouting that became annoying to users.

> Also, I think we should remove the guard location.speed >= 0 to avoid floating around on the route line at a stop sign.

I believe @ericrwolfe wanted this in there to prevent inaccurate course values from slipping through. Perhaps it would help though if we still mutated the user course if it looks good at low speeds.

ericrwolfe:
>> Also, I think we should remove the guard location.speed >= 0 to avoid floating around on the route line at a stop sign.

> I believe @ericrwolfe wanted this in there to prevent inaccurate course values from slipping through. Perhaps it would help though if we still mutated the user course if it looks good at low speeds.

I agree with @frederoni's point. See my comment in https://github.com/mapbox/mapbox-navigation-ios/issues/381#issuecomment-321121509.

bsudekum:
Let's pull that out into a separate PR.

PR #476, opened 2017-08-10
Allow snaping when speed is zero

bsudekum:
This does two things:
* Allows snapping to occur (if deemed possible) when the user speed is zero. This will help at stop lights
* At the beginning of the route, it's difficult to snap the user course and location if we cannot look behind the user. Instead, use the location and let the user begin navigating on the route. This will only happen momentarily and is probably a good thing to allow the user some buffer time when beginning a route.

/cc @frederoni @ericrwolfe @1ec5


bsudekum:
Closing in favor of https://github.com/mapbox/mapbox-navigation-ios/pull/481

Issue #473, opened 2017-08-09
Example-Swift app crashes when running in iPad Pro simulator when tapping Start Navigation button


troyibm:
I get this error when I run Example-Swift in Xcode 8.3.2 on the iPad Pro (9.7 inch) simulator in landscape mode.

`
Terminating app due to uncaught exception 'NSGenericException', reason: 'Your application has presented a UIAlertController (<UIAlertController: 0x7f926fa4fef0>) of style UIAlertControllerStyleActionSheet. The modalPresentationStyle of a UIAlertController with this style is UIModalPresentationPopover. You must provide location information for this popover through the alert controller's popoverPresentationController. You must provide either a sourceView and sourceRect or a barButtonItem.  If this information is not known when you present the alert controller, you may provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation.'
`

troyibm:
worked around this by changing `.actionSheet` to `.alert` in line https://github.com/mapbox/mapbox-navigation-ios/blob/master/Examples/Swift/ViewController.swift#L90

egmixer:
Thanks for sharing your solution

valterh4ck3r:
Thank you so much @troyibm , you saved me :)

SamAlmo:
You are amazing. Thanks for saving my time

PR #477, opened 2017-08-10
Handle iPads in example app

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/473

`UIAlertController` on the iPad needs to know where to anchor the action sheet.

/cc @ericrwolfe


bsudekum:
This also speeds up the action sheet by adding actions in the background. Ref: https://stackoverflow.com/questions/42417808/very-slow-presentation-of-the-uialertcontroller

PR #481, opened 2017-08-11
Use user heading sparingly when speed is low [Controversial!]

bsudekum:
This fixes a few things:

* During snapping, we were using the users speed to find two points near the user to interpret the users expected course for their given location on the route. This is huge trap, if the users speed is zero or near zero, the two points could be on top of each other. Our snapping math would produce a wildly inaccurate because of this. In the worst case, the new course would be zero, which is very bad.
* We've had a continuing reports of the course floating around when the user is at a stoplight. This can be due to the course being inaccurate while not moving. If the course is <= 0 or if the speed is <= 1, we should not use the users course to snap their course to the route. Instead, we can use the heading to see if they should be snapped.
  * We have also had issues with the users direction being off at the beginning of a route. This change will also help their since a lot of the time, a user will not start driving until they know which direction they are facing. In this case, heading is more useful to use since their course is most likely wrong (stand still).

This should be tested heavily. 

### DIFF
--- a/MapboxCoreNavigation/CLHeading.swift
+++ b/MapboxCoreNavigation/CLHeading.swift
@@ -1,0 +1,11 @@
+import Foundation
+import CoreLocation
+
+extension CLHeading {
+    var preferredHeading: CLLocationDirection {
+        guard trueHeading <= 0 || headingAccuracy > 45 else {
+            return magneticHeading
+        }
+        return trueHeading
+    }
+}
--- a/MapboxCoreNavigation/CLHeading.swift
+++ b/MapboxCoreNavigation/CLHeading.swift
@@ -3,7 +3,7 @@ import CoreLocation
 
 extension CLHeading {
     var preferredHeading: CLLocationDirection {
-        guard trueHeading <= 0 || headingAccuracy > 45 else {
+        guard trueHeading >= 0 || headingAccuracy > 45 else {
             return magneticHeading
         }
         return trueHeading
--- a/MapboxCoreNavigation/CLHeading.swift
+++ b/MapboxCoreNavigation/CLHeading.swift
@@ -3,7 +3,7 @@ import CoreLocation
 
 extension CLHeading {
     var preferredHeading: CLLocationDirection {
-        guard trueHeading >= 0 || headingAccuracy > 45 else {
+        guard trueHeading >= 0 || headingAccuracy < 45 else {
             return magneticHeading
         }
         return trueHeading
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -255,6 +255,12 @@ open class RouteController: NSObject {
      */
     var rawLocation: CLLocation?
     
+    var heading: CLHeading?
+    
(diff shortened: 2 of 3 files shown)


MapboxCoreNavigation/CLHeading.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+import CoreLocation
+
+extension CLHeading {
+    var preferredHeading: CLLocationDirection {
+        return magneticHeading
```
bsudekum: This should be improved. What should we be checking for here?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -251,51 +257,52 @@ open class RouteController: NSObject {
...
-            return location
-        }
         
         let nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
         guard let closest = closestCoordinate(on: nearByCoordinates, to: location.coordinate) else { return nil }
         
         let slicedLineBehind = polyline(along: nearByCoordinates.reversed(), from: closest.coordinate, to: nearByCoordinates.reversed().last)
         let slicedLineInfront = polyline(along: nearByCoordinates, from: closest.coordinate, to: nearByCoordinates.last)
-        let userDistanceBuffer: CLLocationDistance = location.speed * RouteControllerDeadReckoningTimeInterval / 2
+        let userDistanceBuffer: CLLocationDistance = max(location.speed * RouteControllerDeadReckoningTimeInterval / 2, 8)
```
bsudekum: This is what caused the strange course going to zero sometimes.

MapboxCoreNavigation/CLHeading.swift
ericrwolfe: Ideally `trueHeading` over `magneticHeading`, espeically if the `headingAccuracy` is high

bsudekum:
@1ec5 @frederoni @ericrwolfe How are people feeling about this?  This has worked pretty well in user testing over the past week or so.

bsudekum:
This seems to work well at the beginning of the route and when the user is actively going off the route. However at low speeds while on the route, there still can be some rotation issues. Going to keep playing with this.

MapboxCoreNavigation/CLHeading.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+import CoreLocation
+
+extension CLHeading {
+    var preferredHeading: CLLocationDirection {
+        guard trueHeading <= 0 || headingAccuracy > 45 else {
+            return magneticHeading
+        }
+        return trueHeading
```
1ec5: I’m a bit confused by the intent of this computed variable: we return `trueHeading` if it’s negative or `headingAccuracy` is really high (really inaccurate)?

bsudekum: Oh, that is a mistake.

bsudekum:
I'm not convinced this fixes anything. I'm still seeing issues when stopped in traffic. Going to close this out.

ericrwolfe:
Superseded by https://github.com/mapbox/mapbox-navigation-ios/pull/540.

PR #488, opened 2017-08-14
Fix storyboard warnings

frederoni:
Fixed the remaining storyboard warnings such as missing trailing constraints and overlapping issues for localized components.

The resume button should look and behave exactly as before but it doesn't suffer from UIButton's mutually exclusive title and image properties.

@bsudekum 👀


bsudekum:
Looks great, however I'm getting a crash:

image

frederoni:
@bsudekum Good catch, fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/488/commits/3d94f8cb94addfdb71badb9118fb0204ebbd6b46

PR #490, opened 2017-08-14
Inform about weak gps signal

frederoni:
Added a new delegate that informs the user about a discarded (non-qualified) location and a corresponding UI. Also simplified the animations by removing all manual constraint animations.

- [x] Add documentation
- [x] Fix 

@bsudekum 👀


MapboxCoreNavigation/RouteController.swift
```diff
@@ -401,6 +404,7 @@ extension RouteController: CLLocationManagerDelegate {
         sessionState.pastLocations.push(location)
         
         guard location.isQualified else {
+            delegate?.routeController?(self, didDiscard: location)
```
bsudekum: Oh I like this design!

bsudekum:
Looks good. My only qualm visually is the spinner- it makes me think we're waiting on something. What do you think about hiding it for this action?

frederoni:
It could continue spinning for as long as we discard locations and then we hide the status view when we get the first qualified location?

bsudekum:
@frederoni I'm suggesting hiding the spinner completely while the gps signal is poor.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -33,6 +33,9 @@ public protocol RouteControllerDelegate: class {
     @objc(routeController:willRerouteFromLocation:)
     optional func routeController(_ routeController: RouteController, willRerouteFrom location: CLLocation)
     
+    @objc(routeController:didDiscardLocation:)
+    optional func routeController(_ routeController: RouteController, didDiscard location: CLLocation)
```
bsudekum: We'll need some documentation for this new delegate method.

PR #497, opened 2017-08-15
Add preferredOptions

bsudekum:
@ericdeveloper and I were talking about an easier way for developers to get started. RouteOptions is particularly a tricky configuration to get right since it requires inside knowledge of the directions API to get it right.

This PR moves the options out into an extension so the developer does not need to worry about these options. 


Examples/Swift/ViewController.swift
```diff
@@ -270,12 +266,8 @@ extension ViewController: WaypointConfirmationViewControllerDelegate {
...
             nextDestination,
         ])
-        options.includesSteps = true
-        options.routeShapeResolution = .full
-        options.profileIdentifier = navigationViewController.route.routeOptions.profileIdentifier
-        
-        // Adding the optional attribute `.congestionLevel` ensures the route line will show the congestion along the route line
-        options.attributeOptions = [.congestionLevel]
+            // Dont forget to apply `preferredOptions`. This adds the necessary options needed to make a navigation experience 💯.
+            .preferredOptions
```
ericrwolfe: Should this instead be a custom initializer? Maybe something like `RouteOptions(forNavigationWithCoordinates:)`? Feels weird to have `.preferredOptions` modify the existing options. Would be confusing if it overrode something the developer wanted to set explicitly.

frederoni: I agree about the overriding. However, [coordinates] won't bridge to Obj-C, how about a class function like `preferredOptions(waypoints:profileIdentifier:)`?

bsudekum:
Updated to use init instead.

MapboxCoreNavigation/RouteOptions.swift
```diff
@@ -9,4 +9,31 @@ extension RouteOptions {
...
         }
     }
+    
+    /**
+     Returns an optimal `RouteOptions` for navigation.
+     */
+    public convenience init(forNavigationWithWaypoints waypoints: [Waypoint], profileIdentifier: MBDirectionsProfileIdentifier? = .automobileAvoidingTraffic) {
+        self.init(waypoints: waypoints, profileIdentifier: profileIdentifier)
+        
+        includesSteps = true
```
bsudekum: @frederoni is there a way we can store these options once in this file?

1ec5: Why not subclass RouteOptions as NavigationOptions? Then NavigationOptions could store its own properties. This approach is intended to be supported by MapboxDirections.swift, and it has the benefit of being more discoverable and more easily documentable.

PR #498, opened 2017-08-15
Expose mapView(:ViewForAnnotation:)

bsudekum:
This exposes the ability to provide a view for any annotation on the map. In the example, I show how to provide a view for the user puck.

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -32,6 +32,10 @@ class ViewController: UIViewController, MGLMapViewDelegate {
     var exampleMode: ExampleMode?
     var nextWaypoint: CLLocationCoordinate2D?
     
+    // In this example, we show you how you can create custom UIView that is used to show the user's location.
+    // Set `showCustomUserPuck` to true to view the custom user puck.
+    var showCustomUserPuck = false
+    
     override func viewDidLoad() {
         super.viewDidLoad()
         
@@ -241,9 +245,11 @@ class ViewController: UIViewController, MGLMapViewDelegate {
 
         present(navigationViewController, animated: true, completion: nil)
     }
-    
+}
+
+extension ViewController: NavigationViewControllerDelegate {
     func navigationMapView(_ mapView: MGLMapView, viewFor annotation: MGLAnnotation) -> MGLAnnotationView? {
-        guard annotation is MGLUserLocation else { return nil }
+        guard annotation is MGLUserLocation && showCustomUserPuck else { return nil }
         
         let reuseIdentifier = "userPuck"
         var annotationView = mapView.dequeueReusableAnnotationView(withIdentifier: reuseIdentifier)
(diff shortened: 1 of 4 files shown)


Examples/Swift/ViewController.swift
```diff
@@ -242,6 +242,34 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         present(navigationViewController, animated: true, completion: nil)
     }
     
+    func navigationMapView(_ mapView: MGLMapView, viewFor annotation: MGLAnnotation) -> MGLAnnotationView? {
+        guard annotation is MGLUserLocation else { return nil }
```
bsudekum: For some reason, I'm never getting an annotation with type `MGLUserLocation `.

Examples/Swift/ViewController.swift
```diff
@@ -242,6 +242,34 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         present(navigationViewController, animated: true, completion: nil)
     }
     
+    func navigationMapView(_ mapView: MGLMapView, viewFor annotation: MGLAnnotation) -> MGLAnnotationView? {
```
1ec5: This method should be implemented in the extension below that conforms to NavigationViewControllerDelegate.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -101,6 +101,8 @@ public protocol NavigationViewControllerDelegate {
      If this method is unimplemented, the navigation map view will represent the destination annotation with the default marker.
      */
     @objc optional func navigationMapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?
+    
+    @objc optional func navigationMapView(_ mapView: MGLMapView, viewFor annotation: MGLAnnotation) -> MGLAnnotationView?
```
1ec5: This method needs documentation.

Examples/Swift/ViewController.swift
1ec5: I wonder if this has to do with some of the user location overriding that the navigation SDK does…

1ec5:
This PR would fix #213.

Examples/Swift/ViewController.swift
1ec5: The reason is that `MGLMapView.showsUserLocation` is set during storyboard initialization, [causing MGLMapView to call `MGLMapViewDelegate.mapView(_:viewFor:)`](https://github.com/mapbox/mapbox-gl-native/blob/3a3a072b7a34e949b9bcc0188c4048e61cdf24dc/platform/ios/src/MGLMapView.mm#L4184). Unfortunately, `navigationMapView` isn’t set until much later.

1ec5: Fixed in b35dfe17220689984bc2a6a10333efa225d9c9f1.

1ec5: Filed mapbox/mapbox-gl-native#9781 to address the root cause.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -47,7 +47,35 @@ open class NavigationMapView: MGLMapView {
...
+            super.showsUserLocation = newValue
+        }
+    }
+    
+    /**
+     Force MGLMapView to ask its delegate for the user location annotation’s view, in case that previously happened before there was a delegate.
+     */
+    func refreshShowsUserLocation() {
+        showsUserLocation = false
+        showsUserLocation = true
```
1ec5: This will result in `-[MGLMapView updateUserLocationAnnotationViewAnimatedWithDuration:]` getting called a lot more frequently than before when setting up the map view. From a cursory glance, I don’t think it’ll be a problem in practice.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -102,6 +102,11 @@ public protocol NavigationViewControllerDelegate {
      */
     @objc optional func navigationMapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?
     
+    /**
+     Returns a view object to mark the given point annotation object on the map.
+     
+     The user location annotation view can also be customized via this method. When annotation is an instance of `MGLUserLocation` (or equal to the map view’s userLocation property), return an instance of `MGLUserLocationAnnotationView` (or a subclass thereof).
```
1ec5: > or equal to the map view’s userLocation property

The developer doesn’t have direct access to `MGLMapView.userLocation`, do they?

bsudekum:
@1ec5 updated.

PR #504, opened 2017-08-16
Expose battery monitoring

frederoni:
~~This PR exposes `isMonitoringBatteryState` so a user can disable it and use their own desired accuracy.~~

@bsudekum @ericrwolfe 👀

### DIFF
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -19,6 +19,12 @@ open class NavigationLocationManager: CLLocationManager {
      */
     public private(set) var isPluggedIn: Bool = false
     
+    /**
+     Indicates whether the location manager’s desired accuracy should update
+     when the battery state changes.
+     */
+    public var automaticallyUpdatesDesiredAccuracy = true
+    
     override public init() {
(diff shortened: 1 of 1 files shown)


bsudekum:
@frederoni where exactly can they disable it?

bsudekum:
@frederoni where exactly can they disable it?

frederoni:
On the location manager before you start navigation
```swift
let locationManager = MyLocationManager()
locationManager.isMonitoringBatteryState = false
…
```
or reaching through the navigation and route controller
```swift
navigationViewController.routeController.locationManager.isMonitoringBatteryState = false
```

bsudekum:
I think we should rethink this. If the goal is to have control over the `kCLLocationAccuracyBestForNavigation` vs `kCLLocationAccuracyBest`, we should expose that. Battery monitoring should always be enabled for telemetry purposes.

frederoni:
Made a few changes here. We always observe the battery state but only update the desired accuracy when `automaticallyUpdatesDesiredAccuracy` is set to true.

However, the actual battery monitoring is something that the RouteController is responsible for.
This PR won't have any impact on telemetry or anything else by default.

PR #506, opened 2017-08-16
Avoid redundantly clearing shield image view

1ec5:
Fixed an issue causing the shield in the turn banner to go missing when swiping backwards. This redundant line was clearing the shield image; subsequently, the `roadCode` setter bailed because the `roadCode` had not changed. The `roadCode` setter is responsible for clearing the shield image when no shield should be displayed.

/cc @bhousel @frederoni


1ec5:
@bsudekum, is it only the shield view that fails to update?

bsudekum:
@1ec5 no, the entire maneuver does not update. Trying swiping forward and then hitting resume.

1ec5:
Yes, I’m seeing that too, also in master: #507.

bsudekum:
Maybe I'm doing something differently, but I'm not seeing this in master.

Issue #507, opened 2017-08-16
Stale turn banner after resuming


1ec5:
Tapping Resume after swiping the turn banner to a future step leaves the future step in the turn banner until the user triggers an alert level change.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/506#pullrequestreview-56697378
/cc @bsudekum

PR #508, opened 2017-08-16
Fix stale turn banner after resuming

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/507

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -137,6 +137,10 @@ class RouteMapViewController: UIViewController {
         mapView.camera = tiltedCamera
         mapView.setUserTrackingMode(.followWithCourse, animated: true)
         mapView.logoView.isHidden = false
+        
+        guard let controller = routePageViewController.currentManeuverPage else { return }
+        controller.step = currentStep
+        routePageViewController.notifyDidReRoute()
     }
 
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -137,6 +137,10 @@ class RouteMapViewController: UIViewController {
         mapView.camera = tiltedCamera
         mapView.setUserTrackingMode(.followWithCourse, animated: true)
         mapView.logoView.isHidden = false
+        
+        guard let controller = routePageViewController.currentManeuverPage else { return }
+        controller.step = currentStep
+        routePageViewController.notifyDidReRoute()
```
1ec5: `notifyDidReRoute()` is quite a misnomer. I’m afraid that we may in the future introduce logic to `notifyDidReRoute()` that only makes sense for rerouting and now recentering. Maybe we should rename the method now to avoid future surprises.

bsudekum: Yeah... I agree, I hate a few of these functions that force updates to various parts of the UI.

bsudekum:
@1ec5 updated function name.

bsudekum:
👍 , tested scenario, works.

PR #509, opened 2017-08-17
Say refs as characters

bsudekum:
_Note, this PR requires https://github.com/Project-OSRM/osrm-text-instructions.swift/pull/33_

Polly sometimes tries to expand a ref. Example:

```
CR 622
```

and polly would say:

```
Creek 622
```

This takes a look at a ref and checks it against our shield list. If the ref is found within the shield list, we know that the ref should be spelled out and not expanded. 


MapboxCoreNavigation/RouteStepFormatter.swift
```diff
@@ -23,6 +23,17 @@ public class RouteStepFormatter: Formatter {
         
         let modifyValueByKey = { (key: OSRMTextInstructions.TokenType, value: String) -> String in
             switch key {
+            case .code:
+                let refComponents = value.addingXMLEscapes.components(separatedBy: .whitespaces)
+                let refsContainingShieldValue = refComponents.filter {
```
1ec5: We should only be looking up the first word in `ShieldImageNamesByPrefix`. For example, hypothetically, if the code is `City 123 I`, we shouldn’t spell out C-I-T-Y.

MapboxCoreNavigation/RouteStepFormatter.swift
```diff
@@ -23,6 +23,17 @@ public class RouteStepFormatter: Formatter {
...
+                    return ShieldImageNamesByPrefix[$0] != nil
+                }
+                
+                guard refsContainingShieldValue.isEmpty else { return value }
+                
+                return refComponents.map {
+                    return Int($0) == nil ? "<say-as interpret-as=\"characters\">\($0.addingXMLEscapes)</say-as>" : "<say-as interpret-as=\"address\">\($0.addingXMLEscapes)</say-as>"
```
1ec5: [`I 69C`](http://www.openstreetmap.org/way/278020221) would be pronounced “eye six nine see” instead of “eye sixty-nine see”.

1ec5: `CA 237 Toll` would be pronounced “see-a two thirty-seven tee-oh-ell-ell”. `US 52 ALT` would be pronounced “you-ess fifty-two a-ell-tee”.

bsudekum:
@1ec5 this is ready for another round of review. Should we also be applying this same logic to `.wayName, .destination, .rotaryName` as well?

bsudekum:
@1ec5 fixed to check if it's caps instead

MapboxCoreNavigation/String.swift
```diff
@@ -40,4 +37,16 @@ extension String {
...
+    var asSSMLAddress: String {
+        return "<say-as interpret-as=\"address\">\(self.addingXMLEscapes)</say-as>"
+    }
+    
+    var asSSMLCharacters: String {
+        return "<say-as interpret-as=\"characters\">\(self.addingXMLEscapes)</say-as>"
+    }
+    
+    var isUppercased: Bool {
+        return (self == self.uppercased()) && (self != self.lowercased())
```
1ec5: Nit: extraneous parentheses; also, drop `self.`.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -138,6 +137,7 @@
 		C561735B1F182113005954F6 /* RouteStep.swift in Sources */ = {isa = PBXBuildFile; fileRef = C561735A1F182113005954F6 /* RouteStep.swift */; };
 		C578DA081EFD0FFF0052079F /* ProcessInfo.swift in Sources */ = {isa = PBXBuildFile; fileRef = C578DA071EFD0FFF0052079F /* ProcessInfo.swift */; };
 		C58159011EA6D02700FC6C3D /* MGLVectorSource.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58159001EA6D02700FC6C3D /* MGLVectorSource.swift */; };
+		C58454161F4B9FD30057D8B3 /* Shields.plist in Resources */ = {isa = PBXBuildFile; fileRef = 351BEC2A1E5BD538006FE110 /* Shields.plist */; };
```
1ec5: Can you verify that Shields.plist is where it was originally (as a resource in MapboxNavigation rather than MapboxCoreNavigation)?

bsudekum: image

PR #510, opened 2017-08-17
Locale and project cleanup

bsudekum:
These locale functions actually need to be made public. Also fix the project file listing. I think there might have been a bad merge conflict resolution recently. 

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxCoreNavigation/Locale.swift
+++ b/MapboxCoreNavigation/Locale.swift
@@ -1,7 +1,11 @@
 import Foundation
 
 extension Locale {
-    static var preferredLocalLanguageCountryCode: String {
+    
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/Locale.swift
```diff
@@ -16,5 +20,8 @@ extension Locale {
         return firstBundleLocale
     }
     
-    static var nationalizedCurrent = Locale(identifier: preferredLocalLanguageCountryCode)
+    /**
+     Returns a the `Locale` from `preferredLocalLanguageCountryCode`.
```
frederoni: Nit: typo

PR #511, opened 2017-08-17
While navigating, show background location indicator and upgrade to xcode 9

bsudekum:
Shows the location bar while the app is in navigation mode to remind the user to cancel the trip when they are done.

simulator screen shot - iphone 7 - 2017-08-17 at 16 45 37

Looks like we also need to update bitrise to ios 11...

/cc @1ec5 @frederoni


bsudekum:
Committing ios 11 test fixtures ands updated bitrise to xcode 9

PR #513, opened 2017-08-18
Fix Obj-C example

frederoni:
Fixed a bug in the Obj-C example.

@bsudekum 👀

### DIFF
--- a/Examples/Objective-C/ViewController.m
+++ b/Examples/Objective-C/ViewController.m
@@ -63,7 +63,7 @@
 }
 
 - (void)alertLevelDidChange:(NSNotification *)notification {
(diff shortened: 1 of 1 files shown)


Examples/Objective-C/ViewController.m
```diff
@@ -63,7 +63,7 @@ - (void)suspendNotifications {
 }
 
 - (void)alertLevelDidChange:(NSNotification *)notification {
-    MBRouteProgress *routeProgress = (MBRouteProgress *)notification.userInfo[MBRouteControllerNotificationProgressDidChange];
+    MBRouteProgress *routeProgress = (MBRouteProgress *)notification.userInfo[MBRouteControllerProgressDidChangeNotificationProgressKey];
```
frederoni: This incorrect notification caused the resulting string to be something like "In 0mm (null)"

Issue #423, opened 2017-07-25
Automatically switch to night style


1ec5:
NavigationMapView should automatically switch to the Navigation Guidance Night v2 style (`mapbox://styles/mapbox/navigation-guidance-night-v2`) when the ambient light level is low or when the Smart Invert Colors option is enabled in iOS 11.

Ideally we’d darken the rest of the UI at the same time, perhaps using a Style. But that can be a separate task if it would require significant changes. Just swapping the style would be a welcome improvement in low-light conditions.

/cc @bsudekum @frederoni @ericrwolfe @aparlato

frederoni:
We can rename DefaultStyle to DayStyle and create a NightStyle.
The ambient light or SIC trigger then just have to toggle `NavigationViewController.view.isHidden`
in order to switch between the day and night style for the UI.

A time-based trigger could also be interesting.

1ec5:
> or when the Smart Invert Colors option is enabled in iOS 11

I still think we should do this, based on [`UIAccessibilityIsInvertColorsEnabled()`](https://developer.apple.com/documentation/uikit/1615167-uiaccessibilityisinvertcolorsena) or the [`UIAccessibilityInvertColorsStatusDidChange`](https://developer.apple.com/documentation/foundation/nsnotification.name/1615196-uiaccessibilityinvertcolorsstatu) notification.

ericrwolfe:
@1ec5 This is an iOS 11 beta feature, can you please ticket separately?

1ec5:
→ #539

TCOA:
Is there a way for a developer to know which style is active?  I've had a request to do some features which require me to change the font depending on which is shown.

1ec5:
Yes, you can access the navigation view controller’s map view using the [`NavigationViewController.mapView`](https://www.mapbox.com/mapbox-navigation-ios/navigation/0.11.0/Classes/NavigationViewController.html#/c:@M@MapboxNavigation@objc%28cs%29MBNavigationViewController%28py%29mapView) property:

```swift
let styleURL = navigationViewController.mapView?.styleURL
```

TCOA:
Thanks, I do get some info, though it is confusing....

When I start the app, I have the mapView, which shows in light background (I would call that 'day') and I get 
THE STYLE FOR DAY/NIGHT IS  mapbox://styles/mapbox/streets-v10

When I put routes to the map, it shows in dark background (I would say 'night' there...) and see
[...]

PR #519, opened 2017-08-21
Automatically use dark style when dark

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/423

todo: 

- [x] Update UI theme more often then on init. Ideally, it'd be cheap enough we can check for this every progress/alert update.
- [x] Use a dark map style
- [x] Actually style dark theme
- [x] Get Solar working. Currently blocked by https://github.com/ceeK/Solar/issues/21

How do we want the developer to handle a custom day/night style? Right now, if you specify a custom style, you do not get a night style.

### DIFF
--- a/MapboxNavigation.xcodeproj/project.pbxproj
+++ b/MapboxNavigation.xcodeproj/project.pbxproj
@@ -313,12 +313,13 @@
 		351BEBDF1E5BCC63006FE110 /* MGLMapView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = MGLMapView.swift; sourceTree = "<group>"; };
 		351BEBE01E5BCC63006FE110 /* Style.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Style.swift; sourceTree = "<group>"; };
 		351BEBE31E5BCC63006FE110 /* RouteManeuverViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteManeuverViewController.swift; sourceTree = "<group>"; };
-		351BEBE41E5BCC63006FE110 /* RouteMapViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteMapViewController.swift; sourceTree = "<group>"; };
+		351BEBE41E5BCC63006FE110 /* RouteMapViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; lineEnding = 0; path = RouteMapViewController.swift; sourceTree = "<group>"; xcLanguageSpecificationIdentifier = xcode.lang.swift; };
 		351BEBE51E5BCC63006FE110 /* RoutePageViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RoutePageViewController.swift; sourceTree = "<group>"; };
 		351BEBE71E5BCC63006FE110 /* RouteTableViewCell.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteTableViewCell.swift; sourceTree = "<group>"; };
 		351BEBE81E5BCC63006FE110 /* RouteTableViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteTableViewController.swift; sourceTree = "<group>"; };
 		351BEBE91E5BCC63006FE110 /* RouteTableViewHeaderView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteTableViewHeaderView.swift; sourceTree = "<group>"; };
-		351BEBEA1E5BCC63006FE110 /* NavigationViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = NavigationViewController.swift; sourceTree = "<group>"; };
+		351BEBEA1E5BCC63006FE110 /* NavigationViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; lineEnding = 0; path = NavigationViewController.swift; sourceTree = "<group>"; xcLanguageSpecificationIdentifier = xcode.lang.swift; };
+		351BEBEB1E5BCC63006FE110 /* String.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = String.swift; sourceTree = "<group>"; };
 		351BEBED1E5BCC63006FE110 /* StyleKitArrows.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = StyleKitArrows.swift; sourceTree = "<group>"; };
 		351BEBEF1E5BCC63006FE110 /* TurnArrowView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = TurnArrowView.swift; sourceTree = "<group>"; };
 		351BEBF01E5BCC63006FE110 /* UIView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = UIView.swift; sourceTree = "<group>"; };
@@ -412,7 +413,7 @@
 		C52D09CB1DEF444D00BE3C5C /* GeometryTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = GeometryTests.swift; sourceTree = "<group>"; };
 		C52D09CD1DEF5E5100BE3C5C /* route.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = route.json; sourceTree = "<group>"; };
 		C52D09D21DEF636C00BE3C5C /* Fixture.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Fixture.swift; sourceTree = "<group>"; };
-		C53208AA1E81FFB900910266 /* NavigationMapView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = NavigationMapView.swift; sourceTree = "<group>"; };
+		C53208AA1E81FFB900910266 /* NavigationMapView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; lineEnding = 0; path = NavigationMapView.swift; sourceTree = "<group>"; xcLanguageSpecificationIdentifier = xcode.lang.swift; };
 		C53C193C1F38E0D5008DB406 /* VisualInstructionFormatter.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = VisualInstructionFormatter.swift; sourceTree = "<group>"; };
 		C53C19701F38EACD008DB406 /* zh-Hans */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = "zh-Hans"; path = "zh-Hans.lproj/Localizable.strings"; sourceTree = "<group>"; };
 		C53C19711F38EADB008DB406 /* it */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = it; path = it.lproj/Localizable.strings; sourceTree = "<group>"; };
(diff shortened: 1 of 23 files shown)


1ec5:
Do you think it’s better to lighten or darken the UI in response to ambient light changes or the sun’s position? I can see arguments either way:

* The interior of a car is always pretty dark during the day, so the threshold for switching styles would have to be very sensitive.
* If I’m traveling through a tunnel, it’d be nice to get the dark style. Maybe we could supplement Solar with feature querying that determines whether the user is in a tunnel?
* Weather conditions – or, er, eclipses – could make the dark style more comfortable even during the day, although I think the dark style is being designed to look more like a nighttime style.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -226,6 +226,20 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
...
+                return DefaultStyle()
+            } else {
+                return NightStyle()
+            }
+        }
+    }
+    
+    var isDaytime = true
+    
+    public var automaticallyAdjustStyleForSunPosition = true
```
1ec5: `automaticallyAdjustsStyleForTimeOfDay`

MapboxNavigation/DefaultStyle.swift
```diff
@@ -107,5 +107,14 @@ open class DefaultStyle: Style {
         
         routeCasingColor = .defaultRouteCasing
     }
+}
+
+/**
+ `NightStyle` is default night style for Mapbox Navigation SDK. Only will be applied if `automaticallyAdjustStyleForSunPosition` and `isDaytime` are true.
```
1ec5: `isDaytime` is an internal property, so it shouldn’t be mentioned in public documentation.

bsudekum:
@1ec5 as far as I'm aware there is not a public API we can use to get the ambient light around the device.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -259,7 +273,7 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         
         super.init(contentViewController: mapViewController, drawerViewController: tableViewController)
         
-        self.styles = styles ?? [DefaultStyle()]
+        self.styles = styles ?? [styleForSunPosition]
```
frederoni: I'd suggest we pass both DayStyle and NightStyle (so we have a ref if the sun sets while in navigation) but later when the style is applied, we'd filter out non-appropriate style. Adding a `dayTime` enum property (day/night) on `Style` could help.

1ec5:
> as far as I'm aware there is not a public API we can use to get the ambient light around the device.

There is a technically public ambient light sensor API in IOHID, but it’s nonetheless off-limits to application developers. An alternative would be to [observe the screen brightness](https://stackoverflow.com/a/15937484/4585461), since most users have Automatic Brightness set, which adjusts the screen’s brightness based on the ambient light sensor. I was also thinking it would be nice to adjust the style based on whether Night Shift is active, but there doesn’t appear to be a public API for that, either.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -368,6 +394,19 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
...
             navigationDelegate?.navigationViewController?(self, didArriveAt: destination)
         }
+        
+        updateStyleIfNeeded()
+    }
+    
+    func updateStyleIfNeeded() {
+        if let styles = styles, styles != [styleForTimeOfDayAndBrightness] {
+            self.styles = [styleForTimeOfDayAndBrightness]
+            applyStyle()
```
bsudekum: @frederoni I'm getting mixed results when setting the style after it's been set. Any ideas as to why it does not apply the new style reliably?

frederoni: I think it's because appearance proxy only sets the properties when the view is added to a window, it doesn't work in real-time. Here's a hack that forces a refresh but it needs to be adapted because the map view crashes if it looses its `UIWindow` when it's active.
```swift
func refreshAppearance() {
        for window in UIApplication.shared.windows {
            for view in window.subviews {
                view.removeFromSuperview()
                window.addSubview(view)
            }
        }
    }
```

Try removing and adding the maneuver view in the same drawing cycle.

MapboxNavigation/Style.swift
```diff
@@ -227,6 +227,11 @@ open class Style: NSObject {
     public var trafficSevereColor: UIColor?
     
     /**
+     Describes the situations in which the style should be used. By default, the style will be used during the daytime.
+     */
+    public var styleType: StyleType = .daytimeStyle
```
bsudekum: @frederoni how should a map style be linked to a UI style? Do you think we should add a new prop to Style that is the associated map style url?

frederoni: Sounds reasonable. You might have to add a dynamic var mapStyleURL to `NavigationMapView` and set the MGLMapViews’s styleURL in the didSet.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -135,13 +135,20 @@
...
 		C588C3C21F33882100520EF2 /* String.swift in Sources */ = {isa = PBXBuildFile; fileRef = 35BF8CA31F28EBD8003F6125 /* String.swift */; };
 		C588C3C31F3388A100520EF2 /* RouteStepFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 35C997401E732C5400544D1C /* RouteStepFormatter.swift */; };
 		C58D6BAD1DDCF2AE00387F53 /* Constants.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58D6BAC1DDCF2AE00387F53 /* Constants.swift */; };
+		C5A6B2DD1F4CE8E8004260EA /* StyleType.swift in Sources */ = {isa = PBXBuildFile; fileRef = C5A6B2DC1F4CE8E8004260EA /* StyleType.swift */; };
+		C5A6B2DE1F4DE57E004260EA /* Solar.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = C57607B01F4CC97D00C27423 /* Solar.framework */; };
+		C5A6B2DF1F4DE57E004260EA /* Solar.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = C57607B01F4CC97D00C27423 /* Solar.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		C5A6B2E01F4DE59E004260EA /* Solar.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = C57607B01F4CC97D00C27423 /* Solar.framework */; };
```
frederoni: This looks like duplicates. I usually drag the framework inside the Frameworks folder and uncheck the targets then add it to "Embedded frameworks" afterward.

MapboxNavigation/Style.swift
```diff
@@ -237,6 +237,11 @@ open class Style: NSObject {
     public var styleType: StyleType = .daytimeStyle
     
     /**
+     Map style to be used for the style.
+     */
+    public var mapStyle: URL?
```
frederoni: Nit: this name indicates that it's a style which it isn't. Can we rename to `mapStyleURL`?

bsudekum:
Making good progress here. Here is what the transition looks like currently:

old

Not bad, but there could be some low hanging fruit we can hit here. My main concern with this PR is what happens when:

* User creates custom UI style
* Modifies the map style on the actual map view
* What happens?

Should developers be able to modify the style on the actual map view? If they do both, who wins? Should mapStyleURL be required?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -373,6 +398,26 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
...
+        
+        applyStyle()
+    }
+    
+    func applyStyle() {
+        styles?.forEach {
+            if $0.styleType == styleForTimeOfDayAndBrightness {
+                $0.apply()
+                mapView?.styleURL = $0.mapStyle
+                UIApplication.shared.statusBarStyle = $0.statusBarStyle ?? .default
```
bsudekum: @frederoni for some reason, this is not working.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -373,6 +399,26 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         if routeProgress.currentLegProgress.alertUserLevel == .arrive {
             navigationDelegate?.navigationViewController?(self, didArriveAt: destination)
         }
+        
+        applyStyle()
```
frederoni: Can we optimize this a bit? Perhaps we could get `intervalUntilTimeOfDayChange` when we present the NavigationViewController and only fire it once when it should happen?

frederoni:
Using light and dark will limit styling later down the road. It prevents adding more than two styles and is more a description than a trait of the style. It also gives a false assumption that one style must be light and the other one dark.

I think these properties should be like custom traits. Day/Night and brightness would fit in perfectly.

Not entirely sure but we should be able to subclass `UITraitCollection` and provide something like `UITraitCollection(timeOfDay:)` but that's probably out of scope of this PR.

bsudekum:
>  It also gives a false assumption that one style must be light and the other one dark.

@frederoni good point. Curious when you created the `Style` class, was did you make the prop on `NavigationViewController` an array? If I'm reading this right, currently as it stands in master, we apply only the last style to the UI, [ref](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/NavigationViewController.swift#L189). What is the thinking here?

MapboxNavigation/NavigationViewController.swift
bsudekum: Are you fearful of constantly updating the style? If so, i added this guard here https://github.com/mapbox/mapbox-navigation-ios/pull/519/commits/e52c62138b8c8b480bb949e1d52893199b03e66d

MapboxNavigation/NavigationViewController.swift
```diff
@@ -259,7 +285,7 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         
         super.init(contentViewController: mapViewController, drawerViewController: tableViewController)
         
-        self.styles = styles ?? [DefaultStyle()]
+        self.styles = styles ?? [DefaultStyle(), DefaultDarkStyle()]
```
bsudekum: I'm sort of confused as to why we have a default value but also need this `??` here.

frederoni:
My thought was to make something like this work:

```swift
let cyclingLandscape = CyclingStyle(traitCollection: UITraitCollection(horizontalSizeClass: .compact)) // Using a specific font or color for when the device is in landscape mode
let cyclingImprovedAccessibility = CyclingStyle(traitCollection: UITraitCollection(preferredContentSizeCategory: .extraExtraExtraLarge))
cyclingImprovedAccessility.textColor = // Something with more contrast
let nightStyle = DefaultStyle()
nightStyle.dayOfTime = .night // Would later become a kind of trait

let styles = [cycling, cyclingImprovedAccessibility, nightStyle]
```

Styles that use a built-in trait collection doesn't need the hack with removing and adding to the window. Rotating to landscape is sufficient to trigger a style change that will transition smoothly to the new style and I hope we can eventually do the same for day/night or brightness.

We should apply all styles that rely on built-in traits but filter out styles that use our custom traits that don't match the current day of time or brightness.

MapboxNavigation/NavigationViewController.swift
frederoni: `NavigationViewController(for:directions:styles:locationManager:)` <- styles here is optional to make it convenient to start a navigation without having to pass a style to the initializer. 

In Swift it allows you to use `NavigationViewController(for:directions:locationManager:)`, unfortunately, in Obj-C, you have to pass `styles:nil` but we need to apply a default style, otherwise it would just look like it looks in IB.

bsudekum:
> We should apply all styles that use built-in traits but filter out styles that use our custom traits if they don't match the current day of time and brightness.

This makes sense to me, but when it becomes time to apply the dark style, we'll loose all the custom traits the user might have applied to the default day style, no?

frederoni:
> when it becomes time to apply the dark style, we'll loose all the custom traits the user might have applied to the default day style

This shouldn't be a problem as long as the NightStyle is a subclass of DayStyle.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -285,7 +285,7 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         
         super.init(contentViewController: mapViewController, drawerViewController: tableViewController)
         
-        self.styles = styles ?? [DefaultStyle(), DefaultDarkStyle()]
```
frederoni: We should pass in both styles by default but when we apply them, we do
```
styles?.filter { $0.timeOfDay == actualTimeOfDay }.forEach { $0.apply() }
```

MapboxNavigation/NavigationViewController.swift
```diff
@@ -226,6 +227,31 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
...
     }
     
+    /**
+     If true, the map style and UI will automatically be updated given the time of day and screen brightness.
+     */
+    public var automaticallyAdjustsStyleForTimeOfDayAndBrightness = false
+    
+    var currentStyleType: StyleType?
+    
+    var styleForTimeOfDayAndBrightness: StyleType {
```
frederoni: Nit: styleTypeFor…

MapboxNavigation/NavigationViewController.swift
```diff
@@ -252,6 +253,31 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
...
+    /**
+     If true, the map style and UI will automatically be updated given the time of day and screen brightness.
+     */
+    public var automaticallyAdjustsStyleForTimeOfDay = false
+    
+    var currentStyleType: StyleType?
+    
+    var styleTypeForTimeOfDay: StyleType {
+        guard automaticallyAdjustsStyleForTimeOfDay else { return .lightStyle }
+        guard UIScreen.main.brightness > 0.25 else { return .darkStyle }
```
ericrwolfe: Can we remove this before merging?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -362,11 +388,13 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
     func resumeNotifications() {
         NotificationCenter.default.addObserver(self, selector: #selector(progressDidChange(notification:)), name: RouteControllerProgressDidChange, object: routeController)
         NotificationCenter.default.addObserver(self, selector: #selector(alertLevelDidChange(notification:)), name: RouteControllerAlertLevelDidChange, object: routeController)
+        NotificationCenter.default.addObserver(self, selector: #selector(forceRefreshAppearanceIfNeeded), name: NSNotification.Name.UIScreenBrightnessDidChange, object: nil)
     }
     
     func suspendNotifications() {
         NotificationCenter.default.removeObserver(self, name: RouteControllerProgressDidChange, object: routeController)
         NotificationCenter.default.removeObserver(self, name: RouteControllerAlertLevelDidChange, object: routeController)
+        NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIScreenBrightnessDidChange, object: nil)
```
ericrwolfe: And this?

Issue #515, opened 2017-08-18
Strange stutter after completing a maneuver 


bsudekum:
There seems to be a strange stutter only after completing a maneuver. Previously we thought this was poppy related.

PR #520, opened 2017-08-22
Fix camera stutter after completing maneuver

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/515

`routePageViewController(:RoutePageViewController:willTransitionTo:)` gets called after each maneuver completion. Because of this, this chunk of code was executing:

```swift
if step == routeController.routeProgress.currentLegProgress.upComingStep {
  view.layoutIfNeeded()
  mapView.camera = tiltedCamera
}
```

I'm unsure which one of these is the culprit, but if the user is already in follow with course mode, we should not re-set it or change the camera.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -96,8 +96,6 @@ class RouteManeuverViewController: UIViewController {
         }
     }
     
-    var isPagingThroughStepList = false
-    
     override func viewDidLoad() {
         super.viewDidLoad()
         turnArrowView.backgroundColor = .clear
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -96,8 +96,6 @@ class RouteManeuverViewController: UIViewController {
         }
     }
     
-    var isPagingThroughStepList = false
```
bsudekum: This was unused.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -528,10 +528,8 @@ extension RouteMapViewController: RoutePageViewControllerDelegate {
         maneuverViewController.updateStreetNameForStep()
         
         updateLaneViews(step: step, alertLevel: .high)
-        
-        maneuverViewController.isPagingThroughStepList = true
 
-        if !isInOverviewMode {
+        if !isInOverviewMode, mapView.userTrackingMode != .followWithCourse {
```
bsudekum: This will be false when the user is swiping.

1ec5: Can you add this as a comment? Thanks!

PR #521, opened 2017-08-22
Alert the user they are in simulation mode

bsudekum:
This notifies the user they are currently viewing a route in simulation mode.

@frederoni have any ideas why the background is missing?

simulator screen shot - iphone 7 - 2017-08-22 at 08 57 49

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -115,6 +115,11 @@ open class RouteController: NSObject {
     }
     
     /**
+     If true, the `RouteController` was initialized with a `SimulatedLocationManager` as the `NavigationLocationManager`. This `Bool` is used to notifiy the user they are currently viewing the `NavigationViewController` in simulation mode.
+     */
+    public var isSimulationModeEnabled = false
+    
+    /**
      If true, location updates will be simulated when driving through tunnels or other areas where there is none or bad GPS reception.
      */
     public var isDeadReckoningEnabled = false
@@ -182,6 +187,10 @@ open class RouteController: NSObject {
         self.resumeNotifications()
         self.startEvents(route: route)
         self.resetSession()
+        
+        if locationManager is SimulatedLocationManager {
+            isSimulationModeEnabled = true
+        }
     }
     
     deinit {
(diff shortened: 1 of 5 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -115,6 +115,11 @@ open class RouteController: NSObject {
     }
     
     /**
+     If true, the `RouteController` was initialized with a `SimulatedLocationManager` as the `NavigationLocationManager`. This `Bool` is used to notifiy the user they are currently viewing the `NavigationViewController` in simulation mode.
+     */
+    public var isSimulationModeEnabled = false
```
frederoni: There's no need to add this to core. Let's just do a `if routeController.locationManager is SimulationLocationManager` in the UI part.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -314,6 +314,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         drawerCornerRadius = 0
         progressBar.dock(on: view)
         self.delegate = self
+        
+        if routeController.isSimulationModeEnabled {
```
frederoni: Try moving this to didAppear. It also improve discoverability as it will animate when you're the view is present. Easier to miss the status message if it's already there when you open the screen.
Also, adding a duration here might be useful.

bsudekum:
@frederoni updated

MapboxNavigation/NavigationViewController.swift
```diff
@@ -323,6 +323,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         routeController.resume()
         
         styles?.forEach { $0.apply() }
+        
+        if routeController.locationManager is SimulatedLocationManager {
+            let title = NSLocalizedString("USER_IN_SIMULATION_MODE", bundle: .mapboxNavigation, value: "Simulating Navigation", comment: "Inform the user they are in simulation mode")
```
1ec5: The comment (and the one in the .strings file) should ideally be detailed enough that the translator can find the string in the UI if necessary:

> The text of a banner that appears during turn-by-turn navigation when route simulation is enabled.

PR #526, opened 2017-08-23
Fix swiping ahead

bsudekum:
Bleh, had a bad regression in https://github.com/mapbox/mapbox-navigation-ios/pull/520. But this seems to fix _everything_.

/cc @frederoni @1ec5

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -533,7 +533,7 @@ extension RouteMapViewController: RoutePageViewControllerDelegate {
 
         
         if !isInOverviewMode {
-            if didSwipe == true, step != routeController.routeProgress.currentLegProgress.upComingStep {
+            if didSwipe, step != routeController.routeProgress.currentLegProgress.upComingStep {
                 mapView.setCenter(step.maneuverLocation, zoomLevel: mapView.zoomLevel, direction: step.initialHeading!, animated: true, completionHandler: nil)
             } else if mapView.userTrackingMode != .followWithCourse {
                 view.layoutIfNeeded()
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -532,15 +532,13 @@ extension RouteMapViewController: RoutePageViewControllerDelegate {
         updateLaneViews(step: step, alertLevel: .high)
 
         
-        if !isInOverviewMode,
-            // This will be false when the user is swiping
-            mapView.userTrackingMode != .followWithCourse {
-            if step == routeController.routeProgress.currentLegProgress.upComingStep {
+        if !isInOverviewMode {
+            if didSwipe == true, step != routeController.routeProgress.currentLegProgress.upComingStep {
```
1ec5: `didSwipe` is a Boolean, so no need to explicitly compare it to `true`.

MapboxNavigation/RoutePageViewController.swift
```diff
@@ -8,7 +8,7 @@ protocol RoutePageViewControllerDelegate: class {
     var upComingStep: RouteStep? { get }
     func stepBefore(_ step: RouteStep) -> RouteStep?
     func stepAfter(_ step: RouteStep) -> RouteStep?
-    func routePageViewController(_ controller: RoutePageViewController, willTransitionTo maneuverViewController: RouteManeuverViewController)
+    func routePageViewController(_ controller: RoutePageViewController, willTransitionTo maneuverViewController: RouteManeuverViewController, didSwipe: Bool)
```
1ec5: Is `didSwipe` supposed to indicate whether the method is being called due to a user gesture? If so, can you add a comment to that effect?

bsudekum: Yep, exactly.

PR #527, opened 2017-08-23
Restore < to format string

1ec5:
Undid an unexplained change to a format string that changed the string’s meaning. A couple localizations had already localized the “<” to something locally appropriate, while others had removed the “<” in keeping with the base string.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/457#discussion_r132809331
/cc @frederoni @bsudekum

### DIFF
--- a/MapboxNavigation/Resources/Base.lproj/Localizable.strings
+++ b/MapboxNavigation/Resources/Base.lproj/Localizable.strings
@@ -16,8 +16,8 @@
 /* Feedback type for Unallowed Turn */
 "FEEDBACK_UNALLOWED_TURN" = "Not Allowed";
 
-/* Format string for less than; 1 = duration remaining */
-"LESS_THAN" = "%@";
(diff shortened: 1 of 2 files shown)


frederoni:
The motivation for removing `<` from the UI was aesthetic and using a language not only engineers speak. Most navigation UIs seems to use 1 as the minimum for time remaining. I'd prefer that or spell out "less than". 
/cc @ericrwolfe

1ec5:
I’m not sure how the arithmetic less-than sign is any more engineering-centric than an abbreviation like “mi” or “ft”. But I’m in favor of spelling out “less than” if we also spell out “miles” and “feet”.

Otherwise, as I mentioned in https://github.com/mapbox/mapbox-navigation-ios/pull/457#discussion_r132809331, if there’s no need for a “<” or “less than” (or the localized equivalent), then there shouldn’t be a format string in the first place.

PR #529, opened 2017-08-23
Add Spanish cardinal directions to ref parsing

1ec5:
[Changeset 51,368,737](http://www.openstreetmap.org/changeset/51368737) adds the first occurrences of Spanish cardinal directions in `destination:ref` tags. This change adds the Spanish cardinal directions alongside the English and French ones, to facilitate shield selection. (Shields.plist currently includes a generic oval shield for Puerto Rico’s routes. In order to choose the correct design among the island’s three shield designs, OSRM would need to look at the route relations instead of the way `ref` tags.)

/ref Project-OSRM/osrm-text-instructions#119
/cc @bsudekum @frederoni

### DIFF
--- a/MapboxNavigation/RouteManeuverViewController.swift
+++ b/MapboxNavigation/RouteManeuverViewController.swift
@@ -54,7 +54,7 @@ class RouteManeuverViewController: UIViewController {
                 return
             }
             
-            if components.count == 2 || (components.count == 3 && ["North", "South", "East", "West", "Nord", "Sud", "Est", "Ouest"].contains(components[2])) {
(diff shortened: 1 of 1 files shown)


1ec5:
As ugly as this solution looks in code, it’s actually as scalable as we need it to be. Per https://github.com/Project-OSRM/osrm-text-instructions/issues/119#issuecomment-324279884, it’s unlikely that cardinal directions are signposted in any other language, given the jurisdictions that use the [MUTCD](https://en.wikipedia.org/wiki/Manual_on_Uniform_Traffic_Control_Devices) and thus pair routes with cardinal directions. 🇺🇸🇨🇦🇵🇷🇻🇮🇳🇿 (Out of curiosity, I also went hunting for examples of routes signposted with cardinal directions in Navajo or Canadian aboriginal syllabary but came up empty-handed.)

PR #531, opened 2017-08-23
Subclass route options

bsudekum:
This creates a sublcass of `RouteOptions` called `NavigationRouteOptions`. 

### DIFF
--- a/MapboxCoreNavigation/NavigationRouteOptions.swift
+++ b/MapboxCoreNavigation/NavigationRouteOptions.swift
@@ -2,7 +2,7 @@ import Foundation
 import MapboxDirections
 
 /**
- A NavigationRouteOptions object specifies turn-by-turn-optimized criteria for results returned by the Mapbox Directions API.
+ A `NavigationRouteOptions` object specifies turn-by-turn-optimized criteria for results returned by the Mapbox Directions API.
  
  `NavigationRouteOptions` is a subclass of `RouteOptions` that has been optimized for navigation. Pass an instance of this class into the `Directions.calculate(_:completionHandler:)` method.
  */
@@ -12,8 +12,8 @@ open class NavigationRouteOptions: RouteOptions {
     /**
      Initializes a navigation route options object for routes between the given waypoints and an optional profile identifier optimized for navigation.
      
-     - parameter waypoints: An array of `Waypoint` objects representing locations that the route should visit in chronological order. The array should contain at least two waypoints (the source and destination) and at most 25 waypoints. (Some profiles, such as `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, [may have lower limits](https://www.mapbox.com/api-documentation/#directions).)
-     - parameter profileIdentifier: A string specifying the primary mode of transportation for the routes. This parameter, if set, should be set to `MBDirectionsProfileIdentifierAutomobile`, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, `MBDirectionsProfileIdentifierCycling`, or `MBDirectionsProfileIdentifierWalking`. `MBDirectionsProfileIdentifierAutomobile` is used by default.
+     - SeeAlso:
+     [RouteOptions](https://www.mapbox.com/mapbox-navigation-ios/directions/0.10.1/Classes/RouteOptions.html)
      */
     public override init(waypoints: [Waypoint], profileIdentifier: MBDirectionsProfileIdentifier? = .automobileAvoidingTraffic) {
         super.init(waypoints: waypoints, profileIdentifier: profileIdentifier)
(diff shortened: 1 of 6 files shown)


MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import MapboxDirections
+
+/**
+ A `NavigationRouteOptions` object is a structure that specifies the criteria for results returned by the Mapbox Directions API.
+ 
+ `NavigationRouteOptions` is a subclass of `RouteOptions` that has been optimized for navigation. Pass an instance of this class into the `Directions.calculate(_:completionHandler:)` method.
```
1ec5: This is kind of silly, but what if we extended Directions to make NavigationRouteOptions a little more discoverable to folks who don’t read documentation?

```swift
extension Directions {
    @objc(calculateDirectionsWithNavigationOptions:completionHandler:)
    open func calculate(_ options: NavigationOptions, completionHandler: @escaping CompletionHandler) -> URLSessionDataTask {
        return calculate(options as RouteOptions, completionHandler)
    }
}
```

Disclaimer: I haven’t tried this; just playing around in a text box in a browser.

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -0,0 +1,52 @@
...
+ */
+@objc(MBNavigationRouteOptions)
+open class NavigationRouteOptions: RouteOptions {
+    
+    /**
+     Initializes a navigation route options object for routes between the given waypoints and an optional profile identifier optimized for navigation.
+     
+     - parameter waypoints: An array of `Waypoint` objects representing locations that the route should visit in chronological order. The array should contain at least two waypoints (the source and destination) and at most 25 waypoints. (Some profiles, such as `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, [may have lower limits](https://www.mapbox.com/api-documentation/#directions).)
+     - parameter profileIdentifier: A string specifying the primary mode of transportation for the routes. This parameter, if set, should be set to `MBDirectionsProfileIdentifierAutomobile`, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, `MBDirectionsProfileIdentifierCycling`, or `MBDirectionsProfileIdentifierWalking`. `MBDirectionsProfileIdentifierAutomobile` is used by default.
```
1ec5: I’m a little uncomfortable with mirroring all the RouteOptions documentation here. Could we get away with a `- see:` command? We might need to beef up [this line](https://github.com/mapbox/mapbox-navigation-ios/blob/5f3856376eaead146a58688bbed40df842b54e9d/scripts/document.sh#L50) that inserts links from the MapboxNavigation docset to the MapboxDirections docset.

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -0,0 +1,52 @@
...
+     Initializes a route options object for routes between the given geographic coordinates and an optional profile identifier optimized for navigation.
+     
+     - parameter coordinates: An array of geographic coordinates representing locations that the route should visit in chronological order. The array should contain at least two locations (the source and destination) and at most 25 locations. Each coordinate is converted into a `Waypoint` object.
+     - parameter profileIdentifier: A string specifying the primary mode of transportation for the routes. This parameter, if set, should be set to `MBDirectionsProfileIdentifierAutomobile`, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, `MBDirectionsProfileIdentifierCycling`, or `MBDirectionsProfileIdentifierWalking`. `MBDirectionsProfileIdentifierAutomobile` is used by default.
+     */
+    public convenience init(coordinates: [CLLocationCoordinate2D], profileIdentifier: MBDirectionsProfileIdentifier? = .automobileAvoidingTraffic) {
+        self.init(waypoints: coordinates.map { Waypoint(coordinate: $0) }, profileIdentifier: profileIdentifier)
+    }
+    
+    public required init?(coder decoder: NSCoder) {
```
1ec5: Nit: it probably isn’t necessary to override this method at this time.

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import MapboxDirections
+
+/**
+ A `NavigationRouteOptions` object is a structure that specifies the criteria for results returned by the Mapbox Directions API.
```
1ec5: We should briefly distinguish between this class and RouteOptions in the first sentence, for the benefit of developers looking at the code completion popup:

> A `NavigationRouteOptions` object specifies turn-by-turn-optimized criteria for results returned by the Mapbox Directions API.

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -0,0 +1,52 @@
...
+ 
+ `NavigationRouteOptions` is a subclass of `RouteOptions` that has been optimized for navigation. Pass an instance of this class into the `Directions.calculate(_:completionHandler:)` method.
+ */
+@objc(MBNavigationRouteOptions)
+open class NavigationRouteOptions: RouteOptions {
```
1ec5: I had been thinking of calling this class NavigationOptions, but NavigationRouteOptions makes more sense. 👍

MapboxCoreNavigation/RouteController.swift
```diff
@@ -182,6 +182,10 @@ open class RouteController: NSObject {
         self.resumeNotifications()
         self.startEvents(route: route)
         self.resetSession()
+        
+        if !(route.routeOptions is NavigationRouteOptions) {
+            print("`Route` was created using `RouteOptions` and not `NavigationRouteOptions`. Although not required, this may lead to a suboptimal navigation experience.")
```
1ec5: Let’s spell out what the developer gains by using NavigationRouteOptions: congestion coloring along the route line, better ETAs, what else?

MapboxCoreNavigation/RouteOptions.swift
```diff
@@ -1,45 +1,13 @@
...
 import MapboxDirections
 
 extension RouteOptions {
     var activityType: CLActivityType {
         switch self.profileIdentifier {
-        case MBDirectionsProfileIdentifier.cycling, MBDirectionsProfileIdentifier.walking:
-            return .fitness
-        default:
-            return .automotiveNavigation
+            case MBDirectionsProfileIdentifier.cycling, MBDirectionsProfileIdentifier.walking:
```
1ec5: Is Xcode 9 indenting `case` now? 🙀

MapboxCoreNavigation/NavigationRouteOptions.swift
bsudekum: It suggests a fix-it here.

MapboxCoreNavigation/NavigationRouteOptions.swift
bsudekum: What would the `- see` command look like here?

1ec5: Sorry, that should be [`- SeeAlso:`](https://developer.apple.com/library/content/documentation/Xcode/Reference/xcode_markup_formatting_ref/SeeAlso.html).

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import MapboxDirections
+
+/**
+ A NavigationRouteOptions object specifies turn-by-turn-optimized criteria for results returned by the Mapbox Directions API.
```
1ec5: Put `NavigationRouteOptions` in backticks for consistency.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -185,6 +185,10 @@ open class RouteController: NSObject {
         self.resumeNotifications()
         self.startEvents(route: route)
         self.resetSession()
+        
+        if !(route.routeOptions is NavigationRouteOptions) {
+            print("`Route` was created using `RouteOptions` and not `NavigationRouteOptions`. Although not required, this may lead to a suboptimal navigation experience. Without `NavigationRouteOptions`, it is not guaranteed you will get congestion along the route line, better ETAs and ETA label color dependent on congestion.")
```
1ec5: None of these features are implemented in Core Navigation, only in MapboxNavigation. Can we move this check to NavigationViewController’s initializer?

Also, use a (lazily initialized) global or static property to track whether this message has been printed before and avoid printing it a second time.

Issue #523, opened 2017-08-22
Swiping turn banner forward should update arrow on route line


1ec5:
When swiping the turn banner to preview a future maneuver, the map should update to display an arrow on the previewed maneuver, not the current maneuver.

/cc @bsudekum @frederoni

PR #532, opened 2017-08-24
Move arrow when maneuver is swiped

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/523

old

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/MGLMapView.swift
+++ b/MapboxNavigation/MGLMapView.swift
@@ -46,9 +46,10 @@ extension MGLMapView {
         }
     }
     
-    func addArrow(_ routeProgress: RouteProgress) {
-        let maneuverCoordinate = routeProgress.currentLegProgress.upComingStep?.maneuverLocation
-        let polylineCoordinates = routeProgress.route.coordinates
+    func addArrow(route: Route, legIndex: Int, stepIndex: Int) {
+        let step = route.legs[legIndex].steps[stepIndex]
+        let maneuverCoordinate = step.maneuverLocation
+        let polylineCoordinates = route.coordinates
         
         guard let style = style else {
             return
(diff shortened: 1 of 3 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -541,6 +541,10 @@ extension RouteMapViewController: RoutePageViewControllerDelegate {
                 mapView.setUserTrackingMode(.followWithCourse, animated: true)
             }
         }
+        
+        if let stepIndex = routeController.routeProgress.currentLeg.steps.index(where: { $0 == step }) {
+            mapView.addArrow(route: routeController.routeProgress.route, legIndex: routeController.routeProgress.legIndex, stepIndex: stepIndex)
```
bsudekum: We'll have to figure out how to handle multiple legs here. I'm not sure even what happens when you try to swipe past the current leg.

1ec5: Could we get away with disallowing swiping to later legs for now?

bsudekum: Yep, that's how this will work right now. You will only be able to swipe on the current leg.

MapboxNavigation/MGLMapView.swift
```diff
@@ -46,24 +46,25 @@ extension MGLMapView {
         }
     }
     
-    func addArrow(_ routeProgress: RouteProgress) {
-        let maneuverCoordinate = routeProgress.currentLegProgress.upComingStep?.maneuverLocation
-        let polylineCoordinates = routeProgress.route.coordinates
+    func addArrow(route: Route, legIndex: Int, stepIndex: Int) {
+        let step = route.legs[legIndex].steps[stepIndex]
```
1ec5: If the user manages to swipe past the last step in the leg or the last step in the route, this line will crash.

bsudekum: Which you cannot do, but valid point, will guard.

MapboxNavigation/RouteMapViewController.swift
ericrwolfe: 👍 This is fine

bsudekum:
Merging after I merge https://github.com/mapbox/mapbox-navigation-ios/pull/270

PR #534, opened 2017-08-24
Only apply appropriate style on load

bsudekum:
Follow up https://github.com/mapbox/mapbox-navigation-ios/pull/519

On load, we were applying all styles. We should only apply the styles that are needed for that time of day.

/cc @frederoni @1ec5

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -423,9 +423,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
     }
     
     func applyStyle() {
-        styles?.forEach {
-            if $0.styleType == styleTypeForTimeOfDay {
-                $0.apply()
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -423,7 +423,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
     }
     
     func applyStyle() {
-        styles?.forEach { $0.apply() }
+        styles?.forEach {
```
1ec5: There’s no need for `forEach` here. Can you make this a normal `for` loop?

bsudekum: it does handle the optional `styles` better, but shrug.

1ec5: The way I usually handle that is:

```swift
for style in styles ?? [] { … }
```

Issue #533, opened 2017-08-24
Warning - Instance method 'navigationMapView(_:imageFor:)' nearly matches optional requirement 'navigationMapView(_:shapeFor:)' of protocol 'NavigationMapViewDelegate'


bsudekum:
I think this occurred after waypoints merged. 

![image](https://user-images.githubusercontent.com/1058624/29685892-c1f1ec72-88cb-11e7-990e-0a1bbb3c35c0.png)

/cc @frederoni @1ec5

1ec5:
Yes, this is fallout from #270.

PR #535, opened 2017-08-25
Better solar install, remove non-optional, default line color when no congestion

bsudekum:
Updates a few minor issues:
* Uses better version of solar
* If the user requests a route with congestion, things could fail. This falls back gracefully
* Removes a `!`
* Fixes: https://github.com/mapbox/mapbox-navigation-ios/issues/533

/cc @1ec5

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -5,4 +5,4 @@ github "rs/SDWebImage" ~> 4.1
 github "Project-OSRM/osrm-text-instructions.swift" ~> 0.3.1
 github "frederoni/aws-sdk-ios" "ac6a35f135aca73c9d308e960a501fcc16a586a8"
 github "mapbox/mapbox-telemetry-ios" ~> 0.2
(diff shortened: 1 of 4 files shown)


Cartfile
```diff
@@ -5,4 +5,4 @@ github "rs/SDWebImage" ~> 4.1
 github "Project-OSRM/osrm-text-instructions.swift" ~> 0.3.1
 github "frederoni/aws-sdk-ios" "ac6a35f135aca73c9d308e960a501fcc16a586a8"
 github "mapbox/mapbox-telemetry-ios" ~> 0.2
-github "https://github.com/bsudekum/Solar" "c162c697652c3dfe389717f6c9070a0421b3f0cd"
+github "https://github.com/1ec5/Solar" "5ce2ca7baaa82b1abb0467642ae237faf674b645"
```
1ec5: This is what I used locally:

```carthage
github "1ec5/Solar" "scheme-test-only"
```

The Cartfile.resolved will end up with a specific hash.

PR #536, opened 2017-08-25
Refactor styling

frederoni:
Made a few changes here since trait collections are usually applied per property, not per style.
This also makes it a lot more convenient to style components that aren't explicitly exposed in the style.

A custom night style would look something like this:
```swift
class CustomNightStyle: DayStyle {
    
    required init() {
        super.init()
        mapStyleURL = URL(string: "mapbox://styles/mapbox/satellite-streets-v9")!
        styleType = .nightStyle
    }
    
    override func apply() {
        super.apply()
        ManeuverView.appearance().backgroundColor = #colorLiteral(red: 0.2974345386, green: 0.4338284135, blue: 0.9865127206, alpha: 1)
        RouteTableViewHeaderView.appearance().backgroundColor = #colorLiteral(red: 0.2974345386, green: 0.4338284135, blue: 0.9865127206, alpha: 1)
    }
}
```

@bsudekum 👀


Examples/Swift/ViewController.swift
```diff
@@ -326,3 +289,16 @@ class CustomAnnotationView: MGLUserLocationAnnotationView {
...
+class CustomStyle: DefaultStyle {
+    override func apply() {
+        super.apply()
+        ManeuverView.appearance().backgroundColor = #colorLiteral(red: 0.2974345386, green: 0.4338284135, blue: 0.9865127206, alpha: 1)
+        RouteTableViewHeaderView.appearance().backgroundColor = #colorLiteral(red: 0.2974345386, green: 0.4338284135, blue: 0.9865127206, alpha: 1)
+        
+        DistanceLabel.appearance().textColor = #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
+        DestinationLabel.appearance().textColor = #colorLiteral(red: 0.8039215803, green: 0.8039215803, blue: 0.8039215803, alpha: 1)
+        TimeRemainingLabel.appearance().textColor = #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
+        ArrivalTimeLabel.appearance().textColor = #colorLiteral(red: 0.8039215803, green: 0.8039215803, blue: 0.8039215803, alpha: 1)
```
bsudekum: While we're at it, want to add a `StyleType` enum to this style and also add a custom dark theme? Would be cool to show custom night/day switching styles in the example.

Examples/Swift/ViewController.swift
```diff
@@ -162,10 +162,10 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         
         exampleMode = .default
         
-        let navigationViewController = NavigationViewController(for: route, locationManager: locationManager())
+        let navigationViewController = NavigationViewController(for: route, styles: [DefaultStyle()], locationManager: locationManager())
```
bsudekum: This is sort of a confusing case. It's sort of half the default case since we're not passing in the default dark style? Thoughts on reverting this change?

frederoni: Ah, yeah, this was just to force the light style while working on it.

bsudekum:
This is looking good, however I'm getting the default dark style in the demo app.

Examples/Swift/ViewController.swift
frederoni: Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/536/commits/29b1269d2aa7c07ee75ea61598ca0c1c4d20b65f

frederoni:
@bsudekum We should be good to ✅ here.

MapboxNavigation/Style.swift
```diff
@@ -564,7 +210,7 @@ public class ProgressBar: UIView {
     
     let bar = UIView()
     
-    dynamic var barColor: UIColor = #colorLiteral(red: 0.1411764771, green: 0.3960784376, blue: 0.5647059083, alpha: 1) {
+    dynamic public var barColor: UIColor = #colorLiteral(red: 0.1411764771, green: 0.3960784376, blue: 0.5647059083, alpha: 1) {
```
bsudekum: This needs documentation.

MapboxNavigation/UIFont.swift
```diff
@@ -23,7 +23,7 @@ extension UIFont {
     }
     
     // Adjusted for preferredContentSizeCategory (`adjustsFontForContentSizeCategory` is supported iOS >= 10.0)
-    var adjustedFont: UIFont {
+    public var adjustedFont: UIFont {
```
bsudekum: documentation needed.

PR #538, opened 2017-08-25
Check for valid indices

bsudekum:
Before looking up the keys.

/cc @frederoni

### DIFF
--- a/MapboxNavigation/RouteTableViewController.swift
+++ b/MapboxNavigation/RouteTableViewController.swift
@@ -56,6 +56,8 @@ class RouteTableViewController: UIViewController {
             return
         }
         
+        guard routeProgress.congestionTravelTimesSegmentsByStep.indices.contains(routeProgress.legIndex),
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteTableViewController.swift
```diff
@@ -56,6 +56,8 @@ class RouteTableViewController: UIViewController {
             return
         }
         
+        guard routeProgress.congestionTravelTimesSegmentsByStep.indices.contains(routeProgress.legIndex),
+            routeProgress.congestionTravelTimesSegmentsByStep[routeProgress.legIndex].indices.contains(routeProgress.currentLegProgress.stepIndex) else { return }
```
1ec5: `indices.contains` is a very roundabout way of checking whether an index is within bounds. How about `routeProgress.legIndex < congestionTravelTimesSegmentsByStep.count`?

bsudekum:
@1ec5 fixed

bsudekum:
@1ec5 I believe this was only showing up because fred was on an outdated git sha.

PR #540, opened 2017-08-25
Snap location and course when speed is low

bsudekum:
If the course is inaccurate or the speed is low and the user is on the route, we should just snap them. The case we're guarding against here is a user sitting in traffic where the course cannot be trusted.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -265,10 +265,6 @@ open class RouteController: NSObject {
         guard let stepCoordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return nil }
         guard let snappedCoordinate = closestCoordinate(on: stepCoordinates, to: location.coordinate) else { return location }
         
-        guard location.speed >= 0 else {
-            return location
-        }
-        
         let nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
         guard let closest = closestCoordinate(on: nearByCoordinates, to: location.coordinate) else { return nil }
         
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -292,11 +288,10 @@ open class RouteController: NSObject {
...
         let absoluteDirection = wrap(wrappedCourse + averageRelativeAngle, min: 0 , max: 360)
         
-        // If the course is inaccurate and the user is on the route,
-        // calculate a rough estimate as to what the course should be at that point on the route.
-        if location.course <= 0 && snappedCoordinate.distance < RouteControllerUserLocationSnappingDistance {
-            let calculatedCourse = wrap((wrappedPointBehind + wrappedPointAhead) / 2, min: 0 , max: 360)
-            return CLLocation(coordinate: snappedCoordinate.coordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: calculatedCourse, speed: location.speed, timestamp: location.timestamp)
+        // If the course is inaccurate or the speed is low and the user is on the route,
+        // snap the users location and course since we know the snapped location and course is more accurate.
+        if location.course <= 0 || location.speed <= 2, snappedCoordinate.distance < RouteControllerUserLocationSnappingDistance {
```
1ec5: Factor out 2 km/h into a constant. Also, for reference, human walking speed is between 4 and 7 km/h.

ericrwolfe: That's some fast walking.

Issue #461, opened 2017-08-07
Fix SDK identifier


ericrwolfe:
Currently, the `sdkIdentifier` used for metrics is [`MapboxEventsNavigationiOS`](https://github.com/mapbox/mapbox-navigation-ios/blob/b892672c31d14aeee5e369e013d8fa1787be5c62/MapboxCoreNavigation/RouteController.swift#L200). This should be `mapbox-navigation-ui-ios` or `mapbox-navigation-ios` depending on if [`usesDefaultUserInterface`](https://github.com/mapbox/mapbox-navigation-ios/blob/b892672c31d14aeee5e369e013d8fa1787be5c62/MapboxCoreNavigation/RouteController.swift#L148) is set. 

This also might mean pulling `usesDefaultUserInterface` into the init function, as it's currently set [after init](https://github.com/mapbox/mapbox-navigation-ios/blob/34934101a0770c730ec61947318d7453c8d79d43/MapboxNavigation/NavigationViewController.swift#L255).

cc @bsudekum

PR #543, opened 2017-08-25
Use proper user agent identifier

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/461

Defers the `startEvents` until `usesDefaultUserInterface` has a chance to be set with a proper value.
 
/cc @1ec5 @ericrwolfe


bsudekum:
Tested with both the standard UI and custom UI example.

Issue #546, opened 2017-08-26
Screen is black after viewing once in background 


bsudekum:
I think this has to do with removing and adding subviews. 

Cc @frederoni

bsudekum:
This occurs when rerouting happens in the background.

bsudekum:
Haven't seen this recently.

bsudekum:
Actually, seeing this again. I can reproduce it without the UI switching from day <-> night which means the view adding/remove hack code is not the culprit. 

@frederoni any ideas what is happening here?

bsudekum:
Doesn't happen any more.

PR #548, opened 2017-08-27
Set styleTypeForTimeOfDay on init

bsudekum:
This fixes an issue that caused the route line to not update after rerouting. This is a crazy codepath, but this guard was truthful too often https://github.com/mapbox/mapbox-navigation-ios/blob/e3860c9f388dd5de7b3599d984fadfe2c910024f/MapboxNavigation/NavigationViewController.swift#L433

[view.window here](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/RouteMapViewController.swift#L542) would then remain false as we update the UI by removing and adding views.

/cc @1ec5 @frederoni @willwhite @lxbarth

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -321,6 +321,8 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         
         tableViewController.routeController = routeController
         tableViewController.headerView.delegate = self
+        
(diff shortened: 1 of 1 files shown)


bsudekum:
This will help subdue  https://github.com/mapbox/mapbox-navigation-ios/issues/546 but it will still go black when transitioning from day <-> night.

PR #549, opened 2017-08-27
Use Style for unknown eta color

bsudekum:
We were not properly setting the ETA color according to the style when it was unknown.

@frederoni What do you think of this approach? Any other way we can fix this?

/cc @pveugen

### DIFF
--- a/MapboxNavigation/RouteTableViewController.swift
+++ b/MapboxNavigation/RouteTableViewController.swift
@@ -16,6 +16,8 @@ class RouteTableViewController: UIViewController {
     @IBOutlet var headerView: RouteTableViewHeaderView!
     @IBOutlet weak var tableView: UITableView!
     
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteTableViewController.swift
```diff
@@ -25,6 +27,10 @@ class RouteTableViewController: UIViewController {
         distanceFormatter.numberFormatter.locale = .nationalizedCurrent
     }
     
+    override func viewDidLoad() {
+        defaultTimeRemainingColor = headerView.timeRemaining.textColor
```
frederoni: Not calling super here can result in all kinds of obscure behavior.

ericrwolfe:
Closing in favor of a traitCollection style refactor @frederoni. PR to come.

PR #550, opened 2017-08-27
Don't show seconds in ETA

bsudekum:
When the nav ui mounts, we momentarily show the ETA with the default `DateComponentsFormatter` options which includes seconds. It looks like setting the options in `viewWillAppear` is not early enough.

/cc @frederoni @1ec5

### DIFF
--- a/MapboxNavigation/RouteTableViewController.swift
+++ b/MapboxNavigation/RouteTableViewController.swift
@@ -19,10 +19,6 @@ class RouteTableViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         setupTableView()
-        dateFormatter.timeStyle = .short
-        dateComponentsFormatter.allowedUnits = [.hour, .minute]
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteTableViewController.swift
```diff
@@ -35,6 +31,11 @@ class RouteTableViewController: UIViewController {
         let arrivalDate = NSCalendar.current.date(byAdding: .second, value: Int(routeProgress.durationRemaining), to: Date())
         headerView.arrivalTimeLabel.text = dateFormatter.string(from: arrivalDate!)
         
+        distanceFormatter.numberFormatter.locale = .nationalizedCurrent
+        dateFormatter.timeStyle = .short
+        dateComponentsFormatter.allowedUnits = [.hour, .minute]
+        dateComponentsFormatter.unitsStyle = .abbreviated
```
1ec5: I think it would be more logical to move these lines to `viewDidLoad()`.

bsudekum:
Wow, I did not mean to merge this.

PR #552, opened 2017-08-28
Limit interpret-as=address to numbers in road names

bsudekum:
Came across a case where polly speaks `N Street NW` as `North Street NW`. The if-first-word-in-ref code we had in place for refs should also be applied to `.wayName, .destination, .rotaryName`.

/cc @ericrwolfe @frederoni @1ec5

### DIFF
--- a/MapboxCoreNavigation/RouteStepFormatter.swift
+++ b/MapboxCoreNavigation/RouteStepFormatter.swift
@@ -25,11 +25,10 @@ public class RouteStepFormatter: Formatter {
             switch key {
             case .wayName, .destination, .rotaryName, .code:
                 let stringComponents = value.addingXMLEscapes.components(separatedBy: .whitespaces)
-                guard var firstRefComponent = stringComponents.first else { return value.asSSMLAddress }
                 
-                firstRefComponent = firstRefComponent.isUppercased ? firstRefComponent.asSSMLCharacters : firstRefComponent.asSSMLAddress
-                
-                return "\(firstRefComponent) \(stringComponents.suffix(from: 1).joined(separator: " ").asSSMLAddress)"
+                return stringComponents.map {
+                    $0.isInt ? $0.asSSMLAddress : $0
+                }.joined(separator: " ")
             default:
                 return value
             }
--- a/MapboxCoreNavigation/RouteStepFormatter.swift
+++ b/MapboxCoreNavigation/RouteStepFormatter.swift
@@ -23,15 +23,13 @@ public class RouteStepFormatter: Formatter {
         
         let modifyValueByKey = { (key: OSRMTextInstructions.TokenType, value: String) -> String in
             switch key {
-            case .code:
-                let refComponents = value.addingXMLEscapes.components(separatedBy: .whitespaces)
-                guard var firstRefComponent = refComponents.first else { return value.asSSMLAddress }
+            case .wayName, .destination, .rotaryName, .code:
+                let stringComponents = value.addingXMLEscapes.components(separatedBy: .whitespaces)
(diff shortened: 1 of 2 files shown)


1ec5:
What if the street is named [`North S Street`](http://www.openstreetmap.org/way/10921209), [`East W Avenue`](http://www.openstreetmap.org/way/287623931) or [`Northwest N Street`](http://www.openstreetmap.org/way/17590370)?

This is symptomatic of a broader issue we have with Polly, which is that it keeps trying to expand abbreviations even though abbreviations are already expanded in OpenStreetMap. Can you look into whether there’s a way to specify in SSML that abbreviation expansion should be disabled?

MapboxCoreNavigation/String.swift
```diff
@@ -52,4 +52,8 @@ extension String {
     var isUppercased: Bool {
         return self == uppercased() && self != lowercased()
     }
+    
+    var isInt: Bool {
+        return Int(self) != nil
```
1ec5: I think this will miss route numbers like [`NY 120A`](https://en.wikipedia.org/wiki/New_York_State_Route_120A), in which `120A` should be wrapped in `interpret-as="address"` in order to say “one-twenty a” instead of “one hundred twenty a”. How about wrapping a word in `address` if it contains a digit?

MapboxCoreNavigation/RouteStepFormatter.swift
```diff
@@ -23,15 +23,12 @@ public class RouteStepFormatter: Formatter {
         
         let modifyValueByKey = { (key: OSRMTextInstructions.TokenType, value: String) -> String in
             switch key {
-            case .code:
-                let refComponents = value.addingXMLEscapes.components(separatedBy: .whitespaces)
-                guard var firstRefComponent = refComponents.first else { return value.asSSMLAddress }
+            case .wayName, .destination, .rotaryName, .code:
+                let stringComponents = value.addingXMLEscapes.components(separatedBy: .whitespaces)
```
1ec5: FYI, this will miss any codes like `A-35`. OSM tagging conventions call for a space, as in `I 35`, but I do see some other regions using hyphens. We could harden this code by enumerating the string with `String.enumerateSubstrings(in:options:_:)` and `byWord`, or with `String.replacingOccurrences(of:with:options:range:)` and `regularExpression`.

bsudekum:
@1ec5 current code produces:

```
<speak>
  <prosody volume='x-loud' rate='1.08'>
    Head east on NY <say-as interpret-as="address">120A</say-as> <say-as interpret-as="address">A-35</say-as>, then in 3 quarters of a mile , Turn right onto NY <say-as interpret-as="address">120A</say-as> <say-as interpret-as="address">A-35</say-as>
  </prosody>
</speak>
```

bsudekum:
link to mp3: https://cloudup.com/c_X5YsKdCFh

MapboxCoreNavigation/String.swift
```diff
@@ -53,7 +53,7 @@ extension String {
         return self == uppercased() && self != lowercased()
     }
     
-    var isInt: Bool {
-        return Int(self) != nil
+    var containsInt: Bool {
```
1ec5: Nit: `containsDecimalDigit` would be more accurate.

MapboxCoreNavigation/String.swift
```diff
@@ -52,4 +52,8 @@ extension String {
     var isUppercased: Bool {
         return self == uppercased() && self != lowercased()
     }
+    
+    var containsInt: Bool {
```
1ec5: Nit: `containsDecimalDigit` would be more accurate.

PR #556, opened 2017-08-29
Filter out locations with big speed jumps

bsudekum:
This filters out new locations by comparing it to previous well known locations. If the new delta is greater that 3x the previous delta, throw out this location as the distance/time is too great.

### DIFF
--- a/MapboxCoreNavigation/Constants.swift
+++ b/MapboxCoreNavigation/Constants.swift
@@ -150,3 +150,8 @@ let milesToMeters = 1609.34
  The mimimum speed value before the user is snapped to the route. This is used to overcome inaccurate course values when a user's speed is low.
  */
 public var RouteControllerMinimumSpeedThresholdForSnappingUserToRoute: CLLocationSpeed = 2
+
+/**
+ Multiplier used to compare the current speed delta from the previous speed delta. When the difference in speed delta is greater than 3 times, the location is thrown out.
+ */
+public var RouteControllerDifferenceInSpeedDeltaMultiplier: CLLocationSpeed = 3
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -262,13 +262,33 @@ open class RouteController: NSObject {
      */
     var rawLocation: CLLocation?
     
+    struct PreviousLocationInfo {
+        var location: CLLocation
+        var locationDelta: CLLocationDistance
+        var durationDelta: TimeInterval
+    }
+    
+    var previousKnownLocation: PreviousLocationInfo?
+    
     /**
      The most recently received user location, snapped to the route line.
      
      This property contains a `CLLocation` object located along the route line near the most recently received user location. This property is set to `nil` if the route controller is unable to snap the user’s location to the route line for some reason.
      */
     public var location: CLLocation? {
-        guard let location = rawLocation, userIsOnRoute(location) else { return nil }
+        
+        guard let location = rawLocation else { return nil }
+        
+        if previousKnownLocation == nil {
+            previousKnownLocation = PreviousLocationInfo(location: location, locationDelta: 0, durationDelta: 0)
+        } else {
+            guard isQualityLocation(location: rawLocation) else {
+                return nil
+            }
+        }
+        
+        guard userIsOnRoute(location) else { return nil }
+        
         guard let stepCoordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return nil }
         guard let snappedCoordinate = closestCoordinate(on: stepCoordinates, to: location.coordinate) else { return location }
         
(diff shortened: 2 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -262,13 +262,33 @@ open class RouteController: NSObject {
...
     public var location: CLLocation? {
-        guard let location = rawLocation, userIsOnRoute(location) else { return nil }
+        
+        guard let location = rawLocation else { return nil }
+        
+        if previousKnownLocation == nil {
+            previousKnownLocation = PreviousLocationInfo(location: location, locationDelta: 0, durationDelta: 0)
+        } else {
+            guard isQualityLocation(location: rawLocation) else {
+                return nil
```
bsudekum: We might want to snap the user to a calculated location in this case.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -315,6 +335,40 @@ open class RouteController: NSObject {
...
+                
+                previousKnownLocation?.location = location
+            }
+        }
+        
+        guard let lastLocation = previousKnownLocation, let location = rawLocation else {
+                return false
+        }
+        
+        guard location.distance(from: lastLocation.location) <= lastLocation.locationDelta * 3 else {
```
1ec5: If the user enters a 500-foot-long tunnel at 60 mph, does the next location get tossed out as a poor-quality location?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -315,6 +335,40 @@ open class RouteController: NSObject {
...
+            return false
+        }
+        
+        guard location.timestamp.timeIntervalSince(lastLocation.location.timestamp) <= lastLocation.durationDelta * 3 else {
```
1ec5: If the user enters a tunnel and doesn’t exit for over 3 seconds, does the next location get tossed out as a poor-quality location?

MapboxCoreNavigation/RouteController.swift
ericrwolfe: Yes, this is the point. We basically throw out the next location update until locations start streaming in at a regular interval again.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -262,13 +262,33 @@ open class RouteController: NSObject {
      */
     var rawLocation: CLLocation?
     
+    struct PreviousLocationInfo {
```
1ec5: This struct essentially holds the delta between two location updates. Let’s call the property `lastQualityLocationInterval` and the struct `LocationInterval`.

MapboxCoreNavigation/RouteController.swift
1ec5: Note that this may impact the following use cases:

* The user has just exited the tunnel and there’s an intersection or exit right after.
* The user has just speed up after a traffic light turns green.

ericrwolfe:
@bsudekum Can we also filter out `speed < 0` and possibly `speed > 0 && course < 0` as part of this PR?

MapboxCoreNavigation/RouteController.swift
ericrwolfe: At worst, we'll be just 1s behind once locations start streaming in, which is better than rerouting and confusing the user.

bsudekum:
@ericrwolfe noting if we do that, we will not hit this if statement https://github.com/mapbox/mapbox-navigation-ios/blob/d261c87b4aea72a368d37f6322a667e4f23e43a9/MapboxCoreNavigation/RouteController.swift#L320-L322


This is pretty useful since we are interpreting the course and location when the speed/course is low/inaccurate.

ericrwolfe:
@bsudekum Good point, let's stick with filtering just `speed < 0` for now then.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -316,6 +333,35 @@ open class RouteController: NSObject {
...
+                
+                lastQualityLocationInterval?.location = location
+            }
+        }
+        
+        guard let lastLocation = lastQualityLocationInterval, let location = rawLocation else {
+                return false
+        }
+        
+        guard location.speed <= lastLocation.speedDelta * 3 else {
```
1ec5: Factor out this factor into a constant.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -300,6 +315,8 @@ open class RouteController: NSObject {
         if location.course <= 0 || location.speed <= RouteControllerMinimumSpeedThresholdForSnappingUserToRoute, snappedCoordinate.distance < RouteControllerUserLocationSnappingDistance {
             let calculatedWrappedCourse = wrap((wrappedPointBehind + wrappedPointAhead) / 2, min: 0 , max: 360)
             return CLLocation(coordinate: snappedCoordinate.coordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: calculatedWrappedCourse, speed: location.speed, timestamp: location.timestamp)
+        } else if isQualityLocation(location: rawLocation) {
```
bsudekum: I think this opens the door for more filtering heuristics.

MapboxCoreNavigation/Constants.swift
```diff
@@ -150,3 +150,8 @@ let milesToMeters = 1609.34
  The mimimum speed value before the user is snapped to the route. This is used to overcome inaccurate course values when a user's speed is low.
  */
 public var RouteControllerMinimumSpeedThresholdForSnappingUserToRoute: CLLocationSpeed = 2
+
+/**
+ Multiplier used to compare the current speed delta from the previous speed delta. When the difference in speed delta is greater than 3 times, the location is thrown out.
+ */
+public var RouteControllerDifferenceInSpeedDeltaMultiplier: CLLocationSpeed = 3
```
1ec5: `DifferenceIn…Delta` is redundant. With the suggestions below, this variable would either be a speed difference of type `CLLocationSpeed` named `RouteControllerMinimumSpeedDelta` or a percentage of type `Double` named `RouteControllerMinimumSpeedDeviation`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -316,6 +333,35 @@ open class RouteController: NSObject {
...
+                
+                lastQualityLocationInterval?.location = location
+            }
+        }
+        
+        guard let lastLocation = lastQualityLocationInterval, let location = rawLocation else {
+                return false
+        }
+        
+        guard location.speed <= lastLocation.speedDelta * RouteControllerDifferenceInSpeedDeltaMultiplier else {
```
1ec5: It’s incorrect to compare a speed (measured in m/s) to a difference in speeds (measured in m/s² assuming location updates every second). We should either evaluate `abs(lastLocation.speed - location.speed)` (speed difference in m/s, or acceleration in m/s²) or compare `abs(lastLocation.speed - location.speed) / lastLocation.speed` (which would be a percentage). Either way, we should get rid of `speedDelta`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -316,6 +333,35 @@ open class RouteController: NSObject {
...
+     
+     We're attempting to throw out coordinates that have a large difference in location or time.
+     */
+    func isQualityLocation(location: CLLocation?) -> Bool {
+        
+        // Any time we return in this func, reset the values.
+        defer {
+            if let location = location {
+                if let lastLocation = lastQualityLocationInterval {
+                    lastQualityLocationInterval?.speedDelta = location.speed - lastLocation.location.speed
```
1ec5: Unless we take the absolute value of this difference somewhere, we’re going to toss out any location update that doesn’t represent a speedup of 3× or more. 💨

ericrwolfe:
Closing in favor of new filtering logic we're experimenting with.

PR #560, opened 2017-08-29
Reroute faster at intersections

bsudekum:
If the users course is within a few degrees of a heading at an intersection along a step, we can assume they are going off course and we can reroute them sooner. 

todo: 
- [x] check performance
- [x] We have to be very sure the user is going down the wrong road at intersection. Maybe we need another check in here to see their distance is increasing away from the intersection.
- [x] actually reroute

This will require a lot of real world testing.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -491,6 +491,7 @@ extension RouteController: CLLocationManagerDelegate {
         }
         
         monitorStepProgress(location)
+        monitorStepIntersectionProgress(location)
         
         // Check for faster route given users current location
         guard checkForFasterRouteInBackground else { return }
@@ -674,6 +675,37 @@ extension RouteController: CLLocationManagerDelegate {
             
             return completion(route, error)
         }
+    }
+    
+    func monitorStepIntersectionProgress(_ location: CLLocation) {
+        let currentStepProgress = routeProgress.currentLegProgress.currentStepProgress
+        guard let intersections = routeProgress.currentLegProgress.currentStepProgress.step.intersections else { return }
+        let interesectionIndex = routeProgress.currentLegProgress.currentStepProgress.intersectionIndex
+        
+        // Don't search for previous intersections
+        for intersection in intersections.suffix(from: interesectionIndex) {
+            let dist = distance(along: currentStepProgress.step.coordinates!, from: location.coordinate, to: intersection.location)
+            
+            // Check if the user is at an intersection
+            if dist < RouteControllerManeuverZoneRadius / 2 {
+                routeProgress.currentLegProgress.currentStepProgress.intersectionIndex += 1
+                guard let newCurrentIntersection = routeProgress.currentLegProgress.currentStepProgress.currentIntersection else { return }
+                
+                // Loop over all possible headings at an intersection.
+                // Check if the user's course is pretty close to an exit heading.
+                for (headingIndex, heading) in newCurrentIntersection.headings.enumerated() {
+                    guard intersection.outletIndex != headingIndex else { continue }
+                    let wrappedCoursae = wrap(location.course, min: 0, max: 360)
+                    let wrappedHeading = wrap(heading, min: 0, max: 360)
+                    if differenceBetweenAngles(wrappedCoursae, wrappedHeading) < RouteControllerMaximumAllowedDegreeOffsetForTurnCompletion {
+                        print("reroute!")
+                    }
+                }
+                return
+            }
+        }
+        
+        print(routeProgress.currentLegProgress.currentStepProgress.intersectionIndex)
     }
     
     func monitorStepProgress(_ location: CLLocation) {
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -491,7 +491,6 @@ extension RouteController: CLLocationManagerDelegate {
         }
         
         monitorStepProgress(location)
-        monitorStepIntersectionProgress(location)
         
         // Check for faster route given users current location
         guard checkForFasterRouteInBackground else { return }
@@ -513,6 +512,9 @@ extension RouteController: CLLocationManagerDelegate {
      If the user is not on the route, they should be rerouted.
      */
     public func userIsOnRoute(_ location: CLLocation) -> Bool {
+        
+        guard !userExitedRouteAtIntersection(location) else { return false }
+        
         // Find future location of user
         let metersInFrontOfUser = location.speed * RouteControllerDeadReckoningTimeInterval
         let locationInfrontOfUser = location.coordinate.coordinate(at: metersInFrontOfUser, facing: location.course)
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -676,6 +679,44 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
     
+    func userExitedRouteAtIntersection(_ location: CLLocation) -> Bool {
+        guard var intersections = routeProgress.currentLegProgress.currentStepProgress.step.intersections else { return false }
+        let currentStepProgress = routeProgress.currentLegProgress.currentStepProgress
+        
+        // The intersections array does not include the upcoming maneuver intersection.
+        if let upcomingStep = routeProgress.currentLegProgress.upComingStep, let upcomingIntersection = upcomingStep.intersections, let firstUpcomginIntersection = upcomingIntersection.first {
+            intersections = intersections + [firstUpcomginIntersection]
```
1ec5: Nit: `+=`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -676,6 +679,44 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
     
+    func userExitedRouteAtIntersection(_ location: CLLocation) -> Bool {
+        guard var intersections = routeProgress.currentLegProgress.currentStepProgress.step.intersections else { return false }
+        let currentStepProgress = routeProgress.currentLegProgress.currentStepProgress
+        
+        // The intersections array does not include the upcoming maneuver intersection.
+        if let upcomingStep = routeProgress.currentLegProgress.upComingStep, let upcomingIntersection = upcomingStep.intersections, let firstUpcomginIntersection = upcomingIntersection.first {
```
1ec5: Typo: `Upcomgin`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -676,6 +679,44 @@ extension RouteController: CLLocationManagerDelegate {
...
+        routeProgress.currentLegProgress.currentStepProgress.intersectionsIncludingUpcomingManeuverIntersection = intersections
+        
+        // Don't search for previous intersections
+        for intersection in intersections.suffix(from: currentStepProgress.intersectionIndex) {
+            let dist = distance(along: currentStepProgress.step.coordinates!, from: location.coordinate, to: intersection.location)
+            // Check if the user is at an intersection
+            if dist <= RouteControllerManeuverZoneRadius / 2 {
+                guard let newCurrentIntersection = routeProgress.currentLegProgress.currentStepProgress.currentIntersection else { return false }
+                
+                // Loop over all possible headings at an intersection.
```
1ec5: Rather than looking at specific outlet headings, it would be simpler to decrease the rerouting radius thresholds when the user is near an intersection. Would that yield acceptable results?

I’m just concerned that we’re relying too heavily on outlet headings, but road headings are less reliable near large intersections in general, due to how they’re modeled in OpenStreetMap. But as long as we’re handling intersections like the following examples well, then this approach could be fine:

1. [Road divides at intersection](https://www.mapbox.com/get-directions/#19/37.35013/-121.95944?coordinates=-121.959435,37.349865;-121.95936,37.350293)
2. [Road curves sharply at intersection](https://www.mapbox.com/get-directions/#18.51/39.3361/-84.27812?coordinates=-84.27780130105117,39.33588655554837;-84.27811584679313,39.33609560406964)
3. [Road curves gently at intersection](https://www.mapbox.com/get-directions/#17.18/39.10636/-84.5548?coordinates=-84.554801,39.106355;-84.554361,39.106354)

In the third example, the heading for Wilder Avenue is reported as 90°. But if the user happens to travel at such a speed that a location update comes in while the user’s course is still 135° or already 45°, then we’ll incorrectly reroute, right? Or is something else preventing that from happening?

bsudekum:
@1ec5 I updated this with your feedback in https://github.com/mapbox/mapbox-navigation-ios/pull/560#discussion_r136382307. This PR initially started with counting intersections but with https://github.com/mapbox/mapbox-navigation-ios/pull/560#discussion_r136382307, it's not totally necessary. However, the counting logic will be useful later. Do you think I should keep them in this PR?

bsudekum:
We won't be able to do things like `Take your next right` until we have support for statements like this in osrm-text-instructions.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -512,12 +521,12 @@ extension RouteController: CLLocationManagerDelegate {
...
      */
     public func userIsOnRoute(_ location: CLLocation) -> Bool {
+        
         // Find future location of user
         let metersInFrontOfUser = location.speed * RouteControllerDeadReckoningTimeInterval
         let locationInfrontOfUser = location.coordinate.coordinate(at: metersInFrontOfUser, facing: location.course)
         let newLocation = CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
-        let radius = max(RouteControllerMaximumDistanceBeforeRecalculating,
-                         location.horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
+        let radius = max(modifiedRerouteDistanceGivenIntersectionLocation, location.horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
```
bsudekum: We might just want to make this `modifiedRerouteDistanceGivenIntersectionLocation` and not incorporate accuracy.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -262,6 +262,14 @@ open class RouteController: NSObject {
      */
     var rawLocation: CLLocation?
     
+    public var modifiedRerouteDistanceGivenIntersectionLocation: CLLocationDistance {
```
1ec5: Let’s call the concept of “how far can I go before triggering rerouting” a “rerouting tolerance”. So this property could be named simply `reroutingTolerance`. (The fact that it accounts for the distance from an intersection is an implementation detail.)

MapboxCoreNavigation/RouteController.swift
```diff
@@ -676,6 +685,30 @@ extension RouteController: CLLocationManagerDelegate {
...
+        }
+        
+        routeProgress.currentLegProgress.currentStepProgress.intersectionsIncludingUpcomingManeuverIntersection = intersections
+        
+        // Don't search for previous intersections
+        for intersection in intersections.suffix(from: currentStepProgress.intersectionIndex) {
+            let dist = distance(along: currentStepProgress.step.coordinates!, from: location.coordinate, to: intersection.location)
+            // Check if the user is at an intersection
+            if dist <= RouteControllerManeuverZoneRadius {
+                routeProgress.currentLegProgress.currentStepProgress.userIsNearAnIntersection = true
```
1ec5: It should be the job of RouteStepProgress to track how far the user is from an intersection, not a judgment about whether they’re close. It would be more flexible to set `distanceToIntersection` here – or maybe even make that a computed variable on RouteStepProgress – and decide whether it counts as “near” up in `reroutingTolerance`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -676,6 +685,30 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
     
+    func monitorIntersectionProgress(_ location: CLLocation) {
```
1ec5: “Monitor” isn’t a good word to use for this method (or `monitorStepProgress(_:)` for that matter), since it doesn’t set up its own location manager or anything. Additionally, it updates the step progress just like `monitorStepProgress(_:)` does; there isn’t a separate IntersectionProgress class to update.

How about `updateDistanceToIntersection(from:)`?

1ec5:
> We won't be able to do things like `Take your next right` until we have support for statements like this in osrm-text-instructions.

Yeah, I’m just thinking we might want to upstream this logic to OSRM Text Instructions at some point or have it accept an option that says how many intersections remain.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -262,6 +262,20 @@ open class RouteController: NSObject {
...
     var rawLocation: CLLocation?
     
+    public var reroutingTolerance: CLLocationDistance {
+        guard let intersections = routeProgress.currentLegProgress.currentStepProgress.intersectionsIncludingUpcomingManeuverIntersection else { return RouteControllerMaximumDistanceBeforeRecalculating }
+        guard let userLocation = rawLocation else { return RouteControllerMaximumDistanceBeforeRecalculating }
+        
+        for intersection in intersections {
+            let distanceToIntersection = distance(along: routeProgress.currentLegProgress.currentStepProgress.step.coordinates!, from: userLocation.coordinate, to: intersection.location)
+            
+            if distanceToIntersection <= RouteControllerManeuverZoneRadius {
```
bsudekum: @1ec5 I moved away from keeping count of the current intersection. This is because this line of code is hit right when the user is reaching the intersection. If we increment the intersection index here, we'd already be looking at the next intersection while the user is still in the current intersection.

We could alternatively check the users course compared to the expected outlet heading, but then again, straight intersections will be difficult.

Noting this for loop should remain light as there are never really that many intersections on a step. _Waits patient to be shown wrong._

1ec5: [108 intersections](https://www.mapbox.com/get-directions/#16.18/39.73485/-105.1595?coordinates=-105.159523,39.734874;-105.022986,39.740455) over a 7½-mile stretch of [Colfax Avenue](https://en.wikipedia.org/wiki/Colfax_Avenue) in Denver. Examples like these would be more common once we implement toll and freeway avoidance.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -262,6 +262,20 @@ open class RouteController: NSObject {
      */
     var rawLocation: CLLocation?
     
+    public var reroutingTolerance: CLLocationDistance {
+        guard let intersections = routeProgress.currentLegProgress.currentStepProgress.intersectionsIncludingUpcomingManeuverIntersection else { return RouteControllerMaximumDistanceBeforeRecalculating }
+        guard let userLocation = rawLocation else { return RouteControllerMaximumDistanceBeforeRecalculating }
+        
+        for intersection in intersections {
+            let distanceToIntersection = distance(along: routeProgress.currentLegProgress.currentStepProgress.step.coordinates!, from: userLocation.coordinate, to: intersection.location)
```
1ec5: If we’re concerned about performance, I think we can use as-the-crow-flies distance instead of the distance along the route line, since anything significantly far from the route line would result in rerouting anyways. Sounds like circular logic, though – needs verification.

ericrwolfe: I was thinking similarly.`distance(along:)` sounds overly expensive here.

bsudekum: As the crow flies sounds good to me.

PR #561, opened 2017-08-29
Do not overload user with info at roundabouts

bsudekum:
Roundabouts will soon have an exit maneuver. In the past, we have tried to make up for the fact that it is not there by over loading the enter roundabout maneuver with as much info as we could grab onto.

Now, when the user is entering the roundabout, we don't need to link the instruction if the instruction contains exit information. 

Before:

```
Enter the roundabout and take the 4th exit onto Foo St and then turn right onto Bar St.
```

Now
```
Enter the roundabout and take the 4th exit onto Foo St.
```

Note: when the new maneuver type is added, this PR needs to account for it.

/cc @miccolis @danpat @1ec5


ericrwolfe:
cc @cammace @danesfeder

MapboxCoreNavigation/SpokenInstructionFormatter.swift
```diff
@@ -76,7 +76,12 @@ public class SpokenInstructionFormatter: NSObject {
                 text = String.localizedStringWithFormat(NSLocalizedString("CONTINUE", bundle: .mapboxCoreNavigation, value: "Continue for %@", comment: "Format for speech string after completing a maneuver and starting a new step; 1 = distance"), escapeIfNecessary(maneuverVoiceDistanceFormatter.string(from: userDistance)))
             }
         } else if alertLevel == .high && upcomingStepDuration < linkedInstructionMultiplier {
-            text = String.localizedStringWithFormat(NSLocalizedString("LINKED_UTTERANCE_FORMAT", bundle: .mapboxCoreNavigation, value: "%@, then %@", comment: "Format for speech string; 1 = current instruction; 2 = the following linked instruction"), upComingInstruction, followOnInstruction)
+            // If the user is entering the roundabout and there is exit information, don't link instruction
+            if let upcomingStep = routeProgress.currentLegProgress.upComingStep, (upcomingStep.maneuverType == .takeRotary || upcomingStep.maneuverType == .takeRoundabout) && upcomingStep.exitIndex != nil {
```
1ec5: Can we simplify this logic by waiting until Project-OSRM/osrm-backend#4358 is deployed and removing this special case altogether?

bsudekum: @1ec5 yep, that's the idea. This PR would be a regression until then. We'll also need to update osrm-text-instructions to account for the new type.

1ec5: I think this code would read more naturally as:

```swift
if followOnInstruction.maneuverType == .exitRoundabout || followOnInstruction.maneuverType == .exitRotary`
```

That way, if the developer fails to specify `includesRoundaboutExitManeuver` in the original route options and there’s a turn close to the exit, that turn won’t be omitted entirely.

bsudekum:
@1ec5 this is mergeable now that `roundabout_exits` is live.

MapboxCoreNavigation/SpokenInstructionFormatter.swift
```diff
@@ -76,7 +76,12 @@ public class SpokenInstructionFormatter: NSObject {
...
             }
         } else if alertLevel == .high && upcomingStepDuration < linkedInstructionMultiplier {
-            text = String.localizedStringWithFormat(NSLocalizedString("LINKED_UTTERANCE_FORMAT", bundle: .mapboxCoreNavigation, value: "%@, then %@", comment: "Format for speech string; 1 = current instruction; 2 = the following linked instruction"), upComingInstruction, followOnInstruction)
+            // If the upcoming step is an .exitRoundabout or .exitRotary, don't link the instruction
+            if let upcomingStep = routeProgress.currentLegProgress.upComingStep, upcomingStep.maneuverType == .exitRoundabout || upcomingStep.maneuverType == .exitRotary {
+                text = upComingInstruction
+            } else {
+                text = String.localizedStringWithFormat(NSLocalizedString("LINKED_UTTERANCE_FORMAT", bundle: .mapboxCoreNavigation, value: "%@,     %@", comment: "Format for speech string; 1 = current instruction; 2 = the following linked instruction"), upComingInstruction, followOnInstruction)
```
1ec5: The `value` changed from `%@, then %@` to `%@,     %@`. Is that intentional?

MapboxCoreNavigation/SpokenInstructionFormatter.swift
```diff
@@ -76,7 +76,12 @@ public class SpokenInstructionFormatter: NSObject {
                 text = String.localizedStringWithFormat(NSLocalizedString("CONTINUE", bundle: .mapboxCoreNavigation, value: "Continue for %@", comment: "Format for speech string after completing a maneuver and starting a new step; 1 = distance"), escapeIfNecessary(maneuverVoiceDistanceFormatter.string(from: userDistance)))
             }
         } else if alertLevel == .high && upcomingStepDuration < linkedInstructionMultiplier {
-            text = String.localizedStringWithFormat(NSLocalizedString("LINKED_UTTERANCE_FORMAT", bundle: .mapboxCoreNavigation, value: "%@, then %@", comment: "Format for speech string; 1 = current instruction; 2 = the following linked instruction"), upComingInstruction, followOnInstruction)
+            // If the upcoming step is an .exitRoundabout or .exitRotary, don't link the instruction
+            if let upcomingStep = routeProgress.currentLegProgress.upComingStep, upcomingStep.maneuverType == .exitRoundabout || upcomingStep.maneuverType == .exitRotary {
```
1ec5: Why `upComingStep`? Wouldn’t the `upComingStep` be `takeRoundabout` and `followOnStep` be `exitRoundabout`?

1ec5: * If `upComingStep` is `takeRoundabout` and `followOnStep` is `exitRoundabout`, we announce `takeRoundabout` then `exitRoundabout`. 🙅‍♂️
* If `upComingStep` is `exitRoundabout` and `followOnStep` is `turn`, we announce just the `exitRoundabout` and never announce the `turn`. 🙅‍♂️

bsudekum:
@1ec5 good call, fixed to look at the `followOnStep`.

PR #563, opened 2017-08-29
Add stylable marker view

frederoni:
Added a stylable marker view. This view is not yet being used.

screenshot 2017-08-29 19 22 20

@bsudekum @mayagao 👀


bsudekum:
@1ec5 how do you feel about this? The goal here is to move more into the `Style` class.

mayagao:
screen shot 2017-08-30 at 12 56 30 pm

@frederoni could we add a little white circle in the middle?

frederoni:
@1ec5 @bsudekum 
If you want animations or a completely different marker view, you'd still want to implement `viewFor:`. This is just a convenient way to change the tint color, stroke color and shadow of the default marker.

frederoni:
@mayagao 

Here's the most recent with the inner circle:

screenshot 2017-08-30 13 47 10

1ec5:
@frederoni, would you mind upstreaming this work to the map SDK? mapbox/mapbox-gl-native#2250

PR #572, opened 2017-09-01
Update styleURL if needed

frederoni:
Fixes the issue we've been seeing where the night style is using the day map style.

@ericrwolfe 👀

### DIFF
--- a/MapboxNavigation/DayStyle.swift
+++ b/MapboxNavigation/DayStyle.swift
@@ -121,6 +121,8 @@ open class DayStyle: Style {
         NavigationMapView.appearance().trafficModerateColor     = .trafficModerate
         NavigationMapView.appearance().trafficHeavyColor        = .trafficHeavy
         NavigationMapView.appearance().trafficSevereColor       = .trafficSevere
+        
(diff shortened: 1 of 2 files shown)


MapboxNavigation/DayStyle.swift
```diff
@@ -121,6 +121,8 @@ open class DayStyle: Style {
         NavigationMapView.appearance().trafficModerateColor     = .trafficModerate
         NavigationMapView.appearance().trafficHeavyColor        = .trafficHeavy
         NavigationMapView.appearance().trafficSevereColor       = .trafficSevere
+        
+        UIApplication.shared.statusBarStyle = statusBarStyle ?? .default
```
bsudekum: What about this property https://github.com/mapbox/mapbox-navigation-ios/blob/51470ad955bd4b41cfe9adfd6451911576a0b5f5/MapboxNavigation/DayStyle.swift#L48 ?

frederoni: Is there anything missing here? We must store the property but wait with applying because of the day/night switch.

PR #573, opened 2017-09-01
Use correct linking multiplier

bsudekum:
I mixed up the constants name in a big way... 🤦‍♂️ 

We are were linking instructions when the next step had a duration of less than 225 seconds, yikes.

/cc @frederoni @ericrwolfe @1ec5


1ec5:
This regression was introduced in #448.

Issue #582, opened 2017-09-05
fatal error: unexpectedly found nil while unwrapping an Optional value in RouteMapViewController.swift


TCOA:
just loaded 0.7.0 and upon running, I get the 'nil' error on this line....
`if notification.userInfo![RouteControllerDidFindFasterRouteKey] as! Bool {`

wrapping in a "!= nil" bypasses the error...

` if notification.userInfo![RouteControllerDidFindFasterRouteKey] != nil {`

though the route is not tracking the Simulator route (not sure if that is related, though still testing...)

bsudekum:
I've actually seen this a few times. It looks like we're sending a reroute somewhere without all the proper user info keys. @TCOA are there specific situations you see this error?

TCOA:
I'm currently presenting the map viewController after receiving a push notification.  Further testing (after I removed the wrap) shows it doesn't crash every time (right now I am also running the example test on the same phone on a long simulation drive, so the phone is pretty busy...).

I'm hoping to find a way to 'add on' the bits to the map instead of presenting a new controller (what I'm doing now is showing the map at first, then after 'conditions' [and a push notification], jump into the turn-by-turn.   Once I figure out how to add/remove just those pieces {and I think the new Custom example will get me there}, I think that will be a much better way)

PR #585, opened 2017-09-05
Dont show reroute UI when failing to reroute

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/582

When failing to reroute, we were showing the reroute UI. Since this notification user object is different from the reroute key, we'd get a crash. This was introduced in https://github.com/mapbox/mapbox-navigation-ios/commit/7334aa565a22b8ab62e027943db529dc019544d2

If we'd like to show something in the UI when failing to reroute, `RouteControllerDidFailToReroute` should call it's own unique function.

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -128,13 +128,11 @@ class RouteMapViewController: UIViewController {
     func resumeNotifications() {
         NotificationCenter.default.addObserver(self, selector: #selector(willReroute(notification:)), name: RouteControllerWillReroute, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(didReroute(notification:)), name: RouteControllerDidReroute, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(didReroute(notification:)), name: RouteControllerDidFailToReroute, object: nil)
     }
     
     func suspendNotifications() {
         NotificationCenter.default.removeObserver(self, name: RouteControllerWillReroute, object: nil)
         NotificationCenter.default.removeObserver(self, name: RouteControllerDidReroute, object: nil)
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -128,13 +128,11 @@ class RouteMapViewController: UIViewController {
     func resumeNotifications() {
         NotificationCenter.default.addObserver(self, selector: #selector(willReroute(notification:)), name: RouteControllerWillReroute, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(didReroute(notification:)), name: RouteControllerDidReroute, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(didReroute(notification:)), name: RouteControllerDidFailToReroute, object: nil)
```
1ec5: `didReroute(_:)` currently calls `statusView.hide(delay:animated:)`, balancing out the call to `statusView.show(_:showSpinner:)` in `willReroute(_:)`. Does this mean the status view will show “Rerouting…” indefinitely?

bsudekum: No, we hide it here: https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/RouteMapViewController.swift#L257

Now we're silently doing nothing when the reroute fails. Even in a case where we do fail to reroute, `RouteControllerDidReroute`  will always be called.

bsudekum: Also, tested.

1ec5: But doesn’t this change remove the call to `didReroute(notification:)` that you linked to, or does that method still get called some other way?

1ec5: If rerouting failed, why does `RouteControllerDidReroute` get posted?

PR #591, opened 2017-09-07
Update cancel button

frederoni:
The hitbox now also covers the entire space.

@bsudekum @ericrwolfe 

screenshot 2017-09-07 17 14 13
screenshot 2017-09-07 17 14 25


bsudekum:
Can any translations be removed?

bsudekum:
Do you think it makes sense to keep the tint in line with rest of the button tints? Looking specifically at the overview, feedback and mute buttons.

image

frederoni:
@bsudekum the tint color is usually don't applied to destructive or cancellation buttons.
I replaced the icon with a slightly more subtle one.




Also fixed the localization extraction script to include the storyboard and discard all "DO NOT TRANSLATE" strings.
/cc @1ec5

PR #593, opened 2017-09-07
Update MapboxDirections.swift, add routeIdentifier

bsudekum:
Adds `routeIdentifier` from https://github.com/mapbox/MapboxDirections.swift/pull/165


MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -49,8 +49,8 @@ struct EventDetails {
         simulation = routeController.locationManager is ReplayLocationManager || routeController.locationManager is SimulatedLocationManager ? true : false
         
         sessionIdentifier = session.identifier.uuidString
-        originalRequestIdentifier = nil
-        requestIdentifier = nil
+        originalRequestIdentifier = session.originalRoute.routeIdentifier
+        requestIdentifier = routeController.routeProgress.route.routeIdentifier
```
ericrwolfe: Looks like we can either use the SessionState.routeIdentifiers here, or remove the properties from SessionState altogether.

bsudekum: 👍, removing the props `originalRequestIdentifier` and `requestIdentifier` completely.

Issue #594, opened 2017-09-08
Refresh ETA


AFcgi:
It's possible to update the ETA? If I stay a the starting point and do nothing, the ETA isn't updated. This happens only when the route progress is changed.

bsudekum:
@AFcgi valid point. I think it should be possible to instead update the bottom bar info once ever ~60s instead of relying on progress updates.

PR #595, opened 2017-09-08
Update ETA info when stale

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/594

When the user is not moving, we should still be able to give them accurate info about their trip.

This also removes a few functions that existed just to call another function.

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -391,6 +391,7 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         super.viewWillDisappear(animated)
         
         UIApplication.shared.isIdleTimerDisabled = false
+        updateETATimer = nil
         routeController.suspendLocationUpdates()
     }
     
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -290,7 +290,7 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         super.init(coder: aDecoder)
     }
     
-    var updateETAInfoTimer: Timer!
+    var updateETATimer: Timer?
     
     required public init(contentViewController: UIViewController, drawerViewController: UIViewController) {
         fatalError("init(contentViewController:drawerViewController:) has not been implemented. " +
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -450,6 +455,14 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         }
     }
     
+    func ETAIsStale() {
```
1ec5: This sounds like the name of a property that returns a value without side-effects, which isn’t the case. How about `updateETA()`?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -290,6 +290,8 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         super.init(coder: aDecoder)
     }
     
+    var updateETAInfoTimer: Timer!
```
1ec5: Why “info”? `updateETATimer` is fine.

1ec5: Can this be an optional instead of an implicitly unwrapped optional? That way the code is explicit about the case where it isn’t set (before `viewWillAppear(_:)`).

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -84,15 +84,7 @@ class RouteTableViewController: UIViewController {
             }
         }
     }
-    
-    func notifyDidChange(routeProgress: RouteProgress) {
-        showETA(routeProgress: routeProgress)
-    }
-    
-    func notifyDidReroute() {
-        tableView.reloadData()
```
ericrwolfe: Don't we still need to refresh this after a reroute?

bsudekum: Good catch

bsudekum:
@1ec5 @ericrwolfe fixed.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -455,12 +455,12 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
...
     }
     
-    func ETAIsStale() {
+    func updateETA() {
         tableViewController?.updateETA(routeProgress: routeController.routeProgress)
     }
     
     func resetETATimer() {
-       updateETAInfoTimer = Timer.scheduledTimer(timeInterval: 30, target: self, selector: #selector(ETAIsStale), userInfo: nil, repeats: true)
+       updateETATimer = Timer.scheduledTimer(timeInterval: 30, target: self, selector: #selector(updateETA), userInfo: nil, repeats: true)
```
1ec5: This timer will leak the whole class unless we invalidate it in `deinit` or `viewDidUnload()`.

bsudekum:
Fixed

MapboxNavigation/NavigationViewController.swift
```diff
@@ -391,6 +391,7 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
         super.viewWillDisappear(animated)
         
         UIApplication.shared.isIdleTimerDisabled = false
+        updateETATimer = nil
```
1ec5: `invalidate()` the timer before nilling it out.

bsudekum:
@1ec5 fixed.

PR #602, opened 2017-09-11
Add slight right/left maneuver icon

bsudekum:
Updating slight right/left lane in a separate PR.

simulator screen shot - iphone 7 - 2017-09-11 at 15 43 53

/cc @1ec5 @frederoni @ericrwolfe @willwhite


1ec5:
The underlying design assets were updated in mapbox/navigation-ui-resources#1.

Issue #600, opened 2017-09-11
Leg summary in step table can be ungrammatical in Spanish


1ec5:
[`LEG_MAJOR_WAYS_FORMAT`](https://github.com/mapbox/mapbox-navigation-ios/blob/5ac9b3b72d176cd21c74c30832dcbfecc7477cf8/MapboxNavigation/Resources/Base.lproj/Localizable.strings#L23) and [`WAYPOINT_SOURCE_DESTINATION_FORMAT`](https://github.com/mapbox/mapbox-navigation-ios/blob/5ac9b3b72d176cd21c74c30832dcbfecc7477cf8/MapboxNavigation/Resources/Base.lproj/Localizable.strings#L41) determine the format of the summary at the top of a leg in the step table when there are multiple legs. Each contains two names joined by the conjunction “and”. In Spanish, the equivalent conjunction is usually “y” but can change to “e” if the following word begins with an “i” or “y”. The navigation SDK nonetheless uses “y” unconditionally.

This problem affects Spanish localizations of all kinds of software (for example, citation-style-language/locales#87), but it’s especially acute for this particular string because so many street names and place names can begin with “I” or “Y”.

We’d need to add a special case in the code, because neither the .strings format nor Transifex supports varying conjunctions this way.

/cc @frederoni

ericrwolfe:
This is a super, super edge case all things considered and would be an extreme over-optimization on the client given other priorities at the moment. 

* Multiple waypoints (edge case)
* Step table view (less commonly used view)
* Multiple way names in summary (edge case)
* Grammatical edge case in Spanish (edge case, relatively speaking)

If this needs to be fixed at all, it should be fixed on the in the directions API.

PR #603, opened 2017-09-12
Localize conjunction

frederoni:
Work in progress for #600 

This PR adds a special case for localizing a conjunction of two words or phrases where the conjunction word depends on the following letter _or_ pronunciation.

In addition to replacing "y" with "e" if the following letter is "y" or "i", we should also use this rule if the following pronunciation is /EYE/. e.g. "Necesito aguja e hilo" ← silent H, while "La mesa es de madera y hierro" uses "y". (Thanks @Guardiola31337)

I added a new string to handle the new rule instead of checking if the language is set to Spanish. A downside of this implementation is that we need to localize the conjunction format for both rules for all languages.

@1ec5 @bsudekum


1ec5:
> I added a new string to handle the new rule instead of checking if the language is set to Spanish. A downside of this implementation is that we need to localize the conjunction format for both rules for all languages.

I think it would actually be preferable to special-case Spanish in the code than to force translators to account for this Spanishism in every language. Unfortunately, Transifex lacks the concept of an optionally translated string.

ericrwolfe:
Closing as premature optimization.

PR #604, opened 2017-09-12
Always map first location update

bsudekum:
If the first location update is not `isQualified` and the user is not moving, the user puck and other information is never shown. This makes sure we update the UI and user puck location at least once if it's the first location update.

/cc @1ec5 @ericrwolfe


ericrwolfe:
New changes look good

PR #607, opened 2017-09-13
Fix congestion segment and guard

bsudekum:
If you add up all the step coordinates, it will be equal to the route geometry count plus the number of steps. This is because each step shares a coordinate with the next step. Previously, our lookups into this array were a little off since we were not removing the last coordinate which is shared. 

This also adds a guard to check that our array lookup is also within bounds.

/cc @1ec5

### DIFF
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -157,8 +157,14 @@ open class RouteProgress: NSObject {
             
             if let segmentCongestionLevels = leg.segmentCongestionLevels, let expectedSegmentTravelTimes = leg.expectedSegmentTravelTimes  {
                 
-                for step in leg.steps.dropFirst() {
-                    let stepCoordinatesCount = maneuverCoordinateIndex + Int(step.coordinateCount) - 1
+                for step in leg.steps {
+                    guard let coordinates = step.coordinates else { continue }
+                    let stepCoordinates = step.maneuverType == .arrive ? Int(step.coordinateCount) : coordinates.dropLast().count
+                    let stepCoordinatesCount = maneuverCoordinateIndex + stepCoordinates - 1
+                    
+                    guard stepCoordinatesCount < segmentCongestionLevels.count else { continue }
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -157,8 +157,14 @@ open class RouteProgress: NSObject {
             
             if let segmentCongestionLevels = leg.segmentCongestionLevels, let expectedSegmentTravelTimes = leg.expectedSegmentTravelTimes  {
                 
-                for step in leg.steps.dropFirst() {
-                    let stepCoordinatesCount = maneuverCoordinateIndex + Int(step.coordinateCount) - 1
+                for step in leg.steps {
+                    guard let coordinates = step.coordinates else { continue }
+                    let stepCoordinates = step.maneuverType == .arrive ? Int(step.coordinateCount) : coordinates.dropLast().count
```
bsudekum: Perhaps it might be smarter to only remove the last if the count is equal to two?

1ec5: This variable is confusingly named: it isn’t the coordinates themselves but rather the count of coordinates. A better name would be `stepCoordinateCount`.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -157,8 +157,14 @@ open class RouteProgress: NSObject {
...
+                    let stepCoordinatesCount = maneuverCoordinateIndex + stepCoordinates - 1
```
1ec5: This isn’t the count of coordinates but rather an index into `segmentCongestionLevels` that corresponds to the last coordinate of this step and the first coordinate (and maneuver) of the following step. A better name would be `nextManeuverCoordinateIndex`.

MapboxCoreNavigation/RouteProgress.swift
1ec5: Conceptually, we want to remove the last coordinate if there’s a following step. Since we’re in a for-in loop, there isn’t a convenient way to do that other than to check that the maneuver type is `arrive`. So I think this is fine. It’s worth checking that multi-leg routes still work as expected, though.

PR #612, opened 2017-09-14
Use Alex voice as fallback voice when available

1ec5:
Prefer the Alex voice over Samantha when Polly is unavailable and the preferred locale is U.S. English. Alex is higher-quality than Samantha, but you have to [download and install](https://support.apple.com/en-us/HT203077) it manually.

/cc @bsudekum


bsudekum:
If it's not downloaded, will it fall back to Samantha?

1ec5:
Yes. It’ll also fall back to Samantha if the locale isn’t `en-US`, since that’s the only locale supported by Alex.

Issue #606, opened 2017-09-13
Confusion about yards vs feet


jeeftor:
I'm a little confused on how to use the distance formatters correctly.  I've tried to implement my own Navigation controller using this repo.

I've defined formatters as such:

```swift
fileprivate let voiceInstructionFormatter = SpokenInstructionFormatter()
fileprivate let textInstructionFormatter = VisualInstructionFormatter()
fileprivate let textDistanceFormatter = DistanceFormatter(approximate: true)
fileprivate let voiceDistanceFormatter = SpokenDistanceFormatter(approximate: true)
```

and set them up this way:

```swift
textDistanceFormatter.numberFormatter.maximumFractionDigits = 0
textDistanceFormatter.numberFormatter.locale = Locale(identifier: "en-US")
voiceDistanceFormatter.numberFormatter.locale = Locale(identifier: "en-US")        
```

My voice formatter is giving me things in `feet` while the text one is giving me `yards`.  Am i missing a setting somewhere?

frederoni:
My assumption is that you're using one of NSLengthFormatter’s methods and therefore bypassing the overriding of yards.

```swift
DistanceFormatter.string(from:) <- should work
DistanceFormatter.string(fromValue:unit:) <- won't work
DistanceFormatter.string(fromMeters:) <- won't work
```

Can you verify if that's the case?

jeeftor:
THANKS YOU!

jeeftor:
I was using `fromMeters:` All is good in the world again!!!

PR #613, opened 2017-09-14
Override string fromMeters

frederoni:
This PR avoids confusion like #606 by linking `string(from:)` and `string(fromMeters:)`.

`DistanceFormatter.string(fromMeters:)` previously used `NSLengthFormatter`’s default which could result in yards. `DistanceFormatter` is a replacement for `MKDistanceFormatter` which leverage some of `NSLengthFormatter`’s functionality but there's no reason why `string(fromMeters:)` shouldn't work the same way as `string(from:)`.

@ericrwolfe @bsudekum


bsudekum:
also /cc @1ec5

PR #614, opened 2017-09-14
Use voice instructions from server

bsudekum:
_Need https://github.com/mapbox/MapboxDirections.swift/pull/175 merged first_

Todo:
- [x] Better handle steps with only less than 3 announcements (right now, a high alert assumes there are 3 announcements)
- [x] Figure out the future of `Alerts`
- [x] Test, figure out where there are gaps between server and local instruction creation.

/cc @mapbox/guidance @1ec5 @frederoni @ericrwolfe


1ec5:
> Figure out the future of Alerts

🚮

bsudekum:
@1ec5 should we just do an index?

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -42,7 +42,8 @@ public class PollyVoiceController: RouteVoiceController {
         
         let routeProgresss = notification.userInfo![RouteControllerAlertLevelDidChangeNotificationRouteProgressKey] as! RouteProgress
         let userDistances = notification.userInfo![RouteControllerAlertLevelDidChangeNotificationDistanceToEndOfManeuverKey] as! CLLocationDistance
-        let instruction = spokenInstructionFormatter.string(routeProgress: routeProgresss, userDistance: userDistances, markUpWithSSML: true)
+        let voiceInstructionIndex = routeProgresss.currentLegProgress.voiceInstructionIndex
+        let instruction = routeProgresss.currentLegProgress.currentStep.instructionsSpokenAlongStep?[voiceInstructionIndex].ssml ?? spokenInstructionFormatter.string(routeProgress: routeProgresss, userDistance: userDistances, markUpWithSSML: true)
```
1ec5: The plain text instruction is still needed for the AVSpeechSynthesizer fallback. I think it would be reasonable to pass the whole VoiceInstruction object into this method.

1ec5: Is `userDistance` still needed?

bsudekum: I'd like to keep `spokenInstructionFormatter.string` around for testing and comparison, so yes still needed.

1ec5:
> should we just do an index?

I guess it depends what our customization story will be. In the future, how will custom milestones interact with the preset ones provided by the API? Would the API be able to add more or fewer than three announcements for a particular step if necessary?

As I see it, the main use case for exposing alert levels publicly was to allow developers to time UI changes around the announcements. We can meet that need by calling a delegate method whenever an announcement will be spoken or has been spoken; we’d need to pass in some way of indicating which announcement was spoken, and that information should be pretty closely tied to the API’s way of referring to the announcements. The API doesn’t currently expose any reliable way to determine the urgency of an announcement, other than its position in the array.

Other current uses of alerts should transition over to the much more flexible RouteProgress class.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -96,7 +95,8 @@ public class PollyVoiceController: RouteVoiceController {
             input.voiceId = voiceId
         }
         
-        input.text = "<speak><prosody volume='\(instructionVoiceVolume)' rate='\(instructionVoiceSpeedRate)'>\(text)</prosody></speak>"
+        // input.text = "<speak><prosody volume='\(instructionVoiceVolume)' rate='\(instructionVoiceSpeedRate)'>\()</prosody></speak>"
```
1ec5: This will render the volume and speaking rate options nonfunctional. What should we do about those options until API support for them lands?

ksummerill:
cc @ksummerill to keep 👀  on this for a customer

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -366,12 +326,6 @@ open class RouteStepProgress: NSObject {
      */
     public var distanceTraveled: CLLocationDistance = 0
     
-    
-    /**
-     Returns distance from user to end of step.
-     */
-    public var userDistanceToManeuverLocation: CLLocationDistance? = nil
```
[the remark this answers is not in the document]
bsudekum: Let's cut it if it's not used.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -230,7 +230,7 @@ open class RouteController: NSObject {
     
     func resumeNotifications() {
         NotificationCenter.default.addObserver(self, selector: #selector(progressDidChange(notification:)), name: RouteControllerProgressDidChange, object: self)
```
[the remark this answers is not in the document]
bsudekum: Let's fix this in it's own PR.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -713,15 +675,14 @@ extension RouteController: CLLocationManagerDelegate {
...
-        var updateStepIndex = false
         
         let userSnapToStepDistanceFromManeuver = distance(along: routeProgress.currentLegProgress.currentStep.coordinates!, from: location.coordinate)
-        let secondsToEndOfStep = routeProgress.currentLegProgress.currentStepProgress.durationRemaining
         var courseMatchesManeuverFinalHeading = false
-
-        let outletRoadClasses = routeProgress.currentLegProgress.currentStepProgress.step.intersections?.first?.outletRoadClasses
+        
+        guard let voiceInstructions = routeProgress.currentLegProgress.currentStep.instructionsSpokenAlongStep else {
+            assert(false, "Voice instructions must be enabled")
```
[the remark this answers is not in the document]
1ec5: > We don’t currently document that NavigationRouteOptions or `includesVoiceInstructions` must be used with this library.

Tracking a documentation update in #630. Unless we remove this requirement, I’d consider #630 to block this PR.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -96,7 +95,11 @@ public class PollyVoiceController: RouteVoiceController {
             input.voiceId = voiceId
         }
         
-        input.text = "<speak><prosody volume='\(instructionVoiceVolume)' rate='\(instructionVoiceSpeedRate)'>\(text)</prosody></speak>"
+        let wrappedSSML = text
+            .replacingOccurrences(of: "<speak>", with: "<speak><prosody volume='\(instructionVoiceVolume)' rate='\(instructionVoiceSpeedRate)'>")
+            .replacingOccurrences(of: "</speak>", with: "</prosody></speak>")
```
1ec5: With my security hat on, this seems like it would enable someone to break the navigation SDK simply by renaming a road to something like `<speak>` or `</speak>`. Using a proper XML parser would be best, although if we’re absolutely certain that the Directions API will be giving us a document with `<speak>` as the root element, then we could address the security concern by restricting finding-and-replacing only at the beginning and end of the string.

MapboxCoreNavigation/Constants.swift
```diff
@@ -55,7 +50,7 @@ public let RouteControllerProgressDidChange = Notification.Name(MBRouteControlle
 /**
  Emitted when the alert level changes. This indicates the user should be notified about the upcoming maneuver.
```
1ec5: Since alert levels are no more, describe this notification in terms of the user passing an ideal point at which to announce an upcoming maneuver.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -434,4 +380,11 @@ open class RouteStepProgress: NSObject {
      The distance in meters the user is to the next intersection they will pass through.
      */
     public var userDistanceToUpcomingIntersection: CLLocationDistance?
+    
+    public var voiceInstructionIndex = 0
+    
+    public var voiceInstruction: SpokenInstruction? {
```
1ec5: Rename to `currentSpokenInstruction` for consistency with `currentStep`, `currentLeg`, etc.

MapboxCoreNavigation/Constants.swift
```diff
@@ -55,7 +50,7 @@ public let RouteControllerProgressDidChange = Notification.Name(MBRouteControlle
...
  */
-public let RouteControllerAlertLevelDidChange = Notification.Name(MBRouteControllerAlertLevelDidChange)
+public let RouteControllerDidPassVoiceInstructionPoint = Notification.Name(MBRouteControllerDidPassVoiceInstructionPoint)
```
1ec5: Rename to `RouteControllerDidPassSpokenInstructionPoint` for consistency. Can you look through the PR for other places that should say “spoken” instead of “voice”?

MapboxNavigation/PollyVoiceController.swift
1ec5: Is it possible to nest `<prosody>` tags in SSML? I could imagine a future in which the Directions API decides to pronounce some portion of an instruction slightly louder than another portion.

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -23,7 +23,9 @@ open class NavigationRouteOptions: RouteOptions {
         includesSteps = true
         routeShapeResolution = .full
         attributeOptions = [.congestionLevel, .expectedTravelTime]
+        includesVoiceInstructions = true
         includesExitRoundaboutManeuver = true
+        distanceMeasurementSystem = .imperial
```
1ec5: Pull this logic out into an extension on Locale, then get the current locale’s measurement system:

https://github.com/mapbox/mapbox-navigation-ios/blob/e1b588b74e9be38b8ba207cf39a553532d5bb650/MapboxCoreNavigation/DistanceFormatter.swift#L31-L37

bsudekum:
In order how this PR is blocked:

1. https://github.com/Project-OSRM/osrm-text-instructions/pull/168
1. Updating api-directions
1. https://github.com/mapbox/MapboxDirections.swift/pull/176
1. https://github.com/mapbox/MapboxDirections.swift/pull/175

Let's start powering through this list. @1ec5 @mcwhittemore can I get your eyes on https://github.com/Project-OSRM/osrm-text-instructions/pull/168 today?

bsudekum:
Began working on prefetching audio instructions here. Blocked by a few PRs. https://github.com/mapbox/mapbox-navigation-ios/compare/voice-instructions...prefetch-audio

1ec5:
This branch is pretty far from compiling. Are you working within a CocoaPods workspace? If so, please try developing directly against the project in this repository.

bsudekum:
Alright this is ready to go. @frederoni @1ec5 @JThramer could you 👀  this?

bsudekum:
@1ec5 @frederoni @JThramer okay, this is now ready for review.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -182,12 +177,7 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         // Set recentlyAnnouncedRouteStep to the current step
         recentlyAnnouncedRouteStep = routeProgress.currentLegProgress.currentStep
         
-        fallbackText = spokenInstructionFormatter.string(routeProgress: routeProgress, userDistance: userDistance, markUpWithSSML: false)
-        
-        // If the user is merging onto a highway, an announcement to merge is a bit excessive
-        if let upComingStep = routeProgress.currentLegProgress.upComingStep, routeProgress.currentLegProgress.currentStep.maneuverType == .takeOnRamp && upComingStep.maneuverType == .merge && routeProgress.currentLegProgress.alertUserLevel == .high {
-            return false
-        }
+        fallbackText = routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction?.text
```
1ec5: FYI, using plain text here will prevent the AVSpeechSynthesizer fallback from honoring pronunciation overrides coming from OSRM. Support for that feature has been awaiting a review in #624. Now we’ll instead have to parse the SSML source code for `<phoneme>` tags using an XML parser, either the SAX-style XMLParser or a third-party library.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -117,7 +116,10 @@ public class PollyVoiceController: RouteVoiceController {
             input.voiceId = voiceId
         }
         
-        input.text = "<speak><amazon:effect name=\"drc\"><prosody volume='\(instructionVoiceVolume)' rate='\(instructionVoiceSpeedRate)'>\(text)</prosody></amazon:effect></speak>"
+        let removedClosingOpenTag = String(text.characters.dropFirst("<speak>".characters.count).dropLast("</speak>".characters.count))
```
1ec5: This removes the first seven and last eight characters, even if those characters aren’t `<speak>` and `</speak>`, respectively. That would lead to malformed XML if the Directions API happened to return something like `<speak version="1.0" xml:lang="en-US">`. (The `version` and `xml:lang` attributes are [required by the SSML specification](https://www.w3.org/TR/speech-synthesis/#S3.1.1), though Polly currently seems to be lenient about this requirement.)

MapboxCoreNavigation/Constants.swift
```diff
@@ -53,9 +43,14 @@ public let RouteControllerDidFindFasterRouteKey = MBRouteControllerDidFindFaster
 public let RouteControllerProgressDidChange = Notification.Name(MBRouteControllerNotificationProgressDidChange)
 
 /**
- Emitted when the alert level changes. This indicates the user should be notified about the upcoming maneuver.
+ Emitted when the user passing an instruction location location.
```
1ec5: Typos.

> Emitted when the user passes an ideal point for saying an instruction aloud.

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -23,31 +23,32 @@ open class NavigationRouteOptions: RouteOptions {
         includesSteps = true
         routeShapeResolution = .full
         attributeOptions = [.congestionLevel, .expectedTravelTime]
-        includesExitRoundaboutManeuver = true
+        includesVoiceInstructions = true
```
1ec5: This will become `includesSpokenInstructions` in mapbox/MapboxDirections.swift#191.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -250,7 +250,7 @@ class RouteMapViewController: UIViewController {
         }
     }
 
-    func notifyAlertLevelDidChange(routeProgress: RouteProgress) {
+    func shouldGiveVoiceInstruction(routeProgress: RouteProgress) {
```
1ec5: RouteMapViewController has no reason to worry about speech. Rename this method to reflect what it does: `updateMapOverlays(for:)`.

MapboxNavigation/RouteTableViewController.swift
```diff
@@ -85,7 +85,7 @@ class RouteTableViewController: UIViewController {
         }
     }
         
-    func notifyAlertLevelDidChange() {
+    func shouldGiveVoiceInstruction() {
```
1ec5: A table view controller has no reason to worry about spoken instructions. Rename this method to reflect what it does: `reload()`.

MapboxNavigation/PollyVoiceController.swift
bsudekum: @1ec5 what would you suggest to change this to?

MapboxNavigation/RouteVoiceController.swift
bsudekum: Considering we're going to soon preload voice instructions, can we punt on this?

MapboxNavigation/PollyVoiceController.swift
1ec5: The original code was slightly more robust; at least we could be sure that it was replacing `<speak>` and `</speak>`.

A better solution is to find and replace these regular expressions: `^<speak(?:\s.[^>]+)?>` and `</speak>$` (with whatever string escaping you need). This only works because we’re still making assumptions about what the API will give us. It’ll still fail to replace if the API gives us a valid XML document (with an XML prologue), but in that case we’ll just fail to apply the desired `<prosody>` tag; at least we won’t end up with a speech synthesis request that errors out.

1ec5: You could also future-proof that regular expression somewhat to account for `xml:lang` and `version` being supplied, as required by the SSML specification: replace `^<speak\s*(?:\s+[^>]*)?>` with `$0<amazon:effect …`

bsudekum:
@1ec5 updated, I think this is ready.

MapboxNavigation/RouteVoiceController.swift
1ec5: Yes, hence the FYI. Preloading voice instructions is sort of orthogonal to this discussion, since AVSpeechSynthesizer announces the entire instruction, including freeform parts like the road name. However, we hope to make the fallback occur so infrequently that it’ll be an edge case, and mispronunciation of road names is only a small part of what makes the AVSpeechSynthesizer fallback jarring to the user.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -117,7 +116,11 @@ public class PollyVoiceController: RouteVoiceController {
             input.voiceId = voiceId
         }
         
-        input.text = "<speak><amazon:effect name=\"drc\"><prosody volume='\(instructionVoiceVolume)' rate='\(instructionVoiceSpeedRate)'>\(text)</prosody></amazon:effect></speak>"
+        let wrappedSSML = text
+            .replacingOccurrences(of: "^<speak\\s*(?:\\s+[^>]*)?>", with: "$0<amazon:effect name=\"drc\"><prosody volume='\(instructionVoiceVolume)' rate='\(instructionVoiceSpeedRate)'>", options: .regularExpression)
+            .replacingOccurrences(of: "</speak>$", with: "</prosody></amazon:effect></speak>",options: .regularExpression)
```
1ec5: Noting for posterity that we need to remove this code once the Directions API begins to add these tags itself.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -1,48 +1,6 @@
...
-     */
-    case high
-    
-    
-    /**
-     The user has completed the route.
-     */
-    case arrive
-}
+import AVFoundation
```
frederoni: Do we use AVFoundation in RouteProgress?

MapboxNavigation/PollyVoiceController.swift
bsudekum: @1ec5 what do you think about making this future proof-able and adding an xml parsing dependency? It might make sense to check and make sure we have all necessary keys.

1ec5: No, I think we should introduce an XML parsing library only when we need it for a particular purpose, so that we can evaluate each candidate library based on how well it’d handle our use case. As it stands, we’re planning to have the Directions API add these tags, so the SDK won’t have any XML parsing needs for now.

bsudekum:
Removed our custom tag addition code in https://github.com/mapbox/mapbox-navigation-ios/pull/614/commits/7254395ef3e76d1023570aa0feae0e692d31cd55. We'll need the new tags deployed to the Directions API before merging this.

/cc @mcwhittemore @1ec5 @ericrwolfe

PR #619, opened 2017-09-16
Move misplaced strings files

1ec5:
Some strings files had been misplaced due to swapped resource slugs on Transifex. As a result, the speech format strings and the feedback UI both had fewer working translations than they should’ve had. I corrected the resource slugs, moved the misplaced files to their correct locations, and updated all strings files based on Transifex. `tx` skipped some translations of Localizable.strings in MapboxNavigation because recent additions to the source file have pushed the translated rate below 80%.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/505#pullrequestreview-63195991
/cc @frederoni @bsudekum


1ec5:
> Should we also pull in the latest from Transifex? There are updates in es,ca,sv, and vi.

I ran `tx pull -a` but saw no changes. Are you seeing something different?

frederoni:
> Are you seeing something different?

Saw quite a few changes with just `tx pull` first but now I only get it when forcing `tx pull -f`

1ec5:
> Saw quite a few changes with just `tx pull` first but now I only get it when forcing `tx pull -f`

You&rsquo;re right, looks like a number of translations got stuck and only force-pulling worked. No idea why.

PR #621, opened 2017-09-16
OSRM Text Instructions v0.3.3

1ec5:
Upgraded to Mapbox iOS SDK v3.6.3 and OSRMTextInstructions.swift v0.3.3. Rely on OSRM Text Instructions for phrases embed instructions. Removed localizable format strings for these phrases, now that OSRM Text Instructions provides translations of them.

/cc @bsudekum


Cartfile
```diff
@@ -2,7 +2,7 @@ binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 3.6
 github "mapbox/MapboxDirections.swift" ~> 0.10
 github "52inc/Pulley" == 1.4
 github "rs/SDWebImage" ~> 4.1
-github "Project-OSRM/osrm-text-instructions.swift" ~> 0.3.2
+github "Project-OSRM/osrm-text-instructions.swift" ~> 0.3.3
```
bsudekum: Can you update the podspec as well?

1ec5: The podspec specifies `~> v0.3`, so running `pod install` should fetch v0.3.3 automatically. Are you seeing otherwise?

bsudekum: Ah right, good call.

PR #622, opened 2017-09-16
Wrap road names in SSML phoneme tag

1ec5:
Wrap road names them in SSML `<phoneme>` tags using pronunciation data supplied by the Directions API to ensure that Polly pronounces them correctly.

Depends on #621 and mapbox/MapboxDirections.swift#174.

/cc @bsudekum @frederoni


Cartfile
```diff
@@ -1,8 +1,8 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 3.6
-github "mapbox/MapboxDirections.swift" ~> 0.10
+github "mapbox/MapboxDirections.swift" "master"
```
bsudekum: Can we wait for a release? Compiling via cocoapods will fail until we update the podspec.

1ec5: Of course, that’s why I mentioned that this PR depends on mapbox/MapboxDirections.swift#174. I only put this line in here to verify that tests pass.

1ec5: This PR now depends on v0.10.5.

PR #617, opened 2017-09-15
Use MapboxVoice instead of aws

bsudekum:
This removes the dependency on AWS and also makes it so anyone using MapboxCoreNavigation can get access to AWS Polly with only a Mapbox access token. 

There is a lot do here, but so far so good.

- [x] release MapboxVoice
- [x] improve falling back to AVSpeechSynthesizer
- [x] Add ssml support back
- [x] localize voices

/cc @1ec5 @ericrwolfe @frederoni @mapbox/navigation @willwhite

PR #624, opened 2017-09-18
Supply phonetic road names to AVSpeechSynthesizer

1ec5:
This change provides AVSpeechSynthesizer with the same phonetic names that #622 provides Polly with. This way the AVSpeechSynthesizer fallback remains consistent with Polly in terms of its pronunciation of difficult names. The closer the two voices’ pronunciation, the less jarring it is when the fallback kicks in.

AVSpeechSynthesizer doesn’t support SSML; instead, on iOS 10 and above, it uses the native NSAttributedString type to hold attributes such as pronunciations. It turns out that the Alex voice (#612) does not support attributed strings, but it manages to correctly pronounce just about all the roads that are currently tagged with pronunciations anyways.

~~Depends on #621, mapbox/MapboxDirections.swift#174, and Project-OSRM/osrm-text-instructions.swift#39.~~

/cc @frederoni @bsudekum

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -198,7 +198,7 @@ public class PollyVoiceController: RouteVoiceController {
             
             // If the task is canceled, don't speak.
             // But if it's some sort of other error, use fallback voice.
-            if let error = error as NSError?, error.code == NSURLErrorCancelled {
+            if let error = error as? URLError, error.code == .cancelled {
                 return
             } else if let error = error {
                 // Cannot call super in a closure
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -3,6 +3,45 @@ import AVFoundation
 import MapboxDirections
 import MapboxCoreNavigation
 
+extension NSAttributedString {
+    @available(iOS 10.0, *)
+    public func pronounced(_ pronunciation: String) -> NSAttributedString {
+        let phoneticWords = pronunciation.components(separatedBy: " ")
+        let phoneticString = NSMutableAttributedString()
+        for (word, phoneticWord) in zip(string.components(separatedBy: " "), phoneticWords) {
+            // AVSpeechSynthesizer doesn’t recognize some common IPA symbols.
+            let phoneticWord = phoneticWord.byReplacing([("ɡ", "g"), ("ɹ", "r")])
+            if phoneticString.length > 0 {
+                phoneticString.append(NSAttributedString(string: " "))
+            }
+            phoneticString.append(NSAttributedString(string: word, attributes: [
+                NSAttributedStringKey(rawValue: AVSpeechSynthesisIPANotationAttribute): phoneticWord,
+            ]))
+        }
+        return phoneticString
+    }
+}
+
+extension SpokenInstruction {
+    @available(iOS 10.0, *)
+    func attributedText(for legProgress: RouteLegProgress) -> NSAttributedString {
+        let attributedText = NSMutableAttributedString(string: text)
+        let step = legProgress.currentStep
+        if let name = step.names?.first,
+            let phoneticName = step.phoneticNames?.first {
+            let nameRange = attributedText.mutableString.range(of: name)
+            attributedText.replaceCharacters(in: nameRange, with: NSAttributedString(string: name).pronounced(phoneticName))
+        }
+        if let step = legProgress.upComingStep,
+            let name = step.names?.first,
+            let phoneticName = step.phoneticNames?.first {
+            let nameRange = attributedText.mutableString.range(of: name)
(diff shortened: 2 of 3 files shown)


bsudekum:
@1ec5 does it make sense to add this just for `AVSpeechSynthesizer` if in theory all users will be using polly via https://github.com/mapbox/mapbox-navigation-ios/pull/617?

1ec5:
> in theory all users will be using polly via #617

The fallback exists for a few reasons, including:

1. The API may not be reachable at the time it’s needed. Since instructions are time-critical, a fallback needs to kick in locally.
1. Some of the languages this library is localized into are supported by AVSpeechSynthesizer but not Polly.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -104,24 +95,35 @@ public class PollyVoiceController: RouteVoiceController {
...
+        guard !awsTask.isCancelled else {
             return
         }
         
         guard let url = awsTask.result else {
-            super.speak(fallbackText, error: "No polly response")
+            print("No polly response")
+            speakWithoutPolly(for: routeProgress, userDistance: userDistance)
             return
         }
```
frederoni: Should we cancel the task before creating a new one?

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -150,13 +152,14 @@ public class PollyVoiceController: RouteVoiceController {
                         audioPlayer.volume = strongSelf.volume
                         audioPlayer.play()
```
frederoni: the return value of `.play()` is discardable but I think we should observe it.

MapboxNavigation/PollyVoiceController.swift
1ec5: Good idea, but I think we should address this issue in #661. Once that PR lands, I’ll rebase and do likewise here.

MapboxNavigation/PollyVoiceController.swift
1ec5: #661 takes care of responding to the return value.

bsudekum:
@1ec5 is this still actionable?

1ec5:
It would still be desirable to ensure that the AVSpeechSynthesizer fallback can handle phonetic names. AVSpeechSynthesizer isn’t going away; after all, Polly is optional, and it doesn’t support most of the languages we do, like Chinese.

For AVSpeechSynthesizer to apply the phonetic names as string attributes, this library currently depends on OSRMTextInstructions.swift, hence all the merge conflicts. Relying on OSRMTextInstructions.swift would create some inconsistencies between the instructions spoken by Polly and those spoken by AVSpeechSynthesizer, due to extra processing we’re doing server-side outside of OSRM Text Instructions. A better solution would be for the Directions API to indicate the ranges within the plain-text instruction that should be pronounced a certain way. We can track that issue internally.

I’ve also been leaving this PR around in order to track a refactoring of PollyVoiceController and RouteVoiceController, but I had you carry it out in #800.

1ec5:
> A better solution would be for the Directions API to indicate the ranges within the plain-text instruction that should be pronounced a certain way. We can track that issue internally.

In the meantime, RouteVoiceController can search for the current route step’s `names` inside the plain-text instruction and apply the step’s `phoneticNames` to those substrings inside an attributed string. The assumption would be that we wouldn’t encounter a single instruction that contains the same name pronounced two different ways. Hopefully the road name is always distinguishable from other instruction text.

1ec5:
Ready for review.

frederoni:
This implementation currently relies on RouteStepFormatter which is getting removed in #767. 
Would it make sense to wait for 767 to land and adapt this PR afterward?

bsudekum:
> Would it make sense to wait for 767 to land and adapt this PR afterward?

Yeah let's do that.

1ec5:
Yes, we should wait until after #767 lands. This PR still includes the modifications to RouteStepFormatter mainly so that the project can still build. But the cruxt of the change is in RouteVoiceController.

1ec5:
All set.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -167,18 +208,28 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
             voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
         }
         
-        let utterance = AVSpeechUtterance(string: instruction.text)
-        
-        // Only localized languages will have a proper fallback voice
+        var utterance: AVSpeechUtterance?
         if Locale.preferredLocalLanguageCountryCode == "en-US" {
-            utterance.voice = AVSpeechSynthesisVoice(identifier: AVSpeechSynthesisVoiceIdentifierAlex)
+            // Alex can’t handle attributed text.
```
bsudekum: 😢

PR #626, opened 2017-09-19
Round distance to nearest 100m

frederoni:
When using the metric measurement system, round to nearest ~~100m~~ 50m in the 100-1000m range.

Imperial changed to 0.1 in rounding increment for miles.

@1ec5 @bsudekum @ericrwolfe


MapboxCoreNavigationTests/DistanceFormatterTests.swift
```diff
@@ -50,9 +51,11 @@ class DistanceFormatterTests: XCTestCase {
         assertDistance(10,      spoken: "0 Meter",          displayed: "0 m")
         assertDistance(15,      spoken: "25 Meter",         displayed: "25 m")
         assertDistance(24,      spoken: "25 Meter",         displayed: "25 m")
+        assertDistance(89,      spoken: "100 Meter",        displayed: "100 m")
+        assertDistance(251,     spoken: "300 Meter",        displayed: "300 m")
         assertDistance(500,     spoken: "500 Meter",        displayed: "500 m")
-        assertDistance(949,     spoken: "950 Meter",        displayed: "950 m")
-        assertDistance(951,     spoken: "950 Meter",        displayed: "950 m")
+        assertDistance(949,     spoken: "900 Meter",        displayed: "900 m")
```
bsudekum: I'd sort of expect this distance to be `950 Meter`

frederoni: Maps.app start rounding to 100m when distance is > 500m. GMaps rounds to 100m when distance > 1000m, as you expected @bsudekum.

On a second thought, I agree with you. If possible, we should round to nearest 100m only on motorways.

Issue #575, opened 2017-09-01
Roundabout icon assumes right-hand driving


1ec5:
For roundabouts, the turn banner shows an icon that goes counterclockwise, even in countries that drive on the left:

<img src="https://user-images.githubusercontent.com/1231218/29976960-ab4957aa-8ef0-11e7-9c6a-124768364dc2.png" width="300" alt="bogeshole">

Even without Project-OSRM/osrm-backend#2772 / Project-OSRM/osrm-backend#4415, we can flip the icon based on whether the roundabout maneuver’s initial and final headings indicate a left or right turn, since those headings correspond to the entrance maneuver, as opposed to the exit maneuver.

I think this issue will be unaffected by Project-OSRM/osrm-backend#4358.

/cc @bsudekum @frederoni @MoKob

1ec5:
The fix would go [on this line](https://github.com/mapbox/mapbox-navigation-ios/blob/b6ebf8ef9a0cfb77f0ece9a1b0515a3c3aeabf4d/MapboxNavigation/TurnArrowView.swift#L76) or thereabouts.

1ec5:
Similar to #640, I think fixing this issue would entail looking at the enter-roundabout step’s maneuver direction: if it’s to the left, then flip the roundabout icon horizontally.

1ec5:
> I think fixing this issue would entail looking at the enter-roundabout step’s maneuver direction: if it’s to the left, then flip the roundabout icon horizontally.

[Here’s an example](https://www.mapbox.com/get-directions/#16.98/39.32167/-84.29224?coordinates=-84.28888880651257,39.321539298960175;-84.29223720379989,39.32166688942459) where the step’s `direction` (derived from the maneuver’s `modifier` in the Directions API response) will be `slightRight` when using a roundabout to continue going straight, whether or not `includesExitRoundaboutManeuver` (`roundabout_exits`) is set. However, note that the `direction` will be the overall direction of travel in the case of a “roundabout turn” maneuver.

1ec5:
Fixed in #911.

/cc @bsudekum

PR #640, opened 2017-09-21
Draw roundabout based on maneuver direction

### LINKED ISSUE #638: Incorrect arrow exit indicator on roundabout
On the current 0.80.0 SDK, the navigation displays an incorrect roundabout exit arrow indicator.

img_0012

The roundabout in question is:
https://www.hitta.se/kartan!~59.21583,17.63506,16z/tr!i=6Kbb9lGk/geocode!l=59.21595:17.63450

frederoni:
Fixes #638 

`takeRoundabout`, `turnAtRoundabout`, and `takeRotary` with any maneuver direction were previously drawn as a roundabout with a right turn.

@bsudekum @1ec5 

/cc @Sealos


MapboxNavigation/TurnArrowView.swift
```diff
@@ -73,7 +73,14 @@ public class TurnArrowView: UIView {
...
             flip = [.right, .slightRight, .sharpRight].contains(direction)
         case .takeRoundabout, .turnAtRoundabout, .takeRotary:
-            StyleKitArrows.drawRoundabout(primaryColor: primaryColor, secondaryColor: secondaryColor, scale: scale)
+            switch direction {
+            case .straightAhead:
+                StyleKitArrows.drawRoundabout_180(primaryColor: primaryColor, secondaryColor: secondaryColor, scale: scale)
+            case .left, .slightLeft, .sharpLeft:
+                StyleKitArrows.drawRoundabout_275(primaryColor: primaryColor, secondaryColor: secondaryColor, scale: scale)
+            default:
+                StyleKitArrows.drawRoundabout(primaryColor: primaryColor, secondaryColor: secondaryColor, scale: scale)
```
1ec5: @mcwhittemore pointed out that this logic doesn’t seem to account for left-hand driving. @frederoni, do we need to set `flip` somehow here?

mcwhittemore: this isn't in the OSRM response...

1ec5: Ah, I see that was ticketed out separately as #575.

Issue #639, opened 2017-09-21
Snap to road on two way streets


Sealos:
Snapping to the road is not taking into consideration the user heading and street heading.

When multiple segments are present on the same street
![img_0003](https://user-images.githubusercontent.com/4553825/30697055-f4ffa118-9edd-11e7-9f31-6bc7ed746666.PNG)

After finishing that multi-segment
![img_0004](https://user-images.githubusercontent.com/4553825/30697060-f6f0ae4a-9edd-11e7-96f4-1a20e71fdad7.PNG)

bsudekum:
I think the issue here is that we use `routeProgress.currentLegProgress.nearbyCoordinates` which includes the previous step, current step and upcoming step coordinates to calculate where to snap. I think it might safe to only look at the current step coordinates if the upcoming step is a maneuver type == .uturn.

PR #642, opened 2017-09-21
Fix uturn step increment and snapping

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/639

This addresses two issues:

1. When a step has coordinates that are close to the current step (think in a uturn), we sometimes snap to the wrong step coordinates. To fix this, when the upcoming maneuver modifier is a uturn, we only look at the current step coordinates for snapping
1. There is a bug in OSRM where exit bearings are calculated incorrectly. This prevents us from incrementing the step counter because the users course never is equal to the exit bearing. This hack should be removed when fixed upstream.

/cc @danpat @1ec5 @frederoni

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -288,13 +288,14 @@ open class RouteController: NSObject {
         guard let stepCoordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return nil }
         guard let snappedCoordinate = closestCoordinate(on: stepCoordinates, to: location.coordinate) else { return location }
         
-        var nearByCoordinates = routeProgress.currentLegProgress.currentAndUpcomgingStepCoordinates
+        var nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
         
         // If the upcoming step is a roundabout, only look at the current step for snapping.
         // Otherwise, we may get false positives from nearby step coordinates
         if let upcomingStep = routeProgress.currentLegProgress.upComingStep, upcomingStep.maneuverDirection == .uTurn, let coordinates = routeProgress.currentLegProgress.currentStep.coordinates {
            nearByCoordinates = coordinates
         }
+
         guard let closest = closestCoordinate(on: nearByCoordinates, to: location.coordinate) else { return nil }
         
         let slicedLineBehind = polyline(along: nearByCoordinates.reversed(), from: closest.coordinate, to: nearByCoordinates.reversed().last)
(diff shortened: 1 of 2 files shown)


1ec5:
> There is a bug in OSRM where exit bearings are calculated incorrectly. This prevents us from incrementing the step counter because the users course never is equal to the exit bearing. This hack should be removed when fixed upstream.

How urgent is this fix? Can we wait until the fix lands in the Directions API?

ericrwolfe:
It's been in this long already now. I think we can wait for an upstream fix.

bsudekum:
I think if we can fix this now, we should. Once it's fixed upstream, the code path won't even be hit. It's very [harmless code](https://github.com/mapbox/mapbox-navigation-ios/pull/642/files#diff-1a2149692b3d7c58b6a0392120d6011aR755).

bsudekum:
removed hack. OSRM fix should be 1-2 weeks out.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -288,7 +288,13 @@ open class RouteController: NSObject {
         guard let stepCoordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return nil }
         guard let snappedCoordinate = closestCoordinate(on: stepCoordinates, to: location.coordinate) else { return location }
         
-        let nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
+        var nearByCoordinates = routeProgress.currentLegProgress.currentAndUpcomgingStepCoordinates
```
bsudekum: Maybe this should remain `nearbyCoordinates`? Although I cant think of any reason to know about the previous step coords

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -344,7 +344,18 @@ open class RouteLegProgress: NSObject {
...
         let nearby = priorCoords + currentCoords + upcomingCoords
-        assert(!nearby.isEmpty, "Polyline must coordinates")
+        assert(!nearby.isEmpty, "Step must have coordinates")
+        return nearby
+    }
+    
+    /**
+     Returns an array of `CLLocationCoordinate2D` of the current and upcoming step geometry.
+     */
+    public var currentAndUpcomgingStepCoordinates: [CLLocationCoordinate2D] {
```
1ec5: Typo: upcoming.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -344,7 +344,18 @@ open class RouteLegProgress: NSObject {
...
+        let upcomingCoords = upComingStep?.coordinates ?? []
+        let currentCoords = currentStep.coordinates ?? []
+        let nearby = currentCoords + upcomingCoords
```
1ec5: Is a separate method really needed to concatenate these two arrays, given that there’s only one call site?

MapboxCoreNavigation/RouteController.swift
1ec5: At a glance, I can’t think of any reason it would make a difference whether we include the previous coordinates or not.

bsudekum:
@1ec5 fixed.

Issue #643, opened 2017-09-21
Make feedback button on by default


bsudekum:
Currently, it's false. Let's have it on for all users.

/cc @ericrwolfe

PR #644, opened 2017-09-21
Make feedback and style switching on by default

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/643

This turns on auto style switching at sunset and the feedback button by default. These features help users and have been tested now are ready.

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -266,11 +266,7 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
     /**
      Shows a button that allows drivers to report feedback such as accidents, closed roads,  poor instructions, etc. Defaults to `false`.
      */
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -266,16 +266,12 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
     /**
      Shows a button that allows drivers to report feedback such as accidents, closed roads,  poor instructions, etc. Defaults to `false`.
      */
-    public var showsReportFeedback: Bool = false {
-        didSet {
-            mapViewController?.reportButton.isHidden = !showsReportFeedback
```
1ec5: Note that this `didSet` made it possible to show or hide the feedback button on the fly.

bsudekum: Eh, I don't really see the use case for this.

PR #646, opened 2017-09-21
Reroute when the user is moving away from maneuver

bsudekum:
If the user is moving away from the maneuver (whips a uturn on a country road half way down a step), we should reroute them once we know they are really moving away.

The gist:
* Calculate distance to maneuver
* add to an array
* check the last distance in the array. If it's greater than the previous distance, add it
* if this array grows to be greater than 3 distances, reroute.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -169,7 +169,7 @@ open class RouteController: NSObject {
     
     var hasFoundOneQualifiedLocation = false
     
-    var previousDistancesFromManeuver: [CLLocationDistance] = []
+    var recentDistancesFromManeuver: [CLLocationDistance] = []
     
     /**
      Intializes a new `RouteController`.
@@ -550,14 +550,14 @@ extension RouteController: CLLocationManagerDelegate {
         if let coordinates = routeProgress.currentLegProgress.currentStep.coordinates {
             let userDistanceToManeuver = distance(along: coordinates, from: location.coordinate)
             
-            guard previousDistancesFromManeuver.count <= 3 else {
+            guard recentDistancesFromManeuver.count <= 3 else {
                 resetPreviousDistanceArray()
                 return false
             }
             
-            if previousDistancesFromManeuver.isEmpty {
-                previousDistancesFromManeuver.append(userDistanceToManeuver)
-            } else if let lastSpeed = previousDistancesFromManeuver.last, userDistanceToManeuver > lastSpeed {
+            if recentDistancesFromManeuver.isEmpty {
+                recentDistancesFromManeuver.append(userDistanceToManeuver)
+            } else if let lastSpeed = recentDistancesFromManeuver.last, userDistanceToManeuver > lastSpeed {
                 previousDistancesFromManeuver.append(userDistanceToManeuver)
             } else {
                 // If we get a descending distance, reset the counter
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -386,6 +386,7 @@ extension RouteController {
             sessionState.arrivalTimestamp = Date()
             sendArriveEvent()
         }
+        resetPreviousDistanceArray()
     }
     
     func willReroute(notification: NSNotification) {
@@ -396,6 +397,7 @@ extension RouteController {
         if let lastReroute = outstandingFeedbackEvents.map({$0 as? RerouteEvent }).last {
             lastReroute?.update(newRoute: routeProgress.route)
         }
+        resetPreviousDistanceArray()
     }
 }
 
@@ -558,7 +560,7 @@ extension RouteController: CLLocationManagerDelegate {
             if recentDistancesFromManeuver.isEmpty {
                 recentDistancesFromManeuver.append(userDistanceToManeuver)
             } else if let lastSpeed = recentDistancesFromManeuver.last, userDistanceToManeuver > lastSpeed {
-                previousDistancesFromManeuver.append(userDistanceToManeuver)
+                recentDistancesFromManeuver.append(userDistanceToManeuver)
             } else {
                 // If we get a descending distance, reset the counter
                 resetPreviousDistanceArray()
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -557,9 +581,14 @@ extension RouteController: CLLocationManagerDelegate {
         return isCloseToCurrentStep
     }
     
+    func resetPreviousDistanceArray() {
+        previousDistancesFromManeuver.removeAll()
```
1ec5: Is a separate method really needed for this statement? It looks like we’re calling it every time we touch `routeProgress.currentLegProgress.stepIndex` – maybe what this class needs is a property that mirrors `routeProgress.currentLegProgress.stepIndex`:

```swift
var currentStepIndex: Int {
    get {
        return routeProgress.currentLegProgress.stepIndex
    }
    set {
        routeProgress.currentLegProgress.stepIndex = newValue
        recentDistancesFromManeuver.removeAll()
    }
}
```

MapboxCoreNavigation/RouteController.swift
```diff
@@ -541,6 +543,28 @@ extension RouteController: CLLocationManagerDelegate {
...
         
+        
+        // Check to see if the user is moving away from the maneuver.
+        // Here, we store an array of distances. If the current distance is greater than the last distance,
+        // add it to the array. If the array grows larger than x, reroute the user.
+        if let coordinates = routeProgress.currentLegProgress.currentStep.coordinates {
+            let userDistanceToManeuver = distance(along: coordinates, from: location.coordinate)
+            
+            guard previousDistancesFromManeuver.count <= 3 else {
+                resetPreviousDistanceArray()
```
1ec5: We should make it so that rerouting automatically calls this method, making this line redundant.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -541,6 +543,28 @@ extension RouteController: CLLocationManagerDelegate {
...
+                return false
+            }
+            
+            if previousDistancesFromManeuver.isEmpty {
+                previousDistancesFromManeuver.append(userDistanceToManeuver)
+            } else if let lastSpeed = previousDistancesFromManeuver.last, userDistanceToManeuver > lastSpeed {
```
1ec5: Are we comparing distances or speeds?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -541,6 +543,28 @@ extension RouteController: CLLocationManagerDelegate {
...
+                previousDistancesFromManeuver.append(userDistanceToManeuver)
```
1ec5: `RouteStepProgress.userDistanceToManeuverLocation` already tracks the most recent distance, so all this class needs is to keep a counter of increasing distances: `countOfRecentBackwardsUpdates` or something along those lines.

MapboxCoreNavigation/RouteController.swift
bsudekum: As much as I like this style, I don't like hiding `routeProgress.currentLegProgress.stepIndex` behind another var. I like explicitly changing values on `routeProgress`.

1ec5: Two alternatives:

* Use KVO to observe for changes to the key path `routeProgress.currentLegProgress` and reset the array.
* Listen for alert level change notifications and reset the array.

bsudekum:
@1ec5 updated.

bsudekum:
Note, this does not solve the case when the user is at the beginning of the route and moving away since distance along has a max value (length of the step). However, if the user is moving away from the departure maneuver, they should be rerouted quickly.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -541,6 +545,28 @@ extension RouteController: CLLocationManagerDelegate {
...
+        
+        // Check to see if the user is moving away from the maneuver.
+        // Here, we store an array of distances. If the current distance is greater than the last distance,
+        // add it to the array. If the array grows larger than x, reroute the user.
+        if let coordinates = routeProgress.currentLegProgress.currentStep.coordinates {
+            let userDistanceToManeuver = distance(along: coordinates, from: location.coordinate)
+            
+            guard recentDistancesFromManeuver.count <= 3 else {
+                resetPreviousDistanceArray()
+                return false
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/646#discussion_r140572776

MapboxCoreNavigation/RouteController.swift
```diff
@@ -541,6 +545,28 @@ extension RouteController: CLLocationManagerDelegate {
...
+            }
+            
+            if recentDistancesFromManeuver.isEmpty {
+                recentDistancesFromManeuver.append(userDistanceToManeuver)
+            } else if let lastSpeed = recentDistancesFromManeuver.last, userDistanceToManeuver > lastSpeed {
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/646#discussion_r140573344

MapboxCoreNavigation/RouteController.swift
```diff
@@ -557,6 +583,10 @@ extension RouteController: CLLocationManagerDelegate {
         return isCloseToCurrentStep
     }
     
+    func resetPreviousDistanceArray() {
+        recentDistancesFromManeuver.removeAll()
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/646#discussion_r140572049

MapboxCoreNavigation/RouteController.swift
bsudekum: @1ec5 solved here: https://github.com/mapbox/mapbox-navigation-ios/pull/646/files#diff-1a2149692b3d7c58b6a0392120d6011aR566

bsudekum:
@1ec5 👍 fixed

MapboxCoreNavigation/RouteController.swift
```diff
@@ -394,6 +397,7 @@ extension RouteController {
         if let lastReroute = outstandingFeedbackEvents.map({$0 as? RerouteEvent }).last {
             lastReroute?.update(newRoute: routeProgress.route)
         }
+        recentDistancesFromManeuver.removeAll()
```
1ec5: Does it make a difference if this is in `willReroute(notification:)` or `didReroute(notification:)`? Wondering if there’s a possibility of a race condition if a location even more distant from the maneuver comes in while waiting for a response from the API.

PR #650, opened 2017-09-25
Issue with access control for a string extension on MapboxCoreNavigation

AFcgi:
I get an compiler error that the `wholeRange` needs another access control to use them.

### DIFF
--- a/MapboxCoreNavigation/String.swift
+++ b/MapboxCoreNavigation/String.swift
@@ -12,7 +12,7 @@ extension String {
         return isEmpty ? nil : self
     }
     
-    var wholeRange: Range<String.Index> {
+    public var wholeRange: Range<String.Index> {
         return startIndex..<endIndex
     }
     


bsudekum:
This should be fixed by https://github.com/mapbox/mapbox-navigation-ios/pull/632. @AFcgi can you verify this by testing master? If this is case, we'll put out a new release.

/cc @1ec5 @fred

1ec5:
I snuck #632 in when pushing v0.8.0 to CocoaPods, but it turns out the podspec refers to the tag (which doesn't have the fix), so we do need to put out a new patch release for CocoaPods.

bsudekum:
@1ec5 so is this PR necessary?

AFcgi:
@bsudekum I don't test it. But I think the #632 is correct.

1ec5:
Closing in favor of #665.

PR #652, opened 2017-09-25
Timeout polly requests sooner

bsudekum:
This times out polly requests after 2 seconds. I'm keeping this low because voice instructions are very time/location based and can become outdated quickly.


bsudekum:
@frederoni yep, I agree. I think we can start to do this once https://github.com/mapbox/mapbox-navigation-ios/pull/614 is in.

PR #653, opened 2017-09-25
Adopt turf-swift

### LINKED ISSUE #396: Use turf-swift
and remove geometry.swift 

https://github.com/mapbox/turf-swift

frederoni:
Fixes #396 

This PR replaces Geometry.swift with [Turf-swift](https://github.com/mapbox/turf-swift)

Blocked by 
- [x] https://github.com/mapbox/turf-swift/pull/14
- [x] https://github.com/mapbox/turf-swift/pull/16
- [x] Submit turf-swift to the CocoaPods public repo
- [x] Add turf-swift as a CocoaPods dependency

@bsudekum @1ec5 👀


MapboxCoreNavigation/RouteStep.swift
```diff
@@ -73,3 +74,15 @@ extension RouteStep {
...
 
 }
+
+extension CLLocation {
+    /**
+     Returns a Boolean value indicating whether the receiver is within a given distance of a route step, inclusive.
+     */
+    func isWithin(_ maximumDistance: CLLocationDistance, of routeStep: RouteStep) -> Bool {
+        guard let closestCoordinate = Polyline(routeStep.coordinates!).closestCoordinate(to: coordinate) else {
+            return false // TODO: Verify - This was true before
```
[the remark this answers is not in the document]
bsudekum: @frederoni 🤔  can we check the count returned from this? If  == 0, return?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -758,10 +762,10 @@ extension RouteController: CLLocationManagerDelegate {
...
         // we count this as within the `RouteControllerMaximumAllowedDegreeOffsetForTurnCompletion`
         if let upcomingStep = routeProgress.currentLegProgress.upComingStep, let finalHeading = upcomingStep.finalHeading, let initialHeading = upcomingStep.initialHeading {
-            let initialHeadingNormalized = wrap(initialHeading, min: 0, max: 360)
-            let finalHeadingNormalized = wrap(finalHeading, min: 0, max: 360)
-            let userHeadingNormalized = wrap(location.course, min: 0, max: 360)
-            let expectedTurningAngle = differenceBetweenAngles(initialHeadingNormalized, finalHeadingNormalized)
+            let initialHeadingNormalized = Turf.wrap(initialHeading, min: 0, max: 360)
+            let finalHeadingNormalized = Turf.wrap(finalHeading, min: 0, max: 360)
+            let userHeadingNormalized = Turf.wrap(location.course, min: 0, max: 360)
+            let expectedTurningAngle = Turf.differenceBetweenAngles(initialHeadingNormalized, finalHeadingNormalized)
```
1ec5: Maybe this method could be in an extension on CLLocationDirection or CLHeading. 🤔

frederoni: Or `<T: FloatingPoint>` 🤔 ?

frederoni: Or `<T: FloatingPoint>`? 🤔

1ec5: That sounds reasonable, although extending something standard like `FloatingPoint` for something as commonplace as wrapping could run the risk of conflicts with other libraries or application code.

MapboxCoreNavigation/RouteStep.swift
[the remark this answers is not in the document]
frederoni: Guarding `routeStep.coordinates.count` prior to the current guard and return false if the count is 0?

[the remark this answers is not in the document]
bsudekum: Oh I see now. I think returning true was a bug before. If we can't find the closest coordinate, we should not say you're withIn x.

1ec5:
This PR will fix #396.

PR #654, opened 2017-09-25
Improve departure announcement

bsudekum:
We've been omitting departure instructions like `Head west on foo street.`

This change now makes departure instructions:

* Upcoming maneuver is > 2000 meters: `Continue west on foo street for 5 miles.`
* Upcoming maneuver is relatively close, less than 2000 meters: `Head west on foo street, then in 0.7 miles, turn right onto bar street.`
* Upcoming maneuver is very close: `Turn right onto bar street`.

Note, this only can occur when on the departure step and the alert level is .depart or .low.

### DIFF
--- a/MapboxCoreNavigation/SpokenInstructionFormatter.swift
+++ b/MapboxCoreNavigation/SpokenInstructionFormatter.swift
@@ -72,9 +72,9 @@ public class SpokenInstructionFormatter: NSObject {
         // We only want to announce this special depature announcement once.
         // Once it has been announced, all subsequnt announcements will not have an alert level of low
         // since the user will be approaching the maneuver location.
-        let isDeparture = routeProgress.currentLegProgress.currentStep.maneuverType == .depart && (alertLevel == .depart || alertLevel == .low)
-        if let currentInstruction = currentInstruction, isDeparture {
-            if routeProgress.currentLegProgress.currentStep.distance > 2_000 {
+        let isStartingDeparture = routeProgress.currentLegProgress.currentStep.maneuverType == .depart && (alertLevel == .depart || alertLevel == .low)
+        if let currentInstruction = currentInstruction, isStartingDeparture {
+            if routeProgress.currentLegProgress.currentStep.distance > RouteControllerMinDistanceForContinueInstruction {
                 text = currentInstruction
             } else if upcomingStepDuration > linkedInstructionMultiplier {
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/SpokenInstructionFormatter.swift
```diff
@@ -72,9 +72,33 @@ public class SpokenInstructionFormatter: NSObject {
         // We only want to announce this special depature announcement once.
         // Once it has been announced, all subsequnt announcements will not have an alert level of low
         // since the user will be approaching the maneuver location.
-        let isDeparture = routeProgress.currentLegProgress.currentStep.maneuverType == .depart && alertLevel == .depart
-        let didCompleteManeuver = routeProgress.currentLegProgress.currentStep.distance > 2_000 && routeProgress.currentLegProgress.alertUserLevel == .low
-        if isDeparture || didCompleteManeuver {
+        let isDeparture = routeProgress.currentLegProgress.currentStep.maneuverType == .depart && (alertLevel == .depart || alertLevel == .low)
```
1ec5: Maybe rename this variable to `isStartingDeparture` to reflect its narrower scope.

MapboxCoreNavigation/SpokenInstructionFormatter.swift
```diff
@@ -72,9 +72,33 @@ public class SpokenInstructionFormatter: NSObject {
...
+        if let currentInstruction = currentInstruction, isDeparture {
+            if routeProgress.currentLegProgress.currentStep.distance > 2_000 {
```
1ec5: Pull this number out into a constant and document it.

1ec5: Should this be based on the step's overall distance or the distance remaining to the maneuver? Also, what happens if the user is a significant distance from the departure point, like in a parking lot – how does that affect the behavior?

bsudekum: > Also, what happens if the user is a significant distance from the departure point, like in a parking lot – how does that affect the behavior?

Let's pull this out into a different PR.

bsudekum:
@1ec5 updated, pulling https://github.com/mapbox/mapbox-navigation-ios/pull/654#discussion_r141128823 into a followup task.

PR #656, opened 2017-09-26
Animate camera for maneuvers and motorways

bsudekum:
Fairly simple rules right now:

* If the current step is a motorway, keep pitch at 45 degrees and change the altitude to 2000 meters
* if the current step is not a motorway and the alert level is high, go into overhead mode
* Otherwise, set pitch to 45 degrees and altitude to 600 meters

Until concurrent animations are fixed, this is not really actionable.

/cc @ericrwolfe @frederoni @1ec5 @kronick @d-prukop

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -123,9 +123,6 @@ class RouteMapViewController: UIViewController {
                 camera.heading = coordinates[0].direction(to: coordinates[1])
             }
             mapView.setCamera(camera, animated: false)
-            mapView.setCamera(cameraCurrentForRouteProgress, withDuration: 0.5, animationTimingFunction: nil, completionHandler: {
-                self.mapView.setUserLocationVerticalAlignment(.bottom, animated: true)
-            })
         }
     }
     
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -43,7 +43,17 @@ class RouteMapViewController: UIViewController {
         }
         return parent.pendingCamera
     }
-    var tiltedCamera: MGLMapCamera {
+    
+    var overheadCamera: MGLMapCamera {
+        get {
+            let camera = mapView.camera
+            camera.altitude = 600
+            camera.pitch = 0
+            return camera
+        }
+    }
+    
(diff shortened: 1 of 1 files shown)


bsudekum:
As a first, mergeable step, we could animate the camera's altitude in and out when entering and leaving a highway.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -282,6 +308,25 @@ class RouteMapViewController: UIViewController {
...
+        
+        guard !isInOverviewMode else {
+            return genericTiltedCamera
+        }
+        
+        guard let _ = routeController.routeProgress.currentLegProgress.upComingStep else {
+            return genericTiltedCamera
+        }
+        
+        if let _ = routeController.routeProgress.currentLegProgress.currentStep.intersections?.first?.outletRoadClasses?.contains(.motorway) {
```
JThramer: Oof. I see what you're doing here, but that's a doozy of a line. In this case, I might be tempted to use a let to define _what_ exactly i'm looking for (some semantic clue to the developer that's more informative then the call chain), and then use _chaining_ to do the test. i.e.:

```swift
if let currentIntersection = routeController.routeProgress.currentLegProgress.currentStep.intersections?.first,
   let outletTypes = currentIntersection.outletRoadClasses,
   outletTypes.contains(.motorway) {
```

IMHO this reads quite a bit more naturally, but YMMV.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -282,6 +308,25 @@ class RouteMapViewController: UIViewController {
...
+            return motorwayCamera
+        } else if routeController.routeProgress.currentLegProgress.alertUserLevel == .high {
```
JThramer: It's also worth calling out that since we're grabbing the currentLegProgress more than once, it may be worth extracting that out into a local variable (above 322) for reuse. IMHO this would improve readability even more.

```swift
let legProgress = routecontroller.routeProgress.currentLegProgress

if let intersection = legProgress.currentStep.intersections?.first,
   let outletTypes = intersection.outletRoadClasses,
   outletTypes.contains(.motorway) {
    return motorwayCamera
} else if legProgress.alertUserLevel == .high {
    return overheadCamera
} else {
    return genericTiledCamera
}
```

PR #657, opened 2017-09-26
Show more instructions before truncating

### LINKED ISSUE #649: The navigation instruction text is cut when it doesn't fit on one line
Version 0.8.0

The navigation instruction text is truncated even though there is plenty of space to have 2 lines.

image

frederoni:
Fixes #649 

Increased to three lines of instructions if the distance is hidden and two lines when the distance is visible.

@bsudekum 👀 

/cc @Sealos


MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -45,7 +45,7 @@ class RouteManeuverViewController: UIViewController {
     }
     
     var numberOfDestinationLines: Int {
-        return distance != nil ? 1 : 2
+        return distance != nil ? 2 : 3
```
bsudekum: @frederoni In the case of 3 lines, it looks like the test fixtures are dropping the distance. Is this the case?

frederoni: Distance is set to nil in those cases

https://github.com/mapbox/mapbox-navigation-ios/blob/9be509be36bf3fe2f6ad64ad70a0e6e71b86ffbe/MapboxNavigationTests/MapboxNavigationTests.swift#L93-L94

bsudekum: You think we could have a value? Just would like to see how it all fits together.

frederoni: `numberOfDestinationLines` would return 2 if you set a distance. See `testManeuverViewLongDestinationWithDistance_iPhone10_3_1_0x0@3x.png` for the expected result for 2 lines with distance. 3 lines with distance are avoided by design.

frederoni: Or you mean we should try 4 lines w/o distance and 3 lines w/ distance?

bsudekum: I guess I'm just looking to know what 
image

looks like with a distance. I think no matter what, we need to always show the distance.

frederoni: It would scale the font to an unglanceable size.
Distance is set to nil if the step.distance <= 0, which rarely happens?
https://github.com/mapbox/mapbox-navigation-ios/blob/2f6c999a167c2e0fa975330535ea9ddba53f72d5/MapboxNavigation/RouteMapViewController.swift#L559

bsudekum: Can we not set it to nil. I'm just trying to understand what 3 lines of text + a distance will look like. distance should never really be nil.

frederoni: screenshot 2017-09-27 11 29 03

There's not enough room with the fixed top and bottom margin so weird things happen with 3 lines + distance and 50% min font scaling. I guess we could increase the font scaling even more but the instruction would not be glanceable.

bsudekum: 🤔 , I'm thinking it might be more important to display the distance than to display 3 lines. Maybe we can move to 3 lines when a .high alert is given?

frederoni: The only noticeable change in this PR is 2 lines of instructions instead of 1 when the distance is shown. Let's make additional tweaks when we have an updated design.

Issue #659, opened 2017-09-26
Snapping to sharp (next) turns


bsudekum:
Followup to https://github.com/mapbox/mapbox-navigation-ios/pull/642

![pasted_image_9_26_17__9_40_am](https://user-images.githubusercontent.com/1058624/30872476-101c5972-a29f-11e7-865b-1f91ebf24288.png)

I got snapped to the next step even though I was still headed north east on market street here. We should look at the sharpness of the upcoming maneuver to see which coordinates to snap to.

/cc @1ec5 @frederoni

PR #660, opened 2017-09-26
Improve steps with short distances

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/659

I ran into a rough intersection (https://github.com/mapbox/mapbox-navigation-ios/issues/659) this weekend that surfaced a few issues. Noting, whether this intersection should only be one maneuver (it should) is outside of the scope of this PR.

1. Our final check before rerouting to see if the user is on the next step should also check on the followOnStep
1. Using `.uTurn` as the red flag to only snap to the current step, is false. it should be the angle of the maneuver. When sharp, look 
1. In some cases on shorter steps, location updates might be very scarce. Because of this, we don't have a user course to compare to see if they user has completed a maneuver. In this case, we can check if the user has first entered the maneuver zone and is moving away from the maneuver and has exited the maneuver zone. In this case, increment the step.

There is a lot to unpack here, but I like these changes. Going to test this a lot before merging.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -292,9 +292,13 @@ open class RouteController: NSObject {
         
         var nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
         
-        // If the upcoming step is a roundabout, only look at the current step for snapping.
+        // If the upcoming maneuver a sharp turn, only look at the current step for snapping.
         // Otherwise, we may get false positives from nearby step coordinates
-        if let upcomingStep = routeProgress.currentLegProgress.upComingStep, upcomingStep.maneuverDirection == .uTurn, let coordinates = routeProgress.currentLegProgress.currentStep.coordinates {
+        if let upcomingStep = routeProgress.currentLegProgress.upComingStep,
+            let initialHeading = upcomingStep.initialHeading,
+            let finalHeading = upcomingStep.finalHeading,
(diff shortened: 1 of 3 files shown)


bsudekum:
Also ref https://github.com/mapbox/mapbox-navigation-ios/pull/642

MapboxCoreNavigation/RouteController.swift
```diff
@@ -805,15 +816,9 @@ extension RouteController: CLLocationManagerDelegate {
...
-            // The objective here is to make sure the user is moving away from the maneuver location
-            // This helps on maneuvers where the difference between the exit and enter heading are similar
-            if  userAbsoluteDistance <= lastKnownUserAbsoluteDistance! {
-                routeProgress.currentLegProgress.currentStepProgress.userDistanceToManeuverLocation = userAbsoluteDistance
-            }
-            
             if routeProgress.currentLegProgress.upComingStep?.maneuverType == ManeuverType.arrive {
                 alertLevel = .arrive
-            } else if courseMatchesManeuverFinalHeading {
+            } else if courseMatchesManeuverFinalHeading || (userAbsoluteDistance > lastKnownUserAbsoluteDistance! && lastKnownUserAbsoluteDistance! > RouteControllerManeuverZoneRadius) {
```
1ec5: Are we guaranteed to have a `lastKnownUserAbsoluteDistance` here? Let’s move the `!`s up to where we define `lastKnownUserAbsoluteDistance` and consider adding a `guard` or `if` statement to handle the case where we don’t have one.

bsudekum:
@1ec5 Fixed.

PR #661, opened 2017-09-26
If audioPlayer is playing, don't also play fallback voice

bsudekum:
We're getting reports of overlapping polly and speech synth instructions. This perhaps is a sledgehammer approach to fixing the bug, however at some point this addition makes sense.

This also makes sure the audio player is ready to play before playing the audio.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -165,9 +165,9 @@ public class PollyVoiceController: RouteVoiceController {
             
             do {
                 strongSelf.audioPlayer = try AVAudioPlayer(data: data)
-                let prepared = strongSelf.audioPlayer?.prepareToPlay()
+                let prepared = strongSelf.audioPlayer?.prepareToPlay() ?? false
                 
-                guard let _ = prepared else {
+                guard prepared else {
                     strongSelf.callSuperSpeak(strongSelf.fallbackText, error: "Audio player failed to prepare")
                     return
                 }
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -122,6 +122,13 @@ public class PollyVoiceController: RouteVoiceController {
     }
     
     func callSuperSpeak(_ text: String, error: String) {
+        guard let audioPlayer = audioPlayer else {
+            super.speak(fallbackText, error: error)
+            return
+        }
+        
+        guard !audioPlayer.isPlaying else { return }
+        
         super.speak(fallbackText, error: error)
     }
     
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -156,30 +156,31 @@ public class PollyVoiceController: RouteVoiceController {
                 return
             }
             
-            guard let data = data else { return }
-            
-            DispatchQueue.main.async {
-                do {
-                    strongSelf.audioPlayer = try AVAudioPlayer(data: data)
-                    strongSelf.audioPlayer?.delegate = self
-                    
-                    if let audioPlayer = strongSelf.audioPlayer {
-                        try strongSelf.duckAudio()
-                        
-                        let readyToyPlay = audioPlayer.prepareToPlay()
-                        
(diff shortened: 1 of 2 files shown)


MapboxNavigation/PollyVoiceController.swift
```diff
@@ -158,8 +165,15 @@ public class PollyVoiceController: RouteVoiceController {
                     
                     if let audioPlayer = strongSelf.audioPlayer {
                         try strongSelf.duckAudio()
-                        audioPlayer.volume = strongSelf.volume
-                        audioPlayer.play()
+                        
+                        let readyToyPlay = audioPlayer.prepareToPlay()
```
frederoni: Nit: typo.
Also, I would assume `prepareToPlay()` should be called on a background thread but initializing an `AVAudioPlayer` in a background thread is what caused the jittering last time IIRC?
Either way, I think 
```swift
…   try AVAudioPlayer(data: data)
…  .prepareToPlay()
```
should be called before we dispatch to the main thread.

bsudekum:
@frederoni fixed

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -122,6 +122,13 @@ public class PollyVoiceController: RouteVoiceController {
     }
     
     func callSuperSpeak(_ text: String, error: String) {
+        guard let audioPlayer = audioPlayer else {
+            super.speak(fallbackText, error: error)
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/624#discussion_r141154550 suggests canceling any existing Polly-related task before handing off to AVSpeechSynthesis, which sounds like a good idea.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -149,23 +158,31 @@ public class PollyVoiceController: RouteVoiceController {
...
             
-            DispatchQueue.main.async {
-                do {
-                    strongSelf.audioPlayer = try AVAudioPlayer(data: data)
-                    strongSelf.audioPlayer?.delegate = self
+            do {
+                strongSelf.audioPlayer = try AVAudioPlayer(data: data)
+                let prepared = strongSelf.audioPlayer?.prepareToPlay()
+                
+                guard let _ = prepared else {
```
frederoni: If `prepared == false`, will this guard fall through?
I think we should unwrap `preparedToPlay() ?? false` so that the guard checks if prepared is true.

bsudekum: I think if prepared is equal to false, it would get caught in this guard. But i can change it if you'd like.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -122,17 +122,26 @@ public class PollyVoiceController: RouteVoiceController {
...
     
     func callSuperSpeak(_ text: String, error: String) {
+        pollyTask?.cancel()
+        
+        guard let audioPlayer = audioPlayer else {
+            super.speak(fallbackText, error: error)
+            return
+        }
+        
+        guard !audioPlayer.isPlaying else { return }
```
1ec5: Should we stop the audio player if some error has caused us to get into this code path?

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -149,23 +158,31 @@ public class PollyVoiceController: RouteVoiceController {
...
+                strongSelf.audioPlayer = try AVAudioPlayer(data: data)
+                let prepared = strongSelf.audioPlayer?.prepareToPlay() ?? false
+                
+                guard prepared else {
+                    strongSelf.callSuperSpeak(strongSelf.fallbackText, error: "Audio player failed to prepare")
+                    return
+                }
+                
+                DispatchQueue.main.async {
+                    let played = strongSelf.audioPlayer?.play() ?? false // Play must be called on UI thread?
```
1ec5: I think this question has been resolved, right?

MapboxNavigation/PollyVoiceController.swift
1ec5: This is happening in #669.

PR #670, opened 2017-09-29
Force kCLLocationAccuracyBestForNavigation

bsudekum:
We're getting reports that the user puck is drifting a lot when the device is unplugged. In this case, we `kCLLocationAccuracyBest`. Let's test using `kCLLocationAccuracyBestForNavigation` in all situations.

/cc @1ec5 @pveugen @frederoni @ericrwolfe

### DIFF
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -19,12 +19,6 @@ open class NavigationLocationManager: CLLocationManager {
      */
     public private(set) var isPluggedIn: Bool = false
     
-    /**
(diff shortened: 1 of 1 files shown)


bsudekum:
> Thought we were keeping the battery state logic for throttling in #402 but it would make more sense to observe the battery state in NavigationMapView when we get to that.

Ah good call. Yeah, let's move it to the proper file for #402 when needed.

1ec5:
> Yeah, let's move it to the proper file for #402 when needed.

This happened in 4f70a07975ce42a050db466d31791083f80fd38e for #402.

Issue #648, opened 2017-09-24
Weird arrow rendering (Flat)


Sealos:
Version 0.8.0

![image](https://user-images.githubusercontent.com/4553825/30782605-85ec573a-a135-11e7-86bd-e42c4a883727.png)

bsudekum:
Fixed in #671

Issue #393, opened 2017-07-16
Thicken casing of maneuver arrow on route line


1ec5:
#390 thickened the route line, so now you can see an unsightly gap between the arrow’s casing and the route line’s casing in which traffic congestion segments are sometimes visible:

![arrow](https://user-images.githubusercontent.com/1231218/28242748-065724e0-6969-11e7-9e5a-8779cc4dba63.png)

We should thicken the casing to fill in this gap. A nice side effect is that it’ll be much easier to distinguish the arrow from an ordinary road.

/cc @bsudekum

1ec5:
This is happening in #671.

bsudekum:
This has been fixed.

PR #671, opened 2017-09-29
Make upcoming maneuver arrow wider

bsudekum:
simulator screen shot oct 3 2017 4 58 34 pm
simulator screen shot oct 3 2017 5 21 36 pm

I also made the route line slighter thinner.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -5,6 +5,14 @@ import Turf
 
 typealias CongestionSegment = ([CLLocationCoordinate2D], CongestionLevel)
 
+let routeLineWidthAtZoomLevels: [Int: MGLStyleValue<NSNumber>] = [
+    10: MGLStyleValue(rawValue: 8),
+    13: MGLStyleValue(rawValue: 9),
+    16: MGLStyleValue(rawValue: 11),
+    19: MGLStyleValue(rawValue: 22),
+    22: MGLStyleValue(rawValue: 28)
+]
+
 /**
  `NavigationMapView` is a subclass of `MGLMapView` with convenience functions for adding `Route` lines to a map.
  */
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -519,6 +521,16 @@ open class NavigationMapView: MGLMapView {
             style.removeLayer(arrowCasingSymbolLayer)
         }
     }
+    
+    func mutateBaseCameraStops(multiplier: Double) -> [Int: MGLStyleValue<NSNumber>] {
```
1ec5: This method doesn’t do anything to `multipler`; it uses `multiplier` to return a new value without any side effects. So rename this method to describe what it returns, avoiding any verbs.

1ec5: I think this method would be more generally useful as an extension on `MGLStyleValue<NSNumber>` or `[Int: MGLStyleValue<NSNumber>]`. Something like this:

```swift
// Extends a dictionary of camera function stops
extension Dictionary<Key, Value> where Key: Numeric, Value: MGLStyleValue<NSNumber> {
  func muliplied(by factor: Double) -> Self {
    …
  }
}
```

in order to say something like this:

```swift
routeLineWidthAtZoomLevels.multiplied(by: 0.9)
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -498,7 +500,7 @@ open class NavigationMapView: MGLMapView {
     /**
      Removes the step arrow from the map.
      */
-    func removeArrow() {
+    public func removeArrow() {
```
1ec5: Is this intentional?

bsudekum: Yeah I think this should be made public.

frederoni:
While you're at it, can we extend the end of the maneuver arrow to go beyond the end of the maneuver location and fix #648?

MapboxNavigation/NavigationMapView.swift
frederoni: Sounds reasonable since `showRoute(_:)` is public.

1ec5:
This PR will fix #393.

bsudekum:
Updated ticket with newer images.

> While you're at it, can we extend the end of the maneuver arrow to go beyond the end of the maneuver location and fix #648?

@frederoni isn't this already happening? 
simulator screen shot oct 3 2017 5 21 42 pm

MapboxNavigation/NavigationMapView.swift
```diff
@@ -404,13 +397,12 @@ open class NavigationMapView: MGLMapView {
         
         let minimumZoomLevel: Float = 14.5
         
-        let shaftLength = max(min(50 * metersPerPoint(atLatitude: maneuverCoordinate.latitude), 50), 10)
+        let shaftLength = max(min(30 * metersPerPoint(atLatitude: maneuverCoordinate.latitude), 30), 10)
         let polyline = Polyline(routeCoordinates)
-        let shaftCoordinates = Array(polyline.trimmed(from: maneuverCoordinate, distance: -shaftLength / 2).coordinates.reversed()
```
frederoni: @bsudekum 

> isn't this already happening?

I guess using the entire shaft length here instead of half of it fixes #648?

frederoni: @bsudekum 

> isn't this already happening?

I guess using the entire shaft length here instead of half of it fixes #648?

PR #672, opened 2017-09-29
Make UIAlertController performant

bsudekum:
This action sheet has not been performant for a long time. I think by moving the creation of this action sheet to init instead of adding it when clicking a button, a lot of memory is freed up. 

This is feeling 💯  better.

/cc @1ec5 @frederoni

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -35,6 +35,8 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     
     var locationManager = CLLocationManager()
     
+    var alertController: UIAlertController!
+    
     // In this example, we show you how you can create custom UIView that is used to show the user's location.
     // Set `showCustomUserPuck` to true to view the custom user puck.
     var showCustomUserPuck = false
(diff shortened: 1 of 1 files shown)


1ec5:
> a lot of memory is freed up

I’d be surprised if this had anything to do with memory usage, but you could post a before-and-after the memory gauges to back up this claim.

Examples/Swift/ViewController.swift
```diff
@@ -53,6 +55,23 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
 
         simulationButton.isSelected = true
         startButton.isEnabled = false
+        
+        alertController = UIAlertController(title: "Start Navigation", message: "Select the navigation type", preferredStyle: .actionSheet)
+        alertController.addAction(UIAlertAction(title: "Default UI", style: .default, handler: { (action) in
```
1ec5: If building up the actions is slow, as claimed above and in #455, then we’re shifting that performance hit to startup, which in principle would hurt perceived performance. Did you determine that `addAction(_:)` is the problem by profiling?

PR #673, opened 2017-09-29
Add more control to mute button

bsudekum:
This also mutes the voice when hitting the mute button. This will allow us to perhaps add a slider or preset volume values.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -87,14 +87,14 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         NotificationCenter.default.addObserver(self, selector: #selector(alertLevelDidChange(notification:)), name: RouteControllerAlertLevelDidChange, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(pauseSpeechAndPlayReroutingDing(notification:)), name: RouteControllerWillReroute, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(didReroute(notification:)), name: RouteControllerDidReroute, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(voiceVolumeDidChange(notification:)), name: RouteVoiceControllerVoiceVolumeDidChange, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(voiceVolumeDidChange(notification:)), name: UserDefaults.didChangeNotification, object: nil)
     }
     
     func suspendNotifications() {
         NotificationCenter.default.removeObserver(self, name: RouteControllerAlertLevelDidChange, object: nil)
         NotificationCenter.default.removeObserver(self, name: RouteControllerWillReroute, object: nil)
         NotificationCenter.default.removeObserver(self, name: RouteControllerDidReroute, object: nil)
-        NotificationCenter.default.removeObserver(self, name: RouteVoiceControllerVoiceVolumeDidChange, object: nil)
+        NotificationCenter.default.removeObserver(self, name: UserDefaults.didChangeNotification, object: nil)
     }
     
     func didReroute(notification: NSNotification) {
@@ -133,12 +133,11 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     }
     
     func voiceVolumeDidChange(notification: Notification) {
-        if let voiceVolume = notification.userInfo?[RouteControllerDidChangeVoiceVolumeKey] as? Float {
-            self.volume = voiceVolume
-            if volume == 0 {
-                speechSynth.stopSpeaking(at: .word)
-                audioPlayer?.stop()
-            }
+        guard volume != NavigationSettings.shared.voiceVolume else { return }
+        volume = NavigationSettings.shared.voiceVolume
+        if volume == 0 {
+            speechSynth.stopSpeaking(at: .word)
+            audioPlayer?.stop()
         }
     }
     
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -133,7 +133,7 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     }
     
     func voiceVolumeDidChange(notification: Notification) {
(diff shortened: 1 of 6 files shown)


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -130,6 +132,16 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         }
     }
     
+    func voiceVolumeDidChange(notification: Notification) {
+        if let voiceVolume = notification.userInfo?[RouteControllerDidChangeVoiceVolumeKey] as? Float {
+            self.volume = voiceVolume
```
1ec5: Nit: drop `self.` or rename `voiceVolume` to `volume`.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -87,12 +87,14 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
         NotificationCenter.default.addObserver(self, selector: #selector(pauseSpeechAndPlayReroutingDing(notification:)), name: RouteControllerWillReroute, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(didReroute(notification:)), name: RouteControllerDidReroute, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(voiceVolumeDidChange(notification:)), name: RouteVoiceControllerVoiceVolumeDidChange, object: nil)
     }
     
     func suspendNotifications() {
         NotificationCenter.default.removeObserver(self, name: RouteControllerAlertLevelDidChange, object: nil)
         NotificationCenter.default.removeObserver(self, name: RouteControllerWillReroute, object: nil)
         NotificationCenter.default.removeObserver(self, name: RouteControllerDidReroute, object: nil)
+        NotificationCenter.default.removeObserver(self, name: RouteVoiceControllerVoiceVolumeDidChange, object: nil)
```
1ec5: No need to post and observe a custom notification; instead:

* The toggle in MapboxNavigation should set the `MBNavigationVoiceVolume` user default (named with a prefix in order to avoid collisions with user defaults set in application code).
* Core Navigation should observe for `NSNotification.Name.didChangeNotification` and handle volume changes based on the `MBNavigationVoiceVolume` user default changing.
  * Alternatively, use KVO to observe for changes to `MBNavigationVoiceVolume` specifically.

bsudekum: @1ec5 I followed your lead [here](https://github.com/mapbox/mapbox-navigation-ios/pull/673/commits/09bb80f10019faa59bf8c8c2d9fed62066200ac8), but for some reason it fails to enter navigation when this [line](https://github.com/mapbox/mapbox-navigation-ios/pull/673/commits/09bb80f10019faa59bf8c8c2d9fed62066200ac8#diff-eaaa217c63f0df828e942d653c6994ceR137) is present. Really strange.

1ec5: `NavigationSettings.voiceVolume`’s setter might need to synchronize the user defaults after setting the default, which it should probably do anyways.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -106,7 +108,7 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     func pauseSpeechAndPlayReroutingDing(notification: NSNotification) {
         speechSynth.stopSpeaking(at: .word)
         
-        guard playRerouteSound && !NavigationSettings.shared.muted else {
+        guard playRerouteSound && volume > 0 else {
```
JThramer: This is a nit-pick, and isn't even really in scope, but meh.


Generally I am a fan of the following convention.

* Use a _comma_ if the guard merely checks a list of preconditions. `This, that, and the other.`
* Use _boolean operators_ if the conditional uses non-vanilla boolean logic. `This AND that, OR this AND the other.`

MapboxNavigation/Settings.swift
```diff
@@ -2,12 +2,12 @@ import Foundation
...
 extension UserDefaults {
     fileprivate func registerNavigationDefaults() {
-        register(defaults: [NavigationSettingsKey.muted.rawValue: false])
+        register(defaults: [NavigationSettingsKey.voiceVolume.rawValue: false])
     }
 }
 
 enum NavigationSettingsKey: String {
-    case muted = "muted"
+    case voiceVolume = "voiceVolume"
```
JThramer: `case voiceVolume`

Explicitly defining the raw string value is redundant if the case name is the same as the string value.

MapboxNavigation/RouteVoiceController.swift
1ec5: Here’s where the map SDK observes changes to user defaults:

https://github.com/mapbox/mapbox-gl-native/blob/d913f4be79f2970c0ba95c297816632241eb374b/platform/ios/src/MGLMapboxEvents.m#L206

frederoni: We don't have to worry about collisions, our defaults will be stored in a different plist and not the standard. This might be the reason why observing standard defaults doesn't trigger KVO.

https://github.com/mapbox/mapbox-navigation-ios/blob/b14107790a7647496e597590846a6c7ec77817cb/MapboxNavigation/Settings.swift#L18

1ec5: Ah, yes, that would do it. Do we need to store the setting in a separate plist? It’s not uncommon for libraries to use the standard user defaults but prefix each key to avoid collisions.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -18,11 +18,6 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     var fallbackText: String!
     var announcementTimer: Timer?
     
-    /**
-     A boolean value indicating whether instructions should be announced by voice or not.
-     */
-    public var isEnabled: Bool = true
```
frederoni: This is not being used. Setting `NavigationViewController.voiceController` to nil or muting should be sufficient.

frederoni:
Refactored this PR a bit for type safe UserDefaults. A `NavigationSettingsDidChange` notification will fire for any change to the settings and looking in the userInfo will indicate what property changed.

frederoni:
Closing in favor of #936

PR #678, opened 2017-10-02
Set distanceFilter to kCLDistanceFilterNone

bsudekum:
_Put on your tinfoil hats._

I'm unsure if this will have an impact, but [many](https://github.com/mapbox/mapbox-gl-native/blob/d4c569a0187d98bca6b71671fa5daf6ee09747da/platform/ios/src/MGLLocationManager.m#L65) different dependencies in the library set a value on `distanceFilter ` that is not  `kCLDistanceFilterNone`. The thinking here is that we inherit this distance.

This PR ensures it is always `kCLDistanceFilterNone.`

### DIFF
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -33,5 +33,6 @@ open class NavigationLocationManager: CLLocationManager {
         }
         
         desiredAccuracy = kCLLocationAccuracyBestForNavigation
+        distanceFilter = kCLDistanceFilterNone
     }
 }


MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -33,5 +33,6 @@ open class NavigationLocationManager: CLLocationManager {
         }
         
         desiredAccuracy = kCLLocationAccuracyBestForNavigation
+        distanceFilter = kCLDistanceFilterNone
```
1ec5: `kCLDistanceFilterNone` is the default value of `distanceFilter`, so this probably has no effect.

PR #679, opened 2017-10-02
Improve location snapping and filtering

bsudekum:
This seeks to improve the user puck snapping feature. The goal here is to not allow bad location updates from moving the user puck off the route when the user is clearly on the route. There are two ideas here I'm moving towards:

1. Snapping should always snap. It should have to worry about filtering bad locations. We should just always snap.
1. Rerouting can occur while snapped. 

It's important though to allow movement when the user is actually moving off the route- taking a left instead of a right for example. I did add a bit of code to allow actual locations and course to be used. If the user's course/location is not what we expect it to be and their location accuracy is high, use the raw location/coordinate. 

Tested this today, it performed well.

### DIFF
--- a/MapboxCoreNavigation/Constants.swift
+++ b/MapboxCoreNavigation/Constants.swift
@@ -80,7 +80,7 @@ public var RouteControllerMaximumDistanceBeforeRecalculating: CLLocationDistance
 /**
  Accepted deviation excluding horizontal accuracy before the user is considered to be off route.
  */
-public var RouteControllerUserLocationSnappingDistance: CLLocationDistance = 20
+public var RouteControllerUserLocationSnappingDistance: CLLocationDistance = 15
 
 /**
  Threshold user must be in within to count as completing a step. One of two heuristics used to know when a user completes a step, see `RouteControllerManeuverZoneRadius`.
@@ -127,7 +127,7 @@ public var RouteControllerDeadReckoningTimeInterval:TimeInterval = 1.0
 /**
  Maximum angle the user puck will be rotated when snapping the user's course to the route line.
  */
-public var RouteControllerMaxManipulatedCourseAngle:CLLocationDirection = 25
+public var RouteControllerMaxManipulatedCourseAngle:CLLocationDirection = 30
 
 /**
  :nodoc This is used internally for debugging metrics
@@ -147,11 +147,16 @@ let RouteControllerLinkedInstructionBufferMultiplier: Double = 1.2
 let milesToMeters = 1609.34
 
 /**
- The mimimum speed value before the user is snapped to the route. This is used to overcome inaccurate course values when a user's speed is low.
+ The minimum speed value before the user's actual location can be considered over the snapped location.
  */
-public var RouteControllerMinimumSpeedThresholdForSnappingUserToRoute: CLLocationSpeed = 2
+public var RouteControllerMinSpeedForLocationSnapping: CLLocationSpeed = 3
 
 /**
  The minimum distance threshold used for giving a "Continue" type instructions.
  */
 public var RouteControllerMinDistanceForContinueInstruction: CLLocationDistance = 2_000
+
+/**
+ Distance in the opposite direction of travel before a reroute occurs.
+ */
+public var RouteControllerMinDistanceForBackwardsProgressReroute: CLLocationDistance = 50
--- a/MapboxCoreNavigation/Constants.swift
+++ b/MapboxCoreNavigation/Constants.swift
@@ -149,7 +149,7 @@ let milesToMeters = 1609.34
 /**
  The minimum speed value before the user's actual location can be considered over the snapped location.
  */
-public var RouteControllerMinSpeedForLocationSnapping: CLLocationSpeed = 3
+public var RouteControllerMinimumSpeedForLocationSnapping: CLLocationSpeed = 3
 
 /**
  The minimum distance threshold used for giving a "Continue" type instructions.
@@ -159,4 +159,9 @@ public var RouteControllerMinDistanceForContinueInstruction: CLLocationDistance 
 /**
  Distance in the opposite direction of travel before a reroute occurs.
  */
-public var RouteControllerMinDistanceForBackwardsProgressReroute: CLLocationDistance = 50
+public var RouteControllerMinimumBackupDistanceForRerouting: CLLocationDistance = 50
+
+/**
+ Minimum number of consecutive location updates moving backwards before the user is rerouted.
+ */
+public var RouteControllerMinimumNumberLocationUpdatesBackwards = 3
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -322,26 +322,25 @@ open class RouteController: NSObject {
...
+        print("course:", location.course, "speed:", location.speed, "distance from route:", snappedCoordinate.distance, "accuracy:", location.horizontalAccuracy, "difference between expected course:", differenceBetweenAngles(calculatedCourseForLocationOnStep, location.course))
         
-        guard snappedCoordinate.distance < RouteControllerUserLocationSnappingDistance else {
-            return location
+        
+        if location.course >= 0 || location.speed >= RouteControllerMinimumSpeedThresholdForSnappingUserToRoute {
+            if differenceBetweenAngles(calculatedCourseForLocationOnStep, location.course) > RouteControllerMaxManipulatedCourseAngle && location.horizontalAccuracy < 20 {
+                userCourse = location.course
+                
+                if snappedCoordinate.distance > RouteControllerUserLocationSnappingDistance && location.horizontalAccuracy < 20 {
```
bsudekum: You can only have the actual raw location if the user course is also different than what we think it is.

MapboxCoreNavigation/Constants.swift
```diff
@@ -147,9 +147,9 @@ let RouteControllerLinkedInstructionBufferMultiplier: Double = 1.2
 let milesToMeters = 1609.34
 
 /**
- The mimimum speed value before the user is snapped to the route. This is used to overcome inaccurate course values when a user's speed is low.
+ The minimum speed value before the user's actual location can be considered over the snapped location.
  */
-public var RouteControllerMinimumSpeedThresholdForSnappingUserToRoute: CLLocationSpeed = 2
+public var RouteControllerMinimumSpeedThresholdForUsersRealLocationInseatOfSnapped: CLLocationSpeed = 3
```
1ec5: This constant name is getting long even by [Objective-C standards](https://github.com/Quotation/LongestCocoa). How about `RouteControllerMaximumSpeedForLocationSnapping`?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -287,7 +287,7 @@ open class RouteController: NSObject {
      This property contains a `CLLocation` object located along the route line near the most recently received user location. This property is set to `nil` if the route controller is unable to snap the user’s location to the route line for some reason.
      */
     public var location: CLLocation? {
-        guard let location = rawLocation, userIsOnRoute(location) else { return nil }
```
bsudekum: Location should not worry about whether the user is on the route. These ideas should be decoupled.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -324,26 +324,22 @@ open class RouteController: NSObject {
...
-            return location
-        }
-
-        let course = averageRelativeAngle <= RouteControllerMaximumAllowedDegreeOffsetForTurnCompletion ? absoluteDirection : location.course
+        var userCourse = calculatedCourseForLocationOnStep
+        var userCoordinate = snappedCoordinate.coordinate
         
-        guard snappedCoordinate.distance < RouteControllerUserLocationSnappingDistance else {
-            return location
+        if location.course >= 0 || location.speed >= RouteControllerMinimumSpeedThresholdForUsersRealLocationInseatOfSnapped {
```
bsudekum: Most of the time we will return a snapped location (line 342):

```swift
return CLLocation(coordinate: userCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: userCourse, speed: location.speed, timestamp: location.timestamp)
```

However, if the user location is accurate and the user's course is not relatively close to what we think it should be, we'll allow the puck to not snap.

bsudekum:
Did many drives with this, it's feeling like a great improvement.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -567,8 +563,12 @@ extension RouteController: CLLocationManagerDelegate {
         if let coordinates = routeProgress.currentLegProgress.currentStep.coordinates {
             let userDistanceToManeuver = Polyline(coordinates).distance(from: location.coordinate)
             
-            guard recentDistancesFromManeuver.count <= 3 else {
-                return false
+            let totalDistances = recentDistancesFromManeuver.reduce(0) { $0 + $1 }
```
1ec5: Nit: this is equivalent to `reduce(0, +)`, because operators like `+` can be used wherever a function can be used.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -567,8 +563,12 @@ extension RouteController: CLLocationManagerDelegate {
...
+            
+            if totalDistances < RouteControllerMinDistanceForBackwardsProgressReroute {
+                guard recentDistancesFromManeuver.count <= 3 else {
```
1ec5: Pull this number out into a constant.

MapboxCoreNavigation/Constants.swift
```diff
@@ -147,11 +147,16 @@ let RouteControllerLinkedInstructionBufferMultiplier: Double = 1.2
 let milesToMeters = 1609.34
 
 /**
- The mimimum speed value before the user is snapped to the route. This is used to overcome inaccurate course values when a user's speed is low.
+ The minimum speed value before the user's actual location can be considered over the snapped location.
  */
-public var RouteControllerMinimumSpeedThresholdForSnappingUserToRoute: CLLocationSpeed = 2
+public var RouteControllerMinSpeedForLocationSnapping: CLLocationSpeed = 3
```
1ec5: Don’t abbreviate “minimum” as “min”.

MapboxCoreNavigation/Constants.swift
```diff
@@ -147,11 +147,16 @@ let RouteControllerLinkedInstructionBufferMultiplier: Double = 1.2
...
 
 /**
  The minimum distance threshold used for giving a "Continue" type instructions.
  */
 public var RouteControllerMinDistanceForContinueInstruction: CLLocationDistance = 2_000
+
+/**
+ Distance in the opposite direction of travel before a reroute occurs.
+ */
+public var RouteControllerMinDistanceForBackwardsProgressReroute: CLLocationDistance = 50
```
1ec5: `RouteControllerMinimumBackupDistanceForRerouting`

MapboxCoreNavigation/RouteController.swift
1ec5: Please add these comments to the code.

bsudekum:
@1ec5 updated

PR #683, opened 2017-10-03
Simulate expected speed

### LINKED ISSUE #440: SimulatedLocationManager should simulate expected speed
SimulatedLocationManager appears to assume a constant speed of 30 meters per second (108 km/h, 67 mph), subject to turn penalties. Instead, it should account for the expected speed along each step or, even better, along each segment according to the route leg’s `segmentExpectedTravelTimes` property. (These per-segment expected travel times could already reflect turn penalties, but I’m not sure.) This would help us reproduce the performance issues that motivated #370 in simulation mode just as in the wild.

/cc @frederoni @ericrwolfe

frederoni:
Fixes #440 - Simulate expected speed

Falls back to turn penalty based speed if `RouteLeg.segmentSpeeds` is missing.

@bsudekum @1ec5 👀

### DIFF
--- a/MapboxCoreNavigation/NavigationRouteOptions.swift
+++ b/MapboxCoreNavigation/NavigationRouteOptions.swift
@@ -22,7 +22,7 @@ open class NavigationRouteOptions: RouteOptions {
         }, profileIdentifier: profileIdentifier)
         includesSteps = true
         routeShapeResolution = .full
-        attributeOptions = [.congestionLevel, .expectedTravelTime]
+        attributeOptions = [.congestionLevel, .expectedTravelTime, .speed]
         includesExitRoundaboutManeuver = true
     }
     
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -22,7 +22,7 @@ open class NavigationRouteOptions: RouteOptions {
         }, profileIdentifier: profileIdentifier)
         includesSteps = true
         routeShapeResolution = .full
-        attributeOptions = [.congestionLevel, .expectedTravelTime]
+        attributeOptions = [.congestionLevel, .expectedTravelTime, .speed]
```
1ec5: This is a large increase in the response payload for a feature that's normally only used in debugging. The caller should take care of adding this option if simulation is enabled.

/cc @frederoni @bsudekum

1ec5: Besides, we can derive the speed from the distance between two nodes and the expected travel time along that segment – no additional data needed.

PR #684, opened 2017-10-04
Optimize simulation

frederoni:
Addresses https://github.com/mapbox/mapbox-navigation-ios/pull/683#discussion_r142565325

Instead of increasing the response payload by adding speed, derive it from expected travel time.

@1ec5 @bsudekum 👀

### DIFF
--- a/MapboxCoreNavigation/SimulatedLocationManager.swift
+++ b/MapboxCoreNavigation/SimulatedLocationManager.swift
@@ -121,11 +121,15 @@ public class SimulatedLocationManager: NavigationLocationManager {
         let distance = min(max(distanceToClosest, 10), safeDistance)
         let coordinatesNearby = polyline.trimmed(from: newCoordinate, distance: 100).coordinates
         
-        // Simulate expected speed
-        if let speeds = routeProgress?.currentLeg.segmentSpeeds,
-            let closestCoordinateOnRoute = Polyline(routeProgress!.route.coordinates!).closestCoordinate(to: lookAheadCoordinate),
-            closestCoordinateOnRoute.index < speeds.count {
-            currentSpeed = speeds[closestCoordinateOnRoute.index]
+        // Simulate speed based on expected segment travel time
+        if let expectedSegmentTravelTimes = routeProgress?.currentLeg.expectedSegmentTravelTimes,
+            let coordinates = routeProgress?.route.coordinates,
+            let closestCoordinateOnRoute = Polyline(routeProgress!.route.coordinates!).closestCoordinate(to: newCoordinate),
+            let nextCoordinateOnRoute = coordinates.after(element: coordinates[closestCoordinateOnRoute.index]),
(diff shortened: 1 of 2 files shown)


bsudekum:
Looks good. Noting that it looks a little less smooth with this change. 
gif2

We should keep this in mind when deciding upon a fps for https://github.com/mapbox/mapbox-navigation-ios/pull/402

MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -121,11 +121,15 @@ public class SimulatedLocationManager: NavigationLocationManager {
         let distance = min(max(distanceToClosest, 10), safeDistance)
         let coordinatesNearby = polyline.trimmed(from: newCoordinate, distance: 100).coordinates
         
-        // Simulate expected speed
-        if let speeds = routeProgress?.currentLeg.segmentSpeeds,
-            let closestCoordinateOnRoute = Polyline(routeProgress!.route.coordinates!).closestCoordinate(to: lookAheadCoordinate),
-            closestCoordinateOnRoute.index < speeds.count {
-            currentSpeed = speeds[closestCoordinateOnRoute.index]
+        // Simulate speed based on expected segment travel time
+        if let expectectedSegmentTravelTimes = routeProgress?.currentLeg.expectedSegmentTravelTimes,
```
1ec5: Typo: “expectected”.

MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -121,11 +121,15 @@ public class SimulatedLocationManager: NavigationLocationManager {
...
+            let coordinates = routeProgress?.route.coordinates,
+            let closestCoordinateOnRoute = Polyline(routeProgress!.route.coordinates!).closestCoordinate(to: newCoordinate),
+            let nextCoordinateOnRoute = coordinates.after(element: coordinates[closestCoordinateOnRoute.index]),
+            let time = expectectedSegmentTravelTimes.optional[closestCoordinateOnRoute.index] {
+            
+            let distance = coordinates[closestCoordinateOnRoute.index].distance(to: nextCoordinateOnRoute)
+            currentSpeed = distance / time
```
1ec5: Any idea why this leads to choppier simulation than the speed attribute? I thought this is the very definition of the API’s speed calculation.

1ec5: Anyways, if it’s a problem, we can switch to the other suggestion in https://github.com/mapbox/mapbox-navigation-ios/pull/683#discussion_r142565325, which is to have ViewController add the speed attribute when the simulation switch is on (and document that developers should do likewise when enabling simulation).

frederoni: I'm seeing roughly the same results. @bsudekum, can you try again using a device or a different route?

bsudekum:
👍 , looks better

PR #689, opened 2017-10-05
Feedback 2.0

frederoni:
WIP Feedback


MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -321,37 +321,49 @@ class FixedLengthQueue<T> {
...
+    }
+    
+    public var description: String? {
+        didSet {
+            eventDictionary["description"] = description
+        }
+    }
+    
+    public var coordinate: CLLocationCoordinate2D? {
+        guard let latitude = eventDictionary["lat"] as? CLLocationDegrees, let longitude = eventDictionary["lng"] as? CLLocationDegrees else { return nil }
```
JThramer: This was a quick-and-dirty way for me to get at the underlying values in the model. My suggestion for production would be that we represent this data in our model class (or protocol) with concrete properties, and synthesize their inclusion into a dictionary model.

JThramer: This was a "quick and dirty" way for me to get at the underlying data in the model. I do feel like ideally, these data points would be represented (in either a model class or protocol) by concrete properties, and we would synthesize the "raw data" dictionary, rather than the other way around.

PR #704, opened 2017-10-12
Hold altitude user changes

bsudekum:
/cc @frederoni


frederoni:
Cherry-picked into #402

PR #718, opened 2017-10-16
Prevent spinning when user is not moving

bsudekum:
I noticed quite a few times this weekend the user puck would spin when I was not moving at stop lights. This should prevent this while still allowing the actual course while rerouting.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -325,7 +325,7 @@ open class RouteController: NSObject {
         var userCourse = calculatedCourseForLocationOnStep
         var userCoordinate = snappedCoordinate.coordinate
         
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -325,7 +325,7 @@ open class RouteController: NSObject {
         var userCourse = calculatedCourseForLocationOnStep
         var userCoordinate = snappedCoordinate.coordinate
         
-        if location.course >= 0 || location.speed >= RouteControllerMinimumSpeedForLocationSnapping {
+        if location.course >= 0 && location.speed >= RouteControllerMinimumSpeedForLocationSnapping {
```
frederoni: Wouldn't this result in even less course snapping?

bsudekum: Nope, this if statement is for when the user course/location is unsnapped.

PR #719, opened 2017-10-16
Add option to add audio recording to feedback event

bsudekum:
Still todo:

- [x] Put behind a flag
- [x] Make starting of animation smoother
- [x] Add text to top of feedback that recording is happening
- [x] Refine writing/reading from disk

/cc @1ec5 @ericrwolfe @frederoni @JThramer

### DIFF
--- a/MapboxNavigation/FeedbackViewController.swift
+++ b/MapboxNavigation/FeedbackViewController.swift
@@ -17,6 +17,8 @@ class FeedbackViewController: UIViewController, UIGestureRecognizerDelegate, AVA
     var sections = [FeedbackSection]()
     let cellReuseIdentifier = "collectionViewCellId"
     
+    var allowRecordedAudioFeedback = false
+    
     typealias SendFeedbackHandler = (FeedbackItem) -> ()
     
     var sendFeedbackHandler: SendFeedbackHandler?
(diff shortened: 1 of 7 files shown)


bsudekum:
This is ready for review.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -70,6 +108,69 @@ class FeedbackViewController: UIViewController {
...
+        }
+    }
+    
+    func finishRecording() {
+        audioRecorder?.stop()
+        audioRecorder = nil
+    }
+    
+    func getDocumentsDirectory() -> URL {
+        let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
```
frederoni: We shouldn't write to the app’s document directory. How about `cachesDirectory`?

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -70,6 +108,69 @@ class FeedbackViewController: UIViewController {
...
+            print(error.localizedDescription)
+        }
+    }
+    
+    func finishRecording() {
+        audioRecorder?.stop()
+        audioRecorder = nil
+    }
+    
+    func getDocumentsDirectory() -> URL {
```
frederoni: `Get` in method names should only be used when getting values indirectly.
Can we change this to an extension on `URL.audioCacheDirectory`?

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -47,18 +72,31 @@ class FeedbackViewController: UIViewController {
...
         sections = [
             [accident, hazard],
             [roadClosed, unallowedTurn],
             [routingError, other]
         ]
+        
+        guard allowRecordedAudioFeedback else { return }
+        
+        recordingSession = AVAudioSession.sharedInstance()
+        recordingSession?.requestRecordPermission() { [unowned self] allowed in
```
frederoni: FYI this will crash if the `Privacy - Microphone Usage Description and value`-key is omitted but that's fine since it's opt-in.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -47,18 +72,31 @@ class FeedbackViewController: UIViewController {
...
+            if allowed {
+                let lpgr : UILongPressGestureRecognizer = UILongPressGestureRecognizer(target: self, action: #selector(self.didTapCell(_:)))
```
frederoni: Nit: No type is being inferred here so we can remove the excessive `: UILongPressGestureRecognizer`

bsudekum:
Strange, this PR is not updating with recent commits to fix above feedback. Can see diff here https://github.com/mapbox/mapbox-navigation-ios/compare/record-audio

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -70,6 +115,64 @@ class FeedbackViewController: UIViewController {
...
+        let settings = [
+            AVFormatIDKey: Int(kAudioFormatMPEG4AAC),
+            AVSampleRateKey: 12000,
+            AVNumberOfChannelsKey: 1,
+            AVEncoderAudioQualityKey: AVAudioQuality.low.rawValue
+        ]
+        
+        do {
+            audioRecorder = try AVAudioRecorder(url: audioFile, settings: settings)
+            audioRecorder?.delegate = self
```
bsudekum: I think it might not be necessary here to set the delegate.

frederoni: Correct, we can remove this and the conformance to AVAudioRecorderDelegate.

PR #724, opened 2017-10-17
Prefetch audio instructions

bsudekum:
_Prerequisite https://github.com/mapbox/mapbox-navigation-ios/pull/614_

It's relatively dumb right now, when we fire the notification to say something:

1. Check if the instruction is a key on `RouteProgress.spokenInstructionsForRoute` and has data.
1. If so, use this data to give the instruction
1. If not, first say the instruction using the normal methods
1. After we say it, we then look and see if the user has any upcoming steps
1. If so, get all instructions for step, download the data and add it to `RouteProgress.spokenInstructionsForRoute`
1. Repeat

/cc @1ec5 @frederoni @ericrwolfe @JThramer

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -31,6 +31,11 @@ public class PollyVoiceController: RouteVoiceController {
      */
     public var timeoutIntervalForRequest:TimeInterval = 2
     
+    /**
+     Number of steps ahead of the current step to cache spoken instructions.
+     */
+    public var stepsAheadToCache: Int = 3
+    
     var pollyTask: URLSessionDataTask?
     
     let sessionConfiguration = URLSessionConfiguration.default
@@ -65,13 +70,19 @@ public class PollyVoiceController: RouteVoiceController {
         audioPlayer?.stop()
         startAnnouncementTimer()
         
-        if let upcomingStep = routeProgresss.currentLegProgress.upComingStep, let instructions = upcomingStep.instructionsSpokenAlongStep {
+        for (stepIndex, step) in routeProgresss.currentLegProgress.leg.steps.suffix(from: routeProgresss.currentLegProgress.stepIndex).enumerated() {
+            let adjustedStepIndex = stepIndex + routeProgresss.currentLegProgress.stepIndex
+            
+            guard adjustedStepIndex < routeProgresss.currentLegProgress.stepIndex + 3 else { continue }
+            guard let instructions = step.instructionsSpokenAlongStep else { continue }
+            
             for instruction in instructions {
                 guard spokenInstructionsForRoute[instruction.ssmlText] == nil else { continue }
                 
                 cacheSpokenInstruction(instruction: instruction.ssmlText)
             }
         }
+        
         
         guard spokenInstructionsForRoute[instruction] == nil else {
             sayInStruction(for: spokenInstructionsForRoute[instruction]!)
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -85,7 +85,7 @@ public class PollyVoiceController: RouteVoiceController {
         
         
         guard spokenInstructionsForRoute[instruction] == nil else {
-            sayInstruction(for: spokenInstructionsForRoute[instruction]!)
+            play(spokenInstructionsForRoute[instruction]!)
             return
         }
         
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -140,6 +140,8 @@ open class RouteProgress: NSObject {
             congestionTravelTimesSegmentsByStep.append(congestionTravelTimesSegmentsByLeg)
         }
     }
+    
+    public var spokenInstructionsForRoute: [String: Data] = [:]
```
frederoni: Can we move this to the PollyVoiceController?

bsudekum: Yeah I guess it doesn’t need to be stored on routeprogress

bsudekum:
@1ec5 @frederoni could I get another review?

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -57,14 +68,31 @@ public class PollyVoiceController: RouteVoiceController {
         
         pollyTask?.cancel()
         audioPlayer?.stop()
+        startAnnouncementTimer()
+        
+        for (stepIndex, step) in routeProgresss.currentLegProgress.leg.steps.suffix(from: routeProgresss.currentLegProgress.stepIndex).enumerated() {
+            let adjustedStepIndex = stepIndex + routeProgresss.currentLegProgress.stepIndex
+            
+            guard adjustedStepIndex < routeProgresss.currentLegProgress.stepIndex + 3 else { continue }
```
frederoni: Should this be using the `stepsAheadToCache` instead of 3?

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -172,29 +207,62 @@ public class PollyVoiceController: RouteVoiceController {
...
-                strongSelf.callSuperSpeak(strongSelf.fallbackText, error: error.localizedDescription)
             }
+            
+            strongSelf.cachePollyTask?.resume()
+            
+            return nil
+        }
+    }
+    
+    func sayInStruction(for data: Data) {
```
frederoni: Typo: s/InSt/Inst

MapboxNavigation/PollyVoiceController.swift
bsudekum: Oh yep.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -31,11 +31,21 @@ public class PollyVoiceController: RouteVoiceController {
      */
     public var timeoutIntervalForRequest:TimeInterval = 2
     
+    /**
+     Number of steps ahead of the current step to cache spoken instructions.
+     */
+    public var stepsAheadToCache: Int = 3
```
1ec5: * “Caching” implies reusing things after the first time they’re needed. A more appropriate word for this feature is “prefetching”.
* The name of this property talks about steps instead of spoken instructions – are we fetching individual steps from the Directions API and caching them now?
* By beginning a name with a plural noun, you’re implying a type of `Array`.

How about `countOfStepsForPrefetchingInstructions`?

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -31,11 +31,21 @@ public class PollyVoiceController: RouteVoiceController {
...
+    
     var pollyTask: URLSessionDataTask?
     
     let sessionConfiguration = URLSessionConfiguration.default
     var urlSession: URLSession
     
+    var cacheURLSession: URLSession
+    var cachePollyTask: URLSessionDataTask?
+    
+    var spokenInstructionsForRoute: [String: Data] = [:]
```
1ec5: Use an NSCache instead of a dictionary: #759.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -57,14 +68,31 @@ public class PollyVoiceController: RouteVoiceController {
...
+        startAnnouncementTimer()
+        
+        for (stepIndex, step) in routeProgresss.currentLegProgress.leg.steps.suffix(from: routeProgresss.currentLegProgress.stepIndex).enumerated() {
+            let adjustedStepIndex = stepIndex + routeProgresss.currentLegProgress.stepIndex
+            
+            guard adjustedStepIndex < routeProgresss.currentLegProgress.stepIndex + stepsAheadToCache else { continue }
+            guard let instructions = step.instructionsSpokenAlongStep else { continue }
+            
+            for instruction in instructions {
+                guard spokenInstructionsForRoute[instruction.ssmlText] == nil else { continue }
```
1ec5: This can be written more clearly as:

```swift
for instruction in instructions where spokenInstructionsForRoute[instruction.ssmlText] != nil { … }
```

PR #730, opened 2017-10-18
Swift Example App - "Multiple Stops" bug

JThramer:
Fixing issue in Example swift app where a "Multiple Stops" action gets added multiple times and is selectable at inappropriate times.

Fun screenie:
img_0276


JThramer:
CC @bsudekum @frederoni

PR #734, opened 2017-10-19
Add more telem keys

bsudekum:
- [x]  `locationEngine`
- [x] `percentTimeInForeground`
- [x] `percentTimeInPortrait`
- [x] `absoluteDistanceToDestination`
- [x] Added `navigation.fasterRoute` event

### DIFF
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -37,6 +37,8 @@ struct EventDetails {
     var applicationState: UIApplicationState
     var userAbsoluteDistanceToDestination: CLLocationDistance?
     var locationEngine: CLLocationManager.Type?
+    var percentTimeInPortrait: Double
+    var percentTimeInForeground: Double
     
     init(routeController: RouteController, session: SessionState) {
         created = Date()
@@ -91,6 +93,9 @@ struct EventDetails {
         if let manager = routeController.locationManager {
             locationEngine = type(of: manager)
         }
+        
+        percentTimeInPortrait = session.timeSpentInPortrait / (session.timeSpentInPortrait + session.timeSpentInLandscape)
+        percentTimeInForeground = session.timeSpentInForeground / (session.timeSpentInForeground + session.timeSpentInBackground)
     }
     
     var eventDictionary: [String: Any] {
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -182,6 +182,12 @@ extension MMEEventsManager {
         return eventDictionary
     }
     
+    func navigationFoundFasterRoute(routeController: RouteController) -> [String: Any] {
+        var eventDictionary = self.addDefaultEvents(routeController: routeController)
+        eventDictionary["event"] = MMEEventTypeNavigationArrive
+        return eventDictionary
+    }
+    
     func navigationFeedbackEvent(routeController: RouteController, type: FeedbackType, description: String?) -> [String: Any] {
         var eventDictionary = self.addDefaultEvents(routeController: routeController)
         eventDictionary["event"] = MMEEventTypeNavigationFeedback
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -184,7 +184,7 @@ extension MMEEventsManager {
     
     func navigationFoundFasterRoute(routeController: RouteController) -> [String: Any] {
         var eventDictionary = self.addDefaultEvents(routeController: routeController)
-        eventDictionary["event"] = MMEEventTypeNavigationArrive
+        eventDictionary["event"] = FasterRouteFoundEvent
         return eventDictionary
     }
     
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -101,7 +101,7 @@ struct EventDetails {
         } else if UIDevice.current.orientation == .landscapeRight || UIDevice.current.orientation == .landscapeLeft {
             totalTimeInLandscape += abs(session.lastTimeInLandscape.timeIntervalSinceNow)
         }
-        percentTimeInPortrait = totalTimeInPortrait / (totalTimeInPortrait + totalTimeInLandscape)
+        percentTimeInPortrait = totalTimeInPortrait + totalTimeInLandscape == 0 ? 1 : totalTimeInPortrait / (totalTimeInPortrait + totalTimeInLandscape)
         
         var totalTimeInForeground = session.timeSpentInForeground
         var totalTimeInBackground = session.timeSpentInBackground
@@ -110,7 +110,7 @@ struct EventDetails {
         } else {
             totalTimeInBackground += abs(session.lastTimeInBackground.timeIntervalSinceNow)
         }
-        percentTimeInForeground = totalTimeInPortrait / (totalTimeInPortrait + totalTimeInLandscape)
+        percentTimeInForeground = totalTimeInPortrait + totalTimeInLandscape == 0 ? 1 : totalTimeInPortrait / (totalTimeInPortrait + totalTimeInLandscape)
     }
     
     var eventDictionary: [String: Any] {
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -93,8 +93,8 @@ struct EventDetails {
         locationEngine = routeController.locationManager
         
         if let departureDate = session.departureTimestamp {
-            percentTimeInPortrait = 1 - abs(session.timeInPortrait / departureDate.timeIntervalSinceNow)
-            percentTimeInForeground = 1 - abs(session.timeInForeground / departureDate.timeIntervalSinceNow)
+            percentTimeInPortrait = (1 - abs(session.timeInPortrait / departureDate.timeIntervalSinceNow)) * 100
+            percentTimeInForeground = (1 - abs(session.timeInForeground / departureDate.timeIntervalSinceNow)) * 100
         }
     }
     
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -173,7 +173,7 @@ extension MMEEventsManager {
         return eventDictionary
     }
     
-    func navigationFoundFasterRoute(routeController: RouteController) -> [String: Any] {
+    func navigationFoundFasterRouteEvent(routeController: RouteController) -> [String: Any] {
         var eventDictionary = self.addDefaultEvents(routeController: routeController)
         eventDictionary["event"] = FasterRouteFoundEvent
         return eventDictionary
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -94,8 +94,23 @@ struct EventDetails {
             locationEngine = type(of: manager)
         }
         
-        percentTimeInPortrait = session.timeSpentInPortrait / (session.timeSpentInPortrait + session.timeSpentInLandscape)
-        percentTimeInForeground = session.timeSpentInForeground / (session.timeSpentInForeground + session.timeSpentInBackground)
+        var totalTimeInPortrait = session.timeSpentInPortrait
+        var totalTimeInLandscape = session.timeSpentInLandscape
+        if UIDevice.current.orientation == .portrait {
+            totalTimeInPortrait += abs(session.lastTimeInPortrait.timeIntervalSinceNow)
+        } else if UIDevice.current.orientation == .landscapeRight || UIDevice.current.orientation == .landscapeLeft {
+            totalTimeInLandscape += abs(session.lastTimeInLandscape.timeIntervalSinceNow)
+        }
+        percentTimeInPortrait = totalTimeInPortrait / (totalTimeInPortrait + totalTimeInLandscape)
+        
+        var totalTimeInForeground = session.timeSpentInForeground
+        var totalTimeInBackground = session.timeSpentInBackground
+        if UIApplication.shared.applicationState == .active {
+            totalTimeInForeground += abs(session.lastTimeInForeground.timeIntervalSinceNow)
+        } else {
+            totalTimeInBackground += abs(session.lastTimeInBackground.timeIntervalSinceNow)
+        }
+        percentTimeInForeground = totalTimeInPortrait / (totalTimeInPortrait + totalTimeInLandscape)
     }
     
     var eventDictionary: [String: Any] {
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -780,10 +806,17 @@ struct SessionState {
...
+    
+    var timeInForeground: TimeInterval
+    var timeInPortrait: TimeInterval
+    var previousForegroundChange: Date?
+    var previousOrientationChangeToPortrait: Date?
 
     init(currentRoute: Route, originalRoute: Route) {
         self.currentRoute = currentRoute
         self.originalRoute = originalRoute
+        self.timeInForeground = 0
```
ericrwolfe: Nit: these can be moved to the property declaration:
`var timeInForeground: TimeInterval = 0`

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -164,6 +182,12 @@ extension MMEEventsManager {
         return eventDictionary
     }
     
+    func navigationFoundFasterRoute(routeController: RouteController) -> [String: Any] {
```
ericrwolfe: Nit: `navigationFoundFasterRoute` -> `navigationFoundFasterRouteEvent`

MapboxCoreNavigation/RouteController.swift
```diff
@@ -804,6 +837,11 @@ extension RouteController {
         events.enqueueEvent(withName: MMEEventTypeNavigationCancel, attributes: events.navigationCancelEvent(routeController: self))
         events.flush()
     }
+    
+    func sendFasterRouteEvent() {
+        events.enqueueEvent(withName: FasterRouteFoundEvent, attributes: events.navigationCancelEvent(routeController: self))
```
ericrwolfe: `events.navigationCancelEvent` -> `events. navigationFoundFasterRouteEvent`

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -164,6 +182,12 @@ extension MMEEventsManager {
...
+        var eventDictionary = self.addDefaultEvents(routeController: routeController)
+        eventDictionary["event"] = FasterRouteFoundEvent
```
ericrwolfe: We should also add `step`, `newDistanceRemaining`, `newDurationRemaining`, and `newGeometry` as we do in the reroute event.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -135,6 +149,10 @@ struct EventDetails {
         modifiedEventDictionary["batteryPluggedIn"] = batteryPluggedIn
         modifiedEventDictionary["batteryLevel"] = batteryLevel
         modifiedEventDictionary["applicationState"] = applicationState.telemetryString
+        modifiedEventDictionary["userAbsoluteDistanceToDestination"] = userAbsoluteDistanceToDestination
+        modifiedEventDictionary["locationEngine"] = locationEngine is SimulatedLocationManager ? "SimulatedLocationManager" : locationEngine is NavigationLocationManager ? "NavigationLocationManager" : locationEngine.description
```
ericrwolfe: `locationEngine.description` won't return what you expect here. Also, locationEngine might be a subclass of NavigationManager.

I tested in a playground and `String.init(describing: type(of: locationEngine))` seems to work.

MapboxCoreNavigation/MMEEventsManager.swift
ericrwolfe: We also may want to have this leverage the `RerouteEvent.update` function: https://github.com/mapbox/mapbox-navigation-ios/blob/10ff07f2baff2574595a92ba425de63a4939a022/MapboxCoreNavigation/MMEEventsManager.swift#L381-L389

MapboxCoreNavigation/RouteController.swift
```diff
@@ -843,6 +881,15 @@ extension RouteController {
 
         return event.id.uuidString
     }
+    
+    func enqueueFoundFasterRouteEvent() -> String {
+        let eventDictionary = events.navigationFoundFasterRoute(routeController: self)
+        let event = FeedbackEvent(timestamp: Date(), eventDictionary: eventDictionary)
```
ericrwolfe: Rather than a `FeedbackEvent`, we should either reuse or subclass `RerouteEvent`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -804,6 +837,11 @@ extension RouteController {
         events.enqueueEvent(withName: MMEEventTypeNavigationCancel, attributes: events.navigationCancelEvent(routeController: self))
         events.flush()
     }
+    
+    func sendFasterRouteEvent() {
```
ericrwolfe: I can't find anything that calls this.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -408,8 +430,12 @@ extension RouteController {
...
     }
-
+    
     func didReroute(notification: NSNotification) {
+        if let _ = notification.userInfo?[FasterRouteFoundEvent] as? Bool {
+            _ = enqueueFoundFasterRouteEvent()
+        }
+        
         if let lastReroute = outstandingFeedbackEvents.map({$0 as? RerouteEvent }).last {
             lastReroute?.update(newRoute: routeProgress.route)
```
ericrwolfe: This also needs to be called for faster routes found.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -843,6 +881,15 @@ extension RouteController {
 
         return event.id.uuidString
     }
+    
+    func enqueueFoundFasterRouteEvent() -> String {
+        let eventDictionary = events.navigationFoundFasterRoute(routeController: self)
+        let event = FeedbackEvent(timestamp: Date(), eventDictionary: eventDictionary)
+        
+        outstandingFeedbackEvents.append(event)
```
ericrwolfe: Rather than enqueuing this, I think we can just send this immediately after updating with the new route information (`step`, `newGeometry`, etc), like an arrive event. This event doesn't need to collect locations so doesn't need to be queued and we don't need to set or return a feedback id.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -267,7 +272,24 @@ open class RouteController: NSObject {
...
+        sessionState.previousOrientationChangeToPortrait = Date()
+    }
 
+    
+    func didChangeApplicationState() {
+        if let previousForegroundChange = sessionState.previousForegroundChange, UIApplication.shared.applicationState == .active {
+            sessionState.timeInForeground += previousForegroundChange.timeIntervalSinceNow
+        }
+        
+        sessionState.previousOrientationChangeToPortrait = Date()
```
ericrwolfe: Should this be `previousForegroundChange`?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -267,7 +272,24 @@ open class RouteController: NSObject {
         locationManager.stopUpdatingLocation()
         locationManager.stopUpdatingHeading()
     }
+    
+    func didChangeOrientation() {
+        if let previousOrientationChangeToPortrait = sessionState.previousOrientationChangeToPortrait, UIDevice.current.orientation == .portrait {
+            sessionState.timeInPortrait += previousOrientationChangeToPortrait.timeIntervalSinceNow
+        }
+        
+        sessionState.previousOrientationChangeToPortrait = Date()
```
ericrwolfe: What happens if I'm in portrait, go into the background, rotate to landscape, and reopen the app?

MapboxCoreNavigation/RouteController.swift
bsudekum: Since `didReroute` is called after finding a faster route, won't this code path be hit?

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -35,6 +35,8 @@ struct EventDetails {
     var batteryPluggedIn: Bool
     var batteryLevel: Float
     var applicationState: UIApplicationState
+    var userAbsoluteDistanceToDestination: CLLocationDistance?
+    var locationEngine: CLLocationManager
```
frederoni: Let's make this a `CLLocationManager.Type` so we don't copy the whole location manager and keep an extra ref to it.

bsudekum:
@ericrwolfe this is ready for another round of 👁

MapboxCoreNavigation/RouteController.swift
```diff
@@ -244,11 +244,38 @@ open class RouteController: NSObject {
...
+        NotificationCenter.default.addObserver(self, selector: #selector(didChangeApplicationState), name: .UIApplicationDidEnterBackground, object: nil)
     }
 
     func suspendNotifications() {
         NotificationCenter.default.removeObserver(self)
     }
+    
+    func didChangeOrientation() {
+        if UIDevice.current.orientation == .portrait {
+            sessionState.timeSpentInPortrait += abs(sessionState.lastTimeInLandscape.timeIntervalSinceNow)
```
ericrwolfe: Shouldn't the logic here be switched with the landscape if block? In this case, we just changed _to_ portrait, so we'd want to increment the time spent in landscape.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -407,9 +434,15 @@ extension RouteController {
 
     func willReroute(notification: NSNotification) {
         _ = enqueueRerouteEvent()
+        sendFasterRouteEvent()
```
ericrwolfe: This sends a faster route event every time a user is about to reroute.

Shouldn't we send this from the faster reroute callback, only if a new route is presented to the user?

ericrwolfe: In fact, this is called before `enqueueFoundFasterRouteEvent()`, which is called from `didReroute()`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -791,19 +839,30 @@ struct SessionState {
 extension RouteController {
     // MARK: Sending events
     func sendDepartEvent() {
+        didChangeOrientation()
+        didChangeApplicationState()
```
ericrwolfe: Won't calling these here cause the `didChangeX` callbacks to occur, even if the orientation or landscape didn't change. For example, `didChangeOrientation()` would be called for changing to `portrait` from `portrait`, and we'd be incorrectly incrementing the time for the wrong state. 

Could we instead just do some quick math right before sending capturing the events with the current state, as we're doing with the distanceCompleted metric?

bsudekum:
@ericrwolfe this is ready for review.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -82,6 +90,27 @@ struct EventDetails {
         batteryPluggedIn = UIDevice.current.batteryState == .charging || UIDevice.current.batteryState == .full
         batteryLevel = UIDevice.current.batteryLevel >= 0 ? UIDevice.current.batteryLevel * 100 : -1
         applicationState = UIApplication.shared.applicationState
+        if let manager = routeController.locationManager {
+            locationEngine = type(of: manager)
+        }
+        
+        var totalTimeInPortrait = session.timeSpentInPortrait
+        var totalTimeInLandscape = session.timeSpentInLandscape
+        if UIDevice.current.orientation == .portrait {
```
ericrwolfe: Checking back in the docs, we should use `isPortrait` and `isLandscape` everywhere instead of trying to parse these values ourselves: https://developer.apple.com/documentation/uikit/uideviceorientation

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -178,11 +212,11 @@ extension MMEEventsManager {
         return eventDictionary
     }
     
-    func navigationRerouteEvent(routeController: RouteController) -> [String: Any] {
+    func navigationRerouteEvent(routeController: RouteController, MMEEventType: String = MMEEventTypeNavigationReroute) -> [String: Any] {
         let timestamp = Date()
         
         var eventDictionary = self.addDefaultEvents(routeController: routeController)
-        eventDictionary["event"] = MMEEventTypeNavigationReroute
+        eventDictionary["event"] = MMEEventType
```
ericrwolfe: Is this generic event type intentional?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -244,11 +244,38 @@ open class RouteController: NSObject {
...
     func suspendNotifications() {
         NotificationCenter.default.removeObserver(self)
     }
+    
+    func didChangeOrientation() {
+        if UIDevice.current.orientation == .portrait {
+            sessionState.timeSpentInLandscape += abs(sessionState.lastTimeInPortrait.timeIntervalSinceNow)
+            
+            sessionState.lastTimeInPortrait = Date()
+        } else if UIDevice.current.orientation == .landscapeLeft || UIDevice.current.orientation == .landscapeRight {
```
ericrwolfe: See comment above about convenience properties for landscape/portrait

MapboxCoreNavigation/RouteController.swift
```diff
@@ -843,6 +890,20 @@ extension RouteController {
 
         return event.id.uuidString
     }
+    
+    func enqueueFoundFasterRouteEvent() -> String {
+        let timestamp = Date()
+        let eventDictionary = events.navigationRerouteEvent(routeController: self, MMEEventType: FasterRouteFoundEvent)
+        
+        sessionState.lastRerouteDate = timestamp
+        sessionState.numberOfReroutes += 1
```
ericrwolfe: We shouldn't increment either of these (`numberOfReroutes` and `lastRerouteDate`) values, since a faster reroute is intentional, and those are meant to track problematic reroutes.

MapboxCoreNavigation/MMEEventsManager.swift
ericrwolfe: Ah, this isn't a constant, its a function parameter. Can you change the variable name to be `eventType` for clarity?

ericrwolfe:
We should also change the events version to `6` with these changes:
https://github.com/mapbox/mapbox-navigation-ios/blob/30932cf15b1f88b15482f846928982ae4d52c076/MapboxCoreNavigation/MMEEventsManager.swift#L7

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -178,11 +214,11 @@ extension MMEEventsManager {
         return eventDictionary
     }
     
-    func navigationRerouteEvent(routeController: RouteController) -> [String: Any] {
+    func navigationRerouteEvent(routeController: RouteController, eventType: String = MMEEventTypeNavigationReroute) -> [String: Any] {
         let timestamp = Date()
         
         var eventDictionary = self.addDefaultEvents(routeController: routeController)
-        eventDictionary["event"] = MMEEventTypeNavigationReroute
+        eventDictionary["event"] = eventType
```
bsudekum: @ericrwolfe confirmed, this change is what is causing the 422s.

bsudekum:
@ericrwolfe ok, this is ready now.

PR #743, opened 2017-10-23
Various feedback tweaks

frederoni:
- Tweaked the highlighted state.
- Increased auto dismiss interval from 5 to 10 seconds.
- Moved the long press recognizer to each individual cell and removed the delaysTouchesBegan so that it doesn't interfere with the regular `collection:didSelectItemAt:`.
- Added reroute feedback

@bsudekum @ericrwolfe 👀


bsudekum:
@frederoni what do you think about removing the progress bar too? It sort of gives me anxiety in an already stressful situation.

frederoni:
I agree that it felt very stressful at 5 seconds but I think we should keep. @ericrwolfe, any thoughts?

PR #744, opened 2017-10-23
Update localizations

frederoni:
Added pt-BR and updated ca, it, vi, sv.

@bsudekum @1ec5 @brunoabinader @gloffreda 👀


brunoabinader:
pt-BR looks :+1:

gloffreda:
What about OSRM speech translation ?

MapboxNavigation/Resources/ca.lproj/Abbreviations.csv.plist
```diff
@@ -3,69 +3,121 @@
...
+&quot;Comunicacions&quot;,&quot;com.&quot;
+&quot;Xarxa Hidrogràfica&quot;,&quot;curs fluv.&quot;
+&quot;Edificació&quot;,&quot;edif.&quot;
+&quot;Equipament&quot;,&quot;equip.&quot;
+&quot;Element Hidrogràfic&quot;,&quot;hidr.&quot;
+&quot;Indret&quot;,&quot;indr.&quot;
+&quot;Relleu del Litoral&quot;,&quot;lit.&quot;
+&quot;Municipi&quot;,&quot;mun.&quot;
+&quot;Orografia&quot;,&quot;orogr.&quot;
+&quot;Massís&quot;,&quot; Serra&quot;,&quot;serra&quot;
```
1ec5: We don't support synonyms like this, do we? I think we need to split this line up in the translated string in Transifex.

1ec5:
> What about OSRM speech translation ?

That will require a new release of [OSRM Text Instructions](https://github.com/Project-OSRM/osrm-text-instructions/) that includes updates to that project's Transifex translations, as well as a change to the Directions API to pull in the new version.

MapboxNavigation/Resources/ca.lproj/Abbreviations.csv.plist
frederoni: Fixed in Transifex. However, this is only the csv version of the plist so the English fallback will be used. Will fix in a follow up PR.

frederoni: Fixed in #750

1ec5:
Project-OSRM/osrm-text-instructions#189 pulls the instruction translations into OSRM Text Instructions.

Issue #645, opened 2017-09-21
Modularize UI components


bsudekum:
Let's make it so developers can mix and match different pieces of the UI as needed. Developers should be able to provide their own view and add a top bar or a bottom bar as needed.

/cc @frederoni @1ec5 @ericrwolfe

ericrwolfe:
Components of the top and bottom bar that might be useful to modularize. Some of these are basically there while some will require more of a lift: 
- [ ] turn arrow view
- [ ] lane guidance view
- [ ] colorized time remaining label
- [ ] banner label that supports inline shields
- [ ] shield view

Most other cases around top/bottom banners can be made up with basic labels that update in response to a progressDidChange notification.

1ec5:
Are we talking about making these views settable as properties on NavigationViewController (so you can swap in a custom view), or being able to use each of the views above as standalone views (so you could use the banner with, say, a MapKit MKMapView)?

ericrwolfe:
> being able to use each of the views above as standalone views

👆

PR #745, opened 2017-10-23
Rework banner

frederoni:
This PR refactors the instruction banner and fixes the designable issues we've been seeing where the IB agent crashes.

- [x] Extract instructions banner from storyboard into an isolated component
- [x] Update maneuver and lane icons
- [x] Structured instructions
- [x] Inline shields and cache shield images (including shield urls)
- [x] Re-enable abbreviations

@bsudekum @ericrwolfe 👀


MapboxNavigation/Style.swift
```diff
@@ -147,11 +147,27 @@ open class DistanceLabel: StylableLabel { }
...
     var unabridgedText: String? {
         didSet {
-            super.text = unabridgedText?.abbreviated(toFit: availableBounds(), font: font)
+            //super.text = unabridgedText?.abbreviated(toFit: availableBounds(), font: font)
+            super.text = unabridgedText
+            sizeToFit()
+        }
+    }
+    
+    var smartText: String? {
```
bsudekum: Is this used?

frederoni: Not yet but the intention here was to support something like `[Shield][String][Image]` depending on what the format will look like in MapboxDirections.swift. Parsing, async loading of images/shields, abbreviations, layout and constructing an attributed string will be done by setting smartText or whatever we call it later.

bsudekum:
@frederoni do you also want to add a `Then` view to the lower left of the maneuver banner while you're at it? This will require moving the buttons to the right side of the screen which I think should be fine.

For the `Then` view, it can be based on `routeProgress.currentStepProgress.followOnStep`.

ericrwolfe:
@bsudekum @frederoni FYI `Then` view will span the full width of the screen:

image

frederoni:
@bsudekum would you mind taking a look at the visual instruction formatter?

https://github.com/mapbox/mapbox-navigation-ios/blob/6af677c1cf1c77f7ddcdb951e34ae20b0cc6d09b/MapboxCoreNavigation/VisualInstructionFormatter.swift#L31-L47

MapboxCoreNavigation/VisualInstructionFormatter.swift
```diff
@@ -27,4 +27,22 @@ public class VisualInstructionFormatter: NSObject {
             return nil
         }
     }
+    
+    public func strings(leg: RouteLeg?, step: RouteStep?) -> (String?, String?) {
+        if let currentLeg = leg, let destinationName = currentLeg.destination.name, let step = step, step.maneuverType == .arrive {
+            return("You have arrived", destinationName)
```
frederoni: @bsudekum should we localize this or get it from the step’s instruction? (`You have arrived at your destination` sometimes it also appends `, on the right`) /cc @ericrwolfe

bsudekum:
Getting a crash:

```
Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Mutating a priority from required to not on an installed constraint (or vice-versa) is not supported.  You passed priority 1000 and the existing priority was 1.'
*** First throw call stack:
```

MapboxCoreNavigation/VisualInstructionFormatter.swift
bsudekum: We should localize this.

bsudekum: We might want to use OSRMTextInstructions for this actually. /cc @1ec5

bsudekum:
@frederoni it looks like you're missing this case: https://github.com/mapbox/mapbox-navigation-ios/blob/6af677c1cf1c77f7ddcdb951e34ae20b0cc6d09b/MapboxCoreNavigation/VisualInstructionFormatter.swift#L18

MapboxCoreNavigation/VisualInstructionFormatter.swift
```diff
@@ -27,4 +27,22 @@ public class VisualInstructionFormatter: NSObject {
...
+            return (primary, secondary)
+        } else if let step = step, let names = step.names {
+            let primary = names.first
+            let secondary = names.suffix(from: 1).joined(separator: NSLocalizedString("DESTINATION_DELIMITER", bundle: .mapboxCoreNavigation, value: " / ", comment: "Delimiter between multiple destinations"))
+            return (primary, secondary)
+        } else if let currentLeg = leg, let destinationName = currentLeg.destination.name {
+            return (destinationName, nil)
+        }
+        
+        return (step?.instructions, nil)
```
bsudekum: Want to add this as a final else?

MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -28,6 +28,9 @@ public class DistanceFormatter: LengthFormatter {
     
     let nonFractionalLengthFormatter = LengthFormatter()
     
+    /// Indicates the most recently used unit
+    public private(set) var unit: LengthFormatter.Unit = .millimeter
```
bsudekum: Is this being elsewhere that warrants this change?

frederoni:
> @frederoni it looks like you're missing this case:

@bsudekum I was following this pseudocode
<details>

```

 if arrive:
      Primary: You have arrived
      Secondary: Street Name
if has highway refs:
      Primary: {Highway or roundabout exit if available:} Highway refs separated by '/'
      Secondary: Destinations separated by '/' || Street names removing ref/parentheses
if has multiple destinations:
      Primary: {Highway or roundabout exit if available:} First destination
      Secondary: rest of destinations
if has multiple names:
      Primary: {Highway or roundabout exit if available:} First name
      Secondary: rest of names
if has destination || street name: 
      Primary: Destination || Street name
      Secondary: null
Else:
      Primary or secondary field wrapping 2 lines if necessary: instruction 

```

</details>

which prioritizes multiple destinations after highway refs, which I think is correct. But you're right, the multiple destinations case is missing.

MapboxCoreNavigation/DistanceFormatter.swift
frederoni: Yep, it's being used here:

https://github.com/mapbox/mapbox-navigation-ios/blob/95eb404200b64d3d9cd4904e87eca113ffe8970d/MapboxNavigation/InstructionsBannerView.swift#L24

frederoni:
A lot of tweaks left to do but this is what it looks like so far:

bsudekum:
@frederoni, looking great. Couple pieces of feedback:

1. I think there is room to reduce the padding top slightly
1. It'd be great if the top of the maneuver arrow and the top of the text were flush, see below. Same goes for the bottom.
1. Both of these combined might allow us to make the distance text slightly larger which is always a good thing.

screen_shot_2017-10-30_at_8_21_51_am

image

MapboxCoreNavigation/VisualInstructionFormatter.swift
```diff
@@ -27,4 +27,56 @@ public class VisualInstructionFormatter: NSObject {
             return nil
         }
     }
+    
+    /**
+     Creates the optimal text to be displayed given a `RouteLeg` and `RouteStep` divided into primary and secondary string.
+     */
+    public func strings(leg: RouteLeg?, step: RouteStep?) -> (String?, String?) {
```
1ec5: Turn this return type into a typealias to the tuple, so that we can document it better.

1ec5: The signature of this method is `strings(_:step:)`. Let’s think up a name that better describes the relationship between the return type (“strings”) and the first paremeter (a leg).

MapboxCoreNavigation/VisualInstructionFormatter.swift
```diff
@@ -27,4 +27,56 @@ public class VisualInstructionFormatter: NSObject {
...
+            return (primary, secondary)
+        } else {
+            return (step?.instructions, nil)
+        }
+    }
+    
+    /**
+     Creates the optimal instructions to be displayed given a `RouteLeg` and `RouteStep` divided into primary and secondary `Instruction`.
+     */
+    public func instructions(leg: RouteLeg?, step: RouteStep?) -> (Instruction?, Instruction?) {
```
1ec5: This method appears to parallel `strings(_:step:)` quite closely. It’s difficult to tell what each case is for. How about a method that classifies the leg-step combination into one of several enumeration values, so that these two methods become switch statements based on that new method?

MapboxCoreNavigation/VisualInstructionFormatter.swift
1ec5: By the way, a tuple return type won’t bridge to Objective-C, although I don’t know whether that’s a concern at the moment. The Objective-C way of doing things would be to have two separate methods, one for the primary text and the other for the secondary text. That would be far easier to implement after https://github.com/mapbox/mapbox-navigation-ios/pull/745#discussion_r148137222.

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -0,0 +1,121 @@
...
+        distanceLabel.translatesAutoresizingMaskIntoConstraints = false
+        distanceLabel.adjustsFontSizeToFitWidth = true
+        distanceLabel.minimumScaleFactor = 0.2
+        addSubview(distanceLabel)
+        self.distanceLabel = distanceLabel
+        
+        let primaryLabel = PrimaryLabel()
+        primaryLabel.translatesAutoresizingMaskIntoConstraints = false
+        primaryLabel.adjustsFontSizeToFitWidth = true
+        primaryLabel.numberOfLines = 3
```
ericrwolfe: This should be `1` if `secondaryLabel.text != nil`, `2` if `secondaryLabel.text == nil`

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -0,0 +1,121 @@
...
+        primaryLabel.minimumScaleFactor = 0.5
+        primaryLabel.lineBreakMode = .byTruncatingTail
+        addSubview(primaryLabel)
+        self.primaryLabel = primaryLabel
+        
+        let secondaryLabel = SecondaryLabel()
+        secondaryLabel.translatesAutoresizingMaskIntoConstraints = false
+        secondaryLabel.adjustsFontSizeToFitWidth = true
+        secondaryLabel.numberOfLines = 3
```
ericrwolfe: This should be `2`

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -0,0 +1,121 @@
...
+        distanceLabel.adjustsFontSizeToFitWidth = true
+        distanceLabel.minimumScaleFactor = 0.2
+        addSubview(distanceLabel)
+        self.distanceLabel = distanceLabel
+        
+        let primaryLabel = PrimaryLabel()
+        primaryLabel.translatesAutoresizingMaskIntoConstraints = false
+        primaryLabel.adjustsFontSizeToFitWidth = true
+        primaryLabel.numberOfLines = 3
+        primaryLabel.minimumScaleFactor = 0.5
```
ericrwolfe: This should be `26/30.0`.

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -0,0 +1,121 @@
...
+        primaryLabel.lineBreakMode = .byTruncatingTail
+        addSubview(primaryLabel)
+        self.primaryLabel = primaryLabel
+        
+        let secondaryLabel = SecondaryLabel()
+        secondaryLabel.translatesAutoresizingMaskIntoConstraints = false
+        secondaryLabel.adjustsFontSizeToFitWidth = true
+        secondaryLabel.numberOfLines = 3
+        secondaryLabel.minimumScaleFactor = 0.2
```
ericrwolfe: This should be `20/26.0`

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -0,0 +1,121 @@
...
+        let maneuverView = ManeuverView()
+        maneuverView.backgroundColor = .clear
+        maneuverView.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(maneuverView)
+        self.maneuverView = maneuverView
+        
+        let distanceLabel = DistanceLabel()
+        distanceLabel.translatesAutoresizingMaskIntoConstraints = false
+        distanceLabel.adjustsFontSizeToFitWidth = true
+        distanceLabel.minimumScaleFactor = 0.2
```
ericrwolfe: The distance label probably shouldn't resize smaller than the default size `22`. If we were to have a min scale, it should be `20` at the smallest.

MapboxCoreNavigation/VisualInstructionFormatter.swift
```diff
@@ -27,4 +27,56 @@ public class VisualInstructionFormatter: NSObject {
...
+            let roadCode = step.codes?.first ?? step.destinationCodes?.first ?? step.destinations?.first
+            return (Instruction([.init(primary, png: nil, roadCode: roadCode)]), Instruction(secondary))
+        } else if let step = step, step.isNumberedMotorway, let codes = step.codes {
+            let primary = codes.joined(separator: NSLocalizedString("REF_DELIMITER", bundle: .mapboxCoreNavigation, value: " / ", comment: "Delimiter between route numbers in a road concurrency"))
+            let secondary = step.destinations?.joined(separator: NSLocalizedString("DESTINATION_DELIMITER", bundle: .mapboxCoreNavigation, value: " / ", comment: "Delimiter between multiple destinations"))
+            let roadCode = step.codes?.first ?? step.destinationCodes?.first ?? step.destinations?.first
+            return (Instruction([.init(primary, png: nil, roadCode: roadCode)]), Instruction(secondary))
+        } else if let destinations = step?.destinations, let names = step?.names {
+            let primary = names.first
```
ericrwolfe: The if check should be `let destinations = step?.destinations ?? step?.names` and this should check `destinations.first`

MapboxCoreNavigationTests/DistanceFormatterTests.swift
```diff
@@ -7,74 +7,67 @@ let oneFeet: CLLocationDistance = 0.3048
...
-        assertDistance(oneMile*5.4,     spoken: "5 miles",              displayed: "5 mi")
+        assertDistance(0,               displayed: "0 ft")
+        assertDistance(oneFeet*50,      displayed: "50 ft")
+        assertDistance(oneFeet*100,     displayed: "100 ft")
+        assertDistance(oneFeet*249,     displayed: "250 ft")
+        assertDistance(oneFeet*305,     displayed: "300 ft")
+        assertDistance(oneMile*0.1,     displayed: "0.1 mi")
+        assertDistance(oneMile*0.25,    displayed: "0.2 mi")
+        assertDistance(oneMile/2,       displayed: "0.5 mi")
+        assertDistance(oneMile*0.75,    displayed: "0.8 mi")
```
ericrwolfe: Shouldn't this round the same as `oneMile*0.25`? We should be showing every `0.1` mile increment under 3 miles.

MapboxCoreNavigation/VisualInstructionFormatter.swift
frederoni: This method is unused so I just removed it.

MapboxCoreNavigation/VisualInstructionFormatter.swift
frederoni: This is an intermediate step to construct instructions made out of components until we move to server-side instructions. It's pretty much the same as before but with components instead of strings and sometimes the primary string contains a highway shield component. The leg-step combination sounds like a good approach but I assume we'll remove this method within a couple of weeks.

MapboxCoreNavigationTests/DistanceFormatterTests.swift
frederoni: Seems to be a floating point issue. 0.251 rounds to 0.3.

MapboxNavigation/InstructionsBannerViewLayout.swift
frederoni: Using minimum font scale 20/26 would result in only 18 pts remaining so it would never render `2` lines of text in the secondary instruction label.

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -0,0 +1,125 @@
...
+        
+        let separatorView = UIView()
+        separatorView.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(separatorView)
+        self.separatorView = separatorView
+    }
+    
+    func setupLayout() {
+        // Distance label
+        distanceLabel.centerXAnchor.constraint(equalTo: maneuverView.centerXAnchor, constant: 0).isActive = true
```
bsudekum: @frederoni can you explain briefly why you chose to setup constraints here for this view here instead of in the storyboard?

frederoni: It's part of #645, componentization of the UI. Also noted in the description of the PR
> Extract instructions banner from storyboard into an isolated component

PR #746, opened 2017-10-23
Nav "Cancel" feedback event update

JThramer:
Adding `rating` and `comment` fields to `navigation.cancel` event.

/cc @bsudekum @frederoni


bsudekum:
@JThramer would it make more sense to add this when you create a PR for post ride feedback?

JThramer:
Closing. Will include as part of post-ride feedback PR.

PR #747, opened 2017-10-23
Removing Alerts section and updating README to reflect new server-side alerts narrative.

JThramer:
Issue:

screen shot 2017-10-23 at 11 18 00 am

/cc @bsudekum @ericrwolfe


bsudekum:
We should remove this, we no longer use alert levels.

PR #750, opened 2017-10-24
Update Catalan

frederoni:
Follow up on #744 to fix unsupported csv format (see https://github.com/mapbox/mapbox-navigation-ios/pull/744#discussion_r146335263)

`Examples/Swift/ca.lproj/Main.strings` was UTF-16 encoded before this change so GH can't show the diff until next change.

@bsudekum @1ec5


MapboxNavigation/Resources/ca.lproj/Abbreviations.csv.plist
```diff
@@ -14,7 +14,7 @@
 &quot;Relleu del Litoral&quot;,&quot;lit.&quot;
 &quot;Municipi&quot;,&quot;mun.&quot;
 &quot;Orografia&quot;,&quot;orogr.&quot;
-&quot;Massís&quot;,&quot; Serra&quot;,&quot;serra&quot;
+&quot;Massís&quot;,&quot; Serra&quot;
```
1ec5: On second thought, “Serra” certainly isn’t an abbreviation for “Massís”. I think what happened was that the translator inadvertently omitted the abbreviation for “Massís” and put the “Serra” → “serra” item on the same line.

frederoni: Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/765

PR #751, opened 2017-10-24
Use system volume for spoken instructions

bsudekum:
This moves to use the system volume. This also allows the physical mute switch to be in the mute position and still get audio instructions. To mute instructions, the developer will have to set `volume?` to 0 or the user has to touch the mute button within the UI.

/cc @1ec5 @frederoni @ericrwolfe @JThramer

### DIFF
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -23,11 +23,9 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     
     
     /**
-     Volume of spoken instructions. By default, this will be set to the system volume.
+     Volume of spoken instructions.
      */
-    public var volume: Float?
-    
-    var spokenInstructionVolume = AVAudioSession.sharedInstance().outputVolume
+    public var volume: Float = 1
     
     
     /**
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -106,7 +96,7 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
             return
         }
         
-        rerouteSoundPlayer.volume = volume
+        rerouteSoundPlayer.volume = spokenInstructionVolume
```
frederoni: AVAudioPlayer’s volume is relative to the hardware volume, using the same would mean that other audio sources would sound louder (or actually the opposite). For example, set hardware volume to 50%, and AVAudioPlayer.volume to AVAudioSession.outputVolume and AVAudioPlayer will use 25% of maximum volume. I'd suggest setting AVAudioPlayer.volume to 1 or even slightly higher.

frederoni: Also, we have a conflicting PR here -> https://github.com/mapbox/mapbox-navigation-ios/pull/673 but I guess we can wait until this issue is ironed out.

bsudekum: @frederoni if it's set to 1, while it always ignore the system volume?

frederoni: @bsudekum no, it will strictly follow the system’s current volume.

PR #754, opened 2017-10-24
Snapping improvements

bsudekum:
When starting a route, if the `pointBehindClosest` is zero, it creates all sorts of problems when calculating the snapped course. In this case, we can just accept the user's course until the distance increases past zero.

### Rerouting often to show a point

Before
gif2

After
gif3


bsudekum:
Changed this around a bit. If the distance is zero, only use the distance in front of the user to calculate the angle.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -296,21 +296,34 @@ open class RouteController: NSObject {
      */
     public var location: CLLocation? {
         guard let location = rawLocation else { return nil }
-        guard let stepCoordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return nil }
-        guard let snappedCoordinate = Polyline(stepCoordinates).closestCoordinate(to: location.coordinate) else { return location }
```
bsudekum: We were calculating the same variable twice.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -296,21 +296,34 @@ open class RouteController: NSObject {
...
 
         var nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
-        let nearByPolyline = Polyline(nearByCoordinates)
 
         // If the upcoming maneuver a sharp turn, only look at the current step for snapping.
         // Otherwise, we may get false positives from nearby step coordinates
         if let upcomingStep = routeProgress.currentLegProgress.upComingStep,
             let initialHeading = upcomingStep.initialHeading,
             let finalHeading = upcomingStep.finalHeading,
-            initialHeading.differenceBetween(finalHeading) < RouteControllerMaxManipulatedCourseAngle,
```
bsudekum: This would find a small angle if the turn was a slight turn as well.

bsudekum:
Also found another bug while working on this: https://github.com/mapbox/mapbox-navigation-ios/pull/754/commits/21316f8f8774ec414fe75cbec9577dfd22dc3a5c

MapboxCoreNavigation/RouteController.swift
```diff
@@ -296,21 +296,34 @@ open class RouteController: NSObject {
...
             let coordinates = routeProgress.currentLegProgress.currentStep.coordinates {
-           nearByCoordinates = coordinates
+            
+            // Calculate if angle is sharp
+            let inAngle = initialHeading * .pi / 180.0
+            let outAngle = finalHeading * .pi / 180.0
```
1ec5: Turf provides `toDegrees()` and `toRadians()` methods for these conversions.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -296,21 +296,34 @@ open class RouteController: NSObject {
...
+            
+            let inX = sin(inAngle)
+            let inY = cos(inAngle)
+            let outX = sin(outAngle)
+            let outY = cos(outAngle)
+            
+            let turnAngle = acos((inX * outX + inY * outY) / 1.0) * (180 / .pi)
```
1ec5: Turf provides a `differenceBetween(_:)` method for calculating the difference between two angles.

bsudekum: @1ec5 thats the issue here. The difference between a slight turn and a sharp turn can be the same angle. This is being removed in https://github.com/mapbox/mapbox-navigation-ios/pull/754/files#diff-1a2149692b3d7c58b6a0392120d6011aL310

MapboxCoreNavigation/RouteController.swift
bsudekum: Hmm when doing something like:

```swift
let inAngle: CLLocationDegrees = initialHeading.toRadians()
let outAngle: CLLocationDegrees = finalHeading.toRadians()
```

I get a `'toRadians' is inaccessible due to 'internal' protection level`. Ref implementation: https://github.com/mapbox/turf-swift/blob/c0651d438dde71519a0469956b69ef641573fcae/Turf/CoreLocation.swift#L21

JThramer: @bsudekum maybe the extension needs to be marked `public` or `open` as well?

JThramer: @bsudekum @frederoni maybe the extension itself needs to be marked `public` or `open` as well?

bsudekum: Currently the function is private and not yet in a release. Fixing this with https://github.com/mapbox/turf-swift/pull/19

bsudekum:
@frederoni @1ec5 @ericrwolfe this is ready I think.

MapboxCoreNavigation/RouteController.swift
1ec5: Ah, so the bug was that `differenceBetween(_:)` would take the shorter path around the circle, which we didn’t want in this case. I think we should upstream this version of the comparison to Turf.

PR #762, opened 2017-10-26
End Of Route Feedback

JThramer:
Hey all, this is getting pretty close to the original deliverable ask, and the deliverable has changed, so I thought I would put a PR up now so you guys can follow along with my progress.

screen shot 2017-10-25 at 6 25 47 pm

Things Left to do:

- [ ] Add `UITextField` and installable constraints to animate it into place
- [ ] Changing `RouteMapViewController`/`EndOfRouteViewController` to replace the header view as the bottom contained view, instead of a modal presentation.
- [ ] Add logic to move carousel off the top at arrival time.
- [ ] Adjust map offset to compensate for new visual bounds, and set appropriate map mode.
- [ ] Hook up UI to actually emit events (re: work done earlier on event models)
- [ ] Final Polish Tweaks

/cc @frederoni @bsudekum @ericrwolfe @1ec5


Examples/Swift/ViewController.swift
```diff
@@ -255,6 +255,8 @@ extension ViewController: WaypointConfirmationViewControllerDelegate {
 
 extension ViewController: NavigationViewControllerDelegate {
     func navigationViewController(_ navigationViewController: NavigationViewController, didArriveAt waypoint: Waypoint) {
+        if waypoint == currentRoute?.legs.last?.destination { showEndOfRouteFeedback(for: waypoint)}
```
bsudekum: [If this is set to true](https://github.com/mapbox/mapbox-navigation-ios/blob/feb895f0527c20c06d9a05a539e78ad503086601/Examples/Swift/ViewController.swift#L241), will this automatically be shown?

Examples/Swift/ViewController.swift
```diff
@@ -266,6 +268,14 @@ extension ViewController: NavigationViewControllerDelegate {
         
         navigationViewController.present(confirmationController, animated: true, completion: nil)
     }
+    
+    private func showEndOfRouteFeedback(for waypoint: Waypoint) {
```
bsudekum: This should probably me moved out of the example and into `NavigationViewController.swift`

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -1368,8 +1377,57 @@
...
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/MapboxDirections.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/OSRMTextInstructions.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/Pulley.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/SDWebImage.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/Polyline.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/AWSPolly.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/AWSCore.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/Turf.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/Solar.framework",
+				"$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/MapboxMobileEvents.framework",
```
bsudekum: Are these changes necessary?

Examples/Swift/ViewController.swift
ericrwolfe: Right now, our concept of route ratings only applies to the entire route rather than individual legs, so let's only show a star rating for the final waypoint rather than intermediate waypoints. This should be configurable with a simple bool like we do for the feedback button currently.

The overall `arrival` UI could have a public delegate method or invoke function, as you've suggested @bsudekum.

MapboxNavigation/Resources/Base.lproj/Localizable.strings
```diff
@@ -40,6 +40,12 @@
 /* Button title for resume tracking */
 "RESUME" = "Resume";
 
+/* Accessibility Star Label */
+"Set \(index + 1) star rating" = "Set \(index + 1) star rating";
```
1ec5: String interpolation is incompatible with localization. Use `NSString.localizedStringWithFormat(_:_:)` in conjunction with `NSLocalizedString()` and a NumberFormatter.

JThramer:
This effort lives on in #848.

Issue #763, opened 2017-10-26
Catalan road abbreviations are miscapitalized


1ec5:
The Catalan localization’s [abbreviations file](https://github.com/mapbox/mapbox-navigation-ios/blob/d9d7d6e9a2f2f6e6ddb4ab776801ff6d10a45197/MapboxNavigation/Resources/ca.lproj/Abbreviations.csv.plist) capitalizes the first letter of each spelled-out word, which prevents `String.abbreviated(by:)` from finding the abbreviation:

https://github.com/mapbox/mapbox-navigation-ios/blob/bf1c7aa5d17efbc06a3055d6deeff27d28376baf/MapboxNavigation/Abbreviations.swift#L23

[One of the Lithuanian abbreviations](https://github.com/mapbox/mapbox-navigation-ios/blob/808760840aa511866535ee72b61c240c85ade1d6/MapboxNavigation/Resources/lt.lproj/Abbreviations.plist#L7) faces the same problem.

/cc @frederoni

1ec5:
The left side of each line is expected to be lowercased; the right side is expected to begin with a capital letter.

1ec5:
I’ve corrected the Catalan and Lithianian localizations to use the correct case.

frederoni:
Fixed  in #765

PR #765, opened 2017-10-26
Update translations

frederoni:
Fixes incorrect Catalan and Lithuanian abbreviations https://github.com/mapbox/mapbox-navigation-ios/pull/750#discussion_r147042574 and updates es, and vi.

@1ec5 👀


1ec5:
I’ve corrected the Catalan and Lithianian localizations to use the correct case (#763) and remove several redundant or unlikely translations (things like “foo” being abbreviated as “bar”).

Issue #469, opened 2017-08-08
Show subsequent step under turn banner


1ec5:
When the next two maneuvers are sufficiently close together, we should display a mini banner beneath the turn banner for the second maneuver. The mini banner would parallel the “linked” voice instructions that the SDK already provides, but we could display it for slightly less closely spaced maneuvers as well.

/cc @mapbox/navigation-ios

bsudekum:
Would it be appropriate to put `Then turn left` into the `statusView`?

ericrwolfe:
> Would it be appropriate to put Then turn left into the statusView? Or would this be it's own UI element.

This wouldn't be visible enough and would conflate the purpose of the status view, to alert the user of something that's out of the ordinary.

@mayagao could work on some mockups of what this could look like next week.

Sealos:
Hello, any updates on this feature 😃 ?

PR #767, opened 2017-10-26
Use banner instructions from server

bsudekum:
The major issue I'm running into is currently we create instructions from the _upcoming step_, while banner instructions from the server puts information about the upcoming maneuver on the _current step_. As it stands now, this PR is off by one. 

This is dependent on https://github.com/mapbox/MapboxDirections.swift/pull/197

/cc @1ec5 @frederoni @ericrwolfe @JThramer @mcwhittemore


bsudekum:
Alright, I got this working but I had to add a new property to `RouteManeuverViewController.swift`, `currentStep` that only deals with the banner instruction for the current step. This is because the banner content for a given step does not actually match up with say the maneuver arrow.

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -13,10 +13,19 @@ class RouteManeuverViewController: UIViewController {
...
         didSet {
             if isViewLoaded {
-                roadCode = step?.codes?.first ?? step?.destinationCodes?.first ?? step?.destinations?.first
+                roadCode = upcomingStep?.codes?.first ?? upcomingStep?.destinationCodes?.first ?? upcomingStep?.destinations?.first
+                updateStreetNameForStep()
+            }
+        }
+    }
+    
+    var currentStep: RouteStep? {
```
1ec5: Setting the `currentStep` should automatically update `upcomingStep` too; otherwise, weird things could happen if we let the two properties get out of sync. Instead of maintaining parallel state, can we extend Route with a method that gets the step after a given step? Would that be efficient enough for our needs here?

bsudekum: We don't have access to route or routeprogress here. Will look around.

bsudekum: @1ec5 changed this around a bit here https://github.com/mapbox/mapbox-navigation-ios/pull/767/commits/774898dd53431370ea60f3b03fa2c2a14298d534

MapboxNavigation/RouteManeuverViewController.swift
```diff
@@ -13,10 +13,10 @@ class RouteManeuverViewController: UIViewController {
     let routeStepFormatter = RouteStepFormatter()
     let visualInstructionFormatter = VisualInstructionFormatter()
     
-    var step: RouteStep? {
+    var currentAndUpcomingStep: (currentStep: RouteStep?, upcomingStep: RouteStep?) {
```
1ec5: A currentStep inside currentAndUpcomingStep is a bit redundant. Either leave the tuple's parts unnamed, or find a different name for the tuple itself, like displayedSteps. (One could picture us needing to add a third step to this tuple once we implement #469.

bsudekum:
After rebasing off of master, this is off by one again.

bsudekum:
I've got the initial pass working, but I need some time to think about how we'll use different steps to get different pieces of the UI we need:

image


The gist of it: you cannot construct a top banner with a single step. The maneuver arrow comes from the upcoming step while the text comes from the current step. We might want to considering shift visual instructions back by 1. This would allow us to construct the UI from a single step.

/cc @mapbox/guidance

1ec5:
> This would allow us to construct the UI from a single step.

Two steps would still be needed in case a “then” banner is required. I don’t think it should be a goal of this class to take only one step as input.

MapboxNavigation/InstructionLabel.swift
```diff
@@ -23,32 +26,20 @@ open class InstructionLabel: StylableLabel {
...
+            if let imageURL = component.imageURLS[scale] {
+                let shieldKey = UIImage.shieldKey(imageURL, height: shieldHeight)
                 if let cachedImage = UIImage.cachedShield(shieldKey) {
                     string.append(attributedString(with: cachedImage))
-                    if let direction = component.direction {
-                        string.append(NSAttributedString(string: " "+direction, attributes: attributes))
-                    }
                 } else {
-                    // Download shield and display road code in the meantime
-                    string.append(NSAttributedString(string: joinChar+roadCode, attributes: attributes))
```
frederoni: I think we should keep this stringified version of the shield while the image is being downloaded which also works if the request fails.

bsudekum:
Fixed off by one issues in https://github.com/mapbox/mapbox-navigation-ios/pull/767/commits/dbbf3d7dafc73a3818bc466ccbce6e3e65e68168

@frederoni I also removed the current step from the listview https://github.com/mapbox/mapbox-navigation-ios/pull/767/commits/dbbf3d7dafc73a3818bc466ccbce6e3e65e68168#diff-05dcf05089a2947433b80faae897f68dR40

mcwhittemore:
@bsudekum are you saying that there is a bug in the turnIcon logic for banners currently?

bsudekum:
@mcwhittemore I'm just getting at that it's not ideal a single step cannot be used to draw banner. This image shows you where we're currently grabbing the necessary information to draw the whole banner UI:
image

Currently, `voiceInstructions` and `bannerInstructions` are shifted one step ahead of a given step. When we store the user's current step, and we want to draw this step, we have to do some magic to draw different parts of the UI. Some are `currentStep` while other pieces are `currentStep + 1`.

MapboxCoreNavigation/VisualInstructionFormatter.swift
```diff
@@ -1,58 +0,0 @@
...
-    /**
-     Creates the optimal text to be displayed given a `RouteLeg` and `RouteStep`.
-     */
-    @objc public func string(leg: RouteLeg?, step: RouteStep?) -> String? {
-        if let currentLeg = leg, let destinationName = currentLeg.destination.name, let step = step, step.maneuverType == .arrive {
-            return destinationName
-        } else if let destinations = step?.destinations {
-            return destinations.joined(separator: NSLocalizedString("DESTINATION_DELIMITER", bundle: .mapboxCoreNavigation, value: " / ", comment: "Delimiter between multiple destinations"))
-        } else if let step = step, step.isNumberedMotorway, let codes = step.codes {
-            return codes.joined(separator: NSLocalizedString("REF_DELIMITER", bundle: .mapboxCoreNavigation, value: " / ", comment: "Delimiter between route numbers in a road concurrency"))
```
1ec5: Remove `REF_DELIMITER`, `DESTINATION_DELIMITER`, and `ARRIVED_AT_DESTINATION` from the base Localizable.strings too.

MapboxCoreNavigation/VisualInstructionFormatter.swift
```diff
@@ -1,58 +0,0 @@
...
-        }
-    }
-    
-    /**
-     Creates the optimal instructions to be displayed given a `RouteLeg` and `RouteStep` divided into primary and secondary `Instruction`.
-     */
-    public func instructions(leg: RouteLeg?, step: RouteStep?) -> (Instruction?, Instruction?) {
-        if let currentLeg = leg, let destinationName = currentLeg.destination.name, let step = step, step.maneuverType == .arrive {
-            return (Instruction([.init(NSLocalizedString("ARRIVED_AT_DESTINATION", bundle: .mapboxCoreNavigation, value: "You have arrived", comment: "Instruction when arrived at a destination"))]),
-                    Instruction(destinationName))
```
1ec5: This will regress #419 until the server-side visual instructions account for the developer-supplied waypoint name.

/cc @mapbox/guidance

MapboxNavigation/UIImage.swift
```diff
@@ -6,17 +6,19 @@ extension UIImage {
...
+    static func shieldKey(_ url: URL, height: CGFloat) -> String {
+        return "\(url)-\(Int(height))-\(Int(scale))"
     }
     
     static func cachedShield(_ shieldKey: String) -> UIImage? {
         return UIImage.shieldImageCache.object(forKey: shieldKey as NSString)
     }
     
-    static func shieldImage(_ network: String, number: String, height: CGFloat, scale: CGFloat = UIScreen.main.scale, completion: @escaping (UIImage?) -> Void) {
+    static func shieldImage(_ url: URL, height: CGFloat, completion: @escaping (UIImage?) -> Void) {
```
1ec5: These shield methods are better suited for an extension on VisualInstruction or VisualInstructionComponent than an extension on UIImage. A class shouldn’t be extended with a method just because that method returns an instance of the class.

MapboxNavigation/URL.swift
```diff
@@ -1,13 +1,6 @@
 import Foundation
 
 extension URL {
-    
-    static func shieldURL(network: String, number: String, height: CGFloat) -> URL? {
-        guard let imageNamePattern = ShieldImageNamesByPrefix[network] else { return nil }
-        let imageName = imageNamePattern.replacingOccurrences(of: " ", with: "_").replacingOccurrences(of: "{ref}", with: number)
-        return URL(string: "https://commons.wikimedia.org/w/api.php?action=query&format=json&maxage=86400&prop=imageinfo&titles=File%3A\(imageName)&iiprop=url%7Csize&iiurlheight=\(Int(round(height)))")
```
1ec5: ㊗️

Examples/Swift/CustomViewController.swift
```diff
@@ -113,7 +112,7 @@ class CustomViewController: UIViewController, MGLMapViewDelegate, AVSpeechSynthe
                 self.arrowView.text = "⬆️"
             }
         }
-        self.instructionLabel.text = visualInstructionFormatter.string(leg: routeProgress.currentLeg, step: routeProgress.currentLegProgress.upComingStep)
+        self.instructionLabel.text = routeProgress.currentLegProgress.currentStepProgress.step.instructionsDisplayedAlongStep?.first?.primaryText
```
1ec5: VisualInstruction is undocumented – should we expect a developer to have any idea how to use it in a custom UI?

MapboxCoreNavigation/VisualInstructionFormatter.swift
1ec5: In https://github.com/mapbox/mapbox-navigation-ios/pull/848#issuecomment-348406695, I suggested displaying this same information in the end-of-route UI, which would partially mitigate the regression. (Even if the developer provides a street address, the user wouldn’t see it until they literally arrive at the destination.)

bsudekum:
I was able to update the visual tests in https://github.com/mapbox/mapbox-navigation-ios/pull/767/commits/20f3ab296fd3545e1dda99e1169cb5143eb40b41, however the images are not loading in time. Check out the diffs.

MapboxCoreNavigation/VisualInstructionFormatter.swift
1ec5: @bsudekum, I think we should override the arrival visual instruction to use the developer-provided waypoint name. For now, any developer-provided waypoint name exists only the client side. The client SDK would be temporarily compensating for missing functionality on the server side.

Examples/Swift/CustomViewController.swift
bsudekum: Conflicted on this. I'm sort of thinking `VisualInstruction` should be documented and open to use for developers.

MapboxNavigationTests/InstructionsBannerViewTests.swift
```diff
@@ -40,49 +40,67 @@ class InstructionsBannerViewTests: FBSnapshotTestCase {
         view.maneuverView.isStart = true
         view.distance = 482
         
-        view.set(Instruction([Instruction.Component("US-45 / Chicago")]),
-                                              secondaryInstruction: nil)
+        let instructions = [
+            VisualInstructionComponent(text: "US-45", imageURL: nil),
```
1ec5: Nit: this particular ref wouldn’t be hyphenated. 😛

MapboxNavigationTests/InstructionsBannerViewTests.swift
```diff
@@ -117,7 +145,10 @@ extension InstructionsBannerViewTests {
     
     func verifyView(_ view: UIView, size: CGSize) {
         view.frame.size = size
-        FBSnapshotVerifyView(view)
+        // Allow the UI to download images
+        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 2) {
```
1ec5: We should mock these HTTP requests to avoid depending on the server to respond in time.

MapboxNavigation/ManeuverView.swift
```diff
@@ -88,7 +88,7 @@ public class ManeuverView: UIView {
         case .arrive:
             switch direction {
             case .right:
-                ManeuversStyleKit.drawArrowright(frame: bounds, resizing: resizing, primaryColor: primaryColor)
+                ManeuversStyleKit.drawArriveright(frame: bounds, resizing: resizing, primaryColor: primaryColor)
```
frederoni: Good catch

MapboxCoreNavigation/VisualInstructionFormatter.swift
bsudekum: Addressed this in https://github.com/mapbox/mapbox-navigation-ios/pull/767/commits/e57be0b1b917187c7f9b7e9ed39adcf48f09929c. If the waypoint is named and it's an arrival step, the secondary instruction is the named waypoint.

MapboxNavigation/UIImage.swift
bsudekum: Moved in https://github.com/mapbox/mapbox-navigation-ios/pull/767/commits/48347b61fabf2302e00dece89d41a95cdc72809b

bsudekum:
Alright @mapbox/navigation-ios, this is ready for review.

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -0,0 +1,55 @@
+import UIKit
+import SDWebImage
+import MapboxDirections
+
+extension VisualInstructionComponent {
+    
+    static let shieldURLCache = NSCache<NSString, NSURL>()
```
frederoni: `shieldURLCache` was used to cache the `imageURL` from the Wikimedia response. We don't need this cache anymore since the URL is always present in the instructions component if it exists.

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -0,0 +1,55 @@
...
+    func shieldImage(height: CGFloat, completion: @escaping (UIImage?) -> Void) {
+        guard let imageURL = imageURL else { return }
+        let shieldKey = self.shieldKey()
+        
+        if let cachedImage = self.cachedShield(shieldKey) {
+            completion(cachedImage)
+            return
+        }
+        
+        if let cachedURL = VisualInstructionComponent.shieldURLCache.object(forKey: shieldKey as NSString) {
```
frederoni: We can drop this entire case.

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -0,0 +1,55 @@
...
+            SDWebImageDownloader.shared().downloadImage(with: (cachedURL as URL), options: [], progress: nil, completed: { (image, data, error, successful) in
+                guard let imageData = data else { return }
+                guard let downscaledImage = UIImage(data: imageData) else {
```
frederoni: UIImage won't scale down automatically when initialized with data since the file extension is unknown. However, the `image` property that's also returned in this completion is already downscaled.

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -0,0 +1,55 @@
...
+            VisualInstructionComponent.shieldURLCache.setObject(imageURL as NSURL, forKey: shieldKey as NSString)
+            
+            SDWebImageDownloader.shared().downloadImage(with: imageURL, options: [], progress: nil, completed: { (image, data, error, successful) in
+                guard let imageData = data else { return }
+                
+                guard let downscaledImage = UIImage(data: imageData) else {
+                    completion(nil)
+                    return
+                }
+                VisualInstructionComponent.shieldImageCache.setObject(downscaledImage, forKey: shieldKey as NSString)
```
frederoni: Let's replace this memory only NSCache with `SDImageCache.shared().imageFromCache/store` to fully utilize memory and disk cache.

bsudekum:
@frederoni updated.

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -0,0 +1,40 @@
...
+        let shieldKey = self.shieldKey()
+        
+        if let cachedImage = self.cachedShield(shieldKey) {
+            completion(cachedImage)
+            return
+        }
+        
+            
+        SDWebImageDownloader.shared().downloadImage(with: imageURL, options: [], progress: nil, completed: { (image, data, error, successful) in
+            guard let imageData = data else { return }
```
frederoni: Nit: drop these two guards and use `image` instead of `data`.

bsudekum: Don't we need to properly scale the image though?

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -0,0 +1,40 @@
...
+            
+            guard let downscaledImage = UIImage(data: imageData, scale: VisualInstructionComponent.scale) else {
+                completion(nil)
+                return
+            }
+            
+            SDImageCache.shared().store(downscaledImage, forKey: shieldKey, toDisk: true, completion: nil)
+            completion(downscaledImage)
```
frederoni: This completion should be moved to inside the `store(…completion:)` to guarantee that the image is cached before we reconstruct the instruction here:

https://github.com/mapbox/mapbox-navigation-ios/blob/8e7f467edcd6990795c23edb08c9e6af31c46762/MapboxNavigation/InstructionLabel.swift#L41-L44

MapboxNavigation/VisualInstructionComponent.swift
frederoni: Yes. The current implementation does scale it appropriately with `UIImage(data:scale)`, however, the image is being initialized with data on the main thread which could cause jittering. SDWebImage already used the data to create a UIImage on a background thread so there is no need for us to do the same again.

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -0,0 +1,42 @@
...
+        guard let imageURL = imageURL else { return nil }
+        return "\(imageURL.absoluteString)-\(VisualInstructionComponent.scale)"
+    }
+    
+    func cachedShield(_ shieldKey: String) -> UIImage? {
+        return SDImageCache.shared().imageFromCache(forKey: shieldKey)
+    }
+    
+    func shieldImage(height: CGFloat, completion: @escaping (UIImage?) -> Void) {
+        guard let imageURL = imageURL else { return }
```
frederoni: Should we call `completion(nil)` in these two guards?

Issue #759, opened 2017-10-25
Store prefetched Polly audio clips in NSCache


1ec5:
`PollyVoiceController.spokenInstructionsForRoute` should be an NSCache, not a simple dictionary. That way we can control the size of the prefetched/cached instructions in memory. We wouldn’t want the cache to create memory pressure and cause the host application to be killed during a long route.

/cc @bsudekum

PR #768, opened 2017-10-26
Use NSCache instead of dictionary

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/759

/cc @1ec5

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -44,7 +44,7 @@ public class PollyVoiceController: RouteVoiceController {
     var cacheURLSession: URLSession
     var cachePollyTask: URLSessionDataTask?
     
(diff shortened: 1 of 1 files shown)


bsudekum:
bump @1ec5

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -44,11 +44,13 @@ public class PollyVoiceController: RouteVoiceController {
     var cacheURLSession: URLSession
     var cachePollyTask: URLSessionDataTask?
     
-    var spokenInstructionsForRoute: [String: Data] = [:]
+    var spokenInstructionsForRoute = NSCache<AnyObject, AnyObject>()
```
frederoni: Nit: use <NSString, Data or NSData> here instead for improved type safety.

Issue #766, opened 2017-10-26
Animate zoom level depending on location on step and step type


bsudekum:
Let's resurrect https://github.com/mapbox/mapbox-navigation-ios/pull/656 and add some simple rules for animating the camera altitude based on road class, distance to maneuver. Rules:

* If the road class is of type `.motorway` and
* The distance to the maneuver is greater than X meters

Then set the altitude to 2000m

If the altitude is 2000m:

* And the distance remaining is less than X meters
* And the upcoming step is not a motorway

Reset the altitude to the default.

If the the altitude is 2000m:

* and the upcoming step is a motorway
* and is also long

Keep the altitude at 2000m

/cc @JThramer @ericrwolfe

1ec5:
There are a lot of conditionals in this logic. That means a lot of edge cases, so look out for possible yo-yoing around each of these thresholds. If that turns out to be a rabbit hole, consider coming up with formulas that match all these thresholds by interpolating values.

ericrwolfe:
The idea here would be to have the voice announcements trigger a camera change, so that when a driver hears an announcement and looks at the screen, the camera will have updated accordingly.

Zooming out for long steps would allow us to heavily throttle the render calls for significant battery changes (when zoomed out, the location puck will move 1px less frequently), and also give the driver more context about the long stretch of road ahead.

1ec5:
> give the driver more context about the long stretch of road ahead

Very much in favor of this. There’s a lot more we can do a long these lines as tail work, but the logic above should do for now, as long as we only reevaluate it on each spoken instruction.

1ec5:
Implemented in #769.

PR #769, opened 2017-10-27
Conditional Maneuver-based camera altitude animation

JThramer:
Implements #766. Quick and dirty first pass to get eyes on the feature.

Can't seem to get demo videos uploaded here, will provide on demand.

/cc @bsudekum @frederoni @ericrwolfe


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +287,30 @@ class RouteMapViewController: UIViewController {
...
+        let upcomingMotorway = routeController.routeProgress.currentLegProgress.upComingStep?.intersections?.first?.outletRoadClasses?.contains(.motorway) ?? false
+        
+        if isZoomedOut {
+            if !upcomingMotorway, routeProgress.distanceRemaining <= threshold {
+                setCamera(altitude: zoomInAltitude)
+            } else if upcomingMotorway, routeProgress.distanceRemaining >= threshold {
+                setCamera(altitude: zoomOutAltitude)
+            }
+        } else if upcomingMotorway, routeProgress.distanceRemaining >= threshold { // we are zoomed in
+           setCamera(altitude: zoomOutAltitude)
```
ericrwolfe: What will happen to the camera if someone is in "Overview" mode (the sqiggly button above the volume button)?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +287,30 @@ class RouteMapViewController: UIViewController {
             mapView.removeArrow()
         }
     }
+
+    func updateCameraAltitude(for routeProgress: RouteProgress) {
+        let threshold = 1000.0 //meters
+        let zoomOutAltitude = 2000.0
+        let zoomInAltitude = 1000.0
```
ericrwolfe: Is this the default altitude? Do we have a constant elsewhere for this?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +287,30 @@ class RouteMapViewController: UIViewController {
...
+        
+        let isZoomedOut = abs(mapView.camera.altitude - 2000) <= 25
```
ericrwolfe: What do the 2000 and 25 values mean here?

JThramer: Since altitude is a float, I noticed that the altitude is not exactly `2000.0000...`, (which is not unusual in the floating point world) so it's basically just a tolerance to ask "are we within 25 meters of our 'zoomed out' altitude?" That tolerance should probably be shaved down.

MapboxNavigation/RouteMapViewController.swift
JThramer: I observed the default altitude (aka I didn't do any gesture manipulation to the camera) to be ~1000 when doing testing this morning. If there is a default defined somewhere, it would probably be trivial to reference that.

MapboxNavigation/RouteMapViewController.swift
JThramer: I didn't consider `userTrackingMode` implications. Will fix.

ericrwolfe:
From testing, it seems like this is checking whether the upcoming step is a highway, rather than the current step the driver is on, which leads to the zooming happening one step early.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +291,32 @@ class RouteMapViewController: UIViewController {
             mapView.removeArrow()
         }
     }
+
+    func updateCameraAltitude(for routeProgress: RouteProgress) {
+        guard mapView.tracksUserCourse else { return } //only adjust when we are actively tracking user course
+        
+        let zoomOutAltitude = RouteMapViewController.zoomedOutManeuverAltitude
+        let defaultAltitude = mapView.defaultAltitude
```
bsudekum: Would it make sense for `mapView.defaultAltitude ` and `RouteMapViewController.zoomedOutManeuverAltitude ` to live next to one another?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +291,32 @@ class RouteMapViewController: UIViewController {
...
+        let isLongRoad = routeProgress.distanceRemaining >= RouteMapViewController.longManeuverDistance
```
bsudekum: Move `RouteMapViewController.longManeuverDistance ` to a constants.swift.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +291,32 @@ class RouteMapViewController: UIViewController {
...
+        
+        let isMotorway : (RouteStep?) -> Bool? = { $0?.intersections?.first?.outletRoadClasses?.contains(.motorway) }
```
bsudekum: We have a `isMotorway` motorway function. I'd say, maybe move it to an extension on `RouteStep`.

MapboxNavigation/Constants.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+struct MapConstants {
```
bsudekum: These should be made public. Although not as pretty, it might be easer to just create variables as seen here https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/Constants.swift. I don't think this will bridge to obj c as is.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +288,32 @@ class RouteMapViewController: UIViewController {
...
+        
+        let zoomOutAltitude = MapConstants.zoomedOutManeuverAltitude
+        let defaultAltitude = MapConstants.defaultAltitude
+        let isLongRoad = routeProgress.distanceRemaining >= MapConstants.longManeuverDistance
+        
+        
+        let currentStepIsMotorway = currentStep.isMotorway
+        let nextStepIsMotorway = upComingStep?.isMotorway ?? false
+        let isExiting = (currentStepIsMotorway && !nextStepIsMotorway) //are we exiting a motorway?
+        let notOnMotorway = (!currentStepIsMotorway && !nextStepIsMotorway) //are we not on a motorway?
```
bsudekum: Are the parens necessary here and the line above?

JThramer: They are superfluous. Good catch, will fix.

MapboxNavigation/Constants.swift
1ec5: Since these constants are strongly associated with NavigationMapView, you could alternatively make them class properties on NavigationMapView. That’ll give you the namespacing you want in both Swift and Objective-C.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +288,32 @@ class RouteMapViewController: UIViewController {
...
+        
+        let currentStepIsMotorway = currentStep.isMotorway
+        let nextStepIsMotorway = upComingStep?.isMotorway ?? false
+        let isExiting = currentStepIsMotorway && !nextStepIsMotorway //are we exiting a motorway?
+        let notOnMotorway = !currentStepIsMotorway && !nextStepIsMotorway //are we not on a motorway?
+        
+        if (notOnMotorway || isExiting) { //if we're exiting or not on a motorway, we should be zoomed in.
+            return setCamera(altitude: defaultAltitude)
+        }
+        if currentStepIsMotorway, isLongRoad { //otherwise, we should be zoomed out if we're on motorway and step distance is long enough
```
bsudekum: Should this be an else if?

JThramer: It won't affect the logic of anything, but it certainly can be.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -22,6 +22,12 @@ let sourceOptions: [MGLShapeSourceOption: Any] = [.maximumZoomLevel: 16]
 @objc(MBNavigationMapView)
 open class NavigationMapView: MGLMapView {
     
+    //MARK: Class Constants
+    public static let defaultAltitude = 1000.0
```
1ec5: Explicitly specify CLLocationDistance as the type. That along with a documentation comment will make it clear how the developer is supposed to interpret this option. (CLLocationDistance is defined to be measured in meters.)

MapboxNavigation/NavigationMapView.swift
```diff
@@ -23,9 +23,21 @@ let sourceOptions: [MGLShapeSourceOption: Any] = [.maximumZoomLevel: 16]
 open class NavigationMapView: MGLMapView {
     
     //MARK: Class Constants
-    public static let defaultAltitude = 1000.0
-    public static let zoomedOutManeuverAltitude = 2000.0
-    public static let longManeuverDistance: CLLocationDistance = 1000.0 //meters
+    
+    /**
+     `defaultAltitude` is the altitude that the map camera initally defaults to.
```
1ec5: Nit: no need to repeat the name of the property inside its own documentation. Sometimes we do that for classes, but I guess that’s because we try to be less terse there.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -23,9 +23,21 @@ let sourceOptions: [MGLShapeSourceOption: Any] = [.maximumZoomLevel: 16]
...
+     */
+    public static let defaultAltitude: CLLocationDistance = 1000.0
+    
+    /**
+     `zoomedOutMotorwayAltitude` is the altitude we conditionally zoom out to when we are on a motorway, and the maneuver length is sufficently long.
```
1ec5: Nit: “we” refers to someone different in the same sentence. How about “the map” for the first occurrence and “the user” for the second.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +288,31 @@ class RouteMapViewController: UIViewController {
...
+        let defaultAltitude = NavigationMapView.defaultAltitude
+        let isLongRoad = routeProgress.distanceRemaining >= NavigationMapView.longManeuverDistance
+        
+        
+        let currentStepIsMotorway = currentStep.isMotorway
+        let nextStepIsMotorway = upComingStep?.isMotorway ?? false
+        let isExiting = currentStepIsMotorway && !nextStepIsMotorway //are we exiting a motorway?
+        let notOnMotorway = !currentStepIsMotorway && !nextStepIsMotorway //are we not on a motorway?
+        
+        if (notOnMotorway || isExiting) { //if we're exiting or not on a motorway, we should be zoomed in.
```
1ec5: Nit: the parentheses are unnecessary.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -22,6 +22,24 @@ let sourceOptions: [MGLShapeSourceOption: Any] = [.maximumZoomLevel: 16]
 @objc(MBNavigationMapView)
 open class NavigationMapView: MGLMapView {
     
+    //MARK: Class Constants
+    
+    /**
+     `defaultAltitude` is the altitude that the map camera initally defaults to.
```
1ec5: Nit: no need to repeat the name of the property inside its own documentation. Sometimes we do that for classes, but I guess that’s because we try to be less terse there.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +288,31 @@ class RouteMapViewController: UIViewController {
...
+        
+        let zoomOutAltitude = NavigationMapView.zoomedOutMotorwayAltitude
+        let defaultAltitude = NavigationMapView.defaultAltitude
+        let isLongRoad = routeProgress.distanceRemaining >= NavigationMapView.longManeuverDistance
+        
+        
+        let currentStepIsMotorway = currentStep.isMotorway
+        let nextStepIsMotorway = upComingStep?.isMotorway ?? false
+        let isExiting = currentStepIsMotorway && !nextStepIsMotorway //are we exiting a motorway?
+        let notOnMotorway = !currentStepIsMotorway && !nextStepIsMotorway //are we not on a motorway?
```
1ec5: In general, it’s less confusing to express Boolean logic in the positive than as double negatives. In this case, if you turn this variable into `isOnOrApproachingMotorway`, you’ll find that you can inline it below:

```swift
if currentStepIsMotorway || nextStepIsMotorway || isExiting
```

which simplifies to:

```swift
if currentStepIsMotorway || nextStepIsMotorway
```

Is that your intended condition for zooming in?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -287,6 +288,29 @@ class RouteMapViewController: UIViewController {
...
+        
+        let zoomOutAltitude = NavigationMapView.zoomedOutMotorwayAltitude
+        let defaultAltitude = NavigationMapView.defaultAltitude
+        let isLongRoad = routeProgress.distanceRemaining >= NavigationMapView.longManeuverDistance
+        
+        
+        let currentStepIsMotorway = currentStep.isMotorway
+        let nextStepIsMotorway = upComingStep?.isMotorway ?? false
+        
+        if currentStepIsMotorway, nextStepIsMotorway, isLongRoad {
```
bsudekum: What if the user starts on a motorway and the next step is not a motorway? Shouldn't we zoom out?

JThramer: The most likely scenario where this would happen is if a user _enters_ on one onramp and immediately _exits_ via the next offramp (never actually getting on the interstate), the camera would not zoom out in this instance, which is arguably desirable.

JThramer: (I've had this happen to me a couple of times in denver, not on voyage)

PR #777, opened 2017-10-30
Adjust layout for iPhone X

bsudekum:
screen shot 2017-10-30 at 3 05 41 pm
screen shot 2017-10-30 at 3 05 46 pm

screen shot 2017-10-30 at 3 04 47 pm

/cc @frederoni


bsudekum:
Well, bummer: 
screen shot 2017-10-30 at 3 20 10 pm

We can however require xcode 9 but not require swift 4.

frederoni:
Let's wait with changes like this until #745 lands. Some of the layout is being moved from storyboard into code and the Pulley will be removed in a subsequent PR.

Issue #773, opened 2017-10-30
Dynamic camera, first pass


ericrwolfe:
The camera should:
* Be more dynamic, communicating maneuver urgency by camera distance/zoom level
* Be timed with voice announcements: what's communicated on screen should match audio cues
* Provide ample context for maneuvers

As a rough first pass, we should implement basic stepped camera movements achieving the following:
- [x] Triggered by/synchronized with voice announcements
- [x] On the completion announcement of a maneuver: zoom out to provide context for long steps with long periods of inactivity (https://github.com/mapbox/mapbox-navigation-ios/pull/769)
- [ ] At the first approach maneuver announcement (`In 2 miles...`): zoom to show the upcoming maneuver on the map, whether by a fixed zoom level based on distance, or a zoom to fit function.
- [x] At the final turn maneuver announcement: zoom in closely to our current default zoom level 

Once this basic set of triggers/logic is in place, we can experiment with more sophisticated zoom logic, camera tilt, anchor points, etc. @d-prukop

cc @jthramer @bsudekum @danesfeder

JThramer:
@ericrwolfe A question about Item 3: Do you want to set the altitude _once_ at maneuver announcement time? or do you want to continuously zoom in as the user gets closer to the upcoming maneuver, a la Lyft/Uber? Both are doable, but the latter might be a little bit tricker, as there may be some `userTrackingMode` implications, as well as the need to hook into `NavigationController.progressDidChange(notification:)`.

/cc @bsudekum @frederoni @1ec5

ericrwolfe:
> Do you want to set the altitude once at maneuver announcement time? or do you want to continuously zoom in as the user gets closer to the upcoming maneuver

@JThramer For this first pass we'll just zoom in once at each step. We'll explore continuous zoom in follow on work. cc @d-prukop

JThramer:
Closing as the development cycle seems to be over for this work.

ericrwolfe:
@JThramer did we ever implement the approach maneuver announcement zooming to show the upcoming maneuver (item 3)? If not, can you open another ticket as tail work with some background on any approaches you explored?

JThramer:
[...]

PR #781, opened 2017-11-01
Dynamic Camera Updates

JThramer:
Implements #773. Adding logic to make maneuver-based altitude even smarter. 🌠 

To-Do: 
~~Implement Item 3 in #773~~

/cc @bsudekum @ericrwolfe @frederoni @1ec5


MapboxCoreNavigation/RouteLeg.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import MapboxDirections
+
+extension RouteLeg {
+    /**
+     Returns the last step in a leg, which is indicitive of a user approaching their destination.
```
bsudekum: I'm not sure I follow this logic, why does a maneuver type of turn mean they are approaching their destination? Should destination be swapped out with `maneuver`?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -297,13 +297,16 @@ class RouteMapViewController: UIViewController {
         let isLongRoad = routeProgress.distanceRemaining >= NavigationMapView.longManeuverDistance
         
         
+        //If the user is at the last turn maneuver, the map should zoom in to the default altitude.
+        let lastTurn = routeProgress.currentLeg.lastTurn
+        guard currentStep != lastTurn else { return setCamera(altitude: defaultAltitude) }
```
bsudekum: What if the last step is 10 miles long? Should this happen when the user starts the last step or is near the end?

JThramer: Yeah, after talking with @ericrwolfe it seems I misunderstood what we meant by "final turn maneuver announcement". Will rework.

PR #784, opened 2017-11-02
Post-reroute feedback

### LINKED ISSUE #780: Post-reroute feedback
Rather than displaying a separate set of 3 feedback items for reroutes, we should show the ordinary feedback popup with the normal set of categories. 

Immediately after a reroute, a "Report Problem" button should slide down from under the turn banner and stay on screen for 10s.

Tapping this button will open the feedback UI, but we should set the `source` property to `reroute` instead of the default `user` value.

cc @frederoni @bsudekum

frederoni:
Fixes #780 and right align floating buttons

- [x] set source to `reroute`

Note that the reroute items were removed after the recording.

report

@ericrwolfe @bsudekum 👀

### DIFF
--- a/MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
+++ b/MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
@@ -82,7 +82,7 @@
                                 <color key="backgroundColor" red="0.14901960784313725" green="0.23921568627450979" blue="0.3411764705882353" alpha="0.80000000000000004" colorSpace="custom" customColorSpace="sRGB"/>
                                 <fontDescription key="fontDescription" type="system" weight="semibold" pointSize="15"/>
                                 <inset key="contentEdgeInsets" minX="10" minY="10" maxX="10" maxY="10"/>
-                                <state key="normal" title="Report a Problem">
+                                <state key="normal" title="Report Problem">
                                     <color key="titleColor" red="1" green="1" blue="1" alpha="1" colorSpace="calibratedRGB"/>
                                 </state>
                                 <connections>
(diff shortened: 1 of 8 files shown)


bsudekum:
@frederoni can we make it a little more subtle? It's pretty large right now.

frederoni:
How about wait with the "Report a Problem" until after the "Rerouting... " spinner disappears?
There's a small delay after `didReroute` before the status view disappears.

bsudekum:
@frederoni yeah that sounds good to me

frederoni:
@bsudekum 
reroute2

bsudekum:
@frederoni awesome. Final piece of feedback, would it make sense to make the y constraint equal to the y value of the normal feedback button? It might help us teach users about the feedback button.

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -77,6 +77,18 @@
                                     <userDefinedRuntimeAttribute type="string" keyPath="styleURL__" value="mapbox://styles/mapbox/navigation-guidance-day-v2"/>
                                 </userDefinedRuntimeAttributes>
                             </view>
+                            <button hidden="YES" opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="I8f-iR-MZm" customClass="MBReportButton">
+                                <rect key="frame" x="115" y="194.5" width="145" height="38"/>
+                                <color key="backgroundColor" red="0.14901960784313725" green="0.23921568627450979" blue="0.3411764705882353" alpha="0.80000000000000004" colorSpace="custom" customColorSpace="sRGB"/>
+                                <fontDescription key="fontDescription" type="system" weight="semibold" pointSize="15"/>
+                                <inset key="contentEdgeInsets" minX="10" minY="10" maxX="10" maxY="10"/>
+                                <state key="normal" title="Report a Problem">
```
ericrwolfe: Can we change this to "Report Problem" for consistency with the feedback UI title?

ericrwolfe:
> @frederoni awesome. Final piece of feedback, would it make sense to make the y constraint equal to the y value of the normal feedback button? It might help us teach users about the feedback button.

@bsudekum @frederoni no, let's keep it right under the banner right now for this first pass. Having it floating in the center of the screen would be awkward and block the map view. 

I'd like to experiment with alternate placements for the feedback button, so we can revisit this then.

frederoni:
@bsudekum @ericrwolfe Can I get some 👀  on https://github.com/mapbox/mapbox-navigation-ios/pull/784/commits/c9a4a326e905845ea5014013f2e813d9048a3d99?
Feel free to merge if it looks good.

ericrwolfe:
Looks good to me.

PR #785, opened 2017-11-02
Replace Pulley with a dropdown

frederoni:
- [x] Remove dependency on Pulley
- [x] Replace the RouteTableViewController with a bottom banner
- [x] Dropdown list of steps from the top banner
- [x] Show highway direction next to highway shield 
- [x] Multiple waypoints
- [x] Show upcoming steps only
- [x] Show map arrow when previewing step
- [x] Shift distance in list of steps by -1
- [x] Preview banner

@bsudekum @ericrwolfe 👀


frederoni:
Simplifies the storyboard a bit: 

screenshot 2017-11-03 23 55 40

frederoni:
Current progress:

Cartfile
```diff
@@ -1,7 +1,6 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 3.6
 github "mapbox/MapboxDirections.swift" ~> 0.11
 github "mapbox/turf-swift" ~> 0.0.3
-github "52inc/Pulley" == 1.4
```
bsudekum: Can you also update the podspec?

MapboxNavigation/BottomBannerView.swift
```diff
@@ -0,0 +1,136 @@
...
+    @IBAction func cancel(_ sender: Any) {
+        delegate?.didCancel()
+    }
+    
+    override open func prepareForInterfaceBuilder() {
+        super.prepareForInterfaceBuilder()
+        separatorView.backgroundColor = .red
+        timeRemainingLabel.text = "22 min"
+        distanceRemainingLabel.text = "4 mi"
+        arrivalTimeLabel.text = "10:09"
```
bsudekum: I assume these are placeholders?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -432,14 +399,11 @@ public class NavigationViewController: NavigationPulleyViewController, RouteMapV
     }
     
     func progressDidChange(notification: NSNotification) {
-        resetETATimer()
```
bsudekum: Can we add this timer functionality back? It's there for when the user starts a navigation session and then does not move for some time.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -107,6 +107,8 @@ class RouteMapViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         
+        resetETATimer()
```
bsudekum: Oh I see it here now.

MapboxNavigation/BottomBannerView.swift
frederoni: Yes, `prepareForInterfaceBuilder()` is only called when rendering through IB.

MapboxNavigation/NavigationViewController.swift
frederoni: The timer just moved to `RouteMapViewController`. Should be the exact same functionality.

MapboxNavigation/StepsViewController.swift
```diff
@@ -0,0 +1,149 @@
...
+        self.steps = steps
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        setupViews()
+    }
+    
+    func setupViews() {
+        view.backgroundColor = #colorLiteral(red: 0.05205290616, green: 0.07546465523, blue: 0.1024953017, alpha: 1)
```
bsudekum: Is this configurable?

frederoni: Good catch, will add a separate view for this.

frederoni:
There are a few breaking changes here. Stylable components that were being used by RouteTableViewController do not exist anymore.

Apart from that, the `NavigationViewController.navigationDelegate` has been renamed to just `delegate` (previously clashed with Pulley.delegate).

frederoni:
Looks like the iPhone X changes regressed here. Probably in https://github.com/mapbox/mapbox-navigation-ios/pull/785/commits/c3324c7c4074f64ef300e764c1e206c3ecda7078

@bsudekum how did you solve the excessive margin for logo and attribution?
screenshot 2017-11-06 21 46 11

bsudekum:
@frederoni hmmm I'm not seeing that:

image

frederoni:
👆 looks like a commit from yesterday. A merge conflict in the storyboard resulted in the floating buttons moving back to the left side.

MapboxNavigation.podspec
```diff
@@ -45,7 +45,6 @@ Pod::Spec.new do |s|
 
   s.dependency "MapboxCoreNavigation", "#{s.version.to_s}"
   s.dependency "Mapbox-iOS-SDK", "~> 3.6"
-  s.dependency "Pulley", "1.4"
```
bsudekum: One last to remove: https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation-Documentation.podspec#L49

bsudekum:
@frederoni can we remove past and current steps from the dropdown list?

Includes current step:

simulator screen shot - iphone x - 2017-11-07 at 09 51 52

Includes past steps: 
simulator screen shot - iphone x - 2017-11-07 at 09 52 12

bsudekum:
@frederoni did we remove the ability to swipe with this change?

bsudekum:
@frederoni what do you think about some indicator on how to close this view? Two examples here: arrow in upper right, `Close` button at bottom.
simulator_screen_shot_-_iphone_x_-_2017-11-07_at_09_52_12

ericrwolfe:
> @frederoni what do you think about some indicator on how to close this view? Two examples here: arrow in upper right, Close button at bottom.

Would love to test this with @v8c3 once the UI lands. I imagine most people who find this with a tap will also be able to find their way out (everything on this screen would be a tap target and close it out). 

Tapping on a list item should take you to the maneuver preview, and would show the "Resume" button as we show with overview to quickly jump back to turn-by-turn. 

Discoverability of this menu in the first place is a different issue we should user test, but would at least be a significant improvement over swiping left/right and swiping the summary view from the bottom. @v8c3 tagging you as something we should user test :).

frederoni:
> can we remove past and current steps from the dropdown list?

👍 

> did we remove the ability to swipe with this change?

Yes

> Tapping on a list item should take you to the maneuver preview, and would show the "Resume" button as we show with overview to quickly jump back to turn-by-turn.

This is already the case.

bsudekum:
> Would love to test this with @v8c3 once the UI lands. I imagine most people who find this with a tap will also be able to find their way out (everything on this screen would be a tap target and close it out).

Per voice with @ericrwolfe, this is not an acceptable answer. We know we need some sort of indicator right now since it's a very confining UI once in it. We need to have some sort of exit button before merging this.

> Tapping on a list item should take you to the maneuver preview, and would show the "Resume" button as we show with overview to quickly jump back to turn-by-turn.

I'm not actually seeing this.

frederoni:
> > Tapping on a list item should take you to the maneuver preview, and would show the "Resume" button as we show with overview to quickly jump back to turn-by-turn.

> I'm not actually seeing this.

Landed in the second last commit today https://github.com/mapbox/mapbox-navigation-ios/pull/785/commits/a8a4a54390f19e8a8bc16813b3f889fea6c3c049 though it only moves the camera to the maneuver. The banner still shows the current step. Is this the correct behavior?

ericrwolfe:
> Landed in the second last commit today a8a4a54 though it only moves the camera to the maneuver. The banner still shows the current step. Is this the correct behavior?

@frederoni The banner should also change to the future maneuver, with a darkened background that matches the list item as a subtle reminder this is a future maneuver.

ericrwolfe:
@frederoni The distances for each of the list items should also be the distance up to the maneuver, rather than the distance after the maneuver. Basically the distances should all be shifted up by one.

So, for example, in this screenshot, the arrival at Linda Ann Place should show "150ft" instead of "0ft":
list

bsudekum:
> Landed in the second last commit today a8a4a54 though it only moves the camera to the maneuver. The banner still shows the current step. Is this the correct behavior?

👍, I forgot to pull, seeing it now. Can you call the `showArrow()` function after selecting? Hitting resume should then update the arrow back to the current step.

bsudekum:
@frederoni looking good. Now that we have safe areas in, would it be possible to have the close button respect them? It's a little close to the iphone x home button thing 

image

bsudekum:
Perhaps for the final item it could read:

```
turn icon | primary text = You have arrived 
no distance if zero | secondary text = waypoint name || step.name
```

image

frederoni:
this PR is ready for another review 👀

bsudekum:
@frederoni I'm seeing something strange, it looks like overlapping views?

gif3

bsudekum:
If the dropdown is down, and the user is rerouted, we should dismiss the view.

bsudekum:
It looks like we're showing the current step name instead of the upcoming step name after pressing resume:

simulator screen shot - iphone x - 2017-11-08 at 12 07 59

PR #788, opened 2017-11-02
Tweak instruction banner

frederoni:
Follow up for #745 

@ericrwolfe @bsudekum 👀


ericrwolfe:
@frederoni looks like the other changes we discussed should go in another PR 😬

PR #789, opened 2017-11-02
Add alternate routes

bsudekum:
Changes `showRoute` to accept an array of routes. This also adds a delegate method for selecting a route.

Issues to fix:

- [x] It's requiring (`!`) the delegate method
- [x] The delegate method is not firing

@1ec5 @frederoni any ideas about these issues?

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -133,7 +133,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     
     @IBAction func clearMapPressed(_ sender: Any) {
         clearMap.isHidden = true
-        mapView.removeRoute()
+        mapView.removeRoutes()
         mapView.removeWaypoints()
         waypoints.removeAll()
         multipleStopsAction.isEnabled = false
@@ -157,13 +157,12 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
                 print(error!.localizedDescription)
                 return
             }
-            guard let route = routes?.first else { return }
+            guard let routes = routes else { return }
             
-            self?.currentRoute = route
+            self?.currentRoute = routes.first
             
             // Open method for adding and updating the route line
-            self?.mapView.showRoute(route)
-            self?.mapView.showWaypoints(route, legIndex: 0)
+            self?.mapView.showRoutes(routes)
         }
     }
 
@@ -239,6 +238,10 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     // This does however require you to increment the leg count on your own. See the example below in `confirmationControllerDidConfirm()`.
     func navigationViewController(_ navigationViewController: NavigationViewController, shouldIncrementLegWhenArrivingAtWaypoint waypoint: Waypoint) -> Bool {
         return false
+    }
+    
+    func navigationMapView(_ mapView: NavigationMapView, didTap routeIndex: Int, routes: [Route]) {
+        print(routeIndex)
     }
 }
 
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -20,6 +20,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     var currentRoute: Route? {
         didSet {
             self.startButton.isEnabled = currentRoute != nil
+            mapView.showWaypoints(currentRoute!)
         }
     }
     
@@ -159,7 +160,6 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
                 return
             }
             guard let routes = routes else { return }
-            
             self?.currentRoute = routes.first
             
             // Open method for adding and updating the route line
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -14,7 +14,7 @@ enum ExampleMode {
     case multipleWaypoints
 }
 
-class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDelegate {
+class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDelegate, NavigationMapViewDelegate {
     
     var waypoints: [Waypoint] = []
     var currentRoute: Route? {
@@ -52,6 +52,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
         
         automaticallyAdjustsScrollViewInsets = false
         mapView.delegate = self
+        mapView.navigationMapDelegate = self
 
         mapView.userTrackingMode = .follow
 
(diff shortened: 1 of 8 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -852,6 +931,8 @@ public protocol NavigationMapViewDelegate: class  {
     
     @objc optional func navigationMapView(_ mapView: NavigationMapView, routeCasingStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
     
+    @objc optional func navigationMapView(_ mapView: NavigationMapView, didTap routeIndex: Int, routes: [Route])
```
1ec5: This method should return the Route itself, not its index in the list of routes.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -383,16 +400,53 @@ open class NavigationMapView: MGLMapView {
...
+            if let layer = style.layer(withIdentifier: routeLayerCasingIdentifier) {
+                style.insertLayer(alternateLayer, below: layer)
+            }
+        }
+    }
+    
+    /**
+     Fired when the user taps a route. If there is an alternate route, and an alternate route is tapped, it is selected.
+     */
+    public func didTapRoute(tap: UITapGestureRecognizer) {
```
1ec5: Why does this method need to be public, given that it’s the action method of a gesture recognizer that’s internal to this class?

Examples/Swift/ViewController.swift
```diff
@@ -157,13 +157,12 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
                 print(error!.localizedDescription)
                 return
             }
-            guard let route = routes?.first else { return }
+            guard let routes = routes else { return }
```
1ec5: If `routes` is an empty array, should we still continue past this point?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -354,24 +363,32 @@ open class NavigationMapView: MGLMapView {
...
     
+    var activeRouteIndex = 0
+    
+    var routes: [Route]?
+    
     /**
      Adds or updates both the route line and the route line casing
      */
-    public func showRoute(_ route: Route, legIndex: Int? = nil) {
+    public func showRoutes(_ routes: [Route], activeRouteIndex: Int = 0, legIndex: Int = 0) {
```
1ec5: Assume that `routes` is ordered from front to back or back to front (your choice), so `activeRouteIndex` becomes redundant.

1ec5: FYI, making `legIndex` non-optional is a breaking change.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -383,16 +400,53 @@ open class NavigationMapView: MGLMapView {
...
         }
+        guard routes.count > 1 else { return }
+        var tmpRoutes = routes
+        tmpRoutes.remove(at: activeRouteIndex)
+        guard let alternateRoute = tmpRoutes.first else { return }
+        
+        self.routes = routes
+        
+        let polyline = MGLPolylineFeature(coordinates: alternateRoute.coordinates!, count: alternateRoute.coordinateCount)
+        polyline.attributes[routeIndexAttribute] = routes.index(of: alternateRoute)
```
1ec5: Once this method runs to completion, the map will contain features that are associated with an index into a temporary array that has gone out of scope. There is no guarantee that this index means anything past this point.

NavigationMapView could hold onto its displayed routes in a property, perhaps in a dictionary that maps Routes to feature IDs or vice versa. Better yet, we can use Turf methods to find the route that comes closest to the tap point without relying on feature querying or indices.

MapboxNavigation/NavigationMapView.swift
bsudekum: I agree, but it more explicit to name it. We apply the same logic to legs (legIndex) and step (stepIndex) as well.

bsudekum: I'm fine with this change. The way this func was designed before seems incorrect.

Examples/Swift/ViewController.swift
bsudekum: No, and we don't(?).

bsudekum:
@1ec5 @frederoni ok this is ready for review

MapboxNavigation/NavigationMapView.swift
1ec5: When it comes to tracking the current step in the leg, we have no choice, because the order of the steps matters and we can’t rearrange them arbitrarily. The order of routes is unimportant and arbitrary, other than having the active one be first or last. (In fact, if not for the need to mark one of them as the active route, I would advocate making it a set of routes.)

As precedent, the map SDK orders style layers from front to back in drawing order.

MapboxNavigation/CGPoint.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+extension CGPoint {
+    func CGPointDistanceSquared(from: CGPoint, to: CGPoint) -> CGFloat {
```
1ec5: This is an instance method in an extension, whereas the method is named as if it’s a standalone function. It’s weird that you can call `point1.CGPointDistanceSquared(from: point2, to: point3)`.

Replace `from` with `self` and remove “CGPoint” from the method name. Also reconsider whether it’s even necessary to break this method out separately.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -561,6 +641,23 @@ open class NavigationMapView: MGLMapView {
         return MGLShapeCollectionFeature(shapes: features)
     }
     
+    func routeAlternateStyleLayer(identifier: String, source: MGLSource) -> MGLStyleLayer {
```
1ec5: `alternateRouteStyleLayer` is more grammatical.

bsudekum:
Route line under labels also regressed in this PR:

image

/cc @brsbl

bsudekum:
Can't reproduce the layering issue now. Anyways, this is ready for review.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -383,16 +393,71 @@ open class NavigationMapView: MGLMapView {
...
+        }.min { (left, right) -> Bool in
+            let leftDistance = Polyline(left.coordinates!).closestCoordinate(to: tapCoordinate)!.distance
+            let rightDistance = Polyline(right.coordinates!).closestCoordinate(to: tapCoordinate)!.distance
+            return leftDistance < rightDistance
+        }
+        
+        if let closestRoute = closestRoute, let closestCoordinate = Polyline(closestRoute.coordinates!).closestCoordinate(to: tapCoordinate), let routeIndex = routes.index(where: { $0 ==  closestRoute}) {
+            let closestPoint = self.convert(closestCoordinate.coordinate, toPointTo: self)
+            let tapDistanceFromClosestRoute = closestPoint.distance(to: tapPoint)
+            if tapDistanceFromClosestRoute <= 50 {
```
JThramer: This should be a constant.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -39,6 +39,11 @@ open class NavigationMapView: MGLMapView {
      */
     public static let longManeuverDistance: CLLocationDistance = 1000.0
     
+    /**
+     Minimum distnace the user can tap for a selection to be valid when selecting an alternate route.
+     */
+    public var minimumTapDistanceToSelectAltRoute: CGFloat = 50
```
JThramer: maximum, not minimum. If the tap distance is greater than this, we do not select the route.

PR #800, opened 2017-11-06
Expose delegate method for failed voice instructions

bsudekum:
This exposes a delegate method for any failure related to voice instructions.

Todo:
- [x] Document.
- [x] Detect when the audioController is interrupted, send error.

/cc @mapbox/guidance @1ec5 @frederoni @JThramer @ericrwolfe


MapboxNavigation/PollyVoiceController.swift
```diff
@@ -250,21 +250,29 @@ public class PollyVoiceController: RouteVoiceController {
...
-                self.callSuperSpeak(self.fallbackText, error: error.localizedDescription)
+                self.callSuperSpeak(self.fallbackText, error: error)
             }
         }
     }
 }
+
+extension NSError {
+    static func withCustom(message: String) -> NSError {
+        return NSError(domain: "MBVoiceErrorDomain", code: -1, userInfo: [
```
bsudekum: Are there any improvements that could be made to this object?

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -29,7 +29,7 @@ public class PollyVoiceController: RouteVoiceController {
     /**
      Number of seconds a Polly request can wait before it is canceled and the default speech synthesizer speaks the instruction.
      */
-    public var timeoutIntervalForRequest:TimeInterval = 2
+    public var timeoutIntervalForRequest:TimeInterval = 5
```
bsudekum: Also going to bump this up. 2 seconds is just too short.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -208,3 +210,7 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         speechSynth.speak(utterance)
     }
 }
+
+@objc public protocol VoiceControllerDelegate {
+    @objc optional func spokenInstructionsDidFail(_ voiceController: RouteVoiceController, error: Error)
```
1ec5: Is it _failure_ if we fall back to the system voice, or is it a _fallback_ situation? Perhaps we should scope this protocol more tightly to be just about PollyVoiceController: PollyVoiceController failing is the same thing as PollyVoiceController falling back to VoiceController.

bsudekum: Another case to think about though is for overlapping instructions. We're not falling back, but it is an error.

bsudekum: Yeah, i think you're right. Going to move to pollyvoicecontroller

bsudekum:
@1ec5 updated to use custom error codes.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -58,14 +58,19 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     public var bufferBetweenAnnouncements: TimeInterval = 3
     
     /**
+     Delegate used for getting metadata information about a particular spoken instruction.
+     */
+    public var voiceControllerDelegate: VoiceControllerDelegate?
```
frederoni: This should be a weak var to avoid a retain cycle.

MapboxNavigation/Error.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+extension NSError {
+    /**
+     Creates a custom `Error` object.
+     */
+    public static func withCustom(message: String, errorCode: SpokenVoiceError = .defaultError) -> NSError {
```
1ec5: Make this method an initializer instead: `NSError(localizedFailureReason:code:)`

MapboxNavigation/Error.swift
```diff
@@ -0,0 +1,13 @@
...
+        return NSError(domain: "MBErrorDomain", code: errorCode.rawValue, userInfo: [
```
1ec5: Pull this domain out into a global constant, along the lines of the map SDK’s `MGLErrorDomain`.

MapboxNavigation/ErrorCodes.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+
+/**
+ Enum used for indicating the type of error occured while speaking an instruction.
+ */
+public enum SpokenVoiceError: Int {
```
1ec5: There should be a single enumeration with all the error codes we may need to use in this library, similar to the map SDK’s `MGLErrorCode`.

MapboxNavigation/ErrorCodes.swift
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     Default error.
+     */
+    case defaultError
```
1ec5: The default case should be `unknown`, set to −1.

MapboxNavigation/ErrorCodes.swift
```diff
@@ -0,0 +1,36 @@
...
+    
+    /**
+     The audio player failed to preprare to play audio data.
+     */
+    case audioPlayerFailedToPrepare
+    
+    /**
+     The audio player failed to play audio data.
+     */
+    case audiPlayerFailedToPlay
```
1ec5: Typo: audio.

MapboxNavigation/ErrorCodes.swift
```diff
@@ -0,0 +1,36 @@
...
+public enum SpokenVoiceError: Int {
+    /**
+     Default error.
+     */
+    case defaultError
+    
+    /**
+     An intruction interrupted another instruction.
+     */
+    case overlappingInstruction
```
1ec5: Is this really an error? Unlike the other cases, one instruction interrupting another instruction can be completely expected, like when the user triggers a reroute while an instruction is being read aloud.

MapboxNavigation/ErrorCodes.swift
```diff
@@ -0,0 +1,36 @@
...
+    
+    /**
+     The audio player failed to preprare to play audio data.
+     */
+    case audioPlayerFailedToPrepare
```
1ec5: Some of these cases are specific to PollyVoiceController (or at least are inapplicable to RouteVoiceController). Perhaps we can generalize these cases so they can be used for more kinds of voice controllers in the future.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -182,16 +187,16 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         return true
     }
     
-    func speak(_ text: String, error: String? = nil) {
+    func speak(_ text: String, error: Error?) {
         // Note why it failed
         if let error = error {
-            print(error)
+            voiceControllerDelegate?.spokenInstructionsDidFail?(self, error: error)
```
1ec5: We need to reapply the refactor that was proposed in #624: the caller should be responsible for handling errors. Making this method the bottleneck for error reporting is a premature optimization.

MapboxNavigation/ErrorCodes.swift
bsudekum: We need to know when something like this happens, as it _shouldn't_. Would you rather have this filed under general?

MapboxNavigation/RouteVoiceController.swift
bsudekum: @1ec5 I'm not seeing a proposal for fixing this line in #624.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -181,12 +184,12 @@ public class PollyVoiceController: RouteVoiceController {
...
     func handle(_ awsTask: AWSTask<NSURL>) {
         guard awsTask.error == nil else {
-            callSuperSpeak(fallbackText, error: awsTask.error!.localizedDescription)
+            callSuperSpeak(fallbackText, error: awsTask.error!)
             return
         }
         
         guard let url = awsTask.result else {
-            callSuperSpeak(fallbackText, error: "No polly response")
+            callSuperSpeak(fallbackText, error: NSError.withCustom(message: "No AWS Task"))
```
1ec5: `localizedFailureReason` is meant to be human-readable and understandable by the end user, in case the developer presents the error via an alert box (not required). The best practice would be to make the NSError’s error code something generic, like `spokenInstructionFailed`, and its localized failure reason also vague, like “Unable to read aloud the instruction ‘…’.” Then we can add another field to the NSError’s user info dictionary that indicates exactly what went wrong (in this case, an AWS failure).

MapboxNavigation/RouteVoiceController.swift
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/624/files#diff-eaaa217c63f0df828e942d653c6994ceL195

MapboxNavigation/ErrorCodes.swift
1ec5: I think I’d rather that overlapping instructions not result in an “error” at all, but it would be fine to have a dedicated delegate method indicating that an instruction was interrupted before completion.

MapboxNavigation/Error.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+extension NSError {
+    /**
+     Creates a custom `Error` object.
+     */
+    public convenience init(localizedFailureReason: String, detailedFailureReason: String, code: MapboxNavigationError = .defaultError) {
+        self.init(domain: MGLErrorDomain, code: code.rawValue, userInfo: [
+            NSLocalizedFailureReasonErrorKey: localizedFailureReason,
+            "MapboxErrorKey": detailedFailureReason
```
bsudekum: @1ec5 What should this key be?

frederoni: `NSLocalizedDescriptionKey`

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -189,7 +189,7 @@ public class PollyVoiceController: RouteVoiceController {
         }
         
         guard let url = awsTask.result else {
-            callSuperSpeak(fallbackText, error: NSError.withCustom(message: "No AWS Task"))
+            callSuperSpeak(fallbackText, error: NSError(localizedFailureReason: "Unable to read instruction aloud", detailedFailureReason: "No AWS Task"))
```
1ec5: What I meant in https://github.com/mapbox/mapbox-navigation-ios/pull/800#discussion_r149533375 is that this error would contain:

* `NSLocalizedFailureReasonErrorKey`: `Unable to say “In 5 miles, turn left onto Main Street”`
* `SpokenInstructionErrorCodeKey`: SpokenInstructionErrorCode.noAwsTask

The application can then do something different by switching on SpokenInstructionErrorCodeKey rather than doing string comparisons.

MapboxCoreNavigation/Constants.swift
```diff
@@ -152,3 +152,6 @@ public var RouteControllerMinimumNumberLocationUpdatesBackwards = 3
 public var RouteControllerNumberOfSecondsForRerouteFeedback: TimeInterval = 10
 
 let FasterRouteFoundEvent = "navigation.fasterRoute"
+
+let MGLErrorDomain = "MGLErrorDomain"
```
1ec5: This constant needs to be public so developers can conditionalize logic in exception handlers based on it. It also needs a little documentation.

MapboxNavigation/Error.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+extension NSError {
+    /**
+     Creates a custom `Error` object.
+     */
+    public convenience init(localizedFailureReason: String, detailedFailureReason: String, code: MapboxNavigationError = .defaultError) {
+        self.init(domain: MGLErrorDomain, code: code.rawValue, userInfo: [
+            NSLocalizedFailureReasonErrorKey: localizedFailureReason,
+            NSLocalizedDescriptionKey: detailedFailureReason
```
1ec5: https://github.com/mapbox/mapbox-navigation-ios/pull/800#discussion_r149804673

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -207,3 +212,7 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         speechSynth.speak(utterance)
     }
 }
+
+@objc public protocol VoiceControllerDelegate {
+    @objc optional func spokenInstructionsDidFail(_ voiceController: RouteVoiceController, error: Error)
```
1ec5: Delegate methods should always accept the delegator object as the first parameter and be named based on the delegator object. So this method should be `voiceController(_:spokenInstructionsDidFailWith:)`. Also, the Objective-C signature needs to include the nouns: `-voiceController:spokenInstructionsDidFailWithError:`.

MapboxNavigation/ErrorCodes.swift
1ec5: > a dedicated delegate method indicating that an instruction was interrupted before completion

For example, `voiceController(_:didInterrupt:with:)`, taking two SpokenInstruction objects.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -57,14 +57,19 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
      Default initializer for `RouteVoiceController`.
      */
     override public init() {
         super.init()
         
         if !Bundle.main.backgroundModes.contains("audio") {
-            print("Voice guidance may not work properly. " +
-                  "Add audio to the UIBackgroundModes key to your app’s Info.plist file")
+            let errorMessage = "Voice guidance may not work properly. Add audio to the UIBackgroundModes key to your app’s Info.plist file"
+            voiceControllerDelegate?.spokenInstructionsDidFail?(self, error: NSError(localizedFailureReason: "Unable to read instruction aloud", detailedFailureReason: errorMessage, code: .audioBackgroundModeMissing)) ?? print(errorMessage)
```
1ec5: I think we should make this flag a hard requirement for the navigation SDK and maybe have it fail earlier, when creating the NavigationViewController. At least make it an assertion, because the user should be able to expect to hear the same spoken instructions while the application is in the background.

bsudekum:
@1ec5 this is ready for review again.

MapboxNavigation/Error.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+
+/**
+ Custom Error code key used in returned in `voiceController(_:didInterrupt:with:)`.
+ */
+public let SpokenInstructionErrorCodeKey = "SpokenInstructionErrorCodeKey"
```
1ec5: A global constant that’s accessible to Objective-C needs to use the class prefix `MB` in both its name and value. Additionally, the word “Key” should only appear in the name of the constant, not its value.

MapboxNavigation/Error.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+
+/**
+ Custom Error code key used in returned in `voiceController(_:didInterrupt:with:)`.
```
1ec5: Fully qualify the reference to this method name so that jazzy can autolink it.

MapboxNavigation/Error.swift
```diff
@@ -0,0 +1,20 @@
...
+public let SpokenInstructionErrorCodeKey = "SpokenInstructionErrorCodeKey"
+
+extension NSError {
+    /**
+     Creates a custom `Error` object.
+     */
+    public convenience init(localizedFailureReason: String, detailedFailureReason: String, code: MapboxNavigationError = .defaultError) {
+        self.init(domain: MGLErrorDomain, code: code.rawValue, userInfo: [
+            NSLocalizedFailureReasonErrorKey: localizedFailureReason,
+            SpokenInstructionErrorCodeKey: detailedFailureReason
```
1ec5: `MBSpokenInstructionErrorCodeKey` should be set to an enumeration that a developer can easily `switch` on, not a string.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/800#discussion_r149804673

MapboxCoreNavigation/Constants.swift
```diff
@@ -157,3 +157,9 @@ public var RouteControllerNumberOfSecondsForRerouteFeedback: TimeInterval = 10
 public var RouteControllerOpportunisticReroutingInterval: TimeInterval = 120
 
 let FasterRouteFoundEvent = "navigation.fasterRoute"
+
+/**
+ Domain which all errors from delegate methods live under.
+ */
+public let MGLErrorDomain = "MGLErrorDomain"
```
1ec5: `MGLErrorDomain` is the name of the constant used by the map SDK. Our non-GL libraries use `MB` as the class prefix, not `MGL`.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -46,6 +46,8 @@ public class PollyVoiceController: RouteVoiceController {
     
     var spokenInstructionsForRoute = NSCache<NSString, NSData>()
     
+    var previousInstruction: String?
```
1ec5: Rename this property to `lastSpokenInstruction` and set it to an Instruction, not just its SSML text.

Examples/Swift/ViewController.swift
```diff
@@ -282,37 +282,47 @@ extension ViewController: NavigationMapViewDelegate {
...
         self.present(actionSheet, animated: true, completion: nil)
     }
+
+    // Called when there is an error with speaking a voice instruction.
+    func voiceController(_ voiceController: RouteVoiceController, spokenInstructionsDidFailWith error: Error) {
+        print(error.localizedDescription)
+    }
+    
+    // Called when an instruction is interrupted by a new voice instruction.
+    func voiceController(_ voiceController: RouteVoiceController, didInterrupt: String, with instruction: String) {
```
1ec5: This method should accept two Instruction objects as parameters, not just strings. That way the implementation can include logic based on the distance between the instructions, for example.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -69,8 +71,12 @@ public class PollyVoiceController: RouteVoiceController {
         guard let instruction = routeProgresss.currentLegProgress.currentStepProgress.currentSpokenInstruction?.ssmlText else { return }
         
         pollyTask?.cancel()
+        if let audioPlayer = audioPlayer, audioPlayer.isPlaying, let previousInstruction = previousInstruction {
+            voiceControllerDelegate?.voiceController?(self, didInterrupt: previousInstruction, with: instruction)
+        }
         audioPlayer?.stop()
```
1ec5: For clarify, move this line up into the `if let`.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -57,14 +57,18 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
+    /**
      Default initializer for `RouteVoiceController`.
      */
     override public init() {
         super.init()
         
         if !Bundle.main.backgroundModes.contains("audio") {
-            print("Voice guidance may not work properly. " +
-                  "Add audio to the UIBackgroundModes key to your app’s Info.plist file")
+            assert(false, "Voice guidance may not work properly. Add audio to the UIBackgroundModes key to your app’s Info.plist file")
```
1ec5: > Voice guidance may not work properly.

With this assertion, voice guidance _definitely_ won’t work properly – the application will crash before voice guidance could ever work. How about:

> This application’s Info.plist file must include “audio” in UIBackgroundModes. This background mode is used for spoken instructions while the application is in the background.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -207,3 +211,8 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         speechSynth.speak(utterance)
     }
 }
+
+@objc public protocol VoiceControllerDelegate {
+    @objc optional func voiceController(_ voiceController: RouteVoiceController, spokenInstructionsDidFailWith error: Error)
```
1ec5: The Objective-C signature of this method is now `-voiceController:spokenInstructionsDidFailWith:`. Objective-C selector pieces should end in nouns, not prepositions like in Swift. Use `@objc(…)` to customize the Objective-C selector.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -181,16 +185,16 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         return true
     }
     
-    func speak(_ text: String, error: String? = nil) {
+    func speak(_ text: String, error: Error?) {
         // Note why it failed
         if let error = error {
-            print(error)
+            voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
```
1ec5: It should be the responsibility of the caller of this method to deal with any error that is being passed in. This method should only be responsible for handling errors that arise in the execution of this method.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/800#discussion_r149505637

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -166,7 +172,7 @@ public class PollyVoiceController: RouteVoiceController {
         }
     }
     
-    func callSuperSpeak(_ text: String, error: String) {
+    func callSuperSpeak(_ text: String, error: Error) {
```
1ec5: `speakWithoutPolly(for:userDistance:)`

https://github.com/mapbox/mapbox-navigation-ios/blob/2097f6d5cb95cdfa154d5c56560a8c0ffa8e107e/MapboxNavigation/PollyVoiceController.swift#L120

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -181,16 +185,16 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         return true
     }
     
-    func speak(_ text: String, error: String? = nil) {
+    func speak(_ text: String, error: Error?) {
```
1ec5: As part of #624, I proposed renaming this method:

https://github.com/mapbox/mapbox-navigation-ios/blob/2097f6d5cb95cdfa154d5c56560a8c0ffa8e107e/MapboxNavigation/RouteVoiceController.swift#L196

and having `alertLevelDidChange(_:)` set up the input data for this method instead of storing the input data on instance properties:

https://github.com/mapbox/mapbox-navigation-ios/blob/2097f6d5cb95cdfa154d5c56560a8c0ffa8e107e/MapboxNavigation/RouteVoiceController.swift#L160-L165

PollyVoiceController overrides `speak(for:userDistance:)` instead of `alertLevelDidChange(_:)`:

https://github.com/mapbox/mapbox-navigation-ios/blob/2097f6d5cb95cdfa154d5c56560a8c0ffa8e107e/MapboxNavigation/PollyVoiceController.swift#L51

Overall, PollyVoiceController and RouteVoiceController end up being a lot easier to understand to those who haven’t been involved in writing these classes.

MapboxNavigation/ErrorCodes.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+/**
+ Enum used for indicating the type of error occured while speaking an instruction.
+ */
+public enum MapboxNavigationError: Int {
```
1ec5: Rename this enumeration to `MBErrorCode` for consistency with `MGLErrorCode` in the map SDK.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -479,6 +482,7 @@
 		C5ADFBF91DDCC9580011824B /* RouteController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteController.swift; sourceTree = "<group>"; };
 		C5ADFBFB1DDCC9AD0011824B /* RouteProgress.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteProgress.swift; sourceTree = "<group>"; };
 		C5CFE4871EF2FD4C006F48E8 /* MMEEventsManager.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = MMEEventsManager.swift; sourceTree = "<group>"; };
+		C5D1C9931FB236900067C619 /* ErrorCodes.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ErrorCodes.swift; sourceTree = "<group>"; };
```
1ec5: Rename this file to ErrorCode.swift for consistency with the `MBErrorCode` enumeration name.

bsudekum:
@1ec5 refactored a bit with your feedback. I also removed the overlap timers in https://github.com/mapbox/mapbox-navigation-ios/pull/800/commits/8689cc99856e7c90e16fb2f7abe6b8e9fd85f74e- this is the servers job now.

MapboxCoreNavigation/MBRouteController.m
```diff
@@ -17,3 +17,5 @@
 NSString *const MBRouteControllerDidReroute                     = @"RouteControllerDidReroute";
 NSString *const MBRouteControllerDidFailToReroute               = @"RouteControllerDidFailToReroute";
 NSString *const MBRouteControllerDidFindFasterRouteKey          = @"RouteControllerDidFindFasterRoute";
+NSString *const MBErrorDomain = @"ErrorDomain";
+NSString *const MBSpokenInstructionErrorCode = @"SpokenInstructionErrorCodeKey";
```
1ec5: Add “Key” to the name of the constant and remove it from the constant’s value.

MapboxNavigation/Error.swift
```diff
@@ -1,19 +1,14 @@
...
-public let SpokenInstructionErrorCodeKey = "SpokenInstructionErrorCodeKey"
+import MapboxCoreNavigation
 
 extension NSError {
     /**
      Creates a custom `Error` object.
      */
-    public convenience init(localizedFailureReason: String, detailedFailureReason: String, code: MapboxNavigationError = .defaultError) {
-        self.init(domain: MGLErrorDomain, code: code.rawValue, userInfo: [
+    public convenience init(localizedFailureReason: String, code: MapboxNavigationError = .defaultError) {
```
1ec5: This initializer is only useful to this SDK’s code, so make it private if possible.

MapboxNavigation/Error.swift
```diff
@@ -1,19 +1,14 @@
...
+        self.init(domain: MBErrorDomain, code: code.rawValue, userInfo: [
             NSLocalizedFailureReasonErrorKey: localizedFailureReason,
-            SpokenInstructionErrorCodeKey: detailedFailureReason
+            MBSpokenInstructionErrorCode: code
```
1ec5: Sorry for not being clearer in https://github.com/mapbox/mapbox-navigation-ios/pull/800#discussion_r149533375, https://github.com/mapbox/mapbox-navigation-ios/pull/800#discussion_r149804673, and https://github.com/mapbox/mapbox-navigation-ios/pull/800#discussion_r150964273:

* This method should accept a `spokenInstructionCode` parameter of type SpokenInstructionErrorCode.
* SpokenInstructionErrorCode should be an enumeration with the following values: `noAwsResult`, `emptyAwsResponse`, and `audioPlayerFailedToPlay`.
* MBErrorCode (renamed from MapboxNavigationError) should have the following values: `unknown` and `spokenInstructionFailed`.
* One of the call sites in PollyVoiceController should call this initializer passing in `spokenInstructionFailed` as the `code` and `noAwsResult` as the `spokenInstructionCode`.

MapboxCoreNavigation/MBRouteController.h
```diff
@@ -17,3 +17,5 @@ extern NSString *const MBRouteControllerWillReroute;
 extern NSString *const MBRouteControllerDidReroute;
 extern NSString *const MBRouteControllerDidFailToReroute;
 extern NSString *const MBRouteControllerDidFindFasterRouteKey;
+extern NSString *const MBErrorDomain;
+extern NSString *const MBSpokenInstructionErrorCode;
```
1ec5: These constants (like all the others here) need documentation comments so that they appear in the jazzy documentation.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -46,7 +47,7 @@ public class PollyVoiceController: RouteVoiceController {
     
     var spokenInstructionsForRoute = NSCache<NSString, NSData>()
     
-    var previousInstruction: String?
+    var previousInstruction: SpokenInstruction?
```
1ec5: Rename this property to `lastSpokenInstruction`. Between spoken instructions, it makes more sense to talk about the last instruction that was said than the “previous” one.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -172,27 +171,29 @@ public class PollyVoiceController: RouteVoiceController {
...
     func handle(_ awsTask: AWSTask<NSURL>) {
         guard awsTask.error == nil else {
-            callSuperSpeak(fallbackText, error: awsTask.error!)
+            speakWithoutPolly(fallbackInstruction, error: awsTask.error!)
             return
         }
         
         guard let url = awsTask.result else {
-            callSuperSpeak(fallbackText, error: NSError(localizedFailureReason: "Unable to read instruction aloud", detailedFailureReason: "No AWS Task"))
+            speakWithoutPolly(fallbackInstruction, error: NSError(localizedFailureReason: "Unable to read instruction aloud"))
```
1ec5: Localized failure reasons should be localizable.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -180,24 +167,19 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         // Set recentlyAnnouncedRouteStep to the current step
         recentlyAnnouncedRouteStep = routeProgress.currentLegProgress.currentStep
         
-        fallbackText = routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction?.text
+        fallbackInstruction = routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction
```
1ec5: Remove this line and remove the `fallbackInstruction` property. From now on, it should be the responsibility of `alertLevelDidChange(_:)` to pass `currentSpokenInstruction` directly into `speak(_:)`.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -68,14 +69,13 @@ public class PollyVoiceController: RouteVoiceController {
         guard shouldSpeak(for: notification) == true else { return }
```
1ec5: Move everything below here into `speak(_:)` and delete this method implementation, including this `guard` statement. It should no longer be necessary for PollyVoiceController to override this method.

MapboxNavigation/RouteVoiceController.swift
1ec5: This method still needs a proper Objective-C signature.

bsudekum:
@1ec5 this is ready for review again.

MapboxCoreNavigation/MBRouteController.h
```diff
@@ -17,5 +17,8 @@ extern NSString *const MBRouteControllerWillReroute;
 extern NSString *const MBRouteControllerDidReroute;
 extern NSString *const MBRouteControllerDidFailToReroute;
 extern NSString *const MBRouteControllerDidFindFasterRouteKey;
+
+/**
+ Constant representing the namespace in which errors created in this library will live under.
```
1ec5: s/namespace/domain/

MapboxCoreNavigation/MBRouteController.m
```diff
@@ -17,5 +17,8 @@
 NSString *const MBRouteControllerDidReroute                     = @"RouteControllerDidReroute";
 NSString *const MBRouteControllerDidFailToReroute               = @"RouteControllerDidFailToReroute";
 NSString *const MBRouteControllerDidFindFasterRouteKey          = @"RouteControllerDidFindFasterRoute";
+
+/**
+ Domain for errors created within this framework.
+ */
```
1ec5: Merge this comment with the one in the header. A public Objective-C symbol should be documented only on the declaration (in the header, in this case), not on the definition.

MapboxNavigation/Error.swift
```diff
@@ -5,10 +5,9 @@ extension NSError {
     /**
      Creates a custom `Error` object.
      */
-    public convenience init(localizedFailureReason: String, code: MapboxNavigationError = .defaultError) {
+    convenience init(code: MBErrorCode, localizedFailureReason: String, spokenInstructionCode: SpokenInstructionErrorCode? = nil) {
         self.init(domain: MBErrorDomain, code: code.rawValue, userInfo: [
-            NSLocalizedFailureReasonErrorKey: localizedFailureReason,
-            MBSpokenInstructionErrorCode: code
+            NSLocalizedFailureReasonErrorKey: localizedFailureReason
```
1ec5: `spokenInstructionCode` appears to be unused. Add an entry here that sets `MBSpokenInstructionErrorCodeKey` to `spokenInstructionCode`, but only if it’s provided.

MapboxNavigation/ErrorCode.swift
```diff
@@ -19,3 +19,18 @@ public enum MapboxNavigationError: Int {
      */
     case noDataInSpokenInstructionResponse
```
1ec5: Rename `noDataInSpokenInstructionResponse` to `emptyAwsResponse`, since it’s assumed that any case in this enumeration is about spoken instructions.

Previous iterations of this PR also had error codes for `noAwsResult` and `audioPlayerFailedToPlay`. Do we still need them?

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -47,7 +47,9 @@ public class PollyVoiceController: RouteVoiceController {
     
     var spokenInstructionsForRoute = NSCache<NSString, NSData>()
     
-    var previousInstruction: SpokenInstruction?
+    var lastSpokenInstruction: SpokenInstruction?
+    
+    let localizedErrorMessage = NSLocalizedString("FAILED_INSTRUCTION", bundle: .mapboxNavigation, value: "Unable to read instruction aloud", comment: "Unable to read instruction aloud")
```
1ec5: The comment should explain to translators how this string is used:

> Error message when the SDK is unable to read a spoken instruction.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -66,17 +68,15 @@ public class PollyVoiceController: RouteVoiceController {
     }
     
     public override func didPassSpokenInstructionPoint(notification: NSNotification) {
```
1ec5: PollyVoiceController should override `speak(_:)`, but should have no reason to override `didPassSpokenInstructionPoint(_:)`. Move everything from this method into `speak(_:)`, except for the statements that are also in `RouteVoiceController.didPassSpokenInstructionPoint(_:)`. Then remove this method override. The implementation in `RouteVoiceController.didPassSpokenInstructionPoint(_:)` will get called on PollyVoiceController whenever we pass a spoken instruction point.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -177,23 +178,23 @@ public class PollyVoiceController: RouteVoiceController {
         voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
         
         guard let audioPlayer = audioPlayer else {
-            super.speak(fallbackInstruction)
+            super.speak(lastSpokenInstruction!)
```
1ec5: Pass in `instruction` instead of `lastSpokenInstuction`. Ditto below.

MapboxNavigation/Resources/Base.lproj/Localizable.strings
```diff
@@ -1,6 +1,9 @@
 ﻿/* Dismiss button title on the steps view */
 "DISMISS_STEPS_TITLE" = "Close";
 
+/* Unable to read instruction aloud */
+"FAILED_INSTRUCTION" = "Unable to read instruction aloud";
```
1ec5: A human-readable error message should read like a complete sentence, so it should end with a period.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -140,39 +137,18 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
     }
     
     open func didPassSpokenInstructionPoint(notification: NSNotification) {
-        guard shouldSpeak(for: notification) == true else { return }
-        
-        speak(fallbackInstruction)
-    }
-    
-    func shouldSpeak(for notification: NSNotification) -> Bool {
-        guard isEnabled, volume > 0, !NavigationSettings.shared.muted else { return false }
```
1ec5: It would’ve been fine to keep `shouldSpeak(for:)` around, but in any case, make sure these `guard` statements are in `RouteVoiceController.didPassSpokenInstructionPoint(_:)`, _not_ `PollyVoiceController.didPassSpokenInstructionPoint(_:)` – the latter method should go away.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -140,39 +137,18 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
-        
-        fallbackInstruction = routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction
-        
-        return true
+        guard let instruction = routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction else { return }
+        speak(instruction)
     }
     
     func speak(_ instruction: SpokenInstruction) {
+        guard isEnabled, volume > 0, !NavigationSettings.shared.muted else { return }
```
1ec5: Move this statement up to `didPassSpokenInstructionPoint(_:)` or keep `shouldSpeak(for:)` around.

1ec5:
Following the refactoring, PollyVoiceController should work like this:

* `RouteVoiceController.didPassSpokenInstructionPoint(_:)` gets called. It consults `RouteVoiceController.shouldSpeak(for:)` as to whether it should continue or bail (or that logic can be inlined in `didPassSpokenInstructionPoint(_:)` – either way’s fine.
* `RouteVoiceController.didPassSpokenInstructionPoint(_:)` calls `speak(_:)`. Since PollyVoiceController has overridden `speak(_:)`, `PollyVoiceController.speak(_:)` gets called.
* Various code paths in `PollyVoiceController.speak(_:)` may call `RouteVoiceController.speak(_:)` via `PollyVoiceController.speakWithoutPolly(_:error:)`.

At no point does `PollyVoiceController.didPassSpokenInstructionPoint(_:)` ever get called, even for PollyVoiceController.

mcwhittemore:
I keep forgetting to say how excited I am about this. I'm very excited! :)

MapboxNavigation/ErrorCode.swift
bsudekum: Yes, `.spokenInstructionAudioPlayerFailedToPlay` are still used.

bsudekum:
@1ec5 alright, I think we're there.

MapboxNavigation/ErrorCode.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+
+/**
+ Enum used for indicating the type of error occured while speaking an instruction.
+ */
+public enum SpokenInstructionErrorCode: Int {
+    /**
+     Default error.
+     */
+    case defaultError = -1
```
1ec5: `unknown`, for consistency with `MBErrorCode.unknown`.

MapboxNavigation/ErrorCode.swift
```diff
@@ -0,0 +1,36 @@
...
+    
+    /**
+     The audio player failed to play audio data.
+     */
+    case spokenInstructionAudioPlayerFailedToPlay
```
1ec5: `spokenInstruction` is redundant here, because it’s part of an enumeration that contains `SpokenInstruction` in its name. Rename this case to `audioPlayerFailedToPlay`.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -85,12 +91,14 @@ public class PollyVoiceController: RouteVoiceController {
...
-            play(spokenInstructionsForRoute.object(forKey: instruction as NSString)! as Data)
+        guard spokenInstructionsForRoute.object(forKey: instruction.ssmlText as NSString) == nil else {
+            play(spokenInstructionsForRoute.object(forKey: instruction.ssmlText as NSString)! as Data)
             return
         }
         
-        speak(instruction, error: nil)
+        guard isEnabled, volume > 0, !NavigationSettings.shared.muted else { return }
+        
+        super.speak(instruction)
```
1ec5: This causes `RouteVoiceController.speak(_:)` to be called, so `PollyVoiceController.speak(_:)` never gets called.

An override should normally call the superclass’s implementation. So call `super.didPassSpokenInstructionPoint(_:)`. That will allow `RouteVoiceController.didPassSpokenInstructionPoint(_:)` to determine which instruction to pass into `PollyVoiceController.speak(_:)`.

If that back-and-forth concerns you, bring back `RouteVoiceController.shouldSpeak(for:)` and move this prefecthing logic to an override of that method, `PollyVoiceController.shouldSpeak(for:)`. (The override can call the super implementation after performing the prefetching.)

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -166,27 +172,29 @@ public class PollyVoiceController: RouteVoiceController {
...
             return
         }
         
         guard !audioPlayer.isPlaying else { return }
         
-        super.speak(fallbackText, error: error)
+        super.speak(instruction)
     }
     
     func handle(_ awsTask: AWSTask<NSURL>) {
```
1ec5: Pass `instruction` into this method, so that it doesn’t need to use `lastSpokenInstruction`. This keeps the class minimally dependent on instance properties.

MapboxNavigation/PollyVoiceController.swift
1ec5: Fixed in a84a28f2df09f12dbe2f8d6b8768b22b248d19d7.

Issue #795, opened 2017-11-05
When pressing recenter, distance resets value until next location tick


bsudekum:
To reproduce:

1. pan map off user
1. hit recenter
1. notice change in the distance label

It might be the default value found in the storyboard. This value should remain constant.

/cc @frederoni

frederoni:
I think this is by design as of https://github.com/mapbox/mapbox-navigation-ios/pull/769/commits/6388c1d8e44b7af40428398f9692da3c30659634? /cc @JThramer

JThramer:
Correct, this is by design. `adjustCameraAltitude(for:)` is called when you hit the Resume Button. That function sets the camera altitude if one of two conditions are matched:

- It sets the camera to `zoomedOutMotorwayAltitude` (2000m) if you are on a motorway (and not about to exit.)
- It sets the camera to `defaultAltitude` (1000m) if you are on the last instruction of your maneuver.

Otherwise the function does nothing.

JThramer:
I think the issue is coming from https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/RouteMapViewController.swift#L606, but I'm still digging into it.

JThramer:
Issue fixed.

PR #801, opened 2017-11-06
update currentManeuverPage with current values immediately after transition

JThramer:
Implements #795. Not the most elegant solution in the world, but without `RouteManeuverViewController` knowing about the concept of `RouteProgress`/`StepProgress`, `notifyDidChange(routeProgress:stepProgress:)` is the only real way (without going down a rabbit hole) to prompt the RMVC to update the `InstructionsBannerView` to the correct values.

-  Fixing issue where carousel would display incorrect information momentarily when recentering by notifying current maneuver cell about it's progress immediately after presentation.

/cc @bsudekum @frederoni

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -73,9 +73,6 @@ class RouteMapViewController: UIViewController {
                     let remaining = self.routeController.routeProgress.currentLegProgress.currentStepProgress.durationRemaining
                     shown.notifyDidChange(routeProgress: progress, secondsRemaining: remaining)
                 }
-                
-               
-
             }
         }
     }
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -68,7 +68,14 @@ class RouteMapViewController: UIViewController {
...
                 controller.step = currentStep
-                routePageViewController.updateManeuverViewForStep()
+                routePageViewController.updateManeuverViewForStep { (shown) in
+                    let progress = self.routeController.routeProgress
+                    let remaining = self.routeController.routeProgress.currentLegProgress.currentStepProgress.durationRemaining
+                    shown.notifyDidChange(routeProgress: progress, secondsRemaining: remaining)
+                }
+                
+               
+
```
bsudekum: Can you remove this whitespace?

Issue #791, opened 2017-11-03
Add delegate method for selecting a waypoint


bsudekum:
By default, the MapboxNavigation should show an action sheet that confirms the user would like to remove the waypoint when selecting one.

/cc @ericrwolfe

JThramer:
Implemented in #806.

PR #806, opened 2017-11-07
791 waypoint removal

JThramer:
Implementing #791. This is based off of @bsudekum's #789, so I'll change bases to `master` once that gets merged in.

This is definitely still a WIP... it's a working implementation, but buggy, as the two gesture recognizers like to interfere with one another. Still some work to do.

✅  : 

- [x] Combine `UITapGestureRecognizer`s into one that run both (waypoint and route) hit tests
- [x] Fix bug where alternate route disappears after selecting other route
- [x] ~~Figure out why destination waypoint is failing waypoint tap recognizer logic~~ (determined out of scope)
- [x] ~~Create delegate method that asks client if map should remove waypoint from route~~
- [x] ~~Create delegate method that notifies client that routes have changed~~
- [x] ~~Create delegate method that notifies client that route request failed.~~
- [x] Create Extension for `RouteOptions`
- [x] Refactor touch calculation methods to take `CGPoint`s instead of `UITapGestureRecognizer`s
- [x] Extract directions call in `ViewController` into its own function.
- [x] Move all inappropriate `NavigationMapView` logic over to implementation (example) side.
- [x] Create optional `navigationMapView(_:didTapWaypoint:)` for generic waypoint handling
- [x] Fix `didTap:`
- [x] Clean it up so it doesn't look so crufty
- [x]  Add logic to `ViewController` that prompts user to delete waypoint

/cc @bsudekum @frederoni @ericrwolfe

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -53,6 +53,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     
     fileprivate lazy var defaultSuccess: RouteRequestSuccess = { [weak self] (routes) in
         guard let current = routes.first else { return }
+        self?.mapView.removeWaypoints()
         self?.routes = routes
         self?.currentRoute = current
         self?.waypoints = current.routeOptions.waypoints
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -266,10 +266,10 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     }
     
     func navigationMapView(_ mapView: NavigationMapView, didSelect waypoint: Waypoint) {
-        guard let opts = currentRoute?.routeOptions else { return }
-        let modifiedOpts = opts.without(waypoint: waypoint)
+        guard let routeOptions = currentRoute?.routeOptions else { return }
+        let modifiedOptions = routeOptions.without(waypoint: waypoint)
         
-        requestRoute(with:modifiedOpts, success: defaultSuccess, failure: defaultFailure)
+        requestRoute(with:modifiedOptions, success: defaultSuccess, failure: defaultFailure)
     }
     
     func navigationMapView(_ mapView: NavigationMapView, didSelect route: Route) {
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -258,14 +258,14 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
         return false
     }
     
-    func navigationMapView(_ mapView: NavigationMapView, didSelectWaypoint waypoint: Waypoint) {
+    func navigationMapView(_ mapView: NavigationMapView, didSelect waypoint: Waypoint) {
         guard let opts = currentRoute?.routeOptions else { return }
         let modifiedOpts = opts.without(waypoint: waypoint)
         
         requestRoute(with:modifiedOpts, success: defaultSuccess, failure: defaultFailure)
     }
     
-    func navigationMapView(_ mapView: NavigationMapView, didSelectRoute route: Route) {
+    func navigationMapView(_ mapView: NavigationMapView, didSelect route: Route) {
         currentRoute = route
     }
 }
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -133,7 +133,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     
     @IBAction func clearMapPressed(_ sender: Any) {
         clearMap.isHidden = true
-        mapView.removeRoute()
+        mapView.removeRoutes()
         mapView.removeWaypoints()
         waypoints.removeAll()
         multipleStopsAction.isEnabled = false
@@ -157,13 +157,12 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
                 print(error!.localizedDescription)
                 return
             }
-            guard let route = routes?.first else { return }
+            guard let routes = routes else { return }
             
-            self?.currentRoute = route
+            self?.currentRoute = routes.first
             
             // Open method for adding and updating the route line
-            self?.mapView.showRoute(route)
-            self?.mapView.showWaypoints(route, legIndex: 0)
+            self?.mapView.showRoutes(routes)
         }
     }
 
@@ -239,6 +238,10 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     // This does however require you to increment the leg count on your own. See the example below in `confirmationControllerDidConfirm()`.
     func navigationViewController(_ navigationViewController: NavigationViewController, shouldIncrementLegWhenArrivingAtWaypoint waypoint: Waypoint) -> Bool {
         return false
+    }
+    
+    func navigationMapView(_ mapView: NavigationMapView, didTap routeIndex: Int, routes: [Route]) {
+        print(routeIndex)
     }
 }
 
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -6,6 +6,8 @@ import Mapbox
 
 let sourceIdentifier = "sourceIdentifier"
 let layerIdentifier = "layerIdentifier"
+private typealias RouteRequestSuccess = (([Route]) -> Void)
+private typealias RouteRequestFailure = ((NSError) -> Void)
 
 enum ExampleMode {
     case `default`
@@ -145,7 +147,19 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
         present(alertController, animated: true, completion: nil)
     }
     
-    // Helper for requesting a route
+    // MARK: Directions Requests
+    
+    private lazy var defaultSuccess: RouteRequestSuccess = { [weak self] (routes) in
+            guard let current = routes.first else { return }
+            self?.currentRoute = current
+            self?.waypoints = current.routeOptions.waypoints
+            self?.mapView.showRoutes(routes)
+    }
+    
+    private lazy var defaultFailure: RouteRequestFailure = { (error) in
+        print(error.localizedDescription)
+    }
+    
     func requestRoute() {
         guard waypoints.count > 0 else { return }
         
@@ -154,17 +168,20 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
         
         let options = NavigationRouteOptions(waypoints: waypoints)
         
-        _ = Directions.shared.calculate(options) { [weak self] (waypoints, routes, error) in
-            guard error == nil else {
-                print(error!.localizedDescription)
-                return
-            }
-            guard let routes = routes else { return }
-            self?.currentRoute = routes.first
-            
-            // Open method for adding and updating the route line
-            self?.mapView.showRoutes(routes)
+        requestRoute(with: options, success: defaultSuccess, failure: defaultFailure)
+    }
+    
+    private func requestRoute(with options: RouteOptions, success: @escaping RouteRequestSuccess, failure: RouteRequestFailure?) {
+        
+        let handler: Directions.CompletionHandler = {(waypoints, potentialRoutes, potentialError) in
+            if let error = potentialError, let fail = failure { return fail(error) }
+            guard let routes = potentialRoutes else { return }
+            return success(routes)
         }
+        
+        mapView.removeRoutes()
+        mapView.removeWaypoints()
+        _ = Directions.shared.calculate(options, completionHandler: handler)
     }
 
     // MARK: - Basic Navigation
@@ -242,25 +259,14 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     }
     
     func navigationMapView(_ mapView: NavigationMapView, didSelectWaypoint waypoint: Waypoint) {
-        requestNew(route: currentRoute, without: waypoint)
+        guard let opts = currentRoute?.routeOptions else { return }
+        let modifiedOpts = opts.without(waypoint: waypoint)
+        
+        requestRoute(with:modifiedOpts, success: defaultSuccess, failure: defaultFailure)
     }
     
     func navigationMapView(_ mapView: NavigationMapView, didSelectRoute route: Route) {
         currentRoute = route
-    }
-    private func requestNew(route: Route?, without waypoint: Waypoint) {
-        guard let route = route else { return }
-        let options = route.routeOptions.without(waypoint: waypoint)
-        
-        mapView?.removeRoutes()
-        mapView?.removeWaypoints()
-        _ = Directions.shared.calculate(options) { (waypoints, routes, error) in
-            guard let routes = routes, let current = routes.first else { return }
-            self.currentRoute = current
-            self.waypoints = current.routeOptions.waypoints
-            self.mapView?.showRoutes([current])
-        }
-        
     }
 }
 
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -282,11 +282,27 @@ extension ViewController: NavigationMapViewDelegate {
         guard let routeOptions = currentRoute?.routeOptions else { return }
         let modifiedOptions = routeOptions.without(waypoint: waypoint)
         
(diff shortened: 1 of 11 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,7 +439,60 @@ open class NavigationMapView: MGLMapView {
...
+     */
+    func didTapWaypoint(tap: UITapGestureRecognizer) {
+        guard let routes: [Route] = routes else { return }
+        let tapPoint = tap.location(in: self)
+        let tapCoordinate = self.convert(tapPoint, toCoordinateFrom: tap.view)
+        guard let candidates = waypoints(on: routes, closeTo: tapCoordinate) else { return }
+
+        //take the best one, if existant, and prompt to delete
+        guard let best = candidates.1.first, let multipoint = candidates.0.first else { return }
+        requestNew(route: multipoint, without: best)
```
bsudekum: This might be a good place to slip in the delegate method. If the user implements it, give them the waypoint they selected. Otherwise, call provided boilerplate functionality.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,7 +439,60 @@ open class NavigationMapView: MGLMapView {
...
+        return super.gestureRecognizerShouldBegin(gestureRecognizer)
+    }
+    
+    private func requestNew(route: Route, without waypoint: Waypoint) {
+        let options = route.routeOptions
+        guard let index = options.waypoints.index(of: waypoint) else { return }
+        options.waypoints.remove(at: index)
+        
+        removeRoutes()
+        removeWaypoints()
```
bsudekum: No need to remove these; if you call showRoute/waypoint, we update routes/waypoints if they are already on the map.

MapboxNavigation/NavigationMapView.swift
JThramer: Should 'boilerplate functionality' just be a `navigationViewController(_:didSelectWaypoint:)`delegate method or something? Perhaps the implementer doesn't want to delete a waypoint when the user selects one...

bsudekum: Right, so if they do not want this to happen, they implement `navigationViewController(_:didSelectWaypoint:)` and do something custom. Otherwise, we show the action sheet, and re-request the route.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,7 +439,60 @@ open class NavigationMapView: MGLMapView {
...
+        _ = Directions.shared.calculate(options) { (waypoints, routes, error) in
+            guard let routes = routes else { return }
+            self.routes = routes
+            self.showRoutes(routes)
```
bsudekum: In the event of an error, we should keep the previous right and gently notify the user the request failed.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -913,6 +916,7 @@
 				C5D9800E1EFBCDAD006DBF2E /* Date.swift */,
 				C51DF8651F38C31C006C6A15 /* Locale.swift */,
 				C578DA071EFD0FFF0052079F /* ProcessInfo.swift */,
+				8DF49B2F1FB2778A00B2F932 /* Route.swift */,
```
JThramer: This shouldn't be here. Investigate.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,36 +426,96 @@ open class NavigationMapView: MGLMapView {
...
+    }
+    
+    private func requestNew(route: Route, without waypoint: Waypoint) {
+        let options = route.routeOptions
+        guard let index = options.waypoints.index(of: waypoint) else { return }
+        options.waypoints.remove(at: index)
+        
+        removeRoutes()
+        removeWaypoints()
+        _ = Directions.shared.calculate(options) { (waypoints, routes, error) in
```
1ec5: It’s contrary to the MVC pattern for a view to kick off network requests. It should be the responsibility of the map view’s delegate (typically a view controller) to issue this request.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,36 +426,96 @@ open class NavigationMapView: MGLMapView {
...
+        //lets filter to see which ones are under threshold
+        let candidates = closest.filter({
+            let coordinatePoint = self.convert($0.coordinate, toPointTo: self)
+            return coordinatePoint.distance(to: tapPoint) < waypointThreshold
+        })
+        
+        return (multipointRoutes, candidates)
+    }
+    
+    private func routes(closeTo tap: UITapGestureRecognizer) -> [Route]? {
```
1ec5: This method should take a point (in screen coordinates), rather than a gesture recognizer. That’ll make it useful for other purposes too.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,36 +426,96 @@ open class NavigationMapView: MGLMapView {
...
         guard let routes = routes else { return }
         
-        let tapPoint = tap.location(in: self)
-        let tapCoordinate = self.convert(tapPoint, toCoordinateFrom: tap.view)
+        let waypointTest = waypoints(on: routes, closeTo: sender) //are there waypoints near the tapped location?
+        if let activeRoute = waypointTest?.multipointRoutes.first ,let waypointToRemove = waypointTest?.waypoints.first { //test passes
+            return requestNew(route: activeRoute, without: waypointToRemove)
+        } else if let routes = self.routes(closeTo: sender) {
+            guard let selectedRoute = routes.first else { return }
+            select(route: selectedRoute)
```
1ec5: This logic should be moved to the delegate implementation. We can make the helper methods public to facilitate that implementation. After all, the developer may want to programmatically select a route, or they may want to determine the route closest to a view that’s floating above the map view.

If implementing this logic represents an undue burden on the developer, we should vend a prefabricated view controller (#808). Alternatively, we could explore supplementing the existing Objective-C delegate protocol with a Swift protocol that can come with its own default implementations.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,36 +426,96 @@ open class NavigationMapView: MGLMapView {
...
+    private func select(route: Route) {
+        guard var routes = routes, let routeIndex = routes.index(where: {$0 == route}) else { return }
+        routes.remove(at: routeIndex)
+        routes.insert(route, at: 0)
+        self.routes = routes //because self.routes is a value type
+        self.showRoutes(routes)
+        navigationMapDelegate?.navigationMapView?(self, didTap: route)
+    }
+    
+    private func requestNew(route: Route, without waypoint: Waypoint) {
```
1ec5: Extend the RouteOptions class with a method `deleting(_:)` that takes a waypoint and returns a new RouteOptions object.

MapboxNavigation/NavigationMapView.swift
1ec5: What if the view controller hosting the map view already has a way of notifying the user of a routing failure?

bsudekum:
@JThramer I agree with @1ec5  here. Let's expose the delegate method and leave it up to the developer to do the rest. Anything related to making a new request can be removed.

MapboxNavigation/NavigationMapView.swift
JThramer: Good idea. Will implement.

MapboxNavigation/NavigationMapView.swift
JThramer: The problem with this is that we do our distance calculation based upon `CGPoint` distance, not `CLLocationDistance`. I need the recognizer so I can convert coordinates back into screen geometry, and then test if the tap was within 50 points of the closest route coordinate.

MapboxNavigation/NavigationMapView.swift
JThramer: Agreed. @bsudekum's idea of just putting this burden on the implementation side will suffice for now, but yeah, it would be nice if the `NavigationMapView` got its own controller that could provide default implementations for some of this behavior. It's time will come :)

MapboxNavigation/NavigationMapView.swift
1ec5: Agreed, my suggestion is to have the method taking a CGPoint, not a CLLocationCoordinate2D.

JThramer: Apologies, I misread. Will Do.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1016,7 +1002,9 @@ public protocol NavigationMapViewDelegate: class  {
     
     @objc optional func navigationMapView(_ mapView: NavigationMapView, routeCasingStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
     
-    @objc optional func navigationMapView(_ mapView: NavigationMapView, didTap route: Route)
+    @objc optional func navigationMapView(_ mapView: NavigationMapView, didSelectRoute route: Route)
+    
+    @objc optional func navigationMapView(_ mapView: NavigationMapView, didSelectWaypoint waypoint: Waypoint)
```
1ec5: The signature should be `navigationMapView(_:didSelect:)` in Swift but `-navigationMapView:didSelectWaypoint:` in Objective-C. You can customize the Objective-C signature using `@objc(…)`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,36 +426,82 @@ open class NavigationMapView: MGLMapView {
...
+    }
+    
+    private func routes(closeTo point: CGPoint) -> [Route]? {
+        let tapCoordinate = convert(point, toCoordinateFrom: self)
+        
+        //do we have routes? If so, filter routes with at least 2 coordinates.
+        guard let routes = routes?.filter({ $0.coordinates?.count ?? 0 > 1 }) else { return nil }
+        
+        //Sort routes by closest distance to tap gesture.
+        let closest = routes.sorted { (left, right) -> Bool in
```
1ec5: How does this sort perform for long routes, such as cross-country routes?

As long as we can figure out which route polyline feature on the map corresponds to which Route object – perhaps by assigning each feature a unique identifier – we can query the map view for visible features near the tap point to determine the nearest route. Feature querying should be more performant, since it’s powered by an R-tree implemented in C++.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -427,36 +426,82 @@ open class NavigationMapView: MGLMapView {
...
-        let closestRoute = routes.filter {
-            guard let coords = $0.coordinates else { return false }
-            return coords.count > 1
-        }.min { (left, right) -> Bool in
-            let leftDistance = Polyline(left.coordinates!).closestCoordinate(to: tapCoordinate)!.distance
-            let rightDistance = Polyline(right.coordinates!).closestCoordinate(to: tapCoordinate)!.distance
+    func didRecieveTap(sender: UITapGestureRecognizer) {
+        guard let routes = routes, let tapPoint = sender.point else { return }
+        
+        let waypointTest = waypoints(on: routes, closeTo: tapPoint) //are there waypoints near the tapped location?
```
1ec5: If there was a waypoint near the tapped location, its annotation would’ve gotten selected, so `mapView.selectedAnnotations` would be non-empty.

MapboxNavigation/NavigationMapView.swift
JThramer: Strongly agree, Will fix. Ideally it would be nice if the map told the implementer that a map feature was selected (much like we tell a implementer that an annotation has been selected), but that's a different battle, for a different war.

MapboxNavigation/NavigationMapView.swift
JThramer: Intermediate waypoints are map features, not `MGLAnnotation`s. https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/NavigationMapView.swift#L516

MapboxNavigation/NavigationMapView.swift
JThramer: Good catch. Done.

MapboxNavigation/NavigationMapView.swift
1ec5: > Ideally it would be nice if the map told the implementer that a map feature was selected (much like we tell an implementer that an annotation has been selected)

This would be a great feature to have in the map SDK, since it isn’t specific to navigation: mapbox/mapbox-gl-native#6515.

Examples/Swift/ViewController.swift
```diff
@@ -154,17 +168,20 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
...
+    
+    private func requestRoute(with options: RouteOptions, success: @escaping RouteRequestSuccess, failure: RouteRequestFailure?) {
+        
+        let handler: Directions.CompletionHandler = {(waypoints, potentialRoutes, potentialError) in
+            if let error = potentialError, let fail = failure { return fail(error) }
+            guard let routes = potentialRoutes else { return }
+            return success(routes)
         }
+        
+        mapView.removeRoutes()
```
bsudekum: No need to remove routes. Only remove them if we error.

Examples/Swift/ViewController.swift
```diff
@@ -241,7 +258,14 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
         return false
     }
     
-    func navigationMapView(_ mapView: NavigationMapView, didTap route: Route) {
+    func navigationMapView(_ mapView: NavigationMapView, didSelect waypoint: Waypoint) {
+        guard let opts = currentRoute?.routeOptions else { return }
```
bsudekum: Can you call this `routeOptions`? Slightly unclear what it is at first glance.

Examples/Swift/ViewController.swift
```diff
@@ -14,23 +16,54 @@ enum ExampleMode {
...
+    @IBOutlet weak var clearMap: UIButton!
+    
+    //MARK: Properties
     var waypoints: [Waypoint] = []
     var currentRoute: Route? {
+        get {
+            return routes?.first
+        }
+        set {
+            self.startButton.isEnabled = (newValue != nil)
```
frederoni: nit: drop self ×4

MapboxCoreNavigation/RouteOptions.swift
```diff
@@ -1,12 +1,19 @@
 import MapboxDirections
 
-extension RouteOptions {
-    var activityType: CLActivityType {
+public extension RouteOptions {
```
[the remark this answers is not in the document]
JThramer: Agree. Willfix.

Examples/Swift/ViewController.swift
JThramer: Agree. Willfix.

JThramer: As it turns out, the last two are required, but the first two are redundant and have been removed.

bsudekum:
@JThramer now that https://github.com/mapbox/mapbox-navigation-ios/pull/828 is in, it might make sense to rebase this PR since this needs MapboxDirections.swift >= v0.12.

MapboxCoreNavigation/RouteOptions.swift
```diff
@@ -1,12 +1,19 @@
...
-    var activityType: CLActivityType {
+    internal var activityType: CLActivityType {
         switch self.profileIdentifier {
         case MBDirectionsProfileIdentifier.cycling, MBDirectionsProfileIdentifier.walking:
             return .fitness
         default:
             return .automotiveNavigation
         }
     }
+    public func without(waypoint: Waypoint) -> RouteOptions {
```
bsudekum: Can you document this function now that it is public?

bsudekum:
In the example app, I'm getting a prompt to remove a waypoint after selecting it, it removes the waypoint from the map, but the route is not recalculated.

Examples/Swift/ViewController.swift
```diff
@@ -14,23 +16,54 @@ enum ExampleMode {
...
+    @IBOutlet weak var clearMap: UIButton!
+    
+    //MARK: Properties
     var waypoints: [Waypoint] = []
     var currentRoute: Route? {
+        get {
+            return routes?.first
+        }
+        set {
+            startButton.isEnabled = (newValue != nil)
```
bsudekum: No need for parens here.

JThramer:
@bsudekum The unreproducible issue you were seeing is a Race Condition, with an observed probability of about 1/20 of repro-ing. I've got a fix ready, just want to test it a bit more to make sure.

Issue #798, opened 2017-11-06
showRoute only shows first 2 legs


apascual:
Hi, I am trying to display directions for multiple waypoints in **NavigationMapView** and looks like a part of the directions (the last leg) is not getting presented in the map.

<img width="372" alt="screen shot 2017-11-06 at 13 51 09" src="https://user-images.githubusercontent.com/3491658/32442074-5a592576-c2fa-11e7-9755-d12a8244269a.png">

The legs and steps looks ok:
```
Leg 0
Head east on Kattesund
Continue straight to stay on Kattesund
Turn right onto Stora Södergatan
Turn left onto Västra Mårtensgatan
Make a slight left
Make a slight right
Make a slight left
You have arrived at your 1st destination
Leg 1
Head south
Turn left
Turn left onto Västra Mårtensgatan
Turn right onto Bankgatan
You have arrived at your 2nd destination, on the left
Leg 2
Head north on Bankgatan
Turn right onto Stora Tvärgatan
Turn left onto Råbygatan
Turn right onto Skolgatan
You have arrived at your destination
```

I am using the latest iOS SDK versions:
```
Using Mapbox-iOS-SDK (3.6.4)
Using MapboxCoreNavigation (0.9.0)
Using MapboxDirections.swift (0.10.6)
Using MapboxMobileEvents (0.2.10)
Using MapboxNavigation (0.9.0)
```

Thanks.

bsudekum:
Curious, which directions profile are you using?

apascual:
@bsudekum I am using "Cycling", but I can also reproduce the issue with "Automobile".

Here is the list of coordinates that I am using:
**1)** 55.702118, 13.190597
**2)** 55.701611, 13.194099
**3)** 55.699075, 13.19603
**4)** 55.700378, 13.201168

**Update:**  Looks like using *RouteOptions* instead of *NavigationRouteOptions* fixes the problem. Is this normal?

bsudekum:
You'll definitely want to use `NavigationRouteOptions` when possible. There are a lot of options we turn on in there. Will check this out.

bsudekum:
Confirmed, this is a bug.

bsudekum:
@apascual a quick fix for this:

```swift
let options = NavigationRouteOptions(waypoints: waypoints)
options.attributeOptions = []
```

`attributeOptions` is used for getting the congestion along a route. But since the biking profile does not support congestion, it's okay to keep this option off.

[...]

PR #807, opened 2017-11-07
Fix multi waypoint leg bug

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/798

Fixes an issue where legs would not show up.

/cc @1ec5

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -587,7 +587,7 @@ open class NavigationMapView: MGLMapView {
                 legCoordinates.insert(firstCoord, at: 0)
             }
             
-            let segments = zip(legCoordinates, legCoordinates).map { [$0.0, $0.1] }
+            let segments = zip(legCoordinates, legCoordinates.suffix(from: 1)).map { [$0.0, $0.1] }
             let congestionSegments = Array(zip(segments, legCongestion))
             
             // Merge adjacent segments with the same congestion level
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -565,7 +565,6 @@ open class NavigationMapView: MGLMapView {
         guard let coordinates = route.coordinates else { return nil }
         
         var linesPerLeg: [[MGLPolylineFeature]] = []
-        var previousLegCongestionIndex = 0
         
         for (index, leg) in route.legs.enumerated() {
             // If there is no congestion, don't try and add it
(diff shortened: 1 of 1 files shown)


bsudekum:
simulator screen shot - iphone x - 2017-11-07 at 17 51 28

eh, a little off.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -577,12 +576,19 @@ open class NavigationMapView: MGLMapView {
...
+                $0.coordinates?.suffix(from: 1)
+            }.joined())
+            
+            // We need to add the first coord of the route in.
+            if let firstCoord = leg.steps.first?.coordinates?.first {
+                legCoordinates.insert(firstCoord, at: 0)
+            }
             
-            previousLegCongestionIndex = legCongestion.count
+            let segments = zip(legCoordinates, legCoordinates).map { [$0.0, $0.1] }
```
1ec5: Zipping `legCoordinates` with itself is pointless. I think this is the source of the off-by-one errors you’re seeing.

1ec5: Also, a tuple is easy enough to work with that I don’t think it’s necessary to map it to an array containing the same values.

bsudekum:
Fixed: 

simulator screen shot - iphone x - 2017-11-08 at 09 54 21

MapboxNavigation/NavigationMapView.swift
```diff
@@ -577,12 +576,19 @@ open class NavigationMapView: MGLMapView {
...
+                $0.coordinates?.suffix(from: 1)
+            }.joined())
+            
+            // We need to add the first coord of the route in.
+            if let firstCoord = leg.steps.first?.coordinates?.first {
+                legCoordinates.insert(firstCoord, at: 0)
+            }
             
-            previousLegCongestionIndex = legCongestion.count
+            let segments = zip(legCoordinates, legCoordinates.suffix(from: 1)).map { [$0.0, $0.1] }
```
1ec5: Remove the call to `map(_:)`. That will obviate at least one `if let` below.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -577,12 +576,19 @@ open class NavigationMapView: MGLMapView {
...
+            let congestionSegments = Array(zip(segments, legCongestion))
```
1ec5: Can I bother you to add https://github.com/mapbox/mapbox-navigation-ios/pull/807#pullrequestreview-74974713 as comments?

PR #815, opened 2017-11-08
Fixup 785

frederoni:
A few of things slipped through #785 

- Extract localizable
- Titles for waypoints
- Sort style attributes

@bsudekum 👀

### DIFF
--- a/MapboxNavigation/DayStyle.swift
+++ b/MapboxNavigation/DayStyle.swift
@@ -59,91 +59,69 @@ open class DayStyle: Style {
             tintColor = .defaultTint
         }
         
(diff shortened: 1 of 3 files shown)


bsudekum:
@frederoni it seems unnecessary to have a seperator if there is only 1 waypoint. How about only showing them past the first leg?

screen shot 2017-11-08 at 4 44 05 pm

frederoni:
> How about only showing them past the first leg?

Sounds good. Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/815/commits/e54ed032461fc34a23a8283ad0b9021e454eca96

PR #816, opened 2017-11-08
More iPhone X improvements

bsudekum:
@frederoni I've got most of it sorted out I think. The only problem I'm seeing is with the bottom bar. Can you look into having the width and height respect the superview but it's contents respect safe areas?

simulator screen shot - iphone x - 2017-11-08 at 18 22 47

### DIFF
--- a/MapboxNavigation/BottomBannerViewLayout.swift
+++ b/MapboxNavigation/BottomBannerViewLayout.swift
@@ -3,7 +3,6 @@ import UIKit
 extension BottomBannerView {
     
     func setupViews() {
-        applyDefaultShadow()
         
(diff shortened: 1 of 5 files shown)


bsudekum:
iPhone X
simulator screen shot - iphone x - 2017-11-08 at 18 24 14

iPhone 8
simulator screen shot - iphone 8 - 2017-11-08 at 18 24 15

frederoni:
@bsudekum Is this the desired behavior?

bsudekum:
@frederoni yep, I’d also check out landscape too. We might want to have a conditional in there as it might not need to respect the bottom safe area.

PR #817, opened 2017-11-08
Manage access token via unversioned file

1ec5:
The Swift and Objective-C example applications now rely on [a Run Script build phase](https://www.mapbox.com/help/ios-private-access-token/) to apply the Mapbox access token at build time. This makes it less likely that those working in this repository will inadvertently expose their own access tokens to the public.

/cc @bsudekum


frederoni:
Should we put a token on Bitrise or conditionalize the build phase?

1ec5:
I added a step to the workflow that lays down a bogus access token.

bsudekum:
@frederoni since this is looking for a token at `~/.mapbox token`, I don't think we need to add this to the gitignore.

PR #819, opened 2017-11-09
Show subsequent step under banner

### LINKED ISSUE #469: Show subsequent step under turn banner

frederoni:
Fixes #469 

Adds a banner for the subsequent step under the instructions banner.

@bsudekum @ericrwolfe @JThramer

### DIFF
--- a/MapboxNavigation/NextBannerView.swift
+++ b/MapboxNavigation/NextBannerView.swift
@@ -1,0 +1,67 @@
+import UIKit
+
+/// :nodoc:
+@objc(MBNextInstructionLabel)
+class NextInstructionLabel: InstructionLabel { }
+
+/// :nodoc:
+@IBDesignable
+@objc(MBNextBannerView)
+class NextBannerView: UIView {
+    
+    weak var maneuverView: ManeuverView!
+    weak var instructionLabel: NextInstructionLabel!
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        commonInit()
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        commonInit()
+    }
+    
+    func commonInit() {
+        setupViews()
+        setupLayout()
+    }
+    
+    func setupViews() {
+        let maneuverView = ManeuverView()
+        maneuverView.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(maneuverView)
+        self.maneuverView = maneuverView
+        
+        let instructionLabel = NextInstructionLabel()
+        instructionLabel.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(instructionLabel)
+        self.instructionLabel = instructionLabel
+        
+        instructionLabel.availableBounds = {
+            let height = ("|" as NSString).size(attributes: [NSFontAttributeName: self.instructionLabel.font]).height
+            let availableWidth = self.bounds.width-self.maneuverView.frame.maxX-(16*2)
+            return CGRect(x: 0, y: 0, width: availableWidth, height: height)
+        }
+    }
+    
+    override func prepareForInterfaceBuilder() {
+        super.prepareForInterfaceBuilder()
+        maneuverView.isEnd = true
+        instructionLabel.text = "San Jose"
+    }
+    
+    func setupLayout() {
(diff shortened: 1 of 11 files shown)


bsudekum:
I'm not sold on the design here yet. I think the main issue is that we're not telling the user what this new view is. I really like Google's design because it's simple, informative and it's not close to visually equal with the main banner:

image

At the point where we're showing this view to the user, it's not actually necessary to tell them the name of the road. One might argue actually giving the user two road names (perhaps even 3 with a destination) is too much and confusing. My vote is still for `Then: maneuver-arrow`.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -388,6 +388,35 @@ class RouteMapViewController: UIViewController {
         instructionsBannerView.maneuverView.step = routeProgress.currentLegProgress.upComingStep
     }
     
+    func updateNextBanner(routeProgress: RouteProgress) {
+        let step = routeProgress.currentLegProgress.upComingStep ?? routeProgress.currentLegProgress.currentStep
+        let nextStep = routeProgress.currentLegProgress.stepAfter(step)
+        let shouldShowNextBanner = nextStep != nil ? nextStep!.distance < 1000 : false
```
bsudekum: Can we guard `let nextStep`?

bsudekum: Let's use `RouteControllerHighAlertInterval` as the constant here.  https://github.com/mapbox/mapbox-navigation- ios/blob/master/MapboxCoreNavigation/Constants.swift#L95

(Maybe we should rename this 😬 )

ericrwolfe:
@frederoni I think the shield in the then banner could be smaller, maybe 60-75% of the primary banner label's shields?

MapboxNavigation/RouteMapViewController.swift
frederoni: Can we reuse `RouteControllerMinimumDistanceForContinueInstruction` ?

bsudekum: This is a little different. I think we will end up using this UI for continue instructions, but for a first pass to make this sync well with the `then`'s coming from the server we should only show this UI when linking instructions with `then`.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -388,6 +388,35 @@ class RouteMapViewController: UIViewController {
         instructionsBannerView.maneuverView.step = routeProgress.currentLegProgress.upComingStep
     }
     
+    func updateNextBanner(routeProgress: RouteProgress) {
+        guard let step = routeProgress.currentLegProgress.upComingStep,
+            routeProgress.currentLegProgress.currentStepProgress.durationRemaining <= RouteControllerHighAlertInterval,
+            let nextStep = routeProgress.currentLegProgress.stepAfter(step)
```
bsudekum: This should actually be `RouteControllerHighAlertInterval * RouteControllerLinkedInstructionBufferMultiplier `

bsudekum:
Added then in https://github.com/mapbox/mapbox-navigation-ios/pull/819/commits/f10c72989b0d6a07ef10ea57697e653ccf10e9b4

MapboxNavigation/NextBannerView.swift
```diff
@@ -0,0 +1,68 @@
...
+    }
+    
+    func setupViews() {
+        let maneuverView = ManeuverView()
+        maneuverView.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(maneuverView)
+        self.maneuverView = maneuverView
+        
+        let instructionLabel = NextInstructionLabel()
+        instructionLabel.shieldHeight = NextInstructionLabel.appearance().font.pointSize
```
bsudekum: Tests are failing because font is nil here.

bsudekum: @frederoni hopefully fixed with https://github.com/mapbox/mapbox-navigation-ios/pull/819/commits/d8b34885ffee4af2e86e869a28b267f8f76563af

PR #820, opened 2017-11-09
Dismiss feedback after timer done

bsudekum:
It doesn't go away right now.

/cc @frederoni @JThramer

### DIFF
--- a/MapboxNavigation/FeedbackViewController.swift
+++ b/MapboxNavigation/FeedbackViewController.swift
@@ -66,8 +66,10 @@ class FeedbackViewController: UIViewController, DismissDraggable, FeedbackCollec
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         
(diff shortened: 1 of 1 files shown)


MapboxNavigation/FeedbackViewController.swift
```diff
@@ -66,8 +66,10 @@ class FeedbackViewController: UIViewController, DismissDraggable, FeedbackCollec
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         
-        UIView.animate(withDuration: autoDismissInterval) {
+        UIView.animate(withDuration: autoDismissInterval, animations: {
             self.progressBar.progress = 0
+        }) { (true) in
+            self.dismissFeedback()
```
frederoni: `enableAutoDismiss()` a few lines down should take care of this. 🤔

bsudekum:
Never mind, this actually is not needed.

Issue #822, opened 2017-11-09
Bottom row feedback text is cutoff


bsudekum:
Only on the iphone x

![simulator screen shot - iphone x - 2017-11-09 at 15 35 13](https://user-images.githubusercontent.com/1058624/32628382-a76ece0e-c563-11e7-9992-ad4755ae2ff7.png)

/cc @frederoni

PR #825, opened 2017-11-09
Expand feedback UI slightly

bsudekum:
Previously, the bottom of the feedback UI cutoff some text on the iphone x.

/cc @frederoni


bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/822

PR #826, opened 2017-11-10
Eliminate imperative mood from property names

### LINKED ISSUE #756: RouteController has properties that look like methods
The RouteController properties `checkForFasterRouteInBackground`, `allowRecordedAudioFeedback`, and `showDebugSpokenInstructionsOnMap` all look like methods, especially in Objective-C. They should be renamed `checksForFasterRouteInBackground`, `allowsRecordedAudioFeedback`, and `showsSpokenInstructionsOnMap`, respectively.

/cc @bsudekum

### LINKED ISSUE #757: RouteController should not control appearance of map
`RouteController.showDebugSpokenInstructionsOnMap` should be moved to RouteMapViewController or NavigationViewController (and renamed: #756), since it controls the appearance of the map. RouteController should not assume the presence of a map view.

/cc @bsudekum

1ec5:
Property names should never begin with an imperative verb, in part because that makes it impossible to distinguish a property from an action method in Objective-C. The same requirement [holds true in Swift](https://swift.org/documentation/api-design-guidelines/#boolean-assertions):

> Uses of Boolean methods and properties should read as assertions about the receiver when the use is nonmutating.

This PR also makes the opportunistic background rerouting interval configurable and moves the spoken instruction debugging option from RouteController to NavigationViewController, where UI options belong.

Fixes #756 and fixes #757.

/cc @bsudekum @JThramer

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -175,7 +175,7 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
         exampleMode = .default
         
         let navigationViewController = NavigationViewController(for: route, locationManager: navigationLocationManager())
-        navigationViewController.routeController.recordsAudioFeedback = true
+        navigationViewController.recordsAudioFeedback = true
         navigationViewController.delegate = self
         
         present(navigationViewController, animated: true, completion: nil)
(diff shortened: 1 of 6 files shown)


JThramer:
Bravo!

PR #828, opened 2017-11-10
MapboxDirections.swift v0.12.0, OSRMTextInstructions v0.5.0

### LINKED ISSUE #821: Spoken instructions ignore application locale
If the system language is set to a language other than English and the application is localized into that language, instructions are nonetheless spoken in English. If Polly is enabled, those instructions are spoken in English with a foreign accent.

This regression occurred due to a combination of relying on the Directions API response for spoken instructions (#614) and failing to set the Directions API language to something other than English (mapbox/MapboxDirections.swift#201).

/cc @bsudekum @frederoni

1ec5:
Also use the “nationalized current locale” for consistency.

Fixes #821. Depends on Project-OSRM/osrm-text-instructions.swift#51.

/cc @bsudekum @frederoni @JThramer


1ec5:
Tests are crashing on CI but not locally:

```
Test Case '-[MapboxCoreNavigationTests.MapboxCoreNavigationTests testArrive]' started.
2017-11-09 23:41:34.022631-0800 xctest[5636:22813] This app has attempted to access privacy-sensitive data without a usage description. The app's Info.plist must contain an NSLocationWhenInUseUsageDescription key with a string value explaining to the user how the app uses this data
Fatal error: Unexpectedly found nil while unwrapping an Optional value
2017-11-09 23:41:34.035252-0800 xctest[5636:22813] Fatal error: Unexpectedly found nil while unwrapping an Optional value
2017-11-09 23:41:52.759 xcodebuild[4798:23324]  IDETestOperationsObserverDebug: Writing diagnostic log for test session to:
/var/folders/90/5stft2v13fb_m_gv3c8x9nwc0000gn/T/com.apple.dt.XCTest/IDETestRunSession-9B016A0B-4FE7-4A8E-9EFB-AED9851CC97F/MapboxCoreNavigationTests-35B98E80-5ADF-4288-A8F4-E8C1F22AA196/Session-MapboxCoreNavigationTests-2017-11-09_234152-IJVu9i.log
2017-11-09 23:41:52.759 xcodebuild[4798:18215] [MT] IDETestOperationsObserverDebug: (49563C5D-C219-46D2-825F-344BBF6B4149) Beginning test session MapboxCoreNavigationTests-49563C5D-C219-46D2-825F-344BBF6B4149 at 2017-11-09 23:41:52.760 with Xcode 9B55 on target <DVTiPhoneSimulator: 0x7fce26a5a370> {
		SimDevice: iPhone 6s Plus (014D6C15-FD93-4320-A812-EA3A6001356E, iOS 11.1, Booted)
} (11.1 (15B87))

Restarting after unexpected exit or crash in MapboxCoreNavigationTests/testArrive(); summary will include totals from previous launches.
```

bsudekum:
Interesting, I'm actually seeing the failure locally:

image

bsudekum:
I think this is failing because the encoded `tunnel.route` object is different now on MapboxDirections.swift v0.12.0.

bsudekum:
Ok now getting an error here: 

image


This line looks incorrect to me:

https://github.com/mapbox/MapboxDirections.swift/blob/master/MapboxDirections/MBRouteLeg.swift#L101

We should not be decoding an array of numbers.

bsudekum:
Something is not right, even after updating `tunnel.route`, I'm still getting an error.

1ec5:
> https://github.com/mapbox/MapboxDirections.swift/blob/master/MapboxDirections/MBRouteLeg.swift#L101
> 
> We should not be decoding an array of numbers.

That line is correct. CongestionLevel is an enumeration, so it’s represented in the archive as a number, because archives have a much more limited set of types (only slightly more flexible than JSON).

JThramer:
@bsudekum @1ec5 the only thing that comes to mind regarding this issue is possibly mapbox/MapboxDirections.swift#204? That bug could cause `RouteOptions` to return as nil… and if you’re trying to unarchive a `Route`, I could see where `RouteOptions.initWithCoder:` returning `nil` for a property (`Route.routeOptions`) that is not Optional could be causing the problem…

1ec5:
> Interesting, I'm actually seeing the failure locally

My bad, I was testing MapboxDirections.swift locally but not the navigation SDK. 😳

> the only thing that comes to mind regarding this issue is possibly mapbox/MapboxDirections.swift#204?

Thanks, I can confirm that building with MapboxDirections.swift on master (including mapbox/MapboxDirections.swift#204) does fix the issue. Let’s publish a patch release of MapboxDirections.swift with that fix.

1ec5:
The test failures should be resolved now, @bsudekum.

PR #834, opened 2017-11-10
Release v0.10.0

bsudekum:
I think we may have to hold on https://github.com/mapbox/mapbox-navigation-ios/pull/806 and https://github.com/mapbox/mapbox-navigation-ios/pull/828 until post release. Alternatively, we can release this Monday, but let's not force something through.

/cc @1ec5 @ericrwolfe @frederoni @JThramer


1ec5:
#828 definitely blocks this release, because #821 is a regression that introduces very politically incorrect behavior to the SDK.

1ec5:
#828 is ready to go.

Issue #840, opened 2017-11-11
Use actual aws-sdk instead of fred's


bsudekum:
We can get off of https://github.com/mapbox/mapbox-navigation-ios/blob/master/Cartfile#L6 now I think.

/cc @frederoni

PR #841, opened 2017-11-13
Use aws/aws-sdk-ios instead of frederoni/aws-sdk-ios

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/840

Tested, works fine.

/cc @frederoni


frederoni:
I get 
```
xcodebuild timed out while trying to read AWSAuthSDK.xcodeproj 😭
```
when running `carthage update`. I think this looked ok because the cache was still was still valid. Next release in https://github.com/aws/aws-sdk-ios will increase the build time by roughly 30 minutes. [Carthage/pull/1990](https://github.com/Carthage/Carthage/pull/1990) would fix this by excluding specific schemes.

frederoni:
> will increase the build time by roughly 30 minutes.

Correction; just the aws-sdk took 93m21.294s on a MacBook Pro 3.3 GHz i7 16 GB.

bsudekum:
Ugh, I forgot that's why we do that. Let's revert. @frederoni while we're on the subject, mind updating your copy?

Issue #812, opened 2017-11-08
Keep FPS throttled if upcoming step is continue/very straight looking


bsudekum:
If it's not complex, we can keep things throttled.

/cc @frederoni

PR #845, opened 2017-11-14
 Allow throttled FPS if upcoming maneuver is straight

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/812

If the upcoming maneuver is easy and straight, we can remain in throttled fps mode.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -161,7 +161,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         let durationUntilNextManeuver = stepProgress.durationRemaining
         let durationSincePreviousManeuver = expectedTravelTime - durationUntilNextManeuver
         
-        guard isPluggedIn else {
+        guard !isPluggedIn else {
             frameInterval = FrameIntervalOptions.pluggedInFrameInterval
             return
         }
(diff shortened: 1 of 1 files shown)


bsudekum:
@1ec5 would you rather look at the initial and final heading for the maneuver?

1ec5:
Oh, I think I misunderstood the description above to mean that we’d throttle or not throttle based on the curviness of the stretch of road between maneuvers. But this PR simply limits the high-FPS behavior to turns, as opposed to straight maneuvers, correct?

bsudekum:
Correct. If the maneuver is straight, remain in throttled FPS mode when approaching the maneuver.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -161,10 +161,16 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
         let durationSincePreviousManeuver = expectedTravelTime - durationUntilNextManeuver
         
-        if !isPluggedIn,
-            durationUntilNextManeuver > FrameIntervalOptions.durationUntilNextManeuver,
+        guard !isPluggedIn else {
+            frameInterval = FrameIntervalOptions.pluggedInFrameInterval
+            return
+        }
+    
+        if durationUntilNextManeuver > FrameIntervalOptions.durationUntilNextManeuver,
```
1ec5: Let’s use `&&` here instead of `,`. In the absence of a `let` expression, the comma tends to obscure the fact that both cases must evaluate to `true`.

PR #848, opened 2017-11-15
End Of Route Feedback Redux

JThramer:
This is a second attempt at getting in an End of Route feedback experience. I figured that it would be easier to set up the interface anew from `master` rather than to try and continue with #762. 

✅ 🔜 :

- [x] Install `UITextField` and animations.
- [x] `UITextField` Dismissal
- [x] Move the EOR scene when keyboard appears
- [x] Add logic to transition map to appropriate bounds/mode
- [x] Fix iPhone X Keyboard Issues
- [x] Add logic to emit feedback events.
- [x] Hook-up EORVC to show at end-of-route
- [x] Style Comment `UITextView`
- [x] PR comments
- [x] Set `hidden` in appropriate places, remove dev colormasks, other polish
- [x] Merge Conflicts
- [x] Set `clipsToBounds` for the `NavigationViewController`
- [x] ~~See if you can get the puck to stop yo-yoing~~ (Unable to fix)
- [x] Talk to @1ec5 about using the Arrival Banner
- [x] Style-able Controls
- [x] ~Change EORVC UI Logic to support nameless destinations~ (Decision was geocoder instead)
- [x] Implement Geocoder
/cc @frederoni @bsudekum @ericrwolfe @1ec5


MapboxNavigation/Resources/Assets.xcassets/feedback-confusing-directions.imageset/Contents.json
```diff
@@ -2,6 +2,10 @@
   "images" : [
     {
       "idiom" : "universal",
+      "filename" : "star.pdf"
```
JThramer: Should not be in this file.

JThramer: ✅

MapboxNavigation/RatingControl.swift
```diff
@@ -0,0 +1,140 @@
...
+    
+    var rating: Int = 0 {
+        didSet {
+            updateSelectionStates()
+            didChangeRating?(rating)
+        }
+    }
+    
+    @IBInspectable var selectedColor: UIColor = #colorLiteral(red: 0.1205472574, green: 0.2422055006, blue: 0.3489340544, alpha: 1)
+    @IBInspectable var normalColor: UIColor = #colorLiteral(red: 0.8508961797, green: 0.8510394692, blue: 0.850877285, alpha: 1)
```
bsudekum: It'd be nice to have this view be style-able one day. Would it make sense to inherit colors from other Style components? Or does it make more sense to pull this out into it's own component?

/cc @frederoni

bsudekum: We can focus on this last btw.

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        
+        animated ? UIView.animate(withDuration: 0.3, animations: layout) : layout()
+    }
+    
+    
+    private func updateInterface() {
+        primary.text = string(for: destination)
+        guard let coordinate = destination?.coordinate else { return }
+        let location = CLLocation(latitude: coordinate.latitude, longitude: coordinate.longitude)
+        geocoder.reverseGeocodeLocation(location) { (places, error) in
```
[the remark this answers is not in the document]
JThramer: Good idea. I'm in favor.

MapboxNavigation/RatingControl.swift
frederoni: As we're slowly moving away from storyboards I'd prefer designable views but not inspectable properties. It's very easy to leave a stray UDRA (User Defined Runtime Attribute) in the storyboard which would lead to a crash since there's no compile-time check. Public class and dynamic public var would make it stylable.

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -0,0 +1,163 @@
...
+    @IBOutlet weak var showCommentView: NSLayoutConstraint!
+    @IBOutlet weak var hideCommentView: NSLayoutConstraint!
+    @IBOutlet weak var ratingCommentsSpacing: NSLayoutConstraint!
+    
+    //MARK: - Properties
+    lazy var placeholder: String = NSLocalizedString("Add an optional comment here.", comment: "Comment Placeholder Text")
+    lazy var endNavigation: String = NSLocalizedString("End Navigation", comment: "End Navigation Button Text")
+    lazy var sendFeedback: String = NSLocalizedString("Send Feedback", comment: "Send Feedback Button Text")
+    
+    lazy var geocoder: CLGeocoder = CLGeocoder()
```
1ec5: Can you think of a way to allow the developer to substitute an alternative geocoder, such as MapboxGeocoder.swift?

MapboxNavigation/RatingControl.swift
1ec5: That would only be a risk if a developer customizes the color in a version of this library that uses a storyboard but later upgrades, correct? Perhaps we can treat adding color inspectables as tail work for when the migration away from storyboards is complete.

MapboxNavigation/RatingControl.swift
```diff
@@ -0,0 +1,140 @@
...
+    
+    private func addButtonSizeConstraints(to view: UIView) {
+        let width = NSLayoutConstraint(item: view, attribute: .width, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1.0, constant: starSize.width)
+        let height = NSLayoutConstraint(item: view, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1.0, constant: starSize.height)
+        
+        NSLayoutConstraint.activate([width, height])
+    }
+    
+    @objc private func ratingButtonTapped(button sender: UIButton) throws {
+        guard let index = stars.index(of: sender) else { throw RatingControlError.tappedButtonNotInArray }
```
1ec5: I’m pretty sure the raw string value of this Error bridges to Objective-C as the NSError’s localized failure reason, in which case it could theoretically be displayed to the user. In fact, I’m surprised that this method would even be called, because the Objective-C selector should be `ratingButtonTapped:error:`, which isn’t a valid action selector. Since the error appears to be responding to a programming error entirely within the developer’s control, let’s make this an assertion instead of throwing an error.

MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -193,10 +195,20 @@ extension MMEEventsManager {
...
     }
     
-    func navigationCancelEvent(routeController: RouteController) -> [String: Any] {
+    //TODO: Change event semantic to `.exit`
+    func navigationCancelEvent(routeController: RouteController, rating: Int = unrated, comment: String? = nil) -> [String: Any] {
         var eventDictionary = self.addDefaultEvents(routeController: routeController)
         eventDictionary["event"] = MMEEventTypeNavigationCancel
         eventDictionary["arrivalTimestamp"] = routeController.sessionState.arrivalTimestamp?.ISO8601 ?? NSNull()
+        
+        let validRating: Bool = (rating >= MMEEventsManager.unrated && rating <= 100)
```
bsudekum: Is it possible for us to ever set a non valid rating?

JThramer: Us? Not currently. Others? Yes. I mean `RouteMapViewController` could be subclassed… also `RouteController.sendCancelEvent(rating:comment:)` is `open` so anyone could call it (which may be undesirable)

JThramer:
Screencap of current progress, as requested.

[Dropbox Link](https://www.dropbox.com/s/ti41dzy6hh1yags/EndOfRoute.mov?dl=0)

/cc @ericrwolfe

ericrwolfe:
Nice work @JThramer. 

* How hard would it be to synchronize the map + UI animation so it doesn't feel staggered?
* The arrival slide in/out animation feels slow. Could we maybe try an animation duration of 0.2s or 0.3s to match typical UIViewController transition speeds
* Let's switch they keyboard button to done and keep enter from automatically ending navigation. Instead let's wait for the send feedback button tap. Let's also keep the confirm button title as "End Navigation" so that it's clear that tapping this will dismiss navigation.
* Can we remove double slide animation on dismiss so that only the parent view controller is dismissed? We shouldn't be able to see the original bottom bar as the animation slides out

JThramer:
@ericrwolfe,

> How hard would it be to synchronize the map + UI animation so it doesn't feel staggered?

Unfortunately, not trivial. I spent quite a bit of time trying to achieve this a few days ago, It just doesn't seem to be currently possible for us to animate the map and a view concurrently in a `UIAnimationGroup`.

> The arrival slide in/out animation feels slow. Could we maybe try an animation duration of 0.2s or 0.3s to match typical UIViewController transition speeds

Not a problem. Will Do.

> Let's switch they keyboard button to done and keep enter from automatically ending navigation. Instead let's wait for the send feedback button tap. Let's also keep the confirm button title as "End Navigation" so that it's clear that tapping this will dismiss navigation.

Not a problem. Will do.

> Can we remove double slide animation on dismiss so that only the parent view controller is dismissed? We shouldn't be able to see the original bottom bar as the animation slides out

Not a problem. Will do.

MapboxNavigation/RatingControl.swift
JThramer: - [ ] Do this

MapboxNavigation/EndOfRouteViewController.swift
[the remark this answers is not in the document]
JThramer: This issue is now moot, as the scope of the story has changed, and we no longer have a `secondary` label to populate `locality` / `administrativeArea`.

MapboxNavigation/EndOfRouteViewController.swift
JThramer: See https://github.com/mapbox/mapbox-navigation-ios/pull/848/files/5a6552fc3d2db1f4b22a63d4f090a0ae2b357608#r152429652

MapboxNavigation/RatingControl.swift
JThramer: - [ ] do this

JThramer:
[Current Demo](https://www.dropbox.com/s/pqkaf7j17evlicx/EndOfRoute-final.mov?dl=0)

ericrwolfe:
Awesome, a few more pieces of feedback:
* Is there an announcement bug in that video where the arrival voice announcement happens when distanceRemaining = 0 rather than as you approach your destination? Ideally we'd trigger the arrival UI when the regular didArrive notification/voice announcement fires, which would hopefully be about 20-50m before the actual destination pin.
* Can we have the stars and the "End Navigation" button colors match the tint color or be stylable like the rest of the UI elements? @frederoni might be able to help here if you have any questions.
* Can we change the placeholder text to "How can we improve?"
* Can we reduce the whitespace between the star rating area and the end navigation button? Either by reducing the size of the text area or by dynamically resizing the parent view when you enter comment mode. We may want to make the rating UI optional for developers to turn off with a bool (we can circle back on this), in which case the height of the parent view should be dynamic.

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -247,12 +264,13 @@
                             <constraint firstItem="drb-do-FoT" firstAttribute="centerX" secondItem="bUe-nv-g0F" secondAttribute="centerX" id="O0m-0k-ilC"/>
                             <constraint firstItem="p9E-v0-Rt6" firstAttribute="bottom" secondItem="UnK-tl-lPe" secondAttribute="bottom" id="OJu-tK-ZSa"/>
                             <constraint firstItem="Gx9-Bj-wuQ" firstAttribute="leading" secondItem="Gvo-VD-DXF" secondAttribute="leading" id="QHw-st-QI3"/>
-                            <constraint firstItem="Gx9-Bj-wuQ" firstAttribute="top" secondItem="drb-do-FoT" secondAttribute="bottom" constant="10" id="SWf-LB-JLK"/>
```
frederoni: Looks like a merge conflict. This makes the Y position of the WayNameLabel ambiguous. Pin it 10pts from the bottom banner.

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -0,0 +1,161 @@
...
+        setPlaceholderText()
+        styleCommentView()
+    }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        roundCornersOfRootView()
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
```
frederoni: Nit: Omit overriding if we only call super

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -0,0 +1,161 @@
...
+        super.viewWillDisappear(animated)
+    }
+
+    //MARK: - IBActions
+    @IBAction func endNavigationPressed(_ sender: Any) {
+        dismissView()
+    }
+    
+    //MARK: - Private Functions
+    private func roundCornersOfRootView() {
```
frederoni: Can we move this to the UIView extension? Perhaps merge it with `applyDefaultCornerRadiusShadow`?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -17,7 +17,7 @@ public protocol NavigationViewControllerDelegate {
     /**
      Called when the user arrives at the destination.
      */
-    @objc optional func navigationViewController(_ navigationViewController : NavigationViewController, didArriveAt waypoint: Waypoint)
+    @objc optional func navigationViewController(_ navigationViewController : NavigationViewController, didArriveAt waypoint: Waypoint, finalDestination: Bool)
```
frederoni: Would it be possible to leave this unchanged and check `waypoint == route.routeOptions.waypoints.last`?

MapboxNavigation/RatingControl.swift
```diff
@@ -0,0 +1,135 @@
...
+    
+    private func setAccessibilityHint(for button: UIButton, at index: Int) {
+        guard rating == (index + 1) else { return } //This applies only to the zero-resettable button.
+        
+        button.accessibilityHint = NSLocalizedString("Tap to reset the rating to zero.", comment: "Rating Reset To Zero Accessability Hint")
+    }
+    
+    
+    private func addButtonSizeConstraints(to view: UIView) {
+        let width = NSLayoutConstraint(item: view, attribute: .width, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1.0, constant: starSize.width)
```
frederoni: Nit: Since we're requiring iOS >= 9, we can use the fluent API. This can be written more succinctly as:
```
view.widthAnchor.constraint(equalToConstant: starSize.width).isActive = true
```

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -11,6 +11,11 @@
         <capability name="Unknown constraint types" minToolsVersion="5.1"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
+    <customFonts key="customFonts">
```
frederoni: Are we using this font? The style currently doesn't support changing font family if we use more than one font.

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -60,10 +65,10 @@
                                     <action selector="rerouteFeedback:" destination="3z2-H5-unA" eventType="touchUpInside" id="yRc-0Z-7Pz"/>
                                 </connections>
                             </button>
-                            <view hidden="YES" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="drb-do-FoT" userLabel="Way Name View" customClass="MBWayNameView">
+                            <view hidden="YES" contentMode="scaleToFill" ambiguous="YES" translatesAutoresizingMaskIntoConstraints="NO" id="drb-do-FoT" userLabel="Way Name View" customClass="MBWayNameView">
```
frederoni: Let's make this unambiguous.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -476,9 +486,68 @@ class RouteMapViewController: UIViewController {
...
+        
+        let layout = view.layoutIfNeeded
+        let noAnimation = { layout(); completion?(true) }
+
+        guard duration > 0.0 else { return noAnimation() }
+        UIView.animate(withDuration: duration, delay: 0.0, options: [.curveLinear], animations: layout, completion: completion)
+    }
+    
+    func hideEndOfRoute(duration: TimeInterval = 1.0, completion: ((Bool) -> Void)? = nil) {
+        self.view.layoutIfNeeded() //flush layout queue
```
frederoni: Nit: drop self ×4

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -476,9 +486,68 @@ class RouteMapViewController: UIViewController {
...
+        let noAnimation = { layout(); complete(true) }
+
+        guard duration > 0.0 else { return noAnimation() }
+       UIView.animate(withDuration: duration, delay: 0.0, options: [.curveLinear], animations: layout, completion: complete)
+    }
+    
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        guard let endOfRouteVC = segue.destination as? EndOfRouteViewController else { return }
+        
+        endOfRouteVC.dismiss = { (stars, comment) in
```
frederoni: Might be mistaken but this looks like a ref cycle. It's not a ref cycle on this line already since `endOfRouteVC` is a scope variable but further down, `endOfRouteViewController = endOfRouteVC` <- strong ref, and using self inside the closure would trigger a ref cycle. We should use a weak self here.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -758,6 +827,63 @@ extension RouteMapViewController: BottomBannerViewDelegate {
     }
 }
 
+//MARK: - Keyboard Handling
+
+extension RouteMapViewController {
+    fileprivate func subscribeToKeyboardNotifications() {
```
frederoni: IMHO, it's easier to keep track of these notifications if they're all in `resumeNotifications/suspendNotifications` but feel free to leave it as it is.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -758,6 +827,63 @@ extension RouteMapViewController: BottomBannerViewDelegate {
...
+    }
+    @objc fileprivate func keyboardWillShow(notification: NSNotification) {
+        guard let userInfo = notification.userInfo else { return }
+        let curve = UIViewAnimationCurve(rawValue: userInfo[UIKeyboardAnimationCurveUserInfoKey] as! Int)
+        let options = (duration: userInfo[UIKeyboardAnimationDurationUserInfoKey] as! Double,
+                       curve: curve!)
+        let keyboardHeight = (userInfo[UIKeyboardFrameEndUserInfoKey] as! CGRect).size.height
+
+        if #available(iOS 11.0, *) {
+            self.endOfRouteShow.constant = -1 * (keyboardHeight - view.safeAreaInsets.bottom) //subtract the safe area, which is part of the keyboard's frame
```
frederoni: Nit: drop self ×2

bsudekum:
Looking good @JThramer, here are few pieces of feedback:

* Implementing geocoding for non-named waypoints (https://github.com/mapbox/mapbox-navigation-ios/pull/848#discussion_r151492491) will allow us to remove the destination as a lng/lat in most cases.
* Floating buttons (overview, mute, feedback) should be made hidden/disabled when the arrival screen is shown.
* When the arrival screen is shown, what does it look like if the pitch is set to zero?
* Can we remove the whitespace below the stars?  image

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
JThramer: - [ ] Fix Storyboard Merge Clobber

MapboxNavigation/EndOfRouteViewController.swift
JThramer: - [ ] Remove Override

MapboxNavigation/EndOfRouteViewController.swift
JThramer: - [ ] Extract to extension

MapboxNavigation/NavigationViewController.swift
bsudekum: What if we keep this function the same but just check `routeProgress.remainingWaypoints` instead?

JThramer: - [ ] Check for last waypoint using `navigationViewController` argument instead of exposing it as an argument

MapboxNavigation/RatingControl.swift
JThramer: - [ ] Use Layout Anchors

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
JThramer: - [ ] No custom fonts

MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
JThramer: Fixed by 1eb38e4

MapboxNavigation/RouteMapViewController.swift
JThramer: - [ ] Remove redundant `self` references

MapboxNavigation/RouteMapViewController.swift
JThramer: - [ ] remove retain cycle

MapboxNavigation/RouteMapViewController.swift
JThramer: - [ ] remove redundant `self` references

JThramer:
Remaining to-do's:

- [x] Dynamic EORVC height
- [ ] Style-able controls
- [ ] Placeholder Text = "How can we improve?"
- [ ] Geocode waypoints
- [ ] Hide RMVC Floating Buttons
- [ ] Pitch to zero at arrival time.

JThramer:
@ericrwolfe:

>Is there an announcement bug in that video where the arrival voice announcement happens when distanceRemaining = 0 rather than as you approach your destination? Ideally we'd trigger the arrival UI when the regular didArrive notification/voice announcement fires, which would hopefully be about 20-50m before the actual destination pin.

Good question. There _may_ be an issue in the logic behind `routeController:didArriveAt:` that might cause the destination maneuver to not get spoken until the _very_ end on the odd route or two, but the vast majority of sessions that I have observed behave exactly as you describe.

bsudekum:
> Good question. There may be an issue in the logic behind routeController:didArriveAt: that might cause the destination maneuver to not get spoken until the very end on the odd route or two, but the vast majority of sessions that I have observed behave exactly as you describe.

@JThramer this was fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/883

bsudekum:
@JThramer looking great. One minor piece of feedback, when the view is shown and the map moves, the user puck drifts a bit. Can you play around with setting the user puck animation to false when this view is shown? Ideally, it would would follow the destination marker as the view is animated.

ericrwolfe:
Awesome work @JThramer, looks slick!

I noticed in the slide down animation the banner view peeks back through. We might want to enable clipsToBounds on the NavigationViewController's view to prevent the off-screen views from showing through mid-animation.

1ec5:
> Implementing geocoding for non-named waypoints (https://github.com/mapbox/mapbox-navigation-ios/pull/848#discussion_r151492491) will allow us to remove the destination as a lng/lat in most cases.

Why not use the same text as the arrival banner? The arrival banner displays the waypoint name if it’s provided by the developer (whether using CLGeocoder or MapboxGeocoder.swift), falling back to the waypoint name returned by OSRM (which has just the street but not the house number).

JThramer:
@1ec5 we geocode the destination name _only_ if `currentLeg.destination.name` is `nil`. Otherwise, we use the value provided to us, I'm not really sure what you mean by "use the same text as the arrival banner" because the arrival banner seems to use `currentLeg.destination.name`.

Could you clarify?

1ec5:
My bad, for some reason I had incorrectly assumed that the developer-provided waypoint name would make it into the Route object, but that can’t happen because the MapboxDirections.swift classes are largely immutable. Instead, take a look at how the SDK hangs onto the developer-provided waypoint name in #419.

1ec5:
> we geocode the destination name _only_ if `currentLeg.destination.name` is `nil`. Otherwise, we use the value provided to us, I'm not really sure what you mean by "use the same text as the arrival banner" because the arrival banner seems to use `currentLeg.destination.name`.

I see. `destination.name` is exactly the right value to use. I guess I’m not convinced the fallback to CLGeocoder is sensible, because we ultimately would want it to match whatever the arrival banner says, and the Directions API has no access to Core Location. So I think we should just hide the label when the destination is unnamed – and also ensure that waypoints are named in all the examples.

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -11,6 +11,15 @@ fileprivate enum ContainerHeight: CGFloat {
     case commentShowing = 260
 }
 
+@objc(MBEndOfRouteContentView)
```
frederoni: Nit: `/// :nodoc:` all these classes.

JThramer: - [ ] Nodoc

JThramer:
@1ec5 so the decision was made in the nav weekly to keep the Geocoder. We should track that idea though, because it would be desirable to have a more flexible EndOfRouteViewController.

JThramer:
[Demo](https://www.dropbox.com/s/qn77epux7ofiwf6/endofroute-readytoland.mov?dl=0) of current state

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import MapboxDirections
+import CoreLocation
```
bsudekum: This can be removed.

JThramer: - [ ] Remove

MapboxCoreNavigation/RouteController.swift
```diff
@@ -81,6 +81,14 @@ public protocol RouteControllerDelegate: class {
...
     @objc(routeController:didUpdateLocations:)
     optional func routeController(_ routeController: RouteController, didUpdate locations: [CLLocation])
+    
+    /**
+     Called when the route controller arrives at a waypoint.
+     - parameter waypoint: The waypoint that the controller has arrived at.
+     - parameter finalDestination: A boolean flag that signals that the waypoint is the final destination.
+    */
+    @objc(routeController:didArriveAtWaypoint:)
+    optional func routeController(_ routeController: RouteController, didArriveAt waypoint: Waypoint)
```
bsudekum: Why do we need this delegate method addition when we already have this? https://mapbox.github.io/mapbox-navigation-ios/navigation/0.7.0/Protocols/NavigationViewControllerDelegate.html#/s:FP16MapboxNavigation32NavigationViewControllerDelegate24navigationViewControllerFTCS_24NavigationViewController11didArriveAtC16MapboxDirections8Waypoint_T_

MapboxNavigation/NavigationViewController.swift
```diff
@@ -276,8 +276,8 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
     var styleTypeForTimeOfDay: StyleType {
         guard automaticallyAdjustsStyleForTimeOfDay else { return .dayStyle }
 
-        guard let location = routeController.location,
-            let solar = Solar(coordinate: location.coordinate),
+        guard let coordinate = routeController.location?.coordinate,
+            let solar = Solar(latitude: coordinate.latitude, longitude: coordinate.longitude),
```
bsudekum: This change should not happen, the api changed when we updated to swift 4/solar 2.1.

MapboxNavigation/RatingControl.swift
```diff
@@ -0,0 +1,141 @@
...
+    private func setAccessibility(for button: UIButton, at index: Int) {
+        setAccessibilityHint(for: button, at: index)
+        
+        let value: String
+        
+        switch rating {
+        case 0:
+            value = "No rating set."
+        case 1:
+            value = "1 star set."
```
bsudekum: Do these need to be localizable?

MapboxNavigation/Waypoint.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+import MapboxDirections
+
+extension Waypoint {
+    var location: CLLocation {
+        return CLLocation.init(latitude: coordinate.latitude, longitude: coordinate.longitude)
```
bsudekum: Is this used anywhere?

bsudekum:
Tapping the current rated star should not dismiss the view:

new

JThramer:
@bsudekum can't find a link to back this up, but I could _swear_ that tapping on the current rating "removing" the rating was part of the original req's? /cc @ericrwolfe

MapboxNavigation/Waypoint.swift
JThramer: Used [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/848/files#diff-292af689ce88d5b5aa06966456abc0e9R597)

MapboxNavigation/RatingControl.swift
JThramer: - [ ] Localize

MapboxNavigation/NavigationViewController.swift
JThramer: -  [ ] Revert Change

MapboxNavigation/RatingControl.swift
```diff
@@ -0,0 +1,141 @@
...
+        
+        let value: String
+        
+        switch rating {
+        case 0:
+            value = NSLocalizedString("No rating set.", comment: "No Rating Set")
+        case 1:
+            value = NSLocalizedString("1 star set.", comment: "One Star Set")
+        default:
+            value = "\(rating) " + NSLocalizedString("stars set.", comment:"(multiple) Stars Set, as in '30 stars set'")
```
frederoni: Use `String.localizedStringWithFormat` in combination with `NSLocalizedString` to support a more generic format. Also, we must specify `bundle:` when using `NSLocalizedString`, otherwise it will look in the main app bundle for a string.

JThramer: - [ ] Better String Localization

MapboxCoreNavigation/RouteController.swift
JThramer: Copying here for posterity: the idea here is that it establishes the `RouteController` as the authority on location updates, and enforces order-of-operations, which would be very helpful to an implementer. You cannot ensure order-of-operations of a notification, but you can easily override a delegate method in a subclass. The idea here is that if someone wanted to do something custom with `NavigationViewController` where they wanted to do some custom logic before firing off the `navigationViewController(_:didArriveAt:)`, they could.

JThramer:
@bsudekum copying for posterity: The original idea behind the "cancellable" behavior in the `RatingControl` is that it doesn't back the user into a state where they _must_ give a rating. Say, they fat-fingered the `RatingControl` while trying to hit the "End Navigation" button.

bsudekum:
@JThramer one last addition- if feedback is disabled, this UI should be hidden.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -196,7 +207,6 @@ open class RouteController: NSObject {
     deinit {
         suspendLocationUpdates()
         checkAndSendOutstandingFeedbackEvents(forceAll: true)
-        sendCancelEvent()
```
bsudekum: Should this still happen?

JThramer: Gah -- that was an oversight.

JThramer: Correction: it kinda was an oversight. a "Navigation Cancel Event" is sent by the EORVC when you hit "End Navigation", but we still need to send a "blank" cancel event if the user sets `showsEndOfRoute` to `false`. Updating PR now.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -266,6 +266,12 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
         }
     }
     
+    @objc public var showsEndOfRoute: Bool = true {
```
bsudekum: Can you add some documentation? It might be more descriptive to call this `showsEndOfRouteFeedbackUI` or something similar.

MapboxCoreNavigation/RouteController.swift
bsudekum: What occurs if the user is only using `MapboxCoreNavigation`? There will be no `EORVC `.

JThramer:
Styling improvements:
simulator screen shot - iphone x - 2017-12-05 at 22 13 06
simulator screen shot - iphone x - 2017-12-05 at 22 13 08

MapboxNavigation/Waypoint.swift
frederoni: Nit: .init is implied here so it can be dropped.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -490,9 +518,97 @@ class RouteMapViewController: UIViewController {
...
+        let animate = {
+            self.view.layoutIfNeeded()
+            self.buttonStack.alpha = 1.0
+        }
+        
+        let complete: (Bool) -> Void = { self.endOfRouteContainer.isHidden = true; completion?($0)}
+        let noAnimation = { animate() ; complete(true) }
+
+        guard duration > 0.0 else { return noAnimation() }
+       UIView.animate(withDuration: duration, delay: 0.0, options: [.curveLinear], animations: animate, completion: complete)
```
frederoni: Nit: Inconsistent indentation

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -490,9 +518,97 @@ class RouteMapViewController: UIViewController {
...
+        }
+        
+        let noAnimation = { animate(); completion?(true) }
+
+        guard duration > 0.0 else { return noAnimation() }
+        UIView.animate(withDuration: duration, delay: 0.0, options: [.curveLinear], animations: animate, completion: completion)
+        
+        // Prevent the user puck from floating around.
+        if let location = routeController.location {
+            mapView.updateCourseTracking(location: location, animated: false)
```
frederoni: Nit: `updateCourseTracking(location:animated)` takes an optional location so no need to put this in an if let.

MapboxNavigation/RouteMapViewController.swift
JThramer: - [ ] fix indent

MapboxNavigation/RouteMapViewController.swift
JThramer: - [ ] remove redundant optional

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -24,13 +26,29 @@ class RouteMapViewController: UIViewController {
     @IBOutlet weak var statusView: StatusView!
     @IBOutlet weak var laneViewsContainerView: LanesContainerView!
     @IBOutlet weak var rerouteFeedbackTopConstraint: NSLayoutConstraint!
+    @IBOutlet weak var endOfRouteContainer: UIView!
```
frederoni: These names don't really follow Cocoa naming conventions but they're pretty isolated and internal.
Preferably, they should end with `View`, `Constraint` etc.

MapboxNavigation/RatingControl.swift
```diff
@@ -0,0 +1,142 @@
...
+            value = String.localizedStringWithFormat("%li %@", rating, starsSet)
+        }
+        
+        button.accessibilityValue = value
+    }
+    
+    private func setAccessibilityHint(for button: UIButton, at index: Int) {
+        guard rating == (index + 1) else { return } //This applies only to the zero-resettable button.
+        
+        button.accessibilityHint = NSLocalizedString("Tap to reset the rating to zero.", comment: "Rating Reset To Zero Accessability Hint")
```
frederoni: Missing bundle parameter.

MapboxNavigation/RatingControl.swift
```diff
@@ -0,0 +1,142 @@
...
+    private func addStars() {
+        for index in 0..<starCount {
+            let button = UIButton(type: .custom)
+            button.setImage(starTemplate, for: .normal)
+            button.adjustsImageWhenHighlighted = false
+            addButtonSizeConstraints(to: button)
+            
+            let setRatingNumber = NSNumber(value: index + 1)
+            let setRatingString = NumberFormatter.localizedString(from: setRatingNumber, number: .none)
+            let localizedStarLabel = NSLocalizedString("Set %@ star rating", comment: "Accessibility Star Label")
```
frederoni: Missing bundle parameter

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+    
+    private func setPlaceholderText() {
+        commentView.text = placeholder
+    }
+}
+
+//MARK: - UITextViewDelegate
+extension EndOfRouteViewController: UITextViewDelegate {
+    func textView(_ textView: UITextView, shouldChangeTextIn range: NSRange, replacementText text: String) -> Bool {
+        guard text == "\n" else { return true }
```
frederoni: Nit: could be improved with `CharacterSet.newlines`

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+            guard isViewLoaded else { return }
+            updateInterface()
+        }
+    }
+
+    //MARK: - Lifecycle Methods
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        clearInterface()
+        stars.didChangeRating = { (new) in self.rating = new }
```
frederoni: `RatingControl` have a strong ref to the `didChangeRating` closure so this is a strong ref cycle. This closure needs a weak self.

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+    @IBOutlet weak var endNavigationButton: UIButton!
+    @IBOutlet weak var stars: RatingControl!
+    @IBOutlet weak var commentView: UITextView!
+    @IBOutlet weak var commentViewContainer: UIView!
+    @IBOutlet weak var showCommentView: NSLayoutConstraint!
+    @IBOutlet weak var hideCommentView: NSLayoutConstraint!
+    @IBOutlet weak var ratingCommentsSpacing: NSLayoutConstraint!
+    
+    //MARK: - Properties
+    lazy var placeholder: String = NSLocalizedString("How can we improve?", comment: "Comment Placeholder Text")
```
frederoni: Nit: missing bundle parameter ×2

MapboxNavigation/EndOfRouteViewController.swift
JThramer: - [ ] weakify self

MapboxNavigation/EndOfRouteViewController.swift
JThramer: - [ ] use sugar

MapboxNavigation/RatingControl.swift
JThramer: - [ ] do that thing that you forgot to do the last time

MapboxNavigation/RatingControl.swift
JThramer: - [ ] here too

MapboxNavigation/RouteMapViewController.swift
JThramer: Yeah, I'm really conflicted about this. There's certainly an argument to be made for homogeneity for the sake of consistency, and I'm not even [consistent](https://github.com/mapbox/mapbox-navigation-ios/pull/848/files/5500b125705d03c92f0e947c7ed53efff0bf935f#diff-292af689ce88d5b5aa06966456abc0e9R14) within my own additions in this PR. I do believe that swift's [guideline](https://swift.org/documentation/api-design-guidelines/#promote-clear-usage) of "things should be named based upon roles, not types" is desirable, however I don't believe I did a good enough job in this PR with "swift-ier nomenclature", so I'll change this to conform to the rest of the class.

This would be a good topic to bring up in the next retro. 

- [ ] refactor names to conform to existing convention

/cc @1ec5 @bsudekum @akitchen

MapboxNavigation/EndOfRouteViewController.swift
JThramer: - [ ] Here too

MapboxNavigation/RouteMapViewController.swift
bsudekum: <- Guilty,  could always be better at naming.

MapboxNavigation/EndOfRouteViewController.swift
JThramer: I'm not so sure this is a retain cycle, since `EndOfRouteViewController` only holds a weak reference to the `RatingControl`, but it was an oversight nonetheless.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -258,18 +258,22 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
...
+            showsEndOfRouteFeedback = showsReportFeedback
         }
     }
     
-    @objc public var showsEndOfRoute: Bool = true {
+    
+    /**
+    Shows End of route Feedback UI when the route controller arrives at the final destination. Defaults to `true.`
+    */
+    @objc public var showsEndOfRouteFeedback: Bool = true {
```
1ec5: This property name makes it sound like it shows feedback that has already been collected. `asksForFeedbackOnArrival` or `showsArrivalFeedbackPrompt` would be more accurate.

MapboxNavigation/RouteMapViewController.swift
1ec5: > I do believe that swift's [guideline](https://swift.org/documentation/api-design-guidelines/#promote-clear-usage) of "things should be named based upon roles, not types" is desirable

Note that this library is written for Cocoa Touch, not pure Swift, so the Swift naming conventions sometimes have to accommodate Cocoa naming conventions. The problem with just calling it a container is that “container” is already a heavily overloaded term. “Container view” trades brevity for clarity.

PR #851, opened 2017-11-15
Don't require access token at ~/.mapbox to run example

bsudekum:
We shouldn't require this file to be there to run the example app. Let's log the missing file but not exit.

/cc @1ec5


1ec5:
Without the build step failing, this Info.plist entry leads to a rather inscrutable error at runtime, upon dropping a pin:

https://github.com/mapbox/mapbox-navigation-ios/blob/ed51113a9c36fa62fd55bf3bcaef52fe23de992f/Examples/Swift/Info.plist#L25-L26

```
The operation couldn’t be completed. Not Authorized - Invalid Token
```

As a fallback strategy, the applications can prompt for an access token at startup, just as the map SDK’s iosapp does. But it ends up being a fair amount of code that an ordinary application based on the navigation SDK’s example applications shouldn’t have to worry about:

https://github.com/mapbox/mapbox-gl-native/blob/dcd7da14f6627a09a5840a3046b5708449962f69/platform/ios/app/MBXViewController.m#L172-L205

bsudekum:
If the user manually sets `MGLMapboxAccessToken ` in their plist and does not have a file `~/.mapbox`, shouldn't we let them by?

1ec5:
Yes, although `MGLMapboxAccessToken` is currently set to a bogus `PASTE MAPBOX ACCESS TOKEN HERE` right out of the box. So we’d need to fail that value still.

PR #853, opened 2017-11-15
Show report feedback button for 5 seconds

bsudekum:
It's hanging around for too long. It gets awkward when you don't want to tap it after a reroute.

/cc @ericrwolfe


ericrwolfe:
cc @danesfeder

PR #864, opened 2017-11-17
Improve voice delegates

bsudekum:
Followup to https://github.com/mapbox/mapbox-navigation-ios/pull/800

This adds:
* ~~A routeProgress object to both voice delegate methods.~~
* Overlap will be triggered in `RouteVoiceController` as well.
* Improves the named arguments in the overlap method.

/cc @1ec5

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -354,6 +354,7 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
         mapViewController.reportButton.isHidden = !showsReportFeedback
 
         self.currentStyleType = styleTypeForTimeOfDay
+        self.voiceController?.routeController = self.routeController
         
         if !(route.routeOptions is NavigationRouteOptions) {
             print("`Route` was created using `RouteOptions` and not `NavigationRouteOptions`. Although not required, this may lead to a suboptimal navigation experience. Without `NavigationRouteOptions`, it is not guaranteed you will get congestion along the route line, better ETAs and ETA label color dependent on congestion.")
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -354,7 +354,6 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
         mapViewController.reportButton.isHidden = !showsReportFeedback
 
         self.currentStyleType = styleTypeForTimeOfDay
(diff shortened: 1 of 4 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -354,6 +354,7 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
         mapViewController.reportButton.isHidden = !showsReportFeedback
 
         self.currentStyleType = styleTypeForTimeOfDay
+        self.voiceController?.routeController = self.routeController
```
bsudekum: I'm worried about the voice controller being inited before the routeController is set on it. Should this line 
https://github.com/mapbox/mapbox-navigation-ios/blob/6dce036818e23d00d47efd7a8897fc703da61b97/MapboxNavigation/NavigationViewController.swift#L217

Be moved inside of the init method here?

bsudekum:
@1ec5 fixed

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -185,9 +183,9 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
 @objc public protocol VoiceControllerDelegate {
-    @objc(voiceController:spokenInstrucionsDidFailWithError:routeProgress:)
-    optional func voiceController(_ voiceController: RouteVoiceController, spokenInstructionsDidFailWith error: Error, routeProgress: RouteProgress)
+    @objc(voiceController:spokenInstrucionsDidFailWithError:)
+    optional func voiceController(_ voiceController: RouteVoiceController, spokenInstructionsDidFailWith error: Error)
     
-    @objc(voiceController:didInterruptSpokenInstruction:withInstruction:routeProgress:)
-    optional func voiceController(_ voiceController: RouteVoiceController, didInterrupt: SpokenInstruction, with instruction: SpokenInstruction, routeProgress: RouteProgress)
+    @objc(voiceController:didInterruptSpokenInstruction:withInstruction:)
+    optional func voiceController(_ voiceController: RouteVoiceController, didInterrupt: SpokenInstruction, with instruction: SpokenInstruction)
```
1ec5: Incidentally, can we get some documentation on these public methods? Something like the comments you already have in ViewController.swift.

1ec5:
Added more documentation in bd686d49003ca65c9f5f1c72c7f01777f09ed790.

Issue #862, opened 2017-11-17
Incorrect then banner


bsudekum:
![simulator screen shot - iphone 8 - 2017-11-16 at 17 38 17](https://user-images.githubusercontent.com/1058624/32925143-3b6b6f5e-caf5-11e7-842f-b62ffc6f89d4.png)

Does it seem a little early to be telling the user about this next step? I'm still on the ramp and have not yet entered 280.

/cc @frederoni @ericrwolfe

1ec5:
For what it’s worth, the southbound CA 85 ramp begins very soon after the on-ramp from De Anza Boulevard merges into I-280. A then instruction isn’t totally unreasonable, but it’s a borderline case I think.

PR #865, opened 2017-11-17
Fix when next banner is visible

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/862

We want to show the nextBanner when:

* the user is close to the end of their current step
* The follow on step is short

This cleans up the logic to make it a little more clear.

/cc @1ec5 @ericrwolfe @JThramer @frederoni

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -423,14 +423,17 @@ class RouteMapViewController: UIViewController {
         }
 
         let instructions = visualInstructionFormatter.instructions(leg: routeProgress.currentLeg, step: nextStep)
-        var instruction = instructions.0
+        let instruction = instructions.0
         
-        if let components = instruction?.components, components.count > 0 {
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -406,15 +406,22 @@ class RouteMapViewController: UIViewController {
...
                 return
         }
     
+        // If the followon step is short and the user is near the end of the current step, show the nextBanner.
+        guard nextStep.expectedTravelTime <= RouteControllerHighAlertInterval * RouteControllerLinkedInstructionBufferMultiplier,
+            routeProgress.currentLegProgress.durationRemaining <= RouteControllerHighAlertInterval * RouteControllerLinkedInstructionBufferMultiplier else {
+            hideNextBanner()
+            return
+        }
+
```
frederoni: While you're at it, can we replace the last `if let` with a guard so we hide the banner if instruction components are missing?

bsudekum:
@frederoni Fixed.

Issue #877, opened 2017-11-23
Navigation DidArrive At WayPooint/Destination not called


ashishke2m:
Hi,
I am facing issues with `navigation view controller:didArriveAtWaypoint`, this method was called when the user arrives at any waypoint or destination as in the previous SDK but in currently released v0.10.1. this method is not called when user arrived at any waypoint or destination.I already assigned a delegate to `Navigationmapviewcontroller` .the name was NavigationDelegate in the previous Version.
This is the major issues which I found please let me know some updates on it ASAP.

frederoni:
`navigationDelegate` was renamed to just `delegate`. Try setting the `delegate` and see if it works.

ashishke2m:
I already renamed it to `delegate`.but it is not working as I just updated SDK with pre-released version v0.11.0-rc.1.But still, it is not calling a method for `navigationviewcontroller:didArriveAtWaypoint`.You can see my code SS.
![screen shot 2017-11-23 at 6 46 51 pm](https://user-images.githubusercontent.com/16642068/33174576-a32d95de-d07e-11e7-8d11-50616e5b3a97.png)

ashishke2m:
I guess it is never satisfying below lines of code as i debug.
```
if routeProgress.currentLegProgress.userHasArrivedAtWaypoint {
            delegate?.navigationViewController?(self, didArriveAt: routeProgress.currentLegProgress.leg.destination)
        }
```
That is the reason is never called as i Tested other methods of deleagte that is **Cancel** and **DidReroute** it is working fine only **didArriveit is not working.**

Please Check it once and update it ASAP because it is important for SDK Users.

ashishke2m:
We are eagerly waiting for your reply.

TCOA:
although things have changed a lot (which does make updating the app a mess...), this hopefully will get you going......

    let center = NotificationCenter.default
    center.addObserver(forName: NSNotification.Name(rawValue: "RouteControllerProgressDidChange"), object: nil, queue: nil) {
                notification in
            let routeProgress = notification.userInfo![RouteControllerProgressDidChangeNotificationProgressKey] as! RouteProgress
            print("userHasArrivedAtWaypoint is ", routeProgress.currentLegProgress.userHasArrivedAtWaypoint)
     }

[...]

PR #883, opened 2017-11-27
Decouple waypoint arrival from voice announcements

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/877

We were piggy backing off of voice announcements to trigger waypoint arrivals. This is not really a good idea in practice since we don't really know when a voice announcement will be given. Instead, when we know the user is within x meters of the waypoint, say that they are there.

/cc @TCOA @ashishke2m @mapbox/navigation-ios


ericrwolfe:
cc @danesfeder

PR #886, opened 2017-11-27
Better initial view when completing leg

bsudekum:
Before, the user puck would float back to the beginning of the route and then jump back to the current location. Now:

gif3

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -121,16 +121,17 @@ class RouteMapViewController: UIViewController {
         mapView.compassView.isHidden = true
         
         mapView.tracksUserCourse = true
-        mapView.updateCourseTracking(location: routeController.location, animated: false)
         mapView.enableFrameByFrameCourseViewTracking(for: 3)
-        
-        guard !isViewLoaded else { return }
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -119,27 +119,28 @@ class RouteMapViewController: UIViewController {
...
+        mapView.updateCourseTracking(location: routeController.location, animated: false)
+        mapView.enableFrameByFrameCourseViewTracking(for: 3)
+        
+        guard !isViewLoaded else { return }
 
         if let camera = pendingCamera {
             mapView.camera = camera
-        } else if let firstCoordinate = route.coordinates?.first {
+        } else if let firstCoordinate = routeController.routeProgress.currentLegProgress.currentStep.coordinates?.first {
             let location = CLLocation(latitude: firstCoordinate.latitude, longitude: firstCoordinate.longitude)
```
frederoni: Can we improve this default case by using the course from first to second coordinate or the course of the location on `RouteController.location` to avoid the initial camera spin?

bsudekum:
@frederoni fixed

PR #887, opened 2017-11-28
Fix custom style demo

bsudekum:
Fixes the custom style demo to show a custom style at any time, not just night.

/cc @frederoni

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -353,7 +353,7 @@ extension ViewController: NavigationViewControllerDelegate {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Examples/Swift/ViewController.swift
```diff
@@ -352,8 +352,27 @@ extension ViewController: NavigationViewControllerDelegate {
     }
 }
 
+/**
+ To find more pieces of the UI to customize, checkout DayStyle.swift and NightStyle.swift
```
frederoni: Nit: Both DayStyle and NightStyle resides in `DayStyle.swift`.  Extract the night style or edit the comment.

PR #893, opened 2017-11-28
Remove unknown constraint types

### LINKED ISSUE #882: Storyboard warnings
image

We should fix this.

/cc @frederoni

frederoni:
Fixes #882 

- Remove unknown constraint types.

@bsudekum @JThramer 👀


MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -20,10 +19,6 @@
                         <rect key="frame" x="0.0" y="0.0" width="375" height="500"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
```
frederoni: This autoresizing mask is making the constraints obsolete.

PR #900, opened 2017-11-29
Swift 4 2

visionstech:


bsudekum:
Hi @visionstech, 

We have already updated to Swift 4. The first release (v0.11.0) with Swift 4 will be shipped today.

Issue #892, opened 2017-11-28
Can't compile Example code - error: Missing Mapbox access token


TCOA:
Why does this have to be such a struggle???

1) followed instructions from https://github.com/mapbox/mapbox-navigation-ios#running-the-example-project
in building, had the same issue as #601 (changed to my personal team - obvious enough...)

However, after putting in my Token in Info.plist, I still get this error.....

> error: Missing Mapbox access token
error: Get an access token from <https://www.mapbox.com/studio/account/tokens/>, then create a new file at /Users/myuser/.mapbox that contains the access token. 

when I try to do that, this stupid Mac tells me "You can't use a name that begins with a dot ".", because these names are reserved for the system.  Please choose another name."

What now???

frederoni:
We implemented an unversioned access token in #817 for security reasons.
There's most likely a setting in Finder that you need to enable to add files that begins with a dot.
However, using a terminal, you can just: `echo "YOUR_ACCESS_TOKEN" > ~/.mapbox`

We should clarify in the readme that this is the preferred way and perhaps also add support for the old way using the Info.plist.

TCOA:
thanks @frederoni  - that worked to compile (sorry, I'm not a fan of Apple - - - actually, since they dropped support for the ][e/gs line - which was my first computer.... I've been forced to use this Mac, but, very sadly, it really isn't very intuitive, even after all these years....)

but, I get no map (just a dot on the screen - top and bottom seem ok). 
![screen shot 2017-11-28 at 8 15 31 am](https://user-images.githubusercontent.com/20958703/33321632-59634784-d414-11e7-8c37-5b7d8f71213c.png)

I can move the dot around, just no location stuff - even though I gave it permission for location and notifications

TCOA:
retried the token (seems the last part didn't copy from the site the first time) - now I get the map and can finally start - again......

TCOA:
trying pod install (again, after following the instructions........

> [!] Unable to satisfy the following requirements:

- `MapboxCoreNavigation (= 0.11.0-rc.1)` required by `MapboxNavigation (0.11.0-rc.1)`

None of your spec sources contain a spec satisfying the dependency: `MapboxCoreNavigation (= 0.11.0-rc.1)`.

You have either:
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.

[...]

PR #902, opened 2017-11-29
Make access token script more lenient

1ec5:
The build script now looks for a file named “mapbox” in addition to “.mapbox”, for those using Finder, which disallows hidden files by default. A missing access token file now results in a warning instead of an error.

Updated the readme to offer this file as an option.

/ref #892
/cc @bsudekum @frederoni


1ec5:
Yes, good catch – fixed in 3cf940f1eb1390f3e17151668d4db7229c5a7089.

PR #904, opened 2017-11-30
Refactor style switching

frederoni:
Extracted style handling into a separate class.
The main issue before was that `RouteController.location` could be nil the very first second which resulted in solar being unable to calculate sunrise and sunset. The first coordinate on the route is now used as a fallback until the location manager received the first update.

@bsudekum @JThramer 👀

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -2,7 +2,6 @@ import UIKit
 import MapboxCoreNavigation
 import MapboxDirections
 import Mapbox
-import Solar
 
 /**
  The `NavigationViewControllerDelegate` provides methods for configuring the map view shown by a `NavigationViewController` and responding to the cancellation of a navigation session.
@@ -224,17 +223,6 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
     @objc public var routeController: RouteController!
     
     /**
-     Styles that will be used for various system traits.
-     
-     See `Style` and `DefaultStyle` for more information.
-     */
-    @objc public var styles: [Style]? {
-        didSet {
-            applyStyle()
-        }
-    }
-    
-    /**
      The main map view displayed inside the view controller.
      
      - note: Do not change this map view’s delegate.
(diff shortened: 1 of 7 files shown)


MapboxNavigation/StyleManager.swift
```diff
@@ -0,0 +1,134 @@
...
+    
+    func resumeNotifications() {
+        suspendNotifications()
+        
+        guard automaticallyAdjustsStyleForTimeOfDay else { return }
+        
+        let location = delegate.locationFor(styleManager: self)
+        guard let solar = Solar(coordinate: location.coordinate),
+            let sunrise = solar.sunrise,
+            let sunset = solar.sunset else {
```
bsudekum: We should `isNightTime()`  over too https://github.com/mapbox/mapbox-navigation-ios/pull/904/files#diff-ab5a908dc637e51fa78cf2be28d033ecL288

bsudekum: Oh it is, nvm. But why do we create two solars?

bsudekum:
Getting a crash when exiting navigation:

image

MapboxNavigation/StyleManager.swift
frederoni: B/c the coordinate is immutable. Take a flight from LA to Montreal and we'd have a discrepancy of ~6 hours if we would reuse the solar object. In addition to using a new solar object, we should also listen for timezone changes via UIApplicationSignificantTimeChangeNotification.

bsudekum:
Looks like UI elements are updating to the dark style, but the dark map style is not: 
simulator screen shot - iphone 8 plus - 2017-12-04 at 12 26 00

MapboxNavigation/NavigationViewController.swift
```diff
@@ -327,30 +298,24 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
...
-        self.directions = directions
-        self.route = route
+        self.styleManager = StyleManager(self)
+        self.styleManager.styles = styles ?? [DayStyle(), NightStyle()]
         
-        self.routeController = RouteController(along: route, directions: directions, locationManager: locationManager ?? NavigationLocationManager())
-        self.routeController.usesDefaultUserInterface = true
-        self.routeController.delegate = self
-
+        mapViewController.view!.pinInSuperview()
```
bsudekum: This sounds scary, why is this necessary?

frederoni: This is just to set up top|left|bottom|right constraints. Same as before but I moved it to an extension on UIView.

https://github.com/mapbox/mapbox-navigation-ios/blob/1a7392048c6ca63582d88297450312e8304db581/MapboxNavigation/UIView.swift#L39-L45

Issue #758, opened 2017-10-25
On-map spoken instructions should be left-justified


1ec5:
Spoken instructions displayed on the map should be left-justified, not center-justified, since they’re anchored to the bottom-left.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/727#discussion_r146900120
/cc @bsudekum

PR #907, opened 2017-11-30
Fix annotatesSpokenInstructions, justify left

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/758

/cc @1ec5

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -141,6 +141,8 @@ class RouteMapViewController: UIViewController {
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         
+        annotatesSpokenInstructions = delegate?.mapViewControllerShouldAnnotateSpokenInstructions(self) ?? false
+        
         showRouteIfNeeded()
         currentLegIndexMapped = routeController.routeProgress.legIndex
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -141,6 +141,8 @@ class RouteMapViewController: UIViewController {
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         
+        annotatesSpokenInstructions = delegate?.mapViewControllerShouldAnnotateSpokenInstructions(self) ?? false
```
1ec5: Why do we have both this delegate method and an internal `annotatesSpokenInstructions` property? The public `NavigationViewController.annotatesSpokenInstructions` property should be hooked up to one of them…

https://github.com/mapbox/mapbox-navigation-ios/blob/a205ce750f3f21a03b6604de1f0e47a65dde08d5/MapboxNavigation/NavigationViewController.swift#L298-L301

bsudekum:
@1ec5 this was added here https://github.com/mapbox/mapbox-navigation-ios/pull/826/files#diff-ab5a908dc637e51fa78cf2be28d033ecR570

1ec5:
Ah, right you are!

PR #908, opened 2017-11-30
Remove exits from the map

bsudekum:
This PR removes all exits from the style if present.

The exit identifiers on the map don't really help. If anything, they create noise since all exits, even those not on the route line, are shown. I also do not think adding exits for the current route is the answer here either.

/cc @mapbox/navigation-ios @aparlato @natslaughter

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -143,6 +143,7 @@ class RouteMapViewController: UIViewController {
         
         showRouteIfNeeded()
         currentLegIndexMapped = routeController.routeProgress.legIndex
+        removeExits()
     }
     
     override func viewWillDisappear(_ animated: Bool) {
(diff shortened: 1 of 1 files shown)


ericrwolfe:
FWIW, we are starting to work on a new iteration of our cartography for navigation, kicking it off with some research sessions with users to test our assumptions around what information is necessary and what information is unnecessary for navigation. 

@bsudekum I fully agree on the spirit of this PR to remove as much unnecessary information from the map as possible, and highway exits are one of the first pieces of information whose utility I'd like to test/quantify. 

As we've discussed before, there are opportunities in the SDK to selectively style the maps dynamically, by replacing placeholder data in navigation-specific map styles, and applying custom labels for intersections and maneuvers. 

Let's use the work from this PR as a foundation to explore those ideas, thinking about the layers we'd like to adjust dynamically based on the driver's current context. We can save the actual reduction of information for the navigation v3 styles themselves, based on @v8c3's research, or look for ways for the SDK to hook into those layers, as @1ec5 has suggested.

cc @eschow

1ec5:
With my amateur cartography hat on: even if we remove the exit numbers themselves, it may still be useful to mark the exits with _something_. Print and Web maps that omit exit numbers still tend to mark exits with diamonds or dots. In a turn-by-turn scenario, it may not be necessary to mark all the exits, but the one relevant to the route should be marked prominently. After all, most exit maneuvers are very slight turns, so even the maneuver arrow leaves some ambiguity as to the maneuver’s location.

I do agree with @bsudekum’s approach of limiting any changes along these lines to the turn-by-turn map. For other uses of NavigationMapView, such as on a preview screen, there’s less need to be minimalistic.

natslaughter:
Although I agree with @bsudekum that exit signs "don't really help," that is only my personal opinion, and I am not the only driver out there. I'm sure that many drivers and even types of drivers would appreciate their presence, signifiers that confirm they are moving along the intended path. This brings up a larger issue of what type of driver is the default navigation-map style intended for? I think establishing this along with user testing will provide an answer on whether or not to include these exit signs. 

A different styling of the signs could also change their visual prominence.

cc @ericrwolfe @eschow @aparlato

PR #910, opened 2017-12-05
Check all future steps

bsudekum:
We currently have a last ditch effort before rerouting to see if the user is somehow on the upcoming or follow on step.

This changes to check if the user is on any future step. If the user is with a certain radius to this new step, we increment the stepIndex to this new step.

/cc @mapbox/navigation

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -584,9 +584,7 @@ extension RouteController: CLLocationManagerDelegate {
         let radius = max(reroutingTolerance, location.horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
         let isCloseToCurrentStep = newLocation.isWithin(radius, of: routeProgress.currentLegProgress.currentStep)
         
-        if isCloseToCurrentStep {
-            return true
-        }
+        guard !isCloseToCurrentStep else { return true }
 
         // Check to see if the user is moving away from the maneuver.
         // Here, we store an array of distances. If the current distance is greater than the last distance,
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -397,7 +397,6 @@ open class RouteController: NSObject {
             guard let coords = step.coordinates else { continue }
             guard let closestCoordOnStep = Polyline(coords).closestCoordinate(to: coordinate) else { continue }
             
-            
             // First time around, currentClosest will be `nil`.
             guard let currentClosestDistance = currentClosest?.distance else {
                 currentClosest = (index: stepIndex, distance: closestCoordOnStep.distance)
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -390,13 +390,12 @@ open class RouteController: NSObject {
     func closestStep(to coordinate: CLLocationCoordinate2D) -> (CLLocationDistance, Int)? {
         var closestStep = Int.max
         var lowestDistance = Double.infinity
-        
-        let remainingSteps = routeProgress.currentLeg.steps.suffix(routeProgress.currentLegProgress.stepIndex)
+        let remainingSteps = routeProgress.currentLeg.steps.suffix(from: routeProgress.currentLegProgress.stepIndex)
         
         for (stepIndex, step) in remainingSteps.enumerated() {
             guard let coords = step.coordinates else { continue }
             guard let closestCoordOnStep = Polyline(coords).closestCoordinate(to: coordinate) else { continue }
-            
+
             if closestCoordOnStep.distance < lowestDistance {
                 lowestDistance = closestCoordOnStep.distance
                 closestStep = stepIndex
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -176,8 +176,6 @@ open class RouteController: NSObject {
     var recentDistancesFromManeuver: [CLLocationDistance] = []
 
     var previousArrivalWaypoint: Waypoint?
-    
-    typealias StepIndexDistance = (index: Int, distance: CLLocationDistance)
 
 
     /**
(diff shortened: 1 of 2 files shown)


bsudekum:
@1ec5 cannot notice any degradation in performance other than in network latency.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -376,6 +376,24 @@ open class RouteController: NSObject {
 
         return CLLocation(coordinate: userCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: userCourse, speed: location.speed, timestamp: location.timestamp)
     }
+    
+    func closestStep(to coordinate: CLLocationCoordinate2D) -> (CLLocationDistance, Int)? {
+        var closestStep = Int.max
+        var lowestDistance = Double.infinity
```
1ec5: Use `nil` to indicate the absence of a closest step. That way the compiler forces you to account for a missing value instead of finding out when a ginormous value crashes the code. Also, combine `closestStep` and `lowestDistance` into a single tuple-typed variable, to ensure that both values are set simultaneously and don’t go out of sync.

1ec5:
Did you try taking a shortcut on that route? The route simulation feature generally hews to the route line, so you wouldn’t be able to test that way.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -786,8 +797,12 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
 
-    func incrementRouteProgress() {
-        routeProgress.currentLegProgress.stepIndex += 1
+    func incrementRouteProgress(forcedStepIndex: Int? = nil) {
```
1ec5: This method is a bit of a misnomer now, since it doesn’t necessarily increment by one. A more flexible name would be `advance(to:)`, which can be called as `advance()` in the normal case.

bsudekum:
new

> Jumping ahead steps

MapboxCoreNavigation/RouteController.swift
```diff
@@ -386,6 +388,29 @@ open class RouteController: NSObject {
 
         return CLLocation(coordinate: userCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: userCourse, speed: location.speed, timestamp: location.timestamp)
     }
+    
+    func closestStep(to coordinate: CLLocationCoordinate2D) -> StepIndexDistance? {
```
1ec5: This method can be moved to an extension on RouteLegProgress, because it doesn’t need any of RouteController’s state other than the current leg progress.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -807,8 +825,12 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
 
-    func incrementRouteProgress() {
-        routeProgress.currentLegProgress.stepIndex += 1
+    func advanceStepIndex(to: Int? = nil) {
```
1ec5: Ideally, this method would take an `Array<RouteStep>.Index`, which would force callers to be safer and avoid potential out of bounds errors. However, that can be a bit of a pain to implement, so your call.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -583,7 +583,10 @@ extension RouteController: CLLocationManagerDelegate {
         let newLocation = CLLocation(latitude: locationInfrontOfUser.latitude, longitude: locationInfrontOfUser.longitude)
         let radius = max(reroutingTolerance, location.horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
         let isCloseToCurrentStep = newLocation.isWithin(radius, of: routeProgress.currentLegProgress.currentStep)
-
+        
+        if isCloseToCurrentStep {
```
JThramer: Nit: I generally like to see conditional branches that early-exit as `guard` statements, but that's largely a style thing.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -306,6 +307,30 @@ open class RouteLegProgress: NSObject {
         assert(!nearby.isEmpty, "Step must have coordinates")
         return nearby
     }
+    
+    typealias StepIndexDistance = (index: Int, distance: CLLocationDistance)
+    
+    func closestStep(to coordinate: CLLocationCoordinate2D) -> StepIndexDistance? {
```
JThramer: I think the performance of this method could be improved substantially if we add a pruning radius argument to this method. That way, we can filter off any step that's outside of our "radius" before enumerating through the array.

L313:
```swift
func closestStep(to coordinate: CLLocationCoordinate2D, within radius: CLLocationDistance = .greatestFiniteMagnitude) -> StepIndexDistance?
```

L315: 
```swift
let remainingSteps = leg.steps.suffix(from: stepIndex)
let filteredSteps = remainingSteps.filter({ $0.distance <= radius})
```

then in L317, enumerate on `filteredSteps`. 

Considering that `Array.suffix` returns an `ArraySlice`, you _shouldn't_ need to manually keep track of array indices by doing this, but don't quote me on that.

/cc @1ec5

JThramer: By "Radius", I'm referring to the check being done [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/910/files#diff-1a2149692b3d7c58b6a0392120d6011aR617)

ericrwolfe:
@cammace @danesfeder this helps with overzealous reroutes coming out of tunnels, parking lots, starting navigation in a different place than originally requested, etc and is something we should bring to Android.

zugaldia:
@ericrwolfe could you cut an Android-specific ticket? Leaving this request here will make it hard to track.

ericrwolfe:
Android ticket ➡️ https://github.com/mapbox/mapbox-navigation-android/issues/586

PR #914, opened 2017-12-06
Added regionType to PollyVoiceController init

### LINKED ISSUE #913: How to change regionType of PollyVoiceController at (Or after) initialisation?
Hello!, I've been trying to implement the PollyVoiceController using a different region than .USEast1 (Which is default)

I've attempted to override the default service configuration of AWS but to no avail. The only thing that works is initializing the AWSService before the VoiceController. Otherwise, setting the regionType does nothing.

This is the only way that I've found it to work, but seems a bit hacky.
```swift
let credentialProvider = AWSCognitoCredentialsProvider(regionType: .EUWest1, identityPoolId: amazonPollyPoolId)
let configuration = AWSServiceConfiguration(region: .EUWest1, credentialsProvider: credentialProvider)
AWSServiceManager.default().defaultServiceConfiguration = configuration

let pollyVoiceController = PollyVoiceController(identityPoolId: amazonPollyPoolId)
pollyVoiceController.regionType = .EUWest1
pollyVoiceController.voiceControllerDelegate = self
```

I submitted #914 which fixes the issue. Another way would be for the variable to be static, or initializing PollyVoiceController with a AWS configuration

Thanks!

Sealos:
Fixes #913

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -49,8 +49,9 @@ public class PollyVoiceController: RouteVoiceController {
     
     let localizedErrorMessage = NSLocalizedString("FAILED_INSTRUCTION", bundle: .mapboxNavigation, value: "Unable to read instruction aloud.", comment: "Error message when the SDK is unable to read a spoken instruction.")
     
-    public init(identityPoolId: String) {
+    public init(regionType: AWSRegionType, identityPoolId: String) {
         self.identityPoolId = identityPoolId
+        self.regionType = regionType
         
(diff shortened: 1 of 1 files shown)


bsudekum:
Thanks for grabbing this @Sealos! What do you think about making this parameter optional and giving it a default value `.USEast1`? It will probably make most sense to rework this line then too: 

https://github.com/Sealos/mapbox-navigation-ios/blob/78c8feedc10de54d121eb864ecc81f1f4833e106/MapboxNavigation/PollyVoiceController.swift#L22

Sealos:
Sounds good!, will do.

PR #915, opened 2017-12-06
Adjustable simulation speed

### LINKED ISSUE #694: Control simulated speed
Right now using real world values, it's a little too slow. Can we have a multiplier? Maybe keyboard shortcuts for toggling the speed too?

/cc @frederoni

frederoni:
Fixes #694 

Slide left or right on the "Simulation Navigation" banner to increase or decrease the speed respectively.

The user puck is not really in sync at these high speeds.

simspeed

@bsudekum @JThramer  👀

### DIFF
--- a/MapboxCoreNavigation/SimulatedLocationManager.swift
+++ b/MapboxCoreNavigation/SimulatedLocationManager.swift
@@ -36,6 +36,8 @@ public class SimulatedLocationManager: NavigationLocationManager {
     fileprivate var locations: [SimulatedLocation]!
     fileprivate var routeLine = [CLLocationCoordinate2D]()
     
+    public var speedMultiplier: Double = 1
+    
     @objc override public var location: CLLocation? {
         get {
             return currentLocation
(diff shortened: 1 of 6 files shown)


bsudekum:
I'm unable to do the swiping in the simulator.

frederoni:
> I'm unable to do the swiping in the simulator.

Oh, it's more like a pan gesture. No fling or velocity involved. Try a linear drag from right to left.
I also speed up the pan gesture in https://github.com/mapbox/mapbox-navigation-ios/pull/915/commits/d32cac696354784dfdd34f062e6fe651e1642aae to make it easier to trigger in a laggy simulator.

bsudekum:
Hrm, I can only seem to get it every once and a while. What about clicking on left or right side?
new

MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -36,6 +36,8 @@ public class SimulatedLocationManager: NavigationLocationManager {
     fileprivate var locations: [SimulatedLocation]!
     fileprivate var routeLine = [CLLocationCoordinate2D]()
     
+    public var speedMultiplier: Double = 1
```
bsudekum: `@objc` and doc.

PR #920, opened 2017-12-07
Add other_map_issue to feedback types

### LINKED ISSUE #916: Add other_map_issue to feedback types
We should add `other_map_issue` to the list of feedback types.

By default, feedback events without a specified category should use `general`. This includes the feedback event that gets sent after a reroute occurs, a user taps the "Report Problem" button but does not tap a specific feedback category. 

The `other_map_issue` feedback type would only apply to the similarly named icon at the bottom center of the feedback UI.

cc @mapbox/navigation-ios @paigemoody

frederoni:
Fixes #916 

@ericrwolfe @bsudekum 👀

### DIFF
--- a/MapboxCoreNavigation/Feedback.swift
+++ b/MapboxCoreNavigation/Feedback.swift
@@ -55,6 +55,11 @@ public enum FeedbackType: Int, CustomStringConvertible {
      */
     case reportTraffic
     
+    /**
+     Identifies the feedback as a general map issue.
+     */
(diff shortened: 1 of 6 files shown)


ericrwolfe:
@frederoni Can we also increment the event version (https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/MMEEventsManager.swift#L7) to `7` from this change?

Issue #921, opened 2017-12-07
End Of Route Label Geocoding issues


JThramer:
@1ec5 has reported that the destination name label on the EORVC is misbehaving. He saw that the label would always either show OSRM names or coordinate pairs. This suggests a problem with the `CLGeocoder` implementation.

The expected order of precedence is: 
`Manually Set Name` ➡️ `OSRM Name` ➡️ `CLGeocoder Name` ➡️ `"You Have Arrived"`

To Do: 
- [x] Investigate `CLGeocoder` failure
- [x] Change coordinate Pair to "You Have Arrived" and hide static label in this case

/cc @bsudekum @frederoni @ericrwolfe

ericrwolfe:
@JThramer Let's remove the fallback to the coordinate pair, and instead just show "You have arrived" in larger text instead of showing the smaller label.

bsudekum:
This is dependent on https://github.com/mapbox/MapboxDirections.swift/pull/222.

1ec5:
Fixed in #922.

PR #922, opened 2017-12-08
Replacing coordinate string logic

JThramer:
This implements the other half of #921. Instead of displaying a coordinate-pair, the VC now displays `You have arrived`.


bsudekum:
This is dependent on https://github.com/mapbox/MapboxDirections.swift/pull/222.

Cartfile
```diff
@@ -1,5 +1,5 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 3.7
-github "mapbox/MapboxDirections.swift" ~> 0.15
+github "mapbox/MapboxDirections.swift" ~> 0.15.1
```
bsudekum: I think this can remain actually. The only change that should occur is in Cartfile.resolved.

MapboxCoreNavigation.podspec
```diff
@@ -40,7 +40,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxCoreNavigation"
 
-  s.dependency "MapboxDirections.swift", "~> 0.15"
+  s.dependency "MapboxDirections.swift", "~> 0.15.1"
```
bsudekum: I think we can keep `0.15` and let semver do the rest.

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -143,19 +144,18 @@ class EndOfRouteViewController: UIViewController {
...
     }
     
-    private func string(for destination: Waypoint?) -> String {
-        guard let destination = destination else { return "Unknown" }
-        guard destination.name?.isEmpty ?? false else { return destination.name! }
-        let coord = destination.coordinate
-        return String(format: "%.2f", coord.latitude) + "," + String(format: "%.2f", coord.longitude)
+    private func styleForUnnamedDestination() {
+        staticYouHaveArrived.alpha = 0.0
+        primary.text = NSLocalizedString("You have arrived", bundle: .mapboxNavigation, comment:"")
```
1ec5: @JThramer, this string isn’t actually localizable. See other calls to `NSLocalizedString()` in this codebase for examples. To wit: the first argument should be a unique string ID, typically in UPPER_SNAKE_CASE, and `value` should be set to the English default. Then you’ll need to add an entry to the relevant .strings file. Until that happens, Transifex will be unable to pick up this string, because it only looks at .strings files, not code.

PR #926, opened 2017-12-13
Expose locationAccuracy on CLLocationManager

bsudekum:
Exposes the value used for `desiredAccuracy` on `CLLocationManager` so it can be overridable.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -17,6 +17,11 @@ open class NavigationLocationManager: CLLocationManager {
     
     var lastKnownLocation: CLLocation?
     
+    /**
+     The accuracy of the `CLLocationManager`.
+     */
+    @objc public var locationAccuracy: CLLocationAccuracy = kCLLocationAccuracyBestForNavigation
+    
     override public init() {
         super.init()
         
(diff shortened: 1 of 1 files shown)


frederoni:
The location manager itself is already exposed through the route controller so we don't need this change to adjust the accuracy, distance filter or any other properties.

```routeController.locationManager.desiredAccuracy```

**Comment by ericrwolfe** [comment]
@frederoni I noticed that you can't update `activityType` without it being overridden by RouteController.init() in https://github.com/mapbox/mapbox-navigation-ios/blob/41d846ae4433eeb33139e1622b484520483892ac/MapboxCoreNavigation/RouteController.swift#L194

Should we set this property in NavigationLocationManager instead, so that developers could also override it if needed?

**Comment by frederoni** [comment]
@ericrwolfe yea, it will only persist if you set it after initializing the route controller or the NavigationViewController if you're using the UI.
```swift
let locationManager = CLLocationManager()
locationManager.activityType = .fitness // Will get overriden
let navigationViewController = NavigationViewController(for: route, locationManager: locationManager)
locationManager.activityType = .fitness // Works
navigationViewController.routeController.locationManager.activityType = .fitness // Works
```

and I kinda like that we try to handle this automatically by converting the profile identifier into an activity type.

PR #930, opened 2017-12-14
Remove copy frameworks script

### LINKED ISSUE #927: Avoid embedding dependencies in framework targets
MapboxNavigation.framework embeds all its dependencies, so an application target’s build phase only needs to directly embed MapboxNavigation.framework and MapboxCoreNavigation.framework, but nothing else. (If MapboxNavigation.framework were to additionally embed MapboxCoreNavigation.framework, then the application target would only need to embed MapboxNavigation.framework alone.)

While this approach simplifies setup, it runs counter to the best practice of [linking frameworks together at the application target level](https://developer.apple.com/library/content/technotes/tn2435/_index.html#//apple_ref/doc/uid/DTS40017543-CH1-PROJ_CONFIG-APPS_WITH_DEPENDENCIES_BETWEEN_FRAMEWORKS). For example, if the application also depends on the map SDK, then things get a bit awkward because a developer would expect to have to embed Mapbox.framework directly, which would lead to two copies of Mapbox.framework in the resulting application bundle. The iTunes Store rejects application submissions with duplicate frameworks.

As far as I can tell, the Swift and Objective-C example applications should embed all the dependencies directly, leaving MapboxNavigation.framework and MapboxCoreNavigation.framework to link their respective dependencies but not embed them.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/804#issuecomment-350819048
/cc @frederoni @bsudekum @LuAndreCast

frederoni:
Fixes #927 avoid nested dependencies

@1ec5 @bsudekum

### DIFF
--- a/MapboxNavigation.xcodeproj/project.pbxproj
+++ b/MapboxNavigation.xcodeproj/project.pbxproj
@@ -984,10 +984,9 @@
 			buildConfigurationList = 351BEBDE1E5BCC28006FE110 /* Build configuration list for PBXNativeTarget "MapboxNavigation" */;
 			buildPhases = (
 				351BEBD21E5BCC28006FE110 /* Sources */,
-				351BEBD31E5BCC28006FE110 /* Frameworks */,
 				351BEBD41E5BCC28006FE110 /* Headers */,
 				351BEBD51E5BCC28006FE110 /* Resources */,
-				35C7E0931E67AD80003ACB96 /* ShellScript */,
+				351BEBD31E5BCC28006FE110 /* Frameworks */,
 			);
 			buildRules = (
 			);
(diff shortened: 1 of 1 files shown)


MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -1368,37 +1367,6 @@
...
-			);
-			inputPaths = (
-				"$(SRCROOT)/Carthage/Build/iOS/Mapbox.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/MapboxDirections.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/SDWebImage.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/Polyline.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/AWSPolly.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/AWSCore.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/Turf.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/Solar.framework",
```
1ec5: Does the developer need to copy these dependencies in their own application now? If so, we should update the installation instructions – this is a significant change, given the number of dependencies.

frederoni: Yes, but this has always been a requirement when using Carthage, and it's stated in their documentation when building for iOS, tvOS, or watchOS.

IIRC, I added this phase because we were unable to test frameworks without an application target but this doesn't seem to be an issue anymore.

1ec5: Yeah, I’m just thinking we should list these dependencies in the readme so developers know which ones are for the navigation SDK. It isn’t obvious that this SDK depends on AWSCore, for example.

PR #931, opened 2017-12-15
Show all banners along step

bsudekum:
Previously we were only showing a single banner along a step because we could not render the correct icon for a banner instruction which was not the actual maneuver. This will be solved once https://github.com/mapbox/MapboxDirections.swift/pull/225 is available and we can use the type/modifier to build an arrow. 

/cc @mapbox/navigation-ios @mapbox/guidance

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -436,11 +436,22 @@ class RouteMapViewController: UIViewController {
         let stepProgress = routeProgress.currentLegProgress.currentStepProgress
         let distanceRemaining = stepProgress.distanceRemaining
         
-        guard let visualInstruction = routeProgress.currentLegProgress.currentStep.instructionsDisplayedAlongStep?.last else { return }
+        guard let visualInstructions = routeProgress.currentLegProgress.currentStep.instructionsDisplayedAlongStep else { return }
         
-        instructionsBannerView.set(visualInstruction.primaryTextComponents, secondaryInstruction: visualInstruction.secondaryTextComponents)
+        for (visualInstructionIndex, visualInstruction) in visualInstructions.enumerated() {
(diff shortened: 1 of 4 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -436,11 +436,22 @@ class RouteMapViewController: UIViewController {
...
         let distanceRemaining = stepProgress.distanceRemaining
         
-        guard let visualInstruction = routeProgress.currentLegProgress.currentStep.instructionsDisplayedAlongStep?.last else { return }
+        guard let visualInstructions = routeProgress.currentLegProgress.currentStep.instructionsDisplayedAlongStep else { return }
+        
+        for (visualInstructionIndex, visualInstruction) in visualInstructions.enumerated() {
+            if routeProgress.currentLegProgress.currentStepProgress.distanceRemaining <= visualInstruction.distanceAlongStep && visualInstructionIndex >= routeProgress.currentLegProgress.currentStepProgress.visualInstructionIndex {
+                
+                instructionsBannerView.set(visualInstruction.primaryTextComponents, secondaryInstruction: visualInstruction.secondaryTextComponents)
+                instructionsBannerView.maneuverView.step = routeProgress.currentLegProgress.upComingStep
```
bsudekum: I'm not thrilled about this. Ideally, there would be a proper initializer here instead.

bsudekum:
@1ec5 @ericrwolfe given the compromises we're making here to make the type/modifier work, do you think we need to go back to the drawing board on the server to make this 💯 ?

1ec5:
> given the compromises we're making here to make the type/modifier work, do you think we need to go back to the drawing board on the server to make this 💯 ?

Yes, I think so. It would be best if a banner that details a given step could be nothing but a reference to that step. A banner that serves a different purpose should say why, such as `"type": "straightaway"` for a “stay on” banner along the majority of a long step.

bsudekum:
Superseded by https://github.com/mapbox/mapbox-navigation-ios/pull/1085

PR #936, opened 2017-12-18
Mute immediately

frederoni:
- Tapping mute now mutes the voice controller immediately.
- Refactored settings for improved type safety and less unwrapping.

@bsudekum @JThramer 👀


MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -0,0 +1,85 @@
+import Foundation
+
+/**
+ Emitted when something changed in the `NavigationSettings`.
+ */
+public let NavigationSettingsDidChange = Notification.Name("MBNavigationSettingsDidChange")
```
JThramer: Consider using an extension here. 

```swift
extension Notification.Name
{
    static let navigationSettingsDidChange = "MBNavigationSettingsDidChange"
}
```

MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -0,0 +1,85 @@
...
+                
+                UserDefaults.standard.set(val, forKey: key.prefixed)
+                NotificationCenter.default.post(name: NavigationSettingsDidChange, object: nil, userInfo: [key: val])
+                
+                found = true
+                break
+            }
+        }
+        
+        if found == false {
```
JThramer:
```swift
if !found {
...
}
```

MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    /**
+     Indicates whether the voice controller should be muted or not.
+     */
+    @objc public dynamic var voiceMuted     : Bool      = false
+    
+    
+    public static let shared = NavigationSettings()
+    
+    lazy var properties: [Mirror.Child] = {
```
JThramer: I'm usually not a big fan of using `Mirror` in production code, but I see why you used it here. Since this is such an unusual pattern, can you add some doc here explaining what `properties` does and the mechanism of action at play in this class?

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -121,19 +121,38 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
     
     func suspendNotifications() {
         NotificationCenter.default.removeObserver(self, name: RouteControllerDidPassSpokenInstructionPoint, object: nil)
         NotificationCenter.default.removeObserver(self, name: RouteControllerWillReroute, object: nil)
         NotificationCenter.default.removeObserver(self, name: RouteControllerDidReroute, object: nil)
+        NotificationCenter.default.removeObserver(self, name: NavigationSettingsDidChange, object: nil)
+    }
+    
+    @objc func settingsDidChange(_ notification: Notification) {
+        let muteKey = #keyPath(NavigationSettings.voiceMuted)
```
JThramer: How about using Smart KeyPaths here?

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -121,19 +121,38 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
...
+        let volumeKey = #keyPath(NavigationSettings.voiceVolume)
+        let muteChanged = notification.userInfo?[muteKey] != nil
```
JThramer: If this notification could be subscribed to in more than one place, consider refactoring these into computed properties in an extension.

MapboxCoreNavigation/NavigationSettings.swift
bsudekum: Should `MBNavigationSettingsDidChange ` be pulled out into a constant?

frederoni: I was following the same pattern we use for route controller notifications b/c `Notification.Name` doesn't bridge to Objective-C, but after a second thought, there's no reason why we shouldn't have properly typed notifications in Swift.

PR #937, opened 2017-12-18
Account for current stepIndex as well

bsudekum:
_Followup to: https://github.com/mapbox/mapbox-navigation-ios/pull/910_

In this final check, we are not accounting for the fact that we're are looking at the [slice](https://github.com/mapbox/mapbox-navigation-ios/compare/fix-reroute-future-check?expand=1#diff-851f438872328e98cbe395ac0733c16bR315) of the route's steps.

Take for example:

The current stepIndex is 5, we look at the remaining steps, find the closest to be 6. The closest step would be found in the iteration through this for loop, with a `currentStepIndex `. We Should return the index count in the for loop _plus_ the current stepIndex to get 6.

/cc @mapbox/navigation-ios


bsudekum:
@1ec5 added test

PR #938, opened 2017-12-18
Allow step to increment before checking for reroute

bsudekum:
While investigating https://github.com/mapbox/mapbox-navigation-ios/pull/937, I found another bug: we're often times doing the `ultimate-is-the-user-on-another-step` check too often. 

I was able to debug https://github.com/mapbox/mapbox-navigation-ios/pull/937 because when the user is moving quickly over straight maneuvers, we call the `ultimate-is-the-user-on-another-step`. This is totally unnecessary, we should only be doing this check when the user is not on the route. 

The change here, changes `monitorStepProgress ` to run before `userIsOnRoute `. This allows us to increment the step if the user has completed the step before checking if the user is on the route. This is important because when the user is moving quickly, the user can be beyond the last coordinate in on the step. In this case, we want to increment the step counter and not reroute them.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -553,13 +553,14 @@ extension RouteController: CLLocationManagerDelegate {
                     ])
             }
         }
+        
+        updateDistanceToIntersection(from: location)
+        monitorStepProgress(location)
+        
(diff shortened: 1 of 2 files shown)


ericrwolfe:
Adding @cammace and @danesfeder to 👀 since this change also seems to affect these two tickets on Android:
* https://github.com/mapbox/mapbox-navigation-android/issues/238
* https://github.com/mapbox/mapbox-navigation-android/issues/578

PR #944, opened 2017-12-19
Transifex 2017 12 19

frederoni:
Pulled in the latest translations from Transifex

@1ec5 @bsudekum @JThramer 👀

### DIFF
--- a/MapboxNavigation/Resources/nl.lproj/Localizable.strings
+++ b/MapboxNavigation/Resources/nl.lproj/Localizable.strings
@@ -44,7 +44,7 @@
 "RESUME" = "Doorgaan";
 
 /* The text of a banner that appears during turn-by-turn navigation when route simulation is enabled. */
-"USER_IN_SIMULATION_MODE" = "Navigatie Simuleren";
+"USER_IN_SIMULATION_MODE" = "Simulating Navigation at %d×";
 
 /* Format for displaying destination and intermediate waypoints; 1 = source ; 2 = destinations */
 "WAYPOINT_DESTINATION_VIA_WAYPOINTS_FORMAT" = "%1$@, via %2$@";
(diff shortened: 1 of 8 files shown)


MapboxNavigation/Resources/nl.lproj/Localizable.strings
```diff
@@ -44,7 +44,7 @@
 "RESUME" = "Doorgaan";
 
 /* The text of a banner that appears during turn-by-turn navigation when route simulation is enabled. */
-"USER_IN_SIMULATION_MODE" = "Navigatie Simuleren";
+"USER_IN_SIMULATION_MODE" = "Simulating Navigation at %d×";
```
JThramer: Shouldn't this be `"Navigatie Simuleren op %dx"`?

frederoni: It should but this string has not yet been translated in Transifex.

MapboxNavigation/Resources/sv.lproj/Localizable.strings
```diff
@@ -44,7 +44,7 @@
 "RESUME" = "Återuppta";
 
 /* The text of a banner that appears during turn-by-turn navigation when route simulation is enabled. */
-"USER_IN_SIMULATION_MODE" = "Simulera navigation";
+"USER_IN_SIMULATION_MODE" = "Simulerar Navigation  %d×";
```
JThramer: I'm nowhere near proficient in Swedish, but it kind of looks like a preposition is missing here?

frederoni: Good catch

1ec5:
Bulgarian is also ready to be added. Not sure why the scripts didn’t pick it up for this PR.

PR #947, opened 2017-12-20
 Assert if app is missing location permissions

bsudekum:
We should fail the app sooner if it does not contain location permissions. Otherwise, strange things can occur while we wait around.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -204,7 +204,7 @@ open class RouteController: NSObject {
         }
         
         if Bundle.main.locationAlwaysUsageDescription == nil && Bundle.main.locationWhenInUseUsageDescription == nil && Bundle.main.locationAlwaysAndWhenInUseUsageDescription == nil {
-            assert(false, "This application’s Info.plist file must include a NSLocationWhenInUseUsageDescription. See https://developer.apple.com/documentation/corelocation for more information.")
+            preconditionFailure("This application’s Info.plist file must include a NSLocationWhenInUseUsageDescription. See https://developer.apple.com/documentation/corelocation for more information.")
         }
     }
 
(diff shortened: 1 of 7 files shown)


bsudekum:
Hrm, either I'm getting the wrong plist here or I'm missing an entry somewhere.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -202,6 +202,11 @@ open class RouteController: NSObject {
         DispatchQueue.main.async {
             self.startEvents(route: route)
         }
+        
+        guard let _ = Bundle.main.bundleIdentifier else { return }
```
bsudekum: Framework tests do not have a "working" Info.plist- values added to it are not actually respected since its not an app.

MapboxCoreNavigationTests/CocoaPodsTest/PodInstall/PodInstall/Info.plist
```diff
@@ -41,5 +41,7 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>NSLocationAlwaysUsageDescription</key>
```
1ec5: This key is ignored in iOS 11.0 and above. You need `NSLocationAlwaysAndWhenInUseUsageDescription` instead:

https://developer.apple.com/documentation/corelocation/choosing_the_authorization_level_for_location_services/requesting_always_authorization

MapboxCoreNavigation/RouteController.swift
```diff
@@ -202,6 +202,11 @@ open class RouteController: NSObject {
...
+        if Bundle.main.locationAlwaysUsageDescription == nil && Bundle.main.locationWhenInUseUsageDescription == nil && Bundle.main.locationAlwaysAndWhenInUseUsageDescription == nil {
+            preconditionFailure("This application’s Info.plist file must include a NSLocationWhenInUseUsageDescription. See https://developer.apple.com/documentation/corelocation for more information.")
```
1ec5: This is a very generic URL. Link to https://developer.apple.com/documentation/corelocation/choosing_the_authorization_level_for_location_services/requesting_always_authorization#overview instead.

Issue #830, opened 2017-11-10
Custom controls


frederoni:
We've had a few requests about adding custom controls to the navigation view controller.
I suggest we expose a grid that's properly aligned with the map view. A public UIStackView in each corner that's customizable by the user. Similar to the current set of controls (overview, mute & feedback), pinned to the top|right, but for all edges and exposed publicly.

/cc @bsudekum @1ec5 @JThramer

1ec5:
Would the grid be limited to RouteMapViewController’s portion of the window, or would other parts of the window be in scope for this feature?

frederoni:
RouteMapViewController is full size as of #785 (perhaps we should consider merging RouteMapVC and NavigationVC) but I'm thinking the size of the map view to begin with. Adding components to the instructions banner or the bottom banner would be trickier.

1ec5:
> perhaps we should consider merging RouteMapVC and NavigationVC

In the future, that might limit our ability to temporarily replace the map view controller with a different view controller for specialized UI. But I agree that flattening the view controller hierarchy would probably simplify a few things like the delegates.

JThramer:
Hey guys,

So bobby asked me to take a good look at this, and after giving it some serious thought over the past few days, here's what I've come up with.

While thinking through this problem, it occurred to me that there are actually two main use-cases to consider:

- Advanced: It seems that for most of our bigger clients, it's highly unlikely that they will use our vanilla UI as-written. Ideally they would use components of our UI framework in an ad-hoc, or "Lego" fashion, engineering a solution that fits their needs, while still taking advantage of as much "code reuse" as they can. For example, they may use our `InstructionBannerView`, but on a ViewController of their own making. Because of this, we should commit to providing _modular, loosely coupled_ UI components with _clear, expressible interfaces_.
- Simple: There are other implementers out there who may not have such an advanced use-case, perhaps they just want to add "a little something extra" to the vanilla UI. In this case, it may be tedious to roll their own `UIViewController` or subclass `RouteMapViewController` just to add "one extra thing", so it _does_ make sense to provide some customizable functionality. However, we want to keep scope in mind here, as the value here would be to provide an easy interface for adding trivial customization, _not_ to create a "[Bootstrap](https://getbootstrap.com) for `RouteMapViewController`". 

[...]

PR #952, opened 2017-12-20
NavigationViewController Responsiveness

JThramer:
This implements part of #830, where I'm creating a "Nav-bar" demo in [navigation-ios-examples](https://github.com/mapbox/navigation-ios-examples) where the NavigationViewController is pushed onto a `UINavigationController` VC stack. So far, there are three main issues that I see surrounding responsiveness:

- [x] NavigationViewController is not subclassable outside of the `MapboxNavigation` module.
- [x] NavigationViewController should be responsive to being pushed as opposed to presented ("X" button)
- [x] ~NavigationViewController should be able to be instantiated in Interface Builder.~ (punted to #968)

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -154,6 +154,19 @@ public protocol NavigationViewControllerDelegate {
      Returns the center point of the user course view in screen coordinates relative to the map view.
      */
     @objc optional func navigationViewController(_ navigationViewController: NavigationViewController, mapViewUserAnchorPoint mapView: NavigationMapView) -> CGPoint
+    
+    /**
+     Returns whether the `BottomBannerView` in the `RouteMapViewController` should show a cancel button.
+     
+     If implemented, this method is called at `RouteMapViewController`'s `viewWillAppear()` time. This is used to initally determine whether the cancel button should be shown when the view first appears. If you need to set the visibility of the cancel button after this point, manipulate the `BottomBannerView.isShowingCancelButton` property directly.
+     
+     - parameter navigationViewController: The navigation view controller that needs to know wheather or not to show the cancel button.
+     - parameter bottomBanner: The bottom banner containing the cancel button.
+     - note: If this method is not implemented, the default behavior is to show the cancel button if the `NavigationViewController` was modally presented.
+     - returns: True to show the cancel button, False to hide.
+     */
+    @objc optional func navigationViewController(_ navigationViewController: NavigationViewController, shouldShowCancelButtonIn bottomBanner: BottomBannerView) -> Bool
+
 }
 
 /**
@@ -420,6 +433,7 @@ open class NavigationViewController: UIViewController, RouteMapViewControllerDel
         UIApplication.shared.applicationIconBadgeNumber = 0
     }
     
+    //MARK: - RouteMapViewControllerDelegate Methods
     func navigationMapView(_ mapView: NavigationMapView, routeCasingStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer? {
         return delegate?.navigationMapView?(mapView, routeCasingStyleLayerWithIdentifier: identifier, source: source)
     }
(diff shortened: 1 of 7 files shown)


JThramer:
So after talking with @frederoni it sounds like the plan here is to spin off that third task into it's own ticket (it's not critical to this story), and get this PR in as-is so we can focus on the "embedded view" part of #830.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -154,6 +154,19 @@ public protocol NavigationViewControllerDelegate {
...
+     Returns whether the `BottomBannerView` in the `RouteMapViewController` should show a cancel button.
+     
+     If implemented, this method is called at `RouteMapViewController`'s `viewWillAppear()` time. This is used to initally determine whether the cancel button should be shown when the view first appears. If you need to set the visibility of the cancel button after this point, manipulate the `BottomBannerView.isShowingCancelButton` property directly.
+     
+     - parameter navigationViewController: The navigation view controller that needs to know wheather or not to show the cancel button.
+     - parameter bottomBanner: The bottom banner containing the cancel button.
+     - note: If this method is not implemented, the default behavior is to show the cancel button if the `NavigationViewController` was modally presented.
+     - returns: True to show the cancel button, False to hide.
+     */
+    @objc optional func navigationViewController(_ navigationViewController: NavigationViewController, shouldShowCancelButtonIn bottomBanner: BottomBannerView) -> Bool
```
1ec5: This method bridges to Objective-C as `-navigationViewController:shouldShowCancelButtonIn:`. To conform to Objective-C naming conventions, we’ll need `@objc(navigationViewController:shouldShowCancelButtonInBottomBannerView:)`.

However, I’m not sure about passing the BottomBannerView as a parameter. Can you think of a situation in which the implementation would need to consult a property on BottomBannerView to decide whether to return true or false? If that does indeed turn out to be useful, we should make the BottomBannerView accessible via NavigationViewController, ideally by making `NavigationViewController.routeMapViewController` public. That would be useful in its own right.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -420,6 +433,7 @@ open class NavigationViewController: UIViewController, RouteMapViewControllerDel
         UIApplication.shared.applicationIconBadgeNumber = 0
     }
     
+    //MARK: - RouteMapViewControllerDelegate Methods
```
1ec5: A common practice in this codebase is to break out a separate extension and have it conform to the protocol. That way, the compiler will complain if any of the protocol’s required methods lies outside the extension.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -961,6 +964,7 @@ extension RouteMapViewController: StatusViewDelegate {
     }
 }
 
+//MARK: - RouteMapViewControllerDelegate
```
1ec5: This protocol already gets its own section in the jump bar menu. A documentation comment would be useful, however. 😉

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -976,9 +980,19 @@ protocol RouteMapViewControllerDelegate: class {
...
     func mapViewControllerDidCancelNavigation(_ mapViewController: RouteMapViewController)
+    func mapViewController(_ mapViewController: RouteMapViewController, shouldShowCancelButtonIn bottomBanner: BottomBannerView) -> Bool
     func mapViewController(_ mapViewController: RouteMapViewController, didSend feedbackId: String, feedbackType: FeedbackType)
     
     func mapViewController(_ mapViewController: RouteMapViewController, mapViewUserAnchorPoint mapView: NavigationMapView) -> CGPoint?
     
     func mapViewControllerShouldAnnotateSpokenInstructions(_ routeMapViewController: RouteMapViewController) -> Bool
 }
+
+extension RouteMapViewControllerDelegate where Self: UIViewController {
```
1ec5: Note that Objective-C doesn’t support generics on protocols, nor default method implementations in protocols. These limitations may become relevant if we decide to make RouteMapViewController public.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -154,6 +154,19 @@ public protocol NavigationViewControllerDelegate {
      Returns the center point of the user course view in screen coordinates relative to the map view.
      */
     @objc optional func navigationViewController(_ navigationViewController: NavigationViewController, mapViewUserAnchorPoint mapView: NavigationMapView) -> CGPoint
+    
+    /**
+     Returns whether the `BottomBannerView` in the `RouteMapViewController` should show a cancel button.
+     
+     If implemented, this method is called at `RouteMapViewController`'s `viewWillAppear()` time. This is used to initally determine whether the cancel button should be shown when the view first appears. If you need to set the visibility of the cancel button after this point, manipulate the `BottomBannerView.isShowingCancelButton` property directly.
```
1ec5: Is an explicit setup method necessary? This note makes it sound like the developer can accomplish the same effect (from the user’s perspective) by setting `isShowingCancelButton` between the call to `init(for:directions:styles:locationManager:)` and `viewDidAppear()`, such as in `prepare(for:)`.

JThramer:
This is ultra-stale, so I'm going to close. If this was a mistake, feel free to re-open.

PR #953, opened 2017-12-21
Flip U-turn icon in right-side regions only

1ec5:
As far as I know, the U-turn icon is already flipped correctly in countries that drive on the left. But checking the turn angle is unintuitive compared to checking the driving side.

The only downside would be U-turns that turn to the wrong side, which are pretty uncommon as far as I know. I can only think of [one example](http://www.openstreetmap.org/way/524469535) of a U-turn to the right in a right-sided country, and that’s more of a fanciful turn lane indication than an actual U-turn.

turn-u

/cc @bsudekum

### DIFF
--- a/MapboxNavigation/ManeuverView.swift
+++ b/MapboxNavigation/ManeuverView.swift
@@ -120,7 +120,7 @@ public class ManeuverView: UIView {
                 flip = true
             case .uTurn:
                 ManeuversStyleKit.drawArrow180right(frame: bounds, resizing: resizing, primaryColor: primaryColor)
-                flip = angle < 0
+                flip = step.drivingSide == .right
             default:
                 ManeuversStyleKit.drawArrowstraight(frame: bounds, resizing: resizing, primaryColor: primaryColor)
             }


MapboxNavigation/ManeuverView.swift
```diff
@@ -120,7 +120,7 @@ public class ManeuverView: UIView {
                 flip = true
             case .uTurn:
                 ManeuversStyleKit.drawArrow180right(frame: bounds, resizing: resizing, primaryColor: primaryColor)
-                flip = angle < 0
+                flip = step.drivingSide == .right
```
bsudekum: Actually, should this be .left. Looking elsewhere:

https://github.com/mapbox/mapbox-navigation-ios/blob/eb5b231c8d9b6fa4a4c46ed222457d725b2963e5/MapboxNavigation/ManeuverView.swift#L89

1ec5: `ManeuversStyleKit.drawArrow180right(frame:resizing:primaryColor:)` draws an arrow to the right by default, right? So we’d want to flip that when driving to the right (U-turning to the left) but keep it as-is when driving to the left (U-turning to the right).

Issue #964, opened 2018-01-02
While using mapbox-core-navigation solely, exiting does not trigger cancel event


bsudekum:
When dismissing `mapbox-core-navigation`, the cancel event should always be triggered.

/cc @ericrwolfe

PR #965, opened 2018-01-02
Always send cancel event even when using mapbox-core-navigation

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/964

This moves to call the cancel feedback event when the RouteController is deinited. Unfortunately had to make some properties public.

/cc @mapbox/navigation-ios @ericrwolfe

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -158,6 +158,18 @@ open class RouteController: NSObject {
             NotificationCenter.default.post(name: .routeControllerDidReroute, object: self, userInfo: userInfo)
         }
     }
+    
+    /**
+     :nodoc:
+     A star rating from 0 to 5 where 0 is the worst and 5 is the best.
+     */
+    public var endOfRouteStarRating: Int?
+    
+    /**
+     :nodoc:
+     The users comment amount their navigation experience.
+     */
+    public var endOfRouteComment: String?
 
     var isRerouting = false
     var lastRerouteLocation: CLLocation?
@@ -211,6 +223,7 @@ open class RouteController: NSObject {
 
     deinit {
         suspendLocationUpdates()
+        sendCancelEvent(rating: endOfRouteStarRating, comment: endOfRouteComment)
         checkAndSendOutstandingFeedbackEvents(forceAll: true)
         suspendNotifications()
         UIDevice.current.isBatteryMonitoringEnabled = false
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -159,17 +159,8 @@ open class RouteController: NSObject {
         }
     }
     
-    /**
-     :nodoc:
-     A star rating from 0 to 5 where 0 is the worst and 5 is the best.
-     */
-    public var endOfRouteStarRating: Int?
-    
-    /**
-     :nodoc:
-     The users comment amount their navigation experience.
-     */
-    public var endOfRouteComment: String?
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -158,6 +158,18 @@ open class RouteController: NSObject {
...
+     :nodoc:
+     A star rating from 0 to 5 where 0 is the worst and 5 is the best.
+     */
+    public var endOfRouteStarRating: Int?
+    
+    /**
+     :nodoc:
+     The users comment amount their navigation experience.
+     */
+    public var endOfRouteComment: String?
```
ericrwolfe: Could we make this a public function to set both at once, so that you can't get the properties publicly? Something like `setEndOfRoute(rating: Int, comment: String?)`.

This way, if we wanted in the future we could control the conditions under which these properties could be set (e.g. only after enough of the route was completed, only if not set before, etc).

MapboxCoreNavigation/RouteController.swift
```diff
@@ -211,6 +223,7 @@ open class RouteController: NSObject {
 
     deinit {
         suspendLocationUpdates()
+        sendCancelEvent(rating: endOfRouteStarRating, comment: endOfRouteComment)
```
ericrwolfe: @bsudekum were you able to verify that a telem event gets sent on cancel even with the call happening during deinit?

bsudekum: Yep, confirmed it is sent.

bsudekum:
@ericrwolfe fixed.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -428,6 +432,14 @@ open class RouteController: NSObject {
             outstandingFeedbackEvents.remove(at: index)
         }
     }
+    
+    /**
+     Set the rating and any comment the user may have about their route. Only used when exiting navigaiton.
+     */
+    @objc public func setEndOfRoute(rating: Int, comment: String?) {
+        endOfRouteStarRating = rating
```
ericrwolfe: We might want to add some basic validation to the rating to conform to the spec:

```
Rating from `0-100` for post ride feedback. `-1 = unrated`
5 star rating system:

- 0 stars (unrated) = `-1`
- 1 star = `0`
- 2 stars = `25`
- 3 stars = `50`
- 4 stars = `75`
- 5 stars = `100`
```

bsudekum: Validation is happening here:
https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/MMEEventsManager.swift#L205

Unsure if we're multiplying yet.

bsudekum: Yep, we're multiplying correct.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -428,6 +432,14 @@ open class RouteController: NSObject {
             outstandingFeedbackEvents.remove(at: index)
         }
     }
+    
+    /**
+     Set the rating and any comment the user may have about their route. Only used when exiting navigaiton.
+     */
+    @objc public func setEndOfRoute(rating: Int, comment: String?) {
```
1ec5: `setEndOfRoute(_:comment:)` is a weird name for a method that doesn’t set an `endOfRoute` property. How about `setRouteRating(_:comment:)` / `-setRouteRating:comment:` or `rateRoute(rating:comment:)` / `-rateRouteWithRating:comment:`?

PR #966, opened 2018-01-03
Prevent user puck spinning at end of route

bsudekum:
When snapping and looking at points ahead/behind user, we should not use the point ahead/behind if it has a distance of zero. In this case, the user is either at the beginning or end of the route. When using a point that has a distance of zero, the calculated course becomes wildly inaccurate causing spinning. 

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -377,7 +377,7 @@ open class RouteController: NSObject {
         var averageRelativeAngle: Double!
         // User is at the beginning of the route, there is no closest point behind the user.
         if pointBehindClosest.distance <= 0 && pointAheadClosest.distance > 0 {
-            averageRelativeAngle = relativeAnglepointBehind
+            averageRelativeAngle = relativeAnglepointAhead
         // User is at the end of the route, there is no closest point in front of the user.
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -373,7 +373,18 @@ open class RouteController: NSObject {
         let wrappedCourse = location.course.wrap(min: -180, max: 180)
         let relativeAnglepointBehind = (wrappedPointBehind - wrappedCourse).wrap(min: -180, max: 180)
         let relativeAnglepointAhead = (wrappedPointAhead - wrappedCourse).wrap(min: -180, max: 180)
-        let averageRelativeAngle = pointBehindClosest.distance > 0 ? (relativeAnglepointBehind + relativeAnglepointAhead) / 2 : relativeAnglepointAhead
+        
+        var averageRelativeAngle: Double!
```
frederoni: Would be safer to define this as a constant `let averageRelativeAngle: Double` which works b/c the assignment below is exhaustive.

bsudekum: Wow, did not know you could do that!

PR #970, opened 2018-01-03
Remain in throttled fps mode longer

bsudekum:
This changes:

* fixes a bug where we would not actually remain in throttled fps mode when the maneuver was straight
* remain in throttled fps mode while the upcoming maneuver is slight right/left
* reduces the time around the maneuver we are in non-throttled mode slightly.

_Note, these changes are only while the device is unplugged._

/cc @mapbox/navigation-ios


bsudekum:
_bump_

PR #972, opened 2018-01-04
Emit didArriveAt waypoint delegate for all waypoints

bsudekum:
As of right now, this function is only returning didArriveAt waypoint for the last waypoint and not all waypoints.

/cc @mapbox/navigation-ios

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -335,7 +335,7 @@ extension ViewController: NavigationViewControllerDelegate {
     // If however you would like to pause and allow the user to provide input, set this delegate method to false.
     // This does however require you to increment the leg count on your own. See the example below in `confirmationControllerDidConfirm()`.
     func navigationViewController(_ navigationViewController: NavigationViewController, shouldIncrementLegWhenArrivingAtWaypoint waypoint: Waypoint) -> Bool {
-        return false
+        return exampleMode == .multipleWaypoints ? false : true
     }
 
     func navigationViewController(_ navigationViewController: NavigationViewController, didArriveAt waypoint: Waypoint) {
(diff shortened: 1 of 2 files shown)


bsudekum:
I also updated this so that when the developer implements their own custom waypoint arrival UI, we do not also show the end of route feedback.

Examples/Swift/ViewController.swift
```diff
@@ -335,7 +335,7 @@ extension ViewController: NavigationViewControllerDelegate {
     // If however you would like to pause and allow the user to provide input, set this delegate method to false.
     // This does however require you to increment the leg count on your own. See the example below in `confirmationControllerDidConfirm()`.
     func navigationViewController(_ navigationViewController: NavigationViewController, shouldIncrementLegWhenArrivingAtWaypoint waypoint: Waypoint) -> Bool {
-        return false
+        return exampleMode == .multipleWaypoints ? false : true
```
JThramer: `return exampleMode != .multipleWaypoints`

MapboxNavigation/NavigationViewController.swift
```diff
@@ -531,7 +531,16 @@ extension NavigationViewController: RouteControllerDelegate {
     }
     
     public func routeController(_ routeController: RouteController, didArriveAt waypoint: Waypoint) {
-        guard routeController.routeProgress.isFinalLeg else { return }
+        guard routeController.routeProgress.isFinalLeg else {
```
JThramer:
```swift
let isFinalLeg = routeController.routeProgress.isFinalLeg
let shouldIncrementLeg = delegate?.navigationViewController?(self, shouldIncrementLegWhenArrivingAtWaypoint: waypoint) ?? false

if !isFinalLeg || !shouldIncrementLeg {
    return delegate?.navigationViewController?(self, didArriveAt: waypoint)
}
```

IDK, this may be "6 one way, half-a-dozen the other"

Examples/Swift/ViewController.swift
bsudekum: Yeah, I thought about this, but I think for the purpose of making the example easy to read, I thought I'd explicitly say true/false.

Issue #973, opened 2018-01-04
Status view can grow in height


bsudekum:
![simulator screen shot - iphone 8 - 2018-01-04 at 11 50 15](https://user-images.githubusercontent.com/1058624/34581512-81b360d6-f145-11e7-963c-08f5d9fe5af1.png)

to reproduce:

* turn off simulation
* start a route
* apply a breakpoint
* wait 2 minutes
* start breakpoint
* many reroutes will now occur, status view will grow

/cc @frederoni

bsudekum:
Removing the animation here

https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/StatusView.swift#L113-L119

seems to fix the issue. Unsure why applying multiple animations would cause this problem. Investigating.

PR #976, opened 2018-01-05
Remove StatusView animations before starting another

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/973

It looks like stacking animations one after another caused the height to grow. It looks like we can prevent this by removing all animations on the layer before applying another.

/cc @mapbox/navigation-ios @frederoni

### DIFF
--- a/MapboxNavigation/StatusView.swift
+++ b/MapboxNavigation/StatusView.swift
@@ -88,6 +88,7 @@ public class StatusView: UIView {
     }
     
     func show(_ title: String, showSpinner: Bool) {
+        layer.removeAllAnimations()
         textLabel.text = title
         activityIndicatorView.hidesWhenStopped = true
         if showSpinner {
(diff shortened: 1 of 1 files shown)


bsudekum:
Closer: it looks like the black bar is now the correct height, but the y position of the text is off 
simulator screen shot - iphone 8 - 2018-01-05 at 10 32 58

frederoni:
There are a few radars open related to the behavior of showing/hiding views inside UIStackViews.
These are triggered by an uneven call to show/hide. In our case, we guard `isHidden == true` when showing but we can hide multiple times consecutively.

http://www.openradar.me/22819594
https://openradar.appspot.com/25087688

Another issue could be that the `StatusView`’s height constraint’s priority is set to 999 (default is 1000). IIRC, this is being used to provide a height-based show/hide animation but could be worth investigating if we can use the default priority and animate the vertical position instead of the height.

bsudekum:
Experiment did not work.

PR #981, opened 2018-01-08
Designable navigation view

frederoni:
This PR adds a new `NavigationView` which will replace the `RouteMapViewController`’s view that's currently laid-out in the storyboard. A designable view of the whole UI encapsulating all other views.

Hooking it up to `RouteMapViewController` will be done in a subsequent PR so this change alone won't effectively change anything aside from:

- `WayNameView` merged into `WayNameLabel`
- Renamed `LanesContainerView` to `LanesView`

screenshot 2018-01-10 17 11 22

@JThramer @bsudekum 👀


JThramer:
Awesome. Looks great so far!

frederoni:
This PR is ready for review. As noted in the description, this change alone won't change anything aside from:
- WayNameView merged into WayNameLabel
- Renamed LanesContainerView to LanesView

MapboxNavigation/LanesView.swift
```diff
@@ -20,6 +20,15 @@ public class LanesContainerView: LanesView {
         commonInit()
     }
     
+    public override func prepareForInterfaceBuilder() {
+        super.prepareForInterfaceBuilder()
+
+        for _ in 0...4 {
```
bsudekum: Are there are always 4?

MapboxNavigation/NavigationView.swift
```diff
@@ -0,0 +1,92 @@
...
+    // Vertically laid-out stack view below the instructions banner consisting of StatusView, NextBannerView, and LanesView.
+    weak var informationStackView: UIStackView!
+    // Vertically laid-out stack view below the information stack view ontop of the map view, docked
+    // to the top right, consisting of Overview, Mute, and Report button.
+    weak var floatingStackView: UIStackView!
+    weak var overviewButton: FloatingButton!
+    weak var muteButton: FloatingButton!
+    weak var reportButton: FloatingButton!
+    weak var rerouteReportButton: ReportButton!
+    weak var separatorView: SeparatorView!
```
bsudekum: Where are all these variables used?

bsudekum:
Initial test feels good. Interested to hear feedback from @1ec5 & @JThramer too.

MapboxNavigation/NavigationView.swift
```diff
@@ -0,0 +1,92 @@
...
+    weak var instructionsBannerContentView: InstructionsBannerContentView!
+    weak var lanesView: LanesView!
+    weak var nextBannerView: NextBannerView!
+    weak var statusView: StatusView!
+    weak var resumeButton: ResumeButton!
+    // Vertically laid-out stack view below the instructions banner consisting of StatusView, NextBannerView, and LanesView.
+    weak var informationStackView: UIStackView!
+    // Vertically laid-out stack view below the information stack view ontop of the map view, docked
+    // to the top right, consisting of Overview, Mute, and Report button.
+    weak var floatingStackView: UIStackView!
```
JThramer: The one major comment I have on this PR is this: Since this is no longer IB-based, it would be much cleaner to use a `lazy var` to set up all of these child views, like so:

```swift
lazy var floatingStackView: UIStackView = { 
    let stack = UIStackView()
    stack.axs = .vertical
    stack.distribution = .equalSpacing
    stack.spacing = 8
    stack.translatesAutoresizingMaskIntoConstraints = false
    return stack
} ()
```

then, `setupViews()` becomes this:

```swift
func setupViews() {
    [..., floatingStackView, overviewButton, muteButton, reportButton, ...].forEach(addSubview(_:))
}
```

However, this does work for now, so if we want to just get this across the line then I'll happily take care of it in my PR.

frederoni: Totally agree about the lazy-loaded ivars. Even more if we decide to lazy-load localized strings. However, I'm going to merge this as is b/c the utilization of this class will be done in a subsequent PR which is when we can measure the impact.

Issue #846, opened 2017-11-14
Add delegate method for overriding a voice announcement


bsudekum:
something like `voiceController.shouldSpeak(_ instruction:step:)`. If the developer returns a String, this string is said instead of the instruction for a step. We'll need to give the developer a `RouteStep` and also the `spokenVoiceIndex` here to allow them to figure out if they want to override this instruction.

/cc @ericrwolfe @ericdeveloper @1ec5

1ec5:
> We'll need to give the developer a `RouteStep` and also the `spokenVoiceIndex` here to allow them to figure out if they want to override this instruction.

Would passing in the current RouteProgress suffice?

bsudekum:
Oh yep

TCOA:
Could you please expand on this (is it available now? How to access?)

I really like the multiple waypoints (just what I need), though the spoken "You have arrived at your first destination" doesn't really work for us (prefer the standard "You have arrived at your destination" for our situation - the user knows they will continue after selecting the info in the Waypoint controller...)

If we can say what we want here, that would be really great!  (Of course, it will take a rework of all the software, but it should be more common to the release version so future mods should be simpler and worth the upfront effort to do it now)

bsudekum:
@TCOA this is not available yet.

PR #988, opened 2018-01-09
add voiceController willSpeak delegate method

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/846

This adds a method for listening on when a voice instruction will be spoken and allows developers to change the instruction as needed. Some things to chew on:

- developers will be on their own when it comes to SSML. It'll be about as difficult or easy as they want to make it.
- I with the dev did not have to include `distanceAlongStep` since it will have no impact on when the instruction is said.

Other than that, I think this a pretty clean addition.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -144,16 +144,17 @@ public class PollyVoiceController: RouteVoiceController {
         if let audioPlayer = audioPlayer, audioPlayer.isPlaying, let lastSpokenInstruction = lastSpokenInstruction {
             voiceControllerDelegate?.voiceController?(self, didInterrupt: lastSpokenInstruction, with: instruction)
         }
+        let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction) ?? instruction
         pollyTask?.cancel()
         audioPlayer?.stop()
-        lastSpokenInstruction = instruction
+        lastSpokenInstruction = modifiedInstruction
         
-        guard spokenInstructionsForRoute.object(forKey: instruction.ssmlText as NSString) == nil else {
-            play(spokenInstructionsForRoute.object(forKey: instruction.ssmlText as NSString)! as Data)
+        guard spokenInstructionsForRoute.object(forKey: modifiedInstruction.ssmlText as NSString) == nil else {
+            play(spokenInstructionsForRoute.object(forKey: modifiedInstruction.ssmlText as NSString)! as Data)
             return
         }
         
-        let input = pollyURL(for: instruction.ssmlText)
+        let input = pollyURL(for: modifiedInstruction.ssmlText)
         
         let builder = AWSPollySynthesizeSpeechURLBuilder.default().getPreSignedURL(input)
         builder.continueWith { [weak self] (awsTask: AWSTask<NSURL>) -> Any? in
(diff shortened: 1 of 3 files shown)


MapboxNavigation/PollyVoiceController.swift
```diff
@@ -144,24 +144,25 @@ public class PollyVoiceController: RouteVoiceController {
...
         }
+        let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction) ?? instruction
         pollyTask?.cancel()
         audioPlayer?.stop()
-        lastSpokenInstruction = instruction
+        lastSpokenInstruction = modifiedInstruction
         
-        guard spokenInstructionsForRoute.object(forKey: instruction.ssmlText as NSString) == nil else {
-            play(spokenInstructionsForRoute.object(forKey: instruction.ssmlText as NSString)! as Data)
+        guard spokenInstructionsForRoute.object(forKey: modifiedInstruction.ssmlText as NSString) == nil else {
```
JThramer: This may be better represented as: 

```swift
if let includedInstructions = spokenInstructionsForRoute.object(forKey: modifiedInstruction.ssmlText as NSString) {
    play(includedInstructions as Data)
    return
}
```

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -275,4 +277,12 @@ public protocol VoiceControllerDelegate {
...
     @objc(voiceController:didInterruptSpokenInstruction:withInstruction:)
     optional func voiceController(_ voiceController: RouteVoiceController, didInterrupt interruptedInstruction: SpokenInstruction, with interruptingInstruction: SpokenInstruction)
+    
+    /** Called when a spoken is about to speak. Useful if it is necessary to give a custom instruction instead. Noting, changing the `distanceAlongStep` property on `SpokenInstruction` will have no impact on when the instruction will be said.
+     
+     - parameter voiceController: The voice controller that experienced the interruption.
+     - parameter instruction: The spoken instruction that will be said.
+     **/
+    @objc(voiceController:willSpeakSpokenInstruction:)
+    optional func voiceController(_ voiceController: RouteVoiceController, willSpeak instruction: SpokenInstruction) -> SpokenInstruction
```
JThramer: You may also want to make the return value optional. That way the implementer can still implement the method, but still conditionally choose whether or not to override the spoken instruction.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -275,4 +277,12 @@ public protocol VoiceControllerDelegate {
...
     @objc(voiceController:didInterruptSpokenInstruction:withInstruction:)
     optional func voiceController(_ voiceController: RouteVoiceController, didInterrupt interruptedInstruction: SpokenInstruction, with interruptingInstruction: SpokenInstruction)
+    
+    /** Called when a spoken is about to speak. Useful if it is necessary to give a custom instruction instead. Noting, changing the `distanceAlongStep` property on `SpokenInstruction` will have no impact on when the instruction will be said.
+     
+     - parameter voiceController: The voice controller that experienced the interruption.
+     - parameter instruction: The spoken instruction that will be said.
+     **/
+    @objc(voiceController:willSpeakSpokenInstruction:)
+    optional func voiceController(_ voiceController: RouteVoiceController, willSpeak instruction: SpokenInstruction) -> SpokenInstruction?
```
1ec5: The arguments to this method don’t provide enough context for a typical application to intelligently replace the spoken instruction with a different instruction. For that, the implementation would need access to the RouteProgress, the one that comes in the `routeControllerDidPassSpokenInstructionPoint` notification.

PR #990, opened 2018-01-09
Support for style w/o location and single style

### LINKED ISSUE #989: Setting a single style leads to undefined behavior if incorrect time of day
I'm getting a situation where I'm building a style switcher and caching each style as a property to load later. When passing them into the NavigationViewController.init() method, sometimes the style loads, and sometimes it displays like so:

simulator screen shot - iphone x - 2018-01-08 at 17 18 35

```
class ListViewController: UITableViewController {

    let styleA = StyleA()
    let styleB = StyleB()
    let styleC = StyleC()

   ...
}
```

Loading navigation from ListViewController like so:
```swift
let viewController = NavigationViewController(for: route, styles: [self.styleA])
```

Each style is a subclass of DayStyle, and implements the init() and apply() methods:
```
class StyleBase: DayStyle {
    required init() {
        super.init()
        statusBarStyle = .lightContent
    }
    
    override func apply() {
        super.apply()
        Button.appearance().textColor = #colorLiteral(red: 0.9842069745, green: 0.9843751788, blue: 0.9841964841, alpha: 1)
    }
}
```

cc @frederoni

frederoni:
Fixes #989 

This change will apply a style even if the receiver doesn't have a fixed
location or if only a day _or_ night style is provided.

@bsudekum @ericrwolfe 👀

### DIFF
--- a/MapboxNavigation/StyleManager.swift
+++ b/MapboxNavigation/StyleManager.swift
@@ -98,8 +98,8 @@ open class StyleManager: NSObject {
     
     func applyStyle() {
         guard let location = delegate?.locationFor(styleManager: self) else {
-            // We can't calculate sunset or sunrise w/o a location so just apply all styles
-            for style in styles {
+            // We can't calculate sunset or sunrise w/o a location so just apply the first style
+            if let style = styles.first {
                 style.apply()
                 delegate?.styleManager?(self, didApply: style)
             }
--- a/MapboxNavigation/StyleManager.swift
+++ b/MapboxNavigation/StyleManager.swift
@@ -97,7 +97,24 @@ open class StyleManager: NSObject {
     }
     
     func applyStyle() {
-        guard let location = delegate?.locationFor(styleManager: self) else { return }
+        guard let location = delegate?.locationFor(styleManager: self) else {
+            // We can't calculate sunset or sunrise w/o a location so just apply all styles
+            for style in styles {
+                style.apply()
+                delegate?.styleManager?(self, didApply: style)
+            }
+            return
+        }
+        
+        // Single style usage
+        guard styles.count > 1 else {
+            if let style = styles.first {
+                style.apply()
+                delegate?.styleManager?(self, didApply: style)
+            }
+            return
+        }
+        
         let styleTypeForTimeOfDay = styleType(for: location)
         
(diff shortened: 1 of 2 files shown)


MapboxNavigation/StyleManager.swift
```diff
@@ -97,7 +97,24 @@ open class StyleManager: NSObject {
...
+            // We can't calculate sunset or sunrise w/o a location so just apply all styles
+            for style in styles {
+                style.apply()
+                delegate?.styleManager?(self, didApply: style)
+            }
+            return
+        }
+        
+        // Single style usage
+        guard styles.count > 1 else {
```
JThramer: Is this block really needed? You could easily do the same thing below via:

```swift
for style in styles {
             if style.styleType == styleTypeForTimeOfDay || styles.count == 1 {
                 style.apply()
                 currentStyleType = style.styleType
                 delegate?.styleManager?(self, didApply: style)
             }
         }
```

MapboxNavigation/StyleManager.swift
```diff
@@ -97,7 +97,24 @@ open class StyleManager: NSObject {
     }
     
     func applyStyle() {
-        guard let location = delegate?.locationFor(styleManager: self) else { return }
+        guard let location = delegate?.locationFor(styleManager: self) else {
+            // We can't calculate sunset or sunrise w/o a location so just apply all styles
+            for style in styles {
+                style.apply()
+                delegate?.styleManager?(self, didApply: style)
```
bsudekum: In the case where we cannot find a style, we're going to rapid fire go through all supplied styles and apply them? What if we just applied the first?

MapboxNavigation/StyleManager.swift
bsudekum: Yes less code is always nice, but IMO explicit guards make things a little more clear and they are easier to debug.

MapboxNavigation/StyleManager.swift
frederoni: I assume styles.count would mostly be 1 but I could imagine e.g. `[TopBarDayStyle(), BottomBarDayStyle(), FloatingButtonsDaysStyle()]` if one class would become unwieldy. 🤔

Issue #939, opened 2017-12-18
Reduce floating when completing a maneuver


bsudekum:
![new](https://user-images.githubusercontent.com/1058624/34131538-e2e9040c-e400-11e7-848f-612e37e1572e.gif)

Although this is the simulator, I'm still seeing hovering IRL. Noting, snapping should be looking at coordinates found on the upcoming step.

Ideally, the user puck would be locked onto the line while the user is actually on the route.

/cc @mapbox/navigation-ios

bsudekum:
Looking at a simulated route: 

![image](https://user-images.githubusercontent.com/1058624/34133101-da8bacf4-e407-11e7-970b-e2432c7aafd3.png)

It sort of makes sense that the puck would be animated around corners in a strange manner.

bsudekum:
This issue is not necessarily restricted to when the user is completing a step. This can also occur mid-step when a step turns 90 degrees left or right.

PR #992, opened 2018-01-09
Animate user puck along route

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/939

Currently, when a location update comes in, we simply animate the user puck from where it is to the new location. This however can cause shortcuts when the points are on either side of sharp turn as seen below:

image

This PR changes the user puck to animate along the route instead of in a straight line. 

todo:
- [ ] it'd probably be smart to allow straight line animations when the previous and current location are far enough apart
- [ ] fix jumping back and forth

/cc @mapbox/navigation-ios  @frederoni

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -90,6 +90,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     
     var previousLocation: CLLocation?
     
+    var routeProgress: RouteProgress?
+    
     struct FrameIntervalOptions {
         fileprivate static let durationUntilNextManeuver: TimeInterval = 7
         fileprivate static let durationSincePreviousManeuver: TimeInterval = 3
@@ -162,9 +164,9 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     @objc func progressDidChange(_ notification: Notification) {
         guard tracksUserCourse else { return }
         
-        let routeProgress = notification.userInfo![RouteControllerProgressDidChangeNotificationProgressKey] as! RouteProgress
+        routeProgress = notification.userInfo![RouteControllerProgressDidChangeNotificationProgressKey] as? RouteProgress
         
-        let stepProgress = routeProgress.currentLegProgress.currentStepProgress
+        guard let stepProgress = routeProgress?.currentLegProgress.currentStepProgress else { return }
         let expectedTravelTime = stepProgress.step.expectedTravelTime
         let durationUntilNextManeuver = stepProgress.durationRemaining
         let durationSincePreviousManeuver = expectedTravelTime - durationUntilNextManeuver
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -310,15 +312,34 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
         let duration: TimeInterval = animated ? 1 : 0
-        UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
-            self.userCourseView?.center = self.convert(location.coordinate, toPointTo: self)
-        }, completion: nil)
         
-        if let userCourseView = userCourseView as? UserCourseView {
-            userCourseView.update(location: location, pitch: camera.pitch, direction: direction, animated: animated, tracksUserCourse: tracksUserCourse)
+        if let route = routes?.first, let coordinates = route.coordinates, let previousLocation = previousLocation {
+            
+            let points = Polyline(coordinates).sliced(from: previousLocation.coordinate, to: location.coordinate).coordinates.map {
```
1ec5: NavigationMapView’s concept of routes is pretty display-focused. For example, the concept of a “first” route is pretty tied to the idea that it renders on top of the others. By contrast, RouteController is currently wholly responsible for calculating coordinates for the course view. Let’s see if we can push the snapped coordinate generation back down into RouteController and ensure that this method obtains its route from RouteProgress, if it even considers the route at all.

bsudekum: I agree with you that things could be moved around here, but a reminder `RouteController` does not do anything UI related. 

We could however look into bisecting lines between points even further in RouteController.

1ec5: > a reminder RouteController does not do anything UI related

That’s true, but RouteController does house the location snapping code that’s primarily used by the user puck. The trick is to make RouteController responsible for anything spatial but make the APIs generic enough that they don’t _have_ to be used with a particular UI feature.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -90,6 +90,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     
     var previousLocation: CLLocation?
     
+    var routeProgress: RouteProgress?
```
1ec5: I have reservations about making the map view responsible for storing the most recent route progress object, especially given that two of the three callers to `updateCourseTracking(location:animated:)` are in code that could easily pass in the route progress.

In the spirit of #993, what if we take the RouteMapViewController code that currently calls `updateCourseTracking(location:animated:)` and move it inside `NavigationMapView.progressDidChange(_:)` and a new method that handles foregrounding? At least that way the fact that this class hangs onto a route progress would be better encapsulated.

frederoni:
The puck looks slightly better connected with the map view but as you mentioned, the jumpiness is slightly worse. Feels a bit like we're patching the symptom here. I don't think it'll ever get perfect until we synchronize the user puck and the map view at the same frame interval.

To achieve that, we need an interpolated location from the RouteController. For the sake of simplicity, let's say we're running at 60FPS here:
screenshot 2018-01-11 14 23 53

30 frames after when we're half way through the turn, requesting the interpolated location from within the DisplayLink’s update call should give us this location:
screenshot_2018-01-11_14_23_53

We already have a similar computed property on the RouteController
<details>


https://github.com/mapbox/mapbox-navigation-ios/blob/d12aa60ecf990ba6ca3ea1d9e40f68446e2fb3e7/MapboxCoreNavigation/RouteController.swift#L503-L528

</details>
<br />
but it needs to be adapted since it wasn't written for a use case like this. It's probably not performant enough to hook up to the DisplayLink and it doesn't snap the course aggressively enough.

bsudekum:
Stale.

PR #993, opened 2018-01-09
Refactoring appearance logic out of RMVC and into Views

JThramer:
This PR concerns #994, where it is desired to refactor view-appearance logic out of the RMVC and into the appropriate views themselves. This makes things a bit more lonely coupled and easier to work with in a piecemeal fashion.

To-Do: 
- [x] InstructionsBannerView
- [x] NextBannerView
- [x] ~BottomBannerView?~ (already pretty well-encapsulated)

Phase 2 To-Do:
- [x] Create TopBannerViewController
- [ ] Plug it in to RMVC
- [ ] Investigate "seconds" ambiguity
- [ ] Make sure you remove all shims (such as `tooFar`)

/cc @bsudekum @1ec5 @frederoni @ericrwolfe

### DIFF
--- a/MapboxNavigation/InstructionsBannerView.swift
+++ b/MapboxNavigation/InstructionsBannerView.swift
@@ -71,6 +71,17 @@ open class BaseInstructionsBannerView: UIControl {
         delegate?.didTapInstructionsBanner(self)
     }
     
+    func update(progress: RouteLegProgress, location: CLLocation, secondsRemaining: TimeInterval) {
+        let stepProgress = progress.currentStepProgress
+        let distanceRemaining = stepProgress.distanceRemaining
+        
+        guard let visualInstruction = progress.currentStep.instructionsDisplayedAlongStep?.last else { return }
+        
+        set(visualInstruction.primaryTextComponents, secondaryInstruction: visualInstruction.secondaryTextComponents)
+        distance = distanceRemaining > 5 ? distanceRemaining : 0
+        maneuverView.step = progress.upComingStep
+    }
+    
     func set(_ primaryInstruction: [VisualInstructionComponent]?, secondaryInstruction: [VisualInstructionComponent]?) {
         primaryLabel.numberOfLines = secondaryInstruction == nil ? 2 : 1
         
--- a/MapboxNavigation/NextBannerView.swift
+++ b/MapboxNavigation/NextBannerView.swift
@@ -1,4 +1,5 @@
 import UIKit
+import MapboxDirections
 
 /// :nodoc:
 @objc(MBNextInstructionLabel)
@@ -68,6 +69,11 @@ open class NextBannerView: UIView {
         instructionLabel.leftAnchor.constraint(equalTo: leftAnchor, constant: 70).isActive = true
         instructionLabel.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
         instructionLabel.rightAnchor.constraint(equalTo: rightAnchor, constant: -16).isActive = true
+    }
+    
+    func update(step: RouteStep, instruction: VisualInstruction) {
+        maneuverView.step = step
+        instructionLabel.instruction = instruction.primaryTextComponents
     }
     
 }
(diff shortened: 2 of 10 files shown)


bsudekum:
@JThramer feel free to break your todo list out into 3 different PRs.

JThramer:
Turned out to be unnecessary @bsudekum. I wasn't able to make as much of an impact on `NextBannerView` as I would have liked, due to there being a bunch of logic rules around whether-or-not the Next banner should show (are the "lanes" showing? etc)... We may want to consider breaking everything in the "top" out into it's own stack-view that could make those determinations itself, but that seems like a big lift for this PR, especially considering #981 is still in-flight.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -71,6 +71,17 @@ open class BaseInstructionsBannerView: UIControl {
         delegate?.didTapInstructionsBanner(self)
     }
     
+    func update(progress: RouteLegProgress, location: CLLocation, secondsRemaining: TimeInterval) {
```
1ec5: `location` and `secondsRemaining` are unused, so  `setRouteProgress(_:)` would suffice as the method name.

MapboxNavigation/NextBannerView.swift
```diff
@@ -70,4 +71,9 @@ open class NextBannerView: UIView {
         instructionLabel.rightAnchor.constraint(equalTo: rightAnchor, constant: -16).isActive = true
     }
     
+    func update(step: RouteStep, instruction: VisualInstruction) {
+        maneuverView.step = step
+        instructionLabel.instruction = instruction.primaryTextComponents
```
1ec5: An `update(list:of:properties:to:update:)` method is still a pretty raw wrapper on the previous approach of setting all the properties directly and individually. This is an opportunity for us to think about when RouteMapViewController updates these properties and why. That’ll naturally lead us to a more descriptive name for the method, so we don’t inadvertently call or neglect to call it in the future.

In this case, I noticed that InstructionsBannerView has separate methods for updating the step (as part of setting the route progress) and setting the instruction; why not follow the same pattern for NextBannerView?

bsudekum: Agreed, noting that `maneuverView.step ` is used for deciding upon the which maneuver arrow to draw.

bsudekum: Ideally, a developer could just give us a RouteProgress, and we'd be able to draw the correct UI for them.

1ec5: Good point. How about making as many of these classes as possible maintain a single RouteProgress-typed property? Each setter would take care of extracting the relevant information from that RouteProgress object. We could create a new protocol for classes that track state via a RouteProgress-typed property.

JThramer:
Status update: Yesterday it was agreed among members of @mapbox/navigation-ios that the views associated with the "Top banner" should be extracted into their own containing class to take this concept even further. I am doing this by creating a `TopBannerViewController` and associated XIB.

MapboxNavigation/TopBannerViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    // MARK: - Model Updating
+    func update(progress: RouteProgress, location: CLLocation, secondsRemaining: TimeInterval) {
+        //update banner
+        instructions.update(progress: progress.currentLegProgress, location: location, secondsRemaining: secondsRemaining)
+        
+        //update lanes
+        if let upComingStep = progress.currentLegProgress?.upComingStep, !progress.currentLegProgress.userHasArrivedAtWaypoint {
+            updateLaneViews(step: upComingStep, durationRemaining: progress.currentLegProgress.currentStepProgress.durationRemaining)
```
JThramer: Investigate why one view uses the `secondsRemaining` argument and another view seems to pull the same value(?) out of `RouteProgress`.

JThramer:
This PR has been blocked by a major architectural change (#981) introduced upstream. I have started work on resolving this blocker, this effort is being tracked by #1062.

JThramer:
This is ultra-stale, and good progress was made in-parallel with the issue that this PR was meant to address. Going to close it.

Issue #818, opened 2017-11-09
Stick to chosen alternative route after rerouting


1ec5:
Upon rerouting, RouteController should choose the route (possibly an alternative route) that’s most similar to the route the user began with.

Let’s say the user is going from San José, California, to Houston, Texas. The Directions API comes up with two routes, one using I-40 and the other using I-10. The user selects I-10 and starts the route. Before leaving San José, they take a small detour to fill up the tank. That’s enough to trigger a reroute, and RouteController dutifully reselects the default, fastest route, discarding any alternative routes that may’ve been requested:

https://github.com/mapbox/mapbox-navigation-ios/blob/64abd85990b62db04cdb2e240d7eab1fb50efba4/MapboxCoreNavigation/RouteController.swift#L720-L722

In Bakersfield, the navigation SDK ends up telling the user to turn onto I-40, despite the user’s original selection.

To determine route similarity, we could compare all the coordinates and sort by the number of shared coordinates. However, that brute-force approach would perform very poorly for longer routes, especially given that multiple rerouting attempts can occur in rapid succession. Instead, we can use the route legs’ descriptions (summaries) as proxies for similarity. Each description names the two most prominent roads along the route leg:

* If two routes between the same waypoints largely travel over the same roads, then it doesn’t really matter which route we choose automatically.
* If the routes travel over different roads, then we can sort the routes by the [Damerau–Levenshtein edit distance](https://github.com/modocache/MDCDamerauLevenshtein/) from the initial route leg description to the new route leg descriptions.

Note that multifaceted routing – shortest route versus fastest versus cheapest – is out of scope for this issue. This issue is only concerned with correctly persisting the user’s selection across rerouting attempts.

/cc @mapbox/navigation-ios

hervouinc:
Up !

hervouinc:
Hi @1ec5, could you please set the milestone for this issue ?

PR #995, opened 2018-01-09
Stick to chosen route

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/818

This PR seeks to stick to the route the user has selected when rerouting. This occurs by comparing the previous route summary to the new route(s) summary. 

/cc @mapbox/navigation-ios @mapbox/navigation-android

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -749,7 +749,7 @@ extension RouteController: CLLocationManagerDelegate {
                 return completion(nil, nil)
             }
 
-            if let route = self?.selectRouteBySimilarity(from: routes) {
+            if let route = self?.mostSimilarRoute(in: routes) {
                 return completion(route, error)
             } else if let route = routes.first {
                 return completion(route, error)
@@ -759,7 +759,7 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
     
-    func selectRouteBySimilarity(from routes: [Route]) -> Route? {
+    func mostSimilarRoute(in routes: [Route]) -> Route? {
         return routes.min { (left, right) -> Bool in
             let leftDistance = left.description.minimumEditDistance(to: routeProgress.route.description)
             let rightDistance = right.description.minimumEditDistance(to: routeProgress.route.description)
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -744,11 +744,27 @@ extension RouteController: CLLocationManagerDelegate {
             if let error = error {
                 return completion(nil, error)
             }
-            guard let route = routes?.first else {
+            
+            guard let routes = routes else {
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/String.swift
```diff
@@ -50,4 +50,37 @@ extension String {
     var containsDecimalDigit: Bool {
         return self.rangeOfCharacter(from: CharacterSet.decimalDigits) != nil
     }
+    
+    // Adapted from https://github.com/raywenderlich/swift-algorithm-club/blob/master/Minimum%20Edit%20Distance/MinimumEditDistance.playground/Contents.swift
```
1ec5: This code is substantially similar to that playground code. We’ll need to include [this license](https://github.com/raywenderlich/swift-algorithm-club/blob/master/LICENSE.txt) either right here in this file or in LICENSE.md. That isn’t a problem at all, but here are two alternatives anyways:

* Write the code from scratch according to the well-documented [Wagner–Fischer algorithm](https://en.wikipedia.org/wiki/Wagner–Fischer_algorithm).
* Depend on [this library](https://cocoapods.org/pods/MDCDamerauLevenshtein) instead, because it provides a [Damerau–Levenshtein distance](https://en.wikipedia.org/wiki/Damerau–Levenshtein_distance) function. Unlike a basic Levenshtein distance, Damerau–Levenshtein treats a transposition (like swapping two words) as a single edit instead of a deletion and addition. We’ve had good results with this approach in a previous application.

bsudekum: I'd really like to steer clear of adding another dependency. I'm 👍  on adding this to a note LICENSE.md.

bsudekum:
Noting, this is a crutch until this feature is implemented on the Mapbox Directions API.

MapboxCoreNavigation/String.swift
1ec5: FYI, Damerau–Levenshtein is quite concise, so if adding a dependency is a problem, it shouldn’t be too difficult to port [this code](https://github.com/modocache/MDCDamerauLevenshtein/blob/91eae31ef9bff5396371471518e554c6020a6913/MDCDamerauLevenshtein/Algorithms/MDCDamerauLevenshteinDistance.m#L16) to Swift and add a similar MIT license block to the codebase.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -744,11 +744,26 @@ extension RouteController: CLLocationManagerDelegate {
...
+                return completion(route, error)
+            } else if let route = routes.first {
+                return completion(route, error)
+            } else {
+                return completion(nil, nil)
+            }
+        }
+    }
+    
+    func selectRouteBySimilarity(from routes: [Route]) -> Route? {
```
1ec5: This method returns a value without any side effects, so start the name with a noun, not a verb: `mostSimilarRoute(in:)`.

PR #1001, opened 2018-01-10
Add nav examples to guides

bsudekum:
Pulls examples from https://github.com/mapbox/navigation-ios-examples and puts them in the guides folder. This will get published to 

image

The update script could be added the documentation script if we'd like or it can be ran independently. 

/cc @mapbox/navigation-ios


bsudekum:
Let's rename `Guides` to `Examples`.

docs/guides/Storyboards.md
```diff
@@ -1,23 +0,0 @@
-# Setting up Mapbox Navigation in a storyboard
```
1ec5: Looks like this PR deletes the storyboard guide, yet it remains in the table of contents under “Examples”. We should keep the storyboard guide as a guide – that is, under docs/guides/ and the “Guides” section, separate from the examples.

1ec5:
There’s no reason why we can’t have both guides and examples:

```
--documentation=docs/{guides,examples}/*.md
```

docs/jazzy.yml
```diff
@@ -11,6 +11,14 @@ skip_undocumented: Yes
...
 custom_categories:
+  - name: Examples
+    children:
+      - Storyboards
+      - Advanced
+      - Basic
+      - Embedded-Navigation
+      - Styled-UI-Elements
+      - Waypoint-Arrival-Screen
   - name: Guides
```
1ec5: Probably makes sense to move “Guides” up above “Examples”. Eventually we’d add more conceptual content here that could serve as a prerequisite for the examples.

docs/jazzy.yml
```diff
@@ -11,6 +11,14 @@ skip_undocumented: Yes
...
 
 custom_categories:
+  - name: Examples
+    children:
+      - Storyboards
+      - Advanced
+      - Basic
+      - Embedded-Navigation
+      - Styled-UI-Elements
+      - Waypoint-Arrival-Screen
```
1ec5: This looks really weird. The script should replace hyphens with spaces in the file names.

docs/examples/StyledUIElements.md
```diff
@@ -0,0 +1,138 @@
+
+_Source available [here](https://github.com/mapbox/navigation-ios-examples/blob/master/Navigation%20Examples/Examples/StyledUIElements.swift)_
```
1ec5: Both Styled-UI-Elements.md and StyledUIElements.md?

docs/jazzy.yml
```diff
@@ -11,6 +11,14 @@ skip_undocumented: Yes
 hide_documentation_coverage: Yes
 
 custom_categories:
+  - name: Examples
+    children:
+      - Storyboards
```
1ec5: “Storyboards” is already listed under “Guides”.

docs/jazzy.yml
```diff
@@ -11,6 +11,14 @@ skip_undocumented: Yes
...
+      - Advanced
+      - Basic
```
1ec5: Rearrange these examples in order of difficulty, so that “Basic” comes before “Advanced”.

bsudekum:
@1ec5 fixed

PR #1002, opened 2018-01-11
Update French Strings

bsudekum:
I think I did this right, simply ran `tx pull`.

/cc @mapbox/navigation-ios


frederoni:
You need to convert the string files to UTF-8 after pulling from Transifex so git and GH can diff them. Just run the `convert_string_files.sh` file from the script folder.

1ec5:
Here’s some documentation on updating the localizations the next time around:

https://github.com/mapbox/mapbox-navigation-ios/blob/master/CONTRIBUTING.md#adding-or-updating-a-localization

Since this PR only updated existing translations, the second set of steps is all that matters.

PR #1005, opened 2018-01-11
Change VoiceController delegate signature

bsudekum:
@1ec5 pointed out it might be necessary to also give the developer access to the current route progress when this delegate method is emitted. 

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -141,15 +141,12 @@ public class PollyVoiceController: RouteVoiceController {
     }
     
     public override func speak(_ instruction: SpokenInstruction) {
-        guard let routeProgress = routeProgress else {
-            assert(false, "`routeProgress should not be nil`")
-            return
-        }
+        assert(routeProgress != nil, "routeProgress should not be nil.")
         
         if let audioPlayer = audioPlayer, audioPlayer.isPlaying, let lastSpokenInstruction = lastSpokenInstruction {
             voiceControllerDelegate?.voiceController?(self, didInterrupt: lastSpokenInstruction, with: instruction)
         }
-        let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction, routeProgress: routeProgress) ?? instruction
+        let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction, routeProgress: routeProgress!) ?? instruction
         pollyTask?.cancel()
         audioPlayer?.stop()
         lastSpokenInstruction = modifiedInstruction
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -141,7 +141,10 @@ public class PollyVoiceController: RouteVoiceController {
     }
     
     public override func speak(_ instruction: SpokenInstruction) {
-        guard let routeProgress = routeProgress else { return }
+        guard let routeProgress = routeProgress else {
+            assert(false, "`routeProgress should not be nil`")
+            return
+        }
         
         if let audioPlayer = audioPlayer, audioPlayer.isPlaying, let lastSpokenInstruction = lastSpokenInstruction {
             voiceControllerDelegate?.voiceController?(self, didInterrupt: lastSpokenInstruction, with: instruction)
--- a/MapboxNavigation/PollyVoiceController.swift
+++ b/MapboxNavigation/PollyVoiceController.swift
@@ -141,10 +141,12 @@ public class PollyVoiceController: RouteVoiceController {
     }
     
     public override func speak(_ instruction: SpokenInstruction) {
+        guard let routeProgress = routeProgress else { return }
+        
         if let audioPlayer = audioPlayer, audioPlayer.isPlaying, let lastSpokenInstruction = lastSpokenInstruction {
             voiceControllerDelegate?.voiceController?(self, didInterrupt: lastSpokenInstruction, with: instruction)
         }
-        let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction) ?? instruction
+        let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction, routeProgress: routeProgress) ?? instruction
         pollyTask?.cancel()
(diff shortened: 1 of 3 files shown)


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -203,8 +203,10 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     @objc open func didPassSpokenInstructionPoint(notification: NSNotification) {
         guard !NavigationSettings.shared.voiceMuted else { return }
         
-        let routeProgress = notification.userInfo![RouteControllerDidPassSpokenInstructionPointRouteProgressKey] as! RouteProgress
-        legProgress = routeProgress.currentLegProgress
+        routeProgress = notification.userInfo![RouteControllerDidPassSpokenInstructionPointRouteProgressKey] as? RouteProgress
```
1ec5: If `RouteControllerDidPassSpokenInstructionPointRouteProgressKey` exists and isn’t a RouteProgress, that’s an internal consistency error. Let’s assert that it’s a RouteProgress (or go back to using `!`).

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -203,8 +203,10 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     @objc open func didPassSpokenInstructionPoint(notification: NSNotification) {
         guard !NavigationSettings.shared.voiceMuted else { return }
         
-        let routeProgress = notification.userInfo![RouteControllerDidPassSpokenInstructionPointRouteProgressKey] as! RouteProgress
-        legProgress = routeProgress.currentLegProgress
+        routeProgress = (notification.userInfo![RouteControllerDidPassSpokenInstructionPointRouteProgressKey] as! RouteProgress)
+        
+        guard let routeProgress = routeProgress else { return }
```
1ec5: This guard statement is no longer necessary.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -203,8 +203,10 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     @objc open func didPassSpokenInstructionPoint(notification: NSNotification) {
         guard !NavigationSettings.shared.voiceMuted else { return }
         
-        let routeProgress = notification.userInfo![RouteControllerDidPassSpokenInstructionPointRouteProgressKey] as! RouteProgress
-        legProgress = routeProgress.currentLegProgress
+        routeProgress = (notification.userInfo![RouteControllerDidPassSpokenInstructionPointRouteProgressKey] as! RouteProgress)
```
1ec5: Nit: extraneous parentheses.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -216,6 +218,8 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
      - parameter instruction: The instruction to read aloud.
      */
     open func speak(_ instruction: SpokenInstruction) {
+        guard let routeProgress = routeProgress else { return }
```
1ec5: Instead of failing silently, assert that there’s a route progress. Better to fail early knowing why than to fail silently without any clues leading to the root cause.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -282,7 +286,8 @@ public protocol VoiceControllerDelegate {
      
      - parameter voiceController: The voice controller that experienced the interruption.
      - parameter instruction: The spoken instruction that will be said.
+     - parameter routeProgress: The current `RouteProgress` when the instruction was generated.
```
1ec5: The instruction was generated on the server at request time. This is the route progress just before the instruction is scheduled to be spoken.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -282,7 +286,8 @@ public protocol VoiceControllerDelegate {
      
      - parameter voiceController: The voice controller that experienced the interruption.
```
1ec5: Copy-pasta: there isn’t necessarily an interruption here.

bsudekum:
@1ec5 @JThramer updated.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -141,10 +141,15 @@ public class PollyVoiceController: RouteVoiceController {
     }
     
     public override func speak(_ instruction: SpokenInstruction) {
+        guard let routeProgress = routeProgress else {
+            assert(false, "`routeProgress should not be nil`")
```
JThramer: assert(routeProgress != nil, "routeProgress should not be nil.")

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -216,6 +221,11 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
      - parameter instruction: The instruction to read aloud.
      */
     open func speak(_ instruction: SpokenInstruction) {
+        guard let routeProgress = routeProgress else {
```
JThramer: here too.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -203,8 +203,13 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     @objc open func didPassSpokenInstructionPoint(notification: NSNotification) {
         guard !NavigationSettings.shared.voiceMuted else { return }
         
-        let routeProgress = notification.userInfo![RouteControllerDidPassSpokenInstructionPointRouteProgressKey] as! RouteProgress
-        legProgress = routeProgress.currentLegProgress
+        routeProgress = notification.userInfo![RouteControllerDidPassSpokenInstructionPointRouteProgressKey] as? RouteProgress
+        
+        guard let routeProgress = routeProgress else {
```
JThramer: and here

MapboxNavigation/PollyVoiceController.swift
bsudekum: I mean, then we have to `!` all instances of routeProgress below

Issue #1003, opened 2018-01-11
Warn developers if they are using an outdated version


bsudekum:
Similar to the Mapbox Maps SDK, we should warn developers who are not on the latest version.

/cc @mapbox/navigation-ios

1ec5:
It’d be nice to pair #1007 with a reference in the readme to [this option](https://github.com/Carthage/Carthage#optionally-add-build-phase-to-warn-about-outdated-dependencies) for Carthage users. That’ll add a build warning, which is still gentle but won’t get lost in the noise of the console.

PR #1007, opened 2018-01-12
print warning when developer is using older version

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1003

This prints a gentle reminder to the console when a developer using an older version of the sdk.
This will be a no-op until the documentation is updated. 

/cc @mapbox/navigation-ios


MapboxCoreNavigation/RouteController.swift
```diff
@@ -720,6 +722,23 @@ extension RouteController: CLLocationManagerDelegate {
             strongSelf.delegate?.routeController?(strongSelf, didRerouteAlong: route)
         }
     }
+    
+    func checkForUpdates() {
+        #if TARGET_IPHONE_SIMULATOR
+            guard let latestVersion = String(describing: Bundle(for: RouteController.self).object(forInfoDictionaryKey: "CFBundleShortVersionString")) else { return }
+            URLSession.shared.dataTask(with: URL(string: "https://www.mapbox.com/mapbox-navigation-ios/navigation/latest_version")!, completionHandler: { (data, response, error) in
```
1ec5: We should put something at this URL before releasing with this PR.

scripts/document.sh
```diff
@@ -50,4 +50,7 @@ REPLACE_REGEXP+='s/BRANDLESS_DOCSET_TITLE/Navigation SDK for iOS $1/, '
 REPLACE_REGEXP+="s/<span class=\"kt\">(${DIRECTIONS_SYMBOLS})<\/span>/<span class=\"kt\"><a href=\"${BASE_URL//\//\\/}\/directions\/${DIRECTIONS_VERSION}\/Classes\/\$1.html\">\$1<\/a><\/span>/, "
 
 find ${OUTPUT} -name *.html -exec \
-    perl -pi -e "$REPLACE_REGEXP" {} \;
\ No newline at end of file
+    perl -pi -e "$REPLACE_REGEXP" {} \;
+
+
+echo $SHORT_VERSION > ./latest_version
```
1ec5: This puts the file in the current working directory, which isn’t the same as the output directory (`$OUTPUT`).

MapboxCoreNavigation/RouteController.swift
```diff
@@ -720,6 +722,23 @@ extension RouteController: CLLocationManagerDelegate {
...
+                if let _ = error { return }
+                guard let httpResponse = response as? HTTPURLResponse, httpResponse.statusCode == 200 else { return }
+                
+                guard let data = data, let currentVersion = String(data: data, encoding: .utf8) else { return }
+                
+                if latestVersion != currentVersion {
```
1ec5: (Note that this string comparison won’t correctly handle any prereleases we do.)

bsudekum: Yeah I think thats fine.

MapboxCoreNavigation/RouteController.swift
bsudekum: Added.

bsudekum: https://www.mapbox.com/mapbox-navigation-ios/navigation/latest_version

MapboxCoreNavigation/RouteController.swift
```diff
@@ -720,6 +722,24 @@ extension RouteController: CLLocationManagerDelegate {
             strongSelf.delegate?.routeController?(strongSelf, didRerouteAlong: route)
         }
     }
+    
+    func checkForUpdates() {
+        #if TARGET_IPHONE_SIMULATOR
+        guard let version = Bundle(for: RouteController.self).object(forInfoDictionaryKey: "CFBundleShortVersionString") else { return }
+        let latestVersion = String(describing: version)
```
1ec5: Nit: inconsistent indentation.

scripts/document.sh
```diff
@@ -53,3 +53,6 @@ REPLACE_REGEXP+="s/<span class=\"kt\">(${DIRECTIONS_SYMBOLS})<\/span>/<span clas
 
 find ${OUTPUT} -name *.html -exec \
     perl -pi -e "$REPLACE_REGEXP" {} \;
+
+
+echo $SHORT_VERSION > $output/latest_version
```
1ec5: `$output` ≠ `$OUTPUT`

PR #1012, opened 2018-01-12
Allow any UIView to be used as a custom user puck

frederoni:
This change provides a default implementation for the user puck’s
rotation, perspective and scale transformations, making it as easy as:
```swift
mapView.userCourseView = UIImageView(image: UIImage(named: "car_icon"))
```
to use a custom user puck.

car

@mapbox/navigation-ios 👀


bsudekum:
We should also add this as an example to https://github.com/mapbox/navigation-ios-examples

Issue #1008, opened 2018-01-12
Consolidate RouteController arrival logic


1ec5:
`RouteController.monitorStepProgress(_:)` and `RouteController.advanceStepIndex(to:)` can result in advancing to the next step and `RouteControllerDelegate.routeController(_:shouldIncrementLegWhenArrivingAtWaypoint:)` to be called:

https://github.com/mapbox/mapbox-navigation-ios/blob/0eb0f675f5b763e61c5fe276d3b35c41635ccc07/MapboxCoreNavigation/RouteController.swift#L816-L822
https://github.com/mapbox/mapbox-navigation-ios/blob/0eb0f675f5b763e61c5fe276d3b35c41635ccc07/MapboxCoreNavigation/RouteController.swift#L853-L857

Unless I’m mistaken, this code causes `RouteControllerDelegate.routeController(_:didArriveAt:)` to be called on the next location update via `RouteController.progressDidChange(_:)`:

https://github.com/mapbox/mapbox-navigation-ios/blob/0eb0f675f5b763e61c5fe276d3b35c41635ccc07/MapboxCoreNavigation/RouteController.swift#L574-L581
https://github.com/mapbox/mapbox-navigation-ios/blob/0eb0f675f5b763e61c5fe276d3b35c41635ccc07/MapboxCoreNavigation/RouteController.swift#L461-L469

It’s confusing that one arrival delegate method gets called for a different location update than another arrival delegate method. Let’s consolidate the arrival logic in one of these methods so that all the arrival-related functionality can happen in lockstep.

/cc @bsudekum @JThramer

PR #1010, opened 2018-01-12
Omit needless words from delegate method name

1ec5:
Renamed `RouteControllerDelegate.routeController(_:shouldIncrementLegWhenArrivingAtWaypoint:)` to `RouteControllerDelegate.routeController(_:shouldAdvanceToNextLegWhenArrivingAt:)` to conform to Swift naming guidelines. (The Objective-C signature has changed to `-routeController:shouldAdvanceToNextLegWhenArrivingAtWaypoint:`.)

Fixes the easy part of #1009.

/cc @bsudekum @JThramer

Issue #1009, opened 2018-01-12
Remove RouteControllerDelegate.routeController(_:shouldIncrementLegWhenArrivingAtWaypoint:)


1ec5:
`RouteControllerDelegate.routeController(_:shouldIncrementLegWhenArrivingAtWaypoint:)` is an unnecessarily unwieldy method name, and I think the method itself is nearly redundant:

* A Swift method argument’s label [shouldn’t end with a type](https://swift.org/documentation/api-design-guidelines/#omit-needless-words). It would be `shouldIncrementLegWhenArrivingAt:`, not `shouldIncrementLegWhenArrivingAtWaypoint:`. The word “waypoint” is only required in the Objective-C name.
* It’s the leg index, not the leg, that’s incremented. But that’s an implementation detail of RouteProgress. Instead, the method name should emphasize the result: `shouldAdvanceToNextLegWhenArrivingAt:`.
* There’s already an arrival delegate method, `RouteControllerDelegate.routeController(_:didArriveAt:)`, and it doesn’t have a return value. Why not have it return a Boolean and remove `RouteControllerDelegate.routeController(_:shouldIncrementLegWhenArrivingAtWaypoint:)` outright?

I attempted to delete `RouteControllerDelegate.routeController(_:shouldIncrementLegWhenArrivingAtWaypoint:)` in favor of a return value on `RouteControllerDelegate.routeController(_:didArriveAt:)` but quickly ran into #1008.

/cc @mapbox/navigation-ios

PR #1013, opened 2018-01-12
Refactor arrival

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1008

This cleans up the arrival logic into a single function and also exposes a TimeInterval that the developer can control to fine to when an arrival is triggered.

todo:

- [ ] make sure the telem arrival event is still triggered.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -455,19 +455,6 @@ open class RouteController: NSObject {
 
 extension RouteController {
     @objc func progressDidChange(notification: NSNotification) {
-        if let progress = notification.userInfo?[RouteControllerProgressDidChangeNotificationProgressKey] as? RouteProgress {
-            let currentDestinationWaypoint = progress.currentLeg.destination
-
-            if progress.currentLegProgress.userHasArrivedAtWaypoint, sessionState.arrivalTimestamp == nil, currentDestinationWaypoint != previousArrivalWaypoint  {
-                delegate?.routeController?(self, didArriveAt: currentDestinationWaypoint)
-                previousArrivalWaypoint = currentDestinationWaypoint
-                
-                if let lastRouteWaypoint = progress.route.legs.last?.destination, lastRouteWaypoint == currentDestinationWaypoint {
-                    sessionState.arrivalTimestamp = Date()
-                    sendArriveEvent()
-                }
-            }
-        }
         if sessionState.departureTimestamp == nil {
             sessionState.departureTimestamp = Date()
             sendDepartEvent()
@@ -482,9 +469,7 @@ extension RouteController {
 
     @objc func willReroute(notification: NSNotification) {
         _ = enqueueRerouteEvent()
-        
     }
-    
     
     @objc func didReroute(notification: NSNotification) {
         if let _ = notification.userInfo?[RouteControllerDidFindFasterRouteKey] as? Bool {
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -455,19 +455,6 @@ open class RouteController: NSObject {
...
-        if let progress = notification.userInfo?[RouteControllerProgressDidChangeNotificationProgressKey] as? RouteProgress {
-            let currentDestinationWaypoint = progress.currentLeg.destination
-
-            if progress.currentLegProgress.userHasArrivedAtWaypoint, sessionState.arrivalTimestamp == nil, currentDestinationWaypoint != previousArrivalWaypoint  {
-                delegate?.routeController?(self, didArriveAt: currentDestinationWaypoint)
-                previousArrivalWaypoint = currentDestinationWaypoint
-                
-                if let lastRouteWaypoint = progress.route.legs.last?.destination, lastRouteWaypoint == currentDestinationWaypoint {
-                    sessionState.arrivalTimestamp = Date()
-                    sendArriveEvent()
```
1ec5: Should we be concerned that `sendArriveEvent()` and `sendDepartEvent()` are being called from different methods now?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -598,6 +575,32 @@ extension RouteController: CLLocationManagerDelegate {
         checkForFasterRoute(from: location)
     }
     
+    func monitorLegProgress(_ location: CLLocation) {
```
1ec5: We need to remove the word “monitor” from these methods. It means nothing and leads to the methods becoming kitchen sinks, because everything is a result of “monitoring”.

For now, `monitorStepProgress(_:)` should become `updateRouteStepProgress(for:)`, and this method should become `updateRouteLegProgress(for:)`. At some point, we should try to break up each method into more granular methods that can be unit tested more easily.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -598,6 +575,32 @@ extension RouteController: CLLocationManagerDelegate {
...
+        previousArrivalWaypoint = currentDestinationWaypoint
+        delegate?.routeController?(self, didArriveAt: routeProgress.currentLeg.destination)
+            
+        if let lastRouteWaypoint = routeProgress.route.legs.last?.destination, lastRouteWaypoint == currentDestinationWaypoint {
+            sessionState.arrivalTimestamp = Date()
+            sendArriveEvent()
+        }
+        
+        if !routeProgress.isFinalLeg,
+            (delegate?.routeController?(self, shouldIncrementLegWhenArrivingAtWaypoint: routeProgress.currentLeg.destination) ?? true) {
```
1ec5: I’d still like for this delegate method to go away (#1009), either as part of this PR or as part of #1010. I was thinking that `routeController(_:didArriveAt:)` would return a Boolean, which would be stored in a local variable above and fed into this check. Do you see any problems with that approach?

MapboxCoreNavigation/RouteController.swift
bsudekum: Moved in https://github.com/mapbox/mapbox-navigation-ios/pull/1013/commits/8ba1cfb0875e8d04c88b14a78e856afa52b38976

MapboxCoreNavigation/RouteController.swift
bsudekum: Changed in https://github.com/mapbox/mapbox-navigation-ios/pull/1013/commits/24a2f6f6ce86f14877c164f296e35286cc0f713b

bsudekum:
@1ec5 this is ready for review

MapboxCoreNavigation/RouteController.swift
1ec5: OK, let’s do this in #1010 then.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -600,6 +583,27 @@ extension RouteController: CLLocationManagerDelegate {
         checkForFasterRoute(from: location)
     }
     
+    func updateRouteLegProgress(for location: CLLocation) {
+        guard routeProgress.currentLegProgress.upComingStep?.maneuverType == .arrive else { return }
+        
+        let polyline = Polyline(routeProgress.currentLegProgress.currentStep.coordinates!)
+        let userSnapToStepDistanceFromManeuver = polyline.distance(from: location.coordinate)
+        let secondsToEndOfStep = userSnapToStepDistanceFromManeuver / location.speed
+        guard secondsToEndOfStep < RouteControllerDurationRemainingWaypointArrival else { return }
```
1ec5: This guard statement should become an if statement, and everything below should be factored out into a helper method. Use guard statements only for edge cases. As it is, you have to read carefully to spot this line and know that the stuff below here only runs when arriving.

PR #1021, opened 2018-01-16
Filter out bad locations

bsudekum:
Instead of just grabbing the last location from the location manager, let's first throw out any unqualified locations and then grab the last location from locations.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -530,18 +530,17 @@ extension RouteController: CLLocationManagerDelegate {
     }
 
     @objc public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
-        guard let location = locations.last else {
-            return
-        }
+        let filteredLocation = locations.filter { $0.isQualified }
+        
+        guard let location = filteredLocation.last else { return }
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -530,21 +530,21 @@ extension RouteController: CLLocationManagerDelegate {
...
-            return
-        }
-        self.rawLocation = location
-
-        sessionState.pastLocations.push(location)
-
-        if location.isQualified {
+        let filteredLocations = locations.filter { $0.isQualified }
+        
+        if !filteredLocations.isEmpty, hasFoundOneQualifiedLocation == false {
```
JThramer: Nit: this property name is a mouthful, how about `hasQualifiedLocation`?

bsudekum: `hasQualifiedLocation` implies something else. This variable is used really only at the beginning of a session when locations are generally bad. We don't emit that we discarded a location until we get at least one good location.

1ec5: Nit: `== false` is unnecessarily explicit in Swift. The `!` operator would suffice.

PR #1029, opened 2018-01-17
Minor step advancement and waypoint arrival fixes

bsudekum:
This fixes a few issues that came out of https://github.com/mapbox/mapbox-navigation-ios/pull/1013/. I will comment inline on each change.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -585,14 +585,10 @@ extension RouteController: CLLocationManagerDelegate {
     }
     
     func updateRouteLegProgress(for location: CLLocation) {
-        guard routeProgress.currentLegProgress.upComingStep?.maneuverType == .arrive else { return }
-        
-        let polyline = Polyline(routeProgress.currentLegProgress.currentStep.coordinates!)
-        let userSnapToStepDistanceFromManeuver = polyline.distance(from: location.coordinate)
-        let secondsToEndOfStep = userSnapToStepDistanceFromManeuver / location.speed
         let currentDestination = routeProgress.currentLeg.destination
+        let legDurationRemaining = routeProgress.currentLegProgress.durationRemaining
         
-        if secondsToEndOfStep < RouteControllerDurationRemainingWaypointArrival, currentDestination != previousArrivalWaypoint {
+        if legDurationRemaining < RouteControllerDurationRemainingWaypointArrival, currentDestination != previousArrivalWaypoint {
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -839,8 +835,7 @@ extension RouteController: CLLocationManagerDelegate {
         let lastKnownUserAbsoluteDistance = routeProgress.currentLegProgress.currentStepProgress.userDistanceToManeuverLocation
 
         if userSnapToStepDistanceFromManeuver <= RouteControllerManeuverZoneRadius &&
-            courseMatchesManeuverFinalHeading ||
-            (userAbsoluteDistance > lastKnownUserAbsoluteDistance && lastKnownUserAbsoluteDistance > RouteControllerManeuverZoneRadius) {
+            (courseMatchesManeuverFinalHeading || (userAbsoluteDistance > lastKnownUserAbsoluteDistance && lastKnownUserAbsoluteDistance > RouteControllerManeuverZoneRadius)) {
```
bsudekum: Take a look at what the code was before: https://github.com/mapbox/mapbox-navigation-ios/pull/1013/files#diff-1a2149692b3d7c58b6a0392120d6011aL836

With the change in https://github.com/mapbox/mapbox-navigation-ios/pull/1013/files#diff-1a2149692b3d7c58b6a0392120d6011aL836, we we're advance the step unnecessarily because it would evaluate to true too often.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -585,14 +585,10 @@ extension RouteController: CLLocationManagerDelegate {
     }
     
     func updateRouteLegProgress(for location: CLLocation) {
-        guard routeProgress.currentLegProgress.upComingStep?.maneuverType == .arrive else { return }
-        
-        let polyline = Polyline(routeProgress.currentLegProgress.currentStep.coordinates!)
-        let userSnapToStepDistanceFromManeuver = polyline.distance(from: location.coordinate)
-        let secondsToEndOfStep = userSnapToStepDistanceFromManeuver / location.speed
         let currentDestination = routeProgress.currentLeg.destination
+        let legDurationRemaining = routeProgress.currentLegProgress.durationRemaining
```
bsudekum: Slight optimization, we do not need to re-compute the duration remaining for the leg. This is also a little more clear because we should not be comparing step duration when we're interested more in the duration of the leg for knowing when the user arrives.

PR #1031, opened 2018-01-18
Add Danish localization

1ec5:
Added a ~~mostly~~ complete Danish localization from Transifex.

/cc @frederoni


bsudekum:
@1ec5 mind also updating https://github.com/mapbox/mapbox-navigation-ios/blob/master/languages.md?

1ec5:
7a192dd414b94f28325d0449161f442a59e709ea adds Danish voice support and updates the language support table.

MapboxNavigation/PollyVoiceController.swift
```diff
@@ -97,6 +97,8 @@ public class PollyVoiceController: RouteVoiceController {
         }
         
         switch (langCode, countryCode) {
+        case ("da", _):
+            input.voiceId = .naja
```
1ec5: Project-OSRM/osrm-text-instructions#208 adds Danish support for voice instruction generation on the server. Unfortunately, we have our own little version of [dependency hell](https://en.wikipedia.org/wiki/Dependency_hell): if this PR goes in before that the OSRM Text Instructions change makes it into the Directions API, then English sentences will be read in a Danish voice. On the other hand, if the Directions API is updated before this PR lands, then Danish instructions will be read in an English voice. We‘ll need to coordinate both the client-side and server-side changes to avoid suboptimal behavior.

1ec5: @bsudekum, what do you think about splitting out this PollyVoiceController change into a separate PR that we can merge when Project-OSRM/osrm-text-instructions#208 is deployed in the Directions API? (If #617 lands first, then the PR would become moot.)

1ec5:
I’ve split out the Danish voice support as #1041, pending a server-side update. See https://github.com/mapbox/mapbox-navigation-ios/pull/1031#discussion_r162471982 for context.

PR #1032, opened 2018-01-18
Avoid asking pedestrians, cyclists to rate drives

1ec5:
If the user chooses to walk or bike, asking them to rate a “drive” would be illogical. “Trip” is just as understandable to motorists.

/cc @JThramer


MapboxNavigation/Resources/Base.lproj/Navigation.storyboard
```diff
@@ -43,7 +43,7 @@
                                     <userDefinedRuntimeAttribute type="string" keyPath="styleURL__" value="mapbox://styles/mapbox/navigation-guidance-day-v2"/>
                                 </userDefinedRuntimeAttributes>
                             </view>
-                            <button hidden="YES" opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="I8f-iR-MZm" customClass="MBReportButton">
+                            <button hidden="YES" opaque="NO" contentMode="scaleToFill" ambiguous="YES" contentHorizontalAlignment="center" contentVerticalAlignment="center" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="I8f-iR-MZm" customClass="MBReportButton">
```
bsudekum: Probably should fix this `ambiguous` addition.

PR #1034, opened 2018-01-18
Pluralize star accessibility values in stringsdict

### LINKED ISSUE #971: Accessibility hint for star ratings assumes English grammar
The logic for setting the accessibility hint for start ratings assumes English pluralization rules, but many languages have different rules:

https://github.com/mapbox/mapbox-navigation-ios/blob/04f8f79acefb21aa6bd7f463a10b47ab9d5d65eb/MapboxNavigation/RatingControl.swift#L110-L117
https://github.com/mapbox/mapbox-navigation-ios/blob/04f8f79acefb21aa6bd7f463a10b47ab9d5d65eb/MapboxNavigation/RatingControl.swift#L75-L76

Pluralized strings should use .stringsdict files, as outlined [in these contributing instructions](https://github.com/mapbox/mapbox-navigation-ios/blob/0a31b806dcf7fffa90ff02aa7bb5b4ede93f67df/CONTRIBUTING.md#adding-user-facing-text).

/cc @JThramer

1ec5:
Use a stringsdict to correctly pluralize the star rating UI’s accessibility values. A localization that distinguishes one from two from many, such as Russian, will be able to translate these values accurately.

Fixes #971.

/cc @JThramer @frederoni

### DIFF
--- a/MapboxNavigation/Resources/Base.lproj/Localizable.strings
+++ b/MapboxNavigation/Resources/Base.lproj/Localizable.strings
@@ -46,20 +46,14 @@
 /* Format string for a short distance or time less than a minimum threshold; 1 = duration remaining */
 "LESS_THAN" = "<%@";
 
-/* No Rating Set */
-"NO_RATING" = "No rating set.";
-
-/* One Star Set */
-"RATING_1_STAR" = "1 star set.";
-
 /* Rating Reset To Zero Accessability Hint */
 "RATING_ACCESSIBILITY_RESET" = "Tap to reset the rating to zero.";
 
-/* Accessibility Star Label */
-"RATING_ACCESSIBILITY_SET" = "Set %@ star rating";
+/* Format for accessibility label of button for setting a rating; 1 = number of stars */
+"RATING_ACCESSIBILITY_SET" = "Set %ld-star rating";
 
-/* Format for rating stars set */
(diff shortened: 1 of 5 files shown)


MapboxNavigation/Resources/Base.lproj/Localizable.strings
```diff
@@ -46,20 +46,14 @@
...
 
-/* Accessibility Star Label */
-"RATING_ACCESSIBILITY_SET" = "Set %@ star rating";
+/* Format for accessibility label of button for setting a rating; 1 = number of stars */
+"RATING_ACCESSIBILITY_SET" = "Set %ld-star rating";
 
-/* Format for rating stars set */
-"RATING_STARS_FORMAT" = "%li stars set.";
+/* Format for accessibility value of label indicating the existing rating; 1 = number of stars */
+"RATING_STARS_FORMAT" = "%ld star(s) set.";
```
frederoni: Does this work as a fallback until every language provides a stringsdict?

1ec5: Yes. I haven’t checked whether the stringsdict gets consulted in that case. If not, maybe the stringsdict should be base-localized instead of living in the English localization. But that could lead to weirdness if a localization has a fully translated Localizable.strings but hasn’t translated the stringsdict.

PR #1036, opened 2018-01-19
Localize zero-rating accessibility label as separate string

### LINKED ISSUE #1035: Unable to import stringsdict into Transifex
Transifex refuses to import the Localizable.stringsdict file added in #1034 with the following error:

```
Invalid plural types for string: RATING_STARS_FORMAT. Language supports: ['one', 'other'], but found: ['zero', 'one', 'other'] instead.
```

We’ll have to roll back part of #1034 to special-case zero.

/cc @JThramer

1ec5:
Restored a string that was removed in #1034 and removed the `zero` case from an English stringsdict file.

Fixes #1035.

/cc @JThramer

### DIFF
--- a/MapboxNavigation/RatingControl.swift
+++ b/MapboxNavigation/RatingControl.swift
@@ -104,7 +104,14 @@ class RatingControl: UIStackView {
     private func setAccessibility(for button: UIButton, at index: Int) {
         setAccessibilityHint(for: button, at: index)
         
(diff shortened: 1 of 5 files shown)


frederoni:
Looks like `Resources/Base.lproj/Localizable.strings` has been converted to UTF-16. Did you run `convert_string_files`?

1ec5:
Not sure how that happened, since I didn’t run Transifex this time; I just edited the file in Xcode. Anyways, it’s fixed now.

PR #1040, opened 2018-01-19
Update MapboxDirections, Maps, and SDWebImage

### LINKED ISSUE #1039: Compiler errors after pod update to MapboxNavigation 0.12.2
I added a new unrelated pod today and ran pod update which also incrementally updated MapboxNavigation from 0.12.1 to 0.12.2. Cleaned and tried to build but I'm getting a couple of compiler errors in MapboxCoreNavigation.

This is my pod file:
`pod 'Mapbox-iOS-SDK', '~> 3.7'`
`pod 'MapboxNavigation', '~> 0.12.0'`
`pod 'ObjectMapper', '~> 3.1'`

Both errors relate to the ManeuverType.
The first error is 
> Initializer for conditional binding must have Optional type, not 'ManeuverType'

on the line
`if let leftType = left.maneuverType, let rightType = right.maneuverType {`

in file RouteStep.swift.
The second (related no doubt) error is
> Cannot use optional chaining on non-optional value of type 'ManeuverType'

on the line
`"upcomingType": upComingStep?.maneuverType?.description ?? NSNull(),`

in the extension to RouteLegProgress in file MMEEventsManager.swift.

Swift 4
Xcode 9.2

frederoni:
Fixes #1039 

We released a couple of breaking changes MapboxDirections.swift 0.16.0 which haven't been adopted in this library yet but CocoaPods decided pull in 0.16.0 even though we specify `~> 0.15` unlike Carthage.

@mapbox/navigation-ios  👀


Examples/Swift/CustomViewController.swift
```diff
@@ -96,22 +96,22 @@ class CustomViewController: UIViewController, MGLMapViewDelegate, AVSpeechSynthe
...
-            case .slightLeft:
-                self.arrowView.text = "↖️"
-            case .sharpLeft, .left:
-                self.arrowView.text = "⬅️"
-            case .uTurn:
-                self.arrowView.text = "⤵️"
-            default:
-                self.arrowView.text = "⬆️"
-            }
+        let direction = step.maneuverDirection
```
bsudekum: This variable can be removed and you can just switch on `step.maneuverDirection `

Examples/Swift/CustomViewController.swift
```diff
@@ -96,22 +96,21 @@ class CustomViewController: UIViewController, MGLMapViewDelegate, AVSpeechSynthe
...
+            self.arrowView.text = "↗️"
+        case .sharpRight, .right:
+            self.arrowView.text = "➡️"
+        case .slightLeft:
+            self.arrowView.text = "↖️"
+        case .sharpLeft, .left:
+            self.arrowView.text = "⬅️"
+        case .uTurn:
+            self.arrowView.text = "⤵️"
+        default:
```
JThramer: There's a subtle edge-case behavioral change here -- if `step.maneuverDirection = nil`, the arrow text becomes "⬆️". 

That may be desirable, I just thought i should point this out.

1ec5: Yes, I think that’s fine. Our other option is ⏹. 😉

MapboxCoreNavigation.podspec
```diff
@@ -40,7 +40,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxCoreNavigation"
 
-  s.dependency "MapboxDirections.swift", "~> 0.15"
+  s.dependency "MapboxDirections.swift", "~> 0.16.0"
```
1ec5: This was the problem: we were specifying only the minor version, so the tadpole operator would float until the next major version (1.0). By contrast, Carthage [intentionally violates semver](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#version-requirement) by being more strict pre-1.0.

bsudekum: Maybe worth changing now?

Issue #1027, opened 2018-01-17
Hide delimiters in banner when surrounded by loaded shield images


1ec5:
In the turn banner, when an image component loads, hide any adjacent delimiter component to prevent unsightly slashes next to images:

![image](https://user-images.githubusercontent.com/1231218/35028582-a9434114-fb0b-11e7-8e19-90287c9f4e82.png)

Blocked by mapbox/MapboxDirections.swift#224.

/cc @bsudekum @ericrwolfe

1ec5:
Looks like we need some padding or a space between each component too.

akitchen:
MVP is done; there is some tail work remaining in the PR description which I would like to ticket out and address separately. Marked as ready for review.

akitchen:
I removed the review flag and am reopening this to look at why the network queue seems to get backed up on very long routes.

bsudekum:
This has been implemented, thanks @akitchen!

1ec5:
Specifically, this was implemented in #1046

PR #1046, opened 2018-01-19
Hides delimiters in banner when surrounded by loaded shield images

akitchen:
This delivers #1027

Includes some refactoring to avoid use of disk cache under test, and adds some synchronous and asynchronous integrated tests of the desired behavior.

Presentation and domain logic are moved into an `InstructionPresenter` class, which already knows about too much but I think it's an incremental improvement on doing all this work in the view.

TO DO list before merging:

 - [x] Update constants / namespace identifiers per @frederoni 's comments
 - [ ] Get feedback on appropriate visibility level for new classes (?)
 - [x] Implement error handling in line with existing patterns 
 - [x] Remove SDWebImage from Cartfile, delete related extensions, etc.
 - [x] Discover more test cases and update implementation so that delimiter hiding logic is more robust

Tail work:

 - [ ] Backfill some tests as indicated as TODOs int the code
 - [ ] Add background execution handler so that we can allow our download operations to continue when the containing app goes into the background (some investigation needed; want to test on a device in order to evaluate the need)
 - [ ] Force download / cache busting ?
 - [ ] Test in the wild (again)


MapboxNavigation/InstructionPresenter.swift
```diff
@@ -0,0 +1,87 @@
...
+                }
+                string.append(NSAttributedString(string: (joinChar + text).abbreviated(toFit: label.availableBounds(), font: label.font), attributes: attributesForLabel(label)))
+            }
+        }
+
+        return string
+    }
+
+    private func instructionHasDownloadedAllShields() -> Bool {
+        for component in instruction! {
```
frederoni: Nit: this seems to be safe considering the guard in `attributedTextForLabel(_:)` but the linter will complain about unwrapping when we decide to enable it.

MapboxNavigationTests/MapboxNavigationTests.swift
```diff
@@ -24,13 +24,24 @@ class MapboxNavigationTests: FBSnapshotTestCase {
...
         SDImageCache.shared().store(shieldImage, forKey: "I280")
     }
-    
+
+    override func tearDown() {
+        super.tearDown()
+
+        SDImageCache.shared().clearMemory()
+
+        let clearDiskSemaphore = DispatchSemaphore.init(value: 1)
```
frederoni: nit: .init can be dropped.

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -0,0 +1,87 @@
+import UIKit
+import MapboxDirections
+
+class InstructionPresenter {
```
frederoni: 👍  Cool. Perhaps an Instructable protocol could be the next step.

akitchen: I'm into the idea -- let's look out for a unified interface when we write/refactor the next view which might otherwise contain logic for setting up & displaying a complex domain object.

akitchen:
Cool. I addressed the code nit and rebased on master again.

akitchen:
I'm continuing to work to drive SDWebImage out of the codebase.  it is currently isolated in the newly-introduced `ImageRepository` class, @frederoni @bsudekum

MapboxNavigation/ImageCache.swift
```diff
@@ -0,0 +1,143 @@
+import Foundation
+
+class ImageCache: BimodalImageCache {
+
+    let memoryCache: NSCache<NSString, UIImage>
+    let diskCacheURL: URL = {
+        let fileManager = FileManager.default
+        let basePath = fileManager.urls(for: .cachesDirectory, in: .userDomainMask).first!
+        return basePath.appendingPathComponent("com.mapbox.directions.downloadedImages")
```
frederoni: Nit: `directions` is somewhat misleading, perhaps we could use `Bundle.mapboxNavigation.bundleIdentifier+suffix`? ×2

MapboxNavigation/ImageCache.swift
```diff
@@ -0,0 +1,143 @@
...
+            if let completion = completion {
+                DispatchQueue.main.async {
+                    completion()
+                }
+            }
+        }
+
+        if toDisk == true {
+            diskAccessQueue.async {
+                self.createCacheDirIfNeeded()
```
frederoni: Might be mistaken but this looks like a retain cycle.

akitchen:
Looking forward to merging this. There's still a bit of work to do, including:

 - [ ] Update constants / namespace identifiers per @frederoni 's comments
 - [ ] Implement error handling in line with existing patterns 
 - [ ] Remove SDWebImage from Cartfile, delete related extensions, etc.
 - [ ] Test in the wild

But I think it's looking good. If there are other PRs which collide, maybe we can integrate them together before the final merge into master. /cc @bsudekum @JThramer

MapboxNavigation/ImageCache.swift
akitchen: This is helpful, you've answered one of my tail work questions.

MapboxNavigation/ImageCache.swift
akitchen: Yes, there is a transient retain cycle there. While this block is in the queue or executing, self will be retained which ensures it sticks around long enough to complete its work. Once the block exits it is disposed of, breaking the cycle.

Does this reasoning hold water or does the team prefer a weakSelf/strongSelf pattern in these disk access methods?

MapboxNavigation/ImageCache.swift
```diff
@@ -0,0 +1,153 @@
...
+            return nil
+        }
+    }
+
+    private func cacheCostForImage(_ image: UIImage) -> Int {
+        return Int(image.size.height * image.size.width * image.scale * image.scale);
+    }
+
+    private func createCacheDirIfNeeded() {
+        if fileManager!.fileExists(atPath: diskCacheURL.absoluteString) == false {
```
bsudekum: Can we gently unwrap `fileManager`  here?

MapboxNavigation/ImageDownloader.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    // MARK: URLSessionDataDelegate
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive response: URLResponse, completionHandler: @escaping (URLSession.ResponseDisposition) -> Void) {
+        guard let response: HTTPURLResponse = response as? HTTPURLResponse, let url = response.url, let operation: ImageDownload = operations[url] else {
+            completionHandler(.cancel)
+            return
+        }
+        operation.urlSession!(session, dataTask: dataTask, didReceive: response, completionHandler: completionHandler)
```
bsudekum: `operation.urlSession!` can also be unwrapped above in the guard.

MapboxNavigation/ImageCache.swift
akitchen: Done - also refactored to rely less on data from self in closures, addressing @frederoni 's previous observation about the transient retain cycles on the disk access queue.

MapboxNavigation/ImageDownloader.swift
akitchen: Instead of unwrapping, I've opted to ask before telling in the latest commit ( which will ensure that the force unwrap is safe )

akitchen:
I just merged master into this branch as the semantic disconnects were a bit much. /cc @frederoni 

Would love to merge this with a final 👍 and follow on with some tail work as enumerated in the OP

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -71,8 +71,8 @@ open class BaseInstructionsBannerView: UIControl {
         delegate?.didTapInstructionsBanner(self)
     }
     
-    func set(_ instruction: VisualInstruction) {
-        let secondaryInstruction = instruction.secondaryTextComponents
+    func set(_ instruction: VisualInstruction?) {
```
akitchen: This optionality is needed in order to unset the instruction in `StepTableViewCell.prepareForReuse()`

/cc @frederoni

akitchen: also, should we use a different semantic here now that we've wrapped the primary and secondary components in the `VisualInstruction` class? Or does this feel idiomatic enough?

frederoni: Just `set` or `set(_:)` in this case almost feels like a reserved keyword but I don't have a strong opinion about that. Maybe `setInstruction(_:)` or `updateInstruction(_:)` would be more idiomatic?

JThramer: 💯 % agree, @frederoni. `update(progress:)` or `update(instruction:)` would be pretty straightforward.

bsudekum: Let's do this as tail work.

PR #1047, opened 2018-01-19
Update French translations

1ec5:
Updated French translations from Transifex. Thanks to @verdy-p for expanding the set of abbreviations, which will reduce the occurrence of banner truncation when using the SDK in France and Québec.

/cc @bsudekum


verdy-p:
Note that when you were updating this, I was asked to add more abbreviations and fix a few of them, and I was doing that. There are additional abbrevs (mostly in the second list, but a few fixes in the first list for street types).

Also I commented about the case of roman numerals that are difficult to read and can be easily transalted to digits on narrow screens. Also I spoke about abbreviations of month names/week day names which are frequent in dates used in street names.

I expended the list of abbreviation of people titles (like "doctor", or "prince") and of several generic terms ("province", "national"...) which are usually abbreviated.

Note that in the first list, the abbreviations are more narrowed and avoid the abbreviating "." ("R" instead of "R." for "Rue"=/street/). While the second list if for the label after generic terms and uses more typographical conventions.

I also spoke about how to use superscripts (typical in abbreviations, but using special Unicode "modifier letters" does not match very well with the exception of ordinal "o" and "a", but French has many abbreviations with superscripts used for rendering the final letters, when the abbreviation "." comes only after the leading letters). For now we cannot indicate with some markup in Transifex that abbreviations can be partly superscripted (or possibly subscripted for some languages that could use this).

Also I don't think that Transifex if the best place to add long lists of abbreviations, it would be easier to add them on some worksheet online (Google Doc ?), so that each added row can be conveniently reviewed, and rows sorted with an additional comment column or some typology column about their intended use, or other contextual information.

1ec5:
Thank you again for the abbreviations and for the detailed feedback.

> I also spoke about how to use superscripts (typical in abbreviations, but using special Unicode "modifier letters" does not match very well with the exception of ordinal "o" and "a", but French has many abbreviations with superscripts used for rendering the final letters, when the abbreviation "." comes only after the leading letters). For now we cannot indicate with some markup in Transifex that abbreviations can be partly superscripted (or possibly subscripted for some languages that could use this).

I’ve opened #1051 to track displaying superior letters for abbreviations in French.

> Also I don't think that Transifex if the best place to add long lists of abbreviations, it would be easier to add them on some worksheet online (Google Doc ?), so that each added row can be conveniently reviewed, and rows sorted with an additional comment column or some typology column about their intended use, or other contextual information to tune your automatic abbreviator.

I’ve opened #1050 to track managing the translations outside of Transifex.

verdy-p:
Keep your translations in Transifex, it's just the single resource for abbreviations that are not well fitted in Transifex and could better be in an online editable (e.g. Google Doc's spreadsheet)

1ec5:
Agreed, only the abbreviations would move; we’d continue to manage the rest of the translations in Transifex. Thanks for clarifying.

Issue #1049, opened 2018-01-20
Fail CI build if strings files are in UTF-16


1ec5:
This project prefers UTF-8 encoding for strings files, since Git is unable to diff UTF-16-encoded files natively. Since there’s already a handy script for converting the strings files to UTF-8, let’s add a step to one of the Bitrise apps that verifies all the files are in the right encoding.

/cc @frederoni

frederoni:
Fixed by adding a step to Bitrise.
Tested in #1053

PR #1053, opened 2018-01-21
Test UTF-16

frederoni:
Test for #1049


frederoni:
Fixed by adding a step to Bitrise.

PR #1054, opened 2018-01-22
Introduce swiftlint

frederoni:
It's too easy to forget force unwraps, space after a colon etc.
Let's try to enforce a minimal set of rules using [SwiftLint](https://github.com/realm/SwiftLint)

- [ ] Review rules
- [ ] Disable fileprivate->private
- [ ] Set up SwiftLint on CI

@mapbox/navigation-ios 👀


bsudekum:
I'm pretty 👍 when it comes to linters, but I know @1ec5 has some reservations. Does anything change on client?

JThramer:
I'm open to trying it out. Linters can be a little heavy-handed sometimes (IUO's specifically come to mind: they're not _always_ an anti-pattern), but I don't hate them.

.swiftlint.yml
```diff
@@ -0,0 +1,6 @@
+opt_in_rules:
+    - force_unwrapping
```
1ec5: I’m OK with this rule as long as the linter also enforces that we assert or throw an error instead of failing silently. 😉

In practice, we’ll all have to be on the lookout for overuse of `guard let` and `if let` statements and `??` chaining. It’s very easy to get lazy with them, leading to mysterious and undebuggable problems down the line.

1ec5: Please disable the `trailing_whitespace` rule, unless there’s a way for it to ignore blank lines that are indented, such as:

```swift
    let foo: String?
    
    let bar: String?
```

We should use the linter for legitimate issues that affect product behavior or maintainability. I can’t think of a single time we’ve ever nitpicked about trailing whitespace on a PR in this repository. Meanwhile, indenting blank lines is the Xcode default.

If we err on the side of passing, we can keep the barrier to entry low for new contributors while making it more likely that they’ll get a 👍 on their first try.

JThramer: This is the issue that I have with the `force_unwrapping` rule: There are plenty of times in which an IUO can be used appropriately, and it seems tedious to start doing workarounds (`guard let thing = potentialThing else { return assertionFailure("Bad wolf") }`) just to make a linter happy, and as @1ec5 pointed out, workarounds can be prone to introducing very nasty, hard-to-fix bugs.

The linter should be there to help guide us towards clean code. This seems to be a step in the wrong direction.

frederoni: I agree. There a many valid cases for `force_unwrapping`. I guess that's why it's disabled by default.

1ec5:
> ~~Set up SwiftLint on CI~~ As a first step, let's just use the linter manually.

Good call. Non-Mapboxers lack access to the Bitrise logs, so they’d have a difficult time identifying any failure that a linter would trigger on a build.

PR #1055, opened 2018-01-22
Support overriding default distance units

### LINKED ISSUE #589: Option to use metric or customary units
`DistanceFormatter` currently switches between metric and customary units based on the locale (specifically the region):

https://github.com/mapbox/mapbox-navigation-ios/blob/e48d724b7e22522bbffc2a90e9435e4d0e31e6ad/MapboxCoreNavigation/DistanceFormatter.swift#L31-L37

But turn-by-turn applications often offer a user setting to switch between metric and customary units, regardless of the locale. We should expose a top-level option that affects the measurement system used by any `DistanceFormatter` instance associated with a NavigationViewController instance.

/ref https://github.com/mapbox/mapbox-navigation-ios/issues/242#issuecomment-327619678
/cc @frederoni @troyibm

frederoni:
Fixes #589 - Option to use metric or customary units

@bsudekum @1ec5 @JThramer  👀

### DIFF
--- a/MapboxCoreNavigation/DistanceFormatter.swift
+++ b/MapboxCoreNavigation/DistanceFormatter.swift
@@ -50,7 +50,7 @@ public class DistanceFormatter: LengthFormatter {
     }
     
     func maximumFractionDigits(for distance: CLLocationDistance) -> Int {
-        if numberFormatter.locale.usesMetric {
+        if NavigationSettings.shared.usesMetric {
             return distance < 3_000 ? 1 : 0
         } else {
             return distance.miles < 3 ? 1 : 0
@@ -58,7 +58,7 @@ public class DistanceFormatter: LengthFormatter {
     }
     
     func roundingIncrement(for distance: CLLocationDistance, unit: LengthFormatter.Unit) -> Double {
-        if numberFormatter.locale.usesMetric {
+        if NavigationSettings.shared.usesMetric {
             if distance < 25 {
                 return 5
             } else if distance < 100 {
@@ -112,7 +112,7 @@ public class DistanceFormatter: LengthFormatter {
     
     func formattedDistance(_ distance: CLLocationDistance, modify unit: inout LengthFormatter.Unit) -> String {
         var formattedDistance: String
-        if numberFormatter.locale.usesMetric {
+        if NavigationSettings.shared.usesMetric {
             let roundedDistance: CLLocationDistance = numberFormatter.number(from: numberFormatter.string(from: distance as NSNumber)!)?.doubleValue ?? distance
             numberFormatter.roundingIncrement = roundingIncrement(for: roundedDistance, unit: unit) as NSNumber
             
--- a/MapboxCoreNavigation/DistanceFormatter.swift
+++ b/MapboxCoreNavigation/DistanceFormatter.swift
@@ -13,6 +13,10 @@ extension CLLocationDistance {
         return self * .feetPerMeter
     }
     
+    var yards: CLLocationDistance {
+        return feet / 3
+    }
+    
     var kilometers: CLLocationDistance {
         return self / 1000
     }
@@ -86,12 +90,6 @@ public class DistanceFormatter: LengthFormatter {
      The user’s `Locale` is used here to set the units.
     */
     @objc public func string(from distance: CLLocationDistance) -> String {
-        // British roads are measured in miles, yards, and feet. Simulate this idiosyncrasy using the U.S. locale.
-        let localeIdentifier = numberFormatter.locale.identifier
-        if localeIdentifier == "en-GB" || localeIdentifier == "en_GB" {
-            numberFormatter.locale = Locale(identifier: "en-US")
-        }
-        
         numberFormatter.positivePrefix = ""
         numberFormatter.positiveSuffix = ""
         numberFormatter.decimalSeparator = nonFractionalLengthFormatter.numberFormatter.decimalSeparator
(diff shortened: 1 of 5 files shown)


MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -23,10 +23,28 @@ public class NavigationSettings: NSObject {
...
+     Specifies whether to mute the voice controller or not.
      */
     @objc public dynamic var voiceMuted     : Bool      = false
     
+    /**
+     Specifies the preferred distance measurement unit.
+     - note: Anything but `kilometer` and `mile` will fall back to the default measurement for the current locale.
+        Meters and feets will be used when the presented distances are small enough. See `DistanceFormatter` for more information.
+     */
+    @objc public dynamic var distanceUnit : LengthFormatter.Unit = Locale.current.usesMetric ? .kilometer : .mile
```
1ec5: Once the Directions API supports yards for the UK (see also mapbox/mapbox-navigation-ios#242), `mile` will become ambiguous. Consider defining a new enumeration along the lines of MapKit’s `MKDistanceFormatterUnits`.

MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -23,10 +23,28 @@ public class NavigationSettings: NSObject {
...
+    
+    var usesMetric: Bool {
+        get {
+            if distanceUnit == .kilometer {
```
JThramer: I would of used a `switch` here, but there's nothing wrong with this.

MapboxCoreNavigation/NavigationSettings.swift
frederoni: Kilometers or miles seems to be a consensus in all navigation and fitness apps I've seen. However, should be fairly straightforward to support yards for walking/cycling in the UK as tail work if needed.

1ec5: The UK uses yards on [a number of road signs](https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/203668/traffic-signs-manual-chapter-07.pdf#page=155), so it’d be prudent to match signage. This is also the stance [CLDR](https://unicode.org/cldr/trac/browser/trunk/common/supplemental/supplementalData.xml?rev=13744#L4833) has taken, so iOS follows this convention.

bsudekum:
Good to merge this?

frederoni:
I think we should do what @1ec5 suggested and have `Navigation/RouteOptions` being the canonical source. We can still override `NavigationSettings.distanceUnits` when navigating to switch units, but it would make it slightly more convenient to only having to set distance when requesting a route.

MapboxCoreNavigation/NavigationSettings.swift
frederoni: Figuring out when to use yards would be an additional feature that we'd have to add support for cases like a sign indicating railway station, ferry, etc. The distance unit would still be set to miles, but the distance formatter would have to become smarter to figure out when to display yards or feet.

frederoni:
Ready for review @1ec5 👀

MapboxCoreNavigation/NavigationSettings.swift
1ec5: The heuristic would be pretty simple: when LengthFormatter wants to use yards, use yards instead of forcing feet, as we do now. Am I missing something?

MapboxCoreNavigation/NavigationSettings.swift
1ec5: Agreed, `switch` would feel better here.

MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -20,12 +20,31 @@ public class NavigationSettings: NSObject {
...
+                return true
+            } else if distanceUnit == .mile {
```
1ec5: Sorry for being stubborn, but even if we have to regress support for yards in the UK with this PR, let’s at least futureproof this code for when we do restore yard support. Replace this if/else statement with a switch statement that includes cases for meters, feet, and yards.

MapboxCoreNavigation/NavigationSettings.swift
frederoni: > Am I missing something?

It's probably me, IDK, U.K is confusing. I was under the impression that U.K only uses yards for some special edge cases or that we regressed but that doesn't seem to be the case.

Is https://github.com/mapbox/mapbox-navigation-ios/pull/1055/commits/9c42982799a2f2c7fef3e57a487471c34379bd72 looking better?

MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -4,6 +4,7 @@ let metersPerMile: CLLocationDistance = 1_609.344
 let secondsPerHour = 60.0 * 60.0
 let yardsPerMile = 1_760.0
 let feetPerMile = yardsPerMile * 3.0
+let yardsPerMeter = 1.0936133
```
1ec5: It’s possible to consolidate this constant with `feetPerMeter` or `yardsPerMile`, relying on the fact that there are three feet in a meter.

MapboxCoreNavigationTests/DistanceFormatterTests.swift
```diff
@@ -66,7 +66,8 @@ class DistanceFormatterTests: XCTestCase {
         NavigationSettings.shared.distanceUnit = .mile
         distanceFormatter.numberFormatter.locale = Locale(identifier: "en-GB")
         
-        assertDistance(0,               displayed: "0 ft")
+        assertDistance(0,               displayed: "0 yd")
```
1ec5: 0&nbsp;feet is more relatable than 0&nbsp;yards. I’m guessing 10&nbsp;yards is a good threshold under which feet is more appropriate.

bsudekum: Do we really want to introduce yards at all? Currently we only use ft, right?

frederoni: Despite the distance to the sign is usually measured in meters, the sign itself, almost exclusively displays the distance in yards between 0-0.1 mile in the U.K. This seems to be the correct behavior.
ref:
https://en.wikipedia.org/wiki/Road_signs_in_the_United_Kingdom
https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/203668/traffic-signs-manual-chapter-07.pdf#page=155
https://www.waze.com/forum/viewtopic.php?f=657&t=114107
https://wiki.waze.com/forum/viewtopic.php?f=3&t=29882

Issue #996, opened 2018-01-10
Tapping "End Navigation" in the arrival UI should call the cancel delegate method


ericrwolfe:
I've implemented a basic Navigation UI that's pushed into a UINavigationController.

The parent view controller implements the following to pop navigation after the cancel button is pressed:
```swift
extension ListViewController: NavigationViewControllerDelegate {
    func navigationViewControllerDidCancelNavigation(_ navigationViewController: NavigationViewController) {
        self.navigationController?.popViewController(animated: true)
        self.navigationController?.setNavigationBarHidden(false, animated: true)
    }
}
```

However, if the arrival UI is shown, tapping "End Navigation" will not call that delegate method, and it will be impossible to exit navigation.

cc @mapbox/navigation-ios @JThramer

ericrwolfe:
Side note: the destination label in the arrival UI should also have some left/right padding (similar to the text box that appears when rating the ride).

PR #1056, opened 2018-01-22
 Send exit delegate when exiting end of route feedback ui

bsudekum:
close: https://github.com/mapbox/mapbox-navigation-ios/issues/996

/cc @mapbox/navigation-ios

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -369,6 +369,14 @@ extension ViewController: NavigationViewControllerDelegate {
         navigationViewController.present(confirmationController, animated: true, completion: nil)
         return false
     }
+    
+    
+    // Called when the user hits the `Cancel` button.
+    // If implemented, you are responsible for also dismissing the UI.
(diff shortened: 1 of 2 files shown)


Examples/Swift/ViewController.swift
```diff
@@ -369,6 +369,14 @@ extension ViewController: NavigationViewControllerDelegate {
         navigationViewController.present(confirmationController, animated: true, completion: nil)
         return false
     }
+    
+    
+    // Called when the user hits the exit button.
+    // If implemented, you are responsible for also dismissing the UI.
+    func navigationViewControllerDidCancelNavigation(_ navigationViewController: NavigationViewController) {
+        print("The user has exited")
```
JThramer: Is this necessary?

bsudekum: Shrug, was just making it obvious to the developer. can remove.

JThramer: Dealer's choice.

Issue #1062, opened 2018-01-23
Hooking up NavigationView to RMVC


JThramer:
It has become necessary to hook-up the new `NavigationView` class (introduced in #981) to the `RouteMapViewController` in order to get to a stable architectural state, so we can unblock #993. 

This ticket will track work done to fully integrate the `NavigationView` into the RMVC. Once done, all RMVC layout should be 100% code-based.

/cc @bsudekum @frederoni @1ec5 @ericrwolfe

JThramer:
Closing this with the merge of #1063.

PR #1063, opened 2018-01-23
NavigationView Hookup

JThramer:
This PR Implements #1062, and is currently a work-in-progress. As outlined elsewhere, this PR is a prerequisite to further architectural work on the framework, and thus has become important injection work.

To Do:

- [x] Refactor IUO's out of `NavigationView`
- [x] Actually hook up `NavigationView` to the `RouteMapViewController`
- [x] Fix lots of bugs
- [x] Hookup EndOfRoute UI / Constraints in RMVC
- [x] Hookup Feedback UI / Constraints in RMVC
- [x] Investigate why RMVC (and thus EORVC) is not receiving `destination`
- [x] Fix `StepsTableViewController` layout
- [x] Fix floating button label appearance with night style
- [x] Fix "Falling from Space" regression at initial map presentation

Post-"Falling From Space" To-Do:
- [x] Make sure things actually work with real locations (recently fixed via #1075, will resolve on rebase)
- [x] Fix the "X" button not working In DayStyle (could not reproduce)
- [x] Bring PR in-sync with `master`
- [x] Apply day style only when returning from presentation of NVC.

Observed issues during testing: 
- [x] Navigation Scene is initially blank when not simulating
- [x] App crashes when rerouting

Final, for-realsies last things to do before we put this to bed:
- [x] Fix status-view layout weirdness
- [x] Fix reroute crash 
- [x] Fix image on recenter button
- [x] Fix text color in night style for "next" text
- [x] ~Fix "flash" of blank UI at presentation~ (can not reproduce)
- [x] Clean Up (Comments, etc)
- [x] Fix issue where way label does not hide when resume button shows.
- [x] Fix issue where user puck does not stay glued to route.
- [x] Fix issue where ViewController mapView is nil after successful completed (not cancelled) route.

The "GOOD GRIEF WHEN WILL IT END?" To-do list: 
- [x] `WayNameLabel` ➡️  `WayNameView` (Medium)
- [x] Adjust the user location puck to the right spot (trivial)
- [x] `StatusView` is showing at EOR (small)
- [x] ~`StatusView` doesn't always show when rerouting~ (CNR - @bsudekum )
- [x] Fix issue where items in the `informationStackView` would look weird appearing and disappearing
- [x] Hopefully the last merge conflict resolution
- [x] Fix issue where rerouting status doesn't always look right
- [x] ~Use property reference for `mapView` instead of local var ([link](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#discussion_r170081780))~ (intentional)
- [x] Put feedback cell vertical padding into constant ([link](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#discussion_r170086373))
- [x] Use inset manipulation to add padding onto constant inset instead of inline construction ([link](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#discussion_r170082314))
- [x] don't force upwrap gesture recognizers ([link](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#discussion_r170082632))
- [x] Feeback looks wonky on iPhone X ([link](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#issuecomment-367757391))
- [x] FIX ANOTHER MERGE CONFLICT YAAAAAAY ([link](https://i.imgur.com/usPqv.gif))
- [x] Fix issue where steps table view would show would animate in front of the instructions banner. ([link](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#issuecomment-368167204))
- [x] FeedbackViewController height looks off ([link](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#issuecomment-368173351))
- [x] Fix merge clobbers
/cc @bsudekum @1ec5 @frederoni @ericrwolfe


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -99,34 +100,39 @@ class RouteMapViewController: UIViewController {
...
-        lanesView.isHidden = true
-        statusView.isHidden = true
-        statusView.delegate = self
-        nextBannerView.isHidden = true
-        isInOverviewMode = false
-        instructionsBannerView.delegate = self
-        bottomBannerView.delegate = self
+//        mapView.tracksUserCourse = true
+//        mapView.delegate = self
+//        mapView.navigationMapDelegate = self
```
JThramer: There are more delegates on the `NavigationMapView` than just `.delegate`, fix this.

MapboxNavigation/NavigationView.swift
```diff
@@ -38,26 +38,89 @@ import MapboxDirections
...
-    // Vertically laid-out stack view below the information stack view ontop of the map view, docked
-    // to the top right, consisting of Overview, Mute, and Report button.
-    weak var floatingStackView: UIStackView!
-    weak var overviewButton: FloatingButton!
-    weak var muteButton: FloatingButton!
-    weak var reportButton: FloatingButton!
-    weak var rerouteReportButton: ReportButton!
-    weak var separatorView: SeparatorView!
+    private struct Images {
+        static let overview = #imageLiteral(resourceName: "overview")
```
frederoni: Nit: Unfortunately, no literals can be used. We have to pass in the `bundle:`

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -105,9 +106,12 @@ class RouteMapViewController: UIViewController, NavigationViewDelegate {
     var annotatesSpokenInstructions = false
 
     override func loadView() {
+        super.loadView()
```
JThramer: @frederoni heh, I totally didn't catch this until I hit it, but LPT: If you override a `loadView()` method of a `UIViewController()`, you must _never_ [call super](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621454-loadview) in your overloaded implementation.

JThramer:
Starting to see [signs of life!](https://www.dropbox.com/s/w587b0sfbw35zna/signsoflife.mov?dl=0)

MapboxNavigation/StatusView.swift
```diff
@@ -8,6 +8,8 @@ protocol StatusViewDelegate: class {
 @IBDesignable
 @objc(MBStatusView)
 public class StatusView: UIView {
+    static let defaultBackgroundColor = UIColor.black.withAlphaComponent(2.0/3.0)
```
frederoni: Can we move this to the style?

JThramer: - [ ] Move to style

bsudekum:
It looks like the way name label has changed colors

image

MapboxNavigation/NavigationView.swift
```diff
@@ -38,26 +38,133 @@ import MapboxDirections
...
+    private static let rerouteReportTitle = NSLocalizedString("REROUTE_REPORT_TITLE", bundle: .mapboxNavigation, value: "Report Problem", comment: "Title on button that appears when a reroute occurs")
+    
+    static let buttonSize = CGSize(width: 50, height: 50)
+    
+    lazy var mapView: NavigationMapView = {
+        let map: NavigationMapView = .forAutoLayout()
+        map.delegate = delegate
+        map.navigationMapDelegate = delegate
+        map.courseTrackingDelegate = delegate
+        map.styleURL = URL(string: "mapbox://styles/mapbox/navigation-guidance-day-v2")
```
bsudekum: Does setting a custom styleURL on a `Style` class still work?

MapboxNavigation/NavigationView.swift
```diff
@@ -38,26 +38,133 @@ import MapboxDirections
...
+        map.showsUserLocation = true
```
bsudekum: It feels odd to be setting this here. Should this be something that is implemented in the view controller where a NavigationMapView is added?

MapboxNavigation/NavigationView.swift
JThramer: This line was added because of it's equivalent in [the storyboard](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/Resources/Base.lproj/Navigation.storyboard#L43), I'm not quite sure if it's even needed.

Examples/Swift/ViewController.swift
```diff
@@ -124,7 +124,6 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
         super.viewWillAppear(animated)
 
         // Reset the navigation styling to the defaults
-        DayStyle().apply()
```
JThramer: This needs to be fixed. This should fire only when the NVC _returns_ from a navigation session.

frederoni:
Rebasing atop master would make it easier to review this PR as it diverges. There will be a few hundred changes that are not relevant to this PR.

bsudekum:
Recenter appears to be un-styled at night:

image

bsudekum:
new3

Various portions of the top banner are visible and empty for a split section when the view is loaded.

JThramer:
Posterity: Can not reproduce https://github.com/mapbox/mapbox-navigation-ios/pull/1063#issuecomment-365454177 , navigation scene seems to present with data payload in place.

@bsudekum said it might of been a pre-rebase artifact. That sounds likely.

bsudekum:
new3

Both the current way name label and the resume button are visible at the same time. We actually hide the current way name label whilst the resume button is visible.

bsudekum:
new3

Also noting when resuming, the puck does not stay tight on the route line. The above is master and the desired behavior.

bsudekum:
Little compiler warning: 

image

frederoni:
Looks great so far but I found a few issues:

- [ ] `libc++abi.dylib: terminating with uncaught exception of type std::domain_error: latitude must not be NaN` when tapping the overview button.
- [ ] `WayNameLabel` is not styled correctly.
- [ ] Feedback view grows by approx the height of one cell every time it's opened.
- [ ] `TimeRemainingLabel` and `ArrivalTimeLabel` is not styled correctly.

Examples/Swift/ViewController.swift
```diff
@@ -298,21 +300,32 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
...
-        mapView!.userTrackingMode = .follow
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.delegate = self
+        mapView.navigationMapDelegate = self
+        mapView.userTrackingMode = .follow
+        let bottomPadding = (view.frame.height + view.frame.origin.y) - bottomBar.frame.origin.y
         
-        view.insertSubview(mapView!, belowSubview: bottomBar)
+        var topPadding: CGFloat = 0.0
+        if #available(iOS 11.0, *) {
```
frederoni: There is a convenient `safeTopAnchor` in an extension on UIView. Can we make use of that instead?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -124,19 +125,113 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-        imageView.frame = bounds
-        addSubview(imageView)
+        let contentFrame = UIEdgeInsetsInsetRect(bounds, contentInset)
+        let courseViewWidth = userCourseView?.frame.width ?? 0
+        let courseViewHeight = userCourseView?.frame.height ?? 0
+        let edgePadding = UIEdgeInsets(top: 50 + courseViewHeight / 2,
+                                       left: 50 + courseViewWidth / 2,
+                                       bottom: 50 + courseViewHeight / 2,
+                                       right: 50 + courseViewWidth / 2)
+        return CGPoint(x: max(min(contentFrame.midX,
```
frederoni: The anchor point has been moved to the center of the map view. Can we move it back to the bottom so that more of the upcoming route is visible?

MapboxNavigation/DayStyle.swift
```diff
@@ -98,8 +98,8 @@ open class DayStyle: Style {
         NavigationMapView.appearance().trafficSevereColor       = .trafficSevere
         NavigationMapView.appearance().trafficUnknownColor      = .trafficUnknown
         NextBannerView.appearance().backgroundColor = #colorLiteral(red: 0.9675388083, green: 0.9675388083, blue: 0.9675388083, alpha: 1)
+        NextBannerView.appearance().textColor = #colorLiteral(red: 0.09803921569, green: 0.09803921569, blue: 0.09803921569, alpha: 1)
         NextInstructionLabel.appearance().font = UIFont.systemFont(ofSize: 20, weight: .medium).adjustedFont
-        NextInstructionLabel.appearance().textColor = #colorLiteral(red: 0.09803921569, green: 0.09803921569, blue: 0.09803921569, alpha: 1)
```
frederoni: What's the reasoning behind moving `NextInstructionsLabel.textColor` to a property on the `NextBannerView`?

Examples/Swift/ViewController.swift
JThramer: This is setting up the map inset, not an auto layout constraint.

MapboxNavigation/DayStyle.swift
JThramer: - [x] Change `NextBannerView.textColor` to `NextAppearanceLabel.appearanceTextColor`

MapboxNavigation/NavigationMapView.swift
JThramer: - [ ] Reposition user puck to appropriate anchor point

Examples/Swift/ViewController.swift
JThramer: - [ ] Try and see if this can be done better with Storyboarding on `ViewController`

JThramer:
>libc++abi.dylib: terminating with uncaught exception of type std::domain_error: latitude must not be NaN when tapping the overview button.

@frederoni I'm having trouble reproducing this. Can you give any more details?

frederoni:
> Feedback view grows by approx the height of one cell each time it's opened.

Happens every time on non-iPhone X devices. Just open the feedback, close, and open again.

> libc++abi.dylib: terminating with uncaught exception of type std::domain_error: latitude must not be NaN when tapping the overview button.

This also reproduces every time on a non-iPhone X device. 🤔 

Add a symbolic breakpoint `UIViewAlertForUnsatisfiableConstraints`. I also print the view hierarchy when this breakpoint triggers for convenience by adding this action `expr -l objc++ -O -- [[UIWindow keyWindow] _autolayoutTrace]` to it.

JThramer:
@frederoni CNR the feedback growth or the crash-on-overview on iPhone 8. Video enclosed.

[CNR Video](https://www.dropbox.com/s/a7e0cmfg4uzjl0a/CNR-1063.mov?dl=0)

However, I am able to repro the unsatisfiable constraints, will work on that next.

JThramer:
Posterity: Talked to Fred this morning, both issues are reproducible on iPhone SE with iOS 11.2. We talked about the `WayNameLabel`, there may be something wrong with the view hierarchy that's causing appearance updates not to be sent to some components, but we agreed that if the root cause cannot be found then we'll bring back `WayNameView`.

JThramer:
@1ec5  @frederoni @bsudekum okay, I’ve tracked down the iPhone SE “feedback collection view grows the second time you tap on it” bug — it’s pretty fascinating. Basically, our `FeedbackViewController` is hardcoded to a 375pt width. 

The two major things to note here is that the `FeedbackViewController` on the Storyboard is [hardcoded to a freeform size](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/Resources/Base.lproj/Navigation.storyboard#L709), and the `CollectionView`’s `flowLayout` is [hardcoded to 125x145](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/Resources/Base.lproj/Navigation.storyboard#L349), which is fine for any 375pt width display, but causes issues for anything smaller, like a 320pt display (iPhone SE). 

Basically, the first time we instantiate the FeedbackVC, the Collection View’s width is 375, even though the `FeedbackViewController`’s root-view width is only 320. I think this is caused by the "Freeform Size" setting in the storyboard. Interestingly, the collection view’s size causes all of our logic (that is designed to ensure the cells show in a 3x2 format) to work properly, but only the first time. The next time the `FeedbackViewController` shows, it’s collection view width is 320 (as it should be) and this causes problematic behavior in our logic, which still sees the old `flowLayout.itemSize` of 125x145 (instead of ~108x118, as it should be...) 

This is caused because we are missing a `flowLayout.estimatedItemSize = <something>` call, which triggers automatic sizing of collection view cells. The _really_ weird thing is `sizeForCellInIndexPath:` is still called, and that gives a correct answer, but the `flowLayout` never updates it’s `itemSize`, and our calculations fail by saying the Collection View should have three rows of two cells instead of two rows of three.

I haven’t figured out how to fix this yet, but it’s good that we know _why_ things are happening the way they are. The _really_ weird thing is that even though our storyboard and logic are screwed up on master too, this issue does _not_ reproduce there. 

Curiouser and curiouser…

bsudekum:
Seeing an issue with the WayNameLabel on the iPhone x as well 
image

bsudekum:
Seeing an issue where the statusView is not shown after rerouting multiple times.

JThramer:
@bsudekum Regarding [your previous comment](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#issuecomment-367162981) The WayNameLabel issue is known.

bsudekum:
Seeing two issues:
1. reroute text sticks around
1. feedback view is clipped

simulator screen shot - iphone x - 2018-02-22 at 09 05 41
simulator screen shot - iphone x - 2018-02-22 at 09 27 16

Examples/Swift/ViewController.swift
```diff
@@ -120,8 +118,14 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
...
         super.viewWillAppear(animated)
 
-        // Reset the navigation styling to the defaults
-        DayStyle().apply()
+        //Reload the mapView.
+        setupMapView()
+
+        // Reset the navigation styling to the defaults if we are returning from a presentation.
+        if (presentedViewController != nil) {
+            DayStyle().apply()
```
vincethecoder: I believe the right way to reset styling to the defaults is use a `styleManager` and probably call `timeOfDayChanged()`... right?

Examples/Swift/ViewController.swift
```diff
@@ -298,21 +302,32 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     }
     
     func setupMapView() {
-        guard mapView == nil else { return }
-        mapView = NavigationMapView(frame: view.bounds)
+        guard self.mapView == nil else { return }
+        let mapView = NavigationMapView(frame: view.bounds)
+        self.mapView = mapView
```
vincethecoder: maybe if we're explicitly dictating `self.mapView` and `mapView` we should probably use a different approach here right? 
Like so:
```
guard mapView == nil else { return }
mapView = NavigationMapView(frame: view.bounds)
```

Examples/Swift/ViewController.swift
```diff
@@ -298,21 +302,32 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
...
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.delegate = self
+        mapView.navigationMapDelegate = self
+        mapView.userTrackingMode = .follow
+        let bottomPadding = (view.frame.height + view.frame.origin.y) - bottomBar.frame.origin.y
         
-        mapView!.autoresizingMask = [.flexibleWidth, .flexibleHeight]
-        mapView!.delegate = self
-        mapView!.navigationMapDelegate = self
-        mapView!.userTrackingMode = .follow
```
vincethecoder: 🎉

Examples/Swift/ViewController.swift
```diff
@@ -298,21 +302,32 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
...
+        var topPadding: CGFloat = 0.0
+        if #available(iOS 11.0, *) {
+            topPadding = view.safeAreaInsets.top
+        } else if let navCon = navigationController {
+            topPadding = navCon.navigationBar.frame.size.height
+        }
+        mapView.contentInset = UIEdgeInsets(top: topPadding + 25, left: 25, bottom: bottomPadding + 25, right: 25)
```
vincethecoder: Could we assign this `25` value to a variable? This would make it more self-documenting. Seems like a magic number.

Examples/Swift/ViewController.swift
```diff
@@ -298,21 +302,32 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
...
         
-        view.insertSubview(mapView!, belowSubview: bottomBar)
+        view.insertSubview(mapView, belowSubview: bottomBar)
         
         let singleTap = UILongPressGestureRecognizer(target: self, action: #selector(didLongPress(tap:)))
-        for recognizer in mapView!.gestureRecognizers! where recognizer is UILongPressGestureRecognizer {
+        for recognizer in mapView.gestureRecognizers! where recognizer is UILongPressGestureRecognizer {
```
vincethecoder: I would suggest we probably safely unwrap `mapView.gestureRecognizers!`.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -53,12 +52,19 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
     
     override func viewDidLoad() {
         super.viewDidLoad()
+        
+        //FIXME: This is a workaround to ensure that the FVC looks good on non 375pt width screens, as the dynamic sizing logic isn't currently working properly.
+        flowLayout.itemSize = collectionView(collectionView, layout: flowLayout, sizeForItemAt: IndexPath(item: 0, section: 0))
```
vincethecoder: should we probably use **size classes** to set the `itemSize` for specific screen sizes?

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -142,7 +148,7 @@ extension FeedbackViewController: UICollectionViewDelegateFlowLayout {
     
     func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
         let width = floor(collectionView.bounds.width / 3)
-        return CGSize(width: width, height: width+5)
+        return CGSize(width: width, height: width+20)
```
vincethecoder: `20` seems like a magic number. Could we assign that to a variable?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -96,19 +115,113 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+                              contentFrame.minX + edgePadding.left),
+                       y: max(max(min(contentFrame.minY + contentFrame.height * 0.8,
+                                      contentFrame.maxY - edgePadding.bottom),
+                                  contentFrame.minY + edgePadding.top),
+                              contentFrame.minY + contentFrame.height * 0.5))
+    }
+    
+    /**
+     Determines whether the map should follow the user location and rotate when the course changes.
+     - seealso: NavigationMapViewCourseTrackingDelegate
```
vincethecoder: `seealso` -> `see also`

MapboxNavigation/FeedbackViewController.swift
vincethecoder: However, if we choose to go with this route, could we consider using the `UICollectionViewDelegateFlowLayout` functions? Like so:
```
// For size
func collectionView(UICollectionView, layout: UICollectionViewLayout, sizeForItemAt: IndexPath)
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -96,19 +115,113 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+                    updateCourseTracking(location: location, animated: false)
+                } else {
+                    userCourseView.center = userAnchorPoint
+                }
+                addSubview(userCourseView)
+            }
+        }
     }
     
+    private lazy var mapTapGesture = UITapGestureRecognizer(target: self, action: #selector(didRecieveTap(sender:)))
```
vincethecoder: Curious, why are we using `UITapGestureRecognizer` vs `Touches` for tapping our mapView? 
> Gesture recognizers may delay the delivery of touch objects to the view while they are recognizing gestures, and by default they cancel delivery of remaining touch objects to the view once they recognize their gesture.
[Reference: UIGestureRecognizers vs touches (accuracy)
](https://stackoverflow.com/a/11851706)

vincethecoder: PS: I know it's existing code, but this is just a general question for everyone. 
cc @mapbox/navigation-ios

MapboxNavigation/NavigationMapView.swift
JThramer: `seealso` is actually a [keyword](https://developer.apple.com/library/content/documentation/Xcode/Reference/xcode_markup_formatting_ref/SeeAlso.html#//apple_ref/doc/uid/TP40016497-CH45-SW1) in Apple Markup.

MapboxNavigation/FeedbackViewController.swift
JThramer: This approach isn't possible because `375`pt width screens (X, 8, 7, 6) share the same size class (`wChR`) as `320`pt width (iPhone SE) screens. Ideally, a `UICollectionViews` auto-size cell logic would take care of variations in the same size-class, but it's not working correctly. That's what this workaround addresses.

Examples/Swift/ViewController.swift
JThramer: This approach won't work because the `ViewController` will always show a `DayStyle` theme, regardless of the time-of-day.

In reality, the call isn't strictly necessary because the only thing the `DayStyle` does that affects the `ViewController` is that it changes the style of the status bar, a thing we could do directly. But I digress.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -135,6 +248,49 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        suspendNotifications()
+    }
+    
+    //MARK: - Overrides
+    
+    open override func prepareForInterfaceBuilder() {
+        super.prepareForInterfaceBuilder()
+        
+        let image = UIImage(named: "feedback-map-error", in: .mapboxNavigation, compatibleWith: nil)
+        let imageView = UIImageView(image: image)
```
vincethecoder: We should safely unwrap the `image` variable, in case it mistakenly gets updated or deleted from the project.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -135,6 +248,49 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        imageView.backgroundColor = .gray
+        imageView.frame = bounds
+        addSubview(imageView)
+    }
+    
+    open override func layoutSubviews() {
+        super.layoutSubviews()
+        
+        //If the map is in tracking mode, make sure we update the camera after the layout pass.
+        if (tracksUserCourse) { updateCourseTracking(location: userLocationForCourseTracking, animated: false) }
```
vincethecoder: should wrap unto a new line.

MapboxNavigation/FeedbackViewController.swift
vincethecoder: Hmmm... so weird tho. Got it.

MapboxNavigation/NavigationMapView.swift
vincethecoder: aah 👍

MapboxNavigation/FeedbackViewController.swift
JThramer: Yeah, It's an artifact of our "half-storyboard, half-code" UI state. As we continue moving everything into code, `sizeForItemAtIndexPath:` should start giving us correct values again.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -410,55 +400,36 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-        return candidates
-    }
-    
-    private func routes(closeTo point: CGPoint) -> [Route]? {
-        let tapCoordinate = convert(point, toCoordinateFrom: self)
-        
-        //do we have routes? If so, filter routes with at least 2 coordinates.
-        guard let routes = routes?.filter({ $0.coordinates?.count ?? 0 > 1 }) else { return nil }
+        // Capture altitude for double tap and two finger tap after animation finishes
+        if sender is UITapGestureRecognizer, sender.state == .ended {
```
vincethecoder: This is way we should use `touchesEnded(_: ,with)`. With that, we can use  Set<UITouch>.count() to determine the number of fingers the user is using and we are guaranteed touches did end. (PS: I don't see where we're setting `numberOfTapsRequired` on our gesture recognizer to `2` either. So this is certainly a concern.)

MapboxNavigation/NavigationMapView.swift
1ec5: The map SDK implements its built-in gestures using gesture recognizers rather than interpreting touches manually, so NavigationMapView would need to use gesture recognizers to avoid conflicts.

The UIGestureRecognizer API offers several ways to resolve conflicts between gesture recognizers. If you encounter a case where the map SDK exhibits improper gesture behavior, please open a ticket upstream in the mapbox-gl-native repository. Thanks!

MapboxNavigation/NavigationMapView.swift
```diff
@@ -785,6 +756,71 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        guard let routes = routes?.filter({ $0.coordinates?.count ?? 0 > 1 }) else { return nil }
+        
+        //Sort routes by closest distance to tap gesture.
+        let closest = routes.sorted { (left, right) -> Bool in
+            
+            //existance has been assured through use of filter.
+            let leftLine = Polyline(left.coordinates!)
+            let rightLine = Polyline(right.coordinates!)
+            let leftDistance = leftLine.closestCoordinate(to: tapCoordinate)!.distance
+            let rightDistance = rightLine.closestCoordinate(to: tapCoordinate)!.distance
```
vincethecoder: lines `805-809` force unwrap. Yikes! 💀

MapboxNavigation/NavigationView.swift
```diff
@@ -38,26 +38,132 @@ import MapboxDirections
...
+    lazy var endOfRouteHeightConstraint: NSLayoutConstraint? = self.endOfRouteView?.heightAnchor.constraint(equalToConstant: 260)
+    
+    lazy var rerouteFeedbackTopConstraint: NSLayoutConstraint = self.rerouteReportButton.topAnchor.constraint(equalTo: self.informationStackView.bottomAnchor, constant: 10)
+    
+    private struct Images {
+        static let overview = UIImage(named: "overview", in: .mapboxNavigation, compatibleWith: nil)!.withRenderingMode(.alwaysTemplate)
+        static let volumeUp = UIImage(named: "volume_up", in: .mapboxNavigation, compatibleWith: nil)!.withRenderingMode(.alwaysTemplate)
+        static let volumeOff =  UIImage(named: "volume_off", in: .mapboxNavigation, compatibleWith: nil)!.withRenderingMode(.alwaysTemplate)
+        static let feedback = UIImage(named: "feedback", in: .mapboxNavigation, compatibleWith: nil)!.withRenderingMode(.alwaysTemplate)
+    }
```
vincethecoder: Maybe we shouldn't force unwrap the images here and allow whoever uses the Images struct to unwrap them. It's safer that way.

MapboxNavigation/NavigationView.swift
```diff
@@ -38,26 +38,132 @@ import MapboxDirections
...
+    
+    private struct Actions {
+        static let cancelButton: Selector = #selector(NavigationView.cancelButtonTapped(_:))
+    }
+    
+    private static let rerouteReportTitle = NSLocalizedString("REROUTE_REPORT_TITLE", bundle: .mapboxNavigation, value: "Report Problem", comment: "Title on button that appears when a reroute occurs")
+    
+    static let buttonSize = CGSize(width: 50, height: 50)
```
vincethecoder: Can we create these buttons and views in a storyboard or a nib? It helps to visually discern our intention. Question, do you agree this remove most of the `.forAutoLayout()` in the code?

MapboxNavigation/NavigationMapView.swift
JThramer: So it's actually not _always_ an anti pattern to use force-unwraps, it is considered good style to use them when you can _guarantee_ an existence check has already been made for the object in question.

In Line 800, we're filtering to ensure that `routes.coordinates != nil`, and that `routes.coordinates.count >= 2`. When we use the bangs in Lines 806 & 807, that is appropriate because we know that `Route.coordinates` exists on those two calls. 

On lines 806-807, we use bangs there because `Polyline.closestCoordinate(to:)` will [return nil](https://github.com/mapbox/turf-swift/blob/master/Sources/Turf/Turf.swift#L260) if the coordinates array passed in is empty. Because we can guarantee that there are at least two coordinates in the polyline, we can assume that the function will never return `nil`.

MapboxNavigation/FeedbackViewController.swift
vincethecoder: @JThramer Yeah, I was having this conversation with @bsudekum earlier on. It makes more sense to keep things mostly in storyboards if we can. To address the potential merge-conflicts we could split the main storyboards into connected storyboards to handle specific tasks. (In case of merge-conflicts concern, I can definitely assist in this domain. 😸 ) 
cc @mapbox/navigation-ios

MapboxNavigation/NavigationMapView.swift
vincethecoder: So, really if the map SDK uses gesture recognizers, then we are forced to use gesture recognizers because using any of the touches delegates will get canceled. Which makes sense.

MapboxNavigation/NavigationMapView.swift
vincethecoder: In fact, I noticed that a few minutes after I scrolled back up to review the variable but you had left a comment. So I take it back 😃

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -98,37 +102,41 @@ class RouteMapViewController: UIViewController {
...
+    
+
+    convenience init(routeController: RouteController, delegate: RouteMapViewControllerDelegate? = nil) {
+        self.init()
+        self.routeController = routeController
+        self.delegate = delegate
+    }
+    
+    
+    override func loadView() {
```
vincethecoder: should we call `super.loadView()` here ?

JThramer: [Nope.](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621454-loadview)

MapboxNavigation/NavigationMapView.swift
vincethecoder: **Follow up**: So it makes sense to use `UITapGestureRecognizer`, however, I'm still concerned that we're not setting `numberOfTapsRequired` to 2 on our gesture recognizer but comment reads that `Capture altitude for double tap and two finger tap after animation finishes`... thoughts?

JThramer:
Fascinating.

@vincethecoder In reply to [these](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#discussion_r170111150) two [comments](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#discussion_r170118125):

This PR was started to implement the [work started](https://github.com/mapbox/mapbox-navigation-ios/issues/1062) by @frederoni to refactor the "Navigation Scene" (`RouteMapViewController`) out of storyboards and into code. It was started because his work affected an [earlier PR](https://github.com/mapbox/mapbox-navigation-ios/pull/993) of mine that was intended to refactor view-appearance-updating code out of `RouteMapViewController`, and into the individual views themselves (something that we've made progress on in-parallel.) 

So why not storyboarding? The impetus behind @frederoni & I's work here is that _storyboards are not extensible._ This is not something that matters much in normal consumer codebases, but in a framework, where extensibility and customization needs are paramount, this is a problem.

The reason why storyboards are not extensible is because they do not return Types, but _repeatable, serialized instances_ of types. Effectively, any client that wishes to subclass one of our `ViewController`s loses all of the visual layout configuration that we've done for that class, which is not desirable.

Ideally, Instead of an inheritance pattern, I'd like to see a protocol-oriented architecture that makes heavy use of composition and default implementation, but that's probably still a little ways off yet.

If we want to resurrect the "storyboarding vs. code-layout" debate, we should do so as a team, but keep in mind that keeping things in-storyboard will make most of #1062 (and this PR) obsolete.

/cc @bsudekum @1ec5 @akitchen @brsbl

MapboxNavigation/NavigationView.swift
JThramer: Replied to [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#issuecomment-367859900)

MapboxNavigation/FeedbackViewController.swift
JThramer: Replied to [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#issuecomment-367859900)

MapboxNavigation/NavigationView.swift
JThramer: Optionals!

Optionals are considered desirable when, _in the course of normal operation_, a developer could encounter a case where the property/variable/pony might not exist. 

In this case, since the images are bundled in the framework, the developer should not normally expect a situation where those images are `nil`. In this sense, we use the force-unwrap operator like an assertion -- we ensure that something that _should_ be there (out of the developer's control), and then free up the developer from having to make that assurance upon every invocation. 

Consider the following -- if our image asset was missing, would it be better for the developer to receive an assertion, a recoverable error, or nothing at all? Often times, it's desirable to "fail fast" when "we" make a mistake with _our_ framework, as wrapping such unlikely scenarios in optionals can lead to subtle, very-hard-to-diagnose bugs down the line.

However, this struct should really be a careless enum -- something I forgot to do whilst working on this. Thanks for reminding me!

/cc @1ec5 @akitchen

bsudekum: I think in this case, force unwrapping is fine.

MapboxNavigation/NavigationMapView.swift
JThramer: Not necessarily disagreeing with you, but it's worth noting that this code was pulled wholesale from [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/1063/files/3d54668965e7ba331c280d31a10a9a62b3c79b33#diff-a43bde32226ca771fb8f732e90f4fc6aL213)

MapboxNavigation/NavigationMapView.swift
JThramer: I know 👍 'd earlier, but I'm going to digress on this one, for the same reason we use force-unwraps [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/1063#discussion_r170130885)

The fact that it's just a `prepareForInterfaceBuilder` call got me thinking for a minute, but `feedback-map-error` is used at runtime as well, so it seems okay that we'd assert that it exists here.

MapboxNavigation/NavigationView.swift
```diff
@@ -38,26 +38,132 @@ import MapboxDirections
...
+    
+    lazy var bannerHideConstraints: [NSLayoutConstraint] = [
+        self.informationStackView.bottomAnchor.constraint(equalTo: self.topAnchor),
+        self.instructionsBannerContentView.topAnchor.constraint(equalTo: self.instructionsBannerView.topAnchor)
+    ]
+    
+    lazy var endOfRouteShowConstraint: NSLayoutConstraint? = self.endOfRouteView?.bottomAnchor.constraint(equalTo: self.safeBottomAnchor)
+    
+    lazy var endOfRouteHideConstraint: NSLayoutConstraint? = self.endOfRouteView?.topAnchor.constraint(equalTo: self.bottomAnchor)
+
```
JThramer: Note to future @jthramer: constant-ize these Magic numbers

Examples/Swift/ViewController.swift
JThramer: There's actually a reason for this pattern -- `self.mapView` is an optional, but since we can guarantee it's existence (because we created it), we use the local reference. It's cleaner than using force-unwraps.

MapboxNavigation/NavigationView.swift
JThramer: Thanks, past @JThramer!

bsudekum:
One last thing i found, 
new3

The list view should animate from underneath the instructions banner.

JThramer:
@bsudekum fixed

bsudekum:
Feedback view is a little tall now 😕 

image

Examples/Swift/ViewController.swift
bsudekum: Is this still actionable?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -65,6 +73,24 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+     Attribute name for the route line that is used for identifying whether a RouteLeg is the current active leg.
+     */
+    public let currentLegAttribute = "isCurrentLeg"
+    
+    /**
+     Attribute name for the route line that is used for identifying different `CongestionLevel` along the route.
+     */
+    public let congestionAttribute = "congestion"
+
+    let routeLineWidthAtZoomLevels: [Int: MGLStyleValue<NSNumber>] = [
```
bsudekum: This variable along with `congestionAttribute` and `currentLegAttribute` got moved out to Constants.swift.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -678,7 +651,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
                 arrow.lineCap = MGLStyleValue(rawValue: cap)
                 arrow.lineJoin = MGLStyleValue(rawValue: join)
                 arrow.lineWidth = MGLStyleValue(interpolationMode: .exponential,
-                                                cameraStops: MBRouteLineWidthByZoomLevel.multiplied(by: 0.70),
```
bsudekum: This change is incorrect.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -830,11 +868,11 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             
             let lines = mergedCongestionSegments.map { (congestionSegment: CongestionSegment) -> MGLPolylineFeature in
                 let polyline = MGLPolylineFeature(coordinates: congestionSegment.0, count: UInt(congestionSegment.0.count))
-                polyline.attributes[MBCongestionAttribute] = String(describing: congestionSegment.1)
```
bsudekum: This change is incorrect and should still use `MBCongestionAttribute`

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1099,6 +1137,20 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
 }
 
+//MARK: - Extensions
+
+extension Dictionary where Key == Int, Value: MGLStyleValue<NSNumber> {
```
bsudekum: This has been moved to it's own file, `Dictionary.swift`.

MapboxNavigation/CGSize.swift
```diff
@@ -0,0 +1,14 @@
...
+import UIKit
+
+extension CGSize: ExpressibleByFloatLiteral {
+    public typealias FloatLiteralType = Double
+    
+    public init(size: Double) {
+        self.init(width: size, height: size)
+    }
+    
+    public init(floatLiteral value: FloatLiteralType) {
```
bsudekum: Do these need to be public?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -4,6 +4,8 @@ import MapboxDirections
 import MapboxCoreNavigation
 import Turf
 
+let sourceOptions: [MGLShapeSourceOption: Any] = [.maximumZoomLevel: 16]
```
bsudekum: This does not need to be in the globals space.

MapboxNavigation/NavigationView.swift
```diff
@@ -74,17 +182,72 @@ open class NavigationView: UIView {
...
+            resumeButton,
+            wayNameView,
+            rerouteReportButton,
+            bottomBannerContentView,
+            instructionsBannerContentView
+        ]
         
-        wayNameLabel.text = "Street Label"
+        subviews.forEach(addSubview(_:))
+//        statusView.hide(delay: 0, animated: false)
```
bsudekum: remove if not needed.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -509,73 +528,122 @@ class RouteMapViewController: UIViewController {
...
     }
 
     func navigationMapView(_ mapView: NavigationMapView, shapeDescribing route: Route) -> MGLShape? {
-        return delegate?.navigationMapView(mapView, shapeDescribing: route)
+        return delegate?.navigationMapView?(mapView, shapeDescribing: route)
     }
     
     func navigationMapView(_ mapView: NavigationMapView, didTap: Route) {
-        delegate?.navigationMapView(mapView, didTap: route)
+        delegate?.navigationMapView?(mapView, didSelect: route)
```
bsudekum: This is a breaking change, correct? If so, let's document in the changelog.

MapboxNavigation/StepsViewController.swift
```diff
@@ -231,6 +231,14 @@ extension StepsViewController: UITableViewDataSource {
         }
     }
     
+    private func updateInstructionsView(_ view: StepInstructionsView, previous: RouteStep?, current: RouteStep, destination: Waypoint) {
```
bsudekum: I think this function might be unused.

bsudekum:
We should figure out what screen sizes our snapshot tests need. https://github.com/mapbox/mapbox-navigation-ios/pull/1063/commits/15a5c66b3faf16a3212aa3556d140b21fb2f35be added a few more images that I'm not sure are needed. We can talk with @frederoni when he gets back.

Examples/Swift/ViewController.swift
JThramer: @bsudekum @frederoni Looked into this -- bringing the creation of the mapView into storyboard doesn't look like a good idea right now, as the storyboard will likely keep a reference to it while the navigation scene is presented. There may be ways to get around this, but it's non-trivial enough to look out-of-scope for this PR.

JThramer:
@bsudekum @vincethecoder @frederoni PR has been Tested and ✅'d against [example](https://github.com/mapbox/navigation-ios-examples) repo.

1ec5:
A little late to the discussion about storyboards, but I just wanted to point out that there’s a third way: XIBs. Used judiciously, XIBs can help us reason about constraints and such without having to reach for pencil and paper every time we change the layout. XIBs are also pretty good in terms of extensibility: you can choose to put a view in a XIB, load it via a subclass, and stick it into the view hierarchy programmatically. The downside to XIBs is that they aren’t as convenient to use with segues and view controllers, not to mention actions that target objects in different XIBs. So adopting XIBs would be something of a half-measure.

PR #1069, opened 2018-01-25
Rewrite language support guide

1ec5:
Rewrote the language support guide to discuss all the aspects of navigation localization in detail, including the ways in which the developer may customize language support.

Some of this documentation will need to be updated once #617 and #1055 land, but there’s already enough confusion among developers that clarifying the documentation right now is warranted.

/cc @bsudekum @frederoni


languages.md
```diff
@@ -1,34 +1,56 @@
...
+| Spanish    | ✅              | ✅                             | ✅                     | ✅
+| Swedish    | ✅              | ✅                             | ✅                     | ✅
+| Turkish    | ❌              | ✅                             | ✅                     | ✅
+| Vietnamese | ✅              | ✅                             | ❌                     | ❌
+
+## Contributing
+
+See the [contributing guide](./CONTRIBUTING.md#adding-or-updating-a-localization) for instructions on adding a new localization or improving an existing localization.
+
+[osrmti]: https://github.com/Project-OSRM/osrm-text-instructions/
```
frederoni: Are these lines supposed to be hidden?

1ec5: Yes, this is an alternative way to create a hyperlink in Markdown. Any occurrence of `[…][osrmti]` above is linked to [this page](https://github.com/Project-OSRM/osrm-text-instructions/). Meanwhile, this line is hidden from the rendered page. I went with this syntax because linking text within a table header causes the table to become fairly unreadable.

Issue #1072, opened 2018-01-25
top/bottom banner view are blank


zijiazhai:
sorry, guys , I am using v0.13.0, but what cause this happen?
![wechatimg3](https://user-images.githubusercontent.com/10751092/35415199-5ccbca3c-01ea-11e8-928e-e59d307b16cc.jpeg)

zijiazhai:
default UI, after I switch simulate locations to without simulate locations.

bsudekum:
It appears this occurs when the location manager does not fire at the beginning of a route.

Sealos:
I'm also experiencing this issue as of 0.13.0.

bsudekum:
@zijiazhai @Sealos I'm having a hard time reproducing this bug. Can you give me a few steps to reproduce? Are you simulating location?

zijiazhai:
if u start navigation with simulation location first, then dismiss navigationController, then switch to start navigation without simulation location, that should reproduce the issue.

zijiazhai:
@bsudekum

PR #1075, opened 2018-01-26
Let at least one location update through even if non-qualified

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1072

We're throwing out the first location, which is often times not a great location. This let's the first location slip through so we update UI elements with some initial data.

/cc @mapbox/navigation-ios @zijiazhai @Sealos

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -530,12 +530,12 @@ extension RouteController: CLLocationManagerDelegate {
             hasFoundOneQualifiedLocation = true
         }
         
-        var loc: CLLocation?
+        var potentialLocation: CLLocation?
         
         // `filteredLocations` contains qualified locations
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -523,19 +523,32 @@ extension RouteController: CLLocationManagerDelegate {
...
+        
         let filteredLocations = locations.filter { $0.isQualified }
         
         if !filteredLocations.isEmpty, hasFoundOneQualifiedLocation == false {
             hasFoundOneQualifiedLocation = true
         }
         
-        guard let location = filteredLocations.last else {
-            if let lastLocation = locations.last, hasFoundOneQualifiedLocation {
+        var loc: CLLocation?
```
JThramer: nit: how about `potential` or `potentialLocation`?

Issue #1057, opened 2018-01-22
Get on the Mapbox Maps iOS SDK v4


bsudekum:
v4 is coming, there will be a decent amount of work to tend to WRT to run time styling. 

Can test this out via pods (or prob carthage too) via: https://raw.githubusercontent.com/mapbox/mapbox-gl-native/1ec5-e8n-8074/platform/ios/Mapbox-iOS-SDK.podspec

bsudekum:
This breaking change will allow us to make https://github.com/mapbox/mapbox-navigation-ios/issues/987 work better

1ec5:
> Can test this out via pods (or prob carthage too) via: https://raw.githubusercontent.com/mapbox/mapbox-gl-native/1ec5-e8n-8074/platform/ios/Mapbox-iOS-SDK.podspec

Note that expressions have landed as a beta.

PR #1076, opened 2018-01-26
Update to Maps SDK for iOS v4.0.0

bsudekum:
Close: https://github.com/mapbox/mapbox-navigation-ios/issues/1057

This PR updates to the Mapbox Maps for iOS dependency to v4.0.0.

todo:
- [x] zoom level dependent styling
- [x] congestion
- [x] active leg opacity
- [x] maneuver arrow
- [x] update changelog

Blocked by mapbox/mapbox-gl-native#11615.

/cc @mapbox/navigation-ios


MapboxNavigation.podspec
```diff
@@ -44,7 +44,7 @@ Pod::Spec.new do |s|
   s.module_name = "MapboxNavigation"
 
   s.dependency "MapboxCoreNavigation", "#{s.version.to_s}"
-  s.dependency "Mapbox-iOS-SDK", "~> 3.6"
+  s.dependency "Mapbox-iOS-SDK", "4.0.0-beta.1"
```
1ec5: Remember to update the Cartfile and MapboxNavigation-Documentation.podspec before merging.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -6,12 +6,13 @@ import Turf
...
 typealias CongestionSegment = ([CLLocationCoordinate2D], CongestionLevel)
 
-let routeLineWidthAtZoomLevels: [Int: MGLStyleValue<NSNumber>] = [
-    10: MGLStyleValue(rawValue: 8),
-    13: MGLStyleValue(rawValue: 9),
-    16: MGLStyleValue(rawValue: 11),
-    19: MGLStyleValue(rawValue: 22),
-    22: MGLStyleValue(rawValue: 28)
+
+let routeLineWidthAtZoomLevels: [Int: NSExpression] = [
```
1ec5: No need to wrap these constant values in NSExpression. The dictionary can be `[Int: Float]` or `[Int: Int]`, which also means you don’t need to declare the type, either.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -6,12 +6,13 @@ import Turf
...
+    10: NSExpression(forConstantValue: 8),
+    13: NSExpression(forConstantValue: 9),
+    16: NSExpression(forConstantValue: 11),
+    19: NSExpression(forConstantValue: 22),
+    22: NSExpression(forConstantValue: 28)
```
1ec5: @anandthakker, currently, it isn’t possible for a zoom expression (`$zoomLevel`) to appear outside a curve or interpolation. But would it be feasible to relax this restriction, so that this curve could be expressed as a single multiplication expression involving `$zoomLevel`?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -677,12 +670,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
                 arrow.minimumZoomLevel = minimumZoomLevel
-                arrow.lineCap = MGLStyleValue(rawValue: cap)
-                arrow.lineJoin = MGLStyleValue(rawValue: join)
-                arrow.lineWidth = MGLStyleValue(interpolationMode: .exponential,
-                                                cameraStops: routeLineWidthAtZoomLevels.multiplied(by: 0.70),
-                                                options: [.defaultValue : MGLConstantStyleValue<NSNumber>(rawValue: 1.5)])
-                arrow.lineColor = MGLStyleValue(rawValue: .white)
+                arrow.lineCap = NSExpression(forConstantValue: cap)
+                arrow.lineJoin = NSExpression(forConstantValue: join)
+                arrow.lineWidth = NSExpression(format: "FUNCTION($zoomLevel, 'mgl_interpolateWithCurveType:parameters:stops:', 'linear', nil, %@)", argumentArray: [routeLineWidthAtZoomLevels.multiplied(by: 1.5)])
```
1ec5: Instead of calling `multiplied(by:)` on the array, multiply the `FUNCTION` call by 1.5 inside the format string.

1ec5: Since you know in advance how many arguments this format string needs, use `NSExpression(format:)` instead of `NSExpression(format:argumentArray:)`. The first argument to `NSExpression(format:)` takes a comma-separated list of arguments following the format string.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -693,12 +684,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             } else {
                 
                 arrowStroke.minimumZoomLevel = minimumZoomLevel
-                arrowStroke.lineCap = MGLStyleValue(rawValue: cap)
-                arrowStroke.lineJoin = MGLStyleValue(rawValue: join)
-                arrowStroke.lineWidth = MGLStyleValue(interpolationMode: .exponential,
-                                                      cameraStops: routeLineWidthAtZoomLevels.multiplied(by: 0.80),
-                                                      options: [.defaultValue : MGLConstantStyleValue<NSNumber>(rawValue: 1.5)])
-                arrowStroke.lineColor = MGLStyleValue(rawValue: .defaultArrowStroke)
+                arrowStroke.lineCap = NSExpression(forConstantValue: cap)
```
1ec5: `NSExpression(forConstantValue: "cap")` or `NSExpression(format: "'cap'")` is also allowed now, although what you’ve got is also perfectly fine if you want compile-time validation.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -885,64 +870,57 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-                                             cameraStops: routeLineWidthAtZoomLevels.multiplied(by: 0.85),
-                                             options: [.defaultValue : MGLConstantStyleValue<NSNumber>(rawValue: 1.5)])
+        lineCasing.lineWidth = NSExpression(format: "FUNCTION($zoomLevel, 'mgl_interpolateWithCurveType:parameters:stops:', 'linear', nil, %@)", argumentArray: [routeLineWidthAtZoomLevels.multiplied(by: 0.85)])
         
-        lineCasing.lineColor = MGLStyleValue(rawValue: routeAlternateColor)
-        lineCasing.lineCap = MGLStyleValue(rawValue: NSValue(mglLineCap: .round))
-        lineCasing.lineJoin = MGLStyleValue(rawValue: NSValue(mglLineJoin: .round))
-        lineCasing.lineOpacity = MGLStyleValue(rawValue: 0.9)
+        lineCasing.lineColor = NSExpression(forConstantValue: routeAlternateColor)
+        lineCasing.lineCap = NSExpression(forConstantValue: NSValue(mglLineCap: .round))
```
1ec5: I think you can say something like `MGLLineCap.round as NSValue` instead of calling `NSValue(mglLineCap:)`. `NSExpression(forConstantValue: "round")` and `NSExpression(format: "'round'")` are also options.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -885,64 +870,57 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
     }
     
     func routeWaypointSymbolStyleLayer(identifier: String, source: MGLSource) -> MGLStyleLayer {
         let symbol = MGLSymbolStyleLayer(identifier: identifier, source: source)
         
-        symbol.text = MGLStyleValue(rawValue: "{name}")
-        symbol.textFontSize = MGLStyleValue(rawValue: 10)
-        symbol.textHaloWidth = MGLStyleValue(rawValue: 0.25)
-        symbol.textHaloColor = MGLStyleValue(rawValue: .black)
+        symbol.text = NSExpression(forConstantValue: "{name}")
```
1ec5: Pretty sure mapbox/mapbox-gl-native#10399 will prevent the legacy `{token}` syntax from working inside an expression. Use a key path expression instead: `NSExpression(forKeyPath: "name")` or `NSExpression(format: "name")`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -885,64 +870,57 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-        line.lineJoin = MGLStyleValue(rawValue: NSValue(mglLineJoin: .round))
+        line.lineWidth = NSExpression(format: "FUNCTION($zoomLevel, 'mgl_interpolateWithCurveType:parameters:stops:', 'linear', nil, %@)", argumentArray: [routeLineWidthAtZoomLevels])
+        line.lineColor = NSExpression(format: "FUNCTION($congestion, 'mgl_stepWithMinimum:stops:', %@, %@)", UIColor(red:1.0, green:0.72, blue:0.85, alpha:1.0), [
+            "unknown": NSExpression(forConstantValue: trafficUnknownColor),
+            "low": NSExpression(forConstantValue: trafficLowColor),
+            "moderate": NSExpression(forConstantValue: trafficModerateColor),
+            "heavy": NSExpression(forConstantValue: trafficHeavyColor),
+            "severe": NSExpression(forConstantValue: trafficSevereColor)
+            ])
+        line.lineOpacity = NSExpression(format: "FUNCTION($currentLegAttribute, 'mgl_interpolateWithCurveType:parameters:stops:', 'linear', nil, %@)", [
```
1ec5: `$currentLegAttribute` would be a reference to an expression variable declared with `mgl_expressionWithContext:`. Use `currentLegAttribute`, without the dollar sign, to refer to an attribute of the current feature.

1ec5: Actually, `currentLegAttribute` is the name of [this variable](https://github.com/mapbox/mapbox-navigation-ios/blob/e942a425208feda1ad2a15ee0c6ad27552440b34/MapboxNavigation/NavigationMapView.swift#L62). You have two options: either insert a key path expression created from the variable:

```swift
NSExpression(format: "FUNCTION(%@, 'mgl_interpolateWithCurveType:parameters:stops:', 'linear', nil, %@)", NSExpression(forKeyPath: currentLegAttribute), …)
```

or inline the key path and delete `currentLegAttribute`:

```swift
NSExpression(format: "FUNCTION(isCurrentLeg, 'mgl_interpolateWithCurveType:parameters:stops:', 'linear', nil, %@)", …)
```

MapboxNavigation/NavigationMapView.swift
anandthakker: Technically, it would be _possible_ to relax the restriction on zoom expressions to:
 - If the expression is a zoom-only expression, no restrictions at all.
 - If the expression is also feature-dependent, it must be in the form `F(zoom) * G(feature)`, where `F` is a step or interpolation expression using `zoom` as the input (and nowhere else), and `G` is not zoom-dependent.

We've chosen so far to go with the existing, less permissive restriction because it's so much simpler to understand.  I think the above, which is maximally permissive (barring significant changes to the data-driven styling implementation), is likely way too complicated.  We could go for something in between, but it'll be tricky to decide just where to draw the line.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -885,64 +870,57 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            true: 1,
+            false: 0
+            ])
```
1ec5: As a Boolean attribute, `isCurrentLeg` can have only two possible values, `true` and `false`, so a full stop dictionary is overkill. With an expression, you can express a condition directly:

```swift
NSExpression(format: "TERNARY(isCurrentLeg = true, 1, 0)")
```

or:

```swift
NSExpression(forConditional: NSPredicate(format: "isCurrentLeg = true"), true: 1, false: 0)
```

1ec5: Then again, the usual way to accomplish this effect in a style is to let each line know which leg index it represents and add two layers to the style, one for the active leg and another for the inactive legs.

Both layers would have constant opacities. The filters (predicates) for the two layers would be updated every time the leg index changes:

```swift
activeLegLayer.predicate = NSPredicate(format: "legIndex = %@", routeProgess.legIndex)
inactiveLegLayer.predicate = NSPredicate(format: "legIndex != %@", routeProgress.legIndex)
```

bsudekum:
This is currently blocked upstream by https://github.com/mapbox/mapbox-gl-native/issues/11277 & https://github.com/mapbox/mapbox-gl-native/issues/10849

MapboxNavigation/NavigationMapView.swift
```diff
@@ -902,64 +887,53 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-        
-        line.lineOpacity = MGLStyleValue(interpolationMode: .categorical, sourceStops: [
-            true: MGLStyleValue(rawValue: 1),
-            false: MGLStyleValue(rawValue: 0)
-            ], attributeName: MBCurrentLegAttribute, options: nil)
-        
-        line.lineJoin = MGLStyleValue(rawValue: NSValue(mglLineJoin: .round))
+        line.lineWidth = NSExpression(format: "FUNCTION($zoomLevel, 'mgl_interpolateWithCurveType:parameters:stops:', 'linear', nil, %@)", argumentArray: [MBRouteLineWidthByZoomLevel])
+        line.lineOpacity = NSExpression(forConditional: NSPredicate(format: "\(MBCurrentLegAttribute) == true"), trueExpression: NSExpression(forConstantValue: 1), falseExpression: NSExpression(forConstantValue: 0.85))
+        line.lineColor = NSExpression(format: "%@.severe", [
```
bsudekum: @anandthakker @1ec5 this is failing with 

```
Assertion failed: (literal), function operator(), file /Users/distiller/project/include/mbgl/style/conversion/data_driven_property_value.hpp, line 50.
```

anandthakker: @bsudekum @1ec5 could we dump the json expression representation that's getting produced by this NSExpression?

MapboxNavigation/Dictionary.swift
```diff
@@ -1,15 +1,15 @@
...
      */
     public func multiplied(by factor: Double) -> Dictionary {
-        var newCameraStop: [Int: MGLStyleValue<NSNumber>] = [:]
-        for stop in self as [Int : MGLStyleValue<NSNumber>] {
-            let f = stop.value as! MGLConstantStyleValue
-            let newValue =  f.rawValue.doubleValue * factor
-            newCameraStop[stop.key] = MGLStyleValue<NSNumber>(rawValue: NSNumber(value:newValue))
+        var newCameraStop: [Int: NSExpression] = [:]
+        for stop in self {
+            let currentValue = stop.value.constantValue as! Double
```
1ec5: Note that this will crash if the dictionary contains any expression that isn’t a constant value. Since this is a public extension on `Dictionary`, we have to assume the developer is going to try to use it wherever the compiler lets them.

I continue to believe we should make this method private or remove it outright. A few explicit multiplications won’t greatly inconvenience the developer.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -629,13 +624,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-                arrow.lineCap = MGLStyleValue(rawValue: cap)
-                arrow.lineJoin = MGLStyleValue(rawValue: join)
-                arrow.lineWidth = MGLStyleValue(interpolationMode: .exponential,
-                                                cameraStops: MBRouteLineWidthByZoomLevel.multiplied(by: 0.70),
-                                                options: [.defaultValue : MGLConstantStyleValue<NSNumber>(rawValue: 1.5)])
-                arrow.lineColor = MGLStyleValue(rawValue: .white)
-                
+                arrow.lineCap = NSExpression(forConstantValue: "cap")
+                arrow.lineJoin = NSExpression(forConstantValue: "join")
+                arrow.lineWidth = NSExpression(format: "FUNCTION($zoomLevel, 'mgl_interpolateWithCurveType:parameters:stops:', 'linear', nil, %@)", argumentArray: [MBRouteLineWidthByZoomLevel.multiplied(by: 0.70)])
```
1ec5: The syntax has now changed to `mgl_interpolate:withCurveType:parameters:stops:($zoomLevel, 'linear', nil, %@)`. (×7)

MapboxNavigation/NavigationMapView.swift
```diff
@@ -664,33 +653,29 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-                arrowSymbolLayer.iconColor = MGLStyleValue(rawValue: .white)
-                arrowSymbolLayer.iconRotationAlignment = MGLStyleValue(rawValue: NSValue(mglIconRotationAlignment: .map))
-                arrowSymbolLayer.iconRotation = MGLStyleValue(rawValue: shaftDirection as NSNumber)
-                arrowSymbolLayer.iconScale = MGLStyleValue(interpolationMode: .exponential,
-                                                           cameraStops: MBRouteLineWidthByZoomLevel.multiplied(by: 0.12),
-                                                           options: [.defaultValue : MGLConstantStyleValue<NSNumber>(rawValue: 0.2)])
-                arrowSymbolLayer.iconAllowsOverlap = MGLStyleValue(rawValue: true)
+                arrowSymbolLayer.iconImageName = NSExpression(forConstantValue: "triangle-tip-navigation")
+                arrowSymbolLayer.iconColor = NSExpression(forConstantValue: UIColor.white)
+                arrowSymbolLayer.iconRotationAlignment = NSExpression(forConstantValue: NSValue(mglIconRotationAlignment: .map))
```
1ec5: Simplify this to `NSExpression(forConstantValue: "map")` for consistency. (×2)

MapboxNavigation/NavigationMapView.swift
```diff
@@ -664,33 +653,29 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-                arrowSymbolLayerCasing.iconColor = MGLStyleValue(rawValue: .defaultArrowStroke)
-                arrowSymbolLayerCasing.iconRotationAlignment = MGLStyleValue(rawValue: NSValue(mglIconRotationAlignment: .map))
-                arrowSymbolLayerCasing.iconRotation = MGLStyleValue(rawValue: shaftDirection as NSNumber)
-                arrowSymbolLayerCasing.iconScale = MGLStyleValue(interpolationMode: .exponential,
-                                                                 cameraStops: MBRouteLineWidthByZoomLevel.multiplied(by: 0.14),
-                                                                 options: [.defaultValue : MGLConstantStyleValue<NSNumber>(rawValue: 0.2)])
-                arrowSymbolLayerCasing.iconAllowsOverlap = MGLStyleValue(rawValue: true)
+                arrowSymbolLayerCasing.iconImageName = NSExpression(forConstantValue: "triangle-tip-navigation")
+                arrowSymbolLayerCasing.iconColor = NSExpression(forConstantValue: UIColor.defaultArrowStroke)
+                arrowSymbolLayerCasing.iconRotationAlignment = NSExpression(forConstantValue: NSValue(mglIconRotationAlignment: .map))
```
1ec5: Reduce repetition by saying, for example:

```swift
arrowSymbolLayerCasing.iconRotationAlignment = arrowSymbolLayer.iconRotationAlignment
```

MapboxNavigation/NavigationMapView.swift
1ec5: Use a key path expression instead: `NSExpression(forKeyPath: "name")` or `NSExpression(format: "name")`.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/1076#discussion_r164260144

MapboxNavigation/NavigationMapView.swift
1ec5: This expression assumes all traffic is severe. The format should be:

```swift
NSExpression(format: "FUNCTION(%@, 'valueForKey:', congestion)")
```

Alternatively, use an `MGL_MATCH` expression:

```swift
NSExpression(format: "MGL_MATCH(congestion, 'unknown', %@, 'low', %@, 'moderate', %@, …)")
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1028,29 +996,30 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-                        localizedStops[zoomLevel] = MGLStyleValue(rawValue: token as NSString)
-                    }
-                }
-                if hasName {
-                    function.stops = localizedStops
-                    layer.text = function
-                }
-            }
+
+// TODO:
```
1ec5: Is this still to do?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -672,7 +672,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
                 arrowSymbolLayerCasing.minimumZoomLevel = minimumZoomLevel
                 arrowSymbolLayerCasing.iconImageName = NSExpression(forConstantValue: "triangle-tip-navigation")
                 arrowSymbolLayerCasing.iconColor = NSExpression(forConstantValue: UIColor.defaultArrowStroke)
-                arrowSymbolLayerCasing.iconRotationAlignment = NSExpression(forConstantValue: NSValue(mglIconRotationAlignment: .map))
+                arrowSymbolLayerCasing.iconRotationAlignment = arrowSymbolLayer.iconRotationAlignment
```
1ec5: The idea behind https://github.com/mapbox/mapbox-navigation-ios/pull/1076#discussion_r178962197 was that we’d do the same thing for the dozens of other properties in this file.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -925,14 +925,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
         line.lineOpacity = NSExpression(forConditional: NSPredicate(format: "\(MBCurrentLegAttribute) == true"), trueExpression: NSExpression(forConstantValue: 1), falseExpression: NSExpression(forConstantValue: 0.85))
-        line.lineColor = NSExpression(format: "%@.severe", [
-                "unknown": trafficUnknownColor,
-                "low": trafficLowColor,
-                "moderate": trafficModerateColor,
-                "heavy": trafficHeavyColor,
-                "severe": trafficSevereColor
-            ])
-        
+        line.lineColor = NSExpression(format: "MGL_MATCH(congestion, 'unknown', %@, 'low' , %@, 'moderate', %@, 'heavy', %@, 'severe', %@, %@)", trafficUnknownColor, trafficLowColor, trafficModerateColor, trafficHeavyColor, trafficSevereColor, UIColor.purple)
```
1ec5: Remove the case for `unknown` and make `trafficUnknownColor` the fallback.

MapboxNavigation/NavigationMapView.swift
1ec5: We can get the JSON expression representation via `NSExpression.mgl_jsonExpressionObject()`. It might be worth printing that out as part of the exception handler. Currently, the exception handler is in code that only knows about mbgl expressions, not NSExpression.

anandthakker: This is an expression parsing bug -- reopened here: https://github.com/mapbox/mapbox-gl-native/issues/10849

MapboxNavigationTests/NSExpressionTests.swift
```diff
@@ -0,0 +1,65 @@
...
+                    80000
+                ],
+                ["get", "abbr"],
+                ["get", "name"]
+            ],
+            5, ["get", "name"]
+        ]
+        let expression = NSExpression.mgl_expression(withJSONObject: jsonExpression)
+        let localizedExpression = expression.localized(into: nil)
+        XCTAssertEqual(localizedExpression.mgl_jsonExpressionObject as! NSArray, localizedJSONExpression as NSArray)
```
1ec5: This test will fail until mapbox/mapbox-gl-native#11615 lands in a build of the map SDK.

1ec5:
Not sure why, but the maneuver arrow’s shaft looks pretty anemic, as though the `lineWidth` expression is having no effect on this particular layer:

arrow

bsudekum:
Fixed arrow width issue with https://github.com/mapbox/mapbox-navigation-ios/pull/1076/commits/c2b276ae61d4b7eb34731bd50161ac8428834d18

MapboxNavigation/NSExpression.swift
```diff
@@ -0,0 +1,80 @@
...
+                localizedStops[zoomLevel] = localizedValue
+                hasLocalizedValue = true
+            }
+        }
+        return hasLocalizedValue ? localizedStops : self
+    }
+}
+
+extension NSExpression {
+    func localized(into locale: Locale?, replacingTokens replacesTokens: Bool) -> NSExpression {
```
1ec5: Nearly all of this functionality has been upstreamed to the map SDK in mapbox/mapbox-gl-native#11651. So the implementation here can be replaced with a simple call to the similarly named method provided by the map SDK.

1ec5: Done in 7bbfe0deb187c350638ff579788ee843bea50474.

1ec5:
The maneuver arrow still doesn’t look quite right. It has the wrong cap, and I think the casing is missing.

arrow

1ec5:
I’ve been pretty involved with this PR, so I’d like to get another pair of 👀 from either @vincethecoder or @JThramer. Thanks!

1ec5:
Tests are failing due to #1327.

Examples/Swift/CustomViewController.swift
```diff
@@ -138,10 +138,10 @@ class CustomViewController: UIViewController, MGLMapViewDelegate, AVSpeechSynthe
             let line = MGLLineStyleLayer(identifier: layerIdentifier, source: lineSource)
 
             // Style the line
-            line.lineColor = MGLStyleValue(rawValue: UIColor(red:0.00, green:0.45, blue:0.74, alpha:0.9))
-            line.lineWidth = MGLStyleValue(rawValue: 5)
-            line.lineCap = MGLStyleValue(rawValue: NSValue(mglLineCap: .round))
-            line.lineJoin = MGLStyleValue(rawValue: NSValue(mglLineJoin: .round))
+            line.lineColor = NSExpression(forConstantValue: UIColor(red:0.00, green:0.45, blue:0.74, alpha:0.9))
+            line.lineWidth = NSExpression(forConstantValue: 5)
+            line.lineCap = NSExpression(forConstantValue: NSValue(mglLineCap: .round))
```
frederoni: nit: `NSExpression(forConstantValue: "round")` seems to be the recommended way according to the documentation and we use throughout the migration except for here.

1ec5: Fixed in 0294f1209b08f3c7329bd0eecb5752aea359b1a7.

Issue #1060, opened 2018-01-22
updateOTA Error


zijiazhai:
Hi, guys I am using navigation-ios 0.13.
I can still see this error sometimes, Thanks.
![snip20180122_16](https://user-images.githubusercontent.com/10751092/35244347-759b8cd0-ff85-11e7-96df-ec9353a001ce.png)

bsudekum:
@zijiazhai we have not seen this crash since https://github.com/mapbox/mapbox-navigation-ios/pull/977/ was introduced in v0.12.2. 

Can you reproduce the crash consistently in v0.13.0? Is there anything notable you are doing when the crash occurs?

zijiazhai:
No, I only saw this error once today in 0.13.3. I will let you guys know if I can get some useful information to reproduce it.

zijiazhai:
This issue still exists in V0.13.1. Seems like when the app crash, the func updateETA was called twice. on the second time, routeController is nil.  Plz take a look @bsudekum , I am not sure how to reproduce yet.
![121519168293_ pic](https://user-images.githubusercontent.com/10751092/36454557-15bfb1b2-1662-11e8-9119-cc37744dae62.jpg)
![111519168281_ pic](https://user-images.githubusercontent.com/10751092/36454559-1848a0b0-1662-11e8-959a-db8b3c5869cb.jpg)

PR #1082, opened 2018-02-01
Make bottomBannerView outlet optional

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1060

We've had issues with this outlet for a while now. By making it optional when `bottomBannerView?.updateETA(routeProgress:)` is called and  `bottomBannerView` is nil, the function will not be called.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -20,7 +20,7 @@ class RouteMapViewController: UIViewController {
     @IBOutlet weak var instructionsBannerContainerView: InstructionsBannerContentView!
     @IBOutlet weak var instructionsBannerView: InstructionsBannerView!
     @IBOutlet weak var nextBannerView: NextBannerView!
-    @IBOutlet weak var bottomBannerView: BottomBannerView!
+    @IBOutlet weak var bottomBannerView: BottomBannerView?
     @IBOutlet weak var statusView: StatusView!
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -488,7 +489,7 @@ class RouteMapViewController: UIViewController {
         var margin: CGFloat = 0.0
         if #available(iOS 11.0, *) { margin = view.safeAreaInsets.bottom }
         let containerHeight = endOfRouteContainerView.frame.height - margin
-        let bottom = self.endOfRouteShowConstraint.isActive ? containerHeight : bottomBannerView.bounds.height
+        let bottom = self.endOfRouteShowConstraint.isActive ? containerHeight : bottomBannerView?.bounds.height ?? 40
```
bsudekum: @mapbox/navigation-ios Is there anything else we can do here other than falling back to 40?

frederoni: This should be factored out to a UIPercentDrivenInteractiveTransition or another interactive transitioning class but I think this works for now. /cc @JThramer

bsudekum: @JThramer @frederoni cool, mind giving this a 🍏  if everything else is good?

Issue #1081, opened 2018-01-31
StepIndex Incrementation leads to assert(false) --> crash


hervouinc:
**Mapbox Navigation SDK version: 0.13.0**

file: `RouteController`
`func updateRouteStepProgress(for location:)`
--> `func advanceStepIndex(to:)`
--> `routeProgress.currentLegProgress.stepIndex += 1`

file: `RouteProgress`
```
@objc public var stepIndex: Int {
        didSet {
            assert(stepIndex >= 0 && stepIndex < leg.steps.endIndex)
            currentStepProgress = RouteStepProgress(step: currentStep)
        }
 }
```

`assert` shoud be replaced by a `guard` or `StepIndex` should not be incremented.

bsudekum:
I think you raise a fair point. We should gently tell the developer they cannot increment further instead of asserting and crashing.

/cc @mapbox/navigation-ios

hervouinc:
Hi @bsudekum , saw you have closed the "Do not assert" PR. 
Can you referenced the PR that fix this issue when available ? Thanks

bsudekum:
@hervouinc I'm trying to reproduce the situation via your comment 

> * the user goes over the last waypoint of the route
> * we choose to not reroute when asked by the routeController to its delegate, at least for a while (based on several local properties) since we know the user can stop after the waypoint
> * the user's location is still handled by the routeController and leads to updateRouteStepProgress and then advanceStepIndex.

Can you clarify step 2 with a code snippet perhaps? 

I'm trying to reproduce it in the sim with something like: 
![image](https://user-images.githubusercontent.com/1058624/35820755-73b1aa18-0a5b-11e8-9f23-123156382c9c.png)

hervouinc:
@bsudekum here is the step 2:

bsudekum:
@hervouinc thanks, was successfully able to reproduce locally.

PR #1083, opened 2018-02-01
Do not assert

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1081

Although incrementing the step/leg index beyond the number of steps/legs denotes a serious issue, we should gently tell the developer something is wrong instead of asserting and crashing. 

/cc @hervouinc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -17,7 +17,7 @@ open class RouteProgress: NSObject {
      */
     @objc public var legIndex: Int {
         didSet {
-            assert(legIndex >= 0 && legIndex < route.legs.endIndex)
+            guard legIndex >= 0 && legIndex < route.legs.endIndex else { return }
             // TODO: Set stepIndex to 0 or last index based on whether leg index was incremented or decremented.
             currentLegProgress = RouteLegProgress(leg: currentLeg)
         }
@@ -158,7 +158,7 @@ open class RouteLegProgress: NSObject {
      */
     @objc public var stepIndex: Int {
         didSet {
-            assert(stepIndex >= 0 && stepIndex < leg.steps.endIndex)
+            guard stepIndex >= 0 && stepIndex < leg.steps.endIndex else { return }
             currentStepProgress = RouteStepProgress(step: currentStep)
         }
     }
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -17,7 +17,10 @@ open class RouteProgress: NSObject {
      */
     @objc public var legIndex: Int {
         didSet {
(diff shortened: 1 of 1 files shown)


JThramer:
I'm not sure I agree with this change. I think a strong argument could be made that it's the developer's responsibility to do basic sanity checks before manipulating the model. If we've messaged you that "hey, the user arrived at waypoint X" (denoting that a leg has completed), it should be trivial to check "Hey, is this my final destination (last leg of route)? or an intermediate destination?"

An `assert()` seems pretty fair to me, as it will crash on debug (alerting developers to issues entirely within their domain of control), but will still fall through in production (where I suppose an argument could be made to add a `guard` _after_ the `assert` to avoid a crash in a release app)

But yeah, this seems like a slippery slope to me. If we start getting defensive about manipulating our model, where does that take us? do we put `guard` statements around every property in every model class we have?

I should call out that the narrative of the initial ticket is somewhat confusing. Did the user run into a bug? or do they just not like the pattern that they called out? If there's an issue in `RouteController.updateStepProgress(for:)`, or `RouteController.advanceStepIndex(to:)`, we should _definitely_ not merge this and fix the base issue instead.

We should discuss this as a team before merging. 

/cc @mapbox/navigation-ios @1ec5

hervouinc:
Hi @JThramer.
Indeed, it crashes in release since the stepIndex is used to access the currentStep in the step array, and stepIndex is out of bounds.

The behaviour we have in our use case is:
- the user goes over the last waypoint of the route
- we choose to not reroute when asked by the routeController to its delegate, at least for a while (based on several local properties) since we know the user can stop after the waypoint
- the user's location is still handled by the routeController and leads to `updateRouteStepProgress` and then `advanceStepIndex`. 

In my opinion, `updateRouteStepProgress` or at least  `advanceStepIndex` should not be called in that situation.

bsudekum:
@hervouinc this information is helpful. Let me see if i can reproduce the assertion locally.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -17,7 +17,10 @@ open class RouteProgress: NSObject {
      */
     @objc public var legIndex: Int {
         didSet {
-            assert(legIndex >= 0 && legIndex < route.legs.endIndex)
+            guard legIndex >= 0 && legIndex < route.legs.endIndex else {
+                print("legIndex - \(legIndex) out of range.")
```
1ec5: In addition to @JThramer’s comments, it should be emphasized that `print()` is no substitute for proper error handling. If not for a few existing occurrences that allow certain features to be optional, I’d be in favor of a linter rule against `print()`.

bsudekum:
Going to go about this by fixing the root problem and keeping the assertion. Incrementing the leg/step index beyond the number of steps is incorrect and should not occur.

PR #1084, opened 2018-02-02
Remove and nil mapView when presenting NavigationViewController

bsudekum:
In the example app, we keep around the preview map view after presenting the NavigationViewController. This map is unnecessary to keep around and uses up a far amount of memory. This PR shows how it can be removed and added and adds some documentation around this concept. In my tests, this frees up around 100mb.

/cc @mapbox/navigation-ios

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -19,15 +19,16 @@ enum ExampleMode {
 class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDelegate, VoiceControllerDelegate {
 
     // MARK: - IBOutlets
-    @IBOutlet weak var mapView: NavigationMapView!
     @IBOutlet weak var longPressHintView: UIView!
 
     @IBOutlet weak var simulationButton: UIButton!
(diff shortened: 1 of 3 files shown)


Examples/Swift/ViewController.swift
```diff
@@ -19,15 +19,16 @@ enum ExampleMode {
 class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDelegate, VoiceControllerDelegate {
 
     // MARK: - IBOutlets
-    @IBOutlet weak var mapView: NavigationMapView!
```
1ec5: @bsudekum, it isn’t necessary to take the map view out of the storyboard in order to add and remove it from the view hierarchy. After all, an object can be managed by the storyboard without even being part of the view hierarchy (by putting it in the “dock” above a scene).

We should put the map view back in the storyboard, since we’re already using a storyboard to manage the rest of the example application’s interface. Make the outlet strong, so that the view controller shares ownership of the map view and the map view won’t go away as soon as we remove it from the view hierarchy.

PR #1085, opened 2018-02-05
Move UI updates methods to class

bsudekum:
This PR should make it easier for developers who are creating their own UI to update the top banner components accordingly. 

todo:

- [x] figure out how to deal with next and lanes view logic
- [x] add tests

/cc @mapbox/navigation-ios


MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -92,4 +92,18 @@ open class BaseInstructionsBannerView: UIControl {
...
+    /**
+     Updates the instructions banner for a given `RouteProgress`.
+     */
+    public func update(for routeProgress: RouteProgress) {
+        let stepProgress = routeProgress.currentLegProgress.currentStepProgress
+        let distanceRemaining = stepProgress.distanceRemaining
+        
+        guard let visualInstruction = routeProgress.currentLegProgress.currentStep.instructionsDisplayedAlongStep?.last else { return }
+        
+        set(visualInstruction.primaryTextComponents, secondaryInstruction: visualInstruction.secondaryTextComponents)
```
1ec5: Argh, this `set(_:secondaryInstruction:)` method again. Can we break it out into two properties that can be set independently? If we really want to prevent one from being set without the other, then we should replace `set(_:secondaryInstruction:)` with a `update(for:)` method that takes a single VisualInstruction.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -92,4 +92,18 @@ open class BaseInstructionsBannerView: UIControl {
         
         distance = 100
     }
+    
+    /**
+     Updates the instructions banner for a given `RouteProgress`.
+     */
+    public func update(for routeProgress: RouteProgress) {
```
1ec5: Looks like `routeProgress` is used only for its `currentLegProgress`, so make this method take a RouteLegProgress instead.

JThramer:
> figure out how to deal with next and lanes view logic

My suggestion here (and the original idea behind #993) is to encapsulate the logic that determines whether the next / lanes / status / whatever UI component should show in _it's own view container._ That way we have a nice encapsulated class that does nothing but determine the visibility of it's children, and users who want _only_ the "top banner" part of our UI have a nice clean class to hook-up to their custom UI.

Cartfile
```diff
@@ -1,5 +1,5 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 3.7
-github "mapbox/MapboxDirections.swift" ~> 0.16.0
+github "mapbox/MapboxDirections.swift" "banner-types"
```
bsudekum: @frederoni noting, I updated this branch.

MapboxNavigation/StepsViewController.swift
```diff
@@ -196,35 +196,30 @@ extension StepsViewController: UITableViewDataSource {
...
     func updateCell(_ cell: StepTableViewCell, at indexPath: IndexPath) {
         let step = sections[indexPath.section][indexPath.row]
-        
-        cell.instructionsView.maneuverView.step = step
        
         let usePreviousLeg = indexPath.section != 0 && indexPath.row == 0
-        let leg = routeProgress.route.legs[indexPath.section]
-        let arrivalSecondaryInstruction = leg.destination.name
         
         if usePreviousLeg {
```
bsudekum: @JThramer did a nice refactor of this if statement in https://github.com/mapbox/mapbox-navigation-ios/pull/1063/files#diff-05dcf05089a2947433b80faae897f68dR207. Wonder if we could reuse this.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -80,16 +81,39 @@ open class BaseInstructionsBannerView: UIControl {
...
-        primaryLabel.instruction = [VisualInstructionComponent(type: .destination, text: "Primary text label", imageURL: nil)]
+        primaryLabel.instruction = [VisualInstructionComponent(type: .destination, text: "Primary text label", imageURL: nil, maneuverType: .none, maneuverDirection: .none)]
         
         distance = 100
     }
+    
+    /**
+     Updates the instructions banner for a given `RouteProgress`.
+     */
+    public func update(for currentLegProgress: RouteLegProgress) {
```
JThramer: Since this method only cares about `currentLegProgress.currentStepProgress`, it would be better if this method took a `RouteStepProgress` argument directly.

Dealer's choice though, I will be tackling this (and things like this) directly when I resurrect #993.

MapboxNavigation/LanesView.swift
```diff
@@ -80,6 +84,30 @@ public class LanesView: UIView {
...
+        } else {
+            hide()
+        }
+    }
+    
+    public func show(animated: Bool = true) {
+        guard isHidden == true else { return }
+        if animated {
+            UIView.defaultAnimation(0.3, animations: {
+                self.isHidden = false
```
JThramer: This doesn't work for animation. You have to animate the `alpha` channel, not the boolean.

Fun fact: `isHidden` automatically sets to `true` when the `alpha` value has sufficiently approached `0`.

MapboxNavigation/ManeuverView.swift
```diff
@@ -21,26 +21,26 @@ public class ManeuverView: UIView {
         }
     }
     
-    @objc public var step: RouteStep? {
+    @objc public var isStart = false {
```
JThramer: These property names aren't super-descriptive.

MapboxNavigation/LanesView.swift
frederoni: Hidden is animatable if the super view is a UIStackView and auto layout is being used. (https://medium.com/@nrewik/easy-animation-with-uistackview-8878b2856ae2)

bsudekum: Ah I guess that makes sense.

JThramer: Yeah, TIL!

PR #1088, opened 2018-02-06
Expose overhead camera view func

bsudekum:
This exposes the function which fits the camera view overhead from the users current location to the end of the route. 

The use case here is a developer who is using MapboxCoreNavigation. On a progress update while the user has enabled overhead view, the dev would call this function.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -1016,6 +1016,9 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         }
     }
     
+    /**
+     Sets the camera directly over a series of coordinates.
+     */
     @objc public func setOverheadCameraView(from userLocation: CLLocationCoordinate2D, along coordinates: [CLLocationCoordinate2D], for bounds: UIEdgeInsets) {
         let slicedLine = Polyline(coordinates).sliced(from: userLocation).coordinates
         let line = MGLPolyline(coordinates: slicedLine, count: UInt(slicedLine.count))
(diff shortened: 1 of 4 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -1011,6 +1011,22 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             style.addLayer(symbol)
         }
     }
+    
+    @objc public func enableOverheadCameraView(from userLocation: CLLocationCoordinate2D, along coordinates: [CLLocationCoordinate2D], for bounds: UIEdgeInsets) {
```
frederoni: `enable` indicates that the behavior is persistent but it's just a one-time update of the coordinate bounds, right? How about `update` or `set`?

bsudekum:
@frederoni updated.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1015,6 +1015,25 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        let line = MGLPolyline(coordinates: slicedLine, count: UInt(slicedLine.count))
+        
+        tracksUserCourse = false
+        let camera = self.camera
+        camera.pitch = 0
+        camera.heading = 0
+        self.camera = camera
+        
+        // Don't keep zooming in
+        guard line.overlayBounds.ne.distance(to: line.overlayBounds.sw) > 200 else { return }
```
JThramer: Could we encapsulate this magic number in a constant somewhere?

bsudekum: Sure

PR #1089, opened 2018-02-07
Update Danish, Ukrainian translations

1ec5:
Updated Danish, Ukrainian translations from Transifex.

/cc @frederoni @Andygol


1ec5:
Normally I don’t add changelog entries for routine translation updates, but we can start doing that if you think it matters to developers.

PR #1092, opened 2018-02-08
remove outdated text

bsudekum:
This text is sort of out of place and not necessary. We also have https://github.com/mapbox/mapbox-navigation-ios/blob/master/custom-navigation.md

/cc @1ec5

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,9 @@ This library relies heavily on notifications for letting the developer know when
 * Emitted when the user is off the route and should be rerouted. Notification contains 1 key:
   * `RouteControllerNotificationShouldRerouteKey` - `CLLocation` - Last location of user
 
+
+## Building your own custom navigation UI
+
 Looking for a more advanced use case? See our installation guide of [MapboxCoreNavigation](./custom-navigation.md).
 
 ## Language support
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -135,30 +135,6 @@ NavigationViewController(for: route, styles: [CustomStyle()])
 * `navigationViewController:didRerouteAlongRoute:`: Fired as soon as the SDK receives a new route.
 * `navigationViewController:didFailToRerouteWithError:`: Fired when SDK receives an error instead of a new route.
 
-## Building your own custom navigation UI
-
-If you need additional flexibility, you can use the following building blocks to build your own custom navigation UI:
-
-* [Interactive map SDK for iOS](https://www.mapbox.com/ios-sdk/) and [macOS](https://mapbox.github.io/mapbox-gl-native/macos/)
-* [Mapbox Studio](https://www.mapbox.com/studio/)
-  * Design custom maps with live traffic overlays
```
1ec5: I think the readme should continue to link to the map SDK and Studio in some capacity, since we do want applications to use the map SDK for all their map needs, not just navigation.

1ec5:
> Let’s link out to custom-navigation.md somewhere in the readme.

It’s a little undiscoverable, but I guess we already link “MapboxCoreNavigation” to that document.

README.md
bsudekum: @1ec5 let's find a new home for this language somewhere else then. It doesn't make sense here.

PR #1095, opened 2018-02-08
Allow developers to filter in locations

bsudekum:
This PR makes it so developers can inspect locations before they are thrown out and conditionally keep the location if they so choose.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -34,16 +34,16 @@ public protocol RouteControllerDelegate: class {
     optional func routeController(_ routeController: RouteController, willRerouteFrom location: CLLocation)
 
     /**
-     Called when a location has been discarded for being inaccurate.
+     Called when a location has been idenetified as unqualified to navigate on.
 
      See `CLLocation.isQualified` for more information about what qualifies a location.
 
      - parameter routeController: The route controller that discarded the location.
-     - parameter location: The location that was discarded
+     - parameter location: The location that will be discarded.
      - return: If `true`, the location is discarded and the `RouteController` will not consider it. If `false`, the location will not be thrown out.
      */
-    @objc(routeController:didDiscardLocation:)
-    optional func routeController(_ routeController: RouteController, didDiscard location: CLLocation) -> Bool
+    @objc(routeController:shouldUpdateToLocation:)
+    optional func routeController(_ routeController: RouteController, shouldUpdateTo location: CLLocation) -> Bool
 
     /**
      Called immediately after the route controller receives a new route.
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -40,9 +40,10 @@ public protocol RouteControllerDelegate: class {
 
      - parameter routeController: The route controller that discarded the location.
      - parameter location: The location that was discarded
+     - return: If `true`, the location is discarded and the `RouteController` will not consider it. If `false`, the location will not be thrown out.
      */
     @objc(routeController:didDiscardLocation:)
-    optional func routeController(_ routeController: RouteController, didDiscard location: CLLocation)
+    optional func routeController(_ routeController: RouteController, didDiscard location: CLLocation) -> Bool
```
1ec5: This is awkward:

* Delegator: I did discard this location!
* Delegate: No.
* Delegator: OK, on second thought, I did not discard the location!

Instead, let’s replace this method with `routeController(_:shouldMoveTo:)`:

* Delegator: Should I move to this location?
* Delegate: No.
* Delegator: OK, I won’t move to this location.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -163,6 +163,17 @@ public protocol NavigationViewControllerDelegate {
      Returns the center point of the user course view in screen coordinates relative to the map view.
      */
     @objc optional func navigationViewController(_ navigationViewController: NavigationViewController, mapViewUserAnchorPoint mapView: NavigationMapView) -> CGPoint
+    
+    /**
+     Called when a location has been discarded for being inaccurate.
+     
+     See `CLLocation.isQualified` for more information about what qualifies a location.
```
1ec5: Is this enough context for the developer to use the method? Or do we need to provide some sort of reason indicating why it’s a candidate for discarding? Without providing a reason, I think the developer only has a choice between discarding all the unqualified locations or including them all, in which case a property would be more convenient than a delegate method.

bsudekum: What would you suggest, some sort of enum for each heuristic we investigate?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -528,9 +539,21 @@ extension NavigationViewController: RouteControllerDelegate {
...
-    @objc public func routeController(_ routeController: RouteController, didDiscard location: CLLocation) {
-        let title = NSLocalizedString("WEAK_GPS", bundle: .mapboxNavigation, value: "Weak GPS signal", comment: "Inform user about weak GPS signal")
-        mapViewController?.statusView.show(title, showSpinner: false)
+    @objc public func routeController(_ routeController: RouteController, shouldUpdateTo location: CLLocation)  -> Bool {
+        func showWeakGPS() {
+            let title = NSLocalizedString("WEAK_GPS", bundle: .mapboxNavigation, value: "Weak GPS signal", comment: "Inform user about weak GPS signal")
+            mapViewController?.statusView.show(title, showSpinner: false)
+        }
+        
+        if let shouldUpdateToIsImplemented = delegate?.navigationViewController?(self, shouldUpdateTo: location), shouldUpdateToIsImplemented {
```
JThramer:
```swift

let shouldUpdate = delegate?.navigationViewController?(self, shouldUpdateTo: location) ?? true //default to yes if unimplmented

if (shouldUpdate) {
    showWeakGPS()
    return true
}
return false
```

bsudekum:
@1ec5 On second thought, I think this needs to be a little more clear that this is called when discarding bad locations. How about calling this `shouldDiscardLocation` instead?

PR #1101, opened 2018-02-12
Tweak course tracking

frederoni:
A few minor changes to the course tracking:

- Expose `NavigationMapViewCourseTrackingDelegate`
- Omit animations when updating user puck on a frame by frame basis.
- Remove the previous course tracking view when adding a new one and update the view instantly.

@bsudekum

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -307,7 +307,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             let padding = UIEdgeInsets(top: point.y, left: point.x, bottom: bounds.height - point.y, right: bounds.width - point.x)
             let newCamera = MGLMapCamera(lookingAtCenter: location.coordinate, fromDistance: altitude, pitch: 45, heading: location.course)
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear) : nil
-            let duration: TimeInterval = animated ? 1 : 0
+            let duration: TimeInterval = animated && !shouldPositionCourseViewFrameByFrame ? 1 : 0
             setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding, completionHandler: nil)
         }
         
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -359,12 +359,12 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     open var tracksUserCourse: Bool = false {
         didSet {
             if tracksUserCourse {
-                enableFrameByFrameCourseViewTracking(for: 3)
+                enableFrameByFrameCourseViewTracking(for: 2)
```
bsudekum: Just curious what this change means

MapboxNavigation/NavigationMapView.swift
```diff
@@ -267,7 +267,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
     
     public weak var navigationMapDelegate: NavigationMapViewDelegate?
-    weak var courseTrackingDelegate: NavigationMapViewCourseTrackingDelegate!
+    public weak var courseTrackingDelegate: NavigationMapViewCourseTrackingDelegate!
```
bsudekum: Want to add little documentation here?

MapboxNavigation/NavigationMapView.swift
frederoni: There's no way to figure out when a gesture finishes with velocity so this enables frame by frame tracking for 2 seconds which means that the user puck will update on every `-[MGLMapView(MGLNavigationAdditions) mapViewDidFinishRenderingFrameFullyRendered:]`. We might be able to simplify this logic when https://github.com/mapbox/mapbox-gl-native/pull/11151 lands.

1ec5: One approach would be to implement `mapView:shouldChangeFrom:to:` to capture when a camera change occurs due to a gesture (since that method only gets called in response to gestures) or set that state in a completion handler; for non-gesture-related camera changes, the velocity animation finishes when `mapView:regionDidChange:animated:` is called.

Issue #1100, opened 2018-02-11
Consider a location qualified when speed is less than 0


bsudekum:
https://github.com/mapbox/mapbox-navigation-ios/blob/f72682186e8bca7ff75b5365bc3b6ba2a83f6e70/MapboxCoreNavigation/CLLocation.swift#L14

We currently discard speeds that -1 even if the location accuracy is still high. I think we should rethink because ["_A negative value indicates an invalid speed_"](https://developer.apple.com/documentation/corelocation/cllocation/1423798-speed) but not necessarily incorrect coordinates. I'd say that most discarded locations are because the speed requirement is not met. In the case the speed is inaccurate, we should not use it when doing things like dead reckoning, but we should still snap, move the user along the route and also consider reroutes.

/cc @mapbox/navigation-ios @mapbox/navigation-android

bsudekum:
I think any loosening of the speed value should also be met with a tightening of the horizontal and vertical accuracy:

https://github.com/mapbox/mapbox-navigation-ios/blob/f72682186e8bca7ff75b5365bc3b6ba2a83f6e70/MapboxCoreNavigation/CLLocation.swift#L12-L13

PR #1103, opened 2018-02-12
 Remove speed from isQualified check

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1100

This removes `speed` from checking whether a location is qualified or not. Since speed is not necessary to know whether a user is on the route, where to snap, or if they have completed a maneuver, we can investigate more meaningful clues like the horizontal accuracy. 

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -9,7 +9,7 @@ extension CLLocation {
             return true
         #else
             return
-                0...80 ~= horizontalAccuracy &&
+                0...100 ~= horizontalAccuracy &&
                 0...30 ~= horizontalAccuracy
         #endif
     }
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -10,8 +10,7 @@ extension CLLocation {
         #else
             return
                 0...100 ~= horizontalAccuracy &&
-                verticalAccuracy > 0 &&
-                speed >= 0
+                0...30 ~= horizontalAccuracy
```
frederoni: Something went wrong here. Between 0-100 and between 0-30 for the same horizontal accuracy.

bsudekum: oof my bad, will fix.

Issue #1026, opened 2018-01-17
Allow overriding speak() function for voice synthesis


ericrwolfe:
This would offer developers who want to announce custom events in their app a way to easily achieve a voice consistent with navigation.

cc @mapbox/navigation-ios @mapbox/navigation-android

1ec5:
`PollyVoiceController.speak(_:)` is currently `public`, but it should be `open` so that a custom subclass of PollyVoiceController can override this method and pass a modified SpokenInstruction to our superclass implementation.

bsudekum:
This has been fixed.

PR #1104, opened 2018-02-13
Modified PollyVoiceController's speak method qualifier

vincethecoder:
Changes made to meet the expected behavior proposed in - [Issue #1026: Allow overriding speak() function for voice synthesis](https://github.com/mapbox/mapbox-navigation-ios/issues/1026)

**Intent**: The `open` access level applied to the `speak` method allows the method to be overridden by subclasses of the  `PollyVoiceController`class, which wasn't attainable with the previous `public` access level.

cc @mapbox/navigation-ios @1ec5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master
 
 * Exposes `setOverheadCameraView(from:along:for:)` which is useful for fitting the camera to an overhead view for the remaining route coordinates.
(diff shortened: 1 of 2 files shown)


frederoni:
@Sealos Both RouteVoiceController and MapboxVoiceController are open. Is there any subclassing behavior that you're missing?

Sealos:
I didn't notice that change, perfect then!

Issue #987, opened 2018-01-09
Expose currentLegAttribute


bsudekum:
https://github.com/mapbox/mapbox-navigation-ios/blob/06d5bc708e4079df25187eb56e43b7954222352a/MapboxNavigation/NavigationMapView.swift#L62

It's difficult to style non-active legs without knowing this string. Ref implementation: 
https://github.com/mapbox/mapbox-navigation-ios/blob/06d5bc708e4079df25187eb56e43b7954222352a/MapboxNavigation/NavigationMapView.swift#L931-L934

/cc @mapbox/navigation-ios

1ec5:
Can NavigationMapView expose a more targeted way of styling legs based on whether they’re active?

PR #1105, opened 2018-02-13
Make necessary strings public

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/987

This exposes the strings that are used in any `attributeName` so the developer can style legs and congestion values however they like.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -858,7 +858,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             
             let lines = mergedCongestionSegments.map { (congestionSegment: CongestionSegment) -> MGLPolylineFeature in
                 let polyline = MGLPolylineFeature(coordinates: congestionSegment.0, count: UInt(congestionSegment.0.count))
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -69,12 +69,21 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
     let instructionSource = "instructionSource"
     let instructionLabel = "instructionLabel"
     let instructionCircle = "instructionCircle"
     let alternateSourceIdentifier = "alternateSource"
     let alternateLayerIdentifier = "alternateLayer"
+    
+    /**
+     Attribute name for the route line that is used for identifying whether a RouteLeg is the current active leg.
+     */
+    public let currentLegAttribute = "isCurrentLeg"
```
1ec5: These constants should be global, so that they get bridged to Objective-C as constants rather than properties on NavigationMapView.

bsudekum: Yeah, these should be moved to the new constants.swift file.

PR #1109, opened 2018-02-13
Add constants file for MapboxNavigation

bsudekum:
The constant `NavigationMapViewMinimumDistanceForOverheadZooming` in https://github.com/mapbox/mapbox-navigation-ios/pull/1088/files#diff-1e0aeb18d31c8c1abfce278ab3489151R170 made it so the the constants file in MapboxCoreNavigation be added to a project. This goes against what is in our readme.

This change adds a new constants file specific to MapboxNavigation and makes it so you can just add MapboxNavgiation to your podfile. More can be moved in here probably.

/cc @mapbox/navigation-ios


frederoni:
IIRC, there was an issue with having two files with the same name when integrating two podspecs from the same root folder. I might be mistaken but please verify integrating this change using CocoaPods before merging.

bsudekum:
Ah interesting, will check it out.

bsudekum:
Tested in another project, no issues.

Issue #1058, opened 2018-01-22
Don't allow for unsnapped user puck


bsudekum:
When snapping the user puck, we allow the puck to remain unsnapped but not rerouted here: https://github.com/mapbox/mapbox-navigation-ios/blob/fe5d5990adb4cb603265cb97d802f1cd9b2e995e/MapboxCoreNavigation/RouteController.swift#L401-L409

But in all cases that this code fires, the user will be rerouted soon. If we were to move this logic to the reroute check, we could reroute users even sooner and with a higher degree of certainty.

/cc @mapbox/navigation-ios

PR #1110, opened 2018-02-14
Refactor snapped location

bsudekum:
This cleans up the snapped location logic and pulls it into smaller functions.

This also changes the code which handles unsnapped locations. Now, it is slightly easier to become unsnapped. If the follow conditions are true:

* Your horizontal accuracy is very low (< 20)
* Your actual course is not -1 and it is different than what we think it should be
* You are moving

These heuristics are also great for rerouting and could potentially make rerouting occur sooner. Instead of unsnapping your location before rerouting, we would just reroute you as soon as we see you are actively not on the route. Will add this in a followup PR. (Ref https://github.com/mapbox/mapbox-navigation-ios/issues/1058)

/cc @mapbox/navigation-ios


MapboxCoreNavigation/RouteController.swift
```diff
@@ -342,36 +342,53 @@ open class RouteController: NSObject {
...
+        var userCoordinate = closest.coordinate
+        if shouldAllowUnsapped(location: location, given: calculatedCourseForLocationOnStep) {
+            userCourse = location.course
+            userCoordinate = location.coordinate
+        }
+        
+        return CLLocation(coordinate: userCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: userCourse, speed: location.speed, timestamp: location.timestamp)
+    }
+    
+    func turnAngleStraightness(angleIn: CLLocationDirection, angleOut: CLLocationDirection) -> CLLocationDirection {
```
1ec5: This would be a good candidate for upstreaming to Turf. As I understand it, the only difference between this method and `CLLocationDirection.difference(from:)` is that the latter takes the short way around, always returning a positive value.

Regardless, let’s extract this method from RouteController, because it doesn’t have any dependencies on `self`. Put it in it an extension of `CLLocationDirection`, as an instance method named `clockwiseDifference(from:)` – assuming it actually does measure clockwise, as implied by `CLLocationDirection`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -392,22 +409,20 @@ open class RouteController: NSObject {
...
-                    userCoordinate =  location.coordinate
-                }
-            }
+        return (wrappedCourse + averageRelativeAngle).wrap(min: 0, max: 360)
+    }
+    
+    /**
+     Determines if the a location is qualified enough to allow the user puck to become unsnapped.
+     */
+    func shouldAllowUnsapped(location: CLLocation, given snappedCourse: CLLocationDirection) -> Bool {
```
1ec5: Typo: “unsnapped”. A more grammatical name would be `shouldSnap(_:facing:)`. “Given” isn’t a word that appears in Swift method names. Note that the logic would be inverted, since positive conditions are easier to reason about than negative conditions.

bsudekum:
@mapbox/navigation-ios this is ready for review.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -154,6 +154,7 @@
 		C57607B21F4CC9E800C27423 /* Solar.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = C57607B01F4CC97D00C27423 /* Solar.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		C578DA081EFD0FFF0052079F /* ProcessInfo.swift in Sources */ = {isa = PBXBuildFile; fileRef = C578DA071EFD0FFF0052079F /* ProcessInfo.swift */; };
 		C58159011EA6D02700FC6C3D /* MGLVectorSource.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58159001EA6D02700FC6C3D /* MGLVectorSource.swift */; };
+		C582FD5F203626E900A9086E /* CLLocationDirections.swift in Sources */ = {isa = PBXBuildFile; fileRef = C582FD5E203626E900A9086E /* CLLocationDirections.swift */; };
```
1ec5: Typo: CLLocationDirection

MapboxCoreNavigation/CLLocationDirections.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+import CoreLocation
+
+extension CLLocationDirection {
+    func clockwiseDifference(from toDirection: CLLocationDirection) -> CLLocationDirection {
```
1ec5: It’s a bit confusing to have an argument called “from” externally but “to” internally. Let’s call this angle `otherDirection`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -434,22 +440,20 @@ open class RouteController: NSObject {
...
-                    userCoordinate =  location.coordinate
-                }
-            }
+        return (wrappedCourse + averageRelativeAngle).wrap(min: 0, max: 360)
+    }
+    
+    /**
+     Determines if the a location is qualified enough to allow the user puck to become unsnapped.
+     */
+    func shouldSnap(_ location: CLLocation, facing snappedCourse: CLLocationDirection) -> Bool {
```
1ec5: I was confused earlier: `snappedCourse` is the _road_’s course, not the user’s course (which is already part of `location`). In that case, let’s rename this method to `shouldSnap(_:routeCourse:)`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -384,36 +384,42 @@ open class RouteController: NSObject {
...
+        }
+        
+        return CLLocation(coordinate: userCoordinate, altitude: location.altitude, horizontalAccuracy: location.horizontalAccuracy, verticalAccuracy: location.verticalAccuracy, course: userCourse, speed: location.speed, timestamp: location.timestamp)
+    }
+    
+    /**
+     Given a location and a series of coordinates, compute what the course should be for a the location.
+     */
+    func interpolatedCourse(from location: CLLocation, along coordinates: [CLLocationCoordinate2D]) -> CLLocationDirection? {
+        let nearByPolyline = Polyline(coordinates)
```
JThramer: We're initializing `Polyline(coordinates)` three times here. Let's take advantage of the local variable that already exists.

PR #1111, opened 2018-02-14
Localize map labels

### LINKED ISSUE #755: Localize the map
The Mapbox Navigation Guidance Day and Night styles show English labels everywhere. That makes little sense in non-English-speaking countries. As soon as the style finishes loading, NavigationMapView should either enable `MGLStyle.localizesLabels` or manually switch all the labels to `{name}`, to match the local language.

I’m sure we should move off English, but I’m not sure which approach to take. It’s important for users to see labels that match signage. OSRM doesn’t currently localize street names (Project-OSRM/osrm-backend#4561), so any labels in a non-local language would create confusion in the presence of differing banner text and spoken instructions. On the other hand, labels in the local language may not be in the same writing system that the user is familiar with.

/cc @mapbox/navigation-ios @captainbarbosa @mapbox/language

1ec5:
Added a public `localizeLabels()` method to NavigationMapView for localizing all the labels: road labels are localized into the local language (the `name` tag in OpenStreetMap, matching the Mapbox Directions API), while place and POI labels are localized into the system’s preferred language if supported by the Mapbox Streets source, falling back to the local language.

The map view embedded in NavigationViewController calls this method automatically, so no developer intervention is typically required for the turn-by-turn map. However, the developer is encouraged to call `localizesLabels()` on any standalone NavigationMapView from within `MGLMapViewDelegate.mapView(_:didFinishLoading:)`.

This change required porting about half of the `MGLStyle.localizesLabels` implementation from Objective-C to Swift. `MGLStyle.localizesLabels` isn’t quite adequate, because it’s a two-way option between the style’s original language (English in the case of the Mapbox Navigation Guidance styles) and the system’s preferred language, with no option to use the local language that matches signage. This map SDK API will be revamped as part of mapbox/mapbox-gl-native#10713.

preview turn-by-turn

Fixes #755.

/cc @bsudekum @frederoni @nickidlugash

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -316,6 +316,8 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     }
     
     func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        self.mapView?.localizeLabels()
+        
         if let routes = routes, let coords = currentRoute?.coordinates, let coordCounts = currentRoute?.coordinateCount {
             mapView.setVisibleCoordinateBounds(MGLPolygon(coordinates: coords, count: coordCounts).overlayBounds, animated: false)
             self.mapView?.showRoutes(routes)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Exposes `setOverheadCameraView(from:along:for:)` which is useful for fitting the camera to an overhead view for the remaining route coordinates.
 * Exposes `PollyVoiceController.speak(_:)` which would allow custom subclass of PollyVoiceController to override this method and pass a modified SpokenInstruction to our superclass implementation.
+* Added a `NavigationMapView.localizeLabels()` method that should be called within `MGLMapViewDelegate.mapView(_:didFinishLoading:)` for standalone `NavigationMapView`s to ensure that map labels are in the correct language. (#1111)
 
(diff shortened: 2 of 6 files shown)


Examples/Swift/ViewController.swift
```diff
@@ -316,6 +316,8 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
     }
     
     func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        self.mapView?.localizeLabels()
```
bsudekum: Is self needed here?

1ec5: Yes, because `localizesLabels()` is defined on NavigationMapView, not MGLMapView. The MGLMapView-typed `mapView` argument takes precedence, so we have to either use the NavigationMapView-typed property or cast the argument to a NavigationMapView.

1ec5:
Road names are never localized in this SDK; that is, they always appear in the local language and are never tailored to the user’s preferred language. In part that’s because it’s important to match what the user sees out the window on signage while driving, whatever the user’s language. But it’s also because OSRM currently only provides the local names in the route data (Project-OSRM/osrm-backend#4561), which is used for the turn banner, so we make sure the map matches the turn banner for consistency.

If you want to localize road labels on the map into English and are OK with these labels potentially differing from what’s in the turn banner and out the window, pass the `en` locale into `MGLStyle.localizeLabels(into:)` inside your implementation of `MGLMapViewDelegate.mapView(_:didFinishLoading:)`.

Issue #1120, opened 2018-02-14
Voice instruction is repeated before rerouting


bsudekum:
This has been occurring since the voice change. Investigating.

PR #1121, opened 2018-02-15
Fix double voice

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1120

This fixes a bug where voice instructions were given twice when rerouting and also moves logic into it's own function. Very little has changed in this PR, mostly moving things around.

The issue came with the check we run before rerouting to see if the user is on a future step. The bug is that we were actually checking the current step and all future steps. Instead, we should only check future steps. When rerouting we'd check if they were on the current step or future step, the func would return the current step and then we'd increment to the current step. Incrementing caused a chain reaction of events to fire including speaking the first instruction of the current step 🙄 .

/cc @mapbox/navigation-ios


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -301,8 +301,10 @@ open class RouteLegProgress: NSObject {
     typealias StepIndexDistance = (index: Int, distance: CLLocationDistance)
     
     func closestStep(to coordinate: CLLocationCoordinate2D) -> StepIndexDistance? {
+        guard stepIndex + 1 < leg.steps.endIndex else { return nil }
+        
         var currentClosest: StepIndexDistance?
-        let remainingSteps = leg.steps.suffix(from: stepIndex)
+        let remainingSteps = leg.steps.suffix(from: stepIndex + 1)
```
bsudekum: This is the actual fix.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -854,19 +857,20 @@ extension RouteController: CLLocationManagerDelegate {
...
-        guard let spokenInstructions = routeProgress.currentLegProgress.currentStepProgress.remainingSpokenInstructions else {
-            print("The directions request was made without `includesVoiceInstructions` enabled. This will prevent users from getting voice instructions. It's recommended to make your directions request via `NavigationRouteOptions()`.")
-            return
-        }
-
+    }
+    
+    func updateSpokenInstructionProgress(for location: CLLocation) {
+        guard let coordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return }
+        let userSnapToStepDistanceFromManeuver = Polyline(coordinates).distance(from: location.coordinate)
```
1ec5: We’re calculating this distance at least three other times on each location update:

https://github.com/mapbox/mapbox-navigation-ios/blob/ebcfcea8bb2502c0f102dca226c8fc022fff96cd/MapboxCoreNavigation/RouteController.swift#L564
https://github.com/mapbox/mapbox-navigation-ios/blob/ebcfcea8bb2502c0f102dca226c8fc022fff96cd/MapboxCoreNavigation/RouteController.swift#L826
https://github.com/mapbox/mapbox-navigation-ios/blob/ebcfcea8bb2502c0f102dca226c8fc022fff96cd/MapboxCoreNavigation/RouteController.swift#L588

Can we consolidate this code into a single calculation that gets persisted on the route progress somehow?

bsudekum:
@1ec5 This is ready for review.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -223,6 +223,12 @@ open class RouteController: NSObject {
     var recentDistancesFromManeuver: [CLLocationDistance] = []
     
     var previousArrivalWaypoint: Waypoint?
+    
+    var userSnapToStepDistanceFromManeuver: CLLocationDistance? {
```
1ec5: Performance-wise, this is no different than before, because this value has to get computed each time the variable is accessed. My suggestion in https://github.com/mapbox/mapbox-navigation-ios/pull/1121#discussion_r168397149 was to compute the distance just once, then either plumb it through to each helper method that needs it or set it on a (stored) property of RouteStepProgress, where the helper methods can access it without any performance penalty.

JThramer: My preference would be to keep the `RouteStep`'s coordinates in a stored `polyline` property, and then have `RouteStep.coordinates` be a computed property that points to the polyline's coordinates.

bsudekum: I agree, however the one pitfall with this approach is that it is one more thing to make sure is updated on every progress update.

bsudekum: @1ec5 @JThramer fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/1121/commits/14ebaa60d215a94a432bc5c420da89087c8fc452

MapboxCoreNavigation/RouteController.swift
```diff
@@ -357,7 +353,13 @@ open class RouteController: NSObject {
      
      This is a raw location received from `locationManager`. To obtain an idealized location, use the `location` property.
      */
-    var rawLocation: CLLocation?
+    var rawLocation: CLLocation? {
+        didSet {
+            guard let coordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return }
+            guard let coordinate = rawLocation?.coordinate else { return }
```
1ec5: If there’s no coordinate, should `userSnapToStepDistanceFromManeuver` be set to `nil`?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -351,7 +353,13 @@ open class RouteController: NSObject {
...
```
1ec5: If there’s no coordinate, should `userSnapToStepDistanceFromManeuver` be set to `nil`?

PR #1124, opened 2018-02-15
Move to constants, make a public

bsudekum:
This makes a few helpful functions public and also moves around global variables to the constants file.

/cc @mapbox/navigation-ios


MapboxNavigation/Constants.swift
```diff
@@ -1,6 +1,35 @@
 import Foundation
+import MapboxDirections
+
+typealias CongestionSegment = ([CLLocationCoordinate2D], CongestionLevel)
+
+/**
+ Line width base values used at zoom levels.
+ */
+public let lineWidthAtZoomLevels: [Int: MGLStyleValue<NSNumber>] = [
```
1ec5: Avoid polluting the global namespace in Objective-C: rename these public methods with the `MB` class prefix.

MapboxNavigation/Constants.swift
```diff
@@ -1,6 +1,35 @@
...
+    13: MGLStyleValue(rawValue: 9),
+    16: MGLStyleValue(rawValue: 11),
+    19: MGLStyleValue(rawValue: 22),
+    22: MGLStyleValue(rawValue: 28)
+]
+
+/**
+ Default opttions applied to features in NavigationMapView.
+ */
+public let sourceOptions: [MGLShapeSourceOption: Any] = [.maximumZoomLevel: 16]
```
1ec5: How is the developer expected to use this constant? Are they expected to use these options for all shape sources they create?

MapboxNavigation/Constants.swift
```diff
@@ -1,6 +1,35 @@
...
 
 /**
  The minium distance remaining on a route before overhead zooming is stopped.
  */
 public var NavigationMapViewMinimumDistanceForOverheadZooming: CLLocationDistance = 200
+
+/**
+ Attribute name for the route line that is used for identifying whether a RouteLeg is the current active leg.
+ */
+public let currentLegAttribute = "isCurrentLeg"
```
1ec5: Note that this constant will most likely go away or be replaced by something like `legIndex` in #1076.

MapboxNavigation/Dictionary.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+extension Dictionary where Key == Int, Value: MGLStyleValue<NSNumber> {
+    /**
+     Returns a proportional `Dictionary` of `lineWidthAtZoomLevels` at a specific factor.
+     */
+    public func multiplied(by factor: Double) -> Dictionary {
```
1ec5: This method will also go away as part of the migration to expressions in #1076.

MapboxNavigation/Dictionary.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+extension Dictionary where Key == Int, Value: MGLStyleValue<NSNumber> {
+    /**
+     Returns a proportional `Dictionary` of `lineWidthAtZoomLevels` at a specific factor.
```
1ec5: As written, this method could just as well operate on circle size stops, not just line width stops:

> Returns a copy of the stop dictionary with each value multiplied by the given factor.

MapboxNavigation/Constants.swift
1ec5: Also, bring back the words “route line”, because these line widths are only applicable to route line styling.

1ec5: I don’t think we should expose this constant publicly at all. The whole point of making the route line style customizable is that the developer would substitute this style value with their own.

MapboxNavigation/Constants.swift
```diff
@@ -6,19 +6,14 @@ typealias CongestionSegment = ([CLLocationCoordinate2D], CongestionLevel)
 /**
  Line width base values used at zoom levels.
  */
-public let lineWidthAtZoomLevels: [Int: MGLStyleValue<NSNumber>] = [
+public let MBRoutelineWidthAtZoomLevels: [Int: MGLStyleValue<NSNumber>] = [
```
1ec5: Nit: “route line” is two words. Also, the usual phrasing for a dictionary would be “by zoom level”. So `MBRouteLineWidthByZoomLevel`.

1ec5:
The assumption we made back when introducing `navigationMapView(_:routeStyleLayerWithIdentifier:source:)` et al. was that a developer would bring their own totally custom line layer. But if the more usual use case is to copy the built-in one and make a few tweaks, maybe what we really need to do is to publicly expose the default line layer’s ID as a constant. Then the developer can call `style.layer(withIdentifier:)` after calling `showRoutes(_:legIndex:)` to do the tweaking.

bsudekum:
@1ec5 could we go with what we have now and make some breaking changes when we pick https://github.com/mapbox/mapbox-navigation-ios/pull/1076 back up?

MapboxCoreNavigation/Dictionary.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+extension Dictionary where Key == Int, Value: MGLStyleValue<NSNumber> {
+    /**
+     Returns a copy of the stop dictionary with each value multiplied by the given factor.
+     */
+    public func multiplied(by factor: Double) -> Dictionary {
+        var newCameraStop: [Int: MGLStyleValue<NSNumber>] = [:]
+        for stop in MBRouteLineWidthByZoomLevel {
```
1ec5: Because this line refers to `MBRouteLineWidthByZoomLevel` instead of `self`, this code:

```swift
let stops = [5: MGLStyleValue(rawValue: 1)]
stops.multiplied(by: 1)
```

magically produces a dictionary equivalent to `MBRouteLineWidthByZoomLevel`, even though `MBRouteLineWidthByZoomLevel` was never used explicitly.

MapboxNavigation/Constants.swift
```diff
@@ -1,6 +1,30 @@
 import Foundation
+import MapboxDirections
+
+typealias CongestionSegment = ([CLLocationCoordinate2D], CongestionLevel)
+
+/**
+ Line width base values used at zoom levels.
```
1ec5: > A stop dictionary representing the default line widths of the route line by zoom level when `NavigationMapViewDelegate.navigationMapView(_:routeStyleLayerWithIdentifier:source:)` is undefined.
> 
> You may use this constant in your implementation of `NavigationMapViewDelegate.navigationMapView(_:routeStyleLayerWithIdentifier:source:)` if you want to keep the default line widths but customize other aspects of the route line.

MapboxCoreNavigation/Dictionary.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+extension Dictionary where Key == Int, Value: MGLStyleValue<NSNumber> {
+    /**
+     Returns a copy of the stop dictionary with each value multiplied by the given factor.
+     */
+    public func multiplied(by factor: Double) -> Dictionary {
+        var newCameraStop: [Int: MGLStyleValue<NSNumber>] = [:]
+        for stop in self as [Int : MGLStyleValue<NSNumber>] {
```
bsudekum: Unsure if I can get away with not casting here.

JThramer: The only thing that comes to mind for me is that making a `private typealias NumericStyleDictionary = [Int: MGLStyleValue<NSNumber>]` (or something) might make the code a little cleaner.

1ec5: The overall extension requires the same type conformance in its `where` condition. If `as` builds without errors (as opposed to `as!` or `as?`), then casting is fine.

bsudekum: No issues.

Issue #559, opened 2017-08-29
Enable night mode while in tunnels


ericrwolfe:
We should be able to tell if a driver's in a tunnel and automatically enable dark mode. 

However, this isn't something that's currently supported by the Directions API 

cc @bsudekum @miccolis @v8c3

daniel-j-h:
Upstream ticket for tunnels and other notifications:

https://github.com/Project-OSRM/osrm-backend/issues/3028

I can remember talking to @1ec5 about tunnels when we added the classes way property and there was an argument against tunnels but I can't remember - @1ec5 could you chime in here?

1ec5:
In this case, I don’t think we need an additional notification from OSRM. We already query the map vector tiles on each location update to get the current road’s name. #297 (which was closed but could easily be resurrected) refactors that code to give us rich information about the current road, including whether it’s in a tunnel or on a bridge.

1ec5:
> However, this isn't something that's currently supported by the Directions API

Project-OSRM/osrm-backend#4816 added a `tunnel` class. Once mapbox/MapboxDirections.swift#237 lands, we can switch to night mode as soon as the user crosses an intersection with the tunnel class and switch back to day mode as soon as they cross an intersection without the tunnel class.

bsudekum:
This shipped in https://github.com/mapbox/mapbox-navigation-ios/pull/1127

Issue #680, opened 2017-10-02
Can't find or build MapboxDirections.framework - Xcode 9 - Swift 4 - Carthage 0.25.0


addzo:
Hello.
I'm running Xcode 9 (GM), Swift 4, and Carthage 0.25.0.
I followed the directions on https://github.com/mapbox/mapbox-navigation-ios and it does not find the framework or build the framework.

I tried building all the projects and searched for the .framework on disk and it does not exist. 
<img width="262" alt="screen shot 2017-10-02 at 4 18 22 pm" src="https://user-images.githubusercontent.com/168576/31102337-de1a13a2-a78e-11e7-8b9b-ebc2693b32d3.png">

Does that framework get pulled from Carthage?
Does it get built in the project?

Thanks

frederoni:
> Does that framework get pulled from Carthage?

Yes
> Does it get built in the project?

No

What did Carthage output after running `carthage bootstrap --platform ios`?

addzo:
Ah.  Well that is interesting:

$ carthage bootstrap --platform ios
*** Downloading binary-only framework Mapbox-iOS-SDK at "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json"
*** Checking out Solar at "5ce2ca7baaa82b1abb0467642ae237faf674b645"
*** Checking out SDWebImage at "4.1.0"
*** Checking out ios-snapshot-test-case at "2.1.4"
*** Checking out Polyline at "v4.1.1"
*** Checking out osrm-text-instructions.swift at "v0.3.5"
*** Checking out MapboxDirections.swift at "v0.10.6"
*** Checking out mapbox-telemetry-ios at "v0.2.11"
*** Checking out Pulley at "1.4"
*** Checking out aws-sdk-ios at "2.6.0"
*** xcodebuild output can be found in /var/folders/6h/gjnb3qpj4zs9h1jtzf73b6640000gn/T/carthage-xcodebuild.3slBUw.log
A shell task (/usr/bin/xcrun xcodebuild -project /Users/brep/Sandbox/mapbox/mapbox-navigation-ios/Carthage/Checkouts/aws-sdk-ios/AWSiOSSDKv2.xcodeproj CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES -list) failed with exit code 72:
xcrun: error: unable to find utility "xcodebuild", not a developer tool or in PATH

[...]

Issue #629, opened 2017-09-19
Investigate dead reckoning code


bsudekum:
https://github.com/mapbox/mapbox-navigation-ios/blob/16312c8ce2d6ce8220f6bb243ec416d70c5761eb/MapboxCoreNavigation/RouteController.swift#L538-L539

Two thoughts:

1. We've learned that using `location.speed` in the past can come back to bite us when the speed is invalid. 
1. We're also looking at a point somewhere in front of the user. If there is a turn ahead and the dead reckoning time interval is large enough, the user would be rerouted.

/cc @1ec5 @frederoni @ericrwolfe

frederoni:
> We've learned that using location.speed in the past can come back to bite us when the speed is invalid.

Yes, I think we should keep a buffer and use the average speed from the 5 most recent locations after filtering out unqualified locations.

> We're also looking at a point somewhere in front of the user. If there is a turn ahead and the dead reckoning time interval is large enough, the user would be rerouted.

In the continuous dead reckoning (that triggers in a tunnel)
https://github.com/mapbox/mapbox-navigation-ios/blob/16312c8ce2d6ce8220f6bb243ec416d70c5761eb/MapboxCoreNavigation/RouteController.swift#L408
we use `coordinate(at:fromStartOf:)` so going off route shouldn't be an issue there. The problem is that we stay on route even if the user makes a wrong turn until we get a real location update.

I think it would be better to assume that the user continues on the right path instead of following the previous course.

bsudekum:
A few more heuristics we came up with last week:

* If the user is traveling at (expected) speed
* (and?) there is no congestion in the tunnel
* (and?) the tunnel is relatively short

Animate the puck through the tunnel.

1ec5:
> We're also looking at a point somewhere in front of the user. If there is a turn ahead and the dead reckoning time interval is large enough, the user would be rerouted.

Given ongoing issues that are suspected to stem from this issue, I think we should remove the dead reckoning feature until we can ensure that the dead-reckoned location still lies along the route. We had to do something similar for the puck’s visual-only dead reckoning in mapbox/mapbox-navigation-ios#1710.

1ec5:
The dead reckoning code does already attempt to snap to the existing route geometry:

https://github.com/mapbox/mapbox-navigation-ios/blob/c55c13beb15d611e4a23eb826a9a966406afd8d3/MapboxCoreNavigation/CLLocation.swift#L135-L136

[...]

Issue #110, opened 2017-03-31
Merge MapboxNavigation and MapboxNavigationUI


bsudekum:
I think we should look into merging MapboxNavigationUI into MapboxNavigation. I think one of the big reasons against this is bloat; you'd have to use AWS polly + core even if you don't use it. If this is the only standing reason, let's investigate conditionally importing these libraries.

/cc @1ec5 @frederoni

1ec5:
There are other reasons to avoid mixing low-level navigation logic with UI view controllers and their logic. As an example, it would be straightforward to modify the non-UI library to run on watchOS or use it in other use cases where a conventional turn-by-turn UI is undesirable. Blurring the lines between the navigation logic and the UI elements will inevitably make it more difficult for a developer to use just the logic but not the UI.

Making the AWS dependencies optional is an orthogonal issue that we should consider regardless, to improve the experience for applications that need to use the UI with other speech engines (say, for Chinese support).

What's the motivation behind merging the two libraries?

bsudekum:
Keeping as is.

PR #1127, opened 2018-02-17
Enable night style when inside tunnel

vincethecoder:
This detects tunnels on a `routeStep` and sets the `style` to `DayStyle`.
cc @mapbox/navigation-ios 

Issue addressed: https://github.com/mapbox/mapbox-navigation-ios/issues/559, https://github.com/mapbox/mapbox-navigation-ios/issues/1170

---
**NOTE**:
When calculating the `distance` to the tunnel entrance using `Polyline`'s `distance(from:to:)`, there was inconsistent results. This issue is similar to the reported issue here: https://github.com/mapbox/turf-swift/issues/27

In this case, the distance to the tunnel entrance reports lower values as it approaches the entrance. However, in the mid-way of the tunnel commute, it suddenly begins to jump to higher values (greater than the tunnel distance) and then reduces to lower values before it exits the tunnel. The screenshot below highlights these inconsistencies:  

screen shot 2018-02-19 at 4 00 39 pm


MapboxCoreNavigation/RouteController.swift
```diff
@@ -635,6 +646,10 @@ extension RouteController: CLLocationManagerDelegate {
         updateRouteStepProgress(for: location)
         updateRouteLegProgress(for: location)
         
+        if routeProgress.currentLegProgress.currentStep.containsTunnel {
```
bsudekum: Move this if into `detectRouteStepInTunnel(for:)` and return early if it's true.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -651,6 +666,27 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let currentLocation = CLLocationCoordinate2D(latitude: location.coordinate.latitude, longitude: location.coordinate.longitude)
+        let distanceToEntrance = currentLocation.distance(to: tunnelStartCoordinate)
+        let distanceToExit = currentLocation.distance(to: tunnelEndCoordinate)
+        
+        // Capute any location which lies within the tunnel route.
+        // Pre-conditions:
+        //   - Current location to exit must be greater than zero meters.
+        //   - Distance to tunnel entrance must be less than tunnel distance.
+        //   - Distance to tunnel exit must also be less than tunnel distance.
+        if (tunnelDistance - distanceToExit) > 0 && distanceToEntrance <= tunnelDistance && distanceToExit <= tunnelDistance {
```
bsudekum: Do you think it's possible to have only 2 conditions here?

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -38,4 +39,46 @@ extension RouteStep {
...
+     */
+    var tunnelDistance: CLLocationDistance? {
+        guard let tunnelIntersection = tunnelIntersection else { return nil }
+        return tunnelIntersection.entry.location.distance(to: tunnelIntersection.exit.location)
+    }
+    
+    /**
+     Returns the entry and exit intersections of the tunnel on the current route step
+     */
+    var tunnelIntersection: (entry: Intersection, exit: Intersection)? {
```
bsudekum: `tunnelIntersection ` denotes a single intersection, while in reality, we're calculating the entry and exit of a tunnel here. Could you come up with a more descriptive variable name for this variable?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -561,6 +561,14 @@ extension NavigationViewController: RouteControllerDelegate {
         }
         return advancesToNextLeg
     }
+    
+    @objc public func routeController(_ routeController: RouteController, didEnterTunnelAt location: CLLocation?) {
+        if let _ = location {
```
bsudekum: The fact we're not using `location` here might clue us in that it may not be necessary to send this param over.

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -38,4 +39,46 @@ extension RouteStep {
...
+        guard let coordinates = coordinates, let tunnelIntersection = tunnelIntersection else { return nil }
+        return Polyline(coordinates).sliced(from: tunnelIntersection.entry.location, to: tunnelIntersection.exit.location)
+    }
+    
+    /**
+     Returns the distance of a tunnel slice for the current route step
+     */
+    var tunnelDistance: CLLocationDistance? {
+        guard let tunnelIntersection = tunnelIntersection else { return nil }
+        return tunnelIntersection.entry.location.distance(to: tunnelIntersection.exit.location)
```
bsudekum: This distance calculation is an absolute point to point distance. Instead, we should be using a function which calculates distance along a line. I believe `Polyline` is a struct and contains a distance variable. It may be possible to remove this and just use `tunnelSlice.distance`.

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -38,4 +39,46 @@ extension RouteStep {
...
+    }
+    
+    /**
+     Returns the entry and exit intersections of the tunnel on the current route step
+     */
+    var tunnelIntersection: (entry: Intersection, exit: Intersection)? {
+        guard let intersections = intersections, containsTunnel else { return nil }
+        for i in 0..<(intersections.count) where intersections.count > 1 {
+            if intersections[i].outletRoadClasses == .tunnel {
+                return (entry: intersections[i], exit: intersections[i+1])
```
bsudekum: Double check that i + 1 is always within bounds before returning here.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -561,6 +561,14 @@ extension NavigationViewController: RouteControllerDelegate {
...
+            self.styleManager.applyStyle(NightStyle())
```
bsudekum: Do you think we could first check if the style needs to be updated before continually updating the style on every location update?

MapboxCoreNavigation/RouteStep.swift
vincethecoder: `line #77` already does that check.

MapboxNavigation/NavigationViewController.swift
vincethecoder: @bsudekum I do a safety check before I apply the style in `applyStyle(_:)` on `line #110`.

MapboxCoreNavigation/RouteController.swift
vincethecoder: Hmm... honestly, 2 conditions would make it less readable -- considering they are related comparisons that meet the comment brief I left on preceeding `lines #680-682`.

MapboxNavigation/NavigationViewController.swift
vincethecoder: Also, this update is only applied to tunnel locations within a tunnel intersection.

MapboxNavigation/NavigationViewController.swift
vincethecoder: Sure!

MapboxNavigation/StyleManager.swift
```diff
@@ -96,7 +96,7 @@ open class StyleManager: NSObject {
         resetTimeOfDayTimer()
     }
     
-    func applyStyle() {
+    func applyStyle(_ newStyle: Style? = nil) {
```
frederoni: We should be able to avoid the special case for custom style below if we’d take a `StyleType` instead of a specific style here.

vincethecoder: @frederoni you're right 👍

MapboxCoreNavigation/RouteController.swift
vincethecoder: @bsudekum Yes indeed. Good looks 👍

vincethecoder:
@bsudekum and @frederoni Thanks for your feedback 👍

MapboxNavigation/NavigationViewController.swift
```diff
@@ -561,6 +561,14 @@ extension NavigationViewController: RouteControllerDelegate {
         }
         return advancesToNextLeg
     }
+    
+    @objc public func didEnterTunnel(_ darkModeEnabled: Bool) {
```
bsudekum: `darkModeEnabled` makes it seem like it _was already_ enabled. I think it also might make sense to remove all notions of UI from this function since MapboxCoreNavigation deals non-UI related functionality.

MapboxNavigation/StyleManager.swift
```diff
@@ -96,6 +96,21 @@ open class StyleManager: NSObject {
         resetTimeOfDayTimer()
     }
     
+    func applyStyle(type styleType: StyleType) {
```
bsudekum: Could this func be combined with the `applyStyle` below?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -134,6 +134,16 @@ public protocol RouteControllerDelegate: class {
...
+    
+    /**
+     Called when a tunnel is detected on a route step.
+     
+     Implement this method to enable dark mode when a commuter enters a tunnel.
+     
+     - parameter darkModeEnabled: The darkModeEnabled flag determines whether a `DarkStyle` should apply.
+     */
+    @objc(didEnterTunnel:)
+    optional func didEnterTunnel(_ darkModeEnabled: Bool)
```
bsudekum: We should remove the notion of dark mode from `RouteController` because MapboxCoreNavigation deals with non-UI related functionality.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -651,6 +662,27 @@ extension RouteController: CLLocationManagerDelegate {
...
+        // Calculated distances from current location to tunnel entrance and exit
+        let currentLocation = CLLocationCoordinate2D(latitude: location.coordinate.latitude, longitude: location.coordinate.longitude)
+        let distanceToEntrance = currentLocation.distance(to: tunnelStartCoordinate)
+        let distanceToExit = currentLocation.distance(to: tunnelEndCoordinate)
+        
+        // Capute any location which lies within the tunnel route.
+        // Pre-conditions:
+        //   - Current location to exit must be greater than zero meters.
+        //   - Distance to tunnel entrance must be less than tunnel distance.
+        if (tunnelDistance - distanceToExit) > 0 && distanceToEntrance <= tunnelDistance {
```
bsudekum: Would something like 

```swift
if distanceToEntrance <= tunnelDistance && distanceToExit <= tunnelDistance {
...
```

I only suggest this since it's easier to understand what we're comparing on each side of the && operator.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -651,6 +662,27 @@ extension RouteController: CLLocationManagerDelegate {
...
+        guard routeProgress.currentLegProgress.currentStep.containsTunnel else { return }
+        guard let tunnelDistance = routeProgress.currentLegProgress.currentStep.tunnelDistance, let tunnelSlice = routeProgress.currentLegProgress.currentStep.tunnelSlice else { return }
+        guard let tunnelStartCoordinate = tunnelSlice.coordinates.first, let tunnelEndCoordinate = tunnelSlice.coordinates.last else { return }
+        
+        // Calculated distances from current location to tunnel entrance and exit
+        let currentLocation = CLLocationCoordinate2D(latitude: location.coordinate.latitude, longitude: location.coordinate.longitude)
+        let distanceToEntrance = currentLocation.distance(to: tunnelStartCoordinate)
+        let distanceToExit = currentLocation.distance(to: tunnelEndCoordinate)
+        
+        // Capute any location which lies within the tunnel route.
```
bsudekum: Nit: no need for `route` in `tunnel route`. Also, `Capute` -> `Compute`.

MapboxCoreNavigation/RouteStep.swift
bsudekum: Isn't line 77 only checking where the number of intersections is greater that 1? Not necessarily that the current index in the loop has a follow on index.

MapboxNavigation/StyleManager.swift
vincethecoder: Yes, I considered it. However, this func `applyStyle(type:)` is cleaner, since it requires the `styleType`. If you notice, the other func `applyStyle ` scope is getting a bit bigger and complex with more conditions.

MapboxCoreNavigation/RouteController.swift
vincethecoder: Makes sense. Will do.

MapboxCoreNavigation/RouteController.swift
vincethecoder: `capture` -- misspelt

MapboxNavigation/StyleManager.swift
vincethecoder: `UPDATES:` I also refactored `applyStyle` so that we have a function  `applyStyle(type:)` to contain `styleType` related logic.

MapboxNavigation/StyleManager.swift
```diff
@@ -96,6 +96,21 @@ open class StyleManager: NSObject {
...
+        guard currentStyleType != styleType else { return }
+        
+        NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(timeOfDayChanged), object: nil)
+        
+        for style in styles {
+            if style.styleType == styleType {
+                style.apply()
+                currentStyleType = styleType
+                delegate?.styleManager?(self, didApply: style)
+                break
```
frederoni: It's rather unlikely but still possible to split up the styles into something like `[TopBannerDayStyle(), BottomBannerDayStyle()]`, with that in mind, I don't think we should break here.

vincethecoder: Oh okay, that makes sense. Will do.

vincethecoder: @frederoni Hmmm... the break can be removed. Since we're only applying a specified `StyleType`, rather than `Style` -- which can be an array, I don't believe it's applicable in this context. Or what do you think?

MapboxCoreNavigation/RouteStep.swift
vincethecoder: Yep - I realized I omitted `for i in 0..<(intersections.count - 1)`

MapboxNavigation/StyleManager.swift
frederoni: Looks good now 👍

MapboxCoreNavigation/RouteController.swift
```diff
@@ -651,6 +663,23 @@ extension RouteController: CLLocationManagerDelegate {
         checkForFasterRoute(from: location)
     }
     
+    func detectRouteStepInTunnel(for location: CLLocation) {
+        guard routeProgress.currentLegProgress.currentStep.containsTunnel else { return }
+        guard let tunnelDistance = routeProgress.currentLegProgress.currentStep.tunnelDistance, let tunnelSlice = routeProgress.currentLegProgress.currentStep.tunnelSlice else { return }
+        guard let tunnelStartCoordinate = tunnelSlice.coordinates.first, let tunnelEndCoordinate = tunnelSlice.coordinates.last else { return }
+        
+        // Calculated distances from current location to tunnel entrance and exit
+        let currentLocation = CLLocationCoordinate2D(latitude: location.coordinate.latitude, longitude: location.coordinate.longitude)
```
bsudekum: I don't think it's necessary to create this variable.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -651,6 +663,23 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let distanceToEntrance = currentLocation.distance(to: tunnelStartCoordinate)
```
bsudekum: `distance(to:)` is the absolute distance. We need to calculate the distance along a the current step.

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -38,4 +39,46 @@ extension RouteStep {
...
+        guard let tunnelSlice = tunnelSlice else { return nil }
+        return tunnelSlice.distance()
+    }
+    
+    /**
+     Returns the entry and exit intersection bounds of the tunnel on the current route step
+     */
+    var tunnelIntersectionBounds: (entry: Intersection, exit: Intersection)? {
+        guard let intersections = intersections, containsTunnel else { return nil }
+        for i in 0..<(intersections.count - 1) where intersections.count > 1 {
```
bsudekum: A couple of things I notice here. 

* I'm not sure `where intersections.count > 1 ` is necessary to check. Looping over a count of 1 is fine.
* The above point does not guard against `i + 1`. We should do that check while in the loop.
* it might be cleaner to do `for (intersectionIndex, intersection)` in intersections.enumerated() {...`

MapboxCoreNavigation/RouteController.swift
```diff
@@ -651,6 +663,23 @@ extension RouteController: CLLocationManagerDelegate {
...
     }
     
+    func detectRouteStepInTunnel(for location: CLLocation) {
+        guard routeProgress.currentLegProgress.currentStep.containsTunnel, let coordinates = routeProgress.currentLegProgress.currentStep.coordinates else { return }
+        guard let tunnelSlice = routeProgress.currentLegProgress.currentStep.tunnelSlice else { return }
+        guard let tunnelStartCoordinate = tunnelSlice.coordinates.first, let tunnelEndCoordinate = tunnelSlice.coordinates.last else { return }
+        
+        // Calculated distances from current location to tunnel entrance and exit
+        let tunnelDistance = tunnelSlice.distance()
+        let distanceToEntrance = Polyline(coordinates).distance(from: tunnelStartCoordinate, to: location.coordinate)
```
bsudekum: nit: pull `Polyline(coordinates)` out into a variable.

bsudekum:
This is looking good. @1ec5 what do you think about distance jump from the screenshot in the initial comment? This jump is pretty worrisome, there is nothing fishy in the code.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -634,6 +645,7 @@ extension RouteController: CLLocationManagerDelegate {
         updateDistanceToIntersection(from: location)
         updateRouteStepProgress(for: location)
         updateRouteLegProgress(for: location)
+        detectRouteStepInTunnel(for: location)
```
1ec5: Based on the simplification suggested in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r169512237, I think we should actually inline this calculation, much as you had originally. That way:

* RouteStepProgress can keep track of the last intersection that the user has passed, either as an object or an index into the step’s `intersections` array.
* RouteControllerDelegate can react to `routeControllerProgressDidChange` or any other notification by checking whether the current step progress’s `lastIntersectionPassed` has `tunnel` in its `outletRoadClasses`.
* No new delegate methods would be required, and future features tied to freeway driving can piggyback on the same idea to check for the `motorway` class without requiring lots of new code.

1ec5: > RouteStepProgress can keep track of the last intersection that the user has passed, either as an object or an index into the step’s `intersections` array.

This would require something slightly different than what https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r169512237 suggests: perhaps when advancing to a new step, we’d iterate over all the intersections, gathering their distances from the start of the step. That way we’d only need to perform a simple lookup on each location update.

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -38,4 +39,38 @@ extension RouteStep {
...
+        return Polyline(coordinates).sliced(from: intersectionBounds.entry.location, to: intersectionBounds.exit.location)
+    }
+    
+    /**
+     Returns the entry and exit intersection bounds of the tunnel on the current route step
+     */
+    var tunnelIntersectionBounds: (entry: Intersection, exit: Intersection)? {
+        guard let intersections = intersections, containsTunnel else { return nil }
+        for i in 0..<(intersections.count - 1) where intersections.count > 1 {
+            if intersections[i].outletRoadClasses == .tunnel {
```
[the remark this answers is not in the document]
vincethecoder: Good call! Just realized after reviewing a sample JSON structure for a route step with multiple tunnels.

MapboxCoreNavigation/RouteController.swift
vincethecoder: @1ec5 hey I came up with a rather simplified approach with detecting multiple tunnel intersections on a route step. Please refer to the `RouteStep` extension - `tunnelIntersectionsBounds` which I consume in the `hasEnteredTunnel(_: coordinates)` function. It handles our edge case.

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -1,4 +1,7 @@
 import MapboxDirections
+import Turf
+
+public typealias IntersectionBounds = (entry: Intersection, exit: Intersection)
```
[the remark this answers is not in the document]
vincethecoder: I considered it that, however, we have intersection bounds for `motorway`,` toll`, `restricted`, `ferry` or `tunnel`... right? This could apply to these road classes.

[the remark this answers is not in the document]
bsudekum: Also, is it required this typealias be public?

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -38,4 +41,31 @@ extension RouteStep {
...
+        }
+        return false
+    }
+    
+    /**
+     Returns an array of the entry and exit intersection bounds of the tunnel on the current route step
+     */
+    var tunnelIntersectionsBounds: [IntersectionBounds]? {
+        guard let intersections = intersections, intersections.count > 1, containsTunnel else { return nil }
+        var intersectionBounds = [(entry: Intersection, exit: Intersection)]()
```
bsudekum: Use `IntersectionBounds` here.

vincethecoder: Ha! missed that one.

MapboxCoreNavigation/RouteStep.swift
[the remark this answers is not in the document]
vincethecoder: nope.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -651,6 +668,31 @@ extension RouteController: CLLocationManagerDelegate {
         checkForFasterRoute(from: location)
     }
     
+    func detectRouteStepInTunnel(for location: CLLocation) {
```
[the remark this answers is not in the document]
vincethecoder: The intention is to make `hasEnteredTunnel(_: coordinates)` a testable function as opposed to embedding it in `detectRouteStepInTunnel (for:)` which seems to be a black box at this point.

bsudekum:
This is looking great. Let's hold on merging this until post v0.14.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -672,22 +677,30 @@ extension RouteController: CLLocationManagerDelegate {
...
         }
     }
     
+    func hasEnteredTunnel(_ intersectionBounds: [IntersectionBounds], coordinates: [CLLocationCoordinate2D]) -> Bool {
+        guard let startLocation = coordinates.first else { return false }
+        
+        let distanceTraveled = routeProgress.distanceTraveled
+        for bounds in intersectionBounds {
+            let distanceToTunnelEntrance = Polyline(coordinates).distance(from: startLocation, to: bounds.entry.location)
+            let distanceToTunnelExit = Polyline(coordinates).distance(from: startLocation, to: bounds.exit.location)
```
1ec5: On each location update, this turns the step’s coordinates into a polyline (which has a memory cost) and computes a distance along the polyline (not cheap) twice for each tunnel along the route – more if there are highway exits and entrances within the tunnel.

Out of curiosity, did you find any problems with the approach suggested in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r169512237? That approach would’ve relied on storing the just-passed intersection on the route step progress, similar to `distanceTraveled`, making the current road segment’s classes much more accessible, and in _O_(1) time.

The suggestion in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r169518181 was to iterate over _all_ the step’s intersections, gathering their distances from the beginning of the step, and to do it just once, as soon as the user begins the step.

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -54,23 +56,16 @@ extension RouteStep {
...
-        guard let intersections = intersections, containsTunnel else { return nil }
-        for i in 0..<(intersections.count - 1) where intersections.count > 1 {
-            if intersections[i].outletRoadClasses == .tunnel {
-                return (entry: intersections[i], exit: intersections[i+1])
+    var tunnelIntersectionsBounds: [IntersectionBounds]? {
+        guard let intersections = intersections, intersections.count > 1, containsTunnel else { return nil }
+        var intersectionBounds = [IntersectionBounds]()
+        for i in 0..<intersections.count {
+            if let outletRoadClasses = intersections[i].outletRoadClasses, outletRoadClasses.contains(.tunnel) && i < intersections.count - 1 {
+                intersectionBounds.append((entry: intersections[i], exit: intersections[i+1]))
```
1ec5: If we do keep this method around, I think it should look like either of the suggestions in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r169501520, because manual index management has been a cause for difficult-to-debug out-of-bounds errors in the past in this library.

vincethecoder: @1ec5 I wanted to go with code readability and simplicity. Where do you sense any difficult-to-debug out-of-bounds errors from this condition?  

Follow up: Perhaps, if you were to compose a self documented functional programming, which can be easily maintained (modified)? Also, wouldn't it be less optimal given we need to do series of operations -- such as `dropLast()`, `contains`, `index(where:)`, but just to mention a few.

1ec5: > I wanted to go with code readability and simplicity. Where do you sense any difficult-to-debug out-of-bounds errors from this condition?

In this case, I’m not concerned that the code would have a difficult-to-debug bug in it, but that it’s easier for an off-by-one error to creep in when reviewing. In general, it’s a best practice in Swift to avoid manual index management, just as it’s best to avoid manual memory access. The functional approach also happens to be more of a literate programming style in this case.

> wouldn't the proposed way be less optimal given we have to do series of chained operations -- such as `dropLast()`, `contains`, `index(where:)`?

The second approach in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r169501520 is just as efficient as this code: `dropLast()` returns a lazy array, so there’s no performance penalty, and `index(where:)` and `index(after:)` are both _O_(1) operations. `contains(_:)` is necessary because RoadClass is a bitmask. (The current code uses it too.)

MapboxCoreNavigation/RouteController.swift
```diff
@@ -687,14 +687,9 @@ extension RouteController: CLLocationManagerDelegate {
...
     func hasEnteredTunnel(_ intersectionBounds: [IntersectionBounds], coordinates: [CLLocationCoordinate2D]) -> Bool {
-        guard let startLocation = coordinates.first else { return false }
-        
         let distanceTraveled = routeProgress.distanceTraveled
         for bounds in intersectionBounds {
-            let distanceToTunnelEntrance = Polyline(coordinates).distance(from: startLocation, to: bounds.entry.location)
-            let distanceToTunnelExit = Polyline(coordinates).distance(from: startLocation, to: bounds.exit.location)
-            
-            if distanceToTunnelEntrance <= distanceTraveled && distanceTraveled <= distanceToTunnelExit {
+            if bounds.distanceToEntry <= distanceTraveled && distanceTraveled <= bounds.distanceToExit {
```
1ec5: This doesn’t address the performance concerns in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r170408923. `Polyline.distance(from:to:)` still gets called twice for every pair of intersections along the step on every location update.

MapboxCoreNavigation/RouteStep.swift
1ec5: Since we now need to iterate over all the intersections, even after finding a tunnel, the `zip` approach looks more desirable:

```swift
// Pair each intersection with its successor.
let intersectionPairs = zip(intersections, intersections.prefix(from: 1))
// Return the first pair of intersections that’s in a tunnel.
return intersectionPairs.first { $0.0.outletRoadClasses?.contains(.tunnel) ?? false }.map { IntersectionBounds($0.0, $0.1, coordinates) }
```

MapboxCoreNavigation/RouteController.swift
vincethecoder: @1ec5 Indeed, currently looking into it.

1ec5:
As a more flexible and memory-efficient alternative to maintaining a global cache of intersections, here’s what the approach suggested in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r169512237 would look like, along with (untested) sketches of the implementation:

* [ ] Upon advancing to a new step in `RouteController.advanceStepIndex(to:)`, iterate over the step’s indices, mapping them to their distances from the beginning of the step. Store the distances on the current step progress:
   ```swift
   let polyline = Polyline(coordinates)
   // Map intersections to distances. This runs in linear time, same as tunnelIntersectionBounds() as currently written.
   let distances = intersections.map { polyline.distance(from: coordinates.first, to: $0) }
   // This will be reset upon advancing to another step.
   currentStepProgress.intersectionDistances = distances
   ```
* [ ] In `RouteController.locationManager(_:didUpdate:)`, just before the `routeControllerProgressDidChange` notification is posted, fix #1170 by setting `intersectionIndex` to the index of the first intersection whose distance is more than the current distance traveled:
   ```swift
   let distances = currentStepProgress.intersectionDistances
   // Get the index of the first intersection that’s farther than the user. This runs in linear time.
   let upcomingIntersectionIndex = distances.index { $0 > currentStepProgress.distanceTraveled } ?? distances.endIndex
   // Get the preceding index, i.e., the just-passed intersection’s index.
   currentStepProgress.intersectionIndex = distances.index(before: upcomingIntersectionIndex)
   ```
   * `intersectionDistances` is guaranteed to be sorted in ascending order, and it’s unlikely that the user would be backing up. So for even better performance, we could start searching from the current intersection index onward.
   * While we’re at it, fix `RouteStepProgress.upcomingIntersection` to return the upcoming intersection, not the one the user just passed:
      ```swift
      return intersections[intersections.index(after: intersectionIndex)]
      ```
* [ ] Implement `RouteStepProgress.currentIntersection`:
   ```swift
   return intersections[intersectionIndex]
   ```
* [ ] Switch styles in `NavigationViewController.progressDidChange(notification:)`:
   ```swift
   let intersection = routeProgress.currentLegProgress.currentStepProgress.currentIntersection
   let isInTunnel = intersection.outletRoadClasses?.contains(.tunnel) ?? false
   styleManager.applyStyle(type: isInTunnel ? .nightStyle : .dayStyle)
   ```

vincethecoder:
Thanks @1ec5. I will review and make the necessary changes.
cc/ @bsudekum btw, what Minh suggested using `IntersectionIndex` to detect tunnel on an routeStep intersection, is what I meant by my prior comment in slack. Upon field testing, the former `2-indices` approach failed in certain use cases.

MapboxCoreNavigation/RouteStep.swift
```diff
@@ -38,4 +41,30 @@ extension RouteStep {
...
+                return true
+            }
+        }
+        return false
+    }
+
+    /**
+     Getter/Setter to retrieve and store all intersection distances.
+     */
+    var intersectionDistances: [CLLocationDistance] {
```
bsudekum: I'm unsure if this is the best place to store the property. What about putting this on RouteProgress?

vincethecoder: @bsudekum we only need to access the intersection distances for the `current step`, that's why I put that property here. In `RouteProgress` is looks out of place -- for instance, we need to call the variable `currentStepDistances`. `RouteProgress` has a variable, `currentStep` which is of type, `RouteStep` so it just made sense to put the current step's intersection distances in there. Or what do you think?

1ec5: While it would be desirable to put this property on RouteStep itself, the problem is the RouteStep is immutable. You’ve worked around this issue by accessing a global variable, but now it looks like it’s possible to get the `intersectionDistances` of a step other than the current step, which will unexpectedly return the intersection distances of the current step.

Due to the immutability of Route, RouteLeg, and RouteStep, we’ve built up a parallel structure of *Progress classes that can store transient state like this.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -400,6 +400,16 @@ public class NavigationViewController: UIViewController, RouteMapViewControllerD
 
         mapViewController?.notifyDidChange(routeProgress: routeProgress, location: location, secondsRemaining: secondsRemaining)
         
+        if let currentIntersection = routeProgress.currentLegProgress.currentStepProgress.currentIntersection,
+            let classes = currentIntersection.outletRoadClasses {
+            if classes.contains(.tunnel) {
+                print("turn on dark mode on intersection index: \(routeProgress.currentLegProgress.currentStepProgress.intersectionIndex)")
```
1ec5: (Remember to remove this `print` statement before merging.)

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -388,21 +388,34 @@ open class RouteStepProgress: NSObject {
...
      */
     @objc public var intersectionIndex: Int = 0
     
+    /**
+     The current intersection the user will travel through.
+     
+     The step must contain `Intersections` for this value not be `nil`.
+     */
+    @objc public var currentIntersection: Intersection? {
+        guard let intersections = intersectionsIncludingUpcomingManeuverIntersection, intersectionIndex >= intersections.startIndex, intersectionIndex <= intersections.endIndex else {
```
1ec5: If `intersectionIndex == intersections.endIndex`, then `intersections[intersectionIndex]` will be out of bounds and crash. Consider using the `~=` and `..<` operators.

bsudekum:
This is looking really good. I think the final piece we'll want to add here is not doing this for any tunnel less than say 100 meters.

vincethecoder:
@bsudekum latest requirements for tunnel distance done ✅

bsudekum:
@vincethecoder taking back my suggestion for checking the min distance. Is it possible to know the duration of the tunnel? This would be a more accurate measure because it accounts for speed. If it is possible to know the duration of a tunnel we could set the min duration to 10 seconds.

vincethecoder:
@bsudekum got it! Will update momentarily.

bsudekum:
@vincethecoder I think this is about ready. For this change, it might be wise that we put this new feature behind a flag with a default set to false. I think we need a little bit more real world testing of the feature. 

How about we add a new property `automaticallyHandleTunnels`. This property will handle both animation (future work) and switching the style while in tunnels.

/cc @frederoni

1ec5:
> How about we add a new property `automaticallyHandleTunnels`. This property will handle both animation (future work) and switching the style while in tunnels.

The tunnel heuristic tracked in #629 is the responsibility of RouteController, because it affects all aspects of location handling. By contrast, the functionality being added in this PR should be controlled by NavigationViewController, because it’s purely presentational.

No method or property name should ever say “handle”, because it’s too vague. NavigationViewController already has an [`automaticallyAdjustsStyleForTimeOfDay`](https://www.mapbox.com/mapbox-navigation-ios/navigation/0.14.1/Classes/NavigationViewController.html#/c:@M@MapboxNavigation@objc%28cs%29MBNavigationViewController%28py%29automaticallyAdjustsStyleForTimeOfDay) property, so how about an `automaticallyAdjustsStyleForTunnels` property to go along with it?

To avoid feature flag proliferation, we’d eventually combine it with `automaticallyAdjustsStyleForTimeOfDay` to become `usesNightStyleAutomatically`. (Ideally, we could also make `NavigationViewController.styleManager` public and put the option there.)

MapboxCoreNavigation/Constants.swift
```diff
@@ -92,3 +92,8 @@ let FasterRouteFoundEvent = "navigation.fasterRoute"
  The number of seconds remaining on the final step of a leg before the user is considered "arrived".
  */
 public var RouteControllerDurationRemainingWaypointArrival: TimeInterval = 3
+
+/**
+ The minimum distance of a tunnel intersection on a given route step.
+ */
+public var RouteProgressMinimumTunnelIntersectionDistance: CLLocationDistance = 100
```
1ec5: This constant is never used inside of RouteProgress. In fact, it’s only used inside MapboxNavigation, for a UI-specific feature. So let’s move the constant to the other Constants.swift and give it a name like `MBMinimumTunnelLengthForUsingNightStyle`.

1ec5: Do you think 100&nbsp;meters is an appropriate distance? I don’t have a good intuition about metric distances, but on a surface street, a tunnel much shorter than [this](https://www.youtube.com/watch?v=ew74X5z3Rgo) would still seem like plenty for switching to the night style. Per https://github.com/mapbox/mapbox-navigation-ios/pull/1127#issuecomment-370557811, factoring in the current speed would help us avoid special cases for freeways versus surface streets. SimulatedLocationManager has some code to determine the expected speed of a given segment along a step:

https://github.com/mapbox/mapbox-navigation-ios/blob/b381563387262b408dfe1939883c813bd616654b/MapboxCoreNavigation/SimulatedLocationManager.swift#L130-L134

vincethecoder:
@bsudekum For feature flag, we can create a field like `kTunnelAutomaticStyleChangesFeatureEnabled` and set it to false by default. (We should start using a plist to track our feature flags... or?)

As a side-note:
We can also use `isDeadReckoningEnabled` flag to detect if tunnel animation is enabled or not.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -388,21 +388,50 @@ open class RouteStepProgress: NSObject {
     /**
      The next intersection the user will travel through.
      
-     The step must contains `Intersections` for this value not be `nil`.
+     The step must contain `Intersections` for this value not be `nil`.
      */
     @objc public var upcomingIntersection: Intersection? {
-        guard let intersections = intersectionsIncludingUpcomingManeuverIntersection, intersectionIndex + 1 < intersections.endIndex else {
+        guard let intersections = intersectionsIncludingUpcomingManeuverIntersection, intersections.startIndex..<intersections.endIndex-1 ~= intersectionIndex else {
```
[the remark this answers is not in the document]
vincethecoder: @1ec5 what do you mean by use the `...` operator? can you clarify please?

bsudekum:
> @bsudekum For feature flag, we can create a field like kTunnelAutomaticStyleChangesFeatureEnabled and set it to false by default. (We should start using a plist to track our feature flags... or?)

Let's make this a store property on `RouteController`:

```swift
public var automaticallySwitchStyleForTunnels: Bool = false
```

> We can also use isDeadReckoningEnabled flag to detect if tunnel animation is enabled or not.

Yeah that sounds like a good idea.

MapboxCoreNavigation/RouteProgress.swift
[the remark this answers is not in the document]
1ec5: These two lines have the same effect of returning 2, but the first is more straightforward:

```swift
(1...2).upperBound
(1..<3-1).upperBound
```

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -417,21 +417,10 @@ open class RouteStepProgress: NSObject {
     }
     
     /**
-     Returns all the intersection distances on the current step.
+     Returns an array of the calculated distances from the current intersection to the next intersection on the current step.
```
1ec5: Nit: this is a property, not a method, so it is _set to_ the array rather than returning the array.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -304,6 +304,11 @@ public class NavigationViewController: UIViewController {
      */
     @objc public var annotatesSpokenInstructions = false
     
+    /**
+     A Boolean value that determines whether the dark style should be automatically activated when a route controller enters a tunnel.
+     */
+    @objc public var automaticallySwitchStyleForTunnels: Bool = false
```
1ec5: Nit: use the declarative form instead of the imperative form when naming a property. In Objective-C, a property’s getter can look identical to an action method:

```objc
// Does this line immediately switch the style?
[navigationViewController automaticallySwitchStyleForTunnels];
```

A clearer name would be `automaticallySwitchesStyleInsideTunnels` or `usesNightStyleInsideTunnels`.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -388,21 +388,39 @@ open class RouteStepProgress: NSObject {
     /**
      The next intersection the user will travel through.
      
-     The step must contains `Intersections` for this value not be `nil`.
+     The step must contain `intersectionsIncludingUpcomingManeuverIntersection` for this value not be `nil`.
```
1ec5: This is still unclear, because `intersectionsIncludingUpcomingManeuverIntersection` is a property of RouteStepProgress, not RouteStep.

> If `intersectionsIncludingUpcomingManeuverIntersection` is `nil`, this property is also `nil`.

MapboxNavigation/Constants.swift
```diff
@@ -30,8 +30,3 @@ public let MBCurrentLegAttribute = "isCurrentLeg"
  Attribute name for the route line that is used for identifying different `CongestionLevel` along the route.
  */
 public let MBCongestionAttribute = "congestion"
-
-/**
- The minimum distance of a tunnel intersection on a given route step.
- */
-public var MBMinimumTunnelLengthForUsingNightStyle: CLLocationDistance = 100
```
1ec5: I think the intent was to address the feedback in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r172624263 by moving this constant, but I don’t see where it got moved to.

MapboxCoreNavigation/RouteProgress.swift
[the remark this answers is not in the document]
bsudekum: @vincethecoder I think this is the last remaining issue to fix before merging.

MapboxNavigation/NavigationViewController.swift
vincethecoder: Honestly, I prefer to use `nightStyleInsideTunnelFeatureEnabled` to hint that variable is a feature flag. Just my `2¢`.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -409,7 +409,7 @@ public class NavigationViewController: UIViewController {
 
         mapViewController?.notifyDidChange(routeProgress: routeProgress, location: location, secondsRemaining: secondsRemaining)
 
-        if automaticallySwitchStyleForTunnels,
+        if enableNightStyleInsideTunnelFeatureEnabled,
```
bsudekum: This property has the word `enable(d)` in it twice. I too am a fan of suggestion - `automaticallySwitchesStyleInsideTunnels`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -628,6 +628,10 @@ extension RouteController: CLLocationManagerDelegate {
         let currentStepProgress = routeProgress.currentLegProgress.currentStepProgress
         let currentStep = currentStepProgress.step
 
+        let intersectionDistances = routeProgress.currentLegProgress.currentStepProgress.intersectionDistances
```
bsudekum: If the user has set `NavigationViewController.automaticallyChangeStyleForTunnel` to false, is it necessary to perform these calculations?

MapboxCoreNavigation/RouteProgress.swift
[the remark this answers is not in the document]
vincethecoder: @1ec5 In this context, since we're doing:

Okay so using the `CountableClosedRange` you suggest, we would end up with this approach:
```
intersections.startIndex...intersections.endIndex-2 ~= intersectionIndex
```
This is a bit buggy because, we could end up with a single intersections array, which would end up with an invalid range where `upperBound` value could be lower than `lowerBound` value.

For example:
intersections with `count` 1 will end up like this: `(0...1-2) -> (0...-1)`. It's safer to use a `(0..<1-1) -> (0...0)` than the former.

cc @bsudekum

MapboxCoreNavigation/RouteController.swift
vincethecoder: @bsudekum FYI 
- [ ] It shouldn't be restricted to automaticallyChangeStyleForTunnel. It should be wrapped within a isDeadReckoningEnabled flag.
- [ ] This calculation updates the `intersectionIndex` - fix for issue https://github.com/mapbox/mapbox-navigation-ios/issues/1170. @1ec5 do you need to use this fix for a task unrelated to dead reckoning?

bsudekum: @vincethecoder I don't think this code should be intertwined with `isDeadReckoningEnabled`. Let's keep these as separate ideas for now.

IMO, the fix here is to move `NavigationViewController.automaticallyChangeStyleForTunnel ` to `RouteController` and rename it to something more appropriate.

vincethecoder: @bsudekum I believe it was mentioned in one of my PR comment to separate the UI-related things in `NavigationViewController` from `RouteController`. Now, I'm totally thrown for loop 🤔  
cc @1ec5 - do you recall this discussion?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -403,6 +408,18 @@ public class NavigationViewController: UIViewController {
...
+
+        if enableNightStyleInsideTunnelFeatureEnabled,
+            let currentIntersection = routeProgress.currentLegProgress.currentStepProgress.currentIntersection,
+            let classes = currentIntersection.outletRoadClasses {
+                if classes.contains(.tunnel) {
+                    styleManager.applyStyle(type:.nightStyle)
+                    routeController.isDeadReckoningEnabled = true
+                } else {
+                    styleManager.timeOfDayChanged()
+                    routeController.isDeadReckoningEnabled = false
```
bsudekum: Why is `isDeadReckoningEnabled` being mutated?

vincethecoder: There is no need to continue mark the intersection as a candidate for dead reckoning.

MapboxCoreNavigation/RouteController.swift
1ec5: Yes, in https://github.com/mapbox/mapbox-navigation-ios/pull/1127#discussion_r172624263 I recommended keeping UI-related stuff out of Core Navigation. I stand by that recommendation.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -942,6 +948,12 @@ extension RouteController: CLLocationManagerDelegate {
         } else {
             routeProgress.currentLegProgress.stepIndex += 1
         }
+        
+        if isDeadReckoningEnabled, let coordinates = routeProgress.currentLegProgress.currentStep.coordinates, let intersections = routeProgress.currentLegProgress.currentStep.intersections {
```
1ec5: If `intersectionDistances` is only set when dead reckoning is enabled, then the documentation for that property should say so.

MapboxCoreNavigation/RouteProgress.swift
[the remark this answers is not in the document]
1ec5: Oops, you’re right, I was subtracting in the wrong direction. 😅

MapboxCoreNavigation/RouteController.swift
vincethecoder: Sure. As soon as we (@bsudekum, @1ec5, myself) gain clarity on whether we are wrapping this inside a `isDeadReckoningEnabled` flag,  I will make a final update on the `intersectionDistances` property accordingly.

bsudekum: This code should not be influenced by dead reckoning.

vincethecoder: Already removed - post our discussion on slack 😺

MapboxNavigation/NavigationViewController.swift
vincethecoder: Per our discussion, this variable will be `usesNightStyleInsideTunnels`.

cc @bsudekum @1ec5

MapboxNavigation/Constants.swift
vincethecoder: @1ec5 we're not using this constant at the moment. The tunnel distance check will be done on my upcoming PR for simulation of the tunnel navigation.

PR #1128, opened 2018-02-19
Round yards

frederoni:
Round yards so that it matches the road sign where applicable.

@bsudekum @vincethecoder @1ec5

### DIFF
--- a/MapboxCoreNavigation/DistanceFormatter.swift
+++ b/MapboxCoreNavigation/DistanceFormatter.swift
@@ -74,6 +74,20 @@ public class DistanceFormatter: LengthFormatter {
                 return 50
             }
             return distance < 3_000 ? 0 : 0.5
+        } else if numberFormatter.locale.identifier == "en-GB" {
+            if unit == .yard {
+                if distance.miles >= 0.1 {
+                    return 0
+                } else if distance.yards < 20 {
+                    return 10
+                } else if distance.yards < 100 {
+                    return 25
+                } else {
+                    return 50
+                }
+            } else {
+                return 0.1
+            }
         } else {
             if unit == .yard {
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -74,6 +74,20 @@ public class DistanceFormatter: LengthFormatter {
...
+        } else if numberFormatter.locale.identifier == "en-GB" {
+            if unit == .yard {
+                if distance.miles >= 0.1 {
+                    return 0
+                } else if distance.yards < 20 {
+                    return 10
+                } else if distance.yards < 100 {
+                    return 25
+                } else {
+                    return 50
```
vincethecoder: @frederoni what's your thought about assigning these values (magic numbers) `0, 10, 25, 50` into constant variables? (This would make it more self-documenting is we chose to go with the constant variable(s) approach.)

JThramer: Agreed. I wonder if we can add a rule to the linter about not allowing inline magic values?

MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -20,6 +20,49 @@ extension CLLocationDistance {
     var kilometers: CLLocationDistance {
         return self / 1000
     }
+    
+    func inMiles() -> Double {
+        return self * .metersPerMile
+    }
+    
+    func inYards() -> Double {
+        return self * .feetPerMeter / 3
```
JThramer: Nit: I would feel better about a `.yardsPerMeter` case that gets rid of the magic value, but it's pretty common knowledge that there are 3 feet in a yard. This might be me getting a little overly pedantic.

PR #1133, opened 2018-02-19
Update custom example

bsudekum:
Updates the custom UI example with recent changes.

/cc @mapbox/navigation-ios

### DIFF
--- a/Examples/Swift/CustomViewController.swift
+++ b/Examples/Swift/CustomViewController.swift
@@ -13,17 +13,16 @@ class CustomViewController: UIViewController, MGLMapViewDelegate, AVSpeechSynthe
     var routeController: RouteController!
 
     let textDistanceFormatter = DistanceFormatter(approximate: true)
-    lazy var speechSynth = AVSpeechSynthesizer()
     var userRoute: Route?
     var simulateLocation = false
+    
+    // Start voice instructions
     let voiceController = MapboxVoiceController()
     
     @IBOutlet var mapView: MGLMapView!
-    @IBOutlet weak var arrowView: UILabel!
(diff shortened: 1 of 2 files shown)


Examples/Swift/CustomViewController.swift
```diff
@@ -65,7 +64,6 @@ class CustomViewController: UIViewController, MGLMapViewDelegate, AVSpeechSynthe
     }
 
     func suspendNotifications() {
-        NotificationCenter.default.removeObserver(self, name: .routeControllerDidPassSpokenInstructionPoint, object: nil)
```
frederoni: Should we pass the instructions to the MapboxVoiceController?

Examples/Swift/CustomViewController.swift
```diff
@@ -13,17 +13,16 @@ class CustomViewController: UIViewController, MGLMapViewDelegate, AVSpeechSynthe
...
+    // Start voice instructions
     let voiceController = MapboxVoiceController()
     
     @IBOutlet var mapView: MGLMapView!
-    @IBOutlet weak var arrowView: UILabel!
-    @IBOutlet weak var instructionLabel: UILabel!
-    @IBOutlet weak var distanceLabel: UILabel!
     @IBOutlet weak var cancelButton: UIButton!
-
+    @IBOutlet weak var instructionsBannerView: InstructionsBannerView!
```
frederoni: 👍

Examples/Swift/CustomViewController.swift
bsudekum: No need to. When MapboxVoiceController is inited, it starts listening on it's own.

PR #1134, opened 2018-02-19
Specify Swift version in podspecs

friedbunny:
CocoaPods 1.4.0 supports a [`swift_version` setting in podspecs](https://blog.cocoapods.org/CocoaPods-1.4.0/#swift-version-dsl), so this adds that.

CocoaPods also says that the `.swift-version` file is being deprecated, but they’re not the only consumers of this file, so let's leave it around for the time being.

/cc @frederoni

### DIFF
--- a/MapboxCoreNavigation.podspec
+++ b/MapboxCoreNavigation.podspec
@@ -44,8 +44,8 @@ Pod::Spec.new do |s|
   s.dependency "MapboxMobileEvents", "~> 0.3"
   s.dependency "Turf", "~> 0.0.4"
 
-  # `swift_version` was introduced in CocoaPods 1.4.0. If a user directly specifies this
-  # podspec and they're using <1.4.0, ruby will throw an unknown method error.
+  # `swift_version` was introduced in CocoaPods 1.4.0. Without this check, if a user were to
+  # directly specify this podspec while using <1.4.0, ruby would throw an unknown method error.
   if s.respond_to?(:swift_version)
     s.swift_version = "4.0"
   end
--- a/MapboxCoreNavigation.podspec
+++ b/MapboxCoreNavigation.podspec
@@ -39,6 +39,7 @@ Pod::Spec.new do |s|
 
   s.requires_arc = true
   s.module_name = "MapboxCoreNavigation"
+  s.swift_version = "4.0"
 
   s.dependency "MapboxDirections.swift", "~> 0.17"
   s.dependency "MapboxMobileEvents", "~> 0.3"
--- a/MapboxCoreNavigation.podspec
+++ b/MapboxCoreNavigation.podspec
@@ -39,10 +39,20 @@ Pod::Spec.new do |s|
 
   s.requires_arc = true
   s.module_name = "MapboxCoreNavigation"
-  s.swift_version = "4.0"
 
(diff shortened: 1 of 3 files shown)


bsudekum:
@friedbunny Will this value be ignored if a user does not use CocoaPods >= 1.4.0? If it does fail, can you add this to `## Requirements` in the readme?

friedbunny:
Wow, CocoaPods continues to amaze — you’re correct, @bsudekum, this is a backwards incompatible change to the podspecs, as unbelievable as that is.

JThramer:
Unbelieveable.

@friedbunny @bsudekum That is indeed unbelievable. Terrible design decision on the part of the CocoaPods team. I guess that's what you get when you use `ruby` as your DSL. Great find!

MapboxCoreNavigation.podspec
```diff
@@ -39,10 +39,20 @@ Pod::Spec.new do |s|
...
   s.dependency "Turf", "~> 0.0.4"
 
+  # The CocoaPods podspec spec is not backwards compatible, so, when they add new parameters
+  # (like `swift_version` in 1.4.0), older versions of CocoaPods blow up in confusion.
+  # Specifying a `cocoapods_version >= 1.4.0` does nothing to solve the problem, as Ruby
+  # interprets unknown parameters as low-level syntax errors.
+  #
+  # Instead, we're forced to use Ruby's built-in semver comparison before declaring parameters
+  # added in recent CocoaPods releases.
+  if Version.new(Pod::VERSION) >= Version.new("1.4.0")
```
JThramer: Holy cow, that is an amazing workaround.

friedbunny: It’s such a terrible hack and on many levels I wish it didn’t work. 😭

JThramer:
@friedbunny you know, now that I think about it, we could also use a ruby existence check, like this:

```ruby
if s.try(:swift_version)
    s.swift_version = 4.0
end
```

Just a thought.

friedbunny:
Explicit version checking in a podspec never struck me as particularly robust, so I’ll be looking at @JThramer’s suggestion to simply check for method availability.

friedbunny:
b15a7ba switches to a method availability check, so this is ready to go.

A few notes:

### _Technically_ backwards incompatible, but...
The “omg podspecs aren’t backwards compatible!” thing is technically true, but is only really relevant to users who directly specify `:podspec => "https://some.url/SomeFramework.podspec"` and therefore download/run the ruby version of our podspec.

Most users will specify a version (or range of versions), which means they rely on the [JSON specs](https://github.com/CocoaPods/Specs/blob/master/Specs/5/8/b/MapboxNavigation/0.13.1/MapboxNavigation.podspec.json) published in [CocoaPods/Specs](https://github.com/CocoaPods/Specs) and execute none of our podspec’s ruby code.

We should _strongly discourage_ ourselves and our users from using direct podspec URLs, as this represents a security risk — ruby in podspecs can execute arbitrary shell commands and the content of a URL could change to include something dangerous.

### Update your CocoaPods installs
The person pushing this podspec to trunk has to be using CocoaPods 1.4.0+, otherwise `swift_version` won’t make it into the JSON podspec.

PR #1140, opened 2018-02-20
Update maps sdk

bsudekum:
Carthage deps are already on v3.7.x

/cc @mapbox/navigation-ios


bsudekum:
> Does this library actually depend on anything that was introduced in v3.7.x?

No, it does not. @1ec5 for consistencey sake, would it make sense to set the carthage maps sdk version to v3.6.x?

1ec5:
> for consistencey sake, would it make sense to set the carthage maps sdk version to v3.6.x?

No, because that would downgrade Carthage users from v3.7.x to v3.6.0.

PR #1141, opened 2018-02-20
Prevent statusView from being shown if it is already visible

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/973

If the `StatusView` is already shown, we should prevent other status updates from showing. The stacking updates caused the view to grow.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -535,10 +535,6 @@ extension NavigationViewController: RouteControllerDelegate {
             mapViewController?.mapView.updateCourseTracking(location: location, animated: true)
             mapViewController?.labelCurrentRoad(at: location)
         }
-    
-        if !(routeController.locationManager is SimulatedLocationManager) {
-            mapViewController?.statusView.hide(delay: 1.5, animated: true)
-        }
     }
     
     @objc public func routeController(_ routeController: RouteController, shouldDiscard location: CLLocation)  -> Bool {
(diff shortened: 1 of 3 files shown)


bsudekum:
This also reduces the some statuses are shown.

1ec5:
> This also reduces the some statuses are shown. There is no need to shame the user for a whole 3 seconds after rerouting.

If the it takes longer than 1½&nbsp;seconds to get a new route (perhaps due to network connectivity issues), will the status bar disappear prematurely?

bsudekum:
@1ec5 yes, but I guess this was the case prior to this PR too though.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -537,7 +537,7 @@ extension NavigationViewController: RouteControllerDelegate {
         }
     
         if !(routeController.locationManager is SimulatedLocationManager) {
-            mapViewController?.statusView.hide(delay: 3, animated: true)
+            mapViewController?.statusView.hide(delay: 1.5, animated: true)
```
1ec5: Can we move this call into `routeController(_:didRerouteAlong:)` so we hide the status view only after the new route comes in?

bsudekum:
@1ec5 @JThramer updated to delay hiding for 2 seconds after getting the new route. Tested with a slow network connection, worked well. Removed some duplication as well.

PR #1143, opened 2018-02-20
Use user's course to reroute sooner

bsudekum:
This PR seeks to reroute users sooner when they are actively moving away from a route.

This reuses the function `interpolatedCourse(from:along:)` and then feeds this into `shouldSnap(course:)`. If `shouldSnap` is false, we know the users course (and other heuristics) is not what we'd expect it to be for a given point on a route. To take out noise, this needs to be true consecutively, x times. The threshold is a sliding scale based on the location accuracy. 

This needs some heavy real world testing before merging.

todo:
- [x] pull out magic numbers into constants
- [x] real world test this

/cc @mapbox/navigation-ios


MapboxCoreNavigation/RouteController.swift
```diff
@@ -689,33 +705,12 @@ extension RouteController: CLLocationManagerDelegate {
         let radius = max(reroutingTolerance, location.horizontalAccuracy + RouteControllerUserLocationSnappingDistance)
         let isCloseToCurrentStep = newLocation.isWithin(radius, of: routeProgress.currentLegProgress.currentStep)
         
-        guard !isCloseToCurrentStep else { return true }
-        
-        // Check to see if the user is moving away from the maneuver.
```
bsudekum: This is now effectively replaced with the `isCourseIsOnRoute` check below.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -675,6 +670,27 @@ extension RouteController: CLLocationManagerDelegate {
...
     
+    /**
+     Monitors the user's course to see if it is consistantly moving away from what we expect the course to be at a given point.
+     */
+    func isCourseIsOnRoute(_ location: CLLocation) -> Bool {
+        let nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
+        
+        guard let calculatedCourseForLocationOnStep = interpolatedCourse(from: location, along: nearByCoordinates) else { return true }
+        
+        let maxUpdatesAwayFromRouteGivenAccuracy = Int(location.horizontalAccuracy / 4)
```
vincethecoder: Why did we choose the number `4`? What's this value?

MapboxCoreNavigation/RouteController.swift
vincethecoder: Maybe we should rename this `isCourseIsOnRoute` -> `isCourseOnRoute` 😸

MapboxCoreNavigation/RouteController.swift
```diff
@@ -220,7 +220,7 @@ open class RouteController: NSObject {
 
     var hasFoundOneQualifiedLocation = false
 
-    var recentDistancesFromManeuver: [CLLocationDistance] = []
+    var movmentsAwayFromRoute = 0
```
vincethecoder: refactor: `movmentsAwayFromRoute ` -> `movementsAwayFromRoute `

MapboxCoreNavigation/RouteController.swift
```diff
@@ -675,6 +670,27 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
     
+    /**
+     Monitors the user's course to see if it is consistantly moving away from what we expect the course to be at a given point.
+     */
+    func isCourseIsOnRoute(_ location: CLLocation) -> Bool {
```
vincethecoder: `func isCourseIsOnRoute(at location: CLLocation) -> Bool` maybe ??

MapboxCoreNavigation/RouteController.swift
```diff
@@ -675,6 +670,27 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
+        
+        guard let calculatedCourseForLocationOnStep = interpolatedCourse(from: location, along: nearByCoordinates) else { return true }
+        
+        let maxUpdatesAwayFromRouteGivenAccuracy = Int(location.horizontalAccuracy / 4)
+        
+        if movmentsAwayFromRoute >= max(3, maxUpdatesAwayFromRouteGivenAccuracy)  {
```
vincethecoder: This value `3` could be set to a variable - to make it self-documenting.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -675,6 +670,27 @@ extension RouteController: CLLocationManagerDelegate {
...
+            return false
+        } else if shouldSnap(location, toRouteWith: calculatedCourseForLocationOnStep) {
```
vincethecoder: For potential debug purpose, I would rather assign this computed condition into a variable. Just a thought...

MapboxCoreNavigation/RouteController.swift
```diff
@@ -675,6 +670,27 @@ extension RouteController: CLLocationManagerDelegate {
...
     
+    /**
+     Monitors the user's course to see if it is consistantly moving away from what we expect the course to be at a given point.
+     */
+    func userCourseIsOnRoute(_ location: CLLocation) -> Bool {
+        let nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
+        
+        guard let calculatedCourseForLocationOnStep = interpolatedCourse(from: location, along: nearByCoordinates) else { return true }
+        
+        let maxUpdatesAwayFromRouteGivenAccuracy = Int(location.horizontalAccuracy / Double(RouteControllerIncorrectCourseMultiplier))
```
1ec5: The horizontal accuracy pertains to the coordinates; it has nothing to do with the course. It makes little sense to divide a measurement in meters by a measurement in degrees.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -675,6 +670,27 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
     
+    /**
+     Monitors the user's course to see if it is consistantly moving away from what we expect the course to be at a given point.
+     */
+    func userCourseIsOnRoute(_ location: CLLocation) -> Bool {
+        let nearByCoordinates = routeProgress.currentLegProgress.nearbyCoordinates
+        
+        guard let calculatedCourseForLocationOnStep = interpolatedCourse(from: location, along: nearByCoordinates) else { return true }
```
1ec5: This is similar enough to the existing dead reckoning code that we should consider how they interact:

https://github.com/mapbox/mapbox-navigation-ios/blob/08bd523fd9c9fc0e045e1635baa88e573dc21dbc/MapboxCoreNavigation/RouteController.swift#L686-L688
https://github.com/mapbox/mapbox-navigation-ios/blob/b381563387262b408dfe1939883c813bd616654b/MapboxCoreNavigation/Constants.swift#L40

The dead reckoning code reroutes the user based on where they’ll be in one second, which is typically when the next location update arrives, based on the current speed.

bsudekum:
Did more testing with this PR this past weekend. I had one false positive reroute, unsure if this would have happened on master. Overall, I'm feeling hopeful this is good addition.

Issue #979, opened 2018-01-06
Add a gesture for repeating an instruction


1ec5:
Some navigation applications repeat the last spoken instruction when you tap on the turn banner. This way, you can slam your thumb on the top part of the screen without taking your eyes off the road. Nowadays, tapping our turn banner brings down the step table, so we’d have to choose a different gesture for repeating the instruction. Long-pressing, maybe? 🤔

/cc @mapbox/navigation-ios

bsudekum:
Another idea I'd like to throw into the mix- after unmuting, and it has been a sufficient amount of time since the previous announcement, repeat the current instruction. Ideally, rapid pressing the mute switch would not announce the instruction.

bsudekum:
@1ec5 @vincethecoder how about as a first pass, when tapping the user puck, the instruction is repeated?

JThramer:
Personally I would suggest tapping the instructions banner to repeat an instruction. This is similar functionality to other navigation applications.

Of course, right now tapping the banner shows the steps table. Perhaps we could switch that to a pan-down gesture?

My 2¢ anyways.

bsudekum:
Yeah this is tricky, a pan down gesture is sort of hidden. Long term, toggling the steps banner view would enabled via a button.

1ec5:
> how about as a first pass, when tapping the user puck, the instruction is repeated?

I don’t think that would be discoverable. No one associates the user puck with instructions of any sort.

> Personally I would suggest tapping the instructions banner to repeat an instruction. This is similar functionality to other navigation applications.

Count me in with the tap-to-repeat crowd. We’ve observed in user studies that people have a difficult time discovering the step table when it’s hidden behind a tap of the banner. People keep trying to swipe up from the bottom bar, which is where we used to put it, or down from the banner. Nothing in the banner indicates that it hides a full list – not even a 🍔 icon.

If revisiting that design is infeasible for some reason, a half-solution would be to repeat the instruction when long-pressing or force-tapping the banner. Note that, in some jurisdictions, it’s unlawful to perform any gesture more complex than a swipe or tap while driving.

1ec5:
#1150 and #1253 made it easier to access the steps table, but neither PR actually added a tap-to-repeat affordance, as requested in this issue.

[...]

Issue #1138, opened 2018-02-20
Make step table more discoverable


1ec5:
#785 moved the entry point for the step table, so that instead of dragging the bottom bar upwards, the user taps the banner at the top. We observed in a small study that users have a difficult time finding the step table. Participants expected to swipe upwards from the bottom bar, as in Google Maps and Apple Maps; failing that, they expected to swipe downwards from the banner. There are several things we can do to improve the discoverability of this feature (pick two):

* [ ] Add a visual cue that the banner reveals additional steps, either a handle at the bottom of the banner or a 🍔 icon on the side.
* [x] Reveal the step table upon swiping downwards from the banner.
* [ ] Reveal the step table upon swiping upwards from the bottom bar.

/cc @mapbox/navigation-ios @mapbox/navigation-android

1ec5:
If we implement either swipe gesture, we can remove the banner’s tap gesture and repurpose it for repeating the current instruction: #979.

bsudekum:
I think before moving forward here, we should design some mockups. I'm weary of adding more visual buttons to the UI that may clutter it and make it confusing for the user.

JThramer:
Just recording this for posterity, but I'm a big fan of swiping, either down from the top banner or up from the bottom banner.

1ec5:
> I'm weary of adding more visual buttons to the UI that may clutter it and make it confusing for the user.

Agreed. Google and Apple take the approach of placing a handle on the bottom bar, subtly hinting that you can swipe on the bar to reveal more content. We could add a pan gesture recognizer and small handle to the banner without fundamentally changing the interaction model or cluttering up the UI.

1ec5:
In chat, it seems like this might be a path forward:

* [ ] Add a subtle handle to the bottom of the banner, as a visual cue that the banner reveals additional steps, similar to what Apple and Google do.
* [ ] Reveal the step table upon swiping downwards from the banner, leaving in place the current tap gesture.

Future steps that don’t need to take place right away:

* [ ] Repeat the current instruction when tapping the banner (to fix #979).
* [ ] Reveal the step table upon swiping upwards from the bottom bar, replacing the functionality that’s currently in the top bar.

brsbl:
This work was merged: https://github.com/mapbox/mapbox-navigation-ios/pull/1150

1ec5:
[...]

PR #1150, opened 2018-02-22
Swipe down to reveal steps list

vincethecoder:
Added functionality to announce current navigation instructions when user puck is tapped.

Issue addressed: https://github.com/mapbox/mapbox-navigation-ios/issues/979

PS: Ongoing discussion on what should respond to user interaction with map interface.

cc @mapbox/navigation-ios 

TODO:

- [x] Reveal the step table upon swiping downwards from the banner, leaving in place the current tap gesture.


bsudekum:
@vincethecoder how shall we proceed here? Per voice yesterday, we discussed this probably is not the best user interaction since it's pretty hidden.

vincethecoder:
@bsudekum Yes, ideally it makes sense to repeat the instructions when the user taps the `instructionsBannerView` and then have tiny list icon at then top-right corner of the `instructionsBannerView` to show the remaining instructions. I believe that's more user friendly.

vincethecoder:
simulator screen shot - iphone 8 - 2018-02-26 at 15 37 01

Added a `step` icon between volume and feedback floating buttons at the right hand side of the map view.

1ec5:
> Added a step icon between volume and feedback floating buttons at the right hand side of the map view.

This might not be discoverable/tappable enough for something as crucial as getting the list of steps. Can we nest the button in the banner somehow?

bsudekum:
Per conversation today in chat, we came to the conclusion that:

1. tapping top banner repeats instruction
1. swiping down on top banner shows steps list

vincethecoder:
As referenced from the ongoing conversation on slack channel, we came to a final conclusion today. 

- [ ] Add a subtle handle to the bottom of the banner, as a visual cue that the banner reveals additional steps, similar to what Apple and Google do.
- [ ] Reveal the step table upon swiping downwards from the banner, leaving in place the current tap gesture.

Issue reference: https://github.com/mapbox/mapbox-navigation-ios/issues/1138#issuecomment-370938050

bsudekum:
I'm getting a crash, to reproduce:

1. start with system muted
1. move along a step
1. unmute
1. tap top banner
1. asserts here

https://github.com/mapbox/mapbox-navigation-ios/blob/0c0a769033330bc526588e85cf2451638b866ce8/MapboxNavigation/MapboxVoiceController.swift#L79

This value is set here:

https://github.com/mapbox/mapbox-navigation-ios/blob/0c0a769033330bc526588e85cf2451638b866ce8/MapboxNavigation/RouteVoiceController.swift#L208-L210

And is nil because the system is muted.

vincethecoder:
@bsudekum cannot reproduce. Can you share the specific route, so that I can simulate it on my branch build? Thanks.

bsudekum:
@vincethecoder I do not think this is specific to a route. The issue is that `RouteVoiceController. routeProgress`  is never set while muted.

vincethecoder:
@bsudekum Sure. I just wanted to test the referenced crash after the change was made. It didn't crash for me. But I understand why this not setting the `RouteVoiceController. routeProgress` will be an issue.

vincethecoder:
@bsudekum moreover, I don't believe the root cause of the identified issue will be in that function scope. We need to investigate why the notification's `routeProgress` is set to `nil` prior to our assert check. A separate issue ticket needs to be raised for that.

bsudekum:
It's set to nil because the this check to `voiceMuted ` is true:

https://github.com/mapbox/mapbox-navigation-ios/blob/0c0a769033330bc526588e85cf2451638b866ce8/MapboxNavigation/RouteVoiceController.swift#L208-L210

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -589,6 +589,25 @@ extension RouteMapViewController: NavigationViewDelegate {
...
     //MARK: InstructionsBannerViewDelegate
     func didTapInstructionsBanner(_ sender: BaseInstructionsBannerView) {
+        repeatInstructions()
+    }
+    
+    func didDragInstructionsBanner(_ sender: BaseInstructionsBannerView) {
+        displayPreviewInstructions()
+    }
+    
+    private func repeatInstructions() {
```
bsudekum: Per https://github.com/mapbox/mapbox-navigation-ios/issues/1138#issuecomment-370938050, we discussed that the first two steps would be implemented first. No need to do anything around repeating the instruction yet.

vincethecoder: Yes. Will revert that work.

vincethecoder: @bsudekum done ✅. This private function will just live here till we are ready to plug it in a delegate or gesture recognizer

bsudekum: @vincethecoder I think we should remove it and add it when it's time to add this feature.

vincethecoder: Ok

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -205,11 +205,11 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     }
     
     @objc open func didPassSpokenInstructionPoint(notification: NSNotification) {
-        guard !NavigationSettings.shared.voiceMuted else { return }
-        
         routeProgress = notification.userInfo![RouteControllerNotificationUserInfoKey.routeProgressKey] as? RouteProgress
         assert(routeProgress != nil, "routeProgress should not be nil.")
         
+        guard !NavigationSettings.shared.voiceMuted else { return }
```
bsudekum: this can be undone.

vincethecoder: @bsudekum is there a reason why we have to undo this?

bsudekum:
Let's break this PR up into two PRs:

1. This pr which implements the gesture handlers
1. Adding subtle button indicator view that clues the user they can pull down.

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -56,6 +56,8 @@ extension BaseInstructionsBannerView {
         self.separatorView = separatorView
         
         addTarget(self, action: #selector(BaseInstructionsBannerView.tappedInstructionsBanner(_:)), for: .touchUpInside)
```
bsudekum: Tests are failing because this function still exists but was removed here: https://github.com/mapbox/mapbox-navigation-ios/pull/1150/files#diff-7bae0cf60f54997d0d5e86fe02b5cddbL60 and renamed to `draggedInstructionsBanner `

vincethecoder: Resolved ✅

bsudekum:
@vincethecoder when running locally, it appears that tapping the top banner no longer shows the upcoming list view; only swiping.

vincethecoder:
TODO: 
- [ ] Update Change log - get release date details: `v0.15.0 (?? Date Unknown ??)`
cc @bsudekum

bsudekum:
@vincethecoder the changelog addition can go under the `master` section.

Issue #901, opened 2017-11-29
currentStepProgress.distanceRemaining can be greater than step.distance


bsudekum:
Seeing this when starting a route and the user is standing still.

```
(lldb) p routeProgress.currentLegProgress.currentStepProgress.distanceRemaining
(CLLocationDistance) $R2 = 32.099941581626325
(lldb) p routeProgress.currentLegProgress.currentStepProgress.step.distance
(CLLocationDistance) $R3 = 31.699999999999999
```

/cc @mapbox/navigation-ios

1ec5:
A distance of 40 centimeters could just be small discrepancies between Turf’s [Haversine distance](https://github.com/mapbox/turf-swift/blob/5b062bdbadaada2598cfdb74d177ffb22ac76607/Turf/Turf.swift#L49-L56) implementation and [OSRM’s](https://github.com/Project-OSRM/osrm-backend/blob/088d4edc6b761d2642a9b1ab7d85b701d3165fb2/src/util/coordinate_calculation.cpp#L35-L62). OSRM applies its own coordinate precision, whereas the Swift implementation just uses Double precision.

bsudekum:
Yeah, you could be right. However when starting a route and the users distance is greater than the step distance, we do not trigger voice/visual announcements that may be equal to the steps distance. 

Would it make sense to put in a `max()` somewhere to make sure we're never above the step's distance?

/cc @frederoni

frederoni:
This could also mean that `RouteStepProgress.distanceTraveled` is negative.
I think it would make sense to prevent negative values for traveled distance if that's the case.

Sealos:
@frederoni It does give negative values, I remember getting them when I submitted #771.

PR #1155, opened 2018-02-25
Always update banner and voice instruction when beginning

bsudekum:
Because of https://github.com/mapbox/mapbox-navigation-ios/issues/901, when starting a trip, the user may not cross the threshold of when a voice instruction or banner should be updated. 

This change ensures that at the beginning of any route, if the spoken/visual index is zero, we make sure to update the user. Otherwise, the banner instruction could be blank or we could delay the voice instruction if the user is standing still.

/cc @mapbox/navigation-ios


bsudekum:
This does however make the assumption that the first voice instruction/banner update does occur at directly at the beginning of any step.

I think this is a fair assumption to make.

PR #1157, opened 2018-02-26
ImageDownloader/InstructionPresenter tail work / cleanup / refactor

akitchen:
Related to #1046 - some light cleanup and refactoring.


akitchen:
There might be a bit more to do here but I'd like to merge this early and often. for example I should extract a constant for the async wait timeout since it takes considerably longer on CI and fails from time to time as a result.

PR #1162, opened 2018-02-26
Hide completed coordinates

bsudekum:
This removes coordinates from the map that the user has passed by.

todo: 

- [ ] Set a max zoom. Zooming past z17 shows some pitfalls with this technique. 
- [ ] Handle maneuver arrows
- [ ] Check for optimizations

new3

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -433,6 +433,12 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         return candidates
     }
     
+    var userPuckScreenCoordinates: CLLocationCoordinate2D? {
+        guard let rect = userCourseView?.frame else { return nil }
+        let point = CGPoint(x: rect.midX, y: rect.midY)
+        return self.convert(point, toCoordinateFrom: self)
+    }
+    
     private func routes(closeTo point: CGPoint) -> [Route]? {
         let tapCoordinate = convert(point, toCoordinateFrom: self)
         
@@ -786,7 +792,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     // MARK: Utility Methods
     
     func shape(describing route: Route, legIndex: Int?) -> MGLShape? {
-        guard let coordinates = route.coordinates else { return nil }
+        guard let routeCoords = route.coordinates else { return nil }
+        
+        let coordinates: [CLLocationCoordinate2D]
+        if let userPuckScreenCoordinates = userPuckScreenCoordinates {
+            coordinates = Polyline(routeCoords).sliced(from: userPuckScreenCoordinates, to: routeCoords.last).coordinates
+        } else {
+            coordinates = routeCoords
+        }
         
         var linesPerLeg: [[MGLPolylineFeature]] = []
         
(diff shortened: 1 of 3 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -433,6 +433,12 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         return candidates
     }
     
+    var userPuckScreenCoordinates: CLLocationCoordinate2D? {
+        guard let rect = userCourseView?.frame else { return nil }
+        let point = CGPoint(x: rect.midX, y: rect.midY)
+        return self.convert(point, toCoordinateFrom: self)
```
bsudekum: Is there a better method we can use here that accounts for the whole frame instead of a point?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -786,7 +792,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     // MARK: Utility Methods
     
     func shape(describing route: Route, legIndex: Int?) -> MGLShape? {
-        guard let coordinates = route.coordinates else { return nil }
+        guard let routeCoords = route.coordinates else { return nil }
+        
+        let coordinates: [CLLocationCoordinate2D]
+        if let userPuckScreenCoordinates = userPuckScreenCoordinates {
```
bsudekum: Handle cases where the view may not be using a `UserPuckView`.

JThramer: Just a wild thought, but why not use `NavigationMapViewDelegate.navigationMapViewUserAnchorPoint(_:)`? That saves us from having to do the screen coordinates conversion, and I think we already define default behavior if the developer doesn't implement the method?

MapboxNavigation/NavigationMapView.swift
1ec5: No; the only other option is to convert a rect, but that won’t account for rotation or pitch.

MapboxNavigation/NavigationMapView.swift
JThramer: Just a wild thought, but why not use `NavigationMapViewDelegate.navigationMapViewUserAnchorPoint(_:)`? That saves us from having to do the screen coordinates conversion, and I think we already define default behavior if the developer doesn't implement the method?

bsudekum: Unsure why, but when I used this delegate, it clipped the route line to the upcoming maneuver.

bsudekum:
Highway clipping is not great; the user moves pretty far in 1 second between location updates.

new3

bsudekum:
To do this right, I think we should hold until gradient work has been added to the Maps SDK, ref https://github.com/mapbox/mapbox-gl-native/compare/gradient

1ec5:
> To do this right, I think we should hold until gradient work has been added to the Maps SDK, ref https://github.com/mapbox/mapbox-gl-native/compare/gradient

Line gradient support in the map SDK is tracked in mapbox/mapbox-gl-native#11718. The plan is to keep the line’s geometry constant as the user moves, but change the line gradient (`MGLLineStyleLayer.lineGradient`) to abruptly go from transparent to opaque at the user’s location along the line, with a smooth transition (`lineGradientTransition`) that matches the course view’s animated interpolation.

Issue #1158, opened 2018-02-26
Remove `Weak GPS` StatusView update


bsudekum:
ref: https://github.com/mapbox/mapbox-navigation-ios/pull/1154

The `Weak GPS` status view addition was meant to manage users expectations while the user puck is not moving because of a bad gps signal. 

The view induces more anxiety and is less helpful than informative. Let's cut it.

PR #1163, opened 2018-02-27
remove weak gps status view update

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1158

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -537,19 +537,6 @@ extension NavigationViewController: RouteControllerDelegate {
         }
     }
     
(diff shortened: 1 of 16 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -537,19 +537,6 @@ extension NavigationViewController: RouteControllerDelegate {
         }
     }
     
-    @objc public func routeController(_ routeController: RouteController, shouldDiscard location: CLLocation)  -> Bool {
```
frederoni: Determining whether to show the weak GPS banner was not the only purpose of this method. We should still implement this method and delegate it to the receiver.

bsudekum: 👍 good call

Issue #1070, opened 2018-01-25
Various constraints fail to mirror in right-to-left languages


1ec5:
Now that #1043 has added Hebrew support to the UI, we need to ensure that the UI layout mirrors properly in right-to-left languages. I spotted a few issues on a quick simulated trip through Tel Aviv:

✅ <img src="https://user-images.githubusercontent.com/1231218/35385291-ae0d54b0-017c-11e8-9c2d-5f90cdbb24af.png" width="300" alt="example" align="top"> <img src="https://user-images.githubusercontent.com/1231218/35385333-cca952ac-017c-11e8-9b3d-15a46c4011a9.png" width="300" alt="start" align="top">

❌ <img src="https://user-images.githubusercontent.com/1231218/35384971-c5f23d44-017b-11e8-8af4-97617c4b7c13.png" width="300" alt="navigation" align="top"> <img src="https://user-images.githubusercontent.com/1231218/35385214-75eea3ea-017c-11e8-90cf-767c4956d139.png" width="300" alt="overview" align="top">

❌ <img src="https://user-images.githubusercontent.com/1231218/35385135-3cbf9c78-017c-11e8-84d0-40cba53ea756.png" width="300" alt="feedback" align="top">

✅ <img src="https://user-images.githubusercontent.com/1231218/35384623-c8bd689c-017a-11e8-8eac-a065745798b3.png" width="300" alt="end of route" align="top">

The storyboard and any programmatic Auto Layout code needs to be audited for the use of left and right constraints, which should be replaced by leading and trailing constraints in most cases.

/cc @mapbox/navigation-ios @danesfeder

JThramer:
Perhaps this would be a good thing to check for in the linter. It's usually just accidental usage of something like `leftAnchor` over `leadingAnchor`.

1ec5:
The lanes in the lane view are also in the wrong order. This is really confusing. In the example below, the user is prone to thinking they should take one of the left lanes to turn left, whereas they should take one of the right lanes to turn right:

<img src="https://user-images.githubusercontent.com/1231218/36819248-7e57e4ba-1c9d-11e8-8eca-865e1aa85079.png" width="300" alt="lanes">

danpat:
@1ec5 are you *sure* this isn't just a https://en.wikipedia.org/wiki/Hook_turn location?

1ec5:
[...]

PR #1175, opened 2018-03-01
Avoid mirroring lanes

### LINKED ISSUE #1174: Lanes in wrong order in Hebrew
When the system language is set to Hebrew, the lanes in the lane view are in the wrong order, potentially creating a confusing end user experience. In the example below, the user is prone to thinking they should take one of the left lanes to turn left, whereas they should take one of the right lanes to turn right:

lanes

/ref https://github.com/mapbox/mapbox-navigation-ios/issues/1070#issuecomment-369422431
/cc @frederoni

1ec5:
Prevent the lanes view from swapping the order of the lanes in right-to-left environments:

lanes

Fixes #1174.

/cc @frederoni


frederoni:
@akitchen 2 seconds timeout doesn't seem to be enough for Bitrise on a slow day like this
```testDownloadingImageAgainAfterFirstDownloadCompletes, Asynchronous wait failed: Exceeded timeout of 2 seconds, with unfulfilled expectations: "Second Image Download".```

1ec5:
> Is it possible to mock the locale and region settings in a snapshot test w/o adding a new test scheme?

It is possible to change the formatter’s locale, but unfortunately that doesn’t affect the value of `LengthFormatter.unitString(forValue:unit:)`. Beyond that, launching with the [`-AppleTextDirection YES`](https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/TestingYourInternationalApp/TestingYourInternationalApp.html#//apple_ref/doc/uid/10000171i-CH7-SW5) environment variable would exercise this PR, but that would still require an additional scheme.

I’d rather not add a new scheme (and therefore a separate Bitrise step) just to test this fix, which is easy to verify manually and relatively unlikely to break. The fixes for #1070 would be more invasive, making testing more desirable, but snapshot tests for those fixes would be more fragile.

PR #1176, opened 2018-03-01
More robust distance formatting

### LINKED ISSUE #1173: No metric distance in banner in Hebrew
When the system language is set to Hebrew and the system region is set to a country that uses the metric system, such as Israel, no distance appears in the turn banner.

customary metric

_Left: Customary units. Right: Metric units._

/cc @frederoni

1ec5:
Implemented `DistanceFormatter.attributedString(for:)`, which applies a `quantity` attribute to the numeric quantity part of the returned attributed string. This method relies on finding the formatted number within the distance string, as opposed to finding the unit, which turned out to be unreliable in some locales.

Simplified `BaseInstructionsBannerView.distance`’s setter to pass this attributed string directly to DistanceLabel instead of attempting to split the distance string on its quantity and unit. DistanceLabel now emphasizes the quantity relative to the unit based on the presence of the quantity attribute.

All this is to work around [a Foundation bug](http://openradar.appspot.com/radar?id=5061404864282624) in which `LengthFormatter.unitString(fromValue:unit:)` is unreliable in metricated Hebrew locales, such as Hebrew (Israel).

hebrew english
_Left: The Hebrew (Israel) locale. Right: The English (United States) locale._

Original description:

~~Worked around [a Foundation bug](http://openradar.appspot.com/radar?id=5061404864282624) by splitting the distance string on whitespace if `LengthFormatter.unitString(fromValue:unit:)` doesn’t return the expected value. Now the distance to the upcoming maneuver appears in the turn banner, as expected, when the system language is set to Hebrew and the system region is set to a metricated region, such as Israel.~~

Fixes #1173.

/cc @JThramer @frederoni

### DIFF
--- a/MapboxCoreNavigation/DistanceFormatter.swift
+++ b/MapboxCoreNavigation/DistanceFormatter.swift
@@ -6,22 +6,22 @@ extension CLLocationDistance {
     static let feetPerMeter: CLLocationDistance = 3.28084
     
     // Returns the distance converted to miles
-    var miles: CLLocationDistance {
+    var miles: Double {
         return self / .metersPerMile
     }
     
     // Returns the distance converted to feet
-    var feet: CLLocationDistance {
+    var feet: Double {
         return self * .feetPerMeter
     }
     
     // Returns the distance converted to yards
-    var yards: CLLocationDistance {
+    var yards: Double {
         return feet / 3
     }
     
     // Returns the distance converted to kilometers
-    var kilometers: CLLocationDistance {
+    var kilometers: Double {
         return self / 1000
     }
     
@@ -33,6 +33,27 @@ extension CLLocationDistance {
     // Returns the distance in meters converted from yards
     func inYards() -> Double {
         return self * .feetPerMeter / 3
+    }
+    
+    func converted(to unit: LengthFormatter.Unit) -> Double {
+        switch unit {
+        case .millimeter:
+            return self / 1_000
+        case .centimeter:
+            return self / 100
+        case .meter:
+            return self
+        case .kilometer:
+            return kilometers
+        case .inch:
+            return feet / 12
+        case .foot:
+            return feet
+        case .yard:
+            return yards
+        case .mile:
+            return miles
+        }
     }
 }
 
@@ -69,6 +90,10 @@ struct RoundingTable {
         }
         return thresholds.last!
     }
+}
+
+extension NSAttributedStringKey {
+    public static let quantity = NSAttributedStringKey(rawValue: "MBQuantity")
 }
 
 /// Provides appropriately formatted, localized descriptions of linear distances.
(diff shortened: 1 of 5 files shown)


MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -36,8 +36,15 @@ open class BaseInstructionsBannerView: UIControl {
...
                 let distanceString = distanceFormatter.string(from: distance)
-                let distanceUnit = distanceFormatter.unitString(fromValue: distance, unit: distanceFormatter.unit)
-                guard let unitRange = distanceString.range(of: distanceUnit) else { return }
+                var distanceUnit = distanceFormatter.unitString(fromValue: distance, unit: distanceFormatter.unit)
+                var unitRange = distanceString.range(of: distanceUnit)
+                // In Hebrew, “meters” looks different in the distance string than it does on its own. <rdar://problem/38028709>
+                // Fall back on splitting the string on whitespace.
+                if unitRange == nil, let spaceRange = distanceString.rangeOfCharacter(from: .whitespaces) {
+                    unitRange = spaceRange.upperBound..<distanceString.endIndex
+                    distanceUnit = String(distanceString[unitRange!])
```
bsudekum: if let instead of force unwrapping.

bsudekum:
A lane test addition here would be 💯

MapboxNavigation/InstructionsBannerView.swift
1ec5: `unitRange` is set to a non-nil value in the line right above this line. `if let` isn’t suitable because `unitRange` is used outside the `if` statement. Specifically, if `distanceString` doesn’t contain the unit string _and_ doesn’t contain a space, we still need to bail unless we come up with a different way to deemphasize the unit string. (Or should we just treat the entire string as the number?)

1ec5:
> A lane test addition here would be 💯

I think you’re thinking of #1175; this PR is unrelated to lanes.

I’ll look into extracting this logic into something unit-testable. 👍 Meanwhile, a snapshot test would be more trouble than it’s worth, because the fix would require a specific locale (he-IL). It is possible to change the formatter’s locale, but unfortunately that doesn’t affect the return value of `LengthFormatter.unitString(forValue:unit:)`. We’d have to create a new scheme and Bitrise step just to test the fix.

1ec5:
OK, I started over from scratch in ceb3c6e89ba545fe005c32532f3ff4113b85cabc: instead of looking for the unit within the distance string, look for the formatted number. This should be more reliable because we’re piggybacking on the distance formatter’s own numeric formatter. As a failsafe, if no quantity is found, the entire string is emphasized. In that worst-case scenario, it’s better to contort the layout a bit than to make the distance illegible.

DistanceFormatter is now responsible for creating the attributed string in the first place. It applies a custom attribute that makes it easier for DistanceLabel to find the quantity and apply presentational attributes to it. I could’ve simplified things further by having DistanceLabel invoke the distance formatter instead of BaseInstructionsBannerView, but I wasn’t sure if it was appropriate for a lightweight class in Style.swift to depend on something in Core Navigation.

The upshot is that `DistanceFormatter.attributedString(for:)` now has unit tests for multiple locales, all in the same scheme. 🎉

MapboxCoreNavigationTests/DistanceFormatterTests.swift
```diff
@@ -14,74 +14,144 @@ class DistanceFormatterTests: XCTestCase {
...
+        assertDistance(500,     displayed: "५०० मी॰",     quantity: "५००")
+        assertDistance(949,     displayed: "९५० मी॰",     quantity: "९५०")
+        assertDistance(951,     displayed: "९५० मी॰",     quantity: "९५०")
+        assertDistance(1000,    displayed: "१ कि॰मी॰",      quantity: "१")
+        assertDistance(1001,    displayed: "१ कि॰मी॰",      quantity: "१")
+        assertDistance(2_500,   displayed: "२.५ कि॰मी॰",    quantity: "२.५")
+        assertDistance(2_900,   displayed: "२.९ कि॰मी॰",    quantity: "२.९")
+        assertDistance(3_000,   displayed: "३ कि॰मी॰",      quantity: "३")
+        assertDistance(3_500,   displayed: "४ कि॰मी॰",      quantity: "४")
+        assertDistance(384_400_000, displayed: "३,८४,४०० कि॰मी॰", quantity: "३,८४,४००")
```
1ec5: In case you’re wondering, this is the distance to the Moon, and the first comma separates the [lakhs](https://en.wikipedia.org/wiki/Lakh) place from the ten-thousands place.

MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -161,4 +186,27 @@ public class DistanceFormatter: LengthFormatter {
...
+    @objc public override func attributedString(for obj: Any, withDefaultAttributes attrs: [NSAttributedStringKey : Any]? = nil) -> NSAttributedString? {
+        guard let distance = obj as? CLLocationDistance else {
+            return nil
+        }
+        
+        let string = self.string(from: distance)
+        let attributedString = NSMutableAttributedString(string: string, attributes: attrs)
+        let convertedDistance = distance.converted(to: threshold(for: distance).unit)
+        if let quantityString = numberFormatter.string(from: convertedDistance as NSNumber) {
+            // NSMutableAttributedString methods accept NSRanges, not NSRanges.
```
frederoni: NSRanges, not NSRanges 🤔

1ec5: Thanks, fixed.

PR #1177, opened 2018-03-01
Add support for map matching

bsudekum:
This adds support for map matching. There is not much to change here, but we should add an example for this.

holding on: https://github.com/mapbox/MapboxDirections.swift/pull/236

/cc @mapbox/navigation-ios


MapboxCoreNavigation/RouteController.swift
```diff
@@ -832,8 +832,10 @@ extension RouteController: CLLocationManagerDelegate {
         if let accessToken = routeProgress.route.accessToken, let apiEndpoint = routeProgress.route.apiEndpoint, let host = apiEndpoint.host {
             directions = Directions(accessToken: accessToken, host: host)
         }
+        
+        let opts = NavigationRouteOptions(waypoints: options.waypoints, profileIdentifier: options.profileIdentifier)
```
bsudekum: I'm not a fan of this change. I think if a developer uses map matching, they should actually never use our reroute logic as this would strip all options they have previously applied.

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -55,3 +55,29 @@ open class NavigationRouteOptions: RouteOptions {
         super.init(coder: decoder)
     }
 }
+
+@objc(MBNavigationMatchOptions)
+open class NavigationMatchOptions: MatchingOptions {
```
frederoni: nit: s/Match/Matching

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -55,3 +55,29 @@ open class NavigationRouteOptions: RouteOptions {
...
+        routeShapeResolution = .full
+        attributeOptions = [.congestionLevel, .expectedTravelTime]
+        includesSpokenInstructions = true
+        locale = Locale.nationalizedCurrent
+        distanceMeasurementSystem = Locale.current.usesMetricSystem ? .metric : .imperial
+        includesVisualInstructions = true
+    }
+    
+    @objc public convenience init(locations: [CLLocation], profileIdentifier: MBDirectionsProfileIdentifier? = .automobileAvoidingTraffic) {
+        self.init(waypoints: locations.map { Waypoint(location: $0) }, profileIdentifier: profileIdentifier)
```
JThramer:
```swift
self.init(waypoints: locations.map(Waypoint.init(location:)), profileIdentifier: profileIdentifier)
```

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -55,3 +55,29 @@ open class NavigationRouteOptions: RouteOptions {
...
+    }
+    
+    @objc public convenience init(coordinates: [CLLocationCoordinate2D], profileIdentifier: MBDirectionsProfileIdentifier? = .automobileAvoidingTraffic) {
+        self.init(waypoints: coordinates.map { Waypoint(coordinate: $0) }, profileIdentifier: profileIdentifier)
```
JThramer: Same here.
```swift
self.init(waypoints: locations.map(Waypoint.init(coordinate:)), profileIdentifier: profileIdentifier)
```

Cartfile
```diff
@@ -1,5 +1,5 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 3.7
-github "mapbox/MapboxDirections.swift" ~> 0.17
+github "mapbox/MapboxDirections.swift" "mapmatching"
```
JThramer: 🎗

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -55,3 +55,29 @@ open class NavigationRouteOptions: RouteOptions {
         super.init(coder: decoder)
     }
 }
+
+@objc(MBNavigationMatchOptions)
+open class NavigationMatchOptions: MatchingOptions {
+    
+    @objc public required init(waypoints: [Waypoint], profileIdentifier: MBDirectionsProfileIdentifier? = .automobileAvoidingTraffic) {
+        super.init(waypoints: waypoints.map {
+            $0.coordinateAccuracy = -1
```
JThramer: Can we turn this magic number into a constant? perhaps it might live in an extension? `CLLocationAccuracy.invalid` (or just `.invalid` in this line) would read a lot cleaner than `-1`.

docs/guides/using-map-matching.md
```diff
@@ -0,0 +1,39 @@
+In some cases, a developer may want to have the user stick to a very specific route, one that does not fit into the scope covered by the Mapbox Directions API. For example, a company would like to use their custom truck routing API but allow people to navigate on it with the Mapbox Navigation SDK. In this case, the Mapbox Map Matching API is an appropriate fit.
```
1ec5: The reader is the developer, so use second-person pronouns in this guide.

1ec5: Add a level 1 heading to the top of this file.

docs/guides/using-map-matching.md
```diff
@@ -0,0 +1,39 @@
...
+Map Matching is the art of taking coordinates and aligning them along a road network. In the truck example above, the developer would hit their own truck routing API, give us the coordinates and then we'd return a Route the developer can plug into the Navigation SDK.
+
+When using the Map Matching SDK with the Navigation SDK, there are a few rules you must adhere to:
+
+### `navigationViewController(:shouldRerouteFrom:)` to false
+
+```swift
+func navigationViewController(_ navigationViewController: NavigationViewController, shouldRerouteFrom location: CLLocation) -> Bool {
+        return false
+    }
```
1ec5: Nit: inconsistent indentation.

docs/guides/using-map-matching.md
```diff
@@ -0,0 +1,39 @@
...
+yourNavigationViewController.route = route
+```
+
+This will cause a waterfall effect, everything downstream should react to the addition of a new route.
+
+### Making the request
+
+Always make sure to use `NavigationMatchOptions` when creating a map matching request. This is subclass of `MatchOptions` which applies a good set of default options for navigation.
+
+It is also import to use `Directions.routableMatch(matchOptions:completionHandler:)` when creating the request. This returns a `Route` instead of `Match` which allows us to navigate on it.
```
1ec5: In mapbox/MapboxDirections.swift#236, we settled on `Directions.calculateRoutes(matching:completionHandler:)`.

docs/guides/using-map-matching.md
1ec5: Add this document to jazzy.yml so it appears under “Guides” in the table of contents.

MapboxCoreNavigation/NavigationRouteOptions.swift
JThramer: Posterity: Apologies @bsudekum, I did not realize that there is no natural `Waypoint.init(location:)` initializer. Disregard this and [this.](https://github.com/mapbox/mapbox-navigation-ios/pull/1177#discussion_r172333746)

PR #1178, opened 2018-03-01
Re-enable kCLLocationAccuracyBestForNavigation while phone is plugged in

bsudekum:
Through some first hand tests, I have come to the conclusion that `kCLLocationAccuracyBestForNavigation` is pretty dangerous if the phone is not held very still in, especially in environments where signal is bad. The result is many false positives. IMO, it's almost dangerous enough to warrant only using `kCLLocationAccuracyBest`, baby steps though.

This brings back logic (https://github.com/mapbox/mapbox-navigation-ios/pull/670) that checks whether the phone is plugged in before using `kCLLocationAccuracyBestForNavigation`. 

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -17,6 +17,8 @@ open class NavigationLocationManager: CLLocationManager {
     
     var lastKnownLocation: CLLocation?
     
+    var batteryStateObservation: NSKeyValueObservation?
+    
     /**
      Indicates whether the device is plugged in or not.
      */
@@ -43,11 +45,15 @@ open class NavigationLocationManager: CLLocationManager {
         
         #if os(iOS)
             guard automaticallyUpdatesDesiredAccuracy else { return }
-            let _ = UIDevice.current.observe(\.batteryState) { [weak self] (device, changed) in
+            batteryStateObservation = UIDevice.current.observe(\.batteryState) { [weak self] (device, changed) in
                 guard let weakSelf = self else { return }
                 weakSelf.isPluggedIn = device.batteryState == .charging || device.batteryState == .full
                 weakSelf.desiredAccuracy = weakSelf.isPluggedIn ? kCLLocationAccuracyBestForNavigation : kCLLocationAccuracyBest
             }
         #endif
+    }
+    
+    deinit {
+        batteryStateObservation = nil
     }
 }
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -31,23 +31,12 @@ open class NavigationLocationManager: CLLocationManager {
     override public init() {
         super.init()
         
-        let always = Bundle.main.locationAlwaysUsageDescription
-        let both = Bundle.main.locationAlwaysAndWhenInUseUsageDescription
-        
-        if always != nil || both != nil {
-            requestAlwaysAuthorization()
-        } else {
-            requestWhenInUseAuthorization()
-        }
+        requestWhenInUseAuthorization()
         
         if #available(iOS 9.0, *) {
             if Bundle.main.backgroundModes.contains("location") {
                 allowsBackgroundLocationUpdates = true
             }
-        }
-        
-        if #available(iOS 11.0, *) {
-            showsBackgroundLocationIndicator = true
         }
         
         desiredAccuracy = kCLLocationAccuracyBest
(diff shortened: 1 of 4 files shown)


bsudekum:
This also removes `requestWhenInUseAuthorization()`.

MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -17,24 +17,48 @@ open class NavigationLocationManager: CLLocationManager {
...
             if Bundle.main.backgroundModes.contains("location") {
                 allowsBackgroundLocationUpdates = true
             }
         }
         
-        desiredAccuracy = kCLLocationAccuracyBestForNavigation
+        desiredAccuracy = kCLLocationAccuracyBest
+        
+        #if os(iOS)
+            UIDevice.current.addObserver(self, forKeyPath: "batteryState", options: [.initial, .new], context: nil)
```
frederoni: The block-based KVO is the recommended approach as of Swift 3.2. It's arguably cleaner, contained in one place, and implicitly removed.

https://github.com/mapbox/mapbox-navigation-ios/blob/1356176ea13c17dc6564d8270a10e34808a2019e/MapboxNavigation/NavigationMapView.swift#L228-L230

MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -17,24 +17,37 @@ open class NavigationLocationManager: CLLocationManager {
...
                 allowsBackgroundLocationUpdates = true
             }
         }
         
-        desiredAccuracy = kCLLocationAccuracyBestForNavigation
+        desiredAccuracy = kCLLocationAccuracyBest
+        
+        #if os(iOS)
+            guard automaticallyUpdatesDesiredAccuracy else { return }
+            let _ = UIDevice.current.observe(\.batteryState) { [weak self] (device, changed) in
```
bsudekum: Is it necessary to remove this observation when deiniting?

frederoni: That should be handled automatically when the NavigationLocationManager deinits but there are a couple of radars open indicating this doesn't work properly on iOS 10 and below https://openradar.appspot.com/35027212. I think we should hold onto the return value and invalidate it properly. 😞 
Would you mind invalidating the observer in `NavigationMapView` while you're at it?

JThramer: @bsudekum The new KVO API (`someInstance.observe(_ keyPath:closure:)`) does not require removing observers on deinit. That's one of the nice things about it.

JThramer: I wasn't aware of the radar though. 🤔

And it looks like manually removing an observer is even more of a PITA with the new API, maybe we can still call `NotificationCenter.removeObserver(self)`?

bsudekum: @JThramer I think @frederoni is saying there could be issues with iOS 10 not properly invalidating the property.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -234,6 +234,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     
     deinit {
         suspendNotifications()
+        batteryStateObservation = nil
```
JThramer: Do we really need this observer? It seems that the logic being done on Line [229](https://github.com/mapbox/mapbox-navigation-ios/blob/86b69b45a4519679abfda1e642cdf2fa53e5be86/MapboxNavigation/NavigationMapView.swift#L229) could be handled via a computed property.

```swift
var isPluggedIn: Bool { return [.charging, .full].contains(UIDevice.current.batteryState) }
```

JThramer: In fact, this could be an extension on `UIDevice`. 

```swift
UIDevice.current.isPluggedIn
```

bsudekum:
@JThramer updated, moved to extension.

MapboxNavigation/NavigationMapView.swift
frederoni: We need to observe it to trigger on change of charging state and modify the frame rate accordingly.

JThramer: @frederoni I hear you, but the local `batteryState` property wasn't acted upon on-change (like with a `didSet` or something), the KVO was just updating the local property. Thus you can replace the bound local property with a computed one.

PR #1182, opened 2018-03-02
Mirror constraints for right-to-left layout

### LINKED ISSUE #1070: Various constraints fail to mirror in right-to-left languages

1ec5:
Replaced various left and right constraint anchors with leading and trailing constraint anchors, respectively. Conditionalized some manual layout on the UI layout direction. Set various asymmetrical icons to mirror in right-to-left environments.

turn-by-turn overview steps
feedback thanks

Fixes #1070.

/cc @frederoni @JThramer


MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -119,15 +119,11 @@ extension BaseInstructionsBannerView {
     func setupAvailableBounds() {
         // Abbreviate if the instructions do not fit on one line
         primaryLabel.availableBounds = {
-            let height = ("|" as NSString).size(withAttributes: [.font: self.primaryLabel.font]).height
-            let availableWidth = self.bounds.width-self.maneuverView.frame.maxX-(8*2)
-            return CGRect(x: 0, y: 0, width: availableWidth, height: height)
+            return CGRect(x: 0, y: 0, width: self.bounds.width, height: self.primaryLabel.font.lineHeight)
```
1ec5: This is the reason for all those test fixture changes down below. In effect, the label now extends to the maximum trailing margin instead of stopping at the natural text boundary. That’s typically the best practice for accommodating right-to-left text. Otherwise, the label ended up being abbreviated far too aggressively.

frederoni: I don't think abbreviations will work correctly when using the full width here.
The `availableWidth` should be equal to `bounds.width-maneuverView.bounds.width-BaseInstructionsBannerView.padding*2`.

1ec5: In left-to-right layout, at least, it seems to have the same behavior as on master:

Before | After
----|----
foothill | foothill
glen este | glen este-withamsville
mlk | mlk

1ec5: > I used `maxX` to simplify the following availableWidth format: 
`H:|-18-maneuverView-18-availableWidth-18-|`
so that the left side padding was accounted for by the maneuverView’s`maxX` value.

Doesn’t the maneuver view already get excluded from the label’s bounds?

frederoni: Hmm, `self.bounds` here refers to `InstructionsBannerView.bounds` no? 🤔 
Using `primaryLabel.bounds.width` might work now when we changed from a `lessThanOrEqualTo` constraint to an `equalTo` constraint.

1ec5: Using `primaryLabel` et al. seems to abbreviate too aggressively when the component gets the whole label width:

Expected | Actual
----|----
reference_testprimaryshieldandsecondary 2x | failed_testprimaryshieldandsecondary 2x
reference_testsinglelineprimaryandsecondary 2x | failed_testsinglelineprimaryandsecondary 2x

1ec5: 8465018814ba78e72ba04158ec430c725aeea1dc pares back this change. I didn’t realize originally that `self` was InstructionBannerView rather than the primary or secondary label, but of course that’s the case. The new code switches on the interface direction to avoid overabbreviating in right-to-left layout.

PR #1184, opened 2018-03-02
Fix initial navigation view frame

### LINKED ISSUE #1168: Navigation view should push onto view
@JThramer is this the default animation when pushing the view? Can make it not grow from the bottom center?

new3

frederoni:
Fixes #1168 - NavigationView’s initial frame is set to zero.

@JThramer @bsudekum 👀

### DIFF
--- a/MapboxNavigation/NavigationView.swift
+++ b/MapboxNavigation/NavigationView.swift
@@ -95,10 +95,10 @@ open class NavigationView: UIView {
     lazy var informationStackView = UIStackView(orientation: .vertical, autoLayout: true)
     
     lazy var floatingStackView: UIStackView = {
-        let stack = UIStackView(orientation: .vertical, autoLayout: true)
-        stack.distribution = .equalSpacing
-        stack.spacing = Constants.buttonSpacing
-        return stack
+        let stackView = UIStackView(orientation: .vertical, autoLayout: true)
+        stackView.distribution = .equalSpacing
+        stackView.spacing = Constants.buttonSpacing
+        return stackView
     }()
     
     lazy var overviewButton = FloatingButton.rounded(image: Images.overview)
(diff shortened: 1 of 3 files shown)


bsudekum:
Seeing a few empty views still

new3

bsudekum:
Actually, it looks like you have pulled this out into another PR #1186.

MapboxNavigation/NavigationView.swift
```diff
@@ -193,11 +193,10 @@ open class NavigationView: UIView {
         let informationChildren: [UIView] = [instructionsBannerView, lanesView, nextBannerView, statusView]
         informationStackView.addArrangedSubviews(informationChildren)
         
-        let constraints = informationChildren.flatMap { (view) -> [NSLayoutConstraint] in
-            return [view.leadingAnchor.constraint(equalTo: informationStackView.leadingAnchor),
-                    view.trailingAnchor.constraint(equalTo: informationStackView.trailingAnchor)]
+        informationChildren.forEach {
```
JThramer: Suggestion: I would caution against setting `isActive = true` in `forEach` closures. I've been burned by this in the past, although I don't know why it occurs. Generally if I'm activating more than one constraint at a time I try to use `NSLayoutConstraint.activate(_:)`.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -505,7 +507,7 @@ func defaultFeedbackHandlers(source: FeedbackSource = .user) -> (send: FeedbackV
             self.navigationView.floatingStackView.alpha = 1.0
         }
         
-        let complete: (Bool) -> Void = { self.navigationView.endOfRouteView?.isHidden = true; completion?($0)}
+        let complete: (Bool) -> Void = { self.navigationView.endOfRouteView?.isHidden = true; self.unembedEndOfRoute(); completion?($0)}
```
JThramer: this was a bad habit for me to get into in the first place when it was two commands. Now that it's three, we should really break this out into a multiline closure.

MapboxNavigation/NavigationView.swift
frederoni: It's especially important to activate/deactive
 conflicting constraints simultaneously for example when switching between size classes. However, activating one by one through loadView, viewDidLoad, or initWithFrame is totally safe.

PR #1186, opened 2018-03-02
Initial information stack view state

### LINKED ISSUE #1172: Turn lane view also shows next banner view
simulator screen shot - iphone x - 2018-02-28 at 13 03 55

/cc @JThramer

frederoni:
Possibly fixes #1172 

I have a hard time trying to reproduce 1172. Successfully reproduced one time before this change but not consistently.

@JThramer @bsudekum 👀

### DIFF
--- a/MapboxNavigation/NavigationView.swift
+++ b/MapboxNavigation/NavigationView.swift
@@ -105,12 +105,21 @@ open class NavigationView: UIView {
     lazy var reportButton = FloatingButton.rounded(image: Images.feedback)
     
     lazy var separatorView: SeparatorView = .forAutoLayout()
-    lazy var lanesView: LanesView = .forAutoLayout()
-    lazy var nextBannerView: NextBannerView = .forAutoLayout()
+    lazy var lanesView: LanesView = {
+        let view: LanesView = .forAutoLayout()
+        view.isHidden = true
+        return view
+    }()
+    lazy var nextBannerView: NextBannerView = {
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationView.swift
```diff
@@ -75,12 +75,11 @@ open class NavigationView: UIView {
     }
     
     lazy var mapView: NavigationMapView = {
-        let map: NavigationMapView = .forAutoLayout()
+        let map: NavigationMapView = .forAutoLayout(frame: self.bounds)
```
frederoni: Initializing an `MGLMapView` with a zero rect will expand to 64x64. This can cause weird behavior and slow 
 autolayout animations. @JThramer 👀

MapboxNavigation/UIView.swift
```diff
@@ -70,8 +70,8 @@ extension UIView {
     }
     
 
-    class func forAutoLayout<ViewType: UIView>() -> ViewType {
-        let view = ViewType.init(frame: .zero)
+    class func forAutoLayout<ViewType: UIView>(frame: CGRect = .zero) -> ViewType {
```
JThramer: Suggestion: Add a hidden argument as well, e.g: `forAutoLayout<ViewType: UIView>(frame: CGRect = .zero, hidden: Bool = false)`. That way, you can avoid having to break out `lanesView` and `nextBannerView` into closures.

PR #1188, opened 2018-03-02
Improve RouteController location mechanism

### LINKED ISSUE #1142: Improve progressDidChange + location
Two issues here:

* `RouteController.location` is poorly named. This location is the snapped location. We should make this more clear.
* The location returned from progressDidChange is the raw unsnapped location. When implementing a custom UI, a developer would have to know to reach into the `routeController` to get the snapped location to update the user puck view. 

Proposals:

1. Rename `RouteController.location` to `snappedLocation`
1. Return both the raw location and snapped location on progress updates.

/cc @mapbox/navigation-ios

JThramer:
Work in progress. Fixes #1142, by breaking out route snapping into more clearly defined components, and also by exposing the raw location update to the notification consumer.

To-do: 
- [x] Fix build error
- [x] Document change
- [x] Move the tests over, silly!
- [x]  Expand on changes in `CHANGELOG.md`

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
 * The `DistanceFormatter.attributedString(for:)` method is now implemented. It returns an attributed string representation of the distance in which the `NSAttributedStringKey.quantity` attribute is applied to the numeric quantity. ([#1176](https://github.com/mapbox/mapbox-navigation-ios/pull/1176))
 * Fixed an issue in which turn lanes were displayed in the wrong order when the system language was set to Hebrew. ([#1175](https://github.com/mapbox/mapbox-navigation-ios/pull/1175))
 
+#### MapboxCoreNavigation
+* `RouteController`'s `MBRouteControllerProgressDidChange` notification now exposes the raw location in it's update, accessible by `MBRouteControllerRawLocationKey`
+
 #### Other changes
 * Fixed a crash while navigating that affected applications that do not use Mapbox-hosted styles or vector tiles. [#1183](https://github.com/mapbox/mapbox-navigation-ios/pull/1183)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 #### Breaking changes
 * `NavigationMapViewDelegate` and `RouteMapViewControllerDelegate`: `navigationMapView(_:didTap:)` is now `navigationMapView(_:didSelect:)` [#1063](https://github.com/mapbox/mapbox-navigation-ios/pull/1063)
+* The Constants that concern Route-Snapping logic have been re-named. The new names are: `RouteSnappingMinimumSpeed`, `RouteSnappingMaxManipulatedCourseAngle`, and `RouteSnappingMinimumHorizontalAccuracy`.
 
 #### User interface
 * `StepsViewController` 's convienence initalizer (`StepsViewController.init(routeProgress:)`) is now public. ([#1167](https://github.com/mapbox/mapbox-navigation-ios/pull/1167))
@@ -12,6 +13,7 @@
 * Fixed an issue in which turn lanes were displayed in the wrong order when the system language was set to Hebrew. ([#1175](https://github.com/mapbox/mapbox-navigation-ios/pull/1175))
 
 #### Core Navigation
+* `RoteController` now has a new property, `snappedLocation`. This property represents the raw location, snapped to the current route, if applicable. If not applicable, the value is `nil.`
 * `RouteController`'s `MBRouteControllerProgressDidChange` notification now exposes the raw location in it's update, accessible by `MBRouteControllerRawLocationKey`
 
 #### Other changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 * The `DistanceFormatter.attributedString(for:)` method is now implemented. It returns an attributed string representation of the distance in which the `NSAttributedStringKey.quantity` attribute is applied to the numeric quantity. ([#1176](https://github.com/mapbox/mapbox-navigation-ios/pull/1176))
 * Fixed an issue in which turn lanes were displayed in the wrong order when the system language was set to Hebrew. ([#1175](https://github.com/mapbox/mapbox-navigation-ios/pull/1175))
 
-#### MapboxCoreNavigation
+#### Core Navigation
 * `RouteController`'s `MBRouteControllerProgressDidChange` notification now exposes the raw location in it's update, accessible by `MBRouteControllerRawLocationKey`
 
 #### Other changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * Fixed an issue in which turn lanes were displayed in the wrong order when the system language was set to Hebrew. ([#1175](https://github.com/mapbox/mapbox-navigation-ios/pull/1175))
 
 #### Core Navigation
-* `RoteController` now has a new property, `snappedLocation`. This property represents the raw location, snapped to the current route, if applicable. If not applicable, the value is `nil.`
+* `RoteController` now has a new property, `snappedLocation`. This property represents the raw location, snapped to the current route, if applicable. If not applicable, the value is `nil`.
 * `RouteController`'s `MBRouteControllerProgressDidChange` notification now exposes the raw location in it's update, accessible by `MBRouteControllerRawLocationKey`
 
 #### Other changes
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -73,4 +73,86 @@ extension CLLocation {
         }
         return closestCoordinate.distance < maximumDistance
     }
+    
+    //MARK: - Route Snapping
+    func snapped(to legProgress: RouteLegProgress) -> CLLocation? {
+        var nearByCoordinates = legProgress.nearbyCoordinates
+        
+        // If the upcoming maneuver a sharp turn, only look at the current step for snapping.
+        // Otherwise, we may get false positives from nearby step coordinates
+        if let upcomingStep = legProgress.upComingStep,
+            let initialHeading = upcomingStep.initialHeading,
+            let finalHeading = upcomingStep.finalHeading,
+            let coordinates = legProgress.currentStep.coordinates {
+            
+            // The max here is 180. The closer it is to 180, the sharper the turn.
+            if initialHeading.clockwiseDifference(from: finalHeading) > 180 - RouteControllerMaxManipulatedCourseAngle {
+                nearByCoordinates = coordinates
+            }
+        }
+        
+        guard let closest = Polyline(nearByCoordinates).closestCoordinate(to: coordinate) else { return nil }
+        guard let calculatedCourseForLocationOnStep = interpolatedCourse(along: nearByCoordinates) else { return nil }
+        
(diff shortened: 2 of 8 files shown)


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -73,4 +73,87 @@ extension CLLocation {
...
+    }
+    
+    /**
+     Determines if the a location is qualified enough to allow the user puck to become unsnapped.
+     */
+    func shouldSnap(toRouteWith course: CLLocationDirection) -> Bool {
+        if course >= 0 &&
+            speed >= RouteSnappingMinimumSpeed &&
+            course.differenceBetween(self.course) > RouteSnappingMaxManipulatedCourseAngle &&
+            horizontalAccuracy < 20 {
```
akitchen: Seems like maybe this little bit o' magic should be given a name?

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 * The `DistanceFormatter.attributedString(for:)` method is now implemented. It returns an attributed string representation of the distance in which the `NSAttributedStringKey.quantity` attribute is applied to the numeric quantity. ([#1176](https://github.com/mapbox/mapbox-navigation-ios/pull/1176))
 * Fixed an issue in which turn lanes were displayed in the wrong order when the system language was set to Hebrew. ([#1175](https://github.com/mapbox/mapbox-navigation-ios/pull/1175))
 
+#### MapboxCoreNavigation
+* `RouteController`'s `MBRouteControllerProgressDidChange` notification now exposes the raw location in it's update, accessible by `MBRouteControllerRawLocationKey`
```
bsudekum: I think we should tell more about the change here. `RouteControllerNotificationUserInfoKey.locationKey` is now the snapped location (before it was raw, raw moved to MBRouteControllerRawLocationKey).

MapboxCoreNavigation/RouteController.swift
```diff
@@ -355,10 +355,25 @@ open class RouteController: NSObject {
...
+     */
+    @objc public var location: CLLocation? {
+        return snappedLocation ?? rawLocation
+    }
+    
+    /**
+     The most recently recieved user location, snapped to the route line, if applicable.
+     - important: If the rawLocation is outside of the route snapping tolerances, this value is nil.
+     */
+    var snappedLocation: CLLocation? {
```
bsudekum: Is this variable necessary? Can what's returned here be inlined into `self.location`?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -355,10 +355,25 @@ open class RouteController: NSObject {
         locationManager.stopUpdatingHeading()
     }
     
+    /**
+     The most recently received user location. Snapped to the route line, if applicable, otherwise raw.
+     - seeAlso: snappedLocation, rawLocation
+     */
+    @objc public var location: CLLocation? {
+        return snappedLocation ?? rawLocation
```
bsudekum: It might be best to not fallback here to raw location. It sort of conflates the two keys on the progress update.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -635,7 +561,8 @@ extension RouteController: CLLocationManagerDelegate {
             currentStepProgress.distanceTraveled = distanceTraveled
             NotificationCenter.default.post(name: .routeControllerProgressDidChange, object: self, userInfo: [
                 RouteControllerNotificationUserInfoKey.routeProgressKey: routeProgress,
-                RouteControllerNotificationUserInfoKey.locationKey: location
+                RouteControllerNotificationUserInfoKey.locationKey: self.location!, // Idealized location, raw is guaranteed
+                RouteControllerNotificationUserInfoKey.rawLocationKey: location //raw, from line 541
```
bsudekum: line numbers are all but temporary.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -635,7 +561,8 @@ extension RouteController: CLLocationManagerDelegate {
             currentStepProgress.distanceTraveled = distanceTraveled
             NotificationCenter.default.post(name: .routeControllerProgressDidChange, object: self, userInfo: [
                 RouteControllerNotificationUserInfoKey.routeProgressKey: routeProgress,
-                RouteControllerNotificationUserInfoKey.locationKey: location
+                RouteControllerNotificationUserInfoKey.locationKey: self.location!, // Idealized location, raw is guaranteed
```
bsudekum: Unwrap if you can.

JThramer: The reason why the force-unwrap is used here is because the raw location is guaranteed.

```swift
guard let location = potentialLocation else { return }
         
self.rawLocation = location
```
[Link](https://github.com/mapbox/mapbox-navigation-ios/pull/1188/files/797437a4705bd129245d706acb8f77da16a8f3cb#diff-1a2149692b3d7c58b6a0392120d6011aR541)

JThramer:
(In reply to [this](https://github.com/mapbox/mapbox-navigation-ios/pull/1188#discussion_r171994222), [this](https://github.com/mapbox/mapbox-navigation-ios/pull/1188#discussion_r171994536), and [this](https://github.com/mapbox/mapbox-navigation-ios/pull/1188#discussion_r171994651))

@bsudekum,

The general impetus behind this PR is to clarify what’s happening with snapped locations in the `routeController` and to break out the logic into easy-to-understand components. The most noticeable change is the movement of route-snapping logic to a CLLocation extension, which improves testability and nicely cleans up `RouteController` a bit. 

The second major change here, as you have noticed, is to `RouteController.location` itself, which was refactored to be easier to test, as well as to be clearer as to what’s happening, *while still maintaining existing behavior.* (aka non-breaking change)

Now, `RouteController.location` returns the user’s _idealized_ location, which is snapped to the `currentRoute`, if applicable, otherwise it is the user’s raw location. You’ll notice that’s exactly the same thing that `RouteController.location` was doing before, but now this concept has been formalized in the code. Basically, this property equates to “the location of the user puck on the map”, which I would argue is desirable knowledge to a developer.

The big reason why `snappedLocation` is now it’s own property is because it’s a much clearer pattern than the old way with `RouteController.location`. If you want to know if the user-puck is snapped to the route, now you can just do `if (snappedLocation != nil)`. Before you had to do something like `if (location != rawLocation)`, which isn’t as expressive.

This is better than simply changing `RouteController.location` to `RouteController.snappedLocation` because that property could in-essence be lying to you at times. The code was designed to return the `rawLocation` if it failed the `shouldSnap` logic, at which time it is _not_ a `snappedLocation`. I make this pattern very clear in this PR by having `snappedLocation` return `nil` if the raw location fails the route snapping thresholds, and then `RouteController.location` becomes this concept of an “idealized” location, which is snapped if applicable, raw if not.

The idea behind this new pattern is that there are basically two use-cases our developers are interested in when it comes to ~location~progress updates. Either they want to know where the user puck is on the map, or they want the raw `CLLocation` update from `CLLocationManager`. This PR now provides notification payloads for those two cases directly, through the `locationKey` (idealized location) and `rawLocationKey`.  

If we really hate this pattern I’m open to changing it, but we should discuss just what exactly the ideal solution is here.

CHANGELOG.md
JThramer: Replied [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/1188#issuecomment-370519058)

MapboxCoreNavigation/RouteController.swift
JThramer: Replied [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/1188#issuecomment-370519058)

MapboxCoreNavigation/RouteController.swift
JThramer: Replied [here.](https://github.com/mapbox/mapbox-navigation-ios/pull/1188#issuecomment-370519058)

bsudekum: I'm not sure this was answered. These values are two very different things and i think it's important that if snapped location is nil, we need to say explicitly it's nil. Would you mind answering inline?

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 * The `DistanceFormatter.attributedString(for:)` method is now implemented. It returns an attributed string representation of the distance in which the `NSAttributedStringKey.quantity` attribute is applied to the numeric quantity. ([#1176](https://github.com/mapbox/mapbox-navigation-ios/pull/1176))
 * Fixed an issue in which turn lanes were displayed in the wrong order when the system language was set to Hebrew. ([#1175](https://github.com/mapbox/mapbox-navigation-ios/pull/1175))
 
+#### MapboxCoreNavigation
```
bsudekum: Change this to `Core Navigation`

CHANGELOG.md
bsudekum: I still think we should note this change. `RouteControllerNotificationUserInfoKey.locationKey ` was the unsnapped (essentially,`rawLocation`) location before.  Now it is the snapped location.

MapboxCoreNavigationTests/LocationTests.swift
```diff
@@ -31,4 +40,43 @@ class LocationTests: XCTestCase {
...
     
+    func testSnappedLocation100MetersAlongRoute() {
+        let progress = setup.progress
+        let firstLocation = setup.firstLocation
+        
+        let initialHeadingOnFirstStep = progress.currentLegProgress.currentStep.finalHeading!
+        let coordinateAlongFirstStep = firstLocation.coordinate.coordinate(at: 100, facing: initialHeadingOnFirstStep)
+        let locationAlongFirstStep = CLLocation(latitude: coordinateAlongFirstStep.latitude, longitude: coordinateAlongFirstStep.longitude)
+        guard let snapped = locationAlongFirstStep.snapped(to: progress.currentLegProgress) else {
+            return XCTFail("Location should of snapped to route")
```
vincethecoder: `Location should ?? of ?? snapped to route` -> `Location should have snapped to route` -- maybe?

JThramer: Great catch, @vincethecoder!

MapboxCoreNavigation/RouteController.swift
bsudekum: @JThramer just flagging the above ^

bsudekum:
@JThramer just flagging, a few old comments: https://github.com/mapbox/mapbox-navigation-ios/pull/1188#discussion_r172320195 & https://github.com/mapbox/mapbox-navigation-ios/pull/1188#discussion_r172320920

MapboxCoreNavigation/RouteController.swift
JThramer: Posterity: Resolved the outstanding questions here via a slack call w/ @bsudekum.

MapboxCoreNavigation/Constants.swift
```diff
@@ -59,7 +53,7 @@ public let RouteControllerLinkedInstructionBufferMultiplier: Double = 1.2
 /**
  The minimum speed value before the user's actual location can be considered over the snapped location.
  */
-public var RouteControllerMinimumSpeedForLocationSnapping: CLLocationSpeed = 3
+public var RouteSnappingMinimumSpeed: CLLocationSpeed = 3
```
bsudekum: We should not this was renamed or revert.

PR #1194, opened 2018-03-03
Fix aggressive abbreviation

### LINKED ISSUE #1185: First banner is abbreviated too aggressively
The first banner that appears when entering turn-by-turn navigation is abbreviated as aggressively as possible, probably because the `availableBounds` calculation occurs before the banner reaches its full width:

s<br>
_This should say “South” but says “S” instead._

/cc @frederoni @JThramer

frederoni:
Fixes #1185 

Fixes aggressive abbreviations by using an RTL and layout state-agnostic width calculation.

@1ec5 👀

### DIFF
--- a/MapboxNavigation/InstructionsBannerViewLayout.swift
+++ b/MapboxNavigation/InstructionsBannerViewLayout.swift
@@ -118,25 +118,15 @@ extension BaseInstructionsBannerView {
     
     func setupAvailableBounds() {
         // Abbreviate if the instructions do not fit on one line
-        primaryLabel.availableBounds = {
-            let availableWidth: CGFloat
-            switch UIApplication.shared.userInterfaceLayoutDirection {
-            case .leftToRight:
-                availableWidth = self.bounds.width - self.maneuverView.frame.maxX - 8 * 2
-            case .rightToLeft:
-                availableWidth = self.maneuverView.frame.minX - 8 * 2
-            }
+        primaryLabel.availableBounds = { [unowned self] in
+            // Available width H:|-padding-maneuverView-padding-availableWidth-padding-|
+            let availableWidth = self.bounds.width - BaseInstructionsBannerView.maneuverViewSize.width - BaseInstructionsBannerView.padding * 3
             return CGRect(x: 0, y: 0, width: availableWidth, height: self.primaryLabel.font.lineHeight)
         }
         
-        secondaryLabel.availableBounds = {
-            let availableWidth: CGFloat
-            switch UIApplication.shared.userInterfaceLayoutDirection {
-            case .leftToRight:
-                availableWidth = self.bounds.width - self.maneuverView.frame.maxX - 8 * 2
-            case .rightToLeft:
-                availableWidth = self.maneuverView.frame.minX - 8 * 2
-            }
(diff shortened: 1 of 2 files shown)


1ec5:
Also, don’t forget the next banner.

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -118,25 +118,15 @@ extension BaseInstructionsBannerView {
...
-            let availableWidth: CGFloat
-            switch UIApplication.shared.userInterfaceLayoutDirection {
-            case .leftToRight:
-                availableWidth = self.bounds.width - self.maneuverView.frame.maxX - 8 * 2
-            case .rightToLeft:
-                availableWidth = self.maneuverView.frame.minX - 8 * 2
-            }
+        primaryLabel.availableBounds = { [unowned self] in
+            // Available width H:|-padding-maneuverView-padding-availableWidth-padding-|
+            let availableWidth = self.bounds.width-BaseInstructionsBannerView.maneuverViewSize.width-BaseInstructionsBannerView.padding*3
```
JThramer: Can we put some whitespace in this mathematical expression?

MapboxNavigation/InstructionsBannerViewLayout.swift
```diff
@@ -118,25 +118,15 @@ extension BaseInstructionsBannerView {
...
-            let availableWidth: CGFloat
-            switch UIApplication.shared.userInterfaceLayoutDirection {
-            case .leftToRight:
-                availableWidth = self.bounds.width - self.maneuverView.frame.maxX - 8 * 2
-            case .rightToLeft:
-                availableWidth = self.maneuverView.frame.minX - 8 * 2
-            }
+        secondaryLabel.availableBounds = { [unowned self] in
+            // Available width H:|-padding-maneuverView-padding-availableWidth-padding-|
+            let availableWidth = self.bounds.width-BaseInstructionsBannerView.maneuverViewSize.width-BaseInstructionsBannerView.padding*3
```
JThramer: Here too.

MapboxNavigation/NextBannerView.swift
```diff
@@ -43,14 +43,8 @@ open class NextBannerView: UIView {
...
-        instructionLabel.availableBounds = {
-            let availableWidth: CGFloat
-            switch UIApplication.shared.userInterfaceLayoutDirection {
-            case .leftToRight:
-                availableWidth = self.bounds.width - self.maneuverView.frame.maxX - 16 * 2
-            case .rightToLeft:
-                availableWidth = self.maneuverView.frame.minX - 16 * 2
-            }
+        instructionLabel.availableBounds = { [unowned self] in
+            let availableWidth = self.bounds.width-BaseInstructionsBannerView.maneuverViewSize.width-BaseInstructionsBannerView.padding*3
```
JThramer: One more.

frederoni:
> But does it address #1185?

@1ec5 by using the bounds instead of the frame, it works around the initial zero frame introduced in #1063 and simplifies RTL.
screenshot 2018-03-05 14 15 38

The initial zero frame is being address in #1184

PR #1197, opened 2018-03-05
Remove unnecessary animation

bsudekum:
This animation caused the user puck to slide around when going to and from overhead mode. Removing this causes the puck to remain tighter snapped to the route line.

/cc @mapbox/navigation-ios


bsudekum:
Okay, I found where this does break down- after rerouting. In this case it does not animate and rather jerkily snaps it self back to the route. Investigating.

bsudekum:
Let's go with this for now and see how people like it. I think it's an improvement over the previous behavior.

Issue #1198, opened 2018-03-05
User puck is a little low when starting 


bsudekum:
It fixes it self after a few moments though.

![image](https://user-images.githubusercontent.com/1058624/37006088-7eda247e-208b-11e8-938a-f4dbc4afa693.png)

/cc @JThramer @frederoni

bsudekum:
![simulator screen shot - iphone x - 2018-03-05 at 16 05 19](https://user-images.githubusercontent.com/1058624/37006809-078d244e-208f-11e8-9c17-c3adba0485d7.png)

JThramer:
I _may_ have an idea as to what's happening here. After the `NavigationMapView` appears in a navigation context (`NavigationViewController`), the `contentInset` mutates from `.bottom = 0` to `.bottom = 80.0`. When I override the `contentInset` reference in the `userAnchorPoint` getter and hard-code the inset with a `.bottom = 80.0`, the problem does not reproduce.

bsudekum:
@JThramer You may have some luck by stepping through the commits in https://github.com/mapbox/mapbox-navigation-ios/pull/1186 & https://github.com/mapbox/mapbox-navigation-ios/pull/1184 as i think this bug came out of there.

JThramer:
@bsudekum awesome, will 👀. Thanks for the tip!

JThramer:
So after some investigation, it looks like [MGLMapView](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/src/MGLMapView.mm#L1002) might be involved in this. If `automaticallyAdjustsScrollViewInsets == true`, it will update the `contentInset` based off of it's `parentViewController`s' `topLayoutGuide` and `bottomLayoutGuide`. I bet that manually setting the `bottomLayoutGuide` as if the `WayNameView` was visible (whether it is or not) will fix this issue.

PR #1202, opened 2018-03-06
Hide wayNameView initially

bsudekum:
Fixes the little blue bubble in: https://github.com/mapbox/mapbox-navigation-ios/issues/1198#issuecomment-370612614

Having a hard time reproducing this often, but I think this is a good move nonetheless. 

/cc @JThramer

### DIFF
--- a/MapboxNavigation/NavigationView.swift
+++ b/MapboxNavigation/NavigationView.swift
@@ -117,10 +117,9 @@ open class NavigationView: UIView {
     lazy var resumeButton: ResumeButton = .forAutoLayout()
     
     lazy var wayNameView: WayNameView = {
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationView.swift
```diff
@@ -120,6 +120,7 @@ open class NavigationView: UIView {
         let view: WayNameView = .forAutoLayout()
         view.clipsToBounds = true
         view.layer.borderWidth = 1.0 / UIScreen.main.scale
+        view.isHidden = true
```
JThramer: you should be able to use `.forAutoLayout(hidden: true)` now and avoid that extra line, but dealer's choice.

Issue #1203, opened 2018-03-06
Initializing a NavigationViewController without a global API token crashes the app


Sealos:
**Mapbox Navigation SDK version:** 0.14.1

Hello!,

Initializing a NavigationViewController with custom styles crashes the application.

https://github.com/mapbox/mapbox-navigation-ios/blob/5d29c1313ecb11ec84088f589f73e5c89494b0aa/MapboxNavigation/NavigationViewController.swift#L236

Initializes the voice controller with an empty token, which in place calls the shared instance of the SpeechSynthesizer (Which crashes without a token)

![image](https://user-images.githubusercontent.com/4553825/37024281-aa73af18-2128-11e8-8005-0101d7a119d5.png)

Possible work around would be to accept a RouteVoiceController on the init 
https://github.com/mapbox/mapbox-navigation-ios/blob/5d29c1313ecb11ec84088f589f73e5c89494b0aa/MapboxNavigation/NavigationViewController.swift#L320

Possible related to https://github.com/mapbox/mapbox-navigation-ios/issues/735

bsudekum:
@Sealos thanks for the report, looking into this.

1ec5:
The Mapbox Voice API requires an access token. You can provide it when initializing MapboxVoiceController.

1ec5:
We should probably document somewhere all the places you’d need to set the access token if you don’t set `MGLMapboxAccessToken` in your Info.plist.

bsudekum:
@1ec5 don't you think various Mapbox dependencies should be able to find and use this access token if provided? Why leave it up to the developer to do this?

1ec5:
All the dependencies do automatically discover the access token from the Info.plist. The difficulty (possibly by design) comes when the developer wants to make an exception for one or more Mapbox services.

JThramer:
@1ec5 we could possibly interrogate the `Info.plist`to see if it already has a value set?

```bash
defaults read $(pwd)/Info.plist "MGLMapboxAccessToken" 
```

1ec5:
@JThramer, that part is already working correctly. If the access token is set in the Info.plist, every library and service honors it. @Sealos is attempting to set the access token for everything _except_ the map SDK. I think we need to know more about the use case before making any changes.

Sealos:
> The Mapbox Voice API requires an access token. You can provide it when initializing MapboxVoiceController.

[...]

PR #1204, opened 2018-03-06
Add voiceController to the initialization of NavigationViewController

Sealos:
When only using the MapboxDirections API without setting a global API token, Mapbox Voice crashes in an attempt to use the shared token (Which is not set).

As a workaround, we can initialize the NavigationViewController with a pre-declared controller: MapboxVoiceController(accessToken: <>).

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -233,7 +233,7 @@ public class NavigationViewController: UIViewController {
      
      See `RouteVoiceController` for more information.
      */
-    @objc public var voiceController: RouteVoiceController? = RouteVoiceController()
+    @objc public lazy var voiceController: RouteVoiceController? = MapboxVoiceController()
     
     /**
      Provides all routing logic for the user.
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -233,7 +233,7 @@ public class NavigationViewController: UIViewController {
      
      See `RouteVoiceController` for more information.
      */
-    @objc public var voiceController: RouteVoiceController? = MapboxVoiceController()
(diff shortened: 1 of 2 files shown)


Sealos:
Closes #1203

frederoni:
Thanks! Can you add `voiceController:` to the ObjC signature and update the ObjC example app?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -233,7 +233,7 @@ public class NavigationViewController: UIViewController {
      
      See `RouteVoiceController` for more information.
      */
-    @objc public var voiceController: RouteVoiceController? = MapboxVoiceController()
+    @objc public var voiceController: RouteVoiceController? = RouteVoiceController()
```
bsudekum: Can this default remain as is?

bsudekum:
I think this a valid way to fix the issue. However, this means developers who are not using a global token will have to have the inside knowledge to pass in an initialized `MapboxVoiceController`. 

I'm wondering if we can fix this without having to change the initialize method and somehow grab a token.

Sealos:
@bsudekum Yeah, the variable assignment can be made lazily, so the default value of the voiceController is not run until needed.

JThramer:
@Sealos,

I opened another PR (#1212) that attempts to solve #1203, it's similar to this one, but there is a key difference that may make it advantageous.

* It doesn't change the initialization signature, keeping us from having to introduce a breaking change. Instead, the developer is free to override `NavigationViewController.voiceController` anytime between initialization and `viewWillAppear:`.

Let me know what you think! 🙇 

/cc @bsudekum

Sealos:
@JThramer while I think that is one way of solving it, I think there is value in providing an injectable voiceController. Nonetheless, I'm fine with both solutions 👍

JThramer:
With the Merge of #1212 I will be closing this PR.

PR #1206, opened 2018-03-06
 Make desiredAccuracy a computed property

bsudekum:
Followup to: https://github.com/mapbox/mapbox-navigation-ios/pull/1178#discussion_r172508305

/cc @JThramer @frederoni

### DIFF
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -24,7 +24,6 @@ open class NavigationLocationManager: CLLocationManager {
     @available(*, deprecated, message: "Manually override desiredAccuracy instead.")
     public var automaticallyUpdatesDesiredAccuracy = true
     
-    
     var accuracyOverride: CLLocationAccuracy?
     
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -21,8 +21,21 @@ open class NavigationLocationManager: CLLocationManager {
...
     
+    var accuracyOverride: CLLocationAccuracy?
+    
+    override open var desiredAccuracy: CLLocationAccuracy {
+        get {
+            if let override = accuracyOverride { return override }
+            return  UIDevice.current.isPluggedIn ? kCLLocationAccuracyBestForNavigation : kCLLocationAccuracyBest
+        }
+        set {
+            accuracyOverride = newValue
```
1ec5: What’s the purpose of this override? Now if client code sets `NavigationLocationManager.desiredAccuracy`, the actual `CLLocationManager.desiredAccuracy` is unaffected, but `NavigationLocationManager.desiredAccuracy` lies; `desiredAccuracy` shadows the actual value.

PR #1215, opened 2018-03-09
Prevent double voice instruction before rerouting

bsudekum:
Followup to: https://github.com/mapbox/mapbox-navigation-ios/pull/1121

We look to see if the user is on future steps before rerouting. In cases where the user is actually closer to a future step but also actively rerouting, we should just reroute. This PR prevents false positives when rerouting by using `RouteControllerUserLocationSnappingDistance`. This way, we only advance you to a future step if we are very very confident that you are actually on one.

image

In the case above, we were getting a voice instruction on the step that is headed to the right because it so happens the user is 23.5m of this step before rerouting (rerouting threshold in this case is 25m).

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -640,12 +640,11 @@ extension RouteController: CLLocationManagerDelegate {
         }
         
         // Check and see if the user is near a future step.
-        guard let nearestStep = routeProgress.currentLegProgress.closestStep(to: location.coordinate),
-            nearestStep.index != routeProgress.currentLegProgress.stepIndex else {
+        guard let nearestStep = routeProgress.currentLegProgress.closestStep(to: location.coordinate) else {
             return false
         }
         
-        if nearestStep.distance < radius {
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -640,12 +640,11 @@ extension RouteController: CLLocationManagerDelegate {
...
         
         // Check and see if the user is near a future step.
-        guard let nearestStep = routeProgress.currentLegProgress.closestStep(to: location.coordinate),
-            nearestStep.index != routeProgress.currentLegProgress.stepIndex else {
+        guard let nearestStep = routeProgress.currentLegProgress.closestStep(to: location.coordinate) else {
             return false
         }
         
-        if nearestStep.distance < radius {
+        if nearestStep.distance < RouteControllerUserLocationSnappingDistance {
```
bsudekum: This could also scale with radius as seen in https://github.com/mapbox/mapbox-navigation-ios/pull/1215/commits/6522939aa18041217e795491772e458bbecbbd63

PR #1218, opened 2018-03-12
Enable simulated navigation for tunnel segments.

### LINKED ISSUE #1275: UI can get stuck in night mode after exiting tunnel
We have a few reports of the UI remaining in night mode even after exiting the tunnel.

/cc @vincethecoder

vincethecoder:
This will enable the simulated navigation whenever the commuter enters a tunnel until they exit the tunnel. 
 
Closes #1275 

TODO: 
- [x] Accompanied Issue Ticket
- [x] Tests
- [x] Documentation
- [x] Class dedicated to Simulated Tunnel Navigation


MapboxCoreNavigation/RouteController.swift
```diff
@@ -566,6 +568,15 @@ extension RouteController: CLLocationManagerDelegate {
                 RouteControllerNotificationUserInfoKey.locationKey: self.location!, //guaranteed value
                 RouteControllerNotificationUserInfoKey.rawLocationKey: location //raw
                 ])
+            
+            if let currentIntersection = routeProgress.currentLegProgress.currentStepProgress.currentIntersection,
+                let classes = currentIntersection.outletRoadClasses {
+                if classes.contains(.tunnel) {
+                    startTunnelAnimation(for: manager, routeProgress: routeProgress, distanceTraveled: distanceTraveled)
```
bsudekum: Why is `distanceTraveled` being passed in here?

vincethecoder: When location simulation is needed in special cases such as `dead reckoning`, the `distanceTraveled ` on the given route determines where the `SimulatedLocationManager` needs to begin the simulation.

MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -65,19 +65,40 @@ public class SimulatedLocationManager: NavigationLocationManager {
...
+     Initalizes a new `SimulatedLocationManager` with the given route and distance traveled.
+     
+     When location simulation is needed in special cases such as dead reckoning, the distance
+     traveled on the given route determines where the `SimulatedLocationManager` needs to begin the simulation.
+     
+     - parameter route: The initial route.
+     - parameter distanceTraveled: The distance traveled on the current route.
+     - returns: A `SimulatedLocationManager`
+     */
+    @objc public init(route: Route, distanceTraveled: CLLocationDistance) {
```
frederoni: Would it be more versatile if we would pass a `from:CLLocationCoordinate2D,to:CLLocationCoordinate2D`?
As it stands now, we don't stop simulation after simulating through a tunnel if you would get stuck in the middle, right? Slicing the route could possibly fix that.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -587,6 +605,49 @@ extension RouteController: CLLocationManagerDelegate {
...
     }
+
+    func beginTunnelAnimation(for manager: CLLocationManager, routeProgress: RouteProgress, distanceTraveled: CLLocationDistance) {
+        guard !(manager is SimulatedLocationManager), simulatedLocationManager == nil else { return }
+
+        simulatedLocationManager = SimulatedLocationManager(route: routeProgress.route, distanceTraveled: distanceTraveled)
+        simulatedLocationManager?.delegate = self
+        simulatedLocationManager?.routeProgress = routeProgress
+        
+        let dispatchGroup = DispatchGroup()
```
frederoni: Can you explain the reasoning behind using this dispatch group?

frederoni:
> refactor to use a delegate approach in NavigationLocationManager. This will allow for ReplayLocationManager and SimulatedLocationManager to utilize this feature.

Would replaying using ReplayLocationManager also make use of SimulatedLocationManager?

MapboxCoreNavigation/RouteController.swift
vincethecoder: @frederoni Sure. 

This ensures we don't encounter any concurrency (dispatch) race between the local `locationManager` and the `simulatedLocationManager`. Basically, when we dispatch using a group, we are re-assured that our local `locationManager` will stop `stopUpdatingLocation` before the `simulatedLocationManager ` starts `startUpdatingLocation`. 

This facilitates a smooth handshake between using our `RouteController.locationManager` and `RouteController.simulatedLocationManager` where we will not have both running at the same time or vice versa.

MapboxCoreNavigation/SimulatedLocationManager.swift
vincethecoder: @frederoni a couple of things.
1. We do stop simulation after simulating through a tunnel https://github.com/mapbox/mapbox-navigation-ios/blob/544c5d1822e7e4f9d42809b2220fa555cae684ba/MapboxCoreNavigation/RouteController.swift#L584
2. We don't need to slice as `SimulatedLocationManager` provides us with robust data based on the `currentDistance` (a.k.a. distanceTraveled). Given the `currentDistance`, the manager can determine our current location on the given `route` - which is used to determine factors such as e_xpectedSegmentTravelTimes, closestCoordinateOnRoute, nextCoordinateOnRoute_ and _time_.

https://github.com/mapbox/mapbox-navigation-ios/blob/544c5d1822e7e4f9d42809b2220fa555cae684ba/MapboxCoreNavigation/SimulatedLocationManager.swift#L137

vincethecoder:
@frederoni Yes, I've considered using a delegate approach in `NavigationLocationManager` but it should be tracked in another PR. The delegate approach will make this PR bloated. `NavigationLocationManager` is used by 3 difference sub classes. It makes more sense to just simulate navigation using this approach and get some field tests going to address all possible tunnel simulation conditions. 

Also, `ReplayLocationManager` and `SimulatedLocationManager` use 2 different approaches in simulating navigation as well. We can find some common ground, but it's definitely shouldn't be tracked in this PR.
https://github.com/mapbox/mapbox-navigation-ios/blob/544c5d1822e7e4f9d42809b2220fa555cae684ba/MapboxCoreNavigation/SimulatedLocationManager.swift#L132
https://github.com/mapbox/mapbox-navigation-ios/blob/544c5d1822e7e4f9d42809b2220fa555cae684ba/MapboxCoreNavigation/ReplayLocationManager.swift#L55

MapboxCoreNavigation/SimulatedLocationManager.swift
frederoni: 👍 
I overlooked this part:
https://github.com/mapbox/mapbox-navigation-ios/blob/544c5d1822e7e4f9d42809b2220fa555cae684ba/MapboxCoreNavigation/RouteController.swift#L583-L585
and thought we would just continue simulating to the destination.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -232,7 +232,9 @@ open class RouteController: NSObject {
     }
     
     var userSnapToStepDistanceFromManeuver: CLLocationDistance?
-
+    
+    var simulatedLocationManager: SimulatedLocationManager?
```
frederoni: Can we rename this to tunnelLocationManager or similar to avoid confusion?

vincethecoder: @frederoni this `simulatedLocationManager` will be used for other context of dead reckoning not just tunnel navigation. I had followed up with @brsbl regarding this simulated navigation effort for dead reckoning.

bsudekum: Ok, how about `animatedLocationManager` then? `simulator` should be reserved for the actual simulator in the sdk I think.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -353,6 +355,11 @@ open class RouteController: NSObject {
     @objc public func suspendLocationUpdates() {
         locationManager.stopUpdatingLocation()
         locationManager.stopUpdatingHeading()
+        
+        // In case simulated navigation is abruptly stopped before completion,
```
bsudekum: I think this sentence is missing a few `the`.

MapboxCoreNavigation/RouteController.swift
JThramer: 🤔 maybe we should have a `MapboxLocationManager` to manage all of these real/faux `CLLocationManager` objects? Something to combine the (three now?) native managers (simulated, native, and tunnel) into one concern. There would also be some advantages here in testability, dynamically enabling/disabling route sim, and handoff between one location manager and another, say when entering/exiting a tunnel.

It doesn't have to happen now (or ever), just an idea.
/cc @akitchen

MapboxCoreNavigation/RouteController.swift
vincethecoder: @frederoni Haha yeah - *facepalm*  (an old habit of eliminating a few prepositions to keep the comments concise 😸) Good looks. Thanks 👍

MapboxCoreNavigation/RouteController.swift
vincethecoder: @JThramer I concur. We should have a LocationManager to handle the 3 variations of location manager. @frederoni, @1ec5 and I had a chat and we concluded there is definitely candidacy for `ReplayLocationManager`, `SimulatedLocationManager` to be tied to the location manager, `MapboxLocationManager`.  I intend to track this in a new issue ticket for an upcoming PR. Stay tuned 😺

bsudekum: I still disagree about this. Names are cheap, and this variable name tips off the developer that this `SimulatedLocationManager` is used for simulation.  While we are simulating, this is particular kind of simulation that we should distinguish from others.

Even if this variable will be used for other types of animation in the future, we can still come up with a better name now that does not lead to confusion.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -71,7 +71,7 @@ extension CLLocation {
         guard let closestCoordinate = Polyline(routeStep.coordinates!).closestCoordinate(to: coordinate) else {
             return false
         }
-        return closestCoordinate.distance < maximumDistance
+        return closestCoordinate.distance <= maximumDistance
```
bsudekum: What spurred this change?

vincethecoder: @bsudekum During dead reckoning test, we encountered reroute because the `closestCoordinate.distance` was, `15` -- equal to the `maximumDistance`. Moreover, the documentation for the function indicates, the maximum distance is inclusive.

>  /**
      Returns a Boolean value indicating whether the receiver is within a given distance of a route step, inclusive.
      */

https://github.com/mapbox/mapbox-navigation-ios/pull/1218/files/732ff8b42b477915ef36eafa78d8cf849ff1fb6a#diff-dfd8aa8fe700af70004116a3a54c9d84L67

vincethecoder:
UPDATES: Currently `blocked` due to upcoming changes in #1260 - [Improve snapping at maneuvers](https://github.com/mapbox/mapbox-navigation-ios/pull/1260/files#diff-1a2149692b3d7c58b6a0392120d6011aR613). Necessary test and possible implementation changes will be made to address it.

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -113,4 +134,69 @@ class RouteControllerTests: XCTestCase {
...
+    func testUserWithinTunnelEntranceRadius() {
+        let navigation = tunnelSetup.routeController
+        
+        // Intersection with a tunnel roadClass
+        let intersectionLocation = navigation.routeProgress.route.legs[0].steps[1].intersections![1].location
+        let tunnelEntranceLocation = CLLocation(latitude: intersectionLocation.latitude, longitude: intersectionLocation.longitude)
+        
+        navigation.locationManager(navigation.locationManager, didUpdateLocations: [tunnelEntranceLocation])
+
+        let currentLegProgress = (navigation.routeProgress.currentLegProgress)!
```
bsudekum: nit, isn't `currentLegProgress` already non-optional?

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -113,4 +134,69 @@ class RouteControllerTests: XCTestCase {
...
+        let currentStep = currentLegProgress.currentStep
+        let currentIntersection = currentStep.intersections![currentLegProgress.stepIndex]
```
bsudekum: This is incorrectly. We're using the stepIndex into the intersection array. Don't we already have something on routeProgress that stores the current intersection?

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -113,4 +134,69 @@ class RouteControllerTests: XCTestCase {
...
+        
+        var currentLocation = location(at: navigation.location!.coordinate)
+        var distanceToTunnelEntrance = currentIntersection.location.distance(to: currentLocation.coordinate)
+        
+        var userIsAtTunnelEntranceRadius = navigation.userWithinTunnelEntranceRadius(at: currentLocation, intersection: currentIntersection, distance: distanceToTunnelEntrance)
```
bsudekum: Looking at `userWithinTunnelEntranceRadius`, why is it necessary to pass in `distance`? If we're passing in the user's currentLocation and an intersection, can't we already calculate the distance inside this function with these arguments?

vincethecoder: It's a preference at this point. You can pass it in as a parameter or let the function do calculation. I rather do the former -- making the function lightweight.

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -113,4 +134,69 @@ class RouteControllerTests: XCTestCase {
...
+    // TODO: Test Tunnel Animation Disabled
+    func testTunnelSimulatedNavigation() {
+        
+    }
+    
+    func testEnableTunnelAnimation() {
+        let navigation = tunnelSetup.routeController
+        
+        // Intersection with a tunnel roadClass
+        navigation.advanceStepIndex(to: 1)
```
bsudekum: We should not be advancing the step index with this method. By updating the index with `locationManager(_:didUpdateLocations:)`, we can verify all pieces of the puzzle are working properly.

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -113,4 +134,69 @@ class RouteControllerTests: XCTestCase {
...
+        let intersectionLocation = navigation.routeProgress.currentLegProgress.currentStep.intersections![1].location
+        navigation.routeProgress.currentLegProgress.currentStepProgress.distanceTraveled = intersectionLocation.distance(to: tunnelSetup.firstLocation.coordinate)
```
bsudekum: Again, don't mutate things on RouteProgress in tests. Just supply a new location.

MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -65,19 +65,40 @@ public class SimulatedLocationManager: NavigationLocationManager {
...
+    private func reset(_ distance: CLLocationDistance) {
         if let coordinates = route?.coordinates {
             routeLine = coordinates
             locations = coordinates.simulatedLocationsWithTurnPenalties()
             
-            currentDistance = 0
+            currentDistance = distance
             currentSpeed = 30
+            
+            currentDistance += currentSpeed * speedMultiplier
```
bsudekum: Why are we updating `currentDistance` twice now within this function?

MapboxCoreNavigationTests/RouteControllerTests.swift
bsudekum: But there is no reason to pass it in. I think fewer arguments would making this function easier.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -605,6 +630,91 @@ extension RouteController: CLLocationManagerDelegate {
...
+            suspendTunnelAnimation(for: manager, at: location)
+        }
+    }
+    
+    func shouldEnableTunnelAnimation(at location: CLLocation, for manager: CLLocationManager, intersection: Intersection?) -> Bool {
+        guard let currentIntersection = intersection else { return false }
+        
+        if let classes = currentIntersection.outletRoadClasses {
+            // Main conditions to enable simulated tunnel animation:
+            // - User location is within minimum tunnel entrance radius AND their location speed is at least 5 m/s
```
bsudekum: It doesn't look like this speed check is still happening.

MapboxCoreNavigationTests/RouteControllerTests.swift
vincethecoder: @bsudekum **Nope**. _See screenshot below_:

screen shot 2018-04-04 at 2 05 30 pm

MapboxCoreNavigation/RouteController.swift
vincethecoder: Yes it is. 
https://github.com/mapbox/mapbox-navigation-ios/pull/1218/files#diff-1a2149692b3d7c58b6a0392120d6011aR765

MapboxCoreNavigation/SimulatedLocationManager.swift
vincethecoder: This is for switching `LocationManager` between `NavigationLocationManager` and `SimulatedLocationManager`. You need to update the `currentDistance` in this function for the location updates function - `tick`.

vincethecoder: @bsudekum Yes we need that, to accompany the enhancements made here: https://github.com/mapbox/mapbox-navigation-ios/pull/1218/files/1d6f00b6d7e604dff75a2e57f9e0fab21d2b9ff3#diff-9bf14ebab2a9283285d3a8e91bd09e86R81

MapboxCoreNavigation/RouteController.swift
bsudekum: Okay, let's move the comment to there then.

MapboxCoreNavigation/SimulatedLocationManager.swift
bsudekum: Ok, let's only mutate distance once then.

MapboxCoreNavigationTests/RouteControllerTests.swift
vincethecoder: @bsudekum I think you misunderstood me. When I say, light-weight - it's in regards to the function scope complexity (not parameters). Besides, I like how we just provide all the values and allow the function to just compare and examine values and send a `state`. Just my 2¢

MapboxCoreNavigation/SimulatedLocationManager.swift
vincethecoder: This is the intent of the code. 
`currentDistance = distance + (currentSpeed * speedMultiplier)`
Will be updated to avoid future confusion.

MapboxCoreNavigationTests/RouteControllerTests.swift
akitchen: I think I'm with Vince here - IMO having this clearly outlined in the method signature improves the clarity of intent and potentially reduces implicit coupling.

MapboxCoreNavigationTests/RouteControllerTests.swift
akitchen: I'd be curious to understand why not? There's a place and time to test things working properly together (an integrated test or collaboration test), but there also is value in testing directly at a different seam in the object graph especially when a pleasant higher-level API is available or aspects of the implementation have undesirable side effects.

In this case the SUT has a call to performSelector(_ withDelay:) which is missing a testing shim. Therefore we need to be careful to dance around it a bit under test. Not sure if that's a factor here or not.

We should be careful not to always reach for the lowest-level way of doing things when a higher-level abstraction exists and is appropriate to use. Is it not appropriate in this case or are you more worried about not having integrated test coverage?

MapboxCoreNavigationTests/RouteControllerTests.swift
akitchen: In this case I'm inclined to agree. So would Demeter.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -648,6 +758,24 @@ extension RouteController: CLLocationManagerDelegate {
...
+    public func userWithinTunnelEntranceRadius(at location: CLLocation, intersection: Intersection?, distance: CLLocationDistance?) -> Bool {
+        // Ensure the upcoming intersection is a tunnel intersection.
+        guard let upcomingIntersection = intersection,
+              let roadClasses = upcomingIntersection.outletRoadClasses, roadClasses.contains(.tunnel),
+              (location.speed >= RouteControllerMinimumSpeedAtTunnelEntranceRadius || !location.isQualified) else {
+                return false
+        }
+        
+        // Distance to the upcoming tunnel entrance
+        guard let distanceToTunnelEntrance = distance else { return false }
```
JThramer: This line is interesting. Why not make `distance` a non-optional?

MapboxCoreNavigationTests/RouteControllerTests.swift
bsudekum: Per pairing, it has to be this way since calculating distance inside the function would either calculate absolute distance or require the addition of a coords argument.

MapboxCoreNavigation/RouteController.swift
vincethecoder: @JThramer that is intentional. `routeProgress.currentLegProgress.currentStepProgress.userDistanceToUpcomingIntersection` happens to be an optional value. Rather than using a default value, I'm just going to exit the function if the value of `userDistanceToUpcomingIntersection` happens to be `nil`.

JThramer: Yeah, that's understood, the crux of what I'm getting at here is that "when it comes to functions that answer a question, it's better to resolve optionality of data at the call-site rather than burying that logic in the function."

Sometimes it's okay to handle optionals inside a function, for example when the possibility of non-existence is part of the "semantic domain" of the function itself -- but I would argue that this is not one of those cases, as "the edge-case of the device not having a CLLocation yet" is not within the problem domain of "Is this location within the tunnel-distance threshold?"

So applying this principle, the function would be written more like this: 


```swift
func shouldEnableTunnelAnimation(at location: CLLocation, for manager: CLLocationManager, intersection: Intersection?) -> Bool {
    guard let currentIntersection = intersection else { return false }
    
    let stepProgress = routeProgress.currentLegProgress.currentStepProgress
    
    guard let classes = currentIntersection.outletRoadClasses,
          let upcoming = stepProgress.upcomingIntersection,
          let upcomingDistance = stepProgress.userDistanceToUpcomingIntersection else { return false }
    
    let isWithinTunnelRadius = userWithinTunnelEntranceRadius(at: location, intersection: upcoming, distance: upcomingDistance)
    let isInsideTunnel = (classes.contains(.tunnel) && (manager is NavigationLocationManager && !location.isQualified))
    
    return isWithinTunnelRadius || isInsideTunnel
}
```

Doesn't that seem cleaner?

vincethecoder: @JThramer I agree we can definitely it's okay to handle optionals inside a function. I have 2 goals here. 

Firstly, make my function very loosely-coupled and testable. So with in mind, you will notice the conditional assignment of the `userDistanceToUpcomingIntersection` makes our function very tightly coupled and our test brittle. 

Secondly, my final effort in this PR will be tripping this tunnel work entirely out of the `RouteController` class. My tunnel detector class (coming soon...) should not even know of `routeProgress.currentLegProgress.currentStepProgress`. 

My goal is to make my functions be simple and easy to plug & play.

----
Further analysis:

Per existing logic, we conditionally set `userDistanceToUpcomingIntersection` and I would rather not make this function very dependable on the routeController properties such as `routeProgress`.  We will be ending up making this work so tightly coupled to the RouteController. Which is what we really want to avoid. 

Makes sense?

Code Reference: (Where `userDistanceToUpcomingIntersection` is set.)

https://github.com/mapbox/mapbox-navigation-ios/blob/1d6f00b6d7e604dff75a2e57f9e0fab21d2b9ff3/MapboxCoreNavigation/RouteController.swift#L955

MapboxCoreNavigation/RouteController.swift
```diff
@@ -605,6 +632,103 @@ extension RouteController: CLLocationManagerDelegate {
...
+
+        return false
+    }
+
+    func enableTunnelAnimation(for manager: CLLocationManager,
+                             routeProgress: RouteProgress,
+                          distanceTraveled: CLLocationDistance,
+                                  callback: RouteControllerSimulationCompletionBlock?) {
+       
+        guard !(manager is SimulatedLocationManager), animatedLocationManager == nil else { return }
```
bsudekum: @vincethecoder Ideally we wouldn't have to check the type of class here. This check currently prevents us from replaying routes. 

Are there any other alternate ways we could bail early here?

vincethecoder: @bsudekum hmmmm... why would you want to replay routes here? Further  explanation would certainly help.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -419,7 +419,7 @@ open class RouteStepProgress: NSObject {
     /**
      Returns an array of the calculated distances from the current intersection to the next intersection on the current step.
      */
-    @objc public var intersectionDistances = [CLLocationDistance]()
+    public var intersectionDistances: [CLLocationDistance]?
```
akitchen: Just curious, is it correct to remove the `@objc` declaration here?

vincethecoder: @akitchen I can't recall if the compiler complained. (PS: **I will be adding the @objc declaration back**,) 

Probably, might not apply to this...however, one problem with exposing the `CLLocationDistance` property is that you could store a non-CLLocationDistance-compatible value to it from Objective-C. This could be resolved with computed properties as we've done with other properties in the RouteProgress class. This variable is a stored property.

akitchen: I think it's a good point, but we can't stop all the ways developers can hurt themselves with Objective-C (or in general). Can't lightweight generics help serve at least as a hint in this regard on the ObjC side?

vincethecoder: @akitchen Yes, when I started my development in Swift 4.0, the `@objc` declaration wasn't happy with the optional array of the CLLocation. However, it's fixed in Swift 4.1.

Currently, our CI builds in Xcode 9.2, using the `@objc` here would trigger a failed build on bitrise.

vincethecoder:
Most of the test results will be captured in this [Driving and Testing for Voyage Ticket](https://github.com/mapbox/driver-operations/issues/20).

MapboxCoreNavigation/RouteController.swift
```diff
@@ -237,7 +237,20 @@ open class RouteController: NSObject {
...
     var userSnapToStepDistanceFromManeuver: CLLocationDistance?
-
+    
+    public var tunnelIntersectionManager: TunnelIntersectionManager?
+    
+    var tunnelIntersectionManagerCompletionHandler: RouteControllerSimulationCompletionBlock?
+    
+    public var tunnelSimulationFeatureEnabled: Bool = false
+    
+    private var tunnelAnimationEnabled: Bool {
```
bsudekum: I'd say either make these private if they can be or document them.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -237,7 +237,20 @@ open class RouteController: NSObject {
     }
 
     var userSnapToStepDistanceFromManeuver: CLLocationDistance?
-
+    
+    public var tunnelIntersectionManager: TunnelIntersectionManager?
+    
+    var tunnelIntersectionManagerCompletionHandler: RouteControllerSimulationCompletionBlock?
+    
+    public var tunnelSimulationFeatureEnabled: Bool = false
```
bsudekum: I'd say either make these private if they can be or document them.

MapboxCoreNavigation/RouteController.swift
bsudekum: Where is the bool used?

MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -65,19 +65,34 @@ open class SimulatedLocationManager: NavigationLocationManager {
...
-    private func reset() {
+    private func reset(_ distance: CLLocationDistance) {
         if let coordinates = route?.coordinates {
             routeLine = coordinates
             locations = coordinates.simulatedLocationsWithTurnPenalties()
             
-            currentDistance = 0
-            currentSpeed = 30
+            currentSpeed = 0
+            currentDistance = distance + (currentSpeed * speedMultiplier)
```
bsudekum: I think we should add a new function here for these changes. Also, the units here do not match. We're adding a distance (meters) to a speed (meters per second).

MapboxCoreNavigation/RouteController.swift
bsudekum: Also, rename this to `tunnelSimulationEnabled`.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -586,13 +611,15 @@ extension RouteController: CLLocationManagerDelegate {
...
+            // Check for a tunnel intersection whenever the current route step progresses.
+            checkForTunnelIntersection(at: location, for: manager)
         }
 
         updateDistanceToIntersection(from: location)
         updateRouteStepProgress(for: location)
         updateRouteLegProgress(for: location)
 
-        guard userIsOnRoute(location) || !(delegate?.routeController?(self, shouldRerouteFrom: location) ?? true) else {
+        guard tunnelAnimationEnabled || userIsOnRoute(location) || !(delegate?.routeController?(self, shouldRerouteFrom: location) ?? true) else {
```
bsudekum: I'm slightly concerned with this. If the tunnel animation taking place, and the user is obviously outside of the tunnel, someplace else, we won't reroute them.

MapboxCoreNavigation/TunnelIntersectionManager.swift
```diff
@@ -0,0 +1,146 @@
...
+    }
+    
+    @objc public func suspendTunnelAnimation(for manager: CLLocationManager,
+                                             at location: CLLocation,
+                                         routeController: RouteController,
+                                                callback: RouteControllerSimulationCompletionBlock?) {
+        
+        guard isAnimationEnabled else { return }
+        
+        // Disable the tunnel animation after at least 3 bad location updates.
```
bsudekum: It think it might make more sense to disable tunnel animation after 3 _good_ location updates. If we get 3 bad location updates in a row, we should not do anything IMO.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -263,6 +276,12 @@ open class RouteController: NSObject {
 
         checkForUpdates()
         checkForLocationUsageDescription()
+        
+        tunnelIntersectionManager = TunnelIntersectionManager()
+        tunnelIntersectionManager?.delegate = self
+        tunnelIntersectionManagerCompletionHandler = { enabled, _ in
+            self.tunnelIntersectionManager?.isAnimationEnabled = enabled
+        }
```
bsudekum: It might be cleaner to wrap this in an `initTunnelManager` function.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -546,8 +567,13 @@ extension RouteController: CLLocationManagerDelegate {
         // `filteredLocations` does not contain good locations and we have found at least one good location previously.
         } else if hasFoundOneQualifiedLocation {
             if let lastLocation = locations.last, delegate?.routeController?(self, shouldDiscard: lastLocation) ?? true {
+                
                 // Allow the user puck to advance. A stationary puck is not great.
                 self.rawLocation = lastLocation
+                
+                // Check for a tunnel intersection at the current step we found the bad location update.
+                checkForTunnelIntersection(at: lastLocation, for: manager)
```
bsudekum: I'd move this this out of this if statement and put it below. This function should use [`location`](https://github.com/mapbox/mapbox-navigation-ios/pull/1218/files#diff-1a2149692b3d7c58b6a0392120d6011aR585).

MapboxCoreNavigation/RouteController.swift
```diff
@@ -1064,3 +1107,25 @@ extension RouteController {
...
+    public func tunnelIntersectionManager(_ manager: CLLocationManager,
+                     willEnableAnimationAt location: CLLocation,
+                                           callback: RouteControllerSimulationCompletionBlock?) {
+        tunnelIntersectionManager?.enableTunnelAnimation(for: manager,
+                                             routeController: self,
+                                               routeProgress: routeProgress,
+                                                    callback: callback)
+    }
+    
+    public func tunnelIntersectionManager(_ manager: CLLocationManager,
```
bsudekum: These functions need documentation since they are public.

MapboxCoreNavigation/RouteController.swift
vincethecoder: Sure

MapboxCoreNavigation/SimulatedLocationManager.swift
vincethecoder: Ok

MapboxCoreNavigation/RouteController.swift
vincethecoder: @bsudekum shouldn't the simulation move along with the puck as indicated in the comment above
>  Allow the user puck to advance. A stationary puck is not great.

vincethecoder: @bsudekum also when you say "_I'd move this out of this if statement and put it below._"... do you mean, moving this out of the conditional statement:
` if let lastLocation = locations.last, delegate?.routeController?(self, shouldDiscard: lastLocation) ?? true` ?

MapboxCoreNavigation/RouteController.swift
vincethecoder: @bsudekum Hey, it's already documented in the `TunnelIntersectionManagerDelegate` class: https://github.com/mapbox/mapbox-navigation-ios/pull/1218/files#diff-987c4bbd828e654bd63d59215f890022R20

MapboxCoreNavigation/RouteController.swift
bsudekum: Move to at least here: https://github.com/mapbox/mapbox-navigation-ios/blob/f00ae83b54d47a7d4de7fbf3dd43ad1758a56d6b/MapboxCoreNavigation/RouteController.swift#L589

MapboxCoreNavigation/SimulatedLocationManager.swift
vincethecoder: @bsudekum this is an already existing function. Remember we had a prior discussion: https://github.com/mapbox/mapbox-navigation-ios/pull/1218#discussion_r179231238

Should we probably leave some comment in the code for it, just to prevent a confusion in the future.

MapboxCoreNavigation/TunnelIntersectionManager.swift
vincethecoder: @bsudekum just for the record, this might delay style switches and delay in puck movements in case the driver exits the tunnel very quickly. (PS: I will still proceed with the suggested changes.)

MapboxCoreNavigation/RouteController.swift
vincethecoder: @bsudekum It already resides here: 
https://github.com/mapbox/mapbox-navigation-ios/blob/f00ae83b54d47a7d4de7fbf3dd43ad1758a56d6b/MapboxCoreNavigation/RouteController.swift#L615

The reason the check for tunnel lives here is to ensure whenever:
> `filteredLocations` does not contain good locations and we have found at least one good location previously.

and... to follow suit as we advance the puck as indicated here:
https://github.com/mapbox/mapbox-navigation-ios/pull/1218/files#diff-1a2149692b3d7c58b6a0392120d6011aL549

... unless I'm missing something, here.

PR #1220, opened 2018-03-13
Code-based FeedbackViewController

### LINKED ISSUE #1165: Refactor FeedbackViewController out of storyboard.
Now that the Navigation Scene is out of the storyboard (#1063), we really should bring the `FeedbackViewController` layout out of the storyboard as well.

/cc @mapbox/navigation-ios

JThramer:
WIP. Still a ways to go, not quite in a buildable state yet, but It feels like I'm making progress.

Fixes #1165.

To-Do: 
- [x] Layout `FeedbackCollectionViewCell`
- [x]  Layout `FeedbackViewController`
- [x]  Fix lots of bugs
- [x]  See if we can do layout properly with `estimatedItemSize`
- [x]  Cleanup & Polish

/cc @mapbox/navigation-ios


MapboxNavigation/FeedbackViewController.swift
```diff
@@ -48,20 +63,27 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
     }
     
     class func loadFromStoryboard() -> FeedbackViewController {
-        let storyboard = UIStoryboard(name: "Navigation", bundle: .mapboxNavigation)
-        return storyboard.instantiateViewController(withIdentifier: "FeedbackViewController") as! FeedbackViewController
+        let feedbackVC = FeedbackViewController()
+        return feedbackVC
+//        let storyboard = UIStoryboard(name: "Navigation", bundle: .mapboxNavigation)
+//        return storyboard.instantiateViewController(withIdentifier: "FeedbackViewController") as! FeedbackViewController
```
vincethecoder: We should probably delete these 2 commented lines.

JThramer: Heh, that is a hack, the entire function will be going away.

Note that this is still a WIP, and as-such this code is still kinda crufty.

JThramer:
Update: Code-based FVC Is looking pretty good now...

simulator screen shot - iphone x - 2018-03-14 at 14 42 50

I'm going to take a little bit of time and see if the cell-sizing can be done the _super_-[proper](https://i.imgur.com/166qrIo.gif) way by using `UICollectionViewFlowLayout.estimatedItemSize` instead of the `itemSize` [hack](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/FeedbackViewController.swift#L59) that was introduced in #1063.

Past that, all that remains is cleanup/polish.

MapboxNavigation/FeedbackCollectionViewCell.swift
```diff
@@ -0,0 +1,128 @@
...
+    lazy var titleLabel: UILabel = {
+        let title: UILabel = .forAutoLayout()
+        title.numberOfLines = 2
+        title.textAlignment = .center
+        title.font = .systemFont(ofSize: 18.0)
+        return title
+    }()
+    lazy var circleView: UIView = .forAutoLayout()
+    
+    lazy var circleViewWidthConstraint = {
```
frederoni: Lazy loading the views makes it easier to digest as it isolates the relevant parts, but a lazy var for each of these constant constraints is excessive. It's arguably harder to read, and it increases the loc count by a factor of three for no reason. Can we inline these in `setupConstraints()`?

MapboxNavigation/FeedbackCollectionViewCell.swift
```diff
@@ -0,0 +1,128 @@
...
+        super.init(coder: aDecoder)
+        commonInit()
+    }
+    
+    func commonInit() {
+        setupViews()
+        setupConstraints()
+    }
+    
+    override func awakeFromNib() {
```
frederoni: nit: drop awakeFromNib if we're not using it.

frederoni:
> See if we can do layout properly with estimatedItemSize

Estimated item size is only used for dynamic cell sizing and since we want exactly 3 columns and the size is fixed and known before the collection view calls the receiver’s  `collectionView:cellForItemAt indexPath:`, we don't need to provide an estimated item size.

We could do this with autolayout and make use of `estimatedItemSize` as well, but then we would have to keep track of `H:|-padding-imageView-padding-|` and `V:|-padding-imageView-label-padding-|` which is more convoluted when we already know the size, rows, and columns.

MapboxNavigation/FeedbackCollectionViewCell.swift
```diff
@@ -0,0 +1,97 @@
...
+
+class FeedbackCollectionViewCell: UICollectionViewCell {
+    static let defaultIdentifier = "MapboxFeedbackCell"
+    
+    struct Constants {
+        static let imageSize: CGSize = 70.0
+        static let circleLabelSpacing: CGFloat = 8.0
+    }
+    
+    lazy var imageView: UIImageView = {
```
JThramer: This doesn't need to be a closure-based loader.

JThramer:
Note to future @JThramer: Don't forget to delete the storyboard markup!

MapboxNavigation/FeedbackCollectionViewCell.swift
```diff
@@ -0,0 +1,94 @@
+import Foundation
+import UIKit
+
+
+class FeedbackCollectionViewCell: UICollectionViewCell {
+    static let defaultIdentifier = "MapboxFeedbackCell"
```
vincethecoder: `defaultCellIdentifier` ?

JThramer: I would argue that this would be redundant name-spacing, since we already know it's context (that it's a cells' identifier) at the call-site: `FeedbackCollectionViewCell.defaultIdentifier`

vincethecoder: Yes, let's keep as is.

PR #1223, opened 2018-03-14
Update user course view when in overview mode

frederoni:
This change fixes course view tracking when in overview mode and stops fitting the camera to the remaining route on each location update. Please disregard the low framerate which is just to keep the gif at a reasonable size.

| Before  | After |
| ------------- | ------------- |
|   |  |

@mapbox/navigation-ios 👀


bsudekum:
I sort of think we need to fit the view while in overhead mode.

frederoni:
> I sort of think we need to fit the view while in overhead mode.

I'm ok with that too, but in that case, I think we should limit the scrollable area so you can't fight against the automatic overhead/overview fitting.

bsudekum:
Isn't that what 

https://github.com/mapbox/mapbox-navigation-ios/blob/fe554f5d2eec7e343ec1112508be6d31ae94120d/MapboxNavigation/NavigationMapView.swift#L1116-L1117

prevents?

frederoni:
@bsudekum this is the behavior I was referring to:

they are related but it's not the same thing. I'm just thinking why would we allow the visible bounds in overview mode to be scrollable if we reset it every second? I think the camera should be fixed _or_ that we allow panning in overview mode.

Issue #1216, opened 2018-03-12
User puck is facing wrong direction when moving backwards on step


bsudekum:
This should not happen. The user puck should be facing the correct direction when moving in the wrong direction along a step.

This is especially disorientating when beginning.

1ec5:
The user would appear to be moving backwards if a spurious location update comes in or if the driver happens to shift into reverse, such as to back out of their driveway. There are a couple insights that may help to address this issue:

* When shifting into reverse, the course changes dramatically but the bearing doesn’t. So if we’re sure the user is in a vehicle, then we could rotate the course 180° before using it for the camera or puck.
* Assume the user is always moving forward along the route line. Cap the difference between the route’s heading and the user’s heading to something like 30° while the user is on-route. Only turn the camera around if a series of heading updates shows that the user is actually making a premature U-turn. (In that case, we’d eventually reroute the user anyways: #646 #690.)

bsudekum:
I'm going to try and fix this with a blind eye to users moving in reverse. We could potentially look at the difference between course and bearing, but I still think this a real edge case.

PR #1224, opened 2018-03-14
Fix user puck facing wrong direction when backwards

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1216

This adds a failing test that asserts the interpolated course that is used for deciding the angle of the user puck should be within a few degrees of the angle created between the end and start of step (this step is a simple straight line). Simple illistration:

```
end * ---<=-----------------------------------------* start
```

Where `<=`  is the direction of the user puck.

This should however be:
```
end * ---=>-----------------------------------------* start
```

Note, this example is assuming the user is moving the opposite direction than intended.

Still need to write the code that fixes the issue, first just wanted to create a failing test case.

/cc @mapbox/navigation-ios


bsudekum:
So it turns out it all comes down to the speed value here:

https://github.com/mapbox/mapbox-navigation-ios/blob/60bde302344cbc51e662e37f74153c9c83c914d9/MapboxCoreNavigation/CLLocation.swift#L150-L158

If the user is moving slowly in the wrong direction, the course is the snapped course instead of the raw user's course.

bsudekum:
By removing the speed check we in shouldSnap,

#### Pros
* More accurate puck course when starting a route

#### Cons
* It'd be possible for more false positives to slip through, I.E. downtown stopped at stoplight, we could possibly see more unsnapped pucks

bsudekum:
Okay made some good progress on this. Noting that this will not fix the case when the user is moving backwards slowly along the route mid route. Only when beginning. However, I think this case will be fixed by https://github.com/mapbox/mapbox-navigation-ios/pull/1143.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -147,11 +148,15 @@ extension CLLocation {
...
+        
+        // If the user is near the beginning of leg, allow for unsnapped more often.
+        let isWithinDepatureStep = distanceToFirstCoordinateOnLeg < RouteControllerManeuverZoneRadius
+
         if course >= 0 &&
-            speed >= RouteSnappingMinimumSpeed &&
-            course.differenceBetween(self.course) > RouteSnappingMaxManipulatedCourseAngle &&
-            horizontalAccuracy < RouteSnappingMinimumHorizontalAccuracy {
+            (speed >= RouteSnappingMinimumSpeed || isWithinDepatureStep) &&
+            (horizontalAccuracy < RouteSnappingMinimumHorizontalAccuracy || isWithinDepatureStep) &&
```
vincethecoder: @bsudekum hmmm, we check for `isWithinDepatureStep` twice in both conditions.

bsudekum: @vincethecoder what would you suggest we do otherwise here?

vincethecoder: @bsudekum this could be refactored as:
```
(speed >= RouteSnappingMinimumSpeed || isWithinDepatureStep || horizontalAccuracy < RouteSnappingMinimumHorizontalAccuracy) && ...```

Comment (bsudekum) [reply MapboxCoreNavigation/CLLocation.swift]: I agree that works, but I think breaking it out line by line makes it more clear what is happening here.

Comment (vincethecoder) [reply MapboxCoreNavigation/CLLocation.swift]: We can eliminate the `isWithinDepatureStep` check in one of the 2 conditions.

Comment (bsudekum) [reply MapboxCoreNavigation/CLLocation.swift]: @vincethecoder Right I get that, but I think by breaking it out, line by line, it's easy for someone to glance at what is happening here.

Comment (vincethecoder) [reply MapboxCoreNavigation/CLLocation.swift]: @bsudekum hmmm the concern is `isWithinDepatureStep` checked in both conditions might be covering a defect that we need to address. This is what I mean:
```
(speed >= RouteSnappingMinimumSpeed [evaluates false] || isWithinDepatureStep [evaluates true]) && (horizontalAccuracy < RouteSnappingMinimumHorizontalAccuracy [evaluates false] || isWithinDepatureStep [evaluates true])
```
In this case, we are forcing the 2 conditions to evaluate to `true`, should `isWithinDepatureStep evaluate as `true` (regardless of the left conditions). This is a bit of a concern. Do you see what I mean?

**Proposed fix for readability**:
```
(speed >= RouteSnappingMinimumSpeed || isWithinDepatureStep) &&
(horizontalAccuracy < RouteSnappingMinimumHorizontalAccuracy) && ...
```
this will be less risky and still readable.

bsudekum: @vincethecoder the proposed code will not work. If `(horizontalAccuracy < RouteSnappingMinimumHorizontalAccuracy` evaluates to false, we should not snap if `isWithinDepatureStep`  is true.

I made simple example you can paste into a playground here: https://gist.github.com/bsudekum/bfcab90b6ed9f04ffd141850251235e4

bsudekum: I think we are splitting hairs here though. I think the proposed code is fine as is.

Issue #1192, opened 2018-03-03
Insufficient contrast for severe congestion text in night theme


1ec5:
In the night theme, the text in the bottom bar that indicates severe congestion is a dark red against a dark blue background. This is too little contrast to read on a phone screen at night.

<img src="https://user-images.githubusercontent.com/1231218/36927108-3970fcda-1e30-11e8-98e4-de4e2070b6ce.png" width="300" alt="severe">

/cc @frederoni @vincethecoder

1ec5:
Fixed in #1228.

PR #1228, opened 2018-03-15
Insufficient contrast for severe congestion text in night theme.

vincethecoder:
This fixes the poor contrast of the severe congestion `TimeRemainingLabel` color when navigation night style (theme) is enabled. 

Fixes issue https://github.com/mapbox/mapbox-navigation-ios/issues/1192

/cc @mapbox/navigation-ios 

BEFORE (_left_) `-->`   AFTER (_right_)

dark red color-before              dark red color

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### User Interface
 * Added support for abbreviated top banner instructions. [#1169](https://github.com/mapbox/mapbox-navigation-ios/pull/1169)
 * Reveal the steps list by swiping down on the top banner. [#1150](https://github.com/mapbox/mapbox-navigation-ios/pull/1150)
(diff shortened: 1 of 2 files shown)


JThramer:
👋 just curious, why is the dark-maroon color still visible in the `After` [screenshot](https://user-images.githubusercontent.com/2340650/37487819-18d90da8-2869-11e8-9cf9-5ddc0fc4742d.png)?

vincethecoder:
@JThramer This the summary of the issue ticket:
> In the night theme, the text in the bottom bar that indicates severe congestion is a dark red against a dark blue background. This is too little contrast to read on a phone screen at night.

The new red color applied to the ETA text will not contrast well against `trafficHeavyColor` vs. `trafficSevereColor`. And we agree that the dark-maroon color should be fine as it's visual contrast does justice in distinguishing between `trafficSevereColor` and `trafficHeavyColor`. 

/cc @1ec5

PR #1230, opened 2018-03-16
Reroute proactively, not opportunistically

1ec5:
“Opportunistic” can have a negative connotation. The automatic rerouting feature is just trying to be helpful, proactively. So the following symbols have been renamed:

* `RouteController.reroutesOpportunistically` to `RouteController.reroutesProactively`
* `RouteControllerOpportunisticReroutingInterval` to `RouteControllerProactiveReroutingInterval`
* `RouteControllerNotificationUserInfoKey.isOpportunisticKey` to `RouteControllerNotificationUserInfoKey.isProactiveKey`

/cc @bsudekum

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## Master
 
-#### User Interface
+#### User interface
 * Added support for abbreviated top banner instructions. [#1169](https://github.com/mapbox/mapbox-navigation-ios/pull/1169)
 * Reveal the steps list by swiping down on the top banner. [#1150](https://github.com/mapbox/mapbox-navigation-ios/pull/1150)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,10 +2,13 @@
 
 ## Master
 
-#### User Interface
+#### User interface
```
JThramer: Why are we Caps-casing everything but "User interface"?

1ec5: Everything should be in sentence case. “Core Navigation” is a framework name, so it just looks like title case.

JThramer: Aha! Totally makes sense.

PR #1235, opened 2018-03-19
Avoid truncating distance label

### LINKED ISSUE #1234: Distance to maneuver is truncated in Chinese
When distances are measured in metric and the system language is Chinese, the distance to the upcoming maneuver can get truncated due to insufficient horizontal space. In the example below, “1,000公尺” briefly becomes “,000公尺”, and the label slightly impinges on the space reserved for the secondary instruction label:

Some supported languages have even longer abbreviations for “meter” or “kilometer” than “公尺”. One possible solution would be to dynamically increase the horizontal space allotted to the maneuver icon and distance label based on the distance label’s intrinsic width. (The more fundamental issue is that we’re cramming such an important, dynamic label into such a tight spot, as if it doesn’t really matter.)

/cc @JThramer @frederoni

frederoni:
Fixes #1234 

This change avoids truncating the distance label by:
- Display "1000 m" as "1 km"
- Constrain leading and trailing anchor on distance label to minimum 8 pts from the bounds to activate font scaling
- Decreased minimum font size from 20 to 16

@1ec5


frederoni:
@1ec5 do you have another case which uses a longer unit than "公尺" that we should test?

frederoni:
> Looks good, would we still have the same issue for a distance over 999km? I guess this is pretty rare.

Yes, it will truncate distances over 999 km or 999 miles if the localized unit consists of two characters or more such as "mi", "km", "公尺" with a similar font character width and spacing.

1ec5:
> do you have another case which uses a longer unit than "公尺" that we should test?

#1237

PR #1241, opened 2018-03-20
Use autoreleased observers

frederoni:
`NSNotificationCenter.addObserver(_:selector:name:object:)` removes the observer automatically for apps targeting iOS 9.0 and higher.

@akitchen 👀

### DIFF
--- a/MapboxNavigation/DataCache.swift
+++ b/MapboxNavigation/DataCache.swift
@@ -10,15 +10,10 @@ public class DataCache: NSObject, BimodalDataCache {
         memoryCache.name = "In-Memory Data Cache"
 
         super.init()
-
-        NotificationCenter.default.addObserver(forName: .UIApplicationDidReceiveMemoryWarning, object: nil, queue: nil) { [weak self] (notif) in
-            self?.clearMemory()
-        }
+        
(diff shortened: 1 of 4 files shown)


bsudekum:
Wow, did not know that. @frederoni we remove these observers all over the place, should we also remove others?

frederoni:
> should we also remove others?

It's still good practice to remove observers but the real issue here is that
```swift
NotificationCenter.default.addObserver(forName: .UIApplicationDidReceiveMemoryWarning, object: nil, queue: nil) { [weak self] (notif) in }
```

will not get removed by 
```swift
```

because the observer is not associated with `self`.

You'd need to keep a ref to the rvalue `NSNotificationCenter.addObserver(forName:object:queue:using:)` and remove it accordingly:
```swift
self.observer = NSNotificationCenter.addObserver(forName:object:queue:using:)
NotificationCenter.default.removeObserver(observer)
```

_or_, add the observer on `self` using `self.addObserver(observer:forKeyPath:options:context:)` which doesn't support the closure based api.

akitchen:
Love it. One thing -- let's also port over the test from the other PR. It is a handy regression test to keep around for our future selves.

akitchen:
Oops, I saw that you already did that since you built upon the previous commits. I just moved the tests into their own methods so that the intent is a bit more clear.

Thanks for patching this up @frederoni !!

PR #1242, opened 2018-03-20
Add speed limit indicator

bsudekum:
todo:
- [x] put behind a flag
- [x] decide upon if units are important
- [ ] check logic on deciding current speed
- [ ] expose and refine style
  * night mode

/cc @mapbox/navigation-ios


bsudekum:
Relies on https://github.com/mapbox/MapboxDirections.swift/pull/250

bsudekum:
Added mph/kph in https://github.com/mapbox/mapbox-navigation-ios/pull/1242/commits/2262b082d90bf3df7c7d3335970a56069a3ec83e. Still needs to be localized.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,98 @@ public class ResumeButton: UIControl {
...
+    
+    var speedLimit: SpeedLimit? {
+        didSet {
+            guard let speedLimit = speedLimit else { return }
+            let fontSize = UIFont.systemFontSize
+            
+            let strings = NSMutableAttributedString()
+            
+            strings.append(NSMutableAttributedString(string: defaultText.uppercased(), attributes: [
+                NSAttributedStringKey.paragraphStyle: paragraphStyle,
```
frederoni: nit: `NSAttributedStringKey` is implied and can be dropped. (×7)

frederoni:
So much easier to review UI changes now. Looks good so far. 👍

bsudekum:
For a first pass do we want to handle none? This could probably be done via a pdf/asset.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,98 @@ public class ResumeButton: UIControl {
...
+            
+            strings.append(NSMutableAttributedString(string: defaultText.uppercased(), attributes: [
+                .paragraphStyle: paragraphStyle,
+                .font: UIFont.systemFont(ofSize: fontSize * 0.9)
+                ]))
+            
+            strings.append(NSMutableAttributedString(string: String(describing: speedLimit.speed), attributes: [
+                .font: UIFont.boldSystemFont(ofSize: fontSize * 2),
+                .paragraphStyle: paragraphStyle,
+                .foregroundColor: userIsOverSpeedLimit ? #colorLiteral(red:0.91, green:0.20, blue:0.25, alpha:1.0) : UIColor.black
```
bsudekum: Do we actually want to change the color when over the speed limit?

frederoni: I think we should. I don't know if there's a general rule but in Sweden, it's a 10% tolerance on the speed limit. Changing to red when you're going 61 on a 60 road could be too aggressive but maybe 66 including the 10% tolerance could work?

bsudekum: This could also be a ramp that goes goes from black to red exponentially.

bsudekum: Yeah I'm thinking about moving this into a followup pr.

1ec5:
> For a first pass do we want to handle speed limit == .none?

My understanding is that the most prominent occurrences of this edge case are in Germany, which conforms to the Vienna Convention. Here’s the standard sign indicating the end of all speed limits:

[](https://en.wikipedia.org/wiki/File:Zeichen_282_-_Ende_sämtlicher_Streckenverbote,_StVO_1970.svg)

For a consistent user experience, I think we’d have to somehow switch between MUTCD and Vienna-style speed limit signs, based on the current region, before we could add a case for this image. Otherwise, would we have to show “SPEED LIMIT ∞”? 😏

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -156,4 +156,14 @@ extension CLLocation {
         }
         return true
     }
+    
+    // Speed representation in miles per hour.
+    public var speedMPH: Double {
```
1ec5: I’d recommend spelling out “miles per hour”, but really this shouldn’t be a property of CLLocation. Instead, extend `CLLocationSpeed` with `milesPerHour` and `kilometersPerHour` properties, just as we extend `CLLocationDistance` with `miles` and `kilometers` properties.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -193,6 +193,14 @@ open class RouteLegProgress: NSObject {
     
     @objc public var userHasArrivedAtWaypoint = false
     
+    /**
+     An approximatation of the closest segment on the current leg.
+     */
+    @objc public var currentSegment: Int {
```
1ec5: What does this value mean? A segment is a kind of geometry, not a number.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -296,25 +296,36 @@ public class NavigationViewController: UIViewController {
...
+
     /**
      A Boolean value that indicates whether the dark style should apply when a route controller enters a tunnel.
      */
     @objc public var usesNightStyleInsideTunnels: Bool = false
     
+    /**
+     If true, the maximum speed limit for the segment of road the user is current traveling on will be shown.
+     */
+    @objc public var showMaximumSpeedLimitSign = false {
```
1ec5: Instead of this extra option on NavigationViewController, show the speed limit sign if the route options specifies the `maximumSpeedLimit` attribution option. Don’t add this attribute option automatically; instead, require the developer to add it to NavigationRouteOptions explicitly.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -311,6 +315,33 @@ class RouteMapViewController: UIViewController {
         statusView.show(title, showSpinner: true)
     }
     
+    @objc func progressDidChange(notification: NSNotification) {
+        guard showMaximumSpeedLimitSign else {
+            navigationView.speedLimitSign.isHidden = true
```
1ec5: Move this line above the `guard` statement and change it so that it can show the sign if necessary:

```swift
navigationView.speedLimitSign.isHidden = !showsMaximumSpeedLimitSign || !hasSpeedLimit
```

Then remove the last line of this method. If it’s a problem that we’re showing the sign before performing layout, then stick the line in a `defer` block.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,95 @@ public class ResumeButton: UIControl {
...
+            let fontSize = UIFont.systemFontSize
+            
+            let strings = NSMutableAttributedString()
+            
+            strings.append(NSMutableAttributedString(string: defaultText.uppercased(), attributes: [
+                .paragraphStyle: paragraphStyle,
+                .font: UIFont.systemFont(ofSize: fontSize * 0.9)
+                ]))
+            
+            strings.append(NSMutableAttributedString(string: String(describing: speedLimit.speed), attributes: [
```
1ec5: Use a NumberFormatter so that, for example, a fractional speed limit uses the correct decimal separator.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,95 @@ public class ResumeButton: UIControl {
...
+/// :nodoc:
+@objc(MBSpeedLimitSign)
+open class SpeedLimitSign: UIView {
+    
+    private static let textInsets = UIEdgeInsets(top: 6, left: 6, bottom: 6, right: 6)
+    
+    lazy var label: SpeedLimitLabel = .forAutoLayout()
+    
+    let defaultText = NSLocalizedString("SPEED_LIMIT", bundle: .mapboxNavigation, value: "Speed\nLimit\n", comment: "Speed limit sign main text.")
+    let paragraphStyle = NSMutableParagraphStyle()
```
1ec5: These two properties are never used outside of this class, so make them `static`.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,95 @@ public class ResumeButton: UIControl {
...
+
+extension SpeedUnit {
+    var localizedSpeedUnit: String {
+        switch self {
+        case .kilometersPerHour:
+            return NSLocalizedString("KILOMETERS_PER_HOUR", bundle: .mapboxNavigation, value: "kph", comment: "Inform the user on limits for speed.")
+        case .milesPerHour:
+            return NSLocalizedString("MILES_PER_HOUR", bundle: .mapboxNavigation, value: "mph", comment: "Inform the user on limits for speed.")
+        default:
+            return NSLocalizedString("KILOMETERS_PER_HOUR", bundle: .mapboxNavigation, value: "kph", comment: "Inform the user on limits for speed.")
```
1ec5: `SpeedUnit` has only two cases, so remove this `default`. That way, if MapboxDirections.swift adds another case – say, knots for marine navigation – we’ll get an error here about a missing case.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,95 @@ public class ResumeButton: UIControl {
...
+        super.layoutSubviews()
+        layer.cornerRadius = bounds.midY / 8
+    }
+}
+
+extension SpeedUnit {
+    var localizedSpeedUnit: String {
+        switch self {
+        case .kilometersPerHour:
+            return NSLocalizedString("KILOMETERS_PER_HOUR", bundle: .mapboxNavigation, value: "kph", comment: "Inform the user on limits for speed.")
```
1ec5: “kph” is a relatively rare abbreviation for kilometers per hour on signage. “km/h” is much more common.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,95 @@ public class ResumeButton: UIControl {
...
+        case .milesPerHour:
+            return NSLocalizedString("MILES_PER_HOUR", bundle: .mapboxNavigation, value: "mph", comment: "Inform the user on limits for speed.")
```
1ec5: Nit: If we’re trying to match MUTCD signage, it would be “MPH”, in all caps.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,95 @@ public class ResumeButton: UIControl {
...
+    var speedLimit: SpeedLimit? {
+        didSet {
+            guard let speedLimit = speedLimit else { return }
+            let fontSize = UIFont.systemFontSize
+            
+            let strings = NSMutableAttributedString()
+            
+            strings.append(NSMutableAttributedString(string: defaultText.uppercased(), attributes: [
+                .paragraphStyle: paragraphStyle,
+                .font: UIFont.systemFont(ofSize: fontSize * 0.9)
```
1ec5: In some languages, a translation of this string may not fit given these font metrics. For example, in Puerto Rico (Spanish), the sign says “VELOCIDAD MAXIMA”. In Québec (French), it says “MAXIMUM”.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,95 @@ public class ResumeButton: UIControl {
...
+    }
+    
+    
+    func commonInit() {
+        addSubview(label)
+        paragraphStyle.lineHeightMultiple = 0.9
+        paragraphStyle.alignment = .center
+        layoutMargins = SpeedLimitSign.textInsets
+        label.lineBreakMode = .byWordWrapping
+        label.numberOfLines = 4
```
1ec5: This skeuomorphic design attempts to reproduce the [MUTCD](https://en.wikipedia.org/wiki/Manual_on_Uniform_Traffic_Control_Devices)-style [speed limit sign](https://en.wikipedia.org/wiki/Speed_limit#Maximum_speed_limit) used in the 50 U.S. states and (with modifications) in Canada and Puerto Rico. This is very different than the design used in some form or another everywhere else in the world.

50 km/h

The plan is for the Directions API to indicate which sign design to use (or which country the step is in), allowing for this SDK to switch between different layouts.

Shorter-term, if the API ends up supporting speed limit attributes in Europe first, then we’ll need to rework this code to unconditionally display a Vienna Convention–style sign.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,95 @@ public class ResumeButton: UIControl {
...
+        commonInit()
+    }
+    
+    
+    func commonInit() {
+        addSubview(label)
+        paragraphStyle.lineHeightMultiple = 0.9
+        paragraphStyle.alignment = .center
+        layoutMargins = SpeedLimitSign.textInsets
+        label.lineBreakMode = .byWordWrapping
```
1ec5: Use three separate labels in a view. That way we can avoid the complexities of NSAttributedString while gaining more control over the layout.

MapboxNavigation/Style.swift
1ec5: Tolerance varies by jurisdiction. In some localities in the U.S., the tolerance is 0&nbsp;mph; in others, it’s 5&nbsp;mph or higher. I don’t think the threshold is something we should determine on the client side. Moreover, the Directions API doesn’t yet distinguish advisory speed limits from legal speed limits.

MapboxNavigation/NavigationViewController.swift
bsudekum: I think we have a good pattern right now of not making developers mutate `NavigationRouteOptions`. I see your logic, but I'm also thinking of time when we use max speeds in other parts of sdk.

MapboxNavigation/Style.swift
bsudekum: We should probably omit fractions huh?

MapboxNavigation/NavigationViewController.swift
1ec5: So until then, developers who leave this setting `false` will nonetheless request speed limit attributes from the server?

MapboxNavigation/Style.swift
1ec5: So if the speed limit is [7½&nbsp;mph](http://overpass-turbo.eu/s/xfT), would we round up or down?

[](http://www.alastaircaldwell.com/alastairs-blog/transam-rally-june-july-2015)

1ec5: More to the point, once we add [an Arabic localization](https://www.transifex.com/mapbox/mapbox-navigation-ios/language/ar/), we’d use Eastern Arabic numerals:

[](https://commons.wikimedia.org/wiki/File:UAE_Speed_Limit_-_60_kmh.svg)

MapboxCoreNavigation/RouteProgress.swift
bsudekum: It describes the user's current index into coordinates or annotations on a leg.

MapboxNavigation/Style.swift
bsudekum: Sounds good to me. I'm thinking of making the view a UIStackView, then appending UILabels per word that is necessary to visualize the sign. Any objections to this thinking?

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,115 @@ public class ResumeButton: UIControl {
...
+        stackView.distribution = .equalSpacing
+        addSubview(stackView)
+        stackView.layoutIfNeeded()
+    }
+    
+    func updateStackView(with labels: [SpeedLimitLabel]) {
+        stackView.subviews.forEach { $0.removeFromSuperview() }
+        stackView.addArrangedSubviews(labels)
+        stackView.layoutIfNeeded()
+    }
```
bsudekum: Converted this to a stack view, and things are no longer really working. @JThramer @frederoni does anything look funny to you here?

MapboxCoreNavigation/RouteProgress.swift
1ec5: The comment and property name are both misleading, then. It’s the current segment _index_.

1ec5: This is very different than how we handle per-segment congestion attributes. I think we should unify the approach: either distribute the speed limit attributes among each step, or change the congestion logic to depend on this `currentSegmentIndex` property and stop distributing the attributes among the steps.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,115 @@ public class ResumeButton: UIControl {
...
+                ])
+            labels.append(speedLabel)
+            
+            let unitLabel: SpeedLimitLabel = .forAutoLayout()
+            unitLabel.attributedText = NSMutableAttributedString(string: "\n\(speedLimit.unit.localizedSpeedUnit)", attributes: [
+                .font: UIFont.boldSystemFont(ofSize: fontSize / 1.3),
+                .paragraphStyle: paragraphStyle
+                ])
+            labels.append(unitLabel)
+
```
vincethecoder: @bsudekum I see a bit of code repetition in the 3 blocks of codes for `defaultLabels `, `speedLabel` and `unitLabel`. Could we keep it DRY?

MapboxNavigation/Style.swift
vincethecoder: @bsudekum what do you mean?
>  things are no longer really working...

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -311,6 +315,23 @@ class RouteMapViewController: UIViewController {
         statusView.show(title, showSpinner: true)
     }
     
+    @objc func progressDidChange(notification: NSNotification) {
```
frederoni: `NavigationViewController` also observes `.routeControllerProgressDidChange` and calls `RouteMapViewController.notifyDidChange(routeProgress:location:secondsRemaining:)`. Let's fold the old `notifyDidChange` into this method in order to avoid having two inputs of the same data coming from two different view controllers.

MapboxNavigation/Style.swift
```diff
@@ -209,6 +210,115 @@ public class ResumeButton: UIControl {
     }
 }
 
+/// :nodoc:
+@objc(MBSpeedLimitSign)
```
frederoni: This class is big enough (and will only get bigger with full MUTCD and Vienna convention conformance) to justify a separate file.

MapboxNavigation/NavigationViewLayout.swift
```diff
@@ -45,6 +45,10 @@ extension NavigationView {
         
         wayNameView.centerXAnchor.constraint(equalTo: centerXAnchor).isActive = true
         wayNameView.bottomAnchor.constraint(equalTo: bottomBannerView.topAnchor, constant: -10).isActive = true
+        
+
```
vincethecoder: extra space can be removed 😸

MapboxNavigation/SpeedLimit.swift
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+extension SpeedUnit {
+    var localizedSpeedUnit: String {
+        switch self {
+        case .kilometersPerHour:
+            return NSLocalizedString("KILOMETERS_PER_HOUR", bundle: .mapboxNavigation, value: "kph", comment: "Inform the user on limits for speed.")
+        case .milesPerHour:
+            return NSLocalizedString("MILES_PER_HOUR", bundle: .mapboxNavigation, value: "mph", comment: "Inform the user on limits for speed.")
```
vincethecoder: @bsudekum perhaps the comments should be unique to each localized string? `Inform the user on limits for speed` should probably indicate one is `mph` or `kph` 💭

MapboxNavigation/SpeedLimit.swift
```diff
@@ -0,0 +1,159 @@
...
+    required public init(coder: NSCoder) {
+        super.init(coder: coder)
+        commonInit()
+    }
+    
+    func commonInit() {
+        axis = .vertical
+        alignment = .center
+        spacing = 0.5
+        if region == .world {
```
bsudekum: I'm not excited how this class is turning out. I think it might make more sense to extract these if statements out further. If @mapbox/navigation-ios has any suggestions, im all ears.

1ec5:
The saga continues in #2291.

PR #1169, opened 2018-02-28
Add abbreviations

bsudekum:
_Relies on: https://github.com/mapbox/MapboxDirections.swift/pull/244_

This adds abbreviations from the server.

todo: 

- [x] Figure out how to use the `abbr_priority` key. Right now, we're just abbreviating in order they are given depending on if the text can fit. 
- [x] Add tests

/cc @mapbox/navigation-ios @danpaz

PR #1244, opened 2018-03-21
Update changelog

1ec5:
Copyedited and updated the changelog to reflect all the developer- or user-impacting changes that have occurred since the previous release, v0.15.0. The blurbs for #1127 had gotten buried under the v0.14.0 section, which was the newest section at the time that PR was first opened. For #1169, I focused on the delimiter bug fix rather than the feature work: while I don’t want to discount the huge amount of work that went into improving banner abbreviations under the hood, abbreviations aren’t new to this library. If that PR works as designed, then there should be no developer or user impact. A couple other notable changes had snuck in without blurbs, so they’ve been added too.

/cc @vincethecoder @bsudekum


CHANGELOG.md
```diff
@@ -1,15 +1,21 @@
...
-## Master
+## master
 
-#### User interface
-* Added support for abbreviated top banner instructions. [#1169](https://github.com/mapbox/mapbox-navigation-ios/pull/1169)
-* Reveal the steps list by swiping down on the top banner. [#1150](https://github.com/mapbox/mapbox-navigation-ios/pull/1150)
-* Fixed the poor contrast of the `TimeRemainingLabel.appearance().trafficSevereColor`, when navigation night style (theme) is enabled. [#1228](https://github.com/mapbox/mapbox-navigation-ios/pull/1228)
-* Data representing spoken voice instructions is now cached locally for reuse. [#1229](https://github.com/mapbox/mapbox-navigation-ios/pull/1229)
+### User interface
+* `DayStyle` is temporarily replaced with `NightStyle` while the user travels through a tunnel. ([#1127](https://github.com/mapbox/mapbox-navigation-ios/pull/1127))
```
frederoni: This is not entirely correct. Styles with `styleType` set to `.night` is temporarily applied while the user travels through a tunnel.

Issue #1196, opened 2018-03-05
Snapping user puck can happen with unqualified locations


bsudekum:
If a location update is unqualified, I think we should still seek to snap the user puck. Other functions we call on every location update (userIsOnRoute, advancing the step, etc) should still use only qualified location updates.

This will prevent situations where the user puck remains at a standstill for many location updates.

/cc @mapbox/navigation-ios

PR #1247, opened 2018-03-21
Allow for user puck advancement when location is unqualified

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1196

Starting by writing a failing test which shows that an unqualified location is not snapped. ~~The actual change here should be minor and will followup soon.~~ Added fix to test in https://github.com/mapbox/mapbox-navigation-ios/pull/1247/commits/75f27f305820bc3518783b1b5fd6420d3d743b94.

/cc @mapbox/navigation-ios


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -5,13 +5,9 @@ import Turf
 extension CLLocation {
     
     var isQualified: Bool {
-        #if (arch(i386) || arch(x86_64)) && (os(iOS) || os(watchOS) || os(tvOS))
```
bsudekum: Now that `isQualified` does not check user course, it's safe to make this check while in the sim.

bsudekum:
@vincethecoder could you check this out?

vincethecoder:
@bsudekum I actually had 👀 on that too... cool!

PR #1248, opened 2018-03-21
Remove log

bsudekum:
This log was getting printed quite often. I think we can silently continue if the audio is not found on disk. 

/cc @mapbox/navigation-ios


akitchen:
Note to future selves that this is where we might want to instrument cache misses in the future.

Issue #1246, opened 2018-03-21
User puck can become uncentered after rerouting


bsudekum:
![simulator screen shot - iphone x - 2018-03-21 at 10 06 43](https://user-images.githubusercontent.com/1058624/37725313-f1264952-2cef-11e8-9f16-1e518a2838bd.png)

Reverting this line back to out of the else fixes the issue:

https://github.com/mapbox/mapbox-navigation-ios/blob/2425fcb5d9a330c099a0fcfceff070e55227fc44/MapboxNavigation/NavigationMapView.swift#L353-L355

/cc @frederoni

PR #1251, opened 2018-03-22
Fix center point

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1246

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -176,7 +176,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             }
             
             if let location = userLocationForCourseTracking {
-                updateCourseTracking(location: location, animated: true, forceUpdatePuck: true)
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -176,13 +176,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             }
             
             if let location = userLocationForCourseTracking {
-                updateCourseTracking(location: location, animated: true)
-            }
-            
-            if let location = userLocationForCourseTracking, tracksUserCourse {
-                UIView.animate(withDuration: 1, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
-                    self.userCourseView?.center = self.convert(location.coordinate, toPointTo: self)
```
bsudekum: I don't think we need to call this twice. This is called just a few lines above in `updateCourseTracking `

Issue #1050, opened 2018-01-20
Stop using Transifex for abbreviations


1ec5:
We’ve been shoehorning the list of turn banner abbreviations into Transifex, with the assumption that translators would be more inclined to provide them directly in the same site as all the other localizable strings than to open a pull request here. But the Transifex interface isn’t well suited for tables – indeed, we have to convert from plist to an ad-hoc format just to be flexible about the number of abbreviations that a localization can provide. While we do document the format, the documentation is undiscoverable in Transifex’s interface. Translators have complained about the format in the past.

We could move the abbreviations into the [OSRM Text Instructions](https://github.com/Project-OSRM/osrm-text-instructions/) repository and pull them into this SDK as part of the release process. Instead of pushing the table up to Transifex, we’d periodically encourage translators at Transifex to head over to GitHub to also provide abbreviations. (Google Documents would have a lower barrier to entry for translators, but I think we’d prefer GitHub-style version control.) If JSON would be too unfriendly to translators, OSRM Text Instructions could use CSV format and convert it to JSON at build time.

This would be a first step toward moving the abbreviation functionality to the server side. (Picture the Directions API providing a list of alternative forms for a word, and the navigation SDK choosing whichever one fits.) That would make it possible for the abbreviations to be based on the language of the region through which the route runs, rather than the user’s current system language. It would also make it much easier for the Android navigation SDK to abbreviate words in the turn banner.

* [x] Commit existing abbreviation files to OSRM Text Instructions: Project-OSRM/osrm-text-instructions#221
* [x] Commit partial abbreviation files to OSRM Text Instructions for languages that haven’t reached 80% completion for the iOS navigation SDK: Project-OSRM/osrm-text-instructions#226
* [x] Delete the abbreviation files from this repository: mapbox/mapbox-navigation-ios#1169
* [x] Delete the abbreviation files from the iOS navigation SDK project on Transifex
* [x] Implement abbreviation in the Directions API response
* [x] Switch over to server-side abbreviations: mapbox/mapbox-navigation-ios#1169
* [x] Document the new abbreviation file format: Project-OSRM/osrm-text-instructions#227
* [x] Notify translators on Transifex that the abbreviation files have moved

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/1047#issuecomment-359126811
/cc @mapbox/navigation @mapbox/guidance

bsudekum:
[...]

PR #1252, opened 2018-03-22
Add Arabic, European Portuguese localizations

1ec5:
Added a partial Arabic localization and a complete European Portuguese localization from Transifex. Thanks to @safaalfulaij for the Arabic translations and @Rui-Santos for the European Portuguese translations!

Arabic European Portuguese

In order to pull these localizations from Transifex, some remnants of the old `localizableabbreviations` resource needed to be removed from the `tx` configuration. (This resource went away as part of #1050.)

/cc @vincethecoder @frederoni

### DIFF
--- a/docs/guides/Localization and Internationalization.md
+++ b/docs/guides/Localization and Internationalization.md
@@ -48,7 +48,7 @@ The upcoming road or ramp destination is named according to the local or nationa
 | Hebrew     | ✅              | ✅                             | —                     | ✅
 | Hungarian  | ✅              | —                             | —                     | ✅
 | Italian    | ✅              | ✅                             | ✅                     | ✅
-| Portuguese | ✅              | ✅                             | ✅                     | ✅
+| Portuguese | ✅             | ✅                             | ✅                     | ✅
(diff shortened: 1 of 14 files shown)


docs/guides/Localization and Internationalization.md
```diff
@@ -47,7 +48,7 @@ The upcoming road or ramp destination is named according to the local or nationa
 | Hebrew     | ✅              | ✅                             | —                     | ✅
 | Hungarian  | ✅              | —                             | —                     | ✅
 | Italian    | ✅              | ✅                             | ✅                     | ✅
-| Portuguese | ✅              | ✅                             | ✅                     | ✅
+| Portuguese | ✅             | ✅                             | ✅                     | ✅
```
bsudekum: Huh, so we noted we already had Portuguese before this addition?

1ec5: Yes, we already have a Brazilian Portuguese localization. I considered spelling out the availability of both variants, but that might make more sense once we make European Portuguese spoken instructions available too.

Issue #685, opened 2017-10-04
Expose more MGLMapViewDelegate methods


frederoni:
We only delegate `mapView(_:viewFor:)` and `mapView(_:imageFor:)` to the `NavigationViewControllerDelegate` currently. Let's add a few more requested (#666) delegate methods.

- [ ] `mapView(_:regionWillChangeAnimated:)`
- [ ] `mapView(_:regionDidChangeAnimated:)`
* [ ] `mapView(_:didFinishLoading:)` (both methods)
* [ ] `mapView(_:viewForAnnotation:)`
* [ ] `mapView(_:didAddAnnotationViews:)`
* [ ] `mapView(_:annotationCanShowCallout:)`
* [ ] `mapView(_:tapOnCalloutForAnnotation:)`

@bsudekum @1ec5

1ec5:
NavigationMapView currently seems to implement only `mapView(_:imageFor:)` and `mapView(_:viewFor:)`, and it does so by turning around and ultimately asking the NavigationViewControllerDelegate’s implementation of those methods. So does NavigationMapViewController even need to be NavigationMapView’s delegate? If not, we can remove all these pass-throughs and let the developer set the map view’s delegate to their own view controller.

frederoni:
@1ec5 It would be great if we could expose all of MGLMapViewDelegate but the RouteMapViewController needs `mapView:didFinishLoadingStyle:` and `mapView:regionDidChangeAnimated:`. 🤔

1ec5:
We could imitate `-mapView:didFinishLoadingStyle:` by key-value observing the `style` key path. We could move RouteMapViewController off `-mapView:regionDidChangeAnimated:` using `didSet` handlers on various properties like `tracksUserCourse`.

akitchen:
see also https://github.com/mapbox/mapbox-navigation-ios/issues/1526

lemwerks:
@akitchen This is something [Roadtrippers](https://github.com/mapbox/support/issues/741) brought up. Their goal is to add annotations you can tap on to open for more details. They can add them to the map, but annotations don’t seem to open. 

@1ec5 made some suggestions that weren't capable at this time. Flagging here to a: give you an idea of the need of this from a long time Enterprise Customer and they want to use the Nav SDK and b. Keep tabs on this development. 

[...]

PR #1256, opened 2018-03-23
Fix custom image for annotation

bsudekum:
Fixes: https://github.com/mapbox/navigation-ios-examples/pull/16#issuecomment-375712695

Unsure how this regressed.

/cc @mapbox/navigation-ios @maxhawkins

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -1210,6 +1210,24 @@ public protocol NavigationMapViewDelegate: class {
     optional func navigationMapView(_ mapView: NavigationMapView, shapeFor waypoints: [Waypoint]) -> MGLShape?
     
     /**
+     Asks the receiver to return an MGLAnnotationImage that describes the image used an annotation.
+     - parameter mapView: The MGLMapView.
+     - parameter annotation: The annotation to be styled.
+     - returns: Optionally, a `MGLAnnotationImage` that defines the image used for the annotation.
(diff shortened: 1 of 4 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -1209,6 +1209,24 @@ public protocol NavigationMapViewDelegate: class {
...
     optional func navigationMapView(_ mapView: NavigationMapView, shapeFor waypoints: [Waypoint]) -> MGLShape?
     
+    /**
+     Asks the receiver to return an MGLAnnotationImage that describes the image used an annotation.
+     - parameter mapView: The MGLMapView.
+     - parameter annotation: The annotation to be styled.
+     - returns: Optionally, a `MGLAnnotationImage` that defines the image used for the annotation.
+     */
+    @objc(navigationMapView:imageForAnnotation:)
+    optional func navigationMapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?
```
1ec5: FYI, this implements part of #685 but takes us further away from being able to allow the developer to just use any MGLMapViewDelegate method, as suggested in https://github.com/mapbox/mapbox-navigation-ios/issues/685#issuecomment-334294064.

bsudekum: Yeah I agree, however lets make the breaking change for all methods in a single pr.

PR #1257, opened 2018-03-23
Use heading when course is invalid

bsudekum:
Now since https://github.com/mapbox/mapbox-navigation-ios/pull/1224 is in, we are unsnapped more at the beginning of the route. When at the beginning of a route, many times the user is still on a wifi network which causes the course to be -1. In this case, the user puck will just point north, which is not a great user experience. 

This change uses `CLLocationHeadin.trueHeading` to override the user's course when it is -1. As a reminder, this will only occur when:

* The user puck is unsnapped.
* the user's course equal to -1.

This change is pretty controversial since we really do not like to rely on heading here but I think this is a pretty good usability change users will enjoy.

/cc @mapbox/navigation-ios


bsudekum:
Going to add a test case for this.

bsudekum:
On second thought, this code is too heavy handed. This should only occur when `snappedLocation` is nil.

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -81,4 +83,36 @@ class RouteControllerTests: XCTestCase {
...
+    func testLocationShouldUseHeading() {
+        let navigation = setup.routeController
+        let firstLocation = setup.firstLocation
+        navigation.locationManager(navigation.locationManager, didUpdateLocations: [firstLocation])
+        
+        XCTAssertEqual(navigation.location!.course, firstLocation.course, "Course should be using course")
+        
+        let invalidCourseLocation = CLLocation(coordinate: firstLocation.coordinate, altitude: firstLocation.altitude, horizontalAccuracy: firstLocation.horizontalAccuracy, verticalAccuracy: firstLocation.verticalAccuracy, course: -1, speed: firstLocation.speed, timestamp: firstLocation.timestamp)
+        
+        let heading = CustomHeading()
```
bsudekum: There is no initializer for CLHeading so I created my own. Unsure about this move.

vincethecoder: @bsudekum I'm reviewing this PR. Btw, what are you unsure of?

bsudekum: @vincethecoder whether we want to use this custom class I created or if we should somehow create our own initializer.

bsudekum:
Noting, this worked well this weekend.

MapboxCoreNavigationTests/RouteControllerTests.swift
vincethecoder: @bsudekum rather than a subclass, I would create an `extension` of `CLHeading` like so:

```
extension CLHeading {
    
    var trueHeading: CLLocationDirection {
        set {
            mbTrueHeading = newValue
        }
        get {
            return mbTrueHeading
        }
    }

    private var mbTrueHeading: CLLocationDirection {
        set {
            trueHeading = newValue
        }
        get {
            return trueHeading
        }
    }
}
```

frederoni: Unfortunately, CLHeading’s properties are not mutable so I don't think we can use an extension for that. Since this is in the test target, we could expose CLHeading’s private initializer in ObjC.

```objc
@interface CLHeading (Private)
- (id)initWithHeading:(double)arg1 accuracy:(double)arg2;
@end
```

vincethecoder: @frederoni 
> Unfortunately, CLHeading’s properties are immutable

Yes, indeed. My suggestion was, 

> Since this is in the test target...

We could get away with either one of these suggested approaches:

```
fileprivate var mbTrueHeading: CLLocationDirection = 50

extension CLHeading {
    
    // Approach 1: Initializer
    convenience init(_ heading: CLLocationDirection) {
        self.init()
        mbTrueHeading = heading
    }
    
    // Approach 2: A getter and setter
    var trueHeading: CLLocationDirection {
        get {
            return mbTrueHeading
        }
        set {
            mbTrueHeading = newValue
        }
    }
}
```

/cc @bsudekum

vincethecoder: @bsudekum ☝️ either a `// Approach 1: Initializer` or `// Approach 2: A getter and setter` above should suffice.

MapboxCoreNavigationTests/CLHeadingPrivate.h
```diff
@@ -0,0 +1,6 @@
+#import <Foundation/Foundation.h>
+#import "CoreLocation/CoreLocation.h"
+
+@interface CLHeading (Private)
+- (id)initWithHeading:(double)arg1 accuracy:(double)arg2;
```
vincethecoder: Hey @bsudekum where's your accompanying implementation?

vincethecoder:
@bsudekum I thought the main purpose of your custom heading class was to set your `trueHeading`, right? 🤔 how would this [initializer](https://github.com/mapbox/mapbox-navigation-ios/pull/1257/files#diff-1f6c5caff373184013796091f270a085R5)  address it?

- (id)initWithHeading:(double)arg1 accuracy:(double)arg2;
https://github.com/mapbox/mapbox-navigation-ios/pull/1257/files#diff-1f6c5caff373184013796091f270a085R5

bsudekum:
@vincethecoder the main goal is to just create a single CLHeading object with a known heading: https://github.com/mapbox/mapbox-navigation-ios/pull/1257/files#diff-361ec3493dc00649eefa7a072d3311d3R109. Do we even need an implementation here? Again, this is just for overriding and creating an object in tests.

vincethecoder:
@bsudekum Yes, you need an implementation file. 

Sidenote: you will be better off with an extension, if this is just for your test suite.

```
fileprivate var mbTrueHeading: CLLocationDirection = 50
fileprivate var mbAccuracy: CLLocationAccuracy = 10

extension CLHeading {
    
    // Approach 1: Initializer
    convenience init(heading: CLLocationDirection, accuracy: CLLocationAccuracy) {
        self.init()
        mbTrueHeading = heading
        mbAccuracy = accuracy
    }
    
    // Approach 2: A getter and setter
    var trueHeading: CLLocationDirection {
        get {
            return mbTrueHeading
        }
        set {
            mbTrueHeading = newValue
        }
    }
}
```

vincethecoder:
@bsudekum Attention: ⚠️ https://stackoverflow.com/a/32904444 ⚠️

PR #1258, opened 2018-03-26
Remove progress bar from bottom banner

frederoni:
This change removes the progress bar from the bottom banner as it's suboptimal for trips with reroutes.
The `ProgressBar` is also being used as a countdown indicator in the feedback form.

@bsudekum 👀


vincethecoder:
@frederoni I believe the `LaneTests` need to be updated as well, right?

bsudekum:
@vincethecoder can you link to the impacted code?

frederoni:
> I believe the LaneTests need to be updated as well, right?

@vincethecoder The lanes are attached to the top banner (InstructionsBanner) and the progress bar were attached to the bottom banner so the LanesTests shouldn't be impacted. The bottom banner doesn't have any snapshot tests yet because of its very simple layout.

PR #1260, opened 2018-03-26
Increase RouteSnappingMaxManipulatedCourseAngle threshold

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/939

I did some tests this weekend and logged each time the puck becomes unsnapped while on a route. Basically, on any 90 degree turn, the puck always becomes unsnapped momentarily. To counteract this, I increased `RouteSnappingMaxManipulatedCourseAngle` to 36 in https://github.com/mapbox/mapbox-navigation-ios/commit/5b84e08a1547adb2e886ab74c19ea668a7dd565e. This was a value I found that worked well from many tests I did.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/Constants.swift
+++ b/MapboxCoreNavigation/Constants.swift
@@ -96,7 +96,7 @@ public var RouteControllerUserLocationSnappingDistance: CLLocationDistance = 15
 /**
  Maximum angle the user puck will be rotated when snapping the user's course to the route line.
  */
-public var RouteSnappingMaxManipulatedCourseAngle: CLLocationDirection = 25
+public var RouteSnappingMaxManipulatedCourseAngle: CLLocationDirection = 36
 
 /**
  Minimum Accuracy (maximum deviation, in meters) that the route snapping engine will accept before it stops snapping.


bsudekum:
Noting, this should be extensively tested on device.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -286,6 +286,17 @@ open class RouteLegProgress: NSObject {
         currentStepProgress = RouteStepProgress(step: leg.steps[stepIndex], spokenInstructionIndex: spokenInstructionIndex)
     }
     
+    /**
+     Returns an array of `CLLocationCoordinate2D` of the current and upcoming step geometry.
+     */
+    @objc public var currentAndUpcomingCoordinates: [CLLocationCoordinate2D] {
+        let currentCoords = currentStep.coordinates ?? []
+        let upcomingCoords = upComingStep?.coordinates ?? []
+        let nearby = currentCoords + upcomingCoords
```
vincethecoder: @bsudekum maybe `nearby` should be named `nearbyCoords`

vincethecoder:
@bsudekum we should sync up regarding changes made to this function scope:
https://github.com/mapbox/mapbox-navigation-ios/pull/1260/files#diff-1a2149692b3d7c58b6a0392120d6011aR613

This would impact #1218 - meaning additional changes need to be made, which requires more field tests to fine tune the current implementation for simulating tunnel navigation.

bsudekum:
Update from field testing: I didn't like it. Two things:

* Only focusing on the current and upcoming step made it feel robotic. Strange snapping occurs in some cases when completing a maneuver.
* https://github.com/mapbox/mapbox-navigation-ios/pull/1260/commits/5b84e08a1547adb2e886ab74c19ea668a7dd565e Is a good change and can probably go in as is.

PR #1263, opened 2018-03-27
Prevent two voice announcements on some final steps

bsudekum:
I noticed on [this drive](https://gist.github.com/bsudekum/0346edd592dfdd305137726cc43acc7e), we're speaking the final instruction twice after turn left onto divis. This uncovered a few issues.

1. `userSnapToStepDistanceFromManeuver` is updated when `rawLocation` is set. However, after setting `userSnapToStepDistanceFromManeuver`, the leg or step index my increment. If this is the case, we need to update `userSnapToStepDistanceFromManeuver` otherwise, it will still be calculating the distance based on the previous step.
1. https://github.com/mapbox/mapbox-navigation-ios/pull/1155 changed voice announcements to always occur when beginning a step. This assumes the first voice announcement is always the distance of the step. This is incorrect as seen in here:

```json
"distance": 61.2,
"voiceInstructions": [{
  "distanceAlongGeometry": 9.2,
  "announcement": "You have arrived at your destination",
  "ssmlAnnouncement": "<speak><amazon:effect name="drc"><prosody rate="1.08">You have arrived at your destination</prosody></amazon:effect></speak>"
}]
```

Now, we're checking if the user is on the first step and it's the first voice announcement before automatically giving the first voice announcement. Also, any time we increment the leg or step index, we are also updating `userSnapToStepDistanceFromManeuver` so it's always accurate.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -376,12 +376,16 @@ open class RouteController: NSObject {
      */
     var rawLocation: CLLocation? {
         didSet {
-            guard let coordinates = routeProgress.currentLegProgress.currentStep.coordinates, let coordinate = rawLocation?.coordinate else {
-                userSnapToStepDistanceFromManeuver = nil
-                return
-            }
-            userSnapToStepDistanceFromManeuver = Polyline(coordinates).distance(from: coordinate)
+            updateDistanceToManeuver()
         }
+    }
+    
+    func updateDistanceToManeuver() {
+        guard let coordinates = routeProgress.currentLegProgress.currentStep.coordinates, let coordinate = rawLocation?.coordinate else {
+            userSnapToStepDistanceFromManeuver = nil
+            return
+        }
+        userSnapToStepDistanceFromManeuver = Polyline(coordinates).distance(from: coordinate)
     }
 
     @objc public var reroutingTolerance: CLLocationDistance {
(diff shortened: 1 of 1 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -852,8 +857,11 @@ extension RouteController: CLLocationManagerDelegate {
         guard let userSnapToStepDistanceFromManeuver = userSnapToStepDistanceFromManeuver else { return }
         guard let spokenInstructions = routeProgress.currentLegProgress.currentStepProgress.remainingSpokenInstructions else { return }
         
+        // Always give the first voice announcement when beginning a leg.
+        let firstInstructionOnFirstStep = routeProgress.currentLegProgress.stepIndex == 0 && routeProgress.currentLegProgress.currentStepProgress.spokenInstructionIndex == 0
```
vincethecoder: This worries me... 🤔 When we start hard-coding `index` in general, it means there's a bigger issue somewhere.

bsudekum: Yep, https://github.com/mapbox/mapbox-navigation-ios/issues/901

vincethecoder: @bsudekum can we use the `distanceTraveled` on each step to determine when we are at the beginning of a step? For instance, if we are the beginning of a step, the distanceTraveled should be `zero` right?

vincethecoder: @bsudekum can we use the `distanceTraveled` on each step to determine when we are at the beginning of a step? For instance, if we are the beginning of a step, the `distanceTraveled` should be `zero` right?

bsudekum: `distanceTraveled` is never guaranteed to be zero. In fact, it's only ever zero when you are behind the first coordinate.

PR #1266, opened 2018-03-29
Prevent automatic snapping to road network

bsudekum:
The location manager has the ability to snap to the road network when it deems it necessary. This presents a lot of problems like this drive I had below:

image
> I took a right onto van ness here

From reading through some links in https://stackoverflow.com/a/46849708/1522419, the only real way we can disable this snapping feature is by setting the activity type to `.otherNavigation `.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/NavigationLocationManager.swift
+++ b/MapboxCoreNavigation/NavigationLocationManager.swift
@@ -44,9 +44,6 @@ open class NavigationLocationManager: CLLocationManager {
         // Other navigation prevents Apple from snapping to the road network
         activityType = .otherNavigation
         
-        // Necessary when `.otherNavigation` is enabled.
-        pausesLocationUpdatesAutomatically = false
-        
         if #available(iOS 9.0, *) {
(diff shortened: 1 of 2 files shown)


bsudekum:
According the [docs](https://developer.apple.com/documentation/corelocation/clactivitytype/1423555-othernavigation), this activity type can be paused in the background. Because of this, I added `pausesLocationUpdatesAutomatically = false` in https://github.com/mapbox/mapbox-navigation-ios/pull/1266/commits/831942f6a7e0ff142bee2497bcb3655eb76760fd

bsudekum:
Rough failure in the core navigation tests:

```
failed: caught "NSInternalInconsistencyException", "Non-UI clients cannont be autopaused"
```

bsudekum:
I'm really unsure about this change after testing it this past weekend. In one sense, Apple snapping the locations to the road network helps us in many cases.

Here is a case this weekend where the issue still persists:

image
> I was on the red line, not on the blue line

bsudekum:
Going to close this out and cut a ticket with next actions.

PR #1270, opened 2018-03-29
Draw slight left/right lanes

### LINKED ISSUE #1267: Show slight arrows on freeway lanes
Deceleration lanes at exits and forks on freeways are currently marked by ordinary 90° left or right turn arrows, which can look strange next to a much gentler maneuver arrow:

lanes

Normally, the convention in OpenStreetMap is to tag freeway exits and forks with `slight_left` or `slight_right` turn lanes, but some localities signpost 90° turn lane signs at exits, too. Besides, we don’t even show a slight right turn icon when there’s a `slight_right` turn lane at an ordinary intersection, because [navigation-ui-resources](https://github.com/mapbox/navigation-ui-resources/) has identical artwork for `arrow-30` and `arrow-45`. A slight turn arrow could be reused inside the exit tab for #1262.

/cc @bsudekum @JThramer @frederoni @danesfeder

frederoni:
Fixes #1267 show slight arrows on freeway lanes

Should we also verify the road class? 🤔 

@1ec5 @bsudekum 👀


MapboxNavigationTests/LaneTests.swift
```diff
@@ -1,6 +1,6 @@
 import XCTest
 import FBSnapshotTestCase
-import MapboxDirections
+@testable import MapboxDirections
```
bsudekum: Getting the error: `Module 'MapboxDirections' was not compiled for testing`

Does something need to change in the cartfile to make this happen?

1ec5: Is this `@testable` needed? This PR doesn’t seem to introduce any calls to private methods.

bsudekum: Yes, this init is private: https://github.com/mapbox/mapbox-navigation-ios/pull/1270/files#diff-ec73732e527cb329ec167966c37b7cbeR50

frederoni: Not sure what the best practice is here but testable imports publicizes private methods in frameworks built with its configuration set to debug (carthage update/bootstrap `--configuration Debug`)

PR #1271, opened 2018-03-30
Make compatible with swift 4.1

bsudekum:
This gentle upgrades the project to xcode 9.3 and swift 4.1. In swift 4.1, `flatMap` was renamed to `companctMap`. To allow previous versions of swift to build, I added an extension that uses the appropriate function.

~~This also updates the project settings presented to me in xcode 9.3. Unsure about a few of them.~~

/cc @mapbox/navigation-ios


MapboxCoreNavigation/sequence.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+extension Sequence where Element: Hashable {
+    func compatibleFlatMap<ElementOfResult>(_ transform: (Element) throws -> ElementOfResult?) rethrows -> [ElementOfResult] {
```
akitchen: I'm not crazy about the name here -- is there another way we can define `compactMap` to delegate to `flatMap` where `compactMap` doesn't exist?

MapboxCoreNavigation/sequence.swift
```diff
@@ -0,0 +1,11 @@
...
+        #if swift(>=4.1)
```
akitchen: I'm also not crazy about using static version checks -- as an old ObjC'er, I prefer to check for API availability. Just my 2¢, but if we address my other comment we will be effectively doing that, so...

frederoni:
I was under the impression that Xcode 9.3 had a backward incompatible project structure 🤔 If that's the case, I think we should postpone this migration.

bsudekum:
@frederoni is right, from the [release notes](https://developer.apple.com/library/content/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html):

> Projects created in Xcode 9.3 use a new project format that is incompatible with earlier versions of Xcode. To open projects in earlier versions, change the project format by selecting the project in the Project navigator, opening the Document inspector, and selecting the desired format from the Project Format pop-up menu. (35207662)

We should hold on any xcode proj updates and just focus on swift 4.1 build time warnings.

bsudekum:
Removed xcode proj 9.3 changes.

bsudekum:
@akitchen I think I made this as least intrusive as possible in https://github.com/mapbox/mapbox-navigation-ios/pull/1271/commits/612a7e027d7defbbc9a3750d60b270c57fbbbbb5. Going forward, we'd always use compactMap, and it'd port backwards nicely in swift 4.0 without a new API. Let me know what you think.

MapboxCoreNavigation/sequence.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+extension Sequence where Element: Hashable {
+    #if !swift(>=4.1)
```
bsudekum: Strangely, `#if swift(<4.1)` does not work.

JThramer: Yeah, it's silly, but 🤷‍♂️

akitchen: Should this file name have a capital S ?

akitchen:
This looks good to me. We can iterate from here if needed.

Issue #1262, opened 2018-03-27
Add Exit signs to the Instruction Banner


JThramer:
When a user is traveling down a motorway, any "Exit" instructions should be styled to be very distinguishable. Currently, our exit instructions look like this:

![Exit instruction](https://user-images.githubusercontent.com/1231218/37171287-368b7e1e-22c2-11e8-9faa-f955a5d70321.png)

Ideally, they would look something more like this: (please excuse the horrible mockup)

![mockup-exit-shield](https://user-images.githubusercontent.com/887225/37987899-b712e96a-31bc-11e8-923f-27c9fd011b1b.png)

where the 'black box' would be represented by a prominent view that clearly communicates to the user that they need to take the specified exit. Ideally this view would be customizable by the client developer, as they would be able to localize the exit instruction, but integrating a simple proof-of-concept into the current instructions banner layout is a good first step.

/cc @mapbox/navigation-ios

JThramer:
Posterity: Just got off a phone/pair with @frederoni, we re-explored the `NSTextAttachment` approach and made some progress -- apparently I was missing a couple of constraints. (Oops!) We plan to pair again on Thursday 3/29.

I'll be cutting a PR for this issue as soon as we settle on one approach over another.

JThramer:
Posterity -- Current progress: 
 
![img_c8c6778b3eea-1](https://user-images.githubusercontent.com/887225/38102940-158dae54-3342-11e8-96f2-b7c8feec7b9f.jpeg)

JThramer:
Posterity: after slack talks with @bsudekum, it sounds like we basically want to implement a simple `[51 ↗] ` looking instruction.

Here is the image asset:
![Arrow](https://upload.wikimedia.org/wikipedia/commons/thumb/2/2a/Android_Emoji_2197.svg/512px-Android_Emoji_2197.svg.png)

PR #1272, opened 2018-03-30
Adding "Exit View" to Instructions Banner

JThramer:
Implements #1262. Still a work in progress, but it's getting there.

Current screenshot:
img_e4f0aeeadae8-1

To-Do:

- [x] Add tests
- [x] Use asset defined in #1262 
- [x] Get exit view styled just the way we want it
- [x] Add logic to only show when appropriate (exit instruction)
- [x] ~Implement asset in PaintCode~ (this was decided against, but asset was added to repo)
- [x] Add `InstructionPresenterDataSource` and add default `UILabel` conformance
- [x] Appearance Proxy
- [x] Implement Bobby's suggestion of basing the exit code label off of the parent's label size
- [x] Implement Fred's request for a base class
- [x] Cleanup and Polish
- [x]  Add timeouts to semaphores
- [x]  Remaining PR Comments

/cc @mapbox/navigation-ios


JThramer:
Since Github is terrible about proxying any image you link to it, the asset is contained here: https://commons.wikimedia.org/wiki/File:Android_Emoji_2197.svg

willwhite:
> Get exit view styled just the way we want it

@brsbl can we pull in some design help to assist with styling?

JThramer:
Current Screenshot: 
img_fcc8448351ca-1

bsudekum:
@JThramer looking great. Couple pieces of feedback:

* The size of the 123A and the arrow should be flipped. I.E. the arrow should be smaller and the text slightly larger.
* Let's go with just an outline around the exit info. The fill can be white.
* How are you thinking about vertical alignment?

JThramer:
@bsudekum 
>The size of the 123A and the arrow should be flipped. I.E. the arrow should be smaller and the text slightly larger.

Easily done. The arrow is constrained to "fill as much space as possible while still keeping aspect ratio", so if I increase the size of the text the arrow should react accordingly.
>Let's go with just an outline around the exit info. The fill can be white.

Not a problem.

>How are you thinking about vertical alignment?

We've got code for a `ShieldAttachment` (`NSTextAttachment` subtype) class that aligns the image to the midpoint of the string. I'll be reusing that.

JThramer:
Current Appearance:
image 1

JThramer:
Posterity: Talked with Design, @Cjballard has signed off on the current design of the exit view, with one change: A reduction in the image height from 50% of view height to 40%. 

/cc @brsbl @willwhite

MapboxNavigation/ExitView.swift
```diff
@@ -0,0 +1,127 @@
+//
+//  ExitView.swift
+//  MapboxNavigation
+//
+//  Created by Jerrad Thramer on 3/23/18.
+//  Copyright © 2018 Mapbox. All rights reserved.
+//
```
bsudekum: nit, this can be removed.

MapboxNavigation/ExitView.swift
```diff
@@ -0,0 +1,127 @@
...
+        view.contentMode = .scaleAspectFit
+        return view
+    }()
+    
+    lazy var exitNumberLabel: UILabel = {
+        let label: UILabel = .forAutoLayout()
+        label.text = exitText
+        label.textColor = .black
+        print("system font size! \(UIFont.systemFontSize)")
+        label.font = UIFont.boldSystemFont(ofSize: 20)
```
bsudekum: Curious, is the rest of the instruction banner hardcoded to a specific size value? We should probably remain consistent here so it doesn't look lopsided if the user has default small/large text.

MapboxNavigation/ExitView.swift
```diff
@@ -0,0 +1,127 @@
...
+        return label
+    }()
+
+    var exitText: String? {
+        didSet {
+            exitNumberLabel.text = exitText
+            invalidateIntrinsicContentSize()
+        }
+    }
+    var pointSize: CGFloat? {
```
bsudekum: Can this be non-optional since it's set during init?

MapboxNavigation/ExitView.swift
JThramer: Yeah, I'll need to find a way to base this off the label's font.

MapboxNavigation/UIView.swift
```diff
@@ -136,6 +136,15 @@ extension UIView {
         }
         return trailingAnchor
     }
+    
+    var imageRepresentation: UIImage {
```
bsudekum: Looks like we should make this optionally return UIImage.

MapboxNavigationTests/InstructionPresenterTests.swift
```diff
@@ -0,0 +1,46 @@
...
+class InstructionPresenterTests: XCTestCase {
+    
+    static let fixtureData = fixtureString.data(using: .utf8)
+    static let fixtureJSON: [String: Any] = try! JSONSerialization.jsonObject(with: InstructionPresenterTests.fixtureData!, options: []) as! [String: Any]
+    static let subject = VisualInstruction(json: InstructionPresenterTests.fixtureJSON, drivingSide: .right)
+
+    func testExitRecognized() {
+        let subject = InstructionPresenterTests.subject
+        XCTAssertTrue(subject.primaryTextComponents.isExit, "Not recognizing exit instruction components as an exit")
+    }
```
akitchen: This is good. Until now this class was implicitly tested via integration & snapshot tests, but as we potentially use it in more places let's look for opportunities to consolidate certain tests in this test class.

MapboxNavigation/ExitView.swift
frederoni: It could default to `PrimaryLabel.appearance().normalFont`

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -140,6 +156,23 @@ class InstructionPresenter {
         attachment.image = shieldImage
         return NSAttributedString(attachment: attachment)
     }
+    
+    private func exitShield(side: ExitSide = .right, text: String) -> NSAttributedString {
+        let exit = ExitView(pointSize: label!.font.pointSize, side: side, text: text)
+        exit.translatesAutoresizingMaskIntoConstraints = false
+        exit.invalidateIntrinsicContentSize()
```
frederoni: None of these invalidation and update calls should be necessary.

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -140,6 +156,23 @@ class InstructionPresenter {
...
+        exit.setNeedsLayout()
+        exit.layoutIfNeeded()
+        let exitAttachment = NSTextAttachment()
+        let exitImage = exit.imageRepresentation
+        exitAttachment.image = exitImage
+        if let label = self.label {
+            let yOrigin = (label.font.capHeight - exitImage.size.height).rounded() / 2
+            exitAttachment.bounds = CGRect(x: 0, y: yOrigin, width: exitImage.size.width, height: exitImage.size.height)
```
frederoni: Any reason we can't reuse shield attachment bounds?

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -140,6 +156,23 @@ class InstructionPresenter {
...
+    private func exitShield(side: ExitSide = .right, text: String) -> NSAttributedString {
+        let exit = ExitView(pointSize: label!.font.pointSize, side: side, text: text)
+        exit.translatesAutoresizingMaskIntoConstraints = false
+        exit.invalidateIntrinsicContentSize()
+        exit.setNeedsLayout()
+        exit.layoutIfNeeded()
+        let exitAttachment = NSTextAttachment()
+        let exitImage = exit.imageRepresentation
+        exitAttachment.image = exitImage
+        if let label = self.label {
```
frederoni: Guard against a nil label at the top of this method.

MapboxNavigation/InstructionPresenter.swift
JThramer: Yep, that's the plan. This is interstitial.

MapboxNavigationTests/InstructionPresenterTests.swift
```diff
@@ -0,0 +1,46 @@
+import Foundation
+import XCTest
+@testable import MapboxNavigation
+import MapboxDirections
```
frederoni: You can do testable import MapboxDirections in order to use the private strongly typed init methods instead of json. (As long as you use `--configuration Debug` when you do `carthage bootstrap/update`.

JThramer:
Current Screenshots: Things are looking pretty good, although I did get an instruction to `Take exit 188.` Will investigate that tomorrow.
img_a8aae27a78c2-1
img_486ab180104d-1

JThramer:
Posterity: It definitely comes from [the directions service](https://api-directions-traf-next-internal.tilestream.net/directions/v5/mapbox/driving-traffic/-105.002381055929,39.7527235296041;-104.998301652637,39.2286961004276?overview=false&steps=true&access_token=pk.eyJ1IjoiYm9iYnlzdWQiLCJhIjoiTi16MElIUSJ9.Clrqck--7WmHeqqvtFdYig&banner_instructions=true) as `Take exit 173` as a text `bannerInstruction`. I'll let this go for the purposes of this PR, but we may want to improve on just showing text that says `Take exit NNN`

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -23,7 +25,7 @@ class InstructionPresenter {
     
     func fittedAttributedComponents() -> [NSAttributedString] {
         guard let label = self.label else { return [] }
-        var attributedPairs = self.attributedPairs()
+        var attributedPairs = InstructionPresenter.attributedPairs(for: instruction, on: label, imageRepository: imageRepository, onImageDownload: completeShieldDownload)
```
akitchen: Let's work on this together tomorrow. I think there might be another way of introducing the seam we need, depending on the tests we need to write.

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -170,6 +231,13 @@ fileprivate struct IndexedVisualInstructionComponent {
 }
 
 extension Array where Element == VisualInstructionComponent {
```
akitchen: Should an array of `VisualInstructionComponents` be given a type name or become a class/struct? Extending `Array` like this makes me think maybe, yes.

JThramer: @akitchen You would think so, and I thought on this too, but aside from adding a `typealias VisualInstructionComponents = [VisualInstructionComponent]` (which is not great, I think "Just pluralize the noun" is kind of an anti-pattern over `[TypeName]`) I'm not sure what we could do here. Instruction Components come from properties in `VisualInstruction`, so perhaps the best place to add extra sugar would be there.

Additionally, I would like to point out that this extension ended up not getting used and thus will be deleted.

akitchen: Works for me! thanks

MapboxNavigation/ExitView.swift
```diff
@@ -0,0 +1,120 @@
...
+    
+    func rebuildConstraints() {
+        NSLayoutConstraint.deactivate(self.constraints)
+        buildConstraints()
+    }
+    
+    func commonInit() {
+        backgroundColor = .clear
+        layer.borderWidth = 1.0
+        layer.borderColor = UIColor.black.cgColor
```
frederoni: This should be an appearance property so that we can style in differently in the `NightStyle`.

JThramer: 🤦‍♂️ I knew I was forgetting something! Will do.

MapboxNavigation/ExitView.swift
```diff
@@ -12,6 +12,26 @@ class ExitView: UIView {
...
+            exitNumberLabel.textColor = foregroundColor
+            setNeedsDisplay()
+        }
+    }
+    @objc dynamic var borderWidth: CGFloat {
+        didSet {
+            layer.borderWidth = borderWidth
+        }
+    }
+    @objc dynamic var cornerRadius: CGFloat {
```
frederoni: Can we consolidate `cornerRadius` and `borderWidth` in a `StylableView` similar to `StylableButton`?
`foregroundColor` fits perfectly and can stay in `ExitView`.

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -1,16 +1,27 @@
 import UIKit
 import MapboxDirections
 
+protocol InstructionPresenterDataSource: class {
+    var availableBounds: (() -> CGRect)! { get }
```
akitchen: Still no solution to eliminate this closure? I guess we can flag it as tail work... now that we have this protocol it feels like a leaky abstraction (and maybe it isn't needed?). Just dropping a note here for posterity; I don't think it should block merging this PR since all we've done here is introduce a protocol to better clarify presenter's data dependencies.

akitchen:
@JThramer what's up with the snapshot tests failing on CI? Do reference images need to be updated or is there something else going on?

Also, could you please verify that we're not accidentally hitting the network in your new tests? the presenter is still using the `ImageRepository.shared` instance internally, which is problematic in the general sense...

akitchen:
>  problematic in the general sense...

Just to be clear, this doesn't need to be fixed before merging, so long as we're not hitting the network in the new tests.

MapboxNavigation/Style.swift
```diff
@@ -258,6 +258,21 @@ open class StylableLabel: UILabel {
     }
 }
 
+/// :nodoc
+@objc(MBStyleableView)
+open class StylableView: UIView {
```
frederoni: nit: s/Stylable/Styleable or the other way around? Both variations seems to be wildly used but we should be consistent. If we settle for styleable, would you mind renaming the existing stylable classes?

MapboxNavigationTests/InstructionPresenterTests.swift
```diff
@@ -0,0 +1,23 @@
...
+
+class InstructionPresenterTests: XCTestCase {
+    
+    func testExitInstructionProvidesExit() {
+                let exitAttribute = VisualInstructionComponent(type: .exit, text: "Exit", imageURL: nil, maneuverType: .takeOffRamp, maneuverDirection: .right, abbreviation: nil, abbreviationPriority: 0)
+                let exitCodeAttribute = VisualInstructionComponent(type: .exitCode, text: "123A", imageURL: nil, maneuverType: .takeOffRamp, maneuverDirection: .right, abbreviation: nil, abbreviationPriority: 0)
+                let label = InstructionLabel(frame: CGRect(origin: .zero, size:CGSize(width: 50, height: 50)))
+        
+                //FIXME: not ideal -- UIAutoLayout?
+                label.availableBounds = { return label.frame }
```
frederoni: Unfortunately, we can't use autolayout here as we need to know the future available bounds. Also, the label’s frame could be based on a previous instruction that was half the size of the available bounds in the banner causing abbrevations to trigger when there is plenty of space left. However, since this is a test case and the size is fixed, we shouldn't have to worry about that. Consider using `CGSize.iPhoneX.width` or something similar to https://github.com/mapbox/mapbox-navigation-ios/blob/7fd21bb748b3a3fce3228d7e6d7c7c1fcbfef539/MapboxNavigation/InstructionsBannerViewLayout.swift#L138-L143 for a more realistic test case.

MapboxNavigation/Style.swift
1ec5: “[Stylable](https://en.wiktionary.org/wiki/stylable#English)” is the preferred spelling.

Cartfile.resolved
```diff
@@ -5,4 +5,4 @@ github "mapbox/mapbox-events-ios" "v0.4.0"
 github "mapbox/mapbox-voice-swift" "v0.0.1"
 github "mapbox/turf-swift" "v0.1.1"
 github "raphaelmor/Polyline" "v4.2.0"
-github "uber/ios-snapshot-test-case" "3.0.0"
+github "uber/ios-snapshot-test-case" "1.1-1as24"
```
frederoni: looks like a typo or merge conflict sneaked in here. We should use 3.0.0

MapboxNavigation/DataCache.swift
```diff
@@ -23,15 +23,7 @@ public class DataCache: NSObject, BimodalDataCache {
...
-        if toDisk == true {
-            fileCache.store(data, forKey: key, completion: completion)
-        } else {
-            if let completion = completion {
-                DispatchQueue.main.async {
-                    completion()
-                }
-            }
-        }
+        toDisk == true ? fileCache.store(data, forKey: key, completion: completion) : completion?()
```
frederoni: This was previously dispatched to the main queue. Are we sure we're on the main queue when we set the image?

JThramer: Yes. See [here](https://github.com/mapbox/mapbox-navigation-ios/pull/1272/files#diff-621dcac99b9b9fa73ac6931ca5a68ccdR22).

(PS: It's interesting to note that we don't actually _set_ the image through the callback. Although the callback does pass back an image, we actually implicitly rely on the image cache, because if you follow the callback chain, we essentially simply [call](https://github.com/mapbox/mapbox-navigation-ios/pull/1272/files#diff-af48d70309ee195e34cc5e6cc6b27caaR203) `attributedText()` again.)

akitchen: Jerrad and I paired on this -- the design decision that felt the most right was to require the caller to control the queue hopping, not burying it inside the lower-level objects. So a UI class needs to be explicit about making assignments on the main queue inside the completion handler, such as we're doing [here](https://github.com/mapbox/mapbox-navigation-ios/pull/1272/files#diff-621dcac99b9b9fa73ac6931ca5a68ccdR22)

We should make sure to call this out in the API documentation.

MapboxNavigation/ImageCache.swift
```diff
@@ -20,18 +20,9 @@ internal class ImageCache: BimodalImageCache {
...
-            }
-        } else {
-            if let completion = completion {
-                DispatchQueue.main.async {
-                    completion()
-                }
-            }
-        }
+        
+        guard toDisk == true, let data = UIImagePNGRepresentation(image) else { return _ = completion?() }
```
akitchen: This reads a little weird for my taste.

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -97,28 +117,52 @@ class InstructionPresenter {
...
+    
+    private func shieldImageForComponent(_ component: VisualInstructionComponent, in repository: ImageRepository, height: CGFloat, completion: ImageDownloadCompletion?) {
         guard let imageURL = component.imageURL, let shieldKey = component.shieldKey() else {
             return
         }
 
-        imageRepository.imageWithURL(imageURL, cacheKey: shieldKey, completion: { (image) in
-            completion(image)
-        })
+        repository.imageWithURL(imageURL, cacheKey: shieldKey, completion: completion! )
```
akitchen: Is this `!` guaranteed to be safe here?

MapboxNavigationTests/Support/ImageDownloadOperationSpy.swift
```diff
@@ -60,5 +60,6 @@ class ImageDownloadOperationSpy: Operation, ImageDownload {
         completionBlocks.forEach { completion in
             completion(image, data, error)
         }
+        RunLoop.current.run(until: Date().addingTimeInterval(5))
```
akitchen: Oof. this is punitive. can we reduce this, with the goal of eliminating it? I thought one tick of the runloop was enough here. was that not the case on your CPU-bound machine?

MapboxNavigation/DayStyle.swift
```diff
@@ -83,6 +83,12 @@ open class DayStyle: Style {
         EndOfRouteStaticLabel.appearance().normalTextColor = #colorLiteral(red: 0.217173934, green: 0.3645851612, blue: 0.489295125, alpha: 1)
         EndOfRouteTitleLabel.appearance().normalFont = .systemFont(ofSize: 36.0)
         EndOfRouteTitleLabel.appearance().normalTextColor = .black
+
```
frederoni: I try to sort these entries w/o blank lines in between. If you don't have a script in Xcode you can always just copy the relevant block and do `pbpaste | sort | pbcopy` in the terminal and paste it back into Xcode.

MapboxNavigation/ExitView.swift
```diff
@@ -0,0 +1,130 @@
...
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.contentMode = .scaleAspectFit
+        return view
+    }()
+    
+    lazy var exitNumberLabel: UILabel = {
+        let label: UILabel = .forAutoLayout()
+        label.text = exitText
+        label.textColor = .black
+        print("system font size! \(UIFont.systemFontSize)")
```
frederoni: Drop this debugging info

MapboxNavigation/InstructionPresenter.swift
frederoni: Curious about an alternative solution. 🤔

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -140,22 +184,58 @@ class InstructionPresenter {
...
+    }
+    
+    private func completeShieldDownload(_ image: UIImage?) {
+        onShieldDownload?(attributedText()) //FIXME: Can we work with the image directly?
+    }
+    
+    private func takeSnapshot(on view: UIView) -> UIImage?{
+        let window = UIApplication.shared.delegate!.window!!
+        
+        window.addSubview(view)
```
frederoni: Add a note describing why we're abusing UIAppearance here.

MapboxNavigationTests/Support/ImageDownloadOperationSpy.swift
akitchen: I pushed an update to adjust this, since things passed for me while under a sustained load average of ~80+

akitchen:
I think we should merge to master from https://github.com/mapbox/mapbox-navigation-ios/pull/1272/commits/a074f0602ec770256f889a7973c8c5424ee6e042 -- this commit seems to introduce a new issue in that none of the tests in `InstructionsBannerViewIntegrationTests` can be run in isolation.

JThramer:
Posterity: Rewound the PR to a074f06 for merging. All changes made past that point are now tracked in #1316.

MapboxNavigationTests/InstructionsBannerViewIntegrationTests.swift
```diff
@@ -89,33 +90,83 @@ class InstructionsBannerViewIntegrationTests: XCTestCase {
...
+        let roadNameRange = NSMakeRange(2, 11)
+        let roadName = attributed.attributedSubstring(from: roadNameRange)
+        XCTAssert(roadName.string == "Main Street", "Banner not populating road name correctly")
+    }
 
     private func simulateDownloadingShieldForComponent(_ component: VisualInstructionComponent) {
         let operation: ImageDownloadOperationSpy = ImageDownloadOperationSpy.operationForURL(component.imageURL!)!
         operation.fireAllCompletions(ShieldImage.i280.image, data: UIImagePNGRepresentation(ShieldImage.i280.image), error: nil)
+        //FIXME: need to get to a place where this isn't necessary, probably by adjusting our test code.
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 1))
```
akitchen: This is OK for the moment but I would like to eliminate the need for this shortly after we merge. This is mostly a note to self.

akitchen:
Might need to give this another once-over after master is re-integrated...

PR #1283, opened 2018-04-04
Update turf dependency

bsudekum:
v0.1.0 came out some time ago, we should probably update to this version.

I also loosened the semver check here and allowed patch versions to come through as needed.

/cc @mapbox/navigation-ios


MapboxNavigation.podspec
```diff
@@ -46,7 +46,6 @@ Pod::Spec.new do |s|
   s.dependency "MapboxCoreNavigation", "#{s.version.to_s}"
   s.dependency "Mapbox-iOS-SDK", "~> 3.6"
   s.dependency "Solar", "~> 2.1"
-  s.dependency "Turf", "~> 0.0.4"
```
bsudekum: Unsure why we needed this here if it's already included in core.

PR #1284, opened 2018-04-05
Don't fallback to default speech synthesizer if prepareToPlay fails

bsudekum:
We should not be falling back to the default speech synthesizer if the audioPlayer does not [prepareToPlay](https://developer.apple.com/documentation/avfoundation/avaudioplayer/1386886-preparetoplay) properly.

Even if `prepareToPlay` fails, the audio can still play as `prepareToPlay` is just there to preload audio.

I was able to hit this code repeatedly by removing alls checks and playing an audio instruction on every update. When rerouting, we would interrupt the current audio, stopping the audioPlayer and rendering it "unprepared".

/cc @mapbox/navigation-ios


bsudekum:
One might argue that prepareToPlay really is not necessary here. I think it's more useful for when the audio will be played soon, but you are waiting on some cue to begin.

PR #1289, opened 2018-04-10
Allow users to select alternate routes while navigating

bsudekum:
_This is an exploration PR, no further actions from the team needed right now_

Adds a popup to the UI that allows users to select alternate routes after rerouting if they are unhappy with their current route.

todo: 
- [ ] add a less intrusive popup view
- [ ] handle touching popup
- [ ] use accurate ETA diff
- [ ] make this more performant
- [ ] better placement when zoomed out

> simulator screen shot - iphone x - 2018-04-09 at 15 13 14
> When zoomed out, we should either hide the popup as a first pass or place it at the midpoint of difference between the two routes. Unsure about that maths right now.

simulator screen shot - iphone x - 2018-04-09 at 15 14 35

/cc @mapbox/navigation-ios


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -11,6 +11,11 @@ open class RouteProgress: NSObject {
      Returns the current `Route`.
      */
     @objc public let route: Route
+    
+    /**
+     An optional alternate route the user could use for navigation.
+     */
+    @objc public var alternateRoute: Route?
```
bsudekum: This is about the least intrusive change I could think of. Making `RouteProgress.route` -> `RouteProgress.routes` would be a pretty huge change.

1ec5: I think we should make that change for consistency with other parts of the API. `route` can still be a computed property serving as an alias for the first route.

bsudekum: Yeah the only issue is that init changes to `routes` which becomes a huge headache. Do we then rely on order of `routes` for the active route?

1ec5: Yes, `NavigationMapView` already does that:

https://github.com/mapbox/mapbox-navigation-ios/blob/b57c8e78355211d59e06ce67714298012b129994/MapboxNavigation/NavigationMapView.swift#L89
https://github.com/mapbox/mapbox-navigation-ios/blob/b57c8e78355211d59e06ce67714298012b129994/MapboxNavigation/NavigationMapView.swift#L423

1ec5:
Remember to ensure that the alternative route line has sufficient contrast against the new Navigation Guidance Day and Navigation Guidance Night map styles.

frederoni:
Looks good so far. What happens if `reroutesProactively` is set to true? Do we use Levenshteins distance to stay on the alternate route?

akitchen:
I suspect this one won't update its build status until another commit is pushed 🤷‍♂️

Issue #1200, opened 2018-03-05
Decouple Telemetry/Feedback from the RouteController.


JThramer:
_Context: This ticket originated as an [issue comment](https://github.com/mapbox/mapbox-navigation-ios/issues/1071#issuecomment-360568842) on #1071. This issue was opened to track movement on that idea after its' parent issue was closed._

Currently, our [telemetry/feedback engine](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/MMEEventsManager.swift) is [very tightly coupled](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/RouteController.swift?utf8=✓#L982) to the [RouteController](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/RouteController.swift).

Ideally, the `RouteController` would interact with the feedback/telemetry engine via A terse, opaque, "fire-and-forget" interface. Something that looks like this: 
```swift
Telemetry.shared.report(reroute: newRouteProgress, old: routeProgress)
```
or 
```swift
Telemetry.shared.report(userFeedback: .confusingInstruction, on: routeProgress, description: nil)
```
Past that point, the `RouteController` cares not about _how_ the report is delivered, with message queueing and the like. That's the telemetry engine's job. It should be fire-and-forget.

That way, the `FeedbackViewController` could also talk directly to the telemetry singleton without leaving the developer the task of manual message passing.

Currently, there is quite a lot of back and forth between the `RouteController` and the telemetry logic, but really the only three things that the feedback engine actually cares about in the route controller is the `routeProgress`, the current location and whether it's simulated, and the `sessionState`, which I think the Telemetry engine should own.

/cc @mapbox/navigation-ios

akitchen:
I like where this is headed -- 'fire and forget'. I would also like us to do more to encapsulate the steps required to build up an event contextually.

The `RouteController` extension you linked to probably deserves to become its own class. There is a lot of navigation-specific event information which needs to live somewhere other than the core mobile telemetry implementation; it just so happens to be partially consolidated in that extension but is also spread around the `RouteController` class itself. As the navigation use cases have matured so should our code...

[...]

PR #1292, opened 2018-04-12
Backfill events tests & decouple events from RouteController

akitchen:
This is still WIP in terms of resolving #1200 , however the plan is to:

 - [ ] Backfill tests for key scenarios that generate telemetry events (& NSNotifications & delegate messages)
   - [x] turnstile event upon initialization
   - [x] successful reroute (scenario 1)
   - [ ] successful reroute (scenario 2)
   - [x] arrival event generation
   - [ ] feedback event data collection workflow
 - [ ] Perform a few key refactorings 
   - [ ] remove excess events knowledge and bookkeeping from RouteController
   - [ ] introduce a RouteContollerEventDelegate to manage event generation and pass delegate messages

I would like to do this in a way which can be merged to master early and often in the spirit of True Continuous Integration™ and in order to avoid semantic disconnects with other work streams.

I'll be tidying up / squashing down the commit log a bit, but the tests we have so far are high value and can/should be merged to master more or less immediately with a 👍 from the team.

/cc @vincethecoder who helped bootstrap this work

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -528,7 +528,7 @@ extension RouteController: CLLocationManagerDelegate {
     }
 
     @objc public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
-        
+        print("locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation])")
         let filteredLocations = locations.filter {
             sessionState.pastLocations.push($0)
             return $0.isQualified
@@ -951,22 +951,27 @@ struct SessionState {
 extension RouteController {
     // MARK: Sending events
     func sendDepartEvent() {
+        print("\(#function)")
         eventsManager.enqueueEvent(withName: MMEEventTypeNavigationDepart, attributes: eventsManager.navigationDepartEvent(routeController: self))
         eventsManager.flush()
     }
 
     func sendArriveEvent() {
+        print("\(#function)")
         eventsManager.enqueueEvent(withName: MMEEventTypeNavigationArrive, attributes: eventsManager.navigationArriveEvent(routeController: self))
         eventsManager.flush()
     }
 
     open func sendCancelEvent(rating: Int? = nil, comment: String? = nil) {
+        print("\(#function)")
         let attributes = eventsManager.navigationCancelEvent(routeController: self, rating: rating, comment: comment)
         eventsManager.enqueueEvent(withName: MMEEventTypeNavigationCancel, attributes: attributes)
         eventsManager.flush()
     }
 
     func sendFeedbackEvent(event: CoreFeedbackEvent) {
+        print("\(#function)")
+
         // remove from outstanding event queue
         if let index = outstandingFeedbackEvents.index(of: event) {
             outstandingFeedbackEvents.remove(at: index)
@@ -982,6 +987,7 @@ extension RouteController {
     // MARK: Enqueue feedback
 
     func enqueueFeedbackEvent(type: FeedbackType, description: String?) -> String {
+        print("\(#function)")
         let eventDictionary = eventsManager.navigationFeedbackEvent(routeController: self, type: type, description: description)
         let event = FeedbackEvent(timestamp: Date(), eventDictionary: eventDictionary)
 
@@ -991,6 +997,7 @@ extension RouteController {
     }
 
     func enqueueRerouteEvent() -> String {
+        print("\(#function)")
         let timestamp = Date()
         let eventDictionary = eventsManager.navigationRerouteEvent(routeController: self)
 
@@ -999,12 +1006,14 @@ extension RouteController {
 
         let event = RerouteEvent(timestamp: Date(), eventDictionary: eventDictionary)
 
+        print("Appending event internally")
         outstandingFeedbackEvents.append(event)
 
         return event.id.uuidString
     }
     
     func enqueueFoundFasterRouteEvent() -> String {
+        print("\(#function)")
         let timestamp = Date()
         let eventDictionary = eventsManager.navigationRerouteEvent(routeController: self, eventType: FasterRouteFoundEvent)
         
@@ -1018,6 +1027,7 @@ extension RouteController {
     }
 
     func checkAndSendOutstandingFeedbackEvents(forceAll: Bool) {
+        print("\(#function)")
         let now = Date()
         let eventsToPush = forceAll ? outstandingFeedbackEvents : outstandingFeedbackEvents.filter {
             now.timeIntervalSince($0.timestamp) > SecondsBeforeCollectionAfterFeedbackEvent
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -528,7 +528,6 @@ extension RouteController: CLLocationManagerDelegate {
     }
 
     @objc public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
-        print("locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation])")
         let filteredLocations = locations.filter {
             sessionState.pastLocations.push($0)
             return $0.isQualified
@@ -999,7 +998,6 @@ extension RouteController {
 
         let event = RerouteEvent(timestamp: Date(), eventDictionary: eventDictionary)
 
-        print("Appending event internally")
         outstandingFeedbackEvents.append(event)
 
         return event.id.uuidString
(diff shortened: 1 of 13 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -998,26 +1020,29 @@ extension RouteController {
...
 
     func checkAndSendOutstandingFeedbackEvents(forceAll: Bool) {
-        let now = Date()
-        let eventsToPush = forceAll ? outstandingFeedbackEvents : outstandingFeedbackEvents.filter {
-            now.timeIntervalSince($0.timestamp) > SecondsBeforeCollectionAfterFeedbackEvent
-        }
-        for event in eventsToPush {
+//        let now = Date()
+
+//        let eventsToPush = forceAll ? outstandingFeedbackEvents : outstandingFeedbackEvents.filter {
```
akitchen: This is the one thing that needs to be reconciled before this set of changes can merge -- do we really need to do this event delay filtering? If so, we need to adjust the testing strategy a bit to take it into account, otherwise this code needs to be cleaned up.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -604,111 +607,125 @@ extension RouteController: CLLocationManagerDelegate {
...
         let durationRemaining = routeProgress.durationRemaining
 
         getDirections(from: location) { [weak self] (route, error) in
-            guard let strongSelf = self else { return }
-            guard let route = route else { return }
+            guard let strongSelf = self else {
+                return
+            }
+
+            guard let route = route else {
```
JThramer: These could be combined into one guard.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -767,6 +784,15 @@ extension RouteController: CLLocationManagerDelegate {
         #endif
     }
 
+    private func checkForLocationUsageDescription() {
+        guard let _ = Bundle.main.bundleIdentifier else {
```
JThramer: `guard Bundle.main.bundleIdentifier != nil`

MapboxCoreNavigation/RouteController.swift
```diff
@@ -767,6 +784,15 @@ extension RouteController: CLLocationManagerDelegate {
...
+            return
+        }
+        if Bundle.main.locationAlwaysUsageDescription == nil && Bundle.main.locationWhenInUseUsageDescription == nil && Bundle.main.locationAlwaysAndWhenInUseUsageDescription == nil {
```
JThramer: Nit: This is fine, but I find it more natural-reading to use commas as opposed to boolean-and's in swift conditionals when the conditional is a simple "if all this stuff is T/F". 

`if thisThing, thatThing, otherThing {` 

vs 

`if thisThing && thatThing && otherThing {`

MapboxCoreNavigation/RouteController.swift
```diff
@@ -998,31 +1026,41 @@ extension RouteController {
...
+
         sessionState.lastRerouteDate = timestamp
-        
+
         let event = RerouteEvent(timestamp: Date(), eventDictionary: eventDictionary)
-        
+
         outstandingFeedbackEvents.append(event)
-        
+//        eventsManager.enqueueEvent(withName: MMEEventTypeNavigationReroute, attributes: event.eventDictionary)
```
JThramer: Comment.

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -1,116 +1,246 @@
...
+        static let jsonRoute = (response["routes"] as! [AnyObject]).first as! [String: Any]
+        static let accessToken = "nonsense"
     }
-    
+
+    let eventsManagerSpy = EventsManagerSpy()
+    let directionsClientSpy = DirectionsSpy(accessToken: "garbage", host: nil)
+    let delegate = RouteControllerDelegateSpy()
+
+    lazy var dependencies: (routeController: RouteController, firstLocation: CLLocation, lastLocation: CLLocation) = {
```
JThramer: This is very slick. I might have to start doing this.

MapboxCoreNavigation/RouteController.swift
vincethecoder: @akitchen will we be getting rid of this, or was this commented out for a reason?

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -1,116 +1,246 @@
...
+            let fromLocation = notification.userInfo![RouteControllerNotificationUserInfoKey.locationKey] as? CLLocation
+            return fromLocation == testLocation
+        }
+
+        let didRerouteNotificationExpectation = expectation(forNotification: .routeControllerDidReroute, object: routeController, handler: nil)
+
+        let routeProgressDidChangeNotificationExpectation = expectation(forNotification: .routeControllerProgressDidChange, object: routeController) { (notification) -> Bool in
+            let location = notification.userInfo![RouteControllerNotificationUserInfoKey.locationKey] as? CLLocation
+            let rawLocation = notification.userInfo![RouteControllerNotificationUserInfoKey.rawLocationKey] as? CLLocation
+            let _ = notification.userInfo![RouteControllerNotificationUserInfoKey.routeProgressKey] as! RouteProgress
```
JThramer: Why are we pulling this out if we're just throwing it away?

MapboxNavigationTests/InstructionsBannerViewIntegrationTests.swift
```diff
@@ -93,33 +93,28 @@ class InstructionsBannerViewIntegrationTests: XCTestCase {
...
 
-    //FIXME: this test is artificially slow as we are polling the run loop following each simulated download. need a better way of synchronizing on the underlying completions.
     func testDelimiterDisappearsOnlyWhenAllShieldsHaveLoaded() {
         let view = instructionsView()
         
-        //setup expectations
         let firstExpectation = XCTestExpectation(description: "First Component Callback")
         let secondExpectation = XCTestExpectation(description: "Second Component Callback")
-        
-        view.primaryLabel.imageRepository = imageRepository
```
JThramer: Hey, what's going on here? Without these two lines, `InstructionLabel` will be hitting the actual `.shared` imageRepository. Am I missing something?

MapboxCoreNavigation/RouteController.swift
vincethecoder: @JThramer Although, it's an elegant approach, being explicit can also help distinguish where the value could be a boolean value or optional. An explicit definition in this conditional statement helps specify our main goal here. Yes?

JThramer: @vincethecoder I'm not sure I understand what you're getting at, the `locationAlwaysUsageDescription == nil` is required in either case, as `locationAlwaysUsageDescription` is not a `Bool`.

`,` is literally just conditional syntactic sugar for `&&`. It works in any conditional where `&&` does. IMHO, boolean and's are appropriate when you have some sort of complex conditional logic, e.g: `if thisThing && thatThing || thatOtherThing` or `(thisThing && thatThing) || orThisThing`, but if you are just checking "all these clauses in this expression are true", then it's more natural-reading to use commas to join the clauses of this expression.

MapboxNavigationTests/InstructionsBannerViewIntegrationTests.swift
akitchen: Yeah, this test is still using the shared instance configured to use the `DownloadOperationSpy` class. It seems to be working well with the label's `.imageDownloadCompletion` pattern as a synchronization point. Let's see how it does. We can continue to refactor as needed since this is just a checkpoint.

akitchen: It's also worth noting that this assignment was assigning the same instance back onto the labels.

MapboxCoreNavigation/RouteController.swift
akitchen: 1) this code was just moved from init
2) this code arguably doesn't need/deserve to exist at all, but we're minimizing changes at this stage in the game

MapboxCoreNavigation/RouteController.swift
akitchen: yep, this is a comment. 

It's likely where the event should be enqueued from, once things are less entangled.

MapboxCoreNavigationTests/RouteControllerTests.swift
akitchen: It would be more explicit to nil check it, agreed.

MapboxCoreNavigationTests/RouteControllerTests.swift
akitchen: This pattern existed here previously but wasn't memoized on a var like this. Something to watch out for!

MapboxCoreNavigation/RouteController.swift
vincethecoder: @JThramer I was inferring when there are instances where a field could be `Bool?`, then we should be explicit. Or there is a series of `Bool` and `Optional<Nil>` values in one conditional statement. 

It's just self-documenting. Yes, we can also short-key click to inspect the variable. In this thread, I just said it is alright to be explicit. It might be additional LOC but it adds clarity to the code -- that's all.

PR #1294, opened 2018-04-12
Give distance label a bit more space

### LINKED ISSUE #1237: Distance label should accommodate more text without truncation
#1235 attempted to fix #1234 by reducing the minimum font size for autoscaling. Unfortunately, that isn’t enough to avoid truncation – or even left truncation – in some languages that have long abbreviations for customary units. Of particular concern are Swedish and Ukrainian. (Ukrainian doesn’t have a localization yet, but it is supported by the Directions API.)

Language | Metric | Customary
----|:--:|:--:
Hebrew | failed_testlongdistance 2x failed_testlongdistance 2x | failed_testlongdistance 2x failed_testlongdistance 2x
Russian | failed_testlongdistance 2x failed_testlongdistance 2x | failed_testlongdistance 2x failed_testlongdistance 2x failed_testlongdistance 2x
Swedish | failed_testlongdistance 2x failed_testlongdistance 2x | failed_testlongdistance 2x failed_testlongdistance 2x failed_testlongdistance 2x failed_testlongdistance 2x
Ukrainian | failed_testlongdistance 2x failed_testlongdistance 2x | failed_testlongdistance 2x failed_testlongdistance 2x failed_testlongdistance 2x failed_testlongdistance 2x
Vietnamese | failed_testlongdistance 2x failed_testlongdistance 2x | failed_testlongdistance 2x failed_testlongdistance 2x failed_testlongdistance 2x

I mentioned a possible fix in https://github.com/mapbox/mapbox-navigation-ios/issues/1234#issue-306117655:

> Some supported languages have even longer abbreviations for “meter” or “kilometer” than “公尺”. One possible solution would be to dynamically increase the horizontal space allotted to the maneuver icon and distance label based on the distance label’s intrinsic width. (The more fundamental issue is that we’re cramming such an important, dynamic label into such a tight spot, as if it doesn’t really matter.)

/cc @frederoni @JThramer

frederoni:
Fixes #1237 round 2

Gives the distance label a bit more space and allows for 20 / 30 font scaling if needed.

@Cjballard @mapbox/navigation-ios


akitchen:
LGTM, though also noticing the units text is different in the two images. Was that intentional?

frederoni:
> also noticing the units text is different in the two images.

Yep, that was intentional. Imperial region settings with Swedish language settings (rarely used) and metric region settings with Ukrainian language settings are the combinations that takes up the most space.

PR #1296, opened 2018-04-13
Add Sequence.swift to MapboxNavigation podspec

### LINKED ISSUE #1295: MapboxNavigation depends on private compactMap(_:) method
I keep running into this error when building;

'compactMap' is inaccessible due to 'internal' protection level

I am using mapbox navigation 0.16.1 via cocoapods on latest xcode and swift.

1ec5:
Include Sequence.swift in MapboxNavigation, fixing a compilation error when building a version of MapboxNavigation that is assembled by CocoaPods instead of being built from the checked-in Xcode project.

Fixes #1295.

/cc @frederoni @JThramer @bsudekum

### DIFF
--- a/MapboxNavigation.podspec
+++ b/MapboxNavigation.podspec
@@ -32,7 +32,7 @@ Pod::Spec.new do |s|
 
   # ――― Source Code ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
-  s.source_files = ["MapboxNavigation/*", "MapboxCoreNavigation/{Date,Geometry,Sequence,String}.swift"]
+  s.source_files = ["MapboxNavigation/*", "MapboxCoreNavigation/{Date,Sequence,String}.swift"]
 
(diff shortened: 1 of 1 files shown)


frederoni:
The reason why CI completed successfully despite running the CocoaPods integration test was that piping `xcodebuild` through `xcpretty` caused the command to return exit code 0 instead of 65. Using `set -o pipefail` returns the correct result through the pipe.

MapboxNavigation.podspec
```diff
@@ -32,7 +32,7 @@ Pod::Spec.new do |s|
 
   # ――― Source Code ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
-  s.source_files = ["MapboxNavigation/*", "MapboxCoreNavigation/{Date,Geometry,String}.swift"]
+  s.source_files = ["MapboxNavigation/*", "MapboxCoreNavigation/{Date,Geometry,Sequence,String}.swift"]
```
frederoni: should be able to drop "Geometry"

PR #1299, opened 2018-04-13
v0.16.2

1ec5:
Updated the Xcode project and CocoaPods podspec to v0.16.2. Also updated the documentation generation script to link to the MapboxDirections.swift v0.19.1 documentation.

/cc @mapbox/navigation-ios


scripts/document.sh
```diff
@@ -29,8 +29,8 @@ THEME=${JAZZY_THEME:-$DEFAULT_THEME}
 BASE_URL="https://www.mapbox.com/mapbox-navigation-ios"
 
 # Link to directions documentation
-DIRECTIONS_VERSION="0.17.0"
-DIRECTIONS_SYMBOLS="Directions|Intersection|Lane|Route|RouteLeg|RouteOptions|RouteStep|SpokenInstruction|VisualInstruction|VisualInstructionComponent|Waypoint"
+DIRECTIONS_VERSION="0.19.1"
+DIRECTIONS_SYMBOLS="Directions|DirectionsOptions|DirectionsResult|Intersection|Lane|Match|MatchOptions|Route|RouteLeg|RouteOptions|RouteStep|SpokenInstruction|Tracepoint|VisualInstruction|VisualInstructionComponent|Waypoint"
```
akitchen: This is interesting to see. I don't recall seeing mention of the need to update this script in https://github.com/mapbox/mapbox-navigation-ios/blob/master/docs/README.md or https://github.com/mapbox/mapbox-navigation-ios/blob/master/RELEASE.md

Does this happen automatically or is it a manual step we need to capture somewhere?

1ec5: Upgrading the MapboxDirections.swift isn’t necessarily part of releasing the navigation SDK, so it doesn’t very naturally go into RELEASE.md.

Updating `DIRECTIONS_VERSION` is the more important part, keeping developers from sinking into an outdated version hole. Updating `DIRECTIONS_SYMBOLS` is rarer but ensures that the autolinking functionality we’ve hacked together works as broadly as possible.

PR #1310, opened 2018-04-18
Rewrite discarding delegate method documentation

1ec5:
The previous documentation had a typo. Also, “qualified” is not a term of art when it comes to describing a location update, so the new documentation instead describes the act of considering or ignoring the location update.

/cc @vincethecoder @bsudekum


1ec5:
The test failures are due to #1312.

1ec5:
Now failing due to #1314.

PR #1315, opened 2018-04-19
Exit View - Checkpoint One

JThramer:
_All of the greatness of #1272, None of the problems™_

Gotta go fast
/cc @mapbox/navigation-ios


JThramer:
It sounds like this is PR is the inverse of what we wanted to do. Closing.

PR #1316, opened 2018-04-19
Exit View Test Issues

JThramer:
This PR is to track the test issues present in #1272  that will not be initially merged into `master`.

/cc @mapbox/navigation-ios


JThramer:
I'm giving up on this. The cause of the crashes here is that the newly-created ImageRepository already has images cached on it, and end so the cache calls end up hitting when they should miss. I have been unable to figure out why this is so, and since yesterday I've been unable to reliably reproduce the state I was in (group would pass, one would fail).

PR #1323, opened 2018-04-23
Expose RouteController.routeProgress to Objective-C

### LINKED ISSUE #1320: How can I navigate to a coordination location after click on button.
Hi all,

I doing mapbox navigation in my IOS apps (objective-c). How can I navigate to a location after click on button in ViewController A without long press location on map and click on "Navigation" button.

Please help. Thank you.

frederoni:
Fixes #1320 Make it possible to update route from Objective-C

@mapbox/navigation-ios 👀

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -185,7 +185,7 @@ open class RouteController: NSObject {
     /**
      Details about the user’s progress along the current route, leg, and step.
      */
-    public var routeProgress: RouteProgress {
+    @objc public var routeProgress: RouteProgress {
         willSet {
             // Save any progress completed up until now
             sessionState.totalDistanceCompleted += routeProgress.distanceTraveled
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master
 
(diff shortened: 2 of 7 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -185,7 +185,7 @@ open class RouteController: NSObject {
     /**
      Details about the user’s progress along the current route, leg, and step.
      */
-    public var routeProgress: RouteProgress {
+    @objc public var routeProgress: RouteProgress {
```
1ec5: Besides exposing the property to Objective-C, this change also switches the property’s getter and setter to use Objective-C method dispatch. The `willSet` and `didSet` should continue to work just like before, but it’s worth double-checking to be sure.

frederoni: https://github.com/mapbox/mapbox-navigation-ios/pull/1323/commits/4f21e5a63878a7029c4a125e6890c3ce9c009fa9 verifies that the notification in the property’s `didSet` gets fired.

tester87:
Hi Frederoni, sorry to ask stupid question. Which mean that now I need to update my CocoaPods 
 Podfile and try to apply sample code as per #1320?

frederoni:
@tester87 correct, that fix is included in v0.17.0.

Issue #590, opened 2017-09-07
Way name label should use current route step


ericrwolfe:
The way name label should use the information in the current route step from the direction's API's route response, rather than querying for rendered features in the map.

This will allow us to be consistent with our turn banner and remove the "ceremonial" highway names from the way name label (replacing them with the proper ref codes). If the user goes off route, we shouldn't be showing the way name label at all.

cc @bsudekum @frederoni

bsudekum:
@1ec5 I know you had your reasons against this, mind reminding me?

I think as long as the user is snapped to the route line, it should be fine to show the current step name.

1ec5:
There’s no guarantee that the way name of the current step applies to the entire step, only that it applies to the first intersection on the step:

* OSRM removes any trace of name changes like “North Main Street” to “South Main Street” that would be too insignificant for a voice instruction but still significant enough for wayfinding via the map. (Knowing you’ve gone all the way to the south quadrant of the city could be useful information.)
* OSRM suppresses temporary name changes (Project-OSRM/osrm-backend#2744).
* OSRM suppresses “obvious” turns onto differently named streets.
* The Directions API recently began suppressing all `new name` instructions.

Given all these processing steps, it’s highly likely that we’d leave a misleading road name in a prominent location on screen. Ultimately, the purpose of the current road name label is different than most of the other UI provided by the navigation SDK: it’s meant to amplify information that’s already labeled on the map, rather than supplementing the map with additional information.

> This will allow us to be consistent with our turn banner and remove the "ceremonial" highway names from the way name label (replacing them with the proper ref codes).

Removing freeway names is the easy part. Replacing them with fully qualified textual route numbers is more difficult. But we do have enough information to display a graphical route shield that’s consistent with the map.

/ref mapbox/mapbox-navigation-ios#193

ericrwolfe:
All of the above can and should be added as annotations within our directions API. cc @willwhite @miccolis @hdaymon

Using the map style to display the current way name is an elaborate hack that is buggy and prone to error. For the majority of time spent on most of my drives (read: highways), the way name either displays the wrong name or none at all.

Let's plan to move forward and switch to the current route step, as this will simplify our logic, provide consistency across platforms and languages, and cover the vast majority of way name use cases. 

The remaining issues that you've brought up can be addressed as tail work in concert with the API team.

1ec5:
> simplify our logic, provide consistency across platforms and languages, and cover the vast majority of way name use cases

[...]

PR #1338, opened 2018-04-25
Custom road name feature-level and integration tests

### LINKED ISSUE #1269: Developers can provide their own way name to populate the label below the puck
This will allow developers to easily add/display a current street label, such as:

img_402cabd8e03c-1

/cc @1ec5

vincethecoder:
Added feature-level and integration tests. (Follow up development efforts for #1309)

Closes #1269

/cc @mapbox/navigation-ios


MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -0,0 +1,192 @@
...
+        }
+        
+        wait(for: [mapviewExpectation], timeout: delay)
+    }
+}
+
+class NavigationViewControllerSpy: NavigationViewController { }
+class NavigationViewControllerFakeSpy: NavigationViewController { }
+
+extension NavigationViewControllerSpy: NavigationViewControllerDelegate {
```
akitchen: Cool, however in this case I wouldn't call this a Spy, given how we're using it. Also, it's not usually considered good form to spy on the test subject itself since we should be most concerned with how it is reacting externally to events or messages.

Could we have the test case class itself (or another test helper class) be the delegate instead?

MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -0,0 +1,192 @@
...
+            
+            mapviewExpectation.fulfill()
+        }
+        
+        wait(for: [mapviewExpectation], timeout: delay)
+    }
+}
+
+class NavigationViewControllerSpy: NavigationViewController { }
+class NavigationViewControllerFakeSpy: NavigationViewController { }
```
akitchen: Let's remove this, I don't think it's needed. also please see my comment below.

MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -0,0 +1,192 @@
...
+        // Identify a location without a custom road name.
+        let fultonStreetLocation = dependencies.poi[2]
+        
+        let mapviewExpectation = expectation(description: "mapview_exp_default_roadname")
+        let customMapView = MGLMapView(frame: (dummyNavigationViewController.mapView?.bounds)!, styleURL: URL(string: "mapbox://styles/mapbox/streets-v9"))
+        
+        let delay = 1.0
+        let fireTime = DispatchTime.now() + delay
+        
+        DispatchQueue.main.asyncAfter(deadline: fireTime) {
```
1ec5: This is an unconventional way to set up an expectation regarding a “did finish” delegate method. For one thing, it subjects the test to a hard-coded time interval to delay the test by. Is this test expected to actually load vector tiles to query? If so, should any of that process be mocked? If not, is a delay necessary?

Normally, an integration test would cause the delegate method to fire organically – in this case, by setting a style – and set an expectation that gets fulfilled inside an implementation of the delegate method. The point is to allow the `mapViewDidFinishLoadingMap(_:)` method to be called on its own; calling that method directly would only test the implementation of that method, which isn’t the focus of this test.

MapboxNavigationTests/NavigationViewControllerTests.swift
vincethecoder: Sure thing.

MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -0,0 +1,192 @@
...
+    func testNavigationViewControllerDelegateRoadNameAtLocationUmimplemented() {
+        
+        let dummyNavigationViewController = dependencies.dummyNavigationViewController
+        let routeController = dummyNavigationViewController.routeController!
+        
+        // Identify a location without a custom road name.
+        let fultonStreetLocation = dependencies.poi[2]
+        
+        let mapviewExpectation = expectation(description: "mapview_exp_default_roadname")
+        let customMapView = MGLMapView(frame: (dummyNavigationViewController.mapView?.bounds)!, styleURL: URL(string: "mapbox://styles/mapbox/streets-v9"))
```
1ec5: RouteMapViewController’s road name label functionality is powered by business logic within that class, dependent on a NavigationMapView that it sets up. Here, we’re creating a new, plain-vanilla MGLMapView, which could deviate significantly from how the NavigationMapView is normally set up inside RouteMapViewController.

MapboxNavigationTests/NavigationViewControllerTests.swift
vincethecoder: @akitchen Perhaps a test helper class is ideal because developers who initially extend the `NavigationViewController` class need to implement `NavigationViewControllerDelegate.navigationViewController(_:roadNameAt:)` in order to customize road names. Hence, this test ensures the validity of this new functionality.

1ec5:
Per chat, we’re going to avoid the rabbit hole of making these tests depend on the Maps API and feature querying. Instead, we’re going to extract the default feature querying implementation and mock it in the tests. The goal is only to ensure that the default implementation gets invoked by default.

The default implementation has significant known issues. Since addressing those issues requires nontrivial changes to the Directions API, testing that functionality rigorously would be a distraction and would only serve to confirm the observations laid out in #590.

akitchen:
Sounds like we're scoping this down to a collaboration test between the `NavigationViewController` and its delegate, which should cover the core logic of the feature. I fully agree w/ @1ec5 on keeping this one tightly focused and not worrying too much about external complexities 👍

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1156,7 +1156,7 @@ public protocol NavigationMapViewDelegate: class {
      - parameter waypoints: The waypoints to be displayed on the map.
      - returns: Optionally, a `MGLShape` that defines the shape of the waypoint, or `nil` to use default behavior.
      */
-    @objc(navigationMapView:shapeDescribingWaypoints:)
```
bsudekum: We should document these changes in the changelog.

MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -0,0 +1,145 @@
...
+extension NavigationViewControllerTests: NavigationViewControllerDelegate {
+    
+    func navigationViewController(_ navigationViewController: NavigationViewController, roadNameAt location: CLLocation) -> String? {
+        return customRoadName[location.coordinate] ?? nil
+    }
+}
+
+extension CLLocationCoordinate2D: Hashable {
+    // Hash value property multiplied by a prime constant.
+    public var hashValue: Int {
```
bsudekum: Is this used?

vincethecoder: Yep, implicitly for our key-value pairs here:
https://github.com/mapbox/mapbox-navigation-ios/pull/1338/files/ea482a1284403cfbb4e4260eee3c62b692bd7162#diff-841adf0b96176970740abfa00fe996b2R7

MapboxNavigation/NavigationMapView.swift
vincethecoder: @bsudekum Sure. In this case where the function signature didn't necessarily change, what kind of changes should we document in the changelog?

MapboxNavigationTests/NavigationViewControllerTests.swift
vincethecoder: @bsudekum Yes, it will. Property 'hashValue' must be declared public because it matches a requirement in public protocol 'Hashable'.

MapboxNavigation/NavigationMapView.swift
vincethecoder: @bsudekum 👉  https://github.com/mapbox/mapbox-navigation-ios/pull/1338/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR12

Issue #1327, opened 2018-04-23
Test failures in ImageDownloaderTests — Semaphore timed out


1ec5:
[This Bitrise build](https://www.bitrise.io/build/3b057052f7f1513a) for #1076 experienced a slew of test failures related to semaphores in the image downloader tests:

<details>
<summary>Test failures</summary>

MapboxNavigationTests.ImageDownloaderTests
  testDownloadingAnImage, XCTAssertTrue failed - Semaphore timed out
  /Users/vagrant/git/MapboxNavigationTests/ImageDownloaderTests.swift:49
  ```
        let semaphoreResult = semaphore.wait(timeout: XCTestCase.NavigationTests.timeout)
        XCTAssert(semaphoreResult == .success, "Semaphore timed out")
  ```
  testDownloadingImageAgainAfterFirstDownloadCompletes, XCTAssertTrue failed - Semaphore timed out
  /Users/vagrant/git/MapboxNavigationTests/ImageDownloaderTests.swift:95
  ```
        let semaphoreResult = semaphore.wait(timeout: XCTestCase.NavigationTests.timeout)
        XCTAssert(semaphoreResult == .success, "Semaphore timed out")
        
  ```
  testDownloadingSameImageWhileInProgressAddsCallbacksWithoutAddingAnotherRequest, failed - Timeout occurred on runUntil(condition:pollingInterval:until:)
  /Users/vagrant/git/MapboxNavigationTests/ImageDownloaderTests.swift:118
  ```
        guard (timeout >= DispatchTime.now()) else {
            XCTFail("Timeout occurred on \(#function)")
            return
  ```
MapboxNavigationTests.ImageRepositoryTests
  test_imageWithURL_downloadsImageWhenNotCached, XCTAssertTrue failed - Semaphore timed out
  /Users/vagrant/git/MapboxNavigationTests/ImageRepositoryTests.swift:27
  ```
        let semaphoreResult = semaphore.wait(timeout: XCTestCase.NavigationTests.timeout)
        XCTAssert(semaphoreResult == .success, "Semaphore timed out")
    }
  ```
  test_imageWithURL_prefersCachedImageWhenAvailable, XCTAssertTrue failed - Semaphore timed out
  /Users/vagrant/git/MapboxNavigationTests/ImageRepositoryTests.swift:27
  ```
        let semaphoreResult = semaphore.wait(timeout: XCTestCase.NavigationTests.timeout)
[...]

PR #1341, opened 2018-04-26
Adjust ImageDownloaderTests to use a new instance each time

akitchen:
Removing this statefulness (the `lazy var`) should make sense here; we're registering & unregistering the `URLProtocol` subclass in a loop while the URL loading system may be doing other work. I'm hopeful this will help with the test instabilities noted in #1327 and elsewhere.

I also noticed that the `ImageDownloader`'s `OperationQueue`'s `maxConcurrentOperationCount` was set semi-arbitrarily, so I removed it. This defaults the configuration to `NSOperationQueueDefaultMaxConcurrentOperationCount`. We could also consider setting it to `2`.

### DIFF
--- a/MapboxNavigationTests/ImageDownloaderTests.swift
+++ b/MapboxNavigationTests/ImageDownloaderTests.swift
@@ -35,12 +35,16 @@ class ImageDownloaderTests: XCTestCase {
     }
 
     func testDownloadingAnImage() {
+        guard let downloader = downloader else {
+            XCTFail()
+            return
+        }
         var imageReturned: UIImage?
         var dataReturned: Data?
         var errorReturned: Error?
         let semaphore = DispatchSemaphore(value: 0)
 
-        downloader?.downloadImage(with: imageURL) { (image, data, error) in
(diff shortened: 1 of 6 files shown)


MapboxNavigationTests/ImageDownloaderTests.swift
```diff
@@ -39,7 +40,7 @@ class ImageDownloaderTests: XCTestCase {
         var errorReturned: Error?
         let semaphore = DispatchSemaphore(value: 0)
 
-        downloader.downloadImage(with: imageURL) { (image, data, error) in
+        downloader?.downloadImage(with: imageURL) { (image, data, error) in
```
1ec5: Should this test assert that `downloader` is non-`nil`, as the other tests do below?

akitchen: I guess I was being terse since this method only touches it once. Preference was for a runtime failure (nil) vs a crash (unexpected failure unwrapping).

Happy to make it consistent with the other methods before merging.

PR #1343, opened 2018-04-26
Clarify language support table

1ec5:
Revised the “Localization and Internationalization” guide to clarify that spoken instructions are only announced in languages supported by the Directions API, regardless of which speech synthesizer supports the language.

Since the Voice API is used by default, the table no longer calls out its use explicitly; instead, the VoiceOver and third-party fallbacks are called out as exceptions. Languages that lack UI support have been added with the caveat that the developer needs to opt into them explicitly.

Finally, the table now notes that instructions are spoken in Mandarin as opposed to Cantonese or another Chinese spoken language.

Thanks to @danesfeder for digging up [this VoiceOver language support document](https://support.apple.com/en-us/HT206175).

/cc @brsbl @danesfeder @frederoni

### DIFF
--- a/docs/guides/Localization and Internationalization.md
+++ b/docs/guides/Localization and Internationalization.md
@@ -27,7 +27,7 @@ func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
 
 Turn instructions are announced in the user interface language when turn instructions are available in that language. Otherwise, if turn instructions are unavailable in that language, they are announced in English instead. To have instructions announced in a language other than the user interface language, set the `RouteOptions.locale` property when calculating the route with which to start navigation.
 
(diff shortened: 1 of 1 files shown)


docs/guides/Localization and Internationalization.md
```diff
@@ -27,39 +27,43 @@ func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
...
-| Italian    | ✅              | ✅                             | ✅                     | ✅
-| Portuguese | ✅             | ✅                             | ✅                     | ✅
-| Polish     | —              | ✅                             | ✅                     | ✅
-| Romanian   | —              | ✅                             | ✅                     | ✅
-| Russian    | ✅              | ✅                             | ✅                     | ✅
-| Spanish    | ✅              | ✅                             | ✅                     | ✅
-| Swedish    | ✅              | ✅                             | ✅                     | ✅
-| Turkish    | —              | ✅                             | ✅                     | ✅
-| Vietnamese | ✅              | ✅                             | —                     | —
+The table below lists the languages that are supported for user interface elements and for spoken instructions. If a language is marked as “manual”, the user will not automatically receive spoken instructions in that language, so you would need to set `RouteOptions.locale` in order to use it.
```
bsudekum: Maybe it would be helpful to outline the process to make:

> If a language is marked as “manual”, the user will not automatically receive spoken instructions in that language, so you would need to set `RouteOptions.locale` in order to use it.

automatic and why this is the case.

1ec5: It isn’t possible to use these particular languages automatically. The top portion of this document does discuss how the other languages are automatic and why these particular languages are manual.

1ec5:
A test is failing on Bitrise due to #1361.

PR #1345, opened 2018-04-27
Modify access level for various classes

### LINKED ISSUE #1344: Modify access control for NavigationViewController
Currently the `NavigationViewController` is [marked as public](https://github.com/mapbox/mapbox-navigation-ios/blob/ce9ae2f057c79c1aa85f9449dec189880979ac47/MapboxNavigation/NavigationViewController.swift#L196), which means that subclasses are not allowed outside of the scope of the module. Would it be possible to change the [access control](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AccessControl.html) such that this class can be `open`, so that subclassing would be possible?

cc @frederoni

frederoni:
Fixes #1344 - Allow subclassing NavigationViewController

This PR also modifies the access level in order to allow subclassing of the following classes:
- DistanceFormatter
- ~NavigationSettings~
- ReplayLocationManager
- SimulatedLocationManager
- LanesView
- ManeuverView

@captainbarbosa @1ec5 👀

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master
 
+* `DistanceFormatter`, `ReplayLocationManager`, `SimulatedLocationManager`, `LanesView`, and `ManueverView` are now subclassable. ([#1345](https://github.com/mapbox/mapbox-navigation-ios/pull/1345]))
 * Upgraded to the [Mapbox Maps SDK for iOS v4.0.0](https://github.com/mapbox/mapbox-gl-native/releases/tag/ios-v4.0.0). If you have customized the route map’s appearance, you may need to migrate your code to use expressions instead of style functions. ([#1076](https://github.com/mapbox/mapbox-navigation-ios/pull/1076))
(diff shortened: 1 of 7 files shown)


MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -15,7 +15,7 @@ extension Notification.Name {
  Properties are prefixed and before they are stored in UserDefaults.standard.
  */
 @objc(MBNavigationSettings)
-public class NavigationSettings: NSObject {
+open class NavigationSettings: NSObject {
```
1ec5: What are the implications for the singleton `NavigationSettings.shared`? Is it even possible to use an instance of a NavigationSettings subclass?

1ec5:
Also, this deserves a changelog entry.

MapboxCoreNavigation/NavigationSettings.swift
frederoni: I figured e.g overriding `voiceMuted`’s didSet could be useful for changing a custom mute icon but a developer can also listen for `.navigationSettingsDidChange` notifications or KVO `.shared.observe(./voiceMuted …`. There are some bad implications though for subclassing this singleton, duplicated notifications being the first that comes to mind so let's avoid subclassing this class by reverting back to public.

PR #1346, opened 2018-04-27
Add Korean localization

1ec5:
Added a Korean localization, courtesy of @tan-tran-ea. Updated a couple translations in Arabic and European Portuguese.

/cc @frederoni

### DIFF
--- a/docs/guides/Localization and Internationalization.md
+++ b/docs/guides/Localization and Internationalization.md
@@ -48,6 +48,7 @@ The upcoming road or ramp destination is named according to the local or nationa
 | Hebrew     | ✅              | ✅                             | —                     | ✅
 | Hungarian  | ✅              | —                             | —                     | ✅
 | Italian    | ✅              | ✅                             | ✅                     | ✅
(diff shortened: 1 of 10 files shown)


1ec5:
I’ve repurposed this PR for adding a Korean localization from Transifex. Thanks for the translations, @tan-tran-ea!

docs/guides/Localization and Internationalization.md
```diff
@@ -48,6 +48,7 @@ The upcoming road or ramp destination is named according to the local or nationa
 | Hebrew     | ✅              | ✅                             | —                     | ✅
 | Hungarian  | ✅              | —                             | —                     | ✅
 | Italian    | ✅              | ✅                             | ✅                     | ✅
+| Korean    | ✅              | —                             | ✅                     | ✅
```
1ec5: The sample applications currently won’t show anything in Korean, but any application that is localized into Korean will have the fully localized navigation UI.

Spoken instructions are marked as unavailable for now, until we release and deploy a version of the Directions API that incorporates Project-OSRM/osrm-text-instructions#243.

Issue #1355, opened 2018-04-30
Rerouting status view remains visible if request fails


bsudekum:
If the user is rerouted when offline or the network request fails, `Rerouting...` remains in the status view. We should remove this status up if the network requests fails for any reason.

Also, if our "is the user on a future step?" check succeeds, we should remove the status update.

PR #1357, opened 2018-04-30
Dismiss reroute status view if request fails

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1355

This prevents the reroute status view update from being visible indefinitely.

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,6 @@
 * `RouteController`’s `routeProgress` is now exposed to Objective-C. [#1323](https://github.com/mapbox/mapbox-navigation-ios/pull/1323)
 * Exit indications are now drawn accurately with a correct exit bearing. [#1288](https://github.com/mapbox/mapbox-navigation-ios/pull/1288)
 * Added a delegate method, `NavigationViewControllerDelegate.navigationViewController(_:roadNameAt:)` which allows you to customize the contents of the road name label displayed towards the bottom of the map view. [#1309](https://github.com/mapbox/mapbox-navigation-ios/pull/1309)
-
-### User interface
-
 * If a reroute request fails, the `Rerouting...` status view update is dismissed. [#1357](https://github.com/mapbox/mapbox-navigation-ios/pull/1357)
 
 ## v0.16.2 (April 13, 2018)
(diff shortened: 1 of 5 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -315,6 +316,10 @@ class RouteMapViewController: UIViewController {
         statusView.show(title, showSpinner: true)
     }
     
+    @objc func rerouteDidFail(notification: NSNotification) {
+        statusView.hide()
```
bsudekum: Ideally, we'd only hide a status view of type reroute. Currently, any and all status updates will be hidden.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 * Exit indications are now drawn accurately with a correct exit bearing. [#1288](https://github.com/mapbox/mapbox-navigation-ios/pull/1288)
 * Added a delegate method, `NavigationViewControllerDelegate.navigationViewController(_:roadNameAt:)` which allows you to customize the contents of the road name label displayed towards the bottom of the map view. [#1309](https://github.com/mapbox/mapbox-navigation-ios/pull/1309)
 
+### User interface
```
1ec5: There are other UI-related changes above. Before or after merging, let’s recategorize everything, since the “master” section has gotten a bit long.

bsudekum:
@1ec5 updated with tests.

MapboxNavigation/RouteMapViewController.swift
1ec5: Still planning to address this detail in this PR? Do we have a way of asking the status view what kind of message it’s displaying?

bsudekum: I think this would require a larger refactor.

1ec5: Let’s either take care of that refactor here or ticket it out. This need has come up multiple times in the past. I think it would be quite unexpected that a request failure would blow away a “Simulating Navigation” banner, for instance.

1ec5: Let’s either take care of that refactor here or ticket it out. This need has come up multiple times in the past. I think it would be quite unexpected that a request failure would blow away a “Simulating Navigation” banner, for instance.

Issue #1361, opened 2018-04-30
ImageDownloaderTests reliability improvements


1ec5:
`testDownloadingImageAgainAfterFirstDownloadCompletes` occasionally fails on Bitrise with an error reminiscent of #1327. Only Xcode 9.3 builds fail this way. For example, [this build](https://www.bitrise.io/build/d0d98774deb30aeb) for #1343 failed:

<details>
<summary>testDownloadingImageAgainAfterFirstDownloadCompletes, XCTAssertTrue failed - Semaphore timed out</summary>

MapboxNavigationTests.ImageDownloaderTests
  testDownloadingImageAgainAfterFirstDownloadCompletes, XCTAssertTrue failed - Semaphore timed out
  /Users/vagrant/git/MapboxNavigationTests/ImageDownloaderTests.swift:120
  ```
        let secondSemaphoreResult = semaphore.wait(timeout: XCTestCase.NavigationTests.timeout)
        XCTAssert(secondSemaphoreResult == .success, "Semaphore timed out")
        
  ```
	 Executed 47 tests, with 1 failure (0 unexpected) in 12.690 (12.786) seconds
</details>

/cc @akitchen

akitchen:
@1ec5 Did this test run include the code changes in https://github.com/mapbox/mapbox-navigation-ios/commit/cb951125de2decd29173c2e719ea19a945f86046 ?

If this is indeed occurring with the URLProtocol registration changes in place, then we need to put a bit more instrumentation in place. It's also likely the case that we're not properly handling a URL loading sad path case (which can be reproduced by running the test without the URLProtocol registered).

1ec5:
Yes, that branch is based on 31848f35775889740af5822fb0c07eb5ff945dec and it has failed a few times now.

akitchen:
oof, that's frustrating. I have a few ideas as to what could be going on here, looking at other URLProtocol implementations.

Meanwhile I'll open a PR to disable the test.

akitchen:
The symptom described here was fixed in #1405

PR #1363, opened 2018-05-01
Temporarily disable flaky test

akitchen:
re: #1361


bsudekum:
Just a fun fact, you can also disable via right clicking and disabling. This will add entry to `<SkippedTests>`.

image

Issue #1306, opened 2018-04-17
Multiple waypoint bug


srikanthsrnvs:
Im using the latest version of mapbox with xcode. I have a list of waypoints and instatiate a navigationViewController with those waypoints. Two issues arise from this;

1) When a waypoint has been visited, it dissapears from the map, and the next waypoint to be visited shows a little circe with a 1, even though this is the second waypoint. Similarly, after visiting the second waypoint, it dissapears and the next waypoint shows on the map as a 1, when it is actually the third waypoint.

2) The voice instructions tell me that im about to visit my first destination even though it might be the 2nd, third or fourth. It only works for the first, and last.

3) Is the tiny circle with the waypoint number an annotation? if so, how can I change its appearence based on my waypoint.name? is there a way to access waypoint through viewForAnnotation?

bsudekum:
Hi @srikanthsrnvs, I'm looking into this issue now.

srikanthsrnvs:
I think I’ve found the reason. Everything works fine when on route, but if you press the button to show all the route waypoints, and then switch back to resume, it screws things up.

srikanthsrnvs:
So the issue is this;


When on multiwaypoint route,

If I press the button to show the entire route after visiting the first waypoint, and then press resume, when it arrives at the next waypoint, it thinks I have arrived at my first waypoint, when it is actually the second. Similarly, for third, fourth etc.

Sometimes the destination annotation also disappears.

PR #1364, opened 2018-05-01
Keep completed waypoints on map

bsudekum:
_Addresses a few issues in https://github.com/mapbox/mapbox-navigation-ios/issues/1306_

Previously, when arriving at a waypoint, we would redraw the waypoints on the map and remove the completed the waypoint. This caused the upcoming waypoint to always be 1. 

Now, completed waypoints are kept on the map but made more translucent.

simulator screen shot - iphone x - 2018-05-01 at 11 15 08
simulator screen shot - iphone x - 2018-05-01 at 11 21 17

> Completed waypoint 1

/cc @mapbox/navigation-ios @srikanthsrnvs


srikanthsrnvs:
@bsudekum its sproadic and random, sometimes it happens, and sometimes it doesnt. But weve had numerous complaints from users that the destination annotation sometimes dissapears. I will try to attach screen shots etc if I manage to reproudce it. Can you help me with another thing I have outlined in the original issue?

PR #1365, opened 2018-05-02
Adjusting flaky test continuation criteria

akitchen:
Happy to let this sit open for a while and evaluate how it does on Xcode 9.3. I'm also planning on refactoring the URLProtocol implementation detail I think might not be correct.


akitchen:
Next step here is to defer the beginning of the actual URL loading and do it on the client's queue, if different. We could also consider using OHHTTPStubs, although it's a heavier abstraction. Something to think/talk about.

PR #1366, opened 2018-05-02
Update style color palette for v3

### LINKED ISSUE #1347: Upgrade to v4 navigation map styles
The v2 Navigation Preview/Guidance Day/Night styles should be replaced by the v3 Navigation Preview/Guidance Day/Night styles once they’re ready.

/cc @mapbox/navigation-ios @aparlato

frederoni:
Partially closes #1347 

This PR updates the color palette for v3 guidance day/night styles.
v2 guidance map styles will still be used by default but convenient methods for any version has been added in the form of (`MGLStyle.navigationGuidance{Day/Night}StyleURL(version:)`.

@1ec5 @bsudekum 👀

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -72,6 +72,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     let instructionCircle = "instructionCircle"
     let alternateSourceIdentifier = "alternateSource"
     let alternateLayerIdentifier = "alternateLayer"
+    let alternateCasingLayerIdentifier = "alternateCasingLayer"
     
     @objc dynamic public var trafficUnknownColor: UIColor = .trafficUnknown
     @objc dynamic public var trafficLowColor: UIColor = .trafficLow
@@ -80,6 +81,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     @objc dynamic public var trafficSevereColor: UIColor = .trafficSevere
     @objc dynamic public var routeCasingColor: UIColor = .defaultRouteCasing
     @objc dynamic public var routeAlternateColor: UIColor = .defaultAlternateLine
+    @objc dynamic public var routeAlternateCasingColor: UIColor = .defaultAlternateLineCasing
     @objc dynamic public var maneuverArrowColor: UIColor = .defaultManeuverArrow
     @objc dynamic public var maneuverArrowStrokeColor: UIColor = .defaultManeuverArrowStroke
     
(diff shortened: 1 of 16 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -887,18 +899,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-        
-        // Take the default line width and make it wider for the casing
-        lineCasing.lineWidth = NSExpression(format: "mgl_interpolate:withCurveType:parameters:stops:($zoomLevel, 'linear', nil, %@)", MBRouteLineWidthByZoomLevel.multiplied(by: 0.85))
-        lineCasing.lineColor = NSExpression(forConstantValue: routeAlternateColor)
-        lineCasing.lineCap = NSExpression(forConstantValue: "round")
-        lineCasing.lineJoin = NSExpression(forConstantValue: "round")
-        lineCasing.lineOpacity = NSExpression(forConstantValue: 0.9)
-        
-        return lineCasing
+    func alternateRouteStyleLayer(identifier: String, lineColor: UIColor, lineWidth: NSExpression, source: MGLSource) -> MGLStyleLayer {
```
bsudekum: What are we gaining by passing through line color/width?

bsudekum: Oh I see.

frederoni: Yea, the alternate route didn't have a casing before so this method is now used for creating the alternate route style layer and the casing layer.

bsudekum:
image

I'm slightly concerned with the contrast here. Now since the strongest color here is white, we do not get such an easy division between ui elements and the map style.

/cc @aparlato @Cjballard @1ec5 @brsbl

1ec5:
The second screenshot demonstrates that we should dim the rest of the screen when the end-of-route feedback view controller comes into view. But I think that would need to be a separate PR focused on the end-of-route experience.

/cc @JThramer

bsudekum:
I'm going to remove this issue from the v0.17.0 milestone as there seems to be some kinks to work out with the style/ui still.

Cjballard:
Exploring a few options for increasing contrast here. We could style the top banner to be Mapbox gray-blue (#273D56) to provide some weight and color at the top of the view. 

banner-darker

We could style the sub-banners as either slightly darker or slightly lighter to make it feel connected – I prefer the slightly darker color, but want to crowdsource some thoughts. From there, tweaked a little bit of spacing and included a 1pt light-gray (#C6D2E1) top border on the bottom tray to provide some contrast with the map style. Just some quick exploring - thoughts? /cc @bsudekum @frederoni @1ec5 

banner-comparison

bsudekum:
@Cjballard noting, we have two styles: day and night. This darker tone is usually used in our night style.

Cjballard:
@bsudekum Definitely. Don't mean to infringe on night mode, but explore how we can put some contrast & color in the current day theme UI. Mapbox core blue felt a little too Google-blue in this context, but I can experiment with some other options.

Cjballard:
An easier solution without changing banner color could be adding a 1pt light-gray (#e2e2e2 or whichever our preferred light gray we already use) line to the top of each trip tray & arrival card element to create some distinction. It's still light, but starts to create separation b/w elements.  @bsudekum 

banner-lighter-style

brsbl:
How about we use #F2F5FA or #BCCBDE per our tertiary brand palette? 
b4db1876-290b-11e7-8bdf-7b7476e1225e

frederoni:
Made a few updates here using colors from the brand palette. The close button in the steps view should probably have a top separator but how about the colors? Too distinct?

|   |  |
| ------------- | ------------- |
|   |   |
|   |   |

Cjballard:
@frederoni Yeah, especially with the light gray/blue that's looking a little thick.

frederoni:
1px line

|   |  |
| ------------- | ------------- |
|   |  |
|  |  |

Cjballard:
@frederoni Much better, looks good to me.

1ec5:
Note that, despite the title of this PR, `navigationGuidanceDayStyleURL` and `navigationGuidanceNightStyleURL` still point to the v2 styles. #1347 continues to track upgrading to the v3 styles.

PR #1368, opened 2018-05-02
Do not use IUO

bsudekum:
/cc @JThramer

### DIFF
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -251,8 +251,8 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         
         let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction, routeProgress: routeProgress!) ?? instruction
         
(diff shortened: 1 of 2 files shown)


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -251,8 +251,8 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
         
         let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction, routeProgress: routeProgress!) ?? instruction
         
-        if #available(iOS 10.0, *), utterance?.voice == nil, let legProgress = routeProgress!.currentLegProgress {
-            utterance = AVSpeechUtterance(attributedString: modifiedInstruction.attributedText(for: legProgress))
+        if #available(iOS 10.0, *), utterance?.voice == nil {
+            utterance = AVSpeechUtterance(attributedString: modifiedInstruction.attributedText(for: routeProgress!.currentLegProgress))
```
JThramer: Nit: it would be _slightly_ cleaner if we did a `guard let routeProgress = self.routeProgress else { return preconditionFailure("routeProgress should not be nil.") }` on line 233, thereby avoiding the force-unwrap here, but this pattern works just fine. Dealer's choice.

PR #1372, opened 2018-05-02
Allow for unlimited amount of alt routes

bsudekum:
The directions api allows for 3 alt routes now. We should scale beyond just two.

simulator screen shot - iphone x - 2018-05-02 at 14 50 01

/cc @mapbox/navigation-ios


MapboxNavigation/NavigationMapView.swift
```diff
@@ -87,6 +87,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     var animatesUserLocation: Bool = false
     var altitude: CLLocationDistance = defaultAltitude
     var routes: [Route]?
+    var altRouteSourceIdentifiers: [String] = []
+    var altRouteLayerIdentifiers: [String] = []
```
1ec5: We only need one layer for all alternative routes. In fact, we might even be able to combine that layer into the one for the primary route: the expression for `lineColor` would get more complex, but there would be a net decrease in code I think.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -87,6 +87,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     var animatesUserLocation: Bool = false
     var altitude: CLLocationDistance = defaultAltitude
     var routes: [Route]?
+    var altRouteSourceIdentifiers: [String] = []
```
1ec5: We only need one source for all the routes. The source’s shape would be an MGLShapeCollectionFeature containing one MGLPolylineFeature for each route. Each polyline feature’s identifier would be the index of that route.

MapboxNavigation/NavigationMapView.swift
bsudekum: Sounds good. Selecting routes will still be doable?

1ec5: Yes, feature querying should get you the specific feature within the overall shape collection feature that was selected. Then you can use the feature’s identifier to distinguish one route from another.

bsudekum:
@mapbox/navigation-ios this could use a review.

bsudekum:
Adding casing back in https://github.com/mapbox/mapbox-navigation-ios/pull/1372/commits/48f0888e633ca0a9b7b87440f95ac68e19c75fb7

Issue #1078, opened 2018-01-29
User puck should not begin turning while stopped at a red light


bsudekum:
Sometimes while stopped at a red light an the maneuver is to take a 90 degree left/right turn, I notice the calculated course is somewhere between the current and upcoming step. 

```
             90°
^----------------->
|
|
| 0°
|
|
|
+
```

In the case above, the user puck would directly in the middle of the intersection, and the calculated course would be 45°. 

One definite answer to why this is occurring is because we allow the calculated course to derive from both the current and upcoming maneuver:

https://github.com/mapbox/mapbox-navigation-ios/blob/39c88ce32e0e4f8cf2df72c6974b2caebc12dc9f/MapboxCoreNavigation/RouteController.swift#L335

Unsure how this would effect things, but one option is to allow the snapped location to be calculated by `nearByCoordinates` and allow the snapped course to be calculated with the current step coordinates only.

/cc @mapbox/navigation-ios

PR #1385, opened 2018-05-04
Dead-reckon user puck

bsudekum:
From a few discussions we've had, for a couple of reasons the user puck lags behind where the user is in the real world by quite a bit in some cases.

This PR seeks to solve this lag by projecting the user puck ahead 1 second at a distance based on their speed.


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -88,25 +88,30 @@ extension CLLocation {
...
-        let userCoordinate = closest.coordinate
+        let userCoordinate = projectedLocation
         guard let firstCoordinate = legProgress.leg.steps.first?.coordinates?.first else { return nil }
         
         guard shouldSnap(toRouteWith: calculatedCourseForLocationOnStep, distanceToFirstCoordinateOnLeg: self.coordinate.distance(to: firstCoordinate)) else { return nil }
         
         return CLLocation(coordinate: userCoordinate, altitude: altitude, horizontalAccuracy: horizontalAccuracy, verticalAccuracy: verticalAccuracy, course: userCourse, speed: speed, timestamp: timestamp)
     }
     
+    var projectedDistance: CLLocationSpeed {
```
1ec5: A speed is not a distance.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -88,25 +88,30 @@ extension CLLocation {
...
+        // Account speed being zero.
+        return max(speed * RouteControllerDeadReckoningTimeInterval, 1)
```
1ec5: If the user is stationary, this puts the user 1&nbsp;meter ahead of their actual location. Not a significant problem in practice, but this will throw a wrench into future location-based tests.

akitchen:
> #1078 will become more important

Indeed, we will probably need to ease in/ease out of this forward projection, as well as revisit the timing/phasing of certain related animations like rotation. Perhaps there are ways CoreMotion can help us (if supported & enabled) but I definitely agree, new & different edge cases could easily emerge as a result of implementing this concept (whether done as proposed here or in a different way)

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -88,25 +88,30 @@ extension CLLocation {
         }
         
         guard let closest = Polyline(nearByCoordinates).closestCoordinate(to: coordinate) else { return nil }
-        guard let calculatedCourseForLocationOnStep = interpolatedCourse(along: nearByCoordinates) else { return nil }
+        guard let projectedLocation = Polyline(nearByCoordinates).trimmed(from: closest.coordinate, distance: projectedDistance).coordinates.last else { return nil }
```
1ec5: `interpolatedCourse(along:)` already applies a weird half-second or 15-meter dead-reckoning interval (whichever is greater). What does this additional “projected distance” do to that dead reckoning?

bsudekum: This is only used for calculating the points ahead/behind the user to create an angle used for the snapped course.

d-prukop:
In response to @1ec5 above: There will definitely be instances where the puck will appear to continue beyond your position. Stop signs and lights will cause this to happen. The thing to keep in mind is that it typically takes longer than a few seconds to come to a stop. Even from a rather low speed. This trending towards stop should mitigate the overshoot issue. If it takes less than a second to come to a complete stop, you probably have other concerns than puck drifting in that moment.

MapboxCoreNavigation/CLLocation.swift
akitchen: Maybe 1/2 second isn't enough. Or maybe we were also suffering from delays due to unrelated performance issues. I'm really not sure, which is why I want to try to measure this in the field today.

1ec5:
Yes, as I said in https://github.com/mapbox/mapbox-navigation-ios/pull/1385#discussion_r186184032, I know it isn’t a practical problem, but this inconsistency will prove problematic as we add tests. (Presumably we’re going to add tests for this change.) The fewer weird edge cases we create, the better.

1ec5:
> There will definitely be instances where the puck will appear to continue beyond your position. Stop signs and lights will cause this to happen. The thing to keep in mind is that it typically takes longer than a few seconds to come to a stop. Even from a rather low speed. This trending towards stop should mitigate the overshoot issue. If it takes less than a second to come to a complete stop, you probably have other concerns than puck drifting in that moment.

Overshoot is very discernable in our implementation of NavigationMapView: #939.

bsudekum:
new3

_green is projected location, red is raw location_

@frederoni any ideas why it seems to be animating back to the raw location here?

bsudekum:
@1ec5 FYI, updated this to not project forward when near the maneuver. This allows this to play nicely with https://github.com/mapbox/mapbox-navigation-ios/pull/1425

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -114,17 +114,26 @@ extension CLLocation {
...
         
-        return nearbyCoordinates
+        return legProgress.nearbyCoordinates
     }
     
+    func projectedDistance(for distanceRemaining: CLLocationDistance) -> CLLocationDistance {
+        // If the user is near the end of the route, do not project the puck forward.
+        // Otherwise, this can cause strange wrapping at maneuvers.
+        guard distanceRemaining > RouteControllerManeuverZoneRadius else {
+            return 0
```
bsudekum: this is looking pretty jagged when it occurs on a straight maneuver. I wonder if we can apply some sort of curve to this value that eases in and out nicely.

bsudekum:
When testing on a highway today, it almost seemed a little too far ahead. Maybe we cap this at some value?

bsudekum:
I think the final item to figure out is coming to a semi-abrupt stop from going 50mph. In this case, as location updates come in and the speed reduces, the puck slowly starts to come back to reality. 

Ideally, as the user slows down we would quickly be able to revert back to reality.

bsudekum:
> I think the final item to figure out is coming to a semi-abrupt stop from going 50mph. In this case, as location updates come in and the speed reduces, the puck slowly starts to come back to reality.

We could just dead reckon further into the future, say only 0.5 seconds.

1ec5:
> as location updates come in and the speed reduces, the puck slowly starts to come back to reality.

This is another form of overshoot, which I raised above. It’s a fundamental shortcoming of dead reckoning. We can sort of hack around it by modulating the dead reckoning as the user approaches an intersection, but that’s only going to have the effect you want if we somehow know whether the user is stopping at that intersection or not. (In other words, we’d have to know whether there’s a stop sign or red light.) Otherwise, we’d end up undershooting, just as we do on master, around every intersection.

akitchen:
A certain amount of overshoot & correction is to be expected, and users are used to this from other maps apps as well as ridesharing apps. The tradeoffs are worth it. 

We could and should continue to improve the experience with iterative adjustments and tuning in areas such as intersections or certain types of maneuvers. In an abrupt stop from 50mph+ however, the user isn't typically looking at the phone (whether driver or passenger). I don't think that's a high priority to address, so long as the correction is gracefully animated.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -71,14 +71,16 @@ extension CLLocation {
     
     //MARK: - Route Snapping
     
-    func snapped(to legProgress: RouteLegProgress) -> CLLocation? {
+    func snapped(to legProgress: RouteLegProgress, previousSpeeds: [CLLocationSpeed]) -> CLLocation? {
```
1ec5: This makes route snapping stateful, no longer purely a function of a single CLLocation. So this method should be moved out of the CLLocation extension and into RouteController, which stores the previous speeds as an ivar.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -114,17 +114,27 @@ extension CLLocation {
...
         }
         
-        return nearbyCoordinates
+        return legProgress.nearbyCoordinates
     }
     
+    func projectedDistance(for distanceRemaining: CLLocationDistance, previousSpeeds: [CLLocationSpeed]) -> CLLocationDistance {
+        var defaultDeadReckonDistance = speed * RouteControllerDeadReckoningTimeInterval
+        
+        // If the user is reducing in speed, start to dead reckon less
```
1ec5: Should we also start dead-reckoning more if the user starts accelerating, to avoid undershooting?

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -114,17 +114,27 @@ extension CLLocation {
...
+            defaultDeadReckonDistance = defaultDeadReckonDistance + averageDelta
+        }
+        return max(defaultDeadReckonDistance, 0)
+    }
     
     /**
      Given a location and a series of coordinates, compute what the course should be for a the location.
      */
-    func interpolatedCourse(along coordinates: [CLLocationCoordinate2D]) -> CLLocationDirection? {
+    func interpolatedCourse(along coordinates: [CLLocationCoordinate2D], alternateCoordinate: CLLocationCoordinate2D? = nil) -> CLLocationDirection? {
```
1ec5: “Alternate coordinate” is too vague. The name should say what the alternative coordinate is used for, or how it relates to other coordinates being passed in.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -114,17 +114,27 @@ extension CLLocation {
...
+        return legProgress.nearbyCoordinates
     }
     
+    func projectedDistance(for distanceRemaining: CLLocationDistance, previousSpeeds: [CLLocationSpeed]) -> CLLocationDistance {
+        var defaultDeadReckonDistance = speed * RouteControllerDeadReckoningTimeInterval
+        
+        // If the user is reducing in speed, start to dead reckon less
+        let averageDelta = previousSpeeds.reduce(0, +) / Double(previousSpeeds.count)
+        if averageDelta < 0 {
+            defaultDeadReckonDistance = defaultDeadReckonDistance + averageDelta
```
1ec5: `defaultDeadReckonDistance` is a distance measured in meters. It’s incorrect to add `averageDelta`, which is measured in meters per second.

bsudekum:
I think https://github.com/mapbox/mapbox-navigation-ios/pull/1385/commits/1c1a92568f443998b2847b365c1bbae052b27846 ended up preventing dead reckoning too much. I'm going to revert this change and take a look at doing something similar after this merges.

Issue #1391, opened 2018-05-06
Incorrect voice instruction read from cache


bsudekum:
I [noticed this route](https://www.mapbox.com/get-directions/#14.99/37.7921/-122.4025?coordinates=-122.400157,37.787735;-122.406105,37.797461) gave me a bogus instruction (I will update this with the actual text, perhaps there is a hint in the instruction that it read) when turning right onto 3rd street. I verified this indeed is an on-device caching issue by making the exact request to the server which produced the correct instruction. Noting where and how we're creating cache keys:

https://github.com/mapbox/mapbox-navigation-ios/blob/d7f8c4d075ca578983cbe02d819d74eb7ac284c4/MapboxNavigation/DataCache.swift#L63-L65

https://github.com/mapbox/mapbox-navigation-ios/blob/d7f8c4d075ca578983cbe02d819d74eb7ac284c4/MapboxNavigation/Cache.swift#L112-L130

/cc @mapbox/navigation-ios

akitchen:
The problem is likely that cacheKeyForKey is too naive; the last path component can easily collide.

bsudekum:
I was able to track down a hash key for the instruction that was read aloud: `8305224406764629362`. Looking for the hash key for the correct instruction now.

bsudekum:
### Correct instruction:

> <speak><amazon:effect name="drc"><prosody rate="1.08">Continue on <say-as interpret-as="address">3rd</say-as> Street for 1 mile</prosody></amazon:effect></speak>

### Incorrect instruction

> <speak><amazon:effect name="drc"><prosody rate="1.08">Continue on <say-as interpret-as="address">I-680</say-as> Street for 6 miles</prosody></amazon:effect></speak>

bsudekum:
Okay, I went through all audio files in the cache on my device, the correct voice instruction was not there.

bsudekum:
Digging a bit deeper, I looked around for a route that had the incorrect instruction. Turns out, a route that includes this instruction, also has an error when this instruction should be spoken, [route](https://www.mapbox.com/get-directions/#11.59/37.9085/-122.3106?coordinates=-122.299111,37.850395;-122.322136,37.96646).

This instruction is telling me to continue for 3 miles instead of 6!

![simulator screen shot - iphone x - 2018-05-08 at 10 44 15](https://user-images.githubusercontent.com/1058624/39773388-cecd6ec0-52ac-11e8-84f7-ac233914288e.png)

[...]

PR #1396, opened 2018-05-07
Use MD5 for creating cache keys

bsudekum:
_Beginning to fix https://github.com/mapbox/mapbox-navigation-ios/issues/1391_

As pointed out by @akitchen, `cacheKeyForKey` is most likely too naive. 

https://github.com/mapbox/mapbox-navigation-ios/blob/d7f8c4d075ca578983cbe02d819d74eb7ac284c4/MapboxNavigation/Cache.swift#L122-L128

- [x]  add failing test case
- [x] actually fix code

/cc @mapbox/navigation-ios


MapboxNavigation/Cache.swift
```diff
@@ -109,22 +109,22 @@ internal class FileCache {
...
 
-    private func cacheKeyForKey(_ key: String) -> String {
+    func cacheKeyForKey(_ key: String) -> String {
         if let keyAsURL = URL(string: key) {
-            return String.init(keyAsURL.lastPathComponent.hashValue)
+            return MD5(keyAsURL.lastPathComponent)
         }
 
-        return String.init(key.hashValue)
+        return MD5(key)
```
bsudekum: This single line is the actual fix. As pointed out by @akitchen, there are some shortcuts String.hashValue takes some shortcuts and things in the middle of the string may not impact the hash.

MapboxNavigation/Cache.swift
```diff
@@ -109,22 +109,22 @@ internal class FileCache {
...
+    func cacheURLWithKey(_ key: String) -> URL {
         let cacheKey = cacheKeyForKey(key)
         return diskCacheURL.appendingPathComponent(cacheKey)
     }
 
-    private func cacheKeyForKey(_ key: String) -> String {
+    func cacheKeyForKey(_ key: String) -> String {
         if let keyAsURL = URL(string: key) {
-            return String.init(keyAsURL.lastPathComponent.hashValue)
+            return MD5(keyAsURL.lastPathComponent)
```
akitchen: I guess I'm surprised to see us still keying off of `lastPathComponent` -- I thought this was part of what was biting us. (?)

MapboxNavigation/Cache.swift
akitchen: Do voice instructions fall through this code path? Asking for a friend...

MapboxNavigation/Cache.swift
bsudekum: Not that I am aware of. It really only came down to two different strings having the same hashValue: https://github.com/mapbox/mapbox-navigation-ios/issues/1391#issuecomment-387490015

bsudekum: I'm happy to remove this check though. I think creating an md5 of the `key` here would be okay.

MapboxNavigationTests/DataCacheTests.swift
```diff
@@ -106,4 +106,27 @@ class DataCacheTests: XCTestCase {
 
         NotificationCenter.default.post(name: .UIApplicationDidReceiveMemoryWarning, object: nil)
     }
+    
+    func testCacheKeyForKey() {
+        let threeMileInstruction = "<speak><amazon:effect name=\"drc\"><prosody rate=\"1.08\">Continue on <say-as interpret-as=\"address\">I-80</say-as> East for 3 miles</prosody></amazon:effect></speak>"
+        let sixMileInstruction = "<speak><amazon:effect name=\"drc\"><prosody rate=\"1.08\">Continue on <say-as interpret-as=\"address\">I-80</say-as> East for 6 miles</prosody></amazon:effect></speak>"
+        XCTAssertNotEqual(cache.fileCache.cacheKeyForKey(threeMileInstruction), cache.fileCache.cacheKeyForKey(sixMileInstruction))
```
akitchen: Great test case, this is the one that nails it.

MapboxNavigationTests/DataCacheTests.swift
```diff
@@ -106,4 +106,27 @@ class DataCacheTests: XCTestCase {
...
+        XCTAssertEqual(cache.fileCache.cacheKeyForKey("-"), cache.fileCache.cacheKeyForKey("-"))
+    }
+    
+    func testCacheKeyPerformance() {
+        let instructionTurn = "Turn left"
+        let instructionContinue = "<speak><amazon:effect name=\"drc\"><prosody rate=\"1.08\">Continue on <say-as interpret-as=\"address\">I-80</say-as> East for 3 miles</prosody></amazon:effect></speak>"
+        measure {
+            for _ in 0...1000 {
+                _ = cache.fileCache.cacheKeyForKey(instructionTurn)
+                _ = cache.fileCache.cacheKeyForKey(instructionContinue)
```
1ec5: Can you post times from master and this branch for comparison?

MapboxNavigation/MD5.swift
```diff
@@ -0,0 +1,273 @@
...
+        output.append(value1)
+        output.append(value2)
+    }
+    
+    return String(output)
+}
+
+/*
+ * Convert a raw string to an array of little-endian words
+ * Characters >255 have their high-byte silently ignored.
```
1ec5: We do care about the high byte. Add test cases for non-ASCII characters, all the way up to emoji like 🛡 (U+1F6E1). Although emoji would be somewhat contrived for an OSM-based data set, there are plenty of multibyte characters in OSM, especially in Chinese and Japanese.

1ec5: Looks like we’re no longer using this code, but a test case would still be good for peace of mind.

MapboxNavigationTests/DataCacheTests.swift
bsudekum: Yeah, it was 0.001 seconds for the same test.

Issue #1392, opened 2018-05-06
Exit view should work in night mode


bsudekum:
**Mapbox Navigation SDK version: master**
![image](https://user-images.githubusercontent.com/1058624/39676110-b57bff7a-5119-11e8-8054-7933fb8c6502.png)

The exit view should be white here.

/cc @JThramer

bsudekum:
Noting this a pretty rare case as it only opens when in a tunnel and the top banner contains an exit.

PR #1398, opened 2018-05-07
Key exit view cache with more information

bsudekum:
Fixes: https://github.com/mapbox/mapbox-navigation-ios/issues/1392

We need to use a more descriptive cache key for exit views since a single exit view can be displayed with many different colors.

Ideally, we'd cache based on the `Style` the exit view is used within, but unfortunately we do not have this information available to us in this section of the code.

/cc @JThramer

### DIFF
--- a/MapboxNavigation/ExitView.swift
+++ b/MapboxNavigation/ExitView.swift
@@ -63,7 +63,8 @@ class ExitView: StylableView {
     
     override var hashValue: Int {
         get {
-             return [foregroundColor?.cgColor.hashValue ?? 0, backgroundColor?.cgColor.hashValue ?? 0, exitText?.hashValue ?? 0, pointSize.hashValue, side.hashValue].reduce(0, +)
+            let criticalProperties: [AnyHashable?] = [foregroundColor, backgroundColor, exitText, pointSize, side]
+            return criticalProperties.compactMap { $0?.hashValue }.reduce(0, ^)
         }
     }
     
--- a/MapboxNavigation/ExitView.swift
+++ b/MapboxNavigation/ExitView.swift
@@ -53,10 +53,17 @@ class ExitView: StylableView {
             invalidateIntrinsicContentSize()
         }
     }
+    
     var pointSize: CGFloat {
         didSet {
             exitNumberLabel.font = exitNumberLabel.font.withSize(pointSize * ExitView.labelFontSizeScaleFactor)
             rebuildConstraints()
+        }
+    }
+    
+    override var hashValue: Int {
+        get {
+             return [foregroundColor?.cgColor.hashValue ?? 0, backgroundColor?.cgColor.hashValue ?? 0, exitText?.hashValue ?? 0, pointSize.hashValue, side.hashValue].reduce(0, +)
         }
     }
     
--- a/MapboxNavigation/ExitView.swift
+++ b/MapboxNavigation/ExitView.swift
@@ -64,7 +64,7 @@ class ExitView: StylableView {
     override var hashValue: Int {
         get {
             let criticalProperties: [AnyHashable?] = [foregroundColor, backgroundColor, exitText, pointSize, side]
-            return criticalProperties.compactMap { $0?.hashValue }.reduce(0, ^)
+            return criticalProperties.reduce(0, { $0 ^ ($1?.hashValue ?? 0)})
         }
     }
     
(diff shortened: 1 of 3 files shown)


MapboxNavigation/InstructionPresenter.swift
```diff
@@ -186,7 +186,11 @@ class InstructionPresenter {
         
         let view = ExitView(pointSize: dataSource.font.pointSize, side: side, text: text)
         let attachment = ExitAttachment()
-        guard let cacheKey = component.cacheKey() else { return nil }
+        guard var cacheKey = component.cacheKey() else { return nil }
+        
+        if let foregroundColor = view.foregroundColor {
+            cacheKey = "\(cacheKey)-\(String(foregroundColor.cgColor.hashValue))"
```
JThramer: So I was taking a look at this earlier today... We _could_ key off of the active style, but it would be a PITA to pass the reference to the `StyleManager` down to the `InstructionPresenter` with the current architecture, totally agree with your assessment.

So with that solution being a no-go, what I would probably suggest here is to implement a custom `hashValue` for `ExitView`. NSHipster [defines](http://nshipster.com/equality/) a good hashing mechanism as "one that XOR's over all critical properties", which I would define in exitView like this: 

```swift
override var hashValue: Int {
    get {
      return [foregroundColor, backgroundColor, side, exitText, pointSize].map({ $0.hashValue }).reduce(0, ^)
    }
}
```

With this logic, we'll cover _all_ our bases, not just `foregroundColor`.

Also: Hash values are [not guaranteed](https://developer.apple.com/documentation/swift/hashable) to be consistent across different runs of your app, so we definitely shouldn't be persisting this cache across different app runs. It looks like we [don't](https://github.com/mapbox/mapbox-navigation-ios/pull/1398/files#diff-af48d70309ee195e34cc5e6cc6b27caaL195) disk cache, but I wanted to call that out anyways, as it's a really nasty pitfall of `Hashable`.

MapboxNavigation/ExitView.swift
```diff
@@ -53,13 +53,20 @@ class ExitView: StylableView {
...
     var pointSize: CGFloat {
         didSet {
             exitNumberLabel.font = exitNumberLabel.font.withSize(pointSize * ExitView.labelFontSizeScaleFactor)
             rebuildConstraints()
         }
     }
     
+    override var hashValue: Int {
+        get {
+             return [foregroundColor?.cgColor.hashValue ?? 0, backgroundColor?.cgColor.hashValue ?? 0, exitText?.hashValue ?? 0, pointSize.hashValue, side.hashValue].reduce(0, +)
```
JThramer: Three things:

1. Why are you keying off of `cgColor`? `UIColor's` hash conformance has good consistency across instances. E.g: Two `UIColor` objects with the same RGB values will result in the same hash.
2. Using addition here is a problem. The reason why XOR is considered the best solution (out of all common boolean/arithmetic algebra functions) for combining hashes is because it has [better distribution](http://eigenjoy.com/2011/05/04/why-is-xor-the-default-way-to-combine-hashes) then other common boolean operations, and it's considered preferable to addition because you're guaranteed to [stay in the same bit-space](https://en.wikipedia.org/wiki/Integer_overflow), as well as having better [entropy properties](https://stackoverflow.com/questions/5533971/does-adding-two-hash-values-generate-a-valid-hash#comment9328749_5533985). Some folks mitigate this by doing some remainder math when they reach an overflow scenario, but I'd be worried about that introducing modulo-bias.
3. I didn't notice that some of those properties are optionals that does change things, but I would still argue that a more functional representation here is both doable and would read nicer than what we have currently.

``` swift
return [foregroundColor, backgroundColor, exitText, pointSize, side].compactMap({ $0?.hashValue }).reduce(0, ^)

// Or, if you're worried about performance (although I don't think there would be any real impact, and it wouldn't be quite as readable) 

return [foregroundColor, backgroundColor, exitText, pointSize, side].reduce(0, { $0 ^ $1?.hashValue ?? 0})
```

JThramer: Supplemental: 

If, for some reason, we _do_ decide to keep arithmetic combination of hashes, we should _absolutely_ use [Overflow Operators](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AdvancedOperators.html#//apple_ref/doc/uid/TP40014097-CH27-ID37) (`&+`) over a regular addition operator, as the regular one will produce a swift error if it encounters an overflow situation.

bsudekum: The problem is that we're mixing types here and that calling compactMap on an array of mixed types produces some compiler errors.

PR #1400, opened 2018-05-08
Adjust ImageDownloader tests

akitchen:
URL loading is deferred until after the current run loop iteration. The two reentrancy tests are adjusted to use the download operation's `isFinished` flag as a continuation criteria, and assertions are done in a more appropriate place for better failure localization.


akitchen:
This has passed on Xcode 9.3 at least 5 times now. I think we can merge and pick it up from here if needed.

PR #1404, opened 2018-05-09
Remove OS_ACTIVITY_MODE from Objective-C example scheme

1ec5:
Rolled back part of #188 that set the `OS_ACTIVITY_MODE` flag when running the example application. This flag was only needed in older Xcode versions to suppress console spew, but it also suppresses legitimate logging from the SDK and its dependencies.

This flag has already been removed from the Swift example application’s scheme as of #1063.

/cc @bsudekum @JThramer


1ec5:
One of the build bots [failed](https://www.bitrise.io/build/1c7e5719f45abeb6) on `testDownloadingSameImageWhileInProgressAddsCallbacksWithoutAddingAnotherRequest`.

PR #1405, opened 2018-05-09
Moves access to active download operations to a dispatch queue.

akitchen:
Reads are done synchronously while writes are performed with a barrier.

The occasionally flaky test was pointing out the lack of thread-safety
here, as `downloadImage(with:)` and URL loading callbacks happen on
different queues.

### DIFF
--- a/MapboxNavigation/ImageDownloader.swift
+++ b/MapboxNavigation/ImageDownloader.swift
@@ -10,20 +10,23 @@ protocol ReentrantImageDownloader {
 
 class ImageDownloader: NSObject, ReentrantImageDownloader, URLSessionDataDelegate {
     private var sessionConfiguration: URLSessionConfiguration = URLSessionConfiguration.default
-    private var operationType: ImageDownload.Type = ImageDownloadOperation.self
 
-    private var queue: OperationQueue
     lazy private var urlSession: URLSession = {
         return URLSession.init(configuration: sessionConfiguration, delegate: self, delegateQueue: nil)
     }()
 
-    private var headers: [String: String] = ["Accept": "image/*;q=0.8"]
+    private var downloadQueue: OperationQueue
+    private var accessQueue: DispatchQueue
 
+    private var operationType: ImageDownload.Type = ImageDownloadOperation.self
     private var operations: [URL: ImageDownload] = [:]
 
+    private var headers: [String: String] = ["Accept": "image/*;q=0.8"]
+
     override init() {
-        self.queue = OperationQueue()
-        self.queue.name = Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloader"
+        self.downloadQueue = OperationQueue()
+        self.downloadQueue.name = Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloader"
+        self.accessQueue = DispatchQueue.init(label: Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloaderInternal")
     }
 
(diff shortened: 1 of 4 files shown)


MapboxNavigation/ImageDownloader.swift
```diff
@@ -10,20 +10,23 @@ protocol ReentrantImageDownloader {
...
     private var operations: [URL: ImageDownload] = [:]
 
+    private var headers: [String: String] = ["Accept": "image/*;q=0.8"]
+
     override init() {
-        self.queue = OperationQueue()
-        self.queue.name = Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloader"
+        self.downloadQueue = OperationQueue()
+        self.downloadQueue.name = Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloader"
+        self.accessQueue = DispatchQueue.init(label: Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloaderInternal")
```
1ec5: Nit: drop `.init`.

MapboxNavigation/ImageDownloader.swift
```diff
@@ -56,10 +61,15 @@ class ImageDownloader: NSObject, ReentrantImageDownloader, URLSessionDataDelegat
     }
 
     func activeOperationWithURL(_ url: URL) -> ImageDownload? {
```
1ec5: While we’re here, would you mind changing the method name to `activeOperations(url:)`? The “with” followed by `_` is weirding me out.

MapboxNavigation/ImageDownloader.swift
```diff
@@ -45,9 +48,11 @@ class ImageDownloader: NSObject, ReentrantImageDownloader, URLSessionDataDelegat
             operation = activeOperation
         } else {
             operation = operationType.init(request: request, in: self.urlSession)
-            self.operations[url] = operation
+            accessQueue.async(flags: .barrier) {
+                self.operations[url] = operation
```
1ec5: Is it necessary to do a weakSelf/strongSelf dance here? (×2)

MapboxNavigation/ImageDownloader.swift
akitchen: Ah yes, autocomplete.

MapboxNavigation/ImageDownloader.swift
akitchen: No, I don't believe so. The implicit retain ensures self (and the access queue) exist long enough for the block to be executed. The block is transient, so when it completes self would be released.

MapboxNavigation/ImageDownloader.swift
akitchen: 👍

MapboxNavigation/ImageDownloader.swift
akitchen: it's also worth noting that this object is owned by the ImageRepository, which is a `static let` singleton.

bsudekum:
@akitchen can we hold this for post v0.17.0?

akitchen:
Up to you. Technically this is a bug fix in already-released code.

akitchen:
It is interesting to note that https://github.com/mapbox/mapbox-navigation-ios/pull/1405/commits/1ee99d0deb0b7966193f521c3da6b9ada241d216 introduces the ability to "pause" image downloading, resulting in the URL loading system waiting for a signal before continuing a URL load beyond the initial response. This is done in order to synchronously test multiple requests for the same resource queueing up while that resource is already being loaded. Previously we were relying on asynchronous work to line up a certain way which didn't always happen, resulting in a flaky test.

PR #1406, opened 2018-05-09
Changes to adopt the named arrival instructions from Directions API

### LINKED ISSUE #1123: Adopt named arrival instructions from Directions API
Spoken instructions for an `arrive` step should include the name of the waypoint. In typical SDK usage, the developer supplies a `Waypoint.name` that’s either the name of a point of interest or an address. The user would find it very useful to hear the address when arriving at the destination. Arriving at a named waypoint is a basic use case, so the SDK should handle it without requiring any developer intervention.

Currently, the arrival instructions come from the Directions API, like any other instruction. In the absence of Directions API support for this feature, the SDK would need to ignore the `arrive` step’s `SpokenInstruction` objects in favor of its own homegrown instruction strings. For localization, we have two options:

* Translate the arrival instructions as part of this SDK’s Transifex project. The strings will be somewhat redundant with OSRM Text Instructions. We won’t be able to take advantage of the grammar support in OSRM Text Instructions, which would be very unfortunate for Russian, because declination is required for the waypoint name.
* Add named waypoint strings to OSRM Text Instructions and depend on [OSRM Text Instructions for Swift](https://github.com/Project-OSRM/osrm-text-instructions.swift/) once again. Russian declination would need to be fixed in Project-OSRM/osrm-text-instructions.swift#49.

/cc @mapbox/navigation-ios @mapbox/guidance

vincethecoder:
Since Directions API will be providing waypoint arrival instructions, we decided to delete the special case for arrival banners.

Closes #1123  

TODO:
- [x] Update the example to show how you can name waypoints.
- [x] Update the dependency MapboxDirections.swift to a version that includes mapbox/MapboxDirections.swift#273

3usjt

vfchy

jipe7

vtofy

/cc @1ec5 @bsudekum


bsudekum:
@vincethecoder I think there maybe a couple more additions needed here

1. Update the example to show how you can name waypoints. Even something as simple as `The user's destination` is fine I think. No need to geocode the dropped pin.
1. Update the dependency MapboxDirections.swift to a version that includes https://github.com/mapbox/MapboxDirections.swift/pull/273

vincethecoder:
⚠️ This PR should be merged post new release of the MapboxDirections.swift. ⚠️ 
/cc @mapbox/navigation-ios

MapboxNavigation/ManeuverView.swift
```diff
@@ -83,7 +83,7 @@ open class ManeuverView: UIView {
             ManeuversStyleKit.drawFork(frame: bounds, resizing: resizing, primaryColor: primaryColor, secondaryColor: secondaryColor)
             flip = [.left, .slightLeft, .sharpLeft].contains(direction)
         case .takeRoundabout, .turnAtRoundabout, .takeRotary:
-            ManeuversStyleKit.drawRoundabout(frame: bounds, resizing: resizing, primaryColor: primaryColor, secondaryColor: secondaryColor, roundabout_angle: CGFloat(visualInstruction.primaryInstruction.degrees))
+            ManeuversStyleKit.drawRoundabout(frame: bounds, resizing: resizing, primaryColor: primaryColor, secondaryColor: secondaryColor, roundabout_angle: CGFloat(visualInstruction.primaryInstruction.finalHeading))
```
bsudekum: I think this change should be reverted, degrees is specific to roundabouts.

Examples/Swift/ViewController.swift
```diff
@@ -154,8 +154,9 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
             multipleStopsAction.isEnabled = false
         }
 
+        let destinationName = "Dropped Pin" // Note you can change this. This will be used in the top banner
         let coordinates = mapView.convert(tap.location(in: mapView), toCoordinateFrom: mapView)
-        let waypoint = Waypoint(coordinate: coordinates)
+        let waypoint = Waypoint(coordinate: coordinates, name: destinationName)
```
bsudekum: Want to use `waypoints.count` here to show this can change for a multi-waypoint route?

MapboxNavigation/ManeuverView.swift
vincethecoder: This is a temporary change, as the current master copy of MapboxDirections.swift is missing `degrees`. Had to use `finalHeading` till we publish the changes which has the `degrees` 
 field. 
https://github.com/mapbox/MapboxDirections.swift/blob/master/MapboxDirections/MBVisualInstruction.swift#L38

bsudekum: Oh wow, indeed it was renamed https://github.com/mapbox/MapboxDirections.swift/commit/03979d6ae59d828ed623217512e9903f8af034cd

Examples/Swift/ViewController.swift
```diff
@@ -154,8 +154,9 @@ class ViewController: UIViewController, MGLMapViewDelegate, CLLocationManagerDel
             multipleStopsAction.isEnabled = false
         }
 
+        let destinationName = "Dropped Pin #\(waypoints.endIndex + 1)" // Note you can change this. This will be used in the top banner
```
bsudekum: > This will be used in the top banner

to 

> This will be used in the top banner when arriving at a destination.

bsudekum:
@vincethecoder MapboxDirections.swift v0.21.0 has been released. Don't forget to update the appropriate podspec files as well.

bsudekum:
Don't forget to also update MapboxCoreNavigation.podspec.

PR #1410, opened 2018-05-10
RouteController - Retain Cycles & Zombie Refs

### LINKED ISSUE #1408: [Crash] RouteController didn't nil locationManager delegate
Used library version MapboxCoreNavigation (0.13.0).

In RouteController you have the code:
`@objc public var locationManager: NavigationLocationManager! {
        didSet {
            oldValue?.delegate = nil
            locationManager.delegate = self
        }
    }`

NavigationLocationManager is subclassed from CCLocationManager. That means `locationManager.delegate = self`
There we assign delegate self. It has **assign**, not **weak** attribute. In this case when we create RouteController and destroy them we got a crash if user change them location. Because locationmanager didn't nil delegate. And he has a reference to deallocated object.

For fix this behavior just add `locationManager.delegate = nil` in deinit method

JThramer:
My favorite game!

Fixes #1408. Two issues are root cause here:

- A strong reference capture in a `RouteController` completion closure that causes a retain cycle,
- A unsafe_unretained reference persisting in the `NavigationLocationManager` that causes a crash.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -378,6 +378,7 @@ open class RouteController: NSObject {
     @objc public func suspendLocationUpdates() {
         locationManager.stopUpdatingLocation()
         locationManager.stopUpdatingHeading()
+        locationManager.delegate = nil
     }
 
     /**
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -286,8 +286,8 @@ open class RouteController: NSObject {
     private func setupTunnelIntersectionManager() {
         tunnelIntersectionManager = TunnelIntersectionManager()
         tunnelIntersectionManager?.delegate = self
-        tunnelIntersectionManagerCompletionHandler = { enabled, _ in
-            self.tunnelIntersectionManager?.isAnimationEnabled = enabled
+        tunnelIntersectionManagerCompletionHandler = { [weak self] enabled, _ in
+            self?.tunnelIntersectionManager?.isAnimationEnabled = enabled
         }
     }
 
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -286,8 +286,8 @@ open class RouteController: NSObject {
     private func setupTunnelIntersectionManager() {
         tunnelIntersectionManager = TunnelIntersectionManager()
         tunnelIntersectionManager?.delegate = self
-        tunnelIntersectionManagerCompletionHandler = { enabled, _ in
-            self.tunnelIntersectionManager?.isAnimationEnabled = enabled
+        tunnelIntersectionManagerCompletionHandler = { [weak self] enabled, _ in
```
vincethecoder: Nice 👍

MapboxCoreNavigation/RouteController.swift
```diff
@@ -378,6 +378,7 @@ open class RouteController: NSObject {
     @objc public func suspendLocationUpdates() {
         locationManager.stopUpdatingLocation()
         locationManager.stopUpdatingHeading()
+        locationManager.delegate = nil
```
vincethecoder: @akitchen I had mentioned this recently, do we want to keep the `locationManager` alive for other intended purposes like _telemetry_ etc.?

bsudekum: After deiniting, I don't think so.

akitchen: No, I was mainly saying that removing this object’s relationship to the location manager doesn’t necessarily power it down or reduce power consumption. Doing this in deinit is a good plan.

akitchen: Doing it here is fine too, if we’re not interested in receiving updates.

Issue #975, opened 2018-01-05
User puck is jumpy on end of route screen


bsudekum:
![new3](https://user-images.githubusercontent.com/1058624/34620941-10c57e9c-f1fc-11e7-8a8c-957effc12949.gif)

/cc @frederoni @JThramer

bsudekum:
Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/1411

PR #1411, opened 2018-05-10
Improve arrival location and end of route screen animation

bsudekum:
This fixes a few issues:

* Animations when showing the end of route view controller: https://github.com/mapbox/mapbox-navigation-ios/issues/975
* The location at which the arrival screen is shown

Previously, we had a hardcoded duration at which the arrival screen was shown. Now this is being handled by the server.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -661,7 +661,7 @@ extension RouteController: CLLocationManagerDelegate {
         let currentDestination = routeProgress.currentLeg.destination
         guard let remainingVoiceInstructions = routeProgress.currentLegProgress.currentStepProgress.remainingSpokenInstructions else { return }
 
-        if routeProgress.currentLegProgress.remainingSteps.count == 1 && remainingVoiceInstructions.count == 0 && currentDestination != previousArrivalWaypoint {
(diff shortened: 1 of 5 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -659,9 +659,9 @@ extension RouteController: CLLocationManagerDelegate {
 
     func updateRouteLegProgress(for location: CLLocation) {
         let currentDestination = routeProgress.currentLeg.destination
-        let legDurationRemaining = routeProgress.currentLegProgress.durationRemaining
+        guard let remainingVoiceInstructions = routeProgress.currentLegProgress.currentStepProgress.remainingSpokenInstructions else { return }
 
-        if legDurationRemaining < RouteControllerDurationRemainingWaypointArrival, currentDestination != previousArrivalWaypoint {
+        if routeProgress.currentLegProgress.remainingSteps.count <= 1 && remainingVoiceInstructions.count == 0 && currentDestination != previousArrivalWaypoint {
```
bsudekum: Noting steps.last is the arrival step, which has a distance of zero. The penultimate step is the real step we're worried about most of the time. In a very rare case, I guess you could end up on the last step via `advanceStep(to:)`.

bsudekum:
@JThramer this could use a review.

JThramer:
👀

PR #1397, opened 2018-05-07
Fixed user puck stuck while in simulation mode.

vincethecoder:
Retrieves a unique successive coordinate after the current coordinate.

Here is the console log from inspecting the `routeProgress?.route.coordinates` at that deadlock location:

```
(lldb) po closestCoordinateOnRoute.index
4

(lldb) po coordinates[closestCoordinateOnRoute.index]
▿ CLLocationCoordinate2D
  - latitude : 37.790279999999996
  - longitude : -122.40714999999999

(lldb) po coordinates[5]
▿ CLLocationCoordinate2D
  - latitude : 37.790279999999996
  - longitude : -122.40714999999999

(lldb) po coordinates[6]
▿ CLLocationCoordinate2D
  - latitude : 37.789969999999997
  - longitude : -122.40708999999998
```

Fixes #1383 

/cc @mapbox/navigation-ios

PR #1414, opened 2018-05-10
Remove tunnel intersection manager callback mechanism

1ec5:
Callback arguments were obscuring the fact that TunnelIntersectionManagerDelegate method implementations and callers were all synchronous. Effectively, properties were getting set to themselves. The callback arguments have been removed and the callbacks inlined into the call sites.

Also fixed a tautology and off-by-one error in the code that transitions from tunnel simulation to standard location management.

/cc ~~@vincentsarago~~ @vincethecoder @bsudekum


MapboxCoreNavigation/TunnelIntersectionManager.swift
```diff
@@ -98,8 +90,7 @@ open class TunnelIntersectionManager: NSObject {
     
     @objc public func enableTunnelAnimation(for manager: CLLocationManager,
                                         routeController: RouteController,
-                                          routeProgress: RouteProgress,
-                                               callback: RouteControllerSimulationCompletionBlock?) {
+                                          routeProgress: RouteProgress) {
         guard !isAnimationEnabled else { return }
         
         isAnimationEnabled = true
```
vincethecoder: @1ec5 we also need to move this down to https://github.com/mapbox/mapbox-navigation-ios/pull/1414/files#diff-987c4bbd828e654bd63d59215f890022L110

And, we need to also add `self.animatedLocationManager?.startUpdatingLocation()` it got removed here.
https://github.com/mapbox/mapbox-navigation-ios/pull/1397/files#diff-9bf14ebab2a9283285d3a8e91bd09e86L54

MapboxCoreNavigation/TunnelIntersectionManager.swift
```diff
@@ -108,22 +99,19 @@ open class TunnelIntersectionManager: NSObject {
         self.animatedLocationManager?.delegate = routeController
         self.animatedLocationManager?.routeProgress = routeProgress
         self.animatedLocationManager?.startUpdatingHeading()
-
```
vincethecoder: @1ec5 we need to also add `self.animatedLocationManager?.startUpdatingLocation()` it got removed here.
https://github.com/mapbox/mapbox-navigation-ios/pull/1397/files#diff-9bf14ebab2a9283285d3a8e91bd09e86L54

MapboxCoreNavigationTests/TunnelIntersectionManagerTests.swift
```diff
@@ -171,9 +170,12 @@ class TunnelIntersectionManagerTests: XCTestCase {
         
         let tunnelExitLocation = location(at: routeController.location!.coordinate, horizontalAccuracy: 20)
 
-        routeController.tunnelIntersectionManager?.delegate?.tunnelIntersectionManager?(routeController.locationManager, willDisableAnimationAt: tunnelExitLocation) { animationEnabled, _ in
-                XCTAssertFalse(animationEnabled, "Animation through tunnel should be disabled.")
```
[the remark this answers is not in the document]
vincethecoder: @1ec5 the reason why this block was never called is `animationEnabled ` was set to `true` in the previous location update.

MapboxCoreNavigationTests/TunnelIntersectionManagerTests.swift
```diff
@@ -140,9 +140,8 @@ class TunnelIntersectionManagerTests: XCTestCase {
         routeController.locationManager(routeController.locationManager, didUpdateLocations: [tunnelLocation])
         
         // Enable the tunnel animation, which should enable the simulated location manager
-        routeController.tunnelIntersectionManager?.delegate?.tunnelIntersectionManager?(routeController.locationManager, willEnableAnimationAt: routeController.location!) { isAnimationEnabled, manager in
-                XCTAssertTrue(isAnimationEnabled, "Animation through tunnel should be enabled.")
```
vincethecoder: @1ec5 the reason why this block was never called is animationEnabled was set to true in the previous location update. In fact, upon the recent decisions to keep both location managers active this means now we don't have a delay in location updates as we did previously.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -286,9 +284,6 @@ open class RouteController: NSObject {
     private func setupTunnelIntersectionManager() {
         tunnelIntersectionManager = TunnelIntersectionManager()
         tunnelIntersectionManager?.delegate = self
-        tunnelIntersectionManagerCompletionHandler = { enabled, _ in
-            self.tunnelIntersectionManager?.isAnimationEnabled = enabled
```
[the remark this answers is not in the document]
1ec5: > However in the future, if we decide to keep only one location manager alive, we need to inform the developer which location manager is active and whether or not animation is enabled.

TunnelIntersectionManager already exposes this information via properties, so the delegator can simply access these properties after calling the delegate methods. A completion handler is only necessary if the delegate method needs to return information asynchronously. That pattern is extremely rare in Cocoa Touch frameworks.

MapboxCoreNavigation/TunnelIntersectionManager.swift
1ec5: Done in d3df73b9a86ef6e9b19ea1af46299bf4b56ff495.

MapboxCoreNavigationTests/TunnelIntersectionManagerTests.swift
```diff
@@ -171,9 +170,12 @@ class TunnelIntersectionManagerTests: XCTestCase {
...
-                XCTAssertFalse(animationEnabled, "Animation through tunnel should be disabled.")
-        }
+        routeController.tunnelIntersectionManager?.delegate?.tunnelIntersectionManager?(routeController.locationManager, willDisableAnimationAt: tunnelExitLocation)
+        XCTAssertTrue(routeController.tunnelIntersectionManager?.isAnimationEnabled ?? false, "Animation through tunnel should remain enabled after 1 location update.")
+        routeController.tunnelIntersectionManager?.delegate?.tunnelIntersectionManager?(routeController.locationManager, willDisableAnimationAt: tunnelExitLocation)
+        XCTAssertTrue(routeController.tunnelIntersectionManager?.isAnimationEnabled ?? false, "Animation through tunnel should remain enabled after 2 location updates.")
+        routeController.tunnelIntersectionManager?.delegate?.tunnelIntersectionManager?(routeController.locationManager, willDisableAnimationAt: tunnelExitLocation)
+        XCTAssertFalse(routeController.tunnelIntersectionManager?.isAnimationEnabled ?? true, "Animation through tunnel should be disabled after 3 location updates.")
```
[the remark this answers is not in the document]
1ec5: 0176348f9587d85cefd2813cb6da1b0331e03afc removes the optional chaining in favor of an explicit assertion.

bsudekum:
This just needs to be cleaned up, then can merge.

Issue #1190, opened 2018-03-02
Imageless banner components should include generic shield icon


1ec5:
If a VisualInstructionComponent lacks an image, it should be drawn as a generic shield so that it doesn’t blend in with the surrounding text. Two or more of these bare numbers can look pretty confusing next to each other. The generic shield could be a white rounded rectangle, as in the Navigation Guidance styles:

<img src="https://user-images.githubusercontent.com/1231218/36926853-d1273528-1e2e-11e8-9661-6d0f097de2e3.png" width="300" alt="20">

<img src="https://user-images.githubusercontent.com/1231218/36927108-3970fcda-1e30-11e8-98e4-de4e2070b6ce.png" width="300" alt="10">

We only support shields inside the U.S. at the moment (#334), so this would be a boon for countries like Canada and Sweden that don’t prefix route numbers with alphabetic prefixes.

/cc @bsudekum @mapbox/guidance

bsudekum:
@1ec5 can you attache the response from the screenshot above? I think there is some confusion on whether there should be shields loaded here or not.

1ec5:
Here’s what the relevant part of the response looks like:

```json
                  "primary": {
                    "type": "off ramp",
                    "modifier": "right",
                    "components": [
                      {
                        "text": "Exit",
                        "type": "exit"
                      },
                      {
                        "text": "25",
                        "type": "exit-number"
                      },
                      {
                        "text": "335",
                        "type": "text"
                      },
                      {
                        "text": "Sud",
                        "type": "text"
                      }
                    ],
                    "text": "Exit 25 335 Sud"
                  }
```

It isn’t surprising that there’s no shield for “335” at the moment, because it’s just a `text` component. When I opened this issue, I figured we could apply shield-like formatting to any component whose `text` consists of only a single alphanumeric word (so no spaces, and at least one digit). But on second thought, I’ve opened an internal issue to request that the Directions API explicitly mark this component as an `icon`. Autodetecting refs on the client side remains an option, but something of a last resort.

1ec5:
No, I think what should happen is that the Directions API reliably assigns these components the `image` type without setting an image, so that the SDK can compose the text atop a bundled image.

The API’s current approach is to vend a separate image for each route in the world. That approach doesn’t scale. There’s a very large and ever growing set of route numbers for which we lack shield artwork. It would be infeasible to maintain a separate image for each one, whereas it would be straightforward to compose a generic shield on the fly.

[...]

PR #1417, opened 2018-05-11
Adding generic shield support to instruction banner.

JThramer:
simulator screen shot - iphone 8 plus - 2018-05-10 at 18 03 09

Adding support for generic shields. Implements #1190. This also has the neat side-effect of rendering normal shields as generic ones until their images load.

To-Do: 
- [x] Fix failing tests
- [x] Add Testing
- [x] `public`-ation
- [x] revert [upstream dependency](https://github.com/mapbox/MapboxDirections.swift/pull/275)
- [x] Changelog entry
- [x] Document `GenericRouteShield`
- [x] Cache `GenericRouteShield` based upon it's foreground color / text
- [x] Remove default case from `VisualInstructionComponent` extension

/cc @mapbox/navigation-ios


MapboxNavigation/GenericRouteShield.swift
```diff
@@ -0,0 +1,85 @@
+import Foundation
+import UIKit
+
+class GenericRouteShield: StylableView {
```
1ec5: Should this class be public, so that developers can customize the colors or roundedness?

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -83,37 +83,44 @@ class InstructionPresenter {
...
-                        continue
-                    }
+            //If we have an icon component, lets turn it into a shield.
+            case .icon:
+                if let shieldString = attributedString(forShieldComponent: component, repository: imageRepository, dataSource: dataSource, onImageDownload: onImageDownload) {
+                    build(component, [joinString, shieldString])
+                } else if let genericShieldString = attributedString(forGenericShield: component, dataSource: dataSource) {
+                    build(component, [joinString, genericShieldString])
+                } else {
+                    fallthrough
```
1ec5: This is a clever use of a switch statement, but do you think the fallthroughs inside conditionals make the code flow a little more difficult to keep track of?

JThramer: @1ec5 It's certainly a complex switch, but I'd make the case that it's a heck of a lot easier to follow than the mess of `if` statements before it.

bsudekum:
This is looking great!

JThramer:
Posterity: determined that test failures are not due to machine. Bisected work on this branch to find root-cause of test issues. First bad commit found to be `fd2a57b37b6d0ce8504dc8468e464de35b5a7cbc`. Not super-helpful (lots of stuff happened in that commit), but it's the first step in helping me narrow down the bad behavior.

MapboxNavigationTests/InstructionsBannerViewSnapshotTests.swift
```diff
@@ -275,13 +276,34 @@ class InstructionsBannerViewSnapshotTests: FBSnapshotTestCase {
...
+    }
 }
 
 extension InstructionsBannerViewSnapshotTests {
     
-    func verifyView(_ view: UIView, size: CGSize) {
+    func verifyView(_ view: UIView, size: CGSize, tolerance: CGFloat = 0.01) {
         view.frame.size = size
-        FBSnapshotVerifyView(view, tolerance: 0.01)
+        FBSnapshotVerifyView(view, suffixes: ["_64"], tolerance: tolerance)
```
bsudekum: This looks suspect.

JThramer: The `suffixes: ["_64"]`? That just tells the snapshot framework that we only have reference images in the `ReferenceImages_64` folder, so it doesn't try to look in the `ReferenceImages_32` folder as a fallback. (Default Behavior)

This basically removes unneeded (since we don't do 32-bit testing) noise from the output log.

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -5,16 +5,20 @@ extension VisualInstructionComponent {
...
+            guard let imageURL = imageURL else { return genericCacheKey }
             return "\(imageURL.absoluteString)-\(VisualInstructionComponent.scale)"
-        case .delimiter:
+        default:
             return nil
         }
     }
+    
+    var genericCacheKey: String {
+        return "generic-" + (text ?? "nil")
```
bsudekum: Are there cases we could end up with different shields that have the same cache key, ie `generic-` because text is nil?

MapboxNavigation/VisualInstructionComponent.swift
```diff
@@ -5,16 +5,20 @@ extension VisualInstructionComponent {
...
         case .exit, .exitCode:
             guard let exitCode = self.text else { return nil}
             return "exit-" + exitCode + "-\(VisualInstructionComponent.scale)-\(hashValue)"
-        case .image, .text:
-            guard let imageURL = imageURL else { return nil }
+        case .image:
+            guard let imageURL = imageURL else { return genericCacheKey }
             return "\(imageURL.absoluteString)-\(VisualInstructionComponent.scale)"
-        case .delimiter:
+        default:
```
bsudekum: It might make more sense to keep `.delimiter` here and remove default so as new cases are added, the compiler will yell at us to handle them.

MapboxNavigation/DayStyle.swift
```diff
@@ -91,6 +91,10 @@ open class DayStyle: Style {
         ExitView.appearance().foregroundColor = .black
         FloatingButton.appearance().backgroundColor = #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
         FloatingButton.appearance().tintColor = tintColor
+        GenericRouteShield.appearance().backgroundColor = .clear
+        GenericRouteShield.appearance().borderWidth = 1.0
+        GenericRouteShield.appearance().cornerRadius = 5.0
+        GenericRouteShield.appearance().foregroundColor = .black
```
bsudekum: How does this look for the night style?

MapboxNavigation/VisualInstructionComponent.swift
bsudekum: A good test case might be, two different steps where the number is given but no text associated with it? Maybe it can't happen but worth checking out.

MapboxNavigation/DayStyle.swift
JThramer: Pretty good, all we end up having to do is needing to change the `foregroundColor`: [Line](https://github.com/mapbox/mapbox-navigation-ios/pull/1417/files/e578eb00dfe627348cb826502f9b98885a6e7260#diff-f33b72188b8d837da672df7b12e4b6c6R200) 

Example

bsudekum: Oh i missed that, 👍

MapboxNavigation/GenericRouteShield.swift
```diff
@@ -0,0 +1,85 @@
+import Foundation
+import UIKit
+
+public class GenericRouteShield: StylableView {
```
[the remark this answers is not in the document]
JThramer: Good point. We should also be hashing off the `foregroundColor` of the `GenericRouteShield`.

MapboxNavigation/VisualInstructionComponent.swift
JThramer: @bsudekum The `InstructionPresenter` doesn't try to render Generic Shields with no text, as they would be an empty box.

PR #1425, opened 2018-05-11
Use current step coordinates when user is traveling slow

bsudekum:
_Addresses https://github.com/mapbox/mapbox-navigation-ios/issues/1078_

This is mostly exasperated when the user is stopped/inching along. This PR changes the coordinates we're using to find the closest snapped location. If the user is moving slowly ( <1 m/s), then only use the current step coordinates to calculate the closest snapped location.

We need to consider the upcoming step because when traveling quickly, there would be an ugly jump from one step to the next if we did not.

/cc @mapbox/navigation-ios


bsudekum:
Noting, this performed well this weekend in a few tests.

Issue #1371, opened 2018-05-02
First step on upcoming leg blank


bsudekum:
![simulator screen shot - iphone x - 2018-05-02 at 14 08 32](https://user-images.githubusercontent.com/1058624/39549369-751a7736-4e12-11e8-9fd0-756c65fefbda.png)

/cc @frederoni

PR #1427, opened 2018-05-11
Fix steps view list

bsudekum:
Fixes: https://github.com/mapbox/mapbox-navigation-ios/issues/1371

This also simplifies the logic.

- [x] figure out why after arriving at a waypoint, the arrival step then disappears. 

/cc @frederoni


MapboxNavigation/StepsViewController.swift
```diff
@@ -48,27 +48,35 @@ open class StepsViewController: UIViewController {
...
+        if !section.isEmpty {
+            sections.append(section)
+        }
+        
+        // Include all steps on any future legs
+        if !routeProgress.isFinalLeg {
+            routeProgress.route.legs.suffix(from: routeProgress.legIndex + 1).forEach {
+                var steps = $0.steps
+                // Don't include the last step, it includes nothing
+                steps.removeLast()
```
frederoni: nit: `popLast()` is safer since it doesn't crash on an empty array.

frederoni:
Looks a lot better. I think
> figure out why after arriving at a waypoint, the arrival step then disappears.

is fixed now as of https://github.com/mapbox/mapbox-navigation-ios/pull/1427/commits/7db80e1811aecb15890d5766f322bd6d0b165171.

Issue #1259, opened 2018-03-26
End of route view controller placement


bsudekum:
![image](https://user-images.githubusercontent.com/1058624/37921659-4b4eb9d0-30df-11e8-8c9c-094008167b3b.png)

@JThramer having a hard time reproducing, but I think this can happen if the route is finished in the background. Seen it ~3 times recently.

bsudekum:
Had this occur again 
![image](https://user-images.githubusercontent.com/1058624/39441508-0a557924-4c63-11e8-9dbe-89ef89614f53.png)

bsudekum:
Had this happen again, really unsure how to reproduce:

![image](https://user-images.githubusercontent.com/1058624/40011592-2128bb16-575d-11e8-9c22-446c0eaa27a4.png)

![image](https://user-images.githubusercontent.com/1058624/40011599-2649aac4-575d-11e8-9bc9-ede5f774ef76.png)

JThramer:
oof, what the heck?

bsudekum:
This looks like the culprit:

https://github.com/mapbox/mapbox-navigation-ios/blob/0c0e54e4bd93d4c30f5e2f0118139b7571852a19/MapboxNavigation/RouteMapViewController.swift#L940

You can reproduce this issue by starting a route, open a new app, open the keyboard, go back into navigation, end the route.

PR #1431, opened 2018-05-14
Do not change EORVC constraints if view is nil

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1259

This keyboard logic is specific to the end of route view controller, specifically when the user touches the info field. If the view is nil (and thus not shown), there is no need to run this logic.

/cc @JThramer

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -915,7 +915,7 @@ extension RouteMapViewController {
         NotificationCenter.default.removeObserver(self, name: .UIKeyboardWillHide, object: nil)
     }
     @objc fileprivate func keyboardWillShow(notification: NSNotification) {
-        guard let _ = navigationView.endOfRouteView else { return }
+        guard navigationView.endOfRouteView != nil else { return }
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -915,6 +915,7 @@ extension RouteMapViewController {
         NotificationCenter.default.removeObserver(self, name: .UIKeyboardWillHide, object: nil)
     }
     @objc fileprivate func keyboardWillShow(notification: NSNotification) {
+        guard let _ = navigationView.endOfRouteView else { return }
```
JThramer: `guard navigationView.endOfRouteView != nil else { return }`

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -932,13 +933,14 @@ extension RouteMapViewController {
     }
     
     @objc fileprivate func keyboardWillHide(notification: NSNotification) {
+        guard let _ = navigationView.endOfRouteView else { return }
```
JThramer: here too

PR #1436, opened 2018-05-16
Update feedback icons and tweak FeedbackViewController in landscape.

### LINKED ISSUE #1426: Broken Feedback UI in landscape mode
Feedback UI:

simulator_screen_shot_-_iphone_x_-_2018-05-11_at_13 13 34

Should become:
simulator screen shot - iphone x - 2018-05-11 at 13 13 34

If rotating the end of route UI, it becomes:
simulator screen shot - iphone x - 2018-05-11 at 13 21 11

frederoni:
Fixes #1426

- Added new vector based feedback icons
- Display 6 items on one row in landscape
- Display 3 items per row in portrait

Co-authored-by: @vincethecoder 

cc @bsudekum @JThramer


vincethecoder:
UPDATES: The issue identified above occurs in iPhone X. 
/cc @frederoni

vincethecoder:
@bsudekum the `No progress bar` defect has been fixed now in the recent commit.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -181,6 +189,7 @@ extension FeedbackViewController: UICollectionViewDataSource {
     
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
         abortAutodismiss()
+        progressBar.progress = 1
```
bsudekum: Minor enhancement, it might make more sense to move this into `dismissFeedback()`.

vincethecoder: @bsudekum so are you saying, let invoke `dismissFeedback()` here instead and move `line # 192` into the `dismssFeedback()` function, because it already calls `abortAutodismiss()` there. https://github.com/mapbox/mapbox-navigation-ios/pull/1436/files#diff-a25e8b3bfe9ceec368b83a004a6d1708R129

bsudekum: I guess I'm saying move it to where the view is cleaned up.

vincethecoder: @bsudekum yeah, I concur. A slight concern about intrusive changes. But, @frederoni I'm curious to know if you could suggest a change which will still maintain the current stability of our SDK. Otherwise, this should actually suffice without any side-effects.

frederoni: As long as the collection view doesn't bounce and the scroll view’s (or collection view in this case) content size matches the bounds, so that isn't scrollable. Also, add a note why we do this b/c it's not an obvious workaround.

PR #1437, opened 2018-05-16
Don't hold onto value from settings observe

bsudekum:
While looking into if we had any memory leaks ([steps I followed](http://blog.mallow-tech.com/2017/12/ways-to-detect-the-memory-leaks-in-swift/)), I found a leak in `RouteVoiceController.resumeNotifications()`.

Unsure why, but holding onto the output from these observe functions seems to be leaking:

image

/cc @mapbox/navigation-ios


bsudekum:
@frederoni this was actually done on device.

PR #1439, opened 2018-05-17
Test for leaks

frederoni:
WIP

Brought in [Quick](https://github.com/Quick/Quick) and [Nimble](https://github.com/Quick/Nimble) which is inspired by [Cedar](https://github.com/cedarbdd/cedar) (@akitchen 👀 ) to check for leaks.

LeakTest is borrowed from [specleaks](https://github.com/leandromperez/specleaks) but it doesn't support Carthage and it doesn't build out-of-the-box.

cc @bsudekum

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,1 +1,3 @@
 github "uber/ios-snapshot-test-case" ~> 3.0.0
+github "Quick/Quick"
+github "Quick/Nimble"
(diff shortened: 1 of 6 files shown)


bsudekum:
Awesome, do tests fail without https://github.com/mapbox/mapbox-navigation-ios/pull/1439/commits/d1c157e5727737e1d18bb7475fd690a5350b54f4?

PR #1441, opened 2018-05-17
Add failing test for tunnel + style refreshes

bsudekum:
In the case where a developer has all tunnel flags turned on, we're force refreshing the map style on every update which causes the map style to flash on and off:

https://github.com/mapbox/mapbox-navigation-ios/blob/026fddd1bc07080f33d0cf3f5e7df3b41a04781a/MapboxNavigation/NavigationViewController.swift#L431-L437

- [x] add failing test
- [x] actually fix the code

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/StyleManager.swift
+++ b/MapboxNavigation/StyleManager.swift
@@ -115,7 +115,7 @@ open class StyleManager: NSObject {
     func applyStyle() {
         guard let location = delegate?.locationFor(styleManager: self) else {
             // We can't calculate sunset or sunrise w/o a location so just apply the first style
-            if let style = styles.first {
+            if let style = styles.first, currentStyleType != style.styleType {
                 currentStyleType = style.styleType
                 style.apply()
                 delegate?.styleManager?(self, didApply: style)
@@ -125,7 +125,7 @@ open class StyleManager: NSObject {
         
         // Single style usage
         guard styles.count > 1 else {
-            if let style = styles.first {
+            if let style = styles.first, currentStyleType != style.styleType {
                 currentStyleType = style.styleType
                 style.apply()
                 delegate?.styleManager?(self, didApply: style)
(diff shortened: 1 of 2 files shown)


MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -72,6 +72,21 @@ class NavigationViewControllerTests: XCTestCase {
         XCTAssertFalse(wayNameView.isHidden, "WayNameView should be visible.")
     }
     
+    var updatedStyleNumberOfTimes = 0
+    func testNavigationShouldNotCallStyleManagerDidRefreshAppearanceMoreThanOnce() {
+        let navigationViewController = NavigationViewController(for: initialRoute, styles: [DayStyle()])
```
frederoni: Nice catch.
Test using two styles, we guard against one style in `resetTimeOfDayTimer()` and the same should be true for `applyStyle`. There’s no reason why we should forcefully refresh appearance when only one style has been provided.

bsudekum: Added test for two styles in https://github.com/mapbox/mapbox-navigation-ios/pull/1441/commits/b0d114f56d304b1f27d6f01bca0fdb1dd3bc3098

bsudekum:
@mapbox/navigation-ios this is ready for review.

MapboxNavigation/StyleManager.swift
```diff
@@ -116,6 +116,7 @@ open class StyleManager: NSObject {
         guard let location = delegate?.locationFor(styleManager: self) else {
             // We can't calculate sunset or sunrise w/o a location so just apply the first style
             if let style = styles.first {
+                currentStyleType = style.styleType
```
vincethecoder: @bsudekum you also check: `currentStyleType != styleType` before updating the `currentStyleType` and invoking the `style .apply()` and `delegate?.styleManager?(self, didApply: style)`.

PR #1446, opened 2018-05-21
Use current step for snapping course

bsudekum:
Building on https://github.com/mapbox/mapbox-navigation-ios/pull/1425, https://github.com/mapbox/mapbox-navigation-ios/pull/1425 was a good addition, but does not seem to catch all cases. There still seems to be situations where the user puck can end up snapping the course in between steps.

This change will snap the puck to the current step coordinates when the user's course is dissimilar from the heading on their upcoming step.

The one place where this could fall apart is when moving very fast through a sharp/windy maneuver. One that comes to mind is a looping on ramp onto a motorway- your course may not be similar to the final heading until the very last second. Still need to test this a bit.

/cc @1ec5


bsudekum:
Tested this a few times yesterday, felt very good. Definitely a little more robotic and slower to complete a 90 degree turn though.

PR #1447, opened 2018-05-22
Do not force refresh style if we don't have the style type

bsudekum:
If tunnel flags are enabled, and we detect that the style needs to change for a given location, we should not try and force update the style if we only have one style or if we don't have that style type.

Example case:
1. inits with tunnel flags and a single style, say night for example.
1. Start in a location where it's day time, the opposite of the style we have.
1. We should not emit a force style update because we do not have a style to update to.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigationTests/NavigationViewControllerTests.swift
+++ b/MapboxNavigationTests/NavigationViewControllerTests.swift
@@ -91,6 +91,24 @@ class NavigationViewControllerTests: XCTestCase {
         updatedStyleNumberOfTimes = 0
     }
     
+    // If tunnel flags are enabled and we need to switch styles, we should not force refresh the map style because we have only 1 style.
+    func testNavigationShouldNotCallStyleManagerDidRefreshAppearanceWhenOnlyOneStyle() {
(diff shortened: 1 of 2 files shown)


MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -91,6 +91,24 @@ class NavigationViewControllerTests: XCTestCase {
...
+    func testNavigationShouldNotCallStyleManagerDidRefreshAppearanceWhenOnlyOneStyle() {
+        let navigationViewController = NavigationViewController(for: initialRoute, styles: [NightStyle()])
+        navigationViewController.usesNightStyleInsideTunnels = true
+        navigationViewController.routeController.tunnelSimulationEnabled = true
+        let routeController = navigationViewController.routeController!
+        navigationViewController.styleManager.delegate = self
+        
+        let someLocation = dependencies.poi.first!
+        
+        routeController.locationManager(routeController.locationManager, didUpdateLocations: [someLocation])
```
bsudekum: We've started navigation with a night style, but these locations warrant a day style.

PR #1448, opened 2018-05-22
Updating destination waypoint when route changes

### LINKED ISSUE #1326: Destination annotation fails to update when route changes
**Mapbox Navigation SDK version: 0.16.2**

### Steps to trigger behavior

 1. Create a new view controller that displays a navigation route between a user's location and a fixed destination.
 2. Add a `UIButton` that triggers an `IBAction` that calculates a new route to a new destination.
 [_Full minimal test case **here.**_](https://github.com/mapbox/navigation-ios-examples/blob/nadia-156562/Navigation-Examples/Examples/Basic.swift#L67-L69)
 3. Observe issue: the destination annotation fails to move to the new destination, even after the route line has updated.

### Expected behavior
- The destination annotation should update to the new destination if the route is changed.
### Actual behavior
- The destination annotation does not update to the new destination if the route is changed, as shown below:

navbug

JThramer:
Fixes #1326. Issue was caused by a missing `showWaypoints(route:)` call.

It's worth calling out that the order-of-operations for the route that the NVC gets initialized with is _really_ obscure and brittle. For instance, unless I manually do `navigationViewController.route = initialRoute` _after_ I wait for the styles to load, I cannot find any annotations on the map, which doesn't seem like reasonable behavior. I thought it out of scope for this issue, but I'll cut a ticket for this as I close this PR out.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigationTests/NavigationViewControllerTests.swift
+++ b/MapboxNavigationTests/NavigationViewControllerTests.swift
@@ -164,7 +164,7 @@ class NavigationViewControllerTests: XCTestCase {
     
     func testDestinationAnnotationUpdatesUponReroute() {
         let styleLoaded = XCTestExpectation(description: "Style Loaded")
-        let navigationViewController = NavigationViewControllerTestable(for: initialRoute, styles: [DayStyle()], styleLoaded: styleLoaded)
+        let navigationViewController = NavigationViewControllerTestable(for: initialRoute, styles: [TestableDayStyle()], styleLoaded: styleLoaded)
         
         //wait for the style to load -- routes won't show without it.
         wait(for: [styleLoaded], timeout: 5)
@@ -256,5 +256,12 @@ class NavigationViewControllerTestable: NavigationViewController {
     
     required init?(coder aDecoder: NSCoder) {
         fatalError("This initalizer is not supported in this testing subclass.")
+    }
+}
+
+class TestableDayStyle: DayStyle {
+    required init() {
+        super.init()
+        mapStyleURL = Fixture.blankStyle
     }
 }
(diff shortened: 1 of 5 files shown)


bsudekum:
It took me a second to understand the link between notifyDidReroute and why this would fix #1326, but then I realized when route is set:

https://github.com/mapbox/mapbox-navigation-ios/blob/026fddd1bc07080f33d0cf3f5e7df3b41a04781a/MapboxNavigation/NavigationViewController.swift#L216-L227

JThramer:
There was an attempt.

Posterity: Apparently my green lights weren't green enough. Holding off on merging this until we can figure out why my tests were passing locally and yet not on CI.

It's also a bit disturbing that results from local testing and CI are not consistent. We should cut a ticket to investigate this and hopefully eliminate any inconsistancies.

akitchen:
> It's also a bit disturbing that results from local testing and CI are not consistent. We should cut a ticket to investigate this and hopefully eliminate any inconsistancies.

These are new tests I believe, so let's get them working properly before merging this PR.

If I'm not mistaken, this new test is loading a style over the network. That's something we want to avoid -- please load one locally from disk instead. This test fails for me on a freshly initialized simulator if I'm offline.

MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -187,3 +233,35 @@ extension NavigationViewControllerTests {
...
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("This initalizer is not supported in this testing subclass.")
+    }
+}
+
+class TestableDayStyle: DayStyle {
+    required init() {
+        super.init()
+        mapStyleURL = Fixture.blankStyle
```
bsudekum: Sort of unfortunate we had to do this deep but, 👍

PR #1452, opened 2018-05-24
Skip playing sound in unit test

### LINKED ISSUE #1450: Intermittently testVoiceDeinit failures
This test: https://github.com/mapbox/mapbox-navigation-ios/blob/f8264f0e6869e07b95e65c982368f738be38524d/MapboxNavigationTests/MapboxVoiceControllerTests.swift#L45-L58

Intermittently fails without a good message on bitrise. Seeing:
```
Failing tests:
	MapboxVoiceControllerTests.testVoiceDeinit()
** TEST FAILED **
```

/cc @frederoni

frederoni:
Fixes #1450

Playing a sound in the simulator fails intermittently so let's not do that in a unit test.

cc @bsudekum

### DIFF
--- a/MapboxNavigationTests/MapboxVoiceControllerTests.swift
+++ b/MapboxNavigationTests/MapboxVoiceControllerTests.swift
@@ -45,15 +45,8 @@ class MapboxVoiceControllerTests: XCTestCase {
     func testVoiceDeinit() {
         var voiceController: MockMapboxVoiceController? = MockMapboxVoiceController()
         let deinitExpectation = expectation(description: "Voice Controller should deinitialize")
-        voiceController?.deinitExpectation = deinitExpectation
-        let bundle = Bundle(for: MapboxVoiceControllerTests.self)
-        let filePath = bundle.path(forResource: "turn_left", ofType: "data")
-        let filePathURL = URL(fileURLWithPath: filePath! )
-        let data = try! Data(contentsOf: filePathURL)
-        voiceController?.play(data)
+        voiceController!.deinitExpectation = deinitExpectation
         voiceController = nil
-        XCTAssertNil(voiceController, "voiceController should be nil after nilling it.")
-        
         wait(for: [deinitExpectation], timeout: 3)
     }
 }


MapboxNavigationTests/MapboxVoiceControllerTests.swift
```diff
@@ -45,15 +45,8 @@ class MapboxVoiceControllerTests: XCTestCase {
...
         let deinitExpectation = expectation(description: "Voice Controller should deinitialize")
-        voiceController?.deinitExpectation = deinitExpectation
-        let bundle = Bundle(for: MapboxVoiceControllerTests.self)
-        let filePath = bundle.path(forResource: "turn_left", ofType: "data")
-        let filePathURL = URL(fileURLWithPath: filePath! )
-        let data = try! Data(contentsOf: filePathURL)
-        voiceController?.play(data)
+        voiceController!.deinitExpectation = deinitExpectation
         voiceController = nil
-        XCTAssertNil(voiceController, "voiceController should be nil after nilling it.")
```
bsudekum: Is it still possible to keep this assertion?

frederoni: We could but it doesn't validate that there's no leak. voiceController is set to nil on the line before. If it was leaking,  deinit wouldn't get called.

akitchen: right; we need to assign a weak pointer before attempting to deallocate it, similar to https://github.com/mapbox/mapbox-navigation-ios/blob/8e094d08a2bbe05ace870f52006e8d02f48dc228/MapboxNavigationTests/LeakTest.swift#L17 or https://github.com/mapbox/mapbox-gl-native/blob/c5625f988a37b6876c4ef5fc8d9de295ee28ed44/platform/ios/Integration%20Tests/MBGLIntegrationTests.m#L15

akitchen:
This is a good start but futzing with the global audio session and playing audio should be stubbed out in all cases in unit tests.

Was this test here originally to guard against a crash/regression?

Issue #1444, opened 2018-05-20
Improvements to user-facing arrival behavior


bsudekum:
There are a few cases we need to handle here:

* User arrives at a waypoint
* User arrives at destination waypoint

In both cases, we should probably stop:
* Snapping (maybe). The issue I see with snapping is that if you go beyond the destination, it remains snapped to the last coord
* rerouting, depending on whether this is the last waypoint or not

akitchen:
Capturing what we've talking about recently:

By default, we want navigation to end when reaching a destination, however this might not be universally true. An optional NavigationDelegate method could allow customization beyond this default for use cases where navigation shouldn't end.

PR #1454, opened 2018-05-24
Stop rerouting when arriving at waypoint

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1444

This change makes it so the user is no longer rerouted after reaching a waypoint. Also, upon arriving, the user puck becomes unsnapped to give a more precise depiction of where they actually are. If they move beyond the waypoint, it's important we show this.

> Example of arriving and moving beyond waypoint

todo:

- [x] add tests
- [x] figure out if this should only occur for the final waypoint

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -712,7 +712,7 @@ extension RouteController: CLLocationManagerDelegate {
     @objc public func userIsOnRoute(_ location: CLLocation) -> Bool {
         
         // If the user has arrived, do not continue monitor reroutes, step progress, etc
-        guard !routeProgress.currentLegProgress.userHasArrivedAtWaypoint  || !(delegate?.routeController?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ?? true) else {
+        guard !routeProgress.currentLegProgress.userHasArrivedAtWaypoint  && !(delegate?.routeController?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ?? true) else {
             return true
         }
 
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -712,7 +712,7 @@ extension RouteController: CLLocationManagerDelegate {
     @objc public func userIsOnRoute(_ location: CLLocation) -> Bool {
         
         // If the user has arrived, do not continue monitor reroutes, step progress, etc
-        guard !routeProgress.currentLegProgress.userHasArrivedAtWaypoint  && !(delegate?.routeController?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ?? true) else {
+        guard !routeProgress.currentLegProgress.userHasArrivedAtWaypoint && (delegate?.routeController?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ?? true) else {
             return true
         }
 
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -134,6 +134,18 @@ public protocol RouteControllerDelegate: class {
     */
     @objc(routeController:didArriveAtWaypoint:)
     optional func routeController(_ routeController: RouteController, didArriveAt waypoint: Waypoint) -> Bool
+    
+    /**
+     Called when the route controller arrives at a waypoint.
+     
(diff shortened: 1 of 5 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -134,6 +134,18 @@ public protocol RouteControllerDelegate: class {
...
+     Called when the route controller arrives at a waypoint.
+     
+     You can implement this method to allow the route controller to continue check and reroute the user if needed. By default, the user will not be rerouted when arriving at a waypoint.
+     
+     - parameter routeController: The route controller that has arrived at a waypoint.
+     - parameter waypoint: The waypoint that the controller has arrived at.
+     - returns: True to prevent the route controller from checking if the user should be rerouted.
+     */
+    @objc(routeController:shouldPreventReroutesWhenArrivingAtWaypoint:)
+    optional func routeController(_ routeController: RouteController, shouldPreventReroutesWhenArrivingAt waypoint: Waypoint) -> Bool
```
bsudekum: Can anyone think of a use case where a developer would return false here? If so, maybe we do not actually need to implement a delegate. 

Stopping rerouting after arriving seems like a good feature for all.

brsbl:
Does a similar ticket exist for Android?

cc @mapbox/navigation-android

danesfeder:
@brsbl I just checked this on our side.  We remove the off route listener after the arrival event on the last leg, so we won't continuously re-route.  There is something going on with our snapped location though after arrival - we should be pushing raw updates so this looks like a bug - will ticket.

MapboxCoreNavigation/RouteController.swift
bsudekum: Ok, will keep.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -698,6 +710,11 @@ extension RouteController: CLLocationManagerDelegate {
      If the user is not on the route, they should be rerouted.
      */
     @objc public func userIsOnRoute(_ location: CLLocation) -> Bool {
+        
+        // If the user has arrived, do not continue monitor reroutes, step progress, etc
+        guard !routeProgress.currentLegProgress.userHasArrivedAtWaypoint && (delegate?.routeController?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ?? true) else {
```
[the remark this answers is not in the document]
bsudekum: Yes. When the legIndex is incremented, `userHasArrivedAtWaypoint ` returns to false.

MapboxCoreNavigation/RouteController.swift
akitchen: I think we should revisit in the process of implementing a sample ride-hailing driver app.

bsudekum:
@1ec5 @akitchen mind giving this a final 👍 👎 ?

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -281,6 +281,47 @@ class RouteControllerTests: XCTestCase {
...
+        
+        // MARK: When navigation begins with a location update
+        routeController.locationManager(routeController.locationManager, didUpdateLocations: [firstLocation])
+        
+        // MARK: It queues and flushes a Depart event
+        XCTAssertTrue(eventsManagerSpy.hasFlushedEvent(with: MMEEventTypeNavigationDepart))
+        // TODO: should there be a delegate message here as well?
+        
+        // MARK: Update to last step
+        routeController.advanceStepIndex(to: routeController.routeProgress.route.legs.first!.steps.count - 1)
```
akitchen: As we talked about before I'm more interested in doing this kind of "journey test" of delegation and telemetry events using location updates only. I have a commit which adjusts that other test but it occurs to me that I haven't pushed it to a PR yet.

https://github.com/mapbox/mapbox-navigation-ios/commit/6a9abcf967d7637540eb3b184c93c8d04fb2c185

We can take this as-is for now and update it later, since it took a bit of a test refactor to make the intent clear.

akitchen:
I'm not so sure about doing this check on every waypoint vs the final destination. I can see now why it felt weird to you too @bsudekum .  But in a multi-passenger pickup & drop off use case we may find this level of granularity helps our customers build what they need.

bsudekum:
> I'm not so sure about doing this check on every waypoint vs the final destination.

I think for consistency sake, we should do this at every way point. The nice thing about this PR is the default behavior is a good feature for everyone. I still am not sure of use cases where developers will ever turn this off.

Issue #1455, opened 2018-05-24
Improve route overview button


brsbl-zz:
We're implementing a route overview button and animation on Android [here](https://github.com/mapbox/mapbox-navigation-android/pull/967#issuecomment-391836478) that is integrated into the trip tray instead of as a button on the map. This saves some map space and feels more where a user would expect to preview the current route they're on. 

Here's how the animation will look on Android:
![ezgif com-video-to-gif 1](https://user-images.githubusercontent.com/8434572/40433589-5bc09d42-5e7a-11e8-8572-1fee89320762.gif)

And it will use these light/dark icons: [routepreview_icons.zip](https://github.com/mapbox/mapbox-navigation-android/files/2036876/routepreview_icons.zip)

Note, the arrival time should also be moved from its current position to the left near the route's distance remaining to better match Android and make room for this button.

Additionally, while we are making this UI/UX change, it's worth investigating this bug in the current route overview implementation where the full route is obscured in overview mode:
![asdf](https://user-images.githubusercontent.com/19296229/40515696-9bdfd97c-5f62-11e8-988a-a25fa7506270.gif)

bsudekum:
The bug can occur when the route is short:

https://github.com/mapbox/mapbox-navigation-ios/blob/53d5802f19bd4c0e34f0e0b6a2927474f6a0d8e3/MapboxNavigation/NavigationMapView.swift#L1032-L1039

We're setting the camera, but breaking early if the distance is short.

1ec5:
> integrated into the trip tray instead of as a button on the map

In other words, is the idea that we’d repurpose the ✖️ button (which currently ends the trip) so that it would conditionally zoom out to the overview or end the trip? If so, that would quite counter to iOS human interface design principles. At the least, we’d need to change the design from a ✖️ to a textual button that says “End” or “Overview” depending. But there are still two usability issues with this design:

* Overview mode and turn-by-turn mode look more or less identical. It’s too easy for the user to accidentally end their trip – a destructive action – because they didn’t realize they were already in overview mode.
* If the user is in turn-by-turn mode and wants to end the trip, the affordance for doing so (tapping twice on the ✖️) is undiscoverable.

Cjballard:
@1ec5 I think the screenshot @brsbl attached was before Android had implemented the button in the trip tray. Here it is with the route preview icon integrated into the trip tray:

[...]

PR #1456, opened 2018-05-24
Fix setOverheadCameraView for short routes

bsudekum:
Addresses the bug in https://github.com/mapbox/mapbox-navigation-ios/issues/1455

Previously, if the user only had a few meters left on the route, we'd flip up the camera, then break early and never actually set the correct center coordinate.

Now, if we detect there is not much left on the route, we simple flip the camera and set the center coordinates to the user.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -1030,13 +1030,24 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         let line = MGLPolyline(coordinates: slicedLine, count: UInt(slicedLine.count))
         
         tracksUserCourse = false
+        
+        // If the user has a short distance left on the route, prevent the camera from zooming all the way.
+        // `MGLMapView.setVisibleCoordinateBounds(:edgePadding:animated:)` will go beyond what is convient for the driver.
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -1030,14 +1030,25 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         let line = MGLPolyline(coordinates: slicedLine, count: UInt(slicedLine.count))
         
         tracksUserCourse = false
+        
+        // If the user has a short distance left on the route, prevent the camera from zooming all the way.
+        // `MGLMapView.setVisibleCoordinateBounds(:edgePadding:animated:)` will go beyond what is convient for the driver.
```
vincethecoder: >  will go beyond what is convient for the driver.

_did you mean to say_ `convenient to the driver`?

bsudekum: Yeah, it'd be inconvenient for the map to continue zooming past what is comprehensible.

vincethecoder: @bsudekum oh okay. There's a typo `convient for` should `convenient to`

PR #1459, opened 2018-05-29
Exit View Padding

### LINKED ISSUE #1451: Fix perceived visual padding on exit view arrow
Allow the padding is probably (maybe not?) equal on both sides, visually the padding looks uneven on one side depending if its in the main or then banner. This is just how an arrow works when it's at 45 degrees I think.

I think the fix here is to reduce the padding on the left side of the arrow on the top banner and reduce the padding on the right side of the arrow on the then banner.

new

> Added orange lines to show whitespace distance.

/cc @JThramer

JThramer:
simulator screen shot - iphone 8 plus - 2018-05-29 at 11 32 50

Fixes #1451. Issue fixed by tightening up the padding a bit, and varying it by two points, depending on text direction, as `UILabel` seems to add about 2 points worth of padding to the trailing side.

To-Do: 
- [ ] Update Snapshot Tests
* Adding variable padding that brings (naturally-perceived) even distribution of elements in view, dependent on text-direction

/cc @mapbox/navigation-ios


bsudekum:
Logic looks sounds, can okay once fixtures are updated.

Issue #1438, opened 2018-05-17
How to hide ‘Report Problem’ button when user is in TBT navigation


leonyuu:
Hi Mapbox,
‘Report Problem’ button randomly appears at the top middle of the screen when user is in TBT navigation. And in my App, I need to hide this 'Report Problem' button but not found any clue. Please see the attachment below.
Thank you very much :)
![img_0857](https://user-images.githubusercontent.com/4963090/40160234-e2f53afa-59d6-11e8-85e5-2ac7b636c70f.PNG)

PR #1460, opened 2018-05-29
Remove report problem button after rerouting

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1438

I have the feeling the "Report Problem" button that is shown after rerouting is more confusing, loud and distracting for drivers than helpful. It takes the same amount of clicks to give feedback after rerouting anyways. 

This PR simply removes the button. The string remains because it's used in the actual feedback UI.

 

/cc @mapbox/navigation-ios @leonyuu

### DIFF
--- a/MapboxNavigation/NavigationView.swift
+++ b/MapboxNavigation/NavigationView.swift
@@ -43,7 +43,6 @@ open class NavigationView: UIView {
         static let feedbackTopConstraintPadding: CGFloat = 10.0
         static let buttonSize: CGSize = 50.0
         static let buttonSpacing: CGFloat = 8.0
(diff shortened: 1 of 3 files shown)


1ec5:
> The string remains because it's used in the actual feedback UI.

@bsudekum, the string is no longer used and should be removed from the bundle as well. The feedback UI uses a distinct string with the same English value.

https://github.com/mapbox/mapbox-navigation-ios/blob/c7d936db261e8a6fc49c0c095060fb19b371ba34/MapboxNavigation/Resources/Base.lproj/Localizable.strings#L70-L71

PR #1464, opened 2018-06-01
Integrated new directions API fields with the tertiary instruction and lane indications

vincethecoder:
These updates make sure we rely on the new directions `LaneIndicationComponents` class for lane information. This information is usually retrieved from the tertiary instruction -- which can either be a `LaneIndicationComponent` or `VisualInstructionComponent`. The instructions banner and integrated tests were also modified to reflect the new changes. 

TODO:
- [ ] Fix broken tests (Lanes, Instructions Snapshot, and Integration tests)

Relies on https://github.com/mapbox/MapboxDirections.swift/pull/258

/cc @JThramer


MapboxNavigation/InstructionPresenter.swift
```diff
@@ -43,16 +43,18 @@ class InstructionPresenter {
         
         guard !stringFits else { return attributedPairs.attributedStrings }
         
-        let indexedComponents = attributedPairs.components.enumerated().map { IndexedVisualInstructionComponent(component: $1, index: $0) }
-        let filtered = indexedComponents.filter { $0.component.abbreviation != nil }
-        let sorted = filtered.sorted { $0.component.abbreviationPriority < $1.component.abbreviationPriority }
-        for component in sorted {
-            let isFirst = component.index == 0
+        let filteredComponents = attributedPairs.components.compactMap { $0 as? VisualInstructionComponent}
+        let visuals = filteredComponents.enumerated().map { IndexedVisualInstructionComponent(component: $1, index: $0) }
```
frederoni: Can there be anything except `VisualInstructionComponent` in attributed pairs? If that's the case, the instruction indices can be shuffled and end up in the wrong order with this change. Indexing should be done before anything else.

vincethecoder: @frederoni  Nope. There are only `VisualIntructionComponent`s in `IndexedVisualInstructionComponent`. See: https://github.com/mapbox/mapbox-navigation-ios/pull/1464/files#diff-af48d70309ee195e34cc5e6cc6b27caaR300

frederoni: What I meant was that this row: 46
```swift
let filteredComponents = attributedPairs.components.compactMap { $0 as? VisualInstructionComponent}
```
filters out components before we index them.
Further down, where we abbreviate the string, it could end up in the wrong order, no?

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -63,12 +65,12 @@ class InstructionPresenter {
...
     
-    typealias AttributedInstructionComponents = (components: [VisualInstructionComponent], attributedStrings: [NSAttributedString])
+    typealias AttributedInstructionComponents = (components: [MBComponentRepresentable], attributedStrings: [NSAttributedString])
     
     func attributedPairs(for instruction: VisualInstruction, dataSource: DataSource, imageRepository: ImageRepository, onImageDownload: @escaping ImageDownloadCompletion) -> AttributedInstructionComponents {
-        let components = instruction.textComponents
+        let components = instruction.components.compactMap { $0 as? VisualInstructionComponent }
         var strings: [NSAttributedString] = []
-        var processedComponents: [VisualInstructionComponent] = []
+        var processedComponents: [MBComponentRepresentable] = []
```
bsudekum: This can simply be `ComponentRepresentable`.

MapboxNavigation/InstructionPresenter.swift
vincethecoder: Gotchu.

MapboxNavigation/InstructionPresenter.swift
vincethecoder: @bsudekum it's a protocol. Hence the prefix `MB`

MapboxNavigation/InstructionPresenter.swift
vincethecoder: @frederoni so, wouldn't the sorting by abbreviation priority address this concern?
https://github.com/mapbox/mapbox-navigation-ios/pull/1464/files/b8f447cc296ce7cc9866794c62a5183d8990f543#diff-af48d70309ee195e34cc5e6cc6b27caaR49

Or, perhaps what would you suggest will be the ideal approach in this context?

MapboxNavigation/InstructionPresenter.swift
bsudekum: I think the MB should be renamed from the swift name of this protocol: https://github.com/mapbox/MapboxDirections.swift/pull/258/files#diff-664678368965fd0cdce04411baf6644cR7

akitchen:
Please update this with master when you have a chance, in order to pick up & trigger CI changes

vincethecoder:
Closing this PR. A new PR branch with less intrusive modification of the existing logic for generic shields.

PR #1475, opened 2018-06-07
Describe return type fom map/reduce

bsudekum:
As pointed in the WWDC talk "Building Faster in Xcode", relying on the compiler to infer the return type  from a map/filter/compactMap can add up to a slower build time. This PR finds a few places we're creating variables from a map/reduce/compactMap and explicitly names the return type.

Thoughts on adopting this as a best practice going forward? Beyond helping the compiler, it does in fact make it easier to review the code by knowing exactly what is returned. 

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -335,7 +335,7 @@ extension UIDevice {
             var systemInfo = utsname()
             uname(&systemInfo)
             let machineMirror = Mirror(reflecting: systemInfo.machine)
-            let identifier = machineMirror.children.reduce("") { identifier, element in
+            let identifier: String = machineMirror.children.reduce("") { (identifier: String, element: Mirror.Child) in
                 guard let value = element.value as? Int8, value != 0 else { return identifier }
                 return identifier + String(UnicodeScalar(UInt8(value)))
             }
--- a/MapboxCoreNavigation/MMEEventsManager.swift
+++ b/MapboxCoreNavigation/MMEEventsManager.swift
@@ -335,12 +335,10 @@ extension UIDevice {
             var systemInfo = utsname()
             uname(&systemInfo)
             let machineMirror = Mirror(reflecting: systemInfo.machine)
-            let identifier: String = machineMirror.children.reduce("") { (identifier: String, element: Mirror.Child) in
+            return machineMirror.children.reduce("") { (identifier: String, element: Mirror.Child) in
                 guard let value = element.value as? Int8, value != 0 else { return identifier }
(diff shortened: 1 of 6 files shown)


MapboxCoreNavigation/MMEEventsManager.swift
```diff
@@ -335,7 +335,7 @@ extension UIDevice {
             var systemInfo = utsname()
             uname(&systemInfo)
             let machineMirror = Mirror(reflecting: systemInfo.machine)
-            let identifier = machineMirror.children.reduce("") { identifier, element in
+            let identifier: String = machineMirror.children.reduce("") { (identifier: String, element: Mirror.Child) in
```
1ec5: Wouldn’t it be better to declare the return type of the closure instead of the type of the local variable?

bsudekum: Yeah, we could return this variable right away instead.

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -38,12 +38,12 @@ class InstructionPresenter {
         guard let source = self.dataSource else { return [] }
         var attributedPairs = self.attributedPairs(for: instruction, dataSource: source, imageRepository: imageRepository, onImageDownload: completeShieldDownload)
         let availableBounds = source.availableBounds()
-        let totalWidth = attributedPairs.attributedStrings.map { $0.size() }.reduce(.zero, +).width
+        let totalWidth: CGFloat = attributedPairs.attributedStrings.map { $0.size() }.reduce(.zero, +).width
```
1ec5: Pretty sure this change doesn’t improve compile time in any measurable way, because the compiler still has to infer the type of the `map` call in order to reduce it and get the `width` of it.

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -38,12 +38,12 @@ class InstructionPresenter {
...
         let stringFits = totalWidth <= availableBounds.width
         
         guard !stringFits else { return attributedPairs.attributedStrings }
         
-        let indexedComponents = attributedPairs.components.enumerated().map { IndexedVisualInstructionComponent(component: $1, index: $0) }
+        let indexedComponents: [IndexedVisualInstructionComponent] = attributedPairs.components.enumerated().map { IndexedVisualInstructionComponent(component: $1, index: $0) }
```
1ec5: This probably isn’t ambiguous, given that `Array.map` always returns an array, and `IndexedVisualInstructionComponent`’s initializers always resolve to `IndexedVisualInstructionComponent` instances.

Issue #1474, opened 2018-06-07
Global app state manipulation


simonseyer:
**Mapbox Navigation SDK version: Master branch**

In the `RouteController.init` the battery monitoring is enabled (`UIDevice.current.isBatteryMonitoringEnabled = true`) and in the `deinit` it is disabled again (`UIDevice.current.isBatteryMonitoringEnabled = false`). 

This manipulation of a global app state is breaking features of us that rely on battery monitoring. This at least has to be configurable. Even better would be if the configuration would be delegated to the integrating app (`delegate.routeController(_ controller, requiresBatteryMonitoring: true/false)`).

I only read a subset of your code yet, but I also recognized a hard dependency to the `CLLocationManger` and its types. This also causes some inconveniences for us because our location is coming from different sources. My suggestion would be to have a protocol which defines just the requirements you have (a location with lat/long & accuracy and true heading I guess) and provide a default implementation with a `CLLocationManager`. 

In general it would be best (from my point of view) to not rely on any system library and to delegate all the required informations/actions to the app code while still providing default implemenations. This would provide a lot of flexibility (that we more or less need to integrate your SDK into our rather complex app) while still keeping the "works out of the box" experience.

akitchen:
Hello @simonseyer and thank you very much for opening this issue. Your feedback is greatly appreciated!

Regarding the issues you raised, we are aware and in agreement regarding both the issue of manipulating global shared state as well as the location manager coupling. Note that the latter we have also been discussing in our Maps SDK ([issue](https://github.com/mapbox/mapbox-gl-native/issues/11983)) (https://github.com/mapbox/mapbox-gl-native/pull/12013) -- your feedback on this approach would be most welcome as to whether something similar in our Navigation SDK would meet your needs.

Regarding the manipulation of `-[UIDevice isBatteryMonitoringEnabled]`, we have been discussing a solution in https://github.com/mapbox/mapbox-navigation-ios/pull/1476 . We use battery monitoring in aspects of our code, but can introduce delegation to control whether to disable it when finished.

1ec5:
Note that battery monitoring is currently used not only for setting CLLocationManager options, but also for throttling NavigationMapView’s frame rate when the device isn’t plugged in.

simonseyer:
Thanks for the immediate reply. I’m happy that you are sharing my opinion about this topic. I can only look into the mentioned ticket/pr on Monday but will be happy to give feedback then.

[...]

PR #1476, opened 2018-06-07
Allow developers to control whether battery monitoring is enabled

bsudekum:
Addresses part 1 of https://github.com/mapbox/mapbox-navigation-ios/issues/1474

If an app sets `UIDevice.isBatteryMonitoringEnabled` on their own, we should not reset this property when initing/deiniting. 

This new prop makes it so developers can control whether this `UIDevice.isBatteryMonitoringEnabled` is enabled internally.

/cc @mapbox/navigation-ios @simonseyer

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -154,7 +154,7 @@ public protocol RouteControllerDelegate: class {
      Implementing this method will allow developers to change whether battery monitoring is disabled when `RouteController` is deinited.
      
      - parameter routeController: The route controller that will change the state of battery monitoring.
-     - returns: A bool representing the value you would like `UIDevice.isBatteryMonitoringEnabled` set to.
+     - returns: A bool indicating whether to disable battery monitoring when the RouteController is deinited.
      */
     @objc(routeControllerShouldDisableBatteryMonitoring:)
     optional func routeControllerShouldDisableBatteryMonitoring(_ routeController: RouteController) -> Bool
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -203,6 +203,13 @@ open class RouteController: NSObject {
      The flag that indicates that the simulated navigation through tunnel(s) is enabled.
      */
     public var tunnelSimulationEnabled: Bool = true
+    
+    /**
+     By default, `RouteContoller` will enable `UIDevice.isBatteryMonitoringEnabled`.
+     
+     If your app requires custom battery monitor, disable this property.
+     */
+    public var isBatteryMonitoringEnabled = true
 
     var didFindFasterRoute = false
 
@@ -277,9 +284,12 @@ open class RouteController: NSObject {
         self.locationManager = locationManager
         self.locationManager.activityType = route.routeOptions.activityType
         self.eventsManager = eventsManager
-        UIDevice.current.isBatteryMonitoringEnabled = true
 
         super.init()
+        
+        if isBatteryMonitoringEnabled {
+            UIDevice.current.isBatteryMonitoringEnabled = true
+        }
 
         self.locationManager.delegate = self
         resumeNotifications()
@@ -303,7 +313,10 @@ open class RouteController: NSObject {
         sendCancelEvent(rating: endOfRouteStarRating, comment: endOfRouteComment)
         sendOutstandingFeedbackEvents(forceAll: true)
         suspendNotifications()
-        UIDevice.current.isBatteryMonitoringEnabled = false
+        
+        if isBatteryMonitoringEnabled {
+            UIDevice.current.isBatteryMonitoringEnabled = false
+        }
     }
 
     func startEvents(accessToken: String?) {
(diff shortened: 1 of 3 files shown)


akitchen:
Please don’t merge this as-is, as it hasn’t been discussed, groomed or prioritized yet.

This PR might not adequately address the issue - instead of introducing more state on RouteController, this should probably be delegated as requested.

akitchen:
My thoughts here: If we require battery monitoring for our own features, we should optionally delegate the decision to disable it when finished.

bsudekum:
@akitchen how does something that looks like:

```swift
optional func routeController(_ routeController: RouteController, willChangeBatteryMonitoringState from: Bool, to: Bool) -> Bool
```

If implemented, we will adhere to whatever value the developer returns.

simonseyer:
Don’t know what the ‘from’ value could be useful for (would always be ‘!to’, otherwise there would be no need to call the delegate). But in general looks good 👍

bsudekum:
@simonseyer @akitchen moved to a delegate in https://github.com/mapbox/mapbox-navigation-ios/pull/1476/commits/c66becd30a2413e3263a24d149aa4dbd17173865.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -277,9 +290,11 @@ open class RouteController: NSObject {
         self.locationManager = locationManager
         self.locationManager.activityType = route.routeOptions.activityType
         self.eventsManager = eventsManager
-        UIDevice.current.isBatteryMonitoringEnabled = true
 
         super.init()
+        
+        let monitorBatteryValue = delegate?.routeController?(self, willChangeBatteryMonitoringState: true) ?? true
```
frederoni: `delegate` is always nil here.

bsudekum: Moved in https://github.com/mapbox/mapbox-navigation-ios/pull/1476/commits/573d582c1be10a9f7e8a296f96147229033bcd5c

bsudekum:
@akitchen updated and exposed the delegate only when deinit is called now.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -146,6 +146,18 @@ public protocol RouteControllerDelegate: class {
...
+    /**
+     Called when the route controller will disable battery monitoring.
+     
+     Implementing this method will allow developers to change whether battery monitoring is disabled when `RouteController` is deinited.
+     
+     - parameter routeController: The route controller that will change the state of battery monitoring.
+     - returns: A bool representing the value you would like `UIDevice.isBatteryMonitoringEnabled` set to.
+     */
+    @objc(routeControllerWillDisableBatteryMonitoring:)
+    optional func routeControllerWillDisableBatteryMonitoring(_ routeController: RouteController) -> Bool
```
akitchen: The logic here might be confusing. Typically, a delegate method with "will" or "should" in the name returns a boolean to indicate whether to proceed with the described action.

So, a `false` response to `routeControllerWillDisableBatteryMonitoring()` would mean *don't* disable it. This might be more clear if we change "will" to "should", but in either case I think the logic should be the other way around from how it is currently implemented here.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -316,7 +316,7 @@ open class RouteController: NSObject {
         sendOutstandingFeedbackEvents(forceAll: true)
         suspendNotifications()
         
-        let monitorBatteryValue = delegate?.routeControllerWillDisableBatteryMonitoring?(self) ?? false
+        let monitorBatteryValue = delegate?.routeControllerShouldDisableBatteryMonitoring?(self) ?? false
```
simonseyer: Now the semantics are still wrong: If I return `false` because it shouldn't be disabled, it will actually be disabled and the other way around.

akitchen: Yes; it's not correct to pass this value into the next line. The heuristic should look like:

 - If the delegate doesn't implement `routeControllerShouldDisableBatteryMonitoring()`, go ahead and disable it.
 - If the delegate implements `routeControllerShouldDisableBatteryMonitoring()` and returns true, disable it.
 - If the delegate implements `routeControllerShouldDisableBatteryMonitoring()` and returns false, don't disable it.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -146,6 +146,18 @@ public protocol RouteControllerDelegate: class {
...
     optional func routeController(_ routeController: RouteController, shouldPreventReroutesWhenArrivingAt waypoint: Waypoint) -> Bool
+    
+    
+    /**
+     Called when the route controller will disable battery monitoring.
+     
+     Implementing this method will allow developers to change whether battery monitoring is disabled when `RouteController` is deinited.
+     
+     - parameter routeController: The route controller that will change the state of battery monitoring.
+     - returns: A bool representing the value you would like `UIDevice.isBatteryMonitoringEnabled` set to.
```
akitchen: The return value here indicates whether to disable battery monitoring when the RouteController is deinited, not the value to set it to. If this method is not implemented, battery monitoring will be disabled when navigation ends.

With that doc change (or something close to it) I think we're good to merge this.

akitchen:
LGTM - thanks for the doc updates @bsudekum

Issue #1473, opened 2018-06-06
InstructionsBannerView includes steps list indicator by default


bsudekum:
If a developer adds a `InstructionsBannerView` class to their custom UI, the steps list indicator is not only visible, but can also be interacted with even though it does nothing. 

We should:

* Make it easier for the steps list view to be visible in a custom UI.
* If the developer has not added the steps list view, the steps list indicator should not be visible.

/cc @mapbox/navigation-ios

PR #1477, opened 2018-06-07
Do not show steps indicator if delegate is not implemented

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1473

If the developer has not implemented the ability to show the steps list view, we should not show the steps list indicator.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/InstructionsBannerView.swift
+++ b/MapboxNavigation/InstructionsBannerView.swift
@@ -28,7 +28,11 @@ open class BaseInstructionsBannerView: UIControl {
     weak var _separatorView: UIView!
     weak var separatorView: SeparatorView!
     weak var stepListIndicatorView: StepListIndicatorView!
-    weak var delegate: InstructionsBannerViewDelegate?
+    weak var delegate: InstructionsBannerViewDelegate? {
+        didSet {
+            stepListIndicatorView.isHidden = false
+        }
+    }
     
     var centerYConstraints = [NSLayoutConstraint]()
     var baselineConstraints = [NSLayoutConstraint]()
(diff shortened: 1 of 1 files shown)


MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -65,18 +65,21 @@ open class BaseInstructionsBannerView: UIControl {
...
-        stepListIndicatorView.isHidden = !stepListIndicatorView.isHidden
-        delegate?.didTapInstructionsBanner?(self)
+        if let delegate = delegate {
+            stepListIndicatorView.isHidden = !stepListIndicatorView.isHidden
+            delegate.didTapInstructionsBanner?(self)
+        }
     }
     
     func set(_ instruction: VisualInstructionBanner?) {
+        stepListIndicatorView.isHidden = delegate == nil
```
JThramer: Nit: Maybe some parenthesis around (delegate == nil) might make this a bit more readable? We should come up with a stylistic rule of thumb for this.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -65,18 +65,21 @@ open class BaseInstructionsBannerView: UIControl {
     }
     
     @objc func draggedInstructionsBanner(_ sender: Any) {
-        if let gestureRecognizer = sender as? UIPanGestureRecognizer, gestureRecognizer.state == .ended {
+        if let gestureRecognizer = sender as? UIPanGestureRecognizer, gestureRecognizer.state == .ended, let delegate = delegate {
```
1ec5: This prevents the two lines below from running if the delegate isn’t set. Therefore, if the delegate isn’t set, then the indicator view doesn’t get hidden.

bsudekum: Yeah I guess that's the point though. We should only toggle the indicator if the developer implements the delegate methods.

PR #1478, opened 2018-06-07
Fix Xcode 10 build error

bsudekum:
Without this change, Xcode 10 fails with the error:

```

Showing Recent Messages
:-1: Multiple commands produce '/Users/bobby/Library/Developer/Xcode/DerivedData/Voyage-bxmobzxqvtumspcofllpykcoymjs/Build/Products/Debug-iphonesimulator/MapboxNavigation/MapboxNavigation.framework/Info.plist':
1) Target 'MapboxNavigation' has copy command from '/Users/bobby/Documents/mapbox/mapbox-navigation-ios/MapboxNavigation/Info.plist' to '/Users/bobby/Library/Developer/Xcode/DerivedData/Voyage-bxmobzxqvtumspcofllpykcoymjs/Build/Products/Debug-iphonesimulator/MapboxNavigation/MapboxNavigation.framework/Info.plist'
2) Target 'MapboxNavigation' has process command with input '/Users/bobby/Documents/mapbox/voyage-ios/Pods/Target Support Files/MapboxNavigation/Info.plist'
```

Cocoapods seems to be including erroneous files like this Plist file. To get around this, we need to explicitly outline the source files.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation.podspec
+++ b/MapboxNavigation.podspec
@@ -32,7 +32,7 @@ Pod::Spec.new do |s|
 
   # ――― Source Code ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
-  s.source_files = ["MapboxNavigation/*", "MapboxCoreNavigation/{Date,Sequence,String}.swift"]
+  s.source_files = ["MapboxNavigation", "MapboxCoreNavigation/{Date,Sequence,String}.swift"]
 
   # ――― Resources ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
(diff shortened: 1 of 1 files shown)


paul019:
@bsudekum @1ec5 I'm a bit confused: On the `MapboxNavigation` site on cocoapods.org is stated, that the library only supports Xcode 9. But here a Xcode 10 build error is fixed.

However, when I try to install `MapboxNavigation` through CocoaPods using Xcode 10 I get various build errors from Xcode, e.g.: `'MapboxDirections/MapboxDirections.h' file not found`. Does `MapboxNavigation` support Xcode 10 or does it only support Xcode 9?

Issue #1480, opened 2018-06-08
NavigationLocationManager shadows CLLocationManager.desiredAccuracy


1ec5:
This SDK offers no way to set the built-in location manager’s `desiredAccuracy` to anything other than the iOS default, `best`. The option that ostensibly lets the user configure this setting has no effect. We should remove the code that implements this option.

As of #1206, `NavigationLocationManager.desiredAccuracy` overrides `CLLocationManager.desiredAccuracy` with a custom getter and setter that get and set the internal `NavigationLocationManager.accuracyOverride` property. The getter falls back to `CLLocationManager.desiredAccuracy`.

https://github.com/mapbox/mapbox-navigation-ios/blob/10195cf07e66823a58499d0185019eeb417e7517/MapboxCoreNavigation/NavigationLocationManager.swift#L31
https://github.com/mapbox/mapbox-navigation-ios/blob/10195cf07e66823a58499d0185019eeb417e7517/MapboxCoreNavigation/NavigationLocationManager.swift#L35

Effectively, `NavigationLocationManager.accuracyOverride` shadows `CLLocationManager.desiredAccuracy`. So if the developer attempts to set `desiredAccuracy`, all they actually set is `accuracyOverride`, not the actual Core Location behavior; after that point, `desiredAccuracy` lies about its actual value.

The following line in the getter originated in the class’s initializer. It used to set the `desiredAccuracy` to `best` or `bestForNavigation` depending on whether the device is plugged in. Now it too lies: if the device is plugged in, the location manager actually has a desired accuracy of `best`, but this getter returns `bestForNavigation`.

https://github.com/mapbox/mapbox-navigation-ios/blob/10195cf07e66823a58499d0185019eeb417e7517/MapboxCoreNavigation/NavigationLocationManager.swift#L32

Telemetry events merely report whatever `NavigationLocationManager.desiredAccuracy`’s getter returns, so if this property were set to any other value, that value is what the events would contain.

https://github.com/mapbox/mapbox-navigation-ios/blob/c45d93837c6a07dc5f3ea5a829f3cd50ca7f7e46/MapboxCoreNavigation/MMEEventsManager.swift#L100

[...]

Issue #1479, opened 2018-06-08
automaticallyUpdatesDesiredAccuracy is nonfunctional


1ec5:
`NavigationLocationManager.automaticallyUpdatesDesiredAccuracy` is marked as deprecated, but it’s already unused. It should be removed. In fact, it should’ve been marked unavailable in the first place if there was no way to automatically update `desiredAccuracy` as a stopgap.

https://github.com/mapbox/mapbox-navigation-ios/blob/10195cf07e66823a58499d0185019eeb417e7517/MapboxCoreNavigation/NavigationLocationManager.swift#L20-L25

/cc @bsudekum

bsudekum:
fixed via https://github.com/mapbox/mapbox-navigation-ios/pull/1481

PR #1481, opened 2018-06-08
Fix desired accuracy

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1480 & https://github.com/mapbox/mapbox-navigation-ios/issues/1479

This changes a few things:
* Removes unused `automaticallyUpdatesDesiredAccuracy `
* Removes `accuracyOverride` and moves to asking the developer to sublcass `NavigationLocationManager` if a developer needs to change how it functions.
* Properly sets the accuracy the `desiredAccuracy ` based on whether the phone is plugged in.

A few problems:
* `desiredAccuracy` does not change when the phone is plugged in/unplugged
* As pointed out in https://github.com/mapbox/mapbox-navigation-ios/issues/1480, we were never actually setting the correct (or any) `desiredAccuracy`. If we want, for now, we can just simple set it `kCLLocationAccuracyBest` to be safe.

/cc @1ec5


MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -9,38 +9,22 @@ import UIKit
 #endif
 
 /**
- `NavigationLocationManager` is the base location manager which handles
- permissions and background modes.
+ `NavigationLocationManager` is the base location manager which handles permissions and background modes.
+ 
+ If it's necessary to change the default behavior in `NavigationLocationManager`, you can subclass this class.
```
1ec5: In fact, the developer doesn’t even need to subclass this class to modify the desired accuracy: all they have to do is set `navigationViewController.locationManager.desiredAccuracy`.

MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -9,38 +9,22 @@ import UIKit
...
-            accuracyOverride = newValue
-        }
-    }
-    
     override public init() {
         super.init()
         
         requestWhenInUseAuthorization()
         
+        desiredAccuracy = UIDevice.current.isPluggedIn ? kCLLocationAccuracyBestForNavigation : kCLLocationAccuracyBest
```
1ec5: What does this do if battery monitoring isn’t enabled? Does it default to `kCLLocationAccuracyBest`?

1ec5: This change assumes `kCLLocationAccuracyBestForNavigation` offers enough of an improvement over `kCLLocationAccuracyBest` that it’s worth enabling, even if there’s increased battery usage (as implied by the `isPluggedIn` check). This change reintroduces the use of `kCLLocationAccuracyBestForNavigation`, which we haven’t been using since #1206. So we need to test this change thoroughly before landing it.

bsudekum: Correct, `isPluggedIn` would return false.

bsudekum: I'm okay with only using `kCLLocationAccuracyBest `

bsudekum:
@1ec5 updated

MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -9,38 +9,20 @@ import UIKit
...
-            accuracyOverride = newValue
-        }
-    }
-    
     override public init() {
         super.init()
         
         requestWhenInUseAuthorization()
         
+        desiredAccuracy = kCLLocationAccuracyBest
```
1ec5: This is the default value on iPhones, right? Do we (or can we) force the value to be `kCLLocationAccuracyBest` on iPads?

bsudekum: I think it would make me sleep better at night knowing we have something set here.

1ec5: Instead of a no-op in the code, let’s add a test that verifies the default value.

bsudekum:
@1ec5 yeah, `kCLLocationAccuracyBest ` is the default on iPads too.

1ec5:
Might want to add a changelog message about this.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Moved `RouteController.tunnelSimulationEnabled` to `TunnelIntersectionManager.tunnelSimulationEnabled`. ([#1489](https://github.com/mapbox/mapbox-navigation-ios/pull/1489))
 * Deprecated `NavigationViewController.usesNightStyleInsideTunnels`. Style switching is enabled as a side effect of `TunnelIntersectionManager.tunnelSimulationEnabled`, which is set to `true` by default. ([#1489](https://github.com/mapbox/mapbox-navigation-ios/pull/1489))
 * Added `RouteControllerDelegate.routeControllerWillDisableBatteryMonitoring(_:)` which allows developers control whether battery monitoring is disabled when `RouteController.deinit()` is called.
+* Setting `NavigationLocationManager` to a custom value now works again. [#1481](https://github.com/mapbox/mapbox-navigation-ios/pull/1481)
```
1ec5: Should be `NavigationLocationManager.desiredAccuracy`. Fixed in 4784fdaf1d04a5f616fbf3722a5bd73c1069803b.

Issue #1484, opened 2018-06-08
Date.ISO8601 is expensive


bsudekum:
![image](https://user-images.githubusercontent.com/1058624/41183856-b468e3be-6b31-11e8-828a-1580665ff01f.png)

Specifically, this line:

https://github.com/mapbox/mapbox-navigation-ios/blob/e1742ee166a71d469f3c10d0122147bb300746e4/MapboxCoreNavigation/Date.swift#L12

We should compute this formatter once and use it as needed.

PR #1485, opened 2018-06-08
Init date formatter once

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1484

This moves the ISO8061 date formatter to a global var instead of recreating every time it is needed.

/cc @mapbox/navigation-ios


MapboxCoreNavigation/Date.swift
```diff
@@ -1,17 +1,15 @@
...
+}()
+
 extension Date {
     var ISO8601: String {
-        return DateFormatter.ISO8601.string(from: self)
-    }
-}
-
-extension DateFormatter {
-    class var ISO8601: DateFormatter {
```
frederoni: Keeping it in the extension on DateFormatter would be slightly more idiomatic and Swifty.

```swift
static let ISO8601: DateFormatter = { ... }
```

Issue #1486, opened 2018-06-08
Don't call `StyleManager.timeOfDayChanged()` on every progress update


bsudekum:
This is a rather expensive function: ![image](https://user-images.githubusercontent.com/1058624/41184773-62ebed64-6b37-11e8-820d-a63ae1930c94.png)

This called on every progress update here:

https://github.com/mapbox/mapbox-navigation-ios/blob/30a35d616737c1f5daa9dfd9cbe0173e191b1ad3/MapboxNavigation/NavigationViewController.swift#L431-L437

bsudekum:
Done in https://github.com/mapbox/mapbox-navigation-ios/pull/1489

PR #1489, opened 2018-06-10
Use tunnel manager delegate for changing styles

bsudekum:
As pointed out in https://github.com/mapbox/mapbox-navigation-ios/issues/1486, calling `styleManager.timeOfDayChanged()` is expensive because we have to re-init `Solar` on every call to check. 

Now with this change, we're only changing the style from/to day/night when the tunnel manager tells us we have entered/exited a tunnel.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -197,7 +197,7 @@ open class RouteController: NSObject {
      
      Will only be enabled if `tunnelSimulationEnabled` is true.
      */
-    public var tunnelIntersectionManager: TunnelIntersectionManager?
+    public var tunnelIntersectionManager: TunnelIntersectionManager = TunnelIntersectionManager()
     
     /**
      The flag that indicates that the simulated navigation through tunnel(s) is enabled.
@@ -288,14 +288,9 @@ open class RouteController: NSObject {
         checkForUpdates()
         checkForLocationUsageDescription()
         
-        setupTunnelIntersectionManager()
+        tunnelIntersectionManager.delegate = self
 
         startEvents(accessToken: route.accessToken)
-    }
-    
-    private func setupTunnelIntersectionManager() {
-        tunnelIntersectionManager = TunnelIntersectionManager()
-        tunnelIntersectionManager?.delegate = self
     }
 
     deinit {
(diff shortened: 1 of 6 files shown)


bsudekum:
Actually, this is still called on every location update:

https://github.com/mapbox/mapbox-navigation-ios/blob/36be30079c2c1cba0301e8a17f027c82075b56f3/MapboxCoreNavigation/RouteController.swift#L652-L661

MapboxCoreNavigation/RouteController.swift
```diff
@@ -586,7 +581,9 @@ extension RouteController: CLLocationManagerDelegate {
                 self.rawLocation = lastLocation
                 
                 // Check for a tunnel intersection at the current step we found the bad location update.
-                checkForTunnelIntersection(at: lastLocation, for: manager)
+                if tunnelSimulationEnabled {
```
vincethecoder: @bsudekum we should probably let the `TunnelIntersectionManager` maintain the status of the `tunnelSimulationEnabled`. This will definitely come in handy for future maintenance efforts.

MapboxCoreNavigation/TunnelIntersectionManager.swift
```diff
@@ -98,7 +109,7 @@ open class TunnelIntersectionManager: NSObject {
         isAnimationEnabled = true
     }
     
-    @objc public func suspendTunnelAnimation(for manager: CLLocationManager, at location: CLLocation, routeController: RouteController) {
```
bsudekum: @vincethecoder noting, I removed the manager from many functions here as they were unused.

bsudekum:
@1ec5 there are a few name changes that occur in here that would be great to get your eyes on.

MapboxCoreNavigation/TunnelIntersectionManager.swift
```diff
@@ -47,31 +47,42 @@ open class TunnelIntersectionManager: NSObject {
     /**
      The flag that indicates whether simulated location manager is initialized.
      */
-    @objc public var isAnimationEnabled: Bool = false
+    @objc var isAnimationEnabled: Bool = false
```
vincethecoder: @bsudekum Hmm 🤔 is there a reason why we deleted the `public` access level declaration here?

bsudekum: @vincethecoder it's not something the developer should be mutating, nor is it something they should be inspecting if they are using the delegate properly.

PR #1490, opened 2018-06-10
Only update banner if something has changed.

bsudekum:
Setting `primaryLabel.instruction` has a relatively expensive didset:

https://github.com/mapbox/mapbox-navigation-ios/blob/36be30079c2c1cba0301e8a17f027c82075b56f3/MapboxNavigation/InstructionLabel.swift#L16-L34

This PR change the set function on InstructionBannerView to only update the text if something has changed.

I'm open to changing this to inspect to see if the individual pieces have changed instead, something like 

```swift
if primaryLabel.instruction != instruction?.primaryInstruction {
  primaryLabel.instruction = instruction?.primaryInstruction
}
```

This would occur for each component. 

/cc @mapbox/navigation-ios


bsudekum:
This is only getting called when it actually needs to be updated. I think the fix is to make the didSet more performant.

Issue #1487, opened 2018-06-08
NavigationMapView.addCongestion() is expensive


bsudekum:
![image](https://user-images.githubusercontent.com/1058624/41185359-7000c060-6b3c-11e8-8555-35969c043c4d.png)

Unsurprisingly, this function with all sorts of zips, loops, and maps is expensive. 

https://github.com/mapbox/mapbox-navigation-ios/blob/74ab1f81051fe81f551980cf68d3748b84ef4ac4/MapboxNavigation/NavigationMapView.swift#L768-L826

At the very least, refactoring it to be readable would be an improvement.

PR #1492, opened 2018-06-12
Cleanup NavigationMapView.addCongestion()

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1487

This cleans up `NavigationMapView.addCongestion()` by doing a few things:

* reduce the amount of variables using a nested map, reduce, filter, for loop, etc.
* Limits the amount of casting using `Array()`
* Tries to make it a little more readable.

It's hard to tell the exact amount of improvement here, but here is what instruments looks like:

image

/cc @mapbox/navigation-ios


bsudekum:
@mapbox/navigation-ios could I get a review on this?

bsudekum:
I think a test addition should be possible.

bsudekum:
@akitchen added a test and also fixed an issue that the test made me aware of in https://github.com/mapbox/mapbox-navigation-ios/pull/1492/commits/6e648f1f87ed581aeba1e51f6c64005c38a9dc58. There are still a few untested code paths, but I think it's a good start.

PR #1496, opened 2018-06-13
[wip] navigation native

frederoni:
This PR introduces navigation native. There's a lot of remaining tasks to be done before this PR can land.

- [ ] Pull in navigation native as a regular dependency and do not check in the binary
- [ ] Set up RouteController and navigator
- [ ] Various CI changes
- [ ] Comprehensive tests
- [ ] …

@mapbox/navigation-ios


MapboxCoreNavigation/RouteController.swift
```diff
@@ -696,10 +709,7 @@ extension RouteController: CLLocationManagerDelegate {
             return true
         }
 
-        let radius = max(reroutingTolerance, RouteControllerManeuverZoneRadius)
-        let isCloseToCurrentStep = location.isWithin(radius, of: routeProgress.currentLegProgress.currentStep)
-
-        guard !isCloseToCurrentStep || !userCourseIsOnRoute(location) else { return true }
+        guard mbNavigatorStatus?.routeState == .offRoute || !userCourseIsOnRoute(location) else { return true }
```
akitchen: Cool! 

Could this be refactored to give us insight into occurrences where these values differ? I think that might be useful.

akitchen:
A few things for us to consider here:

Rather than having a long-running PR, let's close this out and accept that we have a long-running branch (at least for a little while :)

From there, let's open PR's *into* the nav-native branch for functionality that we want others to review or at least be aware of; trivial things can be pushed directly. I think this will give us better insight into what's going on without needing to scroll through a noisy and ever-growing diff.

When we're closer to merging into master, then let's open a PR in order to talk about grooming final nits before merging.

How does that sound?  @frederoni @bsudekum

PR #1501, opened 2018-06-14
Fix squashed distance label in BaseInstructionsBannerView

bsudekum:
The distance label was never aligned to it's parent causing the maneuver arrow and distance to overlap.

Before:

After:

_Noting this is the custom UI example_

/cc @mapbox/navigation-ios


bsudekum:
It'd actually be interesting to know why this never showed up in the drop in UI but occurs in the custom UI.

bsudekum:
There is some slight shifting going on here 
image

🤔 This should probably not be happening.

akitchen:
> There is some slight shifting going on here

I think this is OK; we now have a bit of vertical padding between the two leftmost labels, consistent with the instruction labels on the right side

Issue #1505, opened 2018-06-18
Course view slides around when switching to overview


1ec5:
When the user taps the Overview button, the course view slides around quite noticeably. It doesn’t settle into the correct position on screen until after the map view’s camera animates to completion. This sliding effect destroys the visual illusion of the course view being “stuck to” a geographic location on the map.

`NavigationMapView.setOverheadCameraView(from:along:for:)` zooms out here using MGLMapView’s built-in camera animation functionality:

https://github.com/mapbox/mapbox-navigation-ios/blob/d8beebac6509883c50b70e5b468e449df45f9161/MapboxNavigation/NavigationMapView.swift#L1049

Most likely, this method should also set a flag that causes RouteMapViewController’s implementation of `NavigationMapViewDelegate.mapViewRegionIsChanging(_:)` to update the course view’s position without animation, similar to this code that runs whenever the location changes:

https://github.com/mapbox/mapbox-navigation-ios/blob/d8beebac6509883c50b70e5b468e449df45f9161/MapboxNavigation/NavigationMapView.swift#L345-L347

(It’s a little unfortunate that the course view is laid out in NavigationMapView itself but has to be synchronized by the map view’s delegate. One solution would be to move the course view functionality into RouteMapViewController. Another would be for MGLMapView to expose the internal method that calls `MGLMapViewDelegate.mapViewRegionIsChanging(_:)`.)

This issue has been present probably since we converted the course view from a view-backed annotation to a manually synchronized view in #402. The issue doesn’t affect the Android navigation SDK, which implements the course view as a feature in a GeoJSON source.

/cc @mapbox/navigation-ios

PR #1506, opened 2018-06-19
Fix UserCourseView sliding

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1505

Fixes and adds a few optimizations around going into overhead mode. The sliding issue mainly is due to the fact we were animating to the position of the puck which is not necessary when going into overhead mode.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -1047,8 +1047,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         camera.pitch = 0
         self.camera = camera
         
-        let camerForLine = cameraThatFitsShape(line, direction: 0, edgePadding: bounds)
-        setCamera(camerForLine, animated: true)
+        let cameraForLine = cameraThatFitsShape(line, direction: 0, edgePadding: bounds)
+        setCamera(cameraForLine, animated: true)
     }
     
     /**
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -343,7 +343,9 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear) : nil
             setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding, completionHandler: nil)
         } else {
-            self.userCourseView?.center = self.convert(location.coordinate, toPointTo: self)
+            UIView.animate(withDuration: 0, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
+                self.userCourseView?.center = self.convert(location.coordinate, toPointTo: self)
+            })
         }
         
         if let userCourseView = userCourseView as? UserCourseView {
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -86,6 +86,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     var animatesUserLocation: Bool = false
     var altitude: CLLocationDistance = defaultAltitude
     var routes: [Route]?
+    var isAnimatingToOverheadMode = false
     
     fileprivate var preferredFramesPerSecond: Int = 60 {
         didSet {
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -324,6 +324,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
     
     @objc private func disableUserCourseTracking() {
+        guard tracksUserCourse else { return }
```
bsudekum: `disableUserCourseTracking()` is called on every map move. We don't need to continually call didSet on `tracksUserCourse`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1040,13 +1039,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             return
         }
         
-        // Sadly, `MGLMapView.setVisibleCoordinateBounds(:edgePadding:animated:)` uses the current pitch and direction of the mapview. Ideally, we'd be able to pass in zero.
+        // Sadly, `MGLMapView.cameraThatFitsShape(_:direction:edgePadding:)` uses the current pitch of the mapview.
+        // Because of this, we need to set it before or it will create a camera for the current pitch instead of 0.
         let camera = self.camera
         camera.pitch = 0
-        camera.heading = 0
```
bsudekum: Minor improvement. We are now force setting only one camera property before calculating the correct camera for the line.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -342,9 +343,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear) : nil
             setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding, completionHandler: nil)
         } else {
-            UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
-                self.userCourseView?.center = self.convert(location.coordinate, toPointTo: self)
-            }, completion: nil)
+            self.userCourseView?.center = self.convert(location.coordinate, toPointTo: self)
```
1ec5: This method gets called on every location update. So removing the `UIView.animate(withDuration:delay:options:animation:)` block means the course view never animates, even between user location updates. `duration` is either 1 or 0 depending on whether the caller wants animation.

In #1505, I meant that we should _also_ call `UIView.animate(withDuration:delay:options:animation:)` with a `duration` of 0 elsewhere, to temporarily disable the implicit animation that happens whenever the view’s `center` changes.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1040,13 +1039,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-        // Sadly, `MGLMapView.setVisibleCoordinateBounds(:edgePadding:animated:)` uses the current pitch and direction of the mapview. Ideally, we'd be able to pass in zero.
+        // Sadly, `MGLMapView.cameraThatFitsShape(_:direction:edgePadding:)` uses the current pitch of the mapview.
+        // Because of this, we need to set it before or it will create a camera for the current pitch instead of 0.
         let camera = self.camera
         camera.pitch = 0
-        camera.heading = 0
         self.camera = camera
         
-        setVisibleCoordinateBounds(line.overlayBounds, edgePadding: bounds, animated: true)
+        let camerForLine = cameraThatFitsShape(line, direction: 0, edgePadding: bounds)
```
1ec5: Typo: `camer`.

MapboxNavigation/NavigationMapView.swift
bsudekum: @1ec5 fixed.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -342,9 +344,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear) : nil
             setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding, completionHandler: nil)
         } else {
-            UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
+            // While animating to overhead mode, don't animate the puck.
+            UIView.animate(withDuration: isAnimatingToOverheadMode ? 0 : duration, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
```
1ec5: This is fine, but it might be cleaner to have `isAnimatingToOverheadMode` affect `duration` up above:

```swift
let duration: TimeInterval = animated && !isAnimatingToOverheadMode ? 1 : 0
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1036,17 +1040,22 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             camera.heading = 0
             camera.centerCoordinate = userLocation
             camera.altitude = NavigationMapView.defaultAltitude
-            setCamera(camera, animated: true)
+            setCamera(camera, withDuration: 1, animationTimingFunction: nil) {
+                self.isAnimatingToOverheadMode = false
```
1ec5: Weak self? (×2)

1ec5:
Here’s a changelog blurb:

> Fixed an issue where the user puck slid around after the user pressed the Overview button.

MapboxNavigation/NavigationMapView.swift
bsudekum: I think it will always be false here, but ok.

PR #1510, opened 2018-06-20
Add MapboxNavigationNative

frederoni:
This PR pulls in MapboxNavigationNative as a dependency.

note that this PR is targeting the `navigation-native` branch.

@bsudekum @akitchen


Cartfile
```diff
@@ -1,5 +1,7 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.0
-github "mapbox/MapboxDirections.swift" ~> 0.21.0
+binary "https://s3.amazonaws.com/mapbox/mapbox-navigation-native/ios/builds/Mapbox-iOS-Navigation-Native.json"
+#github "mapbox/MapboxDirections.swift" ~> 0.21.0
```
bsudekum: It's fine to just uncomment this and remove it.

akitchen:
@frederoni et. al -- what do you suggest we do about the Cocopods problem with this one?

frederoni:
@akitchen should be able to work around it with some tweaks to the podspec or Podfile. If not, we could always add a private pod repo for the time being. The issue here is that a podspec can only depend on a published pod.

frederoni:
Sure, let's consider the unpublished pod as tail work.

PR #1513, opened 2018-06-21
Convert CLLocation<->MBFixLocation

frederoni:
Convert CLLocation↔︎MBFixLocation

_note that this PR is targeting the `navigation-native` branch._

cc @bsudekum @akitchen @1ec5


MapboxCoreNavigation/MBFixLocation.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import MapboxNavigationNative
+import CoreLocation
+
+extension MBFixLocation {
+    
+    func asCLLocation() -> CLLocation {
```
bsudekum: Would it make more sense to make the return here optional? How would we know if the location is bogus values fell back to zero?

MapboxCoreNavigation/MBFixLocation.swift
```diff
@@ -0,0 +1,24 @@
...
+        let coordinate = CLLocationCoordinate2D(latitude: 0, longitude: 0)
+        let altitude: CLLocationDistance = self.altitude?.doubleValue ?? 0
+        let horizontalAccuracy: CLLocationAccuracy = self.accuracyHorizontal?.doubleValue ?? 0
+        let course: CLLocationDirection = self.bearing?.doubleValue ?? 0
+        let speed: CLLocationSpeed = self.speed?.doubleValue ?? 0
+        let timestamp: TimeInterval = self.time?.doubleValue ?? Date().timeIntervalSince1970
```
bsudekum: Nit, self is not needed here i think x 7.

frederoni: It's not needed here or on any of the lines where the local name differs from the instance variable. e.g. `let speed: CLLocationSpeed = self.speed`, `speed` shadows `speed` so `self` can't be inferred.

PR #1514, opened 2018-06-21
Integrated new directions API fields with the tertiary instruction and lane indications

vincethecoder:
These updates make sure we rely on the new directions `LaneIndicationComponents` class for lane information. This information is usually retrieved from the tertiary instruction -- which can either be a `LaneIndicationComponent` or `VisualInstructionComponent`. The instructions banner and integrated tests were also modified to reflect the new changes. 

Relies on https://github.com/mapbox/MapboxDirections.swift/pull/258


MapboxNavigation/LanesView.swift
```diff
@@ -67,30 +67,27 @@ open class LanesView: UIView {
...
-            for (i, lane) in allLanes.enumerated() {
-                let laneView = laneArrowView()
-                laneView.lane = lane
-                laneView.maneuverDirection = step.maneuverDirection
-                laneView.isValid = usableLanes.contains(i as Int)
-                stackView.addArrangedSubview(laneView)
-            }
+        let step = currentLegProgress.currentStep
+
+        guard durationRemaining < RouteControllerMediumAlertInterval,
```
vincethecoder: @bsudekum do you think we should reduce the time threshold for `tertiaryInstructions`?

bsudekum: We should not actually be doing any check like this now. The server should tell us exactly when and what to show.

MapboxCoreNavigation.podspec
```diff
@@ -40,7 +40,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxCoreNavigation"
 
-  s.dependency "MapboxDirections.swift", "~> 0.21.0"
+  s.dependency "MapboxDirections.swift", :git => "https://github.com/mapbox/MapboxDirections.swift.git", :branch => "then"
```
bsudekum: Unfortunately, something like this does not work. Only published pods are able to be pulled in as deps here.

vincethecoder: 😞

MapboxNavigation/LanesView.swift
```diff
@@ -67,30 +67,22 @@ open class LanesView: UIView {
...
-            
-            for (i, lane) in allLanes.enumerated() {
-                let laneView = laneArrowView()
-                laneView.lane = lane
-                laneView.maneuverDirection = step.maneuverDirection
-                laneView.isValid = usableLanes.contains(i as Int)
-                stackView.addArrangedSubview(laneView)
-            }
+        let step = currentLegProgress.currentStep
+        guard let firstInstruction = step.instructionsDisplayedAlongStep?.first,
```
bsudekum: What if there is more than one visual instruction on the step?

MapboxNavigation/LanesView.swift
```diff
@@ -67,30 +67,22 @@ open class LanesView: UIView {
...
+                hide()
+                return
         }
         
-        if stackView.arrangedSubviews.count > 0 {
-            show()
-        } else {
-            hide()
-        }
+        let subviews = lanes.map { LaneView(component: $0, direction: step.maneuverDirection) }
```
bsudekum: Why do we need to init new LaneView's? Shouldn't we already have these objects?

bsudekum: Oh I see.

vincethecoder: Yeah, we are given the `LaneIndicationComponent`s from the API.

MapboxNavigation/LanesView.swift
bsudekum: I think we need something like what we have for the currentSpokenInstruction but for visual instruction:

https://github.com/mapbox/mapbox-navigation-ios/blob/03b93adff9419361fefd55d5fc6c0a293df3f479/MapboxCoreNavigation/RouteProgress.swift#L458-L462

We have all the pieces slightly above this computed variable to create `currentVisualInstruction`.

vincethecoder: True. That was an oversight. I will make the necessary changes to account for multiple visual instructions.

MapboxNavigation/LanesView.swift
```diff
@@ -70,14 +70,22 @@ open class LanesView: UIView {
...
-              }), !lanes.isEmpty
-            else {
-                hide()
-                return
+        guard let instructionsDisplayedAlongStep = step.instructionsDisplayedAlongStep else {
+            hide()
+            return
+        }
+        
+        let components: [ComponentRepresentable] = instructionsDisplayedAlongStep
```
bsudekum: @vincethecoder I'm still not seeing how we're finding the current visual instruction that should be displayed. If a step has 3 different tertiary instructions, how do we know which one to show?

vincethecoder: @bsudekum smh - using the current spoken instruction index. Will update in a moment.

MapboxNavigation/LanesView.swift
```diff
@@ -67,30 +67,28 @@ open class LanesView: UIView {
...
+            hide()
+            return
         }
         
-        if stackView.arrangedSubviews.count > 0 {
-            show()
-        } else {
-            hide()
+        let currentInstruction = visualInstructions[spokenInstructionIndex]
+        guard let lanes: [LaneIndicationComponent] = currentInstruction.tertiaryInstruction?.components.compactMap({ component in
```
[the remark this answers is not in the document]
vincethecoder: Sure!

MapboxNavigation/LanesView.swift
```diff
@@ -67,30 +67,28 @@ open class LanesView: UIView {
...
+        guard let visualInstructions = step.instructionsDisplayedAlongStep, spokenInstructionIndex < visualInstructions.count else {
+            hide()
+            return
         }
         
-        if stackView.arrangedSubviews.count > 0 {
-            show()
-        } else {
-            hide()
+        let currentInstruction = visualInstructions[spokenInstructionIndex]
```
bsudekum: It looks like we're still using `spokenInstructionIndex ` instead of the visual index.

vincethecoder:
Waiting on dependency PR https://github.com/mapbox/MapboxDirections.swift/pull/258 to be merged and published. Then this PR can pass the CI tests before merging into master.

bsudekum:
Started to test a few test cases, here is one that I'm looking at, [request](https://api.mapbox.com/directions/v5/mapbox/driving-traffic/-122.185607,37.783910;-122.178333,37.784571.json?geometries=polyline&annotations=congestion&steps=true&overview=full&access_token=TOKEN&banner_instructions=true), [visual link](https://www.mapbox.com/get-directions/#16.52/37.78489/-122.18102?coordinates=-122.187488,37.78413;-122.178271,37.784712).

However, it looks like something is stuck in an infinite loop:

image

vincethecoder:
@bsudekum Upon changes made [here](https://github.com/mapbox/mapbox-navigation-ios/pull/1514/commits/bff3b2f269960be290d78ef693f4f113dc029dbf), here is the new memory report. 

<p float="left">
  
   
</p>

/cc @akitchen - decided to post the memory report, as the [current function](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/InstructionPresenter.swift#L220) is not unit testable at the moment.

bsudekum:
@vincethecoder something doesn't look right here. The memory chart in your screenshot is over the course of ~3 minutes (`(273 - 124) / 60`) but the puck has only moved a few meters from the beginning of the route. I'm still seeing the same memory growth as of https://github.com/mapbox/mapbox-navigation-ios/pull/1514/commits/0bf61eddeb7707cbdfc33f18b0694a69377ce07e

image

bsudekum:
@vincethecoder looking good, looks like you fixed the performance issue. However for the test case in https://github.com/mapbox/mapbox-navigation-ios/pull/1514#issuecomment-400025288, I'm not seeing a lane banner show up.

vincethecoder:
@bsudekum the lane banner does appear for me:

<p float="left">
  
</p>

bsudekum:
new3

@vincethecoder I'm seeing a few issues:
* As discussed in the MapboxDirections.swift PR (https://github.com/mapbox/MapboxDirections.swift/pull/258), it's impossible to have both a lane view and the then banner visible at the same time. This PR changes the logic to display banner info from the server when the sub component is a lane component, but not a then component. We need to refactor when the next banner view is shown.
* The lane view is only visible for a short amount of time. See gif.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -392,16 +389,36 @@ class RouteMapViewController: UIViewController {
...
     }
 
+    func subInstructionsBannerView(for routeProgress: RouteProgress) {
+        let visualInstructionIndex = routeProgress.currentLegProgress.currentStepProgress.visualInstructionIndex
+        
+        if let visualInstructions = routeProgress.currentLegProgress.currentStep.instructionsDisplayedAlongStep,
+               visualInstructionIndex < visualInstructions.count,
+           let tertiaryInstruction = visualInstructions[visualInstructionIndex].tertiaryInstruction {
+            
+            if tertiaryInstruction.components.contains(where: { $0 is LaneIndicationComponent }) {
```
bsudekum: I wonder if we could clean this up further, this check is happening also again on `LanesView.update(for:)`:

https://github.com/mapbox/mapbox-navigation-ios/blob/67f0ab515f76e5855998acf40fdc66c0d91c83f3/MapboxNavigation/LanesView.swift#L69-L85

MapboxNavigation/NavigationView.swift
```diff
@@ -227,6 +228,35 @@ open class NavigationView: UIView {
...
+        
+        guard let visualInstructions = stepProgress.remainingVisualInstructions else { return }
+        
+        for visualInstruction in visualInstructions {
+            if stepProgress.distanceRemaining <= visualInstruction.distanceAlongStep || stepProgress.visualInstructionIndex == 0 {
+                
+                nextBannerView.hide()
+                lanesView.update(for: visualInstruction, currentStepProgress: routeProgress.currentLegProgress.currentStepProgress)
+                
+                if lanesView.isHidden {
```
bsudekum: Logic like this should not occur. We should be render whatever we have to this view since we know we will never have both a lane view and a then view.

vincethecoder: @bsudekum I concur. Was just being extra-cautious.

MapboxNavigation/NavigationView.swift
```diff
@@ -227,6 +228,35 @@ open class NavigationView: UIView {
...
+                    nextBannerView.update(for: routeProgress)
+                }
+                
+                instructionsBannerView.set(visualInstruction)
+                
+                stepProgress.visualInstructionIndex += 1
```
bsudekum: We're updating this index now in two places?

https://github.com/mapbox/mapbox-navigation-ios/blob/a1d71e469df367e81733318b0c0375bf881ffda6/MapboxNavigation/InstructionsBannerView.swift#L114-L131

bsudekum:
@vincethecoder I think it might be helpful to draw out what the flow looks like here. How do we go from a progress update to a rendering the primary instruction and/or the lane view/next banner? 

Something as simple:

```
instructionsBannerView.update(for: routeController.routeProgress.currentLegProgress)
lanesView.update(for: routeController.routeProgress.currentLegProgress)
nextBannerView.update(for: routeController.routeProgress.currentLegProgress)
```
, [here would be ideal](https://github.com/mapbox/mapbox-navigation-ios/blob/245901d26d9b67fe2a05e2252e39744e402a1bee/MapboxNavigation/RouteMapViewController.swift#L400-L404).

Also one common thing that we are doing all over the place is getting the current visual instruction given the `visualInstructionIndex`. It would be very helpful here to create a getting on RouteProgress similar to currentSpokenInstruction:

https://github.com/mapbox/mapbox-navigation-ios/blob/245901d26d9b67fe2a05e2252e39744e402a1bee/MapboxCoreNavigation/RouteProgress.swift#L458-L462

MapboxNavigation/NavigationView.swift
vincethecoder: @bsudekum nope, these are consumed by 2 different `viewcontrollers` -- `RouteMapViewController` and `CustomViewController`.

bsudekum: This is incorrect. CustonViewController is purely an example of using InstructionsBannerView. RouteMapViewController uses it here:

https://github.com/mapbox/mapbox-navigation-ios/blob/245901d26d9b67fe2a05e2252e39744e402a1bee/MapboxNavigation/RouteMapViewController.swift#L20

vincethecoder:
@bsudekum Ideally that was the intention, to create a rather simple elegant solution:

```
instructionsBannerView.update(for: routeController.routeProgress.currentLegProgress)
lanesView.update(for: routeController.routeProgress.currentLegProgress)
nextBannerView.update(for: routeController.routeProgress.currentLegProgress)
```

That would have been easily fulfilled if our `visualInstructionIndex` was not prematurely incremented -- a very hacky pre-existing logic. In lieu of this drawback, the permanent solution merits an enhancement of this code:
https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/InstructionsBannerView.swift#L121

Next steps: Fix this issue on master and then integrate the change to facilitate easy integration of `visualInstructionIndex` in our current solution.

MapboxNavigation/NavigationView.swift
```diff
@@ -1,5 +1,6 @@
 import UIKit
 import MapboxDirections
+import MapboxCoreNavigation
```
bsudekum: nit, remove

MapboxNavigation/LanesView.swift
```diff
@@ -66,31 +66,33 @@ open class LanesView: UIView {
...
-            show()
-        } else {
-            hide()
+        let laneIndications: [LaneIndicationComponent]? = visualInstruction.tertiaryInstruction?.components.compactMap({ component in
+            guard let lane = component as? LaneIndicationComponent else { return nil }
+            return lane
+        })
+        
+        guard let lanes = laneIndications, !lanes.isEmpty
+            else {
```
bsudekum: nit, move this up a line.

MapboxNavigation/NextBannerView.swift
```diff
@@ -91,30 +91,19 @@ open class NextBannerView: UIView {
...
-    
-    public func update(for routeProgress: RouteProgress) {
-        guard shouldShowNextBanner(for: routeProgress) else {
-            hide()
-            return
+    public func update(for currentStepProgress: RouteStepProgress) {
+        let distanceTraveled = currentStepProgress.distanceTraveled
+        guard let visualInstruction = currentStepProgress.currentVisualInstruction,
+              let tertiaryInstruction = visualInstruction.tertiaryInstruction,
+                  distanceTraveled <= visualInstruction.distanceAlongStep,
```
bsudekum: This should not be necessary, this check should already be happening via in RouteController.swift and we should be using the currentVisualInstructionIndex, similar to how the main instruction is getting updated: 

https://github.com/mapbox/mapbox-navigation-ios/blob/f08eb8ee299dd8b8e71bf0c72adaa64c6c4dd174/MapboxNavigation/RouteMapViewController.swift#L345

MapboxNavigation/LanesView.swift
```diff
@@ -66,31 +66,33 @@ open class LanesView: UIView {
...
-                let laneView = laneArrowView()
-                laneView.lane = lane
-                laneView.maneuverDirection = step.maneuverDirection
-                laneView.isValid = usableLanes.contains(i as Int)
-                stackView.addArrangedSubview(laneView)
-            }
+        let distanceTraveled = currentStepProgress.distanceTraveled
+        guard let visualInstruction = currentStepProgress.currentVisualInstruction,
+              let tertiaryInstruction = visualInstruction.tertiaryInstruction,
+                  distanceTraveled <= visualInstruction.distanceAlongStep,
```
bsudekum: This should not be necessary, this check should already be happening via in RouteController.swift and we should be using the currentVisualInstructionIndex, similar to how the main instruction is getting updated: 

https://github.com/mapbox/mapbox-navigation-ios/blob/f08eb8ee299dd8b8e71bf0c72adaa64c6c4dd174/MapboxNavigation/RouteMapViewController.swift#L345

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -347,6 +343,9 @@ class RouteMapViewController: UIViewController {
     @objc func updateInstructionsBanner(notification: NSNotification) {
         guard let routeProgress = notification.userInfo?[RouteControllerNotificationUserInfoKey.routeProgressKey] as? RouteProgress else { return }
         instructionsBannerView.updateInstruction(routeProgress.currentLegProgress.currentStepProgress.currentVisualInstruction)
+        lanesView.update(for: routeProgress.currentLegProgress.currentStepProgress)
+        nextBannerView.update(for: routeProgress.currentLegProgress.currentStepProgress)
```
bsudekum: Let's make all instruction view have the same signature: `foo.updateInstruction(for:)`.

MapboxNavigation/NextBannerView.swift
vincethecoder: @bsudekum 😸  was already making those changes post the recent merge.

MapboxNavigation/RouteMapViewController.swift
vincethecoder: @bsudekum that's not applicable in that context. If the function was using a route step progress, that would make sense. However, this function is currently updating the instruction with `currentVisualInstruction`:

https://github.com/mapbox/mapbox-navigation-ios/pull/1514/files#diff-7bae0cf60f54997d0d5e86fe02b5cddbL96

It will look weird to compose the function like this:
`public func updateInstruction(for instruction: VisualInstructionBanner?)`

Yeah, completely out of place and definitely not well composed.

MapboxNavigation/LanesView.swift
vincethecoder: @bsudekum what do you mean? This was intentional so enhance readability.

bsudekum: @vincethecoder Move the `else {` one line up from line 88 to 87.

bsudekum:
@vincethecoder I'm seeing the wrong maneuver arrow in the then banner, you can see this by simulating the freeway drive:
simulator screen shot - iphone x - 2018-07-12 at 14 13 04
simulator screen shot - iphone x - 2018-07-12 at 14 12 05
simulator screen shot - iphone x - 2018-07-12 at 14 12 35

MapboxNavigation/NextBannerView.swift
```diff
@@ -91,30 +91,15 @@ open class NextBannerView: UIView {
...
+                  !tertiaryInstruction.containsLaneIndications else {
+                    hide()
+                    return
         }
         
-        guard let instruction = routeProgress.currentLegProgress.upComingStep?.instructionsDisplayedAlongStep?.last else { return }
-        
-        maneuverView.visualInstruction = instruction
-        instructionLabel.instruction = instruction.primaryInstruction
+        maneuverView.visualInstruction = visualInstruction
```
bsudekum: I believe this is the issue with the comment above.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -93,7 +93,7 @@ open class BaseInstructionsBannerView: UIControl {
     /**
      Updates the instructions banner info with a given `VisualInstructionBanner`.
      */
-    public func updateInstruction(_ instruction: VisualInstructionBanner?) {
+    public func update(for instruction: VisualInstructionBanner?) {
```
1ec5: Since this method is public, we should mention in the changelog that it has been renamed. Also, it appears to be inaccessible to Objective-C.

MapboxNavigation/LanesView.swift
```diff
@@ -66,31 +66,28 @@ open class LanesView: UIView {
...
+        guard let tertiaryInstruction = visualInstruction?.tertiaryInstruction,
+                  tertiaryInstruction.containsLaneIndications else {
+                    hide()
+                    return
         }
         
-        if stackView.arrangedSubviews.count > 0 {
-            show()
-        } else {
+        let laneIndications: [LaneIndicationComponent]? = tertiaryInstruction.components.compactMap({ component in
```
1ec5: Nit: we can use trailing closure syntax here, since it isn’t confusing.

MapboxNavigation/LanesView.swift
```diff
@@ -66,31 +66,28 @@ open class LanesView: UIView {
...
+            guard let lane = component as? LaneIndicationComponent else { return nil }
```
1ec5: This can be written more succinctly as `{ component as? LaneIndicationComponent }`.

bsudekum:
@vincethecoder I think we're close, just need a few screenshots that @1ec5 is asking for and we can merge.

CHANGELOG.md
```diff
@@ -11,8 +11,9 @@
 * Fixed an issue when selecting a step from the steps list, you could be brought to the wrong step. [#1524](https://github.com/mapbox/mapbox-navigation-ios/pull/1524/)
 * `StyleManager.locationFor(styleManager:)` now allows for an optional CLLocation to be returned. [#1523](https://github.com/mapbox/mapbox-navigation-ios/pull/1523)
 * NavigationViewController now uses the recommended way `.preferredStatusBarStyle` to set the style of the status bar. [#1535](https://github.com/mapbox/mapbox-navigation-ios/pull/1535)
+* Renamed `InstructionsBannerView.updateInstruction(_:)` to `InstructionsBannerView.update(for:)`.
```
1ec5: Link to this PR from any new changeset entries, so that developers can get the full story from the PR description (which should be expanded to explain the API changes too).

CHANGELOG.md
```diff
@@ -11,8 +11,9 @@
...
 * Deleted `InstructionsBannerView.update(_:)`, which previously updated the banner instruction and the current distance.
-* This is replaced with `InstructionsBannerView.updateDistance(for:)` and `InstructionsBannerView.updateInstruction(_:)`. `InstructionsBannerView.updateDistance(for:)` should be called on every location update, while `InstructionsBannerView.updateInstruction(_:)` should be called when `NSNotification.Name. routeControllerDidPassVisualInstructionPoint` is emitted.
+* This is replaced with `InstructionsBannerView.updateDistance(for:)` and `InstructionsBannerView.update(for:)`. `InstructionsBannerView.updateDistance(for:)` should be called on every location update, while `InstructionsBannerView.update(for:)` should be called when `NSNotification.Name. routeControllerDidPassVisualInstructionPoint` is emitted.
```
1ec5: These three points can potentially be combined into one with less repetition.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -93,7 +93,7 @@ open class BaseInstructionsBannerView: UIControl {
     /**
      Updates the instructions banner info with a given `VisualInstructionBanner`.
      */
-    public func updateInstruction(_ instruction: VisualInstructionBanner?) {
+    @objc public func update(for instruction: VisualInstructionBanner?) {
```
1ec5: The Objective-C name of this method is `-updateWithFor:`. It should be `-updateForVisualInstructionBanner:`. (×3)

vincethecoder:
@1ec5 

screen2

screen1

screen3
screen4

/cc @bsudekum

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -166,7 +167,10 @@ class InstructionPresenter {
     }
 
     private func instructionHasDownloadedAllShields() -> Bool {
-        for component in instruction.textComponents {
+        let textComponents = instruction.components.compactMap { $0 as? VisualInstructionComponent }
+        guard !textComponents.isEmpty else { return false }
```
1ec5: This can be written more succinctly as:

```swift
guard let _ = instruction.components.contains(where: { $0 is VisualInstructionComponent }) else { return false }
```

JThramer: `guard instruction.components.contains(where: { $0 is VisualInstructionComponent }) else { return false }` 😏

vincethecoder: @1ec5 Indeed that is succinct. However, for the purpose of this function scope, we need the variable and also an early exit if its empty. Essentially, we need the contents of `textComponent` to iterate through at the proceeding [lines](https://github.com/mapbox/mapbox-navigation-ios/pull/1514/files/13f2d71321b7d5e320721e611c983e6411a2d904#diff-af48d70309ee195e34cc5e6cc6b27caaR173)

MapboxNavigation/ExitView.swift
```diff
@@ -141,4 +141,10 @@ class ExitView: StylableView {
         let labelTrailing = trailingAnchor.constraint(equalTo: exitNumberLabel.trailingAnchor, constant: 8)
         return [imageLeading, imageLabelSpacing, labelTrailing]
     }
+    
+    static func criticalHash(side: ExitSide, dataSource: DataSource) -> String {
```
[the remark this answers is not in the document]
1ec5: Still unclear on this.

MapboxNavigation/InstructionsBannerView.swift
1ec5: The three methods with this name each need a custom Objective-C name of `-updateForVisualInstructionBanner:`.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -93,7 +93,8 @@ open class BaseInstructionsBannerView: UIControl {
     /**
      Updates the instructions banner info with a given `VisualInstructionBanner`.
      */
-    public func updateInstruction(_ instruction: VisualInstructionBanner?) {
+    @objc(updateForVisualInstruction:)
```
1ec5: This should be `-updateForVisualInstructionBanner:`, given the parameter’s type.

MapboxNavigation/LanesView.swift
```diff
@@ -66,31 +66,26 @@ open class LanesView: UIView {
         separatorView.trailingAnchor.constraint(equalTo: trailingAnchor).isActive = true
     }
     
-    func update(for currentLegProgress: RouteLegProgress) {
-        guard let step = currentLegProgress.upComingStep else { return }
-        guard !currentLegProgress.userHasArrivedAtWaypoint else { return }
-        let durationRemaining = currentLegProgress.currentStepProgress.durationRemaining
-        
+    @objc(updateForVisualInstructionBanner:)
```
1ec5: Missing a documentation comment. (×3)

MapboxNavigation/ManeuverView.swift
```diff
@@ -39,7 +39,13 @@ open class ManeuverView: UIView {
...
     }
 
-    @objc public var visualInstruction: VisualInstructionBanner? {
+    @objc public var visualInstruction: VisualInstruction? {
+        didSet {
+            setNeedsDisplay()
+        }
+    }
+    
+    @objc public var drivingSide: DrivingSide = .right {
```
1ec5: These two properties are missing documentation comments.

MapboxNavigation/ExitView.swift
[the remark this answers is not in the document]
JThramer: Yes -- the reason why it's called `criticalHash` is because the `side:` and `dataSource:` arguments contain critical values that affect the appearance of the `ExitView`. 

Considering that we're caching views by their rendered images, _any_ property that affects how an `ExitView` might appear needs to be accounted for in the cache key logic.

MapboxNavigation/ExitView.swift
```diff
@@ -141,4 +141,13 @@ class ExitView: StylableView {
         let labelTrailing = trailingAnchor.constraint(equalTo: exitNumberLabel.trailingAnchor, constant: 8)
         return [imageLeading, imageLabelSpacing, labelTrailing]
     }
+    
+    /**
+     `criticalHash(side:dataSource:)` generates the cache key needed to hold the `ExitView`'s `imageRepresentation` in the `ImageCache` caching engine.
```
JThramer: Nit: Possessive `its'` is with the apostrophe at the end.

akitchen: <grammar police>

> Nit: Possessive its' is with the apostrophe at the end.

wait, what? where?

"its" is the possessive form (think "his" or "hers" or "theirs")
"it's" is the contraction for "it is"

There is no such thing as its'

</grammar police>

Issue #1511, opened 2018-06-20
locationFor(styleManager:) should optionally return CLLocation


bsudekum:
This function requires we return a CLLocation.
https://github.com/mapbox/mapbox-navigation-ios/blob/5592bfe9a9fc11faec30c884f88ba8e28e0e56b7/MapboxNavigation/StyleManager.swift#L9-L12

As a work around, we're basically already optionally returning nil:

https://github.com/mapbox/mapbox-navigation-ios/blob/96b2d6e7902c529b475e939288615935c2197c2b/MapboxNavigation/NavigationViewController.swift#L611-L621

/cc @mapbox/navigation-ios

PR #1523, opened 2018-06-25
Allow locationFor(styleManager:) to optionally return location

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1511

This makes `StyleManager.locationFor(styleManager:)` allows developer to optionally return a location. The optionality we do in https://github.com/mapbox/mapbox-navigation-ios/blob/96b2d6e7902c529b475e939288615935c2197c2b/MapboxNavigation/NavigationViewController.swift#L611-L621

doesn't seem to help us much in the case where `routeController.location ` is nil.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -608,16 +608,8 @@ extension NavigationViewController: TunnelIntersectionManagerDelegate {
 
 extension NavigationViewController: StyleManagerDelegate {
     
-    public func locationFor(styleManager: StyleManager) -> CLLocation {
-        guard let location = routeController.location else {
-            if let coordinate = routeController.routeProgress.route.coordinates?.first {
-                return CLLocation(latitude: coordinate.latitude, longitude: coordinate.longitude)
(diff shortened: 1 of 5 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -608,16 +608,8 @@ extension NavigationViewController: TunnelIntersectionManagerDelegate {
...
-            if let coordinate = routeController.routeProgress.route.coordinates?.first {
-                return CLLocation(latitude: coordinate.latitude, longitude: coordinate.longitude)
-            } else {
-                return CLLocation()
-            }
-        }
-        
-        return location
+    public func locationFor(styleManager: StyleManager) -> CLLocation? {
+        return routeController.location
```
frederoni: `return routeController.location ?? routeController.routeProgress.route.coordinates?.first` to make sure we have a location when we first apply day/night style? Should be safe to assume that the first coordinate on the route is in the same timezone as the user.

PR #1525, opened 2018-06-26
Pin podspec to pre-release

frederoni:
Pin to the pre-release to satisfy the CocoaPods integration test

_note that this PR is targeting the `navigation-native` branch_


frederoni:
This workaround doesn't seem viable. We need a dynamic framework version of navigation native.

Issue #1503, opened 2018-06-14
An app developer should be able to perform client-side customization of banner instructions


akitchen:
In order to satisfy #637 , at least two pieces of functionality are needed, one of which already exists

New functionality: Allow an app developer to customize the visual instruction (AKA the banner text) before it is displayed (for example, adding/removing text or converting to a custom shield). This would take the form of an optional delegate method resembling `willDisplayVisualInstruction:(MBVisualInstructionBanner *)instruction` where the instruction could be altered (or an alternate value returned) before the banner is displayed. We may need to consider adding other delegate methods related to banner sub-components for finer-grained control -- let's discuss based on more concrete use cases.
 
Existing functionality: Allow an app developer to customize a spoken instruction before it is generated and played - https://github.com/mapbox/mapbox-navigation-ios/blob/99cd0f0ceb33ffa538e042396e4758a29fce6581/MapboxNavigation/RouteVoiceController.swift#L306

Once applications implement these two methods, it should be possible to more easily work around cases of arcane road classifications and missing shields.

akitchen:
Per #637 It should be possible to alter the visual instruction so that a shield can be drawn

1ec5:
Fixed in #1530.

> Once applications implement these two methods, it should be possible to more easily work around cases of arcane road classifications and missing shields.

I’m not especially confident that the application would have enough context to intelligently refine data coming from the Directions API using what’s implemented in #1530. That would require the Directions API to expose more structured data that OSRM or Valhalla provides. (And the router in turn might need to expose more of what OpenStreetMap provides.)

PR #1530, opened 2018-06-29
Visual Instruction Presentation Customization

JThramer:
Implements #1503. 

* Adding delegate method to customize attributed string that is computed by `InstructionPresenter`.

To-do: 
- [ ] Add tests
- [ ] Add example

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/InstructionLabel.swift
+++ b/MapboxNavigation/InstructionLabel.swift
@@ -12,6 +12,7 @@ open class InstructionLabel: StylableLabel, InstructionPresenterDataSource {
     var shieldHeight: CGFloat = 30
     var imageRepository: ImageRepository = .shared
     var imageDownloadCompletion: (() -> Void)?
+    var instructionDelegate: VisualInstructionDelegate?
     
     var instruction: VisualInstruction? {
         didSet {
@@ -28,10 +29,28 @@ open class InstructionLabel: StylableLabel, InstructionPresenterDataSource {
             
             let presenter = InstructionPresenter(instruction, dataSource: self, imageRepository: imageRepository, downloadCompletion: update)
             
-            attributedText = presenter.attributedText()
+            let attributed = presenter.attributedText()
+            attributedText = instructionDelegate?.label?(self, willPresent: instruction, as: attributed) ?? attributed
             instructionPresenter = presenter
         }
     }
 
     private var instructionPresenter: InstructionPresenter?
+}
+
+/**
+ The `VoiceControllerDelegate` protocol defines a method that allows an object to customize presented visual instructions.
+ */
+@objc(MBVisualInstructionDelegate)
+public protocol VisualInstructionDelegate: class {
+    
+    /**
+     Called when an InstructionLabel will present a visual instruction.
+     
+     - parameter label: The label that the instruction will be presented on.
+     - parameter instruction: the `VisualInstruction` that will be presented.
+     - parameter presented: the formatted string that is provided by the instruction presenter
+     - returns: optionally, a customized NSAttributedString that will be presented instead of the default, or if nil, the default behavior will be used.
+     */
+    @objc optional func label(_ label: InstructionLabel, willPresent instruction: VisualInstruction, as presented: NSAttributedString) -> NSAttributedString?
 }
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -7,7 +7,7 @@ import Mapbox
  The `NavigationViewControllerDelegate` provides methods for configuring the map view shown by a `NavigationViewController` and responding to the cancellation of a navigation session.
  */
 @objc(MBNavigationViewControllerDelegate)
-public protocol NavigationViewControllerDelegate {
+public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
     /**
      Called when the navigation view controller is dismissed, such as when the user ends a trip.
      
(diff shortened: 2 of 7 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -538,6 +538,10 @@ extension NavigationViewController: RouteMapViewControllerDelegate {
         }
         return roadName
     }
+    
+    @objc public func label(_ label: InstructionLabel, willPresent instruction: VisualInstruction, as presented: NSAttributedString) -> NSAttributedString? {
```
bsudekum: Curious, why does NavigationViewController need to implement the delegate method?

JThramer: This is just plumbing the delegate method through the view-controller hierarchy, just like we do [here](https://github.com/mapbox/mapbox-navigation-ios/pull/1530/files#diff-ab5a908dc637e51fa78cf2be28d033ecL527) for a `NavigationMapViewDelegate` method or [here](https://github.com/mapbox/mapbox-navigation-ios/pull/1530/files#diff-292af689ce88d5b5aa06966456abc0e9R607) for the _same_ method (`label(_:willPresent:as:)`), being routed through `RouteMapViewController`.

It's not really accurate to say that the `NavigationViewController` is _implementing_ the delegate method so much as it's merely passing the message onto _its_ delegate.

JThramer:
Demonstrative screenshot, where in the delegate method, I implemented logic to reverse the presented strings: 
simulator screen shot - iphone 8 plus - 2018-06-28 at 15 39 12

MapboxNavigation/InstructionLabel.swift
```diff
@@ -28,10 +29,28 @@ open class InstructionLabel: StylableLabel, InstructionPresenterDataSource {
...
+    
+    /**
+     Called when an InstructionLabel will present a visual instruction.
+     
+     - parameter label: The label that the instruction will be presented on.
+     - parameter instruction: the `VisualInstruction` that will be presented.
+     - parameter presented: the formatted string that is provided by the instruction presenter
+     - returns: optionally, a customized NSAttributedString that will be presented instead of the default, or if nil, the default behavior will be used.
+     */
+    @objc optional func label(_ label: InstructionLabel, willPresent instruction: VisualInstruction, as presented: NSAttributedString) -> NSAttributedString?
```
1ec5: A delegate method’s first argument should always be the delegator. Is the delegator an InstructionLabel, or something else? Can it be either or two classes? That would be an antipattern in Cocoa.

Does this “will” method’s name make the implementation’s purpose clear enough? Is it clear that it’s an opportunity to override the presentation?

1ec5: The Objective-C signature of this method is currently `-label:willPresent:as:`. I think it should be `-…:willPresentVisualInstruction:label:presentedAttributedString:` or somesuch.

JThramer: Ah, right you are. Good catch. Will fix.

Examples/Swift/ViewController.swift
```diff
@@ -442,6 +452,20 @@ extension ViewController: NavigationViewControllerDelegate {
     }
 }
 
+// Mark: VisualInstructionDelegate
+extension ViewController: VisualInstructionDelegate {
+    func label(_ label: InstructionLabel, willPresent instruction: VisualInstruction, as presented: NSAttributedString) -> NSAttributedString? {
+        guard exampleMode == .customInstruction else { return nil }
```
bsudekum: With `VoiceController(_:WillSpeak:Instruction:)`, returning nil omits the voice instruction. I'm guess that's not the case here? In reality, I can't really think of a use case where a developer would want to show nothing though.

bsudekum:
With https://github.com/mapbox/mapbox-navigation-ios/pull/1514 coming down the line, can you think of anything that might make this play nicely with the lanes view? Maybe returning nil hides/omits the view? https://github.com/mapbox/mapbox-navigation-ios/pull/1530#discussion_r199961726

MapboxNavigationTests/InstructionsBannerViewIntegrationTests.swift
```diff
@@ -69,6 +71,16 @@ class InstructionsBannerViewIntegrationTests: XCTestCase {
 
         super.tearDown()
     }
+    
+    func testCustomVisualInstructionDelegate() {
```
bsudekum: Also add a test for returning nil.

Examples/Swift/ViewController.swift
JThramer: That's kind of the idea. The utility of "I only want to override default behavior under specific circumstances" seems to outweigh the utility of "it's nice to say 'show nothing' via `nil` instead of `NSAttributedString()`". In this _specific_ example you could _also_ implement default behavior by returning `presented`, but I think we should get into the general habit of using a  "returning nil means the delegate [has no answer](http://thecodelesscode.com/case/31) to the delegator, and thus default behavior should be used" pattern in our delegation, as otherwise you kind of force the developer to implement logic if they implement the optional delegate function. Not a super great pattern.

MapboxNavigationTests/InstructionsBannerViewIntegrationTests.swift
JThramer: - [ ] Test for `nil` response

MapboxNavigationTests/InstructionsBannerViewIntegrationTests.swift
```diff
@@ -69,6 +71,27 @@ class InstructionsBannerViewIntegrationTests: XCTestCase {
...
+        let view = instructionsView()
+        view.instructionDelegate = TextReversingDelegate()
```
frederoni: `instructionDelegate` will be deallocated immediately because it's weak. Keep a local ref to `TextReversingDelegate()`. (×2)

Issue #1500, opened 2018-06-14
Developers should be able to refactor RouteController without breaking the generation of telemetry events


akitchen:
As noted in a few tickets (including #1200, which this ticket replaces) the RouteController is overly tightly coupled with the generation of telemetry events.

We want to decouple these concerns, though not with a global singleton pattern as suggested in #1200.

Instead, the steps taken should resemble:

 - [ ] Backfill domain-layer feature tests for key user workflows (Location updates in, telemetry events out)
 - [ ] Sync up with @mapbox/navigation-android to help bring these tests to Android for cross-platform parity testing
 - [ ] Move all event-related code out of `RouteController`, and test-drive it into a new private component resembling a `RouteControllerEventsDelegate`, who receives `RouteControllerDelegate` messages from the `RouteController`, generates telemetry events and other notifications, and passes `RouteControllerDelegate` messages onward

Depending on lessons learned, the third task could become its own ticket. Let's discuss as we learn more about the impact of this decoupling.

PR #1533, opened 2018-07-02
Break out events

frederoni:
Working towards #1500 

- [x] Fix coding failures with `routeLegProgress`, `applicationState`, and `startTimestamp`
- [x] Refactor RouteController for testability and reusability (Vague, but I think we made good progress)
- [x] Move all event-related code out of `RouteController`, and test-drive it into a new private component resembling a `RouteControllerEventsDelegate`, who receives `RouteControllerDelegate` messages from the `RouteController`, generates telemetry events and other notifications, and passes `RouteControllerDelegate` messages onward

-- 

Jerrad's Tasks

- [x] ~~Make sure reroute status hides after reroute~~ (Issue cut -- #1569)
- [x] EventsManager constructor with `Router` dependency injected as arg
- [x] figure out why `sendCancelEvent(rating:comments:)` signature is weird, also determine how EORVC feedback works (bad pattern - issue: #1563)
- [x] LoD pass
- [x] ~~dedupe `checkForFasterRoute` and `reroute`~~ (unacceptable side effects)


bsudekum:
@JThramer Saw an issue with a few drives yesterday where the reroute status view was shown constantly although it never actually needed to reroute nor did it actually end up rerouting.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -294,23 +125,24 @@ open class RouteController: NSObject {
         UIDevice.current.isBatteryMonitoringEnabled = true
 
         super.init()
+        eventsManager.routeController = self
```
JThramer: `eventsManager = EventsManager(router: self)`

JThramer:
@bsudekum hmm, interesting... I'll look into it. Thanks for the report!

MapboxCoreNavigation/RouteController.swift
```diff
@@ -761,7 +522,7 @@ extension RouteController: CLLocationManagerDelegate {
         }
         let durationRemaining = routeProgress.durationRemaining
 
-        getDirections(from: location) { [weak self] (route, error) in
+        getDirections(from: location, along: routeProgress) { [weak self] (route, error) in
```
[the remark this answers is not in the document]
JThramer: On closer inspection, the two methods are divergent enough for this not to be trivial.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -294,23 +125,24 @@ open class RouteController: NSObject {
...
         checkForLocationUsageDescription()
         
         tunnelIntersectionManager.delegate = self
 
-        startEvents(accessToken: route.accessToken)
+        eventsManager.startEvents(accessToken: route.accessToken)
     }
 
     deinit {
+        eventsManager.sendCancelEvent(rating: eventsManager.endOfRouteStarRating, comment: eventsManager.endOfRouteComment)
```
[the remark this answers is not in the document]
JThramer: Answer: This whole pattern is yucky, but this should be ticketed into it's own issue, as it's not a regression. Will do in the morning.

MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -333,30 +337,37 @@ class RouteControllerTests: XCTestCase {
...
-    }
+    //  TODO: Fix ref cycle
+//    func testRouteControllerDoesNotHaveRetainCycle() {
+//        let locationManager = NavigationLocationManager()
+//        let eventsManager = EventsManager()
+//        eventsManager.manager = eventsManagerSpy
+//        routeControllerSpy = RouteControllerSpy(along: initialRoute, directions: directionsClientSpy, locationManager: locationManager, eventsManager: eventsManager)
+//        eventsManager.delegate = routeControllerSpy
+//        let expectation = XCTestExpectation(description: "Deinit")
+//        routeControllerSpy?.deinitCalled = expectation.fulfill
```
akitchen: This is gross anyways. I'd prefer we assign a weak pointer inside an autorelease pool and ensure it comes back nil.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -181,20 +19,20 @@ open class RouteController: NSObject {
...
-    /**
-     The Directions object used to create the route.
-     */
-    @objc public var directions: Directions
-
     /**
      The route controller’s associated location manager.
      */
-    @objc public var locationManager: NavigationLocationManager! {
+    public var locationManager: NavigationLocationManager! {
```
bsudekum: Can the `@objc` remain?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -859,51 +572,31 @@ extension RouteController: CLLocationManagerDelegate {
         }
     }
 
-    func getDirections(from location: CLLocation, completion: @escaping (_ route: Route?, _ error: Error?)->Void) {
+    func getDirections(from location: CLLocation, along progress: RouteProgress, completion: @escaping (_ route: Route?, _ error: Error?)->Void) {
```
bsudekum: It's sort of strange that a routeprogress is needed to get directions. What if we rename this to

```swift
func getDirections(from location: CLLocation, along previousRoute: Route?, ...
```

MapboxCoreNavigation/Router.swift
```diff
@@ -0,0 +1,15 @@
...
+import MapboxDirections
+
+@objc public protocol Router: class, CLLocationManagerDelegate {
+    @objc var eventsManager: EventsManager! { get }
+    @objc var locationManager: NavigationLocationManager! { get }
+    
+    var usesDefaultUserInterface: Bool { get }
+    var routeProgress: RouteProgress { get }
+
+    func locationIsOnRoute(_ location: CLLocation) -> Bool
```
bsudekum: Is this used anywhere?

MapboxCoreNavigation/EventsManager.swift
```diff
@@ -0,0 +1,304 @@
+import Foundation
+import MapboxMobileEvents
+import MapboxDirections
+
+
+@objc(MBEventsManager)
+open class EventsManager: NSObject {
```
bsudekum: Going to need docs here and a few more other public objects.

MapboxCoreNavigation/EventDetails.swift
```diff
@@ -0,0 +1,363 @@
...
+        try container.encode(durationRemaining, forKey: .durationRemaining)
+        try container.encode(rerouteCount, forKey: .rerouteCount)
+        try container.encode(volumeLevel, forKey: .volumeLevel)
+        try container.encode(audioType, forKey: .audioType)
+        try container.encode(screenBrightness, forKey: .screenBrightness)
+        try container.encode(batteryPluggedIn, forKey: .batteryPluggedIn)
+        try container.encode(batteryLevel, forKey: .batteryLevel)
+        try container.encode(applicationState, forKey: .applicationState)
+        try container.encodeIfPresent(userAbsoluteDistanceToDestination, forKey: .userAbsoluteDistanceToDestination)
+        if let locationEngine = locationEngine {
```
bsudekum: Why is this the only optional unwrapped like this?

MapboxCoreNavigation/RouteController.swift
JThramer: The problem is that the `RouteProgress.reroutingOptions(with:)` method relies on `RouteProgress.remainingWaypoints`.

JThramer:
Needs Regression test step to make sure we're not missing telemetry

JThramer:
Issue with rerouting regression has been fixed. Still need to do one more regression test and make sure events are sending okay.

JThramer:
Didn't mean to approve, just blow away the "Changes Requested." No biggie, we'll just have another approval before merge.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -405,6 +197,20 @@ open class RouteController: NSObject {
         locationManager.stopUpdatingHeading()
         locationManager.delegate = nil
     }
+    
+    /**
+     Ends the current navigation session.
+     */
+    @objc public func endNavigation(feedback: EndOfRouteFeedback? = nil) {
+        if !didSendCancelEvent {
```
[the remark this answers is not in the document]
JThramer: AFAIK, `RouteController` is not designed to be reused at this time.

JThramer:
🎵 _Scooter Gang, Scooter Gang, Scooter Gang..._ 🎵 

Scooter Gang.

🎉 After extensive ~drive~scoot-testing today, I have confirmed that the following events are received by the telemetry system:

- `navigation_cancel`
- `navigation_reroute`
- `navigation_feedback`
- `navigation_arrive`
- `navigation_depart`

I'm going to call this ✅ for purposes of regression-testing telemetry events.

MapboxCoreNavigation/RouteController.swift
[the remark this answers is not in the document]
JThramer: @akitchen hmm, so yeah in talking with @bsudekum , it sounds like it might be possible (although out of design scope) to reuse the `RouteController`, possibly by changing `RouteController.routeProgress` (as you’re replacing the thing that the Routecontroller is iterating over)… I could reset that flag on `routeProgress`’ `didSet` as a defensive measure, how's that sound?

MapboxCoreNavigation/EventsManager.swift
```diff
@@ -18,6 +18,8 @@ open class EventsManager: NSObject {
     
     weak var routeController: Router!
     
+    private var didSendCancelEvent = false
```
akitchen: I have concerns with moving this into the EventsManager

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+import CoreLocation
+import MapboxDirections
+
+protocol NavigationService: class {
```
bsudekum: Is this used anywhere?

MapboxCoreNavigation/EventsManager.swift
akitchen: I don't think we should make this component stateful in this way; it complicates potentially reusing it. Not sure if that's a use case but silently failing to send subsequent failures would not be good.

MapboxCoreNavigation/NavigationService.swift
JThramer: 😅 ah, nope... that's for the next stage of changes. Will remove.

MapboxCoreNavigation/EventsManager.swift
frederoni: The events manager is stateful, in particular, its `sessionState`. I fail to see how it was better when the RouteController handled the state of the events manager which definitely complicates potentially reusing it.

The only change here is that the EventsManager will handle `applicationWillTerminate(_:)` on its own instead of having the RouteController reporting to the EventsManager that `applicationWillTerminate(_:)`. It's self-contained and more reusable.

akitchen: rather than add a boolean to keep track of this state, we should key it off of the `SessionState`'s state transition.

Issue #1532, opened 2018-07-02
Visual Instructions Index should not be incremented until the user passes the correct distance remaining value


vincethecoder:
We bypass a pre-existing defect addressed by updating `visualInstructionIndex` whenever `stepProgress.visualInstructionIndex == 0`. The `distanceTraveled` provided at the beginning of a step could be greater than the `currentStepProgress.distanceTraveled`, hence the hack to advance the `visualInstructionIndex`.

This ticket will address the aforementioned issue with an elegant solution to ensure this `visualInstructionIndex` can be shared with the visual instructions consumed by the `sub` components of the instructions banner. 

Resolves: https://github.com/mapbox/mapbox-navigation-ios/pull/1514

bsudekum:
Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/1514

PR #1534, opened 2018-07-03
Visual instructions index incremented only when visual instruction distance along step is completed. 

vincethecoder:
As described [here](https://github.com/mapbox/mapbox-navigation-ios/issues/1532), this PR will increment the visual instruction index only until the user passes the correct distance remaining value. 

Address https://github.com/mapbox/mapbox-navigation-ios/issues/1532


MapboxCoreNavigation/RouteController.swift
```diff
@@ -976,6 +977,15 @@ extension RouteController: CLLocationManagerDelegate {
             }
         }
     }
+    
+    func updateVisualInstructionProgress(for location: CLLocation) {
```
bsudekum: The location argument can be removed.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -114,17 +118,14 @@ open class BaseInstructionsBannerView: UIControl {
...
         
-        for visualInstruction in visualInstructions {
-            if stepProgress.distanceRemaining <= visualInstruction.distanceAlongStep || stepProgress.visualInstructionIndex == 0 {
-                
-                set(visualInstruction)
-                
-                stepProgress.visualInstructionIndex += 1
-                break
-            }
+        if !visualInstructionUpdated {
```
bsudekum: This should probably read, `if updated, do update`. Not `if not updated, do update.`

MapboxCoreNavigation/RouteController.swift
```diff
@@ -976,6 +977,15 @@ extension RouteController: CLLocationManagerDelegate {
...
+        let currentStepProgress = routeProgress.currentLegProgress.currentStepProgress
+        guard let visualInstructions = currentStepProgress.remainingVisualInstructions else { return }
+        
+        let visualInstructionDistances = visualInstructions.map { $0.distanceAlongStep }
+        currentStepProgress.visualInstructionIndex = visualInstructionDistances.index { currentStepProgress.distanceRemaining <= $0 } ?? visualInstructions.startIndex
```
bsudekum: Pull this out into a variable do only update the visualInstructionIndex if this found variable is not equal to visualInstructionIndex.

bsudekum:
With the instruction of https://github.com/mapbox/mapbox-navigation-ios/pull/1514, the lane view and the next banner will now also have to maintain state of the current visual instruction. I'm think we may want to follow patterns we have else where in the sdk and emit a notification when it's time to update the banner. This way each view can simply listen in on these changes and simply update their view.

vincethecoder:
@bsudekum Actually, that wouldn't be the case anymore. The computed property [here](https://github.com/mapbox/mapbox-navigation-ios/pull/1534/files#diff-1a2149692b3d7c58b6a0392120d6011aR272) in the RouteController will take care of updating the visual instruction index, to be used by both the lane view and the next banner. Unless there's a unique case I missed.

vincethecoder:
@bsudekum okay done ✅ in recent commit: https://github.com/mapbox/mapbox-navigation-ios/pull/1534/commits/5bb6080381bc9c23d6615e0680455f896a8d416e

bsudekum:
@vincethecoder awesome! Can you also update the CustomViewController example to use the notification: https://github.com/mapbox/mapbox-navigation-ios/blob/69d7e6a79992b119f00f99e0398d1e40d5ab46ed/Examples/Swift/CustomViewController.swift#L82

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -114,19 +114,6 @@ open class BaseInstructionsBannerView: UIControl {
     public func update(for currentLegProgress: RouteLegProgress) {
```
bsudekum: What does this function do now? We should figure out a good path forward for developers who use this function currently.

vincethecoder:
@bsudekum Good looks! I included the notification in this commit: https://github.com/mapbox/mapbox-navigation-ios/pull/1534/commits/44b3ad36bfa78a8d1e8b4b8ffbe77ab008e1238e

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -116,24 +116,10 @@ open class BaseInstructionsBannerView: UIControl {
...
-            if stepProgress.distanceRemaining <= visualInstruction.distanceAlongStep || stepProgress.visualInstructionIndex == 0 {
-                
-                set(visualInstruction)
-                
-                stepProgress.visualInstructionIndex += 1
-                break
-            }
-        }
-        
+    public func updateDistance(for currentStepProgress: RouteStepProgress) {
```
bsudekum: I think thus function should be removed and combined with `Set(_:)`.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -90,7 +90,7 @@ open class BaseInstructionsBannerView: UIControl {
         }
     }
     
-    func set(_ instruction: VisualInstructionBanner?) {
+    public func set(_ instruction: VisualInstructionBanner?) {
```
bsudekum: We should provide a good changelog note about how `update(for:)` has been renamed to `set(_:)` and should only be used when the banner should be updated. Previously, calling `update(for:)` could return really for you if an update was not necessary. This is not the case now and should be used sparingly to avoid perf hits.

vincethecoder: Nope, the `update(for:)` has been changed to `updateDistance(for:)`

vincethecoder: @bsudekum ☝️

bsudekum: Right, but it's the idea I'm getting at. We need to tell developers how and when they need to update their top banner.

MapboxNavigation/InstructionsBannerView.swift
vincethecoder: @bsudekum we do not want to combine this with `Set(_:)`. It is crucial that we invoke this everytime we receive a location update.

MapboxNavigation/InstructionsBannerView.swift
vincethecoder: That is going to be added in the changelog as a new API update.

bsudekum: @vincethecoder yes, this is what I'm saying, `We should provide a good changelog note...`.

bsudekum: Let's rename this too to make it more inline with `updateDistance(for:)` and make it more clear what you're updating.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fixed an issue when selecting a step from the steps list, you could be brought to the wrong step. [#1524](https://github.com/mapbox/mapbox-navigation-ios/pull/1524/)
 * `StyleManager.locationFor(styleManager:)` now allows for an optional CLLocation to be returned. [#1523](https://github.com/mapbox/mapbox-navigation-ios/pull/1523)
 * NavigationViewController now uses the recommended way `.preferredStatusBarStyle` to set the style of the status bar. [#1535](https://github.com/mapbox/mapbox-navigation-ios/pull/1535)
+* `InstructionsBannerView.updateDistance(for:)` currently updates the instruction banner's distance information on every location update. Also, `InstructionsBannerView.set(_:)` has been optimized to only modify the displayed instruction details when the `visualInstructionIndex` for the current route step is updated by the `RouteController.updateVisualInstructionProgress()`.
```
bsudekum: Let's make this more clear to developers:

* We removed `InstructionsBannerView.update(_:)` which previously updated the banner instruction as well as the distance.
* This has been been replaced with `InstructionsBannerView.updateDistance(for:)` and also `insert_new_func_name`. `InstructionsBannerView.updateDistance(for:)` should be called on every location update, while `insert_new_func_name` should be called when `NSNotification.Name. routeControllerDidPassVisualInstructionPoint` is emitted. 
* No need to talk about visualInstructionIndex or `RouteController.updateVisualInstructionProgress()`

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -90,7 +90,7 @@ open class BaseInstructionsBannerView: UIControl {
         }
     }
     
-    func set(_ instruction: VisualInstructionBanner?) {
+    public func updateInstruction(_ instruction: VisualInstructionBanner?) {
```
bsudekum: Add documentation

MapboxNavigation/InstructionsBannerView.swift
bsudekum: Add documentation

vincethecoder: @bsudekum that already has documentation 

https://github.com/mapbox/mapbox-navigation-ios/pull/1534/files/51a55587555605310ffac51f1f3c25e258c67cba#diff-7bae0cf60f54997d0d5e86fe02b5cddbR119

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -90,7 +90,10 @@ open class BaseInstructionsBannerView: UIControl {
         }
     }
     
-    func set(_ instruction: VisualInstructionBanner?) {
+    /**
+     Updates the instructions banner info with a given `VisualInstructionBanner`.
+     */
+    public func updateInstruction(_ instruction: VisualInstructionBanner?) {
```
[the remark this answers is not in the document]
vincethecoder: @bsudekum what's the value of using a common denominator when the function only needs a specific info based on the signature?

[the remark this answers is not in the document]
vincethecoder: @bsudekum this will not be feasible for the way our current test cases have been set up. We need to create a dummy `routeProgress` for each one of them. This creates a dependency on `routeProgress`. Yes, it will make things easier for our devs for less verbose access of the currentVisualInstruction, however, it's only one level deep. Which isn't too complex (just one level deep).

https://github.com/mapbox/mapbox-navigation-ios/pull/1534/files#diff-11508d8606e1fc6044baab420712bc9cL83

Issue #1539, opened 2018-07-04
NavigationMapView should be responsible for updating the user puck


bsudekum:
Right now, if you are creating a custom navigation expereince, we require you to also manage updating the user puck. This is not an obvious requirement nor is obvious the correct way to manage it. Instead, we can have NavigationMapView listen in on progress changes (it's in fact already listening). This would remove the need to do something like:

https://github.com/mapbox/mapbox-navigation-ios/blob/20711e3fe05a845f5a12a614a38921749f44dae5/Examples/Swift/CustomViewController.swift#L85-L86

/cc @frederoni

PR #1542, opened 2018-07-04
Remove need for developers to implement NavigationMapView.updateCourseTracking(location:animated:)

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1539

It's not very obvious developers need to update the user puck if they implement their own NavigationMapView and this view can easily handle update the puck when progress notifications occur.

This PR makes `NavigationMapView.updateCourseTracking(location:animated:)` private turns on the location manager sooner. This way, when `NavigationMapView` is inited, we already have a location and the view will be properly centered.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -284,9 +284,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
     
     @objc func progressDidChange(_ notification: Notification) {
-        guard tracksUserCourse else { return }
+        guard !isAnimatingToOverheadMode else { return }
         
         let routeProgress = notification.userInfo![RouteControllerNotificationUserInfoKey.routeProgressKey] as! RouteProgress
+        let location = notification.userInfo![RouteControllerNotificationUserInfoKey.locationKey] as! CLLocation
         
         let stepProgress = routeProgress.currentLegProgress.currentStepProgress
         let expectedTravelTime = stepProgress.step.expectedTravelTime
@@ -306,6 +307,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         } else {
             preferredFramesPerSecond = FrameIntervalOptions.pluggedInFramesPerSecond
         }
+        
+        updateCourseTracking(location: location, animated: true)
     }
     
     
@@ -329,7 +332,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         tracksUserCourse = false
     }
     
-    @objc public func updateCourseTracking(location: CLLocation?, animated: Bool) {
+    @objc func updateCourseTracking(location: CLLocation?, animated: Bool) {
         // While animating to overhead mode, don't animate the puck.
         let duration: TimeInterval = animated && !isAnimatingToOverheadMode ? 1 : 0
         animatesUserLocation = animated
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -160,12 +160,23 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
                               contentFrame.minY + contentFrame.height * 0.5))
     }
     
+    open var routeController: RouteController? {
+        didSet {
+            resumeNotifications()
+        }
+    }
+    
     /**
      Determines whether the map should follow the user location and rotate when the course changes.
      - seealso: NavigationMapViewCourseTrackingDelegate
      */
     open var tracksUserCourse: Bool = false {
         didSet {
+            guard let _ = routeController, tracksUserCourse == true else {
+                print("NavigationMapView.routeController must non-nil to set tracksUserCourse to true.")
+                tracksUserCourse = false
+                return
+            }
             if tracksUserCourse {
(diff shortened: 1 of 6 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -306,6 +307,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         } else {
             preferredFramesPerSecond = FrameIntervalOptions.pluggedInFramesPerSecond
         }
+        
+        updateCourseTracking(location: location, animated: true)
```
1ec5: This method is already risky and could become moreso if we rely on notifications to drive course updates as opposed to manual calls to `updateCourseTracking(location:animated:)`. NavigationMapView apparently listens for notifications coming from _any_ RouteController, so every map view in the application will automatically respond to every route controller in the application. There’s a high potential for crosstalk. We’d need to pass in a specific route controller as the `object` here to avoid that problem:

https://github.com/mapbox/mapbox-navigation-ios/blob/ea468da75a44cfd2153b2071968e35f2765809b7/MapboxNavigation/NavigationMapView.swift#L274

MapboxNavigation/NavigationMapView.swift
```diff
@@ -329,7 +332,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         tracksUserCourse = false
     }
     
-    @objc public func updateCourseTracking(location: CLLocation?, animated: Bool) {
+    @objc func updateCourseTracking(location: CLLocation?, animated: Bool) {
```
1ec5: Should we leave this method public? Is there still a use case for manually updating the course view independently of a route controller?

bsudekum: 🤔maybe. I think it might be wise to wrap these types of functions in a more specific helper function like recenterMap:

https://github.com/mapbox/mapbox-navigation-ios/blob/0b69c3247f500c579bfa1b97f95d5f07f844348d/MapboxNavigation/NavigationMapView.swift#L1064-L1067

MapboxNavigation/NavigationMapView.swift
bsudekum: @1ec5 this may require changing the initializer on NavigationMapview. We currently have no `RouterController` props to grab.

bsudekum: I guess making this an optional property is another option. If you only want to update a single map view, set this prop.

1ec5: When we implemented #402, the idea was that moving the course view was the responsibility of the view controller that contains the map view, not the map view itself, since the view controller knows about its route controller for sure. I think we could’ve made that more clear by making the course view a sibling of the map view instead of one of its views. Relying on a global notification is a poor way to make the map view aware of the view controller’s route controller. Changing the initializer is also problematic because map views are often created using the `init(frame:)` or `init(coder:)` initializer.

> If you only want to update a single map view, set this prop.

I don’t think it should be something that you only use in some cases. If we make the route controller a property, then the view controller should always have to set this property on the map view in order to keep the course view in sync. Otherwise, how would the map view handle the route controller swapping that happens around tunnels?

bsudekum: How would someone add a NavigationMapView to a storyboard if they have not yet created a route controller to pass through?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -160,12 +160,23 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    }
+    
     /**
      Determines whether the map should follow the user location and rotate when the course changes.
      - seealso: NavigationMapViewCourseTrackingDelegate
      */
     open var tracksUserCourse: Bool = false {
         didSet {
+            guard let _ = routeController, tracksUserCourse == true else {
+                print("NavigationMapView.routeController must non-nil to set tracksUserCourse to true.")
```
1ec5: `tracksUserCourse` is basically redundant to `routeController` now. We can move all the logic below into `routeController`’s setter, conditioned on whether `routeController` is non-nil.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -160,12 +160,23 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
                               contentFrame.minY + contentFrame.height * 0.5))
     }
     
+    open var routeController: RouteController? {
+        didSet {
+            resumeNotifications()
```
1ec5: What about suspending notifications on the previous route controller?

MapboxNavigation/NavigationMapView.swift
bsudekum: I'd be open to making this an assertion. Fail fast.

1ec5: And we can rename `routeController` to something that makes its purpose clear, like `userCourseController` (which would suggest a protocol by that name based on RouteController’s interface).

bsudekum:
@1ec5 I'm not sure the road we're headed down is any better, in fact it could be worse. I'm going to close this out and move the good parts found in this pr to https://github.com/mapbox/mapbox-navigation-ios/pull/1547.

PR #1549, opened 2018-07-10
Deredeclare CLLocationManagerDelegate conformance on MGLMapView

1ec5:
Deredeclared MGLMapView’s conformance to the CLLocationManagerDelegate protocol. This SDK no longer makes use of the map SDK’s Core Location–dependent functionality, now that #402 has implemented the user course view (user puck) entirely within the navigation SDK. There is no developer or end-user impact to this change.

/cc @bsudekum @JThramer


bsudekum:
Can you confirm simulating locations still works?

1ec5:
Yes, it still works as far as I can tell. Location simulation uses a location manager that’s built into Core Navigation. The map’s camera and user puck are both implemented within MapboxNavigation. If NavigationMapView or RouteMapViewController were to depend on MGLMapView’s conformance to CLLocationManagerDelegate somehow, these changes would’ve resulted in compiler errors, because Swift disallows sending undeclared messages to objects.

1ec5:
I’ve eviscerated that guide – the part about MGLMapView was never relevant to the guide in the first place.

Issue #1554, opened 2018-07-11
Road and POI labels text cannot display when switch endpoint to .cn


m-stephen:
Now I'm focus on navigation project inside China by using our navigation-ios SDK. When I switch the endpoint to .cn, all the roads and POI labels text cannot display:

![image](https://user-images.githubusercontent.com/2045318/42561679-52789cac-852c-11e8-8f9b-d863c63166b9.png)

After deep dive into this bug, I found that during the `func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle)` function in `NavigationMapView.swift`, our navigation SDK do a localized logic called `self.mapView.localizeLabels()`.

In this function, I saw a couple of lines code:

            let localizedText = text.mgl_expressionLocalized(into: locale)
            if localizedText != text {
                layer.text = localizedText
            }

This function make the labels can be localized to the phone's region & language.
However, in .cn map, layers' properties is a little bit different from global side:

China layers sample:
![image](https://user-images.githubusercontent.com/2045318/42562423-446502fc-852e-11e8-8abb-3403e9bbee9c.png)

This is our global [layer properties' link](https://www.mapbox.com/studio/tilesets/mapbox.mapbox-streets-v7/)

As you see, the layer's from .cn only have the `name_zh` but in `text.mgl_expressionLocalized(into: locale)`, it return the `name_zh-Hans`. I know this function is in our map SDK but I'm not very sure it should be fixed in client side or server side.

After talking with @zmofei, he said the .cn data is directly provided by EMG and it's hard to modify in a short term, so I think fix on client side is a better choice. Should we do some downgrade strategy if our layers do not contain `name_zh-Hans` property? @1ec5 @bsudekum 

cc: @chriswu42 @tony-cj

1ec5:
I was unaware that China Streets differs from the rest-of-world Streets source in this way. Are there any other differences that we should be aware of? The navigation SDK makes a lot of assumptions about the Streets source’s fields and layers.

[...]

PR #1558, opened 2018-07-16
Map labels are no longer blank and using China Map Style URLs if the application is configured to use Mapbox China APIs 

m-stephen:
Issue here: https://github.com/mapbox/mapbox-navigation-ios/issues/1554

After thinking of our navigation SDK for supporting .cn map, and I think there should be a .cn network module like our map SDK(Refer to `MGLNetworkConfiguration.h` & `MGLNetworkConfiguration.m`).

So this PR is make a extension of MGLAccountManager for reading configurations of base api URLs.

And another feature is to change the style URLs when the endpoint is switched to .cn map.


MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,31 @@
...
+    // The URL String of China map style.
+    public let mapboxChinaStyleURL = "mapbox://styles/mapbox/streets-zh-v1"
+    
+    private override init() {
+        super.init()
+        apiBaseURL = Bundle.main.object(forInfoDictionaryKey:"MGLMapboxAPIBaseURL") as? String
+    }
+    
+    // Return of whether the map is China map or not
+    public func isChinaMap() -> Bool{
```
bsudekum: Minor enhancement, you can change this to a computed variable, something like: `var isChinaMap: Bool {...`

MapboxNavigation/NavigationMapView.swift
```diff
@@ -946,6 +946,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      to call this method on the value of `NavigationViewController.mapView`.
      */
     @objc public func localizeLabels() {
+        if(NetworkConfiguration.sharedConfiguration.isChinaMap()){
```
bsudekum: Use a a guard statement here.

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,31 @@
...
+        guard apiBaseURL != nil, apiBaseURL == mapboxChinaBaseAPIURL else {
```
bsudekum: I _think_ `guard apiBaseURL == mapboxChinaBaseAPIURL else {` should be enough here.

m-stephen:
I've modify the codes and one question for you @bsudekum.

I saw that api-directions' host is in `MBDirection.swift` and if we wanna build .cn navigation we should change that. So this file `NetworkConfiguration.swift` should be used in `MBDirection` module as well. 

What's your recommendation for the dependency of this file? Or I should say, which part or where should this file at?

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,31 @@
...
+    private var apiBaseURL : String?
+    
+    // The PRC base URL for Mapbox APIs other than the telemetry API.
+    private let mapboxChinaBaseAPIURL = "https://api.mapbox.cn"
+    
+    // The base URL host for Mapbox China
+    public let mapboxChinaBaseURLHost = "api.mapbox.cn"
+    
+    // The URL String of China map style.
+    public let mapboxChinaStyleURL = "mapbox://styles/mapbox/streets-zh-v1"
```
1ec5: This property is currently unused. Let’s leave it out until we need it. Normally it’s the responsibility of the map SDK to handle style URLs anyways.

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,31 @@
...
+    
+    // Value of whether the map is China map or not
+    public var isChinaMap : Bool!{
+        guard apiBaseURL == mapboxChinaBaseAPIURL else {
+            return false
+        }
+        return true
```
1ec5: The `guard` statement is unnecessary; just return the result of the equality comparison.

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,31 @@
...
+    private let mapboxChinaBaseAPIURL = "https://api.mapbox.cn"
+    
+    // The base URL host for Mapbox China
+    public let mapboxChinaBaseURLHost = "api.mapbox.cn"
+    
+    // The URL String of China map style.
+    public let mapboxChinaStyleURL = "mapbox://styles/mapbox/streets-zh-v1"
+    
+    // Value of whether the map is China map or not
+    public var isChinaMap : Bool!{
```
1ec5: The property name is a bit misleading. If the developer sets the style to a third-party URL, the map isn’t necessarily a China map. Conversely, the same URL in `mapboxChinaStyleURL` resolves to an already localized map even for the .com API endpoint.

Currently, in the map SDK, MGLNetworkConfiguration is an internal implementation detail, but MGLAccountManager is public. Access tokens and API endpoints always go hand-in-hand, so I think it would make sense to move this property to an extension on MGLAccountManager. We can name it something like `hasChinaBaseURL`.

1ec5: Normally there’s no reason to declare a computed property as having an implicitly unwrapped optional (`!`) type. `Bool` is correct, because the return value is never `nil`.

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,31 @@
...
+        guard apiBaseURL == mapboxChinaBaseAPIURL else {
+            return false
+        }
+        return true
+    }
+    
+    private override init() {
+        super.init()
+        apiBaseURL = Bundle.main.object(forInfoDictionaryKey:"MGLMapboxAPIBaseURL") as? String
```
1ec5: Is it necessary to store this base URL upfront? I think it would be more straightforward to execute this code on demand, as part of `hasChinaBaseURL`. It’s likely that Bundle already caches these info dictionary values.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -946,6 +946,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      to call this method on the value of `NavigationViewController.mapView`.
      */
     @objc public func localizeLabels() {
+        guard NetworkConfiguration.sharedConfiguration.isChinaMap == false else{
```
1ec5: Are there ever any instances where a user on .cn would want to see non-Chinese labels? If the application is set to the .cn endpoint but the user’s preferred language is English, should we make sure to show English labels?

Another consideration is that the code below tries to always show local names for roads. Unlike most maps, navigation maps have to match the guidance. The Directions API currently shows the local names in all cases, never localized names: Project-OSRM/osrm-backend#4561. But hopefully we can assume that if the application is using the .cn Maps API endpoint, then it’s also using the .cn Directions API endpoint.

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,31 @@
...
+    @objc static let sharedConfiguration = NetworkConfiguration()
+    
+    // The current API base URL of map
+    private var apiBaseURL : String?
+    
+    // The PRC base URL for Mapbox APIs other than the telemetry API.
+    private let mapboxChinaBaseAPIURL = "https://api.mapbox.cn"
+    
+    // The base URL host for Mapbox China
+    public let mapboxChinaBaseURLHost = "api.mapbox.cn"
```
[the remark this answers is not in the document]
m-stephen: If it can make public, that is really helpful for only change the map endpoint. But for a further consideration, I wanna to make this module easily change the map endpoint, api-directions endpoint and integrate Chinese TTS service provided by Baidu.

MapboxNavigation/NetworkConfiguartion.swift
m-stephen: This property is used for my next PR. I'll change the codes in `MGLStyle.swift' sth like:
` (NetworkConfiguration.sharedConfiguration.isChinaMap ? NetworkConfiguration.sharedConfiguration.mapboxChinaStyleURL : "mapbox://styles/mapbox/navigation-guidance-day-v2") `

MapboxNavigation/NetworkConfiguartion.swift
m-stephen: OK, I'll change that.

MapboxNavigation/NavigationMapView.swift
m-stephen: That's a good question. In China, we can only provide Chinese version map via .cn endpoint(That is depends on EMG map, they do not have any other languages map). 

So if a customer use a .cn map, it means their customers can only use Chinese version map, otherwise they should buy our .com maps(It means they want to freely switch the language by settings)

m-stephen: For another question,  if a customer use a .cn map, our navigation SDK should automatically switch the Directions API endpoint to.cn. That's what my next step should do and why I need make this module, to easily switch the endpoints and make a default map style use for China navigation.

MapboxNavigation/NetworkConfiguartion.swift
m-stephen: If you deep diving into our map SDK, this key have already load in +(void)load function in `MGLNetworkConfiguration.m`
image
So that's NP we write here. Of course, if our map SDK didn't do this, we should execute this code on demand.

MapboxNavigation/NetworkConfiguartion.swift
m-stephen: Yep, I'll correct that.

1ec5: An access token is meaningless without an API endpoint. Up to now, the API endpoint has been assumed to be .com, so the access tokens are .com access tokens. I think it would be helpful to keep the options for access tokens and API endpoints as close as possible.

> For my next step, I wanna to use this module to control the whole endpoints switch like Direction APIs, TTS engine. It should be automatically when our customers use our .cn map.

The map SDK, MapboxDirections.swift, and Mapbox Speech for Swift all respect the `MGLMapboxAccessToken` Info.plist key, so it should already be automatic.

MapboxNavigation/NetworkConfiguartion.swift
1ec5: Sure, the value doesn’t change, but my suggestion is to call `object(forInfoDictionaryKey:)` at the point of use anyways, because it shouldn’t affect performance significantly even if we call it several times in a row.

MapboxNavigation/NavigationMapView.swift
1ec5: Everything already respects `MGLMapboxAccessToken` in Info.plist, so we should have them respect the `MGLMapboxAPIBaseURL` key as well: mapbox/MapboxGeocoder.swift#107. The Directions and SpeechSynthesizer initializers do allow the developer to override both settings, but Info.plist should be the central place for configuration changes. (If we need a central place for configuration changes that can change at runtime, then that should be UserDefaults.)

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -16,11 +16,11 @@ class NetworkConfiguration : NSObject{
...
     // Value of whether the map is China map or not
-    public var isChinaMap : Bool!{
-        guard apiBaseURL == mapboxChinaBaseAPIURL else {
-            return false
+    public var hasChinaBaseURL : Bool{
+        if(apiBaseURL == mapboxChinaBaseAPIURL){
+            return true
         }
-        return true
+        return false
```
1ec5:
```swift
return apiBaseURL == mapboxChinaBaseAPIURL
```

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,23 @@
...
+    
+    // The base URL host for Mapbox China
+    public let mapboxChinaBaseURLHost = "api.mapbox.cn"
+    
+    // The URL String of China map style.
+    public let mapboxChinaStyleURL = "mapbox://styles/mapbox/streets-zh-v1"
+    
+    // Value of whether the map is China map or not
+    public var hasChinaBaseURL : Bool{
+        apiBaseURL = Bundle.main.object(forInfoDictionaryKey:"MGLMapboxAPIBaseURL") as? String
```
1ec5: This can be a local variable.

MapboxNavigation/NetworkConfiguartion.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+class NetworkConfiguration : NSObject{
```
1ec5: Per mapbox/mapbox-gl-native#12417, it seems like the `hasChinaAPIBaseURL` property should go on an extension of MGLAccountManager instead of this singleton.

MapboxNavigation/MGLStyle.swift
```diff
@@ -1,24 +1,26 @@
 import Foundation
 import Mapbox
 
+
 extension MGLStyle {
     // Returns the URL to the current version of the Mapbox Navigation Guidance Day style.
     @objc
-    public class var navigationGuidanceDayStyleURL: URL { get { return URL(string: "mapbox://styles/mapbox/navigation-guidance-day-v2")! } }
+    public class var navigationGuidanceDayStyleURL: URL { get { return URL(string: (MGLAccountManager.hasChinaBaseURL ? MGLAccountManager.mapboxChinaDayStyleURL : "mapbox://styles/mapbox/navigation-guidance-day-v2"))! } }
```
1ec5: I agree with using constants for the China URLs since each gets reused. However, consider turning them into URL-typed static constants within MGLStyle. (There’s no need for string constants if we have constant URLs.)

```swift
extension MGLStyle {
    static let streetsChineseURL = URL(string: "mapbox://styles/mapbox/streets-zh-v1")!
}
```

MapboxNavigation/MGLStyle.swift
```diff
@@ -1,24 +1,26 @@
...
     
     // Returns the URL to the current version of the Mapbox Navigation Guidance Night style.
     @objc
-    public class var navigationGuidanceNightStyleURL: URL { get { return URL(string: "mapbox://styles/mapbox/navigation-guidance-night-v2")! } }
+    public class var navigationGuidanceNightStyleURL: URL { get { return URL(string: (MGLAccountManager.hasChinaBaseURL ? MGLAccountManager.mapboxChinaNightStyleURL : "mapbox://styles/mapbox/navigation-guidance-night-v2"))! } }
```
1ec5: For readability, put this code on multiple lines instead of using a ternary operator. (×4)

MapboxNavigation/MGLStyle.swift
```diff
@@ -1,24 +1,26 @@
...
     
     @objc
     // Returns the URL to the given version of the navigation guidance style. Available version are 1, 2, and 3.
```
1ec5: Add a note that the version is ignored when the API base URL is api.mapbox.cn. (×2)

Also, while you’re here, would you mind turning all four of these comments into proper documentation comments using the `/**` syntax? Thanks!

MapboxNavigation/MGLAccountManager.swift
```diff
@@ -0,0 +1,18 @@
...
+
+
+extension MGLAccountManager{
+    
+    static let mapboxChinaBaseAPIURL = "https://api.mapbox.cn"
+    static let mapboxChinaBaseURLHost = "api.mapbox.cn"
+    static let mapboxChinaDayStyleURL = "mapbox://styles/mapbox/streets-zh-v1"
+    static let mapboxChinaNightStyleURL = "mapbox://styles/mapbox/dark-zh-v1"
+    
+    // Value of whether the map is China map or not
```
1ec5: Use the `/**` syntax, so that Quick Help and jazzy can pick this up as a documentation comment.

MapboxNavigation/MGLAccountManager.swift
```diff
@@ -0,0 +1,18 @@
...
+    public class var hasChinaBaseURL : Bool{
```
1ec5: Since this is public, let’s make it available to Objective-C code using `@objc`.

MapboxNavigation/MGLStyle.swift
m-stephen: Sure.

MapboxNavigation/MGLAccountManager.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+import Mapbox
+
+
+extension MGLAccountManager{
+    
+    static let mapboxChinaBaseAPIURL = "https://api.mapbox.cn"
+    static let mapboxChinaBaseURLHost = "api.mapbox.cn"
+    static let mapboxChinaDayStyleURL = "mapbox://styles/mapbox/streets-zh-v1"
```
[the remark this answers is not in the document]
m-stephen: Actually u r right. I'll modify that.

m-stephen:
@1ec5 I have update documentation and move China style constants into MGLStyle extension.

1ec5:
Oh, one more thing: would you mind adding this to the changelog?

> Map labels are no longer blank if the application is configured to use Mapbox China APIs. (#1558)

PR #1560, opened 2018-07-16
Add missing MapboxDirections.swift classes to docs

bsudekum:
Adds a few missing classes to our docs.

/cc @1ec5

### DIFF
--- a/scripts/document.sh
+++ b/scripts/document.sh
@@ -30,7 +30,7 @@ BASE_URL="https://www.mapbox.com/mapbox-navigation-ios"
 
 # Link to directions documentation
 DIRECTIONS_VERSION=$(grep 'MapboxDirections.swift' Cartfile.resolved | grep -oE '"v.+?"' | grep -oE '[^"v]+')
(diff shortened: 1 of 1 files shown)


scripts/document.sh
```diff
@@ -30,7 +30,7 @@ BASE_URL="https://www.mapbox.com/mapbox-navigation-ios"
 
 # Link to directions documentation
 DIRECTIONS_VERSION=$(grep 'MapboxDirections.swift' Cartfile.resolved | grep -oE '"v.+?"' | grep -oE '[^"v]+')
-DIRECTIONS_SYMBOLS="Directions|DirectionsOptions|DirectionsResult|Intersection|Lane|Match|MatchOptions|Route|RouteLeg|RouteOptions|RouteStep|SpokenInstruction|Tracepoint|VisualInstruction|VisualInstructionComponent|Waypoint"
+DIRECTIONS_SYMBOLS="Directions|DirectionsOptions|DirectionsResult|Intersection|Lane|Match|MatchOptions|Route|RouteLeg|RouteOptions|RouteStep|SpokenInstruction|Tracepoint|VisualInstruction|VisualInstructionComponent|Waypoint|RoadClasses|ComponentRepresentable"
```
1ec5: Also VisualInstructionBanner. Are we missing anything else, @vincethecoder?

vincethecoder: @1ec5 ... and `LaneIndicationComponent`

1ec5: This list is meant for symbols that can appear in the public API, for example as argument types. It looks like `LaneIndicationComponent` won’t be exposed anywhere in the public API even after #1514.

PR #1566, opened 2018-07-18
Update Maps SDK, use new camera fitting api

bsudekum:
This updates the maps SDK to v4.2.0 and takes advantage of a newly exposed API, `MGLMapCamera(_:fitting:edgePadding:)`. This makes the transition to overhead view much smoother since we no longer need to first set the pitch and heading to zero.

- [x] wait for v4.2.0 to be released

/cc @mapbox/navigation-ios


bsudekum:
v4.2.0 was released, ready for review.

Issue #1561, opened 2018-07-17
Crash : AVAudioPlayer(AVAudioPlayerPriv) finishedPlaying:


CheemaAneet:
**Mapbox Navigation SDK version:**0.18.1

This seems to be a dup of #1160  and seems to be occuring if the NavigationViewController is closed while a navigation session is underway and a voice instruction is being played at that time. Although not consistently reproducible(coz most times the same sequence of steps don't end up causing this), but I have seen this atleast twice in recent times. 

Crashed: com.apple.main-thread
0  libobjc.A.dylib                0x1834fc910 objc_msgSend + 16
1  AVFAudio                       0x189c814f4 -[AVAudioPlayer(AVAudioPlayerPriv) finishedPlaying:] + 92
2  Foundation                     0x184d7a0ec __NSThreadPerformPerform + 340
3  CoreFoundation                 0x1842d7404 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 24
4  CoreFoundation                 0x1842d6c2c __CFRunLoopDoSources0 + 276
5  CoreFoundation                 0x1842d479c __CFRunLoopRun + 1204
6  CoreFoundation                 0x1841f4da8 CFRunLoopRunSpecific + 552
7  GraphicsServices               0x1861d9020 GSEventRunModal + 100
8  UIKit                          0x18e211758 UIApplicationMain + 236
9  RelayHealth                    0x10502b508 main (main.m:16)
10 libdyld.dylib                  0x183c85fc0 start + 4

CheemaAneet:
Any thoughts on whats causing this ?

bsudekum:
All I can find is https://stackoverflow.com/a/15325328/1522419 but I'm not really convinced nilling out the delegate should be necessary. 

@mapbox/navigation-ios anyone have any thoughts here? @CheemaAneet how often are you seeing the crash? I've really only seen this crash happen once.

CheemaAneet:
@bsudekum Yeah I know its a wierd one to reproduce, this happened to us twice in a very short space of time and never after that. I am not sure of the steps exactly, all I know is that I tried to close the nav controller while a voice instruction was being played and this happened.

I beleive there's some wierd senario thats causing the AVPlayer notification to be sent to an already deallocated instance even though you are cleaning up the observers as i can see, I can only say cleaning up the delegate seems to be your best bet in this case.

akitchen:
[...]

PR #1570, opened 2018-07-20
Nil out audioPlayer delegate when deiniting

bsudekum:
Closes: https://github.com/mapbox/mapbox-navigation-ios/issues/1561

Per, https://github.com/mapbox/mapbox-navigation-ios/issues/1561 we have a hunch that that the spoken instruction audio player's delegate needs to be nilled out when deiniting. This could potentially be the root cause of https://github.com/mapbox/mapbox-navigation-ios/issues/1561 and also other issues we've seen before when multiple voice controllers are speaking instructions when starting/stopping navigation multiple times.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -110,6 +110,7 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     deinit {
         suspendNotifications()
         speechSynth.stopSpeaking(at: .immediate)
+        audioPlayer?.delegate = nil
     }
     
     func resumeNotifications() {


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -110,6 +110,7 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate, AVAudioP
     deinit {
         suspendNotifications()
         speechSynth.stopSpeaking(at: .immediate)
+        audioPlayer?.delegate = nil
```
1ec5: Shouldn’t AVAudioPlayer be responsible for nilling out its own delegate when it deinits? `AVAudioPlayer.delegate` is unsafe-unowned rather than weak, which is unfortunate, but it’d be surprising if AVAudioPlayer doesn’t nil it out anyways in its deinit method. Is there something else keeping the audio player alive, perhaps?

bsudekum: Yeah let's go with #1572 first.

1ec5:
Per https://github.com/mapbox/mapbox-navigation-ios/issues/1561#issuecomment-406532677, it’s _possible_ that the changes here are good hygiene anyways, as an insurance policy. I was just pointing out that AVAudioPlayer wouldn’t message its delegate after it itself goes away.

bsudekum:
I'd be open to testing both.

bsudekum:
@1ec5 going to re-open this and test this out.

PR #1571, opened 2018-07-20
Fix missing instruction on first step

bsudekum:
https://github.com/mapbox/mapbox-navigation-ios/pull/1547 caused a regression where the first instruction on the first step was never announced.

We need to first initialize the voice controller before we begin listening in on spoken instruction notifications.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -386,7 +386,8 @@ open class NavigationViewController: UIViewController {
     
     override open func viewDidLoad() {
         super.viewDidLoad()
-        //initialize voice controller if it hasn't been overridden
+        // Initialize voice controller if it hasn't been overridden.
+        // This is optional and lazy so it can be mutated by the developer after init.
         _ = voiceController
         resumeNotifications()
         view.clipsToBounds = true
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -386,15 +386,14 @@ open class NavigationViewController: UIViewController {
     
     override open func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -386,16 +386,15 @@ open class NavigationViewController: UIViewController {
     
     override open func viewDidLoad() {
         super.viewDidLoad()
+        //initialize voice controller if it hasn't been overridden
+        _ = voiceController
```
1ec5: This is a rather odd way to initialize an ivar. Could we make the ivar non-lazy instead, so that it initializes along with the view controller?

1ec5: On second thought, that would be problematic because the developer might want to set a different voice controller right after initialization.

1ec5: OK, how about we add a comment saying that we’d make the ivar non-lazy and non-optional, except that we don’t want the [access token check](https://github.com/mapbox/mapbox-speech-swift/blob/b0163852eb6e3efd36effc9893e1eab10932036a/MapboxSpeech/MapboxSpeech.swift#L90) to happen before the developer has had an opportunity to swap in a different voice controller that doesn’t rely on Mapbox services.

Maybe we should defer that access token check…

MapboxNavigation/NavigationViewController.swift
```diff
@@ -386,16 +386,16 @@ open class NavigationViewController: UIViewController {
     
     override open func viewDidLoad() {
         super.viewDidLoad()
+        // Initialize voice controller if it hasn't been overridden.
+        // This is optional and lazy so it can be mutated by the developer after init.
```
akitchen: This still feels icky to me. I'd like us to continue to iterate on this in an upcoming PR

PR #1575, opened 2018-07-23
Basic Example EORVC Restoration.

### LINKED ISSUE #1574: EORVC never presents at end-of-route in basic demo
This seems to be an issue I noticed while wrapping up #1533. I was able to reproduce in `master`, so it looks like it's a regression.

STR: 
1. Check out master
2. Set a short route
3. Notice that the Visual instruction says `You will arrive at Dropped Pin #1`, and when you arrive, you see the `Continue To Next Destination` scene, as if you're in the multiple-waypoints demo.

This behavior should only happen for the multiple-waypoints demo. In the basic demo we should see the EORVC show when a user arrives at their destination.

/cc @mapbox/navigation-ios

JThramer:
Fixes #1574.

* Fixing issue where "Continue to next destination" screen was shown for all example runs, including ones where there is only one destination.

To-Do:

- [x] Bobby's Suggestion

/cc @mapbox/navigation-ios


Examples/Swift/ViewController.swift
```diff
@@ -346,6 +346,10 @@ extension ViewController: NavigationViewControllerDelegate {
     func navigationViewController(_ navigationViewController: NavigationViewController, didArriveAt waypoint: Waypoint) -> Bool {
         // When the user arrives, present a view controller that prompts the user to continue to their next destination
         // This type of screen could show information about a destination, pickup/dropoff confirmation, instructions upon arrival, etc.
+        
+        //If we're not in a "Multiple Stops" demo, show the normal EORVC
+        guard waypoints.count > 2 else { return true }
```
bsudekum: Ideally, this would be based off of route progress and would return false for every waypoint which is not the last.

Examples/Swift/ViewController.swift
```diff
@@ -346,7 +346,16 @@ extension ViewController: NavigationViewControllerDelegate {
     func navigationViewController(_ navigationViewController: NavigationViewController, didArriveAt waypoint: Waypoint) -> Bool {
         // When the user arrives, present a view controller that prompts the user to continue to their next destination
         // This type of screen could show information about a destination, pickup/dropoff confirmation, instructions upon arrival, etc.
-        guard let confirmationController = self.storyboard?.instantiateViewController(withIdentifier: "waypointConfirmation") as? WaypointConfirmationViewController else { return true }
+        
+        //If we're not in a "Multiple Stops" demo, show the normal EORVC
+        let waypoints = navigationViewController.routeController.routeProgress.route.routeOptions.waypoints
```
bsudekum: Use `isFinalLeg` instead: https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/RouteProgress.swift#L36

Issue #1440, opened 2018-05-17
Reimplement current road name label atop navigator status/electronic horizon


1ec5:
https://github.com/mapbox/mapbox-navigation-ios/issues/590#issuecomment-327965691 pointed out examples of the current road name label being unreliable, resulting in the user puck being off from the street source, failing to pick up the road name. A solid fix for this issue would require significant changes to the Directions API, but we do have opportunities to improve the quality of our vector tile–powered workaround:

* The invisible line style layer that we query should vary in width the same way that the visible road layers vary in width: based on the zoom level and road class. A lot of the current unreliability is probably due to the invisible layer being only 1&nbsp;pixel wide. https://github.com/mapbox/mapbox-navigation-ios/blob/026fddd1bc07080f33d0cf3f5e7df3b41a04781a/MapboxNavigation/RouteMapViewController.swift#L776
* For numbered roads, we should show the route number next to the road name (previously #291). A road feature’s `ref` attribute provides only the bare number in most cases, not the prefix or shield that we’re accustomed to from the Directions API. But we can use the `shield` and `reflen` attributes to get the shield image from the style and insert it into the label as part of an attributed string. The style images should already look pretty crisp at the size needed for that label.
* I haven’t checked, but the upcoming Streets source v8 probably introduces some changes to the representation of road labels that could help us more straightforwardly determine the current road’s name.

/cc @mapbox/navigation-ios @danesfeder

1ec5:
> The invisible line style layer that we query should vary in width the same way that the visible road layers vary in width: based on the zoom level and road class.

The line style layer isn’t really invisible: it just happens to be occluded most of the time by the road, landcover, and water layers. So we’ll need to ensure that the added line style layer is no wider than the existing layer for that road.

1ec5:
> A lot of the current unreliability is probably due to the invisible layer being only 1 pixel wide. 

Turns out #1076 quietly slipped in a fix for this issue. Now the line is 20 points wide. Not sure if that has improved the reliability at all.

> A road feature’s `ref` attribute provides only the bare number in most cases, not the prefix or shield that we’re accustomed to from the Directions API. But we can use the `shield` and `reflen` attributes to get the shield image from the style and insert it into the label as part of an attributed string.

[...]

PR #1576, opened 2018-07-24
Retrieve additional map attributes to display reliable current road name.

vincethecoder:
Uses the `shield` and `reflen` attributes to get the shield image from the map style and inserts it into the current road name label as part of an attributed string. Includes a data model for the shields with the associated `textColor` based on factors such as `Locale`, `RoadType` and/or `RoadClass`.

(FYI: The style images provided already look crisp at the size needed for the `navigationView.wayNameView`'s label). 

## TODO - Screenshots:
- [x] interstate
- [x] interstate-business
- [x] highway
- [x] state
- [x] communal
- [x] county
- [x] voivodeship
- [x] regional
- [x] trunk
- [x] primary
- [x] secondary
- [x] road
- [x] main
- [x] federal
- [x] national
- [x] expressway
- [x] motorway 

Addresses #1440
shield-pl-voivodeship-50
shield-ro-county-50
shield-sk-highway-50
shield-us-highway-50
shield-us-interstate-business-50
shield-za-national-50
shield-za-regional-50
shield-at-state-b-50
shield-dk-primary-50
shield-dk-secondary-50
shield-e-road
shield-fi-main-50
shield-fi-trunk-50
shield-gr-motorway-50
shield-us-interstate-50
shield-ro-communal-50


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -808,18 +809,56 @@ extension RouteMapViewController: NavigationViewDelegate {
                     } else {
                         currentName = nil
                     }
+                    
+                    if let line = feature as? MGLPolylineFeature {
```
bsudekum: We've already pulled this value out above. See if you can combine these if let statements.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -808,18 +809,56 @@ extension RouteMapViewController: NavigationViewDelegate {
...
             navigationView.wayNameView.isHidden = false
         } else {
             navigationView.wayNameView.isHidden = true
         }
     }
     
+    private func attributedString(withFont font: UIFont, shieldImage: UIImage, text: String, color: UIColor?) -> NSAttributedString {
+        let attachment = RoadNameLabelAttachment()
+        attachment.font = font
+        attachment.text = text
```
bsudekum: What are we getting out of `RoadNameLabelAttachment` over `NSAttributedString` directly? Having a new text prop sounds a little funny when there is already an initializer for string.

vincethecoder: `RoadNameLabelAttachment` should handle the text and image composition. Our primary goal was to use the text prop as the accessibility text/hint for that road name label attachment image. However, it has been revised to allow the `RoadNameLabelAttachment` to handle the task of inserting the text in the image.

bsudekum:
The text should be .white here I think.

vincethecoder:
@bsudekum what are the coordinates you used for testing?

vincethecoder:
@bsudekum yes, that's correct. The captured image the wrong file. Thanks.

shield-us-interstate-50

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -801,25 +802,56 @@ extension RouteMapViewController: NavigationViewDelegate {
...
-                    if let line = feature as? MGLPolylineFeature, let name = line.attribute(forKey: "name") as? String {
-                        currentName = name
-                    } else if let line = feature as? MGLMultiPolylineFeature, let name = line.attribute(forKey: "name") as? String {
-                        currentName = name
-                    } else {
-                        currentName = nil
+                    if let line = feature as? MGLPolylineFeature {
+                        if let name = line.attribute(forKey: "name") as? String {
+                            currentName = name
+                        } else if let line = feature as? MGLMultiPolylineFeature, let name = line.attribute(forKey: "name") as? String {
```
vincethecoder: This line smells suspicious to me. It's existing logic that checks for `line = feature as? MGLPolylineFeature` and `line = feature as? MGLMultiPolylineFeature` within its scope. Needs further investigation why this exists.
/cc @bsudekum @1ec5

bsudekum:
I'm slightly concerned with design here. I think it's difficult to quickly understand what the little bubble with a shield means. It'd be interesting to see what this would look like where we spell out the ref as well to make this more understandable.

It'd look something like `🛡us-101`.

vincethecoder:
@bsudekum, @1ec5 we should probably bring a designer to this conversation too.

/cc @taraniduncan @Cjballard

bsudekum:
To build on the prior comment, I have the same feeling about the single shield in the way name label as I would about a top banner that only has a single shield in it. IMO, when the top banner has only one element, it almost feels like a bug as it's not descriptive enough for the driver to fully understand what we're trying to tell them.

@vincethecoder do you want to give @1ec5's suggestion a try? It'd look something like

```
if name and ref
  display name and ref
else if name and !ref
  display name
else if !name and ref
  display ref
else 
  hide
```

vincethecoder:
@bsudekum sure thing. Can do!

bsudekum:
Yeah, I think this is looking much better @vincethecoder.

Cjballard:
Looks good to me @vincethecoder 👌

PR #1581, opened 2018-07-26
Immediate End Of Route Events

### LINKED ISSUE #1563: Events manager should not rely on RouteController.deinit to send arrival messages
Hey all, so right now the `RouteController` sends the `cancel` event (with rating and comment) at the [very end](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/RouteController.swift#L312) of it's lifecycle, at `deinit` time.

This is dangerous because if a user force-quits the app, or otherwise makes a hasty exit, the `MMEEventsManager` might not have time to send the arrival event.

A better plan here would be to send the cancel event (with rating and comment) as soon as the user dismisses the EORVC, and also setting a "skip" flag that would skip code that otherwise sends a cancel event at the appropriate time (taking care of the edge-case of not using the EORVC or something).

/cc @mapbox/navigation-ios @rclee

JThramer:
Fixes #1563.

*  Refactoring EORVC dismissal logic to send cancel event immediately, instead of relying on RouteController.deinit

/cc @mapbox/navigation-ios


MapboxCoreNavigation/RouteController.swift
```diff
@@ -194,6 +193,20 @@ open class RouteController: NSObject, Router {
         locationManager.stopUpdatingHeading()
         locationManager.delegate = nil
     }
+    
+    /**
+     End the current navigation session.
+     */
+    @objc public func endNavigation(rating: Int = MMEEventsManager.unrated, comment: String? = nil) {
```
bsudekum: Considering all the logic inside of deinit was already private, what do you think about keeping this function private?

bsudekum:
Will a final telem event be sent if the user force quits the app?

MapboxCoreNavigation/RouteController.swift
JThramer: Consider DTO Here

JThramer:
@bsudekum posterity: Yes. Now with the `NSApplicationWillTerminateWarning` observer, we will have the opportunity to send a `cancel` event on a force-quit. (fun fact: `watchdogd` gives us approximately 5 seconds from that notification to send the event before it terminates the process.)

MapboxCoreNavigation/RouteController.swift
```diff
@@ -174,7 +174,10 @@ open class RouteController: NSObject, Router {
     @objc func didChangeApplicationState() {
         eventsManager.reportChange(to: UIApplication.shared.applicationState)
     }
-
+    @objc private func didRecieveTerminationWarning() {
+        endNavigation()
```
bsudekum: Shouldn't this already be handled by deinit?

JThramer: @bsudekum Believe it or not, `deinit` is not called in a force-quit scenario. 

Demo:

```swift
import UIKit

    override func viewDidLoad() {
        super.viewDidLoad()
        NotificationCenter.default.addObserver(self, selector: #selector(ViewController.willTerminate), name: .UIApplicationWillTerminate, object: nil)
        // Do any additional setup after loading the view, typically from a nib.
    }

    override func didReceiveMemoryWarning() {
        super.didReceiveMemoryWarning()
        // Dispose of any resources that can be recreated.
    }
    
    @objc func willTerminate() {
        print("Will Terminate!")
    }
    
    deinit {
        print("Deinit called!")
    }
}
```

if you run this in a sample project, and force-quit the app by "swiping up" on the app in the app-chooser, you'll notice that the only thing printed out onto the console is `Will Terminate!`

akitchen: Yep. relying on `deinit()` or `-dealloc` can be problematic in a variety of scenarios.

PR #1583, opened 2018-07-31
Ensures status view banner text doesn't get stuck on rerouting in simulation mode.

### LINKED ISSUE #1569: Reroute indication under instruction banner not dismissed beyond display threshold, during simulation mode.
A recent investigation of an intermittent reroute reported by @bsudekum, revealed a consistent reroute triggered at a relatively complex tunnel on I-395 (located at Washington DC). 

@jthramer (Jerrad) and @vincethecoder (Kobe) made these major discoveries:

1 - Consistent reroute a few meters upon entering the I-395 tunnel and another reroute event when we exit the tunnel.
2 - In simulation mode, the reroute banner indicator is never dismissed (or replaced) after the reroute threshold.

Based on these findings, Kobe and Jerrad decided to come up with a set of action plans.

- [ ] Avoid any rerouting when the tunnel manager simulation is enabled.
- [ ] Identify and address the reason why reroute is triggered in simulation mode.
       ~~Address the issue with the interpolated location coordinates within the routecontroller provided by the SimulationLocationManager.~~

vincethecoder:
- Fixed banner stuck on rerouting past the [reroute] threshold when navigation is set to simulation mode.

Fixes #1569 

@JThramer


vincethecoder:
This defect is reproducible using the following coordinates:
```
 - Departure Location: 
     latitude: 38.900278999999998, longitude: -77.015082000000007
 - Arrival Location:
     latitude: 38.884351352236479, longitude: -77.084744200135844
```

/cc @JThramer @mapbox/navigation-ios

bsudekum:
Just curious, how do you trigger a reroute when using SimulatedLocationManager? It should just be following the route and never go off route.

vincethecoder:
@bsudekum Yes, it's something we've already raised in this issue ticket #1569. There will be a secondary ticket that will identify and fix the root cause of that bug.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -330,7 +330,11 @@ class RouteMapViewController: UIViewController {
     @objc func didReroute(notification: NSNotification) {
         guard self.isViewLoaded else { return }
         
-        if !(routeController.locationManager is SimulatedLocationManager) {
+        if let locationManager = routeController.locationManager as? SimulatedLocationManager {
```
JThramer: Since we set this string in a couple of places, I'd like to see it exposed as a constant `NSLocalizedString`.

vincethecoder: @JThramer `value: "Simulating Navigation at %f×"` will not localize. Unfortunately, it's imperative we have it this way.

JThramer: What about a `NSLocalizedString.simulatingNavigation` extension property?

vincethecoder: @JThramer Sure! FYI, you cannot create an extension on `NSLocalizedString`. Instead, we can create an extension on String that returns an `NSLocalizedString`.

vincethecoder:
@JThramer the constant NSLocalizedString changes made ✅

MapboxNavigation/String+LocalizedConstants.swift
```diff
@@ -1,5 +1,8 @@
 extension String {
-    static var localizedSimulationSpeedSummary: String {
-        return NSLocalizedString("USER_IN_SIMULATION_MODE", bundle: .mapboxNavigation, value: "Simulating Navigation at %d×", comment: "The text of a banner that appears during turn-by-turn navigation when route simulation is enabled.")
+    struct Localized {
```
JThramer: nit: Use a caseless enum instead of a struct, as Structs can be initialized.

MapboxNavigation/RouteMapViewController.swift
1ec5: NSLocalizedString isn’t a type, it’s a macro (that gets bridged to Swift as a function).

JThramer: @1ec5 yeah, I had forgotten that. We ended up using this `String.Localized` pattern which I think is pretty darn clean anyway, so all turned out well 😅

PR #1591, opened 2018-08-08
Allow opting out of idle timer management

macdrevx:
Removed UIApplication.shared.isIdleTimerDisabled management from NavigationViewController to avoid conflicting with app-specific idle timer strategies.

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -314,6 +314,11 @@ open class NavigationViewController: UIViewController {
             styleManager.automaticallyAdjustsStyleForTimeOfDay = automaticallyAdjustsStyleForTimeOfDay
         }
     }
+
+    /**
+     If `true`, `UIApplication.isIdleTimerDisabled` is set to `true` in `viewWillAppear(_:)` and `false` in `viewWillDisappear(_:)`. If your application manages the idle timer itself, set this property to `false`.
+     */
+    @objc public var shouldManageApplicationIdleTimer = true
     
     var mapViewController: RouteMapViewController?
     
@@ -395,8 +400,10 @@ open class NavigationViewController: UIViewController {
     
     open override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        
-        UIApplication.shared.isIdleTimerDisabled = true
+
+        if shouldManageApplicationIdleTimer {
+            UIApplication.shared.isIdleTimerDisabled = true
+        }
         
         if routeController.locationManager is SimulatedLocationManager {
             let localized = String.Localized.simulationStatus(speed: 1)
@@ -406,8 +413,10 @@ open class NavigationViewController: UIViewController {
     
     open override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
-        
-        UIApplication.shared.isIdleTimerDisabled = false
+
+        if shouldManageApplicationIdleTimer {
+            UIApplication.shared.isIdleTimerDisabled = false
+        }
         
         routeController.suspendLocationUpdates()
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fixed banner stuck on rerouting past the reroute threshold when navigation is set to simulation mode. ([#1583](https://github.com/mapbox/mapbox-navigation-ios/pull/1583))
 * Modified the current road name displayed with new map attribute fields. We use the road feature `ref` attribute's `shield` and `reflen` to retrieve the shield image presented next to the current road name. ([#1576](https://github.com/mapbox/mapbox-navigation-ios/pull/1576)) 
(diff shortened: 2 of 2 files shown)


macdrevx:
Good idea. I'll update the PR.

macdrevx:
@bsudekum how would you feel about making this a flag on NavigationViewController instead of a delegate method? We could declare it as `@objc public var managesApplicationIdleTimer = true`. I think this might be a simpler API for developers to use.

macdrevx:
@bsudekum updated. I went with the flag on `NavigationViewController`, but let me know if it definitely needs to be a delegate-based option instead.

bsudekum:
@macdrevx this a route I initially went down in https://github.com/mapbox/mapbox-navigation-ios/pull/1476, but however ended up going with a delegate. I think to remain consistent, we should keep this similar to https://github.com/mapbox/mapbox-navigation-ios/pull/1476 and make it a delegate.

macdrevx:
@bsudekum updated.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -198,6 +198,17 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
...
+    /**
+     Called to allow the delegate to opt out of changes to UIApplication.shared.isIdleTimerDisabled
+
+     This method is called before NavigationViewController modifies UIApplication.shared.isIdleTimerDisabled. If not implemented, the proposed changes will be accepted automatically.
+
+     - parameter navigationViewController: The navigation view controller that will modify UIApplication.shared.isIdleTimerDisabled
+     - parameter to: The proposed new value of UIApplication.shared.isIdleTimerDisabled
+     - returns: true to accept the proposed change or false to reject it
+     */
+    @objc optional func navigationViewController(_ navigationViewController: NavigationViewController, shouldSetIsIdleTimerDisabledTo to: Bool) -> Bool
```
bsudekum: Correct me if I'm wrong, but I think the double bool here is redundant. 

This could be rewritten as:

```swift
optional func navigationViewControllerShouldSetIsIdleTimerDisabledTo(_ navigationViewController: NavigationViewController) -> Bool
```

macdrevx: The idea here was that `NavigationViewController` will pass in the value that it wants to change `isIdleTimerDisabled` to and then the delegate will return true or false to accept or reject that change.

Without the `to` parameter, the delegate wouldn't have enough information to know whether the `NavigationViewController` wants to enable or disable the idle timer.

bsudekum: Okay, that makes sense.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -198,6 +198,17 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
      */
     @objc(navigationViewController:roadNameAtLocation:)
     optional func navigationViewController(_ navigationViewController: NavigationViewController, roadNameAt location: CLLocation) -> String?
+
+    /**
+     Called to allow the delegate to opt out of changes to UIApplication.shared.isIdleTimerDisabled
```
1ec5: Nit: this property is properly referred to as `UIApplication.isIdleTimerDisabled` in backticks.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -395,8 +406,10 @@ open class NavigationViewController: UIViewController {
     
     open override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        
-        UIApplication.shared.isIdleTimerDisabled = true
+
+        if delegate?.navigationViewController?(self, shouldSetIsIdleTimerDisabledTo: true) ?? true {
```
1ec5: I hadn’t seen the discussion up to now, but I kind of agree with https://github.com/mapbox/mapbox-navigation-ios/pull/1591#issuecomment-411525350 that it would be better to expose this option as a Boolean flag rather than a delegate method. A NavigationViewController instance only ever calls this method twice at predictable times in its life cycle. Battery management is more nuanced than the idle timer; I suspect the application would know upfront whether the SDK should handle the idle timer automatically or not.

Do you have a strong preference for either approach, @macdrevx? Let us know if you’d like us to take care of going back to the Boolean flag for you. Sorry for the back and forth. 😅

macdrevx: @1ec5 @bsudekum I'm happy to do it either way (and happy to change it back if that's what the team prefers). Just let me know once you decide and I can make the change if needed.

bsudekum: @macdrevx @1ec5 I'm okay with a property, thanks for your flexibility @macdrevx.

MapboxNavigation/NavigationViewController.swift
macdrevx: If the team decides to stick with the delegate, I will update this.

MapboxNavigation/NavigationViewController.swift
macdrevx: Sounds good. Changes forthcoming…

macdrevx:
@bsudekum @1ec5 updated 😄

macdrevx:
I'm not sure why the build is failing (I can't access the log). It builds and passes the unit tests for me locally on Xcode 9.4.1.

macdrevx:
You're welcome, glad I could contribute!

PR #1594, opened 2018-08-08
Allow for default user puck view after enabling tracksUserCourse

bsudekum:
When using a single `NavigationMapView` and reusing the map when inside and outside of navigation mode, it needs to be possible to toggle between the course tracking mode in navigation mode and the default tracking modes when outside of navigation.

Once navigation has begun, this if statement is never false:

https://github.com/mapbox/mapbox-navigation-ios/blob/05096e5729c2a79a3eaa36b97c71c34dbc9c10f4/MapboxNavigation/NavigationMapView.swift#L123

because `userLocationForCourseTracking ` is never nil. Because of this, once exiting navigation, the user location view remains in puck mode instead of allowing for the normal view when setting `mapView.userTrackingMode = .follow`.

~~This PR adds a public function to nil out necessary pieces so `mapView. showsUserLocation` can use the default behavior set on MGLMapView.~~

Now, when the user resets the user tracking mode, tracksUserCourse and userLocationForCourseTracking is set to false to make sure the blue dot returns.

/cc @mapbox/navigation-ios


vincethecoder:
@bsudekum kind reminder, a changelog entry about this new public method would be nice.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -199,6 +199,17 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         }
     }
     
+    /**
+     Resets all properties on `MGLMapView` around user course tracking.
+     
+     If reusing the map view outside of navigation, call this before attempting to change `MGLMapView.userTrackingMode` or changing `MGLMapView.showsUserLocation`.
+     */
+    @objc public func disableUseCourseTracking() {
```
1ec5: This one-time function feels kludgy to me. Instead, NavigationMapView should do a better job of distinguishing between course tracking and location/heading tracking when it comes to the `userTrackingMode` property. Outside of turn-by-turn navigation, there’s no need to indicate the user’s course per se.

bsudekum: I agree. The problem is that `tracksUserCourse = false` means two different things depending on your "mode". If you're in navigation made and `tracksUserCourse == false` this means you have moved the map away from the puck (and you get the smaller puck). When not navigating, `tracksUserCourse == false` IMO means that the custom puck view should be hidden.

1ec5: In other words, `tracksUserCourse` is a bit of a misnomer; it should really be called something like `pointOfView` (first-person versus third-person) or `isOverhead`. `userTrackingMode` is kind of vestigial inside RouteMapViewController, since the map SDK only adjusts the center coordinate and heading but not the pitch and zoom level.

> When not navigating, `tracksUserCourse == false` IMO means that the custom puck view should be hidden.

Let’s make that explicit: a new `showsUserCourse` property that controls whether UserCourseView is visible. We can make it mutually exclusive of `showsUserLocation` to keep the map and navigation SDKs from both displaying an indicator simultaneously. What’s nice is that the developer can explicitly show the course view without following the user’s course (which is what `MGLUserTrackingMode.followWithCourse` would kind of do).

bsudekum:
@1ec5 ended up simplifying this quite a bit. Now the relationship is purely on if `tracksUserCourse` as to whether the custom puck is shown.

bsudekum:
@1ec5 @frederoni updated.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -104,27 +104,24 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            if tracksUserCourse {
                 return !(userCourseView?.isHidden ?? true)
             }
             return super.showsUserLocation
         }
         set {
-            if tracksUserCourse || userLocationForCourseTracking != nil {
-                super.showsUserLocation = false
-                
+            super.showsUserLocation  = !tracksUserCourse
```
1ec5: With this change, `mapView.showsUserLocation = true` and `mapView.showsUserLocation = false` have the same effect on `MGLMapView.showsUserLocation` – `newValue` is ignored.

Instead of overriding the setter, define a `didSet` on `showsUserLocation`:

```swift
didSet {
    if showsUserLocation {
        // showsUserLocation and tracksUserCourse cannot both be true at the same time.
        tracksUserCourse = false
    }
    if showsUserLocation && userTrackingMode == .course {
        if userCourseView == nil { … }
        userCourseView?.isHidden = false
    } else {
        userCourseView?.isHidden = true
    }
}
```

then rename `tracksUserCourse` to `showsUserCourse` to avoid confusion with `userTrackingMode`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -156,6 +153,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      */
     open var tracksUserCourse: Bool = false {
```
1ec5: Rename this property to `showsUserCourse` to avoid confusion with `userTrackingMode`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -156,6 +153,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
         didSet {
+            showsUserLocation = !tracksUserCourse
```
1ec5: It should be possible for both `showsUserLocation` and `showsUserCourse` (née `tracksUserCourse`) to be false, so that nothing is overlaid atop the map:

```swift
if tracksUserCourse {
    showsUserLocation = false
}
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -154,12 +149,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    open var showsUserCourse: Bool = false {
         didSet {
-            if tracksUserCourse {
+            if showsUserCourse {
+                // MGLUserLocation and UserCourseView can’t both appear at the same time.
+                showsUserLocation = false
                 enableFrameByFrameCourseViewTracking(for: 3)
                 altitude = NavigationMapView.defaultAltitude
-                showsUserLocation = true
                 courseTrackingDelegate?.navigationMapViewDidStartTrackingCourse?(self)
```
1ec5: All this stuff should only happen if `userTrackingMode` is `followWithCourse`.

1ec5:
As I’ve mentioned a couple times above, I think it would be appropriate for NavigationMapView to honor `MGLMapView.userTrackingMode` and keep its semantics. The developer would be expected to set both `showsUserCourse` and `userTrackingMode` if they want UserCourseView to be visible and the camera to follow the user. The problem is that NavigationMapView must currently switch off user tracking when showing UserCourseView; otherwise, MGLMapView shows its own user location annotation: mapbox/mapbox-gl-native#5598.

nitaliano:
Closing this out in favor of https://github.com/mapbox/mapbox-navigation-ios/pull/1741

PR #1596, opened 2018-08-09
Make audioPlayer public on MapboxVoiceController

bsudekum:
This moves `audioPlayer` off of RouteVoiceController and moves it to MapboxVoiceController and also makes it public. `audioPlayer` was never used directly on RouteVoiceController but rather only used on the subclassed class MapboxVoiceController. 

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/MapboxVoiceController.swift
+++ b/MapboxNavigation/MapboxVoiceController.swift
@@ -9,7 +9,7 @@ import MapboxDirections
  `MapboxVoiceController` extends the default `RouteVoiceController` by providing a more robust speech synthesizer via the Mapbox Speech API. `RouteVoiceController` will be used as a fallback during poor network conditions.
  */
 @objc(MBMapboxVoiceController)
-open class MapboxVoiceController: RouteVoiceController {
+open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
     
     /**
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Fixed an issue where the steps list drag handlebar disappears whenever the user taps the resume button. ([#1588](https://github.com/mapbox/mapbox-navigation-ios/pull/1588))
 * Resolved the partially styled user interface issue that occurs when the style theme is switched between the `NightStyle` and `DayStyle`, after the resume  button was tapped. ([#1589](https://github.com/mapbox/mapbox-navigation-ios/pull/1589))
 * Fixed a rare crash associated with the keyboard when showing the end of route view controller. [#1599](https://github.com/mapbox/mapbox-navigation-ios/pull/1599/)
+* Made the AVAudioPlayer on MapboxVoiceController public. [#1596](https://github.com/mapbox/mapbox-navigation-ios/pull/1596)
```
1ec5: As far as the developer is concerned, this property never existed before:

> * Added a `MapboxVoiceController.audioPlayer` property. You can use this property to interrupt a spoken instruction or adjust the volume.

1ec5:
This could be one half of #1598, though it doesn’t help application code to coordinate AVSpeechSynthesizer utterances when Polly is unavailable.

PR #1599, opened 2018-08-09
Don't force unwrap userInfo keys

bsudekum:
Seeing a few rare crashes around this code. To be safe, we should not force unwrap these userInfo key/values.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -984,10 +984,12 @@ extension RouteMapViewController {
     @objc fileprivate func keyboardWillShow(notification: NSNotification) {
         guard navigationView.endOfRouteView != nil else { return }
         guard let userInfo = notification.userInfo else { return }
-        let curve = UIViewAnimationCurve(rawValue: userInfo[UIKeyboardAnimationCurveUserInfoKey] as! Int)
-        let options = (duration: userInfo[UIKeyboardAnimationDurationUserInfoKey] as! Double,
-                       curve: curve!)
-        let keyboardHeight = (userInfo[UIKeyboardFrameEndUserInfoKey] as! CGRect).size.height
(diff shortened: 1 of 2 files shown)


bsudekum:
@1ec5 it may not be a missing key but also we could be crashing due to a bad force cast.

1ec5:
```
#0. Crashed: com.apple.main-thread
0  libswiftCore.dylib             0x1011c0abc _stdlib_destroyTLS + 247776
1  libswiftCore.dylib             0x10101ff74 (Missing)
2  MapboxNavigation               0x100747610 $S16MapboxNavigation22RouteMapViewControllerC16keyboardWillShow33_AB58FB88B49AF290C13A69A1EE3638C1LL12notificationySo14NSNotificationC_tF (RouteMapViewController.swift:959)
3  MapboxNavigation               0x100748048 $S16MapboxNavigation22RouteMapViewControllerC16keyboardWillShow33_AB58FB88B49AF290C13A69A1EE3638C1LL12notificationySo14NSNotificationC_tFToTm + 4310974536
```

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -984,10 +984,12 @@ extension RouteMapViewController {
     @objc fileprivate func keyboardWillShow(notification: NSNotification) {
         guard navigationView.endOfRouteView != nil else { return }
         guard let userInfo = notification.userInfo else { return }
-        let curve = UIViewAnimationCurve(rawValue: userInfo[UIKeyboardAnimationCurveUserInfoKey] as! Int)
-        let options = (duration: userInfo[UIKeyboardAnimationDurationUserInfoKey] as! Double,
-                       curve: curve!)
```
1ec5: This `!` is the problem, per https://github.com/mapbox/mapbox-navigation-ios/issues/1600#issuecomment-411948817. Let’s whittle this PR down to just handling the case where `curve` is `nil`.

bsudekum:
@1ec5 reworked this to use a default animation if the curve init function fails. Tested with 7, so it fails and falls back to easeIn, looks good.

PR #1605, opened 2018-08-14
Make FeedbackViewController public

bsudekum:
This PR makes the FeedbackViewController public for developers to show within their custom navigation UI.

todo:
- [ ] document.
- [ ] convert public structs to classes.
- [ ] figure out why `self.parent` is nil in many instances.

/cc @mapbox/navigation-ios


MapboxNavigation/FeedbackViewController.swift
```diff
@@ -165,10 +178,48 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
...
+        
+        return (send, dismiss)
+    }
+    
+    func defaultSendFeedbackHandler(source: FeedbackSource = .user, uuid: UUID) -> (FeedbackItem) -> Void {
+        return { [weak self] (item) in
+            guard let strongSelf = self else { return }
+            
+            // todo, can this be moved as a delegate on this view controller?
+            //strongSelf.delegate?.mapViewController(strongSelf, didSendFeedbackAssigned: uuid, feedbackType: item.feedbackType)
```
bsudekum: Would it make sense to make a protocol for FeedbackViewController here? This would remove some plumbing.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -165,10 +178,48 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
...
+            strongSelf.eventsManager.updateFeedback(uuid: uuid, type: item.feedbackType, source: source, description: nil)
+            
+            guard let parent = strongSelf.parent else {
+                strongSelf.dismiss(animated: true)
+                return
+            }
+            
+            strongSelf.dismiss(animated: true) {
+                DialogViewController().present(on: parent)
```
bsudekum: For some reason, paren is nil here.

MapboxNavigation/FeedbackItem.swift
```diff
@@ -7,7 +7,7 @@ extension UIImage {
     }
 }
 
-struct FeedbackItem {
+public struct FeedbackItem {
```
1ec5: Note that a Swift struct can’t bridge to Objective-C, nor can any method or property that uses that type.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -3,29 +3,25 @@ import MapboxCoreNavigation
 import AVFoundation
 
 extension FeedbackViewController: UIViewControllerTransitioningDelegate {
-    func animationController(forDismissed dismissed: UIViewController) -> UIViewControllerAnimatedTransitioning? {
+    public func animationController(forDismissed dismissed: UIViewController) -> UIViewControllerAnimatedTransitioning? {
```
1ec5: If we make these methods public, we’ll need to give them less awkward names (`forDismissing`?) and documentation and consider making them accessible to Objective-C.

MapboxNavigation/FeedbackViewController.swift
JThramer: It _may_ be because you're using a custom transitioning coordinator. If so, you _may_ need to manually add the child VC and do all the lifecycle methods, such as `UIViewController.willMove(toParentViewController:)`

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -71,7 +69,18 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
         return fullHeight
     }
     
-    override func viewDidLoad() {
+    var eventsManager: EventsManager
+    
+    public init(for eventsManager: EventsManager) {
```
1ec5: Use a more descriptive argument label: `init(eventsManager:)`.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -71,7 +69,18 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
...
+        self.eventsManager = eventsManager
+        super.init(nibName: nil, bundle: nil)
+    }
+    
+    required public init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
```
1ec5: Should this method be implemented?

MapboxNavigation/FeedbackViewController.swift
bsudekum: Is this delegate method even helpful? It only returns the uuid and the feedback type.

MapboxNavigation/FeedbackViewController.swift
bsudekum: @JThramer solved with https://github.com/mapbox/mapbox-navigation-ios/pull/1605/commits/bc32b5e75b1921a7e7f53bc20f1f11aacf38844d#diff-a25e8b3bfe9ceec368b83a004a6d1708R200. How do you feel about it?

JThramer: Ah yeah. That makes a lot more sense. 👍

Examples/Swift/CustomViewController.swift
```diff
@@ -31,6 +33,8 @@ class CustomViewController: UIViewController, MGLMapViewDelegate {
         
         mapView.delegate = self
         mapView.compassView.isHidden = true
+        
+        feedbackViewController = FeedbackViewController(eventsManager: routeController.eventsManager)
```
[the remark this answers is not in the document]
bsudekum: Why is that? Is it really that expensive to just keep it in memory?

MapboxNavigation/FeedbackViewController.swift
bsudekum: Noting, we have no control over the naming conventions of this method since it's a func defined on `UIViewControllerTransitioningDelegate `.

bsudekum:
This is ready for review.

MapboxNavigation/FeedbackViewController.swift
1ec5: So it is! 😄 (By the way, the Objective-C method name looks fine; it’s just the Swift name that’s funky.) Oh well, carry on.

MapboxNavigation/FeedbackItem.swift
```diff
@@ -7,7 +7,8 @@ extension UIImage {
     }
 }
 
-struct FeedbackItem {
+@objc(MBFeedbackItem)
```
1ec5: This class needs documentation.

MapboxNavigation/FeedbackItem.swift
```diff
@@ -7,7 +7,8 @@ extension UIImage {
...
+public class FeedbackItem: NSObject {
     var title: String
     var image: UIImage
     var feedbackType: FeedbackType
```
1ec5: Do any of these properties need to be public? What about the initializer?

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -35,6 +34,11 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
     
     let interactor = Interactor()
     
+    /**
+     The feedback items that are visible and selectable by the user.
+     */
+    public var sections: [[FeedbackItem]] = [[.turnNotAllowed, .closure, .reportTraffic, .confusingInstructions, .generalMapError, .badRoute]]
```
1ec5: Why is this an array nested inside an array? What does the outer array represent?

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -71,7 +75,33 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
         return fullHeight
     }
     
-    override func viewDidLoad() {
+    var eventsManager: EventsManager
```
1ec5: This property (or at least its getter) should be public. If application code can pass an EventsManger into the initializer, it should be able to see which EventsManager the object wound up with.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -165,10 +202,48 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
         progressBar.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
         progressBar.bottomAnchor.constraint(equalTo: view.safeBottomAnchor).isActive = true
     }
+    
+    func defaultFeedbackHandlers(source: FeedbackSource = .user) -> (send: (FeedbackItem) -> Void, dismiss: () -> Void) {
```
1ec5: Inline this method into the one call site, to avoid the cognitive overhead of passing these closures around.

MapboxNavigation/FeedbackViewController.swift
frederoni: Nested sections were previously used to forcefully use one or two rows based on device orientation. This is being handled automatically now:
https://github.com/mapbox/mapbox-navigation-ios/blob/4fa51577668b46700bbd07ff37f0c45b636427f4/MapboxNavigation/FeedbackViewController.swift#L287-L289

So should be fine to flatten `[[FeedbackItem]]` to `[FeedbackItem]` and always use the first section. However, keeping the current structure is a bit more versatile.

MapboxNavigation/FeedbackItem.swift
bsudekum: I think the initializer should be enough, no?

bsudekum:
@1ec5 updated

MapboxNavigation/FeedbackItem.swift
1ec5: The initializer is now public, but the arguments passed into the initializer are thereafter inaccessible to client code, making the feedback item something of a dropbox. I think we should make these three properties public for consistency.

MapboxNavigation/FeedbackItem.swift
```diff
@@ -7,13 +7,19 @@ extension UIImage {
...
 public class FeedbackItem: NSObject {
     var title: String
     var image: UIImage
     var feedbackType: FeedbackType
     
-    init(title: String, image: UIImage, feedbackType: FeedbackType) {
+    /**
+     Creates a new `FeedbackItem`.
+     */
+    public init(title: String, image: UIImage, feedbackType: FeedbackType) {
```
1ec5: Add `@objc` to make this initializer accessible to Objective-C code.

MapboxNavigation/FeedbackViewController.swift
1ec5: To make it useful, pass in the feedback item itself instead of the feedback type.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -165,12 +205,42 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
         progressBar.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
         progressBar.bottomAnchor.constraint(equalTo: view.safeBottomAnchor).isActive = true
     }
+    
+    func defaultSendFeedbackHandler(source: FeedbackSource = .user, uuid: UUID) -> (FeedbackItem) -> Void {
```
1ec5: There’s only one call site and it doesn’t specify the `source` argument, so inline the default `.user` value below.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -80,14 +113,18 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
         view.backgroundColor = .white
         progressBar.barColor = #colorLiteral(red: 0.9347146749, green: 0.5047877431, blue: 0.1419634521, alpha: 1)
         enableDraggableDismiss()
+        
+        let uuid = eventsManager.recordFeedback()
+        sendFeedbackHandler = defaultSendFeedbackHandler(uuid: uuid)
+        dismissFeedbackHandler = defaultDismissFeedbackHandler(uuid: uuid)
```
1ec5: Instead of currying closures within `defaultSendFeedbackHandler(uuid:)` and `defaultDismissFeedbackHandler(uuid:)`, pull those methods into simple `send(_:withUUID:)` and `dismiss(_:)` methods and store `uuid` on the class instead. Then, instead of calling `sendFeedbackHandler`, call `send(_:withUUID:)`.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -39,6 +57,8 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
      */
     public var sections: [FeedbackItem] = [.turnNotAllowed, .closure, .reportTraffic, .confusingInstructions, .generalMapError, .badRoute]
     
+    public var delegate: FeedbackViewControllerDelegate?
```
frederoni: nit: weak

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -165,12 +227,32 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
         progressBar.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
         progressBar.bottomAnchor.constraint(equalTo: view.safeBottomAnchor).isActive = true
     }
+    
+    func sendFeedback(_ item: FeedbackItem) {
```
1ec5: Nit: `send(_:)` is the more Swiftian way of naming it these days, or `sendFeedbackItem(_:)` if there’s a conflict with another `send(_:)`.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -165,12 +227,32 @@ class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecog
...
+            dismiss(animated: true)
+            return
+        }
+        
+        dismiss(animated: true) {
+            DialogViewController().present(on: parent)
+        }
+    }
+    
+    func defaultDismissFeedback() {
```
1ec5: Just `dismissFeedbackItem()`, no longer “default”.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -3,29 +3,46 @@ import MapboxCoreNavigation
...
+     */
+    @objc optional func feedbackViewController(_ feedbackViewController: FeedbackViewController, didSendFeedbackAssigned: UUID, feedback: FeedbackItem)
     
-    var sendFeedbackHandler: SendFeedbackHandler?
-    var dismissFeedbackHandler: (() -> Void)?
-    var sections = [FeedbackSection]()
+    /**
+     Called when a `FeedbackViewController` is dismissed for any reason without giving explicit feedback.
+     */
+    @objc optional func feedbackViewControllerDidCancelFeedback(_ feedbackViewController: FeedbackViewController)
```
1ec5: Would it be possible to pass in the FeedbackItem so the application can distinguish between cancellations?

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -3,29 +3,46 @@ import MapboxCoreNavigation
...
+    /**
+     Called when the user opens the feedback form.
+     */
+    @objc optional func  feedbackViewControllerDidOpenFeedback(_ feedbackViewController: FeedbackViewController)
     
-    typealias SendFeedbackHandler = (FeedbackItem) -> Void
+    /**
+     Called when the user submits a feedback event.
+     */
+    @objc optional func feedbackViewController(_ feedbackViewController: FeedbackViewController, didSendFeedbackAssigned: UUID, feedback: FeedbackItem)
```
1ec5: Name this method `feedbackViewController(_:didSend:uuid:)` in Swift and `-feedbackViewController:didSendFeedbackItem:UUID:` in Objective-C.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -3,29 +3,46 @@ import MapboxCoreNavigation
...
-class FeedbackViewController: UIViewController, DismissDraggable, UIGestureRecognizerDelegate {
+/**
+ The `FeedbackViewControllerDelegate` protocol provides methods for responding to feedback events.
+ */
+@objc public protocol FeedbackViewControllerDelegate {
+    
+    /**
+     Called when the user opens the feedback form.
+     */
+    @objc optional func  feedbackViewControllerDidOpenFeedback(_ feedbackViewController: FeedbackViewController)
```
1ec5: Name this method `feedbackViewControllerDidOpen(_:)`, since it gets called when opening the form, not when opening a specific piece of feedback.

MapboxNavigation/FeedbackViewController.swift
bsudekum: Negative.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -3,29 +3,46 @@ import MapboxCoreNavigation
...
+    /**
+     Called when the user opens the feedback form.
+     */
+    @objc optional func  feedbackViewControllerDidOpen(_ feedbackViewController: FeedbackViewController)
     
-    typealias SendFeedbackHandler = (FeedbackItem) -> Void
+    /**
+     Called when the user submits a feedback event.
+     */
+    @objc optional func feedbackViewController(_ feedbackViewController: FeedbackViewController, didSend: FeedbackItem, UUID: UUID)
```
1ec5: * [ ] Distinguish between the argument label and argument name: `didSend feedbackItem: FeedbackItem`
* [ ] The Swift name is currently `feedbackViewController(_:didSend:UUID:)`; it should be `feedbackViewController(_:didSend:uuid:)`
* [ ] The Objective-C name is currently `-feedbackViewController:didSend:UUID:`; it should be `-feedbackViewController:didSendFeedbackItem:UUID:`

MapboxNavigation/FeedbackViewController.swift
1ec5: Then `feedbackViewControllerDidCancel(_:)`.

PR #1606, opened 2018-08-15
Missing telemetry vales

JThramer:
Hey all,

So @bsudekum recently caught that #1533 dropped a few datapoints in the new `EventDetails` class. This PR re-adds them.

* adding platform, operatingSystem, and device datapoints to `EventDeta…ils`. Also alphabetized the properties list.

/cc @mapbox/navigation-ios


JThramer:
Just Waiting on CI.

Issue #1607, opened 2018-08-15
Compiler errors caused by map SDK v4.3.0 — NavigationMapView.preferredFramesPerSecond


1ec5:
Mapbox Maps SDK for iOS v4.3.0 is scheduled for release today. `pod update` and `carthage update` will automatically pull down this release, because the navigation SDK specifies only a minor version with the tadpole operator rather than a patch version.

https://github.com/mapbox/mapbox-navigation-ios/blob/04a8cb5c5af1827982ab587d372c8412d1bf0533/MapboxNavigation.podspec#L47
https://github.com/mapbox/mapbox-navigation-ios/blob/04a8cb5c5af1827982ab587d372c8412d1bf0533/Cartfile#L1

v4.3.0 is slated to include mapbox/mapbox-gl-native#12501, which upstreams some of the frame rate overrides in #709. #935 named the property `NavigationMapView.preferredFramesPerSecond`, which conflicts with the `MGLMapView.preferredFramesPerSecond` property in mapbox/mapbox-gl-native#12501. So anyone who updates their dependencies after v4.3.0 is released will encounter these compiler errors:

```
/path/to/mapbox-navigation-ios/MapboxNavigation/NavigationMapView.swift:91:21: Overriding var must be as accessible as the declaration it overrides
/path/to/mapbox-navigation-ios/MapboxNavigation/NavigationMapView.swift:91:21: Overridden declaration is here (Mapbox.MGLMapView)
/path/to/mapbox-navigation-ios/MapboxNavigation/NavigationMapView.swift:91:21: Property 'preferredFramesPerSecond' with type 'Int' cannot override a property with type 'MGLMapViewPreferredFramesPerSecond'
/path/to/mapbox-navigation-ios/MapboxNavigation/NavigationMapView.swift:91:21: Attempt to override property here (Mapbox.MGLMapView)
```

We’ll need to remove the conflicting property, which is no longer necessary. To work around the incompatibility in v0.19.0, we’ll release a v0.19.1 based on the v0.19.0 tag that cherry-picks the change. Developers who are pinning to v0.19.0 specifically won’t automatically get v0.19.1; they’ll need to either update the dependency manually or explicitly constrain the map SDK dependency to v4.2._x_.

To avoid issues like this in the future, we should coordinate upstreamed changes so that developers can more gracefully migrate.

/cc @mapbox/navigation-ios @mapbox/maps-ios

1ec5:
[...]

PR #1608, opened 2018-08-15
Ensure compatibility with map SDK v4.3.0

1ec5:
Removed `NavigationMapView.preferredFramesPerSecond` in favor of MGLMapView’s implementation, which is coming in iOS map SDK v4.3.0.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,4 +1,4 @@
-binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.2
+binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
 github "mapbox/MapboxDirections.swift" ~> 0.22.0
 github "mapbox/turf-swift" ~> 0.2
 github "mapbox/mapbox-events-ios" ~> 0.4
--- a/MapboxNavigation/MGLMapView+MGLNavigationAdditions.h
+++ b/MapboxNavigation/MGLMapView+MGLNavigationAdditions.h
@@ -4,6 +4,4 @@
 
 - (void)mapViewDidFinishRenderingFrameFullyRendered:(BOOL)fullyRendered;
 
-@property (nonatomic, readonly) CADisplayLink  * _Nullable displayLink;
-
 @end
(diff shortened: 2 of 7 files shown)


MapboxNavigation/MGLMapView+MGLNavigationAdditions.h
```diff
@@ -4,6 +4,4 @@
 
 - (void)mapViewDidFinishRenderingFrameFullyRendered:(BOOL)fullyRendered;
```
bsudekum: One more and we can remove this file!

bsudekum:
My biggest concern here is this change has no real world testing.

Cartfile
```diff
@@ -1,4 +1,4 @@
-binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.2
+binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
```
friedbunny: If semver patch releases of the maps SDK are potentially incompatible (due to close coupling, private API usage, etc.), does it make sense to tighten the pin here to `~> 4.3.0`?

1ec5: With this PR, private API usage is down to a fairly benign minimum – just a forward declaration of `-mapViewDidFinishRenderingFrameFullyRendered:`. The broader root cause in #1607 is that Swift is strict about subclasses having naming collisions with superclasses. It makes you wonder whether we should be adding class prefixes to NavigationMapView properties that are likely to collide with MGLMapView properties in the future. 🤔

I’d rather not specify a patch release because, on balance, fire drills like this are rare. We’d generally want developers to be able to adopt new map SDK versions as they come out; at the same time, we don’t want to prevent developers from using a previous minor version of the map SDK if they require it for some reason. It’s a tough tradeoff, but I think we can avoid future problems just by being more aware of potential pitfalls when adding to MGLMapView or NavigationMapView.

1ec5:
Note that builds will fail until v4.3.0 is out. No way around that interdependency, unfortunately.

bsudekum:
Restarted build now that it's live.

PR #1613, opened 2018-08-17
Add incidents toggle

1ec5:
Added an `MGLMapView.showsIncidents` property to toggle the visibility of any Mapbox Incidents data on a map view. Mapbox Incidents source v1 is a new public tileset that indicates incidents, such as road closures and detours. It is a companion to [Mapbox Traffic source v1](https://www.mapbox.com/vector-tiles/mapbox-traffic-v1/).

The `MGLMapView.showsIncidents` property is based on the existing `MGLMapView.showsTraffic` property. Like `showsTraffic`, `showsIncidents` only toggles the visibility of the relevant layers; it doesn’t add incidents to a style that lacks those layers. (Public styles that come with incidents layers are forthcoming.) Applications that provide a “Shows Traffic” toggle may want to set both `showsTraffic` and `showsIncidents`, since users generally conflate the two kinds of data.

### DIFF
--- a/MapboxNavigation/MGLMapView.swift
+++ b/MapboxNavigation/MGLMapView.swift
@@ -7,33 +7,86 @@ import MapboxCoreNavigation
  An extension on `MGLMapView` that allows for toggling traffic on a map style that contains a [Mapbox Traffic source](https://www.mapbox.com/vector-tiles/mapbox-traffic-v1/).
  */
 extension MGLMapView {
+    /**
+     Returns a set of source identifiers for tilesets that are or include the Mapbox Incidents source.
+     */
+    func sourceIdentifiers(forTileSetIdentifier tileSetIdentifier: String) -> Set<String> {
+        guard let style = style else {
+            return []
+        }
+        return Set(style.sources.compactMap {
+            $0 as? MGLVectorTileSource
+        }.filter {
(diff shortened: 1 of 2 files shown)


1ec5:
Holding onto this one until the tileset’s schema is finalized.

1ec5:
mapbox/mapbox-navigation-android#1226 implements the equivalent functionality on Android.

MapboxNavigation/MGLMapView.swift
```diff
@@ -7,33 +7,86 @@ import MapboxCoreNavigation
...
+            setShowsTileSet(newValue, withIdentifier: "mapbox.mapbox-traffic-v1", layerIdentifier: "traffic")
+        }
+    }
+    
+    /**
+     A Boolean value indicating whether incidents, such as road closures and detours, are visible in the map view’s style.
+     */
+    public var showsIncidents: Bool {
+        get {
+            return showsTileSet(withIdentifier: "mapbox.mapbox-incidents-v1", layerIdentifier: "incidents")
```
1ec5: I jumped the gun and merged this PR without updating the source layer identifier to match a last-minute change to the tileset. Fixed in c7ca0425ede884d88fe51d30d672ab84df3bd063.

PR #1614, opened 2018-08-17
Add incidents to any style

1ec5:
This is a proof of concept of using the runtime styling API to dynamically add the same incidents layers described in #1613 to any style, not just the forthcoming Mapbox-designed styles that come with incidents layers. This is the Streets v10 style showing incidents:

streets

This PR is based on a draft of an incidents-enabled style. Given that the navigation SDK’s default styles will come with incidents layers baked in, it probably won’t be necessary for the navigation SDK to add incidents to arbitrary styles – the SDK tries to be as style-agnostic as possible. As with the [iOS traffic plugin](https://github.com/mapbox/mapbox-plugins-ios/tree/master/TrafficPlugin), it may make sense for this code to be ported to Objective-C and made into a plugin, with an eye toward folding it into the map SDK along the lines of mapbox/mapbox-gl-native#12330.

Depends on #1613.

### DIFF
--- a/Examples/Swift/ViewController.swift
+++ b/Examples/Swift/ViewController.swift
@@ -258,6 +258,7 @@ class ViewController: UIViewController, MGLMapViewDelegate {
     
     func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
         self.mapView?.localizeLabels()
+        self.mapView?.showsIncidents = true
         
         if let routes = routes, let currentRoute = routes.first, let coords = currentRoute.coordinates {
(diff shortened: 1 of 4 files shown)


1ec5:
Updated for compatibility with the released Incidents v1 source and to match the released v4 navigation styles, as seen here on the Outdoors v10 and Light v9 styles:

outdoors light

Aside from a plugin or map SDK option, this PR could form the basis of a [navigation SDK example](https://github.com/mapbox/navigation-ios-examples/).

1ec5:
Closing; this PR was informational in nature anyways.

PR #1617, opened 2018-08-20
Expose additional properties on StyleManager

### LINKED ISSUE #1603: Feature request: Make StyleManager.automaticallyAdjustsStyleForTimeOfDay public
When building custom navigation experiences, StyleManager.automaticallyAdjustsStyleForTimeOfDay needs to be public.

frederoni:
Fixes #1603

This change exposes `automaticallyAdjustsStyleForTimeOfDay`, `delegate`, and
`styles` on StyleManager. It also fixes a bug which could lead to duplicated observers where setting `automaticallyAdjustsStyleForTimeOfDay` called `resumeNotifications()` instead of `resetTimeOfDayTimer()`.

@mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Resolved the partially styled user interface issue that occurs when the style theme is switched between the `NightStyle` and `DayStyle`, after the resume  button was tapped. ([#1589](https://github.com/mapbox/mapbox-navigation-ios/pull/1589))
 * Fixed a rare crash associated with the keyboard when showing the end of route view controller. [#1599](https://github.com/mapbox/mapbox-navigation-ios/pull/1599/)
 * Added a `MapboxVoiceController.audioPlayer` property. You can use this property to interrupt a spoken instruction or adjust the volume. [#1596](https://github.com/mapbox/mapbox-navigation-ios/pull/1596)
+* Added `StyleManager.automaticallyAdjustsStyleForTimeOfDay`, `StyleManager.delegate`, and `StyleManager.styles` properties so that you can control same time-based style switching just as NavigationViewController does. [#1617](https://github.com/mapbox/mapbox-navigation-ios/pull/1617)
 
 ## v0.19.1 (August 15, 2018)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Resolved the partially styled user interface issue that occurs when the style theme is switched between the `NightStyle` and `DayStyle`, after the resume  button was tapped. ([#1589](https://github.com/mapbox/mapbox-navigation-ios/pull/1589))
 * Fixed a rare crash associated with the keyboard when showing the end of route view controller. [#1599](https://github.com/mapbox/mapbox-navigation-ios/pull/1599/)
 * Added a `MapboxVoiceController.audioPlayer` property. You can use this property to interrupt a spoken instruction or adjust the volume. [#1596](https://github.com/mapbox/mapbox-navigation-ios/pull/1596)
+* Exposed `automaticallyAdjustsStyleForTimeOfDay`, `delegate`, and `styles` on `StyleManager` publicly. [#1617](https://github.com/mapbox/mapbox-navigation-ios/pull/1617)
```
1ec5: From the developer’s point of view, these properties never existed in the first place. Also, qualify each property reference so jazzy can automatically link to the documentation when the changelog appears in the docset cover:

> Added `StyleManager.automaticallyAdjustsStyleForTimeOfDay`, `StyleManager.delegate`, and `StyleManager.styles` properties so that you can control same time-based style switching just as `NavigationViewController` does.

MapboxNavigation/StyleManager.swift
```diff
@@ -29,17 +29,40 @@ public protocol StyleManagerDelegate: NSObjectProtocol {
...
+    @objc public weak var delegate: StyleManagerDelegate?
     
-    internal var date: Date?
-    var currentStyleType: StyleType?
-    var styles = [Style]() { didSet { applyStyle() } }
-    var automaticallyAdjustsStyleForTimeOfDay = true {
+    /**
+     Determines whether the style manager should apply a new style given the time of day.
+     
+     - note: Two styles must be provided for this property to have any effect.
```
1ec5: Nit: this doesn’t rise to the level of a `note`. `precondition` seems more appropriate, in any case.

PR #1618, opened 2018-08-23
Navigation Native, Part 2

frederoni:
cc @JThramer 

This branch is targeting ~~`jerrad/146-navigation-service`~~ epic/nav-native.


Cartfile
```diff
@@ -1,5 +1,6 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
-github "mapbox/MapboxDirections.swift" ~> 0.23.0
+binary "https://s3.amazonaws.com/mapbox/mapbox-navigation-native/ios/builds/Mapbox-iOS-Navigation-Native.json"
+github "mapbox/MapboxDirections.swift" "return-response"
```
1ec5: Release a new version of this library before releasing the SDK.

MapboxCoreNavigation/MBFixLocation.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+import MapboxNavigationNative
+import CoreLocation
+
+extension MBFixLocation {
+    
+    func asCLLocation() -> CLLocation {
```
1ec5: Similarly, this should be an initializer on an extension of CLLocation.

MapboxCoreNavigation/NativeRouteController.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+import MapboxNavigationNative
+import MapboxDirections
+
+
+public class NativeRouteController: RouteController {
```
1ec5: This public class doesn’t bridge to Objective-C; should it?

1ec5: All the route controllers are “native”, least of all this one. Since this class is public, can we call it something more descriptive and less misleading? How about `CrossPlatformRouteController` or `PortableRouteController`?

MapboxCoreNavigation/NativeRouteController.swift
```diff
@@ -0,0 +1,79 @@
...
+        }
+        set {
+            routeProgress = RouteProgress(route: newValue)
+            updateNavigator()
+        }
+    }
+    
+    func updateNavigator() {
+        guard let json = route.json else {
+            assertionFailure("Missing json route")
```
1ec5: This is no better than an `assert(_:)` followed by a force-unwrap without a guard.

MapboxCoreNavigation/NativeRouteController.swift
```diff
@@ -0,0 +1,79 @@
...
+            return
+        }
+        
+        let data = try! JSONSerialization.data(withJSONObject: json, options: [])
+        let jsonString = String(data: data, encoding: .utf8)!
+        
+        // TODO: Add support for alternative route
```
1ec5: Remember to ticket out any to-do items.

MapboxCoreNavigation/NativeRouteController.swift
```diff
@@ -0,0 +1,79 @@
...
+            complete(mostSimilar ?? routes.first, potentialError)
+        }
+    }
+    
+    override public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        locations.forEach { navigator.updateLocation(for: $0.asMBFixLocation()) }
+        super.locationManager(manager, didUpdateLocations: locations)
+    }
+    
+    override public func userIsOnRoute(_ location: CLLocation) -> Bool {
```
1ec5: `location` isn’t a route. This method should be called `userIsProceeding(along:)` or `userIsKeeping(to:)`.

MapboxCoreNavigation/NativeRouteController.swift
```diff
@@ -0,0 +1,79 @@
...
+        let status = navigator.getStatusForTimestamp(location.timestamp)
+        return status.routeState.isOnRoute()
+    }
+}
+
+extension MBRouteState {
+    
+    func isOnRoute() -> Bool {
```
1ec5: Can this be a computed property? Or does it have to be a method because this is an extension on an enumeration?

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -12,43 +12,123 @@ public enum SimulationOption: Int {
...
+     */
     var directions: Directions { get }
+    
+    /**
+     The active router, responsible for all route-following.
+     */
     var router: Router! { get }
+    
+    /**
+     The EventsManager, responsible for all telemetry.
```
1ec5: Nit: “the events manager”. The technical class name is already part of the declaration. This is where we translate the class name into plain English.

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -21,6 +21,7 @@ open class NavigationRouteOptions: RouteOptions {
             return $0
         }, profileIdentifier: profileIdentifier)
         includesAlternativeRoutes = true
+        shapeFormat = .polyline6
```
1ec5: Does MapboxNavigationNative handle polyline 5 format gracefully? If not, this is the first property being set here that the SDK strictly requires. Perhaps we should assert on the use of polyline 6 format or just go ahead and set it at the point of use in addition to here.

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -12,43 +12,123 @@ public enum SimulationOption: Int {
...
+     */
     var eventsManager: EventsManager! { get }
+    
+    /**
+     The route being progressed.
```
1ec5: > The route along which the user is expected to travel.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -106,8 +106,7 @@ open class RouteController: NSObject, Router {
      - parameter directions: The Directions object that created `route`.
      - parameter source: The data source for the RouteController.
      */
-    @objc(initWithRoute:directions:locationManager:)
-    public init(along route: Route, directions: Directions = Directions.shared, dataSource source: RouterDataSource) {
+    required public init(along route: Route, directions: Directions = Directions.shared, dataSource source: RouterDataSource) {
```
1ec5: This removes Objective-C compatibility, which is a breaking change as long as the initializer remains public.

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -12,43 +12,123 @@ public enum SimulationOption: Int {
...
+     The active location simulator. Only used during `SimulationOption.always` and `SimluatedLocationManager.onPoorGPS`. If there is no simulation active, this property is `nil`.
+    */
+    private var simulatedLocationSource: SimulatedLocationManager?
+
+    /**
+     The simulation mode of the service.
+     
+     A setting of `.always` will simulate route progress at all times.
+     A setting of `.onPoorGPS` will enable simulation when we do not recieve a location update after the `poorGPSPatience` threshold has elapsed.
+     A setting of `.never` will never enable the location simulator, regardless of circumstances.
```
1ec5: These settings are already documented on SimulationOption, right?

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -12,43 +12,123 @@ public enum SimulationOption: Int {
...
+    
+    /**
+     The route being progressed.
+     */
     var route: Route { get set }
+    
+    /**
+     The simulation mode of the service.
+     */
     var simulationMode: SimulationOption { get }
```
1ec5: Why is the type `SimulationOption` but the property `simulationMode`?

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -61,25 +141,44 @@ public class MapboxNavigationService: NSObject, NavigationService {
...
         self.init(route: route, directions: nil, locationSource: nil, eventsManagerType: nil)
     }
-    
+    /**
+     Intializes a new `NavigationService`.
+     
+     - parameter route: The route to follow.
+     - parameter directions: The Directions object that created `route`.
+     - parameter locationSource: An optional override for the default `NaviationLocationManager`.
+     - parameter eventsManagerType: A type argument used for overriding the events manager.
```
1ec5: What is a “type argument”? I think this is descriptive enough:

> An optional events manager to use  tracking the route.

MapboxCoreNavigation/Router.swift
```diff
@@ -11,6 +11,10 @@ public protocol RouterDataSource {
 @objc public protocol Router: class, CLLocationManagerDelegate {
     @objc unowned var dataSource: RouterDataSource { get }
     @objc var delegate: RouterDelegate? { get set }
+    
+    @objc(initWithRoute:directions:locationManager:)
+    init(along route: Route, directions: Directions, dataSource source: RouterDataSource)
```
1ec5: Given that this initializer is exposed to Objective-C, does it need to be public as well?

MapboxCoreNavigationTests/CocoaPodsTest/PodInstall/Podfile
```diff
@@ -5,4 +5,5 @@ target 'PodInstall' do
   # The branch of MapboxNavigation and MapboxNavigation will be replaced by Bitrise to use the current branch.
   pod 'MapboxNavigation', :path => '../../../'
   pod 'MapboxCoreNavigation', :path => '../../../'
+  pod 'MapboxDirections.swift', :git => 'git@github.com:mapbox/MapboxDirections.swift.git', :branch => 'return-response'
```
1ec5: Remember to remove this explicit dependency once we’re back on an official release.

MapboxCoreNavigation/NativeRouteController.swift
frederoni: How about renaming the old one to LegacyRouteController and Native to just RouteController?

MapboxCoreNavigation.podspec
```diff
@@ -40,6 +40,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxCoreNavigation"
 
+  s.dependency "MapboxNavigationNative"
```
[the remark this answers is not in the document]
akitchen: yes, especially as it is under such active development

MapboxCoreNavigation/NativeRouteController.swift
1ec5: What would the migration path look like for existing clients of the SDK that are using RouteController directly? As long as there’s more than one route controller, I think it would be understandable to qualify the class with an additional word. If we plan to remove the existing RouteController entirely in this release, that would be a different story (with implications for the changelog).

MapboxCoreNavigation/RouteController.swift
JThramer: Oversight. There's no real reason why it can't be `@objc`.

MapboxCoreNavigation/NativeRouteController.swift
frederoni: I agree but `RouteController.userIsOnRoute(_:)` is public in the existing API so this would be a breaking change.

MapboxCoreNavigation/RouteController.swift
frederoni: It bridges to Obj-C because the `Router` protocol is annotated with `@objc`

MapboxCoreNavigation/Router.swift
frederoni: The `Router` protocol is public so this initializer is public as well despite the missing `public/open`.

Cartfile
akitchen: If MapboxDirections.swift is otherwise stable, can we knock this out earlier? Maybe @frederoni can do it while we're working on these other things in Pacific & Mountain time?

Cartfile.resolved
```diff
@@ -1,12 +1,13 @@
+binary "https://s3.amazonaws.com/mapbox/mapbox-navigation-native/ios/builds/Mapbox-iOS-Navigation-Native.json" "3.0.0"
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "4.4.1"
 github "CedarBDD/Cedar" "v1.0"
 github "Quick/Nimble" "v7.3.1"
 github "Quick/Quick" "v1.3.2"
 github "ceeK/Solar" "2.1.0"
-github "mapbox/MapboxDirections.swift" "v0.23.0"
+github "mapbox/MapboxDirections.swift" "9c72e5a7509d656e0d9767c6748ee3c8162f92e8"
```
akitchen: Same thing here, we need an official release

MapboxCoreNavigation.podspec
```diff
@@ -40,6 +40,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxCoreNavigation"
 
+  s.dependency "MapboxNavigationNative", "~> 3.0"
```
frederoni: Bump to 3.1.2 which includes Bitcode and various fixes.

JThramer: @frederoni It's actually resolving to 3.1.3. 

```diff
-binary "https://s3.amazonaws.com/mapbox/mapbox-navigation-native/ios/builds/Mapbox-iOS-Navigation-Native.json" "3.1.2"
+binary "https://s3.amazonaws.com/mapbox/mapbox-navigation-native/ios/builds/Mapbox-iOS-Navigation-Native.json" "3.1.3"
```

MapboxCoreNavigation/RouteController.swift
```diff
@@ -480,6 +481,7 @@ extension RouteController: CLLocationManagerDelegate {
 
     private func checkForUpdates() {
         #if TARGET_IPHONE_SIMULATOR
+        guard (NSClassFromString("XCTestCase") == nil) else { return } // Short-circuit when running unit tests
```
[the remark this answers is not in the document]
akitchen: We'll have a main bundle when tests are run by a "bundle loader" aka an iOS app. And env variables have to be duplicated across multiple schemes.

This is gross, and ultimately is a smell that this code needs to be better factored. I share the blame for the hotfix here (this came up when working to reduce network requests made by the unit tests)

Cartfile.resolved
akitchen: https://github.com/mapbox/MapboxDirections.swift/pull/298

Cartfile
```diff
@@ -1,5 +1,6 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
-github "mapbox/MapboxDirections.swift" ~> 0.23.0
+binary "https://s3.amazonaws.com/mapbox/mapbox-navigation-native/ios/builds/Mapbox-iOS-Navigation-Native.json"
```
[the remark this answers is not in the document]
1ec5: https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json isn’t working _just_ yet, but that’s the URL we want here.

akitchen:
> all the problems we think have gone away by migrating to the MapboxNavigationNative framework.

"benefits from a new off-route detection heuristic provided by Mapbox's cross-platform native navigation library"

?

/cc @mcwhittemore

mcwhittemore:
nav-native is providing route following, location filtering, guidance triggering and location projection support. I'd call this logic move out in the changelog definitely.

akitchen:
Let’s only call out the portions we’re using on iOS — as we move more functionality over we can call it out then

Cartfile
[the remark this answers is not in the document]
akitchen: It's working now, yes? resolved?

Cartfile
```diff
@@ -1,5 +1,5 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
-binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json"
+binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 3.1.4
```
1ec5: My understanding is that this library will respect semver, so it’s OK to float on the minor version as well as the patch version, just as we do in the map SDK.

/cc @kevinkreiser

1ec5: The podspec specifies `~> 3.0`, so let’s be consistent with that.

PR #1602, opened 2018-08-13
Navigation Service, Part 1

JThramer:
It's working!

Part 1 of the "Nav To The Future" initiative.

To-Do: 
- [x] Get tests working
- [x] refactor `EventsManager`
- [x] refactor `TunnelIntersectionManager` (now `TunnelAuthority`)
- [x] Fix weird CI bug where we're failing strings validation step
- [x]  refactor `TunnelIntersectionManagerTests`
- [x] PR Comments
- [x] refactor delegation protocols
- [ ] Update Docs
- [ ] Smoke Test
- [ ] Update `CHANGELOG.md`

/cc @mapbox/navigation-ios @mcwhittemore

PR #1624, opened 2018-08-24
Fix retain cycle if dead reckoning is enabled on RouteController.

### LINKED ISSUE #1623: Retain cycle in RouteController when isDeadReckoningEnabled option is enabled
Hello,

Currently if the isDeadReckoningEnabled option is set to true on a RouteController object and that at least one location update has been received, then a retain cycle is created and there is no way to deallocate the RouteController object.

This is due to *perform(_ aSelector: Selector, with anArgument: Any?, afterDelay delay: TimeInterval)* retaining the receiver and arguments.

**Mapbox Navigation SDK version:** v0.19.2

Woildan:
Proposed solution is to make *suspendLocationUpdates* cancel the next scheduled location interpolation. This way calling *suspendLocationUpdates* or *endNavigation* then releasing all references to the RouteController allows it to be released.

Note that releasing all references to the RouteController object without calling *suspendLocationUpdates* first still leads to a memory leak + parasites notifications being triggered, etc

fixes #1623

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -184,6 +184,7 @@ open class RouteController: NSObject, Router {
         locationManager.stopUpdatingLocation()
         locationManager.stopUpdatingHeading()
         locationManager.delegate = nil
+        NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(interpolateLocation), object: nil)
     }
     
     /**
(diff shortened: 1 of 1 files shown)


bsudekum:
> Note that releasing all references to the RouteController object without calling suspendLocationUpdates first still leads to a memory leak + parasites notifications being triggered, etc

@JThramer what's the current thinking on this? At one point, we used to call this in deinit, should this still be the case?

JThramer:
@bsudekum yeah, that makes sense, although it's worth pointing out that `interpolateLocation` is moving out of the `RouteController` in #1602.

PR #1632, opened 2018-08-29
Cache pod master repo on CircleCI

frederoni:
- [x] Cache the pod master repo on CircleCI for the pod integration test
- [x] Restore Carthage cache before running bootstrap

@mapbox/navigation-ios


bsudekum:
Nice, the objective here is just speed up CI?

frederoni:
> Nice, the objective here is just speed up CI?

Yes. Just transferring the Xcode 10 bot and CocoaPods integration test results in parallellizing those steps which cuts about 15 minutes. The end goal is probably to move all bots to CircleCI.

PR #1633, opened 2018-08-29
Public StepsListViewController

bsudekum:
This makes it possible to have the same steps list view functionality, that is found in the drop in UI, in a custom navigation UI. I think this is about the least intrusive change possible to have this work.

- [x] document

/cc @mapbox/navigation-ios


Examples/Swift/CustomViewController.swift
```diff
@@ -114,4 +118,53 @@ class CustomViewController: UIViewController, MGLMapViewDelegate {
...
+            
+            let controller = StepsViewController(routeProgress: routeController.routeProgress)
+            controller.delegate = self
+            addChildViewController(controller)
+            view.addSubview(controller.view)
+            
+            controller.view.topAnchor.constraint(equalTo: instructionsBannerView.bottomAnchor).isActive = true
+            controller.view.leadingAnchor.constraint(equalTo: view.leadingAnchor).isActive = true
+            controller.view.bottomAnchor.constraint(equalTo: view.bottomAnchor).isActive = true
+            controller.view.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
```
bsudekum: I'm not a fan of how much boilerplate is necessary to make the slide action happen. @frederoni is there anything we can do to encapsulate this better?

frederoni: You could make `UIView.pinInSuperview()` public and replace these 4 lines w/
`controller.view.pinInSuperview()`. But pinInSuperview doesn't respect RTL so make sure you change `left->leading` and `right->trailing` in `UIView.pinInSuperview()` if you decide to go that way.

PR #1635, opened 2018-08-29
Do not fire progress updates when the user is not moving

bsudekum:
If the incoming location is close to identical to the previous location, we can safely filter this location to prevent the map view from going through a render cycle. This is minor change but could add up for the amount of time people spend at stop lights.

To prevent UI elements from become outdated, I added a check to make sure that we're letting at least one progress update per minute.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/Constants.swift
+++ b/MapboxCoreNavigation/Constants.swift
@@ -180,3 +180,8 @@ public var RouteControllerMinimumDistanceForProgressNotification: CLLocationDist
  The minimum difference between the user's course before a location update is filtered out.
  */
 public var RouteControllerMinimumDifferenceBetweenCourse: CLLocationDirection = 1
+
+/**
+ The minimum elapsed time between location updates before a location update is filtered out.
+ */
+public var RouteControllerMinimumElapsedTimeBetweenLocations: TimeInterval = 60
--- a/MapboxCoreNavigation/Constants.swift
+++ b/MapboxCoreNavigation/Constants.swift
@@ -170,3 +170,13 @@ extension Notification.Name {
      */
     public static let routeControllerDidPassVisualInstructionPoint = MBRouteControllerDidPassVisualInstructionPoint
 }
+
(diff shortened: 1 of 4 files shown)


bsudekum:
Tests need some updating.

MapboxCoreNavigation/Constants.swift
```diff
@@ -170,3 +170,18 @@ extension Notification.Name {
...
+
+/**
+ The minimum difference between the user's course before a location update is filtered out.
+ */
+public var RouteControllerMinimumDifferenceBetweenCourse: CLLocationDirection = 1
+
+/**
+ The minimum elapsed time between location updates before a location update is filtered out.
+ */
+public var RouteControllerMinimumElapsedTimeBetweenLocations: TimeInterval = 60
```
1ec5: Do we actually want to throttle location updates to once a second? If we connect to an external GPS device or car, wouldn’t we want to get the more frequent location updates? Or do we think such updates will rarely also look identical in terms of location and course?

bsudekum: So we would only filter when all three cases are met:

* The new location is within 1 meter of the previous location.
* The new course is within 1 degree of the previous course.
* The new location is less than 60 seconds old.

bsudekum: I think if location updates are coming from a high frequency device, and they are very close in terms of proximity, this is even more of a reason to throttle progress events.

bsudekum:
@frederoni @1ec5 could I get one more set of 👀 on this?

bsudekum:
We might actually want to merge and test this post v0.20.0.

1ec5:
Although this PR doesn’t strictly conflict with the recent integration of MapboxNavigationNative, this kind of logic should in principle be handled by MapboxNavigationNative, which assumes it gets all the location updates that come in. Closing as stale.

/cc @mapbox/voyageurs

PR #1639, opened 2018-08-30
Flip roundabout if it's left-hand rule of the road

### LINKED ISSUE #1637: Round about icon direction

0.19.0

**Steps to reproduce:**

1. Run example-swift app
2. Select a route with roundabout
3. Turn on navigation simulation
4. Start default UI navigation
5. Drive through roundabout

**Expected:**

Visual instruction icon should show direction to left

**Actual:**

img_1071

Reproduced in: Lithuania
Driving direction: Right

frederoni:
Fixes #1637 Unflip roundabout icons

Looks like this regressed in #1514 

cc @vincethecoder @bsudekum @1ec5 @margvl


margvl:
I think this issue can effect uTurn as well but I have not tested it

frederoni:
> I think this issue can effect uTurn as well but I have not tested it

@margvl you're right, u-turns were drawn according to the opposite rule of the road as well. Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/1639/commits/fb87d1cc7af14a14e9d0fc1bd58b7f6ab7cc1222

Issue #1634, opened 2018-08-29
Regression: Mapbox Voice Instructions


JThramer:
Hey all,

It looks like we introduced a regression onto `master` as part of #1596. It looks like an `NSException` is being thrown when we try to [prepare the audio player for playing.](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/MapboxVoiceController.swift?utf8=✓#L225)

### Console Log
```
2018-08-29 14:22:15.439976-0600 Example-Swift[56466:8459466] [AudioHAL_Client] HALC_ProxyIOContext.cpp:1399:IOWorkLoop:  HALC_ProxyIOContext::IOWorkLoop: failed to send the final message to the server, Error: 0x10000003
2018-08-29 14:22:15.451383-0600 Example-Swift[56466:8459641] [AudioHAL_Client] HALC_ProxyIOContext.cpp:959:IOWorkLoop:  HALC_ProxyIOContext::IOWorkLoop: failed to get the IOPort
2018-08-29 14:22:30.504840-0600 Example-Swift[56466:8459164] 239: CA_UISoundClientBase::StartPlaying: AddRunningClient failed (status = -66681).
```
### Expected Behavior:
* Polly to speak voice instructions.

### Observed Behavior:
* In `Example-Swift`, Polly speaks the first instruction, `AVSpeechSynthesizer` speaks the remaining voice instructions for the route.
* In `Example-Objective-C`, both voice controllers speak simultaneously.

/cc @mapbox/navigation-ios

bsudekum:
I'm unable to reproduce this on master. @JThramer let me know if you have any clues as to how to trigger this error.

JThramer:
Yeah there is definitely a weird data-race going on, @bsudekum. I *just* got a run where both voices spoke simultaneously  (in `Example-Swift`, no less), but right as I go to debug, the LLDB RPC server crashes.

MFW: 
![seriously?](https://user-images.githubusercontent.com/887225/44817164-cb147500-aba1-11e8-985c-2f8abc683d78.gif)


Still investigating.

JThramer:
@bsudekum I definitely found the source of the "[doubling up](https://github.com/mapbox/mapbox-navigation-ios/blob/master/Examples/Objective-C/ViewController.m#L69)" of voices on the OBJ-C sample. The rest I'm thinking could be cache pollution, but it could also be mental fatigue. 😅

PR #1640, opened 2018-08-30
Objective-C Example Bug: Doublespeak

JThramer:
Two-Minutes Hate, from "1984".
Nobody likes it when two voices talk over each-other. In support of #1634.

* un-breaking problem where two synth-voices make double-speak is double-plus-ungood.
* Also fixed an issue where navigator is never resumed when returning from background because why not.


JThramer:
cc @mapbox/navigation-ios, apologies! 😅

Examples/Objective-C/ViewController.m
```diff
@@ -28,10 +27,11 @@ - (void)viewDidLoad {
     self.lengthFormatter = [[NSLengthFormatter alloc] init];
     self.lengthFormatter.unitStyle = NSFormattingUnitStyleShort;
     
-    self.speechSynth = [[AVSpeechSynthesizer alloc] init];
-    self.speechSynth.delegate = self;
     [self resumeNotifications];
 }
+- (void)viewDidAppear:(BOOL)animated {
+    [self.navigation resume];
```
frederoni: nit: call super
```markdown
Discussion
You can override this method to perform additional tasks associated with presenting the view. If you override this method, you must call super at some point in your implementation.
```

Cartfile.resolved
```diff
@@ -7,4 +7,4 @@ github "mapbox/mapbox-events-ios" "v0.4.51"
 github "mapbox/mapbox-voice-swift" "v0.0.1"
 github "mapbox/turf-swift" "v0.2.0"
 github "raphaelmor/Polyline" "v4.2.0"
-github "uber/ios-snapshot-test-case" "3.0.0"
+github "uber/ios-snapshot-test-case" "3.1.0"
```
frederoni: Can we discard this update since this PR doesn't touch any snapshot tests?

PR #1644, opened 2018-08-31
Restore map template when exiting navigation

1ec5:
Restore the map template’s buttons when exiting navigation via an additional intermediate class around CPMapTemplate. This keeps the CarPlay interface usable after exiting navigation.

There are still a bunch of code smells in this PR, not least of which is the fact that the navigation session is begin within the navigation view controller’s initializer. NavigationMapTemplateController started out as a subclass of CPMapTemplate, since it isn’t possible for the interface controller to present a CPMapTemplate subclass, but it might end up getting folded back into CarPlayNavigationViewController.

Also renamed some CarPlayNavigationViewController properties.

/cc @frederoni @bsudekum

### DIFF
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -5,6 +5,96 @@ import MapboxCoreNavigation
 import CarPlay
 
 @available(iOS 12.0, *)
+protocol NavigationMapTemplateControllerDelegate: class {
+    func navigationMapTemplateController(_ navigationMapTemplateController: NavigationMapTemplateController, willHandle mapButton: CPMapButton)
+    func navigationMapTemplateController(_ navigationMapTemplateController: NavigationMapTemplateController, willHandle barButton: CPBarButton)
+}
+
+@available(iOS 12.0, *)
+public class NavigationMapTemplateController {
+    var mapTemplate: CPMapTemplate
+    weak var delegate: NavigationMapTemplateControllerDelegate?
+    weak var mapDelegate: CPMapTemplateDelegate? {
+        get {
+            return mapTemplate.mapDelegate
+        }
+        set {
+            mapTemplate.mapDelegate = mapDelegate
+        }
+    }
+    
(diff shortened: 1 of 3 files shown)


akitchen:
Testing this out manually, the steps table no longer follows along during simulated navigation. 

Also, the buttons didn't actually restore themselves when the user arrives and navigation is ended.

However, tapping "End" *during* navigation ends nav and restores buttons as expected, although the route is still displayed.

akitchen:
https://github.com/mapbox/mapbox-navigation-ios/pull/1645 will help with the test failures

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -4,6 +4,96 @@ import MapboxCoreNavigation
...
+    var recenterButton: CPMapButton!
+    
+    var exitButton: CPBarButton!
+    var muteButton: CPBarButton!
+    
+    init(mapTemplate: CPMapTemplate) {
+        self.mapTemplate = mapTemplate
+        previousMapButtons = mapTemplate.mapButtons
+        previousLeadingNavigationBarButtons = mapTemplate.leadingNavigationBarButtons
+        previousTrailingNavigationBarButtons = mapTemplate.trailingNavigationBarButtons
```
bsudekum: should we allow developers to mutate `self.previousTrailingNavigationBarButtons` after init?

1ec5: The idea here was to avoid unnecessarily asking the delegate to set up the buttons that were already on the navigation bar, whatever they were. (It’s possible for the user to enter route selection from at least two different screens, so there could in theory be two different sets of navigation bar buttons depending on the flow.) If we were to make these buttons mutable during the navigation session, then I guess this class would ask CarPlayNavigationViewController to ask its delegate for the buttons.

1ec5: Going to merge this PR as a first step and streamline button management directly on the carplay branch as tail work.

PR #1645, opened 2018-08-31
Permit routeless events manager for unique telemetry events

vincethecoder:
Provided optional `sessionState` and `routeController` for specific telemetry events where these fields are not required.

/cc @frederoni @1ec5 @akitchen

### DIFF
--- a/MapboxCoreNavigation/EventsManager.swift
+++ b/MapboxCoreNavigation/EventsManager.swift
@@ -79,17 +79,21 @@ open class EventsManager: NSObject {
             return nil
         }
         
+        guard var event = EventDetails.defaultEvents(router: routeController) else {
+            return nil
+        }
+        
         let rating = potentialRating ?? MMEEventsManager.unrated
-        var event = EventDetails.defaultEvents(router: routeController)
-        event?.event = MMEEventTypeNavigationCancel
-        event?.arrivalTimestamp = sessionState?.arrivalTimestamp
+        
+        event.event = MMEEventTypeNavigationCancel
+        event.arrivalTimestamp = sessionState?.arrivalTimestamp
         
         let validRating: Bool = (rating >= MMEEventsManager.unrated && rating <= 100)
         assert(validRating, "MMEEventsManager: Invalid Rating. Values should be between \(MMEEventsManager.unrated) (none) and 100.")
         guard validRating else { return event }
         
-        event?.rating = rating
-        event?.comment = comment
+        event.rating = rating
+        event.comment = comment
         
         return event
     }
@@ -99,8 +103,11 @@ open class EventsManager: NSObject {
             return nil
         }
         
-        var event = EventDetails.defaultEvents(router: routeController)
-        event?.event = MMEEventTypeNavigationDepart
+        guard var event = EventDetails.defaultEvents(router: routeController) else {
+            return nil
+        }
+        
+        event.event = MMEEventTypeNavigationDepart
         return event
     }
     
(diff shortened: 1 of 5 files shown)


MapboxCoreNavigation/EventsManager.swift
```diff
@@ -74,68 +74,92 @@ open class EventsManager: NSObject {
...
     }
     
-    func navigationCancelEvent(rating potentialRating: Int? = nil, comment: String? = nil) -> EventDetails {
+    func navigationCancelEvent(rating potentialRating: Int? = nil, comment: String? = nil) -> EventDetails? {
+        guard let routeController = routeController else {
+            return nil
+        }
+        
         let rating = potentialRating ?? MMEEventsManager.unrated
         var event = EventDetails.defaultEvents(router: routeController)
```
akitchen: Can this be moved up to the guard statement in order to avoid the `?`'s below?

MapboxCoreNavigation/EventsManager.swift
```diff
@@ -74,68 +74,92 @@ open class EventsManager: NSObject {
...
         return eventDictionary
     }
     
-    func navigationFeedbackEvent(type: FeedbackType, description: String?) -> EventDetails {
+    func navigationFeedbackEvent(type: FeedbackType, description: String?) -> EventDetails? {
+        guard let routeController = routeController else {
+            return nil
+        }
+        
         var event = EventDetails.defaultEvents(router: routeController)
```
akitchen: Same as above; maybe this can move into the guard statement?

MapboxCoreNavigation/EventsManager.swift
```diff
@@ -74,68 +74,92 @@ open class EventsManager: NSObject {
...
     
-    func navigationRerouteEvent(eventType: String = MMEEventTypeNavigationReroute) -> EventDetails {
+    func navigationRerouteEvent(eventType: String = MMEEventTypeNavigationReroute) -> EventDetails? {
+        guard let routeController = routeController else {
+            return nil
+        }
+        
         let timestamp = Date()
         
         var event = EventDetails.defaultEvents(router: routeController)
```
akitchen: Same. it's too many `?`'s for quick readability IMO

MapboxCoreNavigation/EventsManager.swift
```diff
@@ -275,21 +312,17 @@ extension EventsManager {
...
     @objc private func didChangeApplicationState(_ notification: NSNotification) {
-        if sessionState != nil {
-            sessionState.reportChange(to: UIApplication.shared.applicationState)
-        }
+        sessionState?.reportChange(to: UIApplication.shared.applicationState)
     }
     
     @objc private func applicationWillTerminate(_ notification: NSNotification) {
-        if !sessionState.terminated {
+        if let session_state = sessionState, !session_state.terminated {
```
akitchen: This needs some stylistic cleanup. Might as well use the local non-optional below if we can acquire it.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -59,14 +59,16 @@ open class RouteController: NSObject, Router {
     @objc public var routeProgress: RouteProgress {
         willSet {
             // Save any progress completed up until now
-            eventsManager.sessionState.totalDistanceCompleted += routeProgress.distanceTraveled
+            if let _ = eventsManager.sessionState {
```
akitchen: Can we take a local var to the `sessionState` instead?

MapboxCoreNavigation/EventsManager.swift
vincethecoder: sure can.

PR #1646, opened 2018-08-31
Test Quieting

### LINKED ISSUE #1634: Regression: Mapbox Voice Instructions

JThramer:
Hush now.
Quieting the tests. Closes #1634.

* Fixing issue where Voice engine regularly speaks during testing

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/MapboxVoiceController.swift
+++ b/MapboxNavigation/MapboxVoiceController.swift
@@ -28,15 +28,17 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
     
     var audioTask: URLSessionDataTask?
     var cache: BimodalDataCache
+    let audioPlayerType: AVAudioPlayer.Type
     
     var speech: SpeechSynthesizer
     var locale: Locale?
     
     let localizedErrorMessage = NSLocalizedString("FAILED_INSTRUCTION", bundle: .mapboxNavigation, value: "Unable to read instruction aloud.", comment: "Error message when the SDK is unable to read a spoken instruction.")
 
-    @objc public init(speechClient: SpeechSynthesizer = SpeechSynthesizer(accessToken: nil), dataCache: BimodalDataCache = DataCache()) {
+    @objc public init(speechClient: SpeechSynthesizer = SpeechSynthesizer(accessToken: nil), dataCache: BimodalDataCache = DataCache(), audioPlayerType: AVAudioPlayer.Type? = nil) {
         speech = speechClient
         cache = dataCache
+        self.audioPlayerType = audioPlayerType ?? AVAudioPlayer.self
         super.init()
         
         audioPlayer?.delegate = self
@@ -221,7 +223,7 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
         audioQueue.async { [weak self] in
             guard let `self` = self else { return }
             do {
-                self.audioPlayer = try AVAudioPlayer(data: data)
+                self.audioPlayer = try self.audioPlayerType.init(data: data)
                 self.audioPlayer?.prepareToPlay()
                 self.audioPlayer?.delegate = self
                 try self.duckAudio()
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -231,7 +231,7 @@ open class NavigationViewController: UIViewController {
      
      See `RouteVoiceController` for more information.
      */
-    @objc public lazy var voiceController: RouteVoiceController? = MapboxVoiceController()
+    @objc public var voiceController: RouteVoiceController!
     
     /**
      Provides all routing logic for the user.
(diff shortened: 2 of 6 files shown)


MapboxNavigation/MapboxVoiceController.swift
```diff
@@ -28,15 +28,17 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
...
     var speech: SpeechSynthesizer
     var locale: Locale?
     
     let localizedErrorMessage = NSLocalizedString("FAILED_INSTRUCTION", bundle: .mapboxNavigation, value: "Unable to read instruction aloud.", comment: "Error message when the SDK is unable to read a spoken instruction.")
 
-    @objc public init(speechClient: SpeechSynthesizer = SpeechSynthesizer(accessToken: nil), dataCache: BimodalDataCache = DataCache()) {
+    @objc public init(speechClient: SpeechSynthesizer = SpeechSynthesizer(accessToken: nil), dataCache: BimodalDataCache = DataCache(), audioPlayerType: AVAudioPlayer.Type? = nil) {
         speech = speechClient
         cache = dataCache
+        self.audioPlayerType = audioPlayerType ?? AVAudioPlayer.self
```
akitchen: Should this default be declared in the initializer instead?

MapboxNavigation/MapboxVoiceController.swift
```diff
@@ -221,7 +223,7 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
         audioQueue.async { [weak self] in
             guard let `self` = self else { return }
             do {
-                self.audioPlayer = try AVAudioPlayer(data: data)
+                self.audioPlayer = try self.audioPlayerType.init(data: data)
```
akitchen: We can do this as follow-on work, but perhaps we should declare a type/protocol that we can also implement in a test double in order to verify data is sent through correctly

MapboxNavigation/NavigationViewController.swift
```diff
@@ -331,19 +331,20 @@ open class NavigationViewController: UIViewController {
...
+                         voiceController: RouteVoiceController? = nil) {
         
         super.init(nibName: nil, bundle: nil)
         
         self.routeController = RouteController(along: route, directions: directions, locationManager: locationManager ?? NavigationLocationManager())
         self.routeController.usesDefaultUserInterface = true
         self.routeController.delegate = self
         self.routeController.tunnelIntersectionManager.delegate = self
-        
+        self.voiceController = voiceController ?? MapboxVoiceController()
```
akitchen: Seems like this should follow the pattern of being declared in the initializer, like the others

MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -13,14 +13,16 @@ class NavigationViewControllerTests: XCTestCase {
     
     var updatedStyleNumberOfTimes = 0
     
-    lazy var dependencies: (navigationViewController: NavigationViewController, startLocation: CLLocation, poi: [CLLocation], endLocation: CLLocation) = {
+    lazy var dependencies: (navigationViewController: NavigationViewController, startLocation: CLLocation, poi: [CLLocation], endLocation: CLLocation, voice: RouteVoiceController) = {
```
akitchen: This is cool for now, but I think this lazy var pattern is outliving its usefulness. Also, doesn't it hang around in between test runs? Or does XCTest create a new instance of the test class for each test method, as JUnit does? I forget...

Nonetheless, we're going to want to unwind/refactor this (not necessarily in this PR)

akitchen:
Looks like the failures are related to needing to update the ObjC example app ...

MapboxNavigation/MapboxVoiceController.swift
JThramer: I generally like to take this approach to default-arguments for two reasons:

1. It prevents you from having to statically determine "will I or won't I override?" at the call-site. It's much easier to decide to pass nil or not at runtime then it is to figure out overridability logic at compile-time. Basically if you expose the argument in the call, but you still want the default logic, you have to recreate it. Not so with optional defaults.
2. It's way, _way_ more ergonomic for an OBJ-C dev to simply pass "nil" if they want default logic than it is for them to recreate default behavior, and it prevents us from having to make OBJ-C wrapper functions/inits all over the place.

akitchen:
Good points. Should we make this more consistent then?

MapboxNavigation/MapboxVoiceController.swift
akitchen: Good points. Should we make this pattern more consistent? Or do the existing args pass the smell test?

It's probably beyond the scope of this PR though

akitchen:
the test suite still plays audio for me, although less than before. there are some uncontrolled usages as side-effects in other tests that we'll want to also think about. We have the same problem in other areas, it's a general dependency management / inversion of control issue

akitchen:
FWIW, this was with this code merged into another branch, so there could be something else going on

PR #1647, opened 2018-09-03
Fix u-turn direction

frederoni:
screenshot 2018-09-03 20 16 52

The arrow-180 asset used for u-turns in [navigation-ui-resources](https://github.com/mapbox/navigation-ui-resources) is turning clockwise so we should flip it horizontally if it's right-hand rule of the road.

cc @akitchen @1ec5


MapboxNavigationTests/ManeuverViewTests.swift
```diff
@@ -69,3 +88,13 @@ class ManeuverViewTests: FBSnapshotTestCase {
     // TODO: Figure out why the flip transformation do not render in a snapshot so we can test left turns and left side rule of the road.
 }
 
+extension UIView {
+    
+    func asImage() -> UIImage {
```
1ec5: Consider replacing this method with an initializer in a UIImage extension. That style is less “fluent” but also more consistent with the majority of the standard library.

frederoni: I agree but initializers on UIImage would require more steps and it's not particularly elegant since you can't return the instance. It would involve UIView->UIImage->CGImage(or any other type supported by UIImage->UIImage. So I went with a class function on UIImage instead, `UIImage.from(_:)`.

PR #1653, opened 2018-09-05
Also move user puck during course tracking if necessary

### LINKED ISSUE #1651: User Puck Shifted Vertically from true anchor point
Hey all,

I've seen/heard a lot of us talk about a hard-to-reproduce issue where the user-puck "shifts" up or down w.r.t. the actual anchor point. Seems like a good idea to go ahead and cut a ticket to start tracking anecdotal stories of reproduction.

Here are my notes from trying to dig into this by breaking in the `NavigationMapView.userAnchorPoint`  computed method:

| Property  | Good Run  | Bad Run |
| ------------- | ------------- | ------------ |
|` contentInset`  | `{116, 0, 80, 0}`  | `{116, 0, 80, 0}` |
| `frame.size`  | `{75, 75}`  | `{75, 75}` |
| `frame.origin` | `{169.11, 454.99}` | `{169.5, 510.5}`|

Feel free to tack onto this ticket as needed.

/cc @mapbox/navigation-ios

1ec5:
During course tracking, ensure that the user puck moves along with the camera the first time `NavigationMapView.updateCourseTracking(location:animated:)` gets called after adjusting the map view’s layout.

Fixes #1651.

/cc @JThramer @bsudekum


1ec5:
Before, in CarPlay (on the `carplay` branch), the puck would either stay put as the route line shifts to the right to accommodate the maneuver overlay, or it would disappear entirely:

detached
rotated
invisible

This PR synchronizes the puck with the camera and thus the route line:

attached

PR #1659, opened 2018-09-07
Simplify presenting steps from CarPlay to device

frederoni:
We don't the NavigationViewController in the stack when we're navigating with CarPlay.

cc @akitchen


akitchen:
agree that this is much simpler, but I fear it will complicate the case where the user unplugs and we want to continue/resume navigation on the phone... thoughts?

akitchen:
(... because we would no longer be setting up a NavigationViewController)

frederoni:
> continue/resume navigation on the phone

You're right. This should be possible. Perhaps even switching from transport to walking directions if that happens when you exit the car.

PR #1660, opened 2018-09-07
Refine navigation among CarPlay templates

### LINKED ISSUE #1650: Move simulate button from favorites list to route selection
The Simulate / Don’t Simulate toggle button should appear in the map template when previewing routes, not in the favorites list. UI for toggling location simulation is specific to a demo application use case, which is why the Swift example application implements it. But currently if the application implements `CarPlayManagerDelegate.carPlayManager(_:leadingNavigationBarButtonsCompatibleWith:in:)` and `carPlayManager(_:trailingNavigationBarButtonsCompatibleWith:in:)`, it also has to roll its own search and favorites buttons, and the creation of those buttons is currently tangled up in CarPlayManager.

https://github.com/mapbox/mapbox-navigation-ios/blob/d5062353bcf541f07aede967a9c862e7e0955caf/MapboxNavigation/CarPlayManager.swift#L137-L142 https://github.com/mapbox/mapbox-navigation-ios/blob/d5062353bcf541f07aede967a9c862e7e0955caf/MapboxNavigation/CarPlayManager.swift#L147-L149

/cc @frederoni @vincethecoder

1ec5:
Added a back button to the CarPlay route preview screen. The key is to push a new map template onto the stack for previewing rather than blowing away the stack. (The map templates all share the same map view controller; this is only about managing overlays and such.) The same approach might allow us to effortlessly push and pop a navigation-centric map template as an alternative to #1644.

The route preview items have been refined to more closely match Maps. Each route is summarized by the expected travel time and secondarily by the major road names. Setting the major road names in a smaller font allows more of each name to be visible. The user can still compare travel times using the absolute travel times.

The Simulate / Don’t Simulate toggle button has been moved to the route preview screen, making it more discoverable than when it was in the Favorites list.

route overview
route selection

This PR introduces a serious regression: a back button appears during turn-by-turn. Tapping it leads to the search or list template, and tapping cancel from there leads to a map template that still looks like it’s handling turn-by-turn navigation.

back
stuck

Fixes #1650.

/cc @akitchen @vincethecoder

### DIFF
--- a/Examples/Swift/AppDelegate.swift
+++ b/Examples/Swift/AppDelegate.swift
@@ -95,7 +95,7 @@ extension AppDelegate: CarPlayManagerDelegate {
     }
     
     func carPlayManager(_ carPlayManager: CarPlayManager, trailingNavigationBarButtonsCompatibleWith traitCollection: UITraitCollection, in template: CPTemplate) -> [CPBarButton]? {
-        guard let template = template as? CPListTemplate, template.title == "Favorites List" else {
+        guard let template = template as? CPMapTemplate, template != carPlayManager.interfaceController?.rootTemplate else {
             return nil
         }
         
--- a/Examples/Swift/AppDelegate.swift
+++ b/Examples/Swift/AppDelegate.swift
@@ -94,8 +94,8 @@ extension AppDelegate: CarPlayManagerDelegate {
         }
     }
     
-    func carPlayManager(_ carPlayManager: CarPlayManager, trailingNavigationBarButtonsCompatibleWith traitCollection: UITraitCollection, in template: CPTemplate) -> [CPBarButton]? {
-        guard let template = template as? CPMapTemplate, template != carPlayManager.interfaceController?.rootTemplate else {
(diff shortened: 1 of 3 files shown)


1ec5:
> The same approach might allow us to effortlessly push and pop a navigation-centric map template as an alternative to #1644.

b6dbe9f53330c7faef0f5fc7ed38142710ea6b6e overhauls ownership of map templates, particularly the navigation map template. CarPlayMapTemplateController has been removed. CarPlayManager is now directly responsible for setting up the navigation map template, starting the navigation session, and responding to gestures on the navigation view controller’s map (with help from the navigation view controller). The overview and recentering buttons have been consolidated into a single toggle button.

> This PR introduces a serious regression: a back button appears during turn-by-turn. Tapping it leads to the search or list template, and tapping cancel from there leads to a map template that still looks like it’s handling turn-by-turn navigation.

Also fixed in b6dbe9f53330c7faef0f5fc7ed38142710ea6b6e. Making CarPlayManager responsible for all map templates made it possible to clear the stack before pushing the navigation map template.

CarPlayManagerDelegate is now given an opportunity to add navigation bar buttons to the navigation map template alongside the built-in ones. An argument has been added to CarPlayManagerDelegate methods to distinguish between the three map templates.

1ec5:
The only failing test is `testUsingURLStringAsCacheKey`, which is also failing on the carplay branch. Merging.

PR #1661, opened 2018-09-08
Moves route simulation configuration into CarPlayManager, with tests

akitchen:
The tests illustrate a simple BDD example for the sake of discussion (test-driven, no less)

/cc @mapbox/navigation-ios


1ec5:
Wonder if it makes sense to expose a similar API in NavigationViewController, since location simulation isn’t specific to CarPlay.

akitchen:
yes, it probably does make sense to be consistent in how we manage this,
but I'm not sure of whether to make it a separate PR since it's a bit of a
detour from CarPlay work

Also, these changes afford us an opportunity to encapsulate the default
button behavior elsewhere, which feels right.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -103,17 +103,20 @@ public class CarPlayManager: NSObject, CPSearchTemplateDelegate {
     public fileprivate(set) var mainMapTemplate: CPMapTemplate?
     public fileprivate(set) weak var currentNavigator: CarPlayNavigationViewController?
 
+    public var simulatesLocations = false
+    public var simulatedSpeedMultipler = 1.0
```
1ec5: These properties should be documented.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -570,64 +573,74 @@ extension CarPlayManager: CPMapTemplateDelegate {
...
             facing = 270
         }
-        
+
         let newCenter = camera.centerCoordinate.coordinate(at: CarPlayMapViewPanningIncrement, facing: facing)
         camera.centerCoordinate = newCenter
         mapView.setCamera(camera, animated: true)
     }
+
+    private func createRouteController(with route: Route) -> RouteController {
```
1ec5: Nit: label the argument `along` instead of `with`, for consistency with the RouteController initializer.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -570,64 +573,74 @@ extension CarPlayManager: CPMapTemplateDelegate {
...
+        if self.simulatesLocations {
+            let locationManager = SimulatedLocationManager(route: route)
+            locationManager.speedMultiplier = self.simulatedSpeedMultipler
```
1ec5: Idle thought: what if we make 0 mean “don’t simulate”?

MapboxNavigation/CarPlayManager.swift
1ec5: Typo: multiplier.

MapboxNavigation/CarPlayManager.swift
akitchen: I would actually prefer to keep this as-is, in order to differentiate it from the RouteController initializer itself.

MapboxNavigation/CarPlayManager.swift
akitchen: For now I'd like to keep things as-is. But it's an interesting idea. I wonder if it would simplify or complicate setting this value from a UI. It might make sense with a slider or dial, but not with a button as we currently have (of course the multiplier isn't even exposed directly to CarPlay UI right now)

PR #1654, opened 2018-09-05
Enable style manager in CarPlay

1ec5:
The navigation v4 styles are now visible in CarPlay, both on the main map and in turn-by-turn navigation.

/cc @vincethecoder @bsudekum

PR #1664, opened 2018-09-09
Use location manager for overview, style switching in CarPlay

1ec5:
This PR fixes switching between day and night styles in CarPlay (following up on #1654), as well as the Overview button in CarPlay turn-by-turn navigation.

In both the main map view and the navigation map view, switch between day and night styles based on a location retrieved from a location manager, not a route controller or map view. The route controller may disqualify any location in the iOS simulator due to a lack of course readings, while the map view may not have a location by the time the style manager asks for it.

night

When tapping the Overview button in turn-by-turn navigation in CarPlay, use the location manager’s location instead of the route controller’s location, just as RouteMapViewController does. Once again, the route controller disqualifies just about any location in the iOS simulator, so overview mode ends up nonfunctional. Also add the maneuver arrow back onto the map when returning to turn-by-turn mode.

overview night

Depends on #1660.

/cc @frederoni @vincethecoder

### DIFF
--- a/MapboxNavigation/CarPlayMapViewController.swift
+++ b/MapboxNavigation/CarPlayMapViewController.swift
@@ -6,6 +6,12 @@ import CarPlay
 class CarPlayMapViewController: UIViewController, MGLMapViewDelegate {
     
     var styleManager: StyleManager!
+    /// A very coarse location manager used for distinguishing between daytime and nighttime.
+    fileprivate let coarseLocationManager: CLLocationManager = {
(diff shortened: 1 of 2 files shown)


akitchen:
We noticed a situation on the device today where it was night style on the head unit and day style on the phone. This was during simulation, if it matters... could be a new PR or another push to this branch?

1ec5:
It may indeed be relevant that route simulation was enabled. The main map showed the day style as expected. In any case, let’s track the bug in a separate issue.

PR #1665, opened 2018-09-09
Fix crash displaying shields in CarPlay

### LINKED ISSUE #1652: Simulator crashes when displaying shield in maneuver
During turn-by-turn navigation in CarPlay, when we go to update the maneuver to display a visual instruction that includes a shield, the application crashes just after calling `CarPlayNavigationViewController.updateManeuvers(for:)`:

```
#0	0x000000010c4d1071 in CarPlayNavigationViewController.visualInstructionDidChange(_:) at /path/to/mapbox-navigation-ios/MapboxNavigation/CarPlayNavigationViewController.swift:203
#1	0x000000010c4d135c in @objc CarPlayNavigationViewController.visualInstructionDidChange(_:) ()
```
https://github.com/mapbox/mapbox-navigation-ios/blob/76ccc5813665c99f74ce6f33a89cc474d940c3d0/MapboxNavigation/CarPlayNavigationViewController.swift#L203

Here’s the stack trace:

```
#0	0x0000000110ce7705 in objc_exception_throw ()
#1	0x0000000112a9d0f5 in +[NSException raise:format:] ()
#2	0x000000011162898c in __47-[CPInterfaceController _connectionInterrupted]_block_invoke ()
```

Based on the assembly code for that block, the exception is `Connection to host was interrupted`, indicating that the simulator has terminated its connection to the application.

The issue only occurs when the instruction’s attributed string contains a ShieldAttachment.

/cc @bsudekum @JThramer @frederoni

1ec5:
Before adding an attributed string to a maneuver’s attributed instructions, downgrade any instance of a custom text attachment subclass, such as ShieldAttachment, to NSTextAttachment. This ensures that the attributed string can be marshaled over to the CarPlay unit. Unfortunately, even standard NSTextAttachments don’t show up in the CarPlay simulator, but apparently they do show up on an actual CarPlay device.

Along the way, secondary and tertiary text instructions are now included in plain-text maneuver instructions for CarPlay units that lack support for attributed strings.

Fixes #1652.

/cc @JThramer

### DIFF
--- a/MapboxNavigation/NSAttributedString.swift
+++ b/MapboxNavigation/NSAttributedString.swift
@@ -1,12 +1,25 @@
 import Foundation
 
 extension NSAttributedString {
-    static func + (left: NSAttributedString, right: NSAttributedString) -> NSAttributedString
-    {
+    static func + (left: NSAttributedString, right: NSAttributedString) -> NSAttributedString {
         let result = NSMutableAttributedString()
         result.append(left)
         result.append(right)
         return result
     }
+}
 
+extension NSMutableAttributedString {
+    func canonicalizeAttachments() {
+        enumerateAttribute(.attachment, in: NSRange(location: 0, length: length), options: []) { (value, range, stop) in
+            guard let attachment = value as? NSTextAttachment, type(of: attachment) != NSTextAttachment.self else {
+                return
+            }
+            
+            let sanitizedAttachment = NSTextAttachment()
+            sanitizedAttachment.image = attachment.image
+            sanitizedAttachment.bounds = attachment.bounds
+            setAttributes([.attachment: sanitizedAttachment], range: range)
+        }
+    }
 }
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -210,6 +210,7 @@ public class CarPlayNavigationViewController: UIViewController, MGLMapViewDelega
                 instruction.append(attributedSecondary)
             }
             
+            instruction.canonicalizeAttachments()
             primaryManeuver.attributedInstructionVariants = [instruction]
         }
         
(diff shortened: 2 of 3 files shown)


MapboxNavigation/NSAttributedString.swift
```diff
@@ -1,12 +1,25 @@
...
         result.append(left)
         result.append(right)
         return result
     }
+}
 
+extension NSMutableAttributedString {
+    func canonicalizeAttachments() {
+        enumerateAttribute(.attachment, in: NSRange(location: 0, length: length), options: []) { (value, range, stop) in
+            guard let attachment = value as? NSTextAttachment, type(of: attachment) != NSTextAttachment.self else {
```
vincethecoder: @1ec5 this guard... it doesn't read right!!! Would this be why some of the shields are not showing up sometimes?

JThramer: @vincethecoder To me, it's reading as "The attachment is a subclass of `NSAttachment`, but not the actual class `NSAttachment.self`

vincethecoder: @JThramer Yes, I'm reading this as _we are looking for an attachment which is a subclass of `NSTextAttachment` but isn't of type `NSTextAttachment`_. But I suppose that's why we are attempt to cannoicalize the attachment(s).

JThramer: @vincethecoder yeah, this is likely because of limitations of the `NSAttributedString` logic in CarPlay.

1ec5: The application marshals the attributed string over to the CarPlay simulator or device for rendering. At least for the simulator if not also for the device, the attributed string is marshaled over XPC. It isn’t possible to marshal over instances of our custom NSTextAttachment subclasses, especially not the executable code in these subclasses. “Canonicalizing” the attributed string consists of replacing each subclassed attachment with a plain-vanilla NSTextAttachment that is sure to marshal over correctly.

There are other limitations to attributed strings in CarPlay maneuver instructions, such as being unable to change fonts or font sizes. However, those limitations aren’t limited to CarPlay. For example, on macOS, [you can use attributed strings in title bars](https://stackoverflow.com/a/29336473/4585461), but many attributes are ignored.

1ec5:
Confirmed that the shields show up on a CarPlay device, though the generic shields end up illegible in the night style. We’ll figure that out as tail work.

device

PR #1673, opened 2018-09-11
Show rating on CarPlay if connected

frederoni:
The regression was caused by passing a reference to the location manager instead of copying which lead to the NavigationViewController receiving the location updates and the RouteController associated with CarPlay never arrived at the destination.

screenshot 2018-09-11 14 46 09

cc @akitchen @1ec5

### DIFF
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -243,10 +243,10 @@ public class CarPlayManager: NSObject {
         let dateCreatedAttribute = [MMEEventKeyCreated: timestamp]
         eventsManager.manager.enqueueEvent(withName: MMEventTypeCarplayDisconnect, attributes: dateCreatedAttribute)
         eventsManager.manager.flush()
-        isConnectedToCarPlay = false
     }
 
     public func application(_ application: UIApplication, didDisconnectCarInterfaceController interfaceController: CPInterfaceController, from window: CPWindow) {
+        isConnectedToCarPlay = false
         self.interfaceController = nil
         carWindow?.isHidden = true
         let timestamp = Date().ISO8601
@@ -616,17 +616,11 @@ extension CarPlayManager: CarPlayNavigationDelegate {
         delegate?.carPlayManagerDidEndNavigation(self)
     }
 }
+#else
+@objc(MBCarPlayManager)
+class CarPlayManager: NSObject {
+    public static var shared = CarPlayManager()
+    var isConnectedToCarPlay: Bool = false
+}
 #endif
 
-struct CarPlayManagerHelper {
-    
-    static var isConnectedToCarPlay: Bool {
-        #if canImport(CarPlay)
-        if #available(iOS 12.0, *) {
-            return CarPlayManager.shared.isConnectedToCarPlay
-        }
-        #endif
-        return false
-    }
-    
-}
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -110,6 +110,11 @@ public class CarPlayManager: NSObject {
     private var defaultMapButtons: [CPMapButton]?
     
     /**
+     A boolean value indicating whether the phone is connected to a CarPlay device or not.
+     */
+    public var isConnectedToCarPlay: Bool = false
+    
+    /**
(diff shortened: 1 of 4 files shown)


Examples/Swift/AppDelegate.swift
```diff
@@ -72,7 +72,7 @@ extension AppDelegate: CarPlayManagerDelegate {
             
             if !navigationViewControllerExistsInStack {
                 
-                let locationManager = routeController.locationManager
+                let locationManager = routeController.locationManager.copy() as! NavigationLocationManager
```
akitchen: can this fail?

akitchen:
I've been wondering if this would be possible via message passing. copying hadn't occurred to me.

what happens if the copy fails? a crash?

Examples/Swift/AppDelegate.swift
frederoni: It can fail but it won't crash, e.g. using a `ReplayLocationManager`, you'd get something that resembles a `NavigationLocationManager` w/o the state of the replaying location manager. The superclass `(NavigationLocationManager)` is enforced by the route controller. So this could be kinda volatile in some scenarios.

I wonder if another pattern could help us here, e.g. a delegate with multiple receivers since the GPS is shared between CarPlay and device.

Something like:
```swift
var receivers: [CLLocationManagerDelegate]?
func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
    receivers?.forEach { $0.locationManager(manager, didUpdateLocations: locations) }
}
```
which is something usually more seen in Java. 🤔

akitchen: yeah that's basically what I meant by message passing

akitchen:
@frederoni should we take this as-is for now and consider changing it post-release? or do you want to refactor it now? Up to you...

akitchen:
I'm inclined to say let's ship it and refactor from there if desired.

I added @1ec5 as a reviewer to chime in, time permitting

Examples/Swift/AppDelegate.swift
1ec5: A delegator can only have one delegate; otherwise, it’s an observer pattern rather than the delegation pattern, so we’d use NSNotificationCenter.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -236,6 +241,7 @@ public class CarPlayManager: NSObject, CPSearchTemplateDelegate {
         let dateCreatedAttribute = [MMEEventKeyCreated: timestamp]
         eventsManager.manager.enqueueEvent(withName: MMEventTypeCarplayDisconnect, attributes: dateCreatedAttribute)
         eventsManager.manager.flush()
+        isConnectedToCarPlay = false
```
1ec5: Why not set this property to false in `application(_:didDisconnectCarInterfaceController:from:)` instead of here? If the property is set to false here in `sendCarPlayDisconnectEvent(_:)`, I’d expect it to be set to true in the corresponding `sendCarPlayConnectEvent(_:)` method, not in `application(_:didConnectCarInterfaceController:to:)`.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -645,3 +651,16 @@ extension CarPlayManager: CarPlayNavigationDelegate {
     }
 }
 #endif
+
+struct CarPlayManagerHelper {
+    
+    static var isConnectedToCarPlay: Bool {
```
1ec5: Could this property be a static computed property on CarPlayManager itself? Is the reason for splitting this struct out that would be available on all versions of iOS? If so, how about defining a stub CarPlayManager in the `#else` above, with only the `isConnectedToCarPlay` property implemented?

frederoni: > Is the reason for splitting this struct out that would be available on all versions of iOS?

Yes

> If so, how about defining a stub CarPlayManager in the #else above, with only the isConnectedToCarPlay property implemented?

👍

Examples/Swift/AppDelegate.swift
frederoni: Changing this to a loosely coupled observer pattern such as NSNotificationCenter would be a pretty big breaking change and could lead to many unanticipated impacts. Especially when the nav service branch gets merged which uses more than one location manager simultaneously.

PR #1675, opened 2018-09-11
Optional Search in CarPlay

frederoni:
MapboxNavigation now optionally links MapboxGeocoder and optionally compiles relevant classes and extensions. This will minimize the effort needed to get search implemented in a CarPlay application. I've also split up AppDelegate into AppDelegate+CarPlay to keep the default app delegate as readable as possible.

cc @akitchen


Examples/Swift/AppDelegateCarPlay.swift
```diff
@@ -0,0 +1,99 @@
...
+        CarPlayManager.shared.delegate = nil
+        CarPlayManager.shared.application(application, didDisconnectCarInterfaceController: interfaceController, from: window)
+    }
+}
+
+@available(iOS 12.0, *)
+extension AppDelegate: CarPlayManagerDelegate {
+    
+    // MARK: CarPlayManagerDelegate
+    func carPlayManager(_ carPlayManager: CarPlayManager, didBeginNavigationWith routeController: RouteController) {
```
akitchen: This code (and didEnd...) is specific to the drop-in UI -- can we move it somewhere else, perhaps into the `NavigationViewController` itself?

Examples/Swift/AppDelegateCarPlay.swift
```diff
@@ -0,0 +1,99 @@
+import UIKit
```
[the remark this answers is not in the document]
frederoni: Swift’s API design guidelines doesn't state any naming conventions regarding the filename of extensions but inheriting Obj-C’s seems to be the consensus.

Examples/Swift/AppDelegateCarPlay.swift
frederoni: Tried to move this over to NVC but because this happens before the NVC is in the view hierarchy, it needs to be static, and we need to pass the UIWindow to NVC. `UIApplication.shared.keyWindow` returns the CPWindow here so it would have to be a somewhat clumsy workaround.

akitchen: oof. I would strongly prefer that we not reach out to the global like that, preferring to pass it in as an initializer or method argument

PR #1676, opened 2018-09-11
Print xcodebuild log when test fail

frederoni:
Print the raw xcodebuild log when a build fails on CI.

cc @akitchen


akitchen:
awesome! thanks @frederoni

PR #1680, opened 2018-09-13
CarPlay Camera Fixes

JThramer:
Zoom Zoom!

Improves the camera configuration for the initial route preview

[Video of result](https://www.dropbox.com/s/ts3cdrmp8iblf0a/finally.mov?dl=0)

/cc @mapbox/navigation-ios


MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -93,6 +90,24 @@ class CarPlayMapViewController: UIViewController, MGLMapViewDelegate {
...
+        if defaultAltitude {
+            camera.altitude = 16000
+        }
+        camera.pitch = 60
+        mapView.setCamera(camera, animated: false)
+
+    }
+    
+    override func viewSafeAreaInsetsDidChange() {
+        guard isOverviewingRoutes else { return }
```
frederoni: nit: it's good practice to call super on overridden methods unless there’s an explicit reason not to.

JThramer:
Recycling this back to in-progress due to need to re-work logic not to use `.followWithHeading`

akitchen:
LGTM but I'd like to see a final 👍 from @1ec5 or @frederoni 

Also, the target branch `carplay` has shifted underneath you again, @JThramer

MapboxNavigation/CarPlayManager.swift
```diff
@@ -298,13 +298,28 @@ public class CarPlayManager: NSObject {
...
             let mapView = carPlayMapViewController.mapView
             mapView.removeRoutes()
             mapView.removeWaypoints()
-            mapView.userTrackingMode = .followWithCourse
+            mapView.setUserTrackingMode(.followWithCourse, animated: true)
         }
     }
+    public func templateWillDisappear(_ template: CPTemplate, animated: Bool) {
+
+        let isCorrectType = type(of: template) == CPSearchTemplate.self || type(of: template) == CPMapTemplate.self
```
1ec5: FYI, I think this preserves the zooming-in animation when transitioning from the navigation map template to the main map template. I think it’s perfectly fine as long as there’s an initial frame in which the route line and waypoints are still visible; otherwise, it feels gratuitous.

MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -93,6 +88,34 @@ class CarPlayMapViewController: UIViewController, MGLMapViewDelegate {
             mapView.localizeLabels()
         }
     }
+    
+    func resetCamera(animated: Bool = false, defaultAltitude: Bool = false) {
+        let camera = mapView.camera
+        if defaultAltitude {
+            camera.altitude = 16000
```
1ec5: Replace `defaultAltitude` with an `altitude` argument that defaults to 16,000 instead of false.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -408,6 +408,36 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
     
     // MARK: Feature Addition/Removal
+    /**
+     Showcases route array. Adds routes and waypoints to map, and sets camera to point encompassing the route.
+    */
+    public static let defaultPadding: UIEdgeInsets = UIEdgeInsets(top: 10, left: 20, bottom: 10, right: 20)
+    
+    @objc public func showcase(routes: [Route], padding: UIEdgeInsets = NavigationMapView.defaultPadding, animated: Bool = false) {
```
1ec5: Nit: remove the first argument’s label: `_ routes`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -408,6 +408,36 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        removeRoutes()
+        removeWaypoints()
+        
+        showRoutes(routes)
+        showWaypoints(active)
+        
+        center(on: active, facing: 0, padding: padding, animated: animated)
+    }
+    
+    func center(on route: Route, facing direction:CLLocationDirection = 0, padding: UIEdgeInsets = NavigationMapView.defaultPadding, animated: Bool = false) {
```
1ec5: If it isn’t too much trouble, `fit(to:facing:edgePadding:animated:)` is a little more descriptive, since “recentering” can also describe what happens when you leave overview mode in turn-by-turn guidance (hence `recenterButton`).

MapboxNavigation/CarPlayManager.swift
JThramer: That's exactly what I'm seeing in my test runs.

PR #1684, opened 2018-09-14
Avoid irrelevant search results

frederoni:
When searching for something like "Yerba Buena" with focal location set to SF and locale set to `en`, the geocoder’s top result is Argentina. IIRC, this happens because "Yerba Buena" in the US is not translated into English but it's using English as a base language.

This change omits the language query parameter if the locale’s language code is equal to `en` and also sets a focal location.

cc @akitchen @1ec5

### DIFF
--- a/MapboxNavigation/CarPlayManager+Search.swift
+++ b/MapboxNavigation/CarPlayManager+Search.swift
@@ -10,6 +10,13 @@ extension CarPlayManager: CPSearchTemplateDelegate {
     public static let CarPlayGeocodedPlacemarkKey: String = "CPGecodedPlacemark"
     static var recentItems = RecentItem.loadDefaults()
     
+    /// A very coarse location manager used for focal location when searching
+    fileprivate static let coarseLocationManager: CLLocationManager = {
+        let coarseLocationManager = CLLocationManager()
+        coarseLocationManager.desiredAccuracy = kCLLocationAccuracyThreeKilometers
+        return coarseLocationManager
+    }()
+    
     // MARK: CPSearchTemplateDelegate
     
     public func searchTemplate(_ searchTemplate: CPSearchTemplate, updatedSearchText searchText: String, completionHandler: @escaping ([CPListItem]) -> Void) {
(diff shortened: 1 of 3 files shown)


MapboxNavigation/CarPlayManager+Search.swift
```diff
@@ -10,6 +10,13 @@ extension CarPlayManager: CPSearchTemplateDelegate {
     public static let CarPlayGeocodedPlacemarkKey: String = "CPGecodedPlacemark"
     static var recentItems = RecentItem.loadDefaults()
     
+    /// A very coarse location manager used for focal location when searching
+    fileprivate static let coarseLocationManager: CLLocationManager = {
+        let coarseLocationManager = CLLocationManager()
+        coarseLocationManager.desiredAccuracy = kCLLocationAccuracyThreeKilometers
```
1ec5: A more fine-grained accuracy may be desirable for searching for common amenities in a dense urban area.

MapboxNavigation/CarPlayManager+Search.swift
```diff
@@ -57,7 +64,8 @@ extension CarPlayManager: CPSearchTemplateDelegate {
         if shouldSearch {
             
             let options = ForwardGeocodeOptions(query: searchText)
-            options.locale = .autoupdatingCurrent
+            options.focalLocation = CarPlayManager.coarseLocationManager.location
```
1ec5: Good catch!

MapboxNavigation/Locale.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+
+extension Locale {
+    
+    // Returns Locale.autoupdatingCurrent unless it is prefixed with `en`, then nil will be returned.
+    static var autoupdatingCurrentOrNil: Locale? {
```
1ec5: Nit: `autoupdatingNonEnglish`.

MapboxNavigation/Locale.swift
```diff
@@ -0,0 +1,15 @@
...
+        if let languageCode = Locale.autoupdatingCurrent.languageCode,
+            languageCode.hasPrefix("en"){
```
1ec5: English isn’t the only ISO&nbsp;639 code that begins with `en`. Check for equality with `en` instead; Locale will never come up with an alpha-3 language code of `eng`.

MapboxNavigation/CarPlayManager+Search.swift
frederoni: Thought about this too but I reasoned that a map view is most likely already using a location manager with higher accuracy and we don't want the geocoder part of the code to be the cause draining the battery. Even if the desired accuracy is high, it's highly likely that the last received location is accurate enough, probably within 100m.

PR #1686, opened 2018-09-14
Application sleep mode is disabled when carplay is connected.

vincethecoder:
The intention is to prevent the scenario where the system puts the device into a sleep state. Unfortunately, this impacts the map interactions on a carplay head unit. As a recommendation by Apple developer documentation, mapping apps should disable the idle timer to allow the continuous display or interaction with content.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -82,6 +82,17 @@ public protocol CarPlayManagerDelegate {
      * Called when navigation ends so that the containing app can update accordingly.
      */
     @objc func carPlayManagerDidEndNavigation(_ carPlayManager: CarPlayManager) -> ()
+    
+    /**
+     Called when the carplay manager will disable the idle timer.
+     
+     Implementing this method will allow developers to change whether idle timer is disabled when `CarPlayManager` is initialized.
+     
+     - parameter carPlayManager: The carplay manager that will change the state of idle timer.
+     - returns: A bool indicating whether to disable idle timer when the CarPlayManager is initialized.
+     */
+    @objc(carplayManagerShouldDisableIdleTimer:)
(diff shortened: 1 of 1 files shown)


MapboxNavigation/CarPlayManager.swift
```diff
@@ -82,6 +82,17 @@ public protocol CarPlayManagerDelegate {
...
+    
+    /**
+     Called when the carplay manager will disable the idle timer.
+     
+     Implementing this method will allow developers to change whether idle timer is disabled when carplay is connected and the vice-versa when disconnected.
+     
+     - parameter carPlayManager: The carplay manager that will change the state of idle timer.
+     - returns: A bool indicating whether to disable idle timer when carplay is connected and enable when disconnected.
+     */
+    @objc(carplayManagerShouldDisableIdleTimer:)
```
frederoni: nit: this automatically bridges to the same signature, @objc would be sufficient.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -249,6 +265,13 @@ public class CarPlayManager: NSObject {
         carWindow?.isHidden = true
         let timestamp = Date().ISO8601
         sendCarPlayDisconnectEvent(timestamp)
+        
+        // Upon disconnecting the device from carplay, the idle timer ideally should be reset to its initial state.
+        if let shouldDisableIdleTimer = delegate?.carplayManagerShouldDisableIdleTimer?(self) {
+            UIApplication.shared.isIdleTimerDisabled = !shouldDisableIdleTimer
+        } else {
+            UIApplication.shared.isIdleTimerDisabled = true
```
JThramer: should be false

akitchen: yeah I agree, this doesn't seem quite right, but it could either be the inverted boolean or the reuse of the flag with a different semantic meaning. 

Did you two resolve this during your pairing?

PR #1689, opened 2018-09-14
Pan based on screen distance in CarPlay

1ec5:
The panning buttons now shift the map by a fixed number of screen points rather than a fixed physical distance. The previous approach resulted in wildly different behavior depending on the zoom level and relied on the map to be unrotated, which isn’t quite tenable given the recent change to follow with course. Now you can pan left then right and wind up in roughly the same spot as before, even when the map is rotated.

/cc @vincethecoder @JThramer


MapboxNavigation/CarPlayManager.swift
```diff
@@ -613,63 +612,29 @@ extension CarPlayManager: CPMapTemplateDelegate {
...
-        mapView.setUserTrackingMode(.none, animated: false)
-        mapView.resetNorth()
-
-        var facing: CLLocationDirection = 0.0
-
-        if direction.contains(.right) {
-            facing = 90
+        let contentFrame = UIEdgeInsetsInsetRect(mapView.bounds, mapView.safeArea)
+        var center = CGPoint(x: contentFrame.midX, y: contentFrame.midY)
+        let increment = min(mapView.bounds.width, mapView.bounds.height) / 2.0
```
vincethecoder: @1ec5 do we need to consider when `mapView` is pitched in order to determine the ratio to increment the pan `up` move?

1ec5:
@JThramer took care of most of https://github.com/mapbox/mapbox-navigation-ios/pull/1689#issuecomment-421515621. a9aa5924c72355b2354e06811bb35bdd612de63e refactors things so that all the buttons pan by the same amount.

1ec5:
> relied on the map to be unrotated

For future reference, the reason this previous behavior impaired the user experience was that it made the pan buttons useless. Consider: the map is centered on the user’s location, but they want to pan the map upward to see more of what’s in front of them. They enter panning mode, which rotates them 270°, and now they can’t figure out whether they should hit the up button or right button or …

Or if they’re using a CarPlay unit that supports dragging gestures, it’s even weirder, because the map rotates to north upon lifting the finger.

PR #1690, opened 2018-09-15
Generic Route Shield foreground color fix

JThramer:
WIP

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -528,7 +528,7 @@ open class NavigationViewController: UIViewController {
                 let locationManager = routeController.locationManager.copy() as! NavigationLocationManager
                 let directions = routeController.directions
                 let route = routeController.routeProgress.route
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -528,7 +528,7 @@ open class NavigationViewController: UIViewController {
                 let locationManager = routeController.locationManager.copy() as! NavigationLocationManager
                 let directions = routeController.directions
                 let route = routeController.routeProgress.route
-                let navigationViewController = NavigationViewController(for: route, directions: directions, locationManager: locationManager)
+                let navigationViewController = NavigationViewController(for: route, directions: directions, styles: [CarPlayDayStyle(), CarPlayNightStyle()], locationManager: locationManager)
```
1ec5: This is for testing in a phone simulator, I take it?

1ec5:
#1696 is an alternative approach to fixing this issue. Hopefully it isn’t too hacky compared to the fix here.

JThramer:
Closing this as minh fixed it with a much better approach.

PR #1692, opened 2018-09-15
Remove unused ref

frederoni:
We're not using self in this closure

cc @akitchen


akitchen:
Looks like our pod integration test needs an update 
```
xcodebuild: error: Unable to find a destination matching the provided destination specifier:
		{ platform:iOS Simulator, id:9F32C487-FAEC-4ABE-B00D-159EC007E50D }

	The requested device could not be found because no available devices matched the request.
```

akitchen:
actually, this looks like it could be a toolchain version mismatch with what's on CI. Perhaps we should use a device description instead of a GUID identifier? It's probably a bit more flexible.

PR #1694, opened 2018-09-15
Additional metric to ensure idle timer is not modified by NavigationViewController.

vincethecoder:
These changes ensure the `idle timer` managed by the CarPlay manager.

/cc @mapbox/navigation-ios


MapboxNavigation/NavigationViewController.swift
```diff
@@ -518,6 +518,7 @@ open class NavigationViewController: UIViewController {
         if let navigationViewController = window.viewControllerInStack(of: NavigationViewController.self) {
             // Open StepsViewController on iPhone if NavigationViewController is being presented
             navigationViewController.openStepsViewController()
+            navigationViewController.shouldManageApplicationIdleTimer = false
```
1ec5: This code only runs if NavigationViewController is already active on the phone when connecting to CarPlay.

Also, if the user disconnects CarPlay while navigating, then NavigationViewController is unable to keep the screen on for the rest of the trip.

vincethecoder: @1ec5 Yes, that's by design. We need to ensure the `isIdleTimerDisabled` property is reset right when the Phone is disconnected. https://github.com/mapbox/mapbox-navigation-ios/blob/carplay/MapboxNavigation/CarPlayManager.swift#L279

PS: We only need the `isIdleTimerDisabled` for the duration that CarPlay is connected. Afterward, it's imperative that `UIApplication.shared.isIdleTimerDisabled`ideally managed by the application level until another CarPlay connection is established.

1ec5: The scenario I’m concerned about is:

1. The user begins navigation on their phone. `NavigationViewController.shouldManageApplicationIdleTimer` and `UIApplication.isIdleTimerDisabled` are set to true.
1. The user connects the phone to CarPlay. This code sets `NavigationViewController.shouldManageApplicationIdleTimer` to false, but `CarPlayManager.application(_:didConnectCarInterfaceController:to:)` does set `UIApplication.isIdleTimerDisabled` to true.
1. The user disconnects the phone from CarPlay. `CarPlayManager.application(_:didDisconnectCarInterfaceController:from:)` sets `UIApplication.isIdleTimerDisabled` to false. NavigationViewController continues to be visible on the phone, but `NavigationViewController.shouldManageApplicationIdleTimer` is false.
1. After a little while, the phone goes dark because `UIApplication.isIdleTimerDisabled` is false.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -529,7 +530,7 @@ open class NavigationViewController: UIViewController {
                 let directions = routeController.directions
                 let route = routeController.routeProgress.route
                 let navigationViewController = NavigationViewController(for: route, directions: directions, locationManager: locationManager)
-                
+                navigationViewController.shouldManageApplicationIdleTimer = false
```
1ec5: 1. The user connects the phone to CarPlay. The application on the phone hasn’t begun navigation yet, so `UIApplication.isIdleTimerDisabled` was been false. This code sets `NavigationViewController.shouldManageApplicationIdleTimer` to false, and `CarPlayManager.application(_:didConnectCarInterfaceController:to:)` sets `UIApplication.isIdleTimerDisabled` to true.
1. The user disconnects the phone from CarPlay. `CarPlayManager.application(_:didDisconnectCarInterfaceController:from:)` sets `UIApplication.isIdleTimerDisabled` to false. NavigationViewController is still visible on the phone.
1. After a little while, the phone goes dark because `UIApplication.isIdleTimerDisabled` is false.

MapboxNavigation/NavigationViewController.swift
1ec5: Because NavigationViewController is still visible on the phone.

MapboxNavigation/NavigationViewController.swift
1ec5: NavigationViewController is still visible on the phone.

MapboxNavigation/NavigationViewController.swift
vincethecoder: @1ec5 yes, but we don't want the power-saving mode turned off unless it's required for the mapping experience. After disconnection from CarPlay, we certainly need to set thee idle state disabled set to `true`.

MapboxNavigation/NavigationViewController.swift
vincethecoder: @1ec5 yes, but we don't want the power-saving mode turned off unless it's required for the mapping experience. After disconnection from CarPlay, we certainly need to set thee idle state disabled set to `true`.

MapboxNavigation/NavigationViewController.swift
1ec5: Right, after disconnecting from CarPlay, `isIdleTimerDisabled` needs to remain true as long as NavigationViewController remains open on the phone. Have you verified that this is the case with your PR?

vincethecoder: @1ec5 that's what I've been trying to explain that, we don't need to persist this behavior post disconnecting from CarPlay. When it disconnects, we would like to avoid keeping that flag unless the app level indicates it should. This could potentially kill battery life of the device.

MapboxNavigation/NavigationViewController.swift
vincethecoder: https://github.com/mapbox/mapbox-navigation-ios/pull/1694#discussion_r217901723

MapboxNavigation/NavigationViewController.swift
1ec5: Ignoring `NavigationViewController.shouldManageApplicationIdleTimer` for a moment, have you verified that `UIApplication.isIdleTimerDisabled` remains true after CarPlay is disconnected but before ending navigation on the phone?

vincethecoder: @1ec5 `UIApplication.isIdleTimerDisabled` should not remain `true` after CarPlay is disconnected. There's no need to keep it in the state after the phone is disconnected, see this code here: https://github.com/mapbox/mapbox-navigation-ios/blob/carplay/MapboxNavigation/CarPlayManager.swift#L279

1ec5: That is incorrect, and the comment fails to consider the important scenario I’m trying to put forth.

If the user has disconnected CarPlay and turn-by-turn guidance is happening exclusively on the phone, NavigationViewController on the phone should act as if CarPlay had never been connected – that is, it should keep the screen on until the user presses the exit button. It is a poor user experience for the screen to automatically shut off while turn-by-turn navigation is active on any screen, end of story.

MapboxNavigation/NavigationViewController.swift
akitchen: > we don't want the power-saving mode turned off unless it's required for the mapping experience

akitchen:
In order for this to resolve, I'd like to see an integrated test which proves that:

 - When CarPlay is connected, the idle timer is disabled and the NavVC isn't going to disable it under any appearance/disappearance callback scenario
 - When CarPlay is disconnected, the idle time is left disabled and the NavVC takes responsibility for managing it as appropriate.

As an aside, I'm not sure that the NavigationViewController idle timer heuristic is correct:
```
    /**
     If `true`, `UIApplication.isIdleTimerDisabled` is set to `true` in `viewWillAppear(_:)` and `false` in `viewWillDisappear(_:)`. If your application manages the idle timer itself, set this property to `false`.
     */
    @objc public var shouldManageApplicationIdleTimer = true
```

Don't we receive `viewWillDisappear(_:)` for example when the StepsTableViewController is presented modally over the NavigationViewController?

vincethecoder:
@akitchen integration tests addressed in this [commit](https://github.com/mapbox/mapbox-navigation-ios/pull/1694/commits/5ef27a7e2e564f8c81f9061d065d8ce759d87b0c)

MapboxNavigation/NavigationViewController.swift
JThramer: @vincethecoder I think I see where @1ec5 is coming from here. If the current product design (disregarding) is to disable the idle timer (or ask a delegate whether it should) while navigating, we need to keep that behavior consistent. Connecting and disconnecting CarPlay shouldn't irrevocably change the app state.

Maybe we should have a delegate call for the NVC that asks whether or not idle timer should be disabled, CarPlay work notwithstanding?

vincethecoder:
Closing PR - will re-open to the appropriate branch.

Issue #808, opened 2017-11-08
Route preview view controller


1ec5:
This library should vend a view controller that displays a preview of a route (and potentially any alternative routes). There would be controls for beginning the route (pushing NavigationViewController onto the stack), as well as an option to view a static list of steps along the route.

RoutePreviewViewController would serve as an additional, optional entry point to the existing NavigationViewController. The user normally expects to be able to see the route visualized or summarized somehow before committing to it by beginning turn-by-turn navigation. In a navigation-focused application, the main view controller may already contain a NavigationMapView that can display the route, so it RoutePreviewViewController may not be strictly necessary. However, in a non-navigation-focused application, such a UI may not already be present, so RoutePreviewViewController will make it easier for developers to complete that part of the user experience.

/cc @mapbox/navigation-ios

bsudekum:
I think this is a huge rabbit hole and time sink that really doesn't help developers all that much in the end. There are a few basic functions a developer needs here:

* show the route
  * congestion, alterates, etc
* Fit to bounds
* Preview steps

Soon, we have all these pieces exposed (missing: https://github.com/mapbox/mapbox-navigation-ios/pull/785). Looking at google and apple, we'd need to expose all sorts of elements to do this right. This would mean a search view to edit the start location:

<img src="https://user-images.githubusercontent.com/1058624/32554849-260c11a0-c469-11e7-8c90-bebefe94036d.png" width="30%" />

<img src="https://user-images.githubusercontent.com/1058624/32554879-390c5044-c469-11e7-829e-66dc59292f6e.png" width="30%" />

In my opinion we should make the example app implement this in a concise manner and keep focused on more core features.

1ec5:
> However, in a non-navigation-focused application, such a UI may not already be present, so RoutePreviewViewController will make it easier for developers to complete that part of the user experience.

I guess I was envisioning something like a store locator application, where it would be unnecessary to provide a search/waypoint interface, but a drop-in interface for a route confirmation screen would still be convenient. I suppose we could reimplement that view controller as part of [the store locator kit](https://github.com/mapbox/store-locator-ios/), though.

1ec5:
> I think this is a huge rabbit hole and time sink that really doesn't help developers all that much in the end.

It’s important to recognize that the preview view controller would be an optional entry point. A developer who needs a more custom confirmation screen or no confirmation screen at all can still build their own UI and push NavigationViewController onto the stack directly.

[...]

PR #1695, opened 2018-09-15
Fold CarPlay-specific styles into main styles

1ec5:
Removed CarPlayDayStyle and CarPlayNightStyle in favor of using DayStyle and NightStyle in all user interface idioms. Vary ExitView and GenericShieldView by idiom.

Added a `Style.previewMapStyleURL` property, which is used by CarPlayMapViewController but can also be used in the future for the preview view controller proposed by #808. In the meantime, the property is visible to applications for custom preview view controller implementations.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/Style.swift
+++ b/MapboxNavigation/Style.swift
@@ -32,9 +32,23 @@ open class Style: NSObject {
     @objc public var styleType: StyleType = .day
     
     /**
-     Map style to be used for the style.
+     URL of the style to display on the map during turn-by-turn navigation.
      */
     @objc open var mapStyleURL: URL = MGLStyle.navigationGuidanceDayStyleURL
+    
+    #if canImport(CarPlay)
+    /**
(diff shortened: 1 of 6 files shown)


MapboxNavigation/Style.swift
```diff
@@ -32,10 +32,24 @@ open class Style: NSObject {
     @objc public var styleType: StyleType = .day
     
     /**
-     Map style to be used for the style.
+     URL of the style to display on the map during turn-by-turn navigation.
      */
     @objc open var mapStyleURL: URL = MGLStyle.navigationGuidanceDayStyleURL
     
+    #if canImport(CarPlay)
```
frederoni: Do we need to wrap this in a canImport? The URL is still available iOS < 12.

1ec5: The `#else` block below is identical except for the documentation comment. I figured it would make little sense to talk about CarPlay in the documentation if CarPlay is unavailable.

PR #1696, opened 2018-09-15
Fix generic shields in CarPlay

1ec5:
Add CarPlay’s throwaway instruction label to the car window while rendering the instruction. While taking a snapshot of GenericShieldView or ExitView, add the view to the data source’s window if the data source is a view, but not necessarily the application’s main window.

Depends on #1695. I’ll retarget this PR against the `carplay` branch before merging.

/cc @JThramer @frederoni

### DIFF
--- a/MapboxNavigation/InstructionPresenter.swift
+++ b/MapboxNavigation/InstructionPresenter.swift
@@ -244,9 +244,14 @@ class InstructionPresenter {
     }
     
     private func takeSnapshot(on view: UIView) -> UIImage? {
-        let window = UIApplication.shared.delegate!.window!!
+        let window: UIWindow
+        if let hostView = dataSource as? UIView, let hostWindow = hostView.window {
+            window = hostWindow
+        } else {
+            window = UIApplication.shared.delegate!.window!!
+        }
         
-        //We have to temporarily add the view to the view heirarchy in order for UIAppearance to work it's magic.
+        // Temporarily add the view to the view hierarchy for UIAppearance to work its magic.
         window.addSubview(view)
         let image = view.imageRepresentation
         view.removeFromSuperview()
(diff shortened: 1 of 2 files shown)


1ec5:
Unfortunately, I have limited ability to verify this fix. So far I’m setting a breakpoint image here and examining `image` in Quick Look:

https://github.com/mapbox/mapbox-navigation-ios/blob/efe7ae72246d20d09b7403b29f7fb0919869f519/MapboxNavigation/InstructionPresenter.swift#L252

and setting a breakpoint here and examining `sanitizedAttachment.image` in Quick Look:

https://github.com/mapbox/mapbox-navigation-ios/blob/efe7ae72246d20d09b7403b29f7fb0919869f519/MapboxNavigation/NSAttributedString.swift#L21

Please verify that NavigationViewController and its step table continue to show generic shields and exits with the correct colors with the day style applied.

1ec5:
A test relevant to this fix is crashing:

```
    ✗ testUsingURLStringAsCacheKey, Test crashed
```

1ec5:
8bd201c9601fdd524b4809729b9af23f8e01edfe fixes the crash in `testUsingURLStringAsCacheKey`. Not sure how we ended up starting a CarPlay navigation session as part of that test…

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -244,9 +244,14 @@ class InstructionPresenter {
     }
     
     private func takeSnapshot(on view: UIView) -> UIImage? {
-        let window = UIApplication.shared.delegate!.window!!
+        let window: UIWindow
+        if let hostView = dataSource as? UIView, let hostWindow = hostView.window {
+            window = hostWindow
+        } else {
+            window = UIApplication.shared.delegate!.window!!
```
akitchen: Should this be more gentle?

JThramer: We could return nil here, but isn't it a fair assumption make that the application will always have a window on the phone? 

if we ever end up breaking that assumption, it might be a better idea to fail loudly.

1ec5: This line is unchanged in this PR compared to master. This PR only provides a gentler alternative when rendering for CarPlay.

PR #1701, opened 2018-09-15
Arrow Direction Fix

JThramer:
simulator screen shot - iphone x - 2018-09-15 at 16 42 50
simulator screen shot - iphone x - 2018-09-15 at 16 45 51

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/VisualInstruction.swift
+++ b/MapboxNavigation/VisualInstruction.swift
@@ -11,7 +11,7 @@ extension VisualInstruction {
 
 #if canImport(CarPlay)
     @available(iOS 12.0, *)
-    var maneuverImageSet: CPImageSet? {
+    func maneuverImageSet(side: DrivingSide = .right) -> CPImageSet? {
         let colors: [UIColor] = [.black, .white]
         let blackAndWhiteManeuverIcons: [UIImage] = colors.compactMap { (color) in
             let mv = ManeuverView()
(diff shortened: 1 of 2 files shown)


MapboxNavigation/VisualInstruction.swift
```diff
@@ -11,7 +11,7 @@ extension VisualInstruction {
 
 #if canImport(CarPlay)
     @available(iOS 12.0, *)
-    var maneuverImageSet: CPImageSet? {
+    func maneuverImageSet(side: DrivingSide = .right) -> CPImageSet? {
```
1ec5: Is the default argument value needed? Seems like more of a trap than a convenience.

MapboxNavigation/VisualInstruction.swift
```diff
@@ -20,11 +20,23 @@ extension VisualInstruction {
...
         }
         guard blackAndWhiteManeuverIcons.count == 2 else { return nil }
         return CPImageSet(lightContentImage: blackAndWhiteManeuverIcons[1], darkContentImage: blackAndWhiteManeuverIcons[0])
     }
+    
+    func shouldFlipImage(side: DrivingSide) -> Bool {
+        let leftDirection = [.left, .slightLeft, .sharpLeft].contains(maneuverDirection)
+        
+        switch maneuverType {
+        case _ where maneuverDirection == .uTurn, .takeRoundabout, .turnAtRoundabout, .takeRotary:
```
1ec5: Looks like this is effectively switching on `maneuverDirection`, not `maneuverType`.

PR #1706, opened 2018-09-16
Merge master <> carplay

akitchen:
This PR is the results of reconciling `master` with the latest changes on `carplay`. 

This will need to be kept up to date with any carplay-specific changes before final merge into master.


akitchen:
I think we're ready to merge.

/cc @mapbox/navigation-ios

1ec5:
Remember to make a merge commit rather than rebasing or squashing.

PR #1707, opened 2018-09-16
Changes to ensure user puck switches back when pan gesture is complete.

vincethecoder:
Upon tapping the close button, user puck must switch back to original state (`. followWithCourse`) when in pan mode.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -259,6 +259,11 @@ public class CarPlayManager: NSObject {
             guard let strongSelf = self, let mapButtons = strongSelf.defaultMapButtons else {
                 return
             }
+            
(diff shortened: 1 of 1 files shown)


akitchen:
This does switch the puck back, but it also recenters the map when the pan buttons are dismissed. is that avoidable?

vincethecoder:
This will be addressed in another PR. Closing for now.

1ec5:
> This does switch the puck back, but it also recenters the map when the pan buttons are dismissed. is that avoidable?

@akitchen, no, it is not. Either we recenter the map, or we get a dot:

https://github.com/mapbox/mapbox-gl-native/blob/68d463ec2819a2c1e34159c717cfc0166b8ecb02/platform/ios/src/MGLFaux3DUserLocationAnnotationView.m#L52

PR #1708, opened 2018-09-17
Panning Gesture Fixes

JThramer:
Pan.

Fixed behavior with panning, and fixed bug where recenter button shows just by showing the panning interface. It should not show until you actually move.

/cc @mapbox/navigation-ios


akitchen:
Seems much improved. Thanks @JThramer !

PR #1709, opened 2018-09-17
CarPlay

1ec5:
This is a long-running feature branch that integrates the CarPlay framework into the navigation SDK. It includes built-in support for [CarPlay in the navigation SDK](https://blog.mapbox.com/carplay-ios12-mapbox-navigation-sdk-c446c422ad24) to be used in a manner reminiscent of NavigationViewController, as well as a CarPlay-specific example application. CarPlayManager and related classes provide a main map and turn-by-turn navigation, also suitable for use in applications that otherwise roll their own custom navigation UIs. Unlike NavigationViewController, it also provides search and route selection screens.

This feature requires Xcode 10 and iOS 12.

Shoutout to @bsudekum for his initial work on this feature.

/cc @mapbox/navigation-ios


MapboxNavigation/CPFavoritesList.swift
```diff
@@ -0,0 +1,58 @@
+#if canImport(CarPlay)
+import CarPlay
+#endif
+
+public enum CPFavoritesList {
```
1ec5: The `CP` prefix is reserved for the CarPlay framework itself. “CarPlay” should be spelled out.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -621,9 +719,11 @@
 		AE5F8770209A082500F58FDB /* route-with-banner-instructions.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = "route-with-banner-instructions.json"; sourceTree = "<group>"; };
 		AE8B1B96207D2B2B003050F6 /* TunnelIntersectionManagerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TunnelIntersectionManagerTests.swift; sourceTree = "<group>"; };
 		AE997D2121137B8B00EB0AAB /* String+LocalizedConstants.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "String+LocalizedConstants.swift"; sourceTree = "<group>"; };
+		AEA5B13321312CE1003B2961 /* CPFavoritesList.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CPFavoritesList.swift; sourceTree = "<group>"; };
```
1ec5: This enumeration doesn’t belong in the SDK proper. It hard-codes two places as favorites, so it’s of little use to users. If we keep it around, it should be part of the example application.

MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -64,6 +64,28 @@ struct RoundingTable {
...
+            case .millimeter:
+                return Measurement(value: distance.kilometers / 1e6, unit: .millimeters)
+            case .centimeter:
+                return Measurement(value: distance.kilometers / 1e5, unit: .centimeters)
+            case .meter:
+                return Measurement(value: distance.kilometers / 1e3, unit: .meters)
+            case .kilometer:
+                return Measurement(value: distance.kilometers, unit: .kilometers)
+            case .inch:
+                return Measurement(value: distance.feet / 12, unit: .inches)
```
frederoni: feet * 12

MapboxNavigationTests/StyleManagerTests.swift
```diff
@@ -81,27 +81,27 @@ class StyleManagerTests: XCTestCase {
...
-        XCTAssert(styleManager.styleType(for: location) == .day)
-        styleManager.date = noonDate
-        XCTAssert(styleManager.styleType(for: location) == .day)
-        styleManager.date = beforeSunset
-        XCTAssert(styleManager.styleType(for: location) == .day)
-        styleManager.date = afterSunset
-        XCTAssert(styleManager.styleType(for: location) == .night)
-        styleManager.date = midnight
-        XCTAssert(styleManager.styleType(for: location) == .night)
+//        NSTimeZone.default = NSTimeZone.init(abbreviation: "PST")! as TimeZone
```
JThramer: Remove comments.

MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -184,10 +182,14 @@ class NavigationViewControllerTests: XCTestCase {
         //wait for the style to load -- routes won't show without it.
         wait(for: [styleLoaded], timeout: 5)
         navigationViewController.route = initialRoute
+
+        runUntil({
```
JThramer: timeout?

MapboxNavigationTests/InstructionsBannerViewSnapshotTests.swift
```diff
@@ -5,181 +5,181 @@ import MapboxDirections
 @testable import MapboxCoreNavigation
 
 class InstructionsBannerViewSnapshotTests: FBSnapshotTestCase {
-
```
JThramer: Why does this keep happening? Someone's got weird Xcode settings.

MapboxNavigationTests/LaneTests.swift
```diff
@@ -27,33 +27,33 @@ class LaneTests: FBSnapshotTestCase {
...
     }
-    
+
     func testSlightRight() {
         let view = LaneView(frame: CGRect(origin: .zero, size: CGSize(width: 30, height: 30)))
         view.backgroundColor = .white
         view.lane = Lane(indications: [.slightRight])
         view.maneuverDirection = .slightRight
         view.isValid = true
-        FBSnapshotVerifyView(view, suffixes: ["_64"])
```
JThramer: Again, this needs to stay. the `suffixes: ["_64"])` argument prevents FBSnapshotTests from looking in `ReferenceImages_32`, which we don't have. 

Really, all verify calls should be ` FBSnapshotVerifyView(view, suffixes: ["_64"])`.

It basically stops the testing engine from making extra noise on failures.

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -0,0 +1,353 @@
...
+
+    override open func pushTemplate(_ templateToPush: CPTemplate, animated: Bool) {
+        templateStack.append(templateToPush)
+    }
+
+    override open func popTemplate(animated: Bool) {
+        templateStack.removeLast()
+    }
+
+//    open func popToRootTemplate(animated: Bool)
```
JThramer: Comments.

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -0,0 +1,353 @@
...
+        return "CPInterfaceControllerSpy for \(context)"
+    }
+
+    // MARK - CPInterfaceController declarations and overrides
+
+    //Uncomment and override as needed, if needed. At first inspection, some of these result in semaphore_wait_trap when invoked on an instance of CPInterfaceControllerSpy, but `rootTemplate` works as expected once set
+
+    //TODO: solve for this duplication with introspection, à la Cedar::Doubles
+
+//    weak open var delegate: CPInterfaceControllerDelegate?
```
JThramer: comments

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -0,0 +1,353 @@
...
+                }
+            })
+        })
+    }
+}
+
+
+
+//MARK: Test Objects / Classes.
+//TODO: Extract into separate file at some point.
```
JThramer: todo?

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -0,0 +1,353 @@
...
+
+        // the CarPlayNavigationDelegate is notified
+        XCTAssertTrue(exampleDelegate.navigationInitiated, "The CarPlayManagerDelegate should have been told that navigation was initiated.")
+
+        manager.carPlayNavigationViewControllerDidArrive(manager.currentNavigator!)
+
+        XCTAssertTrue(exampleDelegate.navigationEnded, "The CarPlayManagerDelegate should have been told that navigation ended.")
+    }
+
+    //TODO: -mapTemplateDidCancelNavigation:
```
JThramer: did this task fall through the cracks?

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -0,0 +1,353 @@
...
+
+        let expectedEventName = MMEventTypeCarplayDisconnect
+        XCTAssertTrue(eventsManagerSpy.hasEnqueuedEvent(with: expectedEventName))
+        XCTAssertTrue(eventsManagerSpy.hasFlushedEvent(with: expectedEventName))
+    }
+
+    // MARK: Upon connecting to CarPlay, window and interfaceController should be set up correctly
+    // By default we should supply a search template with the Mapbox Geocoder hooked up (?)
+
+//    @available(iOS 12.0, *)
```
JThramer: comments

MapboxNavigationTests/CPMapTemplate+MBTestable.mm
```diff
@@ -0,0 +1,31 @@
...
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 120000
+
+#import <CarPlay/CarPlay.h>
+
+@interface CPMapTemplate (MBTestableInternal)
+
+@end
+
+
+//static char *CurrentTripKey;
```
JThramer: comment

MapboxNavigation/UIEdgeInsets.swift
```diff
@@ -8,6 +9,11 @@ public extension UIEdgeInsets {
                             bottom: left.bottom + right.bottom,
                             right: left.right + right.right )
     }
+    
+    static func >(lhs: UIEdgeInsets, rhs: UIEdgeInsets) -> Bool {
+        return (lhs.top + lhs.left + lhs.bottom + lhs.right)
```
JThramer: This is fine for now, but I would like to call out that this is a weird way to calculate `greaterThan.`

I would have expected `lhs.top > rhs.top && lhs.bottom > rhs.bottom...`

1ec5:
Important: remember to merge this PR with a merge commit, rather than rebasing or squashing.

MapboxNavigationTests/InstructionsBannerViewSnapshotTests.swift
1ec5: This isn’t a great venue to discuss whitespace settings, but for what it’s worth, I’m not as fond of unindenting blank lines. It tends to result in less intuitive merge conflicts by default. In any case, both GitHub and Xcode have options for hiding whitespace changes from the diff.

MapboxNavigationTests/NavigationViewControllerTests.swift
JThramer: 👍

MapboxNavigationTests/LaneTests.swift
JThramer: For Posterity:

Q:
> Can you expand a little bit? We only test on iPhone 6 Plus iOS 11 and 12. Why do we need the 64bit suffix?

A: 
> because by default `FBSnapshotView` looks at `ReferenceImages_32` as a fallback if it can't find the image in `ReferenceImages_64`. Since we don't do 32-bit builds, this does nothing but emit an extra, unnecessary error, which adds noise to the log.

frederoni: FWIW, I'd prefer extra noise on failure than extra noise in tests. Especially now when we have `./scripts/generate_snapshots.sh` so we don't need to manually find the missing snapshots.

JThramer: the "extra noise" conveys exactly 0 extra useful information, because it's always preceded by an error for the `ReferenceImages_64` lookup.

It's just extra cruft that impedes searching through CI logs, it is 100% unnecessary. 

Having things fail loudly is nice, but I'd rather not have any noise in the log at all.

frederoni: I get that the information is not useful, but I'd argue that the suffix parameter isn't useful either. Two wrongs don't make a right 😇 If you see a missing snapshot in the logs, 32 or 64 doesn't matter, just run `./scripts/generate_snapshots.sh` and the error goes away.

JThramer: @frederoni I feel quite strongly about this, as those errors were the source of some major frustration for me not too long ago. I understand the annoyance of having to write an extra argument on the verify call, so I went ahead and made a wrapper call `verify(someView)` that seems to me to be the best of both worlds: It's easier to write then `FBSnapshotVerifyView(someView)`, there are no extra arguments to worry about, and it adds the `suffixes` argument for you.

The change is [here.](
https://github.com/mapbox/mapbox-navigation-ios/pull/1709/commits/aee628955df40b60bd2f45c1c0604a54c5d4cff) What do you think?

PR #1710, opened 2018-09-17
Revert user puck dead reckoning

1ec5:
This reverts commit #1385, which resulted in the puck overshooting at turns and detaching from the route line along curved roads. In some cases, it may have exacerbated other issues with course tracking.

When we revisit this feature in the future, we’ll need to ensure that dead reckoning is only part of the solution: the result of dead reckoning needs to be snapped to the route line. That wasn’t happening before, so this feature tended to foil any attempt at snapping the user’s location to the route line, at least visually. As well, we’ll need to find an alternative to `CLLocation.speed`, which is an instantaneous reading that is documented to be unsuitable for dead reckoning.

/cc @mapbox/navigation-ios


frederoni:
Looks a lot better. Simulating at 9× is still suboptimal but this is an improvement.

PR #1717, opened 2018-09-17
Expose a way to configure the favorites list

frederoni:
Moved the favorites list from the SDK into the Example application and exposed `CarPlayManagerDelegate.carPlayManagerFavoritesList(_:)`

cc @akitchen @1ec5 @vincethecoder


MapboxNavigation/CarPlayManager.swift
```diff
@@ -95,7 +95,16 @@ public protocol CarPlayManagerDelegate {
...
+    
+    /**
+     Called when the carplay manager will present the favorites list.
+     
+     Implementing this method will allow developers to present a custom list of favorites triggered by the default trailing bar button.
+     
+     - parameter carPlayManager: The carplay manager that will present the list.
+     - returns: A `CPListTemplate` containing a list of favorites.
+     */
+    @objc optional func carPlayManagerFavoritesList(_ carPlayManager: CarPlayManager) -> CPListTemplate
```
akitchen: This doesn't explicitly need to be a favorites list in the delegate protocol. It should be bundled into the trailing template items instead.

Examples/Swift/AppDelegate+CarPlay.swift
```diff
@@ -47,6 +47,17 @@ extension AppDelegate: CarPlayManagerDelegate {
         NavigationViewController.carPlayManagerDidEndNavigation(carPlayManager, window: window)
     }
     
+    func carPlayManagerFavoritesList(_ carPlayManager: CarPlayManager) -> CPListTemplate {
+        let mapboxSFItem = CPListItem(text: FavoritesList.POI.mapboxSF.rawValue,
```
JThramer: This could _totally_ wait until after MVP, but it might be nice to have the client-developer return a tuple `(name: String, location: CLLocationCoordinate2D)` instead of burdening the dev with creating the `CarPlay` model-objects themselves. 

Just a thought.

akitchen: 👍

akitchen: That was very much the original idea. This is meant to be example code, but a more obvious factoring would help developers here.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -129,6 +128,7 @@ public class CarPlayManager: NSObject {
 
     public fileprivate(set) var mainMapTemplate: CPMapTemplate?
     public fileprivate(set) weak var currentNavigator: CarPlayNavigationViewController?
+    public static let CarPlayWaypointKey: String = "CPCarPlayWaypoint"
```
1ec5: The `CP` prefix is understood to be the CarPlay framework itself. If we have to prefix, then let’s use `MB`.

Examples/Swift/AppDelegate+CarPlay.swift
```diff
@@ -47,19 +47,42 @@ extension AppDelegate: CarPlayManagerDelegate {
         NavigationViewController.carPlayManagerDidEndNavigation(carPlayManager, window: window)
     }
     
+    func favoritesListTemplate() -> CPListTemplate {
+        let mapboxSFItem = CPListItem(text: FavoritesList.POI.mapboxSF.rawValue,
```
1ec5: For future reference, a more realistic example probably wouldn’t involve a formal symbol for each POI. Instead, there would be an array of POIs.

PR #1722, opened 2018-09-18
v0.21 prep

JThramer:
Twenty-one prep

Updating directions, adding snapshot method wrapper that I forgot about, updating fixtures, adding all-tests target that makes things easier to test, good stuff

/cc @mapbox/navigation-ios


MapboxCoreNavigation.podspec
```diff
@@ -40,7 +40,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxCoreNavigation"
 
-  s.dependency "MapboxDirections.swift", "~> 0.22.0"
+  s.dependency "MapboxDirections.swift", "~> 0.23.0"
```
1ec5: Also update MapboxNavigation-Documentation.podspec with this change.

PR #1725, opened 2018-09-18
Persistent annotations provided by the user

### LINKED ISSUE #1592: NavigationMapView should not remove existing application-specific annotations
SDK version:

Mapbox-iOS-SDK (4.2.0)
MapboxCoreNavigation (0.18.1):
MapboxDirections.swift (> 0.21.0)
MapboxMobileEvents (> 0.4)
Turf (> 0.1)
MapboxDirections.swift (0.21.0):
Polyline (> 4.2)
MapboxMobileEvents (0.4.3)
MapboxNavigation (0.18.1):
Mapbox-iOS-SDK (> 4.0)
MapboxCoreNavigation (= 0.18.1)
MapboxSpeech (> 0.0.1)
Solar (~> 2.1)
MapboxSpeech (0.0.1)
Installed with: CocoaPods
iOS: 11.1.2
iPad model: iPad mini MF069LL/A
Xcode version: 9.4.1

Steps to reproduce:
1. Add multiple annotations by calling navigationViewController.mapView.addAnnotations(annotations) before presenting the navigationViewController.
2. Present the navigationViewController.
3. Notice that the annotations are not added to the mapView.

Here is the part of the code in NavigationMapView from the SDK that removes these annotations (line 510):
screen shot 2018-08-08 at 3 09 33 pm

Prefer:
These annotations are not removed at the time of creation.

frederoni:
Fixes #1592 NavigationMapView should not remove existing application-specific annotations

cc @akitchen @1ec5 @joshpuente

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -533,19 +533,24 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         }
         
         if let lastLeg =  route.legs.last {
-            removeAnnotations(annotations ?? [])
-            let destination = MGLPointAnnotation()
+            removeAnnotations(annotationsToRemove() ?? [])
+            let destination = NavigationAnnotation()
             destination.coordinate = lastLeg.destination.coordinate
             addAnnotation(destination)
         }
     }
+    
+    func annotationsToRemove() -> [MGLAnnotation]? {
+        return annotations?.filter { $0 is NavigationAnnotation }
+    }
+    
     /**
      Removes all waypoints from the map.
      */
     @objc public func removeWaypoints() {
         guard let style = style else { return }
         
-        removeAnnotations(annotations ?? [])
+        removeAnnotations(annotationsToRemove() ?? [])
         
         if let circleLayer = style.layer(withIdentifier: waypointCircleIdentifier) {
             style.removeLayer(circleLayer)
--- a/MapboxNavigationTests/NavigationMapViewTests.swift
+++ b/MapboxNavigationTests/NavigationMapViewTests.swift
@@ -1,8 +1,45 @@
 import XCTest
 @testable import MapboxNavigation
+@testable import MapboxCoreNavigation
 import MapboxDirections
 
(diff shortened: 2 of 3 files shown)


MapboxNavigationTests/NavigationMapViewTests.swift
```diff
@@ -52,4 +89,23 @@ class NavigationMapViewTests: XCTestCase {
         XCTAssertEqual(congestionSegmentsSevere[1].1, .severe)
         XCTAssertEqual(congestionSegmentsSevere[2].1, .low)
     }
+    
+    func testAnnotations() {
```
akitchen: This looks great. My main feedback would be to use a more descriptive test method name, for example `testRemoveWaypointsDoesNotRemoveUserAnnotations()` or some such.

Also, the route used for testing appears to have two waypoints (line 16, above). Is it expected that showing waypoints for the route would only add one waypoint to the map?

frederoni: > Is it expected that showing waypoints for the route would only add one waypoint to the map

Yes, the first one typically represents the user’s current location which we don't show as an annotation.

akitchen: got it, thanks

MapboxNavigation/NavigationMapView.swift
```diff
@@ -533,19 +533,24 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-            let destination = MGLPointAnnotation()
+            removeAnnotations(annotationsToRemove() ?? [])
+            let destination = NavigationAnnotation()
             destination.coordinate = lastLeg.destination.coordinate
             addAnnotation(destination)
         }
     }
+    
+    func annotationsToRemove() -> [MGLAnnotation]? {
+        return annotations?.filter { type(of: $0) == NavigationAnnotation.self }
```
1ec5: Is there a need for NavigationAnnotation to be public? If it’s private, then we can filter on `$0 is NavigationAnnotation.self`.

PR #1728, opened 2018-09-18
remaining coordinates to end of step

d-prukop:
This PR adds a property on the `RouteProgress` object called `coordinatesRemaining`

This property is updated on every location update with a slice of the current route step's polyline from the current location to the end of the step.

### DIFF
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -432,6 +432,11 @@ open class RouteStepProgress: NSObject {
     }
     
     /**
+     Coordinates from current location to end of current step
+     */
+    @objc public var coordinatesRemaining: [CLLocationCoordinate2D] = []
+
+    /**
      Intializes a new `RouteStepProgress`.
 
      - parameter step: Step on a `RouteLeg`.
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -431,6 +431,11 @@ open class RouteStepProgress: NSObject {
         return (1 - fractionTraveled) * step.expectedTravelTime
     }
     
+    /**
+     Coordinates from current location to end of current step
+     */
+    @objc public var coordinatesRemaining: [CLLocationCoordinate2D] = []
```
vincethecoder: @d-prukop this should be an optional property. This will help determine the initial state vs. the final state. Basically, how could we detect that the `coordinatesRemaining` has not been set as opposed to when we slice a polyline and get an empty array in return?

d-prukop:
I'm abandoning this PR. The reason I wanted to add this to the nav SDK was because I saw an opportunity to optimize performance by not walking the polylines over and over again. It turns out that refactoring and testing these optimizations is a much bigger project. I'm moving this function out to my client app.

PR #1732, opened 2018-09-19
Fixed Flying Puck issue

### LINKED ISSUE #1731: UserCourseView should update instantaneously when user gestures cause camera changes
When the user manually rotates the map using gestures, the puck should rotate synchronously with the map. Currently, there’s a bit of a lag that, while smooth, makes the puck look like it’s on a turntable. Similarly and more noticeably, when the user manually pans the map using gestures, the puck should move synchronously with the map instead of casually mosying over to the new location.

Both these issues would be fixed by disabling UserCourseView’s animation (or setting the animation duration to 0&nbsp;seconds) when the map view’s camera changes for reasons other than course tracking. #1506 fixed the same issue, but only when transitioning between overview and turn-by-turn modes.

/cc @vincethecoder

apavani:
Ensures that puck is positioned in the right location immediately after the camera is set to the appropriate the location. 

Closes #1731 

cc @1ec5

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -337,9 +337,16 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             let padding = UIEdgeInsets(top: point.y, left: point.x, bottom: bounds.height - point.y, right: bounds.width - point.x)
             let newCamera = camera ?? MGLMapCamera(lookingAtCenter: location.coordinate, fromDistance: altitude, pitch: 45, heading: location.course)
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear) : nil
-            setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding, completionHandler: nil)
+            setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding) { [weak self] in
+                guard let strongSelf = self else { return }
+                if strongSelf.userAnchorPoint != strongSelf.userCourseView?.center ?? strongSelf.userAnchorPoint {
+                    UIView.animate(withDuration: 0, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
+                        strongSelf.userCourseView?.center = strongSelf.convert(location.coordinate, toPointTo: strongSelf)
+                    })
+                }
+            }
         }
-        if !tracksUserCourse || userAnchorPoint != userCourseView?.center ?? userAnchorPoint {
+        if !tracksUserCourse {
             UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
                 self.userCourseView?.center = self.convert(location.coordinate, toPointTo: self)
             })


akitchen:
This is fantastic, @vincethecoder @apavani !  Thanks for diving in together address this.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -337,9 +337,16 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+                guard let strongSelf = self else { return }
+                if strongSelf.userAnchorPoint != strongSelf.userCourseView?.center ?? strongSelf.userAnchorPoint {
+                    UIView.animate(withDuration: 0, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
+                        strongSelf.userCourseView?.center = strongSelf.convert(location.coordinate, toPointTo: strongSelf)
+                    })
+                }
+            }
         }
-        if !tracksUserCourse || userAnchorPoint != userCourseView?.center ?? userAnchorPoint {
+        if !tracksUserCourse {
```
1ec5: This change effectively reverts #1653. We’ll need to verify the results in CarPlay.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -337,9 +337,16 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             let padding = UIEdgeInsets(top: point.y, left: point.x, bottom: bounds.height - point.y, right: bounds.width - point.x)
             let newCamera = camera ?? MGLMapCamera(lookingAtCenter: location.coordinate, fromDistance: altitude, pitch: 45, heading: location.course)
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: kCAMediaTimingFunctionLinear) : nil
-            setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding, completionHandler: nil)
+            setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding) { [weak self] in
+                guard let strongSelf = self else { return }
+                if strongSelf.userAnchorPoint != strongSelf.userCourseView?.center ?? strongSelf.userAnchorPoint {
+                    UIView.animate(withDuration: 0, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
```
1ec5: If I’m reading this correctly, we’re chaining animations together, snapping the course view back to an accurate location _after_ an initial animation. But isn’t it that initial animation that looks incorrect?

MapboxNavigation/NavigationMapView.swift
1ec5: I just tried this in a CarPlay unit – it’s worse than what I saw in https://github.com/mapbox/mapbox-navigation-ios/issues/1731#issuecomment-423350724 on the simulator for some reason. The main map has the puck shifted upwards by about a third of the screen, and it isn’t possible to show the route at any time (whether in preview, turn-by-turn, or overview mode).

akitchen:
@frederoni can you also take a look please? 

We need to account for how this affects CarPlay as well as transitioning between overview and route following modes.

MapboxNavigation/NavigationMapView.swift
frederoni: As suggested in the ticket (#1731), I think setting the duration on the outer animation closure to 0 when a user is interacting with the map is a better idea.

frederoni:
CarPlay aside, I have a hard time seeing any regressions or improvements with this change compared to master.

MapboxNavigation/NavigationMapView.swift
frederoni: This patch seems to be fixing the turntable part.

<details>
<summary>turntable.patch</summary>

```
diff --git a/MapboxNavigation/NavigationMapView.swift b/MapboxNavigation/NavigationMapView.swift
index 3d2b9ea6..ceaf87da 100644
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -411,6 +411,16 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             guard let location = userLocationForCourseTracking else { return }
             userCourseView?.layer.removeAllAnimations()
             userCourseView?.center = convert(location.coordinate, toPointTo: self)
+            
+            if let view = userCourseView as? UserCourseView,
+                let location = userLocationForCourseTracking {
+                
+                view.update?(location: location,
+                             pitch: self.camera.pitch,
+                             direction: self.direction,
+                             animated: false,
+                             tracksUserCourse: self.tracksUserCourse)
+            }
         }
     }
     

```

</details>

nitaliano:
I'm closing this out, this doesn't seem like the solution.

vincethecoder:
We already have a ticket that's looking into this issue on the SDK board. So this is a duplicate.

Issue #1733, opened 2018-09-20
Avoid inconsistencies when a route doubles back on itself


1ec5:
When a route doubles back on itself and the user visits the same coordinate a second time, the SDK gets confused, momentarily thinking the user is visiting that coordinate for the first time. There are several ways to reproduce this issue, for example:

* A [Michigan left](https://en.wikipedia.org/wiki/Michigan_left), [superstreet](https://en.wikipedia.org/wiki/Superstreet), or other intersection with turn restrictions requires the user to re-approach a given intersection in order to make a left turn.
* A custom route, such as one that solves the [route inspection problem](https://en.wikipedia.org/wiki/Route_inspection_problem), reaches a four-way intersection from two different approaches and leaves on two different outlets. mapbox/navigation-ios-examples#28 demonstrates such a route.
* A loop ramp in a cloverleaf interchange happens to end directly over or under the spot where it begins.
* A route traverses [the upper deck of a double-decker bridge](https://www.openstreetmap.org/way/159588797#map=17/39.09025/-84.52264), then [the lower deck](https://www.openstreetmap.org/way/159588794).
* A route traverses a [spiral bridge](https://en.wikipedia.org/wiki/Spiral_bridge), such as in a parking garage.

Symptoms of this issue are:

* The maneuver arrow along the route line points in the wrong direction.
* During route simulation, the camera and puck jump or turn in the wrong direction.
* Prior to #1710, the puck would also jump even when not simulating the route.

The problem is that we’re passing a coordinate into `Polyline.trimmed(from:distance:)`, which assumes the coordinate can only appear once in the polyline:

https://github.com/mapbox/mapbox-navigation-ios/blob/f7ba632b731e7d8f8ccbad77df3ea07510bab395/MapboxNavigation/NavigationMapView.swift#L592-L593 https://github.com/mapbox/mapbox-navigation-ios/blob/f7ba632b731e7d8f8ccbad77df3ea07510bab395/MapboxCoreNavigation/SimulatedLocationManager.swift#L157

The most robust fix is to rely on indices into the coordinate array rather than the coordinates themselves. This approach will likely be necessary as part of the navigation-native refactoring. But that could make it much more difficult to memoize the remaining portion of the route, as in #1728.

[...]

PR #1735, opened 2018-09-21
Fix maneuver arrow when route doubles back on itself

1ec5:
Partition the route at the step’s index before assembling and trimming the route line around the maneuver.

before after<br>_Before and after_

Fixes the maneuver arrow part of #1733, but not the simulated location part.

/cc @frederoni


frederoni:
re: https://github.com/mapbox/mapbox-navigation-ios/pull/1735#issuecomment-423499811

the maneuver type seems to be encoding Int while decoding tries to decode an NSString.
Until the encoding/decoding is fixed in MapboxDirections, you could `@testable` import MapboxDirections and use: `MatchOptions.response(from:)` to serialize the json instead of encoding/decoding raw data.

MapboxNavigation/Route.swift
```diff
@@ -0,0 +1,41 @@
...
+     Returns a polyline extending a given distance in either direction from a given maneuver along the route.
+     
+     The maneuver is identified by a leg index and step index, in case the route doubles back on itself.
+     
+     - parameter legIndex: Zero-based index of the leg containing the maneuver.
+     - parameter stepIndex: Zero-based index of the step containing the maneuver.
+     - parameter distance: Distance by which the resulting polyline extends in either direction from the maneuver.
+     - returns: A polyline whose length is twice `distance` and whose centroid is located at the maneuver.
+     */
+    func polylineAroundManeuver(legIndex: Int, stepIndex: Int, distance: CLLocationDistance) -> Polyline {
```
frederoni: This looks good but isn't it still technically possible to have a step which could double back on itself?
I haven't really wrapped my head around this yet but why are we trimming instead of slicing? Trimming takes all coordinates within a radius, right? It's just a bit unlikely that a double backed part of the polyline wouldn't be another step. I think there's a single step which double backs on itself in Weehawken New Jersey, from-to `-74.01701276685586,40.76621002800809;-74.02209890200525,40.76787957422633`

If the assumption above is correct, do you think it would be possible to make some changes to the bottom part of this function to use `LineString.sliced(from:to:)` instead of `LineString.trimmed(from:distance:)` and rename `polylineAroundManeuver` to `polylineAlongManeuver`?

1ec5: > This looks really good but isn't it still technically possible to have a step which could double back on itself?

It is true that a single step can double back on itself, as in some of the cases mentioned in #1733. This also concerned me at first. Fortunately, a step can only have one maneuver, and we know for sure the position of that maneuver within the step’s coordinate array (the first index). So as long as the partition occurs at the right place, a step that doubles back on itself will still result in a half-shaft polyline that begins at the maneuver and extends the correct distance in either direction.

> do you think it would be possible to make some changes to the bottom part of this function to use `LineString.sliced(from:to:)` instead of `LineString.trimmed(from:distance:)`

By accepting coordinates as arguments, `LineString.sliced(from:to:)` assumes the polyline never doubles back on itself. Perhaps Turf should provide an indexed alternative to this method, along the lines of mapbox/turf-swift#67.

1ec5:
Added a unit test based on mapbox/navigation-ios-examples#28.

PR #1743, opened 2018-09-26
Optional Geocoder

frederoni:
Despite the Geocoder being optional, it's still required to build `MapboxNavigation`. `MapboxGeocoder.swift` is still being link against in the `Example-CarPlay` target.

This change removes `MapboxGeocoder.swift` from `MapboxNavigation`’s build phase.

@akitchen @1ec5  @chezzdev


chezzdev:
Thanks, @frederoni. This PR fixes broken Carthage distribution.

PR #1749, opened 2018-09-27
Make NavigationMapView’s frame rate more easily configurable

1ec5:
`NavigationMapView` no longer mutates its own frame rate implicitly. A new `NavigationMapView.updatePreferredFrameRate(for:)` method allows the developer to update the frame rate in response to route progress change notifications. Previously, if the developer wanted to customize the frame rate, they would’ve had to install their own notification observer anyways and hope that it would respond to route progress change notifications after NavigationMapView’s built-in observer.

This change gets NavigationMapView out of the business of snooping on the stream of route progress updates coming from every active route controller. This is particularly beneficial when CarPlay is connected, because there are two active route controllers in that case that may deliver staggered route progress updates. However, it does mean that a developer implementing a custom navigation UI will need to call `NavigationMapView.updatePreferredFrameRate(for:)` explicitly to get the power-saving behavior that NavigationMapView used to provide by default.

The `NavigationMapView.minimumFramesPerSecond` property determines the frame rate while the application runs on battery power. By default, the map views in `NavigationViewController` and CarPlay’s navigation activity animate at a higher frame rate on battery power than before: 30&nbsp;frames per second, up from 5&nbsp;frames per second and equivalent to the behavior when running plugged in. This change has the potential to shorten battery life, but users overwhelmingly perceived 5&nbsp;fps animation as a buggy stuttering effect rather than a best-effort animation.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -13,12 +13,18 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     // MARK: Class Constants
     
     struct FrameIntervalOptions {
-        fileprivate static let durationUntilNextManeuver: TimeInterval = 7
-        fileprivate static let durationSincePreviousManeuver: TimeInterval = 3
-        fileprivate static let defaultFramesPerSecond = MGLMapViewPreferredFramesPerSecond.maximum
-        fileprivate static let pluggedInFramesPerSecond = MGLMapViewPreferredFramesPerSecond.lowPower
-        fileprivate static let decreasedFramesPerSecond = MGLMapViewPreferredFramesPerSecond(rawValue: 5)
+        static let durationUntilNextManeuver: TimeInterval = 7
+        static let durationSincePreviousManeuver: TimeInterval = 3
+        static let defaultFramesPerSecond = MGLMapViewPreferredFramesPerSecond.maximum
+        static let pluggedInFramesPerSecond = MGLMapViewPreferredFramesPerSecond.lowPower
     }
+    
+    /**
+     The minimum preferred frames per second at which to render map animations.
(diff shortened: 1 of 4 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -13,13 +13,19 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
     struct FrameIntervalOptions {
-        fileprivate static let durationUntilNextManeuver: TimeInterval = 7
-        fileprivate static let durationSincePreviousManeuver: TimeInterval = 3
-        fileprivate static let defaultFramesPerSecond = MGLMapViewPreferredFramesPerSecond.maximum
-        fileprivate static let pluggedInFramesPerSecond = MGLMapViewPreferredFramesPerSecond.lowPower
-        fileprivate static let decreasedFramesPerSecond = MGLMapViewPreferredFramesPerSecond(rawValue: 5)
+        static let durationUntilNextManeuver: TimeInterval = 7
+        static let durationSincePreviousManeuver: TimeInterval = 3
+        static let defaultFramesPerSecond = MGLMapViewPreferredFramesPerSecond.maximum
+        static let pluggedInFramesPerSecond = MGLMapViewPreferredFramesPerSecond.lowPower
```
akitchen: This is a bit confusing (although this was also the case previously)

When plugged in we are using the low power setting?

1ec5: The map SDK calls it “low power”. It’s 30&nbsp;fps under the hood, which is what the navigation SDK had when plugged in well before #1608.

akitchen: 👍

PR #1750, opened 2018-09-28
Ability to swipe InstructionsBannerView

nitaliano:
* Added `swipeable` property to `InstructionsBannerView`
* Added `didSwipeInstructionsBanner` to `InstructionsBannerViewDelegate`
* Deprecated `didDragInstructionsBanner` for `InstructionsBannerViewDelegate`
* Updates `RouteMapViewController` to implement swipeable banner

@1ec5 @frederoni

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changes to the Mapbox Navigation SDK for iOS
 
 ## master
+* Added `swipeable` property to `InstructionsBannerView`
+* Added `didSwipeInstructionsBanner` to `InstructionsBannerViewDelegate`
+* Deprecated `didDragInstructionsBanner` for `InstructionsBannerViewDelegate`
 
 ### User location
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,7 @@
 # Changes to the Mapbox Navigation SDK for iOS
 
 ## master
-* Added `swipeable` property to `InstructionsBannerView`
-* Added `didSwipeInstructionsBanner` to `InstructionsBannerViewDelegate`
-* Deprecated `didDragInstructionsBanner` for `InstructionsBannerViewDelegate`
+* Added an InstructionsBannerView.swipeable property that allows the user to swipe the banner to the side to preview future steps. The InstructionsBannerViewDelegate.didDragInstructionsBanner(_:) method has been deprecated in favor of InstructionsBannerViewDelegate.didSwipeInstructionsBanner(_:swipeDirection:). (#1750)
 
 ### User location
 
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -37,32 +37,17 @@ open class RouteProgress: NSObject {
     }
 
     /**
+     Returns the remaining legs left on the current route
+     */
+    @objc public var remainingLegs: [RouteLeg] {
+        return Array(route.legs.suffix(from: legIndex + 1))
+    }
+
+    /**
      Returns the remaining steps left on the current route
      */
     @objc public var remainingSteps: [RouteStep] {
-        var steps = [RouteStep]()
-        
-        // get all remaining steps on currentLeg
-        let curStepIndex = currentLegProgress.stepIndex
-        var curLegSteps = currentLeg.steps.suffix(from: curStepIndex)
-        _ = curLegSteps.popLast()
-        
-        for step in curLegSteps {
-            steps.append(step)
-        }
-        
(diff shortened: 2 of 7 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -546,6 +556,15 @@ class RouteMapViewController: UIViewController {
             return populated(named)
         }
     }
+    
+    fileprivate func leg(with step: RouteStep) -> RouteLeg? {
```
1ec5: `leg(containing:)`

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -546,6 +556,15 @@ class RouteMapViewController: UIViewController {
...
+        for leg in route.legs {
+            if leg.steps.contains(step) {
+                return leg
```
1ec5: This is more succinctly written as `route.legs.first { $0.steps.contains(step) }`.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -208,11 +209,16 @@ class RouteMapViewController: UIViewController {
...
         isInOverviewMode = false
+
         updateCameraAltitude(for: router.routeProgress)
         
         mapView.addArrow(route: router.route,
                          legIndex: router.routeProgress.legIndex,
                          stepIndex: router.routeProgress.currentLegProgress.stepIndex + 1)
+
+        if currentPreviewInstructionBannerStepIndex != nil {
+            currentPreviewInstructionBannerStepIndex = nil
```
1ec5: Nit: I think you can set this unconditionally.

MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -18,8 +18,14 @@ public protocol InstructionsBannerViewDelegate: class {
     /**
      Called when the user drags either up or down on the `InstructionsBannerView`.
      */
+    @available(*, deprecated, message: "Please use didSwipeInstructionsBanner instead.")
```
1ec5: This PR causes `didDragInstructionsBanner(_:)` to no longer get called. Deprecating it would be misleading, because that would imply (especially to Objective-C code, via a warning) that it’s OK to keep using the old method for now. Either mark this method `obsoleted` or, preferably, conditionally call the old method if it’s implemented.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -36,6 +36,35 @@ open class RouteProgress: NSObject {
...
+        }
+        
+        // get all remaining steps after currentLeg
+        let remainingLegs = route.legs.suffix(from: legIndex + 1)
+        for leg in remainingLegs {
+            var curLegSteps = leg.steps
+            _ = curLegSteps.popLast()
+            
+            for step in curLegSteps {
+                steps.append(step)
```
1ec5: Implement a `remainingLegs` computed property like this:

```swift
return Array(legs.suffix(from: legIndex + 1))
```

Then this method becomes simply:

```swift
return currentLegProgress.remainingSteps + remainingLegs.flatMap { $0.steps }
```

The one wrinkle is that you might want to merge waypoint arrivals with the subsequent departures. We do that in a few places, but it would be misleading to do that as part of a simply named `RouteProgress.remainingSteps` method. So the caller could do something like this:

```swift
routeProgress.remainingSteps.filter { $0.maneuverType != .depart }
```

nitaliano:
@1ec5 I made all changes

MapboxNavigation/InstructionsBannerView.swift
nitaliano: I am calling `didDragInstructionsBanner(_:)` again

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changes to the Mapbox Navigation SDK for iOS
 
 ## master
+* Added `swipeable` property to `InstructionsBannerView`
+* Added `didSwipeInstructionsBanner` to `InstructionsBannerViewDelegate`
+* Deprecated `didDragInstructionsBanner` for `InstructionsBannerViewDelegate`
```
1ec5: Fully qualify these symbols for jazzy autolinking:

> * Added an `InstructionsBannerView.swipeable` property that allows the user to swipe the banner to the side to preview future steps. The `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method has been deprecated in favor of `InstructionsBannerViewDelegate.didSwipeInstructionsBanner(_:swipeDirection:)`. ([#1750](https://github.com/mapbox/mapbox-navigation-ios/pull/1750))

PR #1768, opened 2018-10-08
Offline routing

frederoni:
Note: this PR targets the `offline-routing` branch.

Started outlining offline routing.

This PR
- [x] Support for Offline Routing via `MapboxOfflineDirections`
- [x] Release v0.24.1 of `MapboxDirections.swift` with support for deserializing routes from external sources (fixes the remaining CI 🔴)
- [x] Release a new version of MapboxNavigationNative

Subsequent PR
- [ ] Mimic a `URLSessionDataTask` for `Directions.calculateOffline` to allow canceling a request and make it easier to combine and use fallbacks between online and offline directions.
- [ ] UI for downloading a region

cc @akitchen @JThramer


Cartfile
```diff
@@ -1,6 +1,6 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 3.1
-github "mapbox/MapboxDirections.swift" ~> 0.24
+github "mapbox/MapboxDirections.swift" "fred/public-result"
```
frederoni: Make a new release of MapboxDirections.swift

MapboxCoreNavigation/Offline.swift
```diff
@@ -0,0 +1,122 @@
...
+    
+    public static let shared = ExtendedDirections()
+    
+    static let offlineSerialQueueLabel = Bundle.mapboxCoreNavigation.bundleIdentifier!.appending(".offline")
+    
+    static let serialQueue = DispatchQueue(label: ExtendedDirections.offlineSerialQueueLabel)
+    
+    var _navigator: MBNavigator!
+    var navigator: MBNavigator {
+        //assert(OperationQueue.current?.underlyingQueue?.label == ExtendedDirections.offlineSerialQueueLabel,
```
frederoni: Fix this assertion

MapboxCoreNavigation/Offline.swift
```diff
@@ -0,0 +1,122 @@
...
+            self._navigator = MBNavigator()
+            
+            let tilePath = OfflineStorage.tilesPath.absoluteString.replacingOccurrences(of: "file://", with: "")
+            let localizationPath = OfflineStorage.translationsPath.absoluteString.replacingOccurrences(of: "file://", with: "")
+            
+            // navigator raises an exception after each try:
+            // 1: uncompressed tiles packed in a tar archive,
+            // 2: uncompressed tiles in directories
+            // 3: compressed tiles in directories
+            // We are using the 3rd option
```
frederoni: Should we capture these exceptions? If so, I think we would have to capture them in Obj-C.

MapboxCoreNavigation/Offline.swift
```diff
@@ -0,0 +1,122 @@
...
+            return true
+        } catch {
+            return false
+        }
+    }
+}
+
+public typealias RouteCompletionHandler = (_ waypoints: [Waypoint]?, _ routes: [Route]?, _ error: NSError?) -> Void
+
+public class ExtendedDirections: NSObject {
```
frederoni: `OfflineDirections` might be a better name. Perhaps the functionality could be added as an extension on the existing `Directions` object.

akitchen: What if `OfflineDirections` and `Directions` both conformed to a `DirectionsProtocol` protocol (although we'd still need to figure out naming).

MapboxCoreNavigation/Offline.swift
akitchen: If that's the case, do we need the `OfflineDirections` class to be written in ObjC? Or another intermediary wrapper perhaps? Obviously we can't have these exceptions (which seem like they could be used for control flow or cascading workflow options) to crash the containing app.

MapboxCoreNavigation/Offline.swift
```diff
@@ -0,0 +1,122 @@
...
+    open func calculate(_ options: RouteOptions, completionHandler: @escaping RouteCompletionHandler) {
+        
+        guard let navigationOptions = options as? NavigationRouteOptions else {
+            Directions.shared.calculate(options, completionHandler: completionHandler)
+            return
+        }
+        
+        switch navigationOptions.preferredRequestOption {
+        case .preferClientSide:
+            fatalError("Not yet implemented")
```
akitchen: I have to wonder - does this option deserve to exist? when would we prefer client side but still "fall back" to server-generated directions? lack of tiles?

MapboxCoreNavigation/Offline.swift
```diff
@@ -0,0 +1,122 @@
...
+        case .preferServerSide:
+            fatalError("Not yet implemented")
```
akitchen: what's the thinking here - preflighting an HTTP request is [discouraged/considered an anti-pattern](https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/WhyNetworkingIsHard/WhyNetworkingIsHard.html) these days

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,116 @@
...
+@objc(MBMapboxOfflineDirections)
+public class MapboxOfflineDirections: Directions, OfflineDirectionsProtocol {
+    
+    required public init(accessToken: String?, host: String?, tilesPath: String, translationsPath: String, completionHandler: @escaping OfflineDirectionsCompletionHandler) {
+        
+        super.init(accessToken: accessToken, host: host)
+        
+        OfflineDirectionsConstants.serialQueue.sync {
+            let tilesPath = tilesPath.replacingOccurrences(of: "file://", with: "")
+            let translationsPath = translationsPath.replacingOccurrences(of: "file://", with: "")
```
1ec5: This makes it look like `tilesPath` and `translationsPath` aren’t actually paths but rather file URL strings. If so, we should stick to the URL type rather than String and call them URLs.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,116 @@
...
+            
+            guard let data = result.json.data(using: .utf8) else {
+                return completionHandler(nil, nil, OfflineRoutingError.corruptRouteData as NSError)
+            }
+            
+            do {
+                let json = try JSONSerialization.jsonObject(with: data, options: []) as! [String: Any]
+                if let errorValue = json["error"] as? String {
+                    DispatchQueue.main.async {
+                        let error = NSError(domain: "..", code: 102, userInfo: [NSLocalizedDescriptionKey: errorValue])
```
1ec5: This is an unusual domain. Can it be replaced with a constant that reliably identifies this SDK?

MapboxCoreNavigationTests/Fixtures/routing/translations/zh-Hans.json
```diff
@@ -0,0 +1 @@
...stination}"},"default":{"default":"进入环岛","name":"通过环岛后驶入{way_name}","destination":"通过环岛后前往{destination}"}}},"roundabout turn":{"default":{"default":"{modifier}转弯","name":"{modifier}转弯，驶入{way_name}","destination":"{modifier}转弯，前往{destination}"},"left":{"default":"左转","name":"左转，驶入{way_name}","destination":"左转，前往{destination}"},"right":{"default":"右转","name":"右转，驶入{way_name}","destination":"右转，前往{destination}"},"straight":{"default":"继续直行","name":"继续直行，驶入{way_name}","destination":"继续直行，前往{destination}"}},"exit roundabout":{"default":{"default":"驶离环岛","name":"驶离环岛，驶入{way_name}","destination":"驶离环岛，前往{destination}"}},"exit rotary":{"default":{"default":"驶离环岛","name":"驶离环岛，驶入{way_name}","destination":"驶离环岛，前往{destination}"}},"turn":{"default":{"default":"{modifier}转弯","name":"{modifier}转弯，驶入{way_name}","destination":"{modifier}转弯，前往{destination}"},"left":{"default":"左转","name":"左转，驶入{way_name}","destination":"左转，前往{destination}"},"right":{"default":"右转","name":"右转，驶入{way_name}","destination":"右转，前往{destination}"},"straight":{"default":"直行","name":"直行，驶入{way_name}","destination":"直行，前往{destination}"}},"use lane":{"no_lanes":{"default":"继续直行"},"default":{"default":"{lane_instruction}"}}}}
```
1ec5: Is it necessary to check in all the translations from OSRM Text Instructions as test fixtures in this repository? Could OSRM Text Instructions be included as a submodule or NPM dependency?

MapboxCoreNavigationTests/OfflineRoutingTests.swift
```diff
@@ -0,0 +1,77 @@
+import XCTest
+import MapboxDirections
+@testable import MapboxCoreNavigation
+
+class OfflineRoutingTests: XCTestCase {
+
+    override func setUp() {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
```
1ec5: Nit: remove boilerplate code.

MapboxCoreNavigationTests/OfflineRoutingTests.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+    
+    func testOfflineDirections() {
+        let bundle = Bundle(for: OfflineRoutingTests.self)
+        let tilesPath = bundle.bundlePath.appending("/routing/liechtenstein")
+        let translationsPath = bundle.bundlePath.appending("/routing/translations")
```
1ec5: Nit: use `URL.appendingPathComponent(_:)`.

MapboxCoreNavigation/MBRouteController.m
```diff
@@ -30,3 +34,16 @@ - (NSString * _Nonnull)md5 {
...
 @end
+
+@implementation MBNavigator (additions)
+
+- (NSUInteger)setupRouter:(NSString *)tilesPath translationsPath:(NSString *)translationsPath {
+    
+    @try {
+        return [self configureRouterForTilesPath:tilesPath translationsPath:translationsPath];
+    } @catch (NSException *exception) {
+        return -1;
```
1ec5: It feels like we’re trying to run around unexpected exceptions coming out of MapboxNavigationNative. If `-configureRouterForTilesPath:translationsPath:` has expected failure modes, it should be the responsibility of MapboxNavigationNative to catch an exception and turn it into a C-style return code and NSError out parameter.

By default, [throwing an exception in Objective-C causes a leak](http://clang.llvm.org/docs/AutomaticReferenceCounting.html#exceptions), even under ARC, because the expectation is that an exception is a catastrophic error that leads to an immediate crash anyways. It doesn’t leak in Objective-C++, but whether the `@throw` happens in Objective-C or Objective-C++ is none of this SDK’s business.

MapboxCoreNavigation/MBRouteController.h
```diff
@@ -82,3 +83,9 @@ extern const MBRouteControllerNotificationUserInfoKey MBRouteControllerIsProacti
 @interface NSString (MD5)
 - (NSString * _Nonnull)md5;
 @end
+
+@interface MBNavigator (additions)
+
+- (NSUInteger)setupRouter:(NSString *)tilesPath translationsPath:(NSString *)translationsPath;
```
1ec5: Documentation would be great here. It wasn’t obvious to me that the return value is a tile count. From the implementation of this method, I initially got the impression that it was an error code of some sort.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,116 @@
+import Foundation
+import MapboxDirections
+import MapboxNavigationNative
+
+public typealias OfflineDirectionsCompletionHandler = (_ numberOfTiles: UInt) -> Void
+
+public enum OfflineRoutingError: Error {
```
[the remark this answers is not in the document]
frederoni: These errors are casted to an NSError but I guess it needs a localized description key?

MapboxCoreNavigation/MBRouteController.m
frederoni: Returning NSErrors from Obj-C has a really poor bridgeability to Swift’s "throws". Noticed that this try-catch block doesn't suppress the underlying exception. It's just a fallback for the kind of resource the offline router is looking for. (1: Uncompress tiles, 2: Compressed tiles, 3: Different structure of compressed tiles). It's not a fatal error but one that gets caught if you have "Catch all Exceptions" enabled.

1ec5: > Returning NSErrors from Obj-C has a really poor bridgeability to Swift’s "throws".

Unless I’m mistaken, the most idiomatic method signature would include an `NSError **` out parameter, which would bridge to Swift as a `throws`, but it wouldn’t return an error.

MapboxCoreNavigation/OfflineDirections.swift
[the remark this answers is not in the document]
1ec5: Hm, yeah, the localized description keys would need to be provided somehow, though I’m not sure exactly how that works when relying on bridging from Error in Swift.

MapboxCoreNavigation/MBRouteController.m
frederoni: > the most idiomatic method signature would include an NSError ** out parameter

this is correct but there are some limitations, a boolean value is the only supported primitive return type:
```objc
- (void)voidNoThrow:(int)arg error:(NSError **)error;      // Bridges to NSErrorPointer
- (NSUInteger)intNoThrow:(int)arg error:(NSError **)error; // Bridges to NSErrorPointer
- (BOOL)boolThrow:(int)arg error:(NSError **)error;        // Throws idiomatically
```

However, there seems to be a consensus about using `-fno-exceptions` eventually, partially due to bloated binary size but the inevitable leak in Obj-C is also a good point. Either way, this try-catch block is already removed.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,116 @@
...
+     Begins asynchronously calculating the route or routes using the given options and delivers the results to a closure.
+     
+     This method retrieves the routes asynchronously via MapboxNavigationNative.
+     
+     Routes may be displayed atop a [Mapbox map](https://www.mapbox.com/maps/). They may be cached but may not be stored permanently. To use the results in other contexts or store them permanently, [upgrade to a Mapbox enterprise plan](https://www.mapbox.com/directions/#pricing).
+     
+     - parameter options: A `RouteOptions` object specifying the requirements for the resulting routes.
+     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread.
+     */
+    func calculateOffline(_ options: RouteOptions, completionHandler: @escaping Directions.RouteCompletionHandler)
```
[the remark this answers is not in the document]
frederoni: Combining sounds good but I think we can be sparse with the prepositions since it doesn't read fluently anyway.

MapboxCoreNavigation/MBRouteController.h
frederoni: This comment is not relevant any longer. `configureRouter` is not wrapped in `setUpRouter` but instead called directly from an initializer in Swift which is documented.

MapboxCoreNavigationTests/Fixtures/routing/translations/zh-Hans.json
frederoni: Included OSRMTI as a submodule.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,140 @@
...
+     Routes may be displayed atop a [Mapbox map](https://www.mapbox.com/maps/). They may be cached but may not be stored permanently. To use the results in other contexts or store them permanently, [upgrade to a Mapbox enterprise plan](https://www.mapbox.com/directions/#pricing).
+     
+     - parameter options: A `RouteOptions` object specifying the requirements for the resulting routes.
+     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread.
+     */
+    func calculate(_ options: RouteOptions, offline: Bool, completionHandler: @escaping Directions.RouteCompletionHandler)
+}
+
+@objc(MBMapboxOfflineDirections)
+public class MapboxOfflineDirections: Directions, OfflineRoutingProtocol {
```
frederoni: Not sure about the name of this class. How about `NavigationDirections`?

frederoni:
> Are any changes to documentation needed?

Yep, we do need some additions to the documentation, but I figured, since we're targeting a feature branch (`offline-routing`) and `api-routing-tiles` is not yet deployed to production, this is subject to change. Updates in the README and a guide on how to use the [offline tile downloader](https://github.com/mapbox/MapboxDirections.swift/pull/303) is slated for a subsequent PR.

annathomasQB:
Hi @frederoni ,

I'm trying to implement offline routing with the examples provided in the documents. But it doesn't seem to work. I'm trying to figure out where I might have gone wrong. I'm getting the tiles downloaded (is the max limit set to 10?) for Faroe Islands, unpacked at URL - all that's happening... I'm guessing the configureRouter too is getting set properly. Is there anything else left to be done ? how do I assign this to the navigationController? I've raised my issue here : [Downloaded Tiles not shown for Offline Routing](https://github.com/mapbox/mapbox-navigation-ios/issues/2256)

PR #1772, opened 2018-10-10
Swipeable Banner Custom UI

nitaliano:
* Adds ` InstructionsBannerView.showStepIndicator` so we have control over whether we want to show the indicator in our custom UI or not. Not everyone will show the steps list from the top so this needs to be configurable.
* Makes `BaseInstructionsBannerView.init(frame: CGRect)` public so we can actually initialize this outside of the SDK.
* Makes `NavigationMapView.enableFrameByFrameCourseViewTracking` public so when you want to move the map camera in navigation, this can help prevent the puck from flying across the screen.

@mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/InstructionsBannerView.swift
+++ b/MapboxNavigation/InstructionsBannerView.swift
@@ -87,7 +87,7 @@ open class BaseInstructionsBannerView: UIControl {
         }
     }
     
-    override init(frame: CGRect) {
+    public override init(frame: CGRect) {
         super.init(frame: frame)
         commonInit()
     }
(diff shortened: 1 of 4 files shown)


MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -48,9 +48,18 @@ open class BaseInstructionsBannerView: UIControl {
     @IBInspectable
     public var swipeable: Bool = false
     
+    @IBInspectable
+    public var showStepIndicator: Bool = true {
```
akitchen: Could you please add a line in the changelog for this addition?

nitaliano: sorry forgot will update!

PR #1779, opened 2018-10-15
Add CI bot to test cocoapods update

frederoni:
In addition to `pod install` we should test `pod update` to make sure we're compliant with recent releases of the dependencies.

cc @akitchen @JThramer @vincethecoder

### DIFF
--- a/MapboxCoreNavigation.podspec
+++ b/MapboxCoreNavigation.podspec
@@ -41,9 +41,9 @@ Pod::Spec.new do |s|
   s.module_name = "MapboxCoreNavigation"
 
   s.dependency "MapboxNavigationNative", "~> 3.1"
-  s.dependency "MapboxDirections.swift", "~> 0.24.0"
-  s.dependency "MapboxMobileEvents", "~> 0.6.0"
-  s.dependency "Turf", "~> 0.2"
+  s.dependency "MapboxDirections.swift", "~> 0.24.0"    # Always pin to a patch release if pre-1.0
+  s.dependency "MapboxMobileEvents", "~> 0.6.0"         # Always pin to a patch release if pre-1.0
+  s.dependency "Turf", "~> 0.2.0"                       # Always pin to a patch release if pre-1.0
 
   # `swift_version` was introduced in CocoaPods 1.4.0. Without this check, if a user were to
   # directly specify this podspec while using <1.4.0, ruby would throw an unknown method error.
--- a/circle.yml
+++ b/circle.yml
@@ -6,7 +6,8 @@ workflows:
     jobs:
       - xcode-10
       - xcode-9
-      - xcode-10-cocoapods-integration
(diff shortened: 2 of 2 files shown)


MapboxCoreNavigation.podspec
```diff
@@ -42,7 +42,7 @@ Pod::Spec.new do |s|
 
   s.dependency "MapboxNavigationNative", "~> 3.1"
   s.dependency "MapboxDirections.swift", "~> 0.24.0"
-  s.dependency "MapboxMobileEvents", "~> 0.5"
+  s.dependency "MapboxMobileEvents", "~> 0.5.0"
```
vincethecoder: @frederoni this should actually be `0.5` not `0.5.0`. Current `0.5` doesn't contain the updated strings reflected in `0.6`

MapboxCoreNavigation.podspec
```diff
@@ -42,7 +42,7 @@ Pod::Spec.new do |s|
 
   s.dependency "MapboxNavigationNative", "~> 3.1"
   s.dependency "MapboxDirections.swift", "~> 0.24.0"
```
vincethecoder: Ideally, we should be dropping the patch version ... it could cause issues down the road. Please update to `0.24`

MapboxCoreNavigation.podspec
frederoni: `0.5` actually breaks the `xcode-10-cocoapods-update` bot. Check the status on the first commit. Are you saying we should upgrade to `0.6`?

vincethecoder: @frederoni yes, we should upgrade at this point.

MapboxCoreNavigation.podspec
frederoni: `0.24` resolves to `0.99.99`. `0.24.0` resolves to `0.24.99`.

vincethecoder: Ok. got it!

MapboxCoreNavigation.podspec
JThramer: We should have a comment in the spec that explicitly calls out the need for `0.x.0` pinning schema.

frederoni:
CI caught the issue in https://github.com/mapbox/mapbox-navigation-ios/pull/1779/commits/1b53c249fcb16043dde1a677710b563568e050a6 so we know it's working.

MapboxCoreNavigation.podspec
akitchen: Indeed - http://guides.cocoapods.org/using/the-podfile.html

MapboxCoreNavigation.podspec
akitchen: Yes, we should be doing this consistently to avoid external breakage

PR #1782, opened 2018-10-16
MapboxNavigationService over-retain fix.

### LINKED ISSUE #1776: Memory leak in 0.22.1
Hi!

After upgrading to Mapbox-navigation-ios version 0.22.1 it looks like there's a memory leak that prevents the NavigationService and it's RouteController from being released. We see this in our own app and notice the same behaviour in the sample app navigation-ios-examples using e.g. the "basic" example; every time a new navigation is started a new NavigationService is created and the old one is still hanging around.

Best,
Marcel

JThramer:
Caaaaaan-dooo!
Fixes #1776. 🇺🇸🎉 

* Fixing issue where using function as closure can result in unwanted scope capture. Rewriting to work as trailing closure.

### DIFF
--- a/MapboxCoreNavigation/CountdownTimer.swift
+++ b/MapboxCoreNavigation/CountdownTimer.swift
@@ -22,7 +22,7 @@ class CountdownTimer {
     
     private(set) var state: State = .disarmed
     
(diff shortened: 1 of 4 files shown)


frederoni:
Awesome, we were on the right track. Can you prove the hypothesis with a leak test?
Also, two required CI bots are not running, please rebase atop master.

PR #1783, opened 2018-10-16
Let's not strip symbols out of debug builds

JThramer:
Dwarfs and DIPs and dSYMs, oh my!

* Fixing issue where we were stripping symbols out of debug-builds of MapboxNavigation.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation.xcodeproj/project.pbxproj
+++ b/MapboxNavigation.xcodeproj/project.pbxproj
@@ -2674,7 +2674,7 @@
 			buildSettings = {
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
-				DEBUG_INFORMATION_FORMAT = dwarf;
+				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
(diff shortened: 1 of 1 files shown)


frederoni:
"dwarf" debug information for debug builds is the default. The dSYM doesn't add any valuable information unless we're trying to symbolicate a crash log that came through ASC/TestFlight from an App which was built with debugging dependencies. Otherwise, it will just add a pretty massive overhead in size and build/test time.

If you're trying to solve `"error: Couldn't IRGen expression, no additional error" `, I think that's unrelated. It goes away with the regular unintuitive "Clean Build Folder/Restart Xcode/Rebuild" dependencies process.

frederoni:
Sorry about the confusing statement. The default in Xcode if you don't provide a value is "dwarf with dsym" but the default Xcode use to override the default when you create a new app or framework is "dwarf".

If Profiling doesn't work, then the "Profile" target’s build configuration should be changed to Debug instead of Release. Profiling usage in Release mode is still recommended for certain tasks, but it's correct that you can't see all symbols.

JThramer:
Posterity: Closing after a talk with Fred. The convenience of this setting is not worth increasing build times by 100%.

PR #1787, opened 2018-10-16
Release v0.19.3

frederoni:
Fixes an issue which caused 0.19.2 to incorrectly install incompatible dependencies upon `pod update`.

cc @1ec5 @akitchen @vincethecoder


akitchen:
What's up with circle CI here? Do we need to merge in the changes bringing us to Circle 2.0?

JThramer:
@akitchen Yes, I'd imagine we need to pull from master. I just had to do it for #1782.

frederoni:
Yep, we were using Bitrise at v0.19. Some tests (including the CocoaPods tests) are unfortunately not currently compatible with the current CircleCI setup. 🤔

PR #1790, opened 2018-10-19
Set initial location

frederoni:
A navigation session always begins from the first coordinate on the route but it can take a few seconds until we get a fixed location. By using the first coordinate on the route if the location source doesn't have a fixed location, we can improve the perceived lag and reduce initial unwanted reroute.

cc @JThramer

### DIFF
--- a/MapboxCoreNavigation/NavigationService.swift
+++ b/MapboxCoreNavigation/NavigationService.swift
@@ -315,6 +315,14 @@ public class MapboxNavigationService: NSObject, NavigationService, DefaultInterf
     }
     
     public func start() {
+        // Jump to the first coordinate on the route if the location source does
+        // not yet have a fixed location.
+        if router.location == nil,
+            let coordinate = route.coordinates?.first {
+            let location = CLLocation(latitude: coordinate.latitude, longitude: coordinate.longitude)
+            router.locationManager?(nativeLocationSource, didUpdateLocations: [location])
+        }
+        
         nativeLocationSource.startUpdatingHeading()
         nativeLocationSource.startUpdatingLocation()
         


MapboxCoreNavigation/NavigationService.swift
```diff
@@ -315,6 +315,14 @@ public class MapboxNavigationService: NSObject, NavigationService, DefaultInterf
     }
     
     public func start() {
+        // Jump to the first coordinate on the route if the location source does
+        // not yet have a fixed location.
+        if router.location == nil,
+            let coordinate = route.coordinates?.first {
+            let location = CLLocation(latitude: coordinate.latitude, longitude: coordinate.longitude)
```
JThramer: While this is cool, I do have a concern that this could cause the `poorGPSTimer` to arm before we get an actual, organic location from the GPS. No bueno.

How about this instead? 

```suggestion
            let location = CLLocation(coordinate: coordinate, altitude: -1, horizontalAccuracy: -1, verticalAccuracy: -1, course: -1, speed: 0, timestamp: Date())
```

akitchen: what do the `-1`'s do here? give us a provisional location which is not completely valid?

JThramer: They mean "Invalid." As in "there is no metric here." Specifically this will make the `RouteController` treat the location as "Non-Qualified", which will [update the UI](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/RouteController.swift#L306), but not much else.

akitchen: That's a more specific way of affirming what I was asking, and makes sense.

JThramer: Actually -- wait, this is passing the message directly onto the router. It bypasses the Timer arming mechanism. My apologies -- I think this will work just fine.

JThramer: However, upon a closer reading of `RouteController`, we may want to make this unqualified anyways, as the `RouteController` behavior changes once it encounters a qualified update. It would be weird if we triggered that logic with a synthetic update.

frederoni: It looks like it doesn't matter if we use a qualified or non-qualified for the first location since we're flagging `hasFoundOneQualifiedLocation` which grabs the first location no matter how inaccurate it is.

JThramer: Posterity: Gained consensus on this with Fred. Will be making change.

PR #1803, opened 2018-10-29
Offline Example

frederoni:
_this branch is targeting the `offline-routing` branch._

Outlined an example on how to implement offline routing.

- [x] Download a pre-defined region
- [x] Draw an arbitrary region on a map view 
- [x] ~~***blocked***~~ Unpack downloaded tar via `NN.unpack(:)` 
- [x] List downloaded packs
- [x] Show unpacking progress
- [x] Release a new version of MapboxDirections.swift with https://github.com/mapbox/MapboxDirections.swift/pull/303

cc @akitchen @1ec5


Examples/Swift/SettingsItems.swift
```diff
@@ -0,0 +1,65 @@
...
+        let downloadSFItem = Item(title: "Download SF region", viewControllerType: nil, payload: {
+            
+            Directions.shared.availableOfflineVersions(completionHandler: { (versions, error) in
+                guard let version = versions?.first else { return }
+                let sfBoundingBox = BoundingBox([CLLocationCoordinate2D(latitude: 37.7890, longitude: -122.4337),
+                                                 CLLocationCoordinate2D(latitude: 37.7881, longitude: -122.4318)])
+                
+                Directions.shared.downloadTiles(for: sfBoundingBox, version: version, completionHandler: { (url, response, error) in
+                    guard let url = url else { return }
+                    print("Downloaded \(url)")
```
akitchen: Shouldn't we unpack at this point?

frederoni: Ended up removing this example since it's basically the same as downloading an arbitrary region.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -138,3 +176,15 @@ fileprivate func currentQueueName() -> String? {
     let name = __dispatch_queue_get_label(nil)
     return String(cString: name, encoding: .utf8)
 }
+
+extension URL {
+    
+    func fileSize() -> UInt64? {
```
JThramer: computed property, not function.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -77,6 +82,39 @@ public class NavigationDirections: Directions, OfflineRoutingProtocol {
...
+    public class func unpackTilePack(at filePath: URL, outputDirectory: URL, progressHandler: UnpackProgressHandler?, completionHandler: UnpackCompletionHandler?) {
+        
+        OfflineDirectionsConstants.offlineSerialQueue.sync {
+            
+            let totalPackedBytes = filePath.fileSize()!
+            
+            // Report 0% progress
+            progressHandler?(totalPackedBytes, totalPackedBytes)
+
+            var timer: CountdownTimer? = CountdownTimer(countdown: .seconds(500), accuracy: .seconds(500), executingOn: OfflineDirectionsConstants.unpackSerialQueue, payload: {
```
JThramer: nit: use the trailing-closure pattern here.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -77,6 +82,39 @@ public class NavigationDirections: Directions, OfflineRoutingProtocol {
         }
     }
     
+    public class func unpackTilePack(at filePath: URL, outputDirectory: URL, progressHandler: UnpackProgressHandler?, completionHandler: UnpackCompletionHandler?) {
+        
+        OfflineDirectionsConstants.offlineSerialQueue.sync {
+            
+            let totalPackedBytes = filePath.fileSize()!
```
JThramer: What if the file doesn't exist? do we error-handle here?

MapboxCoreNavigation/OfflineDirections.swift
JThramer: Also, why is the timer an optional here? AFAIK `CountdownTimer` doesn't have a bailable initializer.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -77,6 +82,39 @@ public class NavigationDirections: Directions, OfflineRoutingProtocol {
...
+            progressHandler?(totalPackedBytes, totalPackedBytes)
+
+            var timer: CountdownTimer? = CountdownTimer(countdown: .seconds(500), accuracy: .seconds(500), executingOn: OfflineDirectionsConstants.unpackSerialQueue, payload: {
+                let remainingBytes = filePath.fileSize()!
+                progressHandler?(totalPackedBytes, remainingBytes)
+            })
+
+            timer?.arm()
+            
+            let tilePath = filePath.absoluteString.replacingOccurrences(of: "file://", with: "")
```
JThramer: Constantize: `fileSchemaIdentifier`?

Examples/Swift/SettingsItems.swift
```diff
@@ -0,0 +1,52 @@
...
+            if let sizeOfDirectory = sizeOfDirectory(at: path) {
+                subtitle = byteCountFormatter.string(fromByteCount: Int64(sizeOfDirectory))
+            }
+            versions.append(Item(title: $0, subtitle: subtitle, canEditRow: true))
+        }
+        
+        return versions
+    }
+    
+    func sizeOfDirectory(at path: URL) -> Int? {
```
JThramer: why not `URL.directorySize`?

Examples/Swift/OfflineViewController.swift
```diff
@@ -0,0 +1,89 @@
...
+        super.viewDidLoad()
+        
+        mapView = MGLMapView(frame: view.bounds)
+        view.addSubview(mapView)
+        
+        backgroundLayer.frame = view.bounds
+        backgroundLayer.fillColor = #colorLiteral(red: 0.1450980392, green: 0.2588235294, blue: 0.3725490196, alpha: 0.196852993).cgColor
+        view.layer.addSublayer(backgroundLayer)
+        
+        resizableView = ResizableView(frame: CGRect(origin: view.center,
```
JThramer: Nit: This is line-broken in a weird way.

MapboxCoreNavigation/OfflineDirections.swift
frederoni: b/c I wanted to disarm it before nilling it out at the end. Not even keeping a reference could be an option? 🤔

MapboxCoreNavigation/OfflineDirections.swift
frederoni: Even better, use `URL.path` to omit the schema.

MapboxCoreNavigation/OfflineDirections.swift
frederoni: We talked about functions indicating a non-trivial calculation vs computed property and since this is I/O, I went with function. However, I changed this to a computed property.

Examples/Swift/OfflineViewController.swift
frederoni: fixed

PR #1808, opened 2018-11-01
Offline routing

frederoni:
This PR brings offline routing to the Navigation SDK

- [x] Offline routing API #1768
- [x] Download examples #1803 
- [x] Offline tile downloader https://github.com/mapbox/MapboxDirections.swift/pull/303
- [ ] ~Automatic failover~
- [ ] ~Wrap `NavigationDirections.calculate(_:offline:completionHandler:)` in a `URLSessionDataTask` or another operation that supports canceling~
- [ ] ~Pause/Resume a download operation~
- [x] Replace OSRMTI submodule by using bundled translation files by default

cc @akitchen @1ec5 @JThramer @vincethecoder


akitchen:
Some thoughts on a few of the remaining checklist items:

> Automatic failover

Should this be a gating feature at this point? I'm not entirely convinced but it should be open to debate. Also, if we are going to do this we should have a small group design discussion about the mechanism. I could see this being a candidate for a boolean delegate method or configuration option.

> Wrap `NavigationDirections.calculate(_:offline:completionHandler:)` in a `URLSessionDataTask` or another operation that supports canceling

As we discussed, if we're going to use an `NSURLSessionDataTask`-like abstraction for this, perhaps the client-side routes should be loaded via a custom URL scheme filtered by an  `NSURLProtocol`, e.g. `mbx-offline://route?options=values` -- there is a bit of complexity there but the abstraction is close to what we need. And a route is technically a resource (which takes time to generate/load) so it feels kind of right.

> Pause/Resume a download operation

I have a feeling this will punt to a further iteration. Especially considering that there are other potential changes to tile transport currently under debate which could change the scope of this.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -457,8 +457,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             source.shape = polylines
             sourceSimplified.shape = mainPolylineSimplified
         } else {
-            let lineSource = MGLShapeSource(identifier: sourceIdentifier, shape: polylines, options: [.lineDistanceMetrics: true])
-            let lineCasingSource = MGLShapeSource(identifier: sourceCasingIdentifier, shape: mainPolylineSimplified, options: [.lineDistanceMetrics: true])
+            let lineSource = MGLShapeSource(identifier: sourceIdentifier, shape: polylines, options: nil)
```
akitchen: Maybe this could be a separate PR? It's not specific to offline routing and we probably want to be able to cite it (and easily revert it) separately

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,194 @@
...
+    public typealias UnpackCompletionHandler = (_ result: UInt64, _ error: Error?) -> ()
+    
+    public override init(accessToken: String? = nil, host: String? = nil) {
+        super.init(accessToken: accessToken, host: host)
+    }
+    
+    public func configureRouter(tilesURL: URL, translationsURL: URL? = nil, completionHandler: @escaping OfflineDirectionsCompletionHandler) {
+        
+        OfflineDirectionsConstants.offlineSerialQueue.sync {
+            // Translations files bundled winthin navigation native
```
akitchen:
```suggestion
            // Translations files bundled within navigation native
```

akitchen:
LGTM. Great work!

akitchen:
I've given a 👍 - @mapbox/navigation-ios any other feedback?

Example/OfflineViewController.swift
```diff
@@ -0,0 +1,91 @@
...
+            
+            let nonEmptyVersions = versions?.filter { !$0.isEmpty }
+            guard let version = nonEmptyVersions?.first else { return }
+            
+            self?.updateTitle("Downloading tiles")
+            
+            Directions.shared.downloadTiles(in: coordinateBounds, version: version, completionHandler: { (url, response, error) in
+                guard let url = url else { return assert(false, "Unable to locate temporary file") }
+                
+                let outputDirectoryURL = Bundle.mapboxCoreNavigation.suggestedTilePathURL(for: version)
```
1ec5: I’m not entirely sure it sets a good precedent to expose `Bundle.mapboxCoreNavigation` publicly. There’s little potential for collision with another library’s extension to Bundle, but maybe it would be better if applications explicitly called `Bundle(for:)`? Or if we expect applications to have to call `Bundle.suggestedTilePathURL(for:)`, then maybe we should make it accessible from a class method and leave `mapboxCoreNavigation` as an implementation detail.

MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -29,7 +29,29 @@ extension Bundle {
...
+            return nil
+        }
+        
+        guard let bundleIdentifier = self.bundleIdentifier else { return nil }
+        let path = URL(fileURLWithPath: cachesDirectory, isDirectory: true).appendingPathComponent(bundleIdentifier)
+        
+        return path.appendingPathComponent("tiles")
+    }
+    
+    public func suggestedTilePathURL(for version: String) -> URL? {
```
1ec5: These public methods should be documented and made accessible to Objective-C (via NSBundle).

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,194 @@
+import Foundation
+import MapboxDirections
+import MapboxNavigationNative
+
+public typealias OfflineDirectionsCompletionHandler = (_ numberOfTiles: UInt64) -> Void
```
1ec5: This type alias appears to be part of a public API, not merely exposed publicly so that it can be used in MapboxNavigation. It should be documented so that the protocol methods using it don’t end up with an enigmatic argument.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,194 @@
...
+            super.calculate(options, completionHandler: completionHandler)
+            return
+        }
+        
+        let url = self.url(forCalculating: options)
+        
+        OfflineDirectionsConstants.offlineSerialQueue.async { [weak self] in
+            
+            guard let result = self?.navigator.getRouteForDirectionsUri(url.absoluteString) else {
+                let error = OfflineRoutingError.unexpectedRouteResult("Unexpected routing result")
```
1ec5: We should make these error messages localizable, since they end up in `Error.localizedDescription` and would likely be presented to the user in an alert dialog.

MapboxCoreNavigationTests/OfflineRoutingTests.swift
```diff
@@ -0,0 +1,125 @@
...
+
+        // Coordinates within Liechtenstein
+        let coordinates = [CLLocationCoordinate2D(latitude: 47.208674, longitude: 9.524650),
+                           CLLocationCoordinate2D(latitude: 47.211247, longitude: 9.526666)]
+
+        let options = NavigationRouteOptions(coordinates: coordinates, profileIdentifier: .automobile)
+        let calculateRouteExpectation = expectation(description: "Calculate route offline")
+        var route: Route?
+
+        directions.calculate(options, offline: true) { (waypoints, routes, error) in
```
1ec5: Can we verify somehow that nothing is fetched from an online source?

MapboxCoreNavigationTests/OfflineRoutingTests.swift
```diff
@@ -0,0 +1,125 @@
...
+        
+        var progressUpdates = 0
+        let unpackExpectation = self.expectation(description: "Tar file should be unpacked")
+        let progressExpectation = self.expectation(description: "Progress should be reported")
+        
+        NavigationDirections.unpackTilePack(at: packURL, outputDirectoryURL: outputDirectoryURL, progressHandler: { (totalBytes, unpackedBytes) in
+            
+            progressUpdates += 1
+            
+            if (progressUpdates >= 2) {
```
1ec5: XCTTestExpectation has an `expectedFulfillmentCount` property that we can use instead of defining our own counter.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -1182,6 +1202,19 @@
 			name = Resources;
 			sourceTree = "<group>";
 		};
+		3529FCEE21A5C52C00AEA9AA /* Recovered References */ = {
```
1ec5: Check whether these recovered references belong anywhere else in the project; if not, feel free to delete the group.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,194 @@
...
+
+struct OfflineDirectionsConstants {
+    static let offlineSerialQueueLabel = Bundle.mapboxCoreNavigation.bundleIdentifier!.appending(".offline")
+    static let unpackSerialQueueLabel = Bundle.mapboxCoreNavigation.bundleIdentifier!.appending(".offline.unpack")
+    static let offlineSerialQueue = DispatchQueue(label: OfflineDirectionsConstants.offlineSerialQueueLabel)
+    static let unpackSerialQueue = DispatchQueue(label: OfflineDirectionsConstants.unpackSerialQueueLabel)
+}
+
+/**
+ Defines additional functionality similar to `Directions` with support for offline routing.
```
1ec5: How does this protocol differ from the OfflineDirectionsProtocol defined by MapboxDirections.swift? I think this could become a source of confusion among developers.

Example/ResizableView.swift
```diff
@@ -0,0 +1,141 @@
...
+    
+}
+
+fileprivate extension CGFloat {
+    
+    var mid: CGFloat {
+        return self / 2
+    }
+    
+    func distance(_ to: CGFloat) -> CGFloat {
```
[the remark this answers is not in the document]
frederoni: The calculations in the pan gesture recognizer were getting unreadable but it has been cleaned up since then so it makes sense to remove this simple distance.

MapboxCoreNavigation/OfflineDirections.swift
frederoni: I agree that it can cause confusion, however, `NavigatorDirectionsProtocol` depends on nav-native so it should be defined here, while `OfflineDirectionsProtocol` is a good fit for MapboxDirections since it's only adding additional directions related API calls. Do you have an idea how to simplify it? The protocol could be moved to MapboxDirections’s `OfflineDirectionsProtocol` but the implementation must reside in MapboxCoreNavigation.

1ec5: I think the confusion would stem from how this protocol is described as “similar to `Directions` with support for offline routing”. So if we need to keep it as a separate protocol, then can we frame it as a kind of NavigationService that has Directions functionality built in?

Example/OfflineViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+        backgroundLayer.frame = view.bounds
+        backgroundLayer.fillColor = #colorLiteral(red: 0.1450980392, green: 0.2588235294, blue: 0.3725490196, alpha: 0.196852993).cgColor
+        view.layer.addSublayer(backgroundLayer)
+        
+        resizableView = ResizableView(frame: CGRect(origin: view.center, size: CGSize(width: 50, height: 50)),
+                                      backgroundLayer: backgroundLayer)
+        
+        view.addSubview(resizableView)
+        
+        navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Download", style: .done, target: self, action: #selector(downloadRegion))
```
1ec5: Looks like we don’t have a Localizable.strings file for the example target yet, only for the framework targets. I didn’t originally expect that we would even translate the example application at all, but we wound up making the storyboard localizable, so we might as well make the rest of it localizable too.

Example/OfflineViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+            Directions.shared.downloadTiles(in: coordinateBounds, version: version, completionHandler: { (url, response, error) in
+                guard let url = url else { return assert(false, "Unable to locate temporary file") }
+                
+                let outputDirectoryURL = Bundle.mapboxCoreNavigation.suggestedTilePathURL(for: version)
+                outputDirectoryURL?.ensureDirectoryExists()
+                
+                NavigationDirections.unpackTilePack(at: url, outputDirectoryURL: outputDirectoryURL!, progressHandler: { (totalBytes, bytesRemaining) in
+
+                    let progress = (Float(bytesRemaining) / Float(totalBytes)) * 100
+                    self?.updateTitle("Unpacking \(Int(progress))%")
```
1ec5: This should ideally use Progress, but at least it should use NumberFormatter and a localized format string.

1ec5:
I’m seeing tons of console spew along these lines, seemingly in a hot loop, while displaying a visual instruction that should include shields in the simulator. The shields never load, but more importantly, I think the level of spew would quickly consume battery and storage.

```
2018-12-04 18:53:40.639392-0800 Example[17124:9955719] Task <C932B850-0162-4891-94CC-D4135E89115F>.<7533> load failed with error Error Domain=NSURLErrorDomain Code=-1002 "unsupported URL" UserInfo={NSLocalizedDescription=unsupported URL, NSErrorFailingURLStringKey=@2x.png, NSErrorFailingURLKey=@2x.png, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalDataTask <C932B850-0162-4891-94CC-D4135E89115F>.<7533>"
), _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <C932B850-0162-4891-94CC-D4135E89115F>.<7533>, NSUnderlyingError=0x600003a123a0 {Error Domain=kCFErrorDomainCFNetwork Code=-1002 "(null)"}} [-1002]
2018-12-04 18:53:40.644171-0800 Example[17124:9956578] Task <1DC0B9B7-FAFC-401C-96B1-11BCCE418505>.<7534> finished with error - code: -1002
```

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -0,0 +1,200 @@
...
+ 
+ Each result produced by the directions object is stored in a `Route` object. Depending on the `RouteOptions` object you provide, each route may include detailed information suitable for turn-by-turn directions, or it may include only high-level information such as the distance, estimated travel time, and name of each leg of the trip. The waypoints that form the request may be conflated with nearby locations, as appropriate; the resulting waypoints are provided to the closure.
+ */
+@objc(MBNavigationDirections)
+public class NavigationDirections: Directions {
+    
+    public typealias UnpackProgressHandler = (_ totalBytes: UInt64, _ remainingBytes: UInt64) -> ()
+    public typealias UnpackCompletionHandler = (_ result: UInt64, _ error: Error?) -> ()
```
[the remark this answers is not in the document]
frederoni: renamed to `numberOfTiles` since it corresponds to the number of unpacked tiles.

1ec5:
The infinite recursion will be fixed by #1888. Locally, with that patch applied, I see generic shields as expected, though I don’t see exit number tabs as I do with server-side routing. The exit number is included in spoken instructions, and the exit number tabs do show up when performing server-side routing on this same branch. Unless an obvious cause materializes very shortly, I think it’s fine to treat that issue as tail work.

PR #1810, opened 2018-11-02
Fix various compiler warnings

frederoni:
Removed unnecessary guard of a non-optional subscript.

cc @1ec5

### DIFF
--- a/Examples/Swift/CustomViewController.swift
+++ b/Examples/Swift/CustomViewController.swift
@@ -170,7 +170,7 @@ class CustomViewController: UIViewController, MGLMapViewDelegate {
         
         // find the upcoming manuever step, and update instructions banner to show preview
         guard stepIndex + 1 < leg.steps.endIndex else { return }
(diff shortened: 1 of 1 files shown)


MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -208,7 +208,7 @@ public class CarPlayNavigationViewController: UIViewController {
         
         // Update the user puck
         mapView?.updatePreferredFrameRate(for: routeProgress)
-        let camera = MGLMapCamera(lookingAtCenter: location.coordinate, fromDistance: 120, pitch: 60, heading: location.course)
+        let camera = MGLMapCamera(lookingAtCenter: location.coordinate, acrossDistance: 120, pitch: 60, heading: location.course)
```
1ec5: I think the existing values are meant to be altitudes, not viewing distances. See #1815.

frederoni: Dropped this commit

PR #1813, opened 2018-11-05
Setup performance tests

frederoni:
Outlined a performance suite consisting of a bench app and tests.
There is only refactoring in this PR at this point.

I have merged the Objective-C example into a single Example app. The Obj-C previously shared the Main.storyboard which seem to have been the cause of the flaky launch process sometimes w/ "Missing Main.storyboard." Removed the RouteTest target (the Bench app will provide the same functionality) Isolated all Fixture.swift-files into a new TestHelper framework for reusability.

- [x] Add initial control route
- [x] Set up a bot on CI

This initial PR doesn't extract any valuable metrics nor does it store it. I think we should keep this already bloated PR as is and add finely tweaked performance tests and configurations in a subsequent PR.

cc @JThramer @akitchen @1ec5 @JThramer


1ec5:
Updated the Transifex configuration in #1828.

PR #1819, opened 2018-11-07
Reduce default minimum frame rate to 20 fps

1ec5:
Reduced the default minimum frame rate (on battery power away from maneuvers) from 30&nbsp;fps to 20&nbsp;fps for improved battery performance. The frame rate while plugged in remains 30&nbsp;fps, and the frame rate during the 10&nbsp;seconds around a maneuver remains at 60&nbsp;fps or 120&nbsp;fps, depending on the device.

/cc @frederoni @JThramer


MapboxNavigation/NavigationMapView.swift
```diff
@@ -24,7 +24,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      
      This property takes effect when the application has limited resources for animation, such as when the device is running on battery power. By default, this property is set to `MGLMapViewPreferredFramesPerSecond.lowPower`.
      */
-    @objc public var minimumFramesPerSecond = MGLMapViewPreferredFramesPerSecond.lowPower
+    @objc public var minimumFramesPerSecond = MGLMapViewPreferredFramesPerSecond(20)
```
frederoni: 👍 
>= 20 should be above the uncanny valley for most pair of eyes.

1ec5:
Just to give a crude sense of the performance improvement, I simulated a half-mile-long (2-minute-long) route along surface streets while on battery power on an iPhone 8 running iOS 12. Over 20&nbsp;seconds of straightaway, away from maneuvers, the Swift example application posted these times:

Frame rate | CPU time | Notes
----|----|----
30fps | 4.41 s | master
**20fps** | **3.04 s** | This PR
5fps | 1.10 s | Before #1749

Overall, for the entire duration of the session (including the flying animation at the beginning), we posted these times:

Frame rate | CPU time | Notes
----|----|----
30fps | 44.48 s | master
**20fps** | **37.46 s** | This PR
5fps | 23.28 s | Before #1749

Issue #1662, opened 2018-09-08
SimulatedLocationManager crash


macdrevx:
**Mapbox Navigation SDK version:**
0.18.1

**Crash log:**

```
# Date: 2018-09-08T03:36:00Z
# OS Version: 11.4.1 (15G77)
# Device: iPad 6
# RAM Free: 12%
# Disk Free: 79.4%

#0. Crashed: com.apple.main-thread
0  libobjc.A.dylib                0x18425891c objc_msgSend + 28
1  MapboxCoreNavigation           0x103819f94 _T020MapboxCoreNavigation24SimulatedLocationManagerC4tick33_BAD5A1656A34E7A88BC59C7E80655DACLLyyF + 1376
2  MapboxCoreNavigation           0x10381a3b0 _T020MapboxCoreNavigation24SimulatedLocationManagerC013startUpdatingE0yyFToTm + 28
3  Foundation                     0x185ac04ac __NSFireDelayedPerform + 408
4  CoreFoundation                 0x185033aa8 __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__ + 28
5  CoreFoundation                 0x18503376c __CFRunLoopDoTimer + 864
6  CoreFoundation                 0x185033010 __CFRunLoopDoTimers + 248
7  CoreFoundation                 0x185030b60 __CFRunLoopRun + 2168
8  CoreFoundation                 0x184f50da8 CFRunLoopRunSpecific + 552
9  GraphicsServices               0x186f36020 GSEventRunModal + 100
10 UIKit                          0x18ef70758 UIApplicationMain + 236
11 OnDemandDriver                 0x102caaecc main (main.swift:25)
12 libdyld.dylib                  0x1849e1fc0 start + 4
```

**Observations**

Our app isn't using `SimulatedLocationManager` directly, but we do use `NavigationViewController`, which has a `RouteController`, which has a `TunnelIntersectionManager`, which has a `SimulatedLocationManager`. As far as I can tell from an inspection of the source code, that's the only place where one is being used.

akitchen:
I think there could be a subtle and difficult-to-reproduce bug in the `tick()` method, which is using `perform(#selector...` after canceling previously scheduled performs and doing some work to set up the next scheduled `perform(#selector...`.

[...]

PR #1820, opened 2018-11-07
Location Manager Simulation

### LINKED ISSUE #1804: Need to set CLLocationManager.delegate to nil in delegate's deinit methods
We're on 0.18.1, but it looks like this is still the case on master.

We've seen a few crashes that look like they could result from CLLocationManager attempting to message a delegate that's been deallocated. By inspection of the code, it looks like the SDK needs to be updated to set `CLLocationManager.delegate` to `nil` when the delegate gets deallocated. Unlike other delegates, this one is `unowned` rather than `weak` ([link](https://developer.apple.com/documentation/corelocation/cllocationmanager/1423792-delegate)).

JThramer:
Fixes #1804 and #1662.

* Fixes issue where strongly-held `unowned` delegate ref in `CLLocationManager` can cause crash.
* Replaces `performSelector:afterDelay:` based timer mechanism with a dispatch timer source, which fixes crash.

To-do: 

- [x] Settle on scheduling parameters
- [x] Unit Tests
- [x] Write changelog

/cc @mapbox/navigation-ios


MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -88,14 +91,18 @@ open class SimulatedLocationManager: NavigationLocationManager {
...
 
-    private func initializeSimulatedLocationManager(for route: Route, currentDistance: CLLocationDistance, currentSpeed: CLLocationSpeed) {
+    private func commonInit(for route: Route, currentDistance: CLLocationDistance, currentSpeed: CLLocationSpeed) {
         
         self.currentSpeed = currentSpeed
         self.currentDistance = currentDistance
         self.route = route
+
+        self.timer = DispatchTimer(countdown: .milliseconds(0), repeating: updateInterval, accuracy: accuracy, executingOn: .main) { [weak self] in
+            DispatchQueue.main.async { self?.tick() }
```
JThramer: Is this async call necessary if we're `executingOn: .main`?

akitchen: if you want to defer to a later run loop iteration ...

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -268,6 +268,8 @@ public class MapboxNavigationService: NSObject, NavigationService, DefaultInterf
     deinit {
         suspendNotifications()
         endNavigation()
+        nativeLocationSource.delegate = nil
+        simulatedLocationSource?.delegate = nil
```
akitchen: This is an important addition, but I can't help but wonder if there are other places where this needs to be done as well?

akitchen:
> Fixes issue where strongly-held unowned delegate ref in CLLocationManager can cause crash.

This is mildly pedantic, but the issue is that the `assign` reference isn't strongly held, it's just a dangling pointer.

MapboxCoreNavigation/NavigationService.swift
JThramer: Apparently `CLLocationManager` is an exception because unlike convention, they store their delegate reference as `unowned` instead of `weak`. Apparently you usually you don't have to worry about this.

I also searched the code -- we don't set location manager delegates outside of the navigation service.

MapboxCoreNavigation/DispatchTimer.swift
```diff
@@ -14,16 +14,18 @@ class CountdownTimer {
         }
     }
     private var deadline: DispatchTime { return .now() + countdownInterval }
+    let repititionInterval: DispatchTimeInterval
```
akitchen: oops, misspelling

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master
 
+* Fixed issue where a released reference that was kept in `CLLocationManager` as `unowned` could cause crash. ([#1820](https://github.com/mapbox/mapbox-navigation-ios/pull/1820))
```
1ec5: The dangling reference was in SDK code; mentioning it in the changelog entry makes it sound to the developer like the crash would occur when application code had a dangling reference.

```suggestion
* Fixed a crash during turn-by-turn navigation. ([#1820](https://github.com/mapbox/mapbox-navigation-ios/pull/1820))
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
...
+* Replacing `performSelector:afterDelay:` mechanism that was [causing a crash](https://github.com/mapbox/mapbox-navigation-ios/issues/1662) with a DispatchTimerSource-based mechanism. ([#1820](https://github.com/mapbox/mapbox-navigation-ios/pull/1820))
```
1ec5: Similarly, this is a purely internal change. If we expect any reliability or stability improvements to result from this change, we can mention those improvements here; otherwise, the change isn’t actionable to the developer.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master
```
1ec5: You’ll probably want to rebase onto master, because the changelog entries need to go in a new master section, not v0.24.0.

CHANGELOG.md
akitchen:
```suggestion
* Replaces the `SimulatedLocationManager` timer mechanism with a `DispatchTimerSource`-based mechanism ([#1820](https://github.com/mapbox/mapbox-navigation-ios/pull/1820)), which fixes a crash that was caused by retaining the `SimulatedLocationManager` while it still had a dangling reference to its delegate (https://github.com/mapbox/mapbox-navigation-ios/issues/1662)
```

1ec5: I think the line above (as amended) adequately describes the developer impact of this PR. This PR’s description is the place to provide details about the crash’s root cause and the fix’s implementation.

PR #1825, opened 2018-11-08
Profile control routes

frederoni:
Profile CPU usage for control routes

So far it's just a short temporary control route to speed up the iterations.

cc @akitchen @1ec5


MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -351,6 +363,27 @@
...
+			proxyType = 1;
+			remoteGlobalIDString = 351BEBD61E5BCC28006FE110;
+			remoteInfo = MapboxNavigation;
+		};
+		352ECD2F2194892E00F0B334 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = C5ADFBC01DDCC7840011824B /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 35CDA85D2190F2A30072B675;
+			remoteInfo = TestHelper;
```
1ec5: Bench now depends on TestHelper, which has a couple `@testable import`s. That prevents Bench from being build in the Release configuration, which is important for accurately measuring most aspects of performance.

akitchen:
This PR has already proved to be super useful. Is there any remaining reason not to merge it? Are there other changes we are thinking about making?

One thing that comes to mind is whether the bench app should live in another repo more suited to tooling...

frederoni:
> Is there any remaining reason not to merge it? Are there other changes we are thinking about making?

Not really. We can iterate from here. Let's keep the short route running on every push and eventually enable all control routes for tags/releases. I'll remove the trace-parser binary for now and potentially replace it with a submodule when we connect the metrics.

frederoni:
Moved to another repo

PR #1833, opened 2018-11-12
Fix rerouting at arrival logic

### LINKED ISSUE #1832: Rerouting occurs after user has arrived
v0.22.3

In RouteController.locationManager(manager:didUpdateLocations), rerouting is based on the value of RouteController.userIsOnRoute :
userIsOnRoute should returns true if routeProgress.currentLegProgress.userHasArrivedAtWaypoint **and** shouldPreventReroutesWhenArrivingAt

floydaddict:
fixes #1832

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -238,7 +238,7 @@ open class RouteController: NSObject, Router {
     @objc public func userIsOnRoute(_ location: CLLocation) -> Bool {
         
         // If the user has arrived, do not continue monitor reroutes, step progress, etc
-        guard !routeProgress.currentLegProgress.userHasArrivedAtWaypoint || (delegate?.router?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ?? DefaultBehavior.shouldPreventReroutesWhenArrivingAtWaypoint) else {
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -238,7 +238,7 @@ open class RouteController: NSObject, Router {
     @objc public func userIsOnRoute(_ location: CLLocation) -> Bool {
         
         // If the user has arrived, do not continue monitor reroutes, step progress, etc
-        guard !routeProgress.currentLegProgress.userHasArrivedAtWaypoint || (delegate?.router?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ?? DefaultBehavior.shouldPreventReroutesWhenArrivingAtWaypoint) else {
+        guard !routeProgress.currentLegProgress.userHasArrivedAtWaypoint || !(delegate?.router?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ?? DefaultBehavior.shouldPreventReroutesWhenArrivingAtWaypoint) else {
```
1ec5: There are a few layers of double negatives in the existing code, made worse by the `guard` statement. If we reexpress the original code (using De Morgan’s laws) as:

```swift
if routeProgress.currentLegProgress.userHasArrivedAtWaypoint
    && !(delegate?.router?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination)
         ?? DefaultBehavior.shouldPreventReroutesWhenArrivingAtWaypoint) {
```

then it’s a little clearer that the method says the user is on the route (preventing rerouting) if `userHasArrivedAtWaypoint` is set and the delegate declines to prevent a reroute (that is, allows rerouting to occur). You’re right; that’s incorrect.

This change says the user is on the route when `userHasArrivedAtWaypoint` is set and the delegate says it wants to prevent rerouting.

PR #1834, opened 2018-11-12
Customize Directions Client in CarPlayManager.

### LINKED ISSUE #1824: Crash calculating a route in CarPlay without access token in Info.plist
CarPlayManager currently hard-codes the shared Directions object, which is only valid if a Mapbox access token is specified in the Info.plist file. An application that needs to specify an access token at runtime will crash because no access token is available to calculate the route:

https://github.com/mapbox/mapbox-navigation-ios/blob/13ef9fa8959a2a478299dc0e00bb3bae729a97d6/MapboxNavigation/CarPlayManager.swift#L420

Elsewhere, CarPlayManager allows the delegate to override the NavigationService, which indirectly allows the developer to customize the access token:

https://github.com/mapbox/mapbox-navigation-ios/blob/ae19419a0b30b97eb35f670cca973462625a6257/MapboxNavigation/CarPlayManager.swift#L502

But that only happens once there’s a route and the user has opted to start navigating along that route. CarPlayManager needs to ask the delegate for the Directions object beforehand, when all it has are the waypoints. At that point, it can’t ask for the NavigationService, because NavigationService requires a Route.

In any case, I’m not sure we’d be able to provide the delegate with enough context to dynamically switch between access tokens on a case-by-case basis. So I think we might as well have CarPlayManager’s initializer optionally accept a Directions object, defaulting to the shared object.

/cc @mapbox/navigation-ios

JThramer:
Fixes #1824. Allowing user to inject custom Directions client @ `CarPlayManager` init time.

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master
 
-* Added the ability to specify a custom directions client to be used by the CarPlayManager. ([#1834](https://github.com/mapbox/mapbox-navigation-ios/pull/1834/))
+* Renamed `CarPlayManager(_:)` to `CarPlayManager(directions:eventsManager:)`, allowing you to pass in a custom `Directions` object to use when calculating routes. ([#1834](https://github.com/mapbox/mapbox-navigation-ios/pull/1834/))
 * Fixed a crash during turn-by-turn navigation. ([#1820](https://github.com/mapbox/mapbox-navigation-ios/pull/1820))
 * Fixed a crash that could happen while simulating a route. ([#1820](https://github.com/mapbox/mapbox-navigation-ios/pull/1820))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master
 
+* Added the ability to specify a custom directions client to be used by the CarPlayManager. ([#1834](https://github.com/mapbox/mapbox-navigation-ios/pull/1834/))
 * Fixed a crash during turn-by-turn navigation. ([#1820](https://github.com/mapbox/mapbox-navigation-ios/pull/1820))
 * Fixed a crash that could happen while simulating a route. ([#1820](https://github.com/mapbox/mapbox-navigation-ios/pull/1820))
 
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -193,11 +193,13 @@ public class CarPlayManager: NSObject {
      */
     @objc public static var isConnected = false
 
-    public var eventsManager: NavigationEventsManager!
+    public let eventsManager: NavigationEventsManager
+    public let directions: Directions
 
-    public init(_ eventsManager: NavigationEventsManager = NavigationEventsManager(dataSource: nil)) {
+    public init(directions: Directions? = nil, eventsManager: NavigationEventsManager? = nil) {
+        self.directions = directions ?? .shared
+        self.eventsManager = eventsManager ?? NavigationEventsManager(dataSource: nil)
         super.init()
-        self.eventsManager = eventsManager
     }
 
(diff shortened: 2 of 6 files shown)


MapboxNavigation/CarPlayManager.swift
```diff
@@ -402,72 +404,82 @@ extension CarPlayManager: CPListTemplateDelegate {
...
+    }
+    
+    internal func fetchNewRoute(options: RouteOptions, completionHandler: @escaping () -> Void) {
+        directions.calculate(options) { [weak self] (waypoints, routes, error) in
+            self?.handleDirectionsResponse(options: options, waypoints: waypoints, routes: routes, error: error, completionHandler: completionHandler)
+        }
+    }
+    
+    
+    internal func handleDirectionsResponse(options routeOptions: RouteOptions, waypoints: [Waypoint]?, routes: [Route]?, error: NSError?, completionHandler: @escaping () -> Void) {
```
1ec5: Let’s try to avoid “handle” in method names:

```suggestion
    internal func didCalculate(_ routes: [Route]?, for routeOptions: RouteOptions, between waypoints: [Waypoint]?, error: NSError?, completionHandler: @escaping () -> Void) {
```

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -170,8 +169,35 @@ class CarPlayManagerTests: XCTestCase {
...
+        let waypoint2 = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 37.7727, longitude: -122.433378))
+        let options = RouteOptions(waypoints: [waypoint1, waypoint2])
+        subject.fetchNewRoute(options: options, completionHandler: {})
+        wait(for: [expectation], timeout: 1.0)
+        
+        XCTAssert(subject.directions == spy, "Directions client is not overridden properly.")
+    }
 }
 
+private class DirectionsInvocationSpy: Directions {
```
1ec5: This spy class is very specific to a single test. Consider moving this class definition inside `testDirectionsOverride()` so the relationship is clearer.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -193,11 +193,13 @@ public class CarPlayManager: NSObject {
      */
     @objc public static var isConnected = false
 
-    public var eventsManager: NavigationEventsManager!
+    public let eventsManager: NavigationEventsManager
+    public let directions: Directions
 
-    public init(_ eventsManager: NavigationEventsManager = NavigationEventsManager(dataSource: nil)) {
+    public init(directions: Directions? = nil, eventsManager: NavigationEventsManager? = nil) {
```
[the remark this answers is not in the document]
1ec5: Unfortunately, the default argument value doesn’t cause the initializer to bridge to Objective-C as two methods, one with and the other without this parameter, as I had thought. So we should remove default values in favor of internal optional chaining across the board or explicitly create multiple related methods the traditional Objective-C way. Either way, the self-documenting benefit mentioned above goes away, at least from the point of view of Objective-C code.

So this part of the PR is fine as is, except that the initializer needs to be documented even more than it did originally (also for the benefit of Swift developers).

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master
 
+* Added the ability to specify a custom directions client to be used by the CarPlayManager. ([#1834](https://github.com/mapbox/mapbox-navigation-ios/pull/1834/))
```
1ec5: This will end up being overwritten by #1836 anyways, but what the hey:

```suggestion
* Renamed `CarPlayManager(_:)` to `CarPlayManager(directions:eventsManager:)`, allowing you to pass in a custom `Directions` object to use when calculating routes. ([#1834](https://github.com/mapbox/mapbox-navigation-ios/pull/1834/))
```

PR #1836, opened 2018-11-14
Allow style customization in CarPlay

### LINKED ISSUE #1823: CarPlay map style should be customizable
CarPlayMapViewController and CarPlayNavigationViewController both hard-code the built-in DayStyle and NightStyle. A developer can specify a custom style for NavigationViewController, so they may want to customize the CarPlay map style, for example to match the style being used on the phone.

https://github.com/mapbox/mapbox-navigation-ios/blob/ae19419a0b30b97eb35f670cca973462625a6257/MapboxNavigation/CarPlayMapViewController.swift#L63 https://github.com/mapbox/mapbox-navigation-ios/blob/ae19419a0b30b97eb35f670cca973462625a6257/MapboxNavigation/CarPlayNavigationViewController.swift#L101

Just as `NavigationViewController(for:styles:navigationService:voiceController:)` takes the initial styles as an argument, CarPlayMapViewController and CarPlayNavigationViewController probably should as well, and that argument should be plumbed through to `CarPlayManager(_:)`. This is a reasonable place to expose the style option, now that CarPlayManager is no longer a singleton (#1792). Still, it may be worth exploring a mechanism to change the style after the fact, since the application delegate typically creates CarPlayManager long before setting up any UI options that might affect the style (such as a street/satellite toggle).

/cc @mapbox/navigation-ios

1ec5:
Applications can now customize the style of the map in CarPlay (nominally the rest of the CarPlay interface, but really just the map style).

Renamed `CarPlayManager(_:)` to `CarPlayManager(styles:eventsManager:)` and `CarPlayNavigationViewController(with:mapTemplate:interfaceController:manager:styles:)` to `CarPlayNavigationViewController(navigationService:mapTemplate:interfaceController:manager:styles:)`. These initializers now accept Style objects as arguments, which are applied throughout the CarPlay interface, similar to NavigationViewController. DayStyle and NightStyle are still used by default but no longer hard-coded everywhere. The application can set `CarPlayManager.styles` at any time to change the map style, for example if it sets the style of NavigationViewController or a standalone NavigationMapView on the phone and needs to synchronize the change on CarPlay.

Renamed `StyleManager(_:)` to `StyleManager()`. The previous initializer looked like it gave ownership of the delegate to StyleManager, whereas it was just a weak reference. Client code is now expected to set the `delegate` property explicitly.

Exposed a few more CarPlayManager members to Objective-C code. Fixed some build warnings.

Fixes #1823.

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,15 @@
 
 ## master
 
-* Renamed `CarPlayManager(_:)` to `CarPlayManager(directions:eventsManager:)`, allowing you to pass in a custom `Directions` object to use when calculating routes. ([#1834](https://github.com/mapbox/mapbox-navigation-ios/pull/1834/))
+### CarPlay
+
+* Renamed `CarPlayManager(_:)` to `CarPlayManager(styles:directions:eventsManager:)` and `CarPlayNavigationViewController(with:mapTemplate:interfaceController:manager:styles:)` to `CarPlayNavigationViewController(navigationService:mapTemplate:interfaceController:manager:styles:)`. These initializers now accept an array of `Style` objects to apply throughout the CarPlay interface, similar to `NavigationViewController`. You can also change the styles at any time by setting the `CarPlayManager.styles` property. ([#1836](https://github.com/mapbox/mapbox-navigation-ios/pull/1836))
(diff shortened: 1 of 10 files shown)


frederoni:
Noticed that this build didn't trigger codecov. Would you mind trying a rebase before merging?

PR #1841, opened 2018-11-14
Adding support for multi-leg routes in CarPlayManager.

### LINKED ISSUE #1822: CarPlayManager should take route options as input
`CarPlayManager.calculateRouteAndStart(from:to:completionHandler:)` only accepts origin and destination waypoints, but some use cases require multiple waypoints or other route options, such as whether the destination must be on the passenger side upon arrival. We should replace this method with `CarPlayManager.preview(_:completionHandler:)` that takes a RouteOptions object as an argument.

/cc @mapbox/navigation-ios

JThramer:
Like it says on the box. Fixes #1822.

/cc @mapbox/navigation-ios


MapboxNavigation/CarPlayManager.swift
```diff
@@ -278,43 +278,54 @@ extension CarPlayManager: CPListTemplateDelegate {
...
-            calculateRouteAndStart(to: waypoint, completionHandler: completionHandler)
+            preview(waypoints: [waypoint], completionHandler: completionHandler)
             return
         }
         
         completionHandler()
     }
     
-    public func calculateRouteAndStart(from fromWaypoint: Waypoint? = nil, to toWaypoint: Waypoint, completionHandler: @escaping () -> Void) {
+    public func preview(waypoints: [Waypoint], completionHandler: @escaping CompletionHandler) {
```
1ec5:
```suggestion
    @objc(previewRoutesBetweenWaypoints:completionHandler:)
    public func previewRoutes(between waypoints: [Waypoint], completionHandler: @escaping CompletionHandler) {
```

MapboxNavigation/CarPlayManager.swift
```diff
@@ -278,43 +278,54 @@ extension CarPlayManager: CPListTemplateDelegate {
...
-                completionHandler()
-                return
+        if waypoints.count == 1 {
+            guard let rootViewController = self.carWindow?.rootViewController as? CarPlayMapViewController,
+                let userLocation = rootViewController.mapView.userLocation,
+                let location = userLocation.location else {
+                    completionHandler()
+                    return
+            }
+            let name = NSLocalizedString("CARPLAY_CURRENT_LOCATION", bundle: .mapboxNavigation, value: "Current Location", comment: "Name of the waypoint associated with the current location")
```
1ec5: I think this method should either always add this current location waypoint or never add it. As it is, if the caller specifies two waypoints, do we know whether the caller intended the first waypoint to be the origin or an intermediate destination?

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changes to the Mapbox Navigation SDK for iOS
 
 ## master
-
+* Adding support for multi-leg routes in `CarPlayManager`. ([#1841](https://github.com/mapbox/mapbox-navigation-ios/pull/1841))
```
1ec5: Since there is a backwards-incompatible public API change, the change needs to be indicated explicitly:

```suggestion
* Renamed `CarPlayManager.calculateRouteAndStart(from:to:completionHandler:)` to `CarPlayManager.previewRoutes(between:completionHandler:)` and added a `CarPlayManager.previewRoutes(for:completionHandler:)` method that accepts an arbitrary `NavigationRouteOptions` object. ([#1841](https://github.com/mapbox/mapbox-navigation-ios/pull/1841))
```

Example/AppDelegate+CarPlay.swift
```diff
@@ -104,7 +104,7 @@ extension AppDelegate: CPListTemplateDelegate {
         // Selected a favorite
         if let userInfo = item.userInfo as? [String: Any],
             let waypoint = userInfo[CarPlayManager.CarPlayWaypointKey] as? Waypoint {
-            carPlayManager.calculateRouteAndStart(to: waypoint, completionHandler: completionHandler)
+            carPlayManager.preview(waypoints: [waypoint], completionHandler: completionHandler)
```
1ec5: This method was renamed to `previewRoutes(between:completionHandler:)`. Fixed in cc255c4d8b66bfab1150e321e696d54cf4debe76.

PR #1845, opened 2018-11-20
Fix CarPlay navigation map vanishing point

### LINKED ISSUE #1681: User puck rotation should account for horizontal content insets
When tracking the user’s course, UserCourseView’s arrow points in a direction determined by a CLHeading object, assuming that UserCourseView is located above the visual center of the map view, and that map’s vanishing point is located directly above UserCourseView. When the map view has a left or right content inset, the route line and UserCourseView are shifted to the side, while the map’s vanishing point remains horizontally centered due to mapbox/mapbox-gl-native#12728.

For example, in CarPlay, there is a wide content inset on the driver’s side to accommodate maneuver panels. When both the route and the user go due north, the route line is slanted towards the driver’s side while UserCourseView’s arrow nonetheless points straight up.

The 3D transform for UserCourseView’s arrow needs to account for horizontal content insets. The correction might end up depending on the pitch. As with mapbox/mapbox-gl-native#5245, it may be necessary to fudge the transform matrix’s `m34`.

1ec5:
Fixed an issue where the user puck pointed away from the route line during turn-by-turn navigation in CarPlay. The map’s vanishing point now accounts for safe area insets, including the side maneuver view. As described in https://github.com/mapbox/mapbox-navigation-ios/issues/1681#issuecomment-440119219, we now place the map view’s horizontal center where we want the vanishing point to be. This approach is preferable to rotating the map because it preserves the angles at which cross streets meet the route line.

Before:

vanishing

After:

full chrome

Before this PR can land, I need to fix a regression where standalone NavigationMapViews on the iPhone fail to respect the safe area insets.

Fixes #1681.

### DIFF
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -448,8 +448,8 @@ extension CarPlayNavigationViewController: NavigationMapViewDelegate {
         
         // Avoid letting the puck go partially off-screen, and add a comfortable padding beyond that.
         let courseViewBounds = mapView.userCourseView?.bounds ?? .zero
-        contentFrame = contentFrame.insetBy(dx: min(50 + courseViewBounds.width / 2.0, contentFrame.width / 2.0),
-                                            dy: min(50 + courseViewBounds.height / 2.0, contentFrame.height / 2.0))
+        contentFrame = contentFrame.insetBy(dx: min(NavigationMapView.courseViewMinimumInsets.left + courseViewBounds.width / 2.0, contentFrame.width / 2.0),
+                                            dy: min(NavigationMapView.courseViewMinimumInsets.top + courseViewBounds.height / 2.0, contentFrame.height / 2.0))
         
         // Get the bottom-center of the remaining frame.
         assert(!contentFrame.isInfinite)
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -24,6 +24,7 @@ public class CarPlayNavigationViewController: UIViewController {
     var navService: NavigationService
     var mapView: NavigationMapView?
     let shieldHeight: CGFloat = 16
+    var mapViewSafeAreaBalancingConstraint: NSLayoutConstraint?
     
     var carSession: CPNavigationSession!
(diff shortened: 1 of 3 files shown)


MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -148,6 +162,12 @@ public class CarPlayNavigationViewController: UIViewController {
         }
         
         previousSafeAreaInsets = view.safeAreaInsets
+        
+        // Adjust the map’s vanishing point to counterbalance the side maneuver panels by extending the view off beyond the other side of the screen.
+        if let mapView = mapView {
+            mapViewRightSafeAreaBalancingConstraint?.constant = -mapView.safeArea.right
+            mapViewLeftSafeAreaBalancingConstraint?.constant = mapView.safeArea.left
```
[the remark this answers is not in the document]
1ec5: → #1849

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -420,6 +440,23 @@ public class CarPlayNavigationViewController: UIViewController {
...
+@available(iOS 12.0, *)
+extension CarPlayNavigationViewController: NavigationMapViewDelegate {
+    public func navigationMapViewUserAnchorPoint(_ mapView: NavigationMapView) -> CGPoint {
+        // Inset by the content inset to avoid application-defined content.
+        var contentFrame = UIEdgeInsetsInsetRect(mapView.bounds, mapView.contentInset)
+        
+        // Avoid letting the puck go partially off-screen, and add a comfortable padding beyond that.
+        let courseViewBounds = mapView.userCourseView?.bounds ?? .zero
+        contentFrame = contentFrame.insetBy(dx: min(50 + courseViewBounds.width / 2.0, contentFrame.width / 2.0),
+                                            dy: min(50 + courseViewBounds.height / 2.0, contentFrame.height / 2.0))
```
vincethecoder: nit: probably a variable for these magic numbers will aid self-documentation here.

1ec5: Indeed, these couple lines were copied from `NavigationMapView.userAnchorPoint` to avoid impacting non-CarPlay maps. While the logic probably doesn’t warrant a separate function, I agree that the additional padding should be centralized in a constant.

Issue #1839, opened 2018-11-14
CarPlay search support for 3rd party applications.


vincethecoder:
**Mapbox Navigation SDK version: 0.24.0**

In lieu of ensuring that `CarPlay` can support 3rd party apps seamlessley. A few discoveries were made that merits a few modifications of the `CarPlayManager`.

* The static framework `MapboxGeocoder` isn't discoverable by the Navigation SDK conditional check `canImport(MapboxGeocoder)` when an app integrates the Navigation SDK into its project workspace using cocoapod `pod install` command. (Since the developer has the liberty to use their preferred `Geocoder`, this framework is intended to support search -- which is classified _optional_.)

* 3rd party apps are limited to use the default `geocoder` supplied in the SDK. Developers cannot provide their own if they preferred.

* `internal` functions such the `resetPanButtons` are not reusable by 3rd party apps.

**Suggested Solutions**:

* The `CarPlayManager` should be an `open` class so that it can be subclassable by other developers who would like to inject their very own `geocoder`.

* The fields and methods highlighted above should be `public`. 

* Lastly, investigate and implement an elegant solution to address the challenge with the integration of the `MapboxGeocoder` for apps that use cocoapods.

/cc @akitchen @frederoni @andrewychen @mapbox/navigation-ios

1ec5:
> The static framework `MapboxGeocoder` isn't discoverable by the Navigation SDK conditional check `canImport(MapboxGeocoder)` when an app integrates the Navigation SDK into their project workspace using cocoapod `pod install` command.

Issues around `canImport()` and CocoaPods have been reported in the past, such as in #1744. As far as I can tell, the reason `canImport()` fails with CocoaPods is that the `FRAMEWORK_SEARCH_PATHS` build setting precludes finding MapboxGeocoder.framework when building MapboxNavigation.framework:

* Standard application target template in Xcode: ```$(inherited) @executable_path/Frameworks```
* Application target set up for Carthage: ```$(inherited) $(PROJECT_DIR)/Carthage/Build/iOS```
[...]

Issue #1744, opened 2018-09-26
Type 'CarPlayManager' has no member 'searchTemplate'


zijiazhai:
Hi, 
After updating to '~> 0.21', I copy, paste the AppDelegate+CarPlay.swift file to my project. There are two errors: 
![image](https://user-images.githubusercontent.com/10751092/46102751-b1f8e700-c194-11e8-9cb8-9c6beca94f6d.png)

After removed those two errors, I can run the app to Car-play simulator successfully.

akitchen:
Hi there -

It sounds like perhaps you are already using MapboxGeocoder in your project, but not using the CarPlayManager+Search.swift extension (and perhaps not using the example search template built on top of the MapboxGeocoder)?

In this case it is fine to remove those lines. Otherwise, you would also want to include CarPlayManager+Search.swift (or a similar implementation) in your project.

Hope that helps!

zijiazhai:
After I uninstall the MapboxGeocoder in my project, two errors goes away.  But, I still can not see the search template button on the left top corner, still can not input address.  How to fix this? Thanks again.

akitchen:
if you want to try the example search template, MapboxGeocoder.swift should be included, as should the code in CarPlayManager+Search.swift

How are you integrating MapboxNavigation into your project? It sounds like the CarPlayManager+Search.swift file is missing

zijiazhai:
Yes, I want to try the example search template. I reinstalled the MapboxGeocoder.swift v0.10 and MapboxNavigation 0.21 by using CocoaPods, I can see both MapboxGeocoder.swift and CarPlayManager+Search.swift in Pods(so CarPlayManager+Search.swift file is not missing? ):
![image](https://user-images.githubusercontent.com/10751092/46154462-d7d8c700-c23a-11e8-8efb-e416c737c617.png)
But, there are two errors shows Type 'CarPlayManager' has no member 'searchTemplate' and carPlayManager. So, why I can not use class of 'extension CarPlayManager: CPSearchTemplateDelegate' with  MapboxGeocoder.swift v0.10 installed.
Sorry for my poor English.

zijiazhai:
guys, any update for my problem? The code inside of #if canImport(MapboxGeocoder) #endif never been executed. Any suggestions I can try to fix this problem?

zijiazhai:
@akitchen @1ec5 @frederoni 
I am so confused why this happen, but can u guys help me fix this problem?

[...]

PR #1846, opened 2018-11-20
Separate search component from the main carplay manager. 

vincethecoder:
Modifications to make the main CarPlay manager and search feature are managed by two different classes.

Addresses #1839 

/cc @akitchen @mapbox/navigation-ios


MapboxNavigation/CarPlaySearchManager.swift
```diff
@@ -0,0 +1,46 @@
...
+    func resetPanButtons(_ mapTemplate: CPMapTemplate)
+    
+    func pushTemplate(_ template: CPTemplate, animated: Bool)
+    
+    func popTemplate(animated: Bool)
+}
+
+@available(iOS 12.0, *)
+@objc(MBCarPlaySearchManager)
+public class CarPlaySearchManager: NSObject {
```
1ec5: Can you speak to why this functionality needs to be in a separate manager class rather than integrated into CarPlayManager and CarPlayManagerDelegate?

akitchen: This could be an incremental step toward moving all search-related functionality into a different framework (like MapboxGeocoder for example). 

Personally I'd like CarPlayManager to not need to know about search at all if we can avoid it. The CarPlayManager should be about lifecycle events and basic plumbing, at least that's how we were originally thinking about it.

vincethecoder: @1ec5 Yes, we hope this will address the issue raised here, https://github.com/mapbox/mapbox-navigation-ios/issues/1744, with this solution. The separation of concerns for the CarPlayManager and the CarPlaySearchManager has been highlighted [here](https://github.com/mapbox/mapbox-navigation-ios/issues/1839).

MapboxNavigation/CarPlaySearchManager+Search.swift
```diff
@@ -149,12 +132,27 @@ extension CarPlayManager: CPSearchTemplateDelegate {
...
+@available(iOS 12.0, *)
+extension CarPlaySearchManager: CPListTemplateDelegate {
+    public func listTemplate(_ listTemplate: CPListTemplate, didSelect item: CPListItem, completionHandler: @escaping () -> Void) {
+        // Selected a search item from the extended list?
+        if let userInfo = item.userInfo as? [String: Any],
+            let placemark = userInfo[CarPlaySearchManager.CarPlayGeocodedPlacemarkKey] as? GeocodedPlacemark,
+            let location = placemark.location {
+            let destinationWaypoint = Waypoint(location: location)
+            delegate?.popTemplate(animated: false)
+            delegate?.previewRoutes(between: [destinationWaypoint], completionHandler: completionHandler)
```
frederoni: these delegates appear to be nil, causing the flow to stall when selecting a `CPListItem`.

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -84,7 +90,8 @@ class CarPlayManagerTests: XCTestCase {
         let searchButton = template.leadingNavigationBarButtons.first!
         searchButton.handler!(searchButton)
 
-        XCTAssert(fakeInterfaceController.topTemplate?.isKind(of: CPSearchTemplate.self) ?? false, "Expecting a search template to be on top")
+        // TODO: Setup leading buttons to be pushed on top of the interface controller
+        // XCTAssert(fakeInterfaceController.topTemplate?.isKind(of: CPSearchTemplate.self) ?? false, "Expecting a search template to be on top")
```
frederoni: this is probably failing because of the same reason mentioned in https://github.com/mapbox/mapbox-navigation-ios/pull/1846/files#r235715392

vincethecoder: This is failing because we need a delegate dedicated to ~`CarPlaySearchManager`~ `CarPlaySearchController`. There is an upcoming commit for this hence the placeholder for this effort.

MapboxNavigation/CarPlaySearchManager+Search.swift
vincethecoder: The wrong function was called here. I corrected it in this new [commit](https://github.com/mapbox/mapbox-navigation-ios/pull/1846/commits/4a5d7d5118972d069e22efd4896b4f70476b3462).

MapboxNavigation/CarPlaySearchController.swift
```diff
@@ -0,0 +1,43 @@
...
+     The most recent search results.
+     */
+    var recentSearchItems: [CPListItem]?
+    
+    /**
+     The most recent search text.
+     */
+    var recentSearchText: String?
+    
+    public weak var delegate: CarPlaySearchControllerDelegate?
```
frederoni: This property and the class should be documented.

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -84,6 +97,7 @@ class CarPlayManagerTests: XCTestCase {
         let searchButton = template.leadingNavigationBarButtons.first!
         searchButton.handler!(searchButton)
 
+        // TODO: Setup leading buttons to be pushed on top of the interface controller
```
frederoni: Is this TODO still relevant?

Example/AppDelegate+CarPlay.swift
```diff
@@ -59,6 +60,23 @@ extension AppDelegate: CarPlayManagerDelegate {
...
+            return nil
+        }
+        
+        switch activity {
+        case .browsing:
+            let searchTemplate = CPSearchTemplate()
+            searchTemplate.delegate = carPlaySearchController
+            let searchButton = carPlaySearchController.searchTemplateButton(searchTemplate: searchTemplate, interfaceController: interfaceController, traitCollection: traitCollection)
+            return [searchButton]
+        default:
```
frederoni: can we make this switch case exhaustive w/o using `default` to prevent regressions if we would add another activity?

vincethecoder: sure!

vincethecoder:
@JThramer Currently, we create a custom search controller `CarPlaySearchController` initialized here: https://github.com/mapbox/mapbox-navigation-ios/pull/1846/files#diff-64b9754c1604a73458ce613b63f76a0bR13

JThramer:
Posterity: Spoke with Kobe about this, digressing.

PR #1848, opened 2018-11-20
Pass token from directions to speech synthesizer

frederoni:
This change allows for setting the access token for the `SpeechSynthesizer` programmatically.

- [ ] Blocked by a release of MapboxDirections including https://github.com/mapbox/MapboxDirections.swift/pull/313
- [ ] Blocked by a release of mapbox-speech-swift including https://github.com/mapbox/mapbox-speech-swift/pull/18

cc @1ec5 @akitchen @JThramer


MapboxNavigation/NavigationViewController.swift
```diff
@@ -371,7 +372,7 @@ open class NavigationViewController: UIViewController {
         self.navigationService = navigationService ?? MapboxNavigationService(route: route)
         self.navigationService.usesDefaultUserInterface = true
         self.navigationService.delegate = self
-        self.voiceController = voiceController ?? MapboxVoiceController()
+        self.voiceController = voiceController ?? MapboxVoiceController(speechClient: SpeechSynthesizer(accessToken: navigationService?.directions.accessToken))
```
akitchen: This is OK for now but ultimately I'd like to see the accessToken offered by the NavigationService directly

1ec5: If I’m not mistaken, the concept of an access token is specific to the online Directions API. It’s also meaningless without an API endpoint hostname (which is filled in by default). It would be suboptimal for NavigationService to expose an API endpoint hostname when it isn’t necessarily backed by a server API.

PR #1850, opened 2018-11-22
v0.25.0

1ec5:
Updated the Xcode project, CocoaPods podspecs, and readme for v0.25.0.

/cc @mapbox/navigation-ios


TestHelper/Info.plist
```diff
@@ -15,8 +15,8 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0</string>
+	<string>0.25.0</string>
 	<key>CFBundleVersion</key>
-	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<string>2</string>
```
vincethecoder: Is this getting hard reset to `2`?

frederoni: yep, when agvtool starts incrementing.

vincethecoder: @frederoni that doesn't look like an `agvtool` variable. It looks like `$(CURRENT_PROJECT_VERSION)` is replaced with a constant value.

frederoni: `agvtool next-version -all` increments the value of `CFBundleVersion` which previously wasn't incrementable so it got bumped to 2.

PR #1854, opened 2018-11-22
Fix metric measurements in CarPlay

### LINKED ISSUE #1853: Metric distances are always 0 meters in CarPlay
When the application region is set to a region that uses the metric system, the distance to the maneuver in CarPlay is always shown as “0&nbsp;m”, regardless of the actual distance:

0 m

The issue appears to be related to rounding thresholds.

https://github.com/mapbox/mapbox-navigation-ios/blob/cc255c4d8b66bfab1150e321e696d54cf4debe76/MapboxCoreNavigation/DistanceFormatter.swift#L217

/ref #681
/cc @frederoni @JThramer

1ec5:
In CarPlay, the meter is once again defined as the distance such that the speed of light in vacuum _c_ is 299&nbsp;792&nbsp;458 when expressed in meters per second, where the second is defined in terms of the cesium frequency Δν<sub>Cs</sub>. Likewise with the centimeter and millimeter.

de anza

Fixes #1853.

/cc @frederoni @JThramer

### DIFF
--- a/MapboxCoreNavigation/DistanceFormatter.swift
+++ b/MapboxCoreNavigation/DistanceFormatter.swift
@@ -68,11 +68,11 @@ struct RoundingTable {
         func measurement(for distance: CLLocationDistance) -> Measurement<UnitLength> {
             switch unit {
             case .millimeter:
-                return Measurement(value: distance.kilometers / 1e6, unit: .millimeters)
+                return Measurement(value: distance * 1_000, unit: .millimeters)
             case .centimeter:
-                return Measurement(value: distance.kilometers / 1e5, unit: .centimeters)
+                return Measurement(value: distance * 100, unit: .centimeters)
             case .meter:
-                return Measurement(value: distance.kilometers / 1e3, unit: .meters)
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -68,11 +68,11 @@ struct RoundingTable {
...
             switch unit {
             case .millimeter:
-                return Measurement(value: distance.kilometers / 1e6, unit: .millimeters)
+                return Measurement(value: distance.kilometers * 1e6, unit: .millimeters)
             case .centimeter:
-                return Measurement(value: distance.kilometers / 1e5, unit: .centimeters)
+                return Measurement(value: distance.kilometers * 1e5, unit: .centimeters)
             case .meter:
-                return Measurement(value: distance.kilometers / 1e3, unit: .meters)
+                return Measurement(value: distance.kilometers * 1e3, unit: .meters)
```
frederoni: nit: `distance` should be sufficient here? meters→kilometers→meters feels excessive.

PR #1860, opened 2018-11-23
Publicize VisualInstruction helper methods

### LINKED ISSUE #1851: Suggestion: Change function maneuverImageSet access level to public inside VisualInstruction extension file
Hello, 
Currently function `func maneuverImageSet(side: DrivingSide) -> CPImageSet?` is used together with CarPlayNavigationViewController, but in cases when project builds it own UI solution it becomes impossible to access certain UI components from MapBox navigation SDK via it internal access level.
So my suggestion would be to change it access level to public.

frederoni:
Fixes #1851 

Made `VisualInstruction.containsLaneIndications`, `VisualInstruction.maneuverImageSet(side:)`, `VisualInstruction.shouldFlipImage(side:)`, and `VisualInstruction.carPlayManeuverLabelAttributedText(bounds:shieldHeight:window:)` public.

cc @1ec5 @JThramer

### DIFF
--- a/MapboxNavigation/VisualInstruction.swift
+++ b/MapboxNavigation/VisualInstruction.swift
@@ -5,13 +5,13 @@ import CarPlay
 
 extension VisualInstruction {
     
-    var containsLaneIndications: Bool {
+    public var containsLaneIndications: Bool {
         return components.contains(where: { $0 is LaneIndicationComponent })
     }
 
 #if canImport(CarPlay)
     @available(iOS 12.0, *)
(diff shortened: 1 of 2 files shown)


MapboxNavigation/VisualInstruction.swift
```diff
@@ -5,13 +5,13 @@ import CarPlay
 
 extension VisualInstruction {
     
-    var containsLaneIndications: Bool {
+    public var containsLaneIndications: Bool {
```
1ec5: Without documentation comments, these symbols won’t appear in the docset, so they’re effectively private.

PR #1861, opened 2018-11-26
Round distances in CarPlay

### LINKED ISSUE #1857: Distance to maneuver should be rounded in CarPlay
Distances shown in CarPlay’s maneuver panel aren’t rounded as they are inside NavigationViewController.

de anza

`RoundingTable.threshold(for:)` is getting called, and `DistanceFormatter.measurement(of:)` even sets `NumberFormatter.roundingIncrement` before calling `Threshold.measurement(for:)`, but `Threshold.measurement(for:)` doesn’t use the number formatter and doesn’t perform any rounding when creating the Measurement:

https://github.com/mapbox/mapbox-navigation-ios/blob/3abfaaebecfbe8274825f9d8ca5a3179143bcaa0/MapboxCoreNavigation/DistanceFormatter.swift#L217 https://github.com/mapbox/mapbox-navigation-ios/blob/3abfaaebecfbe8274825f9d8ca5a3179143bcaa0/MapboxCoreNavigation/DistanceFormatter.swift#L85

`Threshold.roundingIncrement` can be readily plugged into the `Double.rounded(to:)` method shown in https://github.com/mapbox/MapboxDirections.swift/pull/311#discussion_r233297022.

/ref #1853
/cc @mapbox/navigation-ios

frederoni:
Fixes #1857 

- [x] Round metric units
- [x] Round imperial units
- [x] Test remaining locale/regional settings

cc @1ec5

### DIFF
--- a/MapboxCoreNavigation/DistanceFormatter.swift
+++ b/MapboxCoreNavigation/DistanceFormatter.swift
@@ -68,32 +68,25 @@ struct RoundingTable {
         func measurement(for distance: CLLocationDistance) -> Measurement<UnitLength> {
             switch unit {
             case .millimeter:
-                return Measurement(value: distance * 1_000, unit: .millimeters)
+                return Measurement(value: distance, unit: .meters).converted(to: .millimeters)
             case .centimeter:
-                return Measurement(value: distance * 100, unit: .centimeters)
+                return Measurement(value: distance, unit: .meters).converted(to: .centimeters)
             case .meter:
                 return Measurement(value: distance, unit: .meters)
             case .kilometer:
-                let meterMeasurement = Measurement(value: distance, unit: UnitLength.meters)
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -68,21 +68,25 @@ struct RoundingTable {
         func measurement(for distance: CLLocationDistance) -> Measurement<UnitLength> {
             switch unit {
             case .millimeter:
-                return Measurement(value: distance * 1_000, unit: .millimeters)
+                return Measurement(value: distance, unit: .meters).converted(to: .millimeters)
```
1ec5: Good call, that’s a lot more elegant than what we had before. But if we stick to what we had before, then we can round after multiplying but before converting to a Measurement.

PR #1862, opened 2018-11-27
Updates views on BaseInstructionsBannerView to be public

nitaliano:
This PR updates most of the views, and distance members public so it is easier to build custom UI's on top of the `BaseInstructionsBannerView`

**cc** @akitchen @vincethecoder @1ec5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master
 
 * Fixed an issue where the CarPlay navigation map’s vanishing point and user puck initially remained centered on screen, instead of accounting for the maneuver panel, until the navigation bar was shown. ([#1856](https://github.com/mapbox/mapbox-navigation-ios/pull/1856))
+* Updated views on `BaseInstructionsBannerView` to be public to help with being able to build custom instruction banners. This is considered experiential and could change at any time.
 
 ## v0.25.0 (November 22, 2018)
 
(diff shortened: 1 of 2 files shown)


akitchen:
It would be good to add some documentation to the CHANGELOG since this is now new public API. 

Not sure if we need to expose more methods on any of these days, but we'll need to tread carefully when it comes to increasing API surface area.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## master
 
 * Fixed an issue where the CarPlay navigation map’s vanishing point and user puck initially remained centered on screen, instead of accounting for the maneuver panel, until the navigation bar was shown. ([#1856](https://github.com/mapbox/mapbox-navigation-ios/pull/1856))
+* Updated views on `BaseInstructionsBannerView` to be public to help with being able to build custom instruction banners. This is considered experiential and could change at any time.
```
1ec5: jazzy is configured to hide any undocumented method, so I think we can generally fall back to the same policy that the iOS map SDK (and Apple’s own APIs) follow: undocumented methods carry no inherent backward compatibility guarantees. In fact, the containing class is still marked `:nodoc:`, preventing any documentation from appearing even if we added documentation comments to these properties.

To avoid leading developers to use unsupported APIs, I removed this changelog entry in e7c17f019205e2eb9cee86574187bdec71250774.

PR #1868, opened 2018-11-30
Downsize shields in instructions in CarPlay

### LINKED ISSUE #1866: Exit number shields look blurry in CarPlay
We’ve gotten a report that exit number shields (and presumably also generic route shields) look blurry in turn-by-turn navigation on CarPlay devices.

blurry

This is on a 2× CarPlay device, presumably connected to a 2× iPhone.

/ref #1556 #1696
/cc @mapbox/navigation-ios

### LINKED ISSUE #1867: Shields missing in CarPlay simulator
Route shields and exit numbers don’t appear in turn-by-turn mode in the CarPlay simulator, even though they show up in CarPlay devices.

/ref https://github.com/mapbox/mapbox-navigation-ios/issues/1652#issuecomment-419486177
/cc @JThramer

1ec5:
When preparing attributed instruction strings for CarPlay, resize any text attachment image taller than 16 points, preserving the aspect ratio.

Here’s a before and after:

before
after

Shields on the phone are unaffected:

after-phone

Fixes #1867 and speculatively fixes #1866. A more correct fix for #1866 would require adjustments to the constraints in ExitView and GenericRouteShield.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -316,6 +316,13 @@ public class CarPlayNavigationViewController: UIViewController {
             return CGRect(x: 0, y: 0, width: widthOfManeuverView, height: 30)
         }
         
+        // Over a certain height, CarPlay devices downsize the image and CarPlay simulators hide the image.
+        let maximumImageSize = CGSize(width: .infinity, height: shieldHeight)
+        let imageRendererFormat = UIGraphicsImageRendererFormat(for: UITraitCollection(userInterfaceIdiom: .carPlay))
+        if let window = carPlayManager.carWindow {
+            imageRendererFormat.scale = window.screen.scale
+        }
+        
         if let attributedPrimary = visualInstruction.primaryInstruction.carPlayManeuverLabelAttributedText(bounds: bounds, shieldHeight: shieldHeight, window: carPlayManager.carWindow) {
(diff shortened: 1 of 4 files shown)


MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -316,6 +316,13 @@ public class CarPlayNavigationViewController: UIViewController {
             return CGRect(x: 0, y: 0, width: widthOfManeuverView, height: 30)
         }
         
+        // Over a certain height, CarPlay devices downsize the image and CarPlay simulators hide the image.
+        let maximumImageSize = CGSize(width: .infinity, height: shieldHeight)
```
frederoni: IIRC, the image max height of an attachment in  an attributed string drawn by a UILabel should be `fontSize.height + font.ascender + font.descender`.

1ec5: As far as I can tell, none of the accessibility settings on iOS affect the font size of the instructions in CarPlay. CarPlay units do have their own settings (such as the left/right layout setting), but I haven’t seen a setting about the font size in the units I’ve used so far.

frederoni: Hard limit before image gets cropped or scaled depending on other attributes. I thought we had some leeway but, if there seem to be a 16p limit, let's stick with that.

PR #1869, opened 2018-12-01
Speculative route-start course quick-fix

JThramer:
* Speculatively fixing issue where, upon starting a new leg of a route, a situation can occur where the RouteController interpolates user-course incorrectly, causing it to think that an off-route event has occured.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -221,6 +221,11 @@ open class RouteController: NSObject, Router {
      Monitors the user's course to see if it is consistantly moving away from what we expect the course to be at a given point.
      */
     func userCourseIsOnRoute(_ location: CLLocation) -> Bool {
+        // if we have yet to travel along the current leg, don't check for heading conformance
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -221,6 +221,11 @@ open class RouteController: NSObject, Router {
      Monitors the user's course to see if it is consistantly moving away from what we expect the course to be at a given point.
      */
     func userCourseIsOnRoute(_ location: CLLocation) -> Bool {
+        // if we have yet to travel along the current leg, don't check for heading conformance
+        guard routeProgress.currentLegProgress.distanceTraveled > 0 else {
+            return true
```
akitchen: Should we also reset movementsAwayFromRoute here? Seems like that could be part of the bug

JThramer: Yeah, that's a good idea.

akitchen:
Rebased with master / fixed up CHANGELOG.md organization

PR #1877, opened 2018-12-03
Nav-native deferral

JThramer:
Completely deferring to nav-native's authority on on-route status for `PortableRouteController`, also cleaning up code for legacy `RouteController` 

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -250,25 +250,27 @@ open class RouteController: NSObject, Router {
      */
     @objc public func userIsOnRoute(_ location: CLLocation) -> Bool {
         
+        let legProgress = routeProgress.currentLegProgress
+        
         // If the user has arrived, do not continue monitor reroutes, step progress, etc
-        if routeProgress.currentLegProgress.userHasArrivedAtWaypoint &&
-            (delegate?.router?(self, shouldPreventReroutesWhenArrivingAt: routeProgress.currentLeg.destination) ??
+        if legProgress.userHasArrivedAtWaypoint &&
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -277,10 +279,11 @@ open class RouteController: NSObject, Router {
         return false
     }
     
-    internal func userIsWithinRadiusOfRoute(location: CLLocation) -> Bool {
+    internal func userIsWithinEnvelope(of step: RouteStep, at location: CLLocation) -> Bool {
```
1ec5: “Envelope” in a spatial context is a bit novel to this codebase; is there a specific reason for using the term? The implementation still relies on a radius; is that intentional?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -277,10 +279,11 @@ open class RouteController: NSObject, Router {
...
         let radius = max(reroutingTolerance, RouteControllerManeuverZoneRadius)
-        let isCloseToCurrentStep = location.isWithin(radius, of: routeProgress.currentLegProgress.currentStep)
-        return isCloseToCurrentStep
+        let isCloseToCurrentStep = location.isWithin(radius, of: step)
+        let isWithinCourseTolerances = userCourseIsOnRoute(location)
```
1ec5: Did we determine whether the course-related heuristics in `userCourseIsOnRoute(_:)` are germane to the issue we’re attempting to fix?

JThramer:
Closing. It turns out that Nav-native does *not* duplicate the logic contained in `userCourseIsOnRoute(_:)`, which means we really need to fix the underlying issue, not just kick the can over to nav-native.

PR #1884, opened 2018-12-04
Incrementing navigator index for multi-leg routes.

### LINKED ISSUE #1878: PortableRouteController does not increment model leg index of navigator on multi-point routes.
During multi-waypoint routes, the `PortableRouteController` does not increment the native navigator's leg index when arriving at a waypoint. This causes the native navigator to get stuck on a status of `.COMPLETE`, breaking rerouting logic.

/cc @mapbox/navigation-ios

JThramer:
Fixes #1878, where the navigator can get out-of-sync with the route controller when navigating multi-leg routes. This could potentially cause a situation where the navigator always reports you as on-route for any leg other than the first.

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@
 
 ### Other changes
 
-* This release includes a possible fix for spurious rerouting when passing an intermediate waypoint. ([#1869](https://github.com/mapbox/mapbox-navigation-ios/pull/1869))
 * The `NavigationSettings.shared` property is now accessible in Objective-C code as `MBNavigationSettings.sharedSettings`. ([#1882](https://github.com/mapbox/mapbox-navigation-ios/pull/1882))
+* Fixing issue where the native-navigator may not declare off-route status for subsequent legs of a multi-leg route. 
 
 ## v0.25.0 (November 22, 2018)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 ### Other changes
 
 * The `NavigationSettings.shared` property is now accessible in Objective-C code as `MBNavigationSettings.sharedSettings`. ([#1882](https://github.com/mapbox/mapbox-navigation-ios/pull/1882))
-* Fixing issue where the native-navigator may not declare off-route status for subsequent legs of a multi-leg route. 
+* Fixing issue where the native-navigator may not declare off-route status for subsequent legs of a multi-leg route. ([#1884](https://github.com/mapbox/mapbox-navigation-ios/pull/1884))
 
 ## v0.25.0 (November 22, 2018)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 ### Other changes
 
 * The `NavigationSettings.shared` property is now accessible in Objective-C code as `MBNavigationSettings.sharedSettings`. ([#1882](https://github.com/mapbox/mapbox-navigation-ios/pull/1882))
-* Fixing issue where the native-navigator may not declare off-route status for subsequent legs of a multi-leg route. ([#1884](https://github.com/mapbox/mapbox-navigation-ios/pull/1884))
+* Fixed spurious rerouting on multi-leg routes. ([#1884](https://github.com/mapbox/mapbox-navigation-ios/pull/1884))
 
 ## v0.25.0 (November 22, 2018)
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,8 +11,8 @@
 
 ### Other changes
 
-* This release includes a possible fix for spurious rerouting when passing an intermediate waypoint. ([#1869](https://github.com/mapbox/mapbox-navigation-ios/pull/1869))
 * The `NavigationSettings.shared` property is now accessible in Objective-C code as `MBNavigationSettings.sharedSettings`. ([#1882](https://github.com/mapbox/mapbox-navigation-ios/pull/1882))
+* Fixing issue where the native-navigator may not declare off-route status for subsequent legs of a multi-leg route. ([#1884](https://github.com/mapbox/mapbox-navigation-ios/pull/1884))
```
1ec5: The “native navigator” is a private implementation detail.

```suggestion
* Fixed spurious rerouting on multi-leg routes. ([#1884](https://github.com/mapbox/mapbox-navigation-ios/pull/1884))
```

MapboxCoreNavigation/PortableRouteController.swift
```diff
@@ -67,6 +67,24 @@ open class PortableRouteController: RouteController {
         return !offRoute
     }
     
+    /**
+     Advances the leg index. This override also advances the leg index of the native navigator.
+     
+     This is a convienence method provided to advance the leg index of any given router without having to worry about the internal data structure of the router.
+     */
+    override public func advanceLegIndex(location: CLLocation) {
+        precondition(!routeProgress.isFinalLeg, "Can not increment leg index beyond final leg.")
```
1ec5: The superclass implementation already enforces this precondition. At least this redundant check should follow the `super` call.

MapboxCoreNavigation/PortableRouteController.swift
```diff
@@ -67,6 +67,24 @@ open class PortableRouteController: RouteController {
...
+        
+        super.advanceLegIndex(location: location)
+        
+        let status = navigator.getStatusForTimestamp(location.timestamp)
+        let legIndex = status.legIndex
+        let routeIndex = status.routeIndex
+
+        //The first route is the active one in the navigator.
+        navigator.changeRouteLeg(forRoute: routeIndex, leg: legIndex + 1)
```
1ec5: Can we do this to keep the models verifiably synchronized? Or is there a possibility that the navigator’s leg index is non-zero-based?

```suggestion
        navigator.changeRouteLeg(forRoute: routeIndex, leg: routeProgress.legIndex)
```

MapboxCoreNavigation/RouteController.swift
```diff
@@ -282,6 +282,17 @@ open class RouteController: NSObject, Router {
         let isCloseToCurrentStep = location.isWithin(radius, of: routeProgress.currentLegProgress.currentStep)
         return isCloseToCurrentStep
     }
+    
+    /**
+     Advances the leg index.
+     
+     This is a convienence method provided to advance the leg index of any given router without having to worry about the internal data structure of the router.
+     */
+    public func advanceLegIndex(location: CLLocation) {
```
1ec5: Is this method public solely in order to make it available to MapboxNavigation, or is it intended to be used by application code? If application code can use it, let’s make sure to expose it to Objective-C.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -282,6 +282,17 @@ open class RouteController: NSObject, Router {
...
+        precondition(!routeProgress.isFinalLeg, "Can not increment leg index beyond final leg.")
+        routeProgress.legIndex += 1
```
1ec5: Technically, it is possible for application code to reach in and set `RouteProgress.legIndex` to anything at any time. Prior to MapboxNavigationNative integration, doing so would presumably skip legs seamlessly. But now doing so could cause RouteProgress and the navigator to get out of sync. Perhaps PortableRouteController should observe its RouteProgress object’s `legIndex` key path and update the navigator in response.

Applications intentionally mucking with the leg index may be a bit of an edge case, so we can treat it as tail work. But I think accounting for this case will make the code much more robust, so we should definitely consider it.

1ec5: It may sound absurd that an application might set the leg index to an arbitrary value, but that could happen in an application that simulates routes and hooks up buttons to navigate back and forth among steps or legs.

1ec5: Going to treat this as tail work.

PR #1885, opened 2018-12-05
Show lane guidance in CarPlay guidance panel

### LINKED ISSUE #1858: Show lane guidance in CarPlay maneuver panel
The CarPlay maneuver panel currently omits lane guidance, whereas the top banner in NavigationViewController shows lane guidance via the route step’s tertiary instructions. [Apple suggests](https://developer.apple.com/carplay/documentation/CarPlay-Navigation-App-Programming-Guide.pdf#page=31) providing a second maneuver that contains a full-width `symbolSet`, no `instructionVariants`, and `CPManeuverDisplayStyle.symbolOnly`. We might be able to generate the required symbol set by snapshotting a LanesView. However, we’ll need to stop hard-coding some dimensions in LanesView.

/cc @mapbox/navigation-ios

1ec5:
CarPlayNavigationViewController now snapshots a LanesView for display in the CarPlay guidance panel. The layout needs work: currently all the LaneViews are smooshed up against the upper-left corner of the LanesView, and they’re shown at full scale but cropped.

lanes

LanesView is the right size, but the stack view inside it seems to be missing constraints:

lanesview

Additionally, LaneView will need to be modified to accept alternative dimensions for the arrows.

Fixes #1858.

/cc @JThramer @frederoni

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master
 
 * Search functionality on CarPlay is now managed by `CarPlaySearchController`. Added the `CarPlayManagerDelegate.carPlayManager(_:selectedPreviewFor:using:)` method for any additional customization after a trip is selected on CarPlay. ([#1846](https://github.com/mapbox/mapbox-navigation-ios/pull/1846))
+* The CarPlay guidance panel now shows lane guidance. ([#1885](https://github.com/mapbox/mapbox-navigation-ios/pull/1885))
 * Added the `NavigationViewController.showEndOfRouteFeedback(duration:completionHandler:)` method for showing the end-of-route feedback UI after manually ending a navigation session. ([#1932](https://github.com/mapbox/mapbox-navigation-ios/pull/1932))
 * Fixed an issue where setting `Router.route` did not update `SimulatedLocationManager.route`. ([#1928](https://github.com/mapbox/mapbox-navigation-ios/pull/1928))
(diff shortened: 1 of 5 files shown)


frederoni:
There was a missing width constraint when the LanesView isn't wrapped in a container view as you assumed.

The lane icons are not rendered pixel perfect but the constraints seem to be correct now. It's tricky when Xcode’s view hierarchy debugger doesn't work with CarPlay. Not sure if the blurriness comes from the PaintCode assets being incorrectly scaled or `.imageRepresentation` using the incorrect scale. Also, I think there's a separator view we should hide when the LanesView is being used in CarPlay.



Do you think we need to update the lane view icons? Maneuver icons supports both size and scale but the lane views only supports scale:





Actually, it's only the lane_slight_right that supports `scale`.

1ec5:
> Do you think we need to update the lane view icons? Maneuver icons support both size and scale, but the lane views only support scale:

That would be nice; the arrows end up looking pretty anemic at that size.

1ec5:
> Do you think we need to update the lane view icons? Maneuver icons support both size and scale, but the lane views only support scale

Would this be addressed by mapbox/navigation-ui-resources#7?

frederoni:
> Would this be addressed by mapbox/navigation-ui-resources#7?

Good point. Testing these here https://github.com/mapbox/mapbox-navigation-ios/pull/1910

akitchen:
Looks like this has gone stale. Can we revive it? It looks really promising.

1ec5:
I’m also seeing lanes get dropped in some cases. A route from 39.287527°N, 84.462424°W to 39.290445°N, 84.467252°W traverses [this way](https://www.openstreetmap.org/way/627151534) tagged `turn:lanes=left|left|through|through|right|right`, which is correctly interpreted by the Directions API, yet I only see four lane icons in the CarPlay UI:

carplay ios

1ec5:
Closing in favor of #2798.

PR #1895, opened 2018-12-11
Preserve historic route progress

### LINKED ISSUE #1876: Preserve user's historic route progress upon rerouting
Currently, when a user is navigating a route and a re-route event occurs the visual history of the route traveled is removed when the new route is set.

Instead we should merge the new route onto the historic route traveled, so that the user's end-to-end journey is displayed.

frederoni:
Fixes #1876

Preserve historic route progress between reroutes.

History is thrown away by default, implementing `NavigationViewControllerDelegate.navigationViewControllerShouldDiscardHistory(_:)` and returning `true` will pass `previousRouteProgress` onto the new `RouteProgress`.

- [x] Add unit tests to validate `.totalFractionTraveled`, `.totalDistanceTraveled`, and `.totalDistance`.
- [x] Add `.totalDistanceTraveled` total distance traveled along all previous routes including the current route’s `RouteProgress.distanceTraveled`.
- [x] Add `RouteProgress.totalFractionTraveled` the total fraction traveled along all reroutes including the full distance of the current route.
- [x] Add `RouteProgress.totalDistance` distance traveled along all previous routes including the full distance of the current route.
- [x] Add delegate methods to opt-in for historic route progress
- [ ] ~~Show history on the route line.~~

cc @akitchen @JThramer @1ec5 

| History Disabled        | History Enabled           |
| ------------- |-------------| 
|  |  | 

_labels showing "{totalDistanceTraveled}:{totalFractionTraveled}"_


MapboxNavigationTests/HistoricRouteProgressTests.swift
```diff
@@ -0,0 +1,30 @@
...
+        super.setUp()
+        recordMode = false
+        agnosticOptions = []
+    }
+
+    func testHistoricRouteProgress() {
+        // Short initial route
+        let route = Fixture.route(from: "historic-route-progress")
+        // Trace of a detour from the initial origin to the destination
+        let trace = Fixture.locations(from: "historic-route-progress.trace")
```
akitchen: For the sake of future efforts, how was this trace generated?

frederoni: `trace` was generated by a detour version of the `route` (same origin and destination but including a third detour waypoint) using `Fixture.generateTrace(for:)`, will add to the comment above.

frederoni:
@JThramer just storing previous route progresses doesn't work in many scenarios.
Because we're snapping a for a certain threshold on the old route before the reroute triggers, the two routes are often disconnected. It's even worse for overpasses.

Storing snapped coordinates won't work either. I wonder if we have to store all raw coordinates and snap them to the corresponding route progress afterward. 🤔

frederoni:
I dropped the commits with the visual history which will be implemented in a subsequent PR.
`.totalFractionTraveled` and `.totalDistanceTraveled` can reliably be used to calculate progress which takes previous routes into account.

akitchen:
Can we separate the visual and non-visual aspects of this? Perhaps the core navigation and navigation UI work can land separately.

/cc @frederoni @1ec5

frederoni:
> Can we separate the visual and non-visual aspects of this?

The non-visual aspects of this PR have been removed.
This PR is ready for review cc @JThramer @1ec5

akitchen:
I'm a little confused why we're using a snapshot test for testing the non-visual aspects of route history preservation, since a simple test showing a non-zero progress after a re-route would also help illustrate that the data is preserved. Snapshot tests are great for illustrating the behavior and preventing a regression, but they don't really help enforce or illustrate the API or contract. 

Nonetheless I think this is off to a great start. LGTM!

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -64,6 +64,40 @@ open class RouteProgress: NSObject {
...
+    }
+    
+    // Traverses all previous route progresses, excluding `self`.
+    func previousRouteProgresses() -> [RouteProgress] {
+        var current = self
+        var progresses = [RouteProgress]()
+        
+        while let prev = current.previousRouteProgress {
+            progresses.append(prev)
+            current = prev
```
1ec5: The sheer amount of functional programming in this file makes this non-recursive implementation stick out, but building up this array iteratively is a perfectly valid choice since it’s so predictable. 🙂

MapboxCoreNavigation/RouteController.swift
```diff
@@ -85,7 +84,7 @@ open class RouteController: NSObject, Router {
             return routeProgress.route
         }
         set {
-            routeProgress = RouteProgress(route: newValue)
+            routeProgress = RouteProgress(route: newValue, previousRouteProgress: shouldDiscardHistory ? nil : routeProgress)
```
1ec5: Should we be storing the entire previous route progress object, or just a skeletal one that includes only the information we care about? Can we get away with setting `previousRouteProgress` to a RouteProgress that’s been flattened into having a single leg with a single step, to reduce memory usage?

frederoni:
Closing in favor of #1980

PR #1898, opened 2018-12-12
Fixing Maneuver Arrows around waypoints

### LINKED ISSUE #1897: Arrows screw up when a waypoint is near.
Demo

When a maneuver comes right before a waypoint, the waypoint will be occluded and the arrow will point in a strange direction.

JThramer:
img_1438

Fixes #1897. 

To-Do: 
- [x] Write the snapshot test

* Fixing issue where arrow was looking at wrong leg for "following legs"
* Also inserting arrow under waypoints

/cc @mapbox/navigation-ios


akitchen:
Could a snapshot test show that the previous implementation was broken for this use case?

MapboxNavigation/Route.swift
```diff
@@ -20,7 +20,7 @@ extension Route {
         let precedingStepCoordinates = precedingSteps.compactMap { $0.coordinates }.reduce([], +)
         let precedingPolyline = Polyline((precedingLegCoordinates + precedingStepCoordinates).reversed())
         
-        let followingLegs = legs.prefix(upTo: legIndex)
+        let followingLegs = legs.suffix(from: legIndex).dropFirst()
```
1ec5: Thanks for catching this scrumptious bit of copy-pasta. 🍝 By the way, it should also be safe to say `legs.suffix(from: legIndex + 1)`, but this is fine.

JThramer:
For fun, this is a plot of the old behavior that was triggering the issue. Note that the "arrow" (a line) is pointing towards the first coordinate of the first step: 
testmaneuverarrowhandleswaypointscorrectly 3x

Issue #1902, opened 2018-12-14
Enable proactive rerouting by default


1ec5:
Proactive rerouting is currently disabled by default via the `RouteController.reroutesProactively` property:

https://github.com/mapbox/mapbox-navigation-ios/blob/6bc02c320b6f9831e337b522ba2ee09318bfea05/MapboxCoreNavigation/RouteController.swift#L60

This ticket tracks changing the default to true, because users expect traffic-aware directions to react to changes in traffic congestion. I believe this option was originally left disabled by default primarily because we didn’t implement any UI or voice prompt to indicate the receipt of a new route. So at a minimum, we should respond to a proactive rerouting notification by triggering a spoken instruction. Additionally, CarPlay has a built-in affordance for auto-dismissing modal alerts that is meant for indicating the selection of a faster route.

The change will affect the amount of bandwidth consumed by the SDK, but this is a basic user expectation, and the option remains to disable it. Note that the option is currently only capable of finding faster routes, not updating congestion lines for the existing route, which requires a more invasive change.

/cc @mapbox/navigation-ios @nitaliano

PR #1904, opened 2018-12-17
Migrate to nav-native’s route following

frederoni:
This PR further adopts NavigationNative for route following.
The old `RouteController` has been renamed to `LegacyRouteController` and is now deprecated.

Migrating to MapboxNavigationNative’s route following logic.

- [x] Refactor RouteController -> LegacyRouteController
- [x] Refactor PortableRouteController -> RouteController
- [x] Keep customizable arrival thresholds (incl willArrive and didArrive delegate methods)
- [x] Enable remaining events manager tests
- [x] Fix repetetive voice instructions

cc @JThramer


MapboxCoreNavigation/PortableRouteController.swift
```diff
@@ -56,30 +162,208 @@ open class PortableRouteController: RouteController {
...
+        if let location = location {
+            userInfo[.locationKey] = location
+        }
+        userInfo[.isProactiveKey] = didFindFasterRoute
+        NotificationCenter.default.post(name: .routeControllerDidReroute, object: self, userInfo: userInfo)
+        delegate?.router?(self, didRerouteAlong: routeProgress.route, at: dataSource.location, proactive: didFindFasterRoute)
+    }
+    
     /**
      Advances the leg index. This override also advances the leg index of the native navigator.
```
akitchen: I guess this is no longer an override? :)

MapboxCoreNavigation/PortableRouteController.swift
```diff
@@ -94,3 +378,70 @@ open class PortableRouteController: RouteController {
         return navigator.getHistory()
     }
 }
+
+extension PortableRouteController: Router {
+    public func userIsOnRoute(_ location: CLLocation) -> Bool {
```
akitchen: This appears to be the same as `userIsWithinRadiusOfRoute()` [above](https://github.com/mapbox/mapbox-navigation-ios/pull/1904/files#diff-1dff2075b809825263a18d19396f5c0aR343)

MapboxCoreNavigation/PortableRouteController.swift
```diff
@@ -1,14 +1,18 @@
...
 import MapboxNavigationNative
+import MapboxMobileEvents
 import MapboxDirections
+import Polyline
+import Turf
 
 
 @objc(MBPortableRouteController)
-open class PortableRouteController: RouteController {
+open class PortableRouteController: NSObject {
```
akitchen: This is an exciting change. What is going to happen with the existing `RouteController` as a part of this PR?

frederoni: Ideally, we would just delete it, but I assume we have to deprecate it. 🤔

MapboxCoreNavigation/RouteController.swift
```diff
@@ -1,199 +1,153 @@
...
-        self._routeProgress = RouteProgress(route: route)
-        self.dataSource = source
-        UIDevice.current.isBatteryMonitoringEnabled = true
-
-        super.init()
-        
-        checkForUpdates()
-        checkForLocationUsageDescription()
+    @objc func setWaypointArrivalThreshold(_ threshold: TimeInterval) {
+        navigator.configureNavigator(forName: "arrival_threshold_duration", value: "\(threshold)")
```
frederoni: This is less than ideal, however, the property is currently write-only in nav-native.

1ec5: Unless MapboxNavigationNative reserves the right to silently change the threshold, we could store a copy of the new value to be returned in the getter.

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -191,4 +190,12 @@ extension CLLocation {
         }
         return true
     }
+    
+    func shifted(to newTimestamp: Date) -> CLLocation {
+        return CLLocation(coordinate: coordinate, altitude: altitude, horizontalAccuracy: horizontalAccuracy, verticalAccuracy: verticalAccuracy, course: course, speed: speed, timestamp: newTimestamp)
```
1ec5: It occurs to me that this method is lossy if `CLLocation.floor` happens to be set, but I don’t think that poses a practical problem at the moment.

Apple could add new properties to CLLocation in the future, which make this method even lossier. To guard against that, we could have a unit test that shifts a CLLocation to its own timestamp, then checks whether the original and shifted objects are equal.

MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -0,0 +1,660 @@
...
+
+
+protocol RouteControllerDataSource: class {
+    var location: CLLocation? { get }
+    var locationProvider: NavigationLocationManager.Type { get }
+}
+
+
+/**
+ A `RouteController` tracks the user’s progress along a route, posting notifications as the user reaches significant points along the route. On every location update, the route controller evaluates the user’s location, determining whether the user remains on the route. If not, the route controller calculates a new route.
```
1ec5: This documentation refers to RouteController but documents LegacyRouteController. Note that jazzy lacks support for deprecation warnings: realm/jazzy#843. In the past, we’ve gone ahead and removed documentation comments from deprecated symbols, to avoid promoting their continued usage. However, this has historically been such an important class that we might want to leave the comment in place and add a manual deprecation note.

MapboxCoreNavigation/MBNavigator.swift
```diff
@@ -0,0 +1,43 @@
...
+            return "offRoute"
+        case .stale:
+            return "stale"
+        }
+    }
+}
+
+extension MBNavigationStatus {
+    
+    var isFirstInstructionOnFirstStep: Bool {
```
1ec5: `isAtFirstInstructionOfLeg`? Does this logic need to account for multi-leg routes?

MapboxCoreNavigation/Router.swift
```diff
@@ -9,19 +9,54 @@ public protocol RouterDataSource {
...
+     The route controller’s delegate.
+     */
     @objc var delegate: RouterDelegate? { get set }
     
+    /**
+     Intializes a new `RouteController`.
+     
+     - parameter route: The route to follow.
+     - parameter directions: The Directions object that created `route`.
+     - parameter source: The data source for the RouteController.
```
1ec5: Thanks for documenting these symbols!

MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -25,7 +24,13 @@ class MapboxCoreNavigationTests: XCTestCase {
...
         route.accessToken = "foo"
         navigation = MapboxNavigationService(route: route, directions: directions, simulating: .never)
-        let depart = CLLocation(coordinate: CLLocationCoordinate2D(latitude: 37.795042, longitude: -122.413165), altitude: 1, horizontalAccuracy: 1, verticalAccuracy: 1, course: 0, speed: 10, timestamp: Date())
+        
+        // Coordinates from first step
+        let coordinates = route.legs[0].steps[0].coordinates!
+        let locations = coordinates.enumerated().map { CLLocation(coordinate: $0.element,
+                                                                  altitude: -1, horizontalAccuracy: 10,
+                                                                  verticalAccuracy: -1, course: -1, speed: 10,
+                                                                  timestamp: Date().addingTimeInterval(TimeInterval($0.offset))) }
```
1ec5: I’m assuming these offsets are all measured from the same starting time. In general, to ensure accuracy, a single Date instance should be created outside the call to `map(_:)`, and this line would add a time interval to that single Date instance.

Also, it looks like this PR adds a `+` operator for Date and TimeInterval that could be used here.

(×5)

MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -36,7 +41,8 @@ class MapboxCoreNavigationTests: XCTestCase {
         }
         
         navigation.start()
-        navigation.locationManager(navigation.locationManager, didUpdateLocations: [depart])
+        
+        locations.forEach { navigation.locationManager(navigation.locationManager, didUpdateLocations: [$0]) }
```
1ec5: Nit: this call to `forEach(_:)` doesn’t add a whole lot of clarity over a standard `for` loop.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -758,7 +758,7 @@ extension NavigationViewController: StyleManagerDelegate {
 public extension NavigationViewController {
     @available(*, obsoleted: 0.1, renamed: "navigationService", message: "NavigationViewController no-longer directly manages a RouteController. See MapboxNavigationService, which contains a protocol-bound reference to the RouteController, for more information.")
     /// :nodoc: obsoleted
-    @objc public final var routeController: RouteController! {
+    @objc public final var routeController: LegacyRouteController! {
```
1ec5: This use of the now-deprecated LegacyRouteController class introduces build warnings, right? What’s our strategy for suppressing those warnings, which would show up in a CocoaPods-managed application project?

frederoni: These were deprecated 4 months ago. Can we remove them? They provide fix-its but apart from that, they just call `fatalError()`.

MapboxCoreNavigation/MBNavigator.swift
frederoni: This is actually dead code so I removed it. nav-native is handling this now.

MapboxCoreNavigation/CLLocation.swift
frederoni: CLLocation doesn't implement equatable but I added a test for shifting and FixLocation->CLLocation.

MapboxCoreNavigation/Router.swift
```diff
@@ -9,19 +9,54 @@ public protocol RouterDataSource {
...
+     */
     @objc(initWithRoute:directions:locationManager:)
```
JThramer: `initWithRoute:directions:dataSource:`

1ec5:
The <kbd>backwards-incompatible</kbd> label was added in part because the “renaming” of LegacyRouteController to RouteController isn’t a one-for-one replacement. For example, the following change hasn’t been documented:

* As part of replacing LegacyRouteController (né RouteController) with RouteController (née PortableRouteController), the `reroutesProactively` option was effectively removed. This is a useful option, one that #1902 would’ve enabled by default. Developers are now forced to choose between MapboxNavigationNative-based route following functionality and the ability to navigate users around flashmobs, so to speak.

PR #1905, opened 2018-12-18
Moving navigation service stop into deinit method.

### LINKED ISSUE #1899: Navigation Service breaks when we change style.
Fun times.

Basically, the navigation service `stop()`'s when the `NavigationViewController` hits it's `viewWillDisappear` [method](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/NavigationViewController.swift#L448), which is triggered whenever we [update the style.](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/StyleManager.swift#L189)

We could simply restart on `viewWillAppear`, but the upshot of this is that updates would be disabled whenever we're in the background, which is no-bueno.

A solution candidate on this would be to trigger on `deinit` instead, and also to create a dedicated `endNavigation` method (or something) for `NavigationViewController`. Thoughts?

/cc @mapbox/navigation-ios

JThramer:
Can't stop

Fixes #1899.

I really spent some time here looking for a better interface into `NavigationViewController`, but I just couldn't really find one that made sense. Stop is normally called when the navigation service receives a `endNavigation(feedback:)` call, which happens at the EORVC. We could expose an `endNavigation` mechanism to the UI SDK, but that brings along its' own considerations.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -417,6 +417,7 @@ open class NavigationViewController: UIViewController {
     }
     
     deinit {
+        navigationService.stop()
         suspendNotifications()
     }
     
@@ -445,7 +446,6 @@ open class NavigationViewController: UIViewController {
             UIApplication.shared.isIdleTimerDisabled = false
         }
         
-        navigationService.stop()
     }
     
     // MARK: Route controller notifications


frederoni:
Stopping in deinit makes perfect sense. It's up to the developer to stop the nav-service if someone wants to stop it prematurely.

PR #1909, opened 2018-12-19
Upgrade to Mapbox Streets source v8

### LINKED ISSUE #1896: Upgrade to Mapbox Streets source v8
mapbox/mapbox-gl-native#13481 adds support for [Mapbox Streets source v8](https://www.mapbox.com/vector-tiles/mapbox-streets-v8/), which is currently in beta. By the time map SDK v4.7.0 comes out, we’ll need to add corresponding support for Streets source v8; otherwise, anyone using a v8-dependent style will incur twice as much overhead from vector tile downloading and processing, while map labels and the current road name label will go blank. We’ll want to retain support for v7 for the time being, since some popular navigation styles still use v7.

Here are the changes we’ll need to make to accommodate both v7 and v8:

* [x] Detect Streets source v8: https://github.com/mapbox/mapbox-navigation-ios/blob/dc9f1e9be68b18ca4f57882c98e0a58e85461507/MapboxNavigation/MGLVectorTileSource.swift#L9
* [x] Add Streets source v8 instead of v7 when Streets source is missing from the style: https://github.com/mapbox/mapbox-navigation-ios/blob/6bc02c320b6f9831e337b522ba2ee09318bfea05/MapboxNavigation/RouteMapViewController.swift#L777-L782
* [x] Use the `road` layer instead of `road_label` when the style contains Streets source v8: https://github.com/mapbox/mapbox-navigation-ios/blob/6bc02c320b6f9831e337b522ba2ee09318bfea05/MapboxNavigation/NavigationMapView.swift#L1004-L1005 https://github.com/mapbox/mapbox-navigation-ios/blob/6bc02c320b6f9831e337b522ba2ee09318bfea05/MapboxNavigation/RouteMapViewController.swift#L786

/cc @mapbox/navigation-ios @mapbox/maps-ios @brsbl

1ec5:
Both label localization and feature querying for the current road name label now support styles powered by Mapbox Streets source v8. If the current style doesn’t use the Streets source, version 8 is now added automatically instead of v7.

Fixes #1896.

/cc @frederoni @JThramer

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -1006,7 +1006,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             }
             
             // Road labels should match road signage.
-            let locale = layer.sourceLayerIdentifier == "road_label" ? Locale(identifier: "mul") : nil
+            let isLabelLayer = MGLVectorTileSource.roadLabelLayerIdentifiersByTileSetIdentifier.values.contains(layer.sourceLayerIdentifier ?? "")
+            let locale = isLabelLayer ? Locale(identifier: "mul") : nil
             
             let localizedText = text.mgl_expressionLocalized(into: locale)
             if localizedText != text {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master
 
 * The `NavigationDirections.unpackTilePack(at:outputDirectoryURL:progressHandler:completionHandler:)` method is now available to Objective-C code as `-[MBNavigationDirections unpackTilePackAtURL:outputDirectoryURL:progressHandler:completionHandler:]`. ([#1891](https://github.com/mapbox/mapbox-navigation-ios/pull/1891))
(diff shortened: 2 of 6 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -1006,7 +1006,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             }
             
             // Road labels should match road signage.
-            let locale = layer.sourceLayerIdentifier == "road_label" ? Locale(identifier: "mul") : nil
+            let isLabelLayer = MGLVectorTileSource.roadLabelLayerIdentifiersByTileSetIdentifier.values.contains(layer.sourceLayerIdentifier ?? "")
+            let locale = isLabelLayer ? Locale(identifier: "mul") : nil
```
frederoni: What's "mul"? Doesn't seem to be supported by `Locale` 🤔. If so, won't this always evaluate to `nil`?

1ec5: `mul` is the ISO&nbsp;639 identifier meaning “multilingual”. The map SDK maps this code to the `name` field. Since that field displays the local language for each feature, it effectively produces a multilingual map. `Locale(identifier: "mul")` is supported in the sense that its `languageCode` is `mul`, which is all we need for this API.

PR #1910, opened 2018-12-19
Update LanesStyleKit

### LINKED ISSUE #576: Lane icons need to handle more lane configurations
[`LaneArrowView.draw(_:)`](https://github.com/mapbox/mapbox-navigation-ios/blob/622d092808b660085cf149411de7886dfd810b01/MapboxNavigation/LaneArrowView.swift#L34) omits some common lane indication scenarios, which means we probably give misleading information in those cases:

* [x] Three indications on a single lane — There are hundreds of ways in OpenStreetMap that contain `left;through;right` in the [`turn:lanes`](http://taginfo.openstreetmap.org/keys/turn%3Alanes#values) tag, which maps to a `LaneIndication` of `[.left, .straight, .right]` in MapboxDirections.swift ([example](https://www.openstreetmap.org/way/394222593)). There are also `reverse;left;through` lanes in some places.
* [x] U-turn lanes — There are hundreds of ways that contain `reverse;left`, which maps to `[.uTurn, .left]`. For any step that requires a U-turn, this becomes relevant for situations like `reverse;left|left`, when a lane that allows U-turns and left turns is next to a lane that allows only left turns – the driver needs to get into the leftmost lane, not the two left lanes.
* [x] Left-right turn lanes — There are thousands of ways that contain `left;right`, which maps to `[.left, .right]` ([example](http://www.openstreetmap.org/way/517630993); open iD to see the pavement marking). This is especially common at a T-intersection or where a road becomes a one-way road.

[](https://www.tapconet.com/store/product-detail/vkbb/r61-26-ca-triple-lane-control-left-and-right-turn-symbol-sign?c=Bmab&sku=311110)

The U.S. signs above are based on the following standard:

http://mutcd.fhwa.dot.gov/SHSe/Regulatory.pdf#page=29
http://mutcd.fhwa.dot.gov/SHSe/Regulatory.pdf#page=62

Commons has plenty of SVGs of U.S. and European lane control signs that we can repurpose for this UI:

https://commons.wikimedia.org/wiki/Category:Lane_turning_road_signs
https://commons.wikimedia.org/wiki/Category:Mandatory_direction_road_signs
https://commons.wikimedia.org/wiki/Category:R3-6_-_Optional_Movement_Lane_Control
https://commons.wikimedia.org/wiki/Category:R3-18_-_No_Left_or_U_turn

/cc @bsudekum @mayagao @frederoni @daniel-j-h @danesfeder @ericrwolfe

frederoni:
Testing the updated LanesStyleKit with scalable support from https://github.com/mapbox/navigation-ui-resources/pull/7

- [x] Unblocks #1885
- [x] Partially fixes #576
- [x] Fixed a bug which caused `|left|straightAhead|` to be drawn with a right turn.
- [x] Fix u-turn for left hand rule of the road.

~Left turns are not flipped when creating screenshots because the transform is not concatenated into the context. This could be problematic for CarPlay which uses an `.imageRepresentation` of `LaneView`.~ fixed

cc @1ec5 @JThramer


frederoni:
scales incorrectly above 30 points.
testsharpright60x60_iphone_11_4 3x

akitchen:
This looks pretty straightforward, other than the scaling issue noted in https://github.com/mapbox/mapbox-navigation-ios/pull/1910#issuecomment-448675968 . Is that a merge-stopping regression?

frederoni:
The incorrect scale mentioned in https://github.com/mapbox/mapbox-navigation-ios/pull/1910#issuecomment-448675968 is fixed

Issue #1794, opened 2018-10-23
Enable more extensive customization of the bottom banner in the drop-in navigation UI


akitchen:
It should be possible for developers to supply their own view and viewcontroller to popuplate and manage the bottom banner for more complicated/customized use cases. 

Questions exist as to what would become of the existing bottom banner in this case - should we offer developers the opportunity to add it / compose it into their own view hierarchy or would they need to replace its functionality entirely?

akitchen:
@nitaliano and I hacked on a proof-of-concept for this in https://github.com/mapbox/mapbox-navigation-ios/tree/bottom-stack-view , but rather than convert the bottom content view to a stack view, we should be providing a simple container view in which a developer can insert their own view to be managed by their own child viewcontroller

This branch is still relevant as it clarifies the autolayout fixes needed in order to begin decoupling from the existing drop-in UI's assumptions.

akitchen:
delivered in https://github.com/mapbox/mapbox-navigation-ios/pull/1951

PR #1912, opened 2018-12-19
Customizable Bottom Banner - Prep

JThramer:
Oh boy!

Yep, here it comes. Prepares for implementation of componentization in #1794.

To-Do: 
- [x] Investigate failing tests
- [x] Ensure parity of delegate messages with notification messages -- we should use both mechanisms equally well
- [x] Component-ize all UI components and encapsulate related logic into those component types
- [x] ~Encapsulate the bottom banner in a UIViewController and make it customizable~ (Will be completed in #1951 )

/cc @mapbox/navigation-ios


MapboxNavigation/BottomBannerView.swift
```diff
@@ -77,6 +94,31 @@ open class BottomBannerView: UIView {
...
+    }
+    
+    @objc func removeTimer() {
+        timer?.invalidate()
+        timer = nil
+    }
+    
+    @objc func resetETATimer() {
+        removeTimer()
+        timer = Timer.scheduledTimer(timeInterval: 30, target: self, selector: #selector(refreshETA), userInfo: nil, repeats: true)
```
akitchen: I'm a little worried about using a scheduled timer here as it introduces a retain cycle. Previously, when owned by a view controller we broke the cycle on view disappearance, but now I'm worried we won't actually break the cycle if the only call to `removeTimer()` is in `deinit()`

JThramer: @akitchen It's worth pointing out that this code was moved (largely wholesale) from the [RouteMapViewController](https://github.com/mapbox/mapbox-navigation-ios/pull/1912/files#diff-292af689ce88d5b5aa06966456abc0e9L911). Do we want to change the mechanism of action here?

akitchen: yeabut the RouteMapViewController tore down the timer on view disappearance, which was more eager. it's that subtle difference that concerns me

would our DispatchTimer https://github.com/mapbox/mapbox-navigation-ios/blob/31e0742ad39fbb44842b20693135b190f9e00a60/MapboxCoreNavigation/DispatchTimer.swift be a better fit for this new factoring?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -611,35 +611,55 @@ extension RouteController: CLLocationManagerDelegate {
...
+        
+        let info: [RouteControllerNotificationUserInfoKey: Any] = [
+                .routeProgressKey: routeProgress,
+                .spokenInstructionKey: spokenInstructionPoint
+        ]
+        
+        NotificationCenter.default.post(name: .routeControllerDidPassSpokenInstructionPoint, object: self, userInfo: info)
+    }
+    
+    private func announcePassage(of visualInstructionPoint: VisualInstructionBanner, along routeProgress: RouteProgress) {
```
1ec5: Should we tell the application that the user has passed the point for a VisualInstructionBanner or the VisualInstruction itself? There are situations where a tertiary banner can appear independently of (after) the same visual instruction’s primary and secondary banners; is this method called in that case?

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -2803,7 +2815,7 @@
 			buildSettings = {
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
-				DEBUG_INFORMATION_FORMAT = dwarf;
+				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
```
1ec5: Can you split this and the `STRIP_SWIFT_SYMBOLS` change out into a separate PR if it’s intentional?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -450,49 +454,6 @@ open class NavigationViewController: UIViewController {
...
-    @objc func progressDidChange(notification: NSNotification) {
-        let routeProgress = notification.userInfo![RouteControllerNotificationUserInfoKey.routeProgressKey] as! RouteProgress
-        let location = notification.userInfo![RouteControllerNotificationUserInfoKey.locationKey] as! CLLocation
-        let secondsRemaining = routeProgress.currentLegProgress.currentStepProgress.durationRemaining
-
-        mapViewController?.notifyDidChange(routeProgress: routeProgress, location: location, secondsRemaining: secondsRemaining)
-        
-        // If the user has arrived, don't snap the user puck.
-        // In the case the user drives beyond the waypoint,
-        // we should accurately depict this.
```
1ec5: Is this functionality covered now by `navigationService(_:didUpdate:with:rawLocation:)`?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -658,23 +622,49 @@ extension NavigationViewController: NavigationServiceDelegate {
         
         //Check to see if we're in a tunnel.
         checkTunnelState(at: location, along: progress)
+        
+        
+        //Pass the message onto our navigation components
+        navigationComponents.forEach { $0.navigationService?(service, didUpdate: progress, with: location, rawLocation: rawLocation) }
```
1ec5: Nit: using `forEach(_:)` here doesn’t really improve readability over a standard `for` loop.

MapboxNavigation.xcodeproj/project.pbxproj
JThramer: Unintentional. Will revert.

MapboxNavigation/NavigationViewController.swift
JThramer: [Yes.](https://github.com/mapbox/mapbox-navigation-ios/pull/1912/files/f85bcf7ed88881e440863d506c84665a672366d7#diff-ab5a908dc637e51fa78cf2be28d033ecR632)

MapboxCoreNavigation/RouteController.swift
```diff
@@ -611,35 +611,55 @@ extension RouteController: CLLocationManagerDelegate {
...
-                
                 routeProgress.currentLegProgress.currentStepProgress.visualInstructionIndex += 1
+                announcePassage(of: visualInstruction, along: routeProgress)
+                
                 return
             }
         }
     }
+    
+    private func announcePassage(of spokenInstructionPoint: SpokenInstruction, along routeProgress: RouteProgress) {
```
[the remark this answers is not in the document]
JThramer: Agreed, I kind of prefer it this way: "Announce passage of the spoken instruction point along the route's progress"

MapboxCoreNavigation/RouteController.swift
```diff
@@ -611,35 +611,55 @@ extension RouteController: CLLocationManagerDelegate {
...
+        
+        delegate?.router?(self, didPassSpokenInstructionPoint: spokenInstructionPoint, along: routeProgress)
+        
+        let info: [RouteControllerNotificationUserInfoKey: Any] = [
+                .routeProgressKey: routeProgress,
+                .spokenInstructionKey: spokenInstructionPoint
+        ]
+        
+        NotificationCenter.default.post(name: .routeControllerDidPassSpokenInstructionPoint, object: self, userInfo: info)
```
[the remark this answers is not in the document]
JThramer: So what is to be done? do we need a `keypathsAffectingValueForKey:`([ref](https://developer.apple.com/documentation/objectivec/nsobject/1414299-keypathsforvaluesaffectingvaluef?language=objc)) somewhere?

MapboxCoreNavigation/RouteController.swift
JThramer: This is an interesting question, but I'm not sure it's in-scope of this PR, considering that I'm merely exposing the same objects we used to announce through notification, as a delegate message. 

Also, considering that the notification existed before this PR, there are probably deprecation concerns to think of.

MapboxCoreNavigation/RouteController.swift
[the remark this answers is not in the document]
1ec5: Looks like we normally say “for” a route progress, not “along”. But given that this is a subsequent argument, let’s just leave it unlabeled: `routeProgress`.

MapboxCoreNavigation/RouteController.swift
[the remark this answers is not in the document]
1ec5: Yes, the `RouteStepProgress.remainingVisualInstructions` property is affected by the `instructionsDisplayedAlongStep` and `visualInstructionIndex` keys.

MapboxCoreNavigation/RouteController.swift
1ec5: I brought up the issue of VisualInstructionBanner versus VisualInstruction because it might affect the signature of any public methods we introduce in this release regarding visual instructions. I think we need to figure out whether a tertiary instruction appearing independently, away from a maneuver, can trigger a notification and delegate method. If not, I suspect this method should take a VisualInstruction rather than a VisualInstructionBanner; likewise for the notification.

akitchen: This is sort of a confusing case -- the `VisualInstructionBanner` has a `VisualInstruction` as its `tertiaryInstruction`

Maybe there should be separate notifications if the case you cited is a thing we need to worry about. Is this something that happens frequently? Do you have an example handy?

JThramer: > I think we need to figure out whether a tertiary instruction appearing independently, away from a maneuver, can trigger a notification and delegate method.

@1ec5 it looks like this is [not the case](https://github.com/mapbox/mapbox-navigation-ios/pull/1912/files#diff-1a2149692b3d7c58b6a0392120d6011aR633). The only time `announcePassage(of:along:)` is called is when we approach a maneuver.

MapboxCoreNavigation/RouteController.swift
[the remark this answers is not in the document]
1ec5: Ah, I went searching for that pattern but didn’t find it for some reason. 👓

I think we should eliminate that pattern before v1.0 (especially the [labelless `along`](https://github.com/mapbox/mapbox-navigation-ios/blob/dac1496fb758fa8354c527e925113ec9fce96bdf/MapboxCoreNavigation/Router.swift#L21)), because it’s misleading. To reduce the overall churn, let’s start now with just the public methods introduced in this PR; we can deal with the rest of the `along`s separately.

MapboxCoreNavigation/RouteController.swift
1ec5: I see, so NextBannerView knows to unhide itself not because of a specific notification but because it gets updated on every route progress update notification.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -601,4 +601,16 @@ open class RouteStepProgress: NSObject {
         guard visualInstructionIndex < instructionsDisplayedAlongStep.count else { return nil }
         return instructionsDisplayedAlongStep[visualInstructionIndex]
     }
+    
+    @objc override open class func keyPathsForValuesAffectingValue(forKey key: String) -> Set<String> {
```
1ec5: This is fine, but it’s a bit cleaner to implement separate [`keyPathsAffectingValueForRemainingVisualInstructions` and `keyPathsAffectingValueForRemainingSpokenInstructions`](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/KeyValueObserving/Articles/KVODependentKeys.html#//apple_ref/doc/uid/20002179-SW2) properties.

MapboxNavigation/BottomBannerView.swift
```diff
@@ -77,6 +101,34 @@ open class BottomBannerView: UIView {
...
+        updateETA(routeProgress: progress)
+        previousProgress = progress
+    }
+    
+    @objc func removeTimer() {
+        timer?.disarm()
+        timer = nil
+    }
+    
+    @objc func resetETATimer() {
```
frederoni: What's the reasoning behind resetting this timer every second?

JThramer: Good question. Logic was taken wholesale from the RMVC.

JThramer: Posterity: After discussions with @frederoni and @akitchen it was decided to remove this in a fast-followup PR.

1ec5: #1948

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -3330,6 +3341,7 @@
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
+				STRIP_SWIFT_SYMBOLS = YES;
```
akitchen: Should we be taking this change, and other adjacent project file changes related to symbol stripping? It seems they should maybe go into a separate PR if they're intentional.

JThramer: Good catch. Reverted.

MapboxNavigationTests/Support/FBSnapshotTestCase.swift
```diff
@@ -1,12 +1,15 @@
 import Foundation
 import FBSnapshotTestCase
 
+fileprivate let suffix: NSOrderedSet = ["_64"]
+
 @nonobjc extension FBSnapshotTestCase {
+    
     func verify(_ view: UIView, overallTolerance: CGFloat = 0.05) {
-        FBSnapshotVerifyView(view, suffixes: ["_64"], overallTolerance: overallTolerance)
+        FBSnapshotVerifyView(view, suffixes: suffix, perPixelTolerance: overallTolerance)
```
frederoni: perPixelTolerance != overallTolerance
Rename the parameter if you think perPixelTolerance is more appropriate.

MapboxCoreNavigation/MBRouteController.h
```diff
@@ -55,10 +55,20 @@ extern const NSNotificationName MBRouteControllerDidPassVisualInstructionPointNo
 typedef NSString *MBRouteControllerNotificationUserInfoKey NS_EXTENSIBLE_STRING_ENUM;
 
 /**
- A key in the user info dictionary of a `Notification.Name.MBRouteControllerProgressDidChange` or `Notification.Name.RouteControllerDidPassSpokenInstructionPoint` notification. The corresponding value is a `RouteProgress` object representing the current route progress.
+ A key in the user info dictionary of a `Notification.Name.routeControllerProgressDidChange`, `Notification.Name.routeControllerDidPassVisualInstructionPoint`, or `Notification.Name.routeControllerDidPassSpokenInstructionPoint` notification. The corresponding value is a `RouteProgress` object representing the current route progress.
```
frederoni: `Notification.Name.routeControllerProgressDidChange` is a generated Swift symbol so jazzy won't have access to it. Document according to Obj-C symbols here.

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -72,6 +72,25 @@ import MapboxDirections
...
+     - parameter service: The navigation service that passed the instruction point.
+     - parameter instruction: The instruction to be presented.
+     - parameter routeProgress: The routeProgress model through which the service is progressing.
+    */
+    @objc(navigationService:didPassVisualInstructionPoint:routeProgress:)
+    optional func navigationService(_ service: NavigationService, didPassVisualInstructionPoint instruction: VisualInstructionBanner, routeProgress: RouteProgress)
+    
+    
+    /**
+     Called when the navigation services passes a spoken instruction point.
```
1ec5:
```suggestion
     Called when the navigation service detects that the user has passed a point at which an instruction should be spoken.
```

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -72,6 +72,25 @@ import MapboxDirections
...
+     - parameter service: The navigation service that passed the instruction point.
+     - parameter instruction: The instruction to be spoken.
+     - parameter routeProgress: The routeProgress model through which the service is progressing.
```
1ec5:
```suggestion
     - parameter routeProgress: The route progress object that the navigation service is updating.
```

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -72,6 +72,25 @@ import MapboxDirections
     @objc(navigationService:didUpdateProgress:withLocation:rawLocation:)
     optional func navigationService(_ service: NavigationService, didUpdate progress: RouteProgress, with location: CLLocation, rawLocation: CLLocation)
     
+    /**
+     Called when the navigation services passes a visual instruction point.
```
1ec5:
```suggestion
     Called when the navigation service detects that the user has passed a point at which an instruction should be displayed.
```

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -72,6 +72,25 @@ import MapboxDirections
...
+     - parameter service: The navigation service that passed the instruction point.
+     - parameter instruction: The instruction to be presented.
+     - parameter routeProgress: The routeProgress model through which the service is progressing.
```
1ec5:
```suggestion
     - parameter routeProgress: The route progress object that the navigation service is updating.
```

MapboxNavigation/NavigationComponent.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+import MapboxCoreNavigation
+import CoreLocation
+
+@objc protocol NavigationComponent: NavigationServiceDelegate/*, UIContentContainer*/ {}
```
1ec5: Did you intend to uncomment or remove this comment?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -339,6 +340,11 @@ open class NavigationViewController: UIViewController {
     
     var mapViewController: RouteMapViewController?
     
+    var navigationComponents: [NavigationComponent] {
+        guard let mvc = mapViewController else { return [] }
+        return [mvc]
```
1ec5: Nit: this can be written more succinctly as `return mapViewController ?? []`.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -641,12 +602,15 @@ extension NavigationViewController: NavigationServiceDelegate {
...
-    @objc public func navigationService(_ service: NavigationService, didRerouteAlong route: Route, at: CLLocation?, proactive: Bool) {
-        mapViewController?.notifyDidReroute(route: route)
+    @objc public func navigationService(_ service: NavigationService, didRerouteAlong route: Route, at location: CLLocation?, proactive: Bool) {
+        navigationComponents.forEach { $0.navigationService?(service, didRerouteAlong: route, at: location, proactive: proactive) }
+
         delegate?.navigationViewController?(self, didRerouteAlong: route)
     }
     
     @objc public func navigationService(_ service: NavigationService, didFailToRerouteWith error: Error) {
+        navigationComponents.forEach { $0.navigationService?(service, didFailToRerouteWith: error) }
```
1ec5: Functionally equivalent but clearer:

```suggestion
        for component in navigationComponents {
            component.navigationService?(service, didFailToRerouteWith: error)
        }
```

MapboxNavigation/NavigationViewController.swift
```diff
@@ -641,12 +602,15 @@ extension NavigationViewController: NavigationServiceDelegate {
         delegate?.navigationViewController?(self, willRerouteFrom: location)
     }
     
-    @objc public func navigationService(_ service: NavigationService, didRerouteAlong route: Route, at: CLLocation?, proactive: Bool) {
-        mapViewController?.notifyDidReroute(route: route)
+    @objc public func navigationService(_ service: NavigationService, didRerouteAlong route: Route, at location: CLLocation?, proactive: Bool) {
+        navigationComponents.forEach { $0.navigationService?(service, didRerouteAlong: route, at: location, proactive: proactive) }
```
1ec5:
```suggestion
        for component in navigationComponents {
            component.navigationService?(service, didRerouteAlong: route, at: location, proactive: proactive)
        }
```

MapboxNavigation/NavigationViewController.swift
JThramer: It actually can't: 

image

Also, this gets overwritten in the subsequent PR.

1ec5: Ah, I missed that it needed to return `mapViewController` in an array.

PR #1914, opened 2018-12-22
Update MapboxVoiceController.swift

hastiranjkesh:
When we change the value of the voiceMuted variable to true, we also need to call unDuckAudio()


1ec5:
Closing in favor of #1915.

PR #1915, opened 2018-12-22
Update MapboxVoiceController.swift

hastiranjkesh:
In two below scenarios:
1 - Music in another application is playing -> In driving mode when we change the value of the voiceMuted variable to true (while the audio player is playing), we need to call unDuckAudio()
2 - Music in another application is playing -> When we close the navigation controller (driving mode) by tapping on the cancel button, deinit method of MapboxVoiceController will be called, after that, we need to call unDuckAudio()

### DIFF
--- a/MapboxNavigation/MapboxVoiceController.swift
+++ b/MapboxNavigation/MapboxVoiceController.swift
@@ -55,7 +55,7 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
                 do {
                     try strongSelf.unDuckAudio()
                 } catch {
-                    strongSelf.voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
+                    strongSelf.voiceControllerDelegate?.voiceController?(strongSelf, spokenInstructionsDidFailWith: error)
                 }
             }
         }
--- a/MapboxNavigation/MapboxVoiceController.swift
+++ b/MapboxNavigation/MapboxVoiceController.swift
@@ -51,11 +51,11 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
             if settings.voiceMuted {
                 self?.audioPlayer?.stop()
              
-                guard let self = self else { return }
+                guard let strongSelf = self else { return }
                 do {
-                    try self.unDuckAudio()
+                    try strongSelf.unDuckAudio()
                 } catch {
-                    self.voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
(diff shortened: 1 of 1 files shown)


MapboxNavigation/MapboxVoiceController.swift
```diff
@@ -50,12 +50,24 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
         muteToken = NavigationSettings.shared.observe(\.voiceMuted) { [weak self] (settings, change) in
             if settings.voiceMuted {
                 self?.audioPlayer?.stop()
+             
+                guard let self = self else { return }
```
vincethecoder: @hastiranjkesh could we refactor this line ... to say, `guard let strongSelf = self else { return }`

MapboxNavigation/MapboxVoiceController.swift
```diff
@@ -50,12 +50,24 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
...
+                do {
+                    try self.unDuckAudio()
```
vincethecoder: refactor this line to `try strongSelf?.unDuckAudio `

MapboxNavigation/MapboxVoiceController.swift
```diff
@@ -50,12 +50,24 @@ open class MapboxVoiceController: RouteVoiceController, AVAudioPlayerDelegate {
...
+                } catch {
+                    self.voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
```
vincethecoder: This could also be updated to:
`strongSelf.voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)`

1ec5:
The builds are failing because of:

```
/Users/distiller/project/MapboxNavigation/MapboxVoiceController.swift:58:74: error: value of optional type 'MapboxVoiceController?' must be unwrapped to a value of type 'MapboxVoiceController'
                    strongSelf.voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
                                                                         ^
/Users/distiller/project/MapboxNavigation/MapboxVoiceController.swift:58:74: note: coalesce using '??' to provide a default when the optional value contains 'nil'
                    strongSelf.voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
                                                                         ^
                                                                              ?? <#default value#>
/Users/distiller/project/MapboxNavigation/MapboxVoiceController.swift:58:74: note: force-unwrap using '!' to abort execution if the optional value contains 'nil'
                    strongSelf.voiceControllerDelegate?.voiceController?(self, spokenInstructionsDidFailWith: error)
                                                                         ^
                                                                             !
```

Issue #1929, opened 2019-01-12
Do not ask the delegate `shouldRerouteFrom` if already working to obtain a new route following an off-route event


akitchen:
Inside of `RouteController`.`locationManager(_:didUpdateLocations:)` we continue to process GPS updates while a re-route is in progress, potentially continuing to send 'shouldRerouteFrom` messages to the delegate even though we are in the process of obtaining a new route.

Just as we do inside of `reroute(from:along:)`, while rerouting, we should avoid sending this delegate message, and we should probably avoid other processing related to off-route determination since we are already in the process of obtaining a new route.

(related to conversation in #1777 )

**Mapbox Navigation SDK version:**

<= v0.27.0

1ec5:
Fixed in #1930.

PR #1930, opened 2019-01-14
Silence reroute questioning if reroute is already inflight

JThramer:
Quick fix. Implements #1929. 

/cc @mapbox/navigation-ios


JThramer:
@akitchen I generally try and use commas if there are exclusively `AND` statements in the clause ("This, that, and the other"), and use explicit boolean notation if the clause is more complex. ("This and that, but not those or this other thing.")

PR #1933, opened 2019-01-15
Updated configuration in audio ducking for CarPlay.

### LINKED ISSUE #1864: Update audio ducking settings
We’ve gotten a report of missing audio for spoken instructions in turn-by-turn navigation in CarPlay. While we haven’t been able to reproduce the issue, it looks we’re ducking and mixing audio in an outdated way.

We’re setting the AVAudioSession mode to `AVAudioSessionModeSpokenAudio`, but the [CarPlay programming guide](https://developer.apple.com/carplay/documentation/CarPlay-Navigation-App-Programming-Guide.pdf#page=33) recommends `AVAudioSessionModeVoicePrompt`.

https://github.com/mapbox/mapbox-navigation-ios/blob/efcc735fad36de81688f8b9c4df309ed797d6d6d/MapboxNavigation/RouteVoiceController.swift#L163

Additionally, the guide recommends `AVAudioSessionCategoryOptionInterruptSpokenAudioAndMixWithOthers` and `AVAudioSessionCategoryOptionDuckOthers`, whereas we set `AVAudioSessionCategoryAmbient` when mixing audio:

https://github.com/mapbox/mapbox-navigation-ios/blob/efcc735fad36de81688f8b9c4df309ed797d6d6d/MapboxNavigation/RouteVoiceController.swift#L169

Some of these settings are only available in iOS 12, and it may be necessary to change them only for CarPlay, even though both NavigationViewController and CarPlayNavigationViewController use the same RouteController class.

/cc @mapbox/navigation-ios

vincethecoder:
Changes for audio configuration when playing prompts for upcoming maneuvers when device is connected to a CarPlay head unit.

Fixes #1864.

/cc @mapbox/navigation-ios


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -159,14 +161,24 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     }
     
     func duckAudio() throws {
-        let categoryOptions: AVAudioSessionCategoryOptions = [.duckOthers, .interruptSpokenAudioAndMixWithOthers]
-        try AVAudioSession.sharedInstance().setMode(AVAudioSessionModeSpokenAudio)
-        try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, with: categoryOptions)
+        if #available(iOS 12.0, *) {
```
1ec5: Looks like the Xcode 9 build bots are failing, since this is the first availability check for iOS 12 that we’ve added. It’s probably time to jettison the Xcode 9 build bots in favor of Xcode 10, given that we aren’t far from the next version of Xcode.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -159,14 +161,24 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
+            let categoryMode = CarPlayManager.isConnected ? AVAudioSessionModeVoicePrompt : AVAudioSessionModeSpokenAudio
+            try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, mode: categoryMode, options: categoryOptions)
+        } else if #available(iOS 10.0, *) {
+            try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, mode: AVAudioSessionModeSpokenAudio, options: categoryOptions)
```
1ec5: `setCategory(_:mode:options:)` is convenient, but it’s OK to set the mode and category separately as long as you call both setters before activating the audio session.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -82,6 +82,8 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     var volumeToken: NSKeyValueObservation?
     var muteToken: NSKeyValueObservation?
     
+    fileprivate let categoryOptions: AVAudioSessionCategoryOptions = [.duckOthers, .interruptSpokenAudioAndMixWithOthers]
```
1ec5: Note that this is a read-only instance property, not merely a static constant. Consider turning this symbol into either a static constant or moving it into the file scope.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -159,14 +161,24 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     }
     
     func duckAudio() throws {
-        let categoryOptions: AVAudioSessionCategoryOptions = [.duckOthers, .interruptSpokenAudioAndMixWithOthers]
-        try AVAudioSession.sharedInstance().setMode(AVAudioSessionModeSpokenAudio)
-        try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, with: categoryOptions)
+        if #available(iOS 12.0, *) {
+            let categoryMode = CarPlayManager.isConnected ? AVAudioSessionModeVoicePrompt : AVAudioSessionModeSpokenAudio
```
1ec5: `AVAudioSessionModeVoicePrompt` is part of AVFoundation, not the CarPlay framework, so it doesn’t appear to be limited to CarPlay use cases. I wonder if we should always use `AVAudioSessionModeVoicePrompt` on iOS 12 and above.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -159,14 +161,24 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
+            try AVAudioSession.sharedInstance().setMode(AVAudioSessionModeSpokenAudio)
+            try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, with: categoryOptions)
+        }
         try AVAudioSession.sharedInstance().setActive(true)
     }
     
     func mixAudio() throws {
-        try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryAmbient)
+        if #available(iOS 12.0, *), CarPlayManager.isConnected {
+            try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, with: categoryOptions)
```
1ec5: Both `duckAudio()` and `mixAudio()` specify `AVAudioSessionCategoryOptionInterruptSpokenAudioAndMixWithOthers` and `AVAudioSessionCategoryOptionDuckOthers`. However, the names of these methods suggest that maybe one should only duck and the other should only mix. Would that be correct?

MapboxNavigation/RouteVoiceController.swift
vincethecoder: Agree.

MapboxNavigation/RouteVoiceController.swift
akitchen: Me too. I'd also like to eliminate any direct knowledge of CarPlay in this class, if we can.

MapboxNavigation/RouteVoiceController.swift
akitchen: again, is this really specific to CarPlay? We want to avoid creating a coupling here if at all possible.

MapboxNavigation/RouteVoiceController.swift
vincethecoder: @1ec5 I might have been mistaken, the original ticket highlights:
> We’re setting the AVAudioSession mode to AVAudioSessionModeSpokenAudio, but the [CarPlay programming guide](https://developer.apple.com/carplay/documentation/CarPlay-Navigation-App-Programming-Guide.pdf#page=33) recommends AVAudioSessionModeVoicePrompt.

- Which infers we wanted to make these changes only when connected to `CarPlay`, ... correct?

- From what I gathered also, the current `AVAudioSessionModeSpokenAudio` mode will pause your audio when another app plays a short amount of spoken audio. I don't believe that's a desired behavior, right? 

Please cofirm and I will make the necessary changes before going out for a field test.

MapboxNavigation/RouteVoiceController.swift
vincethecoder: That's correct.

MapboxNavigation/RouteVoiceController.swift
vincethecoder: Well, it certainly smashes the LOC in this context. Done ✅

MapboxNavigation/RouteVoiceController.swift
1ec5: CarPlay is our forcing function, and the CarPlay Programming Guide is how we found out about this option, but I guess my question is whether `AVAudioSessionModeVoicePrompt` is a refinement of `AVAudioSessionModeSpokenAudio` in general.

The official documentation set is light on details about `AVAudioSessionMode`, but AVFAudio/AVAudioSession.h has a ton of details in (non-documentation) comments.

1ec5: `AVAudioSessionModeSpokenAudio`:

> Appropriate for applications which play spoken audio and wish to be paused (via audio session interruption) rather than ducked if another app (such as a navigation app) plays a spoken audio prompt.  Examples of apps that would use this are podcast players and audio books.  For more information, see the related category option AVAudioSessionCategoryOptionInterruptSpokenAudioAndMixWithOthers.

`AVAudioSessionModeVoicePrompt`:

> Appropriate for applications which play audio using text to speech. Setting this mode allows for different routing behaviors when connected to certain audio devices such as CarPlay. An example of an app that would use this mode is a turn by turn navigation app that plays short prompts to the user. Typically, these same types of applications would also configure their session to use AVAudioSessionCategoryOptionDuckOthers and AVAudioSessionCategoryOptionInterruptSpokenAudioAndMixWithOthers

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -159,9 +159,12 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
     
     func duckAudio() throws {
-        let categoryOptions: AVAudioSessionCategoryOptions = [.duckOthers, .interruptSpokenAudioAndMixWithOthers]
-        try AVAudioSession.sharedInstance().setMode(AVAudioSessionModeSpokenAudio)
-        try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, with: categoryOptions)
+        if #available(iOS 12.0, *) {
+            try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, mode: AVAudioSessionModeVoicePrompt, options: [.duckOthers])
+        } else {
+            try AVAudioSession.sharedInstance().setMode(AVAudioSessionModeSpokenAudio)
+            try AVAudioSession.sharedInstance().setCategory(AVAudioSessionCategoryPlayback, with: [.duckOthers])
```
1ec5: Sorry for the back-and-forth: from https://github.com/mapbox/mapbox-navigation-ios/pull/1933#discussion_r248424392, I gather that `[.duckOthers, .interruptSpokenAudioAndMixWithOthers]` is correct in conjunction with either `AVAudioSessionModeSpokenAudio` or `AVAudioSessionModeVoicePrompt`, since we’re using this audio session for short text-to-speech prompts as described in the `AVAudioSessionModeVoicePrompt` documentation, as opposed to for long-form audiobook reading as described in the `AVAudioSessionModeSpokenAudio` documentation.

vincethecoder: @1ec5 resolved in the recent commit ✅

PR #1936, opened 2019-01-18
Upgrade the build bot to Xcode 10

vincethecoder:
- Upgraded the build bot to Xcode 10 and got rid of Xcode 9.

/cc @mapbox/navigation-ios

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,7 +5,6 @@ workflows:
   default:
     jobs:
       - xcode-10
-      - xcode-9
       - xcode-10-bench
       - xcode-10-cocoapods-install
       - xcode-10-cocoapods-update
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -110,7 +92,7 @@ step-library:
       run:
         name: Build Example
         command: |
-          xcodebuild -sdk iphonesimulator -destination 'platform=iOS Simulator,OS=11.4,name=iPhone 6 Plus' -project MapboxNavigation.xcodeproj -scheme Example clean build | xcpretty
+          xcodebuild -sdk iphonesimulator -destination 'platform=iOS Simulator,OS=12.1,name=iPhone 6 Plus' -project MapboxNavigation.xcodeproj -scheme Example clean build | xcpretty
```
1ec5: Should we also upgrade the device model here, or do we want to keep testing on lower-end simulators just in case?

vincethecoder: I think it makes sense to upgrade. However, most preferably, let's keep it on 6 till we start experiencing problems. Or what do you think?

frederoni: I'm up for dropping Xcode 9 support but we have to test on the minimum deployment target including the base target. You can use [xcode-install](https://github.com/KrauseFx/xcode-install) to install an older simulator. I guess we should run tests, or at least build the example app for iOS 9 and 12.

CHANGELOG.md
```diff
@@ -2,13 +2,15 @@
 
 ## master
 
+* Added the `CarPlayManagerDelegate.carPlayManager(_:selectedPreviewFor:using:)` method for any additional customization after a trip is selected on CarPlay. ([#1846](https://github.com/mapbox/mapbox-navigation-ios/pull/1846))
```
frederoni: nit: this was probably added accidentally with the merge conflict

PR #1938, opened 2019-01-21
Retain Cycle - RouteMapViewController

hactar:
`RouteMapViewController` creates a `Timer` (`updateETATimer`) to update the ETA. This timer retains a reference to the `RouteMapViewController`. Depending on the timing of when this controller is released, a call to `resetETATimer` can come in after the timer has been removed, which readds the timer, and re-retains the `RouteMapViewController` - because of this `deinit` is never called and the RouteMapViewController remains in memory.

This PR adds a check to see if the RouteMapViewController has been dismissed - in which case it does not allow the creation of a new updateETATimer.

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -583,7 +583,7 @@ extension RouteMapViewController {
 extension RouteMapViewController: NavigationViewDelegate {
     // MARK: NavigationViewDelegate
     func navigationView(_ view: NavigationView, didTapCancelButton: CancelButton) {
-        self?.dismissed = true
+        self.dismissed = true
         delegate?.mapViewControllerDidDismiss(self, byCanceling: true)
     }
     
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -157,6 +157,7 @@ class RouteMapViewController: UIViewController {
     }
 
     deinit {
+        self.dismissed = true
         suspendNotifications()
         removeTimer()
     }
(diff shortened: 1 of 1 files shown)


akitchen:
thank you for the PR!

Note that https://github.com/mapbox/mapbox-navigation-ios/pull/1912 removes the timer from this class, and reintroduces similar functionality via a `DispatchTimer` implementation. See [this comment thread](https://github.com/mapbox/mapbox-navigation-ios/pull/1912#pullrequestreview-192861674) for more info.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -580,6 +583,7 @@ extension RouteMapViewController {
 extension RouteMapViewController: NavigationViewDelegate {
     // MARK: NavigationViewDelegate
     func navigationView(_ view: NavigationView, didTapCancelButton: CancelButton) {
+        self.dismissed = true
```
vincethecoder: @hactar Should we also set `dismissed` to `true` in deinit as well?

hactar: Interesting point @vincethecoder, thanks. I don't see a normal scenario that would lead to deinit being called without self.dismissed being set to true already.

That said, there is no harm in setting it to true there too, and that might protect against scenarios where people are doing weird things with the RouteMapViewController.

I'll update the PR accordingly, tough if I understand @akitchen 's comment correctly this code will be obsolete in future anyway.

andrlee:
@JThramer is this a dup?

JThramer:
@hactar 

As @akitchen pointed out, the `updateETATimer` no longer exists in the `RouteMapViewController` as of v0.35. Closing as stale.

PR #1943, opened 2019-01-23
Mapbox CarPlay Framework

vincethecoder:
Separation of `CarPlay` features/functionality into an independent framework.

TODO
- [x] Run `agvtool new-marketing-version 0.29.0`
- [x] CHANGELOG.md
- [x] README.md with updated Carthage installation instructions that mention MapboxCarPlay.framework
- [x] docs/cover.md with updated Carthage installation instructions and updates to the “Starting points” section
- [x] Resolve merge conflicts
- [x] Generate localized files for CarPlay framework
- [ ] Framework validation PoC

/cc @mapbox/navigation-ios


MapboxNavigation/NavigationViewController.swift
```diff
@@ -46,7 +43,15 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
...
     optional func navigationViewController(_ navigationViewController: NavigationViewController, didArriveAt waypoint: Waypoint) -> Bool
-
+    
+    /**
+     Called when the user completes the final leg on the route.
+     
+     You can use this to detect the status of `CarPlayManager.isConnected`.
+     
+     */
+    @objc optional func navigationViewControllerDidFinishRoute(_ navigationViewController: NavigationViewController, byArrivingAtFinalStep isFinalLeg: Bool) -> Bool
```
frederoni: Can we try to make this self-explanatory? We already have didArrive so adding didFinishRoute is a bit confusing.
How about `…shouldDismissFeedback…` that defaults to true if unimplemented?

frederoni:
@vincethecoder I think we should synchronize the versions of MapboxCoreNavigation, MapboxNavigation, and MapboxCarPlay by running `agvtool new-marketing-version 0.28.0`? cc @1ec5

MapboxNavigation/NavigationMapView.swift
```diff
@@ -12,11 +12,11 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
     
-    struct FrameIntervalOptions {
-        static let durationUntilNextManeuver: TimeInterval = 7
-        static let durationSincePreviousManeuver: TimeInterval = 3
-        static let defaultFramesPerSecond = MGLMapViewPreferredFramesPerSecond.maximum
-        static let pluggedInFramesPerSecond = MGLMapViewPreferredFramesPerSecond.lowPower
+    public struct FrameIntervalOptions {
+        public static let durationUntilNextManeuver: TimeInterval = 7
+        public static let durationSincePreviousManeuver: TimeInterval = 3
+        public static let defaultFramesPerSecond = MGLMapViewPreferredFramesPerSecond.maximum
```
1ec5: Of these enumeration values, only `defaultFramesPerSecond` is used in CarPlay code. Let’s extract `defaultFramesPerSecond` out of this struct into a standalone constant and keep this struct internal.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -56,6 +56,16 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      */
     public weak var courseTrackingDelegate: NavigationMapViewCourseTrackingDelegate?
     
+    /**
+     The possible routes that can be represented on the map view.
+     */
+    public var routes: [Route]?
```
1ec5: I’m pretty uneasy about making this property public, because it’s essentially an implementation detail of `showRoutes(_:legIndex:)`, not something to be tampered with arbitrarily. If we can’t remove the dependency on this property from `CarPlayMapViewController.viewSafeAreaInsetsDidChange()`, then at least remove the documentation comment and use `:nodoc:` to hide the symbol from jazzy.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -56,6 +56,16 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    
+    /**
+     The user location the map view should follow.
+     */
+    @objc public var userLocationForCourseTracking: CLLocation?
```
1ec5: Same deal: we should avoid making this property public, or at least avoid making it discoverable.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -46,7 +43,18 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
...
+    /**
+     Called when the user arrives at the final destination of the current route leg and may display a feedback
+     UI to the user.
+     
+     If implemented, you can use this to detect the status of `CarPlayManager.isConnected` when the user arrives at the final destination.
+   
+     - parameter navigationViewController: The navigation view controller that has arrived at the leg of route.
+     - returns: This will most likely be used to determine whether to show feedback when the user arrives at the final leg of the route.
+     */
+    @objc optional func navigationViewControllerShouldShowFeedback(_ navigationViewController: NavigationViewController) -> Bool
```
1ec5: This seems redundant to `NavigationViewController.showsEndOfRouteFeedback`.

frederoni: `NavigationViewController.showsEndOfRouteFeedback` was replaced by this delegate.
Previously, `showsReportFeedback` was coupled with `showsEndOfRouteFeedback` so you couldn't pick one _or_ the other.

<details>

```swift
@objc public var showsReportFeedback: Bool = true {
        didSet {
            mapViewController?.reportButton.isHidden = !showsReportFeedback
            showsEndOfRouteFeedback = showsReportFeedback
        }
    }
```

</details>

`showsReportFeedback` simply toggles the visibility of the "realtime" feedback button.
Since CarPlay can be connected and disconnected throughout the session, we need to check that with the delegate.

However, this is a good indicator that the naming here isn't clear enough.
`s/navigationViewControllerShouldShowFeedback/navigationViewControllerShouldShowEndOfRouteFeedback`

MapboxCarPlay/Extensions/Locale+CarPlay.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+
+extension Locale {
+    
+    // Returns Locale.autoupdatingCurrent unless its languageCode is equal to `en`, then nil will be returned.
+    static var autoupdatingNonEnglish: Locale? {
```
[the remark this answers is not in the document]
vincethecoder: Sounds good!

MapboxNavigation/Bundle.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 
 extension Bundle {
     
-    class var mapboxNavigation: Bundle {
+    public class var mapboxNavigation: Bundle {
```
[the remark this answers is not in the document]
vincethecoder: Done ✅

1ec5:
Stale – CarPlay integration will remain a part of the navigation SDK for now, but we may reconsider in the future as part of a more comprehensive modularization strategy.

PR #1951, opened 2019-01-29
Bottom Banner Containerization

JThramer:
Get in loser, we're containerizing the UI.

For-realsies, no-joke bottom-banner containerization. Top banner should be an easy fast-follow if desired. 

Successor to #1912.

To-Do:
- [x] Hook up messaging
- [x] Hook up cancel action delegate
- [x] ~Add handling for `preferredContentSize` change~ (Apparently we get this for free)
- [x] Fix issue where UIAppearance won't set color of bottom banner root view
- [x] Replace myriad of init arguments on NVC with `NavigationOptions`
- [x] Tests, probably.

/cc @mapbox/navigation-ios


Example/MBViewController.m
```diff
@@ -114,7 +114,7 @@ - (void)startNavigation:(MBRoute *)route {
     MBNavigationService *service = [[MBNavigationService alloc] initWithRoute:route directions:self.directions locationSource:nil eventsManagerType:nil simulating:MBNavigationSimulationOptionsAlways routerType:nil];
     
     self.navigation = service;
-    MBNavigationViewController *controller = [[MBNavigationViewController alloc] initWithRoute:route styles:nil navigationService:service voiceController: nil];
+    MBNavigationViewController *controller = [[MBNavigationViewController alloc] initWithRoute:route styles:nil navigationService:service voiceController: nil bottomBanner:nil];
```
vincethecoder: Should we consider making this new param more generic? Possibly an array of banners? (...considering banner could be a top or bottom banner.)

MapboxNavigation/NavigationOptions.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+import MapboxCoreNavigation
+
+/*
+    Object representing available options for a Navigation Session.
+ */
+
+@objc(MBNavigationOptions)
+open class NavigationOptions: NSObject {
```
akitchen: 👍

MapboxNavigation/NavigationViewController.swift
```diff
@@ -8,195 +8,10 @@ import CarPlay
 #endif
 
 /**
- The `NavigationViewControllerDelegate` protocol provides methods for configuring the map view shown by a `NavigationViewController` and responding to the cancellation of a navigation session.
+ A ContainerViewController is any UIViewController that suscribes to the NavigationComponent messaging protocol.
```
akitchen: Typo here, no biggie

MapboxNavigation/DayStyle.swift
```diff
@@ -63,7 +63,7 @@ open class DayStyle: Style {
         
         ArrivalTimeLabel.appearance().font = UIFont.systemFont(ofSize: 18, weight: .medium).adjustedFont
         ArrivalTimeLabel.appearance().normalTextColor = .defaultPrimaryText
-        BottomBannerContentView.appearance().backgroundColor = #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
+//        BottomBannerContentView.appearance().backgroundColor = #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1)
```
akitchen: Dead code or hanging chad?

MapboxNavigation/NavigationView.swift
```diff
@@ -69,7 +69,7 @@ open class NavigationView: UIView {
     }
     
     private enum Actions {
-        static let cancelButton: Selector = #selector(NavigationView.cancelButtonTapped(_:))
+//        static let cancelButton: Selector = #selector(NavigationView.cancelButtonTapped(_:))
```
akitchen: Seems like this class has some placeholder/WIP stuff going on (methods below as well).

MapboxNavigation/NavigationViewLayout.swift
```diff
@@ -25,19 +25,19 @@ extension NavigationView {
...
-        
-        bottomBannerView.trailingAnchor.constraint(equalTo: trailingAnchor).isActive = true
-        bottomBannerView.leadingAnchor.constraint(equalTo: leadingAnchor).isActive = true
-        bottomBannerView.bottomAnchor.constraint(equalTo: safeBottomAnchor).isActive = true
+        resumeButton.bottomAnchor.constraint(equalTo: bottomBannerContainerView.topAnchor, constant: -10).isActive = true
+
+        bottomBannerContainerView.trailingAnchor.constraint(equalTo: trailingAnchor).isActive = true
+        bottomBannerContainerView.leadingAnchor.constraint(equalTo: leadingAnchor).isActive = true
+        bottomBannerContainerView.bottomAnchor.constraint(equalTo: bottomAnchor).isActive = true
+//        bottomBannerContentView.topAnchor.constraint(equalTo: bottomBannerView.topAnchor).isActive = true
```
akitchen: More dead code?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -206,7 +206,7 @@ class RouteMapViewController: UIViewController {
     var annotatesSpokenInstructions = false
 
     var overheadInsets: UIEdgeInsets {
-        return UIEdgeInsets(top: navigationView.instructionsBannerView.bounds.height, left: 20, bottom: bottomBannerView.bounds.height, right: 20)
+        return UIEdgeInsets(top: navigationView.instructionsBannerView.bounds.height, left: 20, bottom: 0/*bottomBannerView.bounds.height*/, right: 20)
```
akitchen: Cleanup?

MapboxNavigation/NavigationOptions.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+import MapboxCoreNavigation
+
+/*
+    Object representing available options for a Navigation Session.
```
1ec5: Can you expand on this to avoid confusion with the similarly named NavigationRouteOptions class? Maybe we also need some disambiguation in the NavigationRouteOptions documentation too.

Example/MBViewController.m
```diff
@@ -114,7 +114,9 @@ - (void)startNavigation:(MBRoute *)route {
     MBNavigationService *service = [[MBNavigationService alloc] initWithRoute:route directions:self.directions locationSource:nil eventsManagerType:nil simulating:MBNavigationSimulationOptionsAlways routerType:nil];
     
     self.navigation = service;
-    MBNavigationViewController *controller = [[MBNavigationViewController alloc] initWithRoute:route styles:nil navigationService:service voiceController: nil];
+    MBNavigationOptions *options = [[MBNavigationOptions alloc] init];
+    options.navigationService = service;
```
1ec5: Setting just the `navigationService` property seems like a common enough scenario that maybe we should provide a dedicated initializer for it.

MapboxNavigation/NavigationOptions.swift
1ec5: If we define NavigationOptions in Objective-C as a C struct instead of a class, we get reference semantics in both languages without the mental overhead of allocation in Objective-C. (Note that [these days](https://developer.apple.com/videos/play/wwdc2018/409/) ARC supports object pointers in C structs.) On the other hand, C structs have a fixed size, so they aren’t as forward-compatible, especially as new options are added over time.

So it’s fine that we’ve defined it as a class here, but we should try to reduce the syntactic overhead in Objective-C, for example by adding a `+navigationOptions` initializer that takes no arguments. Additionally, history has shown that value classes like NavigationOptions quickly end up needing conformance to NSSecureCoding, object equality, and the like. These are all things we’d get for free with a struct, but we need to define them explicitly with a class (perhaps as tail work).

(The same considerations apply to RouteProgress et al., but reference semantics are much more useful in that case.)

MapboxNavigation/NavigationOptions.swift
```diff
@@ -0,0 +1,38 @@
...
+     The voice controller that manages the delivery of voice instructions during navigation.
+    */
+    @objc public var voiceController: RouteVoiceController?
+    
+    /*
+     The view controlelr to embed into the bottom section of the UI. Defaults to `BottomBannerViewController`.
+     */
+    @objc public var bottomBanner: ContainerViewController?
+    
+    @objc public convenience init(styles: [Style]? = nil, navigationService: NavigationService? = nil, voiceController: RouteVoiceController? = nil, bottomBanner: ContainerViewController? = nil) {
```
1ec5: As a developer, I would be a little surprised that CarPlayManager and CarPlayNavigationViewController initializers continue to take a long list of arguments instead of a NavigationOptions. There’s some interesting overlap but also some significant differences between the sets of options that each environment would require. It’s fine if we make the CarPlay classes adopt NavigationOptions later on, but any properties in common between the two environments should probably live in a protocol to start out with.

MapboxNavigation/NavigationOptions.swift
```diff
@@ -0,0 +1,38 @@
...
+     */
+    @objc open var navigationService: NavigationService?
+    
+    /*
+     The voice controller that manages the delivery of voice instructions during navigation.
+    */
+    @objc public var voiceController: RouteVoiceController?
+    
+    /*
+     The view controlelr to embed into the bottom section of the UI. Defaults to `BottomBannerViewController`.
```
1ec5:
```suggestion
     The view controller to embed into the bottom section of the UI. A `BottomBannerViewController` is embedded by default.
```

MapboxNavigation/NavigationViewController+Obsolete.swift
```diff
@@ -0,0 +1,67 @@
+import Foundation
+import MapboxCoreNavigation
+import MapboxDirections
+
+//MARK: - Obsoleted Interfaces
+
+public extension NavigationViewController {
+    @available(*, deprecated: 0.1, renamed: "init(route:options:)", message: "Use the new NavigationOptions initalizer.")
```
1ec5: Does the `renamed:` annotation result in a fix-it suggestion? Does it work correctly, considering that the new initializer takes a different set of arguments?

MapboxNavigation/NavigationViewController+Obsolete.swift
```diff
@@ -0,0 +1,67 @@
...
+                         voiceController: RouteVoiceController? = nil) {
+        let bridge = NavigationOptions()
+        bridge.styles = styles
+        bridge.navigationService = navigationService
+        bridge.voiceController = voiceController
+        
+        self.init(for: route, options: bridge)
+    }
+    
+    @available(*, obsoleted: 0.1, renamed: "navigationService", message: "NavigationViewController no-longer directly manages a RouteController. See MapboxNavigationService, which contains a protocol-bound reference to the RouteController, for more information.")
```
1ec5:
```suggestion
    @available(*, obsoleted: 0.1, renamed: "navigationService", message: "NavigationViewController no longer directly manages a RouteController. See MapboxNavigationService, which contains a protocol-bound reference to the RouteController, for more information.")
```

(×3)

MapboxNavigation/NavigationViewController.swift
```diff
@@ -8,195 +8,10 @@ import CarPlay
 #endif
 
 /**
- The `NavigationViewControllerDelegate` protocol provides methods for configuring the map view shown by a `NavigationViewController` and responding to the cancellation of a navigation session.
+ A ContainerViewController is any UIViewController that subscribes to the NavigationComponent messaging protocol.
```
1ec5: A class _subscribes_ to a notification but _conforms_ to a protocol, even if the protocol is used for messaging of some sort.

```suggestion
 A ContainerViewController is any UIViewController that conforms to the NavigationComponent messaging protocol.
```

MapboxNavigation/NavigationOptions.swift
```diff
@@ -0,0 +1,38 @@
...
+    Object representing available options for a Navigation Session.
+ */
+
+@objc(MBNavigationOptions)
+open class NavigationOptions: NSObject {
+    
+    /*
+     The styles that the view controller’s internal `StyleManager` object can select from for display. Defaults to `[DayStyle(), NightStyle()]`
+     */
+    @objc public var styles: [Style]? = nil
```
1ec5: Should this be `open` like the other properties?

MapboxNavigation/NavigationComponent.swift
```diff
@@ -2,4 +2,7 @@ import Foundation
 import MapboxCoreNavigation
 import CoreLocation
 
-@objc protocol NavigationComponent: NavigationServiceDelegate {}
+/*
+ A NavigationComponent is any component that can receive NavigationService messages.
```
1ec5: What is a “NavigationService message”, in terms that a developer who’s unaware of this library’s internals can understand?

MapboxNavigation/BottomBannerViewController.swift
```diff
@@ -62,11 +68,20 @@ open class BottomBannerView: UIView, NavigationComponent {
         removeTimer()
     }
     
-    override open func willMove(toSuperview newSuperview: UIView?) {
-        super.willMove(toSuperview: newSuperview)
-        if superview != nil, newSuperview == nil {
-            removeTimer()
-        }
+    override open func loadView() {
+        let root: BottomBannerView = .forAutoLayout() //Must use local var to prevent generic factory from messing up.
```
1ec5: Huh, interesting. Out of curiosity, how does it mess up?

MapboxNavigation/NavigationOptions.swift
1ec5: We should draw a line somewhere and omit nonessential arguments from this initializer in favor of the properties. It isn’t a goal to allow the developer to specify all the options in a single statement. `navigationService` and `styles` seem more commonly overridden than the others, but I would even be comfortable with having no convenience initializer (just the one factory initializer for Objective-C). The test I’d use is: if the developer isn’t specifically looking to customize some aspect of the experience, would forgetting its property lead to unexpected behavior, or would the default be intuitive?

MapboxNavigation/BottomBannerViewController.swift
frederoni: Somehow, type inference initializes a `UIView` if you do
```swift
let view = BottomBannerView.forAutoLayout()
```

This can be fixed by specifiying the generic type as an argument in the generic function.

Will be interesting to see how Swift 5 handles this.

1ec5: Hm, I guess that behavior sounds reasonable actually. I’d expect `.forAutoLayout<BottomBannerView>()` to be required in that case.

MapboxNavigation/NavigationOptions.swift
JThramer: The idea here is to have struct-like initialization semantics, even though the type is not a struct. Since every initialization argument is optional with a default value of `nil`, an omitted argument in the initializer (just like an unset property), means "use default behavior."

MapboxNavigation/BottomBannerViewController.swift
JThramer: I don't think action is needed here. Clarity at point-of-site is more important than brevity, and here it's very apparent that the root view is a `BottomBannerView`.

MapboxNavigation/NavigationViewController+Obsolete.swift
JThramer: Doh, I think it does. Removed.

MapboxNavigation/BottomBannerViewController.swift
```diff
@@ -2,27 +2,64 @@ import UIKit
...
+    /**
+     A method that is invoked when the user taps on the cancel button.
+     - parameter sender: The button that originated the tap event.
+     */
+    func didTapCancel(_ sender: Any)
 }
 
-/// :nodoc:
+/**
+ The BottomBannerUIController is a UI Element designed to display the ETA, Distance, and Time Remaining, as well as give the user a control the cancel the navigation session.
```
1ec5:
```suggestion
 A user interface element designed to display the estimated arrival time, distance, and time remaining, as well as give the user a control the cancel the navigation session.
```

MapboxNavigation/BottomBannerViewController.swift
```diff
@@ -2,27 +2,64 @@ import UIKit
...
 @IBDesignable
-@objc(MBBottomBannerView)
-open class BottomBannerView: UIView, NavigationComponent {
-    
-    weak var previousProgress: RouteProgress?
+@objc(MBBottomBannerViewController)
+open class BottomBannerViewController: UIViewController, NavigationComponent {
+    
+    /**
+     The Time Remaing label that displayes the estimated time until the user's arrival.
```
1ec5: A couple typos:

```suggestion
     The label that displays the estimated time until the user arrives at the final destination.
```

MapboxNavigation/BottomBannerViewController.swift
```diff
@@ -2,27 +2,64 @@ import UIKit
...
+    */
+    open var cancelButton: CancelButton!
+    
+    /**
+     A vertical divider that seperates the cancel button and informative labels.
+    */
+    open var verticalDividerView: SeparatorView!
+    
+    /**
+     A horizontal divider that adds visual seperation between the bottom banner and it's superview.
```
1ec5:
```suggestion
     A horizontal divider that adds visual separation between the bottom banner and its superview.
```

MapboxNavigation/BottomBannerViewController.swift
```diff
@@ -47,12 +84,32 @@ open class BottomBannerView: UIView, NavigationComponent {
             }
         }
     }
-    
-    override init(frame: CGRect) {
-        super.init(frame: frame)
+    /**
+     Initializes a `BottomBannerViewController` that provides ETA, Distance to arrival, and Time to arrival.
```
1ec5:
```suggestion
     Initializes a `BottomBannerViewController` that provides estimated arrival time, distance to arrival, and time to arrival.
```

(×3)

MapboxNavigation/NavigationCustomizable.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+import MapboxCoreNavigation
+
+/**
+ The `NavigationCustomizable` protocol represents a UI-based mechanism that allows for customization of its visual style, as well as the navigation service that powers it.
+ */
+protocol NavigationCustomizable {
+    /**
+     The styles that the User Interface can select from for display.
```
1ec5: Nit: UI is capitalized, but user interface is not, since it’s a common noun.

```suggestion
     The styles that the user interface can select from for display.
```

1ec5:
Added the <kbd>backwards-incompatible</kbd> label, because this PR broke backwards compatibility in a few ways, some of which were never documented:

* BottomBannerContentView went away. As far as I can tell, if a Style subclass had customized `BottomBannerContentView.backgroundColor`, that code can simply go away because it would be covered by BottomBannerView, correct?
* As part of replacing LegacyRouteController (né RouteController) with RouteController (née PortableRouteController), the `reroutesProactively` option was effectively removed. This is a useful option, one that #1902 would’ve enabled by default. Developers are now forced to choose between MapboxNavigationNative-based route following functionality and the ability to navigate users around flashmobs, so to speak.

PR #1954, opened 2019-02-01
Starting navigation sends route retrieval performance event

1ec5:
Send a route retrieval performance event, indicating how long a round-trip to the Directions API takes, when starting a navigation service (i.e., when starting navigation or getting rerouted). Instead of a single EventDetails class, now there are NavigationEventDetails and PerformanceEventDetails classes that share an EventDetails protocol.

This is a counterpart to mapbox/mapbox-navigation-android#1661 on Android that depends on mapbox/mapbox-events-ios#93 and mapbox/MapboxDirections.swift#335.

/cc @mapbox/navigation-ios @devotaaabel @rclee


Cartfile
```diff
@@ -1,7 +1,7 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 5.0.0
-github "mapbox/MapboxDirections.swift" ~> 0.26.0
+github "mapbox/MapboxDirections.swift" "master"
```
1ec5: Ideally we’ll release MapboxDirections.swift v0.26.2 shortly before landing this PR.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -167,9 +167,13 @@ public class NavigationDirections: Directions {
                         return completionHandler(nil, nil, error as NSError)
                     }
                 } else {
-                    
                     DispatchQueue.main.async {
                         let response = options.response(from: json)
+                        let responseEndDate = Date()
+                        for route in response.1 ?? [] {
```
1ec5: This is just to demonstrate how we’d go about collecting analogous statistics about client-side route calculation if that were to become relevant in the future.

MapboxCoreNavigation/EventDetails.swift
```diff
@@ -5,8 +5,44 @@ import UIKit
...
+    var created: Date { get }
+    var sessionIdentifier: String { get }
+}
+
+struct PerformanceEventDetails: EventDetails {
+    let event: String?
+    let created: Date
+    let sessionIdentifier: String
+    var counters: [String: Double] = [:]
+    var attributes: [String: String] = [:]
```
1ec5: We made a late-breaking decision to store counters and attributes as flat dictionaries instead of as arrays of key-value dictionaries.

1ec5: Never mind, we decided against storing the counters and attributes as flat dictionaries.

MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -107,18 +109,34 @@ open class NavigationEventsManager: NSObject {
...
+    func navigationRouteRetrievalEvent() -> PerformanceEventDetails? {
+        guard let sessionState = sessionState,
+            let responseEndDate = sessionState.currentRoute.responseEndDate,
+            let fetchStartDate = sessionState.currentRoute.fetchStartDate else {
+            return nil
+        }
+
+        var event = PerformanceEventDetails(event: NavigationEventTypeRouteRetrieval, session: sessionState, createdOn: sessionState.currentRoute.responseEndDate)
+        event.counters.append(PerformanceEventDetails.Counter(name: "elapsed_time",
+                                                              value: responseEndDate.timeIntervalSince(fetchStartDate)))
```
1ec5: This is measured in seconds (as a floating-point number), but mapbox/mapbox-navigation-android#1661 implemented the field in nanoseconds as a long integer.

1ec5: mapbox/mapbox-navigation-android#1732

PR #1955, opened 2019-02-02
[CarPlay] enable developers to customize CPTrip and CPTripPreviewTextConfiguration before preview

akitchen:
This is still WIP; so far I've introduced a seam (MapTemplateProvider) that allows us to inject a testable map template class for spying on the results of behaviors which interact with the map template. I'm backfilling tests and implementing this in a way (TDD) which uses tests to drive out the necessary refactoring.

Opening this up early for the sake of discussion; feedback welcome on the overall approach

Next up:
 - [x] define test case for CPTrip customization
 - [x] introduce delegate method for customizing CPTrip before preview
 - [x] define test case for CPTripPreviewTextConfiguration customization
 - [x] introduce delegate method for providing a customized CPTripPreviewTextConfiguration
 - [x] discuss whether to make the MapTemplateProvider seam public (and whether we want this at all)

/cc @mapbox/navigation-ios


MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -6,8 +6,8 @@ import TestHelper
 
 let response = Fixture.JSONFromFileNamed(name: "routeWithInstructions")
 let jsonRoute = (response["routes"] as! [AnyObject]).first as! [String : Any]
-let waypoint1 = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 37.795042, longitude: -122.413165))
-let waypoint2 = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 37.7727, longitude: -122.433378))
+let waypoint1 = Fixture.waypoint1
```
frederoni: nit: There is a `Fixture.route(from:)` that will parse a route including waypoints, no hard-coded waypoints are needed.

vincethecoder: Resolved ✅

PR #1959, opened 2019-02-04
Correct Local Variable Typo

### LINKED ISSUE #1958: Mispelled Variable Name in RouteMapViewController
I was poking around the code day and came across the following type in a local variable convention:

https://github.com/mapbox/mapbox-navigation-ios/blob/187fd6ea886a183a0ee4c31df2ad078d4e082583/MapboxNavigation/RouteMapViewController.swift#L736

https://github.com/mapbox/mapbox-navigation-ios/blob/187fd6ea886a183a0ee4c31df2ad078d4e082583/MapboxNavigation/RouteMapViewController.swift#L758

`snappedLoction` vs. `snappedLocation`

As far as I can tell there's no impact to functionality because the variable is consistently misspelled, but would be worthwhile to fix for future readability and searchability.

Though, because it's an arg, this might _technically_ be a `semver` change.

riastrad:
Fixes #1958 

Minor change to two local uses of the `snappedLocation` variable in `RouteMapViewController`'s `labelCurrentRoad` function.

More of an aesthetic change than anything else, so feel free to reject if it poses too much of a headache for `semver` conversations.

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -733,7 +733,7 @@ extension RouteMapViewController: NavigationViewDelegate {
 
      - parameter location: The user’s current location.
      */
-    func labelCurrentRoad(at rawLocation: CLLocation, for snappedLoction: CLLocation? = nil) {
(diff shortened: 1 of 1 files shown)


riastrad:
cc: @mapbox/navigation-ios

PR #1962, opened 2019-02-05
[CarPlay] Adds custom map buttons delegate for activity navigating.

### LINKED ISSUE #1963: [CarPlay] Provide delegate to customize the trailing bottom map buttons.
**Mapbox Navigation SDK version:** 0.28.0

The current `CarPlayManagerDelegate` implements a delegate for customizing the map buttons per activity. Although the `CarPlayMapViewController` creates the `recenter`, `pan` and `zoom in/out` buttons on the map template, it's undiscoverable.

Hence, the `CarPlayManagerDelegate.CarPlayManagerDelegate.carPlayManager(_:mapButtonsCompatibleWith:in:for:)` should provide an additional parameter for a `CarPlayMapViewController`. 

When that method gets implemented, the developer can choose to use the zoom in/out and pan buttons declared in the `CarPlayMapViewController` or provide a collection of custom map buttons to display in the trailing bottom buttons for the respective `CarPlayActivity` classifications.

The proposed delegate should look like this:
```swift
CarPlayManagerDelegate.carPlayManager(_:mapButtonsCompatibleWith:in:for:
carPlayMapViewController:)
```

The implementation of the delegate should look like this:

```swift

func carPlayManager(_ carPlayManager: CarPlayManager, 
mapButtonsCompatibleWith traitCollection: UITraitCollection, 
in template: CPTemplate, for activity: CarPlayActivity, 
carPlayMapViewController: CarPlayMapViewController) -> [CPMapButton]? {
   ...
   case .browsing:
   var mapButtons = [carPlayMapViewController.recenterButton,
                     carPlayMapViewController.zoomInButton,
                     carPlayMapViewController.zoomOutButton]
            
            if let mapTemplate = template as? CPMapTemplate {
                mapButtons.insert(carPlayMapViewController.createPanMapButton(for: 
                mapTemplate), at: 1)
            }
   return mapButtons 
   ...
}
```
/cc @mapbox/navigation-ios

vincethecoder:
This allows developers to provide custom leading/trailing bar buttons and template buttons displayed on the map. Previously, developers were unable to customize the map template's trailing bottom buttons. This effort addresses that issue.

Resolves #1963 

TODO
- [x] Update Change log
- [x] Update Localization
- [x] ~Update~ Create Tests

/cc @mapbox/navigation-ios


akitchen:
> * Update Tests

Is this tested at all? Have a look at https://github.com/mapbox/mapbox-navigation-ios/pull/1955 and let's consider adopting a testing pattern for this.

Example/AppDelegate+CarPlay.swift
```diff
@@ -72,7 +72,16 @@ extension AppDelegate: CarPlayManagerDelegate {
             searchTemplate.delegate = carPlaySearchController
             let searchButton = carPlaySearchController.searchTemplateButton(searchTemplate: searchTemplate, interfaceController: interfaceController, traitCollection: traitCollection)
             return [searchButton]
-        case .navigating, .previewing:
+        case .navigating:
+            let muteTitle = "Mute"
```
frederoni: Why are these no longer localizable? You may have to the `bundle` parameter or move the localizable strings into the CarPlay framework.

vincethecoder: @frederoni strings are not localized in the `Example` app. This is not the `CarPlay framework` PR #1943 It is an unrelated pull request.

Example/AppDelegate+CarPlay.swift
```diff
@@ -99,6 +108,46 @@ extension AppDelegate: CarPlayManagerDelegate {
...
+    func carPlayManager(_ carPlayManager: CarPlayManager, mapButtonsCompatibleWith traitCollection: UITraitCollection, in template: CPTemplate, for activity: CarPlayActivity, carPlayMapViewController: CarPlayMapViewController) -> [CPMapButton]? {
+        
+        switch activity {
+        case .navigating:
+            let showFeedbackButton = CPMapButton { button in
+                carPlayManager.currentNavigator?.showFeedback()
+            }
+            showFeedbackButton.image = UIImage(named: "carplay_feedback", in: nil, compatibleWith: nil)
+            
+            let overviewButton = CPMapButton { button in
```
JThramer: Could we extract all of this other button creation into `createSuchAndSuchButton(for: dependency)` functions?

JThramer: Posterity: after talking with Kobe about this, it seems like punting is the right choice.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -457,48 +435,20 @@ extension CarPlayManager: CPMapTemplateDelegate {
...
 
         if let rootViewController = self.carWindow?.rootViewController as? CarPlayMapViewController,
             let leadingButtons = delegate?.carPlayManager?(self, leadingNavigationBarButtonsCompatibleWith: rootViewController.traitCollection, in: mapTemplate, for: .navigating) {
             mapTemplate.leadingNavigationBarButtons = leadingButtons
         }
         
-        let muteTitle = NSLocalizedString("CARPLAY_MUTE", bundle: .mapboxNavigation, value: "Mute", comment: "Title for mute button")
-        let unmuteTitle = NSLocalizedString("CARPLAY_UNMUTE", bundle: .mapboxNavigation, value: "Unmute", comment: "Title for unmute button")
-        
-        let muteButton = CPBarButton(type: .text) { (button: CPBarButton) in
```
1ec5: The mute/unmute button is useful out of the box; we don’t want to force developers to copy-pasta swaths of the example application to get such basic functionality. Let’s add factory methods for creating these buttons, which will result in restoring the localized strings.

MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -32,22 +32,54 @@ class CarPlayMapViewController: UIViewController {
...
-        recenterButton.image = UIImage(named: "carplay_locate", in: bundle, compatibleWith: traitCollection)
-        return recenterButton
+        zoomInOut.image = UIImage(named: "carplay_minus", in: bundle, compatibleWith: traitCollection)
+        return zoomInOut
     }()
     
+    /**
+     The map button property for hiding or showing the pan map button.
+     */
+    private(set) public var panMapButton: CPMapButton?
```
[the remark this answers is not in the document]
vincethecoder: @1ec5 So the main intention of this button property is to allow the developer to hide/show the map button. TBH, wouldn't it seem a bit odd to see: 
``` swift 
CarPlayMapViewController.panMapButton(for: mapTemplate).isHidden = true
```
as opposed to:
```swift
CarPlayMapViewController.panButton.isHidden = true
```
Which one would you prefer?

MapboxNavigation/CarPlayManager.swift
```diff
@@ -553,6 +503,16 @@ extension CarPlayManager: CPMapTemplateDelegate {
...
     }
     
+    public func mapTemplateDidShowPanningInterface(_ mapTemplate: CPMapTemplate) {
+        guard let carPlayMapViewController = carWindow?.rootViewController as? CarPlayMapViewController else {
+            return
+        }
+        
+        self.defaultMapButtons = mapTemplate.mapButtons
+        let closeButton = self.dismissPanButton(for: mapTemplate, traitCollection: carPlayMapViewController.traitCollection)
+        mapTemplate.mapButtons = [closeButton]
```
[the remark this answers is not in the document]
vincethecoder: @1ec5 Good catch! Thanks.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 * Added the `MapboxNavigationService.router(_:didPassVisualInstructionPoint:routeProgress:)` and `MapboxNavigationService.router(_:didPassSpokenInstructionPoint:routeProgress:)` methods, which correspond to `Notification.Name.routeControllerDidPassVisualInstructionPoint` and `Notification.Name.routeControllerDidPassSpokenInstructionPoint`, respectively. ([#1912](https://github.com/mapbox/mapbox-navigation-ios/pull/1912))
 * Added an initializer to `DispatchTimer`, along with methods for arming and disarming the timer. ([#1912](https://github.com/mapbox/mapbox-navigation-ios/pull/1912))
 
+### Navigation
+* Renamed `CarPlayManagerDelegate.carPlayManager(_:mapButtonsCompatibleWith:in:for:)` to `CarPlayManagerDelegate.carPlayManager(_:mapButtonsCompatibleWith:in:for:carPlayMapViewController:)`. You can now create and return map buttons that perform built-in actions on the map by calling methods such as `CarPlayMapViewController.zoomInButton()`. ([#1962](https://github.com/mapbox/mapbox-navigation-ios/pull/1962))
+
```
vincethecoder: @1ec5 Ideally, it makes more sense to create the map buttons using a `lazy computed property`. Lazy computed properties are only calculated once.

MapboxNavigation/CarPlayMapViewController.swift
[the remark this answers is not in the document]
1ec5: Where would `panMapButton(for:)` be called from? Will it always be called from a place where the CarPlayMapViewController object is easy to access?

MapboxNavigation/CarPlayManager.swift
```diff
@@ -16,6 +16,8 @@ public enum CarPlayActivity: Int {
     case previewing
     /// The user is actively navigating along a route.
     case navigating
+    /// The user is panning the map.
```
1ec5: Can the user pan the map while navigating or previewing, or only while browsing?

MapboxNavigation/CarPlayManager.swift
```diff
@@ -183,9 +183,9 @@ extension CarPlayManager: CPApplicationDelegate {
         }
     }
 
-    func mapTemplate(for interfaceController: CPInterfaceController, viewController: UIViewController) -> CPMapTemplate {
+    func mapTemplate(for interfaceController: CPInterfaceController, carPlayMapViewController: CarPlayMapViewController) -> CPMapTemplate {
```
1ec5: This method is only called in one place, so `carPlayMapViewController` will only ever be the view controller that we create in `CarPlayManager.application(_:didConnectCarInterfaceController:to:)` and set as the window’s root view controller. We could simplify a lot of this class by formalizing a `carPlayMapViewController` computed property that returns `carWindow?.rootViewController as? CarPlayMapViewController`. For example, the `carPlayMapViewController` argument here becomes unnecessary.

MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -102,35 +143,48 @@ class CarPlayMapViewController: UIViewController {
...
+                mapTemplate.showPanningInterface(animated: true)
             }
-            strongSelf.mapView.setZoomLevel(strongSelf.mapView.zoomLevel + 1, animated: true)
         }
+        
         let bundle = Bundle.mapboxNavigation
-        zoomInButton.image = UIImage(named: "carplay_plus", in: bundle, compatibleWith: traitCollection)
-        return zoomInButton
+        panButton.image = UIImage(named: "carplay_pan", in: bundle, compatibleWith: traitCollection)
+        self.panMapButton = panButton
```
1ec5: An instance method shouldn’t create an object, store it in a property as a side effect, and return that object. The previous lazily computed property was more appropriate if we’re sure that it’s better to store the result, but if we go with a factory method, then it should have no side effects. (×2)

MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -32,22 +36,59 @@ class CarPlayMapViewController: UIViewController {
...
+        }
+        let bundle = Bundle.mapboxNavigation
+        recenter.image = UIImage(named: "carplay_locate", in: bundle, compatibleWith: traitCollection)
+        return recenter
+    }()
+    
+    /**
+     The map button for zooming in the current map view.
+     */
+    @objc public lazy var zoomInButton: CPMapButton = {
```
1ec5: I’m still unsure about making these buttons into lazily computed properties, since that establishes a strict one-to-one correspondence between a button and a map view controller. I know it’s unlikely that the map would ever have multiple zoom-in buttons, for instance, but it seems like an unnecessary constraint to impose on the developer. @JThramer, do you have a preference?

Whatever the case, all the button-typed members of this class should be consistent: either they’re all lazily computed variables or they’re all factory methods without side effects.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/1962#discussion_r254898761

MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -102,35 +143,48 @@ class CarPlayMapViewController: UIViewController {
...
-            guard let strongSelf = self else {
-                return
-            }
-            strongSelf.mapView.setZoomLevel(strongSelf.mapView.zoomLevel - 1, animated: true)
+    /**
+     Creates a new close button to dismiss the visible panning buttons on the map.
+     
+     - parameter mapTemplate: The map template available to the pan map button for display.
+     */
+    @discardableResult public func createDismissPanningButton(for mapTemplate: CPMapTemplate) -> CPMapButton {
```
1ec5: “Create” is a noise word for a method name, like “get”. On the other hand, it would be confusing to start the method name with a verb, like “dismiss”. So how about `mapButton(forDismissingPanningInterfaceIn:)` or `panningInterfaceDismissalButton(for:)`?

MapboxNavigation/CarPlayManager.swift
```diff
@@ -457,48 +434,54 @@ extension CarPlayManager: CPMapTemplateDelegate {
...
-        mapTemplate.leadingNavigationBarButtons.insert(muteButton, at: 0)
-
         if let rootViewController = self.carWindow?.rootViewController as? CarPlayMapViewController,
             let trailingButtons = delegate?.carPlayManager?(self, trailingNavigationBarButtonsCompatibleWith: rootViewController.traitCollection, in: mapTemplate, for: .navigating) {
             mapTemplate.trailingNavigationBarButtons = trailingButtons
+        } else {
+            let exitButton = CPBarButton(type: .text) { [weak self] (button: CPBarButton) in
+                self?.currentNavigator?.exitNavigation(byCanceling: true)
+            }
+            exitButton.title = NSLocalizedString("CARPLAY_END", bundle: .mapboxNavigation, value: "End", comment: "Title for end navigation button")
```
1ec5: What if the developer wants to return an exit button in their `CarPlayManagerDelegate.carPlayManager(_:trailingNavigationBarButtonsCompatibleWith:in:for:)` implementation? There should be a method either on CarPlayManager or CarPlayNavigationViewController that creates and returns such a button, so that the developer doesn’t necessarily have to supply their own icon or localized text.

MapboxNavigation/CarPlayMapViewController.swift
JThramer: @1ec5 I was the one who suggested the pattern to @vincethecoder. It seemed to me that it would be beneficial to put these in lazy var's in case the developer wanted to, for example, hide/disable the button.

MapboxNavigation/CarPlayManager.swift
vincethecoder: only while browsing

vincethecoder: @1ec5 I see what you mean. Updated the panning enum to be more specific description.

MapboxNavigation/CarPlayMapViewController.swift
1ec5: I see. It’s just that the two buttons that you’re most likely to want to show and hide, the panning buttons, are in factory methods because they require a map template as context.

PR #1964, opened 2019-02-07
Obsolete NavigationMapViewDelegate methods

### LINKED ISSUE #1918: Remove redundant map view methods from NavigationViewControllerDelegate
I'm trying to set the destination annotation image of my NavigationController by implementing navigationViewController(navigationViewController::imageFor annotation:) but the method is never called;

After investigation, some NavigationMapViewDelegate methods from NavigationMapView are never invoked like this one:

```swift
optional func navigationMapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?
```

I think one possible solution would be to set the NavigatinMapView delegate to itself and dispatch the missing MGLMapViewDelegate method to navigationMapDelegate.

An example that solves the issue, in NavigatinMapView.swift:

```swift
fileprivate func commonInit() {
    self.delegate = self
    // Rest of the code
    // ...
  }
```

```swift
extension NavigationMapView: MGLMapViewDelegate {
    public func mapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage? {
        return self.navigationMapDelegate?.navigationMapView?(mapView, imageFor: annotation)
    }
}
```

I am using MapboxCoreNavigation 0.27.0. Same issue with 0.26.0.

JThramer:
Like it says on the box. Fixes #1918.

* Moving NavigationMapViewDelegate type into its own file. Also obsoleting delegate methods that are no longer being forwarded from the MGLMapView via its delegate.
* Also taking care of downstream effects.

/cc @mapbox/navigation-ios


MapboxNavigation/NavigationViewControllerDelegate.swift
```diff
@@ -191,5 +175,16 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
      */
     @objc(navigationViewController:roadNameAtLocation:)
     optional func navigationViewController(_ navigationViewController: NavigationViewController, roadNameAt location: CLLocation) -> String?
+    
+    
+    //MARK: Obsolete
+    
+    @available(*, obsoleted: 0.1, message: "The NavigationViewController no longer forwards MGLMapView messages. Use MGLMapViewDelegate.mapView(_:imageFor:) instead.")
```
1ec5: Technically these are MGLMapViewDelegate messages, but anyways I think we should keep these messages as brief as possible, because Xcode only shows a line or two of the message by default (including the standard deprecation warning message), and it shows even less in the source code editor’s margin.

```suggestion
    @available(*, obsoleted: 0.1, message: "Use MGLMapViewDelegate.mapView(_:imageFor:) instead.")
```

(×2)

PR #1979, opened 2019-02-18
Fix a bug with multileg routes

frederoni:
Fixes a regression introduced in #1904 

- Advance leg index when `NavigationServiceDelegate.navigationService(_:didArriveAt:)` returns `true`.

cc @JThramer


1ec5:
Added a blurb to the changelog in 90d37ff109559ff533e8d8430404937a66d61862.

PR #1981, opened 2019-02-18
Carplay Route Failure Delegate Messaging

JThramer:
Epic fail.

Adding a delegate message notifying the developer of a route-fetch failure. Developer can optionally return an alert which will be presented to the user.

To-Do: 
- [x] Default Implementation in Example App
- [x] ~~use `CPAlertTemplate` instead of navigation alert~~ (only `CPNavigationAlert` is possible)
- [x] Tests
- [x] Changelog Entry

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### CarPlay
 
-* The `CarPlayManager` will now notify its delegate if the route request fails, and provides the option to present an alert on the map template.  ([#1981](https://github.com/mapbox/mapbox-navigation-ios/pull/1981))
+* The `CarPlayManager` will now notify its delegate if the route request fails and provide the option to present an alert on the map template.  ([#1981](https://github.com/mapbox/mapbox-navigation-ios/pull/1981))
 * You can now customize the control layer of the map template comprising of the navigation bar's leading and trailing buttons and the map buttons. ([#1962](https://github.com/mapbox/mapbox-navigation-ios/pull/1962))
 * Added new map buttons in the `CarPlayManager` and the `CarPlayMapViewController`. You can now access map buttons that perform built-in actions on the map by accessing read-only properties such as: `CarPlayManager.exitButton`, `CarPlayManager.muteButton`, `CarPlayManager.showFeedbackButton`, `CarPlayManager.overviewButton`, `CarPlayMapViewController.recenterButton`, `CarPlayMapViewController.zoomInButton`, `CarPlayMapViewController.zoomOutButton`, `CarPlayMapViewController.panningInterfaceDisplayButton(for:)`, `CarPlayMapViewController.panningInterfaceDismissalButton(for:)`. ([#1962](https://github.com/mapbox/mapbox-navigation-ios/pull/1962))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### CarPlay
 
+* The `CarPlayManager` will now notify its delegate if the route request fails, and provides the option to present an alert on the map template.  ([#1981](https://github.com/mapbox/mapbox-navigation-ios/pull/1981))
 * You can now customize the control layer of the map template comprising of the navigation bar's leading and trailing buttons and the map buttons. ([#1962](https://github.com/mapbox/mapbox-navigation-ios/pull/1962))
 * Added new map buttons in the `CarPlayManager` and the `CarPlayMapViewController`. You can now access map buttons that perform built-in actions on the map by accessing read-only properties such as: `CarPlayManager.exitButton`, `CarPlayManager.muteButton`, `CarPlayManager.showFeedbackButton`, `CarPlayManager.overviewButton`, `CarPlayMapViewController.recenterButton`, `CarPlayMapViewController.zoomInButton`, `CarPlayMapViewController.zoomOutButton`, `CarPlayMapViewController.panningInterfaceDisplayButton(for:)`, `CarPlayMapViewController.panningInterfaceDismissalButton(for:)`. ([#1962](https://github.com/mapbox/mapbox-navigation-ios/pull/1962))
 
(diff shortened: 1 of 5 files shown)


MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -93,6 +93,18 @@ public protocol CarPlayManagerDelegate {
...
+     Called when the CarPlay manager fails to fetch a route.
+     - parameter carPlayManager: The shared CarPlay manager.
+     - parameter waypoints: the waypoints for which a route could not be retrieved.
+     - parameter options: The route options that were attached to the route request.
+     - parameter error: The error returned from the directions API.
+     
+     - returns Optionally, a CPNavigationAlert to present to the user. If an alert is returned, Carplay will transition back to the map template and display the alert. If `nil` is returned, nothing is done.
+     */
+    @objc(carPlayManager:didFailToFetchRouteBetweenWaypoints:withOptions:becauseOfError:)
+    optional func carplayManager(_ carPlayManager: CarPlayManager, didFailToFetchRouteBetweenWaypoints waypoints: [Waypoint]?, options: RouteOptions, error: NSError) -> CPNavigationAlert?
```
JThramer: Camel-case CarPlay.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -406,16 +406,12 @@ extension CarPlayManager: CPListTemplateDelegate {
...
+                    return
             }
-            let alert = CPNavigationAlert(titleVariants: [error.localizedDescription],
-                                          subtitleVariants: [error.localizedFailureReason ?? ""],
-                                          imageSet: nil,
-                                          primaryAction: okAction,
-                                          secondaryAction: nil,
-                                          duration: 0)
+
+            interfaceController.popToRootTemplate(animated: true)
```
akitchen: I'm not so sure of this one - seems like a POLS violation, and the user may lose context on their current activity.

I'd like to try this out on a device when ready & we can figure out if there are any alternatives in order to successfully display an alert.

MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -93,6 +93,18 @@ public protocol CarPlayManagerDelegate {
     @objc(carPlayManager:searchTemplate:selectedResult:completionHandler:)
     optional func carPlayManager(_ carPlayManager: CarPlayManager, searchTemplate: CPSearchTemplate, selectedResult item: CPListItem, completionHandler: @escaping () -> Void)
     
+    /**
+     Called when the CarPlay manager fails to fetch a route.
+     - parameter carPlayManager: The shared CarPlay manager.
```
akitchen: note, the CarPlayManager instance is no longer a shared instance

MapboxNavigation/CarPlayManager.swift
JThramer: posterity: let's try a `CPAlertTemplate` here so we don't have to blow the stack away.

JThramer: Update: looks like a `CPAlertTemplate` is not possible.

```swift
2019-02-18 15:37:34.803372-0700 Example-CarPlay[23447:1959284] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Unsupported object <CPAlertTemplate: 0x600001272580> <identifier: 6638156E-21B0-435F-BD42-DB2B40173057, userInfo: (null)> passed to pushTemplate:animated:. Allowed classes: {(
    CPSearchTemplate,
    CPMapTemplate,
    CPListTemplate,
    CPGridTemplate
)}
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### CarPlay
 
+* The `CarPlayManager` will now notify its delegate if the route request fails, and provides the option to present an alert on the map template.  ([#1981](https://github.com/mapbox/mapbox-navigation-ios/pull/1981))
```
akitchen:
```suggestion
* The `CarPlayManager` will now notify its delegate if the route request fails and provide the option to present an alert on the map template.  ([#1981](https://github.com/mapbox/mapbox-navigation-ios/pull/1981))
```

MapboxNavigation/CarPlayManager.swift
```diff
@@ -395,28 +395,23 @@ extension CarPlayManager: CPListTemplateDelegate {
...
-                                          subtitleVariants: [error.localizedFailureReason ?? ""],
-                                          imageSet: nil,
-                                          primaryAction: okAction,
-                                          secondaryAction: nil,
-                                          duration: 0)
-            mapTemplate.present(navigationAlert: alert, animated: true)
+
+            let mapTemplate = interfaceController?.rootTemplate as? CPMapTemplate
+            interfaceController?.popToRootTemplate(animated: true)
+            mapTemplate?.present(navigationAlert: alert, animated: true)
```
akitchen: hmm, so we're still stuck with tearing down the navigation stack?

MapboxNavigation/CarPlayManager.swift
```diff
@@ -395,28 +395,23 @@ extension CarPlayManager: CPListTemplateDelegate {
...
+            return
```
akitchen: It's worth calling out that this line is an important change in this PR, right?

MapboxNavigation/CarPlayManager.swift
JThramer: Correct. There is no other way that I can find to display an alert in CarPlay-Nav.

MapboxNavigation/CarPlayManager.swift
JThramer: Having a route and an error are mutually exclusive events, so there is no effective behavioral change that happens here. I just added this as an explicit hint that "we go no further in this case."

PR #1986, opened 2019-02-21
Reuse proactive rerouting from LegacyRouteController

frederoni:
Working towards #1902 - Enable proactive rerouting by default 

- [x] Reuse proactive rerouting from `LegacyRouteController`
- [ ] Add tests


MapboxCoreNavigation/Router.swift
```diff
@@ -50,6 +50,11 @@ public protocol RouterDataSource {
      */
     @objc var location: CLLocation? { get }
     
+    /**
+     If true, the `RouteController` attempts to calculate a more optimal route for the user on an interval defined by `RouteControllerProactiveReroutingInterval`.
+     */
+    @objc var reroutesProactively: Bool { get }
```
1ec5: We’re getting close to being able to refresh the existing route on a timer as well. Should we control both behaviors through a single, more generically named property, on an interval defined by a single, more generically named constant? Or should we keep the two complementary behaviors separate?

akitchen: I think it's OK to keep a separate flag for "reroutesProactively" (although the naming leaves a little to be desired). Faster route polling should happen _directly after_ a directions refresh response, so it's possible we'll be reworking this code again soon.

I don't think we should combine them into one option, to directly answer your question.

MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -653,3 +567,5 @@ open class LegacyRouteController: NSObject, Router, CLLocationManagerDelegate {
         }
     }
 }
+
+extension LegacyRouteController: InternalRouter { }
```
frederoni: This line introduces a deprecation warning. Not sure how to work around that 🤔

MapboxCoreNavigation/Router.swift
noway: what does 'refresh the existing route' mean? separate i reckon, by the sound of it.

PR #1988, opened 2019-02-22
Do not use invalid location course in step advancement heuristic.

Woildan:
I noticed a while ago a minor edge case in the step advancement heuristic that I fixed in our fork but never took time to contribute back, so here it is.

Basically if the user doesn't have a valid course, the wrap of the value gives **359**, which is then used as a valid value to compare against the maneuver final heading.

This can lead to advancing the step too early.
I guess it is better to let the fallback mechanism kick in in this case (user moving away from the maneuver location) to trigger the step advancement.


Woildan:
Sorry will open properly from a dedicated branch.

PR #1994, opened 2019-02-27
Fix depart and arrive event

frederoni:
- [x] Make events agnostic to real-time so we can speed up the tests
- [x] Avoid duplicated depart event (your assumption was correct here @akitchen)
- [x] Set reroute timestamp

Regarding the reroute timestamp, `NavigationEventDetails.created` is now mutable to support the reroute case.

cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/NavigationService.swift
+++ b/MapboxCoreNavigation/NavigationService.swift
@@ -504,10 +504,6 @@ extension MapboxNavigationService {
         return self.router.routeProgress
     }
     
-    public var location: CLLocation? {
-        return self.router.location
-    }
-    
     public var desiredAccuracy: CLLocationAccuracy {
         return self.locationManager.desiredAccuracy
     }
--- a/MapboxCoreNavigation/NavigationService.swift
+++ b/MapboxCoreNavigation/NavigationService.swift
@@ -505,7 +505,7 @@ extension MapboxNavigationService {
     }
     
     public var location: CLLocation? {
-        return self.locationManager.location
+        return self.router.location
     }
     
     public var desiredAccuracy: CLLocationAccuracy {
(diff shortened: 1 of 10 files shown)


TestHelper/CoreLocation.swift
```diff
@@ -136,14 +136,18 @@ extension Array where Element == CLLocation {
     
     // Shifts the [CLLocation]’s first location to now and offsets the remaining locations by one second after the prior.
     public func shiftedToPresent() -> [CLLocation] {
-        let now = Date()
+        return shiftedTo(Date())
+    }
+    
+    // Shifts the [CLLocation]’s first location to the given timestamp and offsets the  remaining locations by one second after the prior.
+    public func shiftedTo(_ timestamp: Date) -> [CLLocation] {
```
1ec5: Nit:

```swift
public func shifted(to timestamp: Date) -> [CLLocation] {
```

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -505,7 +505,7 @@ extension MapboxNavigationService {
     }
     
     public var location: CLLocation? {
-        return self.locationManager.location
+        return self.router.location
```
1ec5: If I’m not mistaken, this is a change from raw locations to snapped locations. Are there implications beyond location-based events? If so, we should note the change in the changelog.

TestHelper/CoreLocation.swift
akitchen: I don't disagree with the improved swiftiness of this suggestion, but wouldn't that necessitate a more objc-friendly variant?

MapboxCoreNavigation/NavigationService.swift
frederoni: Good point. We do want to use raw locations for events. We don't use `NavigationService.location` in any other places in the SDK so it shouldn't have any other implications. Would you mind re-reviewing https://github.com/mapbox/mapbox-navigation-ios/pull/1994/commits/e060c8d905c02f8f867d4ed4c4098ba959ad0046?

TestHelper/CoreLocation.swift
frederoni: Renamed to `shifted(to:)`. This is in the TestHelper which isn't being used in Obj-C currently.

akitchen:
Anything left to do here before we merge this @1ec5 @fred?

/cc @coxchapman

PR #1997, opened 2019-02-28
Optional CarPlay arrival UI

frederoni:
- Added `CarPlayManagerDelegate.carPlayManager(_:shouldPresentArrivalUIFor:)` that lets you determine whether to show arrival UI or not. Defaults to true.

cc @JThramer @akitchen

### DIFF
--- a/Example/AppDelegate+CarPlay.swift
+++ b/Example/AppDelegate+CarPlay.swift
@@ -4,6 +4,7 @@ import MapboxNavigation
 import CarPlay
 import MapboxCoreNavigation
 import MapboxDirections
+import TestHelper
 
 
 /**
@@ -43,6 +44,10 @@ extension AppDelegate: CarPlayManagerDelegate {
         NavigationViewController.carPlayManager(carPlayManager, didBeginNavigationWith: service, window: window)
     }
     
+    func carPlayManager(_ carPlayManager: CarPlayManager, shouldPresentArrivalUIfor waypoint: Waypoint) -> Bool {
+        return false
+    }
+    
     func carPlayManagerDidEndNavigation(_ carPlayManager: CarPlayManager) {
         // Dismiss NavigationViewController if it's present in the navigation stack
         guard let window = window else { return }
@@ -56,7 +61,12 @@ extension AppDelegate: CarPlayManagerDelegate {
                                          detailText: FavoritesList.POI.timesSquare.subTitle)
         mapboxSFItem.userInfo = [CarPlayManager.CarPlayWaypointKey: Waypoint(location: FavoritesList.POI.mapboxSF.location)]
         timesSquareItem.userInfo = [CarPlayManager.CarPlayWaypointKey: Waypoint(location: FavoritesList.POI.timesSquare.location)]
-        let listSection = CPListSection(items: [mapboxSFItem, timesSquareItem])
+        
+        let multilegRouteItem = CPListItem(text: "Multileg route", detailText: "Liechtenstein")
+        let multilegRoute = Fixture.route(from: "multileg-route")
+        multilegRouteItem.userInfo = [CarPlayManager.CarPlayWaypointKey: multilegRoute.routeOptions.waypoints]
+        
+        let listSection = CPListSection(items: [mapboxSFItem, timesSquareItem, multilegRouteItem])
         return CPListTemplate(title: "Favorites List", sections: [listSection])
     }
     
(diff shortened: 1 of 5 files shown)


Example/AppDelegate+CarPlay.swift
```diff
@@ -43,6 +44,10 @@ extension AppDelegate: CarPlayManagerDelegate {
         NavigationViewController.carPlayManager(carPlayManager, didBeginNavigationWith: service, window: window)
     }
     
+    func carPlayManager(_ carPlayManager: CarPlayManager, shouldPresentArrivalUIfor waypoint: Waypoint) -> Bool {
```
akitchen: I think "for" should have a capital "F" here

Example/AppDelegate+CarPlay.swift
```diff
@@ -4,6 +4,7 @@ import MapboxNavigation
 import CarPlay
 import MapboxCoreNavigation
 import MapboxDirections
+import TestHelper
```
akitchen: i think we need a better code organization strategy here; we don't want to start bringing test code into the example app if we can avoid it...

Example/AppDelegate+CarPlay.swift
```diff
@@ -56,7 +61,12 @@ extension AppDelegate: CarPlayManagerDelegate {
                                          detailText: FavoritesList.POI.timesSquare.subTitle)
         mapboxSFItem.userInfo = [CarPlayManager.CarPlayWaypointKey: Waypoint(location: FavoritesList.POI.mapboxSF.location)]
         timesSquareItem.userInfo = [CarPlayManager.CarPlayWaypointKey: Waypoint(location: FavoritesList.POI.timesSquare.location)]
-        let listSection = CPListSection(items: [mapboxSFItem, timesSquareItem])
+        
+        let multilegRouteItem = CPListItem(text: "Multileg route", detailText: "Liechtenstein")
+        let multilegRoute = Fixture.route(from: "multileg-route")
+        multilegRouteItem.userInfo = [CarPlayManager.CarPlayWaypointKey: multilegRoute.routeOptions.waypoints]
```
JThramer: Instead of bringing in the entire route as a fixture (which incurs a problem with importing TestHelper), could we just inline the waypoints?

MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -152,6 +152,15 @@ public protocol CarPlayManagerDelegate {
      */
     @objc func carPlayManagerDidEndNavigation(_ carPlayManager: CarPlayManager) -> ()
     
+    /**
+     Called when the CarPlay navigation view controller detects an arrival.
+     
+     - returns: A boolean value indicating whether to display an arrival UI.
```
vincethecoder: You are missing documentation for the `parameters` here.

MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -152,6 +152,15 @@ public protocol CarPlayManagerDelegate {
...
+     */
+    
+    @objc(carPlayerManager:shouldPresentArrivalUIForWaypoint:)
+    optional func carPlayManager(_ carPlayManager: CarPlayManager, shouldPresentArrivalUIFor waypoint: Waypoint) -> Bool
```
vincethecoder: nit: how about `shouldDisplayArrivalUIForWaypoint`?

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -529,5 +536,12 @@ public protocol CarPlayNavigationDelegate {
      - parameter carPlayNavigationViewController: The CarPlay navigation view controller that was dismissed.
      */
     @objc func carPlayNavigationViewControllerDidArrive(_ carPlayNavigationViewController: CarPlayNavigationViewController)
+    
+    /**
+     Called when the CarPlay navigation view controller detects an arrival.
+     
+     - returns: A boolean value indicating whether to display an arrival UI.
```
vincethecoder: ditto: missing documentation for the `parameters` here as well.

MapboxNavigation/CarPlayManagerDelegate.swift
frederoni: "present" is the verb that's being used for presenting view controllers, documents and templates which we are doing here. "display" is more for styling and when using CoreGraphics directly.

1ec5:
#2000 tracks removing this arrival UI from the SDK and exposing a hook so that the application can implement one if desired.

akitchen:
Should we be closing this out in favor of the solution to #2000 ?

PR #2004, opened 2019-03-05
Safely unwrap spoken instruction

frederoni:
Safely unwrap spoken instruction

cc @JThramer

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -200,8 +200,9 @@ open class RouteController: NSObject {
             routeProgress.currentLegProgress.currentStepProgress.spokenInstructionIndex = Int(voiceInstructionIndex)
             
             // Don't annouce spoken instruction if we are going to reroute
-            if !willReRoute {
-                announcePassage(of: routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction!, routeProgress: routeProgress)
+            if !willReRoute,
+                let spokenInstruction = routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction {
+                announcePassage(of: spokenInstruction, routeProgress: routeProgress)
             }
         }
     }


MapboxCoreNavigation/RouteController.swift
```diff
@@ -200,8 +200,9 @@ open class RouteController: NSObject {
             routeProgress.currentLegProgress.currentStepProgress.spokenInstructionIndex = Int(voiceInstructionIndex)
             
             // Don't annouce spoken instruction if we are going to reroute
-            if !willReRoute {
-                announcePassage(of: routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction!, routeProgress: routeProgress)
+            if !willReRoute,
+                let spokenInstruction = routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction {
```
1ec5: This fixes the symptom of a crash, but it seems very likely that the root cause is navigation-native’s model being out of sync with the SDK’s model somehow. I wonder if it’s significant that `RouteController.updateSpokenInstructionProgress(status:willReRoute:)` gets called before `updateRouteStepProgress(status:)` and `updateRouteLegProgress(status:)`. In LegacyRouteController, both the step and leg indices get incremented before updating spoken instruction progress. We even decide whether to reroute the user before updating the spoken instruction progress:

https://github.com/mapbox/mapbox-navigation-ios/blob/c1405cbb0b95fb6a35c580b691d4794f6b1a0205/MapboxCoreNavigation/LegacyRouteController.swift#L272-L281

The change is consistent with `updateVisualInstructionProgress(status:)`, so I guess it should stand. But it’s going to be more difficult to diagnose issues with missing voice instructions, if that’s how the bug will manifest going forward.

1ec5: OK, we’ve confirmed that the root cause is that `updateSpokenInstructionProgress(status:willReRoute:)` accesses `routeProgress.currentLegProgress.currentStepProgress.currentSpokenInstruction` after updating `spokenInstructionIndex` (and thus `currentSpokenInstruction`) but before updating `stepIndex` or `legIndex` (which would update `currentStepProgress` and `currentLegProgress`, respectively).

frederoni: Confirmed and proposed a fix in #2029

PR #2007, opened 2019-03-06
List downloaded offline packs

frederoni:
To find an already downloaded offline routing pack, you have to scan for directories in a bundle, filter out relevant versions and verify the integrity of that folder or make an API call to `fetchAvailableOfflineVersions(completionHandler:)` and then check for available tiles in the corresponding version’s directory.

I added a helper method to simplify these steps.

cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/OfflineDirections.swift
+++ b/MapboxCoreNavigation/OfflineDirections.swift
@@ -11,6 +11,7 @@ enum OfflineRoutingError: Error, LocalizedError {
     case unexpectedRouteResult(String)
     case corruptRouteData(String)
     case responseError(String)
+    case filePathNotFound(String)
     
     public var localizedDescription: String {
         switch self {
@@ -19,6 +20,8 @@ enum OfflineRoutingError: Error, LocalizedError {
         case .unexpectedRouteResult(let value):
             return value
         case .responseError(let value):
+            return value
+        case .filePathNotFound(let value):
             return value
         }
     }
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -80,6 +83,36 @@ public class NavigationDirections: Directions {
...
+     
+     - returns: A list of versions at the given file path that contains at least one tile.
+     */
+    @objc(downloadedVersionsInBundle:error:)
+    public static func downloadedVersions(in bundle: Bundle? = Bundle.mapboxCoreNavigation) throws -> [String] {
+        
+        let _bundle = bundle ?? Bundle.mapboxCoreNavigation
+        
+        guard let url = _bundle.suggestedTileURL else {
+                let message = NSLocalizedString("OFFLINE_FILEPATH_NOT_FOUND", bundle: .mapboxCoreNavigation, value: "Unable to find the file path for downloaded versions", comment: "Error description when the cache directory is inaccessible")
```
frederoni: `ExtractLocString` before merge

1ec5:
@IodaMikeMapbox, I think the functionality that would be added here will be removed in #2509, correct?

IodaMikeMapbox:
For now we decided to not remove the functionality related to managing tiles manually for offline but it can be removed at any time and no customer should rely on it in production, it can go away in short time because looks like navnative team wants to remove the ability to set tiles manually, instead of that offline regions will be managed in the CommonSdk.

That's why I'm closing the pr.

PR #2013, opened 2019-03-08
Clean up dead code and refactor reused extensions

frederoni:
Cleaned up dead code, unused parameters, and other clean up work to get reliable code coverage.
Also fixed the warnings about classes being a member of multiple targets.

cc @mapbox/navigation-ios


MapboxNavigation/NavigationCustomizable.swift
```diff
@@ -5,10 +5,6 @@ import MapboxCoreNavigation
  The `NavigationCustomizable` protocol represents a UI-based mechanism that allows for customization of its visual style, as well as the navigation service that powers it.
  */
 protocol NavigationCustomizable {
-    /**
-     The styles that the user interface can select from for display.
-    */
-    var styles: [Style]? { get }
```
frederoni: @JThramer what was the intention with this prop?

MapboxNavigation/NavigationOptions.swift
```diff
@@ -48,12 +48,5 @@ open class NavigationOptions: NSObject, NavigationCustomizable {
         self.voiceController = voiceController
         self.bottomBanner = bottomBanner
     }
-    
-    /**
-    Convienence factory-method for convenient bridging to OBJ-C.
-     */
-    class func navigationOptions() -> Self {
```
frederoni: this is not exposed to Obj-C and it's not being used

MapboxNavigation/NavigationCustomizable.swift
JThramer: @frederoni If I remember correctly the intention was to create a protocol that captured the similarities between the iOS NVC and the CPNVC. It's not really used for anything right now and can be removed.

MapboxNavigation.podspec
```diff
@@ -32,7 +32,7 @@ Pod::Spec.new do |s|
 
   # ――― Source Code ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
-  s.source_files = ["MapboxNavigation/**/*.{h,m,swift}", "MapboxCoreNavigation/{Date,Sequence,String}.swift"]
```
frederoni: @1ec5 this generated a warning with CocoaPods validation, right? Something we should double check?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -437,7 +437,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         let line = MGLPolyline(coordinates: coords, count: UInt(coords.count))
         let camera = cameraThatFitsShape(line, direction: direction, edgePadding: padding)
         
-        setCamera(camera, animated: false)
+        setCamera(camera, animated: animated)
```
akitchen: nice catch, I wonder if there was a reason this was explicitly set to `false` before?

MapboxNavigation/NavigationOptions.swift
1ec5: It should be exposed to Objective-C: https://github.com/mapbox/mapbox-navigation-ios/pull/1951#discussion_r252906378.

MapboxNavigation/FeedbackItem.swift
```diff
@@ -41,8 +41,6 @@ public class FeedbackItem: NSObject {
     static let reportTraffic = FeedbackItem(title: reportTrafficTitle, image: .feedbackImage(named:"feedback_traffic"), feedbackType: .reportTraffic)
     static let confusingInstructions = FeedbackItem(title: confusingInstructionTitle, image: .feedbackImage(named:"feedback_confusing"), feedbackType: .confusingInstruction)
     static let badRoute = FeedbackItem(title: badRouteTitle, image: .feedbackImage(named:"feedback_wrong"), feedbackType: .routingError)
-    static let missingRoad = FeedbackItem(title: missingExitTitle, image: .feedbackImage(named:"feedback-missing-road"), feedbackType: .missingRoad)
-    static let missingExit = FeedbackItem(title: missingRoadTitle, image: .feedbackImage(named:"feedback-exit"), feedbackType: .missingExit)
```
1ec5: Also delete the corresponding assets from Assets.xcassets.

MapboxNavigation/FeedbackItem.swift
```diff
@@ -51,6 +49,4 @@ fileprivate let notAllowedTitle = NSLocalizedString("FEEDBACK_NOT_ALLOWED", bund
 fileprivate let reportTrafficTitle = NSLocalizedString("FEEDBACK_REPORT_TRAFFIC", bundle: .mapboxNavigation, value: "Report\nTraffic", comment: "Feedback type for Report Traffic")
 fileprivate let confusingInstructionTitle = NSLocalizedString("FEEDBACK_CONFUSING_INSTRUCTION", bundle: .mapboxNavigation, value: "Confusing\nInstruction", comment: "Feedback type for Confusing Instruction")
 fileprivate let badRouteTitle = NSLocalizedString("FEEDBACK_BAD_ROUTE", bundle: .mapboxNavigation, value: "Bad \nRoute", comment: "Feedback type for Bad Route")
-fileprivate let missingExitTitle = NSLocalizedString("FEEDBACK_MISSING_EXIT", bundle: .mapboxNavigation, value: "Missing\nExit", comment: "Feedback type for Missing Exit")
-fileprivate let missingRoadTitle = NSLocalizedString("FEEDBACK_MISSING_ROAD", bundle: .mapboxNavigation, value: "Missing\nRoad", comment: "Feedback type for Missing Road")
```
1ec5: Also delete the strings from Localizable.strings.

MapboxNavigation/NavigationMapView.swift
1ec5: Probably an oversight. But all the code paths currently call this method with `false` for `animated`, so the change has no actual effect.

MapboxNavigation/NavigationOptions.swift
frederoni: Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/2013/commits/4a781a7482fb8d349dc8a9cbc46e472617d5b1c2

1ec5:
This change made extract_localizable.sh nonexecutable. You must’ve edited it or committed part of it in Xcode: [rdar://problem/25729523](http://www.openradar.me/25729523) (lionheart/openradar-mirror#11401).

PR #2014, opened 2019-03-08
Preserve via points after rerouting

### LINKED ISSUE #2003: Via points lost when rerouting
Via points (waypoints with `separatesLegs` set to `false`) are lost when the user is rerouted. `RouteProgress.reroutingOptions(with:)` comes up with a set of waypoints that consists of the user’s current location and any remaining waypoints:

https://github.com/mapbox/mapbox-navigation-ios/blob/64f07b745decbd2d7ed761bb46c2b24f8aa72289/MapboxCoreNavigation/RouteProgress.swift#L256

But that only considers explicit waypoints that separate legs, not via points:

https://github.com/mapbox/mapbox-navigation-ios/blob/64f07b745decbd2d7ed761bb46c2b24f8aa72289/MapboxCoreNavigation/RouteProgress.swift#L92-L94

We’ll need to figure out which waypoints in the original route options are along the remaining route. One edge case to consider is a route that doubles back on itself; we’ll need to preserve the order in that case.

/cc @mapbox/navigation-ios

1ec5:
Fixed an issue where rerouting would ignore any waypoints from the original route options where the `Waypoint.separatesLegs` property was set to `false`.

~~This is still a work in progress – some test cases fail because the algorithm doesn’t adequately account for edge cases when backtracking.~~

Fixes #2003.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -87,10 +87,21 @@ open class RouteProgress: NSObject {
     }
 
     /**
-     Number of waypoints remaining on the current route.
+     The waypoints remaining on the current route.
+     
+     This property does not include waypoints whose `Waypoint.separatesLegs` property is set to `false`.
      */
     @objc public var remainingWaypoints: [Waypoint] {
         return route.legs.suffix(from: legIndex).map { $0.destination }
+    }
+    
+    /**
+     The waypoints remaining on the current route, including any waypoints that do not separate legs.
+     */
+    func remainingWaypointsForCalculatingRoute() -> [Waypoint] {
+        let (currentLegViaPoints, remainingWaypoints) = route.routeOptions.waypoints(fromLegAt: legIndex)
+        let currentLegRemainingViaPoints = currentLegProgress.remainingWaypoints(among: currentLegViaPoints)
+        return currentLegRemainingViaPoints + remainingWaypoints
     }
 
     /**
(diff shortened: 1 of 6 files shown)


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -454,6 +481,16 @@ open class RouteLegProgress: NSObject {
...
     }
+    
+    /**
+     The waypoints remaining on the current leg, not including the leg’s destination.
+     */
+    func remainingWaypoints(among waypoints: [Waypoint]) -> [Waypoint] {
+        let legPolyline = Polyline(leg.coordinates)
+        let userCoordinateIndex = legPolyline.indexedCoordinateFromStart(distance: distanceTraveled)?.index ?? 0
+        // TODO: What if the index doesn’t monotonically increase due to backtracking?
+        return Array(waypoints.drop { legPolyline.closestCoordinate(to: $0.coordinate)?.index ?? 0 <= userCoordinateIndex })
```
1ec5: This would be more straightforward if Turf would provide a purely distance-based `indexedCoordinate(from:to:)`, or at least if IndexedCoordinate could indicate the distance from the start of the polyline instead of just the snapping distance: mapbox/turf-swift#78.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -454,6 +465,29 @@ open class RouteLegProgress: NSObject {
...
+            // The leg is empty, so none of the waypoints are meaningful.
+            return []
+        }
+        var slice = legPolyline
+        var accumulatedCoordinates = 0
+        return Array(waypoints.drop { (waypoint) -> Bool in
+            var newSlice = slice.sliced(from: waypoint.coordinate)
+            // Work around <https://github.com/mapbox/turf-swift/pull/79>.
+            if newSlice.coordinates.count > 2 && newSlice.coordinates.last == newSlice.coordinates.dropLast().last {
+                newSlice.coordinates.removeLast()
```
1ec5: This workaround for mapbox/turf-swift#79 should be trivial in terms of performance.

PR #2018, opened 2019-03-11
CarPlay Fixes (Short-term)

### LINKED ISSUE #2017: iPhone <> CarPlay continuity
1. Plugging in after starting TBT on the phone should continue on CarPlay, and vice versa

2. There should only be one NavigationService in play at a given time - centralize creation, perhaps via a provider

3. There should only be one VoiceController in play at a given time - controlling a hardware resource, a good candidate for the singleton pattern

@JThramer wsg. @1ec5 and/or @frederoni

JThramer:
Off to the races!

Implements deliverable 2 & 3 in #2017. 

Manifest: 

- [x] Add service-specific subscription on voice controllers
- [x] Add service-specific subscription for `CarPlayNavigationViewController`
- [x] Make `CarPlayManager(_:navigationServiceAlong:) -> NavigationService` non-optional
- [x] Remove Delegate mechanism for CarPlayNavigationViewController
- [x] Test ~changes?~ fixes
- [x] Changelog entry

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -479,7 +479,7 @@ extension CarPlayManager: CPMapTemplateDelegate {
         let route = routeChoice.userInfo as! Route
         var service: NavigationService
         
-        if let override = delegate?.carPlayManager?(self, navigationServiceAlong: route) {
+        if let override = delegate?.carPlayManager(self, navigationServiceAlong: route) {
             service = override
         } else {
             service = MapboxNavigationService(route: route, simulating: simulatesLocations ? .always : .onPoorGPS)
--- a/MapboxNavigation/CarPlayManagerDelegate.swift
+++ b/MapboxNavigation/CarPlayManagerDelegate.swift
@@ -59,7 +59,7 @@ public protocol CarPlayManagerDelegate {
     optional func carPlayManager(_ carPlayManager: CarPlayManager, mapButtonsCompatibleWith traitCollection: UITraitCollection, in template: CPTemplate, for activity: CarPlayActivity) -> [CPMapButton]?
     
     /**
-     Asks the delegate to provide a navigation service.
+     Asks the delegate to provide a navigation service. In multi-screen applications this should be the same instance used to guide the user along the route on the phone.
      
      - parameter carPlayManager: The CarPlay manager instance.
      - parameter route: The route for which the returned route controller will manage location updates.
--- a/MapboxNavigation/CarPlayManagerDelegate.swift
+++ b/MapboxNavigation/CarPlayManagerDelegate.swift
@@ -57,7 +57,7 @@ public protocol CarPlayManagerDelegate {
     optional func carPlayManager(_ carPlayManager: CarPlayManager, mapButtonsCompatibleWith traitCollection: UITraitCollection, in template: CPTemplate, for activity: CarPlayActivity) -> [CPMapButton]?
     
     /**
-     Offers the delegate an opportunity to provide an alternate navigation service, otherwise a default built-in MapboxNavigationService will be created and used.
+     Asks the delegate to provide an alternate navigation service. If no delegate is set, a default built-in MapboxNavigationService will be created and used.
      
      - parameter carPlayManager: The CarPlay manager instance.
      - parameter route: The route for which the returned route controller will manage location updates.
@@ -65,7 +65,7 @@ public protocol CarPlayManagerDelegate {
      */
     
     @objc(carPlayManager:navigationServiceAlongRoute:)
-    optional func carPlayManager(_ carPlayManager: CarPlayManager, navigationServiceAlong route: Route) -> NavigationService
+    func carPlayManager(_ carPlayManager: CarPlayManager, navigationServiceAlong route: Route) -> NavigationService
     
     /**
      Offers the delegate an opportunity to react to updates in the search text.
--- a/MapboxNavigation/CarPlayManagerDelegate.swift
+++ b/MapboxNavigation/CarPlayManagerDelegate.swift
@@ -57,7 +57,7 @@ public protocol CarPlayManagerDelegate {
     optional func carPlayManager(_ carPlayManager: CarPlayManager, mapButtonsCompatibleWith traitCollection: UITraitCollection, in template: CPTemplate, for activity: CarPlayActivity) -> [CPMapButton]?
     
     /**
-     Asks the delegate to provide an alternate navigation service. If no delegate is set, a default built-in MapboxNavigationService will be created and used.
+     Asks the delegate to provide a navigation service. If no delegate is set, a default built-in MapboxNavigationService will be created and used.
      
      - parameter carPlayManager: The CarPlay manager instance.
      - parameter route: The route for which the returned route controller will manage location updates.
--- a/MapboxNavigation/CarPlayManagerDelegate.swift
+++ b/MapboxNavigation/CarPlayManagerDelegate.swift
@@ -8,6 +8,8 @@ import MapboxDirections
  `CarPlayManagerDelegate` is the main integration point for Mapbox CarPlay support.
  
  Implement this protocol and assign an instance to the `delegate` property of the shared instance of `CarPlayManager`.
+ 
+ If no delegate is set, a default built-in MapboxNavigationService will be created and used.
  */
 @available(iOS 12.0, *)
 @objc(MBCarPlayManagerDelegate)
@@ -57,7 +59,7 @@ public protocol CarPlayManagerDelegate {
     optional func carPlayManager(_ carPlayManager: CarPlayManager, mapButtonsCompatibleWith traitCollection: UITraitCollection, in template: CPTemplate, for activity: CarPlayActivity) -> [CPMapButton]?
     
     /**
-     Asks the delegate to provide a navigation service. If no delegate is set, a default built-in MapboxNavigationService will be created and used.
+     Asks the delegate to provide a navigation service.
      
      - parameter carPlayManager: The CarPlay manager instance.
      - parameter route: The route for which the returned route controller will manage location updates.
--- a/MapboxNavigation/CarPlayManagerDelegate.swift
+++ b/MapboxNavigation/CarPlayManagerDelegate.swift
@@ -9,7 +9,7 @@ import MapboxDirections
  
  Implement this protocol and assign an instance to the `delegate` property of the shared instance of `CarPlayManager`.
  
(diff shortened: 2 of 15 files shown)


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -85,14 +85,19 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
         verifyBackgroundAudio()
 
         speechSynth.delegate = self
         
-        resumeNotifications()
+        resumeNotifications(service: navigationService)
+    }
+    
+    public override init() {
+        fatalError("Use init(navigationService:) instead.")
```
JThramer: Don't forget to put a changelog entry about this.

MapboxNavigationTests/MapboxVoiceControllerTests.swift
```diff
@@ -10,7 +10,7 @@ import AVKit
 class MapboxVoiceControllerTests: XCTestCase {
 
     var speechAPISpy: SpeechAPISpy!
-    var controller: MapboxVoiceController?
+//    var controller: MapboxVoiceController?
```
JThramer: remove comments, also 27, also 34

MapboxNavigation/CarPlayManager.swift
```diff
@@ -479,7 +479,7 @@ extension CarPlayManager: CPMapTemplateDelegate {
         let route = routeChoice.userInfo as! Route
         var service: NavigationService
         
-        if let override = delegate?.carPlayManager?(self, navigationServiceAlong: route) {
+        if let override = delegate?.carPlayManager(self, navigationServiceAlong: route) {
             service = override
         } else {
             service = MapboxNavigationService(route: route, simulating: simulatesLocations ? .always : .onPoorGPS)
```
akitchen: Ultimately, we will be moving away from ad hoc creation of a service to something else, right? Retrieval from a registry, forced dependency injection, magic pixie dust... ?

MapboxNavigationTests/MapboxVoiceControllerTests.swift
```diff
@@ -24,22 +24,25 @@ class MapboxVoiceControllerTests: XCTestCase {
...
     }
 
     func testControllerDownloadsAndCachesInstructionDataWhenNotified() {
-        let routeProgress = RouteProgress.init(route: route, legIndex: 0, spokenInstructionIndex: 0)
-        let userInfo = [RouteControllerNotificationUserInfoKey.routeProgressKey : routeProgress]
-        let notification = Notification.init(name: .routeControllerDidPassSpokenInstructionPoint, object: nil, userInfo: userInfo)
+        let service = MapboxNavigationService(route: route)
+        service.routeProgress.legIndex = 0
+        service.routeProgress.currentLegProgress.currentStepProgress.spokenInstructionIndex = 0
+        let subject = MapboxVoiceController(speechClient: speechAPISpy, audioPlayerType: AudioPlayerDummy.self, navigationService: service)
```
akitchen: fancy

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -507,18 +497,5 @@ public protocol CarPlayNavigationDelegate {
...
-     
-     - postcondition: If you return `false` within this method, you must manually advance to the next leg: obtain the value of the `navigationService` and its `NavigationService.routeProgress` property, then increment the `RouteProgress.legIndex` property.
-     - parameter carPlayNavigationViewController: The CarPlay navigation view controller that has arrived at a waypoint.
-     - parameter waypoint: The waypoint that the user has arrived at.
-     - returns: True to automatically advance to the next leg, or false to remain on the now completed leg.
-     */
-    @objc(carPlayNavigationViewController:didArriveAtWaypoint:)
-    optional func carPlayNavigationViewController(_ carPlayNavigationViewController: CarPlayNavigationViewController, didArriveAt waypoint: Waypoint) -> Bool
```
1ec5: This is effectively the removal of `CarPlayManagerDelegate.carPlayNavigationViewControllerDidArrive(_:)`. Since the replacement of `NavigationServiceDelegate.navigationService(_:didArriveAt:)` isn’t obvious, keep a rump method around under the old name of `CarPlayManagerDelegate.carPlayNavigationViewControllerDidArrive(_:)` and add a message directing the developer to use `NavigationServiceDelegate.navigationService(_:didArriveAt:)` instead.

akitchen:
This is currently `Draft` status - what are next steps?

BTW the Example-Carplay app doesn't compile at the moment; we're still missing wiring up of the new delegate method. It's important that the Example app demonstrate the basic use case of connecting to a phone after TBT has begun, in order to see it continue on CarPlay. Or is that considered out-of-scope for this PR?

akitchen:
Note that https://github.com/mapbox/mapbox-navigation-ios/pull/2019 adds the Example-CarPlay app to the CI job list. Not sure if it's viable to do this, but thought I'd at least propose it in order to help tighten the feedback loop here.

MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -8,6 +8,8 @@ import MapboxDirections
  `CarPlayManagerDelegate` is the main integration point for Mapbox CarPlay support.
  
  Implement this protocol and assign an instance to the `delegate` property of the shared instance of `CarPlayManager`.
+ 
+ If no delegate is set, a default built-in MapboxNavigationService will be created and used.
```
akitchen:
```suggestion
 If no delegate is set, a default built-in MapboxNavigationService will be created and used when a trip begins.
```

MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -57,15 +59,15 @@ public protocol CarPlayManagerDelegate {
     optional func carPlayManager(_ carPlayManager: CarPlayManager, mapButtonsCompatibleWith traitCollection: UITraitCollection, in template: CPTemplate, for activity: CarPlayActivity) -> [CPMapButton]?
     
     /**
-     Offers the delegate an opportunity to provide an alternate navigation service, otherwise a default built-in MapboxNavigationService will be created and used.
+     Asks the delegate to provide a navigation service.
```
akitchen:
```suggestion
     Asks the delegate to provide a navigation service. In multi-screen applications this should be the same instance used to guide the user along the route on the phone.
```

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -141,7 +140,7 @@ public class CarPlayNavigationViewController: UIViewController {
         styleManager!.styles = self.styles
         
         makeGestureRecognizersResetFrameRate()
-        resumeNotifications()
+        resumeNotifications(by: navigationService)
```
akitchen: minor suggestion: how about `observeNotifications(by:)` since that's closer to what we're doing, semantically?

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -110,10 +116,10 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
         speechSynth.stopSpeaking(at: .immediate)
     }
     
-    func resumeNotifications() {
-        NotificationCenter.default.addObserver(self, selector: #selector(didPassSpokenInstructionPoint(notification:)), name: .routeControllerDidPassSpokenInstructionPoint, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(pauseSpeechAndPlayReroutingDing(notification:)), name: .routeControllerWillReroute, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(didReroute(notification:)), name: .routeControllerDidReroute, object: nil)
+    func resumeNotifications(by service: NavigationService) {
```
akitchen: Same suggestion as in https://github.com/mapbox/mapbox-navigation-ios/pull/2018/files#r265309688 - how about `observeNotifications(by:)` instead? Not a big deal, but I think it reads a bit more cleanly.

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -431,6 +431,10 @@ class CarPlayManagerSpec: QuickSpec {
         func carPlayManagerDidEndNavigation(_ carPlayManager: CarPlayManager) {
             //no-op
         }
+        
+        func carPlayManager(_ carPlayManager: CarPlayManager, navigationServiceAlong route: Route) -> NavigationService {
+            return MapboxNavigationService(route: route)
```
akitchen: This might be fine but I wonder if we should be using a dummy instead? Init'ing this class does have some side effects.

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -455,6 +459,10 @@ class CarPlayManagerFailureDelegateSpy: CarPlayManagerDelegate {
         return nil
     }
     
+    func carPlayManager(_ carPlayManager: CarPlayManager, navigationServiceAlong route: Route) -> NavigationService {
+        return MapboxNavigationService(route: route)
```
akitchen: Same as above. Is this equivalent to what we were doing before?

akitchen:
oops, @JThramer , last thing remaining here is that Example-CarPlay app doesn't build due to the missing delegate method implementation.

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -508,16 +498,9 @@ public protocol CarPlayNavigationDelegate {
...
-     
-     - postcondition: If you return `false` within this method, you must manually advance to the next leg: obtain the value of the `navigationService` and its `NavigationService.routeProgress` property, then increment the `RouteProgress.legIndex` property.
-     - parameter carPlayNavigationViewController: The CarPlay navigation view controller that has arrived at a waypoint.
-     - parameter waypoint: The waypoint that the user has arrived at.
-     - returns: True to automatically advance to the next leg, or false to remain on the now completed leg.
-     */
+    //MARK: - Deprecated.
+    
+    @available(*, obsoleted: 0.1, message: "Use NavigationViewControllerDelegate.navigationViewController(_:didArriveAt:) or  NavigationServiceDelegate.navigationService(_:didArriveAt:) instead.")
```
1ec5: We should be recommending `NavigationServiceDelegate.navigationService(_:didArriveAt:)`, not `NavigationViewControllerDelegate.navigationViewController(_:didArriveAt:)`, since we know in this case that the application needs the navigation service to work on CarPlay.

PR #2019, opened 2019-03-12
Adjust circle.yml to build Example-CarPlay app

akitchen:
Note this is targeting `jerrad/carplay-short-term-fixup`

In support of this CarPlay work, we should ensure we are building the CarPlay-enabled example app.


akitchen:
Also note that the Example-CarPlay app currently isn't building due to a missing required delegate protocol method implementation. Understand that it's WIP, but wanted to add this to the CI feedback loop.

vincethecoder:
@JThramer this doesn't conform to the CarPlayManagerDelegate's navigation service function.

akitchen:
✌️

PR #2020, opened 2019-03-12
Various fixes to the CarPlay layout

frederoni:
This PR attempts to address
- [ ] Unreliable compass view visibility and position in CarPlay
- [x] Correct padding when previewing a route in CarPlay

- [ ] Fix unbalanced calls to begin/end appearance transitions for UINavigationController and CarPlayMapViewController
- [ ] Fix remaining unsatisfiable constraint
- [x] Reliable navigation bar state in CarPlay

cc @1ec5

### DIFF
--- a/MapboxNavigation/CarPlayMapViewController.swift
+++ b/MapboxNavigation/CarPlayMapViewController.swift
@@ -138,6 +138,11 @@ public class CarPlayMapViewController: UIViewController {
         resetCamera(animated: false, altitude: CarPlayMapViewController.defaultAltitude)
     }
     
+    public override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        print(".")
+    }
+    
     override public func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         styleObservation = nil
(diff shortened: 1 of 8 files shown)


MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -167,15 +169,18 @@ public class CarPlayNavigationViewController: UIViewController {
...
-            let navigationBarIsOpen = view.safeAreaInsets > previousSafeAreaInsets
-            mapView?.compassView.isHidden = navigationBarIsOpen
+        defer {
+            previousSafeAreaInsets = view.safeAreaInsets
+            view.setNeedsUpdateConstraints()
         }
         
-        previousSafeAreaInsets = view.safeAreaInsets
+        // TODO: Verify `?? 0` should result in navigationBarIsOpen = false initially
+        let navigationBarIsOpen = view.safeAreaInsets.top > previousSafeAreaInsets?.top ?? 0
```
1ec5: Is there a possibility that this method could be called twice with the same top safe area inset, perhaps due to a race condition? If so, this check would be unreliable.

In practice, the top safe area inset is zero when the navigation bar is hidden and nonzero when the navigation bar is shown, correct? What if we check the absolute value rather than the value relative to a previous value?

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -167,15 +169,18 @@ public class CarPlayNavigationViewController: UIViewController {
...
+        mapView?.compassView.isHidden = navigationBarIsOpen
         
         // Adjust the map’s vanishing point to counterbalance the side maneuver panels by extending the view off beyond the other side of the screen.
         if let mapView = mapView {
+            // TODO: We observe safeAreaInsets but safeArea is not updated initially
```
1ec5: Isn’t `safeArea` just a convenience for `safeAreaInsets` that accounts for iOS 11 and below? If this to-do is because we’d like to directly constrain to the map view’s `safeArea`, then we could add a dependency between the two properties via `keyPathsForValuesAffecting(_:)`.

MapboxNavigation/CarPlayNavigationViewController.swift
frederoni: > Is there a possibility that this method could be called twice

👍 that was the case. `insets.top > 0` is a reliable way with the current setup to get the state of the nav bar.

MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -138,6 +138,11 @@ public class CarPlayMapViewController: UIViewController {
         resetCamera(animated: false, altitude: CarPlayMapViewController.defaultAltitude)
     }
     
+    public override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        print(".")
```
1ec5: Remember to remove this print statement before merging.

frederoni:
Superseded by #2023 and #2022

PR #2021, opened 2019-03-12
CarPlay Continuity

JThramer:
Continuity!

Implements Deliverable 1 of #2017.

WIP. Will be updating this as polish is added.

To-do: 

- [x] Have the navigation service respect the simulation option of the CarPlay interface
- [x] Probably lots of other stuff too

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -211,6 +211,12 @@ public class CarPlayManager: NSObject {
         self.mapTemplateProvider.delegate = self
     }
     
+    /**
+     Programatically begins a carplay turn-by-turn navigation session.
+     
+     - parameter currentLocation: The current location of the user. This will be used to initally draw the current location icon.
+     - parameter navigationService: The service with which to navigation. CarPlayNavigationViewController will observe the progress updates from this service.
+     */
     public func beginNavigationWithCarPlay(using currentLocation: CLLocationCoordinate2D, navigationService: NavigationService) {
         let route = navigationService.route
         guard let destination = route.routeOptions.waypoints.last else {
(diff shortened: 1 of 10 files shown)


MapboxNavigation/CarPlayManager.swift
```diff
@@ -181,38 +210,31 @@ public class CarPlayManager: NSObject {
...
-        return formatter
-    }()
     
-    /**
-     The main map view displayed inside CarPlay.
-     */
-    @objc public var mapView: NavigationMapView? {
-        let mapViewController = carPlayMapViewController
-        return mapViewController?.mapView
+    public func beginNavigationWithCarPlay(using currentLocation: CLLocationCoordinate2D, navigationService: NavigationService) {
```
vincethecoder: @JThramer We should probably document this.

MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -63,11 +63,12 @@ public protocol CarPlayManagerDelegate {
...
      - parameter carPlayManager: The CarPlay manager instance.
      - parameter route: The route for which the returned route controller will manage location updates.
+     - parameter desiredSimulationMode: The desired simulation mode to use.
      - returns: A navigation service that manages location updates along `route`.
      */
     
-    @objc(carPlayManager:navigationServiceAlongRoute:)
-    func carPlayManager(_ carPlayManager: CarPlayManager, navigationServiceAlong route: Route) -> NavigationService
+    @objc(carPlayManager:navigationServiceAlongRoute:desiredSimulationMode:)
+    func carPlayManager(_ carPlayManager: CarPlayManager, navigationServiceAlong route: Route, desiredSimulationMode: SimulationMode) -> NavigationService
```
1ec5: Should we give `desiredSimulationMode` a default value or make it optional? This is a backwards-incompatible change as is, and it isn’t clear that the developer would know what to set it to.

1ec5: Why is the simulation mode “desired”? Is it that we may or may not honor the request, similar to how CLLocationManager reserves the right to be a contrarian?

PR #2022, opened 2019-03-13
Fix route preview in CarPlay

frederoni:
This PR attempts to fix the padding when previewing a route in CarPlay which wasn't always fully visible before.

There is at least one issue we need to fix before this PR can land.
To reproduce: Preview a route, go back to the initial screen. The route is selected according to `isOverviewingRoute` and accounted for in map view’s content inset, but the route is not drawn on the map. Should the route be selected or deselected?

cc @1ec5


1ec5:
> To reproduce: Preview a route, go back to the initial screen. The route is selected according to isOverviewingRoute and accounted for in map view’s content inset, but the route is not drawn on the map. Should the route be selected or deselected?

This issue only reproduces in the example application when returning to the main map from the favorites list. When returning from the search template, the map immediately tilts and enables user tracking mode, thanks to this special case:

https://github.com/mapbox/mapbox-navigation-ios/blob/2a75525dd4bcd42f44a136b319910eadc39e4c47/MapboxNavigation/CarPlayManager.swift#L333-L342

1ec5:
Tracking the case of returning to the main map in #2030.

MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -191,21 +191,26 @@ public class CarPlayMapViewController: UIViewController {
...
+        var edgePadding = view.safeArea
+        edgePadding += NavigationMapView.defaultPadding
         
-        guard isOverviewingRoutes else {
-            super.viewSafeAreaInsetsDidChange()
-            return
+        if let userCourseView = mapView.userCourseView {
+            let midX = userCourseView.bounds.midX
+            let midY = userCourseView.bounds.midY
+            edgePadding += UIEdgeInsets(top: midY, left: midX, bottom: midY, right: midX)
```
1ec5: This if block seems to be ensuring that the course view is always fully visible on the map. But about the waypoints? And if we’re previewing a route where the route line might extend in all directions from the user’s current location, then this alone won’t ensure that the whole route fits on the map. We have logic elsewhere about resetting or fitting the camera. Maybe we should reuse it instead of trying to reimplement some of it via edge insets.

JThramer:
[Demo video of new, fixed behavior](https://www.dropbox.com/s/awqsdyycfyywyfw/Carplay-zoom-demo.mov?dl=0)

1ec5:
That fixes #2030 as well, then.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -329,18 +329,23 @@ extension CarPlayManager: CPInterfaceControllerDelegate {
         }
     }
     public func templateWillDisappear(_ template: CPTemplate, animated: Bool) {
+        typealias TemplateBooleanClosure = (CPTemplate) -> Bool
+        guard let interface = interfaceController else { return }
+        
+        let listOrSearchOrMap = [CPSearchTemplate.self, CPListTemplate.self, CPMapTemplate.self]
+        let goingToSearch: TemplateBooleanClosure = { type(of: $0) == CPSearchTemplate.self }
```
1ec5: Nit: This closure is only used in one place; I think it would be fine to inline the equality check, as before.

MapboxNavigation/CarPlayManager.swift
```diff
@@ -329,18 +329,23 @@ extension CarPlayManager: CPInterfaceControllerDelegate {
         }
     }
     public func templateWillDisappear(_ template: CPTemplate, animated: Bool) {
+        typealias TemplateBooleanClosure = (CPTemplate) -> Bool
+        guard let interface = interfaceController else { return }
+        
+        let listOrSearchOrMap = [CPSearchTemplate.self, CPListTemplate.self, CPMapTemplate.self]
```
1ec5: What about CPGridTemplate or CPVoiceControlTemplate, in case the application has a need for one of those?

MapboxNavigation/CarPlayManager.swift
```diff
@@ -329,18 +329,23 @@ extension CarPlayManager: CPInterfaceControllerDelegate {
...
+        let goingToSearch: TemplateBooleanClosure = { type(of: $0) == CPSearchTemplate.self }
+        let currentlyOnHomeScreen = interface.templates.count == 1
+        let presentingFromFavOrSearchOrNav = currentlyOnHomeScreen && listOrSearchOrMap.contains { $0 == type(of: template) }
```
1ec5: This is a bit more indirection than necessary:

```swift
let templateIsForBrowsing = [CPSearchTemplate.self, CPListTemplate.self, CPMapTemplate.self, …].contains(type(of: template))
```

MapboxNavigation/CarPlayManager.swift
JThramer: Interesting, but what would we be excluding at that point?

1ec5: Do we need to exclude anything? I’m not sure what the point of this check is in the first place.

MapboxNavigation/CarPlayManager.swift
JThramer: `contains(_:)` does not seem to be available for `[CPTemplate.Type]`. I suspect this may be due to the fact that metatypes may not conform to `Equatable`? 🤔

MapboxNavigation/CarPlayManager.swift
JThramer: That makes two of us. Anyway, suggestion applied 👍

JThramer:
Behavior remains unchanged from demo video posted earlier.

PR #2026, opened 2019-03-14
Carplay Natural End Navigation

JThramer:
Fixing issue where CarPlay would not end navigation gracefully when phone was connected, due to an assumption made in code that was no-longer true. 

/cc @mapbox/navigation-ios

### DIFF
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -423,7 +423,6 @@ extension ViewController: NavigationViewControllerDelegate {
         
         //If we're not in a "Multiple Stops" demo, show the normal EORVC
         if navigationViewController.navigationService.router.routeProgress.isFinalLeg {
-            endCarplayNavigation(canceled: false)
             return true
         }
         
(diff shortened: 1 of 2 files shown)


Example/ViewController.swift
```diff
@@ -439,10 +439,14 @@ extension ViewController: NavigationViewControllerDelegate {
     // Called when the user hits the exit button.
     // If implemented, you are responsible for also dismissing the UI.
     func navigationViewControllerDidDismiss(_ navigationViewController: NavigationViewController, byCanceling canceled: Bool) {
+        endCarPlayNavigation(canceled: canceled)
+        navigationViewController.dismiss(animated: true, completion: nil)
+    }
+    
+    private func endCarPlayNavigation(canceled: Bool) {
         if #available(iOS 12.0, *), let delegate = UIApplication.shared.delegate as? AppDelegate {
             delegate.carPlayManager.currentNavigator?.exitNavigation(byCanceling: canceled)
```
akitchen: I'd like us to find a nicer way to do this longer term, but this works for now.

PR #2028, opened 2019-03-14
v0.30.0-beta.1

1ec5:
Upgraded the example applications to map SDK v4.9.0 and MapboxDirections.swift v0.27.2. Updated the Xcode projects, CocoaPods podspecs, readme, and changelog for v0.30.0-beta.1.

/cc @mapbox/navigation-ios


Bench/Info.plist
```diff
@@ -17,9 +17,9 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>0.29.1</string>
+	<string>0.30.0</string>
```
JThramer: shouldn't the build version string incorporate the beta identifier?

1ec5: I’ve never been completely sure about what we can technically include in the marketing version, but this key is documented to be in x.y.z format.

JThramer: No worries, just checking. As long as we're consistent.

MapboxCoreNavigation.podspec
```diff
@@ -3,7 +3,7 @@ Pod::Spec.new do |s|
   # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
   s.name = "MapboxCoreNavigation"
-  s.version = "0.29.1"
+  s.version = "0.30.0"
```
1ec5: Fixed the version number in the podspecs in efce37be1ef468699c66a04e1a528deea384784c, since CocoaPods is looking for specific tags based on this version string.

PR #2029, opened 2019-03-14
Fix progress updates order of execution

frederoni:
`updateSpokenInstructionProgress(status:willReRoute:)` must happen after step progress and leg progress has been updated to keep the indexes in sync.

cc @1ec5 @JThramer

### DIFF
--- a/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
+++ b/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
@@ -301,6 +301,24 @@ class MapboxCoreNavigationTests: XCTestCase {
         XCTAssertEqual(points[3].visualInstructionIndex, 0)
         XCTAssertEqual(points[3].spokenInstructionIndex, 0)
         XCTAssertEqual(points[3].type, .spoken)
+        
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -224,6 +224,85 @@ class MapboxCoreNavigationTests: XCTestCase {
...
+        XCTAssertEqual(points[1].type, .visual)
+        
+        XCTAssertEqual(points[2].legIndex, 0)
+        XCTAssertEqual(points[2].stepIndex, 0)
+        XCTAssertEqual(points[2].visualInstructionIndex, 0)
+        XCTAssertEqual(points[2].spokenInstructionIndex, 1)
+        XCTAssertEqual(points[2].type, .spoken)
+        
+        XCTAssertEqual(points[3].legIndex, 0)
+        XCTAssertEqual(points[3].stepIndex, 1)
```
frederoni: This assertion fails because stepIndex is 0, indicating that we are going backward in time, spokenInstructionIndex went from 1 to 0 when stepIndex remained on 0.

PR #2032, opened 2019-03-14
Add Japanese localization

1ec5:
Added a Japanese localization from Transifex, courtesy of [joshpa](https://www.transifex.com/user/profile/joshpa/).


1ec5:
Thanks for this localization, @joshandrews! (Didn’t know it was you until I saw Project-OSRM/osrm-text-instructions#276. 😊)

PR #2042, opened 2019-03-18
Marking Invalid Values as Omitted

kevinkreiser:
`speed` , `course` and `horizontalAccuracy` can all hold invalid values. `MBFixLocation` uses `nil` to communicate the omission of an optional value.

Interestingly an invalid `horizontalAccuracy` also means that the `CLLocationCoordinate2D` will also be bunk, but I've not taken any measures there as that is a required parameter by `MBFixLocation`

I've also made the changes in the other direction as well and marked omitted values from `MBFixLocation` as invalid in the constructed `CLLocation`

### DIFF
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -30,10 +30,10 @@ extension CLLocation {
     convenience init(_ location: MBFixLocation) {
         self.init(coordinate: location.coordinate,
                   altitude: location.altitude?.doubleValue ?? 0,
-                  horizontalAccuracy: location.accuracyHorizontal?.doubleValue ?? 0,
+                  horizontalAccuracy: location.accuracyHorizontal?.doubleValue ?? -1,
                   verticalAccuracy: 0,
-                  course: location.bearing?.doubleValue ?? 0,
-                  speed: location.speed?.doubleValue ?? 0,
+                  course: location.bearing?.doubleValue ?? -1,
+                  speed: location.speed?.doubleValue ?? -1,
                   timestamp: location.time)
     }
     
@@ -157,12 +157,5 @@ extension CLLocation {
     
     func shifted(to newTimestamp: Date) -> CLLocation {
         return CLLocation(coordinate: coordinate, altitude: altitude, horizontalAccuracy: horizontalAccuracy, verticalAccuracy: verticalAccuracy, course: course, speed: speed, timestamp: newTimestamp)
-    }
-    
-    convenience init(fixLocation: MBFixLocation) {
-        self.init(coordinate: fixLocation.coordinate, altitude: 0,
-                  horizontalAccuracy: fixLocation.accuracyHorizontal?.doubleValue ?? 0,
-                  verticalAccuracy: 0, course: fixLocation.bearing?.doubleValue ?? 0,
-                  speed: fixLocation.speed?.doubleValue ?? 0, timestamp: fixLocation.time)
     }
 }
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -158,11 +158,4 @@ extension CLLocation {
     func shifted(to newTimestamp: Date) -> CLLocation {
         return CLLocation(coordinate: coordinate, altitude: altitude, horizontalAccuracy: horizontalAccuracy, verticalAccuracy: verticalAccuracy, course: course, speed: speed, timestamp: newTimestamp)
     }
-    
-    convenience init(fixLocation: MBFixLocation) {
```
kevinkreiser: this appeared to be duplicated

MapboxCoreNavigation/CLLocation.swift
```diff
@@ -30,10 +30,10 @@ extension CLLocation {
     convenience init(_ location: MBFixLocation) {
         self.init(coordinate: location.coordinate,
                   altitude: location.altitude?.doubleValue ?? 0,
-                  horizontalAccuracy: location.accuracyHorizontal?.doubleValue ?? 0,
+                  horizontalAccuracy: location.accuracyHorizontal?.doubleValue ?? -1,
                   verticalAccuracy: 0,
-                  course: location.bearing?.doubleValue ?? 0,
-                  speed: location.speed?.doubleValue ?? 0,
+                  course: location.bearing?.doubleValue ?? -1,
+                  speed: location.speed?.doubleValue ?? -1,
```
kevinkreiser: when they are missing from `MBFixLocation` we mark as invalid in the `CLLocation`

MapboxCoreNavigation/MBNavigator.swift
```diff
@@ -7,10 +7,10 @@ extension MBFixLocation {
     convenience init(_ location: CLLocation) {
         self.init(coordinate: location.coordinate,
                   time: location.timestamp,
-                  speed: location.speed as NSNumber,
-                  bearing: location.course as NSNumber,
+                  speed: location.speed >= 0 ? location.speed as NSNumber : nil,
+                  bearing: location.course >= 0 ? location.course as NSNumber : nil,
                   altitude: location.altitude as NSNumber,
-                  accuracyHorizontal: location.horizontalAccuracy as NSNumber,
+                  accuracyHorizontal: location.horizontalAccuracy >= 0 ? location.horizontalAccuracy as NSNumber : nil,
```
kevinkreiser: when they are invalid in the `CLLocation` we mark them as omitted in the `MBFixLocation`

MapboxCoreNavigation/MBNavigator.swift
```diff
@@ -7,10 +7,10 @@ extension MBFixLocation {
     convenience init(_ location: CLLocation) {
         self.init(coordinate: location.coordinate,
                   time: location.timestamp,
-                  speed: location.speed as NSNumber,
-                  bearing: location.course as NSNumber,
+                  speed: location.speed >= 0 ? location.speed as NSNumber : nil,
+                  bearing: location.course >= 0 ? location.course as NSNumber : nil,
                   altitude: location.altitude as NSNumber,
```
frederoni: altitude is nullable but supports negative values to indicate below sea level so this conversion is correct

kevinkreiser: yep that was my thoughts exactly. we want to have correct altitudes when passing through death valley or doing u-boot routing ;)

Issue #2036, opened 2019-03-16
Ending navigation on phone causes pauses navigation begun on CarPlay


1ec5:
When an application is connected to CarPlay, beginning navigation on CarPlay and ending navigation on the phone causes the navigation activity on CarPlay to get stuck.

Steps to reproduce in the CarPlay example application:

1. Connect the phone to CarPlay.
1. Begin navigation on CarPlay. Navigation also begins on the phone.
1. End navigation on the phone. CarPlay remains in the navigation activity, but no location updates come in, so it seems to be frozen. Tapping End does exit out of the navigation activity.

<img src="https://user-images.githubusercontent.com/1231218/54479220-ebbbb980-47d7-11e9-8138-b491b1a75b07.PNG" width="250" alt="phone"> <img src="https://user-images.githubusercontent.com/1231218/54479241-0b52e200-47d8-11e9-9f26-0f3c7aee79c5.PNG" width="400" alt="carplay">

This build of the example application was not patched for #2034 or #2035, so the bug may or may not be related to those issues.

/cc @mapbox/navigation-ios

1ec5:
This issue would be fixed by #2045.

PR #2041, opened 2019-03-18
Car Play Quick Fixes: Data Race Bug

JThramer:
Data Race!

Fixes #2035 and #2034.

* Fixing race-condition bug by waiting until presentation completion to trigger carplay nav
* also changing trip-starting logic to not ask for a NavigationService if the user has already provided one

To-Do: 
- [x] Drive-test this to make sure things are fixed in the way we expect
- [x] Fix test

/cc @mapbox/navigation-ios

PR #2044, opened 2019-03-19
Ending Carplay on Arrival

JThramer:
Pic unrelated.

Like it says on the box. Possible fix for #2036.

 * Adding logic to end CarPlay navigation when we arrive at our final destination on the phone.

To-do: 
- [ ] CHANGELOG entry
- [ ] Tests?

### DIFF
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -412,6 +412,7 @@ extension ViewController: NavigationViewControllerDelegate {
         
         //If we're not in a "Multiple Stops" demo, show the normal EORVC
         if navigationViewController.navigationService.router.routeProgress.isFinalLeg {
+            endCarPlayNavigation(canceled: false)
             return true
         }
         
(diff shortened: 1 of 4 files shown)


Example/ViewController.swift
```diff
@@ -412,6 +412,7 @@ extension ViewController: NavigationViewControllerDelegate {
         
         //If we're not in a "Multiple Stops" demo, show the normal EORVC
         if navigationViewController.navigationService.router.routeProgress.isFinalLeg {
+            endCarPlayNavigation(canceled: false)
```
1ec5: Shouldn’t `navigationViewControllerDidDismiss(_:byCanceling:)` get called in this situation?

JThramer: [It does.](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/CarPlayNavigationViewController.swift#L189-L199)

1ec5:
Fixed in #2041.

1ec5:
(Sorry, wrong tab. 😅)

PR #2045, opened 2019-03-19
CarPlay-initiated navigation delegate fix

JThramer:
From the car.

This PR riffs off of #2044 and introduces even more fixes.

* Fixing issue where the NVC delegate was not set when NVC is created and presented by the CarPlayManager (when starting nav on CarPlay). This manifested as the CarPlay logic freezing when we command a cancel from the phone, because the logic triggering the dismissal of the `CarPlayNavigationViewController` was never called, due to the unset delegate.

~I'm not super happy with the mechanism, would appreciate any suggestions.~ (The mechanism has been improved, and we seem to be at a pretty decent place with this.)

/cc @mapbox/navigation-ios


Example/ViewController.swift
```diff
@@ -65,6 +65,24 @@ class ViewController: UIViewController {
...
+        super.init(nibName: nibNameOrNil, bundle: nibBundleOrNil)
+        setAppDelegateProperty()
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setAppDelegateProperty()
+    }
+    
+    private func setAppDelegateProperty() {
```
1ec5: This name focuses too much on “how” rather than “what” or “why”. Consider renaming it to `commonInit()`, which is a common pattern.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -335,7 +335,7 @@ open class NavigationViewController: UIViewController {
      If the `NavigationViewController` is already in the stack, it will open the `StepsViewController` unless it is already open.
      */
     @available(iOS 12.0, *)
-    public class func carPlayManager(_ carPlayManager: CarPlayManager, didBeginNavigationWith navigationService: NavigationService, window: UIWindow) {
+    public class func carPlayManager(_ carPlayManager: CarPlayManager, didBeginNavigationWith navigationService: NavigationService, window: UIWindow, delegate: NavigationViewControllerDelegate?) {
```
1ec5: We need to find a different name for the `delegate` argument. It makes it look like the caller should pass itself in to become the CarPlayManager’s (strongly held) delegate, which is incorrect on several levels. What role does the view controller play in this scenario?

JThramer: the `ViewController` is the delegate for the `NavigationViewController` in this scenario.

Example/ViewController.swift
```diff
@@ -430,10 +478,10 @@ extension ViewController: NavigationViewControllerDelegate {
     // If implemented, you are responsible for also dismissing the UI.
     func navigationViewControllerDidDismiss(_ navigationViewController: NavigationViewController, byCanceling canceled: Bool) {
         endCarPlayNavigation(canceled: canceled)
-        navigationViewController.dismiss(animated: true, completion: nil)
+        dismissActiveNavigationViewController()
     }
     
-    private func endCarPlayNavigation(canceled: Bool) {
+    func endCarPlayNavigation(canceled: Bool) {
```
JThramer: This should be moved up with the others (L321)

JThramer:
Minor issue: When starting nav from CarPlay, `ViewController.routes` should be set to the routes coming in from CarPlay. Otherwise old route you may have selected on example view controller may persist after navigation end.

Example/ViewController.swift
```diff
@@ -200,23 +220,40 @@ class ViewController: UIViewController {
...
+        nav.mapView?.delegate = self
+        return nav
+    }
+    
+    public func beginNavigationWithCarplay(navigationService: NavigationService) {
+        if activeNavigationViewController != nil {
+            activeNavigationViewController?.isUsedInConjunctionWithCarPlayWindow = true
+        } else {
+            let navigationViewController = createNavigationViewController(navigationService: navigationService)
+            presentAndRemoveMapview(navigationViewController, completion: nil)
```
vincethecoder: @JThramer should we be updating the property `isUsedInConjunctionWithCarPlayWindow` to `true` after we create a new a navigation view controller?

JThramer: Doing it this way makes sure that the StepsViewController is fully presented when the NVC presents, which is less visually jarring IMHO.

CHANGELOG.md
```diff
@@ -12,8 +12,10 @@
...
-* Fixing issue where data-race could unexpectedly cause a NavigationViewController to be created and presented on-screen.  ([#2041](https://github.com/mapbox/mapbox-navigation-ios/pull/2041)
+* Removed `NavigationViewController.carPlayManager(_:didBeginNavigationWith:window:)` that created and presented a `NavigationViewController`. Have your `NavigationViewControllerDelegate` (such as a `UIViewController` subclass, or a discrete delegate) create and present a `NavigationViewController`. ([#2045](https://github.com/mapbox/mapbox-navigation-ios/pull/2045))
+* Removed `NavigationViewController.carPlayManagerDidEndNaviation(_:window:)`. Have your `NavigationViewControllerDelegate` (such as a `UIViewController` subclass, or a discrete delegate) dismiss the active `NavigationViewController`. ([#2045](https://github.com/mapbox/mapbox-navigation-ios/pull/2045))
+* Fixing issue where a data-race could cause CarPlay to create it's own `NavigationService`, resulting in unexpected behavior. ([#2041](https://github.com/mapbox/mapbox-navigation-ios/pull/2041))
```
1ec5:
```suggestion
* Fixed an issue where `CarPlayManager` sometimes created a redundant `NavigationService` object, resulting in unexpected behavior. ([#2041](https://github.com/mapbox/mapbox-navigation-ios/pull/2041))
```

CHANGELOG.md
```diff
@@ -12,8 +12,10 @@
...
+* Fixing issue where data-race could unexpectedly cause a `NavigationViewController` to be created and presented on-screen.  ([#2041](https://github.com/mapbox/mapbox-navigation-ios/pull/2041))
```
1ec5:
```suggestion
* Fixed an issue where `CarPlayManager` sometimes created and presented a redundant `NavigationViewController`. ([#2041](https://github.com/mapbox/mapbox-navigation-ios/pull/2041))
```

Example/ViewController.swift
```diff
@@ -200,23 +220,42 @@ class ViewController: UIViewController {
...
+        nav.delegate = self
+        nav.mapView?.delegate = self
+        return nav
+    }
+    
+    public func beginNavigationWithCarplay(navigationService: NavigationService) {
+        self.routes = [navigationService.route]
+        
+        if activeNavigationViewController != nil {
+            activeNavigationViewController?.isUsedInConjunctionWithCarPlayWindow = true
```
1ec5: The optional here makes the nil check unnecessary.

Example/ViewController.swift
```diff
@@ -200,23 +220,42 @@ class ViewController: UIViewController {
...
         
-        presentAndRemoveMapview(navigationViewController)
+        presentAndRemoveMapview(navigationViewController, completion: beginCarPlayNavigation)
+    }
+    
+    func createNavigationViewController(navigationService: NavigationService) -> NavigationViewController {
+        let route = navigationService.route
+        let options = NavigationOptions( navigationService: navigationService)
+        
+        let nav = NavigationViewController(for: route, options: options)
```
1ec5: Spell out `navigationViewController`.

Example/ViewController.swift
```diff
@@ -200,23 +220,42 @@ class ViewController: UIViewController {
...
-        let options = NavigationOptions(styles: styles, navigationService: navigationService())
+        let options = NavigationOptions(styles: styles, navigationService: navigationService(route: route))
         let navigationViewController = NavigationViewController(for: route, options: options)
         navigationViewController.delegate = self
         
-        presentAndRemoveMapview(navigationViewController)
+        presentAndRemoveMapview(navigationViewController, completion: beginCarPlayNavigation)
+    }
+    
+    func createNavigationViewController(navigationService: NavigationService) -> NavigationViewController {
```
1ec5: Rename to `navigationViewController(navigationService:)`.

Example/ViewController.swift
```diff
@@ -241,35 +280,52 @@ class ViewController: UIViewController {
...
-        presentAndRemoveMapview(navigationViewController)
+        presentAndRemoveMapview(navigationViewController, completion: beginCarPlayNavigation)
     }
 
-    func navigationService() -> NavigationService? {
-        guard let route = routes?.first else { return nil }
+    func navigationService(route: Route) -> NavigationService {
         let simulate = simulationButton.isSelected
-        let mode: SimulationMode = simulate ? .always : .never//.onPoorGPS
+        let mode: SimulationMode = simulate ? .always : .onPoorGPS
```
1ec5: This fixes an issue where the example application didn’t automatically simulate movement within tunnels.

PR #2047, opened 2019-03-20
Sticky course view tracking and persistent content inset

### LINKED ISSUE #1731: UserCourseView should update instantaneously when user gestures cause camera changes

### LINKED ISSUE #1949: User course view lurches to the side when performing a maneuver
`NavigationMapView.shouldPositionCourseViewFrameByFrame` keeps the user course view (puck) constrained to the map on each frame, but it only ever seems to be enabled in response to external changes to the map view rather than internal changes due to location updates. When this property was originally added in #402, the idea was that the map would lasso the puck and force-synchronize it with the map on each frame, rather than allowing it to roam free using its normal 1-second animation, in response to location updates. At some point, this distinction was lost.

As a result of this omission, as the route takes a right turn, we send the puck to the right based on where the next location along the cross street is currently displayed on the map. That movement is animated with a duration of 1&nbsp;second. But at the same time, we kick off a 1-second-long camera animation that rotates the map to look down the cross street. The effect is that the puck appears to lurch to the right before swinging back to the center of the map to meet the route line. This effect is exacerbated by the route simulation feature’s speedup factor, if specified.

right

At a glance, the fix appears to be simple: we already correctly synchronize the puck to the map when `NavigationMapView.shouldPositionCourseViewFrameByFrame` is enabled, but we need to enable that property via `NavigationMapView.enableFrameByFrameCourseViewTracking(for:)` when approaching a maneuver. There’s some existing logic along these lines, but it only adjusts the map view’s frame rate rather than keeping `shouldPositionCourseViewFrameByFrame` enabled:

https://github.com/mapbox/mapbox-navigation-ios/blob/9e83261e2c768ae815dccae2caba0d9464ffb656/MapboxNavigation/NavigationMapView.swift#L273 https://github.com/mapbox/mapbox-navigation-ios/blob/9e83261e2c768ae815dccae2caba0d9464ffb656/MapboxNavigation/NavigationMapView.swift#L295-L296

/cc @mapbox/navigation-ios

### LINKED ISSUE #2030: CarPlay map fits to previewed route when returning to main map from custom template
When returning to the main CarPlay map from a route preview via a template other than the search template, the map doesn’t enter course tracking mode and in some cases doesn’t tilt. Without course tracking mode, the camera seems to fit to a route that is no longer being previewed.

unfit

To reproduce in the CarPlay example application:

1. Tap on the ⭐️ Favorites button in the navigation bar.
1. Select a favorite to preview.
1. Return to the favorites list.
1. Tap Back to return to the main map.
1. Wait until the navigation bar automatically hides.

The issue doesn’t reproduce when using the search template, because this code in CarPlayManager special-cases CPSearchTemplate:

https://github.com/mapbox/mapbox-navigation-ios/blob/2a75525dd4bcd42f44a136b319910eadc39e4c47/MapboxNavigation/CarPlayManager.swift#L333-L342

/ref #2022
/cc @mapbox/navigation-ios

frederoni:
_work in progress_

- [x] Fixes #1731 - UserCourseView should update instantaneously
- [x] Fixes #1949 - User course view lurches to the side
- [ ] ~~Fix #2030 CarPlay map fitting~~ (already fixed in #2022)

This PR brings back the course tracking similar to the default in MGLMapView where the puck and camera are synchronized which is really noticeable when simulating faster than 1×.

tracking

cc @1ec5 @JThramer

### DIFF
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -587,7 +587,8 @@ extension CarPlayManager: CPMapTemplateDelegate {
         topDownCamera.pitch = 0
         mapView.setCamera(topDownCamera, animated: false)
 
-        mapView.showcase([route])
+        let padding = NavigationMapView.defaultPadding + mapView.safeArea
+        mapView.showcase([route], padding: padding)
         
         delegate?.carPlayManager?(self, selectedPreviewFor: trip, using: routeChoice)
     }
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -587,8 +587,7 @@ extension CarPlayManager: CPMapTemplateDelegate {
         topDownCamera.pitch = 0
         mapView.setCamera(topDownCamera, animated: false)
 
-        let padding = NavigationMapView.defaultPadding + mapView.safeArea
-        mapView.showcase([route], padding: padding)
+        mapView.showcase([route])
         
         delegate?.carPlayManager?(self, selectedPreviewFor: trip, using: routeChoice)
     }
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -221,6 +221,10 @@ public class CarPlayNavigationViewController: UIViewController {
             return mapView?.tracksUserCourse ?? false
         }
         set {
+            defer {
+                mapView?.tracksUserCourse = newValue
+            }
+            
             let progress = navigationService.routeProgress
             if !tracksUserCourse && newValue {
                 mapView?.recenterMap()
@@ -232,23 +236,16 @@ public class CarPlayNavigationViewController: UIViewController {
                     let coordinates = progress.route.coordinates else {
                         return
                 }
-                mapView?.enableFrameByFrameCourseViewTracking(for: 3)
+                
+                let contentInsets = carPlayContentInsets(forOverviewing: true)
+                // TODO: The balancing constraints should probably be accounted for here
+                
                 mapView?.setOverheadCameraView(from: userLocation,
                                                along: coordinates,
-                                               contentInsets: contentInset(forOverviewing: false))
+                                               contentInsets: contentInsets,
+                                               animated: false)
             }
         }
-    }
-    
-    func contentInset(forOverviewing overviewing: Bool) -> UIEdgeInsets {
-        var contentInsets = view.safeArea
-        
-        if overviewing {
-            let padding = UIEdgeInsets(top: 40, left: 40, bottom: 40, right: 40)
-            contentInsets += padding
-        }
-        
-        return contentInsets
     }
     
     public func beginPanGesture() {
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -221,10 +221,6 @@ public class CarPlayNavigationViewController: UIViewController {
             return mapView?.tracksUserCourse ?? false
         }
         set {
-            defer {
-                mapView?.tracksUserCourse = newValue
-            }
-            
             let progress = navigationService.routeProgress
             if !tracksUserCourse && newValue {
                 mapView?.recenterMap()
@@ -232,18 +228,11 @@ public class CarPlayNavigationViewController: UIViewController {
                                  legIndex: progress.legIndex,
                                  stepIndex: progress.currentLegProgress.stepIndex + 1)
             } else if tracksUserCourse && !newValue {
-                guard let userLocation = self.navigationService.router.location?.coordinate,
-                    let coordinates = progress.route.coordinates else {
-                        return
+                guard let userLocation = self.navigationService.router.location?.coordinate else {
+                    return
                 }
-                
-                let contentInsets = carPlayContentInsets(forOverviewing: true)
-                // TODO: The balancing constraints should probably be accounted for here
-                
-                mapView?.setOverheadCameraView(from: userLocation,
-                                               along: coordinates,
-                                               contentInsets: contentInsets,
-                                               animated: false)
+                mapView?.enableFrameByFrameCourseViewTracking(for: 3)
+                mapView?.setOverheadCameraView(from: userLocation, along: progress.route.coordinates!, for: self.edgePadding)
             }
         }
     }
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -46,6 +46,8 @@ public class CarPlayNavigationViewController: UIViewController {
     var mapViewLeftSafeAreaBalancingConstraint: NSLayoutConstraint?
     var mapViewRightSafeAreaBalancingConstraint: NSLayoutConstraint?
     
+    var mapViewOverviewRightConstraint: NSLayoutConstraint?
+    
     var carSession: CPNavigationSession!
     var mapTemplate: CPMapTemplate
     var carFeedbackTemplate: CPGridTemplate!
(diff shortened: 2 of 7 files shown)


MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -222,15 +222,29 @@ public class CarPlayNavigationViewController: UIViewController {
...
             }
         }
     }
     
+    func contentInset(forOverviewing overviewing: Bool) -> UIEdgeInsets {
+        var contentInsets = view.safeArea
+        
+        if overviewing {
+            let padding = UIEdgeInsets(top: 40, left: 40, bottom: 40, right: 40)
+            contentInsets += padding
```
frederoni: Max MBRouteLineWidthByZoomLevel should be accounted for.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -331,15 +330,18 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            let point = userAnchorPoint
+            let padding = UIEdgeInsets(top: point.y, left: point.x, bottom: bounds.height - point.y, right: bounds.width - point.x)
+            // Omit padding when https://github.com/mapbox/mapbox-gl-native/pull/14081 has landed
             setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding, completionHandler: nil)
-        }
-        if !tracksUserCourse || userAnchorPoint != userCourseView?.center ?? userAnchorPoint {
-            UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
-                self.userCourseView?.center = self.convert(location.coordinate, toPointTo: self)
+            userCourseView?.center = userAnchorPoint
+        } else {
```
1ec5: This reprises part of #1732. I take it https://github.com/mapbox/mapbox-navigation-ios/pull/1732#discussion_r219017992 is no longer a concern because the puck’s on-screen position is a fixed value now?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -469,11 +469,28 @@ class RouteMapViewController: UIViewController {
...
+        
+        var insets = UIEdgeInsets(top: instructionBannerHeight,
+                                  left: 0,
+                                  bottom: bottomBannerHeight,
+                                  right: 0)
+        
+        if overviewing {
+            let courseView = mapView.userCourseView != nil ? mapView.userCourseView!.bounds.midX : 0
+            let routeLineWidths = MBRouteLineWidthByZoomLevel.compactMap { $0.value.constantValue as? Int }
+            let maxRouteLineWidth = CGFloat(routeLineWidths.max() ?? 0)
```
1ec5: The resulting zoom level is the only one that matters, right?

1ec5:
#2030 was already fixed in #2022.

1ec5:
Apparently this issue regresses #2030, but it should be straightforward to isolate and undo that change.

1ec5:
Current state of this branch when previewing a route to a nearby destination using a search result in search history:

preview

frederoni:
The CarPlay preview is reverted and should behave exactly as before

MapboxNavigation/CarPlayManager.swift
```diff
@@ -587,8 +587,7 @@ extension CarPlayManager: CPMapTemplateDelegate {
         topDownCamera.pitch = 0
         mapView.setCamera(topDownCamera, animated: false)
 
-        let padding = NavigationMapView.defaultPadding + mapView.safeArea
-        mapView.showcase([route], padding: padding)
```
frederoni: this should also be reverted

MapboxNavigation/NavigationMapView.swift
frederoni: Exactly. The position is fixed unless frame by frame tracking is enabled which is needed when content insets or device orientation changes.

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -175,12 +179,33 @@ public class CarPlayNavigationViewController: UIViewController {
...
+        
+        mapViewRightSafeAreaBalancingConstraint?.constant = -mapView.safeArea.right
+        mapViewLeftSafeAreaBalancingConstraint?.constant = mapView.safeArea.left
+        view.setNeedsUpdateConstraints()
+        
+        mapView.enableFrameByFrameCourseViewTracking(for: 1)
+        
+        if !tracksUserCourse {
+            mapView.setContentInset(overviewContentInsets(), animated: false)
+            mapView.fit(to: navigationService.route, facing: 0, padding: .zero, animated: false)
```
1ec5: Now if the user pans, disabling course tracking but without entering overview mode, the map will revert to course tracking once the navigation bar goes away. I guess that’s OK but somewhat unexpected.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1085,12 +1090,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
         }
         
-        let cam = self.camera
-        cam.pitch = 0
-        cam.heading = 0
+        let currentCamera = self.camera
+        currentCamera.pitch = 0
+        currentCamera.heading = 0
+        
+        let newCamera = camera(currentCamera, fitting: line, edgePadding: padding)
```
1ec5: This is equivalent to calling `cameraThatFits(_:direction:edgePadding:)`, since we’re forcing the pitch to be 0° anyways.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -469,11 +471,22 @@ class RouteMapViewController: UIViewController {
...
+        let bottomBannerHeight = bottomBannerContainerView.bounds.height
+        
+        var insets = UIEdgeInsets(top: instructionBannerHeight, left: 0,
+                                  bottom: bottomBannerHeight, right: 0)
+        
+        if overviewing {
+            insets += NavigationMapView.courseViewMinimumInsets
+            
+            let routeLineWidths = MBRouteLineWidthByZoomLevel.compactMap { $0.value.constantValue as? Int }
+            insets += UIEdgeInsets(floatLiteral: Double(routeLineWidths.max() ?? 0))
```
1ec5: 👍 As noted in https://github.com/mapbox/mapbox-navigation-ios/pull/2047#discussion_r267557733, this is rather conservative considering that only one zoom level is active at a given time. But we don‘t have ready access to the evaluated expression (unlike if we were using GL JS), and this method also gets called before calculating a camera rather than after, so choosing the right width isn’t worth the hassle. In practice, it would only be a few extra points on each side at most.

MapboxNavigation/CarPlayNavigationViewController.swift
frederoni: You're right. We probably need a flag `isOverviewing` here similar to RouteMapViewController.
Panning around is currently not possible either with course tracking on/off nor overviewing (navigation bar popping in triggers viewSafeAreaInsetsDidChange() which in turn, fits the camera to the route).

1ec5: @JThramer pointed out that this is actually pretty fortuitous. There isn’t a Resume or Recenter button like in NavigationViewController, and there’s no location tracking button like during the CarPlay browsing activity, so this code path is actually the most convenient way to get back into course tracking mode. Otherwise, the user has to tap the Overview button and tap the location tracking button after that.

JThramer: Disagree. This behavior is actually better than it was previously, as there is no resume button in CarPlay.

Issue #1217, opened 2018-03-12
Show compass on map


1ec5:
The map should show a compass indicating the current direction of travel or current heading (if the user is at rest). For example, if the user is facing south when starting a route, they should see “S” on the map so they know to turn around when the departure step says “Head north”.

This compass could have the same size as the map SDK’s built-in compass view, but it would serve a different purpose. The map SDK’s compass primarily indicates that the map is rotated and offers a way to rotate it back to north. By contrast, this proposal is for a view that bears nothing but a single letter.

There are precedents for this kind of display, both in Apple Maps and even built into rear-view mirrors in some car models. In Google Maps, a similar view serves as a toggle for overview mode. So it may be possible to implement this feature without sacrificing any additional screen real estate.

/cc @mapbox/navigation-ios

akitchen:
Let's get some design input on this before proceeding.

brsbl:
cc @Cjballard

Cjballard:
@1ec5 @brsbl @akitchen I have a few questions about the compass functionality.

Would the compass be ever-present? If so, would we want it to toggle a route overview functionality? In the last two examples, I replaced our current route overview button, but we could preserve that icon in the trip tray so the user has more of a direct CTA for that (like google).

Or would it only show up when a user is idle and before their first instruction like the screenshot example here? Curious to get some feedback here. These are pretty rough outlines but can attach something more hi-fi once we hone in on a solution. 

![compass](https://user-images.githubusercontent.com/7535952/40553334-4878286c-6010-11e8-8844-914113962736.png)

1ec5:
> Would the compass be ever-present?

Yes, that’s the idea anyways. I brought up the departure instruction as the primary use case for a compass, but it can be somewhat useful throughout the trip. If you know that you’re heading south on a vacation, seeing north on-screen is a bit of a red flag.

> If so, would we want it to toggle a route overview functionality?

We do want to retain some UI for toggling the route overview. However, it’s unclear whether we can effectively combine the direction indication with that toggling behavior into a single UI element.

Google’s overview toggle indicates the current direction with an analog design that I think is fairly difficult to use while driving. It spins clockwise as you turn the car counterclockwise and rotate the map counterclockwise – as it should. But there are no labels to use as a frame of reference for those contradictory motions. Also, despite one end of the double arrow being red, I personally don’t see enough contrast to avoid getting momentarily confused about which way it’s pointing.

[...]

PR #2051, opened 2019-03-22
Always hide compass on CarPlay during navigation

1ec5:
Always hide the compass on CarPlay during the navigation activity (in other words, in CarPlayNavigationViewController’s map view). Otherwise, the compass would appear in an odd position to the left of the column of map buttons. Also, the compass has an analog design more appropriate for the browsing activity but not for the navigation activity.

The analog compass remains in the browsing activity. A digital compass will be added as part of #1217 for parity with other CarPlay navigation applications.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -174,9 +174,6 @@ public class CarPlayNavigationViewController: UIViewController {
     public override func viewSafeAreaInsetsDidChange() {
         super.viewSafeAreaInsetsDidChange()
         
-        let navigationBarIsOpen = view.safeAreaInsets.top > 0
-        mapView?.compassView.isHidden = navigationBarIsOpen
-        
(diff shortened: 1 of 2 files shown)


MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -174,9 +174,6 @@ public class CarPlayNavigationViewController: UIViewController {
     public override func viewSafeAreaInsetsDidChange() {
         super.viewSafeAreaInsetsDidChange()
         
-        let navigationBarIsOpen = view.safeAreaInsets.top > 0
-        mapView?.compassView.isHidden = navigationBarIsOpen
```
akitchen: This is hidden by default?

akitchen: ... yes, https://github.com/mapbox/mapbox-navigation-ios/blob/9556c055d01d0fc7f8ad9205e96bf47d91d0f248/MapboxNavigation/CarPlayNavigationViewController.swift#L113

PR #2052, opened 2019-03-22
Remove unused list template code from CarPlayManager

1ec5:
This code was copied into the CarPlay example application code as part of #1717.

/cc @frederoni @vincethecoder


MapboxNavigation/CarPlayManager.swift
```diff
@@ -57,7 +57,6 @@ public class CarPlayManager: NSObject {
 
     public fileprivate(set) var mainMapTemplate: CPMapTemplate?
     public fileprivate(set) weak var currentNavigator: CarPlayNavigationViewController?
-    public static let CarPlayWaypointKey: String = "MBCarPlayWaypoint"
```
vincethecoder: I would move this key into `AppDelegate+CarPlay.swift`. There are a few references to this string in that file.

PR #2058, opened 2019-03-23
Upgrade to Swift 4.2

1ec5:
Ran all the Xcode project’s targets through the Swift 4.2 migrator, simplifying some types. Refreshed the CocoaPods integration test. Removed build setting overrides in the SDK’s podspecs that only affected development pods in the CocoaPods integration test.

Fixes #2057.

/cc @mapbox/navigation-ios @friedbunny

PR #2059, opened 2019-03-25
Declare supported Swift version in podspecs

1ec5:
#2058 was a bit overzealous in removing the [`swift_version`](https://guides.cocoapods.org/syntax/podspec.html#swift_versions) specifier instead of bumping it to 4.2, as this PR does.

### DIFF
--- a/MapboxCoreNavigation.podspec
+++ b/MapboxCoreNavigation.podspec
@@ -45,4 +45,6 @@ Pod::Spec.new do |s|
   s.dependency "MapboxMobileEvents", "~> 0.8.1"         # Always pin to a patch release if pre-1.0
   s.dependency "Turf", "~> 0.3.0"                       # Always pin to a patch release if pre-1.0
 
+  s.swift_version = "4.2"
+
 end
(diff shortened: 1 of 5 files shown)


flashspys:
fixes #2060

JThramer:
What about the swift_version check that was in the [original version](https://github.com/mapbox/mapbox-navigation-ios/blob/v0.29.1/MapboxNavigation.podspec#L51-L55) of this? do we no longer need it?

```ruby
  # `swift_version` was introduced in CocoaPods 1.4.0. Without this check, if a user were to
  # directly specify this podspec while using <1.4.0, ruby would throw an unknown method error.
  if s.respond_to?(:swift_version)
    s.swift_version = "4.0"
  end
```

1ec5:
At this point, I’m pretty sure CocoaPods v1.3._x_ would be unable to build a Swift 4.0 pod in Xcode 10, so I’m no longer concerned about this case.

friedbunny:
> What about the swift_version check that was in the original version of this? do we no longer need it?

As the person who added this, I would say it’s no longer necessary — it covered an edge-case for a now uncommon version of CocoaPods.

PR #2061, opened 2019-03-26
Audit Obj-C bridging

### LINKED ISSUE #1526: NavigationMapView delegates aren't exposed for objective-c
Mapbox Navigation SDK version: 0.18

In NavigationMapView the `navigationMapDelegate` and `courseTrackingDelegate` aren't exposed for objective-c.  The actual delegate methods are exposed but I am unable to implement the actual delegate.

frederoni:
Fixes #1526 

Expose RouteVoiceController.voiceControllerDelegate and NavigationMapView.navigationMapDelegate to Obj-C

@1ec5 @JThramer

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -49,7 +49,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     /**
      The object that acts as the navigation delegate of the map view.
      */
-    public weak var navigationMapDelegate: NavigationMapViewDelegate?
+    @objc public weak var navigationMapDelegate: NavigationMapViewDelegate?
     
     /**
      The object that acts as the course tracking delegate of the map view.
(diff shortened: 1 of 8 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -49,7 +49,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     /**
      The object that acts as the navigation delegate of the map view.
      */
-    public weak var navigationMapDelegate: NavigationMapViewDelegate?
+    @objc public weak var navigationMapDelegate: NavigationMapViewDelegate?
```
1ec5: Should we rename this property to `navigationMapViewDelegate` while we’re at it? The lack of “view” in the property name has always bugged me. If not, we can take care of it as tail work.

TestHelper/Fixture.swift
```diff
@@ -73,7 +73,7 @@ public class Fixture: NSObject {
         return locations.map { CLLocation($0) }
     }
     
-    public class func route(from jsonFile: String) -> Route {
+    @objc public class func route(from jsonFile: String) -> Route {
```
1ec5: The Objective-C selector name should be `-routeFromJSONFilePath:`.

frederoni: File path would indicate a full path or URL, but it's really just a name, a fixture bundled in the TestHelper framework. So `-routeFromJSONFileName:` is the most idiomatic I could come up with.

PR #2066, opened 2019-03-28
Fixed simulation not stopped when simulation mode is set to onPoorGPS bug

vincethecoder:
Modified navigation service to ensure we stop simulation for both `.always` and `.onPoorGPS` modes.

/cc @mapbox/navigation-ios


MapboxCoreNavigation/NavigationService.swift
```diff
@@ -342,7 +342,7 @@ public class MapboxNavigationService: NSObject, NavigationService, DefaultInterf
         nativeLocationSource.stopUpdatingHeading()
         nativeLocationSource.stopUpdatingLocation()
         
-        if simulationMode == .always {
+        if [.always, .onPoorGPS].contains(simulationMode) {
```
1ec5: Should we only stop on `onPoorGPS` if we’re actually simulating? Or is it safe to call `endSimulation()` regardless?

PR #2068, opened 2019-03-28
Document various Core Navigation classes

### LINKED ISSUE #2010: RouterDelegate and NavigationServiceDelegate are undocumented
Neither RouterDelegate nor NavigationServiceDelegate appear in jazzy documentation. RouterDelegate is explicitly marked as undocumented. (jazzy flags this class as being undocumented, despite the `:nodoc:`, because it’s in a standard comment instead of a documentation comment.) NavigationServiceDelegate has no documentation comment, so even its documented methods can’t appear in the docset.

https://github.com/mapbox/mapbox-navigation-ios/blob/8b153c03451c82c4320d4546c2b994b5d16a8b89/MapboxCoreNavigation/RouterDelegate.swift#L7-L8 https://github.com/mapbox/mapbox-navigation-ios/blob/8b153c03451c82c4320d4546c2b994b5d16a8b89/MapboxCoreNavigation/NavigationServiceDelegate.swift#L5

/cc @mapbox/navigation-ios

JThramer:
Documetation!

Fixes #2010.

To-Do:
- [x] `NavigationService` documentation
- [x] `Router` Documentation
- [x] `RouterDelegate` Documentation.
- [x] `NavigationServiceDelegate` documentation
- [ ] `RouterDataSource` documentation

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/NavigationServiceDelegate.swift
+++ b/MapboxCoreNavigation/NavigationServiceDelegate.swift
@@ -12,9 +12,9 @@ import MapboxDirections
  * Customize visual / spoken instructions that are presented to the user
  * Decide if battery monitoring should be disabled
  * Decide if the SDK should pause navigation when the user reaches an intermediate destination (for example, to show a interstitial modal UI)
- * Be told when the SDK fails to fetch a new route
- * Be told when the SDK is beginning or ending route-simulation
- * Be told when the user is about to arrive, or has arrived at their destination
+ * Be informed when the SDK fails to fetch a new route
+ * Be informed when the SDK is beginning or ending route-simulation
+ * Be informed when the user is about to arrive, or has arrived at their destination
  */
 @objc public protocol NavigationServiceDelegate {
     /**
@@ -82,7 +82,6 @@ import MapboxDirections
      - parameter location: the guaranteed location, possibly snapped, associated with the progress update.
      - parameter rawLocation: the raw location, from the location manager, associated with the progress update.
      */
-    
     @objc(navigationService:didUpdateProgress:withLocation:rawLocation:)
     optional func navigationService(_ service: NavigationService, didUpdate progress: RouteProgress, with location: CLLocation, rawLocation: CLLocation)
     
--- a/MapboxCoreNavigation/NavigationServiceDelegate.swift
+++ b/MapboxCoreNavigation/NavigationServiceDelegate.swift
@@ -14,7 +14,7 @@ import MapboxDirections
  * Decide if the SDK should pause navigation when the user reaches an intermediate destination (for example, to show a interstitial modal UI)
  * Be informed when the SDK fails to fetch a new route
  * Be informed when the SDK is beginning or ending route-simulation
- * Be informed when the user is about to arrive, or has arrived at their destination
+ * Be informed when the user is approaching or has arrived at their destination
  */
 @objc public protocol NavigationServiceDelegate {
     /**
--- a/MapboxCoreNavigation/NavigationServiceDelegate.swift
+++ b/MapboxCoreNavigation/NavigationServiceDelegate.swift
@@ -2,6 +2,20 @@ import Foundation
 import CoreLocation
 import MapboxDirections
 
+/**
+ NavigationServiceDelegate is the main delegation protocol in MapboxCoreNavigation. It's purpose is to delegate the operational parameters of the core navigation engine. See discussion for use-case examples.
+
+ You can use this protocol to:
+ * Recieve navigation progress updates
+ * Decide when to reroute the user
+ * Decide what location updates are considered 'qualified'
+ * Customize visual / spoken instructions that are presented to the user
+ * Decide if battery monitoring should be disabled
+ * Decide if the SDK should pause navigation when the user reaches an intermediate destination (for example, to show a interstitial modal UI)
+ * Be told when the SDK fails to fetch a new route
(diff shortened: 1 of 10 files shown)


1ec5:
RouterDelegate also needs to be documented in order to fix #2010.

JThramer:
@1ec5 `RouterDelegate` is not documented by intention, because it's basically a pass-through protocol to the public `NavigationServiceDelegate` facade.

We could document it, but that documentation would literally be a facsimile of `NavigationServiceDelegate`.

What would you recommend?

1ec5:
Oh, Router too.

The documentation should direct people to NavigationServiceDelegate, then. Otherwise, there’s no way to get to the NavigationServiceDelegate documentation from NavigationService.

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -2,6 +2,20 @@ import Foundation
...
+ NavigationServiceDelegate is the main delegation protocol in MapboxCoreNavigation. It's purpose is to delegate the operational parameters of the core navigation engine. See discussion for use-case examples.
+
+ You can use this protocol to:
+ * Recieve navigation progress updates
+ * Decide when to reroute the user
+ * Decide what location updates are considered 'qualified'
+ * Customize visual / spoken instructions that are presented to the user
+ * Decide if battery monitoring should be disabled
+ * Decide if the SDK should pause navigation when the user reaches an intermediate destination (for example, to show a interstitial modal UI)
+ * Be told when the SDK fails to fetch a new route
```
vincethecoder: nit: Be `informed`... rather than be `told`... ?

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -2,6 +2,20 @@ import Foundation
...
+ * Be told when the SDK is beginning or ending route-simulation
+ * Be told when the user is about to arrive, or has arrived at their destination
```
vincethecoder: nit: `Be informed when the user is approaching or has approached their destination` 

Almost identical to the former bullet point... maybe?

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -2,6 +2,20 @@ import Foundation
...
+ You can use this protocol to:
+ * Recieve navigation progress updates
+ * Decide when to reroute the user
+ * Decide what location updates are considered 'qualified'
+ * Customize visual / spoken instructions that are presented to the user
+ * Decide if battery monitoring should be disabled
+ * Decide if the SDK should pause navigation when the user reaches an intermediate destination (for example, to show a interstitial modal UI)
+ * Be informed when the SDK fails to fetch a new route
+ * Be informed when the SDK is beginning or ending route-simulation
+ * Be informed when the user is approaching or has arrived at their destination
```
1ec5: This is a long list. A developer might as well look at the available methods to discover all this functionality. Meanwhile, it could be summarized here as:

> A navigation service delegate interacts with one or more `NavigationService` instances, such as `MapboxNavigationService` objects, during turn-by-turn navigation. This protocol is the main way that your application can synchronize its state with the SDK’s location-related functionality. Each of the protocol’s methods is optional.
> 
> As the user progresses along a route, a navigation service informs its delegate about significant events as they occur, and the delegate has opportunities to influence the route and its presentation. For example, when the navigation service reports that the user has arrived at the destination, your delegate implementation could present information about the destination. It could also customize individual visual or spoken instructions along the route by returning modified instruction objects.
> 
> Assign a `NavigationServiceDelegate` instance to the `NavigationService.delegate` property of the navigation service before you start the service.

MapboxCoreNavigation/Router.swift
```diff
@@ -4,6 +4,10 @@ import MapboxDirections
 
 @objc (MBRouterDataSource)
```
1ec5: This class is undocumented, so the property documentation below will go unseen.

MapboxCoreNavigation/RouterDelegate.swift
```diff
@@ -4,43 +4,156 @@ import MapboxDirections
 
 
 
-//:nodoc: See NavigationServiceDelegate.
+/**
+ RouterDelegate is a mid-level delegate that defines the messaging architechture between the NavigationService and its router. Developers generally shouldn't have to concern themselves with this class unless a custom core-navigation engine (conforming to NavigationService) is being written.
+ 
+ You can use this protocol to:
```
1ec5: This list contradicts the paragraph above it: either developers should ignore this class or they should consider it for all the things below. Here’s a way to phrase it that also avoids discussing the SDK’s architecture, which shouldn’t be the concern of public, class-level documentation:

> A router delegate interacts with one or more `Router` instances, such as `RouteController` objects, during turn-by-turn navigation. This protocol is similar to `NavigationServiceDelegate`, and the main conforming class is `MapboxNavigationService`. Normally you should not need to make a class conform to the `RouterDelegate` protocol or call any of its methods directly, but you would need to call this protocol’s methods if you implement a custom `Router` class.
> 
> As the user progresses along a route, a router informs its delegate about significant events as they occur, and the delegate has opportunities to influence the route and its presentation. For example, when the navigation service reports that the user has arrived at the destination, your delegate implementation could present information about the destination. It could also customize individual visual or spoken instructions along the route by returning modified instruction objects.

PR #2071, opened 2019-04-02
Re-enable line distance metrics

### LINKED ISSUE #2067: Reenable line distance metrics on route line shape source
#1886 disabled line distance metrics on the route line shape source to work around the crash in mapbox/mapbox-gl-native#13378. That crash was fixed in map SDK v4.7.0, so we can reenable it to facilitate gradient route lines.

/ref #1799
/cc @mapbox/navigation-ios

frederoni:
Fixes #2067 Re-enable line distance metrics on the route line source.

_this PR doesn't add a gradient, it merely enables adding a gradient._

cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -467,8 +467,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             source.shape = polylines
             sourceSimplified.shape = mainPolylineSimplified
         } else {
-            let lineSource = MGLShapeSource(identifier: sourceIdentifier, shape: polylines, options: nil)
-            let lineCasingSource = MGLShapeSource(identifier: sourceCasingIdentifier, shape: mainPolylineSimplified, options: nil)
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -462,7 +462,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             source.shape = polylines
             sourceSimplified.shape = mainPolylineSimplified
         } else {
-            let lineSource = MGLShapeSource(identifier: sourceIdentifier, shape: polylines, options: nil)
+            let lineSource = MGLShapeSource(identifier: sourceIdentifier, shape: polylines, options: [.lineDistanceMetrics: true])
             let lineCasingSource = MGLShapeSource(identifier: sourceCasingIdentifier, shape: mainPolylineSimplified, options: nil)
```
frederoni: should we enable line distance metrics on the casing source as well? 🤔

1ec5: Sure, why not? 👨‍🎨

PR #2076, opened 2019-04-04
Delete conflicting files to fix jazzy docset generation

### LINKED ISSUE #1795: Declarations, parameters missing in jazzy documentation
The published jazzy documentation only contains abstracts, no declarations or parameter listings. As a result, permalinks to individual methods are also broken, and overloaded methods are difficult to distinguish from each other.

This issue reproduces in the documentation for [v0.21.0](https://www.mapbox.com/mapbox-navigation-ios/navigation/0.21.0/Protocols/NavigationMapViewDelegate.html#/c:@M@MapboxNavigation@objc%28pl%29MBNavigationMapViewDelegate%28im%29navigationMapView:shapeForRoutes:) and above but not for [v0.20.1](https://www.mapbox.com/mapbox-navigation-ios/navigation/0.20.1/Protocols/NavigationMapViewDelegate.html#/c:@M@MapboxNavigation@objc%28pl%29MBNavigationMapViewDelegate%28im%29navigationMapView:shapeForRoutes:) and below. In v0.21.0 and above, we had to follow the convoluted steps in https://github.com/mapbox/mapbox-navigation-ios/pull/1714#issuecomment-422038913 to publish the docset in the first place, so it’s possible that something along the way is preventing jazzy from matching USRs to the source code. The problem might even be the workaround for realm/jazzy#11 that locally blows away `@availability` attributes, but that might be a red herring.

/cc @mapbox/navigation-ios @mapbox/docs

1ec5:
#2013 duplicated Date.swift across the MapboxCoreNavigation and MapboxNavigation targets, so there were two Date.swift files in the merged, CocoaPods-generated target that jazzy uses to calculate USRs. A similar naming collision occurred with String.swift and Route.swift.

Fixes #1795 a second time.

/cc @mapbox/navigation-ios


MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -1977,6 +1971,7 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
```
frederoni: Xcode 10.2 replaced `English` with `en`. I wonder if this can cause any issues. 🤔

1ec5: Xcode 10.2 adds this entry automatically upon opening the project. I’m not sure why, given that `en` is already in the list and we don’t have any English.lproj folders in the repository. But the project functions fine with or without this change.

1ec5:
#2078 tracks adding a CI step so the docset doesn’t regress again.

MapboxNavigation.xcodeproj/project.pbxproj
1ec5: Huh, today Xcode 10.2 warned me about this entry and offered to remove it. I guess it was having a different mood today. Removed in a016c58a0e7bc4d7d3d53aeaca2a09f474a92840.

PR #2077, opened 2019-04-04
CarPlay compass view

frederoni:
This PR adds a `CarPlayCompassView` that indicates the direction the vehicle is traveling towards.
It snaps the course to eighth cardinal directions in steps of 45° (["N", "NE", "E", "SE", "S", "SW", "W", "NW"])

The CarPlayCompassView is only meant for CarPlay, and it is hidden by default.

cc @mapbox/navigation-ios


1ec5:
> We could also snap the course so that only eight different directions would be present.

This is a reasonable workaround for mapbox/mapbox-gl-native#6872. The style is correct, but the granularity is too high.

MapboxNavigation/CarPayCompassView.swift
```diff
@@ -0,0 +1,46 @@
+import UIKit
+import Mapbox
+
+@objc(MBCarPlayCompassView)
+open class CarPlayCompassView: FloatingButton {
```
1ec5: A circular button on the map may look too much like a tappable button. I suppose that’s why other navigation applications give the compass a rounded rectangle shape, similar to the panels that are overlaid atop the map.

frederoni: Absolutely agree. This should probably be a StylableView or a plain UIView.

MapboxNavigation/CarPayCompassView.swift
```diff
@@ -0,0 +1,68 @@
+import UIKit
+import Mapbox
+import Turf
+import MapboxCoreNavigation
+
+@objc(MBCarPlayCompassView)
+open class CarPlayCompassView: StylableView {
```
1ec5: This class and its `open` members need documentation.

MapboxNavigation/CarPayCompassView.swift
```diff
@@ -0,0 +1,68 @@
...
+    func commonInit() {
+        isHidden = true
+        let label = StylableLabel(frame: .zero)
+        label.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(label)
+        self.label = label
+        
+        course = 0
+        
+        widthAnchor.constraint(equalToConstant: 30).isActive = true
```
1ec5: Is this wide enough to accommodate a four-character-long string?

https://github.com/mapbox/mapbox-gl-native/blob/5b28c1eb9c1bf58bb2ab050d0e52aa7f96e8eef8/platform/darwin/resources/uk.lproj/Foundation.strings#L128
https://github.com/mapbox/mapbox-gl-native/blob/5b28c1eb9c1bf58bb2ab050d0e52aa7f96e8eef8/platform/darwin/resources/he.lproj/Foundation.strings#L128

Maybe the label’s font size needs to shrink if it doesn’t fit.

frederoni: It was a bit out of bounds and it did truncate without font scaling, but it's fixed now.

PR #2092, opened 2019-04-09
Do not prevent location snapping if location course is invalid.

Woildan:
Hello again,

I'm in the process of reducing the number of patches we maintain in our fork.
This one is a bit old so I can't provide as much debug info I'd like to showcase the problem.

It makes sense when reading the code, but for more context, I believe it was causing a bug where the puck would keep "looking up" when our app (not supporting CarPlay) was plugged to a (specific?) CarPlay system.

Basically it doesn't make sense saying we're not gonna snap a location and use it as-is when this location doesn't have a valid course.

Especially, when using `LegacyRouteController`, if there are (unluckily or because of the CarPlay system) no snapped location and both location's `course` and `trueHeading` are not qualified, then we end up displaying the raw location and the puck "looks up".

Edit: Found more info, I confirm that the problem was that thew puck was snapped when user speed was low, and became un-snapped but looking up when speed was high enough.


Woildan:
Could anyone from Mapbox have a look at this?
I'm open to feedback and this actually solves a bug :)

frederoni:
You're right. Using a previous invalid course to calculate the difference would often unsnap incorrectly.
Would you mind adding an entry to the changelog?

Woildan:
@frederoni Changelog entry added + rebased on top of master.

Issue #375, opened 2017-07-12
Switch fractions in distance label to decimals


ericrwolfe:
In practical use, it's extremely difficult to differentiate ¼, ½, and ¾ while driving. We should switch to 1 decimal of precision in the turn banner < 2 miles, and whole numbers over 3 miles.

Voice announcements should continue to use fractions for imperial units, rounded digits for metric.

cc @frederoni @bsudekum

PR #2093, opened 2019-04-09
Always reset rerouting flag even if request fails.

Woildan:
In `LegacyRouteController`, if the re-routing request fails the `isRerouting` flag isn't reset as it is the case in `RouteController` for off-route rerouting and `Router` for proactive rerouting.

If that happens then no subsequent request can be sent by `LegacyRouterController` and the user becomes stuck with an outdated and irrelevant route.

### DIFF
--- a/MapboxCoreNavigation/LegacyRouteController.swift
+++ b/MapboxCoreNavigation/LegacyRouteController.swift
@@ -364,6 +364,7 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
                 return
             }
 
+			strongSelf.isRerouting = false
             if let error = error {
                 strongSelf.delegate?.router?(strongSelf, didFailToRerouteWith: error)
                 NotificationCenter.default.post(name: .routeControllerDidFailToReroute, object: self, userInfo: [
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -373,7 +375,6 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
             }
```
vincethecoder: @Woildan semantically, the solution should go right about the guard for checking the value of `route`. Right above [line #375](https://github.com/mapbox/mapbox-navigation-ios/pull/2093/files#diff-ac5494b213aa895d04f334c942cbf29cL375). We want to respect the error response returned after the `getDirections` call.

Solution should look like this:
```swift
strongSelf.isRerouting = false
guard let route = route else { return }
```

vincethecoder: @Woildan I understand your concern. Well, since the `isRerouting` isn't getting updated, in that case when an error occurs, we should set that value on [line # 366](https://github.com/mapbox/mapbox-navigation-ios/pull/2093/files#diff-ac5494b213aa895d04f334c942cbf29cL366) before inspecting the `error`. That addresses the `route` issue... would you agree?

Woildan: Okay I've made the change.

Woildan:
Hello @vincethecoder, If you're ok with this patch could you merge in master? Thanks!

PR #2096, opened 2019-04-09
Update simulator distance when rerouting occurs.

### LINKED ISSUE #2095: Location simulator currentDistance not properly updated when rerouting occurs.
**Mapbox Navigation SDK version: v0.30.0**

When a reroute occurs while the simulation mode is active, the user is suddenly snapped further away on the route than he was at the rerouting time, even if the received route correctly starts at the rerouting coordinates.

This may be difficult to visualise in your Example app but we have a bunch of debug features in our app that allow us to trigger a proactive rerouting on demand so I've made some videos:

Without patch :

without_fix

With patch :

with_fix

It's problematic only when `SimulatedLocationManager` is active, but since the same object is used for both navigation simulation and bad GPS reception and since we do a (customised) proactive rerouting every 120 seconds, this can happen while in a tunnel with 3G reception but no GPS and so be visible to the user.

Our proactive-rerouting is customised in the sense that our server returns a route with the same polyline as originally requested when launching the navigation, in addition of potentially a faster route alternative. The route with "locked path" is applied immediately since maneuver matches, and we get updated ETA and trafic (visible in second gif). Long story short, we can end up in the case described above easily enough.

Note: this bug has been here for quite a while but better reporting it later than never...

Woildan:
Fixes #2095 

I don't know if there would be a better way of fixing this issue, but this works perfectly for our needs.


frederoni:
> I don't know if there would be a better way of fixing this issue

The fix looks good. Rerouting while simulating is kind of an edge case since the simulated location manager can't trigger an off-route, however, proactive rerouting can, or in case of manually updating the route.

Would you mind adding an entry in the changelog?

Woildan:
@frederoni Changelog entry added + branch rebased on top of master.

PR #2100, opened 2019-04-11
Update CarPlay tracking icon when tracking state changes

### LINKED ISSUE #2097: Overview button should change to recenter icon when panning out of course tracking in CarPlay
When panning the map in CarPlay to disable user tracking mode, the Overview button should change to the arrow icon to match its actual behavior or recentering the map and enabling user tracking mode.

The main browsing map template does show and hide the overview button after panning, but the navigation map template does not. All it does is disable user tracking.

https://github.com/mapbox/mapbox-navigation-ios/blob/4274aced2d02de0d5f416d416aaab3b228d1257b/MapboxNavigation/CarPlayManager.swift#L607-L609 https://github.com/mapbox/mapbox-navigation-ios/blob/4274aced2d02de0d5f416d416aaab3b228d1257b/MapboxNavigation/CarPlayManager.swift#L613-L615

/ref #2048
/cc @mapbox/navigation-ios

frederoni:
Fixes #2097 

cc @1ec5

### DIFF
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -55,8 +55,23 @@ public class CarPlayManager: NSObject {
         }
     }
 
+    var trackingStateObservation: NSKeyValueObservation?
+    
+    deinit {
+        trackingStateObservation = nil
+    }
+    
     public fileprivate(set) var mainMapTemplate: CPMapTemplate?
-    public fileprivate(set) weak var currentNavigator: CarPlayNavigationViewController?
+    public fileprivate(set) weak var currentNavigator: CarPlayNavigationViewController? {
+        didSet {
+            if let controller = currentNavigator {
+                trackingStateObservation = controller.observe(\.tracksUserCourse) { [weak self] (controller, change) in
+                    let imageName = controller.tracksUserCourse ? "carplay_overview" : "carplay_locate"
+                    self?.userTrackingButton.image = UIImage(named: imageName, in: .mapboxNavigation, compatibleWith: nil)
+                }
+            }
+        }
+    }
 
     internal var mapTemplateProvider: MapTemplateProvider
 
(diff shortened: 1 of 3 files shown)


MapboxNavigation/CarPlayManager.swift
```diff
@@ -142,20 +153,24 @@ public class CarPlayManager: NSObject {
     /**
      The bar button that shows the selected route overview on the map.
      */
-    @objc public lazy var overviewButton: CPMapButton = {
-        let overviewButton = CPMapButton { button in
+    @objc public lazy var trackingStateButton: CPMapButton = {
```
1ec5: Maybe call it `userTrackingButton` by analogy with MapKit’s MKUserTrackingButton.

/ref mapbox/mapbox-gl-native#1093

MapboxNavigation/CarPlayManager.swift
```diff
@@ -55,8 +55,19 @@ public class CarPlayManager: NSObject {
...
     }
 
+    var trackingStateObservation: NSKeyValueObservation?
+    
     public fileprivate(set) var mainMapTemplate: CPMapTemplate?
-    public fileprivate(set) weak var currentNavigator: CarPlayNavigationViewController?
+    public fileprivate(set) weak var currentNavigator: CarPlayNavigationViewController? {
+        didSet {
+            if let controller = currentNavigator {
+                trackingStateObservation = controller.observe(\.tracksUserCourse) { [weak self] (controller, change) in
```
1ec5: Do we need to invalidate the NSKeyValueObservation at some point, at least on older iOS versions?

frederoni: Good catch. It would leak on iOS <= 10

MapboxNavigation/CarPlayManager.swift
```diff
@@ -77,6 +92,10 @@ public class CarPlayManager: NSObject {
      */
     @objc public let directions: Directions
     
+    deinit {
```
1ec5: This is redundant to the `deinit` above, causing builds to fail:

```
Invalid redeclaration of 'deinit'
```

PR #2108, opened 2019-04-12
Fix CarPlay documentation warnings

frederoni:
Fixes "Parameter 'template' not found in the function declaration Did you mean 'template_'?"

`template` seems to be a reserved keyword in the generated Obj-C header and
is being renamed to template_ so there is a mismatch in the
documentation parameters.

cc @mapbox/navigation-ios


frederoni:
> This is OK as a workaround, but it’s worth filing a bug report to Apple about.

The bug report is filed.

PR #2110, opened 2019-04-18
[WIP] Fix route preview broken on first render in CarPlay

vincethecoder:
Background: When a device is connected to CarPlay, the route preview on first render displays on the head unit display. (Note - the simulator doesn't exhibit this buggy experience.) 

Issues addressed:
- [ ] The `CarPlayMapViewController`'s map view wasn't updated after a successful `MapboxDirections` service request. 
- [ ] The race condition of fitting the camera prematurely before the camera pitch was set has been addressed. 
- [ ] Mapview camera zooming into the wrong location coordinates after directions route(s) for preview is returned by the `MapboxDirections` service request.

/cc @1ec5 @mapbox/navigation-ios


S2Ler:
@MaximAlien Can you please verify if this issue still relevant in v2.0 and close the PR if not reproducible.

MaximAlien:
I do not think that PR is relevant for `2.x` as camera behavior was fully changed on both Navigation SDK and Maps SDK sides.

PR #2112, opened 2019-04-22
Add speed limit data returned by MapboxDirections.swift to the RouteProgress object

avi-c:
Incorporate the new maxspeed annotations on the RouteLeg object from MapboxDirections.swift. Provide this data to Navigation SDK clients via the maximumSpeedLimitsByLeg array on the RouteProgress object.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,7 +1,5 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.1.1
-# github "https://github.com/avi-c/mapbox/MapboxDirections.swift.git"
-# git "https://github.com/avi-c/MapboxDirections.swift" "maxspeedAnnotations"
 github "mapbox/MapboxDirections.swift" "maxspeedAnnotations"
 github "mapbox/turf-swift" ~> 0.3
 github "mapbox/mapbox-events-ios" ~> 0.8.1
--- a/Cartfile
+++ b/Cartfile
@@ -1,6 +1,8 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.1.1
-github "mapbox/MapboxDirections.swift" ~> 0.27.3
+# github "https://github.com/avi-c/mapbox/MapboxDirections.swift.git"
+# git "https://github.com/avi-c/MapboxDirections.swift" "maxspeedAnnotations"
+github "mapbox/MapboxDirections.swift" "maxspeedAnnotations"
 github "mapbox/turf-swift" ~> 0.3
 github "mapbox/mapbox-events-ios" ~> 0.8.1
 github "ceeK/Solar" ~> 2.1.0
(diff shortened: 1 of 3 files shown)


Cartfile
```diff
@@ -1,6 +1,6 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.1.1
-github "mapbox/MapboxDirections.swift" ~> 0.27.3
+github "mapbox/MapboxDirections.swift" "maxspeedAnnotations"
```
1ec5: This PR depends on mapbox/MapboxDirections.swift#367. Once that PR lands, we’ll need to change this dependency to `master`. Finally, once we publish a new release of MapboxDirections.swift, we’d change this line and corresponding lines in the podspecs to require the new release.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -217,9 +223,43 @@ open class RouteProgress: NSObject {
             }
 
             congestionTravelTimesSegmentsByStep.append(congestionTravelTimesSegmentsByLeg)
+
+            maneuverCoordinateIndex = 0
+            if let segmentMaximumSpeedLimits = leg.segmentMaximumSpeedLimits {
+                for step in leg.steps {
```
1ec5: Consolidate this loop with the loop above that populates the `segmentCongestionLevels` and `expectedSegmentTravelTimes` properties for better performance. We’ll need to refactor a bit of the code, so that the absence of traffic congestion data doesn’t prevent us from processing speed limit data and vice versa.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -217,9 +223,43 @@ open class RouteProgress: NSObject {
...
+
+            maximumSpeedLimitsByLeg.append(maximumSpeedLimitsByStep)
         }
     }
 
+    /**
+     Returns the SpeedLimit for the current position along the route. Returns SpeedLimit.invalid if the speed limit is unknown or missing.
+     */
+    public var currentSpeedLimit: SpeedLimit {
+        let coordinatesLeftOnStepCount = Int(floor((Double(currentLegProgress.currentStepProgress.step.coordinateCount)) * currentLegProgress.currentStepProgress.fractionTraveled))
```
1ec5: This is the same approach taken by #1242. Unfortunately, it assumes that each coordinate is evenly spaced along the step, which is almost never the case. For our purposes, we should assume that the segments are of arbitrary length.

With traffic congestion, it made sense to store congestion levels by segment, because we use those segments to create a shape layer all at once. Displaying the right speed limit at the right time is a very similar problem to presenting the right spoken or visual instruction at the right time, as RouteController and LegacyRouteController do. As you loop over the attributes, set each speed limit to the cumulative distance along the step. Then this computed property can become:

```swift
public var currentSpeedLimit: SpeedLimit {
    …
    let speedLimits = maximumSpeedLimitsByLeg[legIndex][currentLegProgress.stepIndex]
    return speedLimits.last { $0.distance <= distanceTraveled } ?? .invalid
}
```

avi-c: @1ec5 - Are you suggesting that I add a `distance` property to the SpeedLimit object or to keep a collection of (SpeedLimit, distance) pairs that is private to RouteProgress, built in the init() and referenced by currentSpeedLimit()?

1ec5: The latter, a collection of `(SpeedLimit, distance)` pairs. If the additional nesting turns out to be problematic, then the tuples could be grouped by leg and the distances could be measured from the start of the leg, but the tradeoff would be worse performance during the trip.

1ec5:
The saga continues in #2114.

PR #2114, opened 2019-04-24
Add speed limit data returned by MapboxDirections.swift to the RouteProgress object

avi-c:
Shifting from PR #2112 

Incorporate the new maxspeed annotations on the RouteLeg object from MapboxDirections.swift. Provide this data to Navigation SDK clients via the ~~maximumSpeedLimitsByLeg array on the RouteProgress object~~ `RouteLegProgress.currentSpeedLimit` property.


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -171,6 +171,13 @@ open class RouteProgress: NSObject {
      */
     public var congestionTimesPerStep: [[[CongestionLevel: TimeInterval]]]  = [[[:]]]
 
+    /**
+     An array containing speed limits on the route broken up by leg and then step.
+     */
+    public var maximumSpeedLimitsByLeg: [[[SpeedLimit]]] = []
```
1ec5: Rename to `maximumSpeedLimitsByStep` since the objects are grouped by step. Also rename `speedLimitDistancesByLeg` to `positionedSpeedLimitsByStep` (see below).

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -214,10 +221,46 @@ open class RouteProgress: NSObject {
 
                     congestionTimesPerStep[legIndex].append(stepCongestionValues)
                 }
+
+                if let segmentMaximumSpeedLimits = leg.segmentMaximumSpeedLimits {
+                    guard nextManeuverCoordinateIndex < segmentMaximumSpeedLimits.count else { continue }
+
+                    let stepSegmentMaximumSpeedLimits = Array(segmentMaximumSpeedLimits[maneuverCoordinateIndex..<nextManeuverCoordinateIndex])
+                    maximumSpeedLimitsByStep.append(stepSegmentMaximumSpeedLimits)
```
1ec5: Is it necessary to build up this array at all? Seems like ~~`speedLimitDistancesByLeg`~~ `positionedSpeedLimitsByStep` is the more useful property, since that’s what `currentSpeedLimit` uses.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -214,10 +221,46 @@ open class RouteProgress: NSObject {
...
+
+                    let polyline = Polyline(coordinates)
+                    let initialCoordinate = coordinates.first
+                    var maximumSpeedPerCoordinate: [(SpeedLimit, CLLocationDistance)] = []
+
+                    for (index, speedLimit) in stepSegmentMaximumSpeedLimits.enumerated() {
+                        let currentCoordinate = coordinates[index]
+                        let distanceTraveled = polyline.distance(from: initialCoordinate, to: currentCoordinate)
```
1ec5: This code is suboptimal because `Polyline.distance(from:to:)` traverses the entire polyline up to the `to` coordinate on every iteration of the for loop. Instead, calculate just the straight-line distance between the two coordinates (using `CLLocationCoordinate2D.distance(to:)` and accumulate the total distance in a local variable outside the for loop. That approach will also avoid inaccuracies when a route doubles back on itself.

Cartfile
```diff
@@ -1,6 +1,6 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.3
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.1.1
-github "mapbox/MapboxDirections.swift" ~> 0.27.3
+github "mapbox/MapboxDirections.swift" "maxspeedAnnotations"
```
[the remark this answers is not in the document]
avi-c: Understood.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -171,6 +171,13 @@ open class RouteProgress: NSObject {
      */
     public var congestionTimesPerStep: [[[CongestionLevel: TimeInterval]]]  = [[[:]]]
 
+    /**
+     An array containing speed limits on the route broken up by leg and then step.
+     */
+    public var maximumSpeedLimitsByLeg: [[[SpeedLimit]]] = []
+
+    private var speedLimitDistancesByLeg = [[[(SpeedLimit, CLLocationDistance)]]]()
```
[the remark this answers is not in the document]
avi-c: Done.

MapboxCoreNavigation/RouteProgress.swift
avi-c: Did the rename. I removed the `maximumSpeedLimitsByStep` array since I agree it doesn't seem needed now.

MapboxCoreNavigation/RouteProgress.swift
avi-c: Removed.

MapboxCoreNavigation/RouteProgress.swift
avi-c: Good to know about the Polyline.distance algorithm. Switched to the approach you suggested.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -171,6 +171,16 @@ open class RouteProgress: NSObject {
      */
     public var congestionTimesPerStep: [[[CongestionLevel: TimeInterval]]]  = [[[:]]]
 
+    /**
+     Tuple containing a `SpeedLimit` and a corresponding `CLLocationDistance` representing the position of the SpeedLimit along the current route step
```
1ec5: Nit: Declarations are always present in the generated documentation, so all that’s needed is an explanation of the fields’ semantics.

```suggestion
     A speed limit associated with a distance along a `RouteStep`.
```

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -25,7 +25,11 @@ open class NavigationRouteOptions: RouteOptions {
         shapeFormat = .polyline6
         includesSteps = true
         routeShapeResolution = .full
-        attributeOptions = [.congestionLevel, .expectedTravelTime, .maximumSpeedLimit]
+        if profileIdentifier == .walking {
```
1ec5: `profileIdentifier` is mutable, so technically it’s possible for the developer to change the profile identifier to `walking` after initializing the object. We could address that case by conditionally adding `maximumSpeedLimit` in an override of the `URLQueryItems` property instead of this initializer. But maybe it isn’t a big deal.

1ec5: Are speed limit annotations available for cycling directions and other profiles that may come up? What if we special-case the `automobile` and `automobileAvoidingTraffic` profiles instead?

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -25,7 +25,11 @@ open class NavigationRouteOptions: RouteOptions {
...
+            attributeOptions = [.congestionLevel, .expectedTravelTime]
+        } else {
+            attributeOptions = [.congestionLevel, .expectedTravelTime, .maximumSpeedLimit]
```
1ec5: Nit: A less repetitive way of writing this conditional code would be:

```swift
attributeOptions = [.congestionLevel, .expectedTravelTime]
if profileIdentifier == .automobile || profileIdentifier == .automobileAvoidingTraffic {
    attributeOptions.insert(.maximumSpeedLimit)
}
```

Let’s add a unit test to ensure that the right `attributeOptions` results for the profiles we care most about.

MapboxCoreNavigation/NavigationRouteOptions.swift
avi-c: that's probably the right thing to do for now to avoid potential problems that might come up if other profiles are added later.

1ec5: Turns out requesting `maxspeed` annotations is valid for the `cycling` profile, though it comes back as all `unknown`. It’s only `walking` that throws an error when you request `maxspeed` annotations. So just in case the Directions API lights up speed limits for bikeways in the future – they do exist in real life – we could request those annotations in that case, optimistically. The tradeoff would be a somewhat larger response size.

1ec5:
Once mapbox/MapboxDirections.swift#379 lands, this PR will be blocked by #2206.

1ec5:
This PR is no longer blocked by either mapbox/MapboxDirections.swift#379 or #2206. However, mapbox/MapboxDirections.swift#250 does need to land first.

1ec5:
This PR needs to be rebased to resolve conflicts, but mapbox/MapboxDirections.swift#250 depends on mapbox/MapboxDirections.swift#382, which master is incompatible with. #2275 fixes some of the compatibility issues but isn’t ready yet, and it depends on mapbox/MapboxDirections.swift#393, which isn’t ready either. 🤕

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -211,10 +221,55 @@ open class RouteProgress: NSObject {
...
+     
+     Speed limit data is available in [a number of countries and territories worldwide](https://docs.mapbox.com/help/how-mapbox-works/directions/).
+     */
+    public var currentSpeedLimit: Measurement<UnitSpeed>? {
+        guard legIndex < positionedSpeedLimitsByStep.count,
+            currentLegProgress.stepIndex < positionedSpeedLimitsByStep[legIndex].count else {
+            return nil
+        }
+        let speedLimits = positionedSpeedLimitsByStep[legIndex][currentLegProgress.stepIndex]
+        let lastSpeedLimitTuple = speedLimits.last { $0.1 <= currentLegProgress.currentStepProgress.distanceTraveled }
```
1ec5: As with any annotation type vended by the Directions API, each annotation corresponds to the span between two coordinates in the route geometry. So instead of scanning the whole step geometry for a matching position, we can get the index of the closest coordinate on the route line using Turf, then index into the array of speed limits. That will eliminate any unreliability caused by floating point imprecision.

PR #2120, opened 2019-04-30
Clean up CarPlay route preview pane

### LINKED ISSUE #2090: Show statistics when previewing routes in CarPlay
The CarPlay trip preview screen lacks the row of information below the summary text that indicates the estimated arrival time, duration, and distance of a route choice.

preview

`CarPlayManager.didCalculate(_:for:between:error:completionHandler:)` needs to call `CPMapTemplate.update(_:for:with:)` for each of the route choices, or perhaps lazily call it for each route choice as the user selects it.

/cc @mapbox/navigation-ios

1ec5:
Removed the “street address” from the CarPlay route preview pane; it was set redundantly to the destination waypoint’s name. For consistency, call `CarPlayManagerDelegate.carPlayManager(_:willPreview:)` also when programmatically starting a new route in CarPlay. Show statistics like the distance and estimated travel time when previewing a selected route.

preview-single
preview-multi
choose-multi

Fixes #2090.

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,10 @@
 * Fixed a regression where the puck could float around when standing still or moving backwards. ([#2109](https://github.com/mapbox/mapbox-navigation-ios/pull/2109))
 
 ### CarPlay
+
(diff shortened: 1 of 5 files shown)


1ec5:
> shouldn't the bottom image also show the alternative routes on the map?

That hasn’t been implemented yet; tracking in #2122.

PR #2124, opened 2019-05-09
Guard against missing speed limit data when speed limit is queried for

avi-c:
If speed limit annotations were not requested then we need to guard against attempting to look them up should the client request the current speed limit.

We should return SpeedLimit.invalid in these cases.

### DIFF
--- a/MapboxCoreNavigation/RouteProgress.swift
+++ b/MapboxCoreNavigation/RouteProgress.swift
@@ -258,7 +258,7 @@ open class RouteProgress: NSObject {
      Returns the SpeedLimit for the current position along the route. Returns SpeedLimit.invalid if the speed limit is unknown or missing.
      */
     public var currentSpeedLimit: SpeedLimit {
-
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -258,7 +258,7 @@ open class RouteProgress: NSObject {
      Returns the SpeedLimit for the current position along the route. Returns SpeedLimit.invalid if the speed limit is unknown or missing.
      */
     public var currentSpeedLimit: SpeedLimit {
-
+        guard legIndex < positionedSpeedLimitsByStep.count, currentLegProgress.stepIndex < positionedSpeedLimitsByStep[legIndex].count else { return .invalid }
```
1ec5: This is a bit late to verify that `positionedSpeedLimitsByStep` has the same number of items as there are legs and that the leg’s array of positioned speed limits has the same number of items as there are steps in the leg. We should assert about those counts after building up `positionedSpeedLimitsByStep`.

Here, all we really need to check is whether `positionedSpeedLimitsByStep` is populated at all. If speed limits can legitimately be missing (as in the case of walking or cycling directions), then `positionedSpeedLimitsByStep` should be optional.

Issue #1139, opened 2018-02-20
Improve layout in landscape orientation


1ec5:
NavigationViewController is laid out inefficiently when the device is in landscape orientation. Each banner takes up the entire width of the screen, revealing plenty of whitespace, while taking up valuable vertical screen real estate. With enough banners visible (lanes, “then”, simulation), the user puck overlaps the banners, and the floating action buttons at right overlap the bottom bar, obscuring the exit button.

<table>
<tr></tr>
<tr><td>

![landscape](https://user-images.githubusercontent.com/1231218/36442932-27fb8bd4-162c-11e8-8a77-eccd179d7764.png)

</td></tr>
</table>

Many navigation applications, including Apple Maps, switch to a two-column layout in landscape orientation, with the left column containing the banners and the bottom bar. With this layout, the banner gets less horizontal space but more vertical space, which can be used for line wrapping or for displaying additional information like incidents.

Previously: #726.

/cc @mapbox/navigation-ios

bsudekum:
I'm actually unsure how the lanes view and the then banner are both showing up in this pr, as we only show the then banner if the lanes view is hidden

https://github.com/mapbox/mapbox-navigation-ios/blob/c45d93837c6a07dc5f3ea5a829f3cd50ca7f7e46/MapboxNavigation/RouteMapViewController.swift#L276-L279

1ec5:
For the original report above, I was simulating [approximately this route](https://www.mapbox.com/get-directions/#15.55/39.2876/-84.4642?coordinates=-84.4567878134779,39.287417340931626;-84.46744300898638,39.28699625839792), and the stacked views did reproduce reliably. There’s a bit of a gap between these two checks:

https://github.com/mapbox/mapbox-navigation-ios/blob/c45d93837c6a07dc5f3ea5a829f3cd50ca7f7e46/MapboxNavigation/NextBannerView.swift#L78-L82
https://github.com/mapbox/mapbox-navigation-ios/blob/c45d93837c6a07dc5f3ea5a829f3cd50ca7f7e46/MapboxNavigation/LanesView.swift#L78

So one view showed up first, then the other, as I approached the intersection.

1ec5:
[...]

PR #2134, opened 2019-05-23
Remove viewport workaround in CarPlay

frederoni:
Now that https://github.com/mapbox/mapbox-gl-native/pull/14664 gained traction, I took the chance to remove the hacky workaround for the viewport in CarPlay and the NavigationMapView.

This PR is blocked by the PR mentioned above, and a new release of the Maps SDK. As well as some cleanup work and testing.

This PR will unblock #1139

Fixes #2134.

cc @mapbox/navigation-ios @astojilj


astojilj:
@frederoni Thanks. Looks good - simpler, though I'm not familiar with the code to properly review it.
It would be interesting to see how it affects performance - the viewport area should be now 2 times smaller:
from the screenshot looks like it was (1.5 * width + 100pixels) x (1.5 * height + 100pixels), width and height and 100px are in logical coordinates.

100 pixels comes from [CollisionIndex viewportPadding](https://github.com/mapbox/mapbox-gl-native/blob/2cfdd1a2f7042650538d24058fc28ad092739073/src/mbgl/text/collision_index.cpp#L25) that we could also try to tweak.

1ec5:
Awesome – so glad this is becoming a reality!

> 

I’m not sure why, but the vanishing point is offset slightly too far to the right, or the puck is offset slightly too far to the left.

frederoni:
> I’m not sure why, but the vanishing point is offset slightly too far to the right, or the puck is offset slightly too far to the left.

Yea, the userAnchorPoint doesn't exactly match the safeArea’s centerY so there is a small discrepancy that'll need to be addressed before this PR can land. Good pixelhunting 👁

1ec5:
mapbox/mapbox-gl-native#14664 is shipping today as part of map SDK v5.1.0. Since the bug fix isn’t shipping as a semver-major change, the workaround needs to be removed to avoid overcorrecting the vanishing point.

1ec5:
#2143 prevents this SDK from working with v5.1.0 for the time being, but we need to remove that restriction and depend on v5.1._x_ in short order to avoid blocking various applications from upgrading the map SDK.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -337,10 +337,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         if tracksUserCourse {
             let newCamera = camera ?? MGLMapCamera(lookingAtCenter: location.coordinate, altitude: altitude, pitch: 45, heading: location.course)
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: .linear) : nil
-            let point = userAnchorPoint
-            let padding = UIEdgeInsets(top: point.y, left: point.x, bottom: bounds.height - point.y, right: bounds.width - point.x)
-            // Omit padding when https://github.com/mapbox/mapbox-gl-native/pull/14081 has landed
```
1ec5: This comment was a bit misleading: the actual workarounds were the paddings added in `showcase(_:padding:animated:)` and `fit(to:facing:padding:animated:)`, not the one here. The edge padding here is necessary in order to account for the puck’s uncentered location even within the content frame.

/ref https://github.com/mapbox/mapbox-navigation-ios/issues/2145#issuecomment-503800503

1ec5:
These assertions are failing and crashing on iOS 12 for some odd reason:

https://github.com/mapbox/mapbox-navigation-ios/blob/db19ffe8b104ed25475916e95267a66da71da38d/MapboxCoreNavigationTests/NavigationEventsManagerTests.swift#L55-L57

PR #2138, opened 2019-05-24
Fix remaining voice instructions out of bounds crash

frederoni:
This PR fixes an out of bounds array access crash in `RouteStepProgress.getter:remainingVisualInstructions` where `RouteController.updateRouteLegProgress(status:)` would access it at a point in time where the spoken instruction index hadn't been updated.

To keep the indexes in sync, all updating is now consolidated in `RouteController.updateIndexes(status:)` besides the leg index where the receiver of `RouterDelegate.router(_:didArriveAt:)` can determine if the index should be updated or not.

@mapbox/navigation-ios


MapboxCoreNavigation/RouteController.swift
```diff
@@ -194,28 +191,36 @@ open class RouteController: NSObject {
...
         
-        let willChangeVisualIndex = status.bannerInstruction != nil
-        
-        if willChangeVisualIndex || isFirstLocation {
-            let currentStepProgress = routeProgress.currentLegProgress.currentStepProgress
-            currentStepProgress.visualInstructionIndex = Int(status.bannerInstruction?.index ?? 0)
-            
-            if let instruction = currentStepProgress.currentVisualInstruction {
+        // Announce visual instruction if it was updated or it is the first location being reported
+        if didUpdate || isFirstLocation {
```
1ec5: If `didUpdate` is false, then is it safe to announce anything even if `isFirstLocation` is true?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -172,16 +172,13 @@ open class RouteController: NSObject {
...
-        currentStepProgress.intersectionIndex = Int(status.intersectionIndex)
-        
         // Notify observers if the step’s remaining distance has changed.
         update(progress: routeProgress, with: CLLocation(status.location), rawLocation: location)
         
         let willReroute = !userIsOnRoute(location) && delegate?.router?(self, shouldRerouteFrom: location)
                           ?? DefaultBehavior.shouldRerouteFromLocation
         
-        updateRouteStepProgress(status: status)
+        updateIndexes(status: status)
```
vincethecoder: nit: plural of `index` is `indices`

MapboxCoreNavigation/RouteController.swift
```diff
@@ -194,28 +191,36 @@ open class RouteController: NSObject {
         checkForFasterRoute(from: location, routeProgress: routeProgress)
     }
     
-    func updateSpokenInstructionProgress(status: MBNavigationStatus, willReRoute: Bool) {
+    func updateIndexes(status: MBNavigationStatus) {
```
vincethecoder: function signature suggestion:

```swift
func updateIndices(status: MBNavigationStatus) {
...
}
```

MapboxCoreNavigation/RouteController.swift
vincethecoder: yep

frederoni: if `isFirstLocation` is true, that would mean `stepIndex`, `visualInstructionIndex`, and `spokenInstructionIndex` is 0. I can't think if a case where that wouldn't be safe, besides routeless navigation which currently is not supported.

MapboxCoreNavigation/RouteController.swift
frederoni: both are valid pluralizations, `indexes` is even being used in Foundation, e.g. `IndexPath.init(indexes:)`.

MapboxCoreNavigation/RouteController.swift
1ec5: Ah, between the whitespace changes and variable renaming, I had mistakenly thought the logic had changed too. 👍

Incidentally, how well do we handle the pathological case of a route that ends at exactly or roughly the same spot as it begins, or a route with two coincident destinations? If I remember correctly, one of the features of OSRM is that it can give you a no-op route, unlike some other routing engines that make you go in a circle. Perhaps that edge case raises other issues that are out of scope for this PR, but that’s the only case that seems relevant to me.

frederoni: > case of a route that ends at exactly or roughly the same spot as it begins, or a route with two coincident destinations?

Valid concerns but slightly out of scope for this PR.

I found one more place where indexes could be out of sync.

https://github.com/mapbox/mapbox-navigation-ios/blob/b309653a8acf26dcececcde4848e3060fa4ffac1/MapboxCoreNavigation/RouteController.swift#L253-L257

Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/2138/commits/bfe5ba7db6eec2e188ca33b7de386f8ec640f7ba

MapboxCoreNavigation/RouteController.swift
vincethecoder: That's fine. Realized it's accepted in American and Canadian English 😸

MapboxCoreNavigation/RouteController.swift
vincethecoder: nvm. it's an accepted spelling.

PR #2148, opened 2019-06-13
Post notification when style changes

### LINKED ISSUE #2139: Notification needed when style changes between day/night
Users working with day/night styles have no way of being notified when a style is applied  in order to swap out UI elements added as style-specific overlays. The map style and UI within NavigationViewController update automatically given the time of day but there is no way to be notified when the style is applied.

A proposed option would be setting up style manager anywhere except a subclass of `NavigationViewController`. However, the user wouldn’t be notified when the style changes due to entering/exiting a tunnel.

cc @frederoni

frederoni:
Fixes #2139  - Post notification when a style gets applied due to change of time of day or when entering or exiting a tunnel.

cc @1ec5 @JThramer @malwoodsantoro @achhaiba

### DIFF
--- a/MapboxNavigation/Constants.swift
+++ b/MapboxNavigation/Constants.swift
@@ -41,6 +41,7 @@ extension Notification.Name {
     /**
      Posted when `StyleManager` applies a style that was triggered by change of time of day, or when entering or exiting a tunnel.
      
+     This notification is the equivalent of `StyleManagerDelegate.styleManager(_:didApply:)`.
      The user info dictionary contains the key `StyleManagerNotificationUserInfoKey.style` and `StyleManagerNotificationUserInfoKey.styleManager`.
      */
     public static let styleManagerDidApplyStyle = NSNotification.Name.MBStyleManagerDidApplyStyle
--- a/MapboxNavigation/Constants.swift
+++ b/MapboxNavigation/Constants.swift
@@ -35,3 +35,13 @@ public let MBCongestionAttribute = "congestion"
  The minimum volume for the device before a gentle warning is emitted when beginning navigation.
  */
 public let NavigationViewMinimumVolumeForWarning: Float = 0.3
+
+extension Notification.Name {
+    
+    /**
+     Posted when `StyleManager` applies a style that was triggered by change of time of day, or when entering or exiting a tunnel.
+     
+     The user info dictionary contains the key `StyleManagerNotificationUserInfoKey.style` and `StyleManagerNotificationUserInfoKey.styleManager`.
+     */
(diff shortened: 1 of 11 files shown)


MapboxNavigation/StyleManager.swift
```diff
@@ -171,6 +174,13 @@ open class StyleManager: NSObject {
         return solar.date.isNighttime(sunrise: sunrise, sunset: sunset) ? .night : .day
     }
     
+    private func postDidApplyStyleNotification(style: Style) {
```
frederoni: another idea was to post the notification from within the style, something like:
```swift
open override func apply() {
    super.apply()
    
    ....
    defer {
        // Post notification
    }
}
```

but this would post the notification before the full style has been applied to the UIAppearance proxy.

MapboxNavigation/MBConstants.h
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ Key used for constructing errors when spoken instructions fail.
+ */
+extern const _Nonnull NSErrorUserInfoKey MBSpokenInstructionErrorCodeKey;
+
+/**
+ Posted when `MBStyleManager` applies a new style triggered by change of day of time, or when entering or exiting a tunnel.
+ 
+ :nodoc:
```
1ec5: Do these types need to be excluded from the docset?

MapboxNavigation/StyleManager.swift
```diff
@@ -153,6 +155,7 @@ open class StyleManager: NSObject {
                 currentStyle = style
                 style.apply()
                 delegate?.styleManager?(self, didApply: style)
+                postDidApplyStyleNotification(style: style)
```
1ec5: These four lines are repeated in a few places; it would be more maintainable to set `currentStyle`, apply the style, and call the delegate method within the new method.

MapboxNavigation/Constants.swift
```diff
@@ -35,3 +35,13 @@ public let MBCongestionAttribute = "congestion"
  The minimum volume for the device before a gentle warning is emitted when beginning navigation.
  */
 public let NavigationViewMinimumVolumeForWarning: Float = 0.3
+
+extension Notification.Name {
+    
+    /**
+     Posted when `StyleManager` applies a style that was triggered by change of time of day, or when entering or exiting a tunnel.
```
1ec5: This comment should refer to `StyleManagerDelegate.styleManager(_:didApply:)` as being equivalent and vice versa.

MapboxNavigation/MBConstants.h
frederoni: I assumed it had to do with the linter, but after running it, I don't see any reason to exclude these types from the docset.

PR #2149, opened 2019-06-14
Alternate Top Banner Instructions

vincethecoder:
This effort creates a custom top banner instructions a.k.a `Instructions Cards`. This subscribes to the top banner configured in the route options used in creating a Navigation ViewController. 

This top banner instructions component permits customization of the prototype card size, displayed instructions and the desired card style. 

TODO
- [x] Add snapshot tests for cards with maneuver instructions
- [x] Design review complete
- [x] Fix broken tests
- [x] Test drive
~Add snapshot tests for cards with next banner and lane indication(s) instructions~
~Provide the ability to customize the data sent to the `dataSource`~ 

/cc @JThramer @frederoni @1ec5 @d-prukop @okoriep


Example/ViewController.swift
```diff
@@ -504,3 +507,42 @@ extension ViewController: VisualInstructionDelegate {
...
 }
+
+
+extension ViewController: InstructionsCardCollectionDelegate {
+    
+    public func instructionsCardCollection(_ instructionsCardCollection: InstructionsCardCollection, previewFor step: RouteStep) {
+        
+        guard let firstRoute = routes?.first else { return }
+        
+        let service = navigationService(route: firstRoute)
```
frederoni: `service` is not being used other than passing through `firstRoute` to `route`.

MapboxNavigation/InstructionsCard+Font.swift
```diff
@@ -0,0 +1,13 @@
+enum CardFont: String {
+    case regular
+    case bold
+    
+    static func create(_ type: CardFont, with size: CGFloat) -> UIFont! {
+        return type == .bold ? UIFont.boldSystemFont(ofSize: size) : UIFont.systemFont(ofSize: size)
```
frederoni: use `.adjustedFont` to make the font respect dynamic typing.

MapboxNavigation/InstructionsCardCollection.swift
```diff
@@ -0,0 +1,370 @@
...
+    var cardStyle: DayInstructionsCardStyle = DayInstructionsCardStyle()
+    
+    var instructionCollectionView: UICollectionView!
+    var instructionsCardLayout: InstructionsCardCollectionLayout!
+    
+    public private(set) var isInPreview = false
+    private var currentStepIndex: Int?
+    
+    var steps: [RouteStep]? {
+        guard let stepIndex = routeProgress?.currentLegProgress.stepIndex, let steps = routeProgress?.currentLeg.steps else { return nil }
```
frederoni: what about multileg routes?

MapboxNavigation/InstructionsCardStyle.swift
```diff
@@ -0,0 +1,39 @@
+@objc public protocol InstructionsCardStyle: class {
```
frederoni: We could break out the core functionality in `Style` to a Stylable protocol. That way, we could extend this card styling in `InstructionsCardStyle` and provide day style and night style for cards.

MapboxNavigation/InstructionsCardContainerView.swift
```diff
@@ -0,0 +1,306 @@
...
+        if nextBannerView.isHidden {
+            nextBannerView.update(for: instruction)
+            nextBannerView.instructionDelegate = self
+        } else if nextBannerView.isCurrentlyVisible && instruction.tertiaryInstruction?.text == nil {
+            nextBannerView.hide()
+        }
+        
+        if !previewEnabled, let distance = instructionsCardView.distanceFromCurrentLocation {
+            let highlightEnabled = distance < InstructionsCardConstants.highlightDistance
+            updateBackgroundColor(highlightEnabled: highlightEnabled)
```
1ec5: Every time the collection view is updated, we’re recalculating the distances for all remaining steps from the current location to the end of the leg, just to determine whether to highlight a card. This is suboptimal for a few reasons:

* It’s unlikely that the ten steps out is near enough to warrant highlighting, so we don’t need to calculate anything for it.
* Each of the distances has to be recalculated on every location update.

Taking a step back, the point of the highlighting feature is to highlight nearby maneuvers. There’s a simpler approach: on each location update, get the current step and mark it as nearby. If there’s less than _nearbyDistance_ meters left on the step, mark the next maneuver as nearby and subtract the distance to that maneuver from _nearbyDistance_, and so on.

MapboxNavigation/InstructionsCardCollection.swift
```diff
@@ -0,0 +1,370 @@
...
+        }
+        return mutatedSteps
+    }
+    
+    var distancesFromCurrentLocationToManeuver: [CLLocationDistance]? {
+        guard let progress = routeProgress, let steps = steps else { return nil }
+        let distanceRemaining = progress.currentLegProgress.currentStepProgress.distanceRemaining
+        let distanceBetweenSteps = [distanceRemaining] + progress.remainingSteps.map {$0.distance}
+        
+        let distancesFromCurrentLocationToManeuver: [CLLocationDistance] = steps.enumerated().map { (index, _) in
```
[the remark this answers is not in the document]
vincethecoder: Concur 👍

MapboxNavigation/InstructionsCardCollection.swift
vincethecoder: coming soon.

MapboxNavigation/InstructionsCardContainerView.swift
vincethecoder: Yes, the current implementation isn't optimal. The suggested approach will certainly work if we assume 1 - 2 instruction cards are displayed at a time. 

Hence, we need to account for only visible cells for the current leg to ensure we persist the intended user experience.

1ec5: We don’t need to assume the number of cards visible. We’d start out by unhighlighting all the cards and iterating from the current step onward.

MapboxNavigation/InstructionsCardCollection.swift
```diff
@@ -50,11 +50,16 @@ open class InstructionsCardCollection: UIViewController {
...
+        guard let firstDistance = distanceBetweenSteps.first else { return nil }
+        var distancesFromCurrentLocationToManeuver = [CLLocationDistance](repeating: 0, count: steps.count)
+        distancesFromCurrentLocationToManeuver[0] = firstDistance
         
-        let distancesFromCurrentLocationToManeuver: [CLLocationDistance] = steps.enumerated().map { (index, _) in
+        for index in 1..<distancesFromCurrentLocationToManeuver.endIndex {
             let safeIndex = index < distanceBetweenSteps.endIndex ? index : distanceBetweenSteps.endIndex - 1
-            let cardDistance = distanceBetweenSteps[0...safeIndex].reduce(0, +)
-            return cardDistance > 5 ? cardDistance : 0
+            let previousDistance = distanceBetweenSteps[safeIndex-1]
```
1ec5: Assuming `distancesFromCurrentLocationToManeuver` should be the same size as `distanceBetweenSteps`:

```swift
var cumulativeDistance: CLLocationDistance = 0
var distancesFromCurrentLocationToManeuver: [CLLocationDistance] = []
distancesFromCurrentLocationToManeuver.reserveCapacity(distancesBetweenSteps.count)
for distance in distancesBetweenSteps {
    distancesFromCurrentLocationToManeuver = cardDistance > 5 ? cumulativeDistance + distance : 0
    cumulativeDistance.append(distance)
}
```

Also, we should move the `cardDistance > 5` check somewhere else, like in the code that decides whether to highlight or not, because it’s a presentational concern, not something intrinsic to the distances.

MapboxNavigation/InstructionsCardCollection.swift
```diff
@@ -0,0 +1,332 @@
...
+        guard let stepIndex = routeProgress?.currentLegProgress.stepIndex, let steps = routeProgress?.currentLeg.steps else { return nil }
+        var mutatedSteps = steps
+        if mutatedSteps.count > 1 {
+            mutatedSteps = Array(mutatedSteps.suffix(from: stepIndex))
+            mutatedSteps.removeLast()
+        }
+        return mutatedSteps
+    }
+    
+    var distancesFromCurrentLocationToManeuver: [CLLocationDistance]? {
```
1ec5: This property as well as `step` appear to be strictly related to the object model – they don’t use anything else in self. Can we give them more general-purpose names and move them to RouteProgress or an extension of that class? That’ll make it much easier to unit-test this functionality.

Example/ViewController.swift
```diff
@@ -240,6 +240,9 @@ class ViewController: UIViewController {
     func navigationViewController(navigationService: NavigationService) -> NavigationViewController {
         let route = navigationService.route
         let options = NavigationOptions( navigationService: navigationService)
+        let topBanner = InstructionsCardCollection()
+        topBanner.cardCollectionDelegate = self
+        options.topBanner = topBanner
```
JThramer: I'd rather this not be the _default_ example. Why not put this in as one of the options in the action sheet when you start navigation?

MapboxNavigationTests/InstructionsCardSnapshotTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        
+        let service = MapboxNavigationService(route: initialRoute, directions: DirectionsSpy(accessToken: "knut"), simulating: .never)
+        let routeProgress = RouteProgress(route: fakeRoute)
+        let intersectionLocation = routeProgress.route.legs.first!.steps.first!.intersections!.first!.location
+        let fakeLocation = CLLocation(latitude: intersectionLocation.latitude, longitude: intersectionLocation.longitude)
+        subject.routeProgress = routeProgress
+        subject.navigationService(service, didUpdate: routeProgress, with: fakeLocation, rawLocation: fakeLocation)
+        
+        /* TODO: Update and Re-generate for generic instructions cards and complex maneuver instruction cards */
+//        /// Validate the visible collection view cell
```
JThramer: Commented-out code.

Example/ViewController.swift
vincethecoder: Yes, it's best suited as an alternative, considering all factors.

MapboxNavigationTests/InstructionsCardSnapshotTests.swift
vincethecoder: Yep - will be uncommented along with the `WIP` tag.

JThramer: Groovy.

Example/ViewController.swift
```diff
@@ -504,3 +507,39 @@ extension ViewController: VisualInstructionDelegate {
         return presented
     }
 }
+
+
+extension ViewController: InstructionsCardCollectionDelegate {
+    
+    public func instructionsCardCollection(_ instructionsCardCollection: InstructionsCardCollection, previewFor step: RouteStep) {
```
JThramer: Lets talk sometime about using a standard delegate method to hook up this functionality both in the default example and in this UI VC.

vincethecoder: sure.

MapboxNavigation/InstructionsCardCollection.swift
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+    
+    func updateVisibleInstructionCards(at indexPaths: [IndexPath], previewEnabled: Bool = false) {
+        guard let distances = distancesFromCurrentLocationToManeuver else { return }
+        for index in indexPaths.startIndex..<indexPaths.endIndex {
+            let indexPath = indexPaths[index]
+            if let container = instructionContainerView(at: indexPath), indexPath.row < distances.endIndex {
+                let distance = distances[indexPath.row]
+                container.updateInstructionCard(distance: distance, previewEnabled: previewEnabled)
```
1ec5: In https://github.com/mapbox/mapbox-navigation-ios/pull/2149#discussion_r295589920, I pointed out that it’s quite inefficient to calculate the route-line distance from the current step to every remaining maneuver on every location update (once a second). In turns out this was by design: each previewed card (and indeed the follow-on card peeking in from the side) displays a distance measured from the current location at all times, resulting in monotonically increasing distances.

I hope we’ll be able to reconsider this design, which is unfortunate not only for its inherently poor performance but also unintuitiveness to the end user. But for now, there isn’t a good way around performing all those calculations.

MapboxNavigation/InstructionsCardCollection.swift
vincethecoder: Eventually, the goal is to make this a data source provided to the instructions card. This `InstructionsCardCollection` class would not contain any information related to how the distances displayed on each card is derived.

MapboxNavigation/InstructionsCardStyle.swift
vincethecoder: @frederoni - captured in this ticket #2175. Will be resolved in an upcoming PR.

MapboxNavigation/InstructionsCardCollection.swift
vincethecoder: @frederoni captured in this issue ticket #2176

MapboxNavigation/InstructionsCardContainerView.swift
```diff
@@ -0,0 +1,315 @@
...
+        }
+        
+        return firstLayer
+    }
+    
+    public func updateInstruction(for step: RouteStep, distance: CLLocationDistance) {
+        instructionsCardView.updateDistanceFromCurrentLocation(distance)
+        instructionsCardView.step = step
+        
+        // TODO: Merge Instructions Card, Lanes & Next Banner View Instructions
```
1ec5: To do: figure out what this to-do means.

MapboxNavigation/InstructionsCardContainerView.swift
```diff
@@ -0,0 +1,315 @@
...
+    
+    public func updateInstruction(_ instruction: VisualInstructionBanner) {
+        if lanesView.isHidden {
+            lanesView.update(for: instruction)
+        } else if lanesView.isCurrentlyVisible && instruction.tertiaryInstruction == nil {
+            lanesView.hide()
+        }
+        
+        if nextBannerView.isHidden {
+            nextBannerView.update(for: instruction)
```
1ec5: I don’t think this update should be guarded by a conditional here. `NextBannerView.update(for:)` already hides the banner if there’s no information to display. But this conditional means a next banner only updates if it was hidden to begin with. I’m not totally sure that’s always correct.

JThramer:
Posterity: This PR should be good to merge. Multi-leg support for guidance-cards, tracked in #2176, will be follow-on tail work, and as such all classes in this PR are marked `/// :nodoc:` until we are ready to announce the new feature.

/cc @mapbox/navigation-ios

1ec5:
Still getting failures on the Xcode 10.2 / iOS 12.2 CI bot:

```
/Users/distiller/project/MapboxNavigationTests/GuidanceCardsSnapshotTests.swift:59: error: -[MapboxNavigationTests.GuidanceCardsSnapshotTests testLanesManeuver] : failed - Snapshot comparison failed: Optional(Error Domain=FBSnapshotTestControllerErrorDomain Code=4 "Images different" UserInfo={NSLocalizedFailureReason=image pixels differed by more than 0.00% from the reference image, FBDiffedImageKey=<UIImage: 0x6000013f3b80>, {375, 812}, FBReferenceImageKey=<UIImage: 0x6000013f8690>, {375, 812}, FBCapturedImageKey=<UIImage: 0x6000013f3bf0>, {375, 812}, NSLocalizedDescription=Images different})
```

The test fixture was generated in Xcode 10.3 with an iOS 12.1 simulator.

1ec5:
Despite the test fixtures’ file names, the tests are apparently being run on a device other than iPhone X. I see no reason to believe it’d be anything other than iPhone 6 Plus, which is what the CircleCI configuration specifies.

Expected | Actual
----|----
reference_testLanesManeuver_iPhone_X_Portrait_iOS_12 1@3x | failed_testLanesManeuver_iPhone_X_Portrait_iOS_12 1@3x

At the same time, I think it’s unfortunate that the next guidance card would peek in a bit more on one device than another. Either we only want enough of the card to peek in so that the user can tell the cards are swipable, or we want enough to peek in to characterize the following maneuver.

1ec5:
Once we address tail work such as #2175 and #2176, we’ll want to mention this feature and all its API implications in the changelog.

PR #2151, opened 2019-06-17
Add MapboxAccounts

frederoni:
Added MapboxAccounts to handle SKU tokens

- [x] Set up redirect
- [x] Release MapboxSpeech and bump dependency (https://github.com/mapbox/mapbox-speech-swift/pull/27)
- [x] Release MapboxDirections and bump dependency (https://github.com/mapbox/MapboxDirections.swift/pull/374)
- [x] Update readme and changelog
- [ ] ~~Document Info.plist keys in jazzy guide~~
- [x] Publish MapboxAccounts

I've bumped the Xcode requirement to 10.2.1 so we can rely on Swift 5 and migrate to SPM for development as soon as possible now that MapboxSpeech is already on Swift 5.

cc @1ec5 @JThramer @friedbunny


Example/AppDelegate.swift
```diff
@@ -15,6 +16,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     lazy var carPlaySearchController: CarPlaySearchController = CarPlaySearchController()
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        
+        // Opt-in to billing per trip
+        // It’s important to call `MBXAccounts.activateSKUID(_:)` before making requests to the Directions API or the Voice API
+        // to avoid getting billed per request.
+        MBXAccounts.activateSKUID(.navigationSession)
```
1ec5: Is it necessary to call this method on launch, or is it safe for the developer to defer it until right before they make the first Directions API request? Or going in the other direction, is there an Info.plist option for activating the SKU ID? It seems like something that the developer would consider in the same breath as setting their access token.

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -234,14 +237,18 @@ public class MapboxNavigationService: NSObject, NavigationService, DefaultInterf
...
     @objc required public init(route: Route,
                                directions: Directions? = nil,
                                locationSource: NavigationLocationManager? = nil,
                                eventsManagerType: NavigationEventsManager.Type? = nil,
                                simulating simulationMode: SimulationMode = .onPoorGPS,
-                               routerType: Router.Type? = nil)
+                               routerType: Router.Type? = nil,
+                               skuIdentifier: MBXAccountsSKUID? = .navigationSession)
     {
+        MBXAccounts.activateSKUID(skuIdentifier ?? .navigationSession)
```
1ec5: We could detect that the SKU ID hasn’t been activated yet and issue a one-time warning in the console to remind the developer to set it earlier. To differentiate between a Route that comes directly from the Directions API and a Route that’s deserialized from a file or built programmatically, we could check for one of the attributes added in mapbox/MapboxDirections.swift#342.

Example/AppDelegate.swift
frederoni: `.navigationSession` _doesn't_ affect the Maps SDK usage so it's safe to call before using any of the service libraries. `.navigationUser` _does_ affect the Maps SDK usage so it must be called before the Maps SDK is initialized.

I worked around this issue by activating the default (`.navigationSession`) in `+[MBXAccounts(MBXDefault) load]` so that we don't require any additional steps by default. The Info.plist option would be a good addition.

MapboxCoreNavigation/NavigationService.swift
frederoni: This is no longer relevant. See https://github.com/mapbox/mapbox-navigation-ios/pull/2151#discussion_r294803296

Cartfile
```diff
@@ -1,7 +1,8 @@
-binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.0.0
+binary "https://mapbox.s3.amazonaws.com/mapbox-accounts/ios/MapboxAccounts.json" ~> 0.0.1
+binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.1.0-beta.1
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.2.1
-github "mapbox/MapboxDirections.swift" ~> 0.28.0
+github "mapbox/MapboxDirections.swift" "HEAD"
```
frederoni: Make a new release of MapboxDirections

Cartfile
```diff
@@ -1,7 +1,8 @@
...
 github "mapbox/turf-swift" ~> 0.3
 github "mapbox/mapbox-events-ios" ~> 0.9.3
 github "ceeK/Solar" ~> 2.1.0
-github "mapbox/mapbox-speech-swift" ~> 0.1.0
+github "mapbox/mapbox-speech-swift" "HEAD"
```
frederoni: Make a new release of MapboxSpeech

Cartfile.resolved
```diff
@@ -1,14 +1,16 @@
-binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "5.0.0"
+binary "https://mapbox.s3.amazonaws.com/mapbox-accounts/ios/MapboxAccounts.json" "0.0.1"
+binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "5.1.0-beta.1"
```
frederoni: Wait for proper 5.1.0 release

README.md
```diff
@@ -24,7 +24,7 @@ Get up and running in a few minutes with our drop-in turn-by-turn navigation `Na
 
 ## Requirements
 
-The Mapbox Navigation SDK and Core Navigation are compatible with applications written in Swift 4.2 or Objective-C in Xcode 10.0. The Mapbox Navigation and Mapbox Core Navigation frameworks run on iOS 9.0 and above.
+The Mapbox Navigation SDK and Core Navigation are compatible with applications written in Swift 5.0 or Objective-C in Xcode 10.2.1. The Mapbox Navigation and Mapbox Core Navigation frameworks run on iOS 9.0 and above.
 
 The last release compatible with Swift 3.2 was v0.10.1.
```
1ec5: When installing the library using Carthage, does the developer need to ensure that MapboxAccounts.framework is in the Embedded Binaries list? If so, we should also mention this requirement as a backwards-incompatible change in the changelog.

https://github.com/mapbox/mapbox-navigation-ios/blob/0fcaf83b1e5b2dbfba739ba23986bbcecb111914/README.md#L57

MapboxCoreNavigation/MBRouteController.m
```diff
@@ -32,5 +33,13 @@ - (NSString * _Nonnull)md5 {
     
     return  output;
 }
+
 @end
 
+@implementation MBXAccounts (MBXDefault)
+
++ (void)load {
```
1ec5: Making the activation automatic on application launch is an improvement, to be sure: it streamlines the setup experience and removes an error-prone step (where an error takes a long time to discover). But it also removes the developer’s ability to decide when to activate a trip, which may be relevant for an application that puts turn-by-turn navigation behind an in-app purchase or something to that effect.

The fundamental challenge is that the typical workflow requires the developer to use MapboxDirections.swift before involving the navigation SDK in any way, yet that initial MapboxDirections.swift usage also needs to be accounted for in a trip. It makes me wonder whether we could impose an expectation that NavigationRouteOptions or NavigationMatchOptions be used in order to begin a “trip”, as opposed to RouteOptions or MatchOptions. These subclasses are part of the navigation SDK, so we could activate the navigation SKU ID as soon as one is initialized.

/cc @coxchapman

README.md
```diff
@@ -24,7 +24,7 @@ Get up and running in a few minutes with our drop-in turn-by-turn navigation `Na
 
 ## Requirements
 
-The Mapbox Navigation SDK and Core Navigation are compatible with applications written in Swift 4.2 or Objective-C in Xcode 10.0. The Mapbox Navigation and Mapbox Core Navigation frameworks run on iOS 9.0 and above.
+The Mapbox Navigation SDK and Core Navigation are compatible with applications written in Swift 5.0 or Objective-C in Xcode 10.2.1. The Mapbox Navigation and Mapbox Core Navigation frameworks run on iOS 9.0 and above.
```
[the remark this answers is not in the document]
frederoni: You're right. Removed 5.0.

MapboxCoreNavigation/MBRouteController.m
frederoni: > But it also removes the developer’s ability to decide when to activate a trip

Activating a sku identifier only set the identifier. The first service library (Speech or Directions) that makes a subsequent request, will generate the first token and the countdown towards token rotation begins.

Cartfile
```diff
@@ -1,7 +1,8 @@
-binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.0.0
+binary "https://mapbox.s3.amazonaws.com/mapbox-accounts/ios/MapboxAccounts.json" ~> 0.0.1
+binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.1
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.2.1
-github "mapbox/MapboxDirections.swift" ~> 0.28.0
+github "mapbox/MapboxDirections.swift" ~> 0.29
 github "mapbox/turf-swift" ~> 0.3
-github "mapbox/mapbox-events-ios" ~> 0.9.3
+github "mapbox/mapbox-events-ios" "fred/allow-subclassing"
```
frederoni: Don't forget to pin to a proper release when https://github.com/mapbox/mapbox-events-ios/pull/139 has been solved.

1ec5:
Should this readme line be updated to mention MapboxAccounts too?

https://github.com/mapbox/mapbox-navigation-ios/blob/01288ce781b1f07f1e350d4ed4d62263ca8d7835/README.md#L146

MapboxNavigationTests/ImageDownloaderTests.swift
```diff
@@ -12,7 +12,7 @@ class ImageDownloaderTests: XCTestCase {
 
     var downloader: ReentrantImageDownloader?
 
-    let imageURL = URL(string: "https://zombo.com/lulz/selfie.png")!
+    let imageURL = URL(string: "https://github.com/mapbox/mapbox-navigation-ios/blob/master/docs/img/navigation.png")!
```
1ec5: Wow, haha. The test doesn’t actually download this image from the network, right?

frederoni: it is a live test, added in #1046, selfie.png disappeared from the interwebs earlier today. w/i the repo sounds more reasonable but I'm up for stubbing the request now when we've linked OHHTTPStubs.

Cartfile
```diff
@@ -1,7 +1,8 @@
+binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" ~> 0.0.1
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.1
 binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.2.1
-github "mapbox/MapboxDirections.swift" ~> 0.28.0
+github "mapbox/MapboxDirections.swift" ~> 0.29
 github "mapbox/turf-swift" ~> 0.3
-github "mapbox/mapbox-events-ios" ~> 0.9.3
+github "mapbox/mapbox-events-ios" ~> 0.9.4
```
1ec5: #2177 increases the version requirement to v0.9.5, superseding this part of the PR.

MapboxCoreNavigation/MBXAccounts+CoreNavigationAdditions.m
```diff
@@ -0,0 +1,14 @@
+#import "MBXAccounts+CoreNavigationAdditions.h"
+
+@implementation MBXAccounts (CoreNavigationAdditions)
+
++ (void)load {
+    NSString *billingMethodValue = [NSBundle.mainBundle objectForInfoDictionaryKey:@"MBXNavigationBillingMethod"];
+    MBXAccountsSKUID skuIdentifier = MBXAccountsSKUIDNavigationSession;
+    if ([billingMethodValue isEqualToString:@"mau"]) {
```
1ec5: This logic needs to be revised so that `mau` is the default and there’s an option to opt out of a SKU altogether for accounting based on API calls.

1ec5:
`MBXNavigationBillingMethod` will be undocumented, but an Info.plist guide would be helpful for other things like `MGLMapboxAccessToken`.

1ec5:
`SKUTests.testDefaultSKU()` and `SKUTests.testMAUSKU()` have failed ever since https://github.com/mapbox/mapbox-navigation-ios/pull/2151#event-2899306537, primarily due to the map SDK sending no SKU token. The tests need to be updated, but I’m not seeing a straightforward way to test both supported configurations. An end-to-end test we conducted didn’t turn up any glaring issues, but we were unable to test the map SDK’s behavior at the time. So either the map SDK isn’t going along with the SKU token generated by the navigation SDK, or the `MGLMapView.skuToken` extension method in these tests is inaccurate.

Udumft:
With the help from billing team and [some](https://github.com/mapbox/sku-spark) extra [docs](https://github.com/mapbox/billing-team/blob/master/internal-datasets.md) provided, I checked that all requests under my user are logged as `user` type, not individual `api` requests. Tokens from different SDKs match each other within the app and updated when needed.

- [ ] Test that Maps, Speech and Navigation SDKs tokens match
- [ ] Test requests are logged for billing correctly
- [ ] Test that token expires over time and rotates per installation

I think this PR is ready to be merged

MapboxNavigationTests/ImageDownloaderTests.swift
JThramer: @Udumft can we implement this suggestion? Live tests aren't the best idea.

MapboxCoreNavigation/MBXAccounts+CoreNavigationAdditions.m
```diff
@@ -0,0 +1,20 @@
...
+        [MBXAccounts activateSKUID:MBXAccountsSKUIDNavigationUser];
+    } else if (![billingMethodValue isEqualToString:MBXNavigationBillingMethodRequest]) {
+        // Billing method is not the default in the absence of a SKU ID.
+        [NSException raise:@"MBXInvalidNavigationBillingMethod"
+                    format:@"Unrecognized billing method %@. Valid billing methods are: %@, %@.",
+         billingMethodValue, MBXNavigationBillingMethodUser, MBXNavigationBillingMethodRequest];
+    }
+}
+
+@end
```
JThramer: Do these have to be in OBJ-C?

Udumft: Yes, compiler simply does not permit to define such method in Swift.

frederoni: Here's a pretty nice workaround https://stackoverflow.com/a/47621197
It would be both Obj-C and Swift but at least the business logic could be written in Swift.
Maybe that could be cleaner to Swiftify `MBXNavigationBillingMethod`.

Udumft: It still produces some Obj-C code as well as a wrapper class and so on, so... Will we win something if we do it that way?

frederoni: The only benefit I see is that we could `switch` over `MBXNavigationBillingMethod` instead of using `isEqualToString:` if it became a Swift enum but I don't have a strong opinion for or against it.

Udumft: Agreed, `switch` would look much better. But it will introduce an implicit behavior: Obj-C wrapper class that calls "random" Swift class(es) static method which has to have a unique (to avoid possible collisions) name. It compiles slower, it runs slower, it takes more time to understand what is going on in such code. As I don't see benefits that outweigh these cons, I would prefer to just leave this method Obj-C.

MapboxNavigationTests/ImageDownloaderTests.swift
Udumft: I tried this test with invalid image, I tried it with no internet connection and it works every time.
Maybe I misunderstood the point of the test, but since `downloader` is configured with `ImageLoadingURLProtocolSpy` like this: https://github.com/mapbox/mapbox-navigation-ios/blob/6d10ac02bc7ecb038fa770d2d5fad88bf651ad87/MapboxNavigationTests/ImageDownloaderTests.swift#L21-L24
it will always "download" an image. At the same time not configuring the downloader will make `ImageLoadingURLProtocolSpy` checks fail.
Maybe we should re-think these tests, or I missed something?

JThramer:
Posterity: Run a nav-session for an hour and make sure the MapSDK doesn't reset the SKU token undesirably.

Udumft:
I ran a few turn-by-turn navigation sessions over an hour duration and debug log shows that tokens are rotated correctly (according to [this doc](https://github.com/mapbox/pricebook/issues/28)) every hour ("User" SKU type). Tokens for Directions, Speech and Maps SDK match each other after rotation too. Requests on amazon log are all in place.

Udumft:
Billin team [confirmed](https://mapbox.slack.com/archives/C7DD5GLJ0/p1587055121013900?thread_ts=1587023121.010200&cid=C7DD5GLJ0) backed is ready. PR can be merged.

PR #2156, opened 2019-06-20
Fix crash presenting NavigationViewController on iOS 13

### LINKED ISSUE #2154: Crash starting navigation session in Xcode 11 — NavigationSettings.init()
In Xcode 11.0 beta (11M336w) on iOS 13.0, NavigationViewController crashes as soon as it is presented:

```
#0	0x000000010e9fff27 in objc_exception_throw ()
#1	0x000000010ebe9739 in -[NSException raise] ()
#2	0x000000010e45c024 in -[NSObject(NSKeyValueCoding) valueForUndefinedKey:] ()
#3	0x000000010e45ac97 in -[NSObject(NSKeyValueCoding) valueForKey:] ()
#4	0x000000010fcfb41b in NavigationSettings.init() at /path/to/mapbox-navigation-ios/MapboxCoreNavigation/NavigationSettings.swift:73
```

https://github.com/mapbox/mapbox-navigation-ios/blob/0fcaf83b1e5b2dbfba739ba23986bbcecb111914/MapboxCoreNavigation/NavigationSettings.swift#L73

We’re trying to get the object, then the value, for the key `$__lazy_storage_$_properties`. This probably corresponds to the lazy `properties` property, which we try to avoid looking up:

https://github.com/mapbox/mapbox-navigation-ios/blob/0fcaf83b1e5b2dbfba739ba23986bbcecb111914/MapboxCoreNavigation/NavigationSettings.swift#L63

We need to check for `$__lazy_storage_$_properties` instead of or in addition to `properties.storage`.

/cc @mapbox/navigation-ios

1ec5:
Ignore the `$__lazy_storage_$_properties` key on NavigationSettings.

Fixes #2154.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/NavigationSettings.swift
+++ b/MapboxCoreNavigation/NavigationSettings.swift
@@ -60,7 +60,7 @@ public class NavigationSettings: NSObject {
         let properties = Mirror(reflecting: self).children
         return properties.filter({ (child) -> Bool in
             if let label = child.label {
-                return label != "properties.storage"
+                return label != "properties.storage" && label != "$__lazy_storage_$_properties"
             }
             return false
         })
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -60,7 +60,7 @@ public class NavigationSettings: NSObject {
         let properties = Mirror(reflecting: self).children
         return properties.filter({ (child) -> Bool in
             if let label = child.label {
-                return label != "properties.storage"
+                return label != "properties.storage" && label != "$__lazy_storage_$_properties"
```
1ec5: I considered changing this check to `!label.starts(with: "$__lazy_storage_$_")`, but the intent of this check was specifically to filter out the `properties` property, not any lazy arbitrary property.

PR #2181, opened 2019-07-15
Update Jazzy theme

captainbarbosa:
Edits Jazzy generation script to support https://github.com/realm/jazzy/pull/1077.

Requires https://github.com/mapbox/jazzy-theme/pull/9 to merge before merging.

cc @1ec5

### DIFF
--- a/scripts/document.sh
+++ b/scripts/document.sh
@@ -60,10 +60,9 @@ jazzy \
     --documentation="docs/guides/*.md" \
     --root-url "${BASE_URL}/navigation/${RELEASE_VERSION}/" \
     --theme ${THEME} \
-    --output ${OUTPUT}
+    --output ${OUTPUT} \
+    --module_version ${RELEASE_VERSION}
 
-REPLACE_REGEXP='s/MapboxNavigation\s+(Docs|Reference)/Mapbox Navigation SDK for iOS $1/, '
-REPLACE_REGEXP+='s/BRANDLESS_DOCSET_TITLE/Navigation SDK for iOS $1/, '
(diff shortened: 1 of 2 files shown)


captainbarbosa:
Running `scripts/document.sh` results in `${RELEASE_VERSION}` being `0.17.0`, which is slightly unexpected to me given that our release page says we're on `0.35.0`:

Screenshot 2019-07-15 13 44 34

1ec5:
> Running `scripts/document.sh` results in `${RELEASE_VERSION}` being `0.17.0`, which is slightly unexpected to me given that our release page says we're on `0.35.0`:

You may need to run `git fetch --tags` to make sure your clone contains all the tags used to calculate the latest release version.

scripts/document.sh
```diff
@@ -60,10 +60,10 @@ jazzy \
     --documentation="docs/guides/*.md" \
     --root-url "${BASE_URL}/navigation/${RELEASE_VERSION}/" \
     --theme ${THEME} \
-    --output ${OUTPUT}
+    --output ${OUTPUT} \
+    --title "Navigation SDK for iOS" \
```
1ec5: Can we move this configuration setting to jazzy.yml?

captainbarbosa: Yes, but then I think we won't be able to access `${RELEASE_VERSION}`...?

1ec5: Just move `title` to jazzy.yml, but keep `module_version` here.

1ec5:
> Requires mapbox/jazzy-theme#9 to merge before merging.

I missed this bit. This SDK doesn’t depend on jazzy-theme yet; it hosts its own copy of the theme. We’ll need to make the change here:

https://github.com/mapbox/mapbox-navigation-ios/blob/4b22a6860597131cd7b7fb184fbc818750dd2939/docs/theme/templates/header.mustache#L7

PR #2182, opened 2019-07-16
Full-screen modal presentation bugfix

JThramer:
This PR specifies a preference for full screen modal presentation style, which is important in iOS 13.0, since the default seems to have shifted to a card presentation-style. 

The enum was introduced in iOS 3.2 and has no effect below iOS 13.0.

/cc @mapbox/navigation-ios


JThramer:
For some reason this nukes the bottom banner. Investigating.

docs/guides/Reduce Memory Usage.md
```diff
@@ -12,7 +12,7 @@ The Navigation SDK tries to compensate and be as energy conscious as possible. F
 
 Apps that use this SDK often begin by showing a preview map view where the route will go. Then, the user initiates navigation and the `NavigationViewController` is presented. However, the preview map is longer necessary to keep around in memory. 
 
-You can remove the preview map view from the current view using `removeFromSuperview`:
+UIKit will remove the MapView from the View Heirarchy when using a `.fullScreen` presentation style, but if you choose to use another presentation style, you can remove the preview map view from the current view using `removeFromSuperview`:
```
1ec5:
```suggestion
UIKit will remove `NavigationMapView` from the view hierarchy when using `UIModalPresentationStyle.fullScreen`, but if you choose to use another presentation style, you can remove the preview map view from the current view using `UIView.removeFromSuperview()`:
```

Example/ViewController.swift
```diff
@@ -125,9 +125,11 @@ class ViewController: UIViewController {
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-
-        self.mapView = NavigationMapView(frame: view.bounds)
-
+        
+        if mapView == nil {
+            mapView = NavigationMapView(frame: view.bounds)
```
andrlee: does it look like indentation is off? does xcode autoformating on our behalf?

andrlee: i beg your pardon @JThramer but couldn't be the same achieved with `lazy` property? which could also help with DRY on line 510?

1ec5: Regarding indentation, Xcode does automatically indent code. We stick to the default of four spaces per indent, which appears to be correct here. We’ve avoided adding a style linter because stuff like whitespace rarely comes up in review. Occasionally indentation gets flubbed and I’ll nitpick about it if it happens to hurt readability.

The actual fix could bear some explanation in inline comments, given that this is an example application. The application mostly follows [this guide](https://docs.mapbox.com/ios/api/navigation/0.36.0/reduce-memory-usage.html) for reducing memory usage in a multi-scene UI. The guide assumes that the map view is set up in the first place.

Previously, this view controller assumed that the default modal presentation style would cover up the whole screen, causing `viewDidAppear(_:)` to get called when exiting navigation. In iOS 13, the default changed to a card-like presentation, which meant the root view would never reappear because it always remained visible. This PR forces the full screen presentation, which was previously the default, and just in case, it also explicitly restores the map view when canceling navigation. As it happens, full screen presentation means that canceling navigation causes the root view to reappear, so both initializations could occur, hence the nil check.

A lazy property wouldn’t work because we do need the map view to be torn down and restored multiple times as the user enters, exits, and reenters navigation. Also, the `mapView` property not only holds the map view but also installs it in the view hierarchy. So setting this property has a side effect. Perhaps this is too much cleverness for an ostensible example application and explicit teardown and restoration would be more appropriate.

andrlee: Would have been nice to see this in PR description:
This PR forces the full screen presentation, which was previously the default, and just in case, it also explicitly restores the map view when canceling navigation. As it happens, full screen presentation means that canceling navigation causes the root view to reappear, so both initializations could occur, hence the nil check.

Example/ViewController.swift
```diff
@@ -311,6 +313,7 @@ class ViewController: UIViewController {
     }
 
     func presentAndRemoveMapview(_ navigationViewController: NavigationViewController, completion: CompletionHandler?) {
+        navigationViewController.modalPresentationStyle = .fullScreen
```
1ec5: Should we add this line to the “[Configuration](https://github.com/mapbox/mapbox-navigation-ios/#configuration)” section of the readme and any corresponding example code?

Example/ViewController.swift
JThramer: Good point. I'll try and be more explicit next time.

PR #2183, opened 2019-07-18
Fix auto-hiding status view

### LINKED ISSUE #2180: Device Volume Low status not disappearing
**Mapbox Navigation SDK version: 0.35.0**

After starting navigation with low volume a status is presented but it's not hidden after 3s.
It can disappear only if replaced with "Rerouting..." status.

Root cause:
`isCurrentlyVisible` flag in StatusView.swift is set to true only after `show()` animation is completed but `hide()` checks this flag before scheduling its animation when show animation is ongoing

I can suggest moving line 179 (guard) into the beginning of dispatch async body.

1ec5:
Fixed an issue where the “iPhone Volume Low” banner would persist until replaced by another banner about simulation or rerouting. Also expanded some code golf.

Fixes #2180. Thanks to @imaginaris for the fix!

/cc @JThramer

### DIFF
--- a/MapboxNavigation/StatusView.swift
+++ b/MapboxNavigation/StatusView.swift
@@ -176,10 +176,10 @@ public class StatusView: UIControl {
         }
         
         let animate = {
-            guard !self.isHidden, self.isCurrentlyVisible else { return }
-            
             let fireTime = DispatchTime.now() + delay
             DispatchQueue.main.asyncAfter(deadline: fireTime, execute: {
+                guard !self.isHidden, self.isCurrentlyVisible else { return }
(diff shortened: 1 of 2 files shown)


MapboxNavigation/StatusView.swift
```diff
@@ -176,17 +176,21 @@ public class StatusView: UIControl {
...
                 })
             })
         }
         
-        animated ? animate() : hide()
+        if animated {
+            animate()
+        } else {
+            hide()
+        }
```
JThramer: Sad doctor strange.

PR #2185, opened 2019-07-19
Fixing issue where steps controller grows from .zero when presented

### LINKED ISSUE #2165: Expanding the top instructions banner breaks animations
**Mapbox Navigation SDK version: 0.35.0**

When the default instruction banner view is expanded, the instruction labels can animate incorrectly as they are appearing. They can be seen growing from the 0,0 origin.

The offset of the map also seems to change, causing it to recenter to the user's location after closing the instructions. You can see the circular route and volume buttons moving with the banner change as well.

I am supplying `nil` as the `topBanner` to `NavigationOptions`.

Here is a zip of a screen recording showing the issues: [Banner.MOV.zip](https://github.com/mapbox/mapbox-navigation-ios/files/3336670/Banner.MOV.zip)

JThramer:
Fixes #2165. Sizes the steps table container before presentation so it does not grow during animation

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/TopBannerViewController.swift
+++ b/MapboxNavigation/TopBannerViewController.swift
@@ -158,20 +158,27 @@ import MapboxDirections
             return
         }
         
+        var stepsHeightConstraint: [NSLayoutConstraint] = []
+
         
         let controller = StepsViewController(routeProgress: progress)
         controller.delegate = self
-        let topHeight = view.bounds.height
-        let stepsTableExtent = UIScreen.main.bounds.height - topHeight
-        
+
         delegate?.topBanner?(self, willDisplayStepsController: controller)
         embed(controller, in: stepsContainer) { (parent, child) -> [NSLayoutConstraint] in
             child.view.translatesAutoresizingMaskIntoConstraints = false
+            
             let pinningConstraints = child.view.constraintsForPinning(to: self.stepsContainer)
             let hideConstraints = self.stepsContainerHideConstraints
-            let stepsHeight = child.view.heightAnchor.constraint(equalToConstant: stepsTableExtent)
+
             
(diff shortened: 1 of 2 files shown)


MapboxNavigation/TopBannerViewController.swift
```diff
@@ -161,14 +161,17 @@ import MapboxDirections
         
         let controller = StepsViewController(routeProgress: progress)
         controller.delegate = self
+        let topHeight = view.bounds.height
+        let stepsTableExtent = UIScreen.main.bounds.height - topHeight
```
frederoni: Avoid using the bounds of the screen. It could be a split view or embedded navigation view. Can you access the NavigationViewController’s height?

MapboxNavigation/TopBannerViewController.swift
```diff
@@ -158,17 +158,27 @@ import MapboxDirections
...
         delegate?.topBanner?(self, willDisplayStepsController: controller)
         embed(controller, in: stepsContainer) { (parent, child) -> [NSLayoutConstraint] in
             child.view.translatesAutoresizingMaskIntoConstraints = false
+            
             let pinningConstraints = child.view.constraintsForPinning(to: self.stepsContainer)
             let hideConstraints = self.stepsContainerHideConstraints
+
+            
+            if let bannerHostHeight = self.view.superview?.superview?.frame.height { 
+                print("Super: \(self.view.superview!.superview!)")
```
1ec5: This print statement should be removed.

MapboxNavigation/TopBannerViewController.swift
```diff
@@ -158,17 +158,27 @@ import MapboxDirections
             return
         }
         
+        var stepsHeightConstraint: [NSLayoutConstraint] = []
```
1ec5: There’s only one constraint, so it should be optional and not an array.

Issue #2145, opened 2019-06-01
Remove workarounds adding content inset when setting map camera


1ec5:
Once mapbox/mapbox-gl-native#14813 lands and we upgrade to map SDK v5.1.0, `NavigationMapView.userAnchorPoint` and possibly other methods need to be updated to stop adding the map view’s `contentInset` to the edge padding when setting the map’s camera.

/ref https://github.com/mapbox/mapbox-navigation-ios/issues/1651#issuecomment-418675194
/cc @mapbox/navigation-ios @fabian-guerra @d-prukop

1ec5:
In rudimentary testing, `NavigationMapView.userAnchorPoint` and `NavigationMapView.updateCourseTracking(location:camera:animated:)` seem to be behaving fine after upgrading to map SDK v5.1.0. That’s a bit of a surprising outcome, given that there was obviously a bug in the map SDK previously and it got fixed in v5.1.0.

1ec5:
Fixed in #2134.

1ec5:
I must not’ve been paying close enough attention, but `NavigationMapView.updateCourseTracking(location:camera:animated:)` is not quite correct. Upgrading from map SDK v5.0.0 to v5.1.0 causes the map to pivot around the center-bottom of the map instead of the puck’s center.

1ec5:
Possibly related to mapbox/mapbox-gl-native#15232.

1ec5:
> I must not’ve been paying close enough attention, but `NavigationMapView.updateCourseTracking(location:camera:animated:)` is not quite correct. Upgrading from map SDK v5.0.0 to v5.1.0 causes the map to pivot around the center-bottom of the map instead of the puck’s center.

This is probably mapbox/mapbox-gl-native#15233.

astojilj:
> Once [mapbox/mapbox-gl-native#14813](https://github.com/mapbox/mapbox-gl-native/pull/14813) lands and we upgrade to map SDK v5.1.0, `NavigationMapView.userAnchorPoint` and possibly other methods need to be updated to stop adding the map view’s `contentInset` to the edge padding when setting the map’s camera.
> 
> /ref [#1651 (comment)](https://github.com/mapbox/mapbox-navigation-ios/issues/1651#issuecomment-418675194)
> /cc @mapbox/navigation-ios @fabian-guerra @d-prukop

@1ec5, @fabian-guerra 

[...]

Issue #2190, opened 2019-07-26
CarPlay route preview does not fit to the route


1ec5:
When previewing a route on CarPlay in v0.35.0 with map SDK v5.2.0, the route rarely fits the screen. The camera never takes into account the left sidebar:

<img width="539" alt="" src="https://user-images.githubusercontent.com/1231218/61982314-3aad0300-afb1-11e9-8221-3946408f48c3.png">
<img width="539" alt="" src="https://user-images.githubusercontent.com/1231218/61982320-3da7f380-afb1-11e9-8d9c-0dd2aeeaf54e.png">
<img width="539" alt="" src="https://user-images.githubusercontent.com/1231218/61982306-354fb880-afb1-11e9-8054-7c277ab8f104.png">

It’s usually fine when the preview map first appears, but the problem starts as soon as you switch to an alternative route. At the point where `CarPlayManager.mapTemplate(_:selectedPreviewFor:using:)` calls `NavigationMapView.showcase(_:animated:)`, there’s never a left content inset, even though the content inset would’ve already been set to the CarPlay safe area insets in `CarPlayMapViewController.viewSafeAreaInsetsDidChange()`.

This issue appears to be caused primarily by a combination of mapbox/mapbox-gl-native#15232 and mapbox/mapbox-gl-native#15233, but it used to reproduce less frequently in previous versions of the map SDK, so there may be other compounding factors as well.

/cc @mapbox/navigation-ios @mapbox/maps-ios

astojilj:
@1ec5 The workaround in https://github.com/mapbox/mapbox-navigation-ios/pull/2211 makes this look correct. PTAL.

PR #2211, opened 2019-08-12
Simplify anchor point calculation and puck centering

astojilj:
Decouple dependency content insets -> anchor point -> padding/content insets broken after https://github.com/mapbox/mapbox-gl-native/pull/14664 introduced animated interpolation for padding change.

Remove the need to specify center for puck view - use the approach where content inset is keeping it centered.

Take safeArea into account when calculating contentInsets.

[Link to video showing current state.](https://www.dropbox.com/s/wori62tfhp69feg/RPReplay_Final1565767435.mp4?dl=0) 

EDIT 9/9/2019:
Work around mapbox-gl-native/15574. Panning fix.

Fix CarPlay puck positioning, panning away from navigation, navigation -> overview transitions.
Remove syncing safeArea to contentInset, contentInset is to be used to position puck and this causes map position sudden change when panning.

Work around mapbox-gl-native/15574: should be future proof as the mapbox/mapbox-gl-native#15574 exists only with MGLMapView.contentInset.

Addresses: https://github.com/mapbox/mapbox-gl-native/issues/15232, https://github.com/mapbox/mapbox-gl-native/issues/15233

Related to: https://github.com/mapbox/mapbox-navigation-ios/issues/2165,

Fixes: https://github.com/mapbox/mapbox-navigation-ios/issues/2145, #2190

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -148,19 +148,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         if let anchorPoint = navigationMapViewDelegate?.navigationMapViewUserAnchorPoint?(self), anchorPoint != .zero {
             return anchorPoint
         }
-        
-        // Inset by the safe area to avoid notches.
-        // Inset by the content inset to avoid application-defined content.
-        var contentFrame = bounds.inset(by: safeArea).inset(by: contentInset)
-        
-        // Avoid letting the puck go partially off-screen, and add a comfortable padding beyond that.
-        let courseViewBounds = userCourseView?.bounds ?? .zero
-        contentFrame = contentFrame.insetBy(dx: min(NavigationMapView.courseViewMinimumInsets.left + courseViewBounds.width / 2.0, contentFrame.width / 2.0),
-                                            dy: min(NavigationMapView.courseViewMinimumInsets.top + courseViewBounds.height / 2.0, contentFrame.height / 2.0))
-        
-        // Get the bottom-center of the remaining frame.
-        assert(!contentFrame.isInfinite)
-        return CGPoint(x: contentFrame.midX, y: contentFrame.maxY)
+        let contentFrame = bounds.inset(by: contentInset)
+        return CGPoint(x: contentFrame.midX, y: contentFrame.midY)
     }
     
     /**
@@ -194,8 +183,6 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             if let userCourseView = userCourseView {
                 if let location = userLocationForCourseTracking {
                     updateCourseTracking(location: location, animated: false)
-                } else {
-                    userCourseView.center = userAnchorPoint
                 }
                 addSubview(userCourseView)
             }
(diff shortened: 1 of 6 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -396,13 +396,30 @@ class RouteMapViewController: UIViewController {
...
+        var insets = mapView.safeArea
+        insets.top += instructionBannerHeight
+        insets.bottom += bottomBannerHeight
         
         if overviewing {
             insets += NavigationMapView.courseViewMinimumInsets
             
             let routeLineWidths = MBRouteLineWidthByZoomLevel.compactMap { $0.value.constantValue as? Int }
             insets += UIEdgeInsets(floatLiteral: Double(routeLineWidths.max() ?? 0))
+        } else if mapView.tracksUserCourse {
```
1ec5: RouteMapViewController is used by NavigationViewController but not CarPlayNavigationViewController. Developers are also responsible for implementing their own preview map or free-driving map using a standalone NavigationMapView. So this code would need to be duplicated in CarPlayNavigationViewController and application code. It would be better to keep the code in NavigationMapView if possible.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -194,8 +183,6 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             if let userCourseView = userCourseView {
                 if let location = userLocationForCourseTracking {
                     updateCourseTracking(location: location, animated: false)
-                } else {
-                    userCourseView.center = userAnchorPoint
```
1ec5: > Remove the need to specify center for puck view - use the approach where content inset is keeping it centered.

We need to check whether undoing #2047 ends up regressing #1731 and #1949. Also, does the puck consistently remain at `userAnchorPoint` when simulation is sped up?

/ref @frederoni

1ec5:
> Addresses: mapbox/mapbox-gl-native#15232, mapbox/mapbox-gl-native#15233

This is at most a workaround. A downstream change does not fix an upstream regression.

MapboxNavigation/NavigationMapView.swift
astojilj: > Also, does the puck consistently remain at userAnchorPoint when simulation is sped up?

Yes, it remains. It is always centered to padding area and no need to reposition it. Attached video to show:
https://www.dropbox.com/s/wori62tfhp69feg/RPReplay_Final1565767435.mp4?dl=0

Also, the issue here is fixed: https://github.com/mapbox/mapbox-gl-native/issues/15233#issuecomment-521127342

MapboxNavigation/RouteMapViewController.swift
astojilj: After https://github.com/mapbox/mapbox-navigation-ios/pull/2211/commits/98551ddcbf5934226702dc0ce710266d86f809ce

CarPlayNavigationViewController and NavigationViewController contentInset() functions are very close and seems feasible to extract common part to NavigationMapView.
As I don't have enough understanding of navigation-ios API surface. it is better if someone else from navigation-ios team continue on formulating the API, that is if the workarounds taken here are fine.

noway:
When this will be merged?

astojilj:
> When this will be merged? We can no longer use mapbox navagtion 0.34 with mapbox ios sdk 5.0.0 because it fails to compile with Swift 5.1 (which is forced with Xcode 11.0). This is a critical bug for us which blocks any releases we can do to our app. 😣
> 
> Having this bug in production is not an option because it critically affects the user experience

It is currently in review and being tested. Did you try if it fixes the issues in your application?
Thanks.

cc @frederoni @1ec5 @fabian-guerra

noway:
WORKS FABULOUSLY! Even closing the top nav bar stopped crashing the app!

For anyone who's coming across this, the commit id for this PR's branch is:

```
github "mapbox/mapbox-navigation-ios" "98551ddcbf5934226702dc0ce710266d86f809ce"
```

1ec5:
This PR will go in as soon as we’re able to verify that it doesn’t break navigation on a CarPlay device. In the meantime, if you don’t use CarPlay functionality, feel free to use this branch as described above.

MapboxNavigation/RouteMapViewController.swift
1ec5: We’re going to address duplication between CarPlayNavigationViewController and NavigationViewController overall in a separate PR ahead of v1.0.

PR #2215, opened 2019-08-24
Respect heading while walking

### LINKED ISSUE #1942: Rotate map based on heading while walking
When the user is walking, they can more easily turn around than when driving. It’s more important for the map to indicate the direction the user is facing than the direction they’re walking, in the event that they turn around to orient themselves. Essentially, `MGLUserTrackingMode.trackWithHeading` is more appropriate than `MGLUserTrackingMode.trackWithCourse` for the walking use case.

When the current step’s `transportType` is `walking`, the camera’s heading should be set to `CLHeading.trueHeading`, not `CLLocation.course`:

https://github.com/mapbox/mapbox-navigation-ios/blob/68745c937cc50928740dd900bcbd07c487072c9b/MapboxNavigation/NavigationMapView.swift#L336

/cc @mapbox/navigation-ios @lauri-rustanius

1ec5:
This PR contains these user-facing improvements to walking navigation:

* The user’s heading is used instead of the course when reorienting the camera while walking, since heading is more granular and more accurate at low speeds (i.e., walking speeds).
* When rerouting a pedestrian, avoid restricting the route’s initial heading. When driving or cycling, it makes sense to restrict the initial heading, to avoid routes that confusingly begin with a U-turn. However, a pedestrian can turn on a dime, and the direction in an initial “Head [direction] on [street]” instruction is more meaningful to a pedestrian than to a motorist or cyclist.

and these developer-facing improvements:

* Added `RouteControllerNotificationUserInfoKey.headingKey` to the user info dictionary of `Notification.Name.routeControllerWillReroute`, `Notification.Name.routeControllerDidReroute`, and `Notification.Name.routeControllerProgressDidChange` notifications. Core Location treats heading as first-class state instead of subordinating it under location, because heading updates from the magnetometer come in at a different rate than location updates from the GPS receiver.
* Both RouteController and LegacyRouteController can store a `heading`, so now the Router protocol requires it publicly. (Also fixed a latent issue where RouteController ignored heading updates. This issue didn’t matter previously because nothing was using `RouteController.heading`.)
* Deprecated the public but undocumented `NavigationMapView.updateCourseTracking(location:camera:animated:)` method in favor of a publicly documented `NavigationMapView.updateCourseTracking(location:heading:camera:animated:)` argument to the method for transitioning the map camera. (The existing method signature without the extra argument was undocumented but used in examples, so it has been retained in deprecated form.)

The user’s mode of transportation is determined by the current step’s `transportMode`, which is generally based on the routing profile but can be something else like `train` or `ferry` for walking and cycling routes. The improvements above affect only the walking transport mode.

To recap and to do:

* [x] Loosen rerouting heading while walking
* [x] Make heading a first-class piece of context alongside location
* [x] Rotate the map by heading instead of course while walking
* [ ] Initially rotate the map by heading instead of course when enabling course tracking mode while walking
* [ ] Update heading orientation: https://github.com/mapbox/mapbox-navigation-ios/issues/1942#issuecomment-479585406
* [ ] Test loosened rerouting – may be challenging given the lack of heading simulation

Fixes #1942.

### DIFF
--- a/MapboxCoreNavigation/LegacyRouteController.swift
+++ b/MapboxCoreNavigation/LegacyRouteController.swift
@@ -359,7 +359,7 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
 
         self.lastRerouteLocation = location
 
-        getDirections(from: location, along: progress) { [weak self] (route, error) in
+        getDirections(from: location, routeProgress: progress) { [weak self] (route, error) in
             guard let strongSelf = self else {
                 return
             }
--- a/MapboxCoreNavigation/LegacyRouteController.swift
+++ b/MapboxCoreNavigation/LegacyRouteController.swift
@@ -301,11 +301,14 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
             delegate?.router?(self, didUpdate: progress, with: location, rawLocation: rawLocation)
             
             //Fire the notification (for now)
-            NotificationCenter.default.post(name: .routeControllerProgressDidChange, object: self, userInfo: [
-                RouteControllerNotificationUserInfoKey.routeProgressKey: progress,
-                RouteControllerNotificationUserInfoKey.locationKey: location, //guaranteed value
-                RouteControllerNotificationUserInfoKey.rawLocationKey: rawLocation //raw
-                ])
+            var userInfo: [RouteControllerNotificationUserInfoKey: Any] = [
+                .routeProgressKey: progress,
+                .locationKey: location, //guaranteed value
+                .rawLocationKey: rawLocation, //raw
+            ]
+            userInfo[.headingKey] = heading
+            userInfo[.rawHeadingKey] = heading // heading snapping not yet implemented (an unnecessary?)
+            NotificationCenter.default.post(name: .routeControllerProgressDidChange, object: self, userInfo: userInfo)
         }
     }
         
@@ -354,9 +357,11 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
         isRerouting = true
 
         delegate?.router?(self, willRerouteFrom: location)
-        NotificationCenter.default.post(name: .routeControllerWillReroute, object: self, userInfo: [
-            RouteControllerNotificationUserInfoKey.locationKey: location
-        ])
+        var userInfo: [RouteControllerNotificationUserInfoKey: Any] = [
+            .locationKey: location,
+        ]
+        userInfo[.headingKey] = heading
+        NotificationCenter.default.post(name: .routeControllerWillReroute, object: self, userInfo: userInfo)
 
         self.lastRerouteLocation = location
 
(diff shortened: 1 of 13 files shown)


pintonos:
Is there currently a workaround like rotating the arrow dependent of the user location heading? Or is it not possible to change the arrow orientation from outside the box?

bsudekum:
Do you think this should also be applied to bikes?

1ec5:
> Is there currently a workaround like rotating the arrow dependent of the user location heading? Or is it not possible to change the arrow orientation from outside the box?

There isn’t currently a workaround, but this PR implements this behavior out of the box.

> Do you think this should also be applied to bikes?

Not really, because I think it’s a bit of a sledgehammer specific to walking. A pedestrian on foot has a turning radius of 0 and can turn on a dime, whereas a bike may need to find a good place to turn around. Also, I’d be annoyed if I saw the map wiggle around to match my handlebar as I struggle to get going. 😅

I think there will be some opportunities for rerouting heuristics that benefit cyclists as well as pedestrians, but it probably will be orthogonal to using the magnetometer.

MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -300,11 +301,14 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
...
-                RouteControllerNotificationUserInfoKey.locationKey: location, //guaranteed value
-                RouteControllerNotificationUserInfoKey.rawLocationKey: rawLocation //raw
-                ])
+            var userInfo: [RouteControllerNotificationUserInfoKey: Any] = [
+                .routeProgressKey: progress,
+                .locationKey: location, //guaranteed value
+                .rawLocationKey: rawLocation, //raw
+            ]
+            userInfo[.headingKey] = heading
+            userInfo[.rawHeadingKey] = heading // heading snapping not yet implemented (an unnecessary?)
```
JThramer: YAGNI?

1ec5: We could get rid of `rawHeadingKey`, but that could be problematic if we ever do need to implement heading snapping for some reason. We’d want the non-`raw` values to be snapped when a `raw` value is available. That said, the comment refers to the prospect that heading snapping may not be especially useful.

JThramer:
Blocked by https://github.com/mapbox/mapbox-navigation-ios/pull/2319

AhmerKhan1:
Had a call with a delivery customer, Navigan, this morning.  They are having issues with the puck movement on NavSDK 2.0 while on an active route while walking.  They mentioned the movement of the 3D puck is delayed by several seconds.  As they are walking, the puck is stuck in one position.  After several seconds, it jumps and then catches up with them.  While they are seeing some of this delay in a vehicle, the issue is more pronounced while walking.  

Minh mentioned that the Turn-by-turn navigation experience is not optimized for walking.  We may want to use the applications built-in location functionality instead of the PassiveLocationManager for slow speeds.  

I am waiting for the logs to be sent by Navigan, but can we optimize the location movement for slower speeds such as walking/scooters/bikes to remove the effects of the puck delay?

MaximAlien:
Closing this PR in favor of https://github.com/mapbox/mapbox-navigation-ios/pull/3620.

PR #2224, opened 2019-08-30
Adding method for progress updates on the NVC delegate.

### LINKED ISSUE #2212: NavigationService Delagate doesn't delegate, when used by NavigationViewController
**Mapbox Navigation SDK version: 0.36.0**

I'm interested in some information passed by the NavigationServiceDelegate, but the function is not exposed by NavigationViewController.

In NavigationViewController you can read that you can use the NavigationService Delegate.

`To be informed of significant events and decision points as the user progresses along the route, set the 'NavigationService.delegate' property of the 'NavigationService' that you provide when creating the navigation options.`

But this doesn't not work, because navigationservice delegate only calls either your class or NavigationServiceDelegate.

Im Interested in 
`public func router(_ router: Router, didUpdate progress: RouteProgress, with location: CLLocation, rawLocation: CLLocation)`

I did made changes to NavigationViewController to expose that function. is there any reason not to do that or am i missing an easier way to get that information

JThramer:
Fixes #2212 by adding a progress update method to `NavigationViewControllerDelegate`. 

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master
 * Fixed an issue where swiping the banner down after the StepsTableViewController has already displayed could put the UI in an unstable state. ([#2197](https://github.com/mapbox/mapbox-navigation-ios/pull/2197))
-* Added `NavigationViewControllerDelegate.navigationViewController(_:didUpdate:location:rawLocation)` method that allows the client developer to capture progress updates without having to inject themselves between the `NavigationViewController` and the `NavigationService`.  ([#2224](https://github.com/mapbox/mapbox-navigation-ios/pull/2224))
+* Added `NavigationViewControllerDelegate.navigationViewController(_:didUpdate:location:rawLocation)` method that allows you to capture progress updates without having to inject a class between the `NavigationViewController` and the `NavigationService`.  ([#2224](https://github.com/mapbox/mapbox-navigation-ios/pull/2224))
 
 ## v0.36.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master
 * Fixed an issue where swiping the banner down after the StepsTableViewController has already displayed could put the UI in an unstable state. ([#2197](https://github.com/mapbox/mapbox-navigation-ios/pull/2197))
-* Added `NavigationViewControllerDelegate.navigationViewController(_:didUpdate:location:rawLocation)` method that allows the client developer to capture progress updates without having to inject themselves between the `NavigationViewController` and the `NavigationService`. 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master
 * Fixed an issue where swiping the banner down after the StepsTableViewController has already displayed could put the UI in an unstable state. ([#2197](https://github.com/mapbox/mapbox-navigation-ios/pull/2197))
+* Added `NavigationViewControllerDelegate.navigationViewController(_:didUpdate:location:rawLocation)` method that allows the client developer to capture progress updates without having to inject themselves between the `NavigationViewController` and the `NavigationService`.  ([#2224](https://github.com/mapbox/mapbox-navigation-ios/pull/2224))
```
1ec5: Does this work?

```suggestion
* Added `NavigationViewControllerDelegate.navigationViewController(_:didUpdate:location:rawLocation)` method that allows you to capture progress updates without having to inject a class between the `NavigationViewController` and the `NavigationService`.  ([#2224](https://github.com/mapbox/mapbox-navigation-ios/pull/2224))
```

andrlee: i wonder what if we decide to cherry-pick this commit at some point to the patch release branch? would coupling changelog record with the actual fix be a limitation?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -494,6 +494,9 @@ extension NavigationViewController: NavigationServiceDelegate {
             userHasArrivedAndShouldPreventRerouting {
             mapViewController?.mapView.updateCourseTracking(location: location, animated: true)
         }
+
```
andrlee: just out of curiosity, what's the internal guideline wrt `for-loops` and closures like `forEach` usage in nav sdk? specifically referring to line 474.

CHANGELOG.md
1ec5: If we were to cherry-pick this change, we’d most likely revise the changelog entry on both branches to match.

There has been a lot of discussion about how to manage changelogs in the map SDK projects. The main benefit to including the changlog blurb in the change is that we don’t have to go back and try to remember what this PR was about at release time. (In the old days, we would try to put together a Paper document, but it easily fell out of sync and took extra time at release.)

Fortunately, we’ve only had occasional need for one-off patch branches so far, so it hasn’t posed a problem for us. Over time, if the bandwidth of development picks up to the point where we need to maintain long-running release branches, we could consider a different strategy.

MapboxNavigation/NavigationViewController.swift
1ec5: My personal preference is for for-in loops unless a function is being passed into forEach as an argument, so sometimes I nitpick about lines like that. Others have a higher tolerance for these forEach calls, but I view it as an unnecessary barrier to readability.

PR #2230, opened 2019-09-16
1.0 Objective-C Obsoletion

### LINKED ISSUE #2231: Discontinue Objective-C compatibility
We’re discontinuing support for applications written purely in Objective-C to use the iOS navigation SDK. The same change will take place in MapboxDirections.swift: mapbox/MapboxDirections.swift#381. Applications written in Objective-C will need to implement compatibility shims in Swift to continue to interoperate with this library.

Since #3, we’ve half-heartedly supported Objective-C by ensuring bridgeability of key APIs, but there are many gaps in coverage, and the public API in Objective-C feels like a translation from another language. A renewed focus on Swift will eliminate a major source of technical debt. We’ll be able to write more idiomatic, less error-prone Swift code without compromising the developer experience in both languages.

Hybrid application frameworks (such as React Native) are implemented in Objective-C. Discontinuing Objective-C support makes it more difficult – but not infeasible – for a developer to write a hybrid wrapper around the SDK. For instance, a React Native wrapper would need an additional compatibility shim written in Swift to translate the navigation SDK’s API to a more basic API that can bridge to Objective-C. Regardless, we do not officially support hybrid application frameworks.

Discontinuing Objective-C support will have the following immediate impacts on the codebase:

* `@objc` attributes and `MB` class prefixes will disappear, except in some edge cases that require Objective-C method dispatch (for key-value observation).
* Methods whose sole purpose is to bridge to Objective-C will disappear.
* Enumeration declarations will move from Objective-C headers to Swift files.
* References to Objective-C compatibility will be removed from readmes and jazzy guides.

We’ll hopefully be able to take advantage of some Swift language features in the process:

* Convert value classes to structures or enumerations with associated types.
* Convert abstract classes to protocols with default implementations.
* Move global variables and global constants under structures and classes.
* Replace magic default values with optionals.
* Replace NSCoding with Codable.

/cc @mapbox/navigation-ios

JThramer:
Obsolete. With _extreme_ prejudice.

> No man is an island,
> Entire of itself.
> Each is a piece of the continent,
> A part of the main.
> If a clod be washed away by the sea,
> Europe is the less.
> As well as if a promontory were.
> As well as if a manor of thine own
> Or of thine friend's were.
> Each man's death diminishes me,
> For I am involved in mankind.
> Therefore, send not to know
> For whom the bell tolls,
> It tolls for thee.
-- _For Whom The Bell Tolls_, by John Donne

Work in progress.

Running Checklist: 
- [x] Eliminate Objective-C support
   - [x] Delete `@objc` attributes and `MB` class prefixes
   - [x] Fix resultant build errors.
   - [x] Delete methods whose sole purpose is to bridge to Objective-C
   - [x] Fix broken Guidance Cards
   - [x] Fix any other test failures
   - [x] Move enumeration declarations from Objective-C headers to Swift files
   - [x] Delete `CMapboxDirections` target from SPM distribution of `MapboxDirections.swift`
   - [x] Remove references to Objective-C compatibility from readmes and jazzy guides
- [x] Take advantage of Swift language features
    - [x] Convert value classes to structures or enumerations with associated types
    - [x] Convert abstract classes to protocols with default implementations
    - [x] Move global variables and global constants under structures and classes
- [x]  ~~Replace magic default values with optionals~~

Fixes #2231.


Example/ViewController.swift
```diff
@@ -360,6 +358,12 @@ class ViewController: UIViewController {
     }
 }
 
+extension ViewController: UnimplementedLogging {
+    var delegateIdentifier: String {
+        return "navigationMapViewDelegate+navigationViewControllerDelegate+voiceControllerDelegate+waypointConfirmationViewControllerDelegate"
```
1ec5: The intent of these identifiers is to make the console spew more explicit about the protocol that isn’t fully implemented. However, I was concerned that this COM-reminiscent approach would be fragile – imagine copying a conforming class and forgetting to update the identifier, or making the class conform to an additional protocol without updating the identifier. So @JThramer is going to modify `logUnimplemented(level:function:)` to take an explicit protocol name as an argument. This way the console spew will state the precise protocol that the unimplemented method belongs to.

JThramer:
Some lessons learned so far:

There are a few unanticipated complications that may impact our refactoring objectives. Namely: 

- KVO is only possible through NSObject Inheritance (OBJC), and that affects classes like `RouteProgress` and `NavigationSettings`.
- UIAppearance [inherits from](https://developer.apple.com/documentation/uikit/uiappearance) `NSObjectProtocol`, which pins `Style` and `DayStyle` to our current implementation without refactoring our theming mechanism.

Example/ViewController.swift
JThramer: Indeed. And it's done as of 38514ba.

MapboxCoreNavigationTests/DistanceFormatterTests.swift
```diff
@@ -1,170 +1,162 @@
...
+//import XCTest
+//import CoreLocation
+//@testable import MapboxCoreNavigation
+//
+//let oneMile: CLLocationDistance = 1_609.344
+//let oneYard: CLLocationDistance = 0.9144
+//let oneFeet: CLLocationDistance = 0.3048
+//
+//
+//class DistanceFormatterTests: XCTestCase {
```
1ec5: This entire test case needs to be restored and fixed.

MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -47,9 +47,9 @@ class NavigationServiceTests: XCTestCase {
         delegate.reset()
     }
     
-    func testDefaultUserInterfaceUsage() {
-        XCTAssertTrue(dependencies.navigationService.eventsManager.usesDefaultUserInterface, "MapboxCoreNavigationTests should have an implicit dependency on MapboxNavigation due to running inside the Example application target.")
-    }
+//    func testDefaultUserInterfaceUsage() {
+//        XCTAssertTrue(dependencies.navigationService.eventsManager.usesDefaultUserInterface, "MapboxCoreNavigationTests should have an implicit dependency on MapboxNavigation due to running inside the Example application target.")
```
1ec5: This test should be restored and fixed.

MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -165,6 +165,69 @@ public protocol CarPlayManagerDelegate {
      - parameter carPlayManager: The CarPlay manager instance.
      - returns: A Boolean value indicating whether to disable idle timer when carplay is connected and enable when disconnected.
      */
-    @objc optional func carplayManagerShouldDisableIdleTimer(_ carPlayManager: CarPlayManager) -> Bool
+    func carplayManagerShouldDisableIdleTimer(_ carPlayManager: CarPlayManager) -> Bool
+}
+
+@available(iOS 12.0, *)
+extension CarPlayManagerDelegate {
```
1ec5: This default implementation extension doesn’t work: the CarPlay example application fails to build with the following error:

```
/path/to/mapbox-navigation-ios/Example/AppDelegate+CarPlay.swift:47:1: Type 'AppDelegate' does not conform to protocol 'CarPlayManagerDelegate'
/path/to/mapbox-navigation-ios/Example/AppDelegate+CarPlay.swift:47:1: Do you want to add protocol stubs?
```

And the fix-it adds protocol stubs for all the methods implemented below. I think the problem is that the extension is internal whereas the protocol is public.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -329,19 +330,11 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             // Animate course view updates in overview mode
             UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear], animations: { [weak self] in
                 guard let point = self?.convert(location.coordinate, toPointTo: self) else { return }
-                self?.userCourseView?.center = point
+                self?.userCourseView.center = point
             })
         }
         
-        if let userCourseView = userCourseView as? UserCourseView {
-            if let customTransformation = userCourseView.update?(location: location, pitch: self.camera.pitch, direction: direction, animated: animated, tracksUserCourse: tracksUserCourse) {
```
1ec5: [Codecov](https://codecov.io/gh/mapbox/mapbox-navigation-ios/compare/4ae1d627619c385a38cbf9903cf334152a0375de...20fb6a0bc84d175c740a5e27e77c05a8b9499d66/changes) is showing a 72-line drop in test coverage in UserCourseView.swift, because `CourseUpdatable.update(location:pitch:direction:animated:tracksUserCourse:)`, `CourseUpdatable.applyDefaultUserPuckTransformation(location:pitch:direction:animated:tracksUserCourse:)`, and `UserPuckStyleKitView.draw(_:)` don’t get called. I wonder if the deletion of this chunk of code has anything to do with it.

Codecov is right about there being a problem. The puck never draws:

puck

MapboxNavigation/NavigationMapView.swift
```diff
@@ -162,30 +159,34 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    public typealias UserCourseView = UIView & CourseUpdatable
+    
+    /**
+     A `UserCourseView` used to indicate the user’s location and course on the map.
      
-     If the view conforms to `UserCourseView`, its `UserCourseView.update(location:pitch:direction:animated:)` method is frequently called to ensure that its visual appearance matches the map’s camera.
+     The `UserCourseView`'s `UserCourseView.update(location:pitch:direction:animated:)` method is frequently called to ensure that its visual appearance matches the map’s camera.
      */
-    @objc public var userCourseView: UIView? {
+    public var userCourseView: UserCourseView = UserPuckCourseView(frame: CGRect(origin: .zero, size: 75.0)) {
```
1ec5: The course view no longer appears on the map because of this change. `didSet` only gets called after the property is explicitly set, not after it is initialized to a default value. As a result, the course view is never added as a subview of the map view.

1ec5:
512db91d6edb4c460235af1e955767be15e5aaa9 restores the puck to the map view. Using the regular expression `(?:let|var)\s.+=.+\{$\s*didSet`, I verified that we aren’t relying on `didSet` to be triggered by default initialization anywhere else in the codebase.

MapboxNavigation/Constants.swift
```diff
@@ -44,5 +44,40 @@ extension Notification.Name {
...
      The user info dictionary contains the key `StyleManagerNotificationUserInfoKey.style` and `StyleManagerNotificationUserInfoKey.styleManager`.
      */
-    public static let styleManagerDidApplyStyle = NSNotification.Name.MBStyleManagerDidApplyStyle
+    public static let styleManagerDidApplyStyle: Notification.Name = "MBStyleManagerDidApplyStyle"
 }
+
+/**
+ Keys in the user info dictionaries of various notifications posted by instances of `StyleManager`.
+ */
+public struct StyleManagerNotificationUserInfoKey: Hashable, Equatable, RawRepresentable, ExpressibleByStringLiteral {
```
1ec5: Does this need to be ExpressibleByStringLiteral?

MapboxNavigation/DataCache.swift
```diff
@@ -49,7 +49,7 @@ public class DataCache: NSObject, BimodalDataCache {
     /*
      Clears out the memory cache.
      */
-    public func clearMemory() {
+    @objc public func clearMemory() {
```
1ec5: Why did this method gain an `@objc` that it didn’t already have?

MapboxNavigation/NavigationComponent.swift
```diff
@@ -5,43 +5,43 @@ import CoreLocation
 /*
  A navigation component is a member of the navigation UI view hierarchy that responds as the user progresses along a route according to the `NavigationServiceDelegate` protocol.
  */
-@objc public protocol NavigationComponent: NavigationServiceDelegate {}
+public typealias NavigationComponent = NavigationServiceDelegate
```
1ec5: Do we expect that NavigationComponent will never have method and property requirements of its own beyond the protocols that the type conforms to?

This means new navigation component types no longer need to explicitly declare that they’re navigation components. Instead, a type becomes a navigation component automatically by virtue of conforming to NavigationServiceDelegate. Could that lead to a class being used in situations it didn’t expect to be used in?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -440,13 +435,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     /**
      Adds or updates both the route line and the route line casing
      */
-    @objc public func showRoutes(_ routes: [Route], legIndex: Int = 0) {
+    public func showRoutes(_ routes: [Route], legIndex: Int = 0) {
```
1ec5: By the way, now that we only support Swift, we can use more concise method names like:

```swift
public func show(_ routes: [Route], legIndex: Int = 0) {
```

without worrying about conflicts with similarly named methods that take different arguments.

MapboxNavigation/NavigationViewControllerDelegate.swift
```diff
@@ -185,18 +185,123 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
...
+        return nil
+    }
+    
+    func navigationViewController(_ navigationViewController: NavigationViewController, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage? {
+        logUnimplemented(protocolType: NavigationViewControllerDelegate.self,  level: .debug)
+        return nil
+    }
+    
+    func navigationViewController(_ navigationViewController: NavigationViewController, viewFor annotation: MGLAnnotation) -> MGLAnnotationView? {
+        logUnimplemented(protocolType: NavigationViewControllerDelegate.self,  level: .debug)
```
1ec5: How can we automatically verify that all the methods of an UnimplementedLogging-conforming class call `logUnimplemented(protocolType:level:)`? In other languages, this is what you’d use a custom decorator or annotation for. Maybe an automated test could use SourceKitten to compare the number of methods and calls to `logUnimplemented(protocolType:level:)`? Or perhaps UnimplementedLogging could require the class to register each logged method into an array, and the protocol could check at runtime whether any methods are missing from that array.

README.md
```diff
@@ -92,7 +92,7 @@ Consult the [API reference](https://docs.mapbox.com/ios/api/navigation/) for fur
 
 The [API reference](https://docs.mapbox.com/ios/api/navigation/) includes example code for accomplishing common tasks. You can run these examples as part of the [navigation-ios-examples](https://github.com/mapbox/navigation-ios-examples) project.
 
-This repository also contains [Swift and Objective-C testbeds](https://github.com/mapbox/mapbox-navigation-ios/tree/master/Example) that exercise a variety of navigation SDK features:
+This repository also contains [testbeds](https://github.com/mapbox/mapbox-navigation-ios/tree/master/Example) that exercise a variety of navigation SDK features:
```
1ec5:
```suggestion
This repository also contains [a testbed](https://github.com/mapbox/mapbox-navigation-ios/tree/master/Example) that exercises a variety of navigation SDK features:
```

docs/cover.md
```diff
@@ -4,7 +4,7 @@
 
 The Mapbox Navigation SDK gives you all the tools you need to add turn-by-turn navigation to your application. It takes just a few minutes to drop a full-fledged turn-by-turn navigation view controller into your application. Or use the Core Navigation framework directly to build something truly custom.
 
-The Mapbox Navigation SDK and Core Navigation are compatible with applications written in Swift 4 or Objective-C in Xcode 9.0. The Mapbox Navigation and Mapbox Core Navigation frameworks run on iOS 9.0 and above.
+The Mapbox Navigation SDK and Core Navigation are compatible with applications written in Swift 5 in Xcode 9.0. The Mapbox Navigation and Mapbox Core Navigation frameworks run on iOS 10.0 and above.
```
1ec5: Make the same change to README.md.

MapboxNavigation/DataCache.swift
JThramer: https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/DataCache.swift#L14

MapboxNavigation/NavigationMapView.swift
JThramer: Not sure if this is resolved. un-resolving until i'm positive.

1ec5:
A couple test failures have crept in:

```
/Users/distiller/project/MapboxCoreNavigationTests/RouteProgressTests.swift:46: error: -[MapboxCoreNavigationTests.RouteProgressTests testNextRouteStepProgress] : XCTAssertEqual failed: ("439.1") is not equal to ("inf")
/Users/distiller/project/MapboxCoreNavigationTests/RouteProgressTests.swift:34: error: -[MapboxCoreNavigationTests.RouteProgressTests testRouteStepProgress] : XCTAssertEqual failed: ("384.1") is not equal to ("inf")
```

Example/ViewController.swift
```diff
@@ -410,10 +407,16 @@ extension ViewController: NavigationMapViewDelegate {
 // MARK: VoiceControllerDelegate methods
 // To use these delegate methods, set the `VoiceControllerDelegate` on your `VoiceController`.
 extension ViewController: VoiceControllerDelegate {
+    // called when there is an error that requires the speech controller to fall back to a native engine.
+    func voiceController(_ voiceController: RouteVoiceController, didFallBackTo engine: AVSpeechSynthesizer, becauseOf error: SpeechError) {
```
1ec5: No need to work the final argument into a sentence:

```swift
func voiceController(_ voiceController: RouteVoiceController, didFallBackTo engine: AVSpeechSynthesizer, error: SpeechError) {
```

MapboxNavigation/StyleManager.swift
```diff
@@ -1,49 +1,63 @@
...
      */
-    @objc(locationForStyleManager:)
     func location(for styleManager: StyleManager) -> CLLocation?
     
     /**
      Informs the delegate that a style was applied.
+     - note: This delegate method includes a default implementation that prints a warning to the console when this method is called. See `UnimplementedLogging` for details.
      
-     This delegate method is the equivalent of `Notification.Name.styleManagerDidApplyStyle`.
+     - note: This delegate method is the equivalent of `Notification.Name.styleManagerDidApplyStyle`.
```
1ec5: `note`s are pretty loud. Can this just be a normal paragraph before the other `note`?

1ec5:
I got a crash arriving at the final destination:

```
#0	0x00007fff518292c6 in __pthread_kill ()
#1	0x00007fff518d0bf1 in pthread_kill ()
#2	0x00007fff517b9a3c in abort ()
#3	0x00000001080e09e5 in swift::fatalError(unsigned int, char const*, ...) ()
#4	0x00000001080d8de7 in swift::swift_dynamicCastFailure(void const*, char const*, void const*, char const*, char const*) ()
#5	0x00000001080d8e5a in swift::swift_dynamicCastFailure(swift::TargetMetadata<swift::InProcess> const*, swift::TargetMetadata<swift::InProcess> const*, char const*) ()
#6	0x00000001080dafd3 in swift_dynamicCastClassUnconditional ()
#7	0x0000000105e780c3 in closure #1 in RouteMapViewController.endOfRouteViewController.getter at /path/to/mapbox-navigation-ios/MapboxNavigation/RouteMapViewController.swift:20
#8	0x0000000105e77f07 in RouteMapViewController.endOfRouteViewController.getter at /path/to/mapbox-navigation-ios/MapboxNavigation/RouteMapViewController.swift:18
#9	0x0000000105e85928 in RouteMapViewController.embedEndOfRoute() at /path/to/mapbox-navigation-ios/MapboxNavigation/RouteMapViewController.swift:363
#10	0x0000000105e86497 in RouteMapViewController.showEndOfRoute(duration:completion:) at /path/to/mapbox-navigation-ios/MapboxNavigation/RouteMapViewController.swift:384
#11	0x0000000105e31c41 in NavigationViewController.showEndOfRouteFeedback(duration:completionHandler:) at /path/to/mapbox-navigation-ios/MapboxNavigation/NavigationViewController.swift:539
#12	0x0000000105e319f9 in NavigationViewController.navigationService(_:didArriveAt:) at /path/to/mapbox-navigation-ios/MapboxNavigation/NavigationViewController.swift:532
#13	0x0000000105e32ca9 in protocol witness for NavigationServiceDelegate.navigationService(_:didArriveAt:) in conformance NavigationViewController ()
#14	0x0000000107c75cfd in MapboxNavigationService.router(_:didArriveAt:) at /path/to/mapbox-navigation-ios/MapboxCoreNavigation/NavigationService.swift:464
#15	0x0000000107c76529 in protocol witness for RouterDelegate.router(_:didArriveAt:) in conformance MapboxNavigationService ()
#16	0x0000000107c63425 in RouteController.updateRouteLegProgress(status:) at /path/to/mapbox-navigation-ios/MapboxCoreNavigation/RouteController.swift:275
```

The error is:

```
2019-11-12 19:05:27.940108-0800 Example-CarPlay[43862:4033688] [Storyboard] Unknown class MBEndOfRouteViewController in Interface Builder file.
```

Please audit the storyboards for any Objective-C class names.

MapboxCoreNavigation/UnimplementedLogging.swift
```diff
@@ -13,14 +14,31 @@ public protocol UnimplementedLogging {
...
+        
+        guard !alreadyWarned else {
+            return
+        }
+        
         let log = OSLog(subsystem: "com.mapbox.navigation", category: "delegation.\(selfDescription)")
-        let formatted: StaticString = "Unimplemented Delegate Method in %@: %@.%@"
-        os_log(formatted, log: log, type: level, selfDescription, protocolDescription, function)
+           let formatted: StaticString = "Unimplemented Delegate Method in %@: %@.%@. This message will only be logged once."
+           os_log(formatted, log: log, type: level, selfDescription, protocolDescription, function)
```
1ec5: Nit: misleading indentation.

MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -507,7 +507,7 @@ class NavigationServiceTests: XCTestCase {
         
         let ourLogs = logs.filter { $0.0 == "EmptyNavigationServiceDelegate" }
         
-        XCTAssertEqual(ourLogs.count, 520, "Expected logs to be populated and expected number of messages sent")
+        XCTAssertEqual(ourLogs.count, 4, "Expected logs to be populated and expected number of messages sent")
```
1ec5: 😌

MapboxNavigation/Error.swift
```diff
@@ -70,28 +35,34 @@ public enum SpeechError: LocalizedError {
...
     /**
      The speech engine was unable to initalize an audio player.
      - parameter playerType: the type of `AVAudioPlayer` that failed to initalize.
      - parameter instruction: The instruction that failed.
-     - parameter engine: The `SpeechEngine` that attempted the initalization.
+     - parameter engine: The speech engine that attempted the initalization.
      - parameter underlying: the `Error` that was returned by the system audio service.
      */
-    case unableToInitalizePlayer(playerType: AVAudioPlayer.Type, instruction: SpokenInstruction, engine: SpeechEngine, underlying: Error)
+    case unableToInitializePlayer(playerType: AVAudioPlayer.Type, instruction: SpokenInstruction, engine: Any?, underlying: Error)
```
1ec5: “Engine” is fairly odd to see by itself like this. I get that it comes from “text-to-speech engine”, but the iOS term for a TTS engine is “speech synthesizer”, so “synthesizer” would be a better label to use here.

PR #2275, opened 2019-11-25
Upgrade to post-Objective-C MapboxDirections.swift

JThramer:
This PR is to track integration between #2230 (now merged) and mapbox/mapboxdirections.swift#382.

Burn-down list:
- [x] Fix error issues
- [x] JSON Passthrough
- [x] `NSCoder` -> `Codable` refactorings
- [x] copy constructor for `RouteOptions`

Depends on mapbox/MapboxDirections.swift#393.

/cc @mapbox/navigation-ios


MapboxCoreNavigation/RouteController.swift
```diff
@@ -287,7 +288,7 @@ open class RouteController: NSObject {
         let step = stepProgress.step
         
         //Increment the progress model
-        let polyline = Polyline(step.coordinates!)
+        let polyline = Polyline(step.shape!.coordinates)
```
1ec5: Avoid copying the Polyline:

```suggestion
        let polyline = step.shape!
```

MapboxCoreNavigation/RouteController.swift
```diff
@@ -355,9 +356,14 @@ open class RouteController: NSObject {
 
 extension RouteController: Router {
     public func userIsOnRoute(_ location: CLLocation) -> Bool {
+        
+        guard let destination = routeProgress.currentLeg.destination else {
```
1ec5: Instead of defensively swallowing the `nil` destination, let’s take this opportunity to throw an assertion. If the current leg progress contains a leg that was decoded independently of any response, then I’d consider that an internal consistency error.

MapboxCoreNavigation/RouteLeg.swift
```diff
@@ -3,6 +3,6 @@ import Turf
 
 extension RouteLeg {
     var coordinates: [CLLocationCoordinate2D] {
-        return (steps.first?.coordinates ?? []) + steps.dropFirst().flatMap { ($0.coordinates ?? []).dropFirst() }
+        return (steps.first?.shape?.coordinates ?? []) + steps.dropFirst().flatMap { ($0.shape?.coordinates ?? []).dropFirst() }
```
1ec5: Consider defining a `+` operator on two LineStrings that concatenates the LineStrings, deduplicating the last coordinate of the first argument and the first argument of the second argument. (This operator could be part of Turf.) Then this method could become something like:

```swift
return steps.dropFirst().compactMap { $0.shape }.reduce(steps.first.shape ?? Polyline())
```

Even better would be to define a `+=` to avoid excessive copying.

MapboxCoreNavigation/RouteOptions.swift
```diff
@@ -24,6 +24,11 @@ extension RouteOptions {
         return copy
     }
     
+//    init(options: RouteOptions) {
```
1ec5: Is the idea to implement copying functionality? I think it would be better to keep pushing on mapbox/MapboxDirections.swift#388 with an eye towards turning RouteOptions into a struct, but it would be OK to continue to conform to NSCopying.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -436,7 +440,7 @@ open class RouteLegProgress: NSObject {
         let remainingSteps = leg.steps.suffix(from: stepIndex)
 
         for (currentStepIndex, step) in remainingSteps.enumerated() {
-            guard let coords = step.coordinates else { continue }
+            guard let coords = step.shape?.coordinates else { continue }
             guard let closestCoordOnStep = Polyline(coords).closestCoordinate(to: coordinate) else { continue }
```
1ec5: Use `step.shape` instead of copying the coordinates into a new Polyline.

MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -163,8 +163,8 @@ open class SimulatedLocationManager: NavigationLocationManager {
         
         // Simulate speed based on expected segment travel time
         if let expectedSegmentTravelTimes = routeProgress?.currentLeg.expectedSegmentTravelTimes,
-            let coordinates = routeProgress?.route.coordinates,
-            let closestCoordinateOnRoute = Polyline(routeProgress!.route.coordinates!).closestCoordinate(to: newCoordinate),
+            let coordinates = routeProgress?.route.shape?.coordinates,
+            let closestCoordinateOnRoute = Polyline(coordinates).closestCoordinate(to: newCoordinate),
```
1ec5: Use `routeProgress?.route.shape` instead of copying the coordinates into a new Polyline.

MapboxCoreNavigation/RouteOptions.swift
```diff
@@ -3,13 +3,24 @@ import MapboxDirections
 extension RouteOptions {
     internal var activityType: CLActivityType {
         switch self.profileIdentifier {
-        case MBDirectionsProfileIdentifier.cycling, MBDirectionsProfileIdentifier.walking:
+        case DirectionsProfileIdentifier.cycling, DirectionsProfileIdentifier.walking:
```
1ec5: Nit: drop `DirectionsProfileIdentifier`.

MapboxCoreNavigation/Route.swift
```diff
@@ -0,0 +1,13 @@
+import MapboxDirections
+
+extension Route {
+    
+    var json: String? {
```
1ec5: Encoding as JSON is expensive enough that we probably shouldn’t hand-wave away the fact that this computed variable creates a new structure every time it is accessed. Instead, make it a method.

Once we implement mapbox/MapboxDirections.swift#391, the original JSON data can live alongside the decoded Route in a Response object.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -421,9 +425,9 @@ open class RouteLegProgress: NSObject {
      */
     @available(*, deprecated, message: "Use RouteProgress.nearbyCoordinates")
     public var nearbyCoordinates: [CLLocationCoordinate2D] {
```
1ec5: This property has long been deprecated. Can we delete this property and the coordinate array concatenation along with it?

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -63,105 +63,100 @@ class InstructionPresenter {
...
             }
         }
         
         assert(processedComponents.count == strings.count, "The number of processed components must match the number of attributed strings")
         return (components: processedComponents, attributedStrings: strings)
     }
 
-    func attributedString(forExitComponent component: VisualInstructionComponent, maneuverDirection: ManeuverDirection, dataSource: DataSource) -> NSAttributedString? {
-        guard component.type == .exitCode, let exitCode = component.text else { return nil }
+    func attributedString(forExitRepresentation representation: VisualInstruction.Component.TextRepresentation, maneuverDirection: ManeuverDirection, dataSource: DataSource, cacheKey: String) -> NSAttributedString? {
```
1ec5: The new VisualInstruction.Component enumeration makes it tempting to consolidate these “attributed string for” methods into a single method with a switch statement.

MapboxNavigation/LaneView.swift
```diff
@@ -51,12 +51,12 @@ open class LaneView: UIView {
...
     static let defaultFrame: CGRect = CGRect(origin: .zero, size: 30.0)
     
-    convenience init(component: LaneIndicationComponent) {
+    convenience init(indications: LaneIndication, isUsable: Bool) {
         self.init(frame: LaneView.defaultFrame)
         backgroundColor = .clear
-        lane = Lane(indications: component.indications)
-        maneuverDirection = ManeuverDirection(description: component.indications.description)
-        isValid = component.isUsable
+        lane = Lane(indications: indications)
```
1ec5: Store the LaneIndication directly; no need to use the Lane type here.

MapboxCoreNavigation/EventDetails.swift
```diff
@@ -120,14 +120,14 @@ struct NavigationEventDetails: EventDetails {
...
+            let coordinates = dataSource.routeProgress.route.shape?.coordinates,
             let lastCoord = coordinates.last {
             userAbsoluteDistanceToDestination = location.distance(from: CLLocation(latitude: lastCoord.latitude, longitude: lastCoord.longitude))
         } else {
             userAbsoluteDistanceToDestination = nil
         }
         
-        if let geometry = session.originalRoute.coordinates {
+        if let geometry = session.originalRoute.shape?.coordinates {
             originalGeometry = Polyline(coordinates: geometry)
```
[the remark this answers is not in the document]
1ec5: This turned out to be Polyline from the Polyline framework, not the type alias for LineString from the Turf framework.

MapboxCoreNavigation/EventDetails.swift
```diff
@@ -139,7 +139,7 @@ struct NavigationEventDetails: EventDetails {
             originalStepCount = nil
         }
         
-        if let geometry = session.currentRoute.coordinates {
+        if let geometry = session.currentRoute.shape?.coordinates {
             self.geometry = Polyline(coordinates: geometry)
```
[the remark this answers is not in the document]
1ec5: This turned out to be Polyline from the Polyline framework, not the type alias for LineString from the Turf framework.

MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -280,7 +284,10 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
         let step = stepProgress.step
         
         //Increment the progress model
-        let polyline = Polyline(step.coordinates!)
+        guard let polyline = step.shape else {
+            return
```
1ec5: Add a precondition failure here, because the lack of a shape is a serious internal consistency error. `RouteStep.shape` is now mutable unlike before.

MapboxCoreNavigation/RouteOptions.swift
```diff
@@ -3,13 +3,24 @@ import MapboxDirections
...
             return .fitness
         default:
             return .automotiveNavigation
         }
     }
     
+    convenience init(options: RouteOptions) {
```
1ec5: Since this copy constructor lives outside of MapboxGeocoder.swift, it’ll be prone to bitrot. Fortunately, we’ll be able to remove it once mapbox/MapboxDirections.swift#388 lands.

MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -33,7 +33,7 @@ class FeedbackEvent: CoreFeedbackEvent {
 
 class RerouteEvent: CoreFeedbackEvent {
     func update(newRoute: Route) {
-        if let geometry = newRoute.coordinates {
+        if let geometry = newRoute.shape?.coordinates {
             eventDictionary["newGeometry"] = Polyline(coordinates: geometry).encodedPolyline
```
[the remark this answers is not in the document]
1ec5: This turned out to be Polyline from the Polyline framework, not the type alias for LineString from the Turf framework.

1ec5:
This PR blocks #2114: https://github.com/mapbox/mapbox-navigation-ios/pull/2114#issuecomment-568151846.

1ec5:
We’re down to 45 compiler errors in MapboxCoreNavigationTests as of a3df84fb94520700ea0549260451aad8317f6253. All the other targets now build, but I haven’t looked at the test results yet.

1ec5:
Many tests are failing because notifications aren’t getting posted and navigation service delegate methods aren’t getting called.

MapboxCoreNavigation/RouteOptions.swift
1ec5: Indeed, it’s missing several DirectionsOptions properties required for routing, like `includesSteps`.

1ec5:
The latest commits fix the issue where the navigator would never progress along the route, but unfortunately the navigator considers almost every location update to be off-route.

1ec5:
`MBNavigator.getRouteGeometry()` returns locations all in the ballpark of 3.91010900°N, 8.45129600°W, which suggests that MapboxDirections.swift is encoding the geometry as a Polyline-encoded string based on an incorrect precision.

1ec5:
Route preview:

preview

`Route.json`’s return value (which is fed into `RouteController.updateNavigator(with:)`) includes a geometry that, when decoded as Polyline with precision 6, looks like this:

json

1ec5:
The remaining MapboxCoreNavigationTests failures are due to extraneous rerouting. For some reason, the navigation service records a reroute twice when changing the route during simulation: once with the last recorded location before changing the route, and once with the new route’s starting location.

```
		NavigationEventsManagerTests.testDepartRerouteArrive()
		NavigationEventsManagerTests.testDepartRerouteArrive()
		NavigationServiceTests.testProactiveRerouting()
		NavigationServiceTests.testUnimplementedLogging()
```

1ec5:
By the time `NavigationEventsManagerTests.testDepartRerouteArrive()` has finished setting the new route and begins feeding the navigator locations from the second route, `MBNavigator.updateLocation(for:)` is called with the first location of the second route and `MBNavigator.getStatusForTimestamp(_:)` is called immediately afterwards. But that call to `getStatusForTimestamp(_:)` returns an MBNavigationStatus whose `coordinate` is from the previous route. Apart from `location` and `routeState`, which is `offRoute`, all the other fields in the status reflect the new route.

https://github.com/mapbox/mapbox-navigation-ios/blob/daa0d4681353975978347c7a105f963010c32ca8/MapboxCoreNavigation/RouteController.swift#L189-L191

1ec5:
From reading the navigation-native source code, it looks like `MBNavigator.updateLocation(for:)` can [fail](https://github.com/mapbox/mapbox-navigation-native/blob/6efebf5040358c82b7e2a01c1eed0cda364234f3/src/navigator_impl_location.cpp#L75) if the location is internally considered inaccurate. We’ve been discarding the return value of this method call.

Setting `RouteController.route` results in reroute-like behavior, but that isn’t always the intent of setting that property. We should differentiate between `route` getting set as a result of rerouting and `route` getting set intentionally. In the latter case, we should find a way to completely reset the navigator, erasing any previous state.

I’m at a loss as to why the locations only seem to get rejected with the changes in this PR and not on master.

MapboxNavigation/InstructionPresenter.swift
```diff
@@ -63,105 +74,103 @@ class InstructionPresenter {
         return attributedPairs.attributedStrings
     }
     
-    typealias AttributedInstructionComponents = (components: [VisualInstructionComponent], attributedStrings: [NSAttributedString])
+    typealias AttributedInstructionComponents = (components: [VisualInstruction.Component], attributedStrings: [NSAttributedString])
     
     func attributedPairs(for instruction: VisualInstruction, dataSource: DataSource, imageRepository: ImageRepository, onImageDownload: @escaping ImageDownloadCompletion) -> AttributedInstructionComponents {
```
1ec5: The changes to this method cause the compiler to segfault consistently in Xcode 10:

```
1.	While emitting IR SIL function "@$s16MapboxNavigation20InstructionPresenterC15attributedPairs3for10dataSource15imageRepository15onImageDownloadSay0A10Directions06VisualC0C9ComponentOG10components_SaySo18NSAttributedStringCG0E7StringstAK_AA0cd4DataI0_pAA0mK0CySo7UIImageCSgctF".
 for 'attributedPairs(for:dataSource:imageRepository:onImageDownload:)' (at /Users/distiller/project/MapboxNavigation/InstructionPresenter.swift:79:5)
0  swift                    0x0000000110470ee3 PrintStackTraceSignalHandler(void*) + 51
1  swift                    0x00000001104706bc SignalHandler(int) + 348
2  libsystem_platform.dylib 0x00007fff64c1ab5d _sigtramp + 29
3  libsystem_platform.dylib 0x00007ffee39ba650 _sigtramp + 2128214800
4  swift                    0x000000010c44bc73 swift::irgen::SingleScalarTypeInfo<(anonymous namespace)::BridgeObjectTypeInfo, swift::irgen::ReferenceTypeInfo>::copy(swift::irgen::IRGenFunction&, swift::irgen::Explosion&, swift::irgen::Explosion&, swift::irgen::Atomicity) const + 51
5  swift                    0x000000010c486e4e swift::irgen::RecordTypeInfo<(anonymous namespace)::LoadableStructTypeInfo, swift::irgen::LoadableTypeInfo, (anonymous namespace)::StructFieldInfo, true, true>::copy(swift::irgen::IRGenFunction&, swift::irgen::Explosion&, swift::irgen::Explosion&, swift::irgen::Atomicity) const + 94
6  swift                    0x000000010c486e4e swift::irgen::RecordTypeInfo<(anonymous namespace)::LoadableStructTypeInfo, swift::irgen::LoadableTypeInfo, (anonymous namespace)::StructFieldInfo, true, true>::copy(swift::irgen::IRGenFunction&, swift::irgen::Explosion&, swift::irgen::Explosion&, swift::irgen::Atomicity) const + 94
7  swift                    0x000000010c486e4e swift::irgen::RecordTypeInfo<(anonymous namespace)::LoadableStructTypeInfo, swift::irgen::LoadableTypeInfo, (anonymous namespace)::StructFieldInfo, true, true>::copy(swift::irgen::IRGenFunction&, swift::irgen::Explosion&, swift::irgen::Explosion&, swift::irgen::Atomicity) const + 94
8  swift                    0x000000010c486e4e swift::irgen::RecordTypeInfo<(anonymous namespace)::LoadableStructTypeInfo, swift::irgen::LoadableTypeInfo, (anonymous namespace)::StructFieldInfo, true, true>::copy(swift::irgen::IRGenFunction&, swift::irgen::Explosion&, swift::irgen::Explosion&, swift::irgen::Atomicity) const + 94
9  swift                    0x000000010c4d8b45 swift::SILInstructionVisitor<(anonymous namespace)::IRGenSILFunction, void>::visit(swift::SILInstruction*) + 10053
10 swift                    0x000000010c4d3a4a swift::irgen::IRGenModule::emitSILFunction(swift::SILFunction*) + 9898
11 swift                    0x000000010c3915c0 swift::irgen::IRGenerator::emitGlobalTopLevel(bool) + 1232
12 swift                    0x000000010c4b3414 performIRGeneration(swift::IRGenOptions&, swift::ModuleDecl*, std::__1::unique_ptr<swift::SILModule, std::__1::default_delete<swift::SILModule> >, llvm::StringRef, swift::PrimarySpecificPaths const&, llvm::LLVMContext&, swift::SourceFile*, llvm::GlobalVariable**) + 1076
13 swift                    0x000000010c2d5319 performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&, swift::FrontendObserver*, swift::UnifiedStatsReporter*) + 49433
14 swift                    0x000000010c2c592e swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 6862
15 swift                    0x000000010c2639ce main + 1246
16 libdyld.dylib            0x00007fff64a353d5 start + 1
17 libdyld.dylib            0x00000000000000ef start + 2606542107
error: Segmentation fault: 11
```

JThramer: Looks like it's passing a swift-only type to an obj-c method. Will fix.

1ec5:
bf9892a9b9964c5ead4f4429a264f9f8bca48d32 fixes `NavigationEventsManagerTests.testDepartRerouteArrive()` by ignoring location updates that the navigator says to ignore. But that breaks `NavigationServiceTests.testReroutingFromALocationSendsEvents()`.

JThramer:
> [bf9892a](https://github.com/mapbox/mapbox-navigation-ios/commit/bf9892a9b9964c5ead4f4429a264f9f8bca48d32) fixes `NavigationEventsManagerTests.testDepartRerouteArrive()` by ignoring location updates that the navigator says to ignore. But that breaks `NavigationServiceTests.testReroutingFromALocationSendsEvents()`.

@1ec5, could this perhaps be by-design in NN?

1ec5:
I’m totally willing to concede that bf9892a9b9964c5ead4f4429a264f9f8bca48d32 is the wrong fix for `testDepartRerouteArrive()`. But I don’t know how else to ensure that we reroute immediately when programmatically setting a new route but don’t also reroute upon the first subsequent location update.

MapboxNavigation/InstructionPresenter.swift
1ec5: Maybe we should just drop Xcode 10 support, as in #2282.

1ec5:
`testUnimplementedLogging()` is failing because of an `invalid` route state in the navigator. The navigator is complaining about another Codable roundtripping issue:

```
(lldb) po (service.router as! RouteController).navigator.getStatusForTimestamp(trace.last!.timestamp).stateMessage
"text on banner component is required."
```

1ec5:
I’m thinking we should probably log or assert when `MBNavigationStatus.stateMessage` is a non-empty string. It seems to keep cropping up as an indicator of a significant round-tripping problem.

1ec5:
Remaining test failures:

```
	MapboxNavigationTests:
		InstructionsBannerViewIntegrationTests.testDelimiterDisappearsOnlyWhenAllShieldsHaveLoaded()
		InstructionsBannerViewIntegrationTests.testRouteShieldsAreGenericUntilTheyLoad()
		NavigationViewControllerTests.testDestinationAnnotationUpdatesUponReroute()
		NavigationViewControllerTests.testDestinationAnnotationUpdatesUponReroute()
		RouteControllerSnapshotTests.testRouteSnappingOvershooting()
		RouteTests.testPolylineAroundManeuver()
		SimulatedLocationManagerTests.testSimulateRouteDoublesBack()
```

MapboxCoreNavigationTests/CocoaPodsTest/PodInstall/Podfile
```diff
@@ -3,6 +3,7 @@ use_frameworks!
 
 target 'PodInstall' do
   # The branch of MapboxNavigation and MapboxNavigation will be replaced by Bitrise to use the current branch.
+  pod 'MapboxDirections.swift', :git => 'https://github.com/mapbox/MapboxDirections.swift.git', :branch => 'jerrad/delenda-duet-dependencies'
```
1ec5: This change and the corresponding change to Podfile.lock need to be updated once mapbox/MapboxDirections.swift#393 lands in master and reverted once MapboxDirections.swift v1.0 is released.

1ec5:
This PR [finally passes](https://circleci.com/gh/mapbox/mapbox-navigation-ios/13822) the Xcode 11 / iOS 13 build bot. Xcode 10 / iOS 12 build bots are still crashing.

MapboxCoreNavigation/RouteLeg.swift
1ec5: LineString is a struct, and its `coordinates` property is variable, so it is possible to reduce down into a single LineString.

PR #2278, opened 2019-11-27
Specify ManeuverView and LaneView as non-opaque

avi-c:
This is needed since they implement their own draw(rect:) functions and any transparent pixels will appear as black since they are not automatically blended and composited with the rest of the view hierarchy.


avi-c:
Abandoning in favor of https://github.com/mapbox/mapbox-navigation-ios/pull/2279 to simplify git merge to the v0.38 release branch.

Issue #2223, opened 2019-08-30
Switch to night style when Dark Mode is enabled


1ec5:
When Dark Mode is enabled in iOS 13 and above, NavigationViewController should default to the night style regardless of the time of day.

[This guide to Dark Mode](https://developer.apple.com/documentation/appkit/supporting_dark_mode_in_your_interface) recommends using semantic colors and custom color assets, which long-term would be a superior solution to the SDK’s Style system, at least as far as colors are concerned. [There is more granular control](https://developer.apple.com/documentation/appkit/supporting_dark_mode_in_your_interface/choosing_a_specific_interface_style_for_your_ios_app) over the mode by setting [`UIView.overrideUserInterfaceStyle`](https://developer.apple.com/documentation/uikit/uiview/3238086-overrideuserinterfacestyle). These APIs would be more reliable than Style, but they require iOS 13 or above.

#539 would similarly switch styles when Smart Invert Colors is enabled.

/cc @mapbox/navigation-ios

1ec5:
User interface style is a [trait](https://developer.apple.com/documentation/uikit/uitraitcollection/1651063-userinterfacestyle) that we could observe changes to.

zappycode:
Hey I'd like to do this as part of Hacktober. Do I need to be assigned?

MaximAlien:
To better align between iOS and CarPlay we should also consider adding switch that allows users to decide whether they want to change UI appearance whenever it changes in settings.

PR #2283, opened 2019-12-06
Lighten maneuver icons in user notifications in Dark Mode

### LINKED ISSUE #2266: Local Notification direction is hard to see in iOS 13 Dark Mode
The direction arrow shown in the local notification is hard to see in iOS 13 when the device is set to dark mode.  The arrow is black and dark mode notifications are much darker than light mode notifications, making the arrow hard to see.  It would be really nice if the arrow could be white or off-white when the device is set to dark mode.  

The black arrow in light mode looks great. 

Please see my screenshot.  Let me know if you need any additional info from me.

IMG_9011

0.38.0

1ec5:
When Dark Mode is enabled, draw user notifications’ maneuver icons in white instead of black for better contrast.

Fixes #2266.


1ec5:
That’s a good point. Fortunately, once we switch to night style in Dark Mode (#2223), I think the only other place not covered by the Style system would be CarPlay.

PR #2288, opened 2019-12-17
Use built-in ISO date formatter

1ec5:
Now that we’ve dropped support for iOS 9._x_, use ISO8601DateFormatter instead of rolling our own ISO&nbsp;8601 date formatter.

/cc @mapbox/navigation-ios


1ec5:
Reverted in #2327.

PR #2290, opened 2019-12-20
Clean up RouteController user info keys

1ec5:
#2230 left some references to Objective-C symbols in documentation comments. This PR updates the references to the corresponding Swift symbols. It also renames RouteControllerNotificationUserInfoKey to RouteController.NotificationUserInfoKey.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxCoreNavigation/CoreConstants.swift
+++ b/MapboxCoreNavigation/CoreConstants.swift
@@ -119,42 +119,42 @@ public extension Notification.Name {
     /**
      Posted when `RouteController` receives a user location update representing movement along the expected route.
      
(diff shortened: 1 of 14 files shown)


MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -119,47 +119,47 @@ public extension Notification.Name {
     /**
      Posted when `RouteController` receives a user location update representing movement along the expected route.
      
-     The user info dictionary contains the keys `MBRouteControllerRouteProgressKey` and `MBRouteControllerLocationKey`.
+     The user info dictionary contains the keys `RouteControllerNotificationUserInfoKey.routeProgressKey` and `RouteControllerNotificationUserInfoKey.locationKey`.
```
1ec5: It isn’t great that “Key” is repeated in the constant name. Can we remove it?

PR #2296, opened 2020-01-06
Make InstructionsCardViewController reusable independently of NavigationComponent

avi-c:
Also re-expose the Route JSON string for use in debug metrics database.


MapboxCoreNavigation/Route.swift
```diff
@@ -2,7 +2,7 @@ import MapboxDirections
 
 extension Route {
     
-    var json: String? {
+    public var json: String? {
```
1ec5: Per https://github.com/mapbox/mapbox-navigation-ios/pull/2275#discussion_r358536495, this is potentially an expensive operation, so we should communicate that to clients by exposing it as a method instead of a computed variable. It’s also very specific to the needs of MapboxNavigationNative – returning a string instead of Data or a JSON-formatted dictionary – so I’m kind of thinking it would be better for any application that needs this functionality to reimplement it locally. JSONEncoder is public API, part of the Swift standard library.

MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -13,7 +13,7 @@ open class InstructionsCardViewController: UIViewController {
     var instructionsCardLayout: InstructionsCardCollectionLayout!
     
     public private(set) var isInPreview = false
-    private var currentStepIndex: Int?
+    public var currentStepIndex: Int?
```
1ec5: How do we expect the application to use this property?

MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -130,7 +130,7 @@ open class InstructionsCardViewController: UIViewController {
         NSLayoutConstraint.activate(instructionCollectionViewContraints)
     }
     
-    func reloadDataSource() {
+    open func reloadDataSource() {
```
1ec5: In what scenarios would the application need to force a reload because the view controller isn’t automatically reloading?

MapboxNavigation/InstructionsCardViewController.swift
avi-c: @d-prukop - Can you comment on what you use this property for in Apex? Is it possible we can avoid needing it in a client app?

MapboxNavigation/InstructionsCardViewController.swift
d-prukop: @1ec5 I tracked it down. The application actually doesn't need these things. This was part of an effort to prevent unnecessary passing of the navigation service to the guidance cards.

Currently the guidance cards have this code
```
extension InstructionsCardViewController: NavigationComponent {
    public func navigationService(_ service: NavigationService, didUpdate progress: RouteProgress, with location: CLLocation, rawLocation: CLLocation) {
        routeProgress = progress
        reloadDataSource()
    }
    
    public func navigationService(_ service: NavigationService, didPassVisualInstructionPoint instruction: VisualInstructionBanner, routeProgress: RouteProgress) {
        self.routeProgress = routeProgress
        reloadDataSource()
    }
    
    public func navigationService(_ service: NavigationService, didRerouteAlong route: Route, at location: CLLocation?, proactive: Bool) {
        self.currentStepIndex = nil
        self.routeProgress = service.routeProgress
        reloadDataSource()
    }
}
```

and I wanted to try it out with this code

```
extension InstructionsCardViewController {
    func routeProgressDidUpdate(_ progress: RouteProgress) {
        routeProgress = progress
        reloadDataSource()
    }
    
    func routeProgressDidPassVisualInstructionPoint(_ progress: RouteProgress) {
        routeProgress = progress
        reloadDataSource()
    }
    
    func routeProgressDidReroute(_ progress: RouteProgress) {
        currentStepIndex = nil
        routeProgress = progress
        reloadDataSource()
    }
}
```

This way we don't need to expose the entire nav service to the guidance cards or the UI side of the app.

I think we should move to make this change in `InstructionsCardViewController`

MapboxNavigation/InstructionsCardViewController.swift
d-prukop: @avi-c see my comments below. This and the exposing of `reloadDataSource` are related

MapboxNavigation/InstructionsCardViewController.swift
1ec5: Yes, let’s fold the extension into the main implementation.

MapboxCoreNavigation/Route.swift
```diff
@@ -2,7 +2,7 @@ import MapboxDirections
 
 extension Route {
     
-    public var json: String? {
+    public func jsonRepresentation() -> String? {
```
avi-c: @1ec5 - Is this what you had in mind? Does the naming of the function seem appropriate?

1ec5: Something like that, thanks. But I still think the method should be private since it’s such a specific implementation detail. Any application could trivially implement the same functionality if needed, and not just for Route but for RouteLeg, Intersection, etc.

MapboxNavigation/InstructionsCardViewController.swift
1ec5: Per chat, replacing the existing extension with the one proposed in https://github.com/mapbox/mapbox-navigation-ios/pull/2296#discussion_r363524424 would break NavigationComponent conformance and thus NavigationViewController compatibility.

By design, all navigation components need to be aware of the navigation service because that’s how they get access to the route progress, and delegate methods always need to refer to the delegator. (Alternatively, a component could observe notifications, but we moved away from that mechanism because components kept failing to filter for specific senders.) We didn’t consider that it might be desirable to encapsulate NavigationService from UI code. (cc @JThramer)

Making `currentStepIndex` and `reloadDataSource()` public allows an application to reuse InstructionsCardViewController in a non-NavigationViewController environment while wrapping these NavigationServiceDelegate methods in custom methods that don’t necessarily expose NavigationService. Looking ahead to when we make the guidance card feature public, we can document these members as being public for the purpose of subclassing. (In general, we could do a better job of adding subclassing notes to class documentation anyways.)

PR #2301, opened 2020-01-08
Require map SDK v5.5.0 in v0.37.x

1ec5:
This change disallows versions of the map SDK beyond v5.5._x_. It targets the v0.37._x_ release series.

/ref #2299
/cc @mapbox/navigation-ios @mapbox/maps-ios

PR #2302, opened 2020-01-08
Require map SDK v5.6.0

### LINKED ISSUE #2299: Crash on launch — +[NSUserDefaults mme_configuration]: unrecognized selector — conflicting MapboxMobileEvents dependency in Carthage
An application that installs this SDK via Carthage is guaranteed to crash as soon as an access token is provided to MapboxMobileEvents, which typically happens on launch.

## Diagnosis

As of #2208, MapboxNavigation depends on map SDK v5.2.0–v5._x_. As of mapbox/mapbox-gl-native-ios#60, the map SDK depends on a copy of MapboxMobileEvents provided by CocoaPods or Carthage rather than an embedded, statically-linked copy. This change made it into v5.6.0, a minor version update.

Since the map SDK is distributed as a binary artifact via a JSON manifest and `binary` specification, Carthage doesn’t handle this dependency; instead, clients of the map SDK are expected to explicitly depend on MapboxMobileEvents. MapboxCoreNavigation has long depended on MapboxMobileEvents, but it currently depends on v0.9.5, whereas the map SDK apparently requires [v0.10.0–v0.10._x_ for Carthage](https://www.mapbox.com/install/ios/carthage/) (but [specifically v0.10.2 for CocoaPods](https://github.com/mapbox/mapbox-gl-native-ios/blob/ios-v5.6.0/platform/ios/Mapbox-iOS-SDK.podspec#L32)).

Carthage does nothing to resolve this incompatibility; instead, the application crashes at runtime:

```
2020-01-07 09:50:21.014288-0800 Example[82655:123407] +[NSUserDefaults mme_configuration]: unrecognized selector sent to class 0x10be0c100
2020-01-07 09:50:21.031286-0800 Example[82655:123407] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '+[NSUserDefaults mme_configuration]: unrecognized selector sent to class 0x10be0c100'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010bb401ee __exceptionPreprocess + 350
	1   libobjc.A.dylib                     0x000000010b9adb20 objc_exception_throw + 48
	2   CoreFoundation                      0x000000010bb61054 +[NSObject(NSObject) doesNotRecognizeSelector:] + 132
	3   CoreFoundation                      0x000000010bb44f6c ___forwarding___ + 1436
	4   CoreFoundation                      0x000000010bb470f8 _CF_forwarding_prep_0 + 120
	5   Mapbox                              0x000000010ad3b802 -[MGLMapboxEvents init] + 262
	6   Mapbox                              0x000000010ad3ca9a __33+[MGLMapboxEvents sharedInstance]_block_invoke + 38
	7   libdispatch.dylib                   0x0000000110e3c781 _dispatch_client_callout + 8
	8   libdispatch.dylib                   0x0000000110e3d9ad _dispatch_once_callout + 20
	9   Mapbox                              0x000000010ad3c2ee +[MGLMapboxEvents sharedInstance] + 97
	10  Mapbox                              0x000000010ad3c32c +[MGLMapboxEvents setupWithAccessToken:] + 60
	11  libdispatch.dylib                   0x0000000110e3b810 _dispatch_call_block_and_release + 12
	12  libdispatch.dylib                   0x0000000110e3c781 _dispatch_client_callout + 8
	13  libdispatch.dylib                   0x0000000110e48caa _dispatch_main_queue_callback_4CF + 1212
	14  CoreFoundation                      0x000000010baa2e49 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
	15  CoreFoundation                      0x000000010ba9daa9 __CFRunLoopRun + 2329
	16  CoreFoundation                      0x000000010ba9ce66 CFRunLoopRunSpecific + 438
	17  XCTest                              0x000000012b2009f9 -[XCTWaiter waitForExpectations:timeout:enforceOrder:] + 989
	18  XCTest                              0x000000012b19a30f -[XCTFuture _waitForFulfillment] + 724
	19  XCTest                              0x000000012b19b122 -[XCTFuture value] + 32
	20  XCTest                              0x000000012b2207a8 +[XCTRunnerIDESession daemonMediatedSessionForSessionIdentifier:delegate:error:] + 214
	21  XCTest                              0x000000012b21508a _XCTestMain + 858
	22  libXCTestBundleInject.dylib         0x0000000108fdfbe7 __copy_helper_block_e8_32s + 0
	23  CoreFoundation                      0x000000010baa309c __CFRUNLOOP_IS_CALLING_OUT_TO_A_BLOCK__ + 12
	24  CoreFoundation                      0x000000010baa2808 __CFRunLoopDoBlocks + 312
	25  CoreFoundation                      0x000000010ba9d694 __CFRunLoopRun + 1284
	26  CoreFoundation                      0x000000010ba9ce66 CFRunLoopRunSpecific + 438
	27  GraphicsServices                    0x00000001124ffbb0 GSEventRunModal + 65
	28  UIKitCore                           0x0000000117396dd0 UIApplicationMain + 1621
	29  Example                             0x0000000108eb417b main + 75
	30  libdyld.dylib                       0x0000000110ebdd29 start + 1
	31  ???                                 0x0000000000000005 0x0 + 5
)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

## Resolution

The crash occurs after running `carthage update`, as @avi-c did in #2296. A temporary workaround is to only run `carthage bootstrap` with a checked-in Carthage.resolved that sticks to map SDK v5.5.0.

As a matter of good hygiene, we should upgrade MapboxCoreNavigation to v0.10.2. We may also need to patch v0.38.1 to require specifically map SDK v5.5._x_. However, this problem will recur every time the navigation SDK or map SDK needs a new version of MapboxMobileEvents, because the map SDK essentially relies on an implicit dependency with no safeguards around dependency versions.

/cc @mapbox/navigation-ios @mapbox/maps-ios @mapbox/mobile-telemetry

1ec5:
This change requires map SDK v5.6._x_–v5._x_. It targets the v1.0._x_ release series.

By comparison, #2301 requires v5.5._x_ and targets the v0.37._x_ release series.

Fixes #2299.

/cc @mapbox/navigation-ios @mapbox/maps-ios @mapbox/mobile-telemetry


MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -85,9 +85,8 @@ open class NavigationEventsManager: NSObject {
     func start() {
         let eventLoggingEnabled = UserDefaults.standard.bool(forKey: NavigationMetricsDebugLoggingEnabled)
         
-        mobileEventsManager.isDebugLoggingEnabled = eventLoggingEnabled
-        mobileEventsManager.isMetricsEnabledInSimulator = true
-        mobileEventsManager.isMetricsEnabledForInUsePermissions = true
+        
+        UserDefaults.standard.mme_isCollectionEnabled = eventLoggingEnabled
```
rclee: This still needs a change. Either replace `UserDefaults.standard.mme_isCollectionEnabled` with some implementation of `mobileEventsManager.isDebugLoggingEnabled` or remove entirely for the time being. `mme_isCollectionEnabled` should be handled by the Events Library. 

We'll add a plist flag for debug logging in a later release.

rclee: cc: @alfwatt

MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -85,9 +85,8 @@ open class NavigationEventsManager: NSObject {
     func start() {
         let eventLoggingEnabled = UserDefaults.standard.bool(forKey: NavigationMetricsDebugLoggingEnabled)
         
-        mobileEventsManager.isDebugLoggingEnabled = eventLoggingEnabled
-        mobileEventsManager.isMetricsEnabledInSimulator = true
```
1ec5: We decided to remove this setting for the same reason expressed in https://github.com/mapbox/mapbox-navigation-ios/pull/304#pullrequestreview-50951199. If navigation events do need to be sent while running in the simulator, we can reenable that workflow in a separate PR.

MapboxCoreNavigation/NavigationEventsManager.swift
1ec5: I think we should remove this bit, as well as the `NavigationMetricsDebugLoggingEnabled` constant. This debug logging is an advanced workflow that is only relevant when we develop more event types, so we could just enable it locally when needed without involving the SDK.

PR #2303, opened 2020-01-09
v0.37.1

1ec5:
Updated the Xcode project and CocoaPods podspecs for v0.37.1.

/ref #2299
/cc @mapbox/navigation-ios


1ec5:
The Xcode_10.2_iOS_12.2_CP_update build is expected to fail, since we haven’t pushed a v0.37.1 pod to CocoaPods trunk yet.

PR #2310, opened 2020-01-15
Clean up map layer identifiers

1ec5:
Identifiers of map data sources and style layers added by this SDK are now prefixed by the MapboxNavigation bundle identifier to avoid potential collisions with the style or other customizations made by the application. Organized the various identifiers into structs to avoid confusing source identifiers and style layer identifiers with each other. Also refactored the removal of map overlays to be less error-prone.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/MGLStyle.swift
+++ b/MapboxNavigation/MGLStyle.swift
@@ -103,4 +103,24 @@ extension MGLStyle {
         }
         return URL(string: "mapbox://styles/mapbox/navigation-guidance-night-v\(version)")!
     }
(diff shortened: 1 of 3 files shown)


1ec5:
There are no instances of `layer(withIdentifier: SourceIdentifier.` or `source(withIdentifier: StyleLayerIdentifier.`, which hopefully means I didn’t confuse any sources with style layers or vice versa. However, please make sure I didn’t mix up any sources with each other or any style layers with each other.

PR #2317, opened 2020-01-22
Require map SDK v5.5.0 in v0.37.x (redux)

1ec5:
This is a reprise of #2301 that disallows versions of the map SDK other than v5.5.0. In other words, if map SDK v5.5.1 comes out with a critical bug fix, projects that depend on the navigation SDK will not receive v5.5.1 unless we also release a patch release of the navigation SDK.

This change targets the v0.37.x release series.

/ref #2299
/cc @mapbox/navigation-ios @mapbox/maps-ios @mapbox/mobile-telemetry @mapbox/vision-ios


MapboxNavigation.podspec
```diff
@@ -44,7 +44,7 @@ Pod::Spec.new do |s|
   s.module_name = "MapboxNavigation"
 
   s.dependency "MapboxCoreNavigation", "#{s.version.to_s}"
-  s.dependency "Mapbox-iOS-SDK", "~> 5.5.0"
+  s.dependency "Mapbox-iOS-SDK", "= 5.5.0"
```
1ec5: CocoaPods wasn’t affected by the original issue, but I’m pinning the podspec too to ensure consistent dependencies regardless of the dependency manager used.

MapboxNavigation-Documentation.podspec
```diff
@@ -47,7 +47,7 @@ Pod::Spec.new do |s|
 
   s.dependency "MapboxDirections.swift", "~> 0.30.0"
   s.dependency "MapboxGeocoder.swift", "~> 0.10.0"
-  s.dependency "Mapbox-iOS-SDK", "~> 5.5.0"
+  s.dependency "Mapbox-iOS-SDK", "= 5.5.0"
   s.dependency "MapboxMobileEvents", "~> 0.9.5"
```
1ec5: Should we also pin MapboxMobileEvents to v0.9.5 specifically and avoid patch releases? Unlike with the map SDK, this branch is still on a pre-v1.0 release, so it wouldn’t float to v0.10._x_.

/cc @rclee @alfwatt

MapboxCoreNavigationTests/CocoaPodsTest/PodInstall/Podfile.lock
```diff
@@ -1,16 +1,16 @@
 PODS:
   - Mapbox-iOS-SDK (5.5.0)
-  - MapboxCoreNavigation (0.37.0):
+  - MapboxCoreNavigation (0.37.1):
```
chezzdev: Not sure why the previous version should be used here.

1ec5: The CocoaPods installation test lags official releases by design. We haven’t released v0.37.2 yet, so the test can’t verify that running `pod update` would install v0.37.2. However, it would detect if a newer dependency would be pulled in. This is why the test has failed quite often during the v1.0 development process, as breaking changes were introduced between prereleases of MapboxDirections.

1ec5:
Cherry-picked into the release-v0.38 branch as b66616136f2a5159eae95c53cc9b17f3243b9c2b.

PR #2319, opened 2020-01-23
MapboxNavigationNative v9.0.4

### LINKED ISSUE #2125: No RouteProgress updates when shape format is geoJSON
**Mapbox Navigation SDK version:** 0.32.0 (CocoaPods)
**Xcode Version:** 10.2.1
**Tested on various iPads (simulator and real) with iOS 12.2**

Hi everyone,
First of all, thanks for the solid framework you are providing, it's really good to work with!
While working with the results from the Mapbox Navigation API within the SDK, I discovered a bug:

If the `shapeFormat` parameter of the `NavigationRouteOptions` is set to `RouteShapeFormat.geoJSON` the `routeProgress` object is not updated during simulation. In the UI this leads to not updating route statistics, missing voice commands, and not updating direction instructions. And as shown in the example below, all properties of the `routeProgress` object are not updated.
Through debugging I could only narrow it down to a not correct calculated `status` object in `RouteController.locationManager(_:didUpdateLocations:)`. All called update functions afterwards can't read any progress from the retrieved status object.

To reproduce the error just execute this modified version of the basic example:
```swift
import UIKit
import MapboxCoreNavigation
import MapboxNavigation
import MapboxDirections

    override func viewDidLoad() {
        super.viewDidLoad()
        
        let origin = CLLocationCoordinate2DMake(37.77440680146262, -122.43539772352648)
        let destination = CLLocationCoordinate2DMake(37.76556957793795, -122.42409811526268)
        let options = NavigationRouteOptions(coordinates: [origin, destination])
        
        
        /* ADDED TO SIMULATE THE ERROR */
        options.shapeFormat = RouteShapeFormat.geoJSON
        NotificationCenter.default.addObserver(self, selector: #selector(routeProgressUpdate(_:)), name: .routeControllerProgressDidChange, object: nil)
        
        Directions.shared.calculate(options) { (waypoints, routes, error) in
            guard let route = routes?.first, error == nil else {
                print(error!.localizedDescription)
                return
            }
            options.profileIdentifier = MBDirectionsProfileIdentifier.automobile
            
            let navigationService = MapboxNavigationService(route: route, simulating: .always)
            let navigationViewController = NavigationViewController(for: route, options: navigationOptions)
            
            self.present(navigationViewController, animated: true, completion: nil)
        }
    }
    
    /* ADDED TO SIMULATE THE ERROR */
    @objc func routeProgressUpdate(_ notification: NSNotification) {
        let progress = notification.userInfo![RouteControllerNotificationUserInfoKey.routeProgressKey] as! RouteProgress
        print(progress.currentLegProgress.remainingSteps.count)
    }


}
```

1ec5:
Upgraded the MapboxNavigationNative dependency from v6.2.1 to v9.0.4. The changes primarily affect location tracking and off-route calculation during turn-by-turn navigation. These are the areas that will need extensive testing before we can merge and release. (Additionally, MapboxNavigationNative’s internal Valhalla dependency, used for offline route calculation, has been upgraded from roughly v3.0.4 to roughly v3.0.9.)

* [x] Update the dependency
   * [x] Specify the new MapboxNavigationNative version requirement in the Cartfile and podspec
   * [x] Run `carthage update` to update the copy of MapboxNavigationNative used for development in the mapbox-navigation-ios repository
   * [x] Run `cd MapboxCoreNavigationTests/CocoaPodsTest/PodInstall/ && pod update` to update the CocoaPods installation tests
* [x] Build the MapboxNavigation scheme and fix any compiler errors that arise
* [ ] Upgrade again to get the fix for #2336
   * [ ] Round-trip `null` `voiceLocale`: mapbox/mapbox-directions-swift#415
* [ ] Disable filtering
* [ ] Test the upgraded project:
   * [x] Run the MapboxCoreNavigation scheme’s automated tests
   * [ ] Run the Example or Example-CarPlay scheme in an iPhone simulator, enabling the application’s Enable Simulation option
   * [ ] Run the Example or Example-CarPlay scheme on an iPhone device without enabling the Enable Simulation option
   * [ ] Run the CarPlay example application on a CarPlay device
   * [ ] Test-drive the CarPlay example application in a car that has a CarPlay-capable screen

Fixes #2125.

/cc @mapbox/navigation-ios @mapbox/navnative @mapbox/driver-app


1ec5:
The CocoaPods podspec fails to lint because of compiler warnings about MapboxNavigationNative:

```
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBHistoryPlayer.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBGetRouteCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBSetRouteCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MapboxNavigationNativeHistoryPlayer.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBGetTraceAttributesCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBUpdateSensorDataCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBConfigureRouterCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBLocateCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBIterationResult.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBGetStatusCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBUpdateLocationCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBChangeRouteLegCallback.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxNavigationNative' does not include header 'MBPushHistoryCallback.h'

[!] MapboxCoreNavigation did not pass validation, due to 13 warnings (but you can use `--allow-warnings` to ignore them).
You can use the `--no-clean` option to inspect any issue.
```

The CocoaPods build bot isn’t required to pass; it basically tells us whether we need to pass `--allow-warnings` when releasing.

1ec5:
My first impression with this branch in the iOS simulator is that 4× route simulation is noticeably jerky compared to how it used to be. MapboxNavigationNative is probably doing more aggressive filtering that treats these unlikely fast location updates as bogus.

mskurydin:
Could you share the recorded history? We will be able to check all the algorithmic stuff.

1ec5:
After upgrading from v9.0.3 to v9.0.4, one of the tests is failing because twice the expected number of location updates is coming into the navigation service delegate:

> Test Case '-[MapboxCoreNavigationTests.NavigationServiceTests testUnimplementedLogging]' started.
2020-02-14 00:38:17.646596-0800 Example[83813:124462] [delegation.EmptyNavigationServiceDelegate] Unimplemented delegate method in EmptyNavigationServiceDelegate: NavigationServiceDelegate.navigationService(_:didUpdate:with:rawLocation:). This message will only be logged once.
/Users/distiller/project/MapboxCoreNavigationTests/NavigationServiceTests.swift:998: error: -[MapboxCoreNavigationTests.NavigationServiceTests testUnimplementedLogging] : XCTAssertEqual failed: ("8") is not equal to ("4") - Expected logs to be populated and expected number of messages sent
Test Case '-[MapboxCoreNavigationTests.NavigationServiceTests testUnimplementedLogging]' failed (1.226 seconds).

mskurydin:
@1ec5 could you help understand the problem deeper? The only noticeable change for 9.0.4 was speed calculation while doing the look-ahead (during guided driving).

IodaMikeMapbox:
> After upgrading from v9.0.3 to v9.0.4, one of the tests is failing because twice the expected number of location updates is coming into the navigation service delegate:

I investigated the problem, it's related to the our route replay, we simulate the passage of a track with speed multiplier 4 which is roughly similar to have ride at speed 432 km/h. Looks like processing of that speed wasn't supported by v6 which led to not updating progress and we didn't get events of arriving at destination. If we decrease the speedMultiplier to 1 with v6 the sdk will generate three more events which is willArrive, didArrive and shouldPreventReroutesWhenArrivingAt, the total number of events is 7.
As for 8 events. There is location filter in version 9 which, if the speedMulitplier is set to 4, led to a rerouting while traveling on a virtual route. Also filtering solved the problem of not arriving at destination. In total, since the rerouting is one more event, the number of events in v9 was 7+1=8.
For now I fixed the test by decreasing the speedMultiplier to 2 in https://github.com/mapbox/mapbox-navigation-ios/pull/2319/commits/ee5fcb3ccf955888750218e7d39004c183bf0d99, but we should rethink our approach with speed multiplier and sync it with android, the android sdk works correctly even with speed multiplier equal to 8.

1ec5:
The Xcode_10.2_iOS_12.2_CP_update build failure is being tracked in #2394.

Cartfile
```diff
@@ -1,7 +1,7 @@
 binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" ~> 2.2
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.6
-binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.2.1
-github "mapbox/mapbox-directions-swift" ~> 0.31
+binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 9.0.3
+github "mapbox/mapbox-directions-swift" == 1.0.0-alpha.1
```
1ec5, line 4: This should’ve remained v0.31. Looks like a rebase incorrectly changed it back to v1.0.0-alpha.1. Fixed in c3dda0043f8a933668b44cfca997daa2e0e37c2d.

Cartfile
```diff
@@ -1,7 +1,7 @@
 binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" ~> 2.2
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.6
-binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 6.2.1
-github "mapbox/mapbox-directions-swift" ~> 0.31
+binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 9.0.3
```
1ec5, line 3: Pinned to v9.0.4 specifically in 5cae87dc409568826a1dc20b1734f66fccc39a41 and aa149ed5cb3fc00f0b2888c44f9ed32067535d59, because v9.2.1 is available and not backwards-compatible. Compatibility with v9.2.1 would require a bit of work, but we’re doing that anyways as part of upgrading to v13._x_.

/cc @mapbox/navnative

PR #2321, opened 2020-01-23
v0.37.2

1ec5:
Updated the Xcode project and CocoaPods podspecs for v0.37.2.

/cc @mapbox/navigation-ios


avi-c:
Is the failing test a concern? Seems like it is directly related to the PR.

1ec5:
It’s expected that this CocoaPods installation test would fail on a release PR:

```
Resolving dependencies of 
 -> MapboxNavigation (0.37.2)
    - ERROR | [iOS] unknown: Encountered an unknown error (CocoaPods could not find compatible versions for pod "MapboxCoreNavigation":
  In Podfile:
    MapboxNavigation (from `/Users/distiller/project`) was resolved to 0.37.2, which depends on
      MapboxCoreNavigation (= 0.37.2)

None of your spec sources contain a spec satisfying the dependency: `MapboxCoreNavigation (= 0.37.2)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.
```

The reason is that, locally, v0.37.2 does exist, but it hasn’t been pushed to CocoaPods trunk yet.

PR #2329, opened 2020-02-01
Pass API hostname through to speech synthesizer

1ec5:
Fixed an issue where spoken instructions always fell back to VoiceOver when the `MGLMapboxAPIBaseURL` key was set in the Info.plist file and the voice controller was embedded in a NavigationViewController.

/cc @mapbox/navigation-ios


1ec5:
> consider using: https://github.com/mapbox/mobile-telemetry-events-ios/blob/feabf166141e16f0550687207afc931aae5fec86/MapboxMobileEvents/Categories/NSUserDefaults%2BMMEConfiguration.m#L188

What does that code do? Does it automatically select a hostname based on the access token?

For context, this PR ensures that MapboxSpeech uses the same host for Voice API calls as MapboxDirections uses for Directions API calls. The two APIs are tightly coupled in that the Voice API is only guaranteed to accept input that the Directions API outputs. Normally it doesn’t matter much because the Voice API’s input is quite straightforward, but it’s still important to ensure a consistent host for all the public APIs the SDK hits.

PR #2337, opened 2020-02-13
Swap to centralized analytics script

danswick:
I hope I'm not giving you too much whiplash, but we had some discussion internally and decided it would be best to link to a centralized analytics script instead of inlining analytics initialization in each and every repo.

### DIFF
--- a/docs/theme/templates/doc.mustache
+++ b/docs/theme/templates/doc.mustache
@@ -17,12 +17,7 @@
         padding-left: 5px
       }
     </style>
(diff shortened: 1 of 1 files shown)


danswick:
Ah, looks like I don't have adequate permissions to merge here. Would one of you be able to merge, @1ec5 or @captainbarbosa ?

captainbarbosa:
Hmm it appears I'm not authorized either! @1ec5 😁

PR #2338, opened 2020-02-14
Fix crash when using dynamic UIColors in guidance card styling

avi-c:
Problem:
If you provide dynamic UIColor as a color for ExitView or GenericShieldView styling, you will get a crash due to the way the hash value is calculated for storing the rendered image in the image crash.

Fix is to create our own UIColor instances based on the CGColor values of the resolved underlying UIColor.

### DIFF
--- a/MapboxNavigation/ExitView.swift
+++ b/MapboxNavigation/ExitView.swift
@@ -148,14 +148,12 @@ public class ExitView: StylableView {
         let proxy = ExitView.appearance()
         var backgroundColor = proxy.backgroundColor
         var foregroundColor = proxy.foregroundColor
-        let resolvedColorSelector = Selector(("resolvedColorWithTraitCollection:" as NSString) as String)
-        if #available(iOS 13.0, *),
-            let bgColor = backgroundColor, bgColor.responds(to: resolvedColorSelector),
-            let fgColor = foregroundColor, fgColor.responds(to: resolvedColorSelector),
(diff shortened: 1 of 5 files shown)


MapboxNavigation/ExitView.swift
```diff
@@ -148,14 +148,16 @@ public class ExitView: StylableView {
...
-            let currentTraitCollection = UIApplication.shared.keyWindow?.traitCollection {
-            if let backgroundColorInstance = bgColor.perform(resolvedColorSelector, with: currentTraitCollection), let foregroundColorInstance = fgColor.perform(resolvedColorSelector, with: currentTraitCollection) {
-                backgroundColor = backgroundColorInstance.takeRetainedValue() as? UIColor
-                foregroundColor = foregroundColorInstance.takeRetainedValue() as? UIColor
+        let performAsCurrentSelector = Selector(("performAsCurrentTraitCollection:" as NSString) as String)
+        if #available(iOS 13.0, *) {
+            if let currentTraitCollection = UIApplication.shared.keyWindow?.traitCollection, currentTraitCollection.responds(to: performAsCurrentSelector), let backgroundCGColor = backgroundColor?.cgColor, let foregroundCGColor = foregroundColor?.cgColor {
+
+                let colorCopyingClosure = {
+                    backgroundColor = UIColor(cgColor: backgroundCGColor)
```
1ec5: Would calling `resolvedColor(with:)` have the same effect?

PR #2343, opened 2020-02-28
Expose secondary highlight colors used in guidance cards

avi-c:
Nav SDK users are unable to override the secondary highlight colors for the ManeuverView or NextBanner maneuver view. Expose these values and use a transparent version of the primary color as the default value.


avi-c:
For reference, the secondary highlight colors are the dimmed out ones shown as part of the maneuvers illustrated in this image.

Secondary Highlight colors

Issue #2245, opened 2019-09-30
[Regression] Next instruction banners show during entire step, not just at end


JThramer:
We seemed to have regressed when [introducing Guidance Cards](https://github.com/mapbox/mapbox-navigation-ios/pull/1514) by blowing away the logic which [checks that we're approaching a maneuver, in order to show the next view](https://github.com/mapbox/mapbox-navigation-ios/commit/6dc222c618602561504378a2be790d2dc2d840d5#diff-65cceb0e2fd826fafb5551937d1aa83bL91), resulting in us showing the 'next' view during the entire duration of the step.

As a result of this, `RouteControllerLinkedInstructionBufferMultiplier` and `RouteControllerHighAlertInterval` are now orphaned constants.

/cc @mapbox/navigation-ios

PR #2345, opened 2020-03-03
#2245: next instructions display fix

Udumft:
Fixed displaying next instructions banner at the end of current step. Code based on previous implementation. 
NavigationBannerView may be called to respect showing next  banner based on route progress (by calling as `NavigationServiceDelegate`, or called to display such info unconditionally (if called by `update(for:)` directly)

### DIFF
--- a/MapboxNavigation/NextBannerView.swift
+++ b/MapboxNavigation/NextBannerView.swift
@@ -90,9 +90,31 @@ open class NextBannerView: UIView, NavigationComponent {
     }
     
     public func navigationService(_ service: NavigationService, didPassVisualInstructionPoint instruction: VisualInstructionBanner, routeProgress: RouteProgress) {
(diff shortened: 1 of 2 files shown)


Udumft:
Right now, guidance cards use `NextBannerView` and show second maneuver all the time.
'Usual' UI:
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-10 at 15 05 05
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-10 at 15 05 41

Same route w/ Guidance Cards:
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-10 at 15 03 04

I believe it is not too complicated to adopt same behavior for Guidance Cards too, if we wish so.

Udumft:
Is there anything I should add to this PR?

PR #2348, opened 2020-03-09
Speech controller refactor

### LINKED ISSUE #2268: Replace voice controller inheritance with SpeechController registry
MapboxVoiceController and RouteVoiceController have an awkward relationship. RouteVoiceController handles AVSpeechSynthesizer, while MapboxVoiceController handles the Voice API and falls back to AVSpeechSynthesizer by calling superclass method implementations.

We’re encountering a lot of awkwardness in #2230 trying to formalize error states. If a developer wants to integrate with a third-party text-to-speech engine or supply their own audio assets, they have to subclass one of these classes, and error handling gets even weirder.

This functionality should rely on composition rather than inheritance. A single SpeechController class should take the place of these voice controllers. It should own a registry of speech synthesizers. A SpeechSynthesizing protocol would specify methods like `prefetch`, `play`, `duck`, etc.

Supersedes #2131.

/cc @mapbox/navigation-ios

Udumft:
Resolves #2268.

`RouteVoiceController` together with it's subclass `MapboxVoiceController` had 3 goals: tracking navigation events, vocalizing `SpokenInstruction`s and handling fallback mechanism for speech synthesizers.
In this PR these tasks are separated by introducing `SpeechSynthesizing` protocol. 
`RouteVoiceController` now tracks navigation events and triggers configured `speechSynthesizer` to vocalize the instructions.
`SpeechSynthesizersController` handles an arbitrary array of `SpeechSynthesizing` implementations to handle the fallback feature.
`SystemSpeechSynthesizer` and `MapboxSpeechSynthesizer` provide complete and expandable implementations for using iOS built-in speech synthesizer and MapboxSpeech framework respectively.


MapboxNavigation/MapboxSpeechSynthesizer.swift
```diff
@@ -0,0 +1,252 @@
...
+                let urlError = underlyingError as? URLError, urlError.code == .cancelled {
+                self.completion?(error)
+                return
+            } else if let error = error {
+                self.delegate?.voiceController(self, spokenInstructionsDidFailWith: SpeechError.apiError(instruction: modifiedInstruction,
+                                                                                                    options: options,
+                                                                                                    underlying: error))
+                self.completion?(SpeechError.apiError(instruction: modifiedInstruction,
+                                                      options: options,
+                                                      underlying: error))
```
Udumft: Error propagation is still a TODO.

I feel the need to have explicit `completion` for `speak` method to allow it to by async, but at the same time we want to have a `delegate` to track various events. Also, some errors may be blocking for Speech Synthesizer to actually speak, others may be not.
What is the best way to handle errors in this situation? We can:
1. report any errors occur with via `delegate`, and `completion` will receive error only if it was blocking (or nil otherwise)
2. report any errors occur with via `delegate`, and `completion` will just have boolean `success`, and thus the last error reported was a blocker.
3. get rid of `completion` and make it a part of `delegate` protocol (with one of the 2 above points implemented)
4. something else?

MapboxNavigation/SpeechSynthesizerController.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    ///
+    func changedIncomingSpokenInstructions(_ instructions: [SpokenInstruction])
+    ///
+    func speak(_ instruction: SpokenInstruction, completion: SpeechSynthesizerCompletion?)
+    
+    ///
+    func stopSpeaking()
+    ///
+    func interruptSpeaking() // ??
```
Udumft: Not sure if we need to, but maybe it could be useful to have different ways to shut down ongoing phrase? For example: one method to mark that phrase should be stopped gracefully whenever appropriate, and another to abort it immediately.

MapboxNavigation/SpeechSynthesizerController.swift
```diff
@@ -0,0 +1,85 @@
...
+    var volume: Float { get set }
+    ///
+    var isSpeaking: Bool { get }
+    ///
+    var locale: Locale { get set }
+    
+    ///
+    func changedIncomingSpokenInstructions(_ instructions: [SpokenInstruction])
+    ///
+    func speak(_ instruction: SpokenInstruction, completion: SpeechSynthesizerCompletion?)
```
Udumft: Following what was mentioned in original ticket, I don't feel like such protocol should have API to control `ducking` because:
1. It is responsibility of a `SpeechSynthesizer` to know and control when it needs to `duck`, if it needs at all
2. `Ducking` feature is performed via system methods and does not depend on particular synthesizer implementation.

MapboxNavigation/SystemSpeechSynthesizer.swift
```diff
@@ -0,0 +1,170 @@
...
+    public var delegate: SpeechSynthesizerDelegate?
+    public var muted: Bool = false // ???
+    public var volume: Float {
+        get {
+            return NavigationSettings.shared.voiceVolume
+        }
+        set {
+            // ?!?!?!
+        }
+    }
```
Udumft: Handling volume and muting with respect to `NavigationSettings.shared.voiceVolume` is not implemented yet

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -272,7 +191,7 @@ public protocol VoiceControllerDelegate: class, UnimplementedLogging {
      - parameter synthesizer: the Speech engine that was used as the fallback.
      - parameter error: An error explaining the failure and its cause.
      */
-    func voiceController(_ voiceController: RouteVoiceController, didFallBackTo synthesizer: AVSpeechSynthesizer, error: SpeechError)
```
Udumft: This method feels like a dog-nail that we had to support `MapboxVoiceController`'s feature of a 'backup' synthesizer.

MapboxNavigation/MapboxSpeechSynthesizerController.swift
```diff
@@ -0,0 +1,109 @@
+
+import AVFoundation
+import MapboxDirections
+import MapboxCoreNavigation
+import MapboxSpeech
+
+///
+open class MapboxSpeechSynthesizerController: NSObject, SpeechSynthesizerController {
```
1ec5: I think the separation of concerns would be clearer if we rename this class to SpeechSynthesizerController and rename the SpeechSynthesizerController protocol to SpeechSynthesizing. This class doesn’t need to conform to SpeechSynthesizing.

Example/CustomViewController.swift
```diff
@@ -39,7 +39,7 @@ class CustomViewController: UIViewController, MGLMapViewDelegate {
         
         let locationManager = simulateLocation ? SimulatedLocationManager(route: userRoute!) : NavigationLocationManager()
         navigationService = MapboxNavigationService(route: userRoute!, locationSource: locationManager, simulating: simulateLocation ? .always : .onPoorGPS)
-        voiceController = MapboxVoiceController(navigationService: navigationService)
+//        voiceController = MapboxVoiceController(navigationService: navigationService)
```
1ec5: I think we could reimplement this customization hook by exposing the navigation service’s (read-only) speech controller so that the application can append a SpeechSynthesizing instance to the speech controller’s `speechSynthesizers` array.

MapboxNavigation/RouteVoiceController.swift
1ec5: That’s true to some extent, though I think the other consideration was that we wanted to send a telemetry event whenever speech got interrupted. There’s other ways of doing that, though. Do you think it would still be useful for the application to know when all the available speech synthesizers fail?

MapboxNavigation/MapboxSpeechSynthesizer.swift
1ec5: If we have anything bubble up to the delegate, it should be after the speech controller has finished processing all the speech synthesizers. So if there’s an error, try falling back to another speech synthesizer, and only notify the delegate if all the speech synthesizers fail.

MapboxNavigation/MapboxSpeechSynthesizerController.swift
Udumft: I was thinking about this class as some kind of a wrapper or a proxy which manages multiple speech engines. At the same time it is possible to use a single speech engine directly in `RouteVoiceController` if user want's to, so `MapboxSpeechSynthesizerController` should adopt `SpeechSynthesizerController` too.

In this case, `MapboxSpeechSynthesizerController` does not have any crucial role in the mechanism and can be safely removed, or used with any custom set of speech synthesizers.

Yes, looks like naming could be better to display the idea :)

MapboxNavigation/RouteVoiceController.swift
Udumft: For sure, knowing that speech synths had fail is important, but I thought that `SpeechSynthesizerController` describes a single speech Synth implementation, and thus it "doesn't know" about other implementations (so there is no such term as "fallback" in it's context). At this point, responsibility to manage multiple synths and notify when they all failed should be somewhere else. In current situation that is done by `MapboxSpeechSynthesizerController`, which will try to call all synths and if all of them fail - will report the latest error.

Udumft:
I feel the need to clarify my vision on the refactoring being performed to make sure we are on the same page.

As I understood, original problem was that `RouteVoiceController` (together with it's subclass `MapboxVoiceController`) did 3 tasks at the same time: monitoring the actual route progress events to check when voice instructions should be pronounced, handled the speech synth engine(s), and took care about fallback mechanism when multiple engines were used to work together.


> Next, all entity names are what they are at the moment in current draft, just for convenience

The solution is to separate these tasks to make such modules replaceable. What I am trying to achieve:
* `RouteVoiceController` remains responsible for monitoring route progress to pass voice instructions to the speech engine
* Speech engine is hidden behind a protocol (`SpeechSynthesizerController` at the moment) which allows simple actions to actually "speak" an instruction. Each implementation is responsible only for it's own integrity and is independent
* To maintain functionality of "fallback" synthesizers, `MapboxSpeechSynthesizerController` (current name) is introduced, which mimics `SpeechSynthesizerController` functionality, but in fact just aggregates real implementations and thus performs mechanism of a "backup" speech synthesizers.

In result, each element can be replaced by User's implementation. User may implement his own `RouteVoiceController` to monitor events in the fashion he wants and still use any `SpeechSynthesizerController` to vocalize instructions, User may implement his own `SpeechSynthesizerController` and pass it to `MapboxSpeechSynthesizerController` to be a part of a fallback sequence. User may not use `MapboxSpeechSynthesizerController` at all to have only 1 speech engine directly in `RouteVoiceController`.

With that said, when we talk about `SpeechSynthesizerController` delegate, it's methods describe state, errors, and any other events related to this particular implementation. In such delegate, events like "fallback" have no sense because one `SpeechSynthesizerController` is not related to other implementations.

Does it match with you vision? :)

Udumft:
This PR changes will also conflict with #2350. Will do a conflict resolution once it is merged in.

MapboxNavigation/SpeechSynthesizersController.swift
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+    
+    public  let speechSynthesizers: [SpeechSynthesizing]
+    
+    private var currentLegProgress: RouteLegProgress?
+    
+    // MARK: - Lifecycle
+    
+    public init(_ speechSynthesizers: [SpeechSynthesizing]? = nil, accessToken: String? = nil) {
```
1ec5, line 40: The access token (and host) is a detail specific to MapboxSpeechSynthesizer. On master, MapboxVoiceController leaves the default SpeechController’s access token blank, but NavigationViewController ends up creating a different default SpeechSynthesizer with a matching access token, so it isn’t really even necessary for MapboxVoiceController to provide a default SpeechController.

Here, we could make `speechSynthesizers` required and remove the `accessToken` argument; then callers (still within the SDK) would need to create the appropriate MapboxSpeechSynthesizer and SystemSpeechSynthesizer with the context they have on hand.

MapboxNavigation/SpeechSynthesizersController.swift
```diff
@@ -0,0 +1,116 @@
...
+    
+    // MARK: - Public Methods
+    
+    public func changedIncomingSpokenInstructions(_ instructions: [SpokenInstruction]) {
+        speechSynthesizers.forEach { $0.changedIncomingSpokenInstructions(instructions) }
+    }
+    
+    public func speak(_ instruction: SpokenInstruction, during legProgress: RouteLegProgress) {        
+        guard let synthesizer = speechSynthesizers.first else {
+            assert(false, "SpeechSynthesizersController has 0 speechSynthesizers")
```
1ec5, line 57:
```suggestion
            assertionFailure("SpeechSynthesizersController has 0 speechSynthesizers")
```

MapboxNavigation/SpeechSynthesizersController.swift
```diff
@@ -0,0 +1,116 @@
...
+            delegate?.speechSynthesizer(self,
+                                        didSpeak: instruction,
+                                        with: nil)
```
1ec5, line 60: This case is an error on the part of the developer. I think it’s possible for the application to be built with the `-assert-config` flag that disables assertions, in which case this developer error becomes a silent failure. Depending on the severity or recoverability of the issue, it would be best to either issue a `preconditionFailure(_:file:line:)` or throw an error.

MapboxNavigation/SpeechSynthesizersController.swift
```diff
@@ -0,0 +1,116 @@
...
+
+extension SpeechSynthesizersController: SpeechSynthesizingDelegate {
+    
+    public func speechSynthesizer(_ speechSynthesizer: SpeechSynthesizing, didSpeak instruction: SpokenInstruction, with error: SpeechError?) {
+        if let error = error {
+            
+            
+            if let index = speechSynthesizers.firstIndex(where: { $0 === speechSynthesizer }),
+                let legProgress = currentLegProgress,
+                index + 1 < speechSynthesizers.count {
```
1ec5, line 85: Each time a speech synthesizer fails, we search through the list of speech synthesizers for the index of the one that failed, then turn to the one at the next index. This opens the door to certain classes of developer error. For example, if the developer accidentally specifies the same MySpeechSynthesizer twice in the `speechSynthesizers` list and this speech synthesizer fails for some reason, this method recurses infinitely.

This is pretty unfortunate, but I guess it’s a consequence of each speech synthesizer being asynchronous; all we’re trying to do here is to parallelize them. It does seem to do the right thing if the application makes the controller speak another utterance before it has finished going through the list for the first utterance.

CHANGELOG.md
```diff
@@ -18,9 +18,7 @@
...
-* The `MapboxVoiceController.speakWithDefaultSpeechSynthesizer(_:error:)` and `VoiceControllerDelegate.voiceController(_:spokenInstructionsDidFailWith:)` methods now accept a `SpeechError` instance instead of an `NSError` object. ([#2230](https://github.com/mapbox/mapbox-navigation-ios/pull/2230))
-* Added the `VoiceControllerDelegate.voiceController(_:didFallBackTo:becauseOf:)` method for detecting when the voice controller falls back to `AVSpeechSynthesizer`. ([#2230](https://github.com/mapbox/mapbox-navigation-ios/pull/2230))
+* Introduced `SpeechSynthesizing` for implementing custom Speech Synthesizing engines, as well as built-in implementations `SystemSpeechSynthesizer` and `MapboxSpeechSynthesizer` for system SpeechSynthesizer and Mapbox Speech engine respectively. This splits `RouteVoiceController` functionality into separate parts. `MapboxVoiceController` is replaced with alternative `SpeechSynthesizersController` which allows multiple Synthesizers with a fallback mechanism. ([#2268(https://github.com/mapbox/mapbox-navigation-ios/pull/2348)])
```
1ec5, line 21:
```suggestion
* Replaced `RouteVoiceController` and `MapboxVoiceController` with `SpeechSynthesizerController`. `SpeechSynthesizerController` coordinates multiple cascading speech synthesizers. By default, the controller still tries to speak instructions via the Mapbox Voice API (`MapboxSpeechSynthesizer`) before falling back to VoiceOver (`SystemSpeechSynthesizer`), but you can also provide your own speech synthesizer that conforms to the `SpeechSynthesizing` protocol. ([#2268](https://github.com/mapbox/mapbox-navigation-ios/pull/2348))
```

I realize RouteVoiceController technically still exists, but it’s no longer as interesting to the application developer as something that can be swapped out; it’s more likely that the developer would want to swap out individual speech synthesizers.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -207,8 +207,10 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
         
         self.navigationService = options?.navigationService ?? MapboxNavigationService(route: route)
         self.navigationService.delegate = self
-        self.voiceController = options?.voiceController ?? MapboxVoiceController(navigationService: navigationService, speechClient: SpeechSynthesizer(accessToken: navigationService?.directions.accessToken, host: navigationService?.directions.apiEndpoint.host))
+//        self.voiceController = options?.voiceController ?? MapboxVoiceController(navigationService: navigationService, speechClient: SpeechSynthesizer(accessToken: navigationService?.directions.accessToken, host: navigationService?.directions.apiEndpoint.host))
```
1ec5, line 210: This commented-out functionality aligned the speech synthesizer’s credentials to the credentials used with the Directions API. It looks like we should push that logic down into `RouteVoiceController(navigationService:)` now.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -207,8 +207,10 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
 
+        self.voiceController = options?.voiceController ?? RouteVoiceController(navigationService: navigationService)
```
1ec5, line 212: `NavigationOptions.voiceController` was intended as a customization hook to directly swap out speech synthesizers. Now RouteVoiceController is an extra layer of indirection: the developer would create a MySpeechSynthesizer to pass into the RouteVoiceController initializer to pass into the NavigationOptions initializer to pass into the NavigationViewController initializer. But RouteVoiceController itself doesn’t seem to do anything that the developer would need to override, so how about we replace `NavigationOptions.voiceController` with `NavigationOptions.speechSynthesizers` and plumb that value through?

MapboxNavigation/MapboxSpeechSynthesizer.swift
```diff
@@ -0,0 +1,253 @@
...
+            audioPlayer?.play()
+        case .failure(let error):
+            safeUnduckAudio(instruction: instruction)
+            delegate?.speechSynthesizer(self,
+                                        didSpeak: instruction,
+                                        with: error)
+        }
+    }
+    
+    private func downloadAndCacheSpokenInstruction(instruction: SpokenInstruction) {
```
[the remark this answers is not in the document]
Udumft: It is possible, but IMHO current version is clearer and self-documenting

MapboxNavigation/NavigationViewController.swift
Udumft: If we replace `NavigationOptions.voiceController` with `NavigationOptions.speechSynthesizer` we'll loose ability to provide custom subclasses of `RouteVoiceController`. To preserve it and to enable easy speech synthesizer customization we'll have to invent something else, which will also correctly handle custom speech synth + route voice controller simultaneously. 
I couldn't come up with a brilliant idea right away, but maybe we can just make `RouteVoiceController.speechSynthesizer` property writable so user can easily customize speech synthesizing during initialization? This case might require some extensive testing for speech synthesizers do correctly swap mid-navigation.

MapboxNavigation/SpeechSynthesizersController.swift
Udumft: I've added a check to `speechSynthesizers` setter to exclude such case.

Example/CustomViewController.swift
```diff
@@ -18,7 +18,6 @@ class CustomViewController: UIViewController, MGLMapViewDelegate {
     var userRouteOptions: RouteOptions?
 
     // Start voice instructions
```
1ec5, line 20: This comment doesn’t make sense anymore; let’s remove it.

MapboxNavigation/MapboxSpeechSynthesizer.swift
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     Fetches and plays an instruction.
+     */
+    private func fetchAndSpeak(instruction: SpokenInstruction){
+        audioTask?.cancel()
+        
+        let modifiedInstruction = delegate?.speechSynthesizer(self, willSpeak: instruction) ?? instruction
+        let ssmlText = modifiedInstruction.ssmlText
+        let options = SpeechOptions(ssml: ssmlText)
+        options.locale = locale
```
1ec5, line 96: By default, this locale should be the `speechLocale` from the route response, if present, to ensure that we use the voice that the text was designed for. Otherwise, the fallback to `autoupdatingCurrent` is fine, but that shouldn’t be the default. I think this context can be plumbed through the `speak(_:during:)` call.

MapboxNavigation/MapboxSpeechSynthesizer.swift
```diff
@@ -0,0 +1,242 @@
...
+                                        didSpeak: instruction,
+                                        with: error)
+        }
+    }
+    
+    private func downloadAndCacheSpokenInstruction(instruction: SpokenInstruction) {
+        let modifiedInstruction = delegate?.speechSynthesizer(self, willSpeak: instruction) ?? instruction
+        let ssmlText = modifiedInstruction.ssmlText
+        let options = SpeechOptions(ssml: ssmlText)
+        options.locale = locale
```
1ec5, line 165: This also needs to set the `locale` to the route’s `speechLocale`. I think this context can be plumbed through the `didPassSpokenInstructionPoint(notification:)` call.

MapboxNavigation/Error.swift
```diff
@@ -69,5 +68,12 @@ public enum SpeechError: LocalizedError {
      - parameter instruction: The instruction that failed.
      - parameter progress: the offending `RouteProgress` that omits the expected `SpeechLocale`.
      */
-    case undefinedSpeechLocale(instruction: SpokenInstruction, progress: RouteProgress)
+    case undefinedSpeechLocale(instruction: SpokenInstruction, progress: RouteProgress) // to be removed?
```
1ec5, line 71: Given the points about MapboxSpeechSynthesizer needing `speechLocale`, I’m not sure we can remove this case just yet.

MapboxNavigation/SystemSpeechSynthesizer.swift
```diff
@@ -0,0 +1,172 @@
...
+        
+        let modifiedInstruction = delegate?.speechSynthesizer(self, willSpeak: instruction) ?? instruction
+        
+        if utterance?.voice == nil {
+            utterance = AVSpeechUtterance(attributedString: modifiedInstruction.attributedText(for: legProgress))
+        }
+        
+        // Only localized languages will have a proper fallback voice
+        if utterance?.voice == nil {
+            utterance?.voice = AVSpeechSynthesisVoice(language: Locale.preferredLocalLanguageCountryCode)
```
1ec5, line 78: I think we should also try to pass in the `speechLocale` here too. This code has used `Locale.preferredLocalLanguageCountryCode` since #187, even before we moved guidance instruction generation to the server side, so I think this is buggy behavior that we’ve overlooked for a long time.

Cartfile.resolved
```diff
@@ -1,6 +1,6 @@
 binary "https://mapbox-gl-native-ios.s3.amazonaws.com/public/internal/Mapbox-iOS-SDK.json" "5.9.1000"
 binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" "2.3.0"
-binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" "14.1.5"
+binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" "14.1.6"
```
MaximAlien, line 3: Not sure if update to nav. native `14.1.6` should be part of this changelist.

MapboxNavigation/SpeechSynthesizing.swift
```diff
@@ -0,0 +1,111 @@
...
+     Called when one spoken instruction interrupts another instruction currently being spoken.
+     
+     - parameter speechSynthesizer: The speech synthesizer that experienced the interruption.
+     - parameter interruptedInstruction: The spoken instruction currently in progress that has been interrupted.
+     - parameter interruptingInstruction: The spoken instruction that is interrupting the current instruction.
+     */
+    func speechSynthesizer(_ speechSynthesizer: SpeechSynthesizing, didInterrupt interruptedInstruction: SpokenInstruction, with interruptingInstruction: SpokenInstruction)
+    
+    /**
+     Called when a spoken is about to speak. Useful if it is necessary to give a custom instruction instead. Noting, changing the `distanceAlongStep` property on `SpokenInstruction` will have no impact on when the instruction will be said.
```
MaximAlien, line 72: Maybe better to add `spoken instruction`? Not a native English speaker myself though :)
```suggestion
     Called when a spoken instruction is about to be spoken. Useful if it is necessary to give a custom instruction instead. Note that changing the `distanceAlongStep` property on `SpokenInstruction` will have no impact on when the instruction will be said.
```

MapboxNavigation/SystemSpeechSynthesizer.swift
```diff
@@ -0,0 +1,172 @@
...
+        }
+        set {
+            // Do Nothing
+            // AVSpeechSynthesizer uses 'AVAudioSession.sharedInstance().outputVolume' by default
+        }
+    }
+    public var isSpeaking: Bool { return speechSynth.isSpeaking }
+    public var locale: Locale = Locale.autoupdatingCurrent
+    
+    private var speechSynth: AVSpeechSynthesizer
```
MaximAlien, line 34: Nit:
```suggestion
    private var speechSynthesizer: AVSpeechSynthesizer
```

MapboxNavigationTests/SpeechSynthesizersControllerTests.swift
```diff
@@ -0,0 +1,145 @@
...
+        var routeController: RouteVoiceController? = RouteVoiceController(navigationService: dummyService,
+                                                                          speechSynthesizer: MultiplexedSpeechSynthesizer(synthesizers))
+        
+        synthesizers = []
+        routeController = nil
+        
+        wait(for: [deinitExpectation], timeout: 3)
+    }
+    
+    func testSystemSpeechSynthesizer() {
```
MaximAlien, line 119: Not critical, but maybe it'd be good to also verify whether properties like `muted`, `isSpeaking` etc were set correctly as well in test?

CHANGELOG.md
```diff
@@ -88,9 +88,7 @@
...
-* The `MapboxVoiceController.speakWithDefaultSpeechSynthesizer(_:error:)` and `VoiceControllerDelegate.voiceController(_:spokenInstructionsDidFailWith:)` methods now accept a `SpeechError` instance instead of an `NSError` object. ([#2230](https://github.com/mapbox/mapbox-navigation-ios/pull/2230))
-* Added the `VoiceControllerDelegate.voiceController(_:didFallBackTo:becauseOf:)` method for detecting when the voice controller falls back to `AVSpeechSynthesizer`. ([#2230](https://github.com/mapbox/mapbox-navigation-ios/pull/2230))
+* Replaced `RouteVoiceController` and `MapboxVoiceController` with `SpeechSynthesizerController`. `SpeechSynthesizerController` coordinates multiple cascading speech synthesizers. By default, the controller still tries to speak instructions via the Mapbox Voice API (`MapboxSpeechSynthesizer`) before falling back to VoiceOver (`SystemSpeechSynthesizer`), but you can also provide your own speech synthesizer that conforms to the `SpeechSynthesizing` protocol. ([#2268](https://github.com/mapbox/mapbox-navigation-ios/pull/2348))
```
MaximAlien, line 91: Looks like PR references should be used in changelist:
```suggestion
* Replaced `RouteVoiceController` and `MapboxVoiceController` with `SpeechSynthesizerController`. `SpeechSynthesizerController` coordinates multiple cascading speech synthesizers. By default, the controller still tries to speak instructions via the Mapbox Voice API (`MapboxSpeechSynthesizer`) before falling back to VoiceOver (`SystemSpeechSynthesizer`), but you can also provide your own speech synthesizer that conforms to the `SpeechSynthesizing` protocol. ([#2348](https://github.com/mapbox/mapbox-navigation-ios/pull/2348))
```

MapboxNavigationTests/SpeechSynthesizersControllerTests.swift
Udumft: Added a few more tests for parameters, but unfortunately could not find a stable way to test all of them. For example `isSpeaking` does not have a dedicated event after it can be "guaranteed to talk". Or `volume` value for `SystemSpeechSynthesizer` relies on system values and is not controlled by `Multiplexed` synth.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -637,9 +642,19 @@
...
+		2B81EC27241A237E00145086 /* SpeechSynthesizersControllerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SpeechSynthesizersControllerTests.swift; sourceTree = "<group>"; };
+		2B91C9B02416357700E532A5 /* MapboxSpeechSynthesizer.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MapboxSpeechSynthesizer.swift; sourceTree = "<group>"; };
+		2B2B1EDD2424B95600FA18A6 /* ExampleUITests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = ExampleUITests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
+		2B2B1EDF2424B95600FA18A6 /* ExampleUITests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ExampleUITests.swift; sourceTree = "<group>"; };
+		2B2B1EE12424B95600FA18A6 /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
+		2B2B1EEA2424D0A700FA18A6 /* ViewController+UITests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "ViewController+UITests.swift"; sourceTree = "<group>"; };
```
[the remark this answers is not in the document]
Udumft: That should be a leftover code, which should disappear after rebasing

Cartfile.resolved
Udumft: that's `Cartfile.resolved`. This PR does not bump any dependencies. Anyway, after rebasing to the latest commit such change should vanish.

Udumft:
Corrections applied

MapboxNavigation/MapboxSpeechSynthesizer.swift
```diff
@@ -25,9 +25,12 @@ open class MapboxSpeechSynthesizer: NSObject, SpeechSynthesizing {
     public var isSpeaking: Bool {
         return audioPlayer?.isPlaying ?? false
     }
-    public var locale: Locale = Locale.autoupdatingCurrent
+    public var locale: Locale? = Locale.autoupdatingCurrent
```
1ec5, line 28: What does it mean if the developer sets `locale` to `nil`, given that the default value is already `autoupdatingCurrent`? If we should retain this edge case instead of falling back to `autoupdatingCurrent`, then let’s document it.

MapboxNavigation/SystemSpeechSynthesizer.swift
```diff
@@ -48,20 +48,28 @@ open class SystemSpeechSynthesizer: NSObject, SpeechSynthesizing {
...
+            self.delegate?.speechSynthesizer(self,
+                                             encounteredError: SpeechError.undefinedSpeechLocale(instruction: instruction))
+            return
+        }
+        
         var utterance: AVSpeechUtterance?
-        if Locale.preferredLocalLanguageCountryCode == "en-US" {
+        let localeCode = [locale.languageCode, locale.regionCode].compactMap{$0}.joined(separator: "-")
+        
+        if localeCode == "en-US" {
```
1ec5, line 72: It would be cleaner to check if `locale.languageCode == "en" && locale.regionCode == "US"`.

MapboxNavigation/SystemSpeechSynthesizer.swift
```diff
@@ -5,7 +5,7 @@ import MapboxCoreNavigation
 import MapboxSpeech
 
 /**
- `SpeechSynthesizing` implementation, using `AVSpeechSynthesizer`. Supports only english language.
+ `SpeechSynthesizing` implementation, using `AVSpeechSynthesizer`. Supports only English language.
```
1ec5, line 8: This implementation seems to support more than just English now. I don’t think it was ever the intention to have it support only English. Only the Alex voice is English-specific.

MapboxNavigation/MapboxSpeechSynthesizer.swift
Udumft: For vocalizing an instruction a property`locale` or a 'locale' argument to corresponding `speak` method can be used.
I was thinking about `SpeechSynthesizing` usage scenario as to provide user a flexible ways to provide `Locale`s: 
* it can be set by default for all instructions (property is set, method argument is nil)
* it can be overridden for a specific instruction (property is set, method argument set)
* it can be required to manually specify `Locale` for each instruction (property is nil, method argument is set)

The latter case requires property to by `optional`
For `MapboxSpeechSynthesizer` I decided to stick to the first variant, so I've added a default value.

PR #2350, opened 2020-03-12
Directions Interface Compatibility Update

JThramer:
Refactoring PR to enable support of https://github.com/mapbox/mapbox-directions-swift/pull/418 and https://github.com/mapbox/mapbox-directions-swift/pull/406. WIP.

To-Do: 
- [x] Fix remaining compilation issues in `MapboxCoreNavigation` (6 issues)
- [x] Fix remaining compilation issues in `MapboxNavigation` (~~6~~ 4 issues)
- [x] Fix remaining compilation issues in `Example`
- [x] Fix observable bugs in Example (tapping on alternate routes)
- [x] Fix remaining compilation issues in `MapboxNavigationTests` (30 issues)
- [x] Get Navigation tests to pass
- [x] Fix remaining compilation issues in `MapboxCoreNavigationTests`
- [x] Get Core Navigation tests to pass

/cc @mapbox/navigation-ios


MapboxNavigation/NavigationViewController.swift
```diff
@@ -201,15 +201,22 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
      - parameter route: The route to navigate along.
      - parameter options: The navigation options to use for the navigation session.
      */
-    required public init(for route: Route,
+    required public init(for response: RouteResponse,
                          options: NavigationOptions? = nil) {
         super.init(nibName: nil, bundle: nil)
         
+        let route = response.routes!.first! //TODO: Safe-ish, but is there a better way?
```
Udumft: Such logic prevents picking alternative routes for navigation.
`RouteResponse` is useful as calculation result, but for configuring `NavigationViewController` I think it might be  not correct. Such design suggests that user has to pass in the result of a calculation, optionally modifying it's state, which comes beyond original `RouteResponse role. 
For this initializer, I don't think we can get far away from
`init(for route: Route, routeOptions: DirectionsOptions, options: NavigationOptions? = nil`
unless we introduce some new "OptionsWrapper" struct to contain everything needed.

Otherwise we force user to construct his own `RouteResponse` with adjusted properties.

MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -80,7 +80,7 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
     
     required public init(along route: Route, directions: Directions = Directions.shared, dataSource source: RouterDataSource) {
         self.directions = directions
-        self._routeProgress = RouteProgress(route: route)
+        self._routeProgress = RouteProgress(route: route, options: ) //FIXME: Finish This
```
Udumft: Looks like `RouteOptions` should be supplied as additional argument.

MapboxNavigation/NavigationViewController.swift
JThramer: It's a known assumption that the first route in the `routes` model is the `activeRoute`. We allow for mutation of the `RouteResponse` object that allows the developer to customize which route they want to pick.

JThramer:
Blocked from merge by https://github.com/mapbox/mapbox-directions-swift/pull/418, need to fix cartfile

Example/ViewController.swift
```diff
@@ -52,17 +53,17 @@ class ViewController: UIViewController {
 
     // MARK: Directions Request Handlers
 
-    fileprivate lazy var defaultSuccess: RouteRequestSuccess = { [weak self] (routes) in
-        guard let current = routes.first else { return }
+    fileprivate lazy var defaultSuccess: RouteRequestSuccess = { [weak self] (response) in
+        guard let current = response.routes?.first, case let .route(options) = response.options else { return }
```
Udumft, line 57: `current` is unused. Is it just a copy-paste from `resonse.didSet`, and you just meant to check `!response.routes.isEmpty`?

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -34,6 +34,28 @@ open class NavigationRouteOptions: RouteOptions {
...
+     Initializes an equivalent `RouteOptions` object from a `NavigationMapOptions`
+     
+     - seealso: `NavigationMatchOptions`
+     */
+    public convenience init(navigationMatchOptions options: NavigationMatchOptions) {
+        self.init(waypoints: options.waypoints, profileIdentifier: options.profileIdentifier)
+        self.includesSteps = true
+        routeShapeResolution = .full
+        shapeFormat = .polyline6
+        attributeOptions = [.congestionLevel, .expectedTravelTime]
```
Udumft, line 47: redundant line

MapboxCoreNavigation/NavigationRouteOptions.swift
```diff
@@ -34,6 +34,28 @@ open class NavigationRouteOptions: RouteOptions {
         includesExitRoundaboutManeuver = true
     }
 
+    /**
+     Initializes an equivalent `RouteOptions` object from a `NavigationMapOptions`
+     
+     - seealso: `NavigationMatchOptions`
+     */
+    public convenience init(navigationMatchOptions options: NavigationMatchOptions) {
```
Udumft, line 42: I see that body of this initializer is like 80% similar to inits on lines 95 and 16. Maybe it's worth moving duplicated lines to some designated initializer or introduce a `commonInit` method?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -201,28 +206,37 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
      - parameter options: The navigation options to use for the navigation session.
      */
-    required public init(for route: Route,
-                         options: NavigationOptions? = nil) {
+    required public init(for route: Route, routeOptions: RouteOptions,
+                         navigationOptions: NavigationOptions? = nil) {
         super.init(nibName: nil, bundle: nil)
         
-        self.navigationService = options?.navigationService ?? MapboxNavigationService(route: route)
+//        let route = response.routes!.first!
```
Udumft, line 213: Please do not forget to remove commented code

1ec5:
> Some tests a failing due to the dependencies which are not yet released and thus cannot be configured correctly in the .podspec.

This is fine; we’ll have to put up with that failure until we push a new version of the MapboxDirections pod. We can make the CocoaPods checks not required or just use our administrator rights.

PR #2352, opened 2020-03-17
Minimize NSObject inheritance

### LINKED ISSUE #2294: Minimize NSObject inheritance
As a followup to #2230, we should minimize the number of classes that explicitly inherit from NSObject. NSObject inheritance hinders Codable adoption and potentially has other subtle side effects.

The following classes look like good candidates for becoming ordinary Swift classes, though we’ll need to make sure there isn’t any KVO lurking around. Not listed here are any classes that need to conform to NSObjectProtocol because of some Foundation protocol inheritance such as CLLocationManagerDelegate.

* [ ] RouteProgress, RouteLegProgress, and RouteStepProgress
   #2229 registered key-value observers on these classes. Clients can set indices directly and a third object, RouteController, needs to update in response to the change. Though Foundation’s Progress class also inherits from NSObject, it solves this problem by allowing an object to register blocks to handle specific changes. It’s basically a hard-coded subset of KVO, more structured than the delegate pattern. Removing NSObject usage here will unblock codability: #1940.

* [ ] NavigationSettings
   This class also uses KVO, but I think only it registers on its own key paths, for the purpose of sending out notifications. Given the manageable number of properties and outstanding reflection mysteries like #2262, we should replace this mechanism with a series of `didSet`s that call a common method.

* [ ] MapboxNavigationService
* [ ] NavigationOptions
* [x] NavigationEventsManager: #2297
* [ ] EndOfRouteFeedback
* [ ] FeedbackItem
* [ ] CarPlaySearchManager
* [ ] StyleManager
* [ ] DataCache

/cc @mapbox/navigation-ios

Udumft:
Resolves #2294 
Removes some objects NSObject Inheritance. Existing KVO observers are removed in favor of using NSNotifications observations.

There are also some other Public available classes, inheriting NSObject:
- LanesStyleKit
- ManeuversStyleKit
- StyleKitMarker
- SpeedLimitStyleKit
- Style

Didn't check in detail, if such inheritance is legitimate, but quick removing NSObject from declaration showed successful compile. Should I check these too?

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -34,14 +34,15 @@ open class RouteController: NSObject {
     }
     
     private var _routeProgress: RouteProgress {
+        willSet {
+            resetObservation(for: _routeProgress)
+        }
         didSet {
             movementsAwayFromRoute = 0
             updateNavigator(with: _routeProgress)
             updateObservation(for: _routeProgress)
         }
     }
-    
-    private var progressObservation: NSKeyValueObservation?
     
     var movementsAwayFromRoute = 0
     
@@ -149,12 +150,20 @@ open class RouteController: NSObject {
         updateObservation(for: _routeProgress)
     }
     
+    deinit {
+        resetObservation(for: _routeProgress)
+    }
+    
+    func resetObservation(for progress: RouteProgress) {
+        progress.legIndexHandler = nil
+    }
+    
     func updateObservation(for progress: RouteProgress) {
-        progressObservation = progress.observe(\.legIndex, options: [.old, .new]) { [weak self] (progress, change) in
-            guard change.newValue != change.oldValue, let legIndex = change.newValue else {
+        progress.legIndexHandler = { [weak self] (oldValue, newValue) in
+            guard newValue != oldValue else {
                 return
             }
-            self?.updateRouteLeg(to: legIndex)
+            self?.updateRouteLeg(to: newValue)
         }
     }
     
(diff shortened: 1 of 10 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -34,15 +34,16 @@ open class RouteController: NSObject {
     }
     
     private var _routeProgress: RouteProgress {
+        willSet {
+            resetObservation(for: _routeProgress)
+        }
         didSet {
             movementsAwayFromRoute = 0
             updateNavigator(with: _routeProgress)
             updateObservation(for: _routeProgress)
```
Udumft: Shouldn't we not only subscribe to `legIndex` updates here, but also run the subscription handler too to update `RouteController with new `_routeProgress`?

JThramer: I'm not sure what you mean by this? `_routeProgress` is the memoized property that represent's `RouteController`'s copy of the current route progress.

Udumft: When `_routeProgress` is set, we observe it's `legIndex` to call `updateRouteLeg` when it is updated. But shouldn't we call `updateRouteLeg` right now, since we've just updated the entire `_routeProgress` and thus it's `legIndex` is technically updated too.

JThramer: @Udumft No, it's unnecessary, because the [leg is set](https://github.com/mapbox/mapbox-navigation-ios/pull/2352/files#diff-1a2149692b3d7c58b6a0392120d6011aR179) when we call `updateNavigator(with:)`.

PR #2357, opened 2020-04-14
Constrain current route shield; simplify current road name label

### LINKED ISSUE #2355: Custom map style makes current road name label too tall, route shield too big
When using a custom style, the way name can apparently become enlarged, which is undesirable.

Investigate and fix.

/cc @mapbox/navigation-ios

1ec5:
Simplified the code that creates the attributed string to show in the current road name label (WayNameLabel) when the current road is part of a numbered route.

Constrain the route shield in this label to twice the height of the text within the shield, maintaining the image’s aspect ratio. This fixes an issue where the shield was oversized when the current map style was a custom style created in Mapbox Studio.

When the current style uses Mapbox Streets source v8, get the foreground text color [from the road feature](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#--road---shield_text_color-text) instead of using the hard-coded logic that’s based on v7. This corrects the text color for certain shield types that were added in v8.

Fixes #2355. Once this fix lands, we should cherry-pick it into a new branch based on release-v0.39 for a v0.40 release.

/cc @mapbox/navigation-ios


1ec5:
Navigation Guidance Night v4 on master | Navigation Guidance Night v4 with this PR | Navigation style in Studio with this PR
----|----|----
master navigation | pr navigation | pr custom

1ec5:
Cherry-picked to the release-v0.39 branch as 9587e7ee40b84c00f342558dce87ede3a5e26ee9 for v0.39.1.

PR #2359, opened 2020-04-16
Puck desaturation while Location updates are missing

### LINKED ISSUE #2358: Desaturate user puck as user location becomes stale
The arrow in the user puck should desaturate (fade to gray) the longer it takes the location manager to come in with a fresh location.

This is the same as mapbox/mapbox-gl-native#4949, but the navigation SDK has its own puck implementation. It would probably just be a matter of putting the code that sets the color in an animation block. However, we should make sure it’s compatible with the Style/UIAppearance system.

/cc @mapbox/navigation-ios

Udumft:
Resolves #2358 

I tried to find implementation that was done in Maps SDK, but had no luck.
I used Router notifications to receive location updates, and linear interpolation for transitioning between colors.
Not sure if 1 second tick is good enough but since total 'stale' timeout is 1 minute, that shouldn't become visible to a user. Anyway, I thought that these configs might be public so it is possible to adjust it as required.

### DIFF
--- a/MapboxNavigation/UserCourseView.swift
+++ b/MapboxNavigation/UserCourseView.swift
@@ -36,6 +36,19 @@ public extension CourseUpdatable {
  A view representing the user’s location on screen.
  */
 public class UserPuckCourseView: UIView, CourseUpdatable {
+    private var lastLocationUpdate: Date?
+    private var staleTimer: Timer!
+    
+    /// Time interval tick at which Puck view is transitioning into 'stale' state
+    public var staleRefreshInterval: TimeInterval = 1 {
+        didSet {
+            staleTimer.invalidate()
+            initTimer()
+        }
+    }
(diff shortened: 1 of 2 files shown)


MapboxNavigation/UserCourseView.swift
```diff
@@ -36,6 +36,19 @@ public extension CourseUpdatable {
...
+    
+    /// Time interval tick at which Puck view is transitioning into 'stale' state
+    public var staleRefreshInterval: TimeInterval = 1 {
+        didSet {
+            staleTimer.invalidate()
+            initTimer()
+        }
+    }
+    /// Time interval, after which Puck is considered 100% 'stale'
+    public var staleInterval: TimeInterval = 60
```
Udumft, line 50: I thought that it could be handy to make these public. Maybe we should also allow explicit way to "turn off" puck staling?

PR #2360, opened 2020-04-17
Fix Xcode 11.4 builds

1ec5:
Depend on AndriiDoroshko/SnappyShrimp#11, a branch of SnappyShrimp that supports building on Xcode 11.4. Build this configuration on CircleCI.

Also upgraded to Mapbox Maps SDK v5.8.0. After merging, cherry-pick to the release-v0.39 branch.

/cc @mapbox/navigation-ios


1ec5:
Most of the failures were test fixtures for snapshot tests that needed to be renamed to match the new iOS version. But one failure was more difficult to resolve:

> /Users/distiller/project/MapboxNavigationTests/CarPlayManagerTests.swift:398: error: -[MapboxNavigationTests.CarPlayManagerSpec Starting_a_trip__When_configured_to_simulate__starts_navigation_with_a_navigation_service_with_simulation_enabled] : expected to equal <5>, got <1>

`CarPlayManager.currentNavigator` is `nil` by the time the spec tries to start simulating. CarPlayManager holds a weak reference to the CarPlayNavigationViewController because `CarPlayManager.mapTemplate(_:startedTrip:using:)` presents it shortly after setting `currentNavigator` to it. It’s related to this warning in the console:

> 2020-04-17 11:37:07.181709-0700 Example[20040:3588834] Warning: Attempt to present <MapboxNavigation.CarPlayNavigationViewController: 0x7fc93a44f450> on <MapboxNavigation.CarPlayMapViewController: 0x7fc93a638190> whose view is not in the window hierarchy!

The CarPlayNavigationViewController goes away when `CarPlayManager.mapTemplate(_:startedTrip:using:)` goes out of scope, so there’s no opportunity for the test to hang onto it. I don’t encounter this issue when manually starting a trip in CarPlay in the simulator, so I think what changed in the iOS 13.4 SDK is that the application never even shows up in the CarPlay window’s view hierarchy.

Fortunately, these tests only exercise the navigation service, not the UI itself. The TestCarPlayManagerDelegate keeps a reference to the navigation service, so I’m using that reference to start the service instead of going through the CarPlayNavigationViewController.

1ec5:
Cherry-picked to the release-v0.39 branch as 148cf39941df39bb3b5b3f4d842393eba5a390e5 for v0.39.1.

PR #2365, opened 2020-04-24
changing default simulation mode to .never

Skoti:
Hi I would like to change the default simulation mode from `.onPoorGPS` to `.never`.

This is great that Mapbox provides navigation simulation as it is very useful for debugging.
However, the setting is very implicit as it is configured deeply in the `MapboxNavigationService` initializer, which is called by default in `NavigationViewController` initializer.
The last one accepts `NavigationOptions` which one might think these are all the options to possibly customize a navigation service by, which is totally misleading as the navigation service initializer has more (including mentioned simulationMode).
These options also have a possibility to specify a custom instance of NavigationService which one might think is unnecessary by assuming the default one should be safe and configured for the main purpose (real use in the field) of any navigation app.

Thus it is just too easy to miss and ship a simulation mode to production.
I came across this while sitting in an underground garage where GPS could not reach me and the simulation just kicked off! 😄 

I think if one is interested in simulating navigation then they will find a way to turn it on.
It should not be the other way around. The default values should be as simple as possible and work in the most expected way of a general purpose of the framework.

### DIFF
--- a/MapboxCoreNavigation/NavigationService.swift
+++ b/MapboxCoreNavigation/NavigationService.swift
@@ -235,7 +235,7 @@ public class MapboxNavigationService: NSObject, NavigationService {
                          directions: Directions? = nil,
                          locationSource: NavigationLocationManager? = nil,
                          eventsManagerType: NavigationEventsManager.Type? = nil,
-                         simulating simulationMode: SimulationMode = .onPoorGPS,
+                         simulating simulationMode: SimulationMode = .never,
                          routerType: Router.Type? = nil) {
         nativeLocationSource = locationSource ?? NavigationLocationManager()
         self.directions = directions ?? Directions.shared


1ec5:
Hi, thanks for digging in and contributing a patch!

> I came across this while sitting in an underground garage where GPS could not reach me and the simulation just kicked off! 😄

`onPoorGPS` was chosen as the default because the most common case we had come across of poor GPS was in tunnels; it makes sense to simulate in that situation. However, you’re right that it can be confusing for route simulation to kick in while in a parking garage.

> These options also have a possibility to specify a custom instance of NavigationService which one might think is unnecessary by assuming the default one should be safe and configured for the main purpose (real use in the field) of any navigation app.

Indeed, the expectation was that the defaults would be suitable for use in the field, though parking garages are a good counterpoint to that. NavigationOptions – including the ability to enable or disable simulation – does come up [occasionally](https://docs.mapbox.com/ios/navigation/overview/route-generation/#include-multiple-stops) in the documentation, but [the most prominent getting-started examples](https://github.com/mapbox/mapbox-navigation-ios/#configuration) gloss over it in favor of the minimal `NavigationViewController(for:)` initializer. I think the intention when introducing NavigationOptions was that it would be part of the standard workflow for creating a NavigationViewController, so maybe it shouldn’t be completely optional as it is now.

1ec5:
Before `onPoorGPS` was introduced, simulation only kicked in when the route data suggested that there was a tunnel. (This can still happen in a parking garage, because the aisles within a parking garage are sometimes mapped as tunnels in OpenStreetMap.) But it wouldn’t kick in just because location updates stop coming in for a while.

Skoti:
> NavigationOptions – including the ability to enable or disable simulation – does come up occasionally in the documentation, but the most prominent getting-started examples gloss over it in favor of the minimal NavigationViewController(for:) initializer.

If the default simulation mode is going to stay that way then I think it must be highlighted in places like _"Product capabilities"_ and _"Route generation"_ - it is an important setting.

However, I would much more appreciate this to be an opt-in.

---
> (This can still happen in a parking garage, because the aisles within a parking garage are sometimes mapped as tunnels in OpenStreetMap.)

Right now there are just location updates that start it, nothing connected to tunnels.
https://github.com/mapbox/mapbox-navigation-ios/blob/bc36da62797dba2f9b132a430d6026a54d7a9b93/MapboxCoreNavigation/NavigationService.swift#L398-L400

The only thing that tunnels influence is the day/night style:
https://github.com/mapbox/mapbox-navigation-ios/blob/bc36da62797dba2f9b132a430d6026a54d7a9b93/MapboxNavigation/NavigationViewController.swift#L599-L611

---
> Maybe there should be a new SimulationMode, inTunnels, that is only activated within a tunnel when no GPS updates are coming in.

That would be great! 👍

S2Ler:
I'm closing this PR as it got stale. @Skoti Thanks for spinning this discussion with the pull request. I've cut a ticket where we can continue the conversation.

PR #2374, opened 2020-05-27
Offline Directions errors-clarification

### LINKED ISSUE #2273: Misleading guard statement in NavigationDirections.calculate(_:)
https://github.com/mapbox/mapbox-navigation-ios/blob/420b287b0f865e8df8e389921b54973a24524369/MapboxCoreNavigation/OfflineDirections.swift#L151-L157

Currently we're only returning the `OFFLINE_NO_RESULT` error if the directions client is improperly handled (`self` is `nil` when the asynchronous closure executes.)  We should have a separate error for that, and actually check the result to ensure no error was returned from nav native.

/cc @mapbox/navigation-ios

Udumft:
Resolves #2273  
Added new error case for indicating incorrect offline router object handling.

I hope I got original idea, described in the linked issue correctly...

What is the correct way of adding new localizations here? How can I add `OFFLINE_CANCELLED` translation and verify that it's correct?

### DIFF
--- a/MapboxCoreNavigation/OfflineDirections.swift
+++ b/MapboxCoreNavigation/OfflineDirections.swift
@@ -17,17 +17,6 @@ public enum OfflineRoutingError: LocalizedError {
      A Directions API error can occur whether directions are calculated online or offline.
      */
     case standard(DirectionsError)
-    
-    /**
-     The router did not finish the request
-     */
-    case cancelled
-    
-    /**
-     The router returned an empty response.
-     */
-    case noData
-    
     /**
      The router returned a response that isn’t correctly formatted.
     */
@@ -39,10 +28,6 @@ public enum OfflineRoutingError: LocalizedError {
         switch self {
         case .standard(let error):
             return error.localizedDescription
-        case .cancelled:
-            return NSLocalizedString("OFFLINE_CANCELLED", bundle: .mapboxCoreNavigation, value: "Routing was cancelled before response could be acquired.", comment: "Error description when Offline Router was deallocated before receiving the API response")
-        case .noData:
-            return NSLocalizedString("OFFLINE_NO_RESULT", bundle: .mapboxCoreNavigation, value: "Unable to calculate the requested route while offline.", comment: "Error description when an offline route request returns no result")
         case .invalidResponse:
             return NSLocalizedString("OFFLINE_CORRUPT_DATA", bundle: .mapboxCoreNavigation, value: "Found an invalid route while offline.", comment: "Error message when an offline route request returns a response that can’t be deserialized")
         case .unknown(let underlying):
@@ -181,7 +166,7 @@ public class NavigationDirections: Directions {
      
      - parameter options: A `RouteOptions` object specifying the requirements for the resulting routes.
      - parameter offline: Determines whether to calculate the route offline or online.
-     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread.
+     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread. If called `NavigationDirections` instance is deallocated before route calculation is finished - completion won't be called.
      */
     public func calculate(_ options: RouteOptions, offline: Bool = true, completionHandler: @escaping OfflineRouteCompletionHandler) {
         
@@ -203,16 +188,6 @@ public class NavigationDirections: Directions {
         
         NavigationDirectionsConstants.offlineSerialQueue.async { [weak self] in
             guard let result = self?.navigator.getRouteForDirectionsUri(url.absoluteString) else {
-                DispatchQueue.main.async {
-                    completionHandler(session, .failure(.cancelled))
-                }
-                return
-            }
-            
-            guard result.isSuccess else {
-                DispatchQueue.main.async {
-                    completionHandler(session, .failure(.noData))
-                }
                 return
             }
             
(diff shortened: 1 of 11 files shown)


MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -196,6 +203,13 @@ public class NavigationDirections: Directions {
...
             guard let result = self?.navigator.getRouteForDirectionsUri(url.absoluteString) else {
+                DispatchQueue.main.async {
+                    completionHandler(session, .failure(.cancelled))
+                }
+                return
+            }
+            
+            guard result.isSuccess else {
                 DispatchQueue.main.async {
                     completionHandler(session, .failure(.noData))
```
Udumft, line 214: Seem like `.noData` is not the correct option here. `result` has only 2 properties exposed: `isSuccess` (which is so far was unused) and `json` (which may contain a valid response or an error message). In current implementation, I struggle to find a correct definition of "router returned no data" case to fire `.noData` message.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -196,6 +203,13 @@ public class NavigationDirections: Directions {
         
         NavigationDirectionsConstants.offlineSerialQueue.async { [weak self] in
             guard let result = self?.navigator.getRouteForDirectionsUri(url.absoluteString) else {
```
1ec5, line 205: Per https://github.com/mapbox/mapbox-navigation-ios/issues/2273#issuecomment-594084579, the only error case we should handle up here would be that `self` (NavigationDirections) has been deallocated.

Upon closer inspection, I’m wondering if it’s even appropriate to call the completion handler in this case. For mapbox/mapbox-gl-native-ios#210 https://github.com/mapbox/mapbox-gl-native-ios/issues/200#issuecomment-598458116, we found that calling the completion handler after `self` has gone away forces the caller to consider whether `self` has gone away because the parent view controller has gone away or the application is terminating.

In other words, when an application calls this method from a MapViewController that holds a strong reference to NavigationDirections, does it expect the completion handler to magically be called even if that MapViewController has been deallocated (for example, because the user has exited that map screen to go to a different part of the application)? If the application has neglected to make `self` weak inside the completion handler, that could create other unexpected behavior in the application. The completion handler would have to explicitly handle the `.cancelled` case, so at least we would have to document that postcondition on the method, which is unfortunate.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -18,6 +18,11 @@ public enum OfflineRoutingError: LocalizedError {
      */
     case standard(DirectionsError)
     
+    /**
+     The router did not finish the request
+     */
+    case cancelled
```
1ec5, line 24: Amazingly, Apple can’t decide whether it should be “[canceled](https://developer.apple.com/search/?q=canceled)” or “[cancelled](https://developer.apple.com/search/?q=cancelled)”. Either is fine here. 🤷‍♂️

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -34,6 +39,8 @@ public enum OfflineRoutingError: LocalizedError {
         switch self {
         case .standard(let error):
             return error.localizedDescription
+        case .cancelled:
+            return NSLocalizedString("OFFLINE_CANCELLED", bundle: .mapboxCoreNavigation, value: "Routing was cancelled before response could be acquired.", comment: "Error description when Offline Router was deallocated before receiving the API response")
```
1ec5, line 43: Here are [some instructions](https://github.com/mapbox/mapbox-navigation-ios/blob/master/CONTRIBUTING.md#adding-user-facing-text) for adding user-facing text to the SDK. The only remaining step would be to run scripts/extract_localizable.sh to add this string to a strings file.

MapboxCoreNavigation/OfflineDirections.swift
1ec5: This return value is a holdover from 0486778621f774fa88289617dd0c18b12d227ab6. It might be fine to reuse `.noData` for the `.cancelled` case. In the event of a failure, there should be an error that we can pass along.

MapboxCoreNavigation/OfflineDirections.swift
Udumft: Not calling a completion under some conditions may also lead to unexpected behavior. At least such use case must also be documented so that users are aware that completion call is not guaranteed.
Also, if we choose not to rise an explicit error message when `NavigationDirections` were deallocated, `noData` error becomes unused, so we can safely remove it

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -174,7 +166,7 @@ public class NavigationDirections: Directions {
      
      - parameter options: A `RouteOptions` object specifying the requirements for the resulting routes.
      - parameter offline: Determines whether to calculate the route offline or online.
-     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread.
+     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread. If called `NavigationDirections` instance is deallocated before route calculation is finished - completion won't be called.
```
MaximAlien, line 169: Not familiar with context of this problem, so just minor comment from me: probably better to limit length of lines to around 100-120, just for better readability.

Issue #1434, opened 2018-05-15
Alternates routes should be under main routes casing


bsudekum:
The darker blue casing for the main route should be above the line used for alternate routes.

![image](https://user-images.githubusercontent.com/1058624/40085631-040f807a-5850-11e8-8165-5764c5126a7e.png)

1ec5:
For what it’s worth, it’s not uncommon in map design for a single casing to surround a full set of related roadways instead of delineating each individual roadway. The casing above could be seen as a way to distinguish the full set of routes from the rest of the map, as opposed to distinguishing the selected route from the others routes. In other words, all the routes are contrasted as a whole against the base map, while the selected route is highlighted.

Putting casing between the selected route and the other routes will increase the visual noise of the map. I suppose it’s possible that there would also be increased legibility, but I’m on the fence about making that tradeoff.

/cc @aparlato @nickidlugash

captainbarbosa:
This will be fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/2377.

A preview:

<img width="328" alt="Screen Shot 2020-06-17 at 5 19 17 PM" src="https://user-images.githubusercontent.com/10850812/84963707-f6da5500-b0be-11ea-9060-4ae4c1747cc6.png">

captainbarbosa:
Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/2377 ✅

Issue #1307, opened 2018-04-17
Deemphasize portions of the route line that the user has passed


1ec5:
Portions of the route line that the user has already passed should be displayed in a more muted or translucent color, to better emphasize the portion that the user still has yet to reach. This effect would be most prominent in overview mode, but even in straight-ahead mode, the line would look different above and below the user puck.

Contrary to the initial explorations in #1162, we’ll keep the line’s geometry constant as the user moves. However, we’ll change the line gradient (`MGLLineStyleLayer.lineGradient`) to abruptly go from translucent to opaque at the user’s location along the line, with a smooth transition (`lineGradientTransition`) that matches the puck’s animated interpolation. Line gradients will be implemented in the map SDK for mapbox/mapbox-gl-native#11718.

/cc @mapbox/navigation-ios @danesfeder

1ec5:
> Line gradients will be implemented in the map SDK for mapbox/mapbox-gl-native#11718.

This is happening in mapbox/mapbox-gl-native#12575. 🎉

1ec5:
> we’ll change the line gradient (`MGLLineStyleLayer.lineGradient`) to abruptly go from translucent to opaque at the user’s location along the line, with a smooth transition (`lineGradientTransition`) that matches the puck’s animated interpolation.

This is blocked by the fact that data-driven properties don’t support transitions: mapbox/mapbox-gl-js#3170. So even once the map SDK supports line gradients, we’d have to constantly change the `lineGradient` property’s interpolation stops on each frame, which takes us back to square one: #1162.

1ec5:
Also blocked by mapbox/mapbox-gl-native#12917, which appears to be much more straightforward to fix than the other blocker, mapbox/mapbox-gl-js#3170.

asinghal22:
@1ec5 Can you please double check the milestone. I accidently put this on 1.1 milestone. I think it is v0.x next.

1ec5:
No, it’s correctly slated for v1.1.0.

captainbarbosa:
@1ec5 is https://github.com/mapbox/mapbox-gl-js/issues/3170 still considered a requirement to complete this feature?

1ec5:
[...]

PR #2377, opened 2020-05-28
Fade route line during navigation

captainbarbosa:
Fixes https://github.com/mapbox/mapbox-navigation-ios/issues/1307
Fixes https://github.com/mapbox/mapbox-navigation-ios/issues/1434

⚠️ Breaking changes expected to land in v1.0

Adds the option to fade the route line as the user navigates along a route. A custom fade out color can also be specified. the default is clear, to give the appearance of a "vanishing" line as the puck travels along a route.

vanish-route

This also fixes https://github.com/mapbox/mapbox-navigation-ios/issues/1434 because I've broken out the main route / alternate routes into different casing layers:

Screen Shot 2020-06-17 at 5 19 17 PM

The vanishing route color can also be customized by overriding `NavigationMapView.appearance().vanishingRouteColor`:

Screen Shot 2020-06-18 at 6 53 05 PM

Within this change, `NavigationMapViewDelegate.navigationMapView(_:routeStyleLayerWithIdentifier:source:)` and  `NavigationMapViewDelegate.navigationMapView(_:routeCasingStyleLayerWithIdentifier:source:)` are removed in favor of four new delegate methods to customize the route styling:
  * `NavigationMapViewDelegate.navigationMapView(_:mainRouteStyleLayerWithIdentifier:source:)` to style the main route.
  * `NavigationMapViewDelegate.navigationMapView(_:mainRouteCasingStyleLayerWithIdentifier:source:)` to style the casing of the main route.
  * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteStyleLayerWithIdentifier:source:)` to style alternative routes.
  * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteCasingStyleLayerWithIdentifier:source:)` to style the casing of alternative routes.

cc @1ec5 @fabian-guerra


MapboxNavigation/NavigationMapView.swift
```diff
@@ -690,6 +690,45 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+     */
+    
+    public func fadeRouteLine(percentComplete: Double) {
+        guard let style = self.style else { return }
+        
+        /**
+         This is necessary to prevent duplicate keys in the dictionary
+         defining the gradient. It's not for aesthetic purposes.
+         */
+        let gradientToTransparent: Double = 0.0005
```
1ec5, line 704: https://github.com/mapbox/mapbox-navigation-ios/issues/1307#issuecomment-635461995 discusses what it’ll take to eliminate this fudge factor, which doesn’t work if the route is over a certain length.

If for some reason we need to keep a fudge factor, we might be able to refine it a bit, considering that mbgl considers these stops to be floats. Instead of this constant, cast `percentComplete` to `Float` and add `.nextUp` to it. That property is the smallest possible larger value in the type.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -690,6 +690,45 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            casingStops [0] = routeCasingColor.withAlphaComponent(0)
+            stops[percentComplete] = trafficUnknownColor.withAlphaComponent(0)
+            stops[percentComplete + gradientToTransparent] = trafficUnknownColor.withAlphaComponent(1)
+            casingStops[percentComplete] = routeCasingColor.withAlphaComponent(0)
+            casingStops[percentComplete + gradientToTransparent] = routeCasingColor.withAlphaComponent(1)
+        }
+
+        if let routeLine = style.layer(withIdentifier: StyleLayerIdentifier.route) as? MGLLineStyleLayer,
+           let routeLineCasing = style.layer(withIdentifier: StyleLayerIdentifier.routeCasing) as? MGLLineStyleLayer {
+            routeLine.lineGradient = NSExpression(format: "mgl_interpolate:withCurveType:parameters:stops:($lineProgress, 'linear', nil, %@)", stops)
```
1ec5, line 727: By default, `routeLineCasing` is a single LineString feature representing the entire leg, whereas `routeLine` is a FeatureCollection of LineStrings, one per traffic congestion segment. Applying the gradient to this layer causes the exotic dotted line pattern in the screenshot above. If we don’t apply the gradient to this layer, then there may be random-looking, disconnected yellow and red lines behind the puck.

Maybe we could split the difference by making the casing layer become 75% translucent, not transparent, behind the puck. That would preserve the erasing effect but ensure the user can intuit the purpose of those traffic congestion lines.

/cc @d-prukop @LukasPaczos

LukasPaczos: On Android, we've converted the route line to be a single feature as well, to avoid the "dotted" appearance. The traffic is now represented with the percentage-based gradient expression instead of coloring single traffic segment features independently. As a side note, this also significantly reduced the memory footprint of the app because we don't have to split the route geometry into segments and store them.
/cc @cafesilencio

captainbarbosa: On iOS we have a reference to both the entire route line and the individual segments. Both are rendered on the map, it seems. I'm not sure if the segments are used for anything else other than traffic coloring - @1ec5 do you know if we could get by with just rendering the single route (`routeLineCasing`) instead?

cafesilencio:
The original android implementation did the same thing with multiple
segments used for traffic coloring. Currently we have one line feature for
the route line on its own layer with another line feature representing the
route shield line (under the route line) on its own layer.

Please see this BMW issue:
https://github.com/mapbox/bmw-collab/issues/191#issuecomment-634831148
There may be a need to add a third layer with a third line representing
just the traffic.  It seems BMW wants a different width for the route line
and the traffic representation. They have also asked for the ability to
customize the scaling of the lines at different zoom levels but I don't
know the scope of that. As far as I know that isn't expected for the june
30 release of the Android SDK but I'm not 100% sure.

MapboxNavigation/NavigationMapView.swift
1ec5: The segments are only used for traffic congestion coloring. If we can implement them as part of an overall route line feature, then that would be ideal. I think we still need a second layer so that the fill is adequately cased – the style specification doesn’t provide many stroking options otherwise – but its shape source can contain a single feature instead of a multitude of them.

LukasPaczos: > I think we still need a second layer so that the fill is adequately cased

Yes, that's also the case on Android. The route line and traffic are a single feature laid on top of route line casing (route line shield) which is another feature, equal to the same geometry but a little bit wider and with a different color. I missed mentioning that above.

captainbarbosa:
I'm working on nailing down the implementation in a side project first as a minimal test case before porting back to here - stay tuned.

captainbarbosa:
This is the current [state of the prototype](https://gist.github.com/captainbarbosa/99129e0a50ac4480d900ee8f80fa3ac7) - the gradients need to be fine-tuned still but I'm hopeful - seems like just a matter of nailing down the right numbers for the stops.

captainbarbosa:
Looking better - here's a fake route toggling between traffic styled with two difference sources and layers. The difference is subtle, and even more indistinguishable when you end up calculating the gradient stops dictionary with `CGFloat`'s `nextUp`/`nextDown` as @1ec5 suggested in https://github.com/mapbox/mapbox-navigation-ios/pull/2377#discussion_r432772524.

One layer contains individual congestion segments styled with an `MGL_MATCH` expression. The other layer is a singular line feature styled with a line gradient. [Prototype for reference.](https://gist.github.com/captainbarbosa/99129e0a50ac4480d900ee8f80fa3ac7#file-gradient-route-swift-L99)

The next step is to dynamically transform the stops dictionary the gradient is using to simulate the disappearance of traveled-over parts of the line.

captainbarbosa:
I got the gradient to fade today over time. Here's that same fake route line fading from 0-100% on a timer, leaving a white trail in its path:



Next up, I'll work on integrating the code for this prototype into this PR.

captainbarbosa:
As of https://github.com/mapbox/mapbox-navigation-ios/pull/2377/commits/25fcd745327066369246c22693909e205e607403 the route line's traffic is now expressed with a single line gradient. Next, I need to work on how to only display this traffic only for the main route and not the alternate routes:

Screen Shot 2020-06-15 at 5 35 01 PM

captainbarbosa:
This might also take care of https://github.com/mapbox/mapbox-navigation-ios/issues/1434...

captainbarbosa:
Things are looking good (see new updated PR description). There's some additional refactoring I need to do before this PR is ready for review, but the visual implementation looks about right.

captainbarbosa:
Ready for a first review!

captainbarbosa:
@1ec5 suggested this should be an opt-in feature, so I can go ahead and try to include that option here.

fabian-guerra:
@captainbarbosa I think you can make a new pr that targets this branch.

captainbarbosa:
Opt-in setting moving in https://github.com/mapbox/mapbox-navigation-ios/pull/2418 - that should be merged into this PR first.

captainbarbosa:
https://github.com/mapbox/mapbox-navigation-ios/pull/2418 has merged into this PR so now this feature is opt-in. I did some additional manual testing today to make sure the vanishing feature still works as expected when traffic isn’t enabled 👍

MapboxNavigation/NavigationMapView.swift
```diff
@@ -442,39 +469,139 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            // Add all the layers in the correct order
             for layer in style.layers.reversed() {
                 if !(layer is MGLSymbolStyleLayer) &&
                 layer.identifier != StyleLayerIdentifier.arrow && layer.identifier != StyleLayerIdentifier.arrowSymbol && layer.identifier != StyleLayerIdentifier.arrowCasingSymbol && layer.identifier != StyleLayerIdentifier.arrowStroke && layer.identifier != StyleLayerIdentifier.waypointCircle {
-                    style.insertLayer(line, below: layer)
-                    style.insertLayer(lineCasing, below: line)
+                    style.insertLayer(mainRouteLayer, below: layer)
+                    style.insertLayer(mainRouteCasingLayer, below: mainRouteLayer)
+                    style.insertLayer(alternateRoutesLayer, below: mainRouteCasingLayer)
+                    style.insertLayer(alternateRoutesCasingLayer, below: alternateRoutesLayer)
```
1ec5, line 512: This is the part that fixes #1434. I’m a little uneasy about the fact that we’ll be inserting four layers for the route line, but it looks like it’ll be straightforward (and maybe not critical) to recombine a couple of them by conditionalizing properties (or individual stop values) on whether an `isAlternativeRoute` or `isCasing` attribute is set. Not necessary for landing this PR though.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -442,39 +469,139 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        if let minStop = filtered.min(by: { $0.0 < $1.0 }) {
+            filtered[0.0] = vanishingRouteColor
+            filtered[percentTraveled.nextDown] = vanishingRouteColor
+            filtered[percentTraveled] = minStop.value
+        }
+
+        routeGradientStops.line = filtered
+        mainRouteLayer.lineGradient = NSExpression(format: "mgl_interpolate:withCurveType:parameters:stops:($lineProgress, 'linear', nil, %@)", routeGradientStops.line)
+
+        // TODO: Refactor? We're doing the same work twice here...
```
1ec5, line 591: Still to-do?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -442,39 +469,139 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    func alternateRouteCasingStyleLayer(identifier: String, source: MGLSource) -> MGLLineStyleLayer {
+        let alternateRoutesCasingLayer = MGLLineStyleLayer(identifier: identifier, source: source)
+        alternateRoutesCasingLayer.predicate = NSPredicate(format: "isAlternateRoute == true")
+        alternateRoutesCasingLayer.lineColor = NSExpression(forConstantValue: routeAlternateCasingColor)
+        alternateRoutesCasingLayer.lineWidth = NSExpression(format: "mgl_interpolate:withCurveType:parameters:stops:($zoomLevel, 'linear', nil, %@)", MBRouteLineWidthByZoomLevel.multiplied(by: 1.5))
+        alternateRoutesCasingLayer.lineJoin = NSExpression(forConstantValue: "round")
+        return alternateRoutesCasingLayer
+    }
+
+    func fadeRoute(_ fractionTraveled: Double) {
```
1ec5, line 567: Maybe call it something more generic like `updateRouteLine(fractionTraveled:)` (with a named argument).

MapboxNavigation/NavigationMapView.swift
```diff
@@ -899,40 +1035,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-    
-    func routeStyleLayer(identifier: String, source: MGLSource) -> MGLStyleLayer {
-        let line = MGLLineStyleLayer(identifier: identifier, source: source)
-        line.lineWidth = NSExpression(format: "mgl_interpolate:withCurveType:parameters:stops:($zoomLevel, 'linear', nil, %@)", MBRouteLineWidthByZoomLevel)
-        line.lineOpacity = NSExpression(forConditional:
-            NSPredicate(format: "isAlternateRoute == true"),
-                                        trueExpression: NSExpression(forConstantValue: 1),
-                                        falseExpression: NSExpression(forConditional: NSPredicate(format: "isCurrentLeg == true"),
-                                                                      trueExpression: NSExpression(forConstantValue: 1),
-                                                                      falseExpression: NSExpression(forConstantValue: 0)))
```
1ec5, line 911: Previously, on a multileg route, any completed legs would continue to be shown but translucently. Now do they disappear entirely whenever `traversedRouteLineColor` is transparent? Same goes with already-passed intermediate waypoint markers.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -899,40 +1035,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-        line.lineColor = NSExpression(format: "TERNARY(isAlternateRoute == true, %@, MGL_MATCH(congestion, 'low' , %@, 'moderate', %@, 'heavy', %@, 'severe', %@, %@))", routeAlternateColor, trafficLowColor, trafficModerateColor, trafficHeavyColor, trafficSevereColor, trafficUnknownColor)
-        line.lineJoin = NSExpression(forConstantValue: "round")
-        
-        return line
+
+    func generateTrafficGradientStops(for route: Route) {
```
1ec5, line 1039: As far as I can tell, this is a method that has the side effect of setting a property on NavigationMapView. It would be more straightforward to return the generated stops and rely on the caller to set it on the appropriate property.

(If there aren’t too many dependencies on NavigationMapView in this method, we could even move it to an extension on Route, as a computed property `congestionGradientStops`, so that we could potentially reuse the logic for other purposes.)

MapboxNavigation/NavigationMapView.swift
```diff
@@ -899,40 +1035,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            let buffPtr = UnsafeMutableBufferPointer(start: line.coordinates, count: Int(line.pointCount))
+            let lineCoordinates = Array(buffPtr)
+
+            // Get congestion color for the stop.
+            let congestionLevel = line.attributes["congestion"] as! String
+            let congestionColor = getCongestionColor(for: congestionLevel)
+
+            // Measure the line length of the traffic segment.
+            let lineString = LineString(lineCoordinates)
+            guard let distance = lineString.distance() else { return }
```
1ec5, line 1085: Initially, traffic congestion segments are straight lines between two coordinates, but I guess it’s necessary to use `LineString.distance()` because neighboring congestion segments of the same level have already been consolidated?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -899,40 +1035,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+         transition between congestion segments.
+
+            green       gradient       red
+                       transition
+         |-----------|~~~~~~~~~~~~|----------|
+         0         0.499        0.501       1.0
+         */
+
+        for (index, line) in congestionSegments.enumerated() {
+            line.getCoordinates(line.coordinates, range: NSMakeRange(0, Int(line.pointCount)))
```
1ec5, line 1074: This line looks like it replaces `line`’s coordinates with `line`’s coordinates. Can we remove it?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -899,40 +1035,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            //  `UnsafeMutablePointer` is needed here to
```
1ec5, line 1075: Trailing thought. I think the intention is to say that the UnsafeMutableBufferPointer is needed to get the line’s coordinates.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -899,40 +1035,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
-        lineCasing.lineOpacity = NSExpression(forConditional: NSPredicate(format: "isAlternateRoute == true"),
-                                            trueExpression: NSExpression(forConstantValue: 1),
-                                            falseExpression: NSExpression(forConditional: NSPredicate(format: "isCurrentLeg == true"), trueExpression: NSExpression(forConstantValue: 1), falseExpression: NSExpression(forConstantValue: 0.85)))
-        
-        return lineCasing
+
+    /**
+     Given a congestion level, return its associated color.
+     */
+    private func getCongestionColor(for congestionLevel: String) -> UIColor {
```
1ec5, line 1141: This method has no side effects and is synchronous, so rename it to `congestionColor(for:)`.

MapboxNavigation/NavigationMapViewDelegate.swift
```diff
@@ -6,15 +6,46 @@ import MapboxCoreNavigation
...
+
+    /**
+     Asks the receiver to return an MGLStyleLayer for the alternative route lines, given an identifier and source.
+     This method is invoked when the map view loads and any time routes are added.
+     - parameter mapView: The NavigationMapView.
+     - parameter identifier: The style identifier.
+     - parameter source: The source containing the route data that this method would style.
+     - returns: An MGLStyleLayer that is applied to alternative routes.
+    */
+    func navigationMapView(_ mapView: NavigationMapView, alternateRouteStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
```
1ec5, line 38: MapboxDirections uses the term “alternat**ive** route”, so I think we should stick to that for consistency.

CHANGELOG.md
```diff
@@ -30,6 +30,13 @@
 * Fixed ungrammatical spoken instructions at sharp turns in English. ([valhalla/valhalla#2226](https://github.com/valhalla/valhalla/pull/2226))
 * Fixed an issue where spoken and visual instructions sometimes omitted the cardinal direction after a route number in the United States. ([valhalla/valhalla#1917](https://github.com/valhalla/valhalla/pull/1917))
 * A spoken instruction about an exit ramp no longer specifies the side of the road if the ramp branches off the slow lane (on the right side in regions that drive on the right). ([valhalla/valhalla#1990](https://github.com/valhalla/valhalla/pull/1990))
+* Removed `NavigationMapViewDelegate.navigationMapView(_:routeStyleLayerWithIdentifier:source:)`, `NavigationMapViewDelegate.navigationMapView(_:routeCasingStyleLayerWithIdentifier:source:)` in favor of four new delegate methods to customize the route styling ([#2377](https://github.com/mapbox/mapbox-navigation-ios/pull/2377)):
```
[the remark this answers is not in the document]
captainbarbosa: I think this got shuffled in the rebase, just corrected it.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -112,6 +128,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
     var routes: [Route]?
     var isAnimatingToOverheadMode = false
+
+    /**
+     A tuple that represents the dictionary of percentage/color pairs used to
+     calculate the color gradient transitions belonging to the main route line
+     and its casing.
+     */
+    typealias RouteGradientStops = (line: [CGFloat: UIColor], casing: [CGFloat: UIColor])
+    private var routeGradientStops = RouteGradientStops(line: [:], casing: [:]) // TODO: Account for cases where traffic isn't enabled.
```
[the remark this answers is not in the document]
captainbarbosa: Its resolved, I just forgot to remove it :)

MapboxNavigation/NavigationMapView.swift
```diff
@@ -112,6 +128,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     var altitude: CLLocationDistance
     var routes: [Route]?
     var isAnimatingToOverheadMode = false
+
+    /**
+     A tuple that represents the dictionary of percentage/color pairs used to
+     calculate the color gradient transitions belonging to the main route line
+     and its casing.
+     */
+    typealias RouteGradientStops = (line: [CGFloat: UIColor], casing: [CGFloat: UIColor])
```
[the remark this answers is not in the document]
captainbarbosa: They are both altered at the same time when `routeLineTracksTraversal` is enabled, but I don't think the behavior would be affected by using two parallel properties.

MapboxNavigation/NavigationMapView.swift
captainbarbosa: Yep, that's right.

MapboxNavigation/NavigationMapView.swift
captainbarbosa: Line `1074` is required or else the congestion segment stops don't get calculated correctly.

MapboxNavigation/NavigationMapView.swift
captainbarbosa: Oops yeah..! Forgot to finish my sentence there.

MapboxNavigation/NavigationMapView.swift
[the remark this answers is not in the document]
captainbarbosa: Hm, something's off when I try to use two parallel properties:

Tuple (current)

```swift
typealias RouteGradientStops = (line: [CGFloat: UIColor], casing: [CGFloat: UIColor])
private var routeGradientStops = RouteGradientStops(line: [:], casing: [:])
```

Screen Shot 2020-06-24 at 3 58 47 PM

Two parallel properties (proposed)

```swift
private var routeGradientStops: [CGFloat: UIColor] = [:]
private var routeGradientCasingStops: [CGFloat: UIColor] = [:]
```

Screen Shot 2020-06-24 at 3 57 33 PM

Probably worth investigating further, but I wouldn't say its a blocker so I'll note this for follow up work.

captainbarbosa:
I've integrated all feedback, minus this weird behavior which [I've noted for later](https://github.com/mapbox/mapbox-navigation-ios/issues/2423) and is not a blocker for this PR. The CI failure for the `ci/circleci: Xcode_10.2_iOS_12.2_CP_install` job is expected (caused by https://github.com/mapbox/mapbox-navigation-ios/issues/2394), so we're good to merge assuming everything else passes.

PR #2387, opened 2020-06-04
Puck Location fix

Udumft:
Resolves [puck location issue, caused by the bottom banner update](https://github.com/mapbox/support-enterprise/issues/211).
Fixed updating Puck location on various events, including `NavigationMapView` resizing.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -340,16 +340,20 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             return
         }
         
+        let centerUserCourseView = { [weak self] in
+            guard let point = self?.convert(location.coordinate, toPointTo: self) else { return }
+            self?.userCourseView.center = point
+        }
+        
         if tracksUserCourse {
+            centerUserCourseView()
+            
             let newCamera = camera ?? MGLMapCamera(lookingAtCenter: location.coordinate, altitude: altitude, pitch: 45, heading: location.course)
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: .linear) : nil
             setCamera(newCamera, withDuration: duration, animationTimingFunction: function, completionHandler: nil)
         } else {
             // Animate course view updates in overview mode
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -340,16 +340,20 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             return
         }
         
+        let centerUserCourseView = { [weak self] in
+            guard let point = self?.convert(location.coordinate, toPointTo: self) else { return }
+            self?.userCourseView.center = point
+        }
+        
         if tracksUserCourse {
+            centerUserCourseView()
```
Udumft, line 349: This is the smoothest implementation I found so far, to make user puck slide in place gracefully...

1ec5: #2211 removed the line that set the view’s `center`. Can you check that it doesn’t introduce any regressions related to that PR? Thanks!

Udumft: Issue description and reproduction steps for #2211 are a bit cryptic for me, but as I understood, the issue was about incorrect puck positioning during navigation, together with zooming/panning/rotating map view. I checked that user puck is located right on the route regardless of camera position and switching to/from course tracking mode. Is that what #2211 was about?

1ec5: Unfortunately, I’m not super clear on it myself; I’m not entirely convinced that #2211 was the right fix for #2145, and it definitely wasn’t a fix for mapbox/mapbox-gl-native#15232 or mapbox/mapbox-gl-native#15233.

The main things to watch out for are https://github.com/mapbox/mapbox-navigation-ios/issues/2145#issuecomment-512611403 and https://github.com/mapbox/mapbox-navigation-ios/issues/2145#issuecomment-535475988.

Udumft: I could not reproduce symptoms described in linked comments...

fabian-guerra: If I understand correctly, the Nav SDK is not using the Maps SDK user location annotation, which means it will not account for `contentInset` changes. I think this pr is trying to update the user location annotation considering the "insets".

PR #2403, opened 2020-06-16
Add exception in case if leg in route doesn't contain any steps.

### LINKED ISSUE #2399: mapbox-navigation-ios is crashing when using RouteOptions.
**Mapbox Navigation SDK version: 0.40.0**

In case if `RouteOptions` is used instead of `NavigationRouteOptions` like this:
```swift
let origin = CLLocationCoordinate2DMake(37.77440680146262, -122.43539772352648)
let destination = CLLocationCoordinate2DMake(37.76556957793795, -122.42409811526268)
let routeOptions = RouteOptions(coordinates: [origin, destination]) // instead of NavigationRouteOptions

Directions.shared.calculate(routeOptions) { [weak self] (session, result) in
    switch result {
    case .failure(let error):
        print(error.localizedDescription)
    case .success(let response):
        guard let route = response.routes?.first, let strongSelf = self else {
            return
        }
        
        let navigationService = MapboxNavigationService(route: route, routeOptions: routeOptions, simulating: simulationIsEnabled ? .always : .onPoorGPS)
        let navigationViewController = NavigationViewController(for: route, routeOptions: routeOptions, navigationOptions: navigationOptions)
        navigationViewController.mapView?.delegate = strongSelf
        
        strongSelf.present(navigationViewController, animated: true, completion: nil)
    }
}
```
 mapbox-navigation-ios will crash during `MapboxNavigationService` creation. 
Crash log is attached:
[Navigation-Examples_2020-06-11-194331_MacBook-Pro.txt](https://github.com/mapbox/mapbox-navigation-ios/files/4768492/Navigation-Examples_2020-06-11-194331_MacBook-Pro.txt)

MaximAlien:
Fixes #2399.

I don't really like current fix (basically it only gives the ability to have more informative error message other than out of bounds exception), but I guess this is better than propagation of exception handling to upper layers of the API (it'd involve more code changes) or using init?.
Please let me know if you see better ways of handling this. 

I've also found that example app wasn't handling edge case with `RouteOptions`/`NavigationRouteOptions` `includesSteps` property set to `false`.

UPDATE:
Since throwing an exception leads to quite a few changes in public API I've decided to replace it with precondition. Precondition will work for both Debug and Release modes and I don't feel that we need an ability to recover from such type of error.


MaximAlien:
Also as a side note: I have a feeling that we lack logging in Navigation SDKs, it'd be really useful to have logging levels during troubleshooting. As an action item I can create separate ticket to improve logging.

1ec5:
> it'd be really useful to have logging levels during troubleshooting.

Is this because of all the spew from UnimplementedLogging? That mechanism uses Unified Logging, so it should be possible to filter by category.

MaximAlien:
> > it'd be really useful to have logging levels during troubleshooting.
> 
> Is this because of all the spew from UnimplementedLogging? That mechanism uses Unified Logging, so it should be possible to filter by category.

I meant that I have a feeling that we don't have enough logging, not excessive logging from `UnimplementedLogging` with warnings from unimplemented methods. There are several cases when logs will be useful: e.g. in case of rerouting I can see that only notification/delegate will be sent/triggered (which _might_ be handled by end user), other than that there is no way to trace it. There are many other cases of course.

1ec5:
Aside from the delegate methods, some things like rerouting trigger events that get sent to MapboxMobileEvents, and MapboxNavigationNative can record all the method calls sent to it. It does make sense to add logging around the kinds of stuff we already hook up to delegate methods, notifications, and events. I think that would be worth pursuing separately from this PR.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -180,7 +180,7 @@ open class RouteProgress {
         self.route = route
         self.routeOptions = options
         self.legIndex = legIndex
-        self.currentLegProgress = RouteLegProgress(leg: route.legs[legIndex], stepIndex: 0, spokenInstructionIndex: spokenInstructionIndex)
+        self.currentLegProgress = try! RouteLegProgress(leg: route.legs[legIndex], stepIndex: 0, spokenInstructionIndex: spokenInstructionIndex)
```
dersim-davaod, line 183: Concern about explicit `try!` here.
Can we show explicitly the invariant (via adding assert, say)?

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -274,6 +274,10 @@ open class RouteProgress {
     }
 }
 
+enum RouteLegProgressError: Error {
+    case SteplessLegError(String)
```
dersim-davaod, line 278: nit:
```suggestion
    case SteplessLeg(String)
```

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -22,7 +22,7 @@ open class RouteProgress {
         didSet {
             assert(legIndex >= 0 && legIndex < route.legs.endIndex)
```
dersim-davaod, line 23: nit:
use `.contains` syntax:
```swift
(0..<route.legs.endIndex).contains(legIndex)
```
?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -707,7 +707,7 @@ extension NavigationViewController: TopBannerViewControllerDelegate {
         guard let stepIndex = leg.steps.firstIndex(of: step) else { return }
         let nextStepIndex = stepIndex + 1
         
-        let legProgress = RouteLegProgress(leg: leg, stepIndex: stepIndex)
+        let legProgress = try! RouteLegProgress(leg: leg, stepIndex: stepIndex)
```
dersim-davaod, line 710: Same concern about force unwrapping here.
Actually we do the crash in the worst-case scenario. Probably that's ok, but can we try avoid this?

MapboxCoreNavigation/RouteProgress.swift
MaximAlien: I think it's okay to leave this piece of functionality as it is. Don't see any significant difference.

MapboxCoreNavigation/RouteProgress.swift
MaximAlien: Replaced exception with precondition.

PR #2405, opened 2020-06-16
Re-enable MAU

Udumft:
Resolved [mapbox/navigation-sdks/issues/129](https://github.com/mapbox/navigation-sdks/issues/129).
This PR reverts [changes](https://github.com/mapbox/mapbox-navigation-ios/pull/2375), that reverted [MAU biliing implementation](https://github.com/mapbox/mapbox-navigation-ios/pull/2151). Also updates `MapboxAccounts` dependency to version `2.3.0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master
 
+### MAU Billing
+
+* Enabled MAU billing by default, so that SDKs usage of Mapbox APIs is [billed](https://www.mapbox.com/pricing/) together based on [monthly active users](https://docs.mapbox.com/help/glossary/monthly-active-users/) rather than individually by HTTP request. If you prefer to still use request-based billing, set the `MBXNavigationBillingMethod` key in Info.plist to `request` ([#2405](https://github.com/mapbox/mapbox-navigation-ios/pull/2405).
+
 ### Offline navigation
 
 * Fixed a crash that occurred after setting `RouteOptions.shapeFormat` to `RouteShapeFormat.geoJSON`. ([valhalla/valhalla#1867](https://github.com/valhalla/valhalla/pull/1867))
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -17,7 +17,7 @@
...
-* Added a new dependency on MapboxAccounts to prepare for upcoming improvements to how Mapbox [bills](https://www.mapbox.com/pricing/) this SDK’s usage of Mapbox APIs. If you use Carthage to install this SDK, remember to add MapboxAccounts.framework to the “Frameworks, Libraries, and Embedded Content” section and the input and output file lists of Carthage’s Run Script build phase. ([#2151](https://github.com/mapbox/mapbox-navigation-ios/pull/2151))
+* Added a new dependency on MapboxAccounts so that this SDK’s usage of Mapbox APIs is [billed](https://www.mapbox.com/pricing/) together based on [monthly active users](https://docs.mapbox.com/help/glossary/monthly-active-users/) rather than individually by HTTP request. If you use Carthage to install this SDK, remember to add MapboxAccounts.framework to the “Frameworks, Libraries, and Embedded Content” section and the input and output file lists of Carthage’s Run Script build phase. ([#2151](https://github.com/mapbox/mapbox-navigation-ios/pull/2151))
```
1ec5, line 20: This section of the changelog is for v0.40.0, which has already been released. Let’s add a new item that discusses the change to MAU-based billing independently of the changes to the installation instructions, which aren’t new to v1.0.0.

Example/AppDelegate.swift
```diff
@@ -16,7 +16,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if isRunningTests() {
             window!.rootViewController = UIViewController()
         }
-        #if false
+        #if DEBUG
```
1ec5, line 19: This code was taken out so that the example applications would build on their own. Can we leave it disabled for now, or is there some other way to enable the UI tests without making the example applications dependent on the testing targets (and all their dependency libraries)?

Udumft: `Example` App does build on it's own. It is not dependent on any testing target. In fact, it's reverse: `MapboxNavigation` test target needs to build `Example` App and run it with special testing command line argument.

UI test is needed due to feature of Maps functionality which exposes it's SKU token only when requests for tiles or anything else is made. At this point, I don't see any other way to check SKU tokens other than building a special UI test App target.

Issue #1502, opened 2018-06-14
Combine NavigationViewController and RouteMapViewController


bsudekum:
I can't think of any reasons why we have both [NavigationViewController.swift](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/NavigationViewController.swift) and [RouteMapViewController.swift](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigation/RouteMapViewController.swift). I think we should refactor the two files into a single view controller with a few goals in mind:

* This file should be used as a reference on how different pieces can be combined to create a custom navigation view.
* We should think more about making more functions public that are useful and require inside knowledge of the project to do correct. Example, [`recenterMap()`](https://github.com/mapbox/mapbox-navigation-ios/blob/a9d45a4b79eb5058bc8c55250ad1d319af04028f/MapboxNavigation/NavigationMapView.swift#L1057-L1060) function added recently.
* There is a lot going on between the two files. This will be a good opportunity to do some house cleaning.

/cc @mapbox/navigation-ios

1ec5:
For context, the difference between NavigationViewController and RouteMapViewController was originally that:

* NavigationViewController was responsible for pushing other view controllers that weren’t related to a map, such as the step table view controller or feedback view controller.
* In time, NavigationViewController would sit at the same level as a preview view controller (#808), which might also embed a route map view controller.

Things have certainly changed since the original design, especially now that much of the layout has moved into views.

JThramer:
Resurrecting this as a 1.0 release blocker. /cc @1ec5

1ec5:
#2548 tracks making the code to populate the current road name label more reusable by moving it outside of RouteMapViewController.

Udumft:
Resolved in mapbox/mapbox-navigation-ios/pull/2902

PR #2410, opened 2020-06-22
Add location manager for free drive

### LINKED ISSUE #1627: Make route snapping logic available for external developers?
In my context we’re using the `MapboxCoreNavigation.RouteController` for turn-by-turn navigation directly. Our user interface is so customised, that we're unable to use the default mapbox navigation user-interface.

### Intention

I've seen that there is a lot of utility logic in the `CLLocation.swift`extensionn ([link](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/CLLocation.swift)) regarding snapping a potential user location marker to the actual route, which is performed during turn-ty-turn navigation.

### Problem
However, nearly all methods in this extension are not visible outside the mapbox package.

### Question

Could these methods be made `public`, so that the logic could be used by developers using the core navigation library?

taken from our `Cartfile.resolved`
```
github "mapbox/MapboxDirections.swift" "v0.21.0"
github "mapbox/mapbox-navigation-ios" "v0.18.1"
```

### LINKED ISSUE #2456: Indicate current road name during free-driving
The navigation SDK should indicate the name of the road the user is currently on even during “free-driving” outside of turn-by-turn navigation, so that the developer can stick this information in WayNameLabel/WayNameView.

MapboxNavigationNative provides this information via the `MBNavigationStatus.roadName` property. One possibility is to expose this information as part of the location manager being added in #2410.

/ref #2259 #590

IodaMikeMapbox:
The pr adds implementation of `NavigationLocationManager` which processes raw gps values with the `Navigator` in order to map them to the road graph.

Fixes #1627, fixes #2456.

/cc @mapbox/navigation-ios


1ec5:
Tests are failing:

```
Test Case '-[MapboxCoreNavigationTests.FreeDriveLocationManagerTests testFreeDrive]' started.
Assertion failed: file /Users/distiller/project/MapboxCoreNavigation/FreeDriveLocationManager.swift, line 11
2020-08-04 23:23:51.761490+0000 Example[81397:133490] Assertion failed: file /Users/distiller/project/MapboxCoreNavigation/FreeDriveLocationManager.swift, line 11
2020-08-04 23:24:25.657 xcodebuild[80915:127530]  IDETestOperationsObserverDebug: Writing diagnostic log for test session to:
/Users/distiller/Library/Developer/Xcode/DerivedData/MapboxNavigation-gtrukiuwvhbmctfcsthejciepfao/Logs/Test/Run-MapboxCoreNavigation-2020.08.04_23-22-09-+0000.xcresult/Staging/3_Test/Diagnostics/MapboxCoreNavigationTests-596D4DFA-E130-4647-AD32-1447D4864C95/MapboxCoreNavigationTests-755684E2-2AC0-4DB7-8ADA-0247EA8FAA39/Session-MapboxCoreNavigationTests-2020-08-04_232425-w9SQaV.log
2020-08-04 23:24:25.657 xcodebuild[80915:127462] [MT] IDETestOperationsObserverDebug: (AE120DD5-109A-4E2E-9610-E120ACF5258A) Beginning test session MapboxCoreNavigationTests-AE120DD5-109A-4E2E-9610-E120ACF5258A at 2020-08-04 23:24:25.657 with Xcode 11E503a on target <DVTiPhoneSimulator: 0x7fa43f0961b0> {
		SimDevice: iPhone 8 Plus (AC7EC09C-8E14-4B40-ACC1-CE90A75BFBC4, iOS 12.2, Booted)
} (12.2 (16E226))
2020-08-04 23:24:25.661 xcodebuild[80915:127462] [MT] IDETestOperationsObserverDebug: (AE120DD5-109A-4E2E-9610-E120ACF5258A) Finished requesting crash reports. Continuing with testing.
2020-08-04 23:24:26.029388+0000 Example[81408:134457] Launching with XCTest injected. Preparing to run tests.
2020-08-04 23:24:26.172369+0000 Example[81408:134457] Waiting to run tests until the app finishes launching.
2020-08-04 23:24:26.219832+0000 Example[81408:134457] libMobileGestalt MobileGestalt.c:890: MGIsDeviceOneOfType is not supported on this platform.

Restarting after unexpected exit, crash, or test timeout in FreeDriveLocationManagerTests/testFreeDrive(); summary will include totals from previous launches.
```

MapboxCoreNavigation/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,196 @@
...
+    let host = "https://api-routing-tiles-here-staging.tilestream.net"
+    let publicToken = ""
+    assert(publicToken != "")
+    let endpointConfig = TileEndpointConfiguration(
+        host: host,
+        version: tilesVersion,
+        token: publicToken,
+        userAgent: "MapboxNavigationNative",
+        navigatorVersion: "14.1.0", skuTokenSource: Accounts()
+    )
```
1ec5, line 18: This all needs to be cleaned up to not hard-code a staging endpoint, outdated navigation-native version, and tile version.

MapboxCoreNavigation/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    var peer: MBXPeerWrapper?
+}
+
+protocol FreeDriveDebugInfoListener: AnyObject {
+    func didGet(location: CLLocation, with matches: [MapMatch], for rawLocation: CLLocation)
+}
+
+class FreeDriveDebugInfoView: UIView, FreeDriveDebugInfoListener, UITableViewDataSource, UITableViewDelegate {
```
1ec5, line 135: If the intention is to ship this debugging view with the SDK, it needs to be moved into MapboxNavigation. MapboxCoreNavigation is for non-UI components only. There should be no dependency on UIKit.

MapboxCoreNavigation/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,196 @@
...
+
+        public func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
+            delegate?.locationManager?(manager, didFailWithError: error)
+        }
+    }
+}
+
+private class Accounts: SkuTokenSource {
+    init() {
+        MBXAccounts.activateSKUID(.navigationUser)
```
1ec5, line 122: This should be unnecessary due to code in `+[MBXAccounts(CoreNavigationAdditions) load]` that automatically runs as soon as MapboxCoreNavigation.framework loads:

https://github.com/mapbox/mapbox-navigation-ios/blob/0fb35932f62387b96bac1f0653cf8766c3f6d3bb/MapboxCoreNavigation/MBXAccounts+CoreNavigationAdditions.m#L11

MapboxNavigation/FreeDriveDebugger.swift
```diff
@@ -0,0 +1,132 @@
+import UIKit
+import Mapbox
+import MapboxCoreNavigation
+
+class CLLToMGLConverterLocationManager: NSObject, MGLLocationManager, CLLocationManagerDelegate {
```
1ec5, line 5: Can you describe the purpose of this class? It’s a private class, though some of its members are marked public. It’s also in a file called FreeDriveDebugger.swift, which seems temporary.

MapboxCoreNavigation/FreeDriveLocationManager.swift
IodaMikeMapbox: sure, that's why it is a draft

MapboxNavigation/FreeDriveDebugger.swift
IodaMikeMapbox: Initially I used it from the Example app, that's why it needed to be public, for now there is no reason for the methods to be public.

The class by itself is just a converter from one location manager protocol to another. The problem is that there are `NavigationLocationManager` and `MGLLocationManager` protocols which can't be implemented by one class at the same time. `FreeDriveLocationManager` conforms to `NavigationLocationManager` (and therefore to `CLLocationManager`). To show mapped locations on the map we need to provide MGLLocationManager, the `CLLToMGLConverterLocationManager` converts `CLLocationManager` to `MGLLocationManager`

MapboxCoreNavigation/FreeDriveLocationManager.swift
IodaMikeMapbox: The class will be removed, the idea was to use the SkuTokenProvider from #2533

MapboxCoreNavigation/FreeDriveLocationManager.swift
IodaMikeMapbox: moved to `MapboxNavigation`

IodaMikeMapbox:
The pr uses @dersim-davaod's commit from temp-route-tiles-logic branch, as I see the work from that branch was utilised in #2533

dersim-davaod:
@IodaMikeMapbox thanks for notifying. it will be slightly changed in #2533, but that's fine. If the current changes will be merged first, that's also fine, I'll handle a proper rebase.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -8,7 +8,7 @@ import Turf
 class ArrowFillPolyline: MGLPolylineFeature {}
 class ArrowStrokePolyline: ArrowFillPolyline {}
 
-class RouteMapViewController: UIViewController {
+class RouteMapViewController_v2: UIViewController {
```
1ec5, line 11: This rename looks like work towards #1502. Maybe we should keep it out of scope for this PR?

MapboxCoreNavigation/MBXPeerWrapper.h
```diff
@@ -0,0 +1,5 @@
+#import <Foundation/Foundation.h>
+
+@interface MBXPeerWrapper
+
+@end
```
1ec5, line 5: We might not even need to declare the interface; it might be enough to simply forward-declare the class:

```suggestion
@class MBXPeerWrapper;
```

MapboxCoreNavigation/RouteTiles.swift
```diff
@@ -0,0 +1,88 @@
...
+    
+    // MARK: Private properties
+
+    private let directionsCredentials: DirectionsCredentials
+
+    /// URL to request the availalbe version of route tiles format.
+    private var requestVersionsURL: URL? {
+        let requestVersionsURLString = directionsCredentials.host.absoluteString +
+                                       "/route-tiles/v1/versions" +
+                                       "?access_token=" + (directionsCredentials.accessToken ?? "")
```
1ec5, line 50: We should create URLs using URL or URLComponents initializers rather than string concatenation or interpolation whenever possible.

MapboxCoreNavigation/URLSession.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
+import MapboxDirections
+
+extension URLSession {
+    /// The user agent string for any HTTP requests performed directly within this library.
+    static let userAgent: String = {
```
1ec5, line 6: This method was copied wholesale from MapboxNavigation. We should either remove the version in MapboxNavigation in favor of this one or have the version in MapboxNavigation call this version and insert the MapboxNavigation bundle version.

MapboxCoreNavigation/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,134 @@
+import MapboxNavigationNative
+import CoreLocation
+import MapboxDirections
+import MapboxAccounts
+
+public protocol FreeDriveDebugInfoListener: AnyObject {
```
1ec5, line 6: Rename this protocol to FreeDrivingDebuggingInfoDelegate or somesuch. Listeners are called observers or delegates in Swift.

MapboxCoreNavigation/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,134 @@
...
+        let navigator = Navigator(profile: settingsProfile, config: NavigatorConfig() , customConfig: "")
+        let host = Directions.shared.credentials.host.absoluteString
+        let publicToken = Directions.shared.credentials.accessToken ?? ""
+        assert(publicToken != "")
+        let endpointConfig = TileEndpointConfiguration(
+            host: host,
+            version: tilesVersion,
+            token: publicToken,
+            userAgent: "",
+            navigatorVersion: "", skuTokenSource: SkuTokenProvider(with: DirectionsCredentials())
```
1ec5, line 53: Per https://github.com/mapbox/mapbox-navigation-android/pull/2326#issuecomment-567844536, it may actually be desirable to pass in an empty string, rather than the Core Navigator bundle version, to avoid side effects. (We really don’t want the navigator to try to use cURL, which doesn’t exist on iOS.)

MapboxCoreNavigation/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,134 @@
...
+    private var proxyDelegate: ProxyDelegate?
+
+    public required init(navigator: Navigator? = nil) {
+        proxyDelegate = ProxyDelegate()
+        super.init()
+
+        if let navigator = navigator {
+            proxyDelegate?.navNative = navigator
+        } else {
+            let tilesVersion = RouteTilesVersion(with: Directions.shared.credentials)
```
1ec5, line 29: The developer needs to be able to specify a Directions object other than the shared Directions object, in case they need to specify an access token programmatically.

MapboxCoreNavigation/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,134 @@
...
+            proxyDelegate?.debugDelegate
+        }
+        set {
+            proxyDelegate?.debugDelegate = newValue
+        }
+    }
+
+    private var proxyDelegate: ProxyDelegate?
+
+    public required init(navigator: Navigator? = nil) {
```
1ec5, line 22: MapboxNavigationNative types should not be part of the public API. The navigator is an implementation detail, and its interface is written in Objective-C.

MapboxCoreNavigationTests/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,92 @@
...
+class FreeDriveLocationManagerTests: XCTestCase {
+    class LocationsObserver: NSObject, CLLocationManagerDelegate {
+        func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+            guard let location = locations.first else { return }
+            print("Got location: \(location.coordinate.latitude), \(location.coordinate.longitude)")
+        }
+    }
+
+    func testFreeDrive() {
+        let locationManager = FreeDriveLocationManager()
```
1ec5, line 59: This test doesn’t really exercise free driving, other than the ability to set a custom location and get it back out of the location manager. `RouteTilesVersion.directionsCredentials` has a nil `host`, making `RouteTilesVersion.requestVersionsURL` return nil because of an invalid URL, causing `RouteTilesVersion.getAvailableVersions(completionHandlerQueue:completion:)` to bail early before ever kicking off a network request. Not that we’d want a unit test to kick off a network request, but because RouteTilesVersion is unable to return a version, `FreeDriveLocationManager.createNavigator(withTilesVersion:)` never configures the navigator.

MapboxCoreNavigation/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,156 @@
...
+    override public var delegate: CLLocationManagerDelegate? {
+        get {
+            proxyDelegate?.delegate
+        }
+        set {
+            proxyDelegate?.delegate = newValue
+        }
+    }
+
+    private class ProxyDelegate: NSObject, CLLocationManagerDelegate {
```
1ec5, line 103: This is an unusual pattern in Swift, sort of reminiscent of PImpl but also redundant in its conformance to CLLocationManagerDelegate. I think the reason for this approach was the need to avoid making FreeDriveLocationManager both inherit from CLLocationManager and conform to CLLocationManagerDelegate. The inverse approach would be cleaner: FreeDriveLocationManager should continue to conform to CLLocationManagerDelegate directly but _contain_ an object that inherits from CLLocationManager.

MapboxNavigation/FreeDriveDebugInfoView.swift
```diff
@@ -0,0 +1,67 @@
+import MapboxCoreNavigation
+import MapboxNavigationNative
+import UIKit
+
+class FreeDriveDebugInfoView: UIView, FreeDriveDebugInfoListener, UITableViewDataSource, UITableViewDelegate {
```
1ec5, line 5: This class probably belongs in the example application target.

MapboxCoreNavigationTests/FreeDriveLocationManager.swift
1ec5: For now, I’ve changed the test to manually force the location manager to configure the navigator, but that’s quite contrived and won’t catch any failure modes.

MapboxNavigation/FreeDriveDebugger.swift
1ec5: I’ve made the class public, since the whole point of this feature is that applications can integrate a replacement for CLLocationManager into a map view at the application level. The file and class still need to be renamed.

MapboxCoreNavigationTests/FreeDriveLocationManager.swift
```diff
@@ -0,0 +1,102 @@
...
+        let locationUpdateExpectation: XCTestExpectation
+        
+        init(road: Road, locationUpdateExpectation: XCTestExpectation) {
+            self.road = road
+            self.locationUpdateExpectation = locationUpdateExpectation
+        }
+        
+        func locationManager(_ manager: FreeDriveLocationManager, didUpdateLocation location: CLLocation, rawLocation: CLLocation) {
+            print("Got location: \(rawLocation.coordinate.latitude), \(rawLocation.coordinate.longitude) → \(location.coordinate.latitude), \(location.coordinate.longitude)")
+            print("Value: \(road.proximity(of: location.coordinate)) < \(road.proximity(of: rawLocation.coordinate))")
```
1ec5, line 50: Once we determine an acceptable range of values, we should turn these print statements into assertions.

MapboxCoreNavigationTests/FreeDriveLocationManager.swift
1ec5: This is apparently the cause of the test failure on CI, where we haven’t configured a .mapbox file that the build system would read by default.

1ec5:
To stress-test this feature, I set the iOS Simulator to the City Bicycle simulation:

Blaney Homestead

MapboxNavigation.xcodeproj/xcshareddata/xcschemes/MapboxCoreNavigation.xcscheme
```diff
@@ -60,6 +60,9 @@
                <Test
                   Identifier = "BridgingTests/testOfflineRouting">
                </Test>
+               <Test
+                  Identifier = "FreeDriveLocationManagerTests/testFreeDrive()">
+               </Test>
```
1ec5: I’ve disabled the free-driving integration test, because we don’t currently have a way to mock out the routing tile endpoint.

1ec5:
I’ve renamed various things to avoid the term “free driving” in the public API. MapboxCoreNavigation and MapboxNavigation both have their own free driving classes; the distinction between them is easier to make if we call one a location data source, borrowing from RouteDataSource terminology, and the other a location manager, for consistency with MGLLocationManager. This naming convention also avoids the awkwardness of beginning class names with a noun phrase that could easily be misread as a verb phrase. Documentation can continue to call the overall feature utilities for implementing a “free drive mode”, just as it speaks of “Bring Your Own Route” as a technique rather than a specific API.

Example/ViewController.swift
```diff
@@ -541,3 +573,43 @@ extension ViewController: VisualInstructionDelegate {
         return presented
     }
 }
+
+// MARK: Free driving
+extension ViewController {
+    func trackLocations(mapView: NavigationMapView) {
+        let dataSource = PassiveLocationDataSource(directions: Settings.directions)
+        let locationManager = PassiveLocationManager(dataSource: dataSource)
+        mapView.locationManager = locationManager
```
1ec5, line 582: This is the bare minimum necessary to integrate free-driving localization into an application’s pre-navigation map view. The rest of the changes in this file are related to a nifty but inessential feature that plots the user’s raw and idealized path on the map.

MapboxCoreNavigation/URLSession.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+import MapboxDirections
+
+extension URLSession {
+    /**
+     :nodoc:
+     
+     The user agent string for any HTTP requests performed directly within MapboxCoreNavigation or MapboxNavigation.
+     */
+    public static let userAgent: String = {
```
MaximAlien, line 10: Is there any specific reason we use custom `User-Agent`? (`NSURLSession`sets `User-Agent` by default). I can also see that `Directions.swift` uses pretty similar approach when creating `userAgent` so might make sense to have some common way of creating it.

MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
```diff
@@ -0,0 +1,103 @@
...
+        
+        let road = Road(from: CLLocationCoordinate2D(latitude: 47.207966, longitude: 9.527012), to: CLLocationCoordinate2D(latitude: 47.209518, longitude: 9.522167))
+        let delegate = Delegate(road: road, locationUpdateExpectation: locationUpdateExpectation)
+        locationManager.delegate = delegate
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(10)) {
+            locationManager.updateLocation(CLLocation(latitude: 47.208674, longitude: 9.524650))
+        }
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(25)) {
```
MaximAlien, line 86: Probably makes sense to review this test in future since it's time consuming.

MapboxCoreNavigationTests/TileRoutingVersionTests.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    func testGetAvailableVersionsReturnsArrayOfAvailableVersions () {
+        // Given
+        let expectation = XCTestExpectation(description: "")
+        let routeTilesVersion = RouteTilesVersion(with: DirectionsCredentials())
+        
+        // When
+        routeTilesVersion.getAvailableVersions { availableVersions in
+            print(availableVersions)
+            expectation.fulfill()
```
MaximAlien, line 50: There are no actual checks performed except `expectation` usage. Should we verify available versions?

MapboxCoreNavigation/MBXPeerWrapper.h
1ec5: Actually, looks like that doesn’t work. The forward declaration doesn’t satisfy the Swift compiler, even if it appears in the umbrella header.

MapboxCoreNavigation/RouteTiles.swift
```diff
@@ -0,0 +1,94 @@
...
+    
+    // MARK: Private methods
+    
+    /**
+     Gets available tiles versions.
+     
+     - Parameter completionHandlerQueue:
+     - Parameter completion:
+     */
+    func getAvailableVersions(completionHandlerQueue: DispatchQueue = DispatchQueue.main,
```
MaximAlien, line 69: Makes sense to add parameter documentation here, also looks like `getAvailableVersions(completionHandlerQueue:completion:)` is not really private method.

MapboxCoreNavigationTests/TileRoutingVersionTests.swift
```diff
@@ -0,0 +1,56 @@
...
+        let expectedCurrentVersion = "2020_07_03-03_00_00"
+        
+        // When
+        let currentVersion = routeTilesVersion.currentVersion
+        
+        // Then
+        XCTAssertEqual(currentVersion, expectedCurrentVersion)
+    }
+    
+    func testCurrentVersionStoresValuePersistently () {
```
MaximAlien, line 18: Not sure what's the idea of `testCurrentVersionStoresValuePersistently` and `testCurrentVersionReadsValuePersistently`. Makes sense to revise in future.

MapboxCoreNavigation/URLSession.swift
1ec5: In general, we should always customize the user agent string to provide more context in the server logs than the application and Core Foundation versions, which is what URLSession provides by default. This particular user agent string is specific to requests made by MapboxCoreNavigation or MapboxNavigation.

MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -0,0 +1,169 @@
...
+     
+     - parameter directions: The directions service that allows the location data source to access road network data. If this argument is omitted, the shared `Directions` object is used.
+     
+     - postcondition: Call `startUpdatingLocation(completionHandler:)` afterwards to begin receiving location updates.
+     */
+    public required init(directions: Directions = Directions.shared, systemLocationManager: NavigationLocationManager? = nil) {
+        self.directions = directions
+        
+        let settingsProfile = SettingsProfile(application: ProfileApplication.kMobile, platform: ProfilePlatform.KIOS)
+        self.navigator = Navigator(profile: settingsProfile, config: NavigatorConfig() , customConfig: "")
```
MaximAlien, line 25: I remember that @SiarheiFedartsou was mentioning that it can be beneficial to use one global `MBNavigator` and feed it with data whenever it's available. Since we now have 4 (?) places where `MBNavigator` is created it might make sense to consider one global instance approach.

MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
1ec5: Yep, in fact, I disabled this test because it requires network access and a valid Mapbox access token to function at all. We’ll figure out a better integration test as tail work. I just rewrote history in this PR to undo the change that deleted the offline routing tests; I think we could possibly leverage that test’s test fixture to avoid reinventing the wheel.

MapboxCoreNavigation/PassiveLocationDataSource.swift
1ec5: Right, for better or worse, the navigator is intended to be used as something of a singleton if I remember correctly.

#2509 will remove two of these initializations, leaving just the ones in RouteController and PassiveLocationDataSource, so I didn’t bother to centralize the logic just yet.

The long-term vision is indeed to unify PassiveLocationDataSource and RouteController, so that the latter is just a special case of setting a route on the former and keeping track of route progress. For now, we would probably recommend that the developer stop location updates on PassiveLocationDataSource when beginning turn-by-turn navigation with RouteController. This happens to be the case in the example application, because it demonstrates temporarily taking the main map view out of the view hierarchy as a performance optimization.

MapboxCoreNavigationTests/TileRoutingVersionTests.swift
1ec5: I think these tests were a work in progress as part of #2533, but they got folded into this PR at some point. Like the PassiveLocationManagerTests, this class depends on network access, but fortunately it’s straightforward to mock the response in this case.

MapboxCoreNavigationTests/TileRoutingVersionTests.swift
1ec5: I’m going to remove these tests, as well as the `currentVersion` property that they would’ve exercised. I don’t see anything making use of it, and the navigator is already able to persist the version for its lifetime.

1ec5:
Based on the feedback in https://github.com/mapbox/mapbox-navigation-ios/pull/2410#pullrequestreview-468930276, I removed the RouteTilesVersion implementation outright in favor of the preexisting routing tiles API support in MapboxDirections, which is better tested and slightly more strongly typed.

MapboxNavigation.xcodeproj/xcshareddata/xcschemes/MapboxCoreNavigation.xcscheme
1ec5: #2546

MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
1ec5: #2546

PR #2412, opened 2020-06-22
Update nav-native to 9.2.1

dersim-davaod:
This is a mirror of accidentally merged PR #2404.

The base branch was reset and a new PR (the current one) is created.

PR #2416, opened 2020-06-23
Update nav-native to 12.0.3

dersim-davaod:
Upgraded to MapboxNavigationNative v12.0.3 as a followup to #2412.

/cc @mapbox/navigation-ios


MapboxCoreNavigation.podspec
```diff
@@ -40,12 +40,17 @@ Pod::Spec.new do |s|
...
   s.dependency "MapboxDirections", "~> 0.32.0"
   s.dependency "MapboxMobileEvents", "~> 0.10.2"        # Always pin to a patch release if pre-1.0
   s.dependency "Turf", "~> 0.5.0"                       # Always pin to a patch release if pre-1.0
 
   s.swift_version = "5.0"
 
+  # MapboxNavigationNative is built for arm64 and x86_64 architectures (i386 and armv7 are not present), so only 64-bit architecture is checked.
+  s.pod_target_xcconfig = {
+    'ARCHS[sdk=iphonesimulator*]' => '$(ARCHS_STANDARD_64_BIT)'
+  }
```
1ec5, line 54: mapbox/mapbox-navigation-ios#2422 implements a similar workaround for Carthage.

Udumft: Should this bit be in #2422 then?

1ec5:
These changes have been incorporated into #2417, which upgrades all the way to v14.1.3.

PR #2417, opened 2020-06-23
Upgrade to MapboxNavigationNative 14.1.5

MaximAlien:
Due to availability changes in CLLocation API **_had to drop support of Xcode lower than 11.4.1_**. Please let me know if that's okay. Other than that there is workaround for architectures handling in nav. native when running pod lint (added in branch for update to 12.0.3).

/ref #2416


avi-c:
As part of this update we will need to add retrieving and providing a SKU Token from the SDK to Nav Native.

MapboxCoreNavigationTests/CLLocationTests.swift
```diff
@@ -11,42 +11,74 @@ class CLLocationTests: XCTestCase {
...
+        let location = CLLocation(coordinate: coordinate,
+                                  altitude: 10,
                                   horizontalAccuracy: 40,
                                   verticalAccuracy: 50,
-                                  course: 180, speed: 18,
+                                  course: 180,
+                                  speed: 18,
                                   timestamp: Date())
         
         XCTAssertEqual(location.timestamp, location.shifted(to: location.timestamp).timestamp)
```
dersim-davaod, line 56: Does that make sense to use different timestamp for the test? Otherwise looks like the test does not do the job.

MapboxCoreNavigationTests/CLLocationTests.swift
```diff
@@ -57,27 +89,12 @@ class CLLocationTests: XCTestCase {
...
         
-        XCTAssertEqual(location.coordinate.latitude, 1)
-        XCTAssertEqual(location.coordinate.longitude, 2)
-        XCTAssertEqual(location.timestamp, now)
-        XCTAssertEqual(location.speed, -1)
-        XCTAssertEqual(location.course, -1)
-        XCTAssertEqual(location.altitude, -1)
-        XCTAssertEqual(location.horizontalAccuracy, -1)
+        if #available(iOS 13.4, *) {
+            XCTAssertEqual(fixLocation.bearingAccuracy, 2)
```
dersim-davaod, line 95: nit:
Extract values we check into const within "given" section:
```suggestion
            XCTAssertEqual(fixLocation.bearingAccuracy, expectedBearingAccuracy)
```
?

MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
```diff
@@ -53,9 +53,9 @@ class NavigationEventsManagerTests: XCTestCase {
         
         XCTAssertEqual(events.count, 3, "There should be one depart, one reroute, and one arrive event.")
         
-        let departEvent = events.filter { $0.event == MMEEventTypeNavigationDepart }.first!
-        let rerouteEvent = events.filter { $0.event == MMEEventTypeNavigationReroute }.first!
-        let arriveEvent = events.filter { $0.event == MMEEventTypeNavigationArrive }.first!
+        guard let departEvent = events.filter({ $0.event == MMEEventTypeNavigationDepart }).first else { XCTFail(); return }
```
dersim-davaod, line 56: nit: add description for `XCTFail` calls?

MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -122,54 +122,78 @@ class NavigationServiceTests: XCTestCase {
...
+        XCTAssertEqual(navigation.router.location!.coordinate.longitude, lastCoordinate.longitude, accuracy: 0.005, "Longitudes should be almost equal")
 
         let firstLocationOnNextStepWithSpeed = CLLocation(coordinate: firstCoordinateOnUpcomingStep, altitude: 0, horizontalAccuracy: 10, verticalAccuracy: 10, course: 10, speed: 5, timestamp: Date())
         navigation.locationManager!(navigation.locationManager, didUpdateLocations: [firstLocationOnNextStepWithSpeed])
 
-        XCTAssertEqual(navigation.router.location!.coordinate.latitude, firstCoordinateOnUpcomingStep.latitude, accuracy: 0.0005, "User is snapped to upcoming step when moving")
-        XCTAssertEqual(navigation.router.location!.coordinate.longitude, firstCoordinateOnUpcomingStep.longitude, accuracy: 0.0005, "User is snapped to upcoming step when moving")
+        // User is snapped to upcoming step when moving
+        XCTAssertEqual(navigation.router.location!.coordinate.latitude, firstCoordinateOnUpcomingStep.latitude, accuracy: 0.005, "Latitudes should be almost equal")
```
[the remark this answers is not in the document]
dersim-davaod: The other question that comes from the question above: should we make accuracy threshold lower?

.circleci/config.yml
```diff
@@ -196,29 +196,35 @@ workflows:
...
+          xcode: "11.4.1"
+          iOS: "13.4.1"
           device: "iPhone 8 Plus"
       - build-job:
-          name: "Xcode_10.3_iOS_12.1"
-          xcode: "10.3.0"
-          iOS: "12.1"
+          name: "Xcode_11.4.1_iOS_12.2"
+          xcode: "11.4.1"
+          iOS: "12.2"
```
1ec5, line 211: Various test fixtures are conditionalized by iOS version, based on their file names. So we’ll need to rename or delete and rerecord a slew of test fixtures, similar to what took place in #2427.

.circleci/config.yml
```diff
@@ -196,29 +196,35 @@ workflows:
...
           codecoverage: true
       - build-job:
-          name: "Xcode_10.2_iOS_10.3.1"
-          xcode: "10.2.1"
+          name: "Xcode_11.4.1_iOS_10.3.1"
+          xcode: "11.4.1"
```
1ec5, line 215: We need to continue to test against iOS 10._x_ somehow, because the SDK still supports iOS 10._x_. On the other hand, it’s fine to drop Xcode 10 support, but we should update the readme to make that clear.

MapboxCoreNavigationTests/NavigationServiceTests.swift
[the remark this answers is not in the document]
1ec5: A difference of 0.005° longitude is more than 100&nbsp;meters off at the latitudes in this test fixture, whereas 0.0005° is within a dozen meters or so. So I think the accuracy threshold does need to be lower, maybe nearly as low as it was previously.

MapboxCoreNavigationTests/CLLocationTests.swift
```diff
@@ -11,42 +11,74 @@ class CLLocationTests: XCTestCase {
...
+                                  course: -1,
+                                  speed: -1,
+                                  timestamp: now)
+        
+        if #available(iOS 13.4, *) {
+            location = CLLocation(coordinate: coordinate,
                                   altitude: -1,
                                   horizontalAccuracy: -1,
-                                  verticalAccuracy: 0,
+                                  verticalAccuracy: 1,
```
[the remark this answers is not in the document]
1ec5: I’d be surprised if anything in MapboxCoreNavigation or MapboxNavigationNative cares about the vertical accuracy field at this point.

MapboxCoreNavigationTests/NavigationServiceTests.swift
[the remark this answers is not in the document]
MaximAlien: In previous versions of nav. native right after starting navigation `initialized` `routeState` was returned by `RouteController`. This meant that `rawLocation` was used, which is last location, as a result no accuracy checks were needed at that time. Example with `routeState` values with 3 locations:
```
1. Current route state: initialized
2. Current route state: initialized
3. Current route state: tracking
```
As per discussion with @SiarheiFedartsou in nav. native 14.1.0 active guidance logic was changed and now `initialized` state is only returned in `setRouteForRouteResponse(_:route:leg:)` (I've checked to make sure and that's correct). This means that now route state has such sequence:
```
1. Current route state: tracking
2. Current route state: tracking
3. Current route state: tracking
```
This means that instead of `rawLocation` we'll be using `snappedLocation` because of such logic in `RouteController`:
```swift
var snappedLocation: CLLocation? {
    let status = navigator.getStatusForTimestamp(Date())
    guard status.routeState == .tracking || status.routeState == .complete else {
        return nil
    }
    
    return CLLocation(status.location)
}
...
    return snappedLocation ?? rawLocation
}
```

Since `snappedLocation` is returned from `MBNavigator` it explains the fact that last location and `snappedLocation` are not the same (that is the reason accuracy was introduced). I can see two solutions for now:
1. Leaving accuracy argument, but slightly change it to contain more 'realistic' value, e.g. distance between `snappedLocation` and last location?
2. Changing logic in `RouteController` to apply new flow in nav. native. E.g. somehow detecting that we're in `initialized` state even though we're currently already in `tracking` as per status returned from nav. native. 

I might still miss some pitfalls, so if you have any ideas please let me know.

MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -256,7 +280,10 @@ class NavigationServiceTests: XCTestCase {
 
         let willRerouteNotificationExpectation = expectation(forNotification: .routeControllerWillReroute, object: router) { (notification) -> Bool in
             let fromLocation = notification.userInfo![RouteController.NotificationUserInfoKey.locationKey] as? CLLocation
-            return fromLocation == testLocation
+
+            XCTAssertTrue(fromLocation == testLocation)
```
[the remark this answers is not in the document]
MaximAlien: I've added `XCTAssertTrue` instead for plain return value to see where test fails more easily. Previously it wasn't really obvious since in case if false is returned expectation would fail with timeout.

MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -266,7 +293,9 @@ class NavigationServiceTests: XCTestCase {
             let rawLocation = notification.userInfo![RouteController.NotificationUserInfoKey.rawLocationKey] as? CLLocation
             let _ = notification.userInfo![RouteController.NotificationUserInfoKey.routeProgressKey] as! RouteProgress
 
-            return location!.distance(from: rawLocation!) <= 0.0005
+            XCTAssertTrue(location!.distance(from: rawLocation!) <= 0.5)
```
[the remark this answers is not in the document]
MaximAlien: Still in discussion about the best way how to fix this, when finished will address this.

MapboxCoreNavigationTests/CLLocationTests.swift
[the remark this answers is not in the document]
MaximAlien: `verticalAccuracy`, `courseAccuracy`, `speedAccuracy` were just recently introduced. I've added values `1`, `2`, `3` respectively for those properties just for the sake of variety, to see that initializer assigns correct values.

MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
MaximAlien: There is a possibility that this test might slightly change, will check afterwards.

MapboxCoreNavigationTests/CLLocationTests.swift
MaximAlien: Good note, updated test.

MapboxCoreNavigation/FixLocation.swift
```diff
@@ -4,12 +4,25 @@ import MapboxNavigationNative
 
 extension FixLocation {
     convenience init(_ location: CLLocation) {
+        var speedAccuracy: NSNumber? = nil
+        if #available(iOS 10.0, *) {
```
Udumft, line 8: These check seems redundant. `MapboxCoreNavigation` minimum iOS version is 10.0

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -152,7 +152,12 @@ public class NavigationDirections: Directions {
             let tilePath = filePathURL.path
             let outputPath = outputDirectoryURL.path
             
-            let numberOfTiles = Navigator().unpackTiles(forPackedTilesPath: tilePath, outputDirectory: outputPath)
+            let navigator: Navigator = {
+                let settingsProfile = SettingsProfile(application: ProfileApplication.kMobile,
+                                                      platform: ProfilePlatform.KIOS)
+                return Navigator(profile: settingsProfile, customConfig: "")
+            }()
```
Udumft, line 159: Can we use `self.navigator` here instead of initializing a new one?

MapboxCoreNavigationTests/NavigationServiceTests.swift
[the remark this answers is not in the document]
dersim-davaod: Thanks for the comprehensive analysis!

Seems the first option is quicker, but it looks more like cutting the corner.
Should we spend time on trying to update the `RouteController`?

.circleci/config.yml
MaximAlien: I've added step to test against iOS `10.3.1`. Do you think that's enough? Also another note is that since properties like `courseAccuracy` were introduced only in iOS `13.4` minimum usable version of Xcode is `11.4.1`. Please let me know if that's okay.

MapboxCoreNavigation/FixLocation.swift
MaximAlien: Thanks, check was removed.

MapboxCoreNavigation/OfflineDirections.swift
MaximAlien: In this place navigator is used in static context, probably better to leave it as it is.

MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -110,29 +110,27 @@ class NavigationServiceTests: XCTestCase {
         }
     }
 
-    func testAdvancingToFutureStepAndNotRerouting() {
+    func testNotReroutingForAllSteps() {
```
MaximAlien, line 113: I see that when using nav. native `9.0.4` this test was working non deterministically. E.g. here check for advancing to future step is made only for steps 1 and 3 (why not for all?). When I check steps advancing for all steps I see that `stepIndex` is not increased for steps other than 1 and 3 (e.g. we remain on stepIndex 1 (instead of 2) for location with index 1 and on stepIndex 3 for locations 4 and 5). Explanation for this is that stepIndex is tied to value stored in nav. native class `MBNavigationStatus`. 
I've modified test to check whether locations in list are on route, I think there is no reliable way to check in what case `stepIndex` will be increased.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -387,7 +391,7 @@ extension RouteController: Router {
         }
         
         let status = status ?? navigator.getStatusForTimestamp(location.timestamp)
-        let offRoute = status.routeState == .offRoute
+        let offRoute = status.routeState == .offRoute || status.routeState == .invalid
```
MaximAlien, line 394: As per discussion with @SiarheiFedartsou in Slack, for now it's better to treat `invalid` state as `offRoute`.

MaximAlien:
`testDepartRerouteArrive()` is failing due to unexpected route status values returned by nav. native. Details can be found [here](https://github.com/mapbox/mapbox-navigation-native/issues/1330).

.circleci/config.yml
MaximAlien: Thanks, fixtures were updated. I've also decided to remove iOS 13.4.1 build job since there is 13.5 already (this will allow to speed up build time and prevent adding additional fixtures for 13.4.1).

MapboxCoreNavigationTests/CLLocationTests.swift
MaximAlien: Makes sense, updated.

MapboxCoreNavigationTests/NavigationServiceTests.swift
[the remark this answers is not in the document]
MaximAlien: I've updated tests to use original accuracy of `0.0005°` for coordinates and added distance check between coordinates (set to default accuracy of `1` meter). Had to update several tests to reflect current behavior of logic in nav. native `14.1.2`. Please check updated comments in changed tests.

Example/SettingsViewController.swift
```diff
@@ -95,8 +95,8 @@ class SettingsViewController: UITableViewController {
         if let selectedOfflineVersion = Settings.selectedOfflineVersion {
             let tilesURL = Bundle.mapboxCoreNavigation.suggestedTileURL(version: selectedOfflineVersion)
             
-            Settings.directions.configureRouter(tilesURL: tilesURL!) { [weak self] (numberOfTiles) in
-                let message = String.localizedStringWithFormat(NSLocalizedString("ROUTER_CONFIGURED_MSG", value: "Router configured with %ld tile(s).", comment: "Alert message when a router has been configured; 1 = number of map tiles"), numberOfTiles)
+            Settings.directions.configureRouter(tilesURL: tilesURL!) { [weak self] in
+                let message = String.localizedStringWithFormat(NSLocalizedString("ROUTER_CONFIGURED_MSG", value: "Router was configured.", comment: "Alert message when a router has been configured."))
```
1ec5, line 99: This would’ve required rerunning scripts/extract_localizable.sh.

1ec5, line 99: This backwards compatibility issue in MapboxNavigationNative was rolled back in v14.1.4, which I think we can upgrade to pretty smoothly from v14.1.3. I guess we might as well bring back the callback argument.

MaximAlien: Thanks, for now I've reverted changes back after update to `14.1.4`, had to slightly change tests again though.

MaximAlien: Done.

PR #2419, opened 2020-06-24
[Feedback][Part 1] Updates `FeedbackType` and introduces `FeedbackSubType`

nishant-karajgikar:
This diff updates the various categories in the `FeedbackType` enum and introduces infrastructural models ( viz. `FeedbackSubType`) to build a granular feedback workflow.

Simulator Screen Shot - iPhone 8 Plus - 2020-06-24 at 16 35 31


MapboxNavigation/FeedbackItem.swift
```diff
@@ -7,10 +7,61 @@ extension UIImage {
     }
 }
 
+
+public extension FeedbackType {
+
+    // TODO: Localize these strings
```
1ec5, line 13: [To localize these strings](https://github.com/mapbox/mapbox-navigation-ios/blob/master/CONTRIBUTING.md), wrap each one in an `NSLocalizedString(_:bundle:value:comment:)`, as with the original strings that are being deleted, then run scripts/extract_localizable.sh to update all the Localizable.strings files.

MapboxNavigation/Resources/Assets.xcassets/feedback/confusing_audio.imageset/Contents.json
```diff
@@ -0,0 +1,16 @@
...
+      "language-direction" : "left-to-right"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  },
+  "properties" : {
+    "preserves-vector-representation" : true
+  }
```
1ec5, line 15: In the asset catalog, change “Render As” to from “Original Image” to “Template Image” so that the icons will adopt the application’s tint color automatically.

MapboxCoreNavigation/Feedback.swift
```diff
@@ -4,87 +4,119 @@ import Foundation
...
+            return "route_quality"
+        case .illegalRoute:
+            return "illegal_route"
+        case .roadClosure:
+            return "road_closure"
+        }
+    }
+
+    /// Returns a corresponding list of `FeedbackSubType`s for a given `FeedbackType`
+    public var subtypes: [FeedbackSubType] {
```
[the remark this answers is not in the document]
nishant-karajgikar: Gotcha, i didn't realize that developers would override this `subtypes` field themselves. But it's a fair point. I'll move over to using the bespoke subtype enums for each top-level category.

nishant-karajgikar:
> Also delete the existing feedback icons from the asset catalog (except for “feedback”, which is for the feedback button).

Addressed this is in a separate PR #2421

MapboxNavigation/Resources/Assets.xcassets/feedback/confusing_audio.imageset/Contents.json
nishant-karajgikar: Hmm, If i do this the icons seem to disappear. Example:

Simulator Screen Shot - iPhone 8 Plus - 2020-06-24 at 22 20 39


(I've left `Route Preview` as `default` in order to show the difference)

MapboxNavigation/FeedbackItem.swift
nishant-karajgikar: Will be handling in this in a follow up PR.

MapboxCoreNavigation/Feedback.swift
```diff
@@ -4,87 +4,119 @@ import Foundation
...
-     */
-    case reportTraffic
-    
-    /**
-     Identifies the feedback as a general map issue.
-     */
-    case mapIssue
+
+    /// Indicates an incorrect visual artifact.
+    case incorrectVisual
```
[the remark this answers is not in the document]
nishant-karajgikar: I've amended the comment for now. Once this lands on a decision, I'll come back and update this.

nishant-karajgikar:
@1ec5 , I've made the changes we discussed in https://github.com/mapbox/mapbox-navigation-ios/pull/2419#discussion_r445195042 . Would love a review!

MapboxNavigation/Resources/Assets.xcassets/feedback/confusing_audio.imageset/Contents.json
1ec5: The tint color has always been set to clear, bizarrely enough:

https://github.com/mapbox/mapbox-navigation-ios/blob/fa19891a24b90596169046468947a79c7cd76b75/MapboxNavigation/FeedbackViewController.swift#L286

I think we can simply remove that line; the view controller will inherit the application’s tint color.

MapboxCoreNavigation/Feedback.swift
```diff
@@ -3,89 +3,48 @@ import Foundation
...
-    }
-
-
-
 }
 
-/// Enum used to define the many `FeedbackSubType`s that may be nested under a given `FeedbackType`
-public enum FeedbackSubType: String {
-    /// Incorrect Visual Subtypes
+/// Enum denoting the subtypes of the  `Incorrect Visual` top-level category
```
1ec5, line 45: This way jazzy should automatically link the symbol:

```suggestion
/// Enum denoting the subtypes of `FeedbackType.incorrectVisual(subtype:)`.
```

(Ditto for the other enumerations.)

MapboxCoreNavigation/Feedback.swift
```diff
@@ -3,89 +3,48 @@ import Foundation
...
+public enum IncorrectVisualSubtype: String {
+    case none
```
1ec5, line 47: Instead of an explicit `none` subtype, make the associated value optional when declaring the `FeedbackType` case:

```swift
case incorrectVisual(subtype: IncorrectVisualSubtype?)
```

MapboxCoreNavigation/Feedback.swift
```diff
@@ -3,89 +3,48 @@ import Foundation
...
-
-
-
 }
 
-/// Enum used to define the many `FeedbackSubType`s that may be nested under a given `FeedbackType`
-public enum FeedbackSubType: String {
-    /// Incorrect Visual Subtypes
+/// Enum denoting the subtypes of the  `Incorrect Visual` top-level category
+public enum IncorrectVisualSubtype: String {
```
1ec5, line 46: To avoid cluttering the main namespace, consider moving these enumerations inside of `FeedbackType`, so that it’s possible to refer to them as e.g. `FeedbackType.IncorrectVisualSubtype`.

MapboxCoreNavigation/Feedback.swift
```diff
@@ -3,89 +3,48 @@ import Foundation
 /**
  Feedback type is used to specify the type of feedback being recorded with `NavigationEventsManager.recordFeedback(type:description:)`.
  */
-public enum FeedbackType: Int, CustomStringConvertible {
+public enum FeedbackType: CustomStringConvertible {
 
     /// Indicates general feedback. You should provide a `description` string to `NavigationEventsManager.recordFeedback(type:description:)`
     /// to elaborate on the feedback if possible.
     case general
```
1ec5, line 10: Not strictly necessary for this PR, but now that we no longer have to support Objective-C, this case could be removed in favor of making the feedback type optional wherever this enumeration is used.

1ec5: On second thought, that might not be a good idea, because the JSON representation of `general` is the string `general`; we’d want to centralize that representation in `FeedbackType.description` instead of hardcoding it elsewhere.

MapboxCoreNavigation/Feedback.swift
nishant-karajgikar: I'd considered this, but the call site looked a little weird ( `.incorrectVisual(subtype: nil)` for example) to me. But i'm not strongly against this approach.

MapboxNavigation/Resources/Assets.xcassets/feedback/confusing_audio.imageset/Contents.json
nishant-karajgikar: Hmm, i think these images aren't imported correctly as template images. I see this if I remove that line and make the change to "Render as: template":

Simulator Screen Shot - iPhone SE (2nd generation) - 2020-06-29 at 10 50 56

MapboxCoreNavigation/Feedback.swift
nishant-karajgikar: I've made more changes to the `FeedbackType` enum and the subtype enums in #2429 . It would be easier to do this in that PR. I'll add a note.

nishant-karajgikar: Added a note in description of https://github.com/mapbox/mapbox-navigation-ios/pull/2429

MapboxNavigation/Resources/Assets.xcassets/feedback/confusing_audio.imageset/Contents.json
nishant-karajgikar: Added a note in the description of https://github.com/mapbox/mapbox-navigation-ios/pull/2429 to handle this issue.

MapboxNavigation/FeedbackItem.swift
nishant-karajgikar: Flagged this as action item as part of https://github.com/mapbox/mapbox-navigation-ios/pull/2429

MapboxNavigation/Resources/Assets.xcassets/feedback/confusing_audio.imageset/Contents.json
1ec5: The problem is that the glyphs within the circles are white instead of transparent. We’ll need to make them transparent so that the template image will show up correctly.

1ec5: mapbox/navigation-ui-resources#15

MapboxCoreNavigation/Feedback.swift
1ec5: If it looks too weird, we can make the argument optional with a default value of `nil`.

MapboxNavigation/Resources/Assets.xcassets/feedback/confusing_audio.imageset/Contents.json
1ec5: #2433 tracks tail work around tint colors.

MapboxNavigation/FeedbackItem.swift
```diff
@@ -7,10 +7,61 @@ extension UIImage {
...
+
+        return .feedbackImage(named: imageName)
+    }
+
+
+    /// Generates a `FeedbackItem` for a given `FeedbackType`
+    /// - Returns: A `FeedbackItem` model object used to render UI
+    func generateFeedbackItem() -> FeedbackItem {
+        return FeedbackItem(title: self.title, image: self.image, feedbackType: self)
+   }
```
1ec5, line 58: This would be a little cleaner as an initializer on FeedbackItem.

MapboxCoreNavigation/Feedback.swift
nishant-karajgikar: Addressed.

1ec5:
`CarPlayManagerTests.testManagerTellsDelegateWhenNavigationStartsAndEndsDueToArrival()` seems to be [crashing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/1147/workflows/4f8319d8-6361-4712-abbf-c7859da79e86/jobs/16749/steps) intermittently. Not immediately actionable here but something to watch out for going forward as we upgrade MapboxNavigationNative further.

PR #2420, opened 2020-06-24
[Feedback][Part 2] Remove progress bar from FeedbackViewController

nishant-karajgikar:
Builds on #2419 
This PR removes the progress bar from the FeedbackViewController as per new design direction.


nishant-karajgikar:
> Suggested changelog entry:
> 
> * The feedback screen no longer dismisses automatically after 10 seconds. ([#2420](https://github.com/mapbox/mapbox-navigation-ios/pull/2420))

Addressed in #2419

PR #2421, opened 2020-06-24
[Feedback][Part 3] Delete old feedback icons

nishant-karajgikar:
Builds on #2420 
This PR deletes the old/unused feedback icons.


1ec5:
This PR needs to be rebased onto release-v1.0-pre-registry, removing 86bd7b24142f81d1bc62b33fcdeb87d4109769a6, which already landed as part of #2420.

PR #2422, opened 2020-06-24
Disable 32-bit architectures

1ec5:
Unfortunately, MapboxNavigationNative is only available for arm64 and x86_64 architectures at the moment, despite the default and overwhelmingly common architectures being arm64, arm64e, armv7, and armv7s. Until we can upgrade to a release of MapboxNavigationNative that supports the right architectures, we’ll have to limit MapboxCoreNavigation and MapboxNavigation to the same set of architectures.

These changes only fix the issue for Carthage; #2416 includes a fix for CocoaPods.

/cc @mapbox/navigation-ios


zugaldia:
Where are we tracking re-adding the missing archs?

1ec5:
We’re tracking that issue upstream, internally, as part of the MapboxNavigationNative project.

Udumft:
This fix is blocking [amazon metrics](https://github.com/mapbox/navigation-sdks/issues/408) collecting, because it requires a release app built for device farm running, and it is failing due to missing arch.

Udumft:
This fix is blocking the metrics collecting task I am working on right now, because it require a release app to be built to run on a device, and missing ARMv7 arch is blocking the CI process.

zugaldia:
Good catch @Udumft !

PR #2424, opened 2020-06-25
Add http request user-agent

fabian-guerra:
Adds user-agent http header. Format: 
```
Example/0.40.0 MapboxNavigation/0.40.0 MapboxDirections/0.33.0 iOS/13.4.0 (x86_64)
```

### DIFF
--- a/MapboxNavigation/ImageDownloader.swift
+++ b/MapboxNavigation/ImageDownloader.swift
@@ -1,4 +1,5 @@
 import Foundation
+import MapboxDirections
 
 typealias ImageDownloadCompletionBlock = (UIImage?, Data?, Error?) -> Void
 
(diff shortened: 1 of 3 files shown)


MapboxNavigation/ImageDownloader.swift
```diff
@@ -1,7 +1,59 @@
 import Foundation
+import MapboxDirections
 
 typealias ImageDownloadCompletionBlock = (UIImage?, Data?, Error?) -> Void
 
+/// The user agent string for any HTTP requests performed directly within this library.
+let userAgent: String = {
```
1ec5, line 7: There are a couple other spots in MapboxCoreNavigation that kick off URLSession data tasks. Perhaps they should also use this user agent string for consistency. We could move this lazily computed constant to an extension on URLSession.

https://github.com/mapbox/mapbox-navigation-ios/blob/71a5de2c77a7945e2fde984fbd34e3198012dc89/MapboxCoreNavigation/LegacyRouteController.swift#L392 https://github.com/mapbox/mapbox-navigation-ios/blob/71a5de2c77a7945e2fde984fbd34e3198012dc89/MapboxCoreNavigation/NavigationService.swift#L535

Issue #2433, opened 2020-07-01
Feedback category icons are white on black regardless of tint color


1ec5:
#2419 replaced the colorful feedback category icons with new icons that include line art within filled circles. These image assets are currently being rendered as black circles with white line art, whereas they need to adapt to the surrounding view controller’s tint color, especially if the icons are to take the same form in Dark Mode.

* [ ] Convert the SVGs in mapbox/navigation-ui-resources#15 into PDFs
* [ ] Update the asset catalog with the corrected PDFs and set the assets’ “Render As” to “Template Image”
* [ ] Remove the clear tint color override, which was probably a mistake: https://github.com/mapbox/mapbox-navigation-ios/blob/fa19891a24b90596169046468947a79c7cd76b75/MapboxNavigation/FeedbackViewController.swift#L286

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/2419#discussion_r445196887
/cc @mapbox/navigation-ios @nishant-karajgikar @avi-c

1ec5:
@avi-c had a good suggestion to replace the assets with just the glyphs and composite the glyphs onto tinted discs at runtime. That way we can tint the glyphs and discs independently and the glyphs can be a color other than the surrounding view controller’s background color. We’ll want to commit new SVGs to navigation-ui-resources and convert them to PDF for inclusion in this project’s asset catalog.

1ec5:
Fixed in #2435.

PR #2435, opened 2020-07-02
Rework Feedback Buttons to allow for tinting

avi-c:
Fix for #2433


avi-c:
@1ec5 - Not sure why this has more than the single commit even though I targeted the PR to nishant's last PR...

1ec5:
You also need to rebase it onto the same branch using `git rebase`.

PR #2439, opened 2020-07-06
Rework Feedback UI assets to allow tinting.

avi-c:
Replacement for PR #2435 

Needed because I incorrectly targeted nishantk/feedbackOverhaulPt4, in that PR and merged it in, which I since reverted from that branch.


MapboxNavigation/FeedbackCollectionViewCell.swift
```diff
@@ -36,6 +44,7 @@ class FeedbackCollectionViewCell: UICollectionViewCell {
     func commonInit() {
         setupViews()
         setupConstraints()
+        circleColor = .black
```
1ec5, line 47: Instead of black, can we have the default tint color be the tint color of the surrounding view controller?

1ec5:
`CarPlayManagerTests.testManagerTellsDelegateWhenNavigationStartsAndEndsDueToArrival()` [keeps crashing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/1186/workflows/83830d66-6685-46ce-81bd-7402b99e7a4d/jobs/17019/steps) on CI, but that’s unrelated to this PR and can be ticketed separately.

The other build bots are failing because of a compiler error:

```
Testing failed:

	Type 'UIColor?' has no member 'systemBackground'

	Testing cancelled because the build failed.
```

1ec5:
It looks like #2421 needs to be rebased onto release-v1.0-pre-registry. Once it gets rebased, the tests will no longer need to build on Xcode 10 where they’re failing.

1ec5:
#2441 tracks the CarPlay test crash.

PR #2446, opened 2020-07-09
Mobile metrics hook

Udumft:
Added a Circle CI config hook to trigger external Navigation metrics collection workflow.

This is meant to happen on each new commit to `master` branch (even though we are not working on `master` atm), to force metrics data collection. Such approach will create a data timeline for changes applied.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,16 +76,15 @@ step-library:
         command: |
           xcodebuild -sdk iphonesimulator -destination 'platform=iOS Simulator,OS=13.4.1,name=iPhone 8 Plus' -project MapboxNavigation.xcodeproj -scheme Example clean build | xcpretty
 
-  - &trigger-metrics:
-    steps:
-      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -76,6 +76,16 @@ step-library:
         command: |
           xcodebuild -sdk iphonesimulator -destination 'platform=iOS Simulator,OS=13.4.1,name=iPhone 8 Plus' -project MapboxNavigation.xcodeproj -scheme Example clean build | xcpretty
 
+  - &trigger-metrics
+      run:
+        name: Trigger metrics
+        command: |
+          if [ -n "${MOBILE_METRICS_TOKEN}" ]; then
```
MaximAlien, line 83: How do we provide `MOBILE_METRICS_TOKEN` here? Is it environment variable on CircleCI?

Udumft: Correct

PR #2452, opened 2020-07-12
Add Probot Stale configuration

zugaldia:
This PR proposes adding configuration for the [Probot Stale app](https://github.com/probot/stale), a GitHub App that closes abandoned issues and Pull Requests after a period of inactivity. The configuration proposed here is based on the one used by Mapbox Maps SDK and it's somewhat conservative:
- Ignore issues or PRs in a project or a milestone.
- Ignore issues or PRs with the `SEMVER-MAJOR` label.
- PRs go stale after 60 days, and issues after 180 days.
- Bot waits for 5 additional days for any activity before closing the ticket or PR.
- It limits the number of actions to 1 per hour (default is 30).

Note that before this bot is active it needs to be [enabled here](https://github.com/apps/stale). That can happen after this PR has approved and merged.

/cc: @mapbox/navigation-ios

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -15,7 +15,7 @@ limitPerRun: 1
 # Do not archive issues that require a SEMVER major release, those normally
 # have a longer lifeycle.
 exemptLabels:
-  - SEMVER-MAJOR
+  - backwards-incompatible
 
 pulls:
   daysUntilStale: 60
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,36 @@
+# Configuration for Probot (Stale) - https://github.com/probot/stale
+
+# Ignore issues in a project (defaults to false)
+exemptProjects: true
+
+# Ignore issues in a milestone (defaults to false)
+exemptMilestones: true
+
+# Label to use when marking as stale
+staleLabel: archived
+
+# Limit the number of actions to 1 per hour
+limitPerRun: 1
+
+# Do not archive issues that require a SEMVER major release, those normally
+# have a longer lifeycle.
+exemptLabels:
+  - SEMVER-MAJOR
+
+pulls:
+  daysUntilStale: 60
+  daysUntilClose: 5
+  markComment: >
+    This pull request has been automatically marked as stale because it has not had
+    recent activity. It will be closed if no further activity occurs. Thank you
+    for your contributions.
+  closeComment: Stale.
+
+issues:
+  daysUntilStale: 180
+  daysUntilClose: 5
+  markComment: >
+    This issue has been automatically marked as stale because it has not had
+    recent activity. It will be closed if no further activity occurs. Thank you
+    for your contributions.
+  closeComment: Stale.


zugaldia:
@1ec5 @MaximAlien could you help me understand why CI failed on this one?

MaximAlien:
> @1ec5 @MaximAlien could you help me understand why CI failed on this one?

There are 2 failures in Circle CI:
1. Related to missing umbrella headers in nav. native. (it was addressed in `1.0.0` release, after we merge it this validation error will be fixed).
2. Missing iOS simulator for specific iOS version (I've added fix for that to `master`).

Other than that PR looks good to me.

.github/stale.yml
```diff
@@ -0,0 +1,36 @@
...
+
+# Label to use when marking as stale
+staleLabel: archived
+
+# Limit the number of actions to 1 per hour
+limitPerRun: 1
+
+# Do not archive issues that require a SEMVER major release, those normally
+# have a longer lifeycle.
+exemptLabels:
```
1ec5, line 17: I wonder if we should exempt any issue with labels indicating that the issue has already been groomed. The most likely issues to be stale are the bug reports we haven’t even looked at that are quite old.

.github/stale.yml
```diff
@@ -0,0 +1,36 @@
...
+  - SEMVER-MAJOR
+
+pulls:
+  daysUntilStale: 60
+  daysUntilClose: 5
```
1ec5, line 22: What if we start less aggressively by marking as stale the issues that are much older than this, like a year old or more? I think that would quickly get things under control so we can treat the remaining issues with more care.

.github/stale.yml
```diff
@@ -0,0 +1,36 @@
...
+# Label to use when marking as stale
+staleLabel: archived
+
+# Limit the number of actions to 1 per hour
+limitPerRun: 1
+
+# Do not archive issues that require a SEMVER major release, those normally
+# have a longer lifeycle.
+exemptLabels:
+  - SEMVER-MAJOR
```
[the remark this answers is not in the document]
zugaldia: Ah, I was actually looking for such a label, but couldn't find it. Let me amend the PR to use `backwards-incompatible`.

.github/stale.yml
zugaldia: Could you share an example or two? In principle, I'm also interested in archiving issues that had no recent activity even if someone looked at them at some point.

.github/stale.yml
zugaldia: The initial value is somewhat arbitrary - 60 days for PRs worked well for Maps SDK. I suggest we start with that value to be consistent with that experience, and adjust as needed as we learn more. I've set `limitPerRun` to 1 precisely with that in mind. The bot will be slow moving :)

zugaldia:
Merging it to get the ball rolling - we'll adjust it as needed. Thanks @MaximAlien @1ec5 for the quick review.

Issue #2113, opened 2019-04-22
Out-of-band usage of RouteController.location causes instructions to be skipped


1ec5:
As a Cocoa developer, I’d expect a getter such as `RouteController.location` or `RouteController.snappedLocation` to have no side effects (even if it may be expensive to get). But in fact, `RouteController.snappedLocation` calls `MBNavigator.getStatusForTimestamp(_:)` every time without caching the result:

https://github.com/mapbox/mapbox-navigation-ios/blob/fb2873cd6815e6aa795b85b3a3342576839e8821/MapboxCoreNavigation/RouteController.swift#L92

`MBNavigator.getStatusForTimestamp(_:)` is stateful: it’ll return a given spoken or visual instruction just once, so the extra calls to this method end up snarfing instructions that should be accessible to RouteController itself when updating the route progress in response to a location update.

A fix upstream in MapboxNavigationNative will distinguish between the two use cases for accessing the status, fixing this issue. Until then, applications should avoid accessing `location` or `snappedLocation` out of band or redundantly. I think we need to issue a warning when the getters are used this way or at least document the problem until we have a fix.

/cc @mapbox/navigation-ios @kevinkreiser

1ec5:
The following workarounds are available:

* Implement `NavigationServiceDelegate.navigationService(_:didUpdate:with:rawLocation:)` and use the passed-in `location` and `rawLocation` arguments.
* Observe `Notification.name.routeControllerProgressDidChange` notifications from the route controller and use the values associated with `RouteControllerNotificationUserInfoKey.locationKey` and `rawLocationKey` in the `userInfo` dictionary.
* Specify `LegacyRouteController` as the router type when initializing the NavigationService.

andrlee:
update from @kevinkreiser: 
> its quite a big refactor and we'll have to do a lot of work to get the test coverage back up
its basically a major api change
so i guess the only thing blocking it really is that its a decent amount of work

@1ec5 could we roll with one of the workarounds? i'd like to discuss tradeoffs of each workaround if possible.

frederoni:
Simplifying the first workaround with this patch should be a valid workaround.

<details><summary>Patch</summary>

```diff
diff --git a/MapboxCoreNavigation/RouteController.swift b/MapboxCoreNavigation/RouteController.swift
index d009f49b..b6006767 100644
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
[...]

PR #2454, opened 2020-07-13
Test for out-of-band navigator calls.

MaximAlien:
Related to #2113.

This test attempts to prove that even though SBNavigator returns `nil` values for instructions after marking several out-of-band calls to methods which internally call `Navigator.getStatusForTimestamp(_:)` it'll not lead to skipping instructions. Since in subsequent calls both voice and banner instructions are `nil` checks performed in `updateIndexes(status:progress:)` will prevent updating public properties like `spokenInstructionIndex` and `visualInstructionIndex`.

Example:
```
RouteController spoken instruction: Head east on Coventry Street (A4), then turn right onto Haymarket (A4)
RouteController banner instruction: Haymarket
...
Navigator status: <MBNavigationStatus: 0x6000001d14d0>
Navigator status coordinate: CLLocationCoordinate2D(latitude: 51.510055172571825, longitude: -0.13371546842507087)
Navigator status voice instruction: Head east on Coventry Street (A4), then turn right onto Haymarket (A4)
Navigator status step index: 0
Navigator status leg index: 0
Navigator status route index: 0
Navigator status intersection index: 1
Navigator status banner instruction: Haymarket
Remaining status leg distance: 1266.7922
Remaining status leg duration: 259.482696533

Navigator status: <MBNavigationStatus: 0x6000001b8120>
Navigator status coordinate: CLLocationCoordinate2D(latitude: 51.510055172502106, longitude: -0.1337154690437497)
Navigator status voice instruction: nil
Navigator status step index: 0
Navigator status leg index: 0
Navigator status route index: 0
Navigator status intersection index: 1
Navigator status banner instruction: nil
Remaining status leg distance: 1266.7922
Remaining status leg duration: 259.482696533

Navigator status: <MBNavigationStatus: 0x6000001b8120>
Navigator status coordinate: CLLocationCoordinate2D(latitude: 51.510055172502106, longitude: -0.1337154690437497)
Navigator status voice instruction: nil
Navigator status step index: 0
Navigator status leg index: 0
Navigator status route index: 0
Navigator status intersection index: 1
Navigator status banner instruction: nil
Remaining status leg distance: 1266.7922
Remaining status leg duration: 259.482696533
...
RouteController spoken instruction: Head east on Coventry Street (A4), then turn right onto Haymarket (A4)
RouteController banner instruction: Haymarket
```

__Update:__ Looks like original issue is no longer reproducible when using Navigation Native `22.0.5`.


MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -617,6 +617,110 @@ class NavigationServiceTests: XCTestCase {
...
+                    }
+                    
+                    XCTAssertEqual(currentBannerInstruction, bannerInstructionsArray.first)
+                    
+                    // Attempt to snarf instructions by calling methods which internally call Navigator.getStatusForTimestamp(_:) multiple times.
+                    let outOfBandCallsCount = 5
+                    for _ in 0..<outOfBandCallsCount {
+                        let _ = navigationService.router.userIsOnRoute($0.element)
+                        let _ = routeController.location
+                        let _ = routeController.snappedLocation
```
Udumft, line 706: How do you think, maybe it is better to introduce some minimal delay between the calls so that `getStatusForTimestamp:` is called with more or less different timestamps? Would that make any difference?

MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -617,6 +617,110 @@ class NavigationServiceTests: XCTestCase {
...
+                        navigationService.router!.locationManager!(navigationService.locationManager, didUpdateLocations: [$0.element])
```
Udumft, line 707: Why is location update called multiple times (inside the cycle)? Is it to simulate the case when users manually advance the progress?

MaximAlien: Yep, basically idea is to call methods which internally call `getStatusForTimestamp:` to see whether data it returns is correct and doesn't lead to any issues on platform side (like showing voice/visual instructions).

MapboxCoreNavigationTests/NavigationServiceTests.swift
MaximAlien: Good note. As it turns out with certain delay without changing current location `SBNavigator` starts behaving incorrectly. I'll submit test which covers that. I've also created ticket on nav. native side [here](https://github.com/mapbox/mapbox-navigation-native/issues/1445).

MaximAlien:
Looks like this issue is no longer reproducible after switching to `Navigator.getStatusForMonotonicTimestampNanoseconds(_:)` instead of `Navigator.getStatusForTimestamp(_:)` in Navigation Native `22.0.5`. 

Instead of adding new test which verifies whether out-of-band calls do not snarf voice and visual instructions existing test `testOrderOfExecution` was updated.

Rmkasun-8 (commented):
Approval

MapboxCoreNavigationTests/NavigationServiceTests.swift
Rmkasun-8: > Good note. As it turns out with certain delay without changing current location `SBNavigator` starts behaving incorrectly. I'll submit test which covers that. I've also created ticket on nav. native side.

Rmkasun-8 (commented):
Activate

PR #2462, opened 2020-07-16
Fix maneuver location.

### LINKED ISSUE #2451: Map is not centered correctly after switching to different maneuver location.
**Mapbox Navigation SDK version:** `1.0.0-beta.1`

When selecting different maneuver location map is not centered correctly, after selection maneuver is not visible at all. Please see attached example:
output

MaximAlien:
Closes #2451.

Looks like this is regression which was introduced between `0.37.0` and `0.38.1` after modifying `contentInset(forOverviewing:)`. Since `topBannerContainerView.bounds.height` is used as top inset this leads to setting large content inset for map view, this in turn leads to placing maneuver below bottom banner. Fix aims to update content insets only after closing list with maneuvers and finishing animation. 

Original issue is not reproducible when using guidance cards.


MaximAlien:
Result:
output

Issue #2474, opened 2020-07-20
Feedback categories should be localizable


1ec5:
#2419 replaced the feedback categories with new categories and subcategories but hard-coded the new strings as string literals. As a result, v1.0 would have a regression where the feedback UI would be unintelligible to speakers of languages other than English. The new strings should either be extracted into localizable strings or moved into a localizable property list.

/cc @mapbox/navigation-ios @nishant-karajgikar @avi-c

PR #2475, opened 2020-07-20
Add Turkish localization; update translations from Transifex

1ec5:
Added a Turkish localization. Thanks to @CosmosExmo for contributing this localization!

Added a missing Ukrainian resource to the project. It had been committed but not included in the SDK.

Removed feedback category strings that are no longer needed as of #2419. #2474 tracks making the new feedback categories and subcategories localizable.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/Resources/tr.lproj/Localizable.strings
+++ b/MapboxNavigation/Resources/tr.lproj/Localizable.strings
@@ -1,0 +1,129 @@
+/* Delimiter between lines in an address when displayed inline */
+"ADDRESS_LINE_SEPARATOR" = ". ";
+
+/* Title on arrival action sheet */
+"CARPLAY_ARRIVED" = "Hedefe vardınız";
+
+/* Message on arrival action sheet */
+"CARPLAY_ARRIVED_MESSAGE" = "Ne yapmak istersiniz?";
+
+/* Name of the waypoint associated with the current location */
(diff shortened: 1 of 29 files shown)


MapboxNavigation/Resources/tr.lproj/Localizable.strings
```diff
@@ -0,0 +1,129 @@
...
+"REROUTING" = "Yeniden yönlendiriliyor…";
+
+/* Button title for resume tracking */
+"RESUME" = "Devam et";
+
+/* Label above the speed limit in an MUTCD-style speed limit sign. Keep as short as possible. */
+"SPEED_LIMIT_LEGEND" = "Max";
+
+/* The text of a banner that appears during turn-by-turn navigation when route simulation is enabled. */
+"USER_IN_SIMULATION_MODE" = "Simulating Navigation at %@×";
```
MaximAlien, line 122: Not sure if this is expected, but I still see some English in Turkish translation.

1ec5: This file is at least 80% translated, so the repository is configured to pull down the whole thing:

https://github.com/mapbox/mapbox-navigation-ios/blob/f2f78ce9ebd7d411cbb907e479b5bc5327b3bd85/.tx/config#L4

PR #2479, opened 2020-07-20
Make sure to access application state from UI thread.

### LINKED ISSUE #2468: UIApplication.State in NavigationEventDetails should be accessed only from main queue.
**Mapbox Navigation SDK version:** `1.0.0-beta1`

`NavigationEventDetails` struct can be created on global queue. This leads to Main Thread Checker diagnostic errors when accessing/creating `applicationState` property.

MaximAlien:
Closes #2468.

I've added `testNavigationEventDetailsGlobalQueue` to `NavigationEventsManagerTests.swift`. If needed I can create separate file for `NavigationEventDetails` tests. Also if makes sense test to ensure all keys are encoded to dictionary correctly can be added.

### DIFF
--- a/MapboxCoreNavigation/EventDetails.swift
+++ b/MapboxCoreNavigation/EventDetails.swift
@@ -118,6 +118,8 @@ struct NavigationEventDetails: EventDetails {
     var newDistanceRemaining: CLLocationDistance?
     var newDurationRemaining: TimeInterval?
     var newGeometry: String?
+    var totalTimeInForeground: TimeInterval
+    var totalTimeInBackground: TimeInterval
     
     init(dataSource: EventsManagerDataSource, session: SessionState, defaultInterface: Bool) {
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/EventDetails.swift
```diff
@@ -167,9 +180,9 @@ struct NavigationEventDetails: EventDetails {
         }
         percentTimeInPortrait = totalTimeInPortrait + totalTimeInLandscape == 0 ? 100 : Int((totalTimeInPortrait / (totalTimeInPortrait + totalTimeInLandscape)) * 100)
         
-        var totalTimeInForeground = session.timeSpentInForeground
-        var totalTimeInBackground = session.timeSpentInBackground
```
MaximAlien, line 171: Looks like `totalTimeInForeground`, `totalTimeInBackground` and `rating` were not encoded to result dictionary.

PR #2485, opened 2020-07-22
Consume Navigation Native through SDK Registry

chezzdev:
This PR changes the source of Navigation Native dependency from Bintray to SDK Registry.
Although the changes to the release process are still in review at https://github.com/mapbox/mapbox-navigation-native/pull/1480, navigation-ios can already upgrade to manually added version.

cc: @mapbox/navigation-ios 
ref: https://github.com/mapbox/mapbox-navigation-native/issues/1479

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,6 @@ step-library:
       run:
         name: Prepare .netrc file
         command: |
-          echo >> ~/.netrc
           echo "machine api.mapbox.com" >> ~/.netrc
           echo "login mapbox" >> ~/.netrc
           echo "password $SDK_REGISTRY_TOKEN" >> ~/.netrc
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,9 +46,6 @@ step-library:
       run:
         name: Prepare .netrc file
         command: |
-          echo "machine dl.bintray.com" > ~/.netrc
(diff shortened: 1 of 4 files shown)


MaximAlien:
I think we'll also need to update `README.md` and `CONTRIBUTING.md` to remove any Bintray related info.

.circleci/config.yml
```diff
@@ -46,9 +46,6 @@ step-library:
       run:
         name: Prepare .netrc file
         command: |
-          echo "machine dl.bintray.com" > ~/.netrc
-          echo "login $BINTRAY_LOGIN" >> ~/.netrc
-          echo "password $BINTRAY_API_KEY" >> ~/.netrc
           echo >> ~/.netrc
```
1ec5, line 49: This line is also unnecessary, but benign.

chezzdev:
> I think we'll also need to update README.md and CONTRIBUTING.md to remove any Bintray related info.

@MaximAlien good catch, thanks! Updated.

> I can also see that $ pod lib lint MapboxCoreNavigation.podspec is failing because MapboxNavigationNative.podspec is not yet updated. I guess modifications on nav. native side will be required as well.

Yes, it was due to the NavNative podspecs pointing to Bintray. I've re-pushed them with the links pointing to SDK Registry links and tests are passing now.

PR #2495, opened 2020-07-28
Fix the moment of custom feedback event creation

### LINKED ISSUE #2484: Feedback event should be created when tapping feedback button, not when completing the form
For feedback events, `NavigationEventDetails` is populated with coordinates, a screenshot, etc. when the structure is initialized as part of a call to `NavigationEventsManager.recordFeedback(type:description:)`. With the normal built-in feedback flow, this method gets called by side effect when getting `FeedbackViewController.uuid`, which currently only happens once the user taps a feedback category, right before we complete the feedback event with a category:

https://github.com/mapbox/mapbox-navigation-ios/blob/50def9e3dacb22c9ae3c95d54067afbcf1584afd/MapboxNavigation/FeedbackViewController.swift#L230-L234 https://github.com/mapbox/mapbox-navigation-ios/blob/50def9e3dacb22c9ae3c95d54067afbcf1584afd/MapboxNavigation/FeedbackViewController.swift#L126-L128

However, if the user taps the feedback button in CarPlay, `CarPlayNavigationViewController.createFeedbackUI()` creates the event and populates it right away when showing the feedback UI. mapbox/mapbox-navigation-android#3274 also moved the coordinate and screenshot capturing earlier to when the user taps the feedback button and the feedback UI appears.

For consistency with CarPlay and Android, we should do the same on iOS by getting `uuid` when initializing the FeedbackViewController. To avoid future confusion, we should also turn `uuid` into a method that obviously has a side effect.

/cc @mapbox/navigation-ios @avi-c @JunDai

dersim-davaod:
Fixes #2484

### DIFF
--- a/Sources/MapboxNavigation/FeedbackViewController.swift
+++ b/Sources/MapboxNavigation/FeedbackViewController.swift
@@ -133,6 +133,11 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
     var uuid: UUID? {
         return eventsManager?.recordFeedback()
     }
+
+    /**
+     Current feedback's UUID.
+     */
+    var currentFeedbackUUID: UUID?
     
     /**
      Initialize a new FeedbackViewController from a `NavigationEventsManager`.
@@ -156,6 +161,8 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
     func commonInit() {
         self.modalPresentationStyle = .custom
         self.transitioningDelegate = self
+
+        currentFeedbackUUID = eventsManager?.recordFeedback()
     }
     
     override public func viewDidLoad() {
@@ -233,7 +240,7 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
     }
     
     func send(_ item: FeedbackItem) {
-        if let uuid = self.uuid {
+        if let uuid = currentFeedbackUUID {
             delegate?.feedbackViewController(self, didSend: item, uuid: uuid)
             eventsManager?.updateFeedback(uuid: uuid, type: item.feedbackType, source: .user, description: nil)
         }
@@ -250,7 +257,7 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
     
     func dismissFeedbackItem() {
         delegate?.feedbackViewControllerDidCancel(self)
-        if let uuid = self.uuid {
+        if let uuid = currentFeedbackUUID {
             eventsManager?.cancelFeedback(uuid: uuid)
         }
         dismiss(animated: true, completion: nil)
--- a/Sources/MapboxNavigation/FeedbackViewController.swift
+++ b/Sources/MapboxNavigation/FeedbackViewController.swift
@@ -130,10 +130,6 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
      */
     public weak var eventsManager: NavigationEventsManager?
     
-    var uuid: UUID? {
-        return eventsManager?.recordFeedback()
-    }
-
     /**
      Current feedback's UUID.
      */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.4.0
 
+* Fixed the moment of custom feedback event creation. ([#2484](https://github.com/mapbox/mapbox-navigation-ios/pull/2495))
 * Increased the minimum version of `MapboxNavigationNative` to v32.0.0. ([#2910](https://github.com/mapbox/mapbox-navigation-ios/pull/2910))
 * Fixed an issue when feedback UI was always appearing for short routes. ([#2871](https://github.com/mapbox/mapbox-navigation-ios/pull/2871))
 * Fixed automatic day/night switching. ([#2881](https://github.com/mapbox/mapbox-navigation-ios/pull/2881))
--- a/Sources/MapboxNavigation/FeedbackSubtypeViewController.swift
+++ b/Sources/MapboxNavigation/FeedbackSubtypeViewController.swift
@@ -161,7 +161,7 @@ class FeedbackSubtypeViewController: FeedbackViewController {
 
     private func sendReport() {
         if selectedItems.count > 0 {
-            if let uuid = self.uuid {
+            if let uuid = self.currentFeedbackUUID {
                 for item in selectedItems {
                     delegate?.feedbackViewController(self, didSend: item, uuid: uuid)
                     eventsManager?.updateFeedback(uuid: uuid, type: item.feedbackType, source: .user, description: nil)


Udumft:
This may not be related to this change, but to the mechanism in general. 
As I understood, each time we call `eventsManager?.recordFeedback()` there is a new`FeedbackEvent` created and enqueued. So, even if user opened the Feedback UI and then closed it, we still create 2 events (1 at `commonInit()` and 2 in `dismissFeedback()` method. `eventsManager?.recordFeedback()` effectively creates and enqueues new event). Shouldn't we instead create only one`FeedbackEvent` and 'cache' it's `uuid` in `FeedbackViewController` instance for further manipulations?

dersim-davaod:
@Udumft yes, makes sense for me to implement this. Will update PR soon.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -226,30 +224,49 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
...
     }
+
+    /**
+     Gets an unique identifier used to identify the current feedback event.
+
+     The UUID that method returns is the result of calling `NavigationEventsManager.recordFeedback()`.
+
+     - Returns: UUID used to identify the current feedback event.
+     */
+    @discardableResult private func currentFeedbackEventID() -> UUID? {
```
1ec5, line 267: This method’s name should indicate that it has a side effect, not just that it returns a UUID. `recordFeedback()` would be appropriate here.

bamx23:
The feedback UUID is still generated for every call of `currentFeedbackEventID()`, so I guess the @Udumft's comment above is still actual.

Sources/MapboxNavigation/FeedbackViewController.swift
```diff
@@ -231,30 +231,48 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
...
-    
+
+    /**
+     Sends the current feedback event for the specific feedback item.
+
+     - Parameter item: a feedback item for which feedback will be sent.
+     */
     func send(_ item: FeedbackItem) {
-        if let uuid = self.uuid {
+        if let uuid = currentFeedbackUUID {
```
bamx23, line 241: When does this fail? Should we handle such errors explicitly?

Sources/MapboxNavigation/FeedbackViewController.swift
```diff
@@ -156,6 +154,8 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
     func commonInit() {
         self.modalPresentationStyle = .custom
         self.transitioningDelegate = self
+
+        currentFeedbackUUID = self.recordCurrentFeedback()
```
bamx23, line 158: Is it OK to make a screenshot implicitly in the constructor of this class?

bamx23 (approved):
LGTM overall. I haven't checked this PR in the 1Tap, but it seems to solve the issue of wrong screenshots/data.

Sources/MapboxNavigation/FeedbackViewController.swift
dersim-davaod: Agreed to leave it as it is. Possible error handling can be done in scope of separate PR.

Sources/MapboxNavigation/FeedbackViewController.swift
dersim-davaod: Agreed to make currentFeedbackUUID lazy var to avoid creation of cached UUID in the init method.

But there're another topics to discuss:
- in the future we might want to have an opportunity to control the moment of feedback creation explicitly (say, using public API method)
- this also causes creation and immediate removal of feedback screenshot in the `dismissFeedbackItem` method.

dersim-davaod: Reverted, as it is better to call `recordFeedback` in the constructor, otherwise the screenshot will have a different UI at the moment of sending.

Sources/MapboxNavigation/FeedbackViewController.swift
```diff
@@ -129,11 +129,9 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
      The events manager used to send feedback events.
      */
     public weak var eventsManager: NavigationEventsManager?
-    
-    var uuid: UUID? {
-        return eventsManager?.recordFeedback()
-    }
-    
+
+    private var currentFeedbackUUID: UUID?
```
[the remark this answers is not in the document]
dersim-davaod: Fixed.

bamx23 (approved):

Sources/MapboxNavigation/FeedbackViewController.swift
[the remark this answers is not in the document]
1ec5: [Current builds are still failing on the same error](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/2929/workflows/43180726-7aaf-47f5-84ef-eddedeb4d449/jobs/29106).

Sources/MapboxNavigation/FeedbackViewController.swift
```diff
@@ -133,6 +133,11 @@ public class FeedbackViewController: UIViewController, DismissDraggable, UIGestu
     var uuid: UUID? {
```
bamx23, line 133: Do we need this now?

Sources/MapboxNavigation/Resources/de.lproj/Localizable.strings
```diff
@@ -91,6 +94,9 @@
 /* Label indicating the device volume is too low to hear spoken instructions and needs to be manually increased */
 "INAUDIBLE_INSTRUCTIONS_CTA" = "Lautstärke anpassen, um Anweisungen zu hören";
 
+/* Specific route feedback that a speed limit is incorrect. */
+"INCORRECT_SPEED_LIMIT" = "Speed limit incorrect";
```
bamx23, line 98: Why are there so many unrelated changes?

Sources/MapboxNavigation/FeedbackViewController.swift
dersim-davaod: removed.

Sources/MapboxNavigation/Resources/de.lproj/Localizable.strings
dersim-davaod: These changes are not intented for this PR, should be updated, thanks!

S2Ler:
Removed unnecessary commits introduced due to incorrect rebase.

Sources/MapboxNavigation/FeedbackViewController.swift
bamx23: Once again, do we need this now?

S2Ler:
The builds are failing because the machine can't download needed simulators.

Sources/MapboxNavigation/FeedbackViewController.swift
bamx23: сс @S2Ler

S2Ler: Not needed. Removed.

S2Ler: Removed. No longer used.

bamx23:
LGTM, thanks

PR #2497, opened 2020-07-28
Make feedback categories localizable

### LINKED ISSUE #2474: Feedback categories should be localizable

dersim-davaod:
Fixes #2474


MapboxNavigation/FeedbackItem.swift
```diff
@@ -10,21 +10,20 @@ extension UIImage {
...
 
-    // TODO: Localize these strings
     internal var title: String {
         switch self {
             case .general:
-                return "Feedback"
+                return NSLocalizedString("FEEDBACK_TYPE_GENERAL", bundle: .mapboxNavigation, value: "Feedback", comment: "Feedback type for general or unknown issue")
             case .incorrectVisual(_):
-                return "Incorrect Visual"
+                return NSLocalizedString("FEEDBACK_TYPE_INCORRECT_VISUAL", bundle: .mapboxNavigation, value: "Incorrect Visual", comment: "Feedback type for incorrect visual")
```
1ec5, line 18: This is ungrammatical, which will surely hinder some users from choosing this option even when appropriate. It’s also difficult to translate for the same reason. I’d suggest “Looks Incorrect” as an alternative.

/cc @abhishek1508 @asinghal22 @avi-c

dersim-davaod: Updated as @1ec5 suggested.

dersim-davaod: P.S.: @mapbox/navigation-android might want to change this as well: 
https://github.com/mapbox/mapbox-navigation-android/blob/7de359e97901df944a0c53c8f452da2265d1958c/libnavigation-ui/src/main/res/values/strings.xml#L21-L22

MapboxCoreNavigation/Feedback.swift
```diff
@@ -28,8 +28,8 @@ public enum FeedbackType: CustomStringConvertible {
         switch self {
         case .general:
             return "general"
-        case .incorrectVisual(_):
-            return "incorrect_visual"
+        case .looksIncorrect(_):
+            return "looks_incorrect"
```
1ec5, line 32: This needs to remain `incorrect_visual` for compatibility with the backend and consistency with the Android implementation. This string is never presented to users, so translation isn’t a concern.

MapboxNavigation/FeedbackItem.swift
JunDai: thanks @1ec5 and @dersim-davaod . I will make changes accordingly on Android

JunDai:
@1ec5 / @dersim-davaod : I can't remember the reason but the feedback types defined in Android are like [this](https://github.com/mapbox/mapbox-navigation-android/blob/master/libnavigation-core/src/main/java/com/mapbox/navigation/core/telemetry/events/FeedbackEvent.kt#L62-L75) which is different than iOS:

```kotlin
    const val ROAD_CLOSED = "road_closed"
    const val NOT_ALLOWED = "not_allowed"
    const val ROUTING_ERROR = "routing_error"
    const val INCORRECT_VISUAL_GUIDANCE = "incorrect_visual_guidance"
    const val INCORRECT_AUDIO_GUIDANCE = "incorrect_audio_guidance"
    const val POSITIONING_ISSUE = "positioning_issue"
```

this is what we can see from `MODE`:
feedback_subtype_mode

for those feedback types, I think I got them as requirement from the original ticket. It would be nice that iOS can check it, we can sync offline.

JunDai:
@dersim-davaod - as discussed offline with @1ec5 , we would be better to update the feedback types raw values to align with Android. Would you mind to update it?

1ec5:
Yeah, let’s update the raw values on iOS to match. The Android implementation differs from the original specification @nishant-karajgikar and I looked at, but we weren’t sure whether that specification was authoritative. Now that the events are coming in on Android, that implementation is the source of truth anyways.

This is orthogonal to renaming the localizable string. I think the Android implementation should be renamed too. As a translator, I was at a loss for how to translate “Incorrect Visual”.

JunDai:
@1ec5 - yeah, I will have a PR to update the text "Incorrect Visual" to "Looks Incorrect" for UI displaying only.

dersim-davaod:
> we would be better to update the feedback types raw values to align with Android. Would you mind to update it?

@JunDai, sure

avi-c:
I started the process of moving these strings to be localizable but haven't had time to finish it due to pressing issues with some customer conversations. I think I'll need to pass this off to someone. I can provide the work I've done so far to them.

1ec5:
Superseded by #2576.

1ec5:
Superseded by #2576.

Issue #2493, opened 2020-07-28
Crash when setting lineGradient in MGLLineStyleLayer in Release mode.


MaximAlien:
**Mapbox Navigation SDK version:** `1.0.0-beta.1`

Crash occurs when placing destination marker on a map. It's reproducible only in Release mode when running example project and only when using `iPhone 5s, Xs, Xr, Xs Max` devices (maybe other as well). I cannot reproduce crash when running example on iPhone 11 Simulator. Here's stack trace:
[Example_2020-07-27-162644_MacBook-Pro.txt](https://github.com/mapbox/mapbox-navigation-ios/files/4985346/Example_2020-07-27-162644_MacBook-Pro.txt)

```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil'
terminating with uncaught exception of type NSException
abort() called
CoreSimulator 725.10 - Device: iPhone Xs Max (03DF22DF-2111-47B1-96F9-50E03D8FABDA) - Runtime: iOS 12.2 (16E226) - DeviceType: iPhone Xs Max

Application Specific Backtrace 1:
0   CoreFoundation                      0x00000001038e86fb __exceptionPreprocess + 331
1   libobjc.A.dylib                     0x0000000102e94ac5 objc_exception_throw + 48
2   CoreFoundation                      0x0000000103836ddc _CFThrowFormattedException + 194
3   CoreFoundation                      0x0000000103813ae5 -[__NSArrayM insertObject:atIndex:] + 1269
4   Mapbox                              0x0000000102484cfb -[NSExpression(MGLAdditions) mgl_jsonInterpolationExpressionObject] + 1923
5   Mapbox                              0x0000000102482ad2 -[NSExpression(MGLAdditions) mgl_jsonExpressionObject] + 10432
[...]

PR #2500, opened 2020-07-28
Fix crash when passing Dictionary to NSExpression.

MaximAlien:
Closing #2493.

Due to Swift optimization in Release mode keys in `routeGradientStops.lines` dictionary were set to `nil` after NSExpression creation. It was causing crash on certain devices only. Conversion to NSDictionary fixes this.

### DIFF
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -230,6 +230,9 @@ class ViewController: UIViewController {
         let service = navigationService(route: route, options: routeOptions)
         let navigationViewController = self.navigationViewController(navigationService: service)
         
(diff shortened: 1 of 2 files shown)


MaximAlien:
> This is an elegant, if unfortunate, workaround for #2493. Hopefully the crash we’re fixing isn’t indicative of a larger issue with the Swift-Objective-C bridge or the map SDK’s NSExpression-JSON converter.

I have a feeling that this might be related to value vs. class type semantics and how Swift optimizing this in certain cases. Since Swift `Dictionary` is value type and `NSDictionary` is Objective-C class type it might explain the reason why keys are not deallocated when using `NSDictionary`.

PR #2408, opened 2020-06-19
Add junction views.

fabian-guerra:
Adds junction views to navigation UX.

PR #2501, opened 2020-07-28
Show junction views at the distance along route specified by Directions API

avi-c:
This is to ensure we reach the distanceAlongStep specified by the Directions API before showing the JV.

### DIFF
--- a/MapboxNavigation/InstructionsCardViewController.swift
+++ b/MapboxNavigation/InstructionsCardViewController.swift
@@ -299,18 +299,22 @@ extension InstructionsCardViewController: UICollectionViewDelegateFlowLayout {
 extension InstructionsCardViewController: NavigationComponent {
     public func navigationService(_ service: NavigationService, didUpdate progress: RouteProgress, with location: CLLocation, rawLocation: CLLocation) {
         routeProgress = progress
+        junctionView.updateDistance(for: progress.currentLegProgress.currentStepProgress)
         reloadDataSource()
     }
     
     public func navigationService(_ service: NavigationService, didPassVisualInstructionPoint instruction: VisualInstructionBanner, routeProgress: RouteProgress) {
         self.routeProgress = routeProgress
-        junctionView.update(for: instruction, service: service)
+        DispatchQueue.main.async {
+            self.junctionView.update(for: instruction, service: service)
+        }
         reloadDataSource()
     }
     
     public func navigationService(_ service: NavigationService, didRerouteAlong route: Route, at location: CLLocation?, proactive: Bool) {
         self.currentStepIndex = nil
         self.routeProgress = service.routeProgress
+        junctionView.updateDistance(for: service.routeProgress.currentLegProgress.currentStepProgress)
         reloadDataSource()
     }
 }
--- a/MapboxNavigation/JunctionView.swift
+++ b/MapboxNavigation/JunctionView.swift
@@ -37,11 +37,8 @@ public class JunctionView: UIImageView {
         }
         
         guard let guidanceView = quaternaryInstruction?.components.first else { return }
-        if let visualInstructionDistance = visualInstruction?.distanceAlongStep {
-            distanceAlongStep = visualInstructionDistance
-        } else {
-            distanceAlongStep = nil
-        }
+        distanceAlongStep = visualInstruction?.distanceAlongStep
+        
         if case .guidanceView(let guidanceViewImageRepresentation, _) = guidanceView {
             if let cachedImage = imageRepository.cachedImageForKey(guidanceView.cacheKey!) {
                 image = cachedImage
@@ -53,11 +50,8 @@ public class JunctionView: UIImageView {
 
                 guard let guidanceViewImageURL = URL(string: stringURL) else { return }
                 imageRepository.imageWithURL(guidanceViewImageURL, cacheKey: guidanceView.cacheKey!) { [unowned self] (downloadedImage) in
-                    
-                    self.isCurrentlyVisible = true
-                    
+                                        
                     DispatchQueue.main.async {
-                        self.isHidden = !self.isCurrentlyVisible
                         self.image = downloadedImage
                     }
                 }
--- a/MapboxNavigation/JunctionView.swift
+++ b/MapboxNavigation/JunctionView.swift
@@ -7,7 +7,7 @@ import MapboxDirections
 public class JunctionView: UIImageView {
     var isCurrentlyVisible: Bool = false
     var imageRepository: ImageRepository = .shared
-    var distanceAlongStep: CLLocationDistance = -1
+    var distanceAlongStep: CLLocationDistance?
     
     public override init(frame: CGRect) {
         super.init(frame: frame)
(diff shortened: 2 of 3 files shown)


Cartfile
```diff
@@ -1,7 +1,7 @@
 binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" ~> 2.3.0
-binary "https://mapbox-gl-native-ios.s3.amazonaws.com/public/internal/Mapbox-iOS-SDK.json" == 5.9.1000
 github "mapbox/mapbox-directions-swift" ~> 0.33
-binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 14.1.5
+binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps/mapbox-ios-sdk-dynamic.json" ~> 6.0
+binary "https://www.mapbox.com/ios-sdk/MapboxNavigationNative.json" ~> 14.1.6
```
1ec5, line 4: Can you rebase #2408 instead of making this change here? I think it’ll end up causing unnecessary merge conflicts later.

MapboxNavigation/JunctionView.swift
```diff
@@ -7,6 +7,7 @@ import MapboxDirections
 public class JunctionView: UIImageView {
     var isCurrentlyVisible: Bool = false
     var imageRepository: ImageRepository = .shared
+    var distanceAlongStep: CLLocationDistance = -1
```
1ec5, line 10: Make this property optional to avoid magic numbers like −1.

MapboxNavigation/JunctionView.swift
```diff
@@ -50,18 +51,30 @@ public class JunctionView: UIImageView {
...
-                        self.show(true)
                     }
                 }
             }
         }
         
     }
     
+    public func updateDistance(for currentStepProgress: RouteStepProgress) {
+        let distanceTravelled = currentStepProgress.distanceTraveled
```
1ec5, line 66: Nit: use the American spelling, `distanceTraveled`, for consistency with `RouteStepProgress.distanceTraveled`.

MapboxNavigation/JunctionView.swift
```diff
@@ -50,18 +51,30 @@ public class JunctionView: UIImageView {
...
                         self.image = downloadedImage
-                        self.show(true)
                     }
                 }
             }
         }
         
     }
     
+    public func updateDistance(for currentStepProgress: RouteStepProgress) {
```
1ec5, line 65: Add a documentation comment like “Show or hide the junction view based on the distance traveled in an updated step progress.”

MapboxNavigation/JunctionView.swift
```diff
@@ -50,18 +51,30 @@ public class JunctionView: UIImageView {
...
+        let distanceTravelled = currentStepProgress.distanceTraveled
+        if distanceAlongStep > -1 {
+            // show the Junction View if we have progressed enough along the step
+            // hide the Junction View if it is still visible from a previous instruction but shouldn't be yet.
+            if distanceTravelled >= distanceAlongStep, isHidden == true  {
+                show()
+            } else if distanceTravelled < distanceAlongStep, isHidden == false {
+                hide()
+            }
```
1ec5, line 74: Trying to make the logic a little easier to follow:

```suggestion
            var shouldHide = distanceTravelled < distanceAlongStep
            if isHidden && !shouldHide {
                show()
            } else if !isHidden && shouldHide {
                hide()
            }
```

MapboxNavigation/JunctionView.swift
avi-c: Jolly good.

MapboxNavigation/JunctionView.swift
avi-c: This is definitely clearer. I'll update my code.

MapboxNavigation/JunctionView.swift
avi-c: Will do.

MapboxNavigation/JunctionView.swift
avi-c: 👍

Cartfile
avi-c: Sounds good.

MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -299,18 +299,22 @@ extension InstructionsCardViewController: UICollectionViewDelegateFlowLayout {
...
+        junctionView.updateDistance(for: progress.currentLegProgress.currentStepProgress)
         reloadDataSource()
     }
     
     public func navigationService(_ service: NavigationService, didPassVisualInstructionPoint instruction: VisualInstructionBanner, routeProgress: RouteProgress) {
         self.routeProgress = routeProgress
-        junctionView.update(for: instruction, service: service)
+        DispatchQueue.main.async {
+            self.junctionView.update(for: instruction, service: service)
+        }
```
1ec5, line 310: I’m curious about this update being asynchronous whereas the other two calls to `update(for:service:)` remain synchronous.

MapboxNavigation/JunctionView.swift
```diff
@@ -32,14 +33,18 @@ public class JunctionView: UIImageView {
...
+            distanceAlongStep = nil
         }
         
         guard let guidanceView = quaternaryInstruction?.components.first else { return }
-        
+        if let visualInstructionDistance = visualInstruction?.distanceAlongStep {
+            distanceAlongStep = visualInstructionDistance
+        } else {
+            distanceAlongStep = nil
+        }
```
1ec5, line 44: Nit: `distanceAlongStep` is already optional, so we don’t need to unwrap it:

```suggestion
        distanceAlongStep = visualInstruction?.distanceAlongStep
```

MapboxNavigation/JunctionView.swift
```diff
@@ -49,19 +54,31 @@ public class JunctionView: UIImageView {
                 guard let guidanceViewImageURL = URL(string: stringURL) else { return }
                 imageRepository.imageWithURL(guidanceViewImageURL, cacheKey: guidanceView.cacheKey!) { [unowned self] (downloadedImage) in
                     
-                    self.isCurrentlyVisible = true
-                    self.isHidden = !self.isCurrentlyVisible
```
1ec5, line 53: Does this affect whether the view is visible while the image is still loading? If the view is currently visible, the view always gets hidden here, so what happens if there are two junctions in short succession that both need the junction view to be visible?

avi-c: This code is from Fabian's original PR, I just moved it into a dispatch block since the function can be invoked off the main thread. As to the substance of your comment I think you are right that the original code was to prevent the view from being visible when the image wasn't loaded yet. However, I don't think we need to deal with that in the loading code and let the show/hide logic sort it out. It works fine in my testing but I haven't been able to come up with a test case that includes having two maneuvers show up in quick succession.

avi-c:
@1ec5 - If you approve then I'll merge this down to Fabian's PR and keep working from there.

PR #2502, opened 2020-07-28
Pickup Guidance Card container corner radius from  card style.

avi-c:


avi-c:
@nishant-karajgikar could you check this small PR out? I think it's the last thing needed to merge in your styling work,

Issue #2504, opened 2020-07-29
Crash when using routeLineTracksTraversal and arriving to final destination.


MaximAlien:
**Mapbox Navigation SDK version:** `1.0.0-beta.1`

When using `routeLineTracksTraversal` and arriving to final destination crash can occur sometimes. Here's stack trace:
[Example_2020-07-29-130802_MacBook-Pro.txt](https://github.com/mapbox/mapbox-navigation-ios/files/4997220/Example_2020-07-29-130802_MacBook-Pro.txt)
```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '‘stops’ dictionary argument to ‘mgl_interpolate:withCurveType:parameters:stops:’ function must not be empty.'
terminating with uncaught exception of type NSException
abort() called
CoreSimulator 725.10 - Device: iPhone 11 (DB6D57A8-946D-4970-8FA3-B043629D86CD) - Runtime: iOS 13.5 (17F61) - DeviceType: iPhone 11

Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fff23e3cf0e __exceptionPreprocess + 350
1   libobjc.A.dylib                     0x00007fff50ba89b2 objc_exception_throw + 48
2   CoreFoundation                      0x00007fff23e3cd4c +[NSException raise:format:] + 188
3   Mapbox                              0x0000000106452aa1 -[NSExpression(MGLAdditions) mgl_jsonInterpolationExpressionObject] + 1321
4   Mapbox                              0x0000000106450ad2 -[NSExpression(MGLAdditions) mgl_jsonExpressionObject] + 10432
5   Mapbox                              0x0000000106469dab _ZN24MGLStyleValueTransformerIN4mbgl5ColorEU8__strongP7UIColorS1_S4_E15toPropertyValueINS0_5style22ColorRampPropertyValueEEENSt3__19enable_ifIXsr3std7is_sameIT_S8_EE5valueESB_E4typeEP12NSExpression + 71
6   Mapbox                              0x000000010647ad33 -[MGLLineStyleLayer setLineGradient:] + 104
[...]

PR #2506, opened 2020-07-29
Fix vanishing route line arrival crash.

MaximAlien:
Closing #2504.


MaximAlien:
> Did I understand correctly that issue was related to creating line gradients without gradient stops?

Yes, sometimes filtered gradient stops are empty, this is a no go when creating `NSExpression`.

PR #2509, opened 2020-07-30
Remove offline

IodaMikeMapbox:
In preparation for the OfflineService to land in the Maps and Nav native libs, and platform delivery in Nav SDK v1.1, we should remove all of the legacy offline APIs that were implemented as part of the Nav UI SDK v0.x.

The pr removes:
 - `NavigationDirections`
 - offline settings from the example app, but leaves settings screen. Settings button will be disabled if there is no settings to show on the settings screen

Resolves mapbox/navigation-sdks#490


Example/ViewController.swift
```diff
@@ -92,6 +92,8 @@ class ViewController: UIViewController {
         super.viewDidLoad()
         
         navigationItem.rightBarButtonItem = UIBarButtonItem(image: UIImage(named: "settings"), style: .plain, target: self, action: #selector(openSettings))
+
+        navigationItem.rightBarButtonItem?.isEnabled = SettingsViewController.numberOfSettings > 0
```
Udumft, line 96: As I understood, this is done to temporary(?) disable settings. At the same time `numberOfSettings` is a let constant defined as zero. So, this condition will never be true. Maybe we should instead rely here on `SettingsViewController.settings().isEmpty` instead? Thus it will be automatically enabled when there are settings available.

Example/ViewController.swift
```diff
@@ -311,7 +315,7 @@ class ViewController: UIViewController {
     func navigationService(route: Route, options: RouteOptions) -> NavigationService {
         let simulate = simulationButton.isSelected
         let mode: SimulationMode = simulate ? .always : .onPoorGPS
-        return MapboxNavigationService(route: route, routeOptions: options, directions: Settings.directions, simulating: mode)
+        return MapboxNavigationService(route: route, routeOptions: options, simulating: mode)
```
Udumft, line 318: `directions:` argument here is removed because it has default value of `Directions.shared`, which is the same as `Settings.directions`. This maintains the existing logic, but I see there are 2 ways:
(1) either we get rid of `Settings` at all since it is used only once and does nothing except introducing a synonym to existing singleton, or
(2) we preserve `Settings` for the future use, but in this case we should also preserve it in this line.

Example/ViewController.swift
IodaMikeMapbox: I thought about using `SettingsViewController.settings().isEmpty` and decided to add a property because the `settings` method in that case creates setting items which won't be used, and that's why in theory it can be too heavy just for getting the number of settings

Udumft: Good point. Maybe then rename it to `numberOfSections` to show it's relation to `sections()` below?

MaximAlien:
Also just to make sure: I can see that there are several offline related entries (OFFLINE_) in `Localizable.strings`. We'll use the same ones in upcoming offline service implementation?

1ec5:
> Also just to make sure: I can see that there are several offline related entries (OFFLINE_) in `Localizable.strings`. We'll use the same ones in upcoming offline service implementation?

I think most of them won’t be relevant to the offline service feature, since the user won’t have the opportunity to choose an arbitrary bounding box to download.

asinghal22:
I am not sure why this is part of v1.0 scope. Removing legacy should be part of v1.1. Am I missing something here. @1ec5

Example/ViewController.swift
IodaMikeMapbox: renamed

Example/ViewController.swift
IodaMikeMapbox: removed `Settings` file

IodaMikeMapbox:
> I can see that `let MBSelectedOfflineVersion = "MBSelectedOfflineVersion"` is never used in Example. Do we still need it?

removed

> Also just to make sure: I can see that there are several offline related entries (OFFLINE_) in Localizable.strings. We'll use the same ones in upcoming offline service implementation?

removed

IodaMikeMapbox:
> I am not sure why this is part of v1.0 scope. Removing legacy should be part of v1.1. Am I missing something here. @1ec5

@asinghal22 
We should remove all of the legacy offline APIs that were implemented as part of the Nav UI SDK v0.x. This will give us a clear marketing strategy and avoid confusion/breaking changes when v1.1 rolls out. What's important, those legacy features were never released publicly, so we are not lowering the bar.

For more info please check mapbox/navigation-sdks#484

asinghal22:
> We should remove all of the legacy offline APIs that were implemented as part of the Nav UI SDK v0.x. This will give us a clear marketing strategy and avoid confusion/breaking changes when v1.1 rolls out. What's important, those legacy features were never released publicly, so we are not lowering the bar.

How about this link?

https://docs.mapbox.com/ios/navigation/overview/offline-routing/#offline-routing-in-the-navigation-sdk-example-app

1ec5:
@asinghal22 is correct, we did publicly release a version of offline navigation on iOS. What we never released on iOS was the _hybrid_ navigation feature. In any event, if we don’t remove this offline navigation functionality ahead of v1.0, we won’t be able to remove it until v2.0, and that may prevent us from introducing a planned revamped offline navigation feature in v1.1.

MapboxCoreNavigationTests/OfflineRoutingTests.swift
```diff
@@ -1,130 +0,0 @@
-import XCTest
-import MapboxDirections
-import TestHelper
-@testable import MapboxCoreNavigation
-
-class OfflineRoutingTests: XCTestCase {
-    func testOfflineDirections() {
-        let bundle = Bundle(for: Fixture.self)
-        let tilesURL = URL(fileURLWithPath: bundle.bundlePath.appending("/tiles/liechtenstein"))
```
1ec5, line 9: #2546 proposes reusing this test fixture, if not the overall test, to facilitate automated integration testing of the PassiveLocationDataSource in free driving mode.

Example/ViewController.swift
```diff
@@ -578,7 +557,7 @@ extension ViewController: VisualInstructionDelegate {
 // MARK: Free driving
 extension ViewController {
     func trackLocations(mapView: NavigationMapView) {
-        let dataSource = PassiveLocationDataSource(directions: Settings.directions)
+        let dataSource = PassiveLocationDataSource(directions: Directions.shared)
```
1ec5, line 560: Nit: This is the default value already:

```suggestion
        let dataSource = PassiveLocationDataSource()
```

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -2,14 +2,16 @@ import Foundation
 import MapboxDirections
 import MapboxNavigationNative
 
-/**
+/** :nodoc:
  A closure to call when the `NavigationDirections` router has been configured completely.
  */
+@available(*, deprecated, message: "Not recommended to use")
```
1ec5, line 8: Let’s give no message for these deprecations. Deprecating a symbol already means we don’t recommend using it. In theory, we could say that offline navigation functionality will come back in v1.1, but it’s best not to make forward-looking statements in the API, and chances are the API will look a little different anyways.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -58,30 +60,34 @@ public enum OfflineRoutingError: LocalizedError {
     }
 }
 
+/// :nodoc:
```
1ec5, line 63: Nit: `:nodoc:` is only needed for public symbols. jazzy doesn’t attempt to include private symbols in the docset anyways.

MapboxCoreNavigation/OfflineDirections.swift
```diff
@@ -92,13 +98,15 @@ public typealias UnpackCompletionHandler = (_ numberOfTiles: UInt64, _ error: Er
...
 
-/**
+/** :nodoc:
  A `NavigationDirections` object provides you with optimal directions between different locations, or waypoints. The directions object passes your request to a built-in routing engine and returns the requested information to a closure (block) that you provide. A directions object can handle multiple simultaneous requests. A `RouteOptions` object specifies criteria for the results, such as intermediate waypoints, a mode of transportation, or the level of detail to be returned. In addition to `Directions`, `NavigationDirections` provides support for offline routing.
  
  Each result produced by the directions object is stored in a `Route` object. Depending on the `RouteOptions` object you provide, each route may include detailed information suitable for turn-by-turn directions, or it may include only high-level information such as the distance, estimated travel time, and name of each leg of the trip. The waypoints that form the request may be conflated with nearby locations, as appropriate; the resulting waypoints are provided to the closure.
  */
+@available(*, deprecated, message: "Not recommended to use")
```
1ec5, line 109: For this particular deprecation, let’s recommend Directions as an alternative. It doesn’t do offline routing, but at least it does routing.

```suggestion
@available(*, deprecated, message: "Use the Directions class instead.")
```

PR #2402, opened 2020-06-15
[Guidance Cards][Part 2]Defer to UIAppearance proxy for color/font literals

nishant-karajgikar:
Builds on #2401 

This PR changes the UI logic in the `InstructionCardContainerView` and `InstructionCardView` to defer to the color/font literals supplied by the UIAppearance proxy.

PR #2510, opened 2020-07-30
[Guidance Cards][Part 4] Remove InstructionsCardStyle class

avi-c:
Completes the change to UIAppearance proxy based styling for Guidance Cards.

Depends on #2402.


avi-c:
Removes all references and implementation for the old styling mechanism now that all existing properties can be styled via UIAppearance.

1ec5:
Some of the tests are still referring to InstructionsCardStyle:

```
/Users/distiller/project/MapboxNavigationTests/InstructionsCardCollectionTests.swift:166:34: error: use of undeclared type 'InstructionsCardStyle'
class TestInstructionsCardStyle: InstructionsCardStyle {
                                 ^~~~~~~~~~~~~~~~~~~~~
/Users/distiller/project/MapboxNavigationTests/InstructionsCardCollectionTests.swift:130:53: error: use of undeclared type 'DayInstructionsCardStyle'
        XCTAssertTrue(instructionsCardView.style is DayInstructionsCardStyle)
                                                    ^~~~~~~~~~~~~~~~~~~~~~~~
/Users/distiller/project/MapboxNavigationTests/InstructionsCardCollectionTests.swift:132:30: error: value of type 'InstructionsCardView' has no member 'style'
        instructionsCardView.style = TestInstructionsCardStyle()
        ~~~~~~~~~~~~~~~~~~~~ ^~~~~
/Users/distiller/project/MapboxNavigationTests/InstructionsCardCollectionTests.swift:133:44: error: value of type 'InstructionsCardView' has no member 'style'
        XCTAssertTrue(instructionsCardView.style is TestInstructionsCardStyle)
                      ~~~~~~~~~~~~~~~~~~~~ ^~~~~
```

avi-c:
Looks like tests fail because I removed the style classes. Looking at the test, it's not very comprehensive in that it only sets the style value of the instruction card view and then checks that it was set, no UI tests or anything.

@1ec5 -  Is it reasonable just to remove the automated test and rely on manual QA that the styling is possible?

1ec5:
> Looks like tests fail because I removed the style classes. Looking at the test, it's not very comprehensive in that it only sets the style value of the instruction card view and then checks that it was set, no UI tests or anything.
> Is it reasonable just to remove the automated test and rely on manual QA that the styling is possible?

Yeah, that’s reasonable given your observation that the existing tests weren’t doing much more than exercising some lines of code. I think we should be fine in terms of code coverage too, since the code is basically being removed in this PR.

PR #2529, opened 2020-08-06
Expose the corner radius, border width and height properties to guidance card styling

avi-c:
Make public some properties that were incorrectly kept private to Nav SDK.


avi-c:
@1ec5 - Is the timeout failure for the unrelated test case ok to just ignore?

avi-c:
@Udumft or @1ec5 - can you override the need for the checks to pass? I don't have sufficient access to merge without all the checks succeeding.

Udumft:
There is a [fix](https://github.com/mapbox/mapbox-navigation-ios/pull/2532) for this issue already merged in `master`. You could merge in/rebase on top of that to make CI green.

PR #2532, opened 2020-08-07
Tests fix

### LINKED ISSUE #2525: MapboxNavigationTests.SKUTests.testSKUTokensMatch fails intermittently
https://github.com/mapbox/mapbox-navigation-ios/pull/2408#issuecomment-666664896, https://github.com/mapbox/mapbox-navigation-ios/pull/2477#issuecomment-668943748, and https://github.com/mapbox/mapbox-navigation-ios/pull/2522#issuecomment-669425464 all have examples of the test case `-[MapboxNavigationTests.SKUTests testSKUTokensMatch]` failing intermittently on CI.

I think this failure is unrelated to the symbol collisions we were seeing earlier in https://github.com/mapbox/mapbox-navigation-ios/pull/2477#issuecomment-666043451, because the other SKU token tests are passing. Instead, I suspect that the messages about “Unbalanced calls to begin/end appearance transitions” and “Failed to bind EAGLDrawable” indicate that the map view or its containing view controller haven’t completely initialized yet. It could be a race condition in the test or a problem with the example application that only crops up when run headlessly. But hopefully it doesn’t indicate a problem with the SKU token functionality itself.

/cc @mapbox/navigation-ios

Udumft:
Resolves #2525 
Attempt to fix `SKUTokensMatch` and `testNavigationEventDetailsGlobalQueue` tests.

I could not reproduce the SKU test failure locally, so had to test it on the CI.
I also increased the timeout period for `testNavigationEventDetailsGlobalQueue` because it failed sometimes , but this test is not about performance measurement. It uses background queue and sometimes it fails to fall into strict `0.1` timeout.

### DIFF
--- a/MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
+++ b/MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
@@ -75,6 +75,7 @@ class NavigationEventsManagerTests: XCTestCase {
             CLLocationCoordinate2D(latitude: 38.853108, longitude: -77.043331),
             CLLocationCoordinate2D(latitude: 38.910736, longitude: -76.966906),
         ])
+        let eventTimeout = 0.3
         let route = Fixture.route(from: "DCA-Arboretum", options: routeOptions)
         let dataSource = MapboxNavigationService(route: route, routeOptions: routeOptions)
         let sessionState = SessionState(currentRoute: route, originalRoute: route)
@@ -87,7 +88,7 @@ class NavigationEventsManagerTests: XCTestCase {
             expectation.fulfill()
         }
         
-        wait(for: [expectation], timeout: 0.3)
+        wait(for: [expectation], timeout: eventTimeout)
         
         // Sanity check to verify that no issues occur when creating NavigationEventDetails from main queue.
         let _ = NavigationEventDetails(dataSource: dataSource, session: sessionState, defaultInterface: false)
--- a/MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
+++ b/MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
@@ -87,7 +87,7 @@ class NavigationEventsManagerTests: XCTestCase {
             expectation.fulfill()
         }
         
-        wait(for: [expectation], timeout: 0.1)
+        wait(for: [expectation], timeout: 0.3)
         
         // Sanity check to verify that no issues occur when creating NavigationEventDetails from main queue.
         let _ = NavigationEventDetails(dataSource: dataSource, session: sessionState, defaultInterface: false)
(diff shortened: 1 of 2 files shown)


MapboxNavigationTests/Support/TokenTestViewController.swift
```diff
@@ -45,10 +45,11 @@ class TokenTestViewController: UIViewController {
         super.viewDidLoad()
         
         DispatchQueue.global().async {
-            self.directionsToken = Directions.skuToken
-            self.speechSynthesizerToken = SpeechSynthesizer.skuToken
             
             _ = self.semaphore.wait(timeout: .now() + 4)
+
+            self.directionsToken = Directions.skuToken
+            self.speechSynthesizerToken = SpeechSynthesizer.skuToken
```
Udumft, line 52: It results that polling these `skuToken`s in different moments of time fetches different token values...
That makes me think if that may result in single application be counted as 2 (or more?) users with different tokens? Could not come up with correct way to check that.

MapboxNavigationTests/Support/TokenTestViewController.swift
```diff
@@ -45,10 +45,11 @@ class TokenTestViewController: UIViewController {
         super.viewDidLoad()
         
         DispatchQueue.global().async {
```
IodaMikeMapbox, line 47:
```suggestion
        DispatchQueue.global().asyncAfter(deadline: .now() + 4) {
```

MapboxNavigationTests/Support/TokenTestViewController.swift
```diff
@@ -45,10 +45,11 @@ class TokenTestViewController: UIViewController {
...
-            self.directionsToken = Directions.skuToken
-            self.speechSynthesizerToken = SpeechSynthesizer.skuToken
             
             _ = self.semaphore.wait(timeout: .now() + 4)
```
IodaMikeMapbox, line 49:
```suggestion

```

MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
```diff
@@ -87,7 +87,7 @@ class NavigationEventsManagerTests: XCTestCase {
             expectation.fulfill()
         }
         
-        wait(for: [expectation], timeout: 0.1)
```
dersim-davaod, line 90: can we extract it into the constant?

MapboxNavigationTests/Support/TokenTestViewController.swift
dersim-davaod: Additional note: I'd prefer explicit comment regarding `asyncAfter(deadline: .now() + 4)` code here.

Udumft: Semaphore approach allows test to be potentially finished faster than 4 seconds, so I thing this variant is better

PR #2535, opened 2020-08-10
Building extrusion highlights

d-prukop:
Added the ability to highlight buildings in the NavigationMapView using a geographic coordinate.


MapboxNavigation/NavigationViewController.swift
```diff
@@ -147,6 +147,11 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
      */
     public var shouldManageApplicationIdleTimer = true
     
+    /**
+     Coordinate which is used to determine what associated building extrusion should be highlighted.
+     */
+    public var buildingExtrusionCoordinate: CLLocationCoordinate2D? = nil
```
1ec5, line 153: Since a single route can have multiple destinations, we’ll need to refactor this code so that NavigationViewController exposes only a feature flag and the logic to choose a coordinate lives inside NavigationMapView (based on a waypoint or route progress being passed in).

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -25,6 +25,8 @@
 		16E3625C201265D600DF0592 /* ImageDownloadOperationSpy.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16E3625B201265D600DF0592 /* ImageDownloadOperationSpy.swift */; };
 		16EF6C1E21193A9600AA580B /* CarPlayManagerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16EF6C1D21193A9600AA580B /* CarPlayManagerTests.swift */; };
 		16EF6C22211BA4B300AA580B /* CarPlayMapViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16EF6C21211BA4B300AA580B /* CarPlayMapViewController.swift */; };
+		1F25A90F24A52C1600737F01 /* URLSession.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F25A90E24A52C1600737F01 /* URLSession.swift */; };
+		1FFDFD92249C1AA80091746A /* JunctionView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FFDFD91249C1AA70091746A /* JunctionView.swift */; };
```
1ec5, line 29: Looks like some merging happened here, but I’m not seeing a merge commit. We might need to rebase this PR just to be sure it won’t undo any changes that have already landed.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1357,114 @@ extension NavigationMapView {
...
+            unhighlightBuildingExtrusions()
+        }
+    }
+    
+    public func unhighlightBuildingExtrusions() {
+        extrudeBuildings([(0, buildingExtrusionColor)], extrudeAll: true)
+    }
+    
+    private func addStyleLayersIfNecessary() {
+        if let buildingsSource = style?.source(withIdentifier: "composite") {
```
1ec5, line 1376: It’s possible for the style to incorporate the Mapbox Streets source explicitly without using source compositing. In that case, the source identifier could be something arbitrary. The current road name label feature dynamically finds the Mapbox Streets source, then adds it only if it isn’t already part of the style:

https://github.com/mapbox/mapbox-navigation-ios/blob/3f361bb8735bf2671dc302484e75feb964976a89/MapboxNavigation/RouteMapViewController.swift#L645-L656

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1357,114 @@ extension NavigationMapView {
...
+                style?.addLayer(highlightedBuildingLayer)
+            }
+        }
+    }
+
+    private func buildingId(coordinate: CLLocationCoordinate2D) -> Int? {
+        let screenCoordinate = convert(coordinate, toPointTo: self)
+        let features = visibleFeatures(at: screenCoordinate)
+        
+        if let feature = features.first, let buildingId = feature.identifier as? Int {
```
1ec5, line 1398: By default, `MGLMapView.visibleFeatures(at:)` returns visible features in all layers at the screen coordinate, sorted from top to bottom in z order. However, this code considers only the first returned feature, regardless of whether it’s a building. This could cause the method to fail if, for instance, a POI label or destination annotation happens to overlap the query coordinate.

There are a couple ways to make this code more robust:

* Filter `MGLStyle.layers` to layers whose `sourceLayerIdentifier` is [`building`](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#building), then pass those layers into `visibleFeatures(at:styleLayerIdentifiers:)`. But this assumes the style shows buildings to begin with, which isn’t the case for Satellite Streets or some other likely styles.
* Add a faintly visible building layer to the style, not really visible to the naked eye but opaque enough for feature querying to find it. For example, this is how the current road name feature makes road name querying more robust using white lines: https://github.com/mapbox/mapbox-navigation-ios/blob/3f361bb8735bf2671dc302484e75feb964976a89/MapboxNavigation/RouteMapViewController.swift#L658-L665
* Filter the returned features for feature attributes that one would expect on a building, such as `extrude`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1357,114 @@ extension NavigationMapView {
...
+        if let highlightedBuildingFillExtrusionLayer = style?.layer(withIdentifier: StyleLayerIdentifier.buildingExtrusion) as? MGLFillExtrusionStyleLayer {
+            var predicate = "extrude = 'true' && type = 'building' && underground = 'false'"
+            if extrudeAll == false {
+                // form a predicate to filter out the other buildings from the datasource so only the desired ones are included
+                var identifiersBoolean = "($featureIdentifier == \(buildings.first!.0)"
+                
+                if buildings.count > 1 {
+                    for building in buildings[1...buildings.count-1] {
+                        identifiersBoolean += " || $featureIdentifier == \(building.0)"
+                    }
```
1ec5, line 1421: This can be written more succinctly as:

```suggestion
                    for (featureIdentifier, _) in buildings {
                        identifiersBoolean += " || $featureIdentifier == \(featureIdentifier)"
                    }
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1357,114 @@ extension NavigationMapView {
...
+                }
+                
+                identifiersBoolean += ")"
+                
+                predicate += " && \(identifiersBoolean)"
+            }
+            
+            highlightedBuildingFillExtrusionLayer.predicate = NSPredicate(format: predicate)
```
1ec5, line 1429: This can be written more succinctly as:

```suggestion
            if extrudeAll {
                highlightedBuildingFillExtrusionLayer.predicate = NSPredicate(format: "extrude = 'true' AND type = 'building' AND underground = 'false'")
            } else {
                // Form a predicate to filter out the other buildings from the datasource so only the desired ones are included.
                highlightedBuildingFillExtrusionLayer.predicate = NSPredicate(format: "extrude = 'true' AND type = 'building' AND underground = 'false' AND $featureIdentifier IN %@", buildings.map { $0.0 })
            }
```

avi-c:
Do we want to squash the commit history?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1357,114 @@ extension NavigationMapView {
...
+            
+            let fillExtrusionHeightStops = [0: NSExpression(forConstantValue: 0),
+                                            15: NSExpression(forConstantValue: 0),
+                                            15.25: NSExpression(format: highlightedBuildingHeightExpression)]
+            
+            let fillExtrusionBaseStops = [0: NSExpression(forConstantValue: 0),
+                                          15: NSExpression(forConstantValue: 0),
+                                          15.25: NSExpression(forKeyPath: "min_height")]
+            
+            let opacityStops = [15: 0.5, 17: 0.8]
```
1ec5, line 1462: According to these stops, the buildings would be completely flat until z15, then grow to a realistic height by z15.25. Is the zoom level at the time of arrival guaranteed to be at least z15? Do we need it to depend on the zoom level?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1357,114 @@ extension NavigationMapView {
...
+            // the rest of the buildings will be extruded but kept at a uniform color
+            var colorsList = [UIColor]()
+            var highlightedBuildingHeightExpression = "MGL_MATCH($featureIdentifier, "
+            var highlightedBuildingColorExpression = "MGL_MATCH($featureIdentifier, "
+            
+            for building in buildings {
+                highlightedBuildingHeightExpression += "\(building.0), height, "
+                highlightedBuildingColorExpression += "\(building.0), %@, "
+                colorsList.append(building.1)
+            }
```
1ec5, line 1441: We can simplify these expressions based on an assumption that there’s only two sets of buildings: the buildings we want to highlight and everything else.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1357,114 @@ extension NavigationMapView {
...
+        let features = visibleFeatures(at: screenCoordinate)
+        
+        if let feature = features.first, let buildingId = feature.identifier as? Int {
+            return buildingId
+        }
+        
+        return nil
+    }
+    
+    private func extrudeBuildings(_ buildings: [(Int, UIColor)], extrudeAll: Bool) {
```
1ec5, line 1405: I don’t think we can assume that the building ID is integral or that it fits into 32&nbsp;bits on 32-bit architectures. The predicates in this method would work regardless of the feature identifier’s type; strings would work too, as long as we pass the identifiers into `NSPredicate(format:arguments:)` instead of using string interpolation. But even if we were to assume that the identifiers are integral, it would be possible for them to overflow the `Int` type. [`UInt64` or `Int64`](https://docs.mapbox.com/ios/api/maps/6.0.0/working-with-geojson-data.html) would be a safer representation.

MapboxNavigation.xcodeproj/project.pbxproj
MaximAlien: Rebase was already made. This part was added by Dave in one of commits when `BuildingExtrusionStyler.swift` was added to the project. In any case I don't consider it too critical, since the same files were added back in the same commit.

MapboxNavigation/NavigationViewController.swift
MaximAlien: I'm not sure we can get away with feature flag to do this. Since in most cases route destination coordinate is not equal to highlighted building coordinate we'll have to store highlighted building coordinates separately. In such case probably list with separate coordinates will have to be used (along with route which is passed to `NavigationViewController`). Or am I missing something here?

MapboxNavigation/NavigationMapView.swift
MaximAlien: I'm not author of initial implementation, but seems that we can safely skip first gradient stop. @avi-c, do you maybe remember what was the original idea?
```suggestion
            let fillExtrusionHeightStops = [15: NSExpression(forConstantValue: 0),
                                            15.25: NSExpression(format: highlightedBuildingHeightExpression)]
            
            let fillExtrusionBaseStops = [15: NSExpression(forConstantValue: 0),
                                          15.25: NSExpression(forKeyPath: "min_height")]
            
            let opacityStops = [15: 0.5, 17: 0.8]
```

MapboxNavigation/NavigationMapView.swift
MaximAlien: Turns out that it's possible to add only one `MGLFillExtrusionStyleLayer` with `sourceLayerIdentifier` set to `building`. In case if extrusion layer has different ID buildings will not be extruded. This means that only 3rd approach is fine in this case. I've added filtering which takes into account `extrude` and `type` attributes.

MapboxNavigation/NavigationMapView.swift
d-prukop: This is meant to run on items in `buildings` after the item at index 0. I think this would duplicate the first item.

MaximAlien:
Here are some results which contain changes made by Dave and arrival tweaks:
output output2

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1389,130 @@ extension NavigationMapView {
...
+        }
+        
+        if let feature = features.first, let buildingId = feature.identifier as? UInt64 {
+            return buildingId
+        }
+        
+        return nil
+    }
+    
+    private func highlightBuildings(_ buildings: [(identifier: UInt64, highlightColor: UIColor)]?, in3D: Bool, extrudeAll: Bool) {
```
MaximAlien, line 1477: Since `extrudeAll` feature is not really supported I propose to remove it completely. If needed we can add it in future, but for now it'll simplify current logic.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -147,13 +147,30 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
+    public var highlightBuildingsIn3D: Bool = true
+    
     var isConnectedToCarPlay: Bool {
         if #available(iOS 12.0, *) {
             return CarPlayManager.isConnected
         } else {
             return false
         }
     }
+    var buildingHighlightCoordinates: [CLLocationCoordinate2D] = []
```
MaximAlien, line 173: I don't see that `buildingHighlightCoordinates` is used anywhere, do we really need it?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1389,130 @@ extension NavigationMapView {
...
+                highlightedBuildingFillExtrusionLayer.predicate = NSPredicate(format: "extrude = 'true' AND type = 'building' AND underground = 'false' AND $featureIdentifier IN %@", buildings!.map { $0.identifier })
+                opacityStops = [13: 0.5, 17: 0.8]
+            }
+            
+            // we have some specific building IDs that we will be the highlight color
+            // the rest of the buildings will be extruded but kept at a uniform color
+            
+            var highlightedBuildingHeightExpression = "MGL_MATCH($featureIdentifier, -1, height, \(extrudeAll ? "height" : "0"))"
+            var highlightedBuildingColorExpression = "MGL_MATCH($featureIdentifier, -1, %@, %@)"
+            var colorsList: [UIColor] = [.green]
```
MaximAlien, line 1494: Not sure it's a good idea to hardcode `.green` in list.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -125,6 +127,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     @objc dynamic public var traversedRouteColor: UIColor = .defaultTraversedRouteColor
     @objc dynamic public var maneuverArrowColor: UIColor = .defaultManeuverArrow
     @objc dynamic public var maneuverArrowStrokeColor: UIColor = .defaultManeuverArrowStroke
+    @objc dynamic public var buildingColor: UIColor = .defaultBuildingColor
```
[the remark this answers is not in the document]
d-prukop: I'd like to keep this here but I'll unify the naming around `defaultBuildingColor`

MapboxNavigation/NavigationMapView.swift
```diff
@@ -232,6 +236,25 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
 
     private lazy var routeGradient = [CGFloat: UIColor]()
     
+    /**
+     This property enables the ability to highlight buildings.
+     
+     Use the `highlightBuildings(for coordinates:, in3D:)` function to perform building highlighting.
+     */
+    public var buildingHighlightingEnabled: Bool = false {
```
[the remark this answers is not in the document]
d-prukop: I'd need to see an example of how this will work.

MapboxNavigation/NavigationMapView.swift
d-prukop: I'd like to keep this capability in for further design and UX experimentation

MapboxNavigation/NavigationMapView.swift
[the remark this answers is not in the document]
d-prukop: Actually changed it to `buildingDefaultColor` to better align with `buildingHighlightColor`

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1389,130 @@ extension NavigationMapView {
...
+    /**
+     Removes the highlight from all buildings highlighted by `highlightBuildings(for coordinates:, in3D:)`.
+     */
+    public func unhighlightBuildings() {
+        highlightBuildings(nil, in3D: false, extrudeAll: false)
+    }
+    
+    private func highlightBuildings(with attributes: [BuildingHighlightAttributes], in3D: Bool) {
+        let buildingAttributes = attributes.map {(value) -> BuildingHighlightAttributes in
+            var output = value
```
MaximAlien, line 1421: I guess that return result is not really used anywhere. Maybe it'd be better to simplify current logic to stop using `BuildingHighlightAttributes` for now? We can handle edge cases like rendering highlighted building only when they appear on screen as a tail work.

MaximAlien:
Overall this looks good to me. We expose public `highlightDestinationBuildings` property in `NavigationViewController` which allows to highlight buildings upon arrival. We also give the ability to call `highlightBuildings(for:in3D:)` and `unhighlightBuildings` methods inside `NavigationMapView`. There are still some issues I've found:
1. Issue with zooming in when arriving to intermediate waypoint. When arriving to final destination zoom-in works correctly.
1. Issue with adding layer back in case of style updates (marked by FIXME).
1. Issue with not highlighting buildings when they're not currently visible on map.
1. Crash when adding final destination to the same building (can reproduce it only when running internal sample).
1. `ViewController.swift` - in case if we get waypoints limit buildings are not removed.
1. Flickering in some cases when pressing map view.

They can be taken care in scope of other PRs.

PR which adds showcase sample can be found [here](https://github.com/mapbox/navigation-ios-examples/pull/71/). I'll also add it to documentation as well.

@1ec5, please feel free to take a look and let us know whether you have any comments. Many thanks to @d-prukop for hard work on this PR.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1378,130 @@ extension NavigationMapView {
...
+    }
+    
+    /**
+     Removes the highlight from all buildings highlighted by `highlightBuildings(for coordinates:, in3D:)`.
+     */
+    public func unhighlightBuildings() {
+        highlightBuildings(nil)
+    }
+    
+    private func highlightBuildings(with attributes: [BuildingHighlightAttributes], in3D: Bool) {
```
Udumft, line 1408: What is the motivation to keep `highlightBuildings(...)` methods family and `buildingId(coordinate:)` private? Those might be not as convenient as above, but it may be handy for users who need finer control over highlighting.

Example/ViewController.swift
```diff
@@ -247,6 +262,12 @@ class ViewController: UIViewController {
         let navigationViewController = NavigationViewController(for: route, routeOptions: routeOptions, navigationOptions: options)
         navigationViewController.delegate = self
         
+        // Example of building highlighting using the NavigationViewController.
+        navigationViewController.highlightDestinationBuildings = true
+        
+        // Extrude buildings in 2D.
```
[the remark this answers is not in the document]
MaximAlien: I think in this case idea is to use 2D and pass false to `highlightBuildingsIn3D` so comment is correct.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -289,6 +299,19 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         if (tracksUserCourse) {
             updateCourseTracking(location: userLocationForCourseTracking, camera:self.camera, animated: false)
         }
+        
+        styleObservation = self.observe(\.style, options: .new) { [weak self] (mapView, change) in
```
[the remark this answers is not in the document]
MaximAlien: Good catch, added `styleObservation = nil` to `deinit`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1365,129 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        if extrudeAll {
+            highlightedBuildingsLayer.predicate = NSPredicate(format: "extrude = 'true' AND type = 'building' AND underground = 'false'")
+        } else {
+            // Form a predicate to filter out the other buildings from the datasource so only the desired ones are included.
+            highlightedBuildingsLayer.predicate = NSPredicate(format: "extrude = 'true' AND type = 'building' AND underground = 'false' AND $featureIdentifier IN %@", buildings.map { $0.identifier })
+        }
+        
+        // Buildings with IDs will be highlighted with provided color. Rest of the buildings will be highlighted, but kept at a uniform color.
+        let highlightedBuildingHeightExpression = "MGL_MATCH($featureIdentifier, \(buildings.map { "\($0.identifier), \(in3D ? "height" : "0"), " }.joined(separator: ""))\(extrudeAll ? "height" : "0"))"
+        let highlightedBuildingColorExpression = "MGL_MATCH($featureIdentifier, \(buildings.map { "\($0.identifier), %@, " }.joined(separator: ""))%@)"
```
1ec5, line 1468: Avoid using concatenation or string interpolation to build an expression format string (or any kind of format string, for that matter). If the style happened to have a layer whose identifiers were strings, the format string would fail to parse, resulting in a crash.

The first line below avoids the unwieldiness of a match expression, taking advantage of the fact that the values of `in3D` and `extrudeAll` are uniform for all buildings. The second line uses the strongly typed initializer for match expressions to avoid string concatenation.

```suggestion
        let highlightedBuildingHeightExpression = NSExpression(format: "TERNARY(%@ AND (%@ OR $featureIdentifier IN %@), height, 0)", in3D, extrudeAll, buildings.map { $0.identifier })
        let colorsByBuilding = Dictionary(buildings.map { (NSExpression(forConstantValue: $0.identifier), NSExpression(forConstantValue: $0.highlightColor)) }) { (_, last) in last }
        let highlightedBuildingColorExpression = NSExpression(forMGLMatchingKey: NSExpression(forVariable: "featureIdentifier"), in: colorsByBuilding, default: NSExpression(forConstantValue: buildingDefaultColor))
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1367,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        var colorsList = buildings.map { $0.highlightColor }
+        
+        colorsList.append(buildingDefaultColor)
+        
+        let fillExtrusionHeightStops = [0: NSExpression(forConstantValue: 0),
+                                        13: NSExpression(forConstantValue: 0),
+                                        13.25: NSExpression(format: highlightedBuildingHeightExpression)]
```
1ec5, line 1475: I think the purpose of these stops is to make the buildings appear to grow out of the ground when they first appear. But if somehow the camera stays put at z13.1, the buildings will stay short. Can we accomplish the same effect without the edge case by setting `fillExtrusionHeight` to `TERNARY($zoomLevel > 13.25, %@, 0)` and relying on the duration in `fillExtrusionHeightTransition` to make the buildings’ appearance less sudden?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1367,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        highlightedBuildingsLayer.fillExtrusionOpacity = NSExpression(forConstantValue: 0.05)
+        highlightedBuildingsLayer.fillExtrusionHeightTransition = MGLTransition(duration: 0.8, delay: 0)
+        highlightedBuildingsLayer.fillExtrusionOpacityTransition = MGLTransition(duration: 0.8, delay: 0)
+        
+        style?.addLayer(highlightedBuildingsLayer)
+        
+        return highlightedBuildingsLayer
+    }
+
+    private func buildingId(at coordinate: CLLocationCoordinate2D) -> Int64 {
```
1ec5, line 1435: Nit: Spell out “identifier” instead of [abbreviating](https://swift.org/documentation/api-design-guidelines/#avoid-abbreviations) it as “id”. That will also skirt any inconsistency between “ID” and “Id”.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1367,122 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+}
+
+extension NavigationMapView {
+       
+    /**
+     Receives coordinates for searching the map for buildings. If buildings are found, they will be highlighted in 2d or 3d depending on the `in3D` value.
+     `in3D` defaults to true
+     `buildingHighlightingEnabled` must be set to true before this function will work.
+     */
+    public func highlightBuildings(at coordinates: [CLLocationCoordinate2D], in3D extrudesBuildings: Bool = true) {
```
1ec5, line 1396: Should this method return any sort of indication as to whether it found a building that can be extruded? That would enable the caller to stop calling this method on an interval.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -289,6 +299,19 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            guard change.newValue != nil else {
+                return
+            }
+
+            // FIXME: Since building might be already highlighted and we do not have any info
+            // regarding its identification after style change such highlighted building will be lost.
+            
+            // In case if style was changed make sure to remove existing layer so that it can be
+            // re-created in future.
+            self?.removeBuildingsLayer()
```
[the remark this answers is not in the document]
MaximAlien: I remember that I had several crashes when using `highlightedBuildingsLayer` not as member variable, but when accessing via `MGLStyle.layer(withIdentifier:)` directly. In any case now it seems to be working, so I'll remove call to remove layer when changing style, on the other hand observer will remain to handle case when layer has to be re-added.

MapboxNavigation/NavigationMapView.swift
1ec5: Does the Android implementation have the ability to highlight multiple buildings in different colors at the same time? If not, we should ticket that out as tail work in the mapbox/mapbox-navigation-android repository.

MapboxNavigation/NavigationMapView.swift
MaximAlien: I think in initial version it's better to keep API as simple as possible. In case if there is such need we'll expose additional methods.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1380,128 @@ extension NavigationMapView {
...
 }
 
+// MARK: - Building Extrusion Highlights
+
+private struct BuildingHighlightAttributes: Hashable {
+    static let kDefaultIdentifier: Int64 = -1
+    
+    var coordinate: CLLocationCoordinate2D = kCLLocationCoordinate2DInvalid
+    var highlightColor: UIColor = .defaultBuildingHighlightColor
+    var identifier: Int64 = kDefaultIdentifier
```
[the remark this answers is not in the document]
MaximAlien: Yep, using `Int64` already.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Not sure if there is such need at this point. During implementation of some kind of polling mechanism to check whether there are any buildings which must be rendered it can be revised.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1350,3 +1380,128 @@ extension NavigationMapView {
...
+    /**
+     Receives coordinates for searching the map for buildings. If buildings are found, they will be highlighted in 2d or 3d depending on the `in3D` value.
+     `in3D` defaults to true
+     `buildingHighlightingEnabled` must be set to true before this function will work.
+     */
+    public func highlightBuildings(for coordinates: [CLLocationCoordinate2D], in3D: Bool = true) {
+        let buildingAttributes = coordinates
+            // Create building attributes by searching for building ID by provided coordinate.
+            .map({ BuildingHighlightAttributes(coordinate: $0,
+                                               highlightColor: buildingHighlightColor,
```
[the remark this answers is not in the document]
MaximAlien: I think best idea is to keep current initial implementation as simple as possible. I think it's even better to use `Set<Int64>` instead of `Dictionary<Int64, CLLocationCoordinate2D>` since `CLLocationCoordinate2D` is only used to get building identifier.

MapboxNavigation/NavigationMapView.swift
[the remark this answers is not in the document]
MaximAlien: Good idea, changed to optional.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1365,143 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+                                               identifier: buildingId(for: $0)) })
+            // Do not add building attribute, in case if it wasn't found.
+            .filter({ $0.identifier != BuildingHighlightAttributes.kDefaultIdentifier })
+            // Since there might be case in which the same building is highlighted several times, remove duplicated IDs.
+            .withoutDuplicates
```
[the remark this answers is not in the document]
MaximAlien: At the end even `CLLocationCoordinate2D` not really needed, so `Set<Int64>` seems like a better option.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -147,6 +147,18 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
      */
     public var shouldManageApplicationIdleTimer = true
     
+    /**
+     Enables highlighting of the destination building when arriving.
+     */
+    public var highlightDestinationBuildings: Bool = false
+    
+    /**
+     When destination building highlighting is enabled, this option controls whether the building will be presented in 3d or 2d.
```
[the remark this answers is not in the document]
MaximAlien: I don't think that `highlightsBuildingsIn3D` is ignored when `highlightsDestinationBuildings` is enabled, it just doesn't have any effect since `highlightsDestinationBuildings` is set to `false` by default and thus buildings are not extruded.

MapboxNavigation/NavigationMapView.swift
MaximAlien: I did some tests with changed `fillExtrusionHeight` and do not see any significant difference. Maybe @d-prukop or @avi-c can comment on this one?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -289,6 +295,11 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         if (tracksUserCourse) {
             updateCourseTracking(location: userLocationForCourseTracking, camera:self.camera, animated: false)
         }
+    
+        // TODO: Add style observation to be able to re-add highlighted building whenever it happens.
```
MaximAlien, line 299: Turns out removing styleObserver in `deinit` cases crash. Since for now functionality which allows to re-add buildings when style is changed is not ready yet I've removed observer completely and added `TODO` statement =.

1ec5: The map view doesn‘t take care of readding other things like the route line when the style changes, either, so I don’t think readding the buildings should be the map view’s responsibility. RouteMapViewController is currently responsible for readding the route line, so we should piggyback on that style observer to readd the buildings there.

MapboxNavigation/NavigationViewController.swift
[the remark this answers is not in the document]
1ec5: Yes, that’s what I mean by it being ignored. Essentially we have a three-way setting, or a setting that depends on another setting. We should avoid such situations in Swift.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -289,6 +295,11 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        
+        // FIXME: Since building might be already highlighted and we do not have any info
+        // regarding its identification after style change such highlighted building will be lost.
```
1ec5, line 302: If the user approaches the destination at around twilight, the UI style could change from DayStyle to NightStyle, making the map style change as well, causing the building to go away. That could make the map less usable.

We could supplement RouteMapViewController’s style observer with an `NSKeyValueObserving.willChangeValue(forKey:)` implementation that stores the extrusion layer’s predicate and expressions to reapply them after the style change. Better yet, `NavigationViewController.updateMapStyle(_:animated:)` could handle reapplying the buildings.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1353,96 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        }
+        
+        if let feature = features.first, let identifier = feature.identifier as? Int64 {
+            return identifier
+        }
+        
+        return nil
+    }
+    
+    private func highlightBuildings(_ buildingIdentifiers: Set<Int64>, in3D: Bool = false, extrudeAll: Bool = false) {
```
1ec5, line 1413: Building identifiers aren’t buildings, so this first argument should be labeled `withIdentifiers:`.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -572,11 +584,45 @@ extension NavigationViewController: NavigationServiceDelegate {
...
+        let expectedZoomLevel = 16.5
+        if currentZoomLevel < expectedZoomLevel {
+            currentZoomLevel = expectedZoomLevel
+        }
+        
+        let mapView = mapViewController.mapView
+        mapView.setCenter(location.coordinate,
+                          zoomLevel: currentZoomLevel,
+                          direction: location.course,
+                          animated: true,
```
1ec5, line 613: Let’s verify that the camera gets reset correctly when beginning the next leg of the route.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -262,6 +266,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         addGestureRecognizer(mapTapGesture)
         
         installUserCourseView()
+        
+        styleURL = MGLStyle.navigationDayStyleURL
```
1ec5, line 270: This is worth a changelog blurb that NavigationMapView now uses `MGLStyle.navigationDayStyleURL` by default.

1ec5, line 270: This overrides the style URL explicitly passed into `init(frame:styleURL:)` and possibly also `init(coder:)`. I think we should move it into `init(frame:)` and find a way to only conditionally set it in `init(coder:)`. This might be why we hadn’t customized NavigationMapView’s default style URL up to now.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1353,96 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
     }
 }
+
+// MARK: - Building Extrusion Highlighting
+
+extension NavigationMapView {
+       
+    /**
+     Receives coordinates for searching the map for buildings. If buildings are found, they will be highlighted in 2d or 3d depending on the `in3D` value.
+     `in3D` defaults to true
```
1ec5, line 1363: Make this a `- parameter` command.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1353,96 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+     `buildingHighlightingEnabled` must be set to true before this function will work.
```
1ec5, line 1364: Make this a `- precondition:` command. I think the precondition is actually that the setting should be set to `true` well before this method is called; otherwise, it would be convenient for us to simply set it to true at the beginning of this method (which wouldn’t be such a bad idea anyways).

MapboxNavigation/NavigationMapView.swift
MaximAlien: Good catch. There is one really strange issue with new approach though. When I add `let styleURL = decoder.decodeObject(of: [URL.self], forKey: "styleURL") as? URL` anywhere in the code in `NavigationMapView` Xcode gets `Segmentation fault: 11` error during build. I think it'd be better to create separate ticket to handle this, because of mentioned issue and to prevent any side effects.

MapboxNavigation/NavigationMapView.swift
MaximAlien: This comment is actually no longer accurate, I've added ability to control building highlight via `DestinationBuildingHighlightType` enum. On the other hand user can perform building highlight via `NavigationMapView` independently.

MapboxNavigation/NavigationMapView.swift
MaximAlien: I'll move `FIXME` statement to `RouteMapViewController` to find out best way to re-add highlighted building layer there. Will create ticket for that as well.

MapboxNavigation/NavigationViewController.swift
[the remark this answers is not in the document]
MaximAlien: Change two boolean variables to `DestinationBuildingHighlightType`. What do you think about such public API/naming?

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@
   * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteStyleLayerWithIdentifier:source:)` to style alternative routes.
   * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteCasingStyleLayerWithIdentifier:source:)` to style the casing of alternative routes.
   * Fixed an issue where the casing for the main route would not overlap alternative routes. ([#2377](https://github.com/mapbox/mapbox-navigation-ios/pull/2377))
+* Added the NavigationViewController.highlightsDestinationBuildings property and NavigationMapView.highlightBuildings(at:in3D:) method to highlight the destination building in UIColor.defaultBuildingHighlightColor as the user approaches it.
+* Added the NavigationViewController.highlightsBuilidngsIn3D property to additionally extrude the destination building in 3D.
```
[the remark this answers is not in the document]
MaximAlien: This one was changes once again because of public API renaming. Please check new version.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Thanks for idea. I'll create separate ticket to tackle this as a tail-work. More testing will be required to check how extrusion behaves when style was dynamically changed, that is the reason I've added such ability in sample [here](https://github.com/mapbox/navigation-ios-examples/pull/71).

MapboxNavigation/NavigationViewController.swift
```diff
@@ -148,16 +148,29 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     public var shouldManageApplicationIdleTimer = true
     
     /**
-     Enables highlighting of the destination building when arriving.
+     Enum denoting the types of the destination building highlighting on arrival.
      */
-    public var highlightsDestinationBuildings: Bool = false
+    public enum DestinationBuildingHighlightType: Int {
+        /**
+         Do not highlight destination building on arrival.
```
1ec5, line 155: Perhaps say something here about the default fallback. I think we still fall back to displaying a 📍, right?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -148,16 +148,29 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     public var shouldManageApplicationIdleTimer = true
     
     /**
-     Enables highlighting of the destination building when arriving.
+     Enum denoting the types of the destination building highlighting on arrival.
      */
-    public var highlightsDestinationBuildings: Bool = false
+    public enum DestinationBuildingHighlightType: Int {
```
1ec5, line 153: This name is a bit long. What if we call it something like `WaypointStyle` and name its cases `pin`, `building`, and `extrudedBuilding`? Would that capture enough nuance? I think it would also make it easier to support non-building treatments in the future, such as making a particular address label more prominent.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -147,6 +147,31 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
      */
     public var shouldManageApplicationIdleTimer = true
     
+    /**
+     Enum denoting the types of the destination building highlighting on arrival.
+     */
+    public enum DestinationBuildingHighlightType: Int {
```
1ec5, line 153: Not critical, but there might be an appropriate category in the jazzy table of contents for this new symbol.

MapboxNavigation/NavigationViewController.swift
MaximAlien: Such naming looks good to me, it's generic enough to give the ability to extend it with other options other than buildings.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Will skip adding `MGLStyle.navigationDayStyleURL` in this PR to prevent and regression as per this [comment](https://github.com/mapbox/mapbox-navigation-ios/pull/2535#discussion_r470764973). Separate ticket will be created to tackle this.

MapboxNavigation/NavigationViewController.swift
MaximAlien: Yep, added default fallback note.

MapboxNavigation/NavigationViewController.swift
MaximAlien: Added `WaypointType` to `Maps` section in `jazzy.yml`.

MapboxNavigation/NavigationMapView.swift
1ec5: URL is a pure Swift type, so it’s incompatible with `NSCoder.decodeObject(of:forKey:)`. But it should work with `TopLevelDecoder.decode(_:from:)`.

MapboxNavigation/NavigationMapView.swift
1ec5: I hadn’t tried out these expressions before suggesting them. This is probably closer to running without issues:

```swift
NSExpression(format: "TERNARY(%@ = TRUE AND (%@ = TRUE OR $featureIdentifier IN %@), height, 0)", in3D as NSValue, extrudeAll as NSValue, buildings.map { $0.identifier })
```

MaximAlien: This one works, thanks.

MapboxNavigation/NavigationMapView.swift
MaximAlien: I think @d-prukop can provide more info on this one.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1335,3 +1346,95 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+
+    private func buildingIdentifier(at coordinate: CLLocationCoordinate2D) -> Int64? {
+        let screenCoordinate = convert(coordinate, toPointTo: self)
+
+        // To increase a chance of selecting correct building identifier filter out
+        // features which do not contain appropriate attributes.
+        let features = visibleFeatures(at: screenCoordinate).filter {
+            let extrude = $0.attribute(forKey: "extrude") as? String
+            let type = $0.attribute(forKey: "type") as? String
+            return extrude == "true" && type == "building"
```
1ec5, line 1397: @langsmith pointed out that `type` isn’t necessarily `building` for all buildings. In fact, the values are rather freeform, coming from OpenStreetMap’s [`building`](https://wiki.openstreetmap.org/wiki/Key:building) key.

This check is here because the call to `visibleFeatures(at:)` doesn’t filter the results to a specific style layer upfront, and the returned features don’t indicate which style layer they come from: mapbox/mapbox-gl-native#7162. The current road name label’s feature querying code solves this issue by figuring out upfront which style layers contain features from the Mapbox Streets source.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/2535#discussion_r468106341

1ec5: From https://github.com/mapbox/mapbox-navigation-ios/pull/2535#discussion_r468106341:

> Filter `MGLStyle.layers` to layers whose `sourceLayerIdentifier` is [`building`](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#building), then pass those layers into `visibleFeatures(at:styleLayerIdentifiers:)`. But this assumes the style shows buildings to begin with, which isn’t the case for Satellite Streets or some other likely styles.

Showing buildings is now a precondition of this method, so this approach is now feasible.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Interesting finding, on the other hand seems that `TopLevelDecoder.decode(_:from:)` is still part of Combine. In any case I'll consider this approach as well.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Moved to first approach to mirror Android implementation, also removed check type = 'building' to increase chances of extrusion.

MapboxNavigation/NavigationMapView.swift
1ec5: Oops, I meant `KeyedDecodingContainer.decode(_:)`. We need to create a local decoding container in order to decode that key.

PR #2536, opened 2020-08-11
Public API signatures. Style hooks

d-prukop:


MapboxNavigation/NavigationViewController.swift
```diff
@@ -793,3 +800,13 @@ extension NavigationViewController: CarPlayConnectionObserver {
         }
     }
 }
+
+// MARK: - MGLMapViewDelegate
+
+extension NavigationViewController: MGLMapViewDelegate {
```
MaximAlien, line 806: I'm not sure that such approach will work in all cases. Since end user can set `MGLMapViewDelegate` delegate it'll override one which was set internally in `NavigationViewController` and current method will never be called. I did not hear any explicit requirement that we should always show buildings extrusion (even without actual navigation, just when showing `NavigationMapView`), so probably calling `setCenter(_:zoomLevel:direction:animated:completionHandler:)` on mapView will be enough to highlight building correctly after arriving to destination. @1ec5, what do you think?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -147,10 +147,11 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
      */
     public var shouldManageApplicationIdleTimer = true
     
-    /**
-     Coordinate which is used to determine what associated building extrusion should be highlighted.
-     */
-    public var buildingExtrusionCoordinate: CLLocationCoordinate2D? = nil
+    public var highlightDestinationBuildings: Bool = false {
```
MaximAlien, line 150: I think exposing `highlightDestinationBuildings` is a good idea. I wasn't actually aware that `Waypoint` contains `targetCoordinate` which allows to store highlighted buildings coordinates.

Example/ViewController.swift
```diff
@@ -239,13 +236,15 @@ class ViewController: UIViewController {
     func startBasicNavigation() {
         guard let response = response, let route = response.routes?.first, case let .route(routeOptions) = response.options else { return }
         
+        let waypoints = routeOptions.waypoints
```
MaximAlien, line 239: I don't think we need `let waypoints = routeOptions.waypoints` here.

MapboxNavigation/NavigationViewController.swift
1ec5: RouteMapViewController installs a key-value observer for `MGLMapView.style` instead of registering as an MGLMapViewDelegate. I think NavigationViewController could cleanly do the same.

MapboxNavigation/NavigationViewController.swift
1ec5: In that case, we should document that the developer should always pass in a `targetCoordinate`, even if it’s redundant with the `coordinate`, to get highlighted buildings. Hopefully that doesn’t have any downsides.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1455,7 +1487,7 @@ extension NavigationMapView {
             highlightedBuildingColorExpression += "%@)"
             
             // TODO: Verify whether extrusion only for specific building or all buildings is required.
```
MaximAlien, line 1489: If we're choosing path that we can highlight either all or specific building with different colors I think `TODO` statement can be removed.

MaximAlien:
I did some testing and here's output:
1. There are some rendering issues we were previously discussing with Dave. E.g. in case of arrival on iOS device (iPhone SE) building is not highlighted, since this is related to `zoomLevel` there are several ways to fix this:
- increase `zoomLevel` to value which will allow to render all buildings (I think it's around `16.5`).
- increase `zoomLevel` only in case of arrival.
I personally prefer 2nd one, since it's faster. I can also update current PR with some changes related to arrival.
Check to verify whether style was loaded is not really needed I think in such case.
2. At this point we render some buildings when using `NavigationMapView`, would it be better if we highlight buildings only in case of navigation? Or even after arrival? I've noticed performance downgrade in case if we always render buildings.
3. Dave mentioned a case in which after arrival buildings might not be visible on screen. Because of this some kind of poller/listener mechanism should be implemented to continuously check whether there is a need to highlight building. Not sure if we'll be able to implement reliable implementation in upcoming days.

MapboxNavigation/NavigationViewController.swift
d-prukop: Passing in the `coordinate` will most likely result in no building highlight. `coordinate` can be automatically written in if no `targetCoordinate` is supplied.

Example/ViewController.swift
d-prukop: Deleted

MapboxNavigation/NavigationMapView.swift
d-prukop: Deleted

PR #2544, opened 2020-08-17
Add optional second level of detail for user navigation feedback

avi-c:
Add new subclass of FeedbackViewController, FeedbackSubtypeViewController that will display items with more specific detail for each category of user feedback on a navigation session.


MapboxNavigation/FeedbackViewController.swift
```diff
@@ -305,3 +326,225 @@ extension String {
...
+            case .roadClosure(_):
+                return [FeedbackType.roadClosure(subtype: .streetPermanentlyBlockedOff),
+                        FeedbackType.roadClosure(subtype: .roadMissingFromMap),
+                        FeedbackType.roadClosure(subtype: .other)].map { $0.generateFeedbackItem() }
+            }
+        }
+    }
+
+    override var draggableHeight: CGFloat {
+        return 400
```
MaximAlien, line 410: I think using hardcoded height constant prevents showing `reportIssueLabel` at the top of the screen in landscape mode. Maybe it'd be better to use `UIScreen` height for landscape and `400` for portrait?
```suggestion
if UIDevice.current.orientation == .portrait {
    return 400
}

return UIScreen.main.bounds.height
```

MapboxNavigation/FeedbackItem.swift
```diff
@@ -13,18 +13,74 @@ public extension FeedbackType {
...
+        case .routeQuality(.routeNonDrivable):
+            return "Route Non Drivable"
+        case .routeQuality(.routeNotPreferred):
+            return "Route Not Preferred"
+        case .routeQuality(.alternativeRouteNotExpected):
+            return "Alternative Route Not Expected"
+        case .routeQuality(.routeIncludedMissingRoads):
+            return "Route Includes Missing Roads"
+        case .routeQuality(.routeHadRoadsTooNarrowToPass):
+            return "Route Had Roads Too Narrow To Pass"
```
1ec5, line 61: Should these items be in title case, or would sentence case be appropriate for labels that could be fairly long (especially when translated)? Either way, “to” should be lowercased in title case.

MapboxNavigation/FeedbackSubtypeCollectionViewCell.swift
```diff
@@ -0,0 +1,125 @@
...
+            if originalTransform == nil {
+                originalTransform = self.circleView.transform
+            }
+
+            UIView.defaultSpringAnimation(0.3, animations: {
+                if self.isHighlighted {
+                    self.circleView.transform = self.circleView.transform.scaledBy(x: 0.85, y: 0.85)
+                } else {
+                    guard let t = self.originalTransform else { return }
+                    self.circleView.transform = t
```
MaximAlien, line 80:
```suggestion
                    guard let transform = self.originalTransform else { return }
                    self.circleView.transform = transform
```

MapboxNavigation/FeedbackItem.swift
```diff
@@ -13,18 +13,74 @@ public extension FeedbackType {
...
+        case .routeQuality(.other):
+            return "Other"
+        case .illegalRoute(.none):
+            return "Illegal Route"
+        case .illegalRoute(.routedDownAOneWay):
+            return "Routed Down A One Way"
+        case .illegalRoute(.turnWasNotAllowed):
+            return "Turn Was Not Allowed"
```
1ec5, line 69: The HIG doesn’t seem to say it explicitly, but there’s a strong preference for conjunctions to make the text more approachable. “Wasn’t” would be appropriate here.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -305,3 +326,225 @@ extension String {
         return ceil(boundingBox.height)
     }
 }
+
+public class FeedbackSubtypeViewController: FeedbackViewController {
```
MaximAlien, line 330: What about moving `FeedbackSubtypeViewController` into separate file? I think it'd be beneficial for readability since `FeedbackViewController.swift` is pretty large already.

MaximAlien, line 330: Worth updating `User Feedback` section in `jazzy.yml` to contain `FeedbackSubtypeViewController` and modifying `CHANGELOG.md`.

MapboxNavigation/FeedbackViewController.swift
MaximAlien: Just tested on iPad as well and seems that we should also handle additional cases there. E.g. does it make sense to show `FeedbackSubtypeViewController` full screen in both landscape and portrait there to have more items visible on screen?

MapboxNavigation/FeedbackItem.swift
avi-c: sentence case works for me too. Switching to that.

MapboxNavigation/FeedbackItem.swift
```diff
@@ -13,18 +13,74 @@ public extension FeedbackType {
     // TODO: Localize these strings
```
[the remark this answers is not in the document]
avi-c: I believe so.

MapboxNavigation/FeedbackViewController.swift
```diff
@@ -276,7 +283,21 @@ extension FeedbackViewController: UICollectionViewDataSource {
 extension FeedbackViewController: UICollectionViewDelegate {
     public func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
         let item = sections[indexPath.row]
-        send(item)
+
+        if detailedFeedbackEnabled, eventsManager != nil {
+            let feedbackViewController = FeedbackSubtypeViewController(eventsManager: eventsManager!, feedbackType: item.feedbackType)
```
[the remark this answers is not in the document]
avi-c: Done.

MapboxNavigation/FeedbackSubtypeCollectionViewCell.swift
avi-c: We can't use `transform` as it resolves to the view's own `transform` property, but to follow the spirit of your comment I changed `t` to `originalTransform`.

MapboxNavigation/FeedbackViewController.swift
avi-c: Done.

avi-c: Good idea. Done.

MapboxNavigation/FeedbackItem.swift
avi-c: 👍

MapboxNavigation/FeedbackSubtypeCollectionViewCell.swift
```diff
@@ -0,0 +1,125 @@
+//
+//  FeedbackSubtypeCollectionViewCell.swift
+//  MapboxNavigation
+//
+//  Created by Avi Cieplinski on 8/17/20.
+//  Copyright © 2020 Mapbox. All rights reserved.
+//
```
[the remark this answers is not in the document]
avi-c: 👍

MapboxNavigation/FeedbackSubtypeCollectionViewCell.swift
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    public var separatorColor: UIColor = .lightGray {
+        didSet {
+            separatorView.backgroundColor = separatorColor
+        }
+    }
+
+    var longPress: UILongPressGestureRecognizer?
```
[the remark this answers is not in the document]
avi-c: Done

MapboxNavigation/FeedbackViewController.swift
avi-c: I think you are right that fullscreen is best. More simple code-wise and also better for possibly long lists.

Issue #2545, opened 2020-08-17
Route and highlighted building are removed after changing style.


MaximAlien:
After dynamically changing style in `NavigationMapView` current route and highlighted building are getting removed. It is expected that current route and highlighted building remain on `NavigationMapView`. Example:
![output](https://user-images.githubusercontent.com/1496498/90454140-3fa29000-e0a7-11ea-896d-b5c87b2a9dbd.gif)

1ec5:
> It is expected that current route and highlighted building remain on `NavigationMapView`.

By design, changing the map style blows away any map style customizations, including the route line and highlighted building style layers. RouteMapViewController is currently responsible for restoring the route line and intermediate waypoint layers by calling `show(_:legIndex:)`. We should ensure that it also restores the building layer.

By coincidence, this issue doesn’t affect the final destination waypoint, which is marked by a point annotation, because annotations do persist across style changes. mapbox/mapbox-gl-native#6180 tracks enabling style layers to persist across style changes like annotations do.

MaximAlien:
In scope of this fix I've added style observation for `NavigationMapView` as well. Since `NavigationMapView` can be used independently from `NavigationViewController` fix inside `RouteMapViewController` will not cover all cases. Even though in case if style changes several times all subsequent calls to `highlightBuildings(at:in3D:)` will not perform any further work if layer is already present. Please let me know if you have better ideas or see other edge cases here.

1ec5:
> Since `NavigationMapView` can be used independently from `NavigationViewController` fix inside `RouteMapViewController` will not cover all cases.

This is also true of the route line and runtime styling in general. I think it would actually be confusing for some runtime styling to persist but not this specific runtime styling customizations. Better for runtime styling to behave consistently, so that the developer knows to call all the relevant runtime styling methods, including `highlightBuildings(at:in3D:)`, in their `MGLMapViewDelegate.mapView(_:didFinishLoading:)`.

PR #2550, opened 2020-08-18
Fix building highlighting when changing style.

MaximAlien:
Closing #2545.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -258,6 +258,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         installUserCourseView()
         
         styleObservation = observe(\.style, options: .new) { [weak self] (mapView, change) in
+            guard change.newValue != nil else { return }
             guard let self = self else { return }
             
             if let routes = self.routes, let currentRoute = routes.first {
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -125,6 +125,12 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     var altitude: CLLocationDistance
     var routes: [Route]?
     var isAnimatingToOverheadMode = false
+    var styleObservation: NSKeyValueObservation?
+    
+    /**
+     A tuple that represents currently highlighted buildings and their extrusion (either 2D or 3D).
+     */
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -266,6 +276,20 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         addGestureRecognizer(mapTapGesture)
         
         installUserCourseView()
+        
+        styleObservation = observe(\.style, options: .new) { [weak self] (mapView, change) in
```
1ec5, line 280: Does observing a key path on self causes the map view to retain itself?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -266,6 +276,20 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            guard change.newValue != nil else { return }
+            guard let self = self else { return }
+            
+            if let routes = self.routes, let currentRoute = routes.first {
+                self.show(routes)
+                self.showWaypoints(on: currentRoute)
+            }
```
1ec5, line 287: It would be better to gradually reduce the usage of `NavigationMapView.routes`. We’ve successfully removed NavigationMapView’s listening for route progress change. In the same vein, we should aim to remove extra state that the map view keeps around that could easily get out of sync with RouteController.

1ec5:
> current behavior brings not really good UX, since those who're just getting familiar with the Navigation SDK might encounter it without any prior knowledge of why annotations are staying on map view whereas routes and waypoints do not.

On the one hand, I see how it’s unintuitive that changing the style clears out runtime styling changes, but on the other hand, we can only solve so much of that issue. The developer is free to apply their own runtime styling changes, so the navigation SDK can only solve so much of that issue for them, and we’d do so clumsily by enforcing a z-order that the developer might not want.

I think I’d still prefer that we move the style observation to RouteMapViewController or NavigationViewController, where it’s already a normal thing to observe the style. Then we could add something to the documentation for `show(_:legIndex:)` etc. that reiterates that these methods’ changes need to be reapplied when switching styles, for the case where a developer implements a standalone map or custom navigation UI and needs to call these methods directly.

1ec5:
On the other _other_ hand, changing the style currently seems to get `routes` out of sync with what’s visually on the map. Since `routes` is private, there’s no recourse other than remembering to call `removeRoutes()` or `show(_:legIndex:)`. So at a minimum, we do need to observe the `styleURL` key path and clear out `routes` if it changes.

MaximAlien:
Current PR is no longer relevant.

Issue #2552, opened 2020-08-19
Flickering when highlighting building.


MaximAlien:
In some cases after calling `unhighlightBuildings` and highlighting buildings once again flickering happens. For example:
![output](https://user-images.githubusercontent.com/1496498/90578953-6df0a000-e179-11ea-800c-b77d3e8cc466.gif)
It is expected that `unhighlightBuildings` completely removes all info related to previously highlighted building and it'll not appear on screen afterwards.

PR #2558, opened 2020-08-20
Fix map view flickering.

MaximAlien:
Closing #2552.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -87,9 +87,9 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         
         static let instruction = "\(identifierNamespace).instruction"
         
-        static let mainRouteCasingSource = "\(identifierNamespace).mainRouteCasingSource"
+        static let mainRouteCasing = "\(identifierNamespace).mainRouteCasing"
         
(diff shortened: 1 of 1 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -1370,9 +1372,22 @@ extension NavigationMapView {
         style?.removeLayer(highlightedBuildingsLayer)
     }
     
+    private func addBuildingsSource() -> MGLSource? {
+        let buildingsSource = style?.source(withIdentifier: SourceIdentifier.buildingExtrusion)
+        if buildingsSource == nil {
+            let buildingsSource = MGLVectorTileSource(identifier: SourceIdentifier.buildingExtrusion,
+                                                      configurationURL: URL(string: "mapbox://mapbox.mapbox-streets-v8")!)
```
1ec5, line 1379: I just realized this code runs even if the style already has a Streets source v8 source, either as part of the `composite` source or independently. This is potentially problematic: the style specification requires that the overall style contain no conflicts between the source layer identifiers of the various sources. By adding a redundant copy of the Streets source, we guarantee conflicts between those copies. In the best-case scenario, this will only result in console spew, but it could also result in runtime styling changes not taking effect uniformly.

1ec5: I missed that this code only runs if `streetsSources.isEmpty` is true. Sorry for the false alarm.

PR #2560, opened 2020-08-20
Location projection looking forward during free driving

d-prukop:
This PR adds an adjustment so that the timestamp used in `Navigator.status(at:)` is now a time in the future.

### DIFF
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -600,6 +600,13 @@ extension ViewController {
             
             var coordinates: [CLLocationCoordinate2D] = [location.coordinate]
             trackPolyline?.appendCoordinates(&coordinates, count: UInt(coordinates.count))
+            
+            if mapView?.userTrackingMode != .followWithCourse, location.horizontalAccuracy < 65.0, location.speed > 8.0, let camera = mapView?.camera {
+                mapView?.userTrackingMode = .followWithCourse
+                camera.altitude = 750.0
+                camera.pitch = 45.0
+                mapView?.setCamera(camera, animated: true)
+            }
         }
         
         if let rawLocation = notification.userInfo?[PassiveLocationDataSource.NotificationUserInfoKey.rawLocationKey] as? CLLocation {
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -600,13 +600,6 @@ extension ViewController {
             
             var coordinates: [CLLocationCoordinate2D] = [location.coordinate]
             trackPolyline?.appendCoordinates(&coordinates, count: UInt(coordinates.count))
-            
-            if mapView?.userTrackingMode != .followWithCourse, location.horizontalAccuracy < 65.0, location.speed > 8.0, let camera = mapView?.camera {
-                mapView?.userTrackingMode = .followWithCourse
-                camera.altitude = 750.0
-                camera.pitch = 45.0
-                mapView?.setCamera(camera, animated: true)
-            }
         }
         
         if let rawLocation = notification.userInfo?[PassiveLocationDataSource.NotificationUserInfoKey.rawLocationKey] as? CLLocation {
--- a/MapboxCoreNavigation/PassiveLocationDataSource.swift
+++ b/MapboxCoreNavigation/PassiveLocationDataSource.swift
@@ -117,7 +117,8 @@ extension PassiveLocationDataSource: CLLocationManagerDelegate {
             return
         }
         
-        let status = navigator.status(at: lastRawLocation.timestamp)
+        let timestampForStatus = Date().addingTimeInterval(TimeInterval(1.1))
+        let status = navigator.status(at: timestampForStatus)
         let lastLocation = CLLocation(status.location)
         
         delegate?.passiveLocationDataSource(self, didUpdateLocation: lastLocation, rawLocation: lastRawLocation)
--- a/MapboxCoreNavigation/PassiveLocationDataSource.swift
+++ b/MapboxCoreNavigation/PassiveLocationDataSource.swift
@@ -117,7 +117,7 @@ extension PassiveLocationDataSource: CLLocationManagerDelegate {
             return
         }
         
-        let timestampForStatus = Date().addingTimeInterval(TimeInterval(1.1))
+        let timestampForStatus = Date().addingTimeInterval(RouteControllerDeadReckoningTimeInterval)
         let status = navigator.status(at: timestampForStatus)
         let lastLocation = CLLocation(status.location)
         
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -193,7 +193,8 @@ open class RouteController: NSObject {
         
         locations.forEach { navigator.updateLocation(for: FixLocation($0)) }
         
-        let status = navigator.status(at: location.timestamp)
+        let timestampForStatus = Date().addingTimeInterval(RouteControllerDeadReckoningTimeInterval)
+        let status = navigator.status(at: timestampForStatus)
         
         // Notify observers if the step’s remaining distance has changed.
         update(progress: routeProgress, with: CLLocation(status.location), rawLocation: location)


MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -117,7 +117,8 @@ extension PassiveLocationDataSource: CLLocationManagerDelegate {
             return
         }
         
-        let status = navigator.status(at: lastRawLocation.timestamp)
+        let timestampForStatus = Date().addingTimeInterval(TimeInterval(1.1))
```
1ec5, line 120: @mapbox/navnative, is it a safe assumption that we should always ask the navigator for the location 1.1&nbsp;seconds ahead during free-driving, but not during active turn-by-turn navigation? Would this duration vary by any factors within the navigation SDK’s control?

SiarheiFedartsou: I believe this look ahead time is somehow configurable by end user in Android SDK. @mapbox/navigation-android am I right?

Also I don't see why we should do it for free-drive only. It should be done for both active guidance and free drive. But it can introduce more map matching errors, because every prediction increases chances of wrong snappings - that's why it is good idea to give end user ability to tune this parameter.

LukasPaczos: Yes, we add prediction (1.1s by default) to both free drive and active guidance and this value is configurable with runtime options.

1ec5: OK, that addresses my concern about there being a discrepancy between navigators, given the intention to reuse the navigator between PassiveLocationDataSource and RouteController at some point.

As for the default, can we explain why it’s 1.1&nbsp;seconds and not 1.0 seconds, the actual duration of each camera animation?

https://github.com/mapbox/mapbox-navigation-ios/blob/8a3aec8b152ad1cdede6ef8494300b3e2c4406f6/MapboxNavigation/NavigationMapView.swift#L372

Example/ViewController.swift
```diff
@@ -600,6 +600,13 @@ extension ViewController {
             
             var coordinates: [CLLocationCoordinate2D] = [location.coordinate]
             trackPolyline?.appendCoordinates(&coordinates, count: UInt(coordinates.count))
+            
+            if mapView?.userTrackingMode != .followWithCourse, location.horizontalAccuracy < 65.0, location.speed > 8.0, let camera = mapView?.camera {
+                mapView?.userTrackingMode = .followWithCourse
+                camera.altitude = 750.0
+                camera.pitch = 45.0
```
1ec5, line 607: Might make sense to align these values to CarPlay:

https://github.com/mapbox/mapbox-navigation-ios/blob/8a3aec8b152ad1cdede6ef8494300b3e2c4406f6/MapboxNavigation/CarPlayMapViewController.swift#L10

Example/ViewController.swift
```diff
@@ -600,6 +600,13 @@ extension ViewController {
             
             var coordinates: [CLLocationCoordinate2D] = [location.coordinate]
             trackPolyline?.appendCoordinates(&coordinates, count: UInt(coordinates.count))
+            
+            if mapView?.userTrackingMode != .followWithCourse, location.horizontalAccuracy < 65.0, location.speed > 8.0, let camera = mapView?.camera {
+                mapView?.userTrackingMode = .followWithCourse
```
1ec5, line 605: The example application doesn’t have a Recenter button, so this might be annoying to undo when stopping. Should we automatically change the camera back when at rest or add a button to change the tracking mode back?

MapboxCoreNavigation/PassiveLocationDataSource.swift
1ec5: @LukasPaczos confirmed that the intention is to align the fast-forwarding interval with the animation interval to avoid choppiness. The animation interval is based on location updates coming in once a second from Core Location, also to avoid choppiness. I think that speaks to a more appropriate default of 1.0&nbsp;seconds, but I’d love to hear why it should be 1.1&nbsp;seconds instead. It’s very unlikely that a 1.0-second interval would result in a user-perceptible lag where 1.1&nbsp;seconds would not.

LukasPaczos: Like I mentioned on slack, I think we can safely change the prediction on Android to 1.0 if that simplifies the implementation on iOS. We're already multiplying the animation duration by 1.1 in the Maps SDK to have smooth transitions, so it shouldn't impact the Android experience in any significant way.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -193,7 +193,8 @@ open class RouteController: NSObject {
         
         locations.forEach { navigator.updateLocation(for: FixLocation($0)) }
         
-        let status = navigator.status(at: location.timestamp)
+        let timestampForStatus = Date().addingTimeInterval(RouteControllerDeadReckoningTimeInterval)
+        let status = navigator.status(at: timestampForStatus)
```
1ec5, line 197: I’m concerned that we’re taking a blanket approach to fast-forwarding location tracking based solely on the map’s built-in animation behavior. The motivation behind this change was to align the camera and puck’s animated position on screen with the user’s real-world location, accounting for the built-in 1&nbsp;second animation. However, other mechanisms depend on these location updates, namely guidance instructions, rerouting, arrival, tunnel-related style switching, and any other UI or business logic at the application level.

The ideal solution would be to move this lookahead logic to MapboxNavigation: `NavigationViewController.navigationService(_:didUpdate:with:rawLocation:)` would ignore the passed-in locations (which would not be influenced by any lookahead) in favor of calling `RouteController.projectedLocation(for:)`, passing in a lookahead that matches the animation duration. This would more appropriately keep UI-related concerns together:

https://github.com/mapbox/mapbox-navigation-ios/blob/ebab7d0881d33a975e265e8db20ca88d2eb7918a/MapboxNavigation/NavigationViewController.swift#L546 https://github.com/mapbox/mapbox-navigation-ios/blob/ebab7d0881d33a975e265e8db20ca88d2eb7918a/MapboxCoreNavigation/RouteController.swift#L340-L346

There are two downsides to this approach:

* If the application implements a custom UI instead of using NavigationViewController, it’s the application’s responsibility to call `RouteController.projectedLocation(for:)`, which may not be obvious. We can address this expectation in documentation and examples.
* More concerningly, `RouteController.projectedLocation(for:)` must call `Navigator.status(at:)` directly; calling `Navigator.status(at:)` out of band causes the navigator to skip instructions and potentially behave nondeterministically: #2113. I don’t think the workaround in https://github.com/mapbox/mapbox-navigation-ios/issues/2113#issuecomment-513833620 would help at all, because we would still need two different statuses at any given time. @SiarheiFedartsou, do you know if the underlying MapboxNavigationNative issue has been resolved?

MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -117,7 +117,8 @@ extension PassiveLocationDataSource: CLLocationManagerDelegate {
             return
         }
         
-        let status = navigator.status(at: lastRawLocation.timestamp)
+        let timestampForStatus = Date().addingTimeInterval(RouteControllerDeadReckoningTimeInterval)
+        let status = navigator.status(at: timestampForStatus)
```
1ec5, line 121: PassiveLocationManager is map view–specific, but PassiveLocationDataSource is not. To decouple map view behavior from location tracking behavior, we’d need PassiveLocationDataSource to expose a method similar to `RouteController.projectedLocation(for:)`, with all the same caveats.

MapboxCoreNavigation/RouteController.swift
1ec5: > The ideal solution would be to move this lookahead logic to MapboxNavigation: `NavigationViewController.navigationService(_:didUpdate:with:rawLocation:)` would ignore the passed-in locations (which would not be influenced by any lookahead) in favor of calling `RouteController.projectedLocation(for:)`, passing in a lookahead that matches the animation duration.

#2583 tracks removing `RouteController.projectedLocation(for:)`, because MapboxNavigationNative won’t support this (already buggy) functionality for much longer. #2587 tracks decoupling the actual and projected location so that the map view can animate the camera and puck to the right target position without affecting non-animated components.

Issue #2563, opened 2020-08-21
FeedbackViewController is not dismissed when rotating device (iPad only).


MaximAlien:
To follow default behavior on iPhone which dismisses `FeedbackViewController` whenever device is rotated, iPad should follow the same flow. Since `FeedbackViewController` is not currently dismissed after rotation it cases UI issues on iPad:

![Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2020-08-21 at 10 58 45](https://user-images.githubusercontent.com/1496498/90920588-66a9dc00-e39d-11ea-81ac-73056a898d64.png)

PR #2564, opened 2020-08-21
Add fix which allows to dismiss FeedbackViewController on both iPad and iPhone.

MaximAlien:
Closing #2563.


avi-c:
I merged in #2544 and since this PR is ready I will merge it in on master too.

PR #2566, opened 2020-08-24
Fix conflicts when generating documentation from combined module

1ec5:
The two `Bundle.mapboxNavigation` extension properties conflict when the documentation script gloms MapboxNavigation and MapboxCoreNavigation together into a single module. Make the one in MapboxNavigation public instead, since it has more certainty about the presence of MapboxNavigation.

/cc @mapbox/navigation-ios


1ec5:
Cherry-picked into the release-v1.0 branch in c159d2820fa3873a0dc0a441c975d2c5e62f7606.

PR #2291, opened 2019-12-27
Show speed limit indicator during turn-by-turn navigation

1ec5:
Added a speed limit indicator to the upper-left corner of the map view during turn-by-turn navigation (upper-right corner in CarPlay). To hide the speed limit indicator, set the `NavigationViewController.showsSpeedLimits` property to `false`. To customize the indicator’s colors, configure `SpeedLimitView`’s appearance proxy inside a `Style` subclass.

phone-mutcd-day phone-vienna-day phone-mutcd phone-vienna

The speed limit indicator appears beneath the compass on CarPlay:

carplay-day-vienna carplay-night-mutcd

Fixes #962.

/ref #1242 mapbox/navigation-ui-resources#13
/cc @mapbox/navigation-ios @avi-c

PR #2569, opened 2020-08-25
Update screenshot

1ec5:
The screenshot hasn‘t been updated in a while. This screenshot shows the speed limit indicator added in #2291 as well as a state-specific route shield implemented last year in the Directions API. I took this screenshot before #2377 landed but will need to update it again, now that the route line no longer appears behind the puck.

iPhone X night + frame iPhone X night

/cc @mapbox/navigation-ios


1ec5:
Holding off on an updated screenshot until #2574 is fixed.

1ec5:
I’m having some difficulty getting the Directions API to give me shield icons, so I think we can just run with this screenshot as if `NavigationViewController.routeLineTracksTraversal` were disabled, as is the default.

1ec5:
Cherry-picked into the release-v1.0 branch as 4e6a91f9259350b13a2247a0090aa9a4620f014c.

PR #2576, opened 2020-08-28
Make Second Level Feedback strings localizable

### LINKED ISSUE #2474: Feedback categories should be localizable

avi-c:
Move hard coded strings to localizable ones.

Fixes #2474.

### DIFF
--- a/MapboxNavigation/FeedbackItem.swift
+++ b/MapboxNavigation/FeedbackItem.swift
@@ -16,71 +16,71 @@ public extension FeedbackType {
         case .general:
             return "Feedback"
         case .incorrectVisual(.none):
-            return "Incorrect visual"
+            return NSLocalizedString("INCORRECT_VISUAL_FEEDBACK", bundle: .mapboxNavigation, value: "Incorrect visual", comment: "General category of route feedback where visual instruction was incorrect.")
         case .incorrectVisual(.turnIconIncorrect):
-            return "Turn icon incorrect"
+            return NSLocalizedString("INCORRECT_VISUAL_TURN_ICON_INCORRECT_FEEDBACK", bundle: .mapboxNavigation, value: "Turn icon incorrect", comment: "Specific route feedback for incorrect turn arrow being shown.")
         case .incorrectVisual(.streetNameIncorrect):
-            return "Street name incorrect"
+            return NSLocalizedString("INCORRECT_VISUAL_STREET_NAME_INCORRECT_FEEDBACK", bundle: .mapboxNavigation, value: "Street name incorrect", comment: "Specific route feedback for incorrect street name.")
         case .incorrectVisual(.instructionUnnecessary):
-            return "Instruction unnecessary"
+            return NSLocalizedString("INCORRECT_VISUAL_INSTRUCTION_UNNECESSARY_FEEDBACK", bundle: .mapboxNavigation, value: "Instruction unnecessary", comment: "Specific route feedback for an unnecessary instruction.")
         case .incorrectVisual(.instructionMissing):
-            return "Instruction missing"
+            return NSLocalizedString("INCORRECT_VISUAL_INSTRUCTION_MISSING_FEEDBACK", bundle: .mapboxNavigation, value: "Instruction missing", comment: "Specific route feedback that an instruction was missing.")
         case .incorrectVisual(.maneuverIncorrect):
-            return "Maneuver incorrect"
+            return NSLocalizedString("INCORRECT_VISUAL_MANEUVER_INCORRECT_FEEDBACK", bundle: .mapboxNavigation, value: "Maneuver incorrect", comment: "Specific route feedback that a maneuver specified was incorrect.")
         case .incorrectVisual(.exitInfoIncorrect):
-            return "Exit info incorrect"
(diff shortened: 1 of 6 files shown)


MapboxNavigation/FeedbackItem.swift
```diff
@@ -16,71 +16,71 @@ public extension FeedbackType {
         case .general:
             return "Feedback"
         case .incorrectVisual(.none):
-            return "Incorrect visual"
+            return NSLocalizedString("INCORRECT_VISUAL_FEEDBACK", bundle: .mapboxNavigation, value: "Incorrect visual", comment: "General category of route feedback where visual instruction was incorrect.")
```
1ec5, line 19: Can you make the same change to ”Looks Incorrect” as in https://github.com/mapbox/mapbox-navigation-ios/pull/2497#discussion_r462691238?

avi-c: Done.

MapboxNavigation/FeedbackSubtypeViewController.swift
```diff
@@ -149,11 +149,12 @@ class FeedbackSubtypeViewController: FeedbackViewController {
...
+            reportButton.setTitle(NSLocalizedString("NAVIGATION_REPORT_CANCEL", bundle: .mapboxNavigation, value: "Cancel", comment: "Title for button that cancels user's submission of feedback on navigation session issues."), for: .normal)
+            
         } else if selectedItems.count == 1 {
-            reportButton.setTitle(NSLocalizedString("NAVIGATION_REPORT_ISSUE", comment: "Title for button that submits user's feedback on navigation session issues."), for: .normal)
+            reportButton.setTitle(NSLocalizedString("NAVIGATION_REPORT_ISSUE", bundle: .mapboxNavigation, value: "Send Item", comment: "Title for button that submits user's feedback on navigation session issues."), for: .normal)
         } else {
-            reportButton.setTitle(NSLocalizedString("NAVIGATION_REPORT_ISSUES", comment: "Title for button that submits user's feedback on multiple navigation session issues."), for: .normal)
+            reportButton.setTitle(NSLocalizedString("NAVIGATION_REPORT_ISSUES", bundle: .mapboxNavigation, value: "Send Items",  comment: "Title for button that submits user's feedback on multiple navigation session issues."), for: .normal)
```
1ec5, line 157: This logic doesn’t work in some of the supported languages, such as Russian. We use a stringsdict to enable more flexible localization of grammatical plurals and genders:

https://github.com/mapbox/mapbox-navigation-ios/blob/6e54f9d38bb09e7cb1cdac7d194364e69f7d3ea8/MapboxNavigation/RatingControl.swift#L105-L109 https://github.com/mapbox/mapbox-navigation-ios/blob/dc9f1e9be68b18ca4f57882c98e0a58e85461507/MapboxNavigation/Resources/en.lproj/Localizable.stringsdict#L21-L36

MapboxNavigation/FeedbackItem.swift
```diff
@@ -16,71 +16,71 @@ public extension FeedbackType {
...
+            return NSLocalizedString("ROUTE_QUALITY_ROAD_CLOSURE_FEEDBACK", bundle: .mapboxNavigation, value: "Road closure", comment: "General route feedback that route contained closed road.")
         case .roadClosure(.streetPermanentlyBlockedOff):
-            return "Street permanently blocked off"
+            return NSLocalizedString("ROUTE_QUALITY_ROAD_CLOSURE_PERMANENT_FEEDBACK", bundle: .mapboxNavigation, value: "Street permanently blocked off", comment: "Specifc route feedback that route contained road that is permanently closed.")
         case .roadClosure(.roadMissingFromMap):
-            return "Road missing from map"
+            return NSLocalizedString("ROUTE_QUALITY_ROAD_MISSING_FROM_MAP_FEEDBACK", bundle: .mapboxNavigation, value: "Road missing from map", comment: "Specifc route feedback that route contained road not shown on map.")
         case .roadClosure(.other):
-            return "Other"
+            return NSLocalizedString("ROAD_CLOSURE_OTHER_FEEDBACK", bundle: .mapboxNavigation, value: "Other", comment: "Specific route feedback that a road closure type was encountered but not listed as a choice.")
```
1ec5, line 83: This was probably a pain to make localizable. As tail work, to reduce verbosity, we should move these strings into a localizable plist. That would also bring us a little closer to making the feedback categories more configurable.

MapboxNavigation/Resources/Base.lproj/Localizable.strings
```diff
@@ -70,15 +85,57 @@
...
 "LESS_THAN" = "<%@";
 
+/* Title for button that cancels user's submission of feedback on navigation session issues. */
+"NAVIGATION_REPORT_CANCEL" = "Cancel";
+
+/* Title for button that submits user's feedback on navigation session issues. */
+"NAVIGATION_REPORT_ISSUE" = "Send Item";
+
+/* Title for button that submits user's feedback on multiple navigation session issues. */
+"NAVIGATION_REPORT_ISSUES" = "Send Items";
```
1ec5, line 137: Once we add the string format to Localizable.stringsdict, these two entries will collapse into a single entry, set to “Send Item(s)”, that will be overridden by the properly pluralized form if available.

PR #2579, opened 2020-08-31
Add automated test of PassiveLocationDataSource

### LINKED ISSUE #2546: Add automated tests of PassiveLocationDataSource
#2410 adds an integration test of PassiveLocationDataSource, but the test is disabled because it requires network access, a valid Mapbox access token, and loads of time. We should figure out how to mock a routing tiles API response that the navigator can route over, however crudely.

We might be able to piggyback on the test fixture already in place for OfflineDirections. #2509 would remove OfflineRoutingTests, but it doesn’t currently remove the test fixture. It weighs in at only 819 kB, far less than the routing tiles for the setting of the disabled PassiveLocationDataSourceTests, so we won’t have to resort to something like LFS.

/cc @mapbox/navigation-ios @SiarheiFedartsou @LukasPaczos

IodaMikeMapbox:
The pr add a test of `PassiveLocationDataSourec` which doesn't require network access and doesn't take much time.
Also the pr fixes the method for manually setting the location.

Resolves #2546

### DIFF
--- a/MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
+++ b/MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
@@ -62,26 +62,8 @@ class PassiveLocationDataSourceTests: XCTestCase {
     func testManualLocations() {
         let tilesVersion = "preloadedtiles" // any string
 
-        // Copy tiles to the cache directory, tiles should be placed in a folder with name
-        guard var tilesCacheURL = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first else {
-            preconditionFailure("No Caches directory to create the tile directory inside")
-        }
-        if let bundleIdentifier = Bundle.main.bundleIdentifier ?? Bundle.mapboxCoreNavigation.bundleIdentifier {
-            tilesCacheURL.appendPathComponent(bundleIdentifier, isDirectory: true)
-        }
-        tilesCacheURL.appendPathComponent(".mapbox", isDirectory: true)
-        tilesCacheURL.appendPathComponent(tilesVersion, isDirectory: true)
         let bundle = Bundle(for: Fixture.self)
         let filePathURL: URL = URL(fileURLWithPath: bundle.bundlePath.appending("/tiles/liechtenstein"))
-        let fileManager = FileManager.default
-        var isDir : ObjCBool = false
-        if !fileManager.fileExists(atPath: tilesCacheURL.path, isDirectory:&isDir) {
-            do {
-                try fileManager.copyItem(atPath: filePathURL.path, toPath: tilesCacheURL.path)
-            }catch let error{
-                print(error.localizedDescription)
-            }
-        }
 
         // Create PassiveLocationDataSource and configure it with the tiles version (version is used to find the tiles in the cache folder)
         let locationManager = PassiveLocationDataSource()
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
```diff
@@ -59,45 +60,52 @@ class PassiveLocationDataSourceTests: XCTestCase {
...
+        let filePathURL: URL = URL(fileURLWithPath: bundle.bundlePath.appending("/tiles/liechtenstein"))
+        let fileManager = FileManager.default
+        var isDir : ObjCBool = false
+        if !fileManager.fileExists(atPath: tilesCacheURL.path, isDirectory:&isDir) {
+            do {
+                try fileManager.copyItem(atPath: filePathURL.path, toPath: tilesCacheURL.path)
+            }catch let error{
+                print(error.localizedDescription)
+            }
+        }
```
1ec5, line 84: This is doing a lot of work that PassiveLocationDataSource normally does. Can we restructure PassiveLocationDataSource to take an arbitrary path to the tiles? That option can be private and visible only to this test.

MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
```diff
@@ -59,45 +60,52 @@ class PassiveLocationDataSourceTests: XCTestCase {
...
+
+        // Create PassiveLocationDataSource and configure it with the tiles version (version is used to find the tiles in the cache folder)
         let locationManager = PassiveLocationDataSource()
-        
-        let startingExpectation = expectation(description: "Location manager should start without an error")
-        locationManager.startUpdatingLocation { (error) in
-            XCTAssertNil(error)
-            startingExpectation.fulfill()
+        do {
+            try locationManager.configureNavigator(withTilesVersion: tilesVersion)
```
1ec5, line 89: If we can configure the PassiveLocationDataSource to use the Liechtenstein tiles, then we won’t need to call this private method. Avoiding an explicit call to this method would give us more confidence in the overall functionality in a more realistic setting.

MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
```diff
@@ -59,45 +60,52 @@ class PassiveLocationDataSourceTests: XCTestCase {
...
-                        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(500)) {
-                            locationManager.updateLocation(CLLocation(latitude: 47.209842, longitude: 9.522377))
-                        }
-                    }
-                }
-            }
+            locationManager.updateLocation(CLLocation(latitude: 47.209082, longitude: 9.524319))
+            locationManager.updateLocation(CLLocation(latitude: 47.209229, longitude: 9.523838))
+            locationManager.updateLocation(CLLocation(latitude: 47.209612, longitude: 9.522629))
+            locationManager.updateLocation(CLLocation(latitude: 47.209842, longitude: 9.522377))
```
1ec5, line 106: Each of these location updates has more or less the same timestamp. Would that throw off the navigator and make this an unrealistic test? We could set these CLLocations’ timestamps to be more offset from each other, which would at least come closer to the asynchronous test we had before.

1ec5 (approved):
Looks good, thanks!

Issue #2574, opened 2020-08-27
Traffic congestion colors are misplaced and smeared along the route line


1ec5:
`NavigationMapView.generateTrafficGradientStops(for:)` seems to be calculating incorrect stops for the gradient that colors the route line according to traffic congestion. To the end user, it would seem like the route line has random splotches of red and yellow that don’t look like traffic congestion and don’t correspond to real-world conditions.

The following examples are in the vicinity of 39.2872°N, −84.4656°W near Tri-County Mall in Springdale, Ohio, comparing the example application running navigation SDK v1.0.0-rc.1 with map SDK v6.0.0 against a webpage running mapbox/mapbox-gl-directions#145 (which is based on v4.0.0).

<img width="1280" alt="kemper" src="https://user-images.githubusercontent.com/1231218/91367528-93099200-e7bb-11ea-9fa9-591a4a45f95b.png">

<img width="1280" alt="tri-county" src="https://user-images.githubusercontent.com/1231218/91367537-98ff7300-e7bb-11ea-957d-7cfab2ea53f9.png">

The inaccuracy is more apparent when getting a multileg route in the example application: the segments move considerably between a given single-leg route and a multileg route that includes it. These screenshots were taken within a few seconds of each other; the inconsistency reproduces reliably:

<img src="https://user-images.githubusercontent.com/1231218/91368104-0bbd1e00-e7bd-11ea-8029-ae90590a6d9e.png" height="400" alt="2"> <img src="https://user-images.githubusercontent.com/1231218/91368105-0d86e180-e7bd-11ea-8714-8810bb036c26.png" height="400" alt="3">

This regression was introduced in #2377. It’s possible that the refactoring suggested in #2423 would improve the situation.

/cc @mapbox/navigation-ios

1ec5:
`NavigationMapView.generateTrafficGradientStops(for:)` hard-codes the first leg, so it’s incapable of representing the congestion data that accompanies subsequent legs.

https://github.com/mapbox/mapbox-navigation-ios/blob/b1f685062a4d5b4d3ec0611c0be551b335b78cb1/MapboxNavigation/NavigationMapView.swift#L1104

[...]

Issue #2512, opened 2020-07-31
Route line potentially underlaps roads or buildings


1ec5:
The route line is inserted at the wrong place in the map style’s layer stack, potentially underlapping a non-symbol layer. In the default Mapbox Navigation Preview and Guidance styles, this layer happens to be a disputed boundary layer, which doesn’t necessarily get in the way, but a custom style could put anything at that position. So the route line could potentially be obscured by roads or buildings, affecting basic usability.

As I once explained in [this Stack Overflow post](https://stackoverflow.com/a/43876838/4585461), the intention is primarily to place the route line over any linear features like roads and rivers but under any labels and icons. It’s a little more nuanced than that, because one-way arrows are also implemented as icons in symbol layers. So the algorithm is to iterate over the stack from top to bottom, looking for the first non-symbol layer and inserting the route line _above_ that layer. Instead, #270 regressed the logic, inserting it _under_ that layer:

https://github.com/mapbox/mapbox-navigation-ios/blob/d17882f50adad5bbf3f38de9c5bce0adff5efa91/MapboxNavigation/NavigationMapView.swift#L507-L510

The fix is simply to replace the call to `MGLStyle.insert(_:below:)` with `MGLStyle.insert(_:above:)`.

/cc @mapbox/navigation-ios

MaximAlien:
Fixed in #2580.

PR #2580, opened 2020-08-31
Fix traffic congestion rendering issues.

### LINKED ISSUE #2423: Investigate route gradient stop property declaration in NavigationMapView
Following up on https://github.com/mapbox/mapbox-navigation-ios/pull/2377#discussion_r445168323,

`NavigationMapView` declares two gradient stop dictionaries in the form of a tuple, used to calculate the traffic line gradients for both the main route line and its casing:

```swift
typealias RouteGradientStops = (line: [CGFloat: UIColor], casing: [CGFloat: UIColor])
private var routeGradientStops = RouteGradientStops(line: [:], casing: [:])
```

It was suggested in https://github.com/mapbox/mapbox-navigation-ios/pull/2377#discussion_r445168323 that this be changed to two parallel properties instead, such as:

```swift
private var routeGradientStops: [CGFloat: UIColor] = [:]
private var routeGradientCasingStops: [CGFloat: UIColor] = [:]
```

However, this declaration causes the gradient stops to be miscalculated - we should look into why this is happening.

| Screen Shot 2020-06-24 at 3 58 47 PM | Screen Shot 2020-06-24 at 3 57 33 PM |
|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Tuple                                                                                                                                                                   | Two parallel properties                                                                                                                                                 |

\* note - the line caps are not rounded in these two screenshots, but since those were taken that has been corrected.

cc @1ec5

MaximAlien:
Closing #2574.

I've slightly re-worked original implementation. Now there are such methods which allow to add route layers:
```swift
addMainRouteLayer(_:source:lineGradient:)
addMainRouteCasingLayer(_:source:lineGradient:below:)
addAlternateRoutesLayer(_:source:below:)
addAlternateRoutesCasingLayer(_:source:below:)
```

Such methods are used to create line gradient:
```swift
routeLineGradient(_:fractionTraveled:)
routeCasingGradient(_:)
```

In case if route layer was already added route line gradient with congestion segments will be retrieved out of it.
For route casing color is always the same (without congestions), but fading is also added depending on route progress.

Current PR also fixes #2423 and #2512.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -610,9 +610,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     
     @discardableResult func addAlternativeRoutesLayer(_ style: MGLStyle, source: MGLSource, below layer: MGLStyleLayer) -> MGLStyleLayer {
         let customAlternativeRoutesLayer = navigationMapViewDelegate?.navigationMapView(self,
-                                                                                      alternativeRouteStyleLayerWithIdentifier: StyleLayerIdentifier.alternativeRoutes,
-                                                                                      source: source)
-        
+                                                                                        alternativeRouteStyleLayerWithIdentifier: StyleLayerIdentifier.alternativeRoutes,
+                                                                                        source: source)
         let currentAlternativeRoutesLayer = style.layer(withIdentifier: StyleLayerIdentifier.alternativeRoutes)
         
         if let alternativeRoutesLayer = customAlternativeRoutesLayer, let _ = currentAlternativeRoutesLayer {
@@ -641,9 +640,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     
     @discardableResult func addAlternativeRoutesCasingLayer(_ style: MGLStyle, source: MGLSource, below layer: MGLStyleLayer) -> MGLStyleLayer {
         let customAlternativeRoutesCasingLayer = navigationMapViewDelegate?.navigationMapView(self,
-                                                                                            alternativeRouteCasingStyleLayerWithIdentifier: StyleLayerIdentifier.alternativeRoutesCasing,
-                                                                                            source: source)
-        
+                                                                                              alternativeRouteCasingStyleLayerWithIdentifier: StyleLayerIdentifier.alternativeRoutesCasing,
+                                                                                              source: source)
         let currentAlternativeRoutesCasingLayer = style.layer(withIdentifier: StyleLayerIdentifier.alternativeRoutesCasing)
         
         if let alternativeRoutesCasingLayer = customAlternativeRoutesCasingLayer, let _ = currentAlternativeRoutesCasingLayer {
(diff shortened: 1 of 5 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -487,126 +477,202 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        style.insertLayer(mainRouteCasingLayer, below: layer)
+        
         return mainRouteCasingLayer
     }
-
-    func alternativeRouteStyleLayer(identifier: String, source: MGLSource) -> MGLLineStyleLayer {
-        let alternateRoutesLayer = MGLLineStyleLayer(identifier: identifier, source: source)
+    
+    @discardableResult func addAlternateRoutesLayer(_ style: MGLStyle, source: MGLSource, below layer: MGLStyleLayer) -> MGLStyleLayer {
+        let customAlternateRoutesLayer = navigationMapViewDelegate?.navigationMapView(self,
```
MaximAlien, line 612: At this point with use `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteStyleLayerWithIdentifier:source:)` and `NavigationMapViewDelegate.navigationMapView(_:alternateRouteCasingStyleLayerWithIdentifier:source:)`. @1ec5, maybe it'd be better to use the same convention here? As I can see `alternate` is mostly used in `NavigationMapView`, maybe we should use the same naming in public API method?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -487,126 +477,202 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+                                                                                      alternativeRouteStyleLayerWithIdentifier: StyleLayerIdentifier.alternateRoutes,
```
MaximAlien, line 613: At this point we use `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteStyleLayerWithIdentifier:source:)` and `NavigationMapViewDelegate.navigationMapView(_:alternateRouteCasingStyleLayerWithIdentifier:source:)`. @1ec5, maybe it'd be better to use the same convention here? As I can see `alternate` is mostly used in `NavigationMapView`, maybe we should use the same naming in public API method?

1ec5: It should be “alternative” consistently. “Alternate” is easily read as a verb in this context, and it has a slightly different meaning anyways (alternating back and forth, not the alternative to something).

MapboxNavigation/NavigationMapView.swift
```diff
@@ -487,126 +477,202 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        
         self.routes = routes
+        
+        let allRoutesShape = navigationMapViewDelegate?.navigationMapView(self, shapeFor: routes) ?? shape(for: routes, legIndex: legIndex)
+        let allRoutesSource = addAllRoutesSource(style, shape: allRoutesShape)
+        
+        let mainRouteCasingShape = navigationMapViewDelegate?.navigationMapView(self, shapeFor: [mainRoute]) ?? shape(for: [mainRoute], legIndex: legIndex)
+        let mainRouteCasingSource = addMainRouteCasingSource(style, shape: mainRouteCasingShape)
+        
+        let mainRouteLayer = addMainRouteLayer(style, source: allRoutesSource, lineGradient: routeLineGradient(mainRoute, fractionTraveled: 0.0))
```
1ec5, line 491: The new helper methods make this method much easier to follow, thanks!

Issue #2526, opened 2020-08-05
NavigationMapView.routeLineTracksTraversal only works inside NavigationViewController


1ec5:
`NavigationMapView.routeLineTracksTraversal` is billed as an option to enable the “erasing route line” or “vanishing route line” feature on any NavigationMapView (#1307). But this property only works when the map view is embedded inside a NavigationViewController, because the only thing that reads the property is in RouteMapViewController. Nor does the feature work in CarPlay, which has a CarPlayNavigationViewController instead of a NavigationViewController.

https://github.com/mapbox/mapbox-navigation-ios/blob/aaed03d2c8e186abc34df296fd89a1218dea2013/MapboxNavigation/RouteMapViewController.swift#L460-L462

Unusually, `NavigationMapView.fadeRoute(_:)` is an internal method not exposed publicly, so there’s no workaround for an application that implements its own custom UI for turn-by-turn navigation.

https://github.com/mapbox/mapbox-navigation-ios/blob/aaed03d2c8e186abc34df296fd89a1218dea2013/MapboxNavigation/NavigationMapView.swift#L603

We should rename `fadeRoute(_:)` to `updateRoute(for:)`, have it take a RouteProgress instead of a raw number, and expose it publicly. Since `routeLineTracksTraversal` is dependent on functionality outside of NavigationMapView, namely progress updates, we should either move it to NavigationViewController or move that conditional inside `updateRoute(for:)`. Finally, we should call `updateRoute(for:)` from within `CarPlayNavigationViewController.progressDidChange(_:)` too.

/cc @mapbox/navigation-ios @d-prukop @avi-c @captainbarbosa

PR #2581, opened 2020-08-31
Make updateRoute(_:) public.

MaximAlien:
Closing #2526.

We only mention in `CHANGELOG.md` that `A portion of the route line now disappears behind the user puck as the user travels along the route during turn-by-turn navigation.` If needed such description can be extended to mention newly added APIs as well.


1ec5:
> We only mention in `CHANGELOG.md` that `A portion of the route line now disappears behind the user puck as the user travels along the route during turn-by-turn navigation.` If needed such description can be extended to mention newly added APIs as well.

Good point, let’s mention it, because the feature is otherwise off by default.

> A portion of the route line now disappears behind the user puck as the user travels along the route during turn-by-turn navigation if `NavigationViewController.routeLineTracksTraversal` is set to `true`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -672,17 +660,24 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
 
     // MARK: - Vanishing route line methods
     
-    func fade(_ route: Route, fractionTraveled: Double) {
+    /**
+     Update route style layer and its casing to disappear as the user location puck travels along certain route.
```
1ec5, line 664:
```suggestion
     Updates the route style layer and its casing style layer to gradually disappear as the user location puck travels along the displayed route.
```

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -481,8 +481,8 @@ extension RouteMapViewController: NavigationComponent {
             mapView.showVoiceInstructionsOnMap(route: route)
         }
         
-        if mapView.routeLineTracksTraversal {
-            mapView.fade(route, fractionTraveled: progress.fractionTraveled)
+        if let parent = parent as? NavigationViewController, parent.routeLineTracksTraversal {
```
1ec5, line 484: To avoid making this assumption, we can add a `routeLineTracksTraversal` property to this class and turn `NavigationViewController.routeLineTracksTraversal` into a computed property that passes through to `RouteMapViewController.routeLineTracksTraversal`.

PR #2582, opened 2020-08-31
Building highlights before reaching the destination

d-prukop:
Building highlights when the distance remaining on the current leg is less than 250 meters.


MapboxNavigation/NavigationMapView.swift
```diff
@@ -1362,13 +1368,17 @@ extension NavigationMapView {
      - parameter extrudesBuildings: Switch which allows to highlight buildings in either 2D or 3D. Defaults to true.
      */
     public func highlightBuildings(at coordinates: [CLLocationCoordinate2D], in3D extrudesBuildings: Bool = true) {
-        highlightBuildings(with: Set(coordinates.compactMap({ buildingIdentifier(at: $0) })), in3D: extrudesBuildings)
+        buildingHighlightPollingTimer?.invalidate()
+        buildingHighlightPollingTimer = Timer.scheduledTimer(withTimeInterval: 0.5, repeats: true) { [weak self] timer in
```
1ec5, line 1372: Every half-second seems somewhat aggressive, especially if the user ends up moving slowly (or stopped at a light) within that final third of a mile. Is there anything in `NavigationMapView.highlightBuildings(with:in3D:extrudeAll:)` that bails, avoiding style changes, if the call to that method is redundant?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -29,7 +29,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     /**
      Returns the altitude that the map camera initally defaults to.
      */
-    public var defaultAltitude: CLLocationDistance = 1000.0
+    public var defaultAltitude: CLLocationDistance = 850.0
```
1ec5, line 32: Should we make the map generally zoom in closer during urban driving, as this change does, or should we distinguish between the default altitude and the altitude as the user approaches the destination, also making that altitude configurable? It would be pretty straightforward to make an additional property for that.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -546,6 +546,10 @@ extension NavigationViewController: NavigationServiceDelegate {
             mapViewController?.mapView.updateCourseTracking(location: location, animated: true)
         }
         
+        if waypointStyle != .annotation, progress.currentLegProgress.distanceRemaining < 500 {
```
MaximAlien, line 549: I think Android is currently using threshold of [250 meters](https://github.com/mapbox/mapbox-navigation-android/pull/3515/files#diff-e8bba261553ad36e3c4c76f514f7f798R136).

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1362,13 +1368,17 @@ extension NavigationMapView {
      - parameter extrudesBuildings: Switch which allows to highlight buildings in either 2D or 3D. Defaults to true.
      */
     public func highlightBuildings(at coordinates: [CLLocationCoordinate2D], in3D extrudesBuildings: Bool = true) {
-        highlightBuildings(with: Set(coordinates.compactMap({ buildingIdentifier(at: $0) })), in3D: extrudesBuildings)
+        buildingHighlightPollingTimer?.invalidate()
```
MaximAlien, line 1371: Since `SimulatedLocationManager` uses `updateInterval` of 1 second timer invalidation might cause inability to highlight building if polling timer timeout is higher than 1 second. This also applies to real-life use case when location manager is going to deliver location data. We might have to reconsider current implementation and think whether `NavigationMapView` is really best place to store polling timer.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -614,9 +618,6 @@ extension NavigationViewController: NavigationServiceDelegate {
                           direction: location.course,
                           animated: true,
                           completionHandler: {
-                            // Highlight buildings which were marked as target destination coordinate in waypoint.
-                            mapView.highlightBuildings(at: self.routeOptions.waypoints.compactMap({ $0.targetCoordinate }), in3D: self.waypointStyle == .extrudedBuilding ? true : false)
```
MaximAlien, line 618: I think `zoomInAndHighlightBuilding(for:)` name is not really relevant if we start polling and decrease altitude. On the other hand insets will still have to be used to re-center map correctly. What about replacing current implementation with something like:
```swift
// Update insets to be able to correctly center map view after presenting end of route view.
mapViewController.updateMapViewContentInsets(animated: true, completion: {
    // Update user course view to correctly place it in map view.
    self.mapView?.updateCourseTracking(location: location, animated: false)
})
```
and renaming it to something more appropriate?

MaximAlien: Please also note that we only zoom-in and recenter in case if end of route view is presented. In case of intermediate waypoint such logic will not be performed. I think we'll have to perform more testing to support both cases.

MaximAlien: What I meant here by renaming current method and removing zoom-in is that if `defaultAltitude` is low enough zoom-in will not be needed at all. In current case we renamed method but still call it internally. I was imagining more like:

```swift
private func frameDestinationArrival(for location: CLLocation?) {
    if waypointStyle == .annotation { return }
    guard let mapViewController = self.mapViewController else { return }
    guard let location = location else { return }
    
    // Update insets to be able to correctly center map view after presenting end of route view.
    mapViewController.updateMapViewContentInsets(animated: true, completion: {
        // Update user course view to correctly place it in map view.
        self.mapView?.updateCourseTracking(location: location, animated: false)
    })
}
```

But for this `defaultAltitude` has to be lowered (in my case `850` doesn't work and buildings are not highlighted at all). With `600` highlighting works. I guess we have to play around with best value.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -270,6 +274,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         installUserCourseView()
     }
     
+    deinit {
+        buildingHighlightPollingTimer?.invalidate()
```
1ec5, line 278: The downside to managing the timer in NavigationMapView is that we may need to invalidate it in more places to avoid memory leaks. For one thing, what should happen if client code sets a new route or sets a new style?

MapboxNavigation/NavigationMapView.swift
1ec5: Agreed. In general, we should avoid storing any additional state on NavigationMapView because we aren’t full in control of the lifecycle of its contents.

MaximAlien (approved):
Even though I have concerns about current implementation (polling timer usage, `currentLeg` comparisons to detect whether map should be zoomed-in) probably the best idea would be to include it in upcoming `v1.0.0-rc.3` release. This will give us a chance to test it in 1TAP and possibly address any other issues/feedback before public release.

1ec5:
Let’s track the timer usage in a followup issue. In general, the risk with a timer or observation within NavigationMapView is that the map view itself would leak, which would be a major performance issue.

1ec5:
#2612 follows up on the timer usage.

PR #2588, opened 2020-09-02
Add passive location manager to CarPlay pre-navigation map

1ec5:
In CarPlay, the map view’s camera and user location indicator also snap to the road network during the browsing (free-driving) and previewing activities, just as in the navigating activity, thanks to a PassiveLocationManager.

/cc @mapbox/navigation-ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * Upgraded to [Mapbox Maps SDK for iOS v6.1.0-beta.2](https://github.com/mapbox/mapbox-gl-native-ios/releases/tag/ios-v6.1.0-beta.2). ([#2565](https://github.com/mapbox/mapbox-navigation-ios/pull/2565))
 
+### Other changes
+
+* In CarPlay, the map view’s camera and user location indicator also snap to the road network during the browsing and previewing activities, just as in the navigating activity. ([#2588](https://github.com/mapbox/mapbox-navigation-ios/pull/2588))
+
 ## v1.0.0
 
 ### Packaging
--- a/MapboxNavigation/CarPlayMapViewController.swift
+++ b/MapboxNavigation/CarPlayMapViewController.swift
@@ -1,4 +1,5 @@
 import Foundation
+import MapboxCoreNavigation
 #if canImport(CarPlay)
 import CarPlay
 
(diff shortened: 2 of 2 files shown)


1ec5:
This change means there can easily be two navigators active at the same time, one targeting the phone and one targeting the CarPlay screen, both running in the same process on the phone. However, both navigators would be guaranteed to be configured the same way and receive the same location updates at the same time. There isn’t any way for CarPlayMapViewController to share much of anything with the phone’s preview map view, because the latter map view is part of the application rather than built into the SDK.

If the presence of two concurrent navigators turns out to be a problem, we could investigate making PassiveLocationDataSource rely on a singleton under the hood, but that’ll just force the application developer to think about any crosstalk and race condition issues that might have led MapboxNavigationNative to use a singleton under its hood.

/cc @SiarheiFedartsou

SiarheiFedartsou:
> However, both navigators would be guaranteed to be configured the same way and receive the same location updates at the same time.

But it doesn't guarantee that `getStatus` calls will return the same values even if you feed the same locations to both Navigators. We have various non-deterministic logic inside: for example we download and load tiles in memory in background thread. In case of two active Navigators we can end up in situation when one Navigator has loaded tile in memory, but another Navigator instance hasn't. In this case they may return different results in `getStatus`.
Probably it can be somehow solved if we would allow to pass tile cache instance via `Navigator` constructor, but it seems to be difficult to achieve now, because of bindgen restrictions.

IodaMikeMapbox (approved):

yaakoh (approved):

1ec5:
Once this is in, it’ll be nice to get a speed limit view hooked up as well: #2847.

MaximAlien:
`PassiveLocationManager` was added to `CarPlayMapViewController` in `release-v2.0` branch in scope of https://github.com/mapbox/mapbox-navigation-ios/pull/2826.

zugaldia:
What's next to land this PR?

MaximAlien:
BTW similar functionality is available in `2.x` in `CarPlayMapViewController.setupPassiveLocationManager`.

zugaldia:
Ah, I missed this is a 1.x feature :+1: (there's less pressure on timeline in that case).

MaximAlien:
Based on this [comment](https://github.com/mapbox/mapbox-navigation-ios/pull/2588#issuecomment-819016036) this change is no longer required, as `PassiveLocationManager` is already available on CarPlay.

Issue #2589, opened 2020-09-02
Replace Route with RouteResponse plus route index


1ec5:
The original vision behind mapbox/mapbox-directions-swift#391 was that a client such as the navigation SDK would operate mostly in terms of a whole RouteResponse plus a route index into that response. That way, we don’t have to copy attributes of the response into each individual route, as seen in mapbox/mapbox-directions-swift#452. Unfortunately, we never got quite that far in mapbox/mapbox-navigation-ios#2350:

* [ ] `MapboxNavigationService(route:routeOptions:directions:locationSource:eventsManagerType:simulating:routerType:)` should take a RouteResponse, not an individual Route.
* [ ] `NavigationViewController(for:routeOptions:navigationOptions:)` should take a NavigationService instead of a Route.

This much sounds simple, but there are potentially quite a few knock-on effects throughout the codebase. It’s too late to get these backwards-incompatible changes into v1.0. We can get them into a subsequent minor release as long as we take steps to ensure backwards compatibility. A convenience initializer for MapboxNavigationService can wrap a Route in a RouteResponse, and a convenience initializer for NavigationViewController can wrap a Route in a RouteResponse and MapboxNavigationService.

/cc @mapbox/navigation-ios

Udumft:
Corresponding changes happen in these PRs for [directions](https://github.com/mapbox/mapbox-directions-swift/pull/562) and for [NavSDK](https://github.com/mapbox/mapbox-navigation-ios/pull/3182).

PR #2366, opened 2020-04-24
Route Refresh

Udumft:
Traffic congestion segments along the route line and the estimated arrival time periodically update to reflect current conditions when using the `DirectionsProfileIdentifier.automobileAvoidingTraffic` profile. These updates correspond to the new `Notification.Name.routeControllerDidRefreshRoute` notification, `NavigationServiceDelegate.navigationService(_:didRefresh:)` method, and `NavigationViewControllerDelegate.navigationViewController(_:didRefresh:)` method. To disable these updates, set `RouteOptions.refreshingEnabled` to `false`.

The `NavigationViewController.route` and `NavigationService.route` properties are now read-only. To change the route that the user is traveling along, set the `NavigationViewController.indexedRoute` or `NavigationService.indexedRoute` property instead, pairing the route with the index of the route in the original `RouteResponse` object.

The following methods now require a route index to be passed in as an argument:

* `NavigationViewController(for:routeIndex:routeOptions:navigationOptions:)`
* `NavigationViewController(route:routeIndex:routeOptions:navigationService:)`
* `CarPlayManagerDelegate.carPlayManager(_:navigationServiceAlong:routeIndex:routeOptions:desiredSimulationMode:)`
* `MapboxNavigationService(route:routeIndex:routeOptions:)`
* `MapboxNavigationService(route:routeIndex:routeOptions:directions:locationSource:eventsManagerType:simulating:routerType:)`
* `RouteProgress(route:routeIndex:options:legIndex:spokenInstructionIndex:)`
* `RouteProgress(route:routeIndex:options:legIndex:spokenInstructionIndex:)`
* `Router(along:routeIndex:options:directions:dataSource:)`
* `RouteController(along:routeIndex:options:directions:dataSource:)`

Resolves #2284 

----

Work-In-Progress
So far added basic implementation for calling Route Refresh. Experimenting with `RouteProgress` to refresh it's Route contents (suggestions are welcome).
Thinking about route refresh period setting to be related with better route search period

PR #2591, opened 2020-09-02
Pair route with route index

1ec5:
This PR updates #2366 to work with the latest changes in mapbox/mapbox-directions-swift#420 that decapsulate route refreshing a bit. I’ve opened a separate PR because the changes are massive and gangly, and I’m open to reconsidering the approach to make it less disruptive. See the changes to CHANGELOG.md for details.

With this PR, classes that previously kept track of the route now also keep track of the route’s index in the overall route response and use it when refreshing the route. This entailed adding a required route index argument to a litany of public and private methods and initializers, though it was possible to hard-code 0 in many places due to the prior use of `routes.first`. With these changes, it becomes clear how we’d transition the public API to treating the route as merely an indexed item in the overall route response: #2589. We could bite that off here too; it all comes down to whether the (route, route index) tuple would be easier or harder to transition to.

Additionally and critically, the code to merge the skeletal refreshed route into the original route has moved from mapbox/mapbox-directions-swift#420 into this PR.

/cc @mapbox/navigation-ios


1ec5:
It’s expected that the Xcode_11.4.1_iOS_12.2_CP_update and Xcode_11.4.1_iOS_12.2_CP_install build bots would fail, because this PR depends on mapbox/mapbox-directions-swift#420, which hasn’t been released yet.

CHANGELOG.md
```diff
@@ -53,6 +54,17 @@
...
+* The following methods now require a route index to be passed in as an argument ([#2366](https://github.com/mapbox/mapbox-navigation-ios/pull/2366)):
+   * `NavigationViewController(for:routeIndex:routeOptions:navigationOptions:)`
+   * `NavigationViewController(route:routeIndex:routeOptions:navigationService:)`
+   * `CarPlayManagerDelegate.carPlayManager(_:navigationServiceAlong:routeIndex:routeOptions:desiredSimulationMode:)`
+   * `MapboxNavigationService(route:routeIndex:routeOptions:)`
+   * `MapboxNavigationService(route:routeIndex:routeOptions:directions:locationSource:eventsManagerType:simulating:routerType:)`
+   * `RouteProgress(route:routeIndex:options:legIndex:spokenInstructionIndex:)`
+   * `RouteProgress(route:routeIndex:options:legIndex:spokenInstructionIndex:)`
+   * `Router(along:routeIndex:options:directions:dataSource:)`
+   * `RouteController(along:routeIndex:options:directions:dataSource:)`
```
1ec5, line 67: This blurb nicely encapsulates why I’m nervous about this PR. Adding these extra arguments is pretty safe from a runtime stability standpoint, other than in some of the CarPlay code that relies on casting `Any` into tuples, but it’s always a big deal to add complexity to the basic setup workflow in NavigationViewController particularly.

Cartfile
```diff
@@ -2,7 +2,7 @@ binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" ~> 2.3.0
 binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.json" ~> 18.0.2
 binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps/mapbox-ios-sdk-dynamic.json" == 6.1.0-beta.2
 binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon-ios.json" == 4.0.0
-github "mapbox/mapbox-directions-swift" ~> 0.33
+github "mapbox/mapbox-directions-swift" "2284-route-refresh-directions"
```
1ec5, line 5: This is temporary. We’ll release v0.34.0 before or right after merging #2366 so that the CocoaPods integration tests pass.

Example/ViewController.swift
```diff
@@ -260,7 +260,7 @@ class ViewController: UIViewController {
     func startBasicNavigation() {
         guard let response = response, let route = response.routes?.first, case let .route(routeOptions) = response.options else { return }
         
-        let service = navigationService(route: route, options: routeOptions)
+        let service = navigationService(route: route, routeIndex: 0, options: routeOptions)
```
1ec5, line 263: In many cases, the developer will be able to hard-code 0 as the route index after getting the `first` route out of the route array. But I made the argument required because I didn’t want any code that deals with multiple routes to go unnoticed during an upgrade.

MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -130,6 +132,13 @@ public extension Notification.Name {
      */
     static let routeControllerProgressDidChange: Notification.Name = .init(rawValue: "RouteControllerProgressDidChange")
     
+    /**
+     Posted when `RouteController` receives updated information about the current route.
+     
+     The user info dictionary contains the key `RouteController.NotificationUserInfoKey.routeProgressKey`.
+     */
+    static let routeControllerDidRefreshRoute: Notification.Name = .init(rawValue: "RouteControllerDidRefreshRoute")
```
1ec5, line 140: Is it weird that a notification about refreshing a _route_ comes with a route _progress_ object as user info? The idea is that the route progress object contains the route as well as other members that are affected by refreshing, so it’s handy to have immediate access to the whole route progress object.

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -30,7 +30,7 @@ public protocol NavigationServiceDelegate: class, UnimplementedLogging {
     /**
      Called immediately before the navigation service calculates a new route.
      
-     This method is called after `navigationService(_:shouldRerouteFrom:)` is called, simultaneously with the `NavigationServiceWillReroute` notification being posted, and before `navigationService(_:didRerouteAlong:)` is called.
+     This method is called after `navigationService(_:shouldRerouteFrom:)` is called, simultaneously with the `Notification.Name.routeControllerWillReroute` notification being posted, and before `navigationService(_:didRerouteAlong:)` is called.
```
1ec5, line 33: A bunch of these references were always inaccurate. I don’t think we’ve ever prefixed the notification names with `NavigationService`.

MapboxCoreNavigation/Router.swift
```diff
@@ -12,6 +12,11 @@ public protocol RouterDataSource: class {
     var locationProvider: NavigationLocationManager.Type { get }
 }
 
+/**
+ A route and its index in a `RouteResponse` that sorts routes from most optimal to least optimal.
+ */
+public typealias IndexedRoute = (Route, Int)
```
1ec5, line 18: This is the most generic place I could find for a type that ends up getting used all over the place. The “indexed” pattern is inspired by Turf for Swift. The idea is to force code that sets the route to a new value to also set the route index to a potentially new value at the same time, avoiding stale state or race conditions.

MapboxNavigation/CPTrip.swift
```diff
@@ -36,7 +36,7 @@ extension CPTrip {
             let routeChoice = CPRouteChoice(summaryVariants: summaryVariants,
                                             additionalInformationVariants: [route.description],
                                             selectionSummaryVariants: [route.description])
-            let info: (Route, RouteOptions) = (route: route, options: routeOptions)
+            let info: (Route, Int, RouteOptions) = (route: route, routeIndex: routeIndex, options: routeOptions)
             routeChoice.userInfo = info
```
1ec5, line 40: It’s frustrating that the only way to associate arbitrary data with UI elements that CarPlay controls is to stuff it in an untyped property and cast it later. As tail work, it might be better to define a set of dictionary keys and set this property to a dictionary, similar to when posting a Notification.

MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -199,7 +199,7 @@ class CarPlayManagerTests: XCTestCase {
             CLLocationCoordinate2D(latitude: 37.764793, longitude: -122.463161),
             CLLocationCoordinate2D(latitude: 34.054081, longitude: -118.243412),
         ])
-        choice.userInfo = (Fixture.route(from: "route-with-banner-instructions", options: options), options)
+        choice.userInfo = (Fixture.route(from: "route-with-banner-instructions", options: options), 0, options)
```
1ec5, line 202: The route index is almost always 0 in the unit tests and integration tests, because `Fixture.route(from:options:)` always returns the first route, and few of the fixtures come with multiple routes anyways.

chezzdev (approved):
The changes look good to me.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -525,6 +535,14 @@ extension NavigationViewController: NavigationServiceDelegate {
         delegate?.navigationViewController(self, didFailToRerouteWith: error)
     }
     
+    public func navigationService(_ service: NavigationService, didRefresh routeProgress: RouteProgress) {
+        for component in navigationComponents {
+            component.navigationService(service, didRefresh: routeProgress)
+        }
+        
+        delegate?.navigationViewController(self, didRefresh: routeProgress)
```
MaximAlien, line 543: What about adding this functionality to `ViewController.swift`? I think it'd be beneficial for testing purposes to have it ready there (I can submit my changes if it makes sense as I did some testing to try it out).

MapboxNavigation/NavigationViewController.swift
```diff
@@ -525,6 +535,14 @@ extension NavigationViewController: NavigationServiceDelegate {
         delegate?.navigationViewController(self, didFailToRerouteWith: error)
     }
     
+    public func navigationService(_ service: NavigationService, didRefresh routeProgress: RouteProgress) {
```
MaximAlien, line 538: I think in this place when should redraw all congestion segments (if they were updated) by calling `mapView?.updateRoute(routeProgress)`. At the same time current implementation of congestion segments calculation re-uses already existing `lineGradient` stops which means that route in most recent `RouteProgress` will not be used. Previously I was aiming to prevent re-calculation because for routes with many congestion segments such logic might cause performance drawbacks, on the other hand in current situation there seems no other way.

MapboxNavigation/NavigationViewController.swift
MaximAlien: We can also add separate example in `navigation-ios-examples` to showcase this specific feature.

1ec5: Sure, feel free to push what you have.

MapboxNavigation/NavigationViewController.swift
1ec5: Do you mean in RouteMapViewController?

https://github.com/mapbox/mapbox-navigation-ios/blob/a303150d3a3762799048914d58de793b6ffbd9f4/MapboxNavigation/RouteMapViewController.swift#L522-L524

CHANGELOG.md
MaximAlien: If I'm not mistaken in all public API methods which were modified to contain `routeIndex` `RouteOptions` argument is also present. Would it work if we add `routeIndex` as part of options?

MapboxNavigation/NavigationViewController.swift
MaximAlien: I think most recent update in `RouteMapViewController` should work (removing all route layers, showing them for traveled distance = 0.0, updating them for traveled distance from `RouteProgress`). On the other hand it defats the idea of reusing layers (I think the only data we need from `RouteProgress` in such case is `shape`). We can leave current implementation, but revise it in future.

MapboxCoreNavigation/CoreConstants.swift
MaximAlien: I think it's not really obvious, I'd expect route Route object to be there. We also have `routeControllerProgressDidChange`, but it doesn't contain only `RouteProgress` updates.

Example/ViewController.swift
```diff
@@ -240,6 +240,11 @@ class ViewController: UIViewController {
         waypoints.insert(userWaypoint, at: 0)
 
         let options = NavigationRouteOptions(waypoints: waypoints)
+        
+        // Get periodic updates regarding changes in estimated arrival time and traffic congestion segments along the route line.
+        options.profileIdentifier = .automobileAvoidingTraffic
```
1ec5, line 245: This is the default value in NavigationRouteOptions.

Example/ViewController.swift
```diff
@@ -240,6 +240,11 @@ class ViewController: UIViewController {
...
+        options.refreshingEnabled = true
```
1ec5, line 246: Per https://github.com/mapbox/mapbox-directions-swift/pull/420#discussion_r483287478, NavigationRouteOptions should set this value by default.

MapboxNavigation/NavigationViewController.swift
1ec5: Yeah, let’s deal with any performance optimizations as tail work.

CHANGELOG.md
1ec5: A single RouteOptions may be associated with more than one route, so I don’t think that would work. Unless you mean NavigationOptions?

Cartfile
1ec5: We need to also update the CocoaPods podspecs and CocoaPods integration tests once v1.0.0-rc.1 is pushed to CocoaPods trunk.

PR #2592, opened 2020-09-03
Cherry pick commits for v1.0.0-rc.3.

MaximAlien:
Cherry pick commits for v1.0.0-rc.3.


1ec5:
> Turns out I cannot push changes to release-1.0 directly, thus created this PR.

Not sure why that is; I’ve been able to push directly to the release branch a few times now.

MapboxNavigation/NavigationViewControllerDelegate.swift
```diff
@@ -118,7 +118,7 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate{
 
      If this method is unimplemented, the navigation view controller’s map view draws the casing for the alternative route lines using an `MGLLineStyleLayer`.
     */
-    func navigationViewController(_ navigationViewController: NavigationViewController, alternateRouteCasingStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
+    func navigationViewController(_ navigationViewController: NavigationViewController, alternativeRouteCasingStyleLayerWithIdentifier identifier: String, source: MGLSource) -> MGLStyleLayer?
```
1ec5, line 121: I forgot to mention it in https://github.com/mapbox/mapbox-navigation-ios/pull/2580#discussion_r481957736, but renaming this method is a backwards-incompatible change that we should mention in the changelog. Normally, the Swift compiler will complain if a class fails to implement a method declared by a protocol it conforms to. However, we use default implementations to suppress that behavior, simulating the Objective-C optional methods that are characteristic of the Cocoa delegate pattern. So the only way the developer would know about the rename would be the changelog entry and – if they’re really paying attention – some console spew from UnimplementedLogging.

1ec5 (approved):
(No need to hold up this PR for the changelog update suggested below.)

PR #2593, opened 2020-09-03
Add tiles manager

IodaMikeMapbox:
The pr adds a convenient way to create navigators configured with the tiles and tile endpoint config.

For now the sdk contains three places where the navigator objects can be created.
Untitled Diagram

Regular arrows shows who creates the objects, for example User app creates an instance of `NavigationViewController`, etc. Hollow arrow shows inheritance. The chart does not display all dependencies.

The problem is that we don't have a place which can control all of these three places and manage the navigators according with the recommendations from the navnative team. The recommendations are:
1. Navigator with a route set will map raw coords most likely to the route instead of to the road graph which is not suitable for free drive mode.
1. Navigator which is used for route tracking will be work more accurate and will have higher performance if it will be configured with navigation tiles as well.
1. It's not recommended to have more than one navigators at the same time because of performance.

The idea for this pr is to use `FileManager` as a point of synchronization for the navigators created from different places.
Untitled Diagram (1)

TilesManager is able to:
1. Get the list of the downloaded tiles by checking content of the `.mapbox/` folder in the cache directory.
1. Fetch the list of available versions using MapboxDirections.
1. Delete outdated tiles.

Resolves mapbox/navigation-sdks#430

### DIFF
--- a/MapboxCoreNavigation/NavigatorTilesManager.swift
+++ b/MapboxCoreNavigation/NavigatorTilesManager.swift
@@ -1,0 +1,151 @@
+import Foundation
+import MapboxDirections
+import MapboxNavigationNative
+
+private let tilesManager = NavigatorTilesManager()
+
+extension Navigator {
+    static func createWith(profile: SettingsProfile, config: NavigatorConfig, customConfig: String) -> Navigator {
+        let navigator = Navigator(profile: profile, config: config, customConfig: customConfig)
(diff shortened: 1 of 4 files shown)


Udumft (commented):
This approach looks good to me. I understand that this is just a draft, but here's some NITs I found:

1ec5:
Stale; @Udumft is working on a replacement.

Issue #2489, opened 2020-07-26
NextBannerView is overlapping with StatusView.


MaximAlien:
**Mapbox Navigation SDK version:** `1.0.0-beta.1`

In some cases NextBannerView is overlapping with StatusView which leads to broken UI. See example:
<img width="395" alt="Screen Shot 2020-07-26 at 12 14 57 PM" src="https://user-images.githubusercontent.com/1496498/88487448-02306400-cf3a-11ea-9a21-10cf6e2b151c.png">

1ec5:
This issue reproduces easily by choosing a route that has two very closely spaced maneuvers right at the beginning of the route. For example, set the iOS Simulator’s simulated location to Apple, then get a route out to De Anza Boulevard. The two turn maneuvers in the IL2 parking lot cause this next banner to appear and persist across multiple maneuvers. Only showing and hiding the step banner clears it out.

PR #2607, opened 2020-09-04
Fix NextBannerView presentation.

MaximAlien:
Closing #2489.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -678,7 +678,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         let fractionTraveled = routeProgress.fractionTraveled
         
         // In case if route was fully travelled - remove main route and its casing.
-        if Int(fractionTraveled) == 1 {
+        if fractionTraveled >= 1.0 {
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -678,7 +678,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         let fractionTraveled = routeProgress.fractionTraveled
         
         // In case if route was fully travelled - remove main route and its casing.
-        if fractionTraveled == 1.0 {
+        if Int(fractionTraveled) == 1 {
```
1ec5, line 681: Was the bug that `fractionTraveled` could increase beyond 1.0 and get missed by this logic? If so, `>=` would be a little clearer, but this works.

1ec5 (approved):

MapboxNavigation/NavigationMapView.swift
MaximAlien: Yep, there are cases sometimes when `fractionTraveled` can be `1.0000000000000002` or similar. I guess you're right `>=` will be clearer, will change to use this check.

PR #2610, opened 2020-09-08
Make timestamps consumed by Navigator.getStatus consistent

### LINKED ISSUE #2583: RouteController should get navigation statuses based on a consistent timestamp
RouteController currently calls `Navigator.status(at:)` (a wrapper around `Navigator.getStatusForTimestamp(_:)`) with inconsistent or user-defined timestamps. Historically, the expectation was that the navigator would be able to march to and fro along the timeline of the route. However, this has never quite been the case in MapboxNavigationNative. A forthcoming version of MapboxNavigationNative will remove the ability to ask for the navigation status at an arbitrary time, to avoid misleading clients into thinking it has a capability that doesn’t quite exist.

## Projected location

For forward compatibility, we need to remove `RouteController.projectedLocation(for:)`, which allows applications to request the location at an arbitrary time:

https://github.com/mapbox/mapbox-navigation-ios/blob/ebab7d0881d33a975e265e8db20ca88d2eb7918a/MapboxCoreNavigation/RouteController.swift#L340-L346

## Current location

RouteController internally gets navigation statuses based on the time of the last Core Location update, which accounts for potential interruptions in location updates:

https://github.com/mapbox/mapbox-navigation-ios/blob/ebab7d0881d33a975e265e8db20ca88d2eb7918a/MapboxCoreNavigation/RouteController.swift#L196

However, a very commonly used property of RouteController gets the status as of the current time, which can easily fall after the last Core Location update:

https://github.com/mapbox/mapbox-navigation-ios/blob/893f57f0bbb5d6f03fac0fe8ed1408519bb150fe/MapboxCoreNavigation/RouteController.swift#L89-L90

This is likely the usual trigger of #2113. A straightforward fix would be to remember the last location’s timestamp and get the status based on that timestamp.

## Leg advancing

`RouteController.advanceLegIndex(location:)` is also problematic because it can take an arbitrary location and passes its timestamp to the navigator:

https://github.com/mapbox/mapbox-navigation-ios/blob/ebab7d0881d33a975e265e8db20ca88d2eb7918a/MapboxCoreNavigation/RouteController.swift#L348-L351

This method is required by the Router protocol, but it’s only used internally by LegacyRouteController to increment `RouteProgress.legIndex`. The example application used to call this method, too, to demonstrate manually advancing to the next leg after presenting an arrival interstitial at an intermediate waypoint. However, nowadays, it increments the leg index directly and restarts the navigation service:

https://github.com/mapbox/mapbox-navigation-ios/blob/ebab7d0881d33a975e265e8db20ca88d2eb7918a/Example/ViewController.swift#L508-L511

I’m unsure if that works quite as it’s supposed to. It would be better to encapsulate those details in `RouteController.advanceLegIndex(location:)`, except that the method should take no parameters and always increment the leg index by one. RouteController can easily implement that functionality by calling `RouteController.updateRouteLeg(to:)`.

/cc @mapbox/navigation-ios @mapbox/navnative @d-prukop

chezzdev:
The PR addresses 3 points described in https://github.com/mapbox/mapbox-navigation-ios/issues/2583 to enhance the consistency across timestamps provided to `Navigator.getStatus`.

Fixes #2583.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,8 +76,8 @@
 * If the user’s raw course as reported by Core Location differs significantly from the direction of the road ahead, the camera and user puck are oriented according to the raw course. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * `RouteController` now tracks the user’s location more accurately within roundabouts. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * Fixed an issue where departure instructions were briefly missing when beginning turn-by-turn navigation. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
-* Removed the `RouteController.projectedLocation(for:)` method in favor of `RouteController.location`. It is no longer possible to predict the user’s location at an arbitrary time. ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
-* Renamed the `Router.advanceLegIndex(location:)` method to `Router.advanceLegIndex()`. It is no longer possible to advance to an arbitrary leg using this method. ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
+* Removed the `RouteController.projectedLocation(for:)` method in favor of `RouteController.location`. It is no longer possible to predict the user’s location at an arbitrary time. ([#2610](https://github.com/mapbox/mapbox-navigation-ios/pull/2610))
+* Renamed the `Router.advanceLegIndex(location:)` method to `Router.advanceLegIndex()`. It is no longer possible to advance to an arbitrary leg using this method. ([#2610](https://github.com/mapbox/mapbox-navigation-ios/pull/2610))
 
 ### Other changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@
 * If the user’s raw course as reported by Core Location differs significantly from the direction of the road ahead, the camera and user puck are oriented according to the raw course. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * `RouteController` now tracks the user’s location more accurately within roundabouts. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * Fixed an issue where departure instructions were briefly missing when beginning turn-by-turn navigation. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
+* Removed `RouteController.projectedLocation(for:)` method and `location` parameter from `advanceLegIndex` method to avoid inconsistencies with utilized timestamps ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
 
 ### Other changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,8 @@
 * If the user’s raw course as reported by Core Location differs significantly from the direction of the road ahead, the camera and user puck are oriented according to the raw course. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * `RouteController` now tracks the user’s location more accurately within roundabouts. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * Fixed an issue where departure instructions were briefly missing when beginning turn-by-turn navigation. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
-* Removed `RouteController.projectedLocation(for:)` method and `location` parameter from `advanceLegIndex` method to avoid inconsistencies with utilized timestamps ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -76,6 +76,7 @@
 * If the user’s raw course as reported by Core Location differs significantly from the direction of the road ahead, the camera and user puck are oriented according to the raw course. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * `RouteController` now tracks the user’s location more accurately within roundabouts. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * Fixed an issue where departure instructions were briefly missing when beginning turn-by-turn navigation. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
+* Removed `RouteController.projectedLocation(for:)` method and `location` parameter from `advanceLegIndex` method to avoid inconsistencies with utilized timestamps ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
```
1ec5, line 79:
```suggestion
* Removed the `RouteController.projectedLocation(for:)` method in favor of `RouteController.location`. It is no longer possible to predict the user’s location at an arbitrary time. ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
* Renamed the `Router.advanceLegIndex(location:)` method to `Router.advanceLegIndex()`. It is no longer possible to advance to an arbitrary leg using this method. ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
```

MapboxCoreNavigation/RouteController.swift
```diff
@@ -188,7 +194,6 @@ open class RouteController: NSObject {
     /// updateRouteLeg is used to notify nav-native of the developer changing the active route-leg.
     private func updateRouteLeg(to value: Int) {
         let legIndex = UInt32(value)
-        navigator.changeRouteLeg(forRoute: 0, leg: legIndex)
```
1ec5, line 191: This redundant call to `MBNavigator.changeRouteLeg(forRoute:leg:)` was introduced in c5ee2acaacbb845f20294c2a50014c609938853a for #2138. It looks like an oversight to me too, rather than a workaround for something.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -91,15 +91,21 @@ open class RouteController: NSObject {
...
+
+        let status = navigator.status(at: locationUpdateDate)
         guard status.routeState == .tracking || status.routeState == .complete else {
             return nil
         }
         return CLLocation(status.location)
     }
-    
+
+    private var lastLocationUpdateDate: Date?
```
1ec5, line 105: Can this property be computed from `rawLocation.timestamp` to avoid parallel state?

1ec5 (approved):
In addition to the feedback below, can we have ViewController call `Router.advanceLegIndex()` and restart the navigator instead of reaching into the route progress and modifying the leg index directly, just to demonstrate that this refactor works correctly?

https://github.com/mapbox/mapbox-navigation-ios/blob/ebab7d0881d33a975e265e8db20ca88d2eb7918a/Example/ViewController.swift#L508-L511

CHANGELOG.md
chezzdev: Applied, thanks.

MapboxCoreNavigation/RouteController.swift
chezzdev: Yes, good idea.

CHANGELOG.md
```diff
@@ -76,7 +76,8 @@
...
 * `RouteController` now tracks the user’s location more accurately within roundabouts. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
 * Fixed an issue where departure instructions were briefly missing when beginning turn-by-turn navigation. ([#2417](https://github.com/mapbox/mapbox-navigation-ios/pull/2417))
-* Removed `RouteController.projectedLocation(for:)` method and `location` parameter from `advanceLegIndex` method to avoid inconsistencies with utilized timestamps ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
+* Removed the `RouteController.projectedLocation(for:)` method in favor of `RouteController.location`. It is no longer possible to predict the user’s location at an arbitrary time. ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
+* Renamed the `Router.advanceLegIndex(location:)` method to `Router.advanceLegIndex()`. It is no longer possible to advance to an arbitrary leg using this method. ([#2583](https://github.com/mapbox/mapbox-navigation-ios/issues/2583))
```
1ec5, line 80: Nit: Typically we’re linking to PRs where the problem and solution are laid out in detail.

```suggestion
* Removed the `RouteController.projectedLocation(for:)` method in favor of `RouteController.location`. It is no longer possible to predict the user’s location at an arbitrary time. ([#2610](https://github.com/mapbox/mapbox-navigation-ios/pull/2610))
* Renamed the `Router.advanceLegIndex(location:)` method to `Router.advanceLegIndex()`. It is no longer possible to advance to an arbitrary leg using this method. ([#2610](https://github.com/mapbox/mapbox-navigation-ios/pull/2610))
```

1ec5 (approved):

CHANGELOG.md
chezzdev: Let's keep it consistent, thanks for the heads-up.

PR #2612, opened 2020-09-08
Removed the timer and state management for building highlights

d-prukop:
Removed the timer and state management of building highlights within the NavigationMapView.

### DIFF
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -77,6 +77,8 @@ class ViewController: UIViewController {
         print(error.localizedDescription)
         self?.presentAlert(message: error.localizedDescription)
     }
+    
+    private var foundAllBuildings = false
 
     // MARK: - Init
     
@@ -162,7 +164,7 @@ class ViewController: UIViewController {
     
         // Example of highlighting buildings in 2d and directly using the API on NavigationMapView.
         let buildingHighlightCoordinates = waypoints.compactMap { $0.targetCoordinate }
-        mapView.highlightBuildings(at: buildingHighlightCoordinates, in3D: false)
+        foundAllBuildings = mapView.highlightBuildings(at: buildingHighlightCoordinates, in3D: false)
 
         requestRoute()
     }
@@ -443,6 +445,15 @@ extension ViewController: MGLMapViewDelegate {
     
     func mapView(_ mapView: MGLMapView, lineWidthForPolylineAnnotation annotation: MGLPolyline) -> CGFloat {
         return annotation == trackPolyline || annotation == rawTrackPolyline ? 4 : 1
+    }
+    
+    func mapView(_ mapView: MGLMapView, regionDidChangeWith reason: MGLCameraChangeReason, animated: Bool) {
+        if foundAllBuildings == false, let navMapView = mapView as? NavigationMapView, reason != .programmatic {
+            let buildingHighlightCoordinates = waypoints.compactMap { $0.targetCoordinate }
+            if buildingHighlightCoordinates.count > 0 {
+                foundAllBuildings = navMapView.highlightBuildings(at: buildingHighlightCoordinates, in3D: false)
+            }
+        }
     }
 }
 
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -447,8 +447,8 @@ extension ViewController: MGLMapViewDelegate {
         return annotation == trackPolyline || annotation == rawTrackPolyline ? 4 : 1
     }
     
-    func mapView(_ mapView: MGLMapView, regionDidChangeWith reason: MGLCameraChangeReason, animated: Bool) {
-        if foundAllBuildings == false, let navMapView = mapView as? NavigationMapView, reason != .programmatic {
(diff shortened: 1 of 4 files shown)


Example/ViewController.swift
```diff
@@ -444,6 +446,15 @@ extension ViewController: MGLMapViewDelegate {
     func mapView(_ mapView: MGLMapView, lineWidthForPolylineAnnotation annotation: MGLPolyline) -> CGFloat {
         return annotation == trackPolyline || annotation == rawTrackPolyline ? 4 : 1
     }
+    
+    func mapView(_ mapView: MGLMapView, regionDidChangeWith reason: MGLCameraChangeReason, animated: Bool) {
+        if foundAllBuildings == false, let navMapView = mapView as? NavigationMapView, reason != .programmatic {
```
1ec5, line 451: Would it be problematic for this view controller to react to any camera change, not just programmatic ones? `MGLMapViewDelegate.mapView(_:regionDidChangeWith:animated:)` is an undocumented method that may be removed from future map SDK versions. ViewController is only in the example applications in this repository, but we would need to avoid using this method in any developer-facing examples in navigation-ios-examples.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1433,42 +1425,24 @@ extension NavigationMapView {
...
-    public func highlightBuildings(at coordinates: [CLLocationCoordinate2D], in3D extrudesBuildings: Bool = true) {
-        guard coordinates != currentCoordinatesForBuildingsToHighlight else { return }
-        currentCoordinatesForBuildingsToHighlight = coordinates
-        buildingHighlightPollingTimer?.invalidate()
-        highlightedBuildingIdentifiers = Set()
-        findAndHighlightBuildings(at: coordinates, in3D: extrudesBuildings)
-        buildingHighlightPollingTimer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: true) { [weak self] timer in
-            self?.findAndHighlightBuildings(at: coordinates, in3D: extrudesBuildings)
-        }
+    public func highlightBuildings(at coordinates: [CLLocationCoordinate2D], in3D extrudesBuildings: Bool = true) -> Bool {
```
1ec5, line 1431: If the order of the coordinates is unimportant, the argument should be a set rather than an array.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -589,18 +591,22 @@ extension NavigationViewController: NavigationServiceDelegate {
             currentLeg = progress.currentLeg
             passedApproachingDestinationThreshold = false
             mapViewController?.supressAutomaticAltitudeChanges = false
```
1ec5, line 593: Nitpickiest of nits: `supressAutomaticAltitudeChanges` should be spelled `suppressAutomaticAltitudeChanges`, with an extra P.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -589,18 +591,22 @@ extension NavigationViewController: NavigationServiceDelegate {
...
+            foundAllBuildings = false
             if let mapView = mapView {
                 mapView.altitude = mapView.defaultAltitude
             }
         }
         
-        if let mapView = mapView, passedApproachingDestinationThreshold == false, waypointStyle != .annotation, let currentLegWaypoint = progress.currentLeg.destination?.targetCoordinate, progress.currentLegProgress.distanceRemaining < approachingDestinationThreshold {
+        if let mapView = mapView, passedApproachingDestinationThreshold == false, progress.currentLegProgress.distanceRemaining < approachingDestinationThreshold {
             passedApproachingDestinationThreshold = true
             mapViewController?.supressAutomaticAltitudeChanges = true
```
1ec5, line 602: Do we still want to set this value to true even if the waypoint style is `annotation`?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1433,42 +1425,24 @@ extension NavigationMapView {
      
      - parameter coordinates: Coordinates which represent buildings locations.
```
1ec5, line 1426:
```suggestion
     - parameter coordinates: Coordinates which represent building locations.
```

Example/ViewController.swift
d-prukop: Changed this to `mapViewRegionIsChanging(_ mapView:)`

MapboxNavigation/NavigationMapView.swift
d-prukop: I couldn't store `CLLocationCoordinate2D` in a `Set`. Leaving this as-is.

MaximAlien (commented):
I did some more testing again on `iPhone SE` device and still see that there are cases in which building is not highlighted beforehand. I think we should perform more testing on different devices (incl. iPads and real-life driving) to find and fix those issues (possibly in `1.1` release).

MaximAlien:
I've slightly refactored initial implementation to be able to highlight building in case if distance to destination is within threshold right after navigation was started. Buildings also should not be selected randomly now (after arrival to destination). There are still some edge cases (marked with `FIXME`). We can tackle them in upcoming releases.

Issue #2561, opened 2020-08-21
Map info button moves after tapping follow mode (?) button in top right corner


zacharyblank:
**Mapbox Navigation SDK version:**
1.0.0

After tapping the follow mode (?) button in the top right corner the map info button moves to the middle of the screen as seen here:

![IMG_7A930A97B020-1](https://user-images.githubusercontent.com/759763/90906183-17a47c80-e386-11ea-929b-33fa84f88ab3.jpeg)

PR #2613, opened 2020-09-09
Fix info button alignment and improve bottom banner height constraints for various device models.

MaximAlien:
Closing #2561.

### DIFF
--- a/MapboxNavigation/BottomBannerViewLayout.swift
+++ b/MapboxNavigation/BottomBannerViewLayout.swift
@@ -67,8 +67,6 @@ extension BottomBannerViewController {
     }
     
     fileprivate func setupVerticalCompactLayout(_ c: inout [NSLayoutConstraint]) {
-        c.append(bottomBannerView.heightAnchor.constraint(equalToConstant: 50))
-        
         c.append(cancelButton.widthAnchor.constraint(equalTo: bottomBannerView.heightAnchor))
         c.append(cancelButton.topAnchor.constraint(equalTo: bottomBannerView.topAnchor))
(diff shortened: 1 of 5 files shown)


MapboxNavigation/BottomBannerViewLayout.swift
```diff
@@ -95,9 +94,6 @@ extension BottomBannerViewController {
     }
     
     fileprivate func setupVerticalRegularLayout(_ c: inout [NSLayoutConstraint]) {
-        let size = bottomBannerView.heightAnchor.constraint(equalToConstant: 80)
```
MaximAlien, line 98: Since `bottomBannerView` is added inside `BottomBannerViewController` which is in turn embedded into `bottomBannerContainerView` setting height constraint has no effect here.

1ec5 (approved):
Looks alright at a glance, assuming you’ve considered safe area insets on the relevant devices. If this was a regression versus v0.40.0, remember to add something to the changelog about it.

MaximAlien:
> Looks alright at a glance, assuming you’ve considered safe area insets on the relevant devices. If this was a regression versus v0.40.0, remember to add something to the changelog about it.

I've intentionally left `attributionButton` and `logoView` not to use `safeTrailingAnchor` and `safeLeadingAnchor` to prevent overlapping with feedback button in landscape mode. Current look:

Screen Shot 2020-09-10 at 11 45 52 AM
Screen Shot 2020-09-10 at 11 46 05 AM

Also even though this issue was still present in `0.40.0` I'll mention it in `CHANGELOG.md`.

Issue #2511, opened 2020-07-30
Guidance card UI is not aligned correctly in landscape mode.


MaximAlien:
**Mapbox Navigation SDK version:** `1.0.0-beta.1`
When starting navigation in portrait mode and then switching to landscape guidance card does not align properly to fill almost all screen width, this leads to inconsistency when switching between cards. Also in landscape mode Feedback button is not accessible.

Issue which occurs after rotation:
![output1](https://user-images.githubusercontent.com/1496498/88978113-eb5c7b00-d273-11ea-9151-d25a6a65758c.gif)

Expected landscape layout after rotation (except missing Feedback button, which should be visible):
![output2](https://user-images.githubusercontent.com/1496498/88978133-fca58780-d273-11ea-8ad2-acd2a41d21ba.gif)

PR #2614, opened 2020-09-10
Update size of collection view cell with instructions when device is rotated.

MaximAlien:
Closing #2511.


avi-c:
I think it's fine to make this change, but I know the previous behavior was there deliberately from the past. We never really resolved what the guidance cards should do when the device had a wide screen (phone or iPad). It seems weird to make a super wide guidance card that is nearly empty, but also seems weird and possibly confusing if there are several guidance cards fully on screen at once since the user has to make sure to read just the left-most one. The existing trade off wa just to keep the car at a constant size.

Changing the behavior to once again resize to the whole screen width (minus the bit of the card peaking in) is probably the best trade off since it is best for phone-sized UIs and that's what our demo app is optimized for. Anyone trying to build a nav solution for an iPad sized device will need to be doing a bunch more UI customization thinking anyway.

MaximAlien:
> I think it's fine to make this change, but I know the previous behavior was there deliberately from the past. We never really resolved what the guidance cards should do when the device had a wide screen (phone or iPad). It seems weird to make a super wide guidance card that is nearly empty, but also seems weird and possibly confusing if there are several guidance cards fully on screen at once since the user has to make sure to read just the left-most one. The existing trade off wa just to keep the car at a constant size.
> 
> Changing the behavior to once again resize to the whole screen width (minus the bit of the card peaking in) is probably the best trade off since it is best for phone-sized UIs and that's what our demo app is optimized for. Anyone trying to build a nav solution for an iPad sized device will need to be doing a bunch more UI customization thinking anyway.

Main reason `InstructionsCardContainerView` resizing was added is that previously after rotation instructions cards were not resized at all, which in turn led to inconsistency in updating upcoming maneuvers/swiping.

There is also a way for end user to use own size for instruction cards by implementing `instructionsCardCollection(_:cardSizeFor:)`.

PR #2615, opened 2020-09-10
Codable Progress classes

### LINKED ISSUE #1940: Route progress classes should be codable
RouteProgress, RouteLegProgress, and RouteStepProgress should conform to NSSecureCoding or Codable. MapboxDirections.swift classes such as Route are already NSSecureCoding-conformant.

/cc @mapbox/navigation-ios

Udumft:
Resolves #1940 

Added `Codable` compliance to `RouteLegProgress`, `RouteStepProgress` and `RouteProgress` classes.

Due to some `tuple` properties ([`IndexedRoute`](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/Router.swift#L18) and [`TimedCongestionLevel`](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/RouteProgress.swift#L165)) in `RouteProgress` I had to implement custom `encode/decode` methods.

### DIFF
--- a/MapboxCoreNavigation/RouteLegProgress.swift
+++ b/MapboxCoreNavigation/RouteLegProgress.swift
@@ -1,0 +1,255 @@
+
+import Foundation
+import MapboxDirections
+
+/**
+ `RouteLegProgress` stores the user’s progress along a route leg.
+ */
+open class RouteLegProgress: Codable {
+    /**
+     Returns the current `RouteLeg`.
+     */
+    public let leg: RouteLeg
+
+    /**
+     Index representing the current step.
+     */
+    public var stepIndex: Int {
+        didSet {
+            assert(stepIndex >= 0 && stepIndex < leg.steps.endIndex)
+            currentStepProgress = RouteStepProgress(step: currentStep)
+        }
+    }
+
+    /**
+     The remaining steps for user to complete.
+     */
+    public var remainingSteps: [RouteStep] {
+        return Array(leg.steps.suffix(from: stepIndex + 1))
+    }
+
+    /**
+     Total distance traveled in meters along current leg.
+     */
+    public var distanceTraveled: CLLocationDistance {
+        return leg.steps.prefix(upTo: stepIndex).map { $0.distance }.reduce(0, +) + currentStepProgress.distanceTraveled
+    }
+
+    /**
+     Duration remaining in seconds on current leg.
+     */
+    public var durationRemaining: TimeInterval {
+        return remainingSteps.map { $0.expectedTravelTime }.reduce(0, +) + currentStepProgress.durationRemaining
+    }
+
+    /**
+     Distance remaining on the current leg.
+     */
+    public var distanceRemaining: CLLocationDistance {
+        return remainingSteps.map { $0.distance }.reduce(0, +) + currentStepProgress.distanceRemaining
+    }
+
+    /**
+     Number between 0 and 1 representing how far along the current leg the user has traveled.
+     */
+    public var fractionTraveled: Double {
+        return distanceTraveled / leg.distance
+    }
+
+    public var userHasArrivedAtWaypoint = false
+
+    /**
+     Returns the `RouteStep` before a given step. Returns `nil` if there is no step prior.
+     */
+    public func stepBefore(_ step: RouteStep) -> RouteStep? {
+        guard let index = leg.steps.firstIndex(of: step) else {
+            return nil
+        }
+        if index > 0 {
+            return leg.steps[index-1]
+        }
+        return nil
+    }
+
+    /**
+     Returns the `RouteStep` after a given step. Returns `nil` if there is not a step after.
+     */
+    public func stepAfter(_ step: RouteStep) -> RouteStep? {
+        guard let index = leg.steps.firstIndex(of: step) else {
+            return nil
+        }
+        if index+1 < leg.steps.endIndex {
+            return leg.steps[index+1]
+        }
+        return nil
+    }
+
+    /**
+     Returns the `RouteStep` before the current step.
+
+     If there is no `priorStep`, nil is returned.
+     */
+    public var priorStep: RouteStep? {
+        guard stepIndex - 1 >= 0 else {
+            return nil
+        }
+        return leg.steps[stepIndex - 1]
+    }
+
+    /**
+     Returns the current `RouteStep` for the leg the user is on.
+     */
+    public var currentStep: RouteStep {
+        return leg.steps[stepIndex]
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -296,401 +296,68 @@ open class RouteProgress {
...
-    
-    public var keyPathsAffectingValueForRemainingVisualInstructions: Set<String> {
-        return ["step.instructionsDisplayedAlongStep", "visualInstructionIndex"]
+        let route = try container.decode(Route.self, forKey: .indexedRoute)
+        let routeIndex = try container.decode(Int.self, forKey: .indexedRouteIndex)
+        self.indexedRoute = (route, routeIndex)
+        self.routeOptions = try container.decode(RouteOptions.self, forKey: .routeOptions)
+        self.legIndex = try container.decode(Int.self, forKey: .legIndex)
+        self.currentLegProgress = try container.decode(RouteLegProgress.self, forKey: .currentLegProgress)
+        self.congestionTravelTimesSegmentsByStep = decodeCongestion(try container.decode([[[TimedCongestionLevelCodable]]].self,
```
Udumft, line 347: Even though `congestionTravelTimesSegmentsByStep` and `congestionTimesPerStep` can be recalculated using `calculateLegsCongestion()`, I still decided to explicitly encode/decode them because these are `public var` properties and can be potentially modified by User.

MapboxCoreNavigationTests/RouteProgressTests.swift
```diff
@@ -272,4 +272,76 @@ class RouteProgressTests: XCTestCase {
...
+        let data = try! encoder.encode(routeProgress)
+        let decoder = JSONDecoder()
+        decoder.userInfo[.options] = routeOptions
+        let decoded = try! decoder.decode(RouteProgress.self, from: data)
+
+        XCTAssertEqual(routeProgress.indexedRoute.0.distance, decoded.indexedRoute.0.distance)
+        XCTAssertEqual(routeProgress.indexedRoute.0.speechLocale, decoded.indexedRoute.0.speechLocale)
+        XCTAssertEqual(routeProgress.indexedRoute.0.shape, decoded.indexedRoute.0.shape)
+        XCTAssertEqual(routeProgress.indexedRoute.0.expectedTravelTime, decoded.indexedRoute.0.expectedTravelTime)
+        XCTAssertEqual(routeProgress.indexedRoute.0.routeIdentifier, decoded.indexedRoute.0.routeIdentifier)
```
Udumft, line 290: I could not directly compare original and decoded `Route` due to differences in floating point [de]serialization precision.

Also, I noticed weird behavior: consider a `Waypoint` with a non-nil `name` as a `source` or `destination` of some `RouteLeg`. Decoding `[RouteLeg]` alone works perfectly, while decoding such array as a part of `Route` object - results in `Waypoint.name` to become nil.  I could not find how this happens.

MapboxCoreNavigation/RouteProgress.swift
1ec5: I think it was an oversight that the setters are public. Nothing in MapboxNavigation appears to read or write to these properties. Should we privatize these properties or at least their setters in v1.0.0 to avoid the headache and size impact of encoding these ordinarily computed properties?

MapboxCoreNavigationTests/RouteProgressTests.swift
1ec5: > I could not directly compare original and decoded Route due to differences in floating point [de]serialization precision.

In case it helps, MapboxDirectionsTests uses [`JSONSerialization.objectsAreEqual(_:_:approximate:)`](https://github.com/mapbox/mapbox-directions-swift/blob/b7c26f516cdc46637e5a2eb5b775694516f8b7b3/Tests/MapboxDirectionsTests/JSONSerialization.swift#L31) to deep-compare encoded structures while skirting around floating-point imprecision. 

> Also, I noticed weird behavior: consider a `Waypoint` with a non-nil `name` as a `source` or `destination` of some `RouteLeg`. Decoding `[RouteLeg]` alone works perfectly, while decoding such array as a part of `Route` object - results in `Waypoint.name` to become nil.  I could not find how this happens.

It might be related to passing in a RouteOptions without those names in its waypoints. MapboxDirections tries to copy the names from the request to the response, because the Directions API would otherwise put just the snapped street name into this particular field in the response.

MapboxCoreNavigation/RouteLegProgress.swift
```diff
@@ -0,0 +1,255 @@
+
+import Foundation
+import MapboxDirections
+
+/**
+ `RouteLegProgress` stores the user’s progress along a route leg.
+ */
+open class RouteLegProgress: Codable {
```
1ec5, line 8: Thanks for splitting these classes out into their own files. Were there any changes to them other than adding the Codable-related methods?

1ec5 (commented):
Suggested changelog entry:

> * `RouteProgress`, `RouteLegProgress`, and `RouteStepProgress` now conform to the `Codable` protocol. ([#2615](https://github.com/mapbox/mapbox-navigation-ios/pull/2615))

MapboxCoreNavigation/RouteLegProgress.swift
Udumft: No other changes made.

MapboxCoreNavigation/RouteProgress.swift
Udumft: I can barely imagine a case when user needs to manually adjust those properties, so IMHO we could at least make it `readonly`. However, such change better be deployed in `1.0` but this current PR is targeted for `1.1`...

MapboxCoreNavigationTests/RouteProgressTests.swift
Udumft: Yes, I saw `JSONSerialization.objectsAreEqual(_:_:approximate:)` approach but I wanted to compare decoded objects too

MapboxCoreNavigation/RouteProgress.swift
1ec5: As long as the read-only change is in a separate commit or separate PR, we can cherry-pick it separately to the release-v1.0 branch.

Udumft:
[15b9eed](https://github.com/mapbox/mapbox-navigation-ios/pull/2615/commits/15b9eed382d189a0ff5f32f28df0dc53f782e235) commit to be cherry-picked to `1.0.0` release due to minor public API modification.

1ec5 (approved):

Issue #2617, opened 2020-09-11
Add positioning issue category to feedback view controller


1ec5:
#2419 overhauled the set of categories in the feedback view controller, but a Positioning Issue category was later added to the Android navigation SDK in mapbox/mapbox-navigation-android#3167. For parity with Android, we should add that category.

We’ll need another case in the `FeedbackType` enumeration, a corresponding set of subtypes, and an asset in https://github.com/mapbox/navigation-ui-resources/tree/master/img/feedback/ .

/cc @mapbox/navigation-ios @avi-c @JunDai

1ec5:
mapbox/navigation-ui-resources#17 adds the asset, and #2621 adds the button to the feedback view controller.

PR #2621, opened 2020-09-14
Add new positioning category for nav feedback

avi-c:
Fix for #2617


MapboxNavigation/FeedbackItem.swift
```diff
@@ -81,6 +81,10 @@ public extension FeedbackType {
             return NSLocalizedString("ROUTE_QUALITY_ROAD_MISSING_FROM_MAP_FEEDBACK", bundle: .mapboxNavigation, value: "Road missing from map", comment: "Specifc route feedback that route contained road not shown on map.")
         case .roadClosure(.other):
             return NSLocalizedString("ROAD_CLOSURE_OTHER_FEEDBACK", bundle: .mapboxNavigation, value: "Other", comment: "Specific route feedback that a road closure type was encountered but not listed as a choice.")
+        case .positioning(.none):
+            return NSLocalizedString("POSITIONING", bundle: .mapboxNavigation, value: "Positioning", comment: "General category of route feedback where user position is incorrect.")
+        case .positioning(.userPosition):
+            return NSLocalizedString("POSITIONING", bundle: .mapboxNavigation, value: "User Position", comment: "Specific route feedback that user positioning is incorrect.")
```
1ec5, line 87: These two strings should have different keys.

Remember to run scripts/extract_localizable.sh to update the Localizable.strings file. Unfortunately, there won’t be much time to get translations of this string, but we’ll pull in any missing translations in v1.1.

MapboxNavigation/Resources/Base.lproj/Localizable.strings
```diff
@@ -136,6 +136,12 @@
 /* Accessibility value of label indicating the absence of a rating */
 "NO_RATING" = "No rating set.";
 
+/* General category of route feedback where user position is incorrect. */
+"POSITIONING" = "Positioning";
+
+/* Specific route feedback that user positioning is incorrect. */
+"POSITIONING_USER" = "User Position";
```
1ec5, line 143: This string will be corrected in #2625.

1ec5 (approved):

avi-c:
@1ec5 - I merged everything into this single PR. Hopefully that's ok with you and once you review it we can merge this one.

MapboxCoreNavigation/EventDetails.swift
```diff
@@ -352,7 +352,9 @@ enum EventDetailsError: Error {
 
 extension EventDetails {
     func asDictionary() throws -> [String: Any] {
-        let data = try JSONEncoder().encode(self)
+        let encoder = JSONEncoder()
+        encoder.keyEncodingStrategy = .convertToSnakeCase
+        let data = try encoder.encode(self)
```
1ec5, line 357: Remember to remove this bit, since we don’t actually want the keys to be snake case after all.

1ec5:
Before merging, let’s find some way to verify that the JSON-encoded feedback event looks reasonably similar to the example Android feedback event that was shared.

avi-c:
What I can say is that the events dictionary that I am now passing to the telem library does look similar to the Android JSON. It has the entry for the feedbackType and an array of subtypes. I haven't been able to see what is sent by the telem library or find an event in the server logs yet.

avi-c:
The telem team has confirmed that one of the new events is indeed showing up on their stream, so I think we are ok to merge this.

MapboxCoreNavigation/EventDetails.swift
1ec5: #2628

1ec5:
Cherry-picked into the release-v1.0 branch as 18b89363fd8d36210372354383b034f1e65b6ddd.

Issue #2618, opened 2020-09-11
Subtype not sent with feedback events


1ec5:
The feedback subtypes are all specified in camelCase per Swift API design guidelines. However, the Android navigation SDK sends them in snake_case. The iOS navigation SDK doesn’t seem to be using [`JSONEncoder.KeyEncodingStrategy.convertToSnakeCase`](https://developer.apple.com/documentation/foundation/jsonencoder/keyencodingstrategy/converttosnakecase) anywhere to convert between the two cases.

https://github.com/mapbox/mapbox-navigation-ios/blob/b0fbb24a33bc7ca06bbeb109bb6d03353ca902e7/MapboxCoreNavigation/Feedback.swift#L47-L55

The Android navigation SDK is also specifying some raw key names with typos in them, so if the schema or backend has replicated those typos, then need to override the key names on iOS:

https://github.com/mapbox/mapbox-navigation-android/blob/af8a3734d7a3922a01dc0b417200b53dcea72a38/libnavigation-core/src/main/java/com/mapbox/navigation/core/telemetry/events/FeedbackEvent.kt#L125
https://github.com/mapbox/mapbox-navigation-android/blob/af8a3734d7a3922a01dc0b417200b53dcea72a38/libnavigation-core/src/main/java/com/mapbox/navigation/core/telemetry/events/FeedbackEvent.kt#L150

/cc @mapbox/navigation-ios @avi-c

1ec5:
When the user selects the subtype and sends the feedback, we try to update the feedback event, but I don’t see where the subtype ever makes it into the JSON-formatted event:

https://github.com/mapbox/mapbox-navigation-ios/blob/b0fbb24a33bc7ca06bbeb109bb6d03353ca902e7/MapboxCoreNavigation/NavigationEventsManager.swift#L157 https://github.com/mapbox/mapbox-navigation-ios/blob/b0fbb24a33bc7ca06bbeb109bb6d03353ca902e7/MapboxCoreNavigation/CoreFeedbackEvent.swift#L28

1ec5:
Fixed in #2622.

PR #2622, opened 2020-09-14
Send Feedback event subtypes

avi-c:
Fix for #2618 

Add JSON keys for Feedback event subtypes. Update the feedback events with the subtypes before sending them back to the telemetry endpoint.


MapboxCoreNavigation/EventDetails.swift
```diff
@@ -352,7 +352,9 @@ enum EventDetailsError: Error {
 
 extension EventDetails {
     func asDictionary() throws -> [String: Any] {
-        let data = try JSONEncoder().encode(self)
+        let encoder = JSONEncoder()
+        encoder.keyEncodingStrategy = .convertToSnakeCase
```
1ec5, line 356: I think I was mistaken in https://github.com/mapbox/mapbox-navigation-ios/issues/2618#issue-699592630: so far, we’ve only identified that the values of this one key are in the wrong case, but the keys aren’t necessarily in the wrong case. After all, the general feedback event code hasn’t changed for quite a long time. Looking at the backend’s event schema, I think camel case is actually correct for the keys.

MapboxCoreNavigation/Feedback.swift
```diff
@@ -26,22 +26,85 @@ public enum FeedbackType: CustomStringConvertible {
 
     /// Indicates a problem with positioning the user
     case positioning(subtype: PositioningSubtype?)
-    
+
+    /*
+      case streetPermanentlyBlockedOff
+      case roadMissingFromMap
+      case other
```
1ec5, line 33: What’s this comment for?

MapboxCoreNavigation/Feedback.swift
```diff
@@ -26,22 +26,85 @@ public enum FeedbackType: CustomStringConvertible {
...
     public var description: String {
         switch self {
         case .general:
             return "general"
-        case .incorrectVisual(_):
+        case .incorrectVisual(subtype: .none):
             return "incorrect_visual"
-        case .confusingAudio(_):
+        case .incorrectVisual(subtype: .turnIconIncorrect):
+            return "turn_icon_incorrect"
```
1ec5, line 42: This replaces the type with the subtype in the event. Let’s verify that that’s what the server expects and what the Android navigation SDK is sending. At a glance, I’m unsure how this code would be able to send multiple subtypes for the same event if the user selects multiple rows in the feedback detail view controller – would it be multiple events?

MapboxCoreNavigation/Feedback.swift
avi-c: I'll remove it. I was dropping in the enum as a reference as I built the case statement

MapboxCoreNavigation/EventDetails.swift
avi-c: Let's remove this then until it's clear it's needed.

MapboxCoreNavigation/Feedback.swift
avi-c: As I understand it, the way it sends multiple events is that for each `item` selected in the UI, it creates a new FeedbackEvent (by getting a new UUID for one), then updates that latest event with the actual event details.

1ec5: Based on a cursory reading of the events schema, the SDK needs to send a single feedback event that contains both the type and any subtypes (in the `feedbackSubType` property).

1ec5:
Cherry-picked into the release-v1.0 branch as part of 18b89363fd8d36210372354383b034f1e65b6ddd.

PR #2623, opened 2020-09-14
Custom Route shapes fix

### LINKED ISSUE #2595: Unable to provide custom route line shapes
Methods like `NavigationMapViewDelegate.navigationMapView(_:shapeFor:)` and `NavigationMapViewDelegate.navigationMapView(_:simplifiedShapeFor:)` were originally added to allow the developer to essentially provide a custom route line colored by something other than traffic congestion. However, as of #2377, these methods are no longer called by NavigationMapView, so the route line can only be colored by traffic congestion. When these methods are implemented, we should use them instead of creating a route line colored by a gradient.

/cc @mapbox/navigation-ios

Udumft:
Fixes #2595 

Restored calling `NavigationMapViewDelegate.navigationMapView(_:simplifiedShapeFor:)` method for shape customization and verified that custom shapes mechanism works.

There is a discussion going on around these methods on how to improve the API and make customization easy and straightforward. Eventually we should remove these methods, but decided to leave them for now because there is no alternative ideas developed atm and some clients already use that API.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -473,7 +473,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         let allRoutesShape = navigationMapViewDelegate?.navigationMapView(self, shapeFor: routes) ?? shape(for: routes, legIndex: legIndex)
         let allRoutesSource = addAllRoutesSource(style, shape: allRoutesShape)
         
-        let mainRouteCasingShape = navigationMapViewDelegate?.navigationMapView(self, shapeFor: [mainRoute]) ?? shape(for: [mainRoute], legIndex: legIndex)
+        let mainRouteCasingShape = navigationMapViewDelegate?.navigationMapView(self, simplifiedShapeFor: mainRoute) ?? shape(forCasingOf: mainRoute, legIndex: legIndex)
+        
         let mainRouteCasingSource = addMainRouteCasingSource(style, shape: mainRouteCasingShape)
         
         let mainRouteLayer = addMainRouteLayer(style, source: allRoutesSource, lineGradient: routeLineGradient(mainRoute, fractionTraveled: 0.0))
--- a/MapboxNavigation/NavigationMapViewDelegate.swift
+++ b/MapboxNavigation/NavigationMapViewDelegate.swift
@@ -83,6 +83,8 @@ public protocol NavigationMapViewDelegate: class, UnimplementedLogging {
     
     /**
      Asks the receiver to return an MGLShape that describes the geometry of the route.
+        
+     Resulting `MGLShape` will then be styled using `NavigationMapView.navigationMapView(_: mainRouteStyleLayerWithIdentifier: source:)` provided style or a default congestion style if above delegate method was not implemented. In latter case, consider modifing your custom `MGLShape` `attributes` to have 'isAlternateRoute' key set to 'false'. Otherwise style predicate condition will filter out the shape.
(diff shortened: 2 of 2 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -473,7 +473,8 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         let allRoutesShape = navigationMapViewDelegate?.navigationMapView(self, shapeFor: routes) ?? shape(for: routes, legIndex: legIndex)
         let allRoutesSource = addAllRoutesSource(style, shape: allRoutesShape)
         
-        let mainRouteCasingShape = navigationMapViewDelegate?.navigationMapView(self, shapeFor: [mainRoute]) ?? shape(for: [mainRoute], legIndex: legIndex)
+        let mainRouteCasingShape = navigationMapViewDelegate?.navigationMapView(self, simplifiedShapeFor: mainRoute) ?? shape(forCasingOf: mainRoute, legIndex: legIndex)
```
1ec5, line 476: Should `allRoutesShape` also call `navigationMapView(_:simplifiedShapeFor:)`? What should call `navigationMapView(_:shapeFor:)`? Previously, a developer would’ve implemented `navigationMapView(_:shapeFor:)` to provide the shape data for the whole route (normally to be used as the casing) and `navigationMapView(_:simplifiedShapeFor:)` to provide the shape data for the route broken up by arbitrary segments (normally to be colored by traffic congestion).

Udumft: As I found in the code, `navigationMapView(_:shapeFor:)` is meant to provide route shapes, and `navigationMapView(_:simplifiedShapeFor:)` meant to provide the casing shape. At least when I provided my custom Shapes like that they were colored as route and casing respectfully. Even if my custom shape for `navigationMapView(_:shapeFor:)` had just one `Polyline` it still was colored with congestion segments. 
Doesn't it supposed to be like that?

1ec5: > As I found in the code, `navigationMapView(_:shapeFor:)` is meant to provide route shapes, and `navigationMapView(_:simplifiedShapeFor:)` meant to provide the casing shape.

Yes, sorry, I got the two methods mixed up in https://github.com/mapbox/mapbox-navigation-ios/pull/2623#discussion_r488007899. If you provide only one shape in `navigationMapView(_:shapeFor:)`, I don’t think there should be colored congestion segments.

IodaMikeMapbox (approved):

MapboxNavigation/NavigationMapView.swift
Udumft: So... Is this convo resolved? :)

1ec5 (approved):
Remember to add a changelog entry about restoring the ability to customize the route shape.

MaximAlien:
@1ec5, @Udumft After merging this PR to `main` I can see that route casing is no longer present:

Screen Shot 2020-10-28 at 3 28 10 PM

Before route line casing was visible:

Screen Shot 2020-10-28 at 3 28 37 PM

When I check `shape(forCasingOf:legIndex:)` implementation I see that `MGLPolylineFeature` doesn't use `isAlternateRoute`. Since `shape(forCasingOf:legIndex:)` is called only main route line in default case shouldn't we set `polyline.attributes["isAlternateRoute"] = false`?

Udumft:
Shame on me for missing this one :( but thank you for pointing that out! Fix submitted in [mapbox/mapbox-navigation-ios/pull/2715](https://github.com/mapbox/mapbox-navigation-ios/pull/2715)

Rmkasun-8 (commented):
Navigation on mapview

Rmkasun-8 (commented):
Change

Rmkasun-8 (commented):
Authored

PR #2625, opened 2020-09-14
Match Android for Feedback event JSON keys and the sending of a single event per category

avi-c:
Match Android feedback mechanism of sending a single event per category of feedback with an optional array of subtypes. Update JSON keys to match Android as well.

### DIFF
--- a/MapboxCoreNavigation/CoreFeedbackEvent.swift
+++ b/MapboxCoreNavigation/CoreFeedbackEvent.swift
@@ -26,6 +26,19 @@ class CoreFeedbackEvent: Hashable {
 class FeedbackEvent: CoreFeedbackEvent {
     func update(type: FeedbackType, source: FeedbackSource, description: String?) {
         eventDictionary["feedbackType"] = type.description
+
+        // if there is a subtype for this event then append the subtype description to our list for this type of feedback
+        if let subtypeDescription = type.subtypeDescription {
+            var subtypeList = [String]()
+            if let existingSubtypeList = eventDictionary["feedbackSubType"] as? [String] {
+                subtypeList.append(contentsOf: existingSubtypeList)
+            }
+
+            if !subtypeList.contains(subtypeDescription) {
+                subtypeList.append(subtypeDescription)
+            }
+            eventDictionary["feedbackSubType"] = subtypeList
+        }
         eventDictionary["source"] = source.description
         eventDictionary["description"] = description
     }
--- a/MapboxCoreNavigation/Feedback.swift
+++ b/MapboxCoreNavigation/Feedback.swift
@@ -27,17 +27,29 @@ public enum FeedbackType: CustomStringConvertible {
     /// Indicates a problem with positioning the user
     case positioning(subtype: PositioningSubtype?)
 
-    /*
-      case streetPermanentlyBlockedOff
-      case roadMissingFromMap
-      case other
-     */
+    /// Description of the category for this type of feedback
     public var description: String {
         switch self {
         case .general:
             return "general"
-        case .incorrectVisual(subtype: .none):
-            return "incorrect_visual"
+        case .incorrectVisual(_):
+            return "incorrect_visual_guidance"
+        case .confusingAudio(_):
+            return "incorrect_audio_guidance"
+        case .routeQuality(_):
+            return "routing_error"
+        case .illegalRoute(_):
+            return "not_allowed"
+        case .roadClosure(_):
+            return "road_closed"
+        case .positioning(_):
+            return "positioning_issue"
+        }
+    }
+
+    /// Optional detailed description of the subtype of this feedback
+    public var subtypeDescription: String? {
+        switch self {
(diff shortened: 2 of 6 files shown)


MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -26,6 +26,19 @@ class CoreFeedbackEvent: Hashable {
 class FeedbackEvent: CoreFeedbackEvent {
     func update(type: FeedbackType, source: FeedbackSource, description: String?) {
         eventDictionary["feedbackType"] = type.description
+
+        // if there is a subtype for this event then append the subtype description to our list for this type of feedback
+        if let subtypeDescription = type.subtypeDescription {
+            var subtypeList = [String]()
+            if let existingSubtypeList = eventDictionary["feedbackSubType"] as? [String] {
+                subtypeList.append(contentsOf: existingSubtypeList)
+            }
```
1ec5, line 35: Is it necessary to merge the existing and new subtypes? FeedbackSubtypeViewController only calls into this code path once all the subtypes have been selected, but technically it would be possible for an application to call `NavigationEventsManager.updateFeedback(uuid:type:source:description:)` multiple times for a single event, which would result in some duplication.

MapboxNavigation/FeedbackSubtypeViewController.swift
```diff
@@ -160,8 +160,8 @@ class FeedbackSubtypeViewController: FeedbackViewController {
 
     private func sendReport() {
         if selectedItems.count > 0 {
-            selectedItems.forEach { item in
-                if let uuid = self.uuid {
+            if let uuid = self.uuid {
+                selectedItems.forEach { item in
```
1ec5, line 164: Nit: A little more straightforward this way:

```suggestion
                for item in selectedItems {
```

MapboxNavigation/Resources/Base.lproj/Localizable.strings
```diff
@@ -136,10 +136,12 @@
 /* Accessibility value of label indicating the absence of a rating */
 "NO_RATING" = "No rating set.";
 
-/* General category of route feedback where user position is incorrect.
-   Specific route feedback that user positioning is incorrect. */
+/* General category of route feedback where user position is incorrect. */
 "POSITIONING" = "Positioning";
 
+/* Specific route feedback that user positioning is incorrect. */
+"POSITIONING_USER" = "User Position";
```
1ec5, line 143: It’s a bit weird to address the user as “User”. “Your position” would be more intuitive (and also consistent in terms of capitalization).

MapboxNavigation/FeedbackSubtypeViewController.swift
avi-c: It's written this way because it's the call to `if let uuid = self.uuid` that causes a new event to be created and enqueued for transmitting. Then these event update calls are what append the individual subtypes to the one feedback event that's waiting to be sent.

There shouldn't be any duplication since a subcategory is added only if it isn't already in the list.

MapboxCoreNavigation/CoreFeedbackEvent.swift
avi-c: see below...

MapboxNavigation/Resources/Base.lproj/Localizable.strings
avi-c: Good point.

MapboxNavigation/FeedbackSubtypeViewController.swift
1ec5: Oh, the feedback was just about the use of `forEach` instead of a standard for loop. Not sure why GitHub highlighted the `if let` also.

MapboxCoreNavigation/Feedback.swift
```diff
@@ -92,20 +92,14 @@ public enum FeedbackType: CustomStringConvertible {
...
         case .roadClosure(subtype: .roadMissingFromMap):
             return "road_is_missing_from_map"
-        case .roadClosure(subtype: .other):
-            return "road_closure"
-        case .positioning(subtype: .none):
-            return "positioning_issue"
         case .positioning(subtype: .userPosition):
             return "positioning_issue"
+        default:
+            return nil
```
1ec5, line 102: Ideally we could avoid `default` cases so that someone adding another category in the future can rely on the compiler’s switch exhaustiveness check to find this switch statement. But if explicitly mapping all the `other` and `none` cases to `nil` would be inconvenient, I guess we can leave it this way.

MapboxNavigation/Resources/Base.lproj/Localizable.strings
1ec5: Remember to rerun extract-localizable.sh to pick up the change to the string’s base value in the source code.

1ec5 (approved):

1ec5:
Cherry-picked into the release-v1.0 branch as part of 18b89363fd8d36210372354383b034f1e65b6ddd.

PR #2627, opened 2020-09-14
Document guidance cards, junction views

### LINKED ISSUE #2519: Need documentation for Guidance Card styling API
The revised Guidance Card styling API needs documenting. It now uses UIAppearance proxies to support custom styling.

See changes in https://github.com/mapbox/mapbox-navigation-ios/pull/2401 and related tickets.

1ec5:
Added class documentation comments for the minimum set of types necessary to consider the guidance cards feature from #2149 and #2296 to be publicly available. With these changes, we can finally add a blurb to the changelog announcing the introduction of this feature, even though it has been in the codebase for quite some time. This PR also adds some documentation about junction views for styling purposes.

Guidance cards, represented by the `InstructionsCardViewController` class, is an alternative presentation of maneuver instructions that is intended to be better visually integrated with the map than the traditional presentation of banners flush with the edges of the screen, represented by the default `TopBannerViewController` class. To enable guidance cards, an application would pass an instance of `InstructionsCardViewController` into `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:)` when creating the NavigationViewController. Using guidance cards in a custom turn-by-turn navigation UI requires some additional setup and delegate method implementations.

Fixes most of #2519.

/cc @mapbox/navigation-ios @avi-c @d-prukop

### DIFF
--- a/MapboxNavigation/TopBannerViewController.swift
+++ b/MapboxNavigation/TopBannerViewController.swift
@@ -42,6 +42,11 @@ public extension TopBannerViewControllerDelegate {
     }
 }
 
+/**
+ A view controller that displays the current maneuver instruction as a “banner” flush with the edges of the containing view. The user swipes to one side to preview a subsequent maneuver.
+ 
+ This class is the default top banner view controller used by `NavigationOptions` and `NavigationViewController`. `InstructionsCardViewController` provides an alternative, user notification–like interface.
+ */
 open class TopBannerViewController: UIViewController {
     weak var delegate: TopBannerViewControllerDelegate? = nil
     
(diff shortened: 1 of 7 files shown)


MapboxNavigation/TopBannerViewController.swift
```diff
@@ -42,6 +42,11 @@ public extension TopBannerViewControllerDelegate {
     }
 }
 
+/**
```
MaximAlien, line 45: Not critical: indentation is a bit shifted here.

MaximAlien (approved):

1ec5:
Cherry-picked into the release-v1.0 branch as db74f46011c4904103f2986e974e42cdc92577cd.

Issue #2632, opened 2020-09-16
PassiveLocationDataSource randomly crashes downloading routing tiles — navigator::cache::TileLoaderImpl::DownloadTile, navigator::HttpTileGetter::get, etc.


1ec5:
We’ve received some reports of random crashes when starting a PassiveLocationManager or PassiveLocationDataSource in v1.0.0-rc.1 through v1.0.0-rc.4. Typical stack traces look something like this:

```
0  MapboxNavigationNative         0x102bb4c34 navigator::cache::TileLoaderImpl::DownloadTile(valhalla::baldr::GraphId, std::__1::shared_ptr<navigator::base::CancellationToken>, navigator::cache::TileLoader::DownloadPolicy) + 784
1  MapboxNavigationNative         0x102bae51c std::__1::__packaged_task_func<std::__1::__bind<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > (navigator::cache::TileLoader::*)(valhalla::baldr::GraphId, std::__1::shared_ptr<navigator::base::CancellationToken>, navigator::cache::TileLoader::DownloadPolicy), navigator::cache::TileLoader*, valhalla::baldr::GraphId const&, std::__1::shared_ptr<navigator::base::CancellationToken>&, navigator::cache::TileLoader::DownloadPolicy&>, std::__1::allocator<std::__1::__bind<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > (navigator::cache::TileLoader::*)(valhalla::baldr::GraphId, std::__1::shared_ptr<navigator::base::CancellationToken>, navigator::cache::TileLoader::DownloadPolicy), navigator::cache::TileLoader*, valhalla::baldr::GraphId const&, std::__1::shared_ptr<navigator::base::CancellationToken>&, navigator::cache::TileLoader::DownloadPolicy&> >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > ()>::operator()() + 12256
2  MapboxNavigationNative         0x102baeaa8 std::__1::packaged_task<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > ()>::operator()() + 13676
3  MapboxNavigationNative         0x102b30a94 navigator::base::ThreadPool::ProcessTasks() + 1036 (__hidden#1595_:1036)
[...]

PR #2633, opened 2020-09-16
Avoid reconfiguring passive navigator

1ec5:
Avoid refetching offline routing tile versions and reconfiguring the passive location data source’s navigator’s router multiple times, because a router instance only expects to be configured once.

When we later upgrade to MapboxNavigationNative v20._x_, we’ll need to move this configuration logic into the code that initializes the navigator (currently in `PassiveLocationDataSource(directions:systemLocationManager:)`). That will be interesting to consider as part of unifying the navigators in #2593.

This is a speculative fix for #2632.

/cc @mapbox/navigation-ios @mapbox/navnative

### DIFF
--- a/MapboxCoreNavigation/PassiveLocationDataSource.swift
+++ b/MapboxCoreNavigation/PassiveLocationDataSource.swift
@@ -47,6 +47,13 @@ open class PassiveLocationDataSource: NSObject {
     let navigator: Navigator
     
     /**
+     Whether the navigator’s router has been configured.
+     
+     Set this property to `true` before calling `Navigator.configureRouter(for:)` and reset it to `false` if something causes the router to be unconfigured.
+     */
+    var isConfigured = false
+    
+    /**
      The location data source’s delegate.
      */
     public weak var delegate: PassiveLocationDataSourceDelegate?
@@ -56,6 +63,10 @@ open class PassiveLocationDataSource: NSObject {
      */
     public func startUpdatingLocation(completionHandler: ((Error?) -> Void)? = nil) {
         systemLocationManager.startUpdatingLocation()
+        
+        guard !isConfigured else {
+            return
+        }
         
         directions.fetchAvailableOfflineVersions { [weak self] (versions, error) in
             guard let self = self, let latestVersion = versions?.first(where: { !$0.isEmpty }), error == nil else {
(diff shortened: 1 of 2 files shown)


MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -46,6 +46,13 @@ open class PassiveLocationDataSource: NSObject {
      */
     let navigator: Navigator
     
+    /**
+     Whether the navigator’s router has been configured.
+     
+     Set this property to `true` before calling `Navigator.configureRouter(for:)` and reset it to `false` if something causes the router to be unconfigured.
```
1ec5, line 52: Apparently there’s no way to ask the navigator whether the router has been configured. Fortunately, there’s also no situation where something else might call `configureRouter(for:)` on our behalf, making this property go stale, and we don’t currently call `removeTilesForPackedTilesPath(_:lowerLeft:upperRight:)` or `clearCache()`.

MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
```diff
@@ -70,7 +70,9 @@ class PassiveLocationDataSourceTests: XCTestCase {
         wait(for: [startingExpectation], timeout: 2)
         
         // FIXME: Starting the location manager should automatically configure the navigator.
+        XCTAssertFalse(locationManager.isConfigured)
         XCTAssertNoThrow(try locationManager.configureNavigator(withTilesVersion: "1234"))
+        XCTAssertTrue(locationManager.isConfigured)
```
1ec5, line 75: This test is currently disabled – #2579 reenables it after reworking it a bit – but I verified locally that the new assertions pass.

MaximAlien (approved):

1ec5:
Cherry-picked into the release-v1.0 branch as 2c2966d65f56b37d099397f55829df389dcdaf53.

1ec5:
> When we later upgrade to MapboxNavigationNative v20._x_, we’ll need to move this configuration logic into the code that initializes the navigator (currently in `PassiveLocationDataSource(directions:systemLocationManager:)`). That will be interesting to consider as part of unifying the navigators in #2593.

@kmadsen pointed out that, in principle, it may not even be safe to defer the `Navigator.configureRouter(for:)` call to when the location manager is started. An early iteration of #2410 did configure the navigator’s router when the data source was initialized, but I deferred it to when starting the data source to avoid an impact on launch time. In practice, I think the current setup happens to be safe because we do nothing with the navigator between the time we initialize it in `PassiveLocationDataSource(directions:systemLocationManager:)` and the time we call `Navigator.configureRouter(for:)` in `configureNavigator(withTilesVersion:)`. But the upcoming refactor for MapboxNavigationNative v20._x_ will ensure that there’s no such gap.

PR #2634, opened 2020-09-16
Add initial implementation of delay information views.

MaximAlien:
Closing [#573](https://github.com/mapbox/navigation-sdks/issues/573).

Current implementation depends on changes in Mapbox Directions. PR can be found [here](https://github.com/mapbox/mapbox-directions-swift/pull/462).

Look of initial implementation:

Simulator Screen Shot - iPhone SE (2nd generation) - 2020-09-16 at 16 07 36


MapboxNavigation/NavigationMapView.swift
```diff
@@ -1525,3 +1607,51 @@ extension NavigationMapView {
         highlightedBuildingsLayer.fillExtrusionOpacity = NSExpression(format: "mgl_interpolate:withCurveType:parameters:stops:($zoomLevel, 'linear', nil, %@)", opacityStops)
     }
 }
+
+// MARK: - Delay Information
+
+extension NavigationMapView {
```
MaximAlien, line 1613: Since `NavigationMapView` now contains several features we can expose them as separate extension files (e.g. for route lines, building highlighting, vanishing route line, delay information).

MapboxNavigation/DelayInformationView.swift
```diff
@@ -0,0 +1,34 @@
+import UIKit
+
+class DelayInformationView: UIView {
```
MaximAlien, line 3: I don't really like creating `DelayInformationView` which is then later converted to `UIImage` so that it's possible to use it in `MGLSymbolStyleLayer` (there are some additional issues like text overlapping etc). 

On the other hand seems that `MGLPointAnnotation` will not be performant enough when using large amount of delay annotations. It'll also require implementation of `MGMapViewDelegate` methods inside `NavigationMapView`.

@1ec5 if you have any better ideas please let me know.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -60,6 +60,29 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      */
     public weak var courseTrackingDelegate: NavigationMapViewCourseTrackingDelegate?
     
+    /**
+     Controls whether delays are shown along the route line.
+     */
+    public var showsDelay: Bool = false {
```
MaximAlien, line 66: For now boolean flag is used to show/hide delay information views. Since delay information is provided for `Route`, `Leg` and `Step` we might want to provide API which allows to show delay for each of those. As per comment [here](https://github.com/mapbox/navigation-sdks/issues/373#issuecomment-672027950) default minimum dealy threshold value should be set to `4 minutes`.

MapboxNavigation/DelayInformationView.swift
LukasPaczos: @MaximAlien have you seen https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-symbol-icon-text-fit? Might be useful here.

LukasPaczos: /cc @langsmith ^^^ in case we've not considered this for the Android impl.

MaximAlien:
UI part for delay information is not in scope for `1.2.0`. Removing milestone for now.

MaximAlien:
Closing this PR in favor of https://github.com/mapbox/mapbox-navigation-ios/pull/2734.

PR #2635, opened 2020-09-16
Improve performance of Wayname calculation

### LINKED ISSUE #2259: Current road name label can display names of adjacent footpaths
When the user drives down a road that has a named footpath or bike path running alongside it, the current road name label can display that path’s name instead of the road’s. The SDK queries the map based on a location snapped to the route line, so normally it would be fine if some other kind of pathway runs alongside the roadway. But an incorrect result could still occur if the user has deviated far enough from the route that snapping ceases, but not far enough that the user gets rerouted, or if they’ve driven past the destination.

`RouteMapViewController.labelCurrentRoad(at:for:)` lays down a line layer bound to the Streets source and ensures that it gets obscured by the original style’s own road layers:

https://github.com/mapbox/mapbox-navigation-ios/blob/4ae1d627619c385a38cbf9903cf334152a0375de/MapboxNavigation/RouteMapViewController.swift#L648-L655

The line style layer needs to have a source layer of [`road`](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#road) (to avoid snagging other named features like rivers). The layer should also have a filter that requires [`class`](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#--road---class-text) to be an appropriate value. The set of appropriate values depends on the current route step’s mode, as seen in https://github.com/mapbox/mapbox-navigation-android/issues/2227#issuecomment-543290260.

/cc @mapbox/navigation-ios

avi-c:
Add a predicate filter for the road label MGLLineStyleLayer. Prevent the repeated recalculation of sliced current step line shape.

Fixes #2259.

### DIFF
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -687,10 +687,9 @@ extension RouteMapViewController: NavigationViewDelegate {
 
             if ![DirectionsProfileIdentifier.walking, DirectionsProfileIdentifier.cycling].contains( router.routeProgress.routeOptions.profileIdentifier) {
                 // filter out to road classes valid for motor transport
-                let roadPredicates = ["motorway", "motorway_link", "trunk", "trunk_link", "primary", "primary_link", "secondary", "secondary_link", "tertiary", "tertiary_link", "street", "street_limited", "roundabout", "mini_roundabout"].compactMap { roadClass -> NSPredicate? in
-                    return NSPredicate(format: "%K == %@", "class", roadClass)
-                }
-                let compoundPredicate = NSCompoundPredicate(orPredicateWithSubpredicates: roadPredicates)
+                let roadPredicates = ["motorway", "motorway_link", "trunk", "trunk_link", "primary", "primary_link", "secondary", "secondary_link", "tertiary", "tertiary_link", "street", "street_limited", "roundabout"]
+
+                let compoundPredicate = NSPredicate(format: "class IN %@", roadPredicates)
                 streetLabelLayer.predicate = compoundPredicate
             }
             style.insertLayer(streetLabelLayer, at: 0)
--- a/MapboxNavigation/RouteMapViewController.swift
+++ b/MapboxNavigation/RouteMapViewController.swift
@@ -684,6 +684,13 @@ extension RouteMapViewController: NavigationViewDelegate {
             streetLabelLayer.lineOpacity = NSExpression(forConstantValue: 1)
             streetLabelLayer.lineWidth = NSExpression(forConstantValue: 20)
             streetLabelLayer.lineColor = NSExpression(forConstantValue: UIColor.white)
+
+            // filter out to road classes valid for motor transport
+            let roadPredicates = ["motorway", "trunk", "primary", "secondary", "tertiary", "street"].compactMap { roadClass -> NSPredicate? in
+                return NSPredicate(format: "%K == %@", "class", roadClass)
+            }
+            let compoundPredicate = NSCompoundPredicate(orPredicateWithSubpredicates: roadPredicates)
+            streetLabelLayer.predicate = compoundPredicate
             style.insertLayer(streetLabelLayer, at: 0)
         }
 
@@ -692,6 +699,7 @@ extension RouteMapViewController: NavigationViewDelegate {
         var smallestLabelDistance = Double.infinity
         var currentName: String?
         var currentShieldName: NSAttributedString?
+        let slicedLine = stepShape.sliced(from: closestCoordinate)!
 
         for feature in features {
             var allLines: [MGLPolyline] = []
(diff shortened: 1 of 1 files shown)


MapboxNavigation/RouteMapViewController.swift
```diff
@@ -684,6 +684,13 @@ extension RouteMapViewController: NavigationViewDelegate {
             streetLabelLayer.lineOpacity = NSExpression(forConstantValue: 1)
             streetLabelLayer.lineWidth = NSExpression(forConstantValue: 20)
             streetLabelLayer.lineColor = NSExpression(forConstantValue: UIColor.white)
+
+            // filter out to road classes valid for motor transport
+            let roadPredicates = ["motorway", "trunk", "primary", "secondary", "tertiary", "street"].compactMap { roadClass -> NSPredicate? in
```
1ec5, line 689: It looks like several relevant classes are omitted from this list, such as `motorway_link` (freeway ramps) and `roundabout`. The full list of supported road classes is at:

https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v7/#road
https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#--road---class-text

It’s also worth considering whether we should expand the list or omit the predicate altogether if the current profile is `walking` or `cycling`.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -706,7 +714,6 @@ extension RouteMapViewController: NavigationViewDelegate {
                 guard line.pointCount > 0 else { continue }
                 let featureCoordinates =  Array(UnsafeBufferPointer(start: line.coordinates, count: Int(line.pointCount)))
                 let featurePolyline = LineString(featureCoordinates)
-                let slicedLine = stepShape.sliced(from: closestCoordinate)!
```
1ec5, line 709: Thanks, moving the slicing out of the loop should have a measurable impact on areas where there are lots of overlapping lines.

MapboxNavigation/RouteMapViewController.swift
avi-c: I included the list of motor vehicle road classes that are used to style with in the `road-label` layer of our navigation map style but can add these other ones as well. I'll put in a check for the routing profile as well.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -684,6 +684,15 @@ extension RouteMapViewController: NavigationViewDelegate {
             streetLabelLayer.lineOpacity = NSExpression(forConstantValue: 1)
             streetLabelLayer.lineWidth = NSExpression(forConstantValue: 20)
             streetLabelLayer.lineColor = NSExpression(forConstantValue: UIColor.white)
+
+            if ![DirectionsProfileIdentifier.walking, DirectionsProfileIdentifier.cycling].contains( router.routeProgress.routeOptions.profileIdentifier) {
```
1ec5, line 688: Btw, I think it’s possible to shorten this line based on the type of `profileIdentifier`:

```suggestion
            if ![.walking, .cycling].contains(router.routeProgress.routeOptions.profileIdentifier) {
```

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -684,6 +684,15 @@ extension RouteMapViewController: NavigationViewDelegate {
...
+                // filter out to road classes valid for motor transport
+                let roadPredicates = ["motorway", "motorway_link", "trunk", "trunk_link", "primary", "primary_link", "secondary", "secondary_link", "tertiary", "tertiary_link", "street", "street_limited", "roundabout", "mini_roundabout"].compactMap { roadClass -> NSPredicate? in
+                    return NSPredicate(format: "%K == %@", "class", roadClass)
+                }
+                let compoundPredicate = NSCompoundPredicate(orPredicateWithSubpredicates: roadPredicates)
```
1ec5, line 693: Something along these lines should work and would probably perform slightly better:

```suggestion
                let roadPredicates = ["motorway", "motorway_link", "trunk", "trunk_link", "primary", "primary_link", "secondary", "secondary_link", "tertiary", "tertiary_link", "street", "street_limited", "roundabout", "mini_roundabout"]
                let compoundPredicate = NSPredicate(format: "class IN %@", roadPredicates)
```

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -684,6 +684,15 @@ extension RouteMapViewController: NavigationViewDelegate {
             streetLabelLayer.lineOpacity = NSExpression(forConstantValue: 1)
             streetLabelLayer.lineWidth = NSExpression(forConstantValue: 20)
             streetLabelLayer.lineColor = NSExpression(forConstantValue: UIColor.white)
+
+            if ![DirectionsProfileIdentifier.walking, DirectionsProfileIdentifier.cycling].contains( router.routeProgress.routeOptions.profileIdentifier) {
+                // filter out to road classes valid for motor transport
+                let roadPredicates = ["motorway", "motorway_link", "trunk", "trunk_link", "primary", "primary_link", "secondary", "secondary_link", "tertiary", "tertiary_link", "street", "street_limited", "roundabout", "mini_roundabout"].compactMap { roadClass -> NSPredicate? in
```
1ec5, line 690: I’m not sure about `mini_roundabout` being necessary, since it’s tagged as a node in OpenStreetMap and would thus be a point feature in the Streets source. We’re only querying for polyline features.

1ec5 (approved):
Suggested changelog entry:

* Fixed an issue where the current road name label (`WayNameLabel`) could momentarily show the name of a footpath or bike path even while driving. ([#2635](https://github.com/mapbox/mapbox-navigation-ios/pull/2635))
* Improved the performance of `WayNameLabel` in dense, urban areas. ([#2635](https://github.com/mapbox/mapbox-navigation-ios/pull/2635))

MapboxNavigation/RouteMapViewController.swift
avi-c: I'll adopt these changes. Worth getting the best performance.

PR #2643, opened 2020-09-21
Upscaling FPS on plugged in device

### LINKED ISSUE #2609: Reduced FPS when device is plugged-in
https://github.com/mapbox/mapbox-navigation-ios/blob/0fd7fc9d1c084250564f161a806f583291e63ba3/MapboxNavigation/NavigationMapView.swift#L19

This doesn't make sense to me. Can you elaborate why we switch to a low-power mode although we are charging the battery?

Udumft:
Resolves #2609 
FPS was a constant when device is plugged in (= 30 FPS), while on battery mode FPS could be scaled up to 60. 
Added upscaling FPS when device is plugged in with same conditions as on battery mode while preserving minimum FPS = 20  with battery usage case.


1ec5 (approved):
Suggested changelog entry:

* Fixed an issue where `NavigationMapView` redrew at a low frame rate even when the device was plugged in. ([#2643](https://github.com/mapbox/mapbox-navigation-ios/pull/2643))

MaximAlien (approved):

PR #2653, opened 2020-09-26
Add offline support to Navigation SDK and example application.

MaximAlien:
PR adds ability to use Offline Service in Navigation SDK and example application.

UI which allows to manage offline regions (list, download and remove):

Screen Shot 2020-09-25 at 6 03 50 PM


OfflineService/OfflineDataRegionTableViewCell.xib
```diff
@@ -0,0 +1,125 @@
...
+                                        <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                        <nil key="textColor"/>
+                                        <nil key="highlightedColor"/>
+                                    </label>
+                                    <progressView opaque="NO" contentMode="scaleToFill" verticalHuggingPriority="750" progress="0.5" translatesAutoresizingMaskIntoConstraints="NO" id="LX2-zi-58t">
+                                        <rect key="frame" x="0.0" y="23" width="914" height="2"/>
+                                    </progressView>
+                                </subviews>
+                                <constraints>
+                                    <constraint firstAttribute="height" constant="25" id="n0y-GK-vkA"/>
```
IodaMikeMapbox, line 85:
```suggestion
                                    <constraint firstAttribute="height" priority="750" constant="25" id="n0y-GK-vkA"/>
```
to prevent the conflict with `OfflineDataRegionTableViewCell.downloadProgressContainerHeightConstraint`

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,269 @@
...
+            }
+
+            let offlineDataRegions = expected?.value as? Array<Any>
+            offlineDataRegions?.forEach({
+                if let metadata = $0 as? OfflineDataRegionMetadata {
+                    self.offlineDataItems.append(OfflineDataItem(dataRegionMetadata: metadata,
+                                                                     mapPackMetadata: nil,
+                                                                     navigationPackMetadata: nil))
+                }
+            })
```
IodaMikeMapbox, line 198:
```suggestion
            offlineDataRegions?.forEach {
                if let metadata = $0 as? OfflineDataRegionMetadata {
                    self.offlineDataItems.append(OfflineDataItem(dataRegionMetadata: metadata,
                                                                     mapPackMetadata: nil,
                                                                     navigationPackMetadata: nil))
                }
            }
```

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,269 @@
...
+        return 200.0
+    }
+    
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+        
+        let offlineDataRegion = offlineDataItems[indexPath.row]
+        presentActionsAlertController(offlineDataRegion)
+    }
+
```
IodaMikeMapbox, line 166:
```suggestion
    
    // MARK: - Private methods
```

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,269 @@
...
+            OfflineService.registerObserver(for: self)
+
+            DispatchQueue.main.async {
+                self.tableView.reloadData()
+            }
+        })
+    }
+    
+    func updateOfflineDataRegions(for domain: OfflineDataDomain, metadata: OfflineDataRegionMetadata, delete: Bool = false) {
```
IodaMikeMapbox, line 208:
```suggestion
    private func updateOfflineDataRegions(for domain: OfflineDataDomain, metadata: OfflineDataRegionMetadata, delete: Bool = false) {
```

OfflineService/OfflineDataRegionTableViewCell.swift
```diff
@@ -0,0 +1,153 @@
...
+    
+    override func setSelected(_ selected: Bool, animated: Bool) {
+        super.setSelected(selected, animated: animated)
+    }
+    
+    override func prepareForReuse() {
+        setupUI()
+    }
+    
+    func setupUI() {
```
IodaMikeMapbox, line 55:
```suggestion
    private func setupUI() {
```

OfflineService/OfflineDataRegionTableViewCell.swift
```diff
@@ -0,0 +1,153 @@
...
+        
+        downloadProgressLabel.text = ""
+        downloadProgressLabel.font = UIFont.systemFont(ofSize: 11.0)
+        downloadProgressLabel.textColor = .darkGray
+        downloadProgressLabel.textAlignment = .center
+        
+        downloadProgressView.progress = 0.0
+    }
+    
+    func setupConstraints() {
```
IodaMikeMapbox, line 91:
```suggestion
    private func setupConstraints() {
```

OfflineService/OfflineDataRegionTableViewCell.swift
```diff
@@ -0,0 +1,153 @@
...
+                                                    attribute: .height,
+                                                    relatedBy: .equal,
+                                                    toItem: nil,
+                                                    attribute: .notAnAttribute,
+                                                    multiplier: 0.0,
+                                                    constant: 0)
+        downloadProgressContainerHeightConstraint?.isActive = true
+    }
+    
+    func updateDownloadProgressContainerHeight(constant: CGFloat) {
```
IodaMikeMapbox, line 103:
```suggestion
    private func updateDownloadProgressContainerHeight(constant: CGFloat) {
```

IodaMikeMapbox (changes requested):
Nice work! Still a few concerns.

Also what do you think about splitting ui logic and offline regions management logic? I mean at least place them in different classes like `OfflineServiceViewController` and `OfflineServiceDataSource` respectively

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,311 @@
...
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        setupUI()
+        listAvailableRegions()
+    }
+    
+    deinit {
+        // FIXME: In some cases this call might cause crash.
```
IodaMikeMapbox, line 156: looks like crash happens when you try to unregister object which wasn't registered

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,311 @@
...
+            let offlineDataRegions = expected?.value as? Array<Any>
+            offlineDataRegions?.forEach({
+                if let metadata = $0 as? OfflineDataRegionMetadata {
+                    self.offlineDataItems.append(OfflineDataItem(dataRegionMetadata: metadata,
+                                                                 mapPackMetadata: nil,
+                                                                 navigationPackMetadata: nil))
+                }
+            })
+
+            OfflineService.registerObserver(for: self)
```
IodaMikeMapbox, line 237: I think we should register `self` on view did load

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,311 @@
...
+    
+    public var peer: MBXPeerWrapper?
+    
+    // MARK: - UIViewController delegate methods
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        setupUI()
+        listAvailableRegions()
```
IodaMikeMapbox, line 152: we should show already downloaded regions even if they are not in the list gotten with `offlineService?.listAvailableRegions`, that is needed for cases when the region was downloaded and not presented in the list in response of the `offlineService?.listAvailableRegions`, e.g. it may happen when there is no internet connection

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,269 @@
...
+    static let title = NSLocalizedString("OFFLINE_SERVICE_TITLE", value: "Offline Service", comment: "Title for UIViewController and UIAlertController.")
+    static let close = NSLocalizedString("CLOSE_TITLE", value: "Close", comment: "Close title.")
+}
+
+class OfflineServiceViewController: UITableViewController, OfflineServiceObserver {
+    
+    let cellIdentifier = NSStringFromClass(OfflineDataRegionTableViewCell.self)
+    var offlineDataItems = [OfflineDataItem]()
+    var offlineService: OfflineService?
+    let tilesVersiom = "1.0"
```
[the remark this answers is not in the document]
MaximAlien: No longer used.

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,269 @@
...
+            print("Failed to create tiles folder with error: \(error)")
+        }
+
+        guard let outputDirectory = outputDirectoryURL?.path else { return }
+
+        offlineService = OfflineService.getInstanceForPath(outputDirectory, options: OfflineServiceOptions(username: username,
+                                                                                                           accessToken: accessToken,
+                                                                                                           baseURL: baseURL))
+
+        offlineService?.listAvailableRegions(forCallback: { [weak self] (expected) in
```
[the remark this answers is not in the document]
MaximAlien: Thanks, simplified method call.

OfflineService/OfflineDataRegionTableViewCell.xib
MaximAlien: Yep, decreased priority to prevent layout constraint warnings.

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,269 @@
...
+import MapboxCoreNavigation
+
+struct OfflineDataItem {
+    
+    var dataRegionMetadata: OfflineDataRegionMetadata
+    var mapPackMetadata: OfflineDataPackMetadata? = nil
+    var navigationPackMetadata: OfflineDataPackMetadata? = nil
+}
+
+struct OfflineServiceConstants {
```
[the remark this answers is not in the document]
MaximAlien: Thanks for your suggestion, but I think several constant values will be used across multiple entities, so having private `Constants` enum will not be agile enough. I've created `OfflineServiceConstants` as a separate struct to contain those values.

Example/ViewController.swift
```diff
@@ -109,9 +111,11 @@ class ViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         
+        // Trigger OfflineServiceManager singleton creation to make sure that Maps SDK is notified whenever offline packs are available.
+        let _ = OfflineServiceManager.instance
```
MaximAlien, line 115: OfflineServiceManager shared instance creation allows to notify Maps SDK/MBGLCore whether there are any offline packs which are already downloaded. Current OfflineService implementation is not able detect whether offline packs are available implicitly so without this check offline packs will not be visible.

I'm not really fond of using singleton, but it allows to prevent any errors with handling OfflineService lifecycle. I think another way of using `OfflineService` might involve its local instances which are created whenever they're needed (e.g. to list available packs etc), but major thing is that user has to *"login"* to `OfflineService` before `MGLMapView` object is created to be able to show offline packs correctly. 

We might consider this approach in wrapper implementation for OfflineService, which is tracked in another task. In any case let me know what you think about current implementation.

/cc @1ec5, @IodaMikeMapbox

OfflineService/OfflineServiceViewController.swift
MaximAlien: That's correct, on the other hand there is clash in data which is retrieved from `OfflineService.listAvailableRegions(forCallback:)` and `OfflineServiceObserver.onAvailable(for:metadata:pack:)` methods and previously I was calling first one and only in case if data was received observer was created. In case of error observer wasn't created - this lead to the crash. In any case I'd expect `OfflineServiceObserver` not to crash in such case, probably something we'd like to flag to Core SDK team.

OfflineService/OfflineServiceViewController.swift
MaximAlien: Moved to `OfflineServiceViewController.listAvailableRegions()` for now.

OfflineService/OfflineServiceViewController.swift
MaximAlien: Even though 1Tap for Android doesn't seem to support this I think it's a good idea to show already downloaded regions. Added this ability.

MaximAlien:
Adding latest result (for now valid region for San Francisco is available when using staff username and token). In example offline region is available for San Francisco for zoom levels between `12.0` and `16.0`, after downloading region Wi-Fi is disabled and current location is changed from Berlin to San Francisco.

output

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,300 @@
...
+            do {
+                guard let outputDirectoryURL = Bundle.mapboxCoreNavigation.suggestedTileURL?.appendingPathComponent("unpacked") else { return }
+                guard let packName = URL(string: pack.path)?.lastPathComponent else { return }
+                guard let packData = FileManager.default.contents(atPath: pack.path) else { return }
+                
+                try FileManager.default.createDirectory(at: outputDirectoryURL, withIntermediateDirectories: true, attributes: nil)
+                let temporaryPackURL = outputDirectoryURL.appendingPathComponent(packName).appendingPathExtension("tar")
+                try packData.write(to: temporaryPackURL)
+                
+                NavigationDirections.unpackTilePack(at: temporaryPackURL, outputDirectoryURL: outputDirectoryURL, progressHandler: { (totalBytes, unpackedBytes) in
```
MaximAlien, line 62: I have some concerns regarding this API, since we have to unpack archive so it can be consumed by Navigation Native all possible regions are unpacked in the same folder. For example here's approximate structure for several regions:
Screen Shot 2020-10-09 at 3 56 46 PM

As you can see due to folder naming in unarchived folders we might get conflicts which might lead to overwrite of certain files. There is also another problem with removal of Navigation offline region when using OfflineService API: even though original archive will be removed we no longer have control over unpacked files, thus they won't be removed and user will still be able to use them. This problem is partly fixed by calling `removeUnpackedTilesDirectory()` each time app starts - it'll remove whole unpacked folder and will unpack all available regions once again.

/cc @1ec5, @IodaMikeMapbox, @SiarheiFedartsou

MaximAlien:
> Nice work! Still a few concerns.
> 
> Also what do you think about splitting ui logic and offline regions management logic? I mean at least place them in different classes like `OfflineServiceViewController` and `OfflineServiceDataSource` respectively

`OfflineServiceDataSource` was added to provide updates with most recent `OfflineDataItem`s downloaded/removed using Offline Service.

OfflineService/OfflineServiceViewController.swift
MaximAlien: No longer relevant.

MaximAlien:
@1ec5, @IodaMikeMapbox offline navigation works in free drive and when requesting route. On the other hand for active guidance there is no way of providing `tilesUrl` and `tilesVersion` to navigator without breaking APIs in `RouteController` and possibly other classes. Currently navigator is created like this:
https://github.com/mapbox/mapbox-navigation-ios/blob/a9af590d181f3543e619db485b3b3502c83b9dd2/MapboxCoreNavigation/RouteController.swift#L24-L27

For now I don't see any good way of providing required info to navigator in `RouteController` without landing https://github.com/mapbox/mapbox-navigation-ios/pull/2593. Do you have any better ideas?

OfflineService/OfflineServiceDataSource.swift
```diff
@@ -0,0 +1,142 @@
...
+ for any events (downloading and removing certain Maps and Navigation packs) related to specific regions.
+ */
+open class OfflineServiceDataSource: OfflineServiceObserver {
+    
+    /**
+     The `OfflineServiceDataSource` delegate.
+     */
+    public weak var delegate: OfflineServiceDataSourceDelegate?
+    
+    private var offlineDataItems = [OfflineDataItem]()
```
IodaMikeMapbox, line 16: looks unused

OfflineService/OfflineServiceDataSource.swift
```diff
@@ -0,0 +1,142 @@
...
+
+            let offlineDataRegions = expected?.value as? Array<Any>
+            var offlineDataItems = [OfflineDataItem]()
+            offlineDataRegions?.forEach {
+                if let metadata = $0 as? OfflineDataRegionMetadata {
+                    offlineDataItems.append(OfflineDataItem(dataRegionMetadata: metadata))
+                }
+            }
+            
+            if !offlineDataItems.isEmpty {
```
IodaMikeMapbox, line 115: empty array is still a result, we may want to stop showing previously gotten regions

OfflineService/OfflineServiceViewController.swift
```diff
@@ -0,0 +1,210 @@
...
+        actionPayloads
+            .map({ payload in UIAlertAction(title: payload.0, style: payload.1, handler: payload.2) })
+            .forEach(alertController.addAction(_:))
+        
+        present(alertController, animated: true, completion: nil)
+    }
+    
+    // MARK: - OfflineServiceDataSourceDelegate methods
+    
+    func offlineServiceDataSource(_ dataSource: OfflineServiceDataSource, didUpdate offlineDataItems: [OfflineDataItem]) {
```
IodaMikeMapbox, line 84: there might be a situation when the following call of this method will contain a sublist of regions from the previous call, in that case we should remove some items from the `offlineDataItems`. I understand that in the current usage it's not a problem because the view controller most likely will be recreated between the calls, but there is no guarantees that in the future changes there will no be such situations

IodaMikeMapbox (changes requested):
The main complaint is the hardcoded values. And I don't fully share your way for configuring navigators, not that I need fame, but there is a pr that does it better, the pr is #2593

OfflineService/OfflineServiceConstants.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+struct OfflineServiceConstants {
```
IodaMikeMapbox, line 3: > Thanks for your suggestion, but I think several constant values will be used across multiple entities, so having private Constants enum will not be agile enough. I've created OfflineServiceConstants as a separate struct to contain those values.

The suggestion wasn't just about placing them inside the body of corresponding class.
And as I see there is no constant values used across multiple entities, moreover these constants is available for every class, the constants are too agile

Example/AppDelegate.swift
```diff
@@ -17,10 +17,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             window!.rootViewController = UIViewController()
         }
         
+        listMapboxFrameworks()
```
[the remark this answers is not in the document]
MaximAlien: This is convenience method which allows to quickly find out what versions of Mapbox frameworks are used in Example app. Since `AppDelegate` is main point of entry I think it's acceptable to add this method here, also not sure it's worth moving it to another utils entity. Do you have any other ideas about place for this method?

Example/CustomViewController.swift
```diff
@@ -37,7 +37,12 @@ class CustomViewController: UIViewController, MGLMapViewDelegate {
         super.viewDidLoad()
         
         let locationManager = simulateLocation ? SimulatedLocationManager(route: userIndexedRoute!.0) : NavigationLocationManager()
-        navigationService = MapboxNavigationService(route: userIndexedRoute!.0, routeIndex: userIndexedRoute!.1, routeOptions: userRouteOptions!, locationSource: locationManager, simulating: simulateLocation ? .always : .onPoorGPS)
+        navigationService = MapboxNavigationService(route: userIndexedRoute!.0,
+                                                    routeIndex: userIndexedRoute!.1,
+                                                    routeOptions: userRouteOptions!,
+                                                    locationSource: locationManager,
+                                                    simulating: simulateLocation ? .always : .onPoorGPS,
+                                                    tilesVersion: "2020_10_11-03_00_00")
```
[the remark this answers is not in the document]
MaximAlien: Moved `tilesVersion` to `OfflineServiceConstants`.

OfflineService/OfflineServiceDataSource.swift
MaximAlien: > The main complaint is the hardcoded values. And I don't fully share your way for configuring navigators, not that I need fame, but there is a pr that does it better, the pr is #2593

Even though current implementation adds support to use tiles downloaded via offline service for route request, active guidance and free drive by extending public APIs I agree that PR in #2593 can address navigators configuration better and provide singe point of entry for all Navigators we currently use. My concerns can be found in https://github.com/mapbox/mapbox-navigation-ios/pull/2653#issuecomment-708842296 as well.

I encourage Core Navigation team members to review your Navigator configuration PR as well.

MaximAlien:
> The main complaint is the hardcoded values. And I don't fully share your way for configuring navigators, not that I need fame, but there is a pr that does it better, the pr is #2593

Even though current implementation adds support to use tiles downloaded via offline service for route request, active guidance and free drive by extending public APIs I agree that PR in #2593 can address navigators configuration better and provide singe point of entry for all Navigators we currently use. My concerns can be found in https://github.com/mapbox/mapbox-navigation-ios/pull/2653#issuecomment-708842296 as well.

I encourage Core Navigation team members to review your Navigator configuration PR as well.

OfflineService/OfflineServiceDataSource.swift
```diff
@@ -0,0 +1,142 @@
...
+    }
+    
+    public func onExpired(for domain: OfflineDataDomain, metadata: OfflineDataRegionMetadata, pack: OfflineDataPack) {
+        NSLog("[OfflineServiceObserver] \(#function), identifier: \(metadata.id)")
+    }
+    
+    public func onErrored(for domain: OfflineDataDomain, metadata: OfflineDataRegionMetadata, pack: OfflineDataPack) {
+        NSLog("[OfflineServiceObserver] \(#function), identifier: \(metadata.id)")
+        
+        self.delegate?.offlineServiceDataSource(self, didFail: OfflineServiceError.genericError(message: "Error occured for: \(metadata.id)"))
```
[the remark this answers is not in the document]
MaximAlien: As I understand you removed `OfflineServiceDataSource.swift` in favor of API which wraps Offline Service. Let's address it in https://github.com/mapbox/mapbox-navigation-ios/pull/2704 then.

OfflineService/OfflineServiceViewController.swift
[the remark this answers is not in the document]
MaximAlien: Let's address it in https://github.com/mapbox/mapbox-navigation-ios/pull/2704 then.

OfflineService/OfflineServiceDataSource.swift
```diff
@@ -0,0 +1,142 @@
+import Foundation
+import MapboxCommon
+import MapboxCoreNavigation
+
+/**
+ `OfflineServiceDataSource` provides information about available offline regions downloaded via Offline Service. It also subscribes
+ for any events (downloading and removing certain Maps and Navigation packs) related to specific regions.
+ */
+open class OfflineServiceDataSource: OfflineServiceObserver {
```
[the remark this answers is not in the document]
MaximAlien: Let's address it in https://github.com/mapbox/mapbox-navigation-ios/pull/2704.

OfflineService/OfflineServiceDataSource.swift
MaximAlien: Let's address it in https://github.com/mapbox/mapbox-navigation-ios/pull/2704.

Example/AppDelegate.swift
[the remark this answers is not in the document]
IodaMikeMapbox: does it make sense in that case use `NSLog` instead of `print` to see those versions in the device log?

IodaMikeMapbox (approved):
I still have stylistic concerns, but they are not important enough to not approve

Issue #2658, opened 2020-09-28
layoutSubviews() is called too often on iOS 14.0.


MaximAlien:
**Mapbox Navigation SDK version:** `v1.0.0-rc.5`

On `iOS 14.0` `NavigationMapView.layoutSubviews()` is called too often, this leads to animated updates of `UserCourseView`. E.g. `iOS 13.5` `NavigationMapView.layoutSubviews()` was called only once during basic navigation session, whereas on iOS `14.0` `19` (!) times within 1 second. Because of this there is a lag in animated `UserCourseView` update. For example:

![output](https://user-images.githubusercontent.com/1496498/94490707-9f5a8380-019b-11eb-9837-a7076ed86103.gif)

PR #2659, opened 2020-09-28
Prevent updates to `UserCourseView` when layoutSubviews() is called.

MaximAlien:
Closing #2658.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -270,8 +270,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         super.layoutSubviews()
         
         //If the map is in tracking mode, make sure we update the camera after the layout pass.
-        if (tracksUserCourse) {
-            updateCourseTracking(location: userLocationForCourseTracking, camera:self.camera, animated: false)
+        if (tracksUserCourse) {            
+            if #available(iOS 14.0, *) {
+                // Since layoutSubviews() is called too often on iOS 14.0, it leads to lags in `UserCourseView` animated updates.
+                // Workaround is to allow `UserCourseView` updates only on iOS versions lower than 14.0.
+            } else {
+                updateCourseTracking(location: userLocationForCourseTracking, camera:self.camera, animated: false)
+            }
         }
     }
     


MapboxNavigation/NavigationMapView.swift
```diff
@@ -270,8 +270,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         super.layoutSubviews()
         
         //If the map is in tracking mode, make sure we update the camera after the layout pass.
-        if (tracksUserCourse) {
-            updateCourseTracking(location: userLocationForCourseTracking, camera:self.camera, animated: false)
+        if (tracksUserCourse) {            
+            if #available(iOS 14.0, *) {
+                // Since layoutSubviews() is called too often on iOS 14.0, it leads to lags in `UserCourseView` animated updates.
+                // Workaround is to allow `UserCourseView` updates only on iOS versions lower than 14.0.
```
1ec5, line 276: If the device orientation changes or the user closes the step table, the puck will be stuck in an incorrect state until the next location update. I think that would be worth documenting as a known issue and treating as tail work.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -270,8 +270,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            } else {
+                updateCourseTracking(location: userLocationForCourseTracking, camera:self.camera, animated: false)
```
1ec5, line 278: Just spitballing about alternatives that would allow us to keep updating the puck without it getting stuck on iOS 14:

* If the root cause is that kicking off a redundant view animation preempts any existing one, then `updateCourseTracking(location:camera:animated:)` should bail if `userLocationForCourseTracking` is unchanged.
* Perhaps for iOS 14 and above, NavigationMapView can keep a timestamp of the last time this method called `updateCourseTracking(location:camera:animated:)` and avoid calling it again within 1&nbsp;second (the typical interval between location updates).
* At one time, I think we had `updateCourseTracking(location:camera:animated:)` pass `UIView.AnimationOptions.beginFromCurrentState` into `UIView.animate(withDuration:delay:options:animations:completion:)`.

1ec5 (approved):
Meant to approve this PR. I think one of the approaches in https://github.com/mapbox/mapbox-navigation-ios/pull/2659#discussion_r496280866 would be worth pursuing, but it can wait until v1.0.1 if the implementation is nonobvious or risky.

MapboxNavigation/NavigationMapView.swift
MaximAlien: BTW I did some tests on iPad and iPhone Simulators and do not experience inaccurate state during device rotation. This is positive flow though. We also get some undesired `UserCourseView` shift when `updateCourseTracking(location:camera:animated:)` is called inside `layoutSubviews()`:

output

This can be taken care of in #2660.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Thanks for outlining possible alternatives. Since additional testing will require more time I think it'd be better to proceed with current implementation and consider other options in #2660.

PR #2661, opened 2020-09-29
Skip shape annotations when inserting route line into style

### LINKED ISSUE #2657: Route line obscures labels and shields
Calling `NavigationMapView.show(_:legIndex:)` places the route line above any labels and shields, obscuring them:

labels

Reproduces in navigation SDK v1.0.0-rc.5 with map SDK v6.2.1. Does not reproduce in navigation SDK v1.0.0-rc.4 with map SDK v6.1.0.

/cc @mapbox/navigation-ios

1ec5:
Detect and skip polyline and polygon annotation layers when scouring the style’s layers for the right layer to insert the route line above.

overlap

Fixes #2657.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -536,6 +536,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             var parentLayer: MGLStyleLayer? = nil
             for layer in style.layers.reversed() {
                 if !(layer is MGLSymbolStyleLayer) && !identifiers.contains(layer.identifier) {
+                    // MGLMapView automatically adds an MGLLineStyleLayer or MGLFillStyleLayer to the top of the layer stack when adding a polyline or polygon annotation, respectively. There’s no way to insert them lower in the layer stack, so they aren’t good indicators of where to insert the route line.
+                    // Detect and skip such a layer by checking if its source lacks a dedicated MGLSource subclass.
+                    if let vectorLayer = layer as? MGLVectorStyleLayer,
+                       let sourceIdentifier = vectorLayer.sourceIdentifier,
+                       let source = style.source(withIdentifier: sourceIdentifier), type(of: source) == MGLSource.self {
+                        continue
+                    }
+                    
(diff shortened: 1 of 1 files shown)


MaximAlien (approved):

avi-c (approved):

MapboxNavigation/NavigationMapView.swift
```diff
@@ -536,6 +536,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             var parentLayer: MGLStyleLayer? = nil
             for layer in style.layers.reversed() {
                 if !(layer is MGLSymbolStyleLayer) && !identifiers.contains(layer.identifier) {
+                    // MGLMapView automatically adds an MGLLineStyleLayer or MGLFillStyleLayer to the top of the layer stack when adding a polyline or polygon annotation, respectively. There’s no way to insert them lower in the layer stack, so they aren’t good indicators of where to insert the route line.
+                    // Detect and skip such a layer by checking if its source lacks a dedicated MGLSource subclass.
+                    if let vectorLayer = layer as? MGLVectorStyleLayer,
+                       let sourceIdentifier = vectorLayer.sourceIdentifier,
+                       let source = style.source(withIdentifier: sourceIdentifier), type(of: source) == MGLSource.self {
```
1ec5, line 543: Another approach would’ve been to hard-code the [`com.mapbox.annotations.shape`](https://github.com/mapbox/mapbox-gl-native/blob/bcbaa3585bcde2de76ad778e8520e02baf9ba84c/src/mbgl/annotation/annotation_manager.cpp#L32) prefix that mbgl applies to each shape annotation layer’s identifier, but that would’ve been an implementation detail difficult to have any control over.

Looking for an unspecialized MGLSource is roundabout in its own way. However, the risk is only that some other exotic kind of source not covered by the style specification would also be skipped, which I think is an acceptable risk.

1ec5:
Cherry-picked into the release-v1.0 branch as c992102e6293e07bd844f98e0e72d561892914af.

PR #2667, opened 2020-09-30
Exclude arm64 simulator from CocoaPods-managed targets in Xcode 12

1ec5:
Temporarily modified the podspecs to exclude the arm64 simulator from the CocoaPods-generated target and application target, based on the workaround in https://github.com/CocoaPods/CocoaPods/issues/10065#issuecomment-694266259 https://github.com/CocoaPods/CocoaPods/issues/10104#issuecomment-700918704 but conditionalized for Xcode 12, just like the Carthage workaround in #2649. Unlike that workaround, this one isn’t sensitive to patch releases of Xcode, so we don’t need the extra case for Xcode 12.0.1 seen in #2663.

Unfortunately, this workaround automatically modifies the `EXCLUDED_ARCHS[sdk=iphonesimulator*]` and `EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200` build settings of the application target, which could eventually be incompatible with another pod if that other pod employs the same workaround. #2672 tracks removing this workaround.

~~This workaround apparently requires Xcode 11.7 and above to build from source, so I’ve also made that a minimum requirement for both CocoaPods and Carthage users for consistency.~~

Before merging:

* [x] Lint podspecs with Xcode 11
* [x] Lint podspecs with Xcode 12
* [x] Push test podspecs to CocoaPods trunk with Xcode 12
* [x] Build and run internal dogfooding application in Xcode 11
   * [x] Debug
      * [x] Simulators
      * [x] Devices
   * [x] Release
      * [x] Simulators
      * [x] Devices
* [x] Build and run internal dogfooding application in Xcode 12
   * [x] Debug
      * [x] Simulators
      * [x] Devices
   * [x] Release
      * [x] Simulators
      * [x] Devices

Fixes #2665.

/ref #2652
/cc @mapbox/navigation-ios

PR #2668, opened 2020-09-30
MIgrate nav docs script over ios-sdk repo

danswick:
Related: https://github.com/mapbox/mapbox-directions-swift/pull/467

This PR will move the [shell script for generating documentation](https://github.com/mapbox/ios-sdk/blob/publisher-production/scripts/update-navigation.sh) from /ios-sdk into this repo. The script will generate the documentation, switch and commit the new version on the publisher-production branch.

To do:
- [ ] Verify version branch naming convention
- [ ] Test script
- [ ] Document!

cc @katydecorah and @1ec5 

Ref mapbox/ios-sdk#980

### DIFF
--- a/scripts/update-navigation.sh
+++ b/scripts/update-navigation.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -e
+set -o pipefail
+set -u
+
+function step { >&2 echo -e "\033[1m\033[36m* $@\033[0m"; }
+function finish { >&2 echo -en "\033[0m"; }
+trap finish EXIT
+
+OUTPUT="/tmp/`uuidgen`"
+RELEASE_BRANCH=${1:-master}
+
+step "Updating mapbox-navigation-ios…"
+git fetch --depth=1 --prune
+git fetch --tags
+git checkout $RELEASE_BRANCH
+VERSION=$( git describe --tags --match=v*.*.* --abbrev=0 | sed 's/^v//' )
+
+# I don't think mapbox-navigation-ios uses this branch naming convention
+# git checkout $VERSION
+
+step "Updating jazzy…"
+gem install jazzy
+
+step "Generating new docs for ${VERSION}…"
+OUTPUT=${OUTPUT} scripts/document.sh
+
+step "Moving new docs folder to ./$VERSION"
+rm -rf "./$VERSION"
+mkdir -p "./navigation/api/$VERSION"
+mv -v $OUTPUT/* "./navigation/api/$VERSION"
+
+step "Switching branch to publisher-production"
+git checkout origin/publisher-production
+step "Committing API docs for $VERSION"
+git add "./$VERSION"
+git commit -m "[navigation] Add Mapbox Directions for Swift API docs for $VERSION [ci skip]" --no-verify
+
+step "Finished updating documentation"
--- a/scripts/update-navigation.sh
+++ b/scripts/update-navigation.sh
@@ -12,7 +12,7 @@ OUTPUT="/tmp/`uuidgen`"
 RELEASE_BRANCH=${1:-master}
 
 step "Updating mapbox-navigation-ios…"
-git fetch --depth=1 --prune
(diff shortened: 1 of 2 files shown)


scripts/update-navigation.sh
```diff
@@ -0,0 +1,40 @@
...
+RELEASE_BRANCH=${1:-master}
+
+step "Updating mapbox-navigation-ios…"
+git fetch --depth=1 --prune
+git fetch --tags
+git checkout $RELEASE_BRANCH
+VERSION=$( git describe --tags --match=v*.*.* --abbrev=0 | sed 's/^v//' )
+
+# I don't think mapbox-navigation-ios uses this branch naming convention
+# git checkout $VERSION
```
1ec5, line 21: The tag convention here is `v$VERSION`.

scripts/update-navigation.sh
```diff
@@ -0,0 +1,40 @@
...
+step "Moving new docs folder to ./$VERSION"
+rm -rf "./$VERSION"
+mkdir -p "./navigation/api/$VERSION"
+mv -v $OUTPUT/* "./navigation/api/$VERSION"
+
+step "Switching branch to publisher-production"
+git checkout origin/publisher-production
+step "Committing API docs for $VERSION"
+git add "./$VERSION"
+git commit -m "[navigation] Add Mapbox Directions for Swift API docs for $VERSION [ci skip]" --no-verify
```
1ec5, line 38: Should this be the navigation SDK documentation?

scripts/update-navigation.sh
```diff
@@ -0,0 +1,40 @@
...
+
+function step { >&2 echo -e "\033[1m\033[36m* $@\033[0m"; }
+function finish { >&2 echo -en "\033[0m"; }
+trap finish EXIT
+
+OUTPUT="/tmp/`uuidgen`"
+RELEASE_BRANCH=${1:-master}
+
+step "Updating mapbox-navigation-ios…"
+git fetch --depth=1 --prune
```
1ec5, line 15: I’d just make this a normal `fetch`. Otherwise it could cause the repository to get in a bad state for working with other branches for development purposes later.

1ec5 (commented):
Great, I think we should also name the script something more descriptive, like publish-documentation.sh.

1ec5:
The CircleCI build failures are due to Homebrew unreliability that #2667 will work around by downloading Carthage directly from GitHub.

scripts/update-navigation.sh
danswick: Whatever works! This is just copied over from the old one (which I don't think ever actually ran).

scripts/update-navigation.sh
danswick: Ah yes, this is just sloppy copypasta. My mistake. As far as I can tell looking through `ios-sdk` PRs, I don't think the nav automation has ever successfully run, so we have a pretty clean slate to implement this in a way that works best for you.

1ec5: The navigation SDK update script has successfully run on a number of occasions. We’ve always run it from a local machine, though, never on a CI environment.

scripts/update-navigation.sh
```diff
@@ -0,0 +1,40 @@
...
+
+step "Updating jazzy…"
+gem install jazzy
+
+step "Generating new docs for ${VERSION}…"
+OUTPUT=${OUTPUT} scripts/document.sh
+
+step "Moving new docs folder to ./$VERSION"
+rm -rf "./$VERSION"
+mkdir -p "./navigation/api/$VERSION"
```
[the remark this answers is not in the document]
danswick: Thanks for pointing that out, @MaximAlien. You are correct and I've updated this value.

danswick:
Thank you @1ec5 and @MaximAlien for your comments. I've made some updates to address the issues you've pointed out. I tested locally and all seemed to work. I would appreciate another tester if anyone has some spare cycles.

PR #2670, opened 2020-10-01
Fix CarPlay crash on iOS 14 after a navigation session

### LINKED ISSUE #2655: CarPlay crash on iOS 14 after a navigation session
**Mapbox Navigation SDK version:** `v1.0.0-rc.5`

https://github.com/mapbox/mapbox-navigation-ios/blob/e810a461d571083ed9682a8d6302bf27815c9415/MapboxNavigation/CarPlayManager.swift#L685

This method is deprecated in iOS 14 in favor of `popToRootTemplate(animated: Bool, completion: ((Bool, Error?) -> Void)? = nil)`.
Without this callback, this and similar template methods might crash with an exception like this: `An error was encountered during a template operation, but no completion block was specified`. This crash happens only if the template method fails and don't a completion block to handle an error. Providing empty completion block for iOS14 will solve crashing, but also will hide the root problem: an error at the line mention above, that happens after navigation session is finished: `No templates were available to be popped`.

IodaMikeMapbox:
Resolves #2655

### DIFF
--- a/MapboxNavigation/CarPlayManager.swift
+++ b/MapboxNavigation/CarPlayManager.swift
@@ -426,7 +426,7 @@ extension CarPlayManager {
             }
 
             let mapTemplate = interfaceController?.rootTemplate as? CPMapTemplate
-            interfaceController?.popToRootTemplate(animated: true)
+            popToRootTemplate(interfaceController: interfaceController, animated: true)
             mapTemplate?.present(navigationAlert: alert, animated: true)
             return
         case let .success(response):
@@ -487,8 +487,7 @@ extension CarPlayManager: CPMapTemplateDelegate {
         if simulatesLocations == true {
             service.simulationSpeedMultiplier = simulatedSpeedMultiplier
         }
-
-        interfaceController.popToRootTemplate(animated: false)
+        popToRootTemplate(interfaceController: interfaceController, animated: false)
         let navigationMapTemplate = self.mapTemplate(forNavigating: trip)
         interfaceController.setRootTemplate(navigationMapTemplate, animated: true)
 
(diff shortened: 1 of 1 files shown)


MapboxNavigation/CarPlayManager.swift
```diff
@@ -664,6 +663,24 @@ extension CarPlayManager: CPMapTemplateDelegate {
         let shiftedCenterCoordinate = mapView.centerCoordinate.coordinate(at: distance, facing: shiftedDirection)
         mapView.setCenter(shiftedCenterCoordinate, animated: true)
     }
+
+    private func popToRootTemplate(interfaceController: CPInterfaceController?, animated: Bool) {
+        guard let interfaceController = interfaceController else { return }
+        // TODO: uncomment
+//        if #available(iOS 14.0, *) {
```
MaximAlien, line 670: Why this part is commented and marked as `TODO`? What about using `open func popToRootTemplate(animated: Bool, completion: ((Bool, Error?) -> Void)? = nil)` for iOS `14.0` and keep using `open func popToRootTemplate(animated: Bool)` for previous versions?

MaximAlien (approved):
Approving, on the other hand I'd rather improve TODO statement to make it more descriptive (point to method which we should use for iOS 14 when Xcode 11 support is dropped) instead to leaving commented code.
Please also cherry-pick current fix to `release-1.0` branch.

MapboxNavigation/CarPlayManager.swift
IodaMikeMapbox: because the xcode <12 don't know about the `open func popToRootTemplate(animated: Bool, completion: ((Bool, Error?) -> Void)? = nil)` and hiding it inside an if statement doesn't mean that it shouldn't know it. I'll enhance the comment

IodaMikeMapbox: ok, you've just done it, never mind

MaximAlien:
Cherry picked to `release-v.1.0` branch in https://github.com/mapbox/mapbox-navigation-ios/commit/c7f7564791b19f315ace5787ef400fb9377e0131.

Issue #2498, opened 2020-07-28
Style.fontFamily is ignored


1ec5:
#330 added a `Style.fontFamily` property that is documented to affect the font of all labels in the UI. #536 and #2013 removed the code that used and implemented this property, but it’s still there in the public API, confusing developers who want to make the UI match the overall application’s look and feel. We should either remove the property or restore the functionality by explicitly applying the font to every StylableLabel subclass. If we remove the property, we should do it in v1.0.0, since it would be a backwards-incompatible change.

/cc @mapbox/navigation-ios @ZiZasaurus @frederoni

1ec5:
As a workaround, set the `font` or `normalFont` property of the appearance proxy of every StylableLabel subclass and the `textFont` property of the appearance proxy of every StylableButton subclass.

1ec5:
A fix is ready for review in #2679.

Issue #2645, opened 2020-09-21
ETA label is too small


1ec5:
The ETA label is the same size as the distance traveled label. It should be considerably larger as seen in #2569.

b30c71533e884bd807367690a06fbc569b5b5168 | 7866e550a9b363d2fe438669c2da2429361a4ce4
----|----
<img src="https://user-images.githubusercontent.com/1231218/91210315-e13a6a80-e6c1-11ea-85da-b2d8ee6c4229.png" width="300" alt="big"> | <img src="https://user-images.githubusercontent.com/1231218/93814668-b3d1d580-fc09-11ea-8b53-05ccb34e14e4.png" width="300" alt="small">

This regression was introduced toward the end of the v1.0 release cycle.

/cc @mapbox/navigation-ios

avi-c:
Looks like the UIAppearance font property is not being applied properly. Indeed, even the small size seen now for the distance travelled label is incorrect at 16pts and not the 18pts it was before. 16 points is what's inherited from the base `StylableLabel` class, not the 18 intended for the`DistanceRemainingLabel` subclass.

Not sure why this changed partway through the release though.

PR #2679, opened 2020-10-05
Apply UIAppearance proxy overrides for font style to normalFont property and not font property

avi-c:
Make sure to apply UIAppearance style values to the normalFont property and not font property since UILabel font properties are not properly exposed as UIAppearance stylable.

Similar to what was already being done for text color.

Fix for #2645.

### DIFF
--- a/MapboxNavigation/DayStyle.swift
+++ b/MapboxNavigation/DayStyle.swift
@@ -66,7 +66,7 @@ open class DayStyle: Style {
             tintColor = .defaultTint
         }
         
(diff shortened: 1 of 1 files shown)


1ec5:
This fixes #2498, not #2645.

1ec5 (approved):
Suggested changelog entries:

* Fixed an issue in which the estimated arrival time label had too small of a font. ([#2679](https://github.com/mapbox/mapbox-navigation-ios/pull/2679))
* Fixed an issue where various labels ignored the `Style.fontFamily` property. ([#2679](https://github.com/mapbox/mapbox-navigation-ios/pull/2679))

dersim-davaod (approved):
LGTM

Issue #2671, opened 2020-10-01
UserCourseView is not following route progress on CarPlay on iOS 14.0.


MaximAlien:
On CarPlay shipped with `iOS 14` UserCourseView is not following route progress correctly during location simulation. 

CarPlay iOS 14 behavior:
![1](https://user-images.githubusercontent.com/1496498/94842133-df09b100-03cf-11eb-8685-d224df4dff1e.gif)

CarPlay iOS 13 behavior:
![2](https://user-images.githubusercontent.com/1496498/94842151-e6c95580-03cf-11eb-8b12-1b20bb7ee08f.gif)

1ec5:
If I remember correctly, there’s supposed to be a sweeping zooming-in animation at the beginning of either the browsing or navigation activity. It looks as if that zooming animation is occurring but at only 1&nbsp;frame per second, with the frames backing up instead of skipping.

@zugaldia pointed out that mapbox/mapbox-gl-native-ios#411 landed recently. That change is good overall, but I suppose there’s a risk that something exotic like CarPlay could regress as a result.

PR #2683, opened 2020-10-07
Add workaround which prevents UI lags because of excessive calls to `viewDidLayoutSubviews()` on iOS 14.

MaximAlien:
Closing #2671. 

I don't really like current workaround as it only partly fixes original problem (there will still be UI lags every `10` seconds based on `contentInsetUpdateDelay` variable). Here's example (as you can see every 10 seconds there is still visible lag):

output

I've also noticed that there are no drawbacks when we do not set `contentInset` on `MGLMapView` at all on iOS 14, so we might want to consider such approach as well.

Best way would be to address issue reported on Maps SDK [side](https://github.com/mapbox/mapbox-gl-native-ios/issues/500).

### DIFF
--- a/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -96,6 +96,9 @@ public class CarPlayNavigationViewController: UIViewController, NavigationMapVie
     
     var styleObservation: NSKeyValueObservation?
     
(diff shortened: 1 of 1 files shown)


MaximAlien:
Root cause of this issue was fixed in https://github.com/mapbox/mapbox-gl-native-ios/issues/500 after update to Maps SDK `6.2.2`.

Issue #2637, opened 2020-09-16
End of route view UI is broken on iOS 10.


MaximAlien:
End of route view is not correctly laid out on `iOS 10`.

Actual:

<img width="225" alt="Screen Shot 2020-09-16 at 3 27 42 PM" src="https://user-images.githubusercontent.com/1496498/93399000-7cda7900-f831-11ea-95a8-fc1fb4179a78.png">

Expected:

<img width="225" alt="Screen Shot 2020-09-16 at 3 31 11 PM" src="https://user-images.githubusercontent.com/1496498/93399104-b4e1bc00-f831-11ea-869f-349dc05730fa.png">

1ec5:
Does this issue reproduce in v0.40.0? If not, was the regression introduced as part of #2596 or something else?

MaximAlien:
Yes, looks like it's regression in `v1.0.0`. At the same time I've added patch in #2596 and it doesn't cause any problems. Looks like something else causes this. 

Here's what we get in `v0.40.0`:

<img width="200" alt="Screen Shot 2020-09-17 at 10 51 04 AM" src="https://user-images.githubusercontent.com/1496498/93509003-fd51b600-f8d3-11ea-9635-936bd0961c32.png">

ShanMa1991:
I don't think it's  the regression problem, it's similar to mapbox/mapbox-navigation-ios#2680, not only shows up in iOS10.  It seems that the judgment of whether it has arrived at the destination, didn't pass the result to change the topBannerContainer content, or to adjust the userCourseView center to the destination. It should be related with the NavigationStatus of RouterController, and the execution sequence to update the route and to show end of route.

MaximAlien:
This is UI related issue which is reproducible only on `iOS 10`. As you can see on original screenshot feedback view is clipped. I'd recommend testing end of route view on various iOS versions to get a sense what's wrong on `iOS 10`.
<img width="225" alt="93399000-7cda7900-f831-11ea-95a8-fc1fb4179a78" src="https://user-images.githubusercontent.com/1496498/95602885-6d9eb380-0a0a-11eb-9e1e-9bed25d7130b.png">

ShanMa1991:
I see. Thanks for pointing it out the feedback container problem! I'll work on that.

PR #2690, opened 2020-10-12
Solve the End of route view UI broken before iOS 11

ShanMa1991:
To solve the issue for fix mapbox/mapbox-navigation-ios#2637 , I adjusted the constraint of the UI components inside the `End Of Route View Controller`. This is caused by the `safe area` is not the same as the `End Of Route View Container` under iOS 11. So I changed both the `MBSeparatorView.top` and the `Arrival Labels Container.top`  from the `safeArea.top` to the `Root.top`. It will remove the extra white padding above the `End Of Route View Container`  and display the Rating component correctly.

fixed_ios10_frame

fixed_ios10

/cc @mapbox/navigation-ios


MaximAlien (approved):

1ec5 (approved):
Suggested changelog entry:

* Fixed an issue where the rating stars in the end-of-route feedback panel are cut off on iOS 10. ([#2690](https://github.com/mapbox/mapbox-navigation-ios/pull/2690))

Issue #2555, opened 2020-08-19
Indicate reduced accuracy location mode


1ec5:
Starting in iOS 14, the system Location Services prompt will offer the user an option to share only crude location updates with the application. Reduced accuracy is unlikely to be quite as useful for turn-by-turn navigation as full accuracy, though it might be a valid use case for free-driving, depending on the application.

mapbox/mapbox-gl-native-ios#372 tracks adding an accuracy halo to the user puck during course tracking, similar to the halo around the user dot during location or heading tracking. The UserPuckCourseView drawing code is more or less a straight copy of the course mode of the map SDK’s MGLFaux3DUserLocationAnnotationView, so it should be straightforward to keep UserPuckCourseView in sync.

Accurate location tracking is critical to turn-by-turn navigation functionality, so we should revive the “Weak GPS Signal” status bar that was introduced in #490. We can call `NavigationViewController.showStatus(title:spinner:duration:animated:interactive:)` to display that status bar.

/ref #2554
/cc @mapbox/navigation-ios @mapbox/maps-ios

jill-cardamon:
Clarification on status banner: Could you even generate a route with user's precise location?
It would require a lot of steps for a person to turn precise location off in the middle of turn-by-turn navigation, so this banner is a bit of an edge case. This case might be more likely to show up during free driving.

Issue #2691, opened 2020-10-13
Duration of notifyUserAboutLowVolumeIfNeeded() status


jill-cardamon:
The current duration of `navigationViewController.notifyUserAboutLowVolumeIfNeeded()` is only 3 seconds. If someone is not focusing on their screen, they will potentially miss it. The banner is not turned off by an actual change in volume (and only shows up at the beginning of a trip). Initially introduced in #2121 

Seems like volume adjustment has been an issue for a while and loops back to the problem noted in #1826

PR #2649, opened 2020-09-23
Fix PassiveLocationManager crash with map SDK v6.2.0 on iOS 14

1ec5:
PassiveLocationManager is now compatible with map SDK v6.2.0 on iOS 14 and no longer triggers the crash in mapbox/mapbox-gl-native-ios#470 when updating the user dot. PassiveLocationManager and PassiveLocationDataSource implement members added to MGLLocationManager in map SDK v6.2.0 for basic compatibility with iOS 14’s approximate location feature.

This PR maintains compatibility with map SDK v6.0.0–v6.1._x_ even while adding support for v6.2.0–v6._x_. The navigation SDK simultaneously supports applications that need to stay on v6.1.0 because of mapbox/mapbox-gl-native-ios#450 and those that need to upgrade to v6.2.0 due to iOS 14 location changes. Maintaining backwards compatibility required some Objective-C runtime gymnastics, but it’s minor compared to the tricks we pulled in the past to make this SDK compatible with Objective-C code.

Updated Cartfile.resolved and Podfile.lock to reflect the latest dependencies that developers would obtain by default from Carthage and CocoaPods, respectively. Added a script based on https://github.com/Carthage/Carthage/issues/3019#issuecomment-665136323 to work around a `lipo` error when building dependencies using Carthage and Xcode 12. Added an Xcode 12 build bot.

Fixes #2554.

/cc @mapbox/navigation-ios @mapbox/maps-ios @mapbox/1tap-ios

PR #2693, opened 2020-10-13
added banner to notify user to enable precise location during turn-by-turn navigation

jill-cardamon:
Added banner to notify a user to enable precise location during turn-by-turn navigation using `CLLocationManager` and `locationDidChangeAuthorization(..)`, bringing this value up the delegate chain. Banner displays for 9 seconds (about the time it takes for a user to notice the banner and change their settings).
- [x] Add banner
- [x] Add notifications
- [x] Make this work on Xcode 11
- [x] Update changelog

Simulator Screen Shot - iPhone 8 - 2020-10-13 at 14 37 43

fix mapbox/mapbox-navigation-ios#2555
/cc @mapbox/navigation-ios


MapboxNavigation/NavigationViewController.swift
```diff
@@ -687,6 +688,17 @@ extension NavigationViewController: NavigationServiceDelegate {
         return navigationComponents.allSatisfy { $0.navigationServiceShouldDisableBatteryMonitoring(service) }
     }
     
+    public func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        let title = NSLocalizedString("ENABLE_PRECISE_LOCATION", bundle: .mapboxNavigation, value: "Enable precise location to  navigate", comment: "Label indicating precise location is off and needs to be turned on to navigate")
+        if #available(iOS 14.0, *), manager.accuracyAuthorization == .reducedAccuracy {
```
1ec5, line 693: Xcode 11 builds are [failing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/1984/workflows/ceeb96e7-7c3a-498d-866f-030a7aa211d5/jobs/22062) due to references to `CLLocationManager.accuracyAuthorization`, which the iOS 13 SDK built into Xcode 11 doesn’t know about:

```
/Users/distiller/project/MapboxNavigation/NavigationViewController.swift:693:45: error: value of type 'CLLocationManager' has no member 'accuracyAuthorization'
        if #available(iOS 14.0, *), manager.accuracyAuthorization == .reducedAccuracy {
                                    ~~~~~~~ ^~~~~~~~~~~~~~~~~~~~~
/Users/distiller/project/MapboxNavigation/NavigationViewController.swift:693:71: error: cannot infer contextual base in reference to member 'reducedAccuracy'
        if #available(iOS 14.0, *), manager.accuracyAuthorization == .reducedAccuracy {
                                                                     ~^~~~~~~~~~~~~~~
```

The `#available` is important because it prevents this code from running on iOS 13, but that isn’t quite enough for the compiler. I ran into a similar problem when authoring #2649. Fortunately, we can apply two different workarounds:

* Because CLLocationManager is an Objective-C class, every property has a corresponding key path (as in KVO). Instead of getting `CLLocationManager.accuracyAuthorization` directly, call `value(forKey:)` to get the value for the key `accuracyAuthorization`.
* `CLAccuracyAuthorization.reducedAccuracy` is also new to the iOS 14 SDK, so #2649 implements an `MBNavigationAccuracyAuthorization` enumeration as a compatibility shim in Objective-C.

https://github.com/mapbox/mapbox-navigation-ios/blob/64dd90e79e882516d50e3577d854f80bbc1da2b8/MapboxNavigation/PassiveLocationManager.swift#L48-L50 https://github.com/mapbox/mapbox-navigation-ios/blob/8403eebf864d7693d61c59cfce5819f1e241b155/MapboxNavigation/CLLocationManager%2BMGLNavigationAdditions.h#L7

MapboxNavigation/NavigationViewController.swift
```diff
@@ -687,6 +688,17 @@ extension NavigationViewController: NavigationServiceDelegate {
         return navigationComponents.allSatisfy { $0.navigationServiceShouldDisableBatteryMonitoring(service) }
     }
     
+    public func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        let title = NSLocalizedString("ENABLE_PRECISE_LOCATION", bundle: .mapboxNavigation, value: "Enable precise location to  navigate", comment: "Label indicating precise location is off and needs to be turned on to navigate")
```
langsmith, line 692:
```suggestion
        let title = NSLocalizedString("ENABLE_PRECISE_LOCATION", bundle: .mapboxNavigation, value: "Enable precise location to navigate", comment: "Label indicating precise location is off and needs to be turned on to navigate")
```

MapboxNavigation/NavigationViewController.swift
```diff
@@ -687,6 +688,17 @@ extension NavigationViewController: NavigationServiceDelegate {
...
+        if #available(iOS 14.0, *), manager.accuracyAuthorization == .reducedAccuracy {
+            // It takes about 9 seconds to recognize the banner and change location authorization in settings
+            showStatus(title: title, spinner: false, duration: 9, animated: true, interactive: false)
```
1ec5, line 695: Unlike some of the other things we use the status view for, approximate location mode is persistent, requiring the user to take action to disable it. Nine seconds seems like a long time when staring at a simulator, but if you’re a driver juggling multiple tasks, you might still miss the status. As tail work or perhaps as part of this PR, consider leaving the status visible. Apparently if the duration is infinite, the status won’t autohide:

https://github.com/mapbox/mapbox-navigation-ios/blob/e8abb39222563efc56d4721bb25a0167fba95091/MapboxNavigation/StatusView.swift#L122

There’s a separate method on StatusView to manually hide a status, though it isn’t currently exposed as part of the NavigationStatusPresenter protocol.

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -139,6 +139,17 @@ public protocol NavigationServiceDelegate: class, UnimplementedLogging {
...
+    /**
+     Called when the location manager's accuracy authorization changed.
+     
+     You can implement this method to allow the navigation service to check if the user changed accuracy authorization, especially if reducedAccuracy is enabled. This method is only relevant for iOS 14 and above.
+     
+     - parameter service: The navigation service that will alert that user that reducedAccuracy is enabled.
+     - parameter manager: The location manager.
+     - returns: A bool indicating whether to alert the user to enable preciseAccuracy.
+     */
+    func locationManagerDidChangeAuthorization(_ manager: CLLocationManager)
```
1ec5, line 151: Part of the delegate pattern is that the first part of the method name always names the delegator, so that a single class can serve as the delegate for multiple delegators. Following that pattern, this method would be named `navigationServiceDidChangeAuthorization(_:)` and take the NavigationService as the first argument, not the CLLocationManager. It would then fall to the delegate’s implementation of this method to get the location manager from the navigation service.

MapboxNavigation/NavigationViewController.swift
1ec5, line 692: Nit: there’s an extra space between “to” and “navigate”. Apparently UIKit doesn’t collapse extra spaces, so a gap shows up on screen.

1ec5, line 692: This is a new user-facing string. You’ve already followed most of [the steps for adding a new user-facing string](https://github.com/mapbox/mapbox-navigation-ios/blob/64dd90e79e882516d50e3577d854f80bbc1da2b8/CONTRIBUTING.md#adding-user-facing-text), but the base localization file still needs to be updated using a script.

1ec5 (commented):
Suggested changelog entry:

* Added the `NavigationServiceDelegate.navigationServiceDidChangeAuthorization(_:)` method to find out when the user changes the application’s Location Services permissions. ([#2693](https://github.com/mapbox/mapbox-navigation-ios/pull/2693))

MapboxNavigation/NavigationViewController.swift
jill-cardamon: I noticed this. I tried to set a longer duration, but it wouldn't go away if locationAuthorization was changed. I will look into exposing that part of the protocol!

MapboxNavigation/NavigationViewController.swift
jill-cardamon: Thanks for catching this!!

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -146,9 +146,8 @@ public protocol NavigationServiceDelegate: class, UnimplementedLogging {
      
      - parameter service: The navigation service that will alert that user that reducedAccuracy is enabled.
      - parameter manager: The location manager.
-     - returns: A bool indicating whether to alert the user to enable preciseAccuracy.
      */
-    func locationManagerDidChangeAuthorization(_ manager: CLLocationManager)
+    func navigationServiceDidChangeAuthorization(_ service: NavigationService, _ manager: CLLocationManager)
```
1ec5, line 150: Only the first argument should ever have a `_` label, and only if the first part of the method name already refers to the argument or makes the method read more like a sentence.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -688,13 +688,16 @@ extension NavigationViewController: NavigationServiceDelegate {
...
-        let title = NSLocalizedString("ENABLE_PRECISE_LOCATION", bundle: .mapboxNavigation, value: "Enable precise location to  navigate", comment: "Label indicating precise location is off and needs to be turned on to navigate")
-        if #available(iOS 14.0, *), manager.accuracyAuthorization == .reducedAccuracy {
-            // It takes about 9 seconds to recognize the banner and change location authorization in settings
-            showStatus(title: title, spinner: false, duration: 9, animated: true, interactive: false)
+    public func navigationServiceDidChangeAuthorization(_ service: NavigationService, _ manager: CLLocationManager) {
+        let title = NSLocalizedString("ENABLE_PRECISE_LOCATION", bundle: .mapboxNavigation, value: "Enable precise location to navigate", comment: "Label indicating precise location is off and needs to be turned on to navigate")
+        
+        // CLLocationManager.accuracyAuthorization was introduced in the iOS 14 SDK in Xcode 12, so Xcode 11 doesn’t recognize it.
+        let accuracyAuthorization = MBNavigationAccuracyAuthorization(rawValue: manager.value(forKey: "accuracyAuthorization") as! Int)
```
1ec5, line 695: For additional safety, use `as?` and handle the case where `accuracyAuthorization` isn’t an integer by doing nothing.

MapboxCoreNavigation/NavigationService.swift
```diff
@@ -418,6 +418,16 @@ extension MapboxNavigationService: CLLocationManagerDelegate {
         //Finally, pass the update onto the router.
         router.locationManager?(manager, didUpdateLocations: locations)
     }
+    
+    public func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        if #available(iOS 14.0, *) {
+            NotificationCenter.default.post(name: .locationAuthorizationDidChange, object: nil)
```
jill-cardamon, line 424: I didn't send any UserInfo, but I can add that if it's necessary for consistency. It seems like a better idea to check for the authorization higher in the stack if a developer wants to observe this notification.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -367,6 +367,7 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
         showStatus(title: title, spinner: false, duration: 3, animated: true, interactive: false)
     }
     
+
```
ShanMa1991, line 370: I think we can remove the extra white space.

MapboxNavigation/PassiveLocationManager.swift
```diff
@@ -48,7 +48,7 @@ open class PassiveLocationManager: NSObject, MGLLocationManager {
         // CLLocationManager.accuracyAuthorization was introduced in the iOS 14 SDK in Xcode 12, so Xcode 11 doesn’t recognize it.
         guard let accuracyAuthorizationValue = dataSource.systemLocationManager.value(forKey: "accuracyAuthorization") as? Int,
               let accuracyAuthorization = MBNavigationAccuracyAuthorization(rawValue: accuracyAuthorizationValue) else {
-            return MBNavigationAccuracyAuthorization(rawValue: 0)!
+            return .fullAccuracy
```
ShanMa1991, line 51: Because we use the result from the key-value call and check the `.reducedAccuracy` only. Maybe it's not necessary to make the change in the `PassiveLocationManager`

jill-cardamon: The change doesn't affect my code at all. Minh and I noticed the use of forced downcasting and realized that it could be replaced with `.fullAccuracy` to make this return safer.

MapboxCoreNavigation/NavigationService.swift
1ec5: It would be helpful to include user info so that developers don’t need to look around for the location manager to get the authorization status. I’m a bit concerned that the developer’s first reaction would be to create a throwaway location manager, which would be wasteful. If we don’t include any user info, then there’s no reason to define `locationAuthorizationKey`.

MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -139,6 +139,16 @@ public protocol NavigationServiceDelegate: class, UnimplementedLogging {
...
     
+    /**
+     Called when the location manager's accuracy authorization changed.
+     
+     You can implement this method to allow the navigation service to check if the user changed accuracy authorization, especially if reducedAccuracy is enabled. This method is only relevant for iOS 14 and above.
+     
+     - parameter service: The navigation service that will alert that user that reducedAccuracy is enabled.
+     - parameter manager: The location manager.
+     */
+    func navigationServiceDidChangeAuthorization(_ service: NavigationService, manager: CLLocationManager)
```
1ec5, line 150: Nit: Since there’s a second argument, it’s more natural to write `navigationService(_:didChangeAuthorizationFor:)`.

```swift
func navigationService(_ service: NavigationService, didChangeAuthorizationFor locationManager: CLLocationManager) {
```

MapboxNavigation/NavigationViewController.swift
```diff
@@ -687,6 +688,21 @@ extension NavigationViewController: NavigationServiceDelegate {
         return navigationComponents.allSatisfy { $0.navigationServiceShouldDisableBatteryMonitoring(service) }
     }
     
+    public func navigationServiceDidChangeAuthorization(_ service: NavigationService, _ manager: CLLocationManager) {
+        // CLLocationManager.accuracyAuthorization was introduced in the iOS 14 SDK in Xcode 12, so Xcode 11 doesn’t recognize it.
+        guard let accuracyAuthorizationValue = manager.value(forKey: "accuracyAuthorization") as? Int
+        else { return }
```
1ec5, line 694: FYI, it’s common to put this all on one line, or at least to put `else` on the same line as the rest of the `guard` statement.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Fixed an issue where the route line flickered when refreshing. ([#2642](https://github.com/mapbox/mapbox-navigation-ios/pull/2642))
 * Fixed an issue where the End of route view UI is broken prior to iOS 11 . ([#2690](https://github.com/mapbox/mapbox-navigation-ios/pull/2690))
```
1ec5, line 11: Suggested changelog entries:

* Added the `NavigationServiceDelegate.navigationService(_:didChangeAuthorizationFor:)` method and `Notification.Name.locationAuthorizationDidChange` to detect when the user changes the Location Services permissions for the current application, including for approximate location on iOS 14. ([#2693](https://github.com/mapbox/mapbox-navigation-ios/pull/2693))
* When approximate location is enabled on iOS 14, a banner appears reminding the user to disable approximate location to continue navigating. ([#2693](https://github.com/mapbox/mapbox-navigation-ios/pull/2693))

MapboxNavigation/NavigationViewController.swift
```diff
@@ -687,6 +688,21 @@ extension NavigationViewController: NavigationServiceDelegate {
...
+        let accuracyAuthorization = MBNavigationAccuracyAuthorization(rawValue: accuracyAuthorizationValue)
+        
+        if #available(iOS 14.0, *), accuracyAuthorization == .reducedAccuracy {
+            let title = NSLocalizedString("ENABLE_PRECISE_LOCATION", bundle: .mapboxNavigation, value: "Enable precise location to navigate", comment: "Label indicating precise location is off and needs to be turned on to navigate")
+            showStatus(title: title, spinner: false, duration: 20, animated: true, interactive: false)
```
1ec5, line 699: Is there a way to make this status stay up until precise location is enabled, or is that the focus of another issue?

1ec5 (commented):
Almost there!

MapboxCoreNavigation/NavigationService.swift
1ec5: Similar to other code in this PR, we can set `locationAuthorizationKey` to `manager.value(forKey: "accuracyAuthorization")`, without checking the value in any way.

MapboxNavigation/NavigationViewController.swift
jill-cardamon: not sure what happened with formatting but it's technically on the same line

MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -180,6 +180,14 @@ public extension Notification.Name {
      The user info dictionary indicates which keys and values changed.
      */
     static let navigationSettingsDidChange: Notification.Name = .init(rawValue: "NavigationSettingsDidChange")
+    
+    /**
+     Posted when user changes location authorization settings.
+     
+     The user info dictionary indicates which keys and values changed.
```
1ec5, line 187:
```suggestion
     The user info dictionary contains the key `MapboxNavigationService.NotificationUserInfoKey.locationAuthorizationKey`.
```

MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -268,3 +276,20 @@ extension PassiveLocationDataSource {
...
+        public typealias RawValue = String
+        public var rawValue: String
+        public init(rawValue: String) {
+            self.rawValue = rawValue
+        }
+    }
+
+    /**
+     A key in the user info dictionary of a `Notification.Name.locationAuthorizationDidChange` notification. The corresponding value is a `Int` object the current location authorization setting. */
+    public static let locationAuthorizationKey: NotificationUserInfoKey = .init(rawValue: "locationAuthorization")
```
1ec5, line 294: This constant should be part of `MapboxNavigationService.NotificationUserInfoKey`, so that you’d use it as `MapboxNavigationService.NotificationUserInfoKey.locationAuthorizationKey` (or just `NotificationUserInfoKey.locationAuthorizationKey` if inside the implementation of MapboxNavigationService).

MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -268,3 +276,20 @@ extension PassiveLocationDataSource {
...
+    public struct NotificationUserInfoKey: Hashable, Equatable, RawRepresentable {
+        public typealias RawValue = String
+        public var rawValue: String
+        public init(rawValue: String) {
+            self.rawValue = rawValue
+        }
+    }
+
+    /**
+     A key in the user info dictionary of a `Notification.Name.locationAuthorizationDidChange` notification. The corresponding value is a `Int` object the current location authorization setting. */
```
1ec5, line 293: It’s technically an `Int` when we put it in the dictionary, due to our internal workaround, but the developer would be expected to cast it to a `CLAccuracyAuthorization`.

If the developer needs to similarly work around the lack of `CLAccuracyAuthorization` on iOS 13, they’d conditionally implement an enumeration on iOS 13 that would match a typedef on iOS 14. But documenting that workaround is probably outside the scope of the documentation here, since the workaround requires a modicum of Objective-C code that might not be appropriate for every application.

```suggestion
     A key in the user info dictionary of a `Notification.Name.locationAuthorizationDidChange` notification. The corresponding value is a `CLAccuracyAuthorization` indicating the current location authorization setting. */
```

MapboxNavigation/NavigationViewController.swift
1ec5: Before merging, please open an issue to track persisting this banner until precise location is reenabled, so that we don’t lose track of that important detail. #2691 is sort of related, but I think persisting the accuracy authorization banner and automatically hiding it at the right time would be more tractable, because we’re already observing the notification that we need to observe to know when to hide the banner.

1ec5 (approved):
This PR is almost there, but there are a few more things to take care of before merging.

MapboxNavigation/NavigationViewController.swift
1ec5: #2723

jill-cardamon:
Fixed a change (commit hash: 9295eb949904ef79c5fe1b72c45ed40d2e70605b) that I accidentally deleted.

PR #2694, opened 2020-10-14
Route Alerts

Udumft:
Adds Route Alerts processing and deploying it to user.

Implementation is based on entities, provided by `MapboxDirections` and reported by `Navigation Native`.
`Incidents` support is subject to be changed.

This PR depends on [Updating Navigation Native to v22](https://github.com/mapbox/mapbox-navigation-ios/pull/2648) as it needs to further update it to v24. Also, MapboxDirections should be [updated](https://github.com/mapbox/mapbox-directions-swift/pull/466) to reflect `Route Alerts` data.

I had to update Routing Tiles fixture because the old one was apparently quite old and did not contain `admin. iso_3166_1_alpha3` name which became obligatory. Updated tiles take much more disk space so I wonder if this can be reduced somehow...


avi-c:
I have successfully updated Apex to use this branch and the MapboxDirections branch it pulls in. I then used it to find and label all the TollBooth features.

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -144,6 +144,13 @@ open class RouteProgress: Codable {
         return currentLegProgress.upcomingStep ?? upcomingLeg?.steps.first
     }
     
+    /**
+     Upcoming `RouteAlerts` as reported by the navigation engine.
+     
+     Contents of the array depends on user's current progress along the route and are modified on each location update. You can use this to get info about incoming POIs.
+     */
+    public internal(set) var upcomingRouteAlerts: [UpcomingRouteAlertInfo] = []
```
avi-c, line 152: Are we able to provide any information on how far ahead "upcoming" is supposed to be? Is it tunable or related to speed of travel or current step?

Cartfile
```diff
@@ -1,8 +1,8 @@
 binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" ~> 2.3.0
-binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.json" ~> 22.0
+binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.json" ~> 24.0
 binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps/mapbox-ios-sdk-dynamic.json" ~> 6.0
 binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon-ios.json" == 7.1.0
-github "mapbox/mapbox-directions-swift" ~> 1.0
+github "mapbox/mapbox-directions-swift" "18061abb9a185fb5200159eb757c1a81dd7656b6"
```
avi-c, line 5: Will we be moving to a specific version once the directions changes are merged into MapboxDirections?

Cartfile
```diff
@@ -1,8 +1,8 @@
 binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" ~> 2.3.0
-binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.json" ~> 22.0
+binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.json" ~> 24.0
```
avi-c, line 2: Maybe we want v25.0 now?

MapboxCoreNavigation/RouteController.swift
```diff
@@ -336,8 +336,9 @@ open class RouteController: NSObject {
         }
     }
     
-    private func update(progress: RouteProgress, with location: CLLocation, rawLocation: CLLocation) {
+    private func update(progress: RouteProgress, with location: CLLocation, rawLocation: CLLocation, upcomingRouteAlerts routeAlerts: Array<UpcomingRouteAlert>) {
```
avi-c, line 339: Are alerts always present or do we want this to be an optional?

Cartfile
Udumft: Actually, yes, you are right. In v25 NN [renamed Incident type](https://github.com/mapbox/mapbox-navigation-native/pull/1787) JSON key which affects the feature.

Cartfile
Udumft: Correct. This is temporary while Directions didn't release with [corresponding update](https://github.com/mapbox/mapbox-directions-swift/pull/466).

MapboxCoreNavigation/RouteController.swift
Udumft: Array itself is non-nil. It is also important to update it each time since "upcoming" alerts do not include already reported ones, so most of the time this array is reset to empty.

MapboxCoreNavigation/RouteProgress.swift
Udumft: Good question. At least I am not aware of such configurations, and this is decided by NN. I'll investigate

Udumft: So I found the doc should be a bit corrected: `upcomingRouteAlerts` contain all events, that user still didn't pass entirely. In other words, at the beginning of the route this array holds all events from all the route. Event is removed from this array only when user has passed it's `end coordinate`

avi-c:
@MaximAlien @1ec5 if possible it would be nice to get this merged today so we can be closer to Android parity

MaximAlien:
> @MaximAlien @1ec5 if possible it would be nice to get this merged today so we can be closer to Android parity

@avi-c, not sure we can do this just yet. Since this PR is based on Navigation Native update (which currently fails when running CocoaPods linter on Xcode 12) I think it'd make sense to wait for a fix in Mapbox Common and Navigation Native. I've created ticket here: https://github.com/mapbox/mapbox-sdk-common/issues/570. As I also can see this PR updates to Navigation Native 24.0.0, it'll have to contain same fix as well.

I can also see that there are several compilation errors like:
```
/Users/distiller/project/MapboxCoreNavigation/BorderCrossingInfo.swift:6:11: cannot find type 'AdministrationRegion' in scope
```

Udumft:
About compile error mentioned above: this is related to `mapbox-directions` not yet released [required changes](https://github.com/mapbox/mapbox-directions-swift/pull/466), so cocoapods is unable to checkout correct source code.

avi-c:
@MaximAlien As @Udumft points out above, getting a review on https://github.com/mapbox/mapbox-directions-swift/pull/466 and getting it merged would be great to unblock this one.

avi-c:
I just merged https://github.com/mapbox/mapbox-directions-swift/pull/466 so the Intersection attributes are present now.

1ec5:
> I had to update Routing Tiles fixture because the old one was apparently quite old and did not contain `admin. iso_3166_1_alpha3` name which became obligatory. Updated tiles take much more disk space so I wonder if this can be reduced somehow...

It turns out this is the root cause of mapbox/mapbox-directions-swift#491. mapbox/mapbox-directions-swift#492 works around the issue by making the alpha-3 property optional. Per mapbox/mapbox-directions-swift#477, Swift code on Apple platfoms is unlikely to need the alpha-3 code anyways.

MaximAlien (commented):
It's been a while since this PR was updated. I think good idea would be to rebase onto `main` to get latest changes.

MapboxCoreNavigationTests/CocoaPodsTest/PodInstall/Podfile.lock
```diff
@@ -19,8 +19,8 @@ PODS:
     - MapboxMobileEvents (~> 0.10.2)
     - MapboxSpeech (~> 1.0)
     - Solar (~> 2.1)
-  - MapboxNavigationNative (22.0.5):
-    - MapboxCommon (= 7.1.2)
+  - MapboxNavigationNative (25.1.0):
+    - MapboxCommon (= 7.1.1)
```
Udumft, line 23: Oddly, NN v25 uses older `MapboxCommon` version than v22. Could this be a reason for CI Pod check? Verbose output shows an issue with linking `MapboxCommon` slices:
```
-> MapboxCoreNavigation (1.2.0-alpha.1)
    - ERROR | [iOS] xcodebuild: Returned an unsuccessful exit code.
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | xcodebuild:  note: Building targets in parallel
    - NOTE  | xcodebuild:  note: Using codesigning identity override: -
    - NOTE  | [iOS] xcodebuild:  note: Planning build
    - NOTE  | [iOS] xcodebuild:  note: Constructing build description
    - NOTE  | [iOS] xcodebuild:  warning: Skipping code signing because the target does not have an Info.plist file and one is not being generated automatically. (in target 'App' from project 'App')
    - NOTE  | [iOS] xcodebuild:  warning: [CP] Unable to find matching .xcframework slice in 'MapboxCommon/MapboxCommon.xcframework MapboxCommon framework ios-i386_x86_64-simulator ios-armv7_arm64 ios-x86_64-maccatalyst' for the current build architectures (arm64 x86_64 i386).
    - NOTE  | xcodebuild:  ld: warning: directory not found for option '-F/Users/viktorkononov/Library/Developer/Xcode/DerivedData/App-dxvmejvrbgstrehborjnnbpvaujh/Build/Products/Release-iphonesimulator/XCFrameworkIntermediates/MapboxCommon'
    - NOTE  | xcodebuild:  clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Udumft:
I had to update Nav Native to version 26.3 due to some dependencies mismatch: NN v25 (where RouteAlerts feature originates) used `MapboxCommons` v7.1.1 which causes pods issues. According to the schedule, we were to update it anyways for NavSDK v1.2

MapboxCoreNavigation/BorderCrossingInfo.swift
```diff
@@ -0,0 +1,23 @@
...
+        self.init(countryCode: adminInfo.iso_3166_1, countryCodeAlpha3: adminInfo.iso_3166_1_alpha3)
+    }
+}
+
+/**
+ `BorderCrossingInfo` encapsulates a border crossing, specifying crossing region codes.
+ */
+public struct BorderCrossingInfo {
+    public let from: AdministrationRegion
+    public let to: AdministrationRegion
```
1ec5, line 17: Is it expected that there will be more information about the border crossing, or could it be more efficiently represented as a tuple?

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -144,6 +144,13 @@ open class RouteProgress: Codable {
         return currentLegProgress.upcomingStep ?? upcomingLeg?.steps.first
     }
     
+    /**
+     Upcoming `RouteAlerts` as reported by the navigation engine.
+     
+     Contents of the array depends on user's current progress along the route and are modified on each location update. Array always contains only events, that user stil not passed. Keep in mind that some events may have non-zero length and are also incleded while user is mid-traversing it. You can use this to get info about incoming POIs.
```
1ec5, line 150:
```suggestion
     The contents of the array depend on user's current progress along the route and are modified on each location update. This array contains only the alerts that the user has not passed. Some events may have non-zero length and are also included while the user is traversing it. You can use this property to get information about incoming points of interest.
```

MapboxCoreNavigation/TunnelInfo.swift
```diff
@@ -0,0 +1,14 @@
+
+import Foundation
+import MapboxNavigationNative
+
+/**
+ `TunnelInfo` is used for naming incoming tunnels, together with route alerts.
+ */
+public struct TunnelInfo {
+    public let name: String
```
1ec5, line 9: Since this struct stores only a name, do we need the struct, or can we inline the name?

If we do expect to extend this type with more information in the future, then consider calling it Tunnel. (Compare to MapboxDirections’ [Road](https://github.com/mapbox/mapbox-directions-swift/blob/4184df3c07694f75738a99728cdfa5504c8e4632/Sources/MapboxDirections/RouteStep.swift#L307) struct.)

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
```diff
@@ -0,0 +1,84 @@
...
+ `UpcomingRouteAlertInfo` encapsulates information about various incoming events. Common attributes like location, distance to the event, length and other is provided for each POI, while specific meta data is supplied via `alert` property.
+ */
+public struct UpcomingRouteAlertInfo {
+    public enum RouteAlertType {
+        case incident(Incident)
+        case tunnel(TunnelInfo)
+        case borderCrossing(BorderCrossingInfo)
+        case tollCollection(TollCollection)
+        case serviceArea(RestStop)
+    }
```
1ec5, line 16: This enumeration indicates more than just the type, so we should rename it to just `RouteAlert`. There will probably be a conflict with MapboxNavigationNative, but we can solve that conflict by selective imports and namespaces. In the worst-case scenario, a separate file could import MapboxNavigationNative and declare `InternalRouteAlert` as a type alias of `MapboxNavigationNative.RouteAlert`.

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
```diff
@@ -0,0 +1,84 @@
+import Foundation
+import CoreLocation
+import MapboxNavigationNative
+import MapboxDirections
+
+/**
+ `UpcomingRouteAlertInfo` encapsulates information about various incoming events. Common attributes like location, distance to the event, length and other is provided for each POI, while specific meta data is supplied via `alert` property.
+ */
+public struct UpcomingRouteAlertInfo {
```
1ec5, line 9: Is “upcoming” the right terminology to use? Is there a situation where the developer might end up referring to a past UpcomingRouteAlert? This is sort of like a “positioned” route alert in that it combines the alert itself with positioning information.

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
```diff
@@ -0,0 +1,84 @@
...
+        switch upcomingAlert.alert.type {
+        case .kIncident:
+            guard let incidentInfo = upcomingAlert.alert.incidentInfo else {
+                return nil
+            }
+            self.alert = .incident(Incident(incidentInfo))
+        case .kTunnelEntrance:
+            guard let tunnelInfo = upcomingAlert.alert.tunnelInfo else {
+                return nil
+            }
```
1ec5, line 63: Is this initializer failable because the alert may come back with no metadata? Or are these guard statements meant to guard against internal inconsistency? If the latter, perhaps these code paths should be `preconditionFailure`s.

MapboxCoreNavigation/IncidentInfo.swift
```diff
@@ -0,0 +1,22 @@
...
+                  creationTime: incidentInfo.creationTime?.ISO8601 ?? "",
+                  startTime: incidentInfo.startTime?.ISO8601 ?? "",
+                  endTime: incidentInfo.endTime?.ISO8601 ?? "",
+                  impact: incidentInfo.impact ?? "",
+                  subtype: incidentInfo.subType,
+                  subtypeDescription: incidentInfo.subTypeDescription,
+                  alertCodes: incidentInfo.alertcCodes.map { $0.intValue },
+                  lanesBlocked: incidentInfo.lanesBlocked.map { Int($0) ?? -1 },
+                  geometryIndexStart: -1,
+                  geometryIndexEnd: -1)
```
1ec5, line 20: mapbox/mapbox-directions-swift#482 will simplify this initializer a bit.

1ec5 (commented):
Also remember to update the changelog to mention the added types and properties, as well as the upgraded MapboxDirections and MapboxNavigationNative dependencies.

MapboxCoreNavigation/BorderCrossingInfo.swift
Udumft: To be honest: I don't know. At least now having it as `struct` is justified by introducing a convenience initializer which makes it easier to use than a `tuple`.

MapboxCoreNavigation/TunnelInfo.swift
Udumft: Situation here is same as for `BorderCrossingInfo`. I wanted to have similar configuration for each Route Alert possible. IMHO if we have some mix of `struct`s, `tuple`s and inlined values that would look inconsistent. That's why I decided to stick with `struct`s everywhere because this is the most universal approach and it allows extending functionality and data without breaking backwards compatibility.

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
Udumft: `UpcomingRouteAlertInfo` are provided to a user on location updates and the feed contains only not passed events. So for the context of how SDK supplies and uses it, "upcoming" seems correct.

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
Udumft: Hm... Originally I decided to make this one failable to cover cases when NN submits unknown alert type, so that SDK could just omit it. But it seems that such cases are impossible because NN version will always be in sync with SDK version, so having `default` case here might even become a potential point to forget add support for new alert types.
About optional alert info structures - these should not be nil if alert type is correct, so I think yes, that might be considered as an error.

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
1ec5: I think it’s important for the name of a public type to be intrinsic to the data contained inside of it; to some extent, the name should be independent of the situations in which the type is or isn’t used. It would be awkward if the application keeps its own array of passed “upcoming” alerts. (For a contrived example, consider a counter of how many “upcoming” alerts the user has already passed.)

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
1ec5: Ideally we’d add default unknown cases, but unfortunately, MapboxNavigationNative incorrectly declares all its structs as frozen.

Udumft: I thought that having a `default` case might create a room for an error once any new alert type is added to NN. If such happens now - compiler will alarm.

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
```diff
@@ -0,0 +1,85 @@
...
+    init(_ upcomingAlert: UpcomingRouteAlert) {
+        self.distance = upcomingAlert.alert.distance
+        self.distanceToStart = upcomingAlert.distanceToStart
+        self.length = upcomingAlert.alert.length?.doubleValue
+        self.beginCoordinate = upcomingAlert.alert.beginCoordinate
+        self.endCoordinate = upcomingAlert.alert.endCoordinate
+        self.beginSegmentIndex = upcomingAlert.alert.beginGeometryIndex
+        self.endSegmentIndex = upcomingAlert.alert.endGeometryIndex
+        
+        switch upcomingAlert.alert.type {
```
1ec5, line 55: Is it possible to add a default or default unknown case even though the compiler thinks that the enumeration is frozen and the switch statement is exhaustive?

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
1ec5: Will it reliably produce an error if MapboxNavigationNative enables module stability?

MapboxCoreNavigation/UpcomingRouteAlertInfo.swift
Udumft: Adding `@unknown default` case triggers a warning since all cases are already handled and `RouteTypeAlert` is frozen ATM. If any issues arise with module stability and bindgen products - it will affect not just this one switch, and more general approach would be required.

CHANGELOG.md
```diff
@@ -18,6 +22,7 @@
 * Added Ukrainian localization. ([#2735](https://github.com/mapbox/mapbox-navigation-ios/pull/2735))
 * Created the `UserHaloCourseView` similar to `UserCourseView` for approximate location on iOS 14 during the navigation to represent user location. Allow the switch between `UserHaloCourseView` and `UserCourseView` when precise mode is changed. ([#2664](https://github.com/mapbox/mapbox-navigation-ios/pull/2664))
 * Added the ability to provide more detailed `Speed limit incorrect` feedback during turn-by-turn navigation. ([#2725](https://github.com/mapbox/mapbox-navigation-ios/pull/2725))
+* Added support for `Route Alerts` by exposing `RouteProgress.upcomingRouteAlerts` property. Use `UpcomingRouteAlertInfo` with `alert` with one of the following types: `Incident`, `TunnelInfo`, `BorderCrossingInfo`, `TollCollection`, `RestStop` ([#2694](https://github.com/mapbox/mapbox-navigation-ios/pull/2694)
```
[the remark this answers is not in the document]
Udumft: Done

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## v1.2.0
 
+### Packaging
+
+* `MapboxNavigationNative` version bumped to v26.3, `MapboxCommon` updated to v9.1.0 and `MapboxDirections.swift` to v1.2  ([#2694](https://github.com/mapbox/mapbox-navigation-ios/pull/2694)
```
1ec5, line 7:
```suggestion
* Increased the minimum versions of `MapboxNavigationNative` to v26.3, `MapboxCommon` to v9.1.0 and `MapboxDirections` to v1.2. ([#2694](https://github.com/mapbox/mapbox-navigation-ios/pull/2694))
```

MapboxCoreNavigation/RouteAlert.swift
```diff
@@ -0,0 +1,88 @@
...
+ */
+public struct RouteAlert {
+    public enum Alert {
+        case incident(Incident)
+        case tunnel(Tunnel)
+        case borderCrossing(BorderCrossing)
+        case tollCollection(TollCollection)
+        case serviceArea(RestStop)
+        case restrictedArea
+    }
```
1ec5, line 17: This enumeration and its cases should be documented, since the public `alert` property uses them. It can be just a perfunctory documentation comment, such as “Incident with the given incident details” for `incident`.

1ec5 (approved):
I’ve pushed MapboxDirections v1.2.0-alpha.3 to CocoaPods trunk, so the CocoaPods integration tests should pass after rerunning. Thanks for your persistence on this feature!

CHANGELOG.md
Udumft: Changed

MapboxCoreNavigation/RouteAlert.swift
Udumft: Doc comments added

PR #2704, opened 2020-10-21
Offline wrapper

IodaMikeMapbox:
Resolves mapbox/navigation-sdks#638

The pr is blocked by #2653 

Technically it's not directly blocked by the #2653, we can merge it without adopting the usage logic, but I don't believe that it can/will be properly tested/reviewed without:
 - comparison with the same functionality from the CommonSdk;
 - api usage example, because there are too much context for a reviewer to see all or most of the subtle parts of the api.


OfflineService/OfflineServiceViewController.swift
```diff
@@ -42,9 +39,11 @@ class OfflineServiceViewController: UITableViewController, OfflineServiceDataSou
     }
     
     private func setupOfflineServiceDataSource() {
-        offlineServiceDataSource = OfflineServiceDataSource()
-        offlineServiceDataSource?.delegate = self
-        offlineServiceDataSource?.startObservingAvailableRegions()
+        OfflineService.shared.register(observer: self)
```
MaximAlien, line 42: It should be possible to configure `OfflineService` with values provided by end user. Currently it's possible to configure either `OfflineService` or `OfflineServiceUser` due to protection level errors.

MapboxCoreNavigation/OfflineService.swift
```diff
@@ -0,0 +1,379 @@
...
+        DispatchQueue.main.async { [weak self] in
+            guard let self = self else { return }
+            let region = self._regions[metadata.id] ?? OfflineRegion(region: metadata)
+            self.observers.forEach {
+                $0.didBeginVerifying(region: region, forDomain: .common(domain: domain))
+            }
+        }
+    }
+
+    public func onAvailable(for domain: OfflineDataDomain, metadata: OfflineDataRegionMetadata, pack: OfflineDataPack) {
```
MaximAlien, line 266: In case of offline mode already downloaded regions are not shown, can it be addressed to match previous implementation?

OfflineService/OfflineServiceViewController.swift
```diff
@@ -160,51 +156,73 @@ class OfflineServiceViewController: UITableViewController, OfflineServiceDataSou
...
-        let offlineDataRegion = offlineDataItems[indexPath.row]
-        showActions(offlineDataRegion)
+        let offlineRegion = offlineRegions[indexPath.row]
+        showActions(offlineRegion)
     }
     
     // MARK: - Private methods
 
-    private func showActions(_ offlineDataRegion: OfflineDataItem) {
+    private func showActions(_ offlineRegion: OfflineRegion) {
```
MaximAlien, line 165: After introducing new logic it's no longer possible to correctly remove region in some cases: e.g. action which allows to delete region appears before fully downloading region, region also remains as downloaded. Can you check previous behavior to mimic it?

MapboxCoreNavigation/OfflineService.swift
```diff
@@ -0,0 +1,379 @@
...
+            }
+        }
+    }
+}
+
+extension OfflineService: MapboxCommon.OfflineServiceObserver {
+    public func onPending(for domain: OfflineDataDomain, metadata: OfflineDataRegionMetadata, pack: OfflineDataPack) {
+        DispatchQueue.main.async { [weak self] in
+            guard let self = self else { return }
+            let region = self._regions[metadata.id] ?? OfflineRegion(region: metadata)
```
MaximAlien, line 221: Is `metadata.id` really reliable way of using it as key in a dictionary? As I can see it contains names like `austin_day` which can be controlled by user during creation, this will in turn will lead to clash in regions which have the same id, but different content. I'd perform more testing with different ids to make sure.

OfflineService/OfflineDataRegionTableViewCell.swift
```diff
@@ -1,5 +1,5 @@
 import UIKit
-import MapboxCommon
+import MapboxCoreNavigation
 
 class OfflineDataRegionTableViewCell: UITableViewCell {
```
MaximAlien, line 4: In current implementation both last updated and pack size values are not visible when they're not downloaded, can we show them to match Android functionality. I think it's also important to show last updated value because it's used as tilesVersion for offline navigation.

MaximAlien:
Would it be possible to cover wrapper APIs with basic tests?

OfflineService/OfflineDataRegionTableViewCell.swift
```diff
@@ -56,43 +56,63 @@ class OfflineDataRegionTableViewCell: UITableViewCell {
     }
     
     // MARK: - Public UI update related methods
-    
-    func showDownloadProgress(for domain: OfflineDataDomain, dataPack: OfflineDataPack, metadata: OfflineDataRegionMetadata) {
+
+    func showDownloadProgress(for domain: OfflineRegionDomain, dataPack: OfflineRegionPack, region: OfflineRegion) {
         DispatchQueue.main.async {
```
MaximAlien, line 61: I think after changes which were introduced in `OfflineService` `onDownloading(for:metadata:pack:)` is already calling `didStartDownloading(region:forDomain:)` on main thread, so there is no need to perform double dispatch.

MapboxCoreNavigation/OfflineService.swift
```diff
@@ -0,0 +1,379 @@
+import Foundation
+import MapboxCommon
+
+private typealias CommonOfflineService = MapboxCommon.OfflineService
+
+public struct OfflineServiceUser {
+    private enum Constants {
```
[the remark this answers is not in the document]
IodaMikeMapbox: agree

MapboxCoreNavigation/OfflineRegion.swift
```diff
@@ -0,0 +1,172 @@
+import Foundation
+import MapboxCommon
+
+public enum OfflineRegionStatus {
```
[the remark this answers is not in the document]
IodaMikeMapbox: oh, I forgot this part

MapboxCoreNavigation/OfflineService.swift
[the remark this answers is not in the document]
IodaMikeMapbox: added comments, removed the constants

MapboxCoreNavigation/OfflineService.swift
```diff
@@ -0,0 +1,379 @@
...
+        static let username = "1tap-nav"
+        static let baseURL = "https://api.mapbox.com"
```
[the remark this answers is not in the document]
IodaMikeMapbox: removed

MapboxCoreNavigation/OfflineRegion.swift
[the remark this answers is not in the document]
IodaMikeMapbox: used as a pack's status

OfflineService/OfflineServiceViewController.swift
```diff
@@ -42,9 +39,11 @@ class OfflineServiceViewController: UITableViewController, OfflineServiceDataSou
     }
     
     private func setupOfflineServiceDataSource() {
```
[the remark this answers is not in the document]
IodaMikeMapbox: renamed

OfflineService/OfflineServiceViewController.swift
IodaMikeMapbox: added method for configuring the user

MapboxCoreNavigation/OfflineService.swift
IodaMikeMapbox: can't reproduce, could you provide more info about the case? It do show downloaded regions without an internet connection

OfflineService/OfflineServiceViewController.swift
IodaMikeMapbox: fixed

MapboxCoreNavigation/OfflineService.swift
IodaMikeMapbox: @kkaefer are the ids not unique for the regions?

OfflineService/OfflineDataRegionTableViewCell.swift
IodaMikeMapbox: fixes

IodaMikeMapbox:
> Would it be possible to cover wrapper APIs with basic tests?

I think about it, there are some thought, but the primary logic can't be tested without an internet connection

MapboxCoreNavigation/OfflineRegion.swift
```diff
@@ -0,0 +1,159 @@
...
+     The total number of bytes of this offline pack
+     */
+    public var totalBytes: UInt64? {
+        commonPackMetadata?.bytes
+    }
+
+    /**
+     The URL the offline pack can be downloaded from.
+     */
+    public var url: String? {
```
1ec5, line 63: This property should be of type URL.

MapboxCoreNavigation/OfflineRegion.swift
```diff
@@ -0,0 +1,159 @@
...
+     When this region was last updated
+     */
+    public var lastUpdated: Date {
+        region.last_updated
+    }
+
+    /**
+     A GeoJSON geometry definition describing the extent of this region
+     */
+    public var geography: MBXGeometry {
```
1ec5, line 134: MBXGeometry is a generated interface for a GeoJSON object. We should convert it to a Turf GeoJSON type.

MapboxCoreNavigation/OfflineRegionStatus.swift
```diff
@@ -0,0 +1,159 @@
...
+ *     (failed)─────│ Verifying │─────delete()──────┘   │
+ *                  └───────────┘                       │
+ *                        │                         (succeeds)
+ *                   (successful)                       │
+ *                        │                             │
+ *                  ┌─────▼─────┐                  ┌────┴─────┐
+ *                  │ Available ├───────delete()───▶ Deleting │
+ *                  └─────▲─────┘                  └────┬─────┘
+ *                        │                             │
+ *                        └─────────download()──────────┘
```
1ec5, line 39: Can you check if this ASCII art appears correct in jazzy docs? I think we might need to wrap it in some markup to keep it as preformatted text.

MapboxCoreNavigation/OfflineRegion.swift
```diff
@@ -0,0 +1,159 @@
...
+/**
+ A metadata record for an offline region.
+
+ The key for looking up offline regions should be a compound of id and revision.
+ */
+public class OfflineRegion: Equatable {
+    /**
+     A machine-readable identifier
+     */
+    public var id: String {
```
1ec5, line 99: Swift API design guidelines call for expanding this term as “identifier”.

MapboxCoreNavigation/OfflineRegion.swift
```diff
@@ -0,0 +1,159 @@
...
+     The URL the offline pack can be downloaded from.
+     */
+    public var url: String? {
+        commonPackMetadata?.url
+    }
+
+    /**
+     A numeric version of the format. An SDK version only supports one format version and uses this field to verify compatibility.
+     */
+    public var format: UInt32? {
```
1ec5, line 70: Is there any pattern to the format? Should there be an enumeration?

IodaMikeMapbox:
Plans changed, we don't need to launch the offline service api

PR #2705, opened 2020-10-21
Allow to customize the bottom banner view height 

### LINKED ISSUE #2561: Map info button moves after tapping follow mode (?) button in top right corner

### LINKED ISSUE #2698: Navigation bottom banner height is fixed and makes impossible to customise the bottom banner height
**Mapbox Navigation SDK version:** 1.0.0

As explained in the [Custom Banner Example](https://docs.mapbox.com/ios/navigation/examples/custom-banner/), I have a custom Navigation Bottom Banner that I pass to the Mapbox Navigation Options to replace the stock Mapbox bottom banner.
This custom banner has its own height that is set with AutoLayout - no other constraints are set on this view.
That used to work great in Mapbox 0.40.0, but in 1.0.0 the bottom banner height is fixed, even when the bottom banner is passed from outside, as in my case.

I was able to track down the responsible to the fix #2613 .
The fix should apply the height constraints just if the bottom banner view is the internal auto-generated view.
I can work on a patch if you like.

ShanMa1991:
This PR adds ability to allow customers to build their own navigation map with the customized bottom banner view height, and fix the height increase of bottom banner during navigation overview mode.

- [x] remove the `reinstallRequiredConstraints` to allow customize the bottom banner height
- [x] fix the attribution button move from issue #2561 
- [x] update the [custom-banne-tutorial](https://docs.mapbox.com/ios/navigation/examples/custom-banner/) in the way to change the bottom banner height

For example, in the [custom-banne-tutorial](https://docs.mapbox.com/ios/navigation/examples/custom-banner/), in the `CustomBottomBarViewController`, customers could change the `bottomBanner` height in as follow:
`bottomBanner.view.heightAnchor.constraint(equalTo: some constant customer want to set).isActive = true` instead of setting `bottomBanner.bannerView.heightAnchor`

The problem is caused by the `attributionButton` and `logoView` with a default position defined by the MGLMapView, which is set in the bottom of the safe area. But when the overview mode is on during navigation, the navigation map view is concentrated into a small area, which change the location of the `attributionButton` and `logoView`.

By getting the bottom banner height before the overview, and re-setting the bottom banner and  the `attributionButton` layout constraints after the `contentInset` changed, it would allow the `attributionButton` position dynamically fixed related with the bottom banner.  By implementing this pr, the fixes over view mode and allows the bottom banner height changing is as follow:

overview

fix #2698 
fix #2561 
/cc @mapbox/navigation-ios @zugaldia

### DIFF
--- a/MapboxNavigation/NavigationViewLayout.swift
+++ b/MapboxNavigation/NavigationViewLayout.swift
@@ -38,9 +38,6 @@ extension NavigationView {
     }
     
     public func reinstallRequiredConstraints() {
-        if let bottomBannerContainerHeightConstraint = bottomBannerContainerHeightConstraint {
-            bottomBannerContainerView.removeConstraint(bottomBannerContainerHeightConstraint)
-        }
         
         var height: CGFloat = 100.0
         
@@ -54,6 +51,7 @@ extension NavigationView {
             height = 80.0
         }
         
+        var bottomBannerContainerHeightConstraint: NSLayoutConstraint?
         bottomBannerContainerHeightConstraint = bottomBannerContainerView.heightAnchor.constraint(equalToConstant: height)
         bottomBannerContainerHeightConstraint?.isActive = true
     }
--- a/MapboxNavigation/NavigationViewLayout.swift
+++ b/MapboxNavigation/NavigationViewLayout.swift
@@ -6,12 +6,7 @@ extension NavigationView {
         mapView.leadingAnchor.constraint(equalTo: leadingAnchor).isActive = true
         mapView.bottomAnchor.constraint(equalTo: bottomAnchor).isActive = true
         mapView.trailingAnchor.constraint(equalTo: trailingAnchor).isActive = true
-        
-        mapView.logoView.bottomAnchor.constraint(equalTo: bottomBannerContainerView.topAnchor, constant: -10).isActive = true
-        mapView.logoView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: 10).isActive = true
-        mapView.attributionButton.bottomAnchor.constraint(equalTo: bottomBannerContainerView.topAnchor, constant: -10).isActive = true
-        mapView.attributionButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -10).isActive = true
-        
+  
         topBannerContainerView.leadingAnchor.constraint(equalTo: leadingAnchor).isActive = true
         topBannerContainerView.trailingAnchor.constraint(equalTo: trailingAnchor).isActive = true
         topBannerContainerView.topAnchor.constraint(equalTo: topAnchor).isActive = true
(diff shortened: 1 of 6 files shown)


MapboxNavigation/NavigationViewLayout.swift
```diff
@@ -38,9 +38,6 @@ extension NavigationView {
     }
     
     public func reinstallRequiredConstraints() {
-        if let bottomBannerContainerHeightConstraint = bottomBannerContainerHeightConstraint {
```
MaximAlien, line 41: I'm not sure this fix allows to resolve current issue. Even though you remove `bottomBannerContainerHeightConstraint` default `height` values are still used (can be checked when rotating device).

I think good idea would be to create two projects:
1. CocoaPods based project which uses Navigation SDK `0.40.0` which uses `CustomBottomBarViewController. You can then try to change `bannerView` height there by calling:
```swift
bottomBanner.bannerView.heightAnchor.constraint(equalToConstant: 10.0).isActive = true
```

2. CocoaPods based project which uses Navigation SDK `1.0.0` which uses `CustomBottomBarViewController`. You can apply the same change in height and see if it's was applied correctly (as it's stated in original issue).

Next step I think would be to debug how fix in https://github.com/mapbox/mapbox-navigation-ios/pull/2613 breaks the ability to control bottom bar height: try to remove `reinstallRequiredConstraints` completely and apply new fix without hardcoding default height values for bottom bar, but at the same time without causing regression in https://github.com/mapbox/mapbox-navigation-ios/issues/2561. 

It might also make sense to create test sample which uses only `MBLMapView` with visible `logoView` and `attributionButton` and plain UIView below them to understand how layout behaves when those `MBLMapView` views are constrained to bottom view.

MaximAlien: I also remember that when applying fix for #2613 I thought that it'd be enough to set https://github.com/mapbox/mapbox-navigation-ios/blob/f95bab2a5fabdf55a05e960fdbf5e01ff2c6bca1/MapboxNavigation/NavigationViewLayout.swift#L10-L13 to fix the issue, but bottom bar height started increasing simultaneously. I think good idea would be to check possible reasons of this as well.

ShanMa1991: Yes. The bottom banner view hight could be adjusted in both Navigation SDK `0.40.0` and `1.0.0` by the 
`bottomBanner.bannerView.heightAnchor.constraint(equalToConstant: 10.0).isActive = true`
But it will cause the overlapping of the `logoView` and the `attributionButton`, and caused the break of the above constraint. 

Because the `CustomBottomBannerView` is inside the `MBBannerContainerView`. The top and bottom of the `MBBannerContainerView` is AutoLayout set with the imageView, button and the navigationView. So it's safer the change the `CustomBottomBannerView` height by setting top and bottom anchor responding to the `MBBannerContainerView`, which is
 `bannerView.topAnchor.constraint(equalTo: view.topAnchor, constant: 10.0)).isActive = true` 
and the 
`bannerView.bottomAnchor.constraint(equalTo: safeArea.bottomAnchor, constant: -10.0).isActive = true` 
so whole `MBBannerContainerView` could still  follow the view height change in the `reinstallRequiredConstraints`. And the the `CustomBottomBannerView` only change it's height inside the `MBBannerContainerView`. It would not cause the problem in the `logoView` or the `attributionButton` overlapping either.

ShanMa1991: Yes, the nextstep is to rewrite the `reinstallRequiredConstraints` , to replace the height constant with the auto layout constraints. So the user could edit the `superview.height` in `CustomBottomBannerView`, to adjust the whole `BottomBannerContainerView`'s height

v1-m (approved):

MaximAlien:
@ShanMa1991, thanks for your work on this. I did some tests and bottom banner height can be changed. There are no visible issues in `MGMapView` components either. Since root cause of second issue is on Mapbox Maps SDK side issue was created here: https://github.com/mapbox/mapbox-gl-native-ios/issues/533.

MaximAlien (approved):

PR #2710, opened 2020-10-26
Fix speed multiplier

### LINKED ISSUE #2636: Incorrect speed multiplier value after arriving to intermediate waypoint.
After arriving to intermediate waypoint when using location simulation with speed multiplier higher than `1` it'll be reset to `1` (even though `SimulatedLocationManager` still uses increased value). It is expected that UI always shows most up-to-date `speedMultiplier` value.

Example:

output

IodaMikeMapbox:
Resolves #2636


MaximAlien (approved):
Overall looks good. Since this fix is required for `1.1.0` I've updated changelog and merged it to `main`.

PR #2712, opened 2020-10-26
Revert changes which were added to support Offline Service.

MaximAlien:
This PR reverts changes made in https://github.com/mapbox/mapbox-navigation-ios/pull/2653 to remove Offline Service functionality. Some changes were left intact as I think that they'd be beneficial in `1.1.0`. Changelog was also slightly updated to contain fixes from `1.0.1` in `1.1.0`.


Example/ResizableView.swift
```diff
@@ -0,0 +1,129 @@
+import UIKit
+
+class ResizableView: UIControl {
```
MaximAlien, line 3: Not sure there is any sense in keeping `ResizableView.swift`, `SettingsItems.swift`, `SettingsViewController.swift` at this point.

IodaMikeMapbox (approved):

PR #2714, opened 2020-10-28
Fix wrong format specifier of speed multiplier in simulation (Hebrew)

pasovsky:
**Mapbox Navigation SDK version:** `v1.0.0-rc.5`

When in simulation mode, a small status banner indicating the simulation speed multiplier is displayed.

In Hebrew, an integer format specifier was used for the multiplier but a string is being passed to the formatter.
So instead of showing "Simulating navigation at 1x" (in Hebrew)
it would show "Simulating navigation at -1,103,473,346x"

The code for passing the string to the localized string is `showSimulationStatus(speed: Int)` in `StatusView.swift`

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxNavigation/Resources/he.lproj/Localizable.strings
+++ b/MapboxNavigation/Resources/he.lproj/Localizable.strings
@@ -47,7 +47,7 @@
 "RESUME" = "חזור";
 
 /* The text of a banner that appears during turn-by-turn navigation when route simulation is enabled. */
-"USER_IN_SIMULATION_MODE" = "הדמה ניווט ב %dx";
+"USER_IN_SIMULATION_MODE" = "הדמה ניווט ב %@x";
 
 /* Format for displaying destination and intermediate waypoints; 1 = source ; 2 = destinations */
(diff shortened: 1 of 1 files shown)


1ec5 (approved):
Thanks! Somehow I missed this translation and the Simplified Chinese one in #2136. We manage translations [at Transifex](https://www.transifex.com/mapbox/mapbox-navigation-ios/), so I’ve updated the translations based on your fix there. Since you haven’t given maintainers of this repository permission to push to your fork’s PR branch, I’ll follow up with the updates to the Simplified Chinese localization and changelog in a separate commit.

1ec5:
The builds appear to be failing because the third-party PR doesn’t have permission to use the secret token in .netrc for downloading the map SDK. I’ll merge regardless.

1ec5:
Updated the Simplified Chinese localization and changelog in 600bf7323b81095dd909ab7eb23fe39ea3faa954. Thanks again!

pasovsky:
@1ec5 Thanks for taking care of the PR!

Issue #2678, opened 2020-10-05
Navigation with waypoints, can't change leg color anymore in 1.0.0


abouroubi:
**Mapbox Navigation SDK version: 1.0.0**

I was using the SDK version 1.0.0-alpha1, I'm starting navigation with 3 waypoints each time, with the 1.0.0-alpha1, if I wanted to change the color of the current leg, and diffenrentiate it to otehr legs I used this code, and it worked:
```swift
self.mapView?.routeCasingColor = #colorLiteral(red: 0.337254902, green: 0.6588235294, blue: 0.9843137255, alpha: 0.1)
self.mapView?.trafficUnknownColor  = #colorLiteral(red: 0.3384324702, green: 0.6636026477, blue: 0.9891712385, alpha: 0.2964898768)
```
The first line controled the color of the remaining legs, and the second line was the color of the current active leg (where the user is).

After updating to 1.0.0, this doesn't work anymore, the whole route have the second color.

1ec5:
Possibly related: #2595.

1ec5:
As discussed in https://github.com/mapbox/mapbox-navigation-ios/pull/2377#discussion_r445175661, the new `NavigationViewController.routeLineTracksTraversal` property’s implementation seems to have required removing the styling of inactive legs as translucent lines.

abouroubi:
Sorry but I don't really understand, you decided that `NavigationViewController.routeLineTracksTraversal` is a more important feature than the customization of the inactive legs so you removed it ?

1ec5:
No, it seems to have been an oversight. We’re tracking this issue as a regression. In fact, we were planning on introducing an explicit option to give different routes and legs different colors in a future minor release, something more explicit than how `routeCasingColor` and `trafficUnknownColor` happened to combine to produce that effect in previous releases. But fixing the regression is important.

Once #2623 lands, there will be a workaround in the form of specifying a different polyline feature for each leg, but until then, I don’t know of a straightforward workaround.

abouroubi:
I get it, thank you

abouroubi:
Hi @1ec5 ,
Can you please help me with this issue ? I'm still trying to figure out how I can change the color of the current leg, but I couldn't find a solution.

Some help will be appreciated

1ec5:
[...]

PR #2719, opened 2020-10-30
Allow styling each route differently

MaximAlien:
Initial implementation of route line styling functionality. Example:

output

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -75,21 +75,11 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         
         static let instruction = "\(identifierNamespace).instruction"
         
-        static let mainRouteCasing = "\(identifierNamespace).mainRouteCasing"
-        
         static let buildingExtrusion = "\(identifierNamespace).buildingExtrusion"
     }
     
     struct StyleLayerIdentifier {
         static let namespace = Bundle.mapboxNavigation.bundleIdentifier ?? ""
-
-        static let mainRoute = "\(identifierNamespace).mainRoute"
-        static let mainRouteCasing = "\(identifierNamespace).mainRouteCasing"
-        static let alternativeRoutes = "\(identifierNamespace).alternativeRoutes"
-        static let alternativeRoutesCasing = "\(identifierNamespace).alternativeRoutesCasing"
-
-        static let route = "\(identifierNamespace).route"
-        static let routeCasing = "\(identifierNamespace).routeCasing"
 
         static let waypointCircle = "\(identifierNamespace).waypointsCircle"
         static let waypointSymbol = "\(identifierNamespace).waypointsSymbol"
@@ -131,6 +121,18 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     var routes: [Route]?
     var isAnimatingToOverheadMode = false
     
+    var mainRouteLayerIdentifier: String? {
+        guard let route = routes?.first?.description else { return nil }
+        
+        return "\(route.description)_route"
+    }
+    
+    var mainRouteCasingLayerIdentifier: String? {
+        guard let route = routes?.first?.description else { return nil }
+        
+        return "\(route.description)_casing"
+    }
+    
     var shouldPositionCourseViewFrameByFrame = false {
         didSet {
             if shouldPositionCourseViewFrameByFrame {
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationMapView.swift
```diff
@@ -469,62 +468,101 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+                
+                let routeCasingSource = addRouteSource(style, identifier: routeCasingSourceIdentifier, shape: mainRouteCasingShape)
+                
+                parentLayer = addMainRouteCasingLayer(style,
+                                                      source: routeCasingSource,
+                                                      identifier: routeCasingIdentifier,
+                                                      lineGradient: routeCasingGradient(0.0),
+                                                      below: mainRouteLayer)
+                
+                continue
```
MaximAlien, line 507: I think this is the most important change. As documentation was slightly changed in https://github.com/mapbox/mapbox-navigation-ios/pull/2623 to address that `NavigationMapViewDelegate.navigationMapView(_:shapeFor:)` and `NavigationMapViewDelegate.navigationMapView(_:simplifiedShapeFor:)` are meant to be used only for main route in comment:
```
Resulting `MGLShape` will then be styled using `NavigationMapView.navigationMapView(_: mainRouteStyleLayerWithIdentifier: source:)` provided style or a default congestion style if above delegate method was not implemented. In latter case, consider modifing your custom `MGLShape` `attributes` to have 'isAlternateRoute' key set to 'false'. Otherwise style predicate condition will filter out the shape.
```
I think it has become more misleading, as `NavigationMapViewDelegate.navigationMapView(_:shapeFor:)` allows to change shape not only for main route, but for all routes.

After adding the ability to provide route style for every route in this PR `NavigationMapViewDelegate.navigationMapView(_:shapeFor:)` and `NavigationMapViewDelegate.navigationMapView(_:simplifiedShapeFor:)` will be called only for main route, this means there will be no ability to provide custom shape for alternative routes and their casings.

Because of this I think `NavigationMapViewDelegate.navigationMapView(_:shapeFor:)` can be updated to be similar to `NavigationMapViewDelegate.navigationMapView(_:simplifiedShapeFor:)` and provide `Route` instead of `[Route]`, but it'd break public API.

@1ec5, @Udumft please let me know if that sounds reasonable and any changes are needed.

Example app was also updated to showcase the ability to provide custom layers and shapes in https://github.com/mapbox/navigation-ios-examples/pull/83.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -469,62 +468,101 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      */
     public func show(_ routes: [Route], legIndex: Int = 0) {
```
MaximAlien, line 469: @1ec5, do you remember how `legIndex` can be used here? As I can see after adding vanishing route line functionality it's no longer used when creating shape for the route, and the only place I can see where it was used before was to control `public let MBCurrentLegAttribute = "isCurrentLeg"` property in`MGLPolylineFeature`. 

After running few tests I don't see that setting `isCurrentLeg` attribute to `true` or `false` affects anything though.

MapboxNavigation/NavigationMapView.swift
1ec5: > Because of this I think `NavigationMapViewDelegate.navigationMapView(_:shapeFor:)` can be updated to be similar to `NavigationMapViewDelegate.navigationMapView(_:simplifiedShapeFor:)` and provide `Route` instead of `[Route]`, but it'd break public API.

We can add a new delegate method and deprecate the old one, as long as the old one continues to function reasonably appropriately.

MapboxNavigation/NavigationMapView.swift
1ec5: This is functionality that broke when implementing #1307. Any leg other than the current leg is supposed to be dimmed out: #2678. That no longer matters as much for already-passed legs, but it does matter for legs ahead of the current leg.

MaximAlien: Thanks for clarifying this. In this case probably good idea would be to tackle it in https://github.com/mapbox/mapbox-navigation-ios/issues/2733, fix for it might resolve https://github.com/mapbox/mapbox-navigation-ios/issues/2678 as well.

MapboxNavigation/NavigationMapView.swift
MaximAlien: To play it safe I'm inclined to leave current implementation without adding new/deprecating old APIs: end user will be able to change shape only for main route instead of all routes (main + alternative). This will match current docs as well.

1ec5 (approved):
Remember to add a changelog entry. Thanks!

Issue #2626, opened 2020-09-14
Top banner swiping is not mirrored for right-to-left languages


1ec5:
TopBannerViewController allows the user to swipe to the left to reveal the subsequent maneuver coming in from the right. In right-to-left languages, such as Arabic or Hebrew, the user should swipe to the right to reveal the subsequent maneuver coming in from the left. However, the interface isn’t currently mirrored, so the user has to swipe in a direction that is counterintuitive for them.

InstructionsCardViewController correctly shows the subsequent maneuver peeking in from the left, but swiping either left or right tends to make the carousel get confused and fast-forward to the end of the route.

/cc @mapbox/navigation-ios @mapbox/1tap-ios

1ec5:
Might be good to look at at the same time as #2695.

MarcoCarnevali:
@1ec5 can we add an option to hide subsequent maneuvers ?

jill-cardamon:
Looks like top banner swiping is hard-coded as right and left:
https://github.com/mapbox/mapbox-navigation-ios/blob/b7280efb6cc77eb111a4afd4bd5d956a2bfd3191/MapboxNavigation/NavigationViewController.swift#L802-L810

1ec5:
> can we add an option to hide subsequent maneuvers ?

Can you elaborate? Do you mean hiding a particular maneuver? Or disabling the feature where the next maneuver’s card always peeks in from the side? Or the feature where a subsequent maneuver in close proximity is attached to the current card? Either way, it would be an unrelated feature request, so please open a separate issue so we can track it. Thanks!

truburt:
@jill-cardamon , can we close this one?

jill-cardamon:
yep will close today!

PR #2724, opened 2020-11-04
Right to Left Language instruction card swipe bug

jill-cardamon:
### Description
This PR is to fix top banner swiping not being mirrored for right-to-left languages for instructions cards #2626.
I attempted to fix this issue, but the code was starting to look messy/hard to read with a bunch of cases to check whether the language was right to left. A higher-level abstraction should be able to fix this issue.
- [x] Find and implement abstraction/workaround
- [x] Fix issue which was return incorrect card width after multiple rotations on iPad.
- [x] Update Changelog

### Implementation
Added `UICollectionViewFlowLayout` extension to override `flipsHorizontallyInOppositeLayoutDirection`.

To fix the iPad issue, I added `InstructionsCardViewController.viewWillTransition(to:with:)`.

### Screenshots or Gifs
ezgif-1-a0f73e8d4da4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@
 * Fixed an issue which was preventing the ability to scroll between instructions cards on iOS 14 using workaround. ([#2755](https://github.com/mapbox/mapbox-navigation-ios/pull/2755))
 * Fixed an instructions cards layout issue that arose when changing orientation (portrait to landscape). ([#2755](https://github.com/mapbox/mapbox-navigation-ios/pull/2755))
 * Fixed swiping for right-to-left languages for the traditional top banner to be more intuitive. ([#2755](https://github.com/mapbox/mapbox-navigation-ios/pull/2755))
+* Fixed swiping for right-to-left languages for the Guidance Card UI to be more intuitive. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
+* Fixed an issue which was returning incorrect card width after multiple rotations on iPads. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
 
 ### Location tracking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@
 * Fixed an instructions cards layout issue that arose when changing orientation (portrait to landscape). ([#2755](https://github.com/mapbox/mapbox-navigation-ios/pull/2755))
 * Fixed swiping for right-to-left languages for the traditional top banner to be more intuitive. ([#2755](https://github.com/mapbox/mapbox-navigation-ios/pull/2755))
 * Fixed swiping for right-to-left languages for the Guidance Card UI to be more intuitive. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
-* Fixed an issue which was returning incorrect card width after multiple rotations on iPads. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
+* Fixed an issue which was returning incorrect card width after multiple rotations on iPads by adding `InstructionsCardViewController.viewWillTransition(to:with:)`. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
 
 ### Location tracking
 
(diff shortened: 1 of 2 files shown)


MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -177,12 +183,16 @@ open class InstructionsCardViewController: UIViewController {
     
     func snapToIndexPath(_ indexPath: IndexPath) {
         guard let itemCount = steps?.count, itemCount >= 0 && indexPath.row < itemCount else { return }
-        instructionsCardLayout.collectionView?.scrollToItem(at: indexPath, at: .left, animated: true)
+        instructionsCardLayout.collectionView?.isPagingEnabled = false
```
MaximAlien, line 186: Xcode 12.1 (iOS 14.1) and Xcode 12.2 (iOS 14.2) are available. We should test whether scrolling issue was fixed in those versions. If not, I think it's acceptable to use such workaround. It should be thoroughly tested though: in landscape and portrait modes during rotation for iPhones (with/without notches) and iPads. 

There should also be `@available` check to run this code only for iOS 14 and leave it unchanged for iOS 13 and lower.

jill-cardamon: Sounds good. I will test with Xcode 12.2. I tested with Xcode 12.1 and it wasn't fixed.

Will add the check!

jill-cardamon: Wasn't fixed with Xcode 12.2 (iOS 14.2)

MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -165,6 +169,8 @@ open class InstructionsCardViewController: UIViewController {
         let remainingSteps = legProgress.remainingSteps
         guard let currentCardStep = remainingSteps.first else { return }
         
+        print("!!! indexPaths in updateVisibleInstructionCards \(indexPaths)")
```
1ec5, line 172: Remember to remove print statements before merging.

MaximAlien (commented):
After landing https://github.com/mapbox/mapbox-navigation-ios/issues/2755 do we really need current PR?

MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -121,6 +121,10 @@ open class InstructionsCardViewController: UIViewController {
     
     @objc func orientationDidChange(_ notification: Notification) {
         instructionsCardLayout.invalidateLayout()
+        
+        instructionsCardLayout.collectionView?.isPagingEnabled = false
```
MaximAlien, line 125: Since https://github.com/mapbox/mapbox-navigation-ios/pull/2755 has landed to `main` rebasing will be needed to resolve outstanding conflicts. This part is no longer needed I guess.

jill-cardamon: The point of this PR is to fix swiping for right-to-left languages when using instructions cards. I haven't pushed a commit of the fix yet, but it would remove most of the added code from this commit history (since I separated into the PR I merged this morning). I can either close this PR and do a separate one, or just push the changes on this branch that I currently have. I think that the conflict will not exist after I do that.

MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -290,6 +305,10 @@ extension InstructionsCardViewController: UICollectionViewDataSource {
         let distance = firstStep ? distanceRemaining : step.distance
         cell.configure(for: step, distance: distance)
         
+        if UIApplication.shared.userInterfaceLayoutDirection == .rightToLeft {
```
MaximAlien, line 308: As in previous comment, let's try to use property recommended by Apple.

I can also see warning related to incorrectly handling layout constraint:
```
2020-12-21 14:58:39.322972-0800 Example[42832:1629484] the item height must be less than the height of the UICollectionView minus the section insets top and bottom values, minus the content insets top and bottom values.
```

I think decrease of UICollectionView height should fix it (should be tested though):
```swift
public func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
    
    return CGSize(width: cardSize.width, height: cardSize.height - 10)
    // return cardSize
}
```

MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -100,7 +100,7 @@ open class InstructionsCardViewController: UIViewController {
         instructionCollectionView.backgroundColor = .clear
         instructionCollectionView.isPagingEnabled = true
         instructionCollectionView.translatesAutoresizingMaskIntoConstraints = false
-
```
MaximAlien, line 103: Please try to minimize changes to empty lines.

MaximAlien (approved):
Since not that much code was changed in this PR I propose to squash it to keep history cleaner.

MaximAlien:
I can see that such tests started failing: 
```
GuidanceCardsSnapshotTests.testLanesManeuver()
GuidanceCardsSnapshotTests.testRegularManeuver()
GuidanceCardsSnapshotTests.testTertiaryManeuver()
```

Can you make sure that they're related to your current changes and fix them if needed?

Issue #2695, opened 2020-10-14
InstructionsCard orientation & swipe bug


MarcoCarnevali:
Hello, while using the new InstructionsCard view I noticed 2 relevant bugs: 
1. When using navigation on portrait and then change the orientation to landscape this layout issue occur: 

![image_from_ios](https://user-images.githubusercontent.com/9656572/96027764-a39fc580-0e58-11eb-9c76-ae17836ce2ee.png)

2. At the top the second card is visible. I'm not sure if this is a bug or is expected but I'm unable to scroll to see the next card instruction. Seems a weird UX
![image_from_ios-2](https://user-images.githubusercontent.com/9656572/96027848-ccc05600-0e58-11eb-940f-645dc327841b.png)

**Mapbox Navigation SDK version: 1.0.0**

jill-cardamon:
Hello, here is an update on the Instruction Card bugs:

1. Landscape layout issue:
    - I have reproduced the issue. It only appears on iOS 14.
    - Still investigating, but I know that the issue is likely related to the card size and animations.

2. Swipe to next card issue:
    - I have reproduced the issue and identified the problem. It only appears on iOS 14.
    - The scroll view will not end dragging due to something going wrong with `snapToIndexPath(..)`: https://github.com/mapbox/mapbox-navigation-ios/blob/b7280efb6cc77eb111a4afd4bd5d956a2bfd3191/MapboxNavigation/InstructionsCardViewController.swift#L247
    - The `snapToIndexPath(..)` function calls `UICollectionView`'s `scrollToItem(at:at:animated:)`, and `scrollToItem(at:at:animated:)` only works when paging is disabled.
    - Found a patch fix by disabling paging, calling `scrollToItem(at:at:animated:)`, and then enabling paging.
    - Other developers are having similar issues with `scrollToItem(at:at:animated:)`, so it could be potentially fixed by an iOS 14 update. We will continue to look for a long-term solution.

MarcoCarnevali:
@jill-cardamon is this fixed?

jill-cardamon:
@MarcoCarnevali The second bug you reported has been fixed. The first bug is being fixed with #2732.

MarcoCarnevali:
Ok, Thanks for the update!

truburt:
@jill-cardamon , can this be closed as well?

PR #2727, opened 2020-11-06
Update snapshot tests for iOS 14

jill-cardamon:
Snapshot tests are specific to a particular iOS version, so a lot of tests are failing locally. This PR is to update these tests to run for iOS 14.
- [x] copy screenshots from iOS 13 for iOS 14
- [x] for failed tests, replace them with screenshots that the test harness came up with

PR #2732, opened 2020-11-09
Make sure to resize the gradient view under the current Instruction Card during a layout change

avi-c:
Fix for one of the issues in #2695

### DIFF
--- a/MapboxNavigation/InstructionsCardContainerView.swift
+++ b/MapboxNavigation/InstructionsCardContainerView.swift
@@ -72,9 +72,7 @@ public class InstructionsCardContainerView: StylableView {
     @objc dynamic public var highlightedBackgroundColor: UIColor = UIColor(red: 0.26, green: 0.39, blue: 0.98, alpha: 1.0)
     
     public weak var delegate: InstructionsCardContainerViewDelegate?
-    
-    private var gradientLayer: CAGradientLayer!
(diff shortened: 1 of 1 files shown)


avi-c:
I think the issue with the failing tests is the need to update the image snapshots. I don't know why only those particular fail but I think the main difference between the snapshot and the rendered image are the use of corner radii and some font changes.

jill-cardamon:
Makes sense. Lately, tests have been failing for iOS 14/Xcode 12, so it's weird that they're failing older iOS versions.

avi-c:
@1ec5 - If I want to update the test assets, would that happen under this ticket or do we need a different one?

1ec5:
Can we update it as part of #2727?

avi-c:
Returning to this since it's fallen through the cracks. I am still confident that the code is correct but the tests need updating.

@jill-cardamon - Can we combine your updating of the iOS 14 snapshots with including this PR so that everything can happen at once? Maybe we can merge this PR into your branch and regenerate the snapshots there, then merge it all in with #2727

1ec5:
Sorry that this PR fell through the cracks. Do you think we could try to revive it for v2.0, now that UI tests are once again working there?

S2Ler:
This PR merged already as part of 99dfa32

PR #2734, opened 2020-11-10
Route Duration Annotations

avi-c:
Add callout annotations showing route durations for comparing alternate routes during preview.

Will include noting which routes have tolls when that information is available.


Extensions/MGLMapView+Example.swift
```diff
@@ -0,0 +1,76 @@
...
+
+extension MGLMapView {
+    func coordinateBoundsInset(_ inset: CGSize) -> MGLCoordinateBounds {
+        return convert(bounds.insetBy(dx: inset.width, dy: inset.height), toCoordinateBoundsFrom: nil)
+    }
+}
+
+
+extension MGLStyle {
+    func addDebugLineLayer(identifier: String, coordinates: [CLLocationCoordinate2D], color: UIColor = UIColor.purple) {
```
1ec5, line 13: Ultimately, we’ll need to move these extensions from the example application to MapboxNavigation so that developers don’t have to copy the whole file.

Extensions/MGLMapView+Example.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+import Mapbox
+import Turf
+
+extension MGLMapView {
+    func coordinateBoundsInset(_ inset: CGSize) -> MGLCoordinateBounds {
+        return convert(bounds.insetBy(dx: inset.width, dy: inset.height), toCoordinateBoundsFrom: nil)
```
1ec5, line 7: `MGLMapView.convert(_:toCoordinateBoundsFrom:)` is inaccurate on a tilted map: mapbox/mapbox-gl-native#2259.

More broadly, I think we should investigate positioning the labels based on screen coordinates rather than geographic coordinates. Core GL’s own symbol collision avoidance code is based on projected coordinates, which account for rotation and tilting, so it would make sense to be consistent with core GL labeling on that respect.

Imagine taking the maneuver coordinates for a given route, converting them all to screen points, and stringing them together in a UIBezierPath. Calculating the distance between two points is suddenly just a Pythagorean distance problem.

1ec5 (commented):
Nice, it really looks the part! A couple initial thoughts:

avi-c:
Made a number of updates to my algorithm that simplify it and reduce the total number of needed additions and extensions. I did add the need for an algorithm to simplify a polyline and chose to use the one from: https://github.com/tomislav/Simplify-Swift. I resorted to compiling in the source file since the library is not available through Carthage but we can figure out another solution, including re-implementing the algorithm.

Extensions/MGLMapView+Example.swift
avi-c: Removed this extension since I no longer needed it. I am still doing a lot of my calculations in geographic coordinates, but do filter them to only use ones that are onscreen. Also, when coordinates are chosen for the annotations I use the on-screen position to decide where to put the "stem" of the bubble callout.

Extensions/MGLMapView+Example.swift
avi-c: Done.

Example/ViewController.swift
```diff
@@ -48,7 +54,237 @@ class ViewController: UIViewController {
...
+        if let image = UIImage(named: "RouteInfoAnnotationRightHanded") {
+            let regularRouteImage = image.tint(UIColor.white).resizableImage(withCapInsets: capInsets, resizingMode: .stretch)
+            style.setImage(regularRouteImage, forName: "RouteInfoAnnotationRightHanded")
+
+            let selectedRouteImage = image.tint(#colorLiteral(red: 0.337254902, green: 0.6588235294, blue: 0.9843137255, alpha: 1)).resizableImage(withCapInsets: capInsets, resizingMode: .stretch)
+            style.setImage(selectedRouteImage, forName: "RouteInfoAnnotationRightHanded-Selected")
+        }
+    }
+
+    private func updateRouteAnnotations(_ routes: [Route]?, style: MGLStyle) {
```
MaximAlien, line 87: This method is making pretty heavy computation each time. I'd probably consider outlining only specific actions when we should update delay view location (e.g. it doesn't make sense to show delay view on low zoom levels, since route is not really visible anyway). We can also move certain calculations to background thread.

Example/ViewController.swift
```diff
@@ -48,7 +54,237 @@ class ViewController: UIViewController {
...
+        // remove any existing route annotation
+        removeRouteAnnotationsLayerFromStyle(style)
```
MaximAlien, line 89: When route is selected `ViewController.response` will trigger `clearMapView()`, which will in turn clear delay view. Since this feature is going to be part of `MapboxNavigation` it makes sense to move existing logic into `NavigationMapView` instead of improving implementation in Example app.

Example/Simplify.swift
```diff
@@ -0,0 +1,201 @@
...
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import Foundation
```
MaximAlien, line 24: Personally, I'm not really fond of storing 3rdparty sources as is. Not sure it's good idea to have separate dependency either. It'd be great if we can have our own implementation of simplification functionality.

MaximAlien (commented):
Thanks for this PR. I like current look of delay views. In my opinion we should aim for implementation on Navigation SDK side. There are also other changes being made to `NavigaitonMapView` at this point (route styling, congestion consistency, vanishing route line calculation improvements) and it has become pretty large class. I think it'd make sense to extract certain areas of functionality into separate entities (e.g. at least `NavigationMapView` extensions).

Example/Simplify.swift
avi-c: Agreed. It isn't something I am very familiar with though. @1ec5 - Is there a mapbox implementation that we could port over to the Nav SDK so I could drop any form of dependency?

MapboxNavigation/MGLStyle.swift
```diff
@@ -87,4 +88,129 @@ extension MGLStyle {
             removeSource(source)
         }
     }
+
+    /**
+     Convenience method for adding a circle at a given coordinate.
+
+     Useful for debugging or visualizing data.
+     */
+    public func addDebugCircleLayer(identifier: String, coordinate: CLLocationCoordinate2D, color: UIColor = UIColor.purple) {
```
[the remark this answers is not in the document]
avi-c: Yeah, they are just for debugging, so I can move them around, but we can also drop them from being in the project. I find them very useful but could just keep them in my own repo or gist or something. @1ec5 - Do you think there is a place for this type of debugging code in the actual project or just something to keep for ourselves?

Example/ViewController.swift
```diff
@@ -94,6 +330,10 @@ class ViewController: UIViewController {
         if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
             appDelegate.currentAppRootViewController = self
         }
+
+        dateComponentsFormatter.maximumUnitCount = 3
```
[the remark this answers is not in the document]
avi-c: I'll look at consolidating all of these within MapboxNavigation.

Example/ViewController.swift
```diff
@@ -48,7 +54,237 @@ class ViewController: UIViewController {
...
+            13: NSExpression(forConstantValue: 16),
+            15.5: NSExpression(forConstantValue: 20)
+        ]
+        shapeLayer.textFontSize = NSExpression(format: "mgl_interpolate:withCurveType:parameters:stops:($zoomLevel, 'linear', nil, %@)", fontSizeByZoomLevel)
+
+        shapeLayer.textColor = NSExpression(forConditional: NSPredicate(format: "selected == true"),
+                                            trueExpression: NSExpression(forConstantValue: UIColor.white),
+                     falseExpression: NSExpression(forConstantValue: UIColor.black))
+
+        shapeLayer.textFontNames = NSExpression(forConstantValue: ["DIN Pro Medium"])
```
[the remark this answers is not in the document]
avi-c: There are some good discussions to have on exactly what visual treatment we want. I did make the text size vary by zoom as I saw that being done in some other nav apps.

I'd prefer to investigate the designs in a separate ticket later though since it's bound to take a bit longer to come to agreement on the details.

Example/ViewController.swift
```diff
@@ -48,7 +54,237 @@ class ViewController: UIViewController {
...
+        point.attributes = ["selected": true, "tailPosition": selectedRouteTailPosition.rawValue, "text": labelText, "imageName": selectedRouteTailPosition == .left ? "RouteInfoAnnotationLeftHanded-Selected" : "RouteInfoAnnotationRightHanded-Selected"]
+        features.append(point)
+
+        // add the features to the style
+        addRouteAnnotationSymbolLayer(features: features, style: style)
+    }
+
+    private let annotationLayerIdentifier = "RouteETAAnnotations"
+
+    private func addRouteAnnotationSymbolLayer(features: [MGLPointFeature], style: MGLStyle) {
```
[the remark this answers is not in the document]
avi-c: Should I move this PR to come on top of some of the other work you have in progress? What would you suggest as a way to coordinate this rework/refactor of the `NavigationMapView` class?

Example/ViewController.swift
MaximAlien: It'd be also good to rebase current changes onto `main` as it now contains at least one fix to example app which can be useful here (previously after pressing on route clear button would disappear).

Example/ViewController.swift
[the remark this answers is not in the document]
MaximAlien: Unfortunately there are multiple PRs which contain changes to `NavigationMapView`: https://github.com/mapbox/mapbox-navigation-ios/pull/2741, https://github.com/mapbox/mapbox-navigation-ios/pull/2737 and https://github.com/mapbox/mapbox-navigation-ios/pull/2719. Easiest way would be to rebase onto `main` for now. I can then help with tackling conflicts (if there are any) after mentioned PRs are merged into `main`.

Example/ViewController.swift
avi-c: @1ec5 @MaximAlien - This brings up a question I have. Can we perform map style updates from a background thread?

MapboxNavigation/MGLStyle.swift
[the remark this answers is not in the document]
avi-c: They are useful not just for our own debugging, they are a good way to see geo info on the map with just simple coordinates. If @1ec5 thinks it is ok to add debug functions to the SDK then that's fine with me. If not, I would move them to the Example app and leave them there.

Example/ViewController.swift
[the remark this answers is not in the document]
avi-c: I moved the CPTrip `DateComponentsFormatter` extensions to a dedicated file within MapboxNavigation and changed to use those instead. Removed the use of a local `DateComponentsFormatter` within the Route Annotations file.

Example/ViewController.swift
avi-c: Completed the rebase.

Example/ViewController.swift
avi-c: Actually, I don't think this is actually that heavy in terms of performance. I profiled several recomputations of the route annotation coordinates by repeatedly scrolling the map view (which causes a re-layout) and the time spent in this routine is only about 1% of the CPU time. It also will only happen when the map is moved or otherwise redrawn.

It's also much, much smaller than the delay in retrieving the new route from the Directions API so causes only imperceptible delays in the case when a new route is being fetched. 

I'd argue it's better just to leave it as is than to try and add in some concurrency.

Hopefully @1ec5 can weigh in on this.

Example/Simplify.swift
avi-c: Will move to a simplify algorithm I added to Turf: https://github.com/mapbox/turf-swift/pull/124
Once that PR is approved and merged I will update the Turf version in Nav SDK and use it instead of another dependency.

MapboxNavigation/MGLStyle.swift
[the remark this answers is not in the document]
1ec5: I didn’t realize they were debugging methods when I suggested moving them to MapboxNavigation. It makes sense to keep them limited to the example application. Sorry for the confusion.

Example/ViewController.swift
1ec5: It may be worth moving the calculations to a background thread just in case, but most likely the reason it hasn’t come up as a performance bottleneck yet is that it only runs once per camera change and not continuously (which is fine).

Example/Simplify.swift
1ec5: Turf v1.2.0-beta.1 has been released. MapboxDirections doesn’t require it yet, but you can have the navigation SDK require it by modifying the Cartfile in this repository. The CocoaPods integration test will fail until there’s a final release of Turf v1.2.0, but at least you’ll be able to prepare this PR for the moment when that happens.

```diff
diff --git a/Cartfile b/Cartfile
index ba3e3e59..1cf9c2ca 100644
--- a/Cartfile
+++ b/Cartfile
@@ -3,7 +3,7 @@ binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/Ma
 binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps/mapbox-ios-sdk-dynamic.json" ~> 6.0
 binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon-ios.json" == 7.1.2
 github "mapbox/mapbox-directions-swift" ~> 1.0
-github "mapbox/turf-swift" ~> 1.0
+github "mapbox/turf-swift" ~> 1.2.0-beta.1
 github "mapbox/mapbox-events-ios" ~> 0.10.2
 github "ceeK/Solar" ~> 2.1.0
 github "mapbox/mapbox-speech-swift" ~> 1.0
diff --git a/Cartfile.resolved b/Cartfile.resolved
index 1b02b2c1..6042bedd 100644
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -10,8 +10,8 @@ github "ceeK/Solar" "2.1.0"
 github "linksmt/OHHTTPStubs" "563f48d3fab84ef04639649c770b00f4fa502cca"
 github "mapbox/MapboxGeocoder.swift" "v0.10.2"
 github "mapbox/mapbox-directions-swift" "v1.1.0"
-github "mapbox/mapbox-events-ios" "v0.10.6"
+github "mapbox/mapbox-events-ios" "v0.10.7"
 github "mapbox/mapbox-speech-swift" "v1.0.0"
-github "mapbox/turf-swift" "v1.1.0"
+github "mapbox/turf-swift" "v1.2.0-beta.1"
 github "raphaelmor/Polyline" "v5.0.2"
 github "uber/ios-snapshot-test-case" "5.0.2"
```

avi-c:
@MaximAlien @1ec5 - I reworked the algorithm a bit because I wasn't satisfied with how the annotations were landing.

It has a number of steps but should actually be much simpler computation wise since it is mainly things like coordinate comparisons and a few closestCoordinateTo computations.

See what you think. I tried to make sure the comments were giving a clear step-by-step overview of what is happening.

avi-c:
@1ec5 @MaximAlien - I'll fix the merge conflicts but would love to get this merged and out of the way this week. If there are changes you'd like to see please update and I'll address those too.

Example/ViewController.swift
```diff
@@ -44,13 +50,229 @@ class ViewController: UIViewController {
...
+                        return false
+                    }
+                }
+                return true
+            }
+
+            excludedSteps.append(contentsOf: alternateSteps)
+            let visibleAlternateSteps = alternateSteps.filter { $0.intersects(visibleBoundingBox) }
+
+            var coordinate = kCLLocationCoordinate2DInvalid
```
1ec5, line 120: Use nil instead of the C constant `kCLLocationCoordinate2DInvalid`. The compiler won’t remind you to check for whether the coordinate is valid, but it would remind you about an unhandled `nil` case for an optional.

Example/ViewController.swift
```diff
@@ -44,13 +50,229 @@ class ViewController: UIViewController {
...
+            // We will consider a good spot to be somewhere near the middle of the line, making sure that the coordinate is visible on-screen.
+            if let continuousLine = visibleAlternateSteps.continuousShape(), continuousLine.coordinates.count > 0 {
+                coordinate = continuousLine.coordinates[0]
+
+                // Pick a coordinate using some randomness in order to give visual variety.
+                // Take care to snap that coordinate to one that lays on the original route line.
+                // If the chosen snapped coordinate is not visible onthe screen, then we walk back along the route coordinates looking for one that is.
+                // If none of the earlier points are on screen then we walk forward along the route coordinates until we find one that is.
+                if let distance = continuousLine.distance(), let sampleCoordinate = continuousLine.indexedCoordinateFromStart(distance: distance * CLLocationDistance.random(in: 0.3...0.8))?.coordinate, let routeShape = route.shape, let snappedCoordinate = routeShape.closestCoordinate(to: sampleCoordinate) {
```
1ec5, line 131: What’s the significance of the start and end ranges of this random distance? Should it depend on the zoom level or the style? Does the range hold up better in some kinds of routes versus others?

I’m still concerned that, by focusing on geographic coordinates rather than screen coordinates as suggested in https://github.com/mapbox/mapbox-navigation-ios/pull/2734#discussion_r521687903, the whole algorithm is vulnerable to a variety of factors that are hard to control for at this high level of the stack.

In particular, `LineString.indexedCoordinateFromStart(distance:)` can only return one of the route line’s vertices, but there’s no guarantee about the density of vertices along the route line. For example, in some places in OpenStreetMap, a vertex may occur only once every quarter mile at an intersection, while in other places, it may occur every few car lengths because turn lanes and parking restrictions have been mapped in fine detail.

Example/ViewController.swift
```diff
@@ -44,13 +50,229 @@ class ViewController: UIViewController {
...
     }
-    
+
+    // This function generates the text for the label to be shown on screen. It will include estimated duration and info on Tolls, if applicable
+    private func annotationLabelForRoute(_ route: Route, tolls: Bool) -> String {
+        var eta = DateComponentsFormatter.shortDateComponentsFormatter.string(from: route.expectedTravelTime) ?? ""
+
+        let hasTolls = (route.tollIntersections?.count ?? 0) > 0
+        if hasTolls {
+            eta += "\n" + NSLocalizedString("ROUTE_HAS_TOLLS", value: "Tolls", comment: "This route does have tolls")
```
1ec5, line 263: Remember to run scripts/extract_localizable.sh to update the Base localization.

Example/ViewController.swift
```diff
@@ -44,13 +50,229 @@ class ViewController: UIViewController {
...
+            if let symbol = Locale.current.currencySymbol {
+                eta += " " + symbol
```
1ec5, line 265: Normally, we would avoid concatenating multiple “lego brick” strings into a larger string and instead define a format string. But it looks as if this space is being used for layout rather than to piece together parts of a larger localizable string. Perhaps that would be clearer if this code would make use of the map SDK’s format string type instead of a plain-text string.

Example/ViewController.swift
```diff
@@ -44,13 +50,229 @@ class ViewController: UIViewController {
...
-    
+
+    // This function generates the text for the label to be shown on screen. It will include estimated duration and info on Tolls, if applicable
+    private func annotationLabelForRoute(_ route: Route, tolls: Bool) -> String {
+        var eta = DateComponentsFormatter.shortDateComponentsFormatter.string(from: route.expectedTravelTime) ?? ""
+
+        let hasTolls = (route.tollIntersections?.count ?? 0) > 0
+        if hasTolls {
+            eta += "\n" + NSLocalizedString("ROUTE_HAS_TOLLS", value: "Tolls", comment: "This route does have tolls")
+            if let symbol = Locale.current.currencySymbol {
```
1ec5, line 264: This currency symbol would be based on the user’s current systemwide region setting, which could differ from the geographic region that contains the route. A more accurate answer would come from a Locale created using the identifier in the tolled intersection’s `Intersection.regionCode` property, or more simply using the return value from passing zeros into `RouteLeg.regionCode(atStepIndex:intersectionIndex:)`.

Example/ViewController.swift
avi-c: The intent is to pick a coordinate along the line some distance away from the start and end points of the route, but with some variation of the distance. The goal is to not hard code something like halfway through the route.

I went with the indexed coordinate in order to reduce the computation burden of calculating a point along the line somewhere between two vertices. It might be that the overall algorithm is simplified enough that we wouldn't need to be calculating as many points and it would be worth it.

In my testing, the annotations generally work out very well, but do sometimes collide. I suspect that the code required to calculate things in screen space so that they would never collide would be significantly more complicated than just doing the geographic calculations.

My hope is that the logic we have is good enough for most of the time and that any customer who found the collision avoidance to not work to their liking would just figure out what works best for them.

Example/ViewController.swift
avi-c: The formatting here is indeed to just give some padding between the time string and the string that is appended on if the route has some tolls.

This is to match other nav apps that add this extra notation in a line underneath the quoted duration.

I did it as this concatenating logic so that the more common case where there are no tolls would be as easy as possible and then it would be a simple string addition for cases where it is needed.

Example/ViewController.swift
avi-c: Since the annotation shows up only once per route line (regardless of how many tolls may be along the route), I chose the currency symbol based on what would be familiar to the user based on their settings. I could look up the currency symbol at the location where one of the tolls is, but then the question would be which toll location? If they occurred across several geographies then we'd have to pick a single one.

For this purpose where all we are trying to do is give the user a bit more information so that they can decide between multiple alternate routes I think it's ok for the symbol to not necessarily be correct for the location of the toll, it need only be recognizable to the user as "This route requires additional payment from me".

avi-c:
Responded to the review feedback, including moving the annotations from the Example project over to NavigationMapView.

Example/ViewController.swift
```diff
@@ -44,13 +50,229 @@ class ViewController: UIViewController {
...
+        for (index, route) in routes.enumerated() {
+            let allSteps = route.legs.compactMap { return $0.steps }.reduce([], +)
+            let alternateSteps = allSteps.filter { step -> Bool in
+                for existingStep in excludedSteps {
+                    if step == existingStep {
+                        return false
+                    }
+                }
+                return true
+            }
```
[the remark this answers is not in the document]
avi-c: Done.

Example/ViewController.swift
```diff
@@ -44,13 +50,229 @@ class ViewController: UIViewController {
...
+        // pick a random tail direction to keep things varied
+        guard let randomTailPosition = [RouteDurationAnnotationTailPosition.left, RouteDurationAnnotationTailPosition.right].randomElement() else { return }
+
+        var features = [MGLPointFeature]()
+
+        // Run through our heuristic algorithm looking for a good coordinate along each route line to place it's route annotation
+        // First, we will look for a set of RouteSteps unique to each route
+        var excludedSteps = [RouteStep]()
+        for (index, route) in routes.enumerated() {
+            let allSteps = route.legs.compactMap { return $0.steps }.reduce([], +)
```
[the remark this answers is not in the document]
avi-c: Fixed the instances I found.

Example/ViewController.swift
avi-c: Done.

Example/ViewController.swift
avi-c: Done.

Example/ViewController.swift
```diff
@@ -44,13 +50,229 @@ class ViewController: UIViewController {
...
+
+    private func updateAnnotationSymbolImages(_ style: MGLStyle) {
+        let capInsetHeight = CGFloat(22)
+        let capInsetWidth = CGFloat(11)
+        let capInsets = UIEdgeInsets(top: capInsetHeight, left: capInsetWidth, bottom: capInsetHeight, right: capInsetWidth)
+        if let image = UIImage(named: "RouteInfoAnnotationLeftHanded") {
+            let regularRouteImage = image.tint(UIColor.white).resizableImage(withCapInsets: capInsets, resizingMode: .stretch)
+            style.setImage(regularRouteImage, forName: "RouteInfoAnnotationLeftHanded")
+
+            let selectedRouteImage = image.tint(#colorLiteral(red: 0.337254902, green: 0.6588235294, blue: 0.9843137255, alpha: 1)).resizableImage(withCapInsets: capInsets, resizingMode: .stretch)
```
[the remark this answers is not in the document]
avi-c: Moved over, including the tint color.

Example/ViewController.swift
avi-c: Also, we could choose to omit the currency symbol if we think it is confusing in any way. It's there to reinforce with a symbol the textual message of "Has Tolls".

Example/ViewController.swift
1ec5: At a glance, probably the biggest risk at the moment would be that the map happens to be rotated or tilted differently than the positioning code was designed for. As long as these degrees of freedom are possible, it’ll remain difficult to completely account for them in geographic coordinate space. But to the extent that this poses a problem, we can treat it as tail work, perhaps awaiting new map SDK functionality to make such calculations easier.

MapboxNavigation/Route.swift
```diff
@@ -41,4 +41,62 @@ extension Route {
...
+
+        return intersectionsWithTolls
+    }
+
+    // returns the list of line segments along the route that fall within given bounding box. Returns nil if there are none. Line segments are defined by the route shape coordinates that lay within the bounding box
+    func shapes(within: Turf.BoundingBox) -> [LineString]? {
+        guard let coordinates = shape?.coordinates else { return nil }
+        var lines = [[CLLocationCoordinate2D]]()
+        var currentLine: [CLLocationCoordinate2D]?
+        for coordinate in coordinates {
```
1ec5, line 59: Another approach would be to pass `coordinates` and `coordinates.suffix(from: 1)` into `zip(_:_:)`, then convert the resulting pairs to LineStrings, then map them to `BoundingBox.contains(_:ignoreBoundary:)` or `intersection(_:_:)`. I’m not sure if this approach would be more performant, but it could be more elegant and reusable.

MapboxNavigation/Turf.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import Mapbox
+import Turf
+
+extension Turf.BoundingBox {
+    init(coordinateBounds: MGLCoordinateBounds) {
+        self.init(coordinateBounds.sw, coordinateBounds.ne)
```
1ec5, line 7: I somehow missed that the BoundingBox initializer has [unlabeled arguments](https://github.com/mapbox/turf-swift/blob/87d12f6c7a9ca28d451c1f85fe3a811d72485b50/Sources/Turf/BoundingBox.swift#L25). We should fix that independently of this PR.

MapboxNavigation/UIImage.swift
```diff
@@ -33,4 +33,29 @@ extension UIImage {
             (text as NSString).draw(in: rect.offsetBy(dx: 0, dy: (constrainedSize.height - font.lineHeight) / 2).integral, withAttributes: textFontAttributes)
         }
     }
+
+    func tint(_ tintColor: UIColor) -> UIImage {
+        let imageSize = size
+        let imageScale = scale
+        let contextBounds = CGRect(origin: .zero, size: imageSize)
+
+        UIGraphicsBeginImageContextWithOptions(imageSize, false, imageScale)
```
1ec5, line 42: By the way, `UIGraphicsImageRenderer` is also available on iOS 10 and above. It does have some shortcomings compared to the C-style `UIGraphicsBeginImageContextWithOptions(_:_:_:)` around transformations, but we’re already using it in several parts of the SDK.

MapboxNavigation/DateComponentsFormatter+MGLNavigationAdditions.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+extension DateComponentsFormatter {
+    public static let fullDateComponentsFormatter: DateComponentsFormatter = {
```
1ec5, line 4: Moving these formatters out of CPTrip is a good idea!

MapboxNavigation/Array.swift
```diff
@@ -32,3 +35,22 @@ extension Array where Element: NSAttributedString {
...
+    // Find the longest contiguous series of RouteSteps connected to the first one.
+    //
+    // tolerance: -- Maximum distance between the end of one RouteStep and the start of the next to still consider them connected. Defaults to 100 meters
+    func continuousShape(tolerance: CLLocationDistance = 100) -> LineString? {
+        guard count > 0 else { return nil }
+        guard count > 1 else { return self[0].shape }
+        var continuousLine = [CLLocationCoordinate2D]()
+
+        for index in 0...count-2 {
+            if let currentStepFinalCoordinate = self[index].shape?.coordinates.last, currentStepFinalCoordinate.distance(to: self[index+1].maneuverLocation) < tolerance, let coordinates = self[index].shape?.coordinates {
```
1ec5, line 49: Manual index arithmetic makes me nervous. Can we express this code differently using Sequence methods like `prefix(…)` and `zip(…)`?

```suggestion
        for index in 0...count-2 {
            if let coordinates = self[index].shape?.coordinates, let currentStepFinalCoordinate = coordinates.last, currentStepFinalCoordinate.distance(to: self[index+1].maneuverLocation) < tolerance {
```

then, since `maneuverLocation` is guaranteed to be the first coordinate in `shape`:

```swift
zip(compactMap { $0.shape }, suffix(from: 1).compactMap { $0.shape })
    .filter { $0.coordinates.last?.distance(to: $1.coordinates.first ?? Double.max) < tolerance }
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -128,7 +128,9 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             userCourseView = reducedAccuracyActivatedMode ? UserHaloCourseView(frame: CGRect(origin: .zero, size: 75.0)) : UserPuckCourseView(frame: CGRect(origin: .zero, size: 75.0))
         }
     }
-    
+
+    @objc dynamic public var routeDurationAnnotationSelectedColor: UIColor = .defaultRouteDurationAnnotationSelectedColor
```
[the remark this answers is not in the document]
avi-c: Your point about some styling properties being exposed and others not is a good one. I had exposed the ones I had needed to change, but it's a bit short sighted not to expose the others in order to allow for more complete customization of colors and font.

I made the naming change you suggested and also exposed background colors, text colors, and font as well.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -958,6 +960,226 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             SourceIdentifier.arrowSymbol,
         ].compactMap { style.source(withIdentifier: $0) }))
     }
+
+    public func showRouteDurationAnnotations(_ routes: [Route]?) {
```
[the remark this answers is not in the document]
avi-c: Added some documentation.

Right now the user will need to call the `showRouteDurationAnnotations` function themselves. I did it this way since I was trying to minimize the changes I was making to existing behavior and wanted to give people more fine-grained control on when the annotations show up, when they are refreshed, and when they are removed.

Not sure if that's the best way to go but it fits my "low impact" intention.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -958,6 +960,226 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        guard let style = style else { return }
+        updateAnnotationSymbolImages(style)
+        updateRouteDurationAnnotations(routes, style: style)
```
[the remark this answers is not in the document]
avi-c: Fixed.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -958,6 +960,226 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        // add the features to the style
+        self.addRouteAnnotationSymbolLayer(features: features)
+    }
+
+    private let annotationLayerIdentifier = "RouteDurationAnnotations"
+
+    private func addRouteAnnotationSymbolLayer(features: [MGLPointFeature]) {
+        guard let style = style else { return }
+        let dataSource: MGLShapeSource
+        if let source = style.source(withIdentifier: annotationLayerIdentifier + "-source") as? MGLShapeSource {
```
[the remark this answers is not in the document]
avi-c: Fixed.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -958,6 +960,226 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+
+            var tailPosition = randomTailPosition
+
+            // convert our coordinate to screen space so we can make a choice on which side of the coordinate the label ends up on
+            let unprojectedCoordinate = convert(annotationCoordinate, toPointTo: nil)
+
+            // pick the orientation of the bubble "stem" based on how close to the edge of the screen it is
+            if tailPosition == .left && unprojectedCoordinate.x > bounds.width * 0.75 {
+                tailPosition = .right
+            } else if tailPosition == .right && unprojectedCoordinate.x < bounds.width * 0.25 {
```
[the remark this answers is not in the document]
avi-c: Opened https://github.com/mapbox/mapbox-navigation-ios/issues/2785 to track this future investigation.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -958,6 +960,226 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        if let source = style.source(withIdentifier: annotationLayerIdentifier + "-source") as? MGLShapeSource {
+            dataSource = source
+        } else {
+            dataSource = MGLShapeSource(identifier: annotationLayerIdentifier + "-source", features: features, options: nil)
+            style.addSource(dataSource)
+        }
+
+        let shapeLayer: MGLSymbolStyleLayer
+
+        if let layer = style.layer(withIdentifier: annotationLayerIdentifier + "-shape") as? MGLSymbolStyleLayer {
```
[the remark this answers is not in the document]
avi-c: Fixed.

MapboxNavigation/Route.swift
avi-c: Still reviewing this now.

MapboxNavigation/Array.swift
avi-c: @1ec5 - I believe I have incorporated your feedback in the change I just made. Let me know if I missed an improvement to this function that you had suggested.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -958,6 +970,246 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             SourceIdentifier.arrowSymbol,
         ].compactMap { style.source(withIdentifier: $0) }))
     }
+
+    /**
+     Shows a callout containing the duration of each route.
+     Useful as a way to give the user more information when picking between multiple route alternatives.
+     If the route contains any tolled segments then the callout will specify that as well.
+     */
+    public func showRouteDurationAnnotations(_ routes: [Route]?) {
```
1ec5, line 979: An array of routes is not an array of route duration annotations, so the argument should have a label:

```swift
public func showRouteDurations(along routes: [Route]?) {
```

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -958,6 +970,246 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+
+            let selectedRouteImage = image.tint(routeDurationAnnotationSelectedColor).resizableImage(withCapInsets: capInsets, resizingMode: .stretch)
+            style.setImage(selectedRouteImage, forName: "RouteInfoAnnotationRightHanded-Selected")
+        }
+    }
+
+    /**
+     Remove any old route duration callouts and generate new ones for each passed in route.
+     */
+    private func updateRouteDurationAnnotations(_ routes: [Route]?) {
```
1ec5, line 1013:
```swift
private func updateRouteDurations(along routes: [Route]?) {
```

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -958,6 +970,246 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        shapeLayer.textOffset = shapeLayer.iconOffset
+        shapeLayer.iconAllowsOverlap = NSExpression(forConstantValue: true)
+
+        style.addLayer(shapeLayer)
+    }
+
+    /**
+     Removes all visible route duration callouts.
+     */
+    public func removeRouteDurationAnnotations() {
```
1ec5, line 1177:
```swift
public func removeRouteDurations() {
```

MapboxNavigation/Turf.swift
1ec5: This was fixed in mapbox/turf-swift#132, which will be coming in v2._x_.

Sources/MapboxNavigation/Array.swift
```diff
@@ -32,3 +35,25 @@ extension Array where Element: NSAttributedString {
...
+
+extension Array where Element == RouteStep {
+    // Find the longest contiguous series of RouteSteps connected to the first one.
+    //
+    // tolerance: -- Maximum distance between the end of one RouteStep and the start of the next to still consider them connected. Defaults to 100 meters
+    func continuousShape(tolerance: CLLocationDistance = 100) -> LineString? {
+        guard count > 0 else { return nil }
+        guard count > 1 else { return self[0].shape }
+
+        let filteredStepShapes = zip(compactMap { $0.shape }, suffix(from: 1).compactMap { $0.shape }).filter({
```
1ec5, line 47: Avoid mapping the array twice:

```suggestion
        let stepShapes = compactMap { $0.shape }
        let filteredStepShapes = zip(stepShapes, stepShapes.suffix(from: 1)).filter({
```

Sources/MapboxNavigation/MGLStyle.swift
```diff
@@ -87,4 +88,129 @@ extension MGLStyle {
             removeSource(source)
         }
     }
+
+    /**
+     Convenience method for adding a circle at a given coordinate.
+
+     Useful for debugging or visualizing data.
+     */
+    public func addDebugCircleLayer(identifier: String, coordinate: CLLocationCoordinate2D, color: UIColor = UIColor.purple) {
```
1ec5, line 97: Mention these new public methods (by fully-qualified name, like `MGLStyle.addDebugCircleLayer(identifier:coordinate:color:)` in the changelog, or make them private if they aren’t intended to be used outside the SDK.

Sources/MapboxNavigation/NavigationMapView.swift
1ec5, line 979: Mention these three new methods in the changelog.

1ec5 (approved):
Just some minor stylistic suggestions, documentation notes, and performance optimizations. (GitHub split up the review, so see some comments above too.)

CHANGELOG.md
```diff
@@ -35,6 +35,16 @@
   * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteCasingStyleLayerWithIdentifier:source:)` to style the casing of alternative route.
 * Fixed an issue where the route line periodically peeked out from behind the user puck even though `NavigationViewController.routeLineTracksTraversal` was enabled. ([#2737](https://github.com/mapbox/mapbox-navigation-ios/pull/2737))
 * Created the `UserHaloCourseView` similar to `UserCourseView` for approximate location on iOS 14 during the navigation to represent user location. Allow the switch between `UserHaloCourseView` and `UserCourseView` when precise mode is changed. ([#2664](https://github.com/mapbox/mapbox-navigation-ios/pull/2664))
+* Added option to show route duration callouts when previewing route alternatives ([#2734](https://github.com/mapbox/mapbox-navigation-ios/pull/2734)):
+  * `NavigationMapView.showRouteDurations(along routes: [Route]?)` to show duration annotation callouts on the map for the provided Routes.
```
1ec5, line 39:
```suggestion
  * `NavigationMapView.showRouteDurations(along:)` to show duration annotation callouts on the map for the provided Routes.
```

1ec5 (approved):

PR #2737, opened 2020-11-12
Update vanishing point calculation with a location based implementation

ShanMa1991:
### Description

This pr is to solve the inconsistency of route progress, puck and vanishing route line during the navigation, when speed is fast or when zooming in.  This is pr is mainly following the the Android implementation as [#3422](https://github.com/mapbox/mapbox-navigation-android/pull/3422) and [3661](https://github.com/mapbox/mapbox-navigation-android/pull/3661).

- [x] fix [647](https://github.com/mapbox/navigation-sdks/issues/647) 
- [x] fix the constant gap between the puck and the vanishing route line point
- [x] change the Changelog
- [x] update the calculation of the `fractionTraveled` calculation method to be synchronized with puck

### Implementation
Instead of using the sum of previous route steps’ distance as the traveled distance, and  calculating the `fractionTraveled ` based on the distance fraction, this pr stores the coordinates and route line index along the route. The distance is calculated by the coordinates instead of the line string. The `fractionTraveled ` is based on the route line index and the distance, which improve the accuracy.  It also keeps the location and `fractionTraveled`  of route line synchronized with puck to avoid the inconsistency. 

Add a NSTimer (Timer) to update the vanishing route line, slicing the `fractionTraveled ` into small pieces and updating each of them during a small time interval. It allows the smoothly transition during each 1 second update of the route progress.

### Screenshots or Gifs
From the .gif below, we could see that the vanishing route line is at the same location of the puck view, and also move smoothly when zoom in and at high speed

fixGif

/cc @mapbox/navigation-ios


ShanMa1991:
The problem so far mainly comes from two parts during navigation tests.

First, the method of `line-slice-along` methods defined in `Turf.js`  used in Android team is different from `LineString.trimmed(from:distance:)` in `Turf-swift`.  The `Turf-swift` method asks for a location and a distance as parameter, while the `Turf.js` asks for two distance as parameter. It caused some problems in dealing with the overlapping coordinates points and raise the index problem. Especially in the `updateUpcomingRoutePointIndex(routeProgress:)`, it would fail to update the `primaryRouteRemainingDistancesIndex` correctly due to the index problem. I'll keep working on the bug fixing to solve this problem.

Second, for the `routeProgress` update, the `NavigationMapView` should receive the notification of the routeProgress change to re-calculate the `primaryRouteRemainingDistancesIndex` and then the `fractionTraveled` based on the coordinates and the location. I'll keep working on this part to keep track of the `routeProgress` and the calculation.

ShanMa1991:
The constant location gap between the vanishing route line point and the puck may be caused by the calling of the `func updateUpcomingRoutePointIndex(routeProgress: )` and the calling of the `func updateTraveledRouteLine(point:)`. They are  called after the updating the `routeProgress` and  the puck location. It may have a sequence problem and the caller problem during the navigation. I'll keep working on that to fix it.

cafesilencio:
One thing I would suggest is looking at [MapRouteLineTest.kt](https://github.com/mapbox/mapbox-navigation-android/blob/master/libnavigation-ui/src/test/java/com/mapbox/navigation/ui/route/MapRouteLineTest.kt) on the Android side and using the unit tests against your implementation to see if you get the same results. The routes we test with can be found [here](https://github.com/mapbox/mapbox-navigation-android/tree/master/libnavigation-ui/src/test/resources).

ShanMa1991:
After updating the callers to receive the `routeProgress`, tested it with really long route simulation and different, The vanishing route line is synchronized with puck. And there're are some improvements to do:

1. The simulated location manager provides routeProgress with location update every 1 second. The `updateRoute` with vanishing route line update is every 1 second as well. So the end point of the vanishing routeline is jumping every 1 second from A to B. While the puck view is smoothly animated transition from A to B during this 1 second. So from vision effect, that the puck view is driving and sliding but the route line is jumping during each 1 second. 

To solve this problem, the vanishing route line could add the animating effect to allow the smoothly transition during the 1 second, similar toas the puck view.

2. Add the tests similar to [MapRouteLineTest.kt](https://github.com/mapbox/mapbox-navigation-android/blob/master/libnavigation-ui/src/test/java/com/mapbox/navigation/ui/route/MapRouteLineTest.kt) , to check the correctness of the `fractionTraveled`

LukasPaczos:
> To solve this problem, the vanishing route line could add the animating effect to allow the smoothly transition during the 1 second

Is the animation of the puck constant and completely predictable? On the Android side, the puck's transition animation is variable, so we opted to expose a listener from the Maps SDK that's invoked whenever the transition animation of the puck ticks. On each of those ticks, we can precisely update the vanishing point.

cafesilencio:
To add to what Lukas has written, on the Android side we [add a listener called OnIndicatorPositionChangedListener to the map location component.](https://github.com/mapbox/mapbox-navigation-android/blob/d2c3e0950cfd41b3d63b37b597832b83f5335eab/libnavigation-ui/src/main/java/com/mapbox/navigation/ui/route/NavigationMapRoute.java#L313) This was recently added to the Maps SDK. I assume it was added on the IOS version. See also [here.](https://github.com/mapbox/mapbox-navigation-android/blob/d2c3e0950cfd41b3d63b37b597832b83f5335eab/libnavigation-ui/src/main/java/com/mapbox/navigation/ui/route/NavigationMapRoute.java#L334) The call is consumed [here.](https://github.com/mapbox/mapbox-navigation-android/blob/d2c3e0950cfd41b3d63b37b597832b83f5335eab/libnavigation-ui/src/main/java/com/mapbox/navigation/ui/route/MapRouteLine.kt#L1145) 

During the puck's animation this is called with every point the puck is moving to.  It is called many times per second. So the vanishing point adjusts a very small amount along with the puck's movements. I haven't had a chance to review your code yet so maybe you're already implementing this. If so, disregard this. But in case it's not clear how to get a smooth vanishing point transition, this is how it's done.

ShanMa1991:
Thanks for the suggestion above! I'll try the methods as provided to see how to get a smooth vanishing point transition synchronized with the puck animation.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -671,6 +676,188 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
 
     // MARK: - Vanishing route line methods
+
+    private var primaryRoutePoints: RoutePoints?
+    private var primaryRouteLineGranularDistances: RouteLineGranularDistances?
+    private var primaryRouteRemainingDistancesIndex: Int?
+    public var newFractionTraveled: Double = 0.0
+    
+    private struct RoutePoints {
```
MaximAlien, line 685: Is it necessary to store both `nestedList` and `flatList`? Would it be possible to store only list of coordinates?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -671,6 +676,188 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    private func initPrimaryRoutePoints(route: Route) {
+        primaryRoutePoints = parseRoutePoints(route: route)
+        primaryRouteLineGranularDistances = calculateRouteGranularDistances(coordinates: primaryRoutePoints?.flatList ?? [])
+    }
+    
+    /**
+    Tranform the route data into nested arrays of legs -> steps -> coordinates.
+    The first and last point of adjacent steps overlap and are duplicated.
+    */
+    private func parseRoutePoints(route: Route) -> RoutePoints {
```
MaximAlien, line 709: What about reusing already implemented functionality which allows to create list of coordinates along the route? E.g. there is code in:
https://github.com/mapbox/mapbox-navigation-ios/blob/8bb50396f6715f36b29cc729915ee0a65404eec4/MapboxNavigation/NavigationMapView.swift#L1148-L1154 which partly does this job. We can extract that part into method and reuse it. Also as in previous comment, do we really need to create `RoutePoints`? What about just returning list of coordinates?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -671,6 +676,188 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        return RoutePoints(nestedList: nestedList, flatList: flatList)
+    }
+    
+    /**
+    Find and cache the index of the upcoming [RouteLineDistancesIndex].
+    */
+    public func updateUpcomingRoutePointIndex(routeProgress: RouteProgress) {
+        let currentLegProgress = routeProgress.currentLegProgress
+        let currentStepProgress = routeProgress.currentLegProgress.currentStepProgress
+        if currentLegProgress != nil && currentStepProgress != nil && primaryRoutePoints != nil {
```
MaximAlien, line 731: Since these properties are not optionals there is no need in additional check.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -671,6 +676,188 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            var allRemainingPoints = getSlicedLinePointsCount(currentLegProgress: currentLegProgress, currentStepProgress: currentStepProgress)
+            
+            /**
+            Add to the count of remaining points all of the remaining points on the current leg, after the current step.
+            */
+            let currentLegSteps = completeRoutePoints.nestedList[routeProgress.legIndex]
+            if currentLegProgress.stepIndex < currentLegSteps.count {
+                let startIndex = currentLegProgress.stepIndex + 1
+                let endIndex = currentLegSteps.count - 1
+                allRemainingPoints += currentLegSteps.prefix(endIndex).suffix(from: startIndex).flatMap{ $0.flatMap{ $0 } }.count
```
MaximAlien, line 745: As per [docs](https://developer.apple.com/documentation/swift/sequence/2905332-flatmap) `flatMap` was deprecated. Let's replace it with `compactMap`.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -671,6 +676,188 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    }
+    
+    private func getSlicedLinePointsCount(currentLegProgress: RouteLegProgress, currentStepProgress: RouteStepProgress) -> Int {
+        let startDistance = currentStepProgress.distanceTraveled
+        let stopDistance = currentStepProgress.step.distance
+        var slicedLine: Turf.LineString?
+        
+        /**
+        Implement the Turf.lineSliceAlong(lineString, startDistance, stopDistance) to return a sliced lineString.
+        */
```
MaximAlien, line 772: Please try to keep correct indentation for comments. To fix it you can select current comment and press <kbd>⌃</kbd> + <kbd>i</kbd>.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -671,6 +676,188 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            return newYValue
+        }
+    }
+    
+    /**
+     Updates the route line appearance from the origin point to the indicated point
+     - parameter point: current position of the puck
+     */
+    public func updateTraveledRouteLine(point: CLLocationCoordinate2D) {
+        if primaryRouteLineGranularDistances != nil && primaryRouteRemainingDistancesIndex != nil {
```
MaximAlien, line 839: To exit early we should use `guard` statement instead of `nil` checks. Also it'll allow us to stop using force-unwrapping.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -671,6 +676,188 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        } else {
+            return newYValue
+        }
+    }
+    
+    /**
+     Updates the route line appearance from the origin point to the indicated point
+     - parameter point: current position of the puck
+     */
+    public func updateTraveledRouteLine(point: CLLocationCoordinate2D) {
```
MaximAlien, line 838: I don't think that this function should be part of public API, so `public` should be removed.
```suggestion
    func updateTraveledRouteLine(_ coordinate: CLLocationCoordinate2D) {
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -671,6 +676,188 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            indexArray[index - 1] = RouteLineDistancesIndex(point: prev, distanceRemaining: distance)
+        }
+        indexArray[points.count - 1] = RouteLineDistancesIndex(point: points[points.count - 1], distanceRemaining: 0.0)
+        return RouteLineGranularDistances(distance: distance, distanceArray: indexArray.compactMap{ $0 })
+    }
+
+    /**
+    Calculates the distance between 2 points using [EPSG:3857 projection](https://epsg.io/3857).
+    */
+    private func calculateDistance(point1: CLLocationCoordinate2D, point2: CLLocationCoordinate2D) -> Double {
```
MaximAlien, line 810: I think these calculation methods should be tested either directly of indirectly.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -534,6 +534,8 @@ extension RouteMapViewController: NavigationComponent {
         }
         
         if routeLineTracksTraversal {
+            mapView.updateUpcomingRoutePointIndex(routeProgress: progress)
+            mapView.updateTraveledRouteLine(point: location.coordinate)
             mapView.updateRoute(progress)
```
MaximAlien, line 539: As discussed, lets consider moving `updateUpcomingRoutePointIndex(routeProgress:)` into `updateRoute(_:)`. It'll provide single point of entry for vanishing route line updates.

MaximAlien (commented):
Thanks for this PR. Please check initial iteration of my comments. As discussed earlier let's concentrate of following tasks:
1. Since after updating calculation algorithm user course view is in sync with vanishing route line let's try to cover implemented functionality with tests (minor fixes).
2. `NavigationMapView` becomes pretty large, do you think we can extract added functionality into separate entity or `NavigationMapView` extension?
3. Find out whether Maps SDK for iOS provides callback similar to `onIndicatorPositionChanged` as on Android.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -252,6 +252,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         
         installUserCourseView()
         showsUserLocation = false
+        
+        if let firstRoute = routes?.first {
```
MaximAlien, line 256: In case of `commonInit()` `routes` property will not be initialized yet, this means that this check will always fail and is not really needed. Let's test it to make sure and remove if it's not really needed.

ShanMa1991:
Right now the `func updateRoute(_ routeProgress:)` only receive route progress every one second, and it limited the vanishing line update in frequency. I'll use the `func updateTraveledRouteLine(point:)` to receive location change without time constraint and update route line. The `func updateUpcomingRoutePointIndex(routeProgress:)` will be responsible to receive route progress change every second to provide the `primaryRouteRemainingDistancesIndex` whenever it's needed to be used in calculating `fractionTraveled`. So in this way, the route line layer could receive updates from location without the 1 second frequency constraint.

After that, we could work on providing the location change to the update the vanishing route line. No matter from the puck view location, or from the animation and frame rate.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: I think storing the `nestedList` would be used in the  `func updateUpcomingRoutePointIndex(routeProgress:)` , because we'll need the `nestedList[routeProgress.legIndex]` to find out the remaining coordinates number based on the route progress updates. Otherwise, the route, route leg, route step, if we flatten them all, it's hard to track them when there's any route progress update with the index.

MapboxNavigation/RouteMapViewController.swift
ShanMa1991: Thanks for the suggestion! I'll remove the unuseful code.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion! I'll remove the unuseful code.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll replace the `flatMap `

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion! I'll replace the this part with the `guard let` instead of the `nil` check.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll change the function scope and the parameter name.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: yes. So in case of it failed, the `updateRoute` did a check for the `fractionTraveled` calculated from the new method. If it failed, the `fractionTraveled` is default to 0.0, and the route line will be based on the old `routeProgress.fractionTraveled`. In the test, sometimes the `initPrimaryRoutePoints(route:)` has some problem during the initialization, so the check should be necessary. I'll look into this problem.

ShanMa1991:
Because the MapSDK on iOS side doesn't have the `onIndicatorPositionChanged` as on Android. So during the 1 second, the `func updateRoute()` could not be triggered at a higher frequency as the Android to update the `routeLayer.lineGradient`. So in the next step, to keep the puck and the vanishing route line synced during the 1 second when zoom in, I'll try increase the frame rate to update the vanishing route line at a similar frequency as the puck, and use the animation rate of the puck to sliced the `fractionTraveled` into small pieces to update the  `routeLayer.lineGradient` gradually. Because the animation of the puck is trackable with a constant rate.

Besides, in the `NavigationMapViewTests`, it still needs a correct and expected `fractionTraveled` to testify the calculated result.

ShanMa1991:
The NSTimer could not repeat the function at a constant time interval and may call the action multiple times. It cause the lineGradient calculation problem, which is based on even distributed time interval. I'll work on to fix this problem.

ShanMa1991:
Now the problem is that the user puck view is updated evenly during every second, but the vanishing route line is updated with the Timer. But the Timer called the time interval not at an evenly distributed rate. So the changing puck view and the changing vanishing route line is still not synchronous when zoom pretty in. 

So the next step is to update the Timer event at a constant time interval to allow the two synced. Maybe try `CADisplayLink`

ShanMa1991:
By updating the Timer, the vanishing route line bow could implement the smoothly transition as above [above](https://github.com/mapbox/mapbox-navigation-ios/pull/2737#issue-520042216) . But because the time interval of the Timer is not evenly distributed, when zoom pretty in, the puck view and the vanishing route line may still sometimes not at the same location. And the route line casing layer also sometimes has a flash under layer.

MapboxNavigationTests/Fixtures/vanish_point_test.json
```diff
@@ -0,0 +1 @@
...,"entry":[true],"in":0}]}],"annotation":{"distance":[49.34180914849393,33.05802569090612,10.689795908138624,39.59839199650681,34.80992351675273,14.209672171473654,55.332462814681335,8.615785848371143,40.91659557025385,5.914738766868038,52.97482177741399,85.20461216655733,12.501699782507817,35.42252405667138,29.311743333674485,35.66534898263901,25.758372926868166,44.32194001964289,22.517423451854935,28.451254156899267,40.20997808786729,30.687302775025532,38.532552346674045,105.03289645548367],"congestion":["low","unknown","unknown","low","unknown","unknown","heavy","low","low","unknown","low","heavy","low","low","low","low","low","low","low","low","unknown","unknown","unknown","low"]}}],"routeOptions":{"baseUrl":"https://api.mapbox.com","user":"mapbox","profile":"driving-traffic","coordinates":[[-122.5237734,37.9753973],[-122.5264995,37.9709171]],"alternatives":true,"language":"en","continue_straight":false,"roundabout_exits":false,"geometries":"polyline6","overview":"full","steps":true,"annotations":"congestion,distance","voice_instructions":true,"banner_instructions":true,"voice_units":"imperial","access_token":"someToken","uuid":"ckd9ao6hl13a97ars2byaymo7"},"voiceLocale":"en-US"}
```
MaximAlien, line 1: Do we really need to add test data twice in both TestHelper and MapboxNavigationTests?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -251,6 +251,12 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         
         installUserCourseView()
         showsUserLocation = false
+        
+        if routeLineTracksTraversal {
+            if let firstRoute = routes?.first{
```
MaximAlien, line 256: This can be merged into one if statement.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -709,6 +720,190 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
 
     // MARK: - Vanishing route line methods
+
+    var primaryRoutePoints: RoutePoints?
+    var primaryRouteLineGranularDistances: RouteLineGranularDistances?
+    var primaryRouteRemainingDistancesIndex: Int?
+    var routeLineTracksTraversal: Bool = false
+    private var startDate = Date()
+    public var fractionTraveled: Double = 0.0
```
MaximAlien, line 729: Do we need to publicly expose `fractionTraveled`?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -709,6 +720,190 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
 
     // MARK: - Vanishing route line methods
```
MaximAlien, line 722: Would it be possible to move all vanishing route line related methods into another file? `NavigationMapView.swift` becomes pretty heavy.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -722,16 +917,35 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            let timePassedInMilliseconds = Date().timeIntervalSince(self.startDate) * 1000
+            if timePassedInMilliseconds >= 970 {
+                timer.invalidate()
+                return
+            }
+            let newFractionTraveled = self.preFractionTraveled + traveledDifference * timePassedInMilliseconds.truncatingRemainder(dividingBy: 1000) / 1000
+            let mainRouteLayerGradient = self.routeLineGradient(routeProgress.route, fractionTraveled: newFractionTraveled)
+            if mainRouteLayerGradient == nil {
+                return
+            }
```
MaximAlien, line 943: It'd be better to use `guard let` here.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -709,6 +720,190 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        return RoutePoints(nestedList: nestedList, flatList: flatList)
+    }
+    
+    /**
+    Find and cache the index of the upcoming [RouteLineDistancesIndex].
+    */
+    public func updateUpcomingRoutePointIndex(routeProgress: RouteProgress) {
+        let currentLegProgress = routeProgress.currentLegProgress
+        let currentStepProgress = routeProgress.currentLegProgress.currentStepProgress
+        if primaryRoutePoints != nil {
```
MaximAlien, line 779: Please return early here by using `guard let` statement.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -709,6 +720,190 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+            indexArray[index - 1] = RouteLineDistancesIndex(point: prev, distanceRemaining: distance)
+        }
+        indexArray[points.count - 1] = RouteLineDistancesIndex(point: points[points.count - 1], distanceRemaining: 0.0)
+        return RouteLineGranularDistances(distance: distance, distanceArray: indexArray.compactMap{ $0 })
+    }
+
+    /**
+     Calculates the distance between 2 points using [EPSG:3857 projection](https://epsg.io/3857).
+    */
+    func calculateDistance(point1: CLLocationCoordinate2D, point2: CLLocationCoordinate2D) -> Double {
```
MaximAlien, line 858: Can you implement unit-test for this method by adding pre-defined coordinates and expected distance?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -709,6 +720,189 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     }
 
     // MARK: - Vanishing route line methods
+
+    var primaryRoutePoints: RoutePoints?
+    var primaryRouteLineGranularDistances: RouteLineGranularDistances?
+    var primaryRouteRemainingDistancesIndex: Int?
+    var routeLineTracksTraversal: Bool = false
+    private var startDate = Date()
```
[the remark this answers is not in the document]
ShanMa1991: I see. Thanks for the suggestion. I'll make it local.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -709,6 +720,190 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    func initPrimaryRoutePoints(route: Route) {
+        primaryRoutePoints = parseRoutePoints(route: route)
+        primaryRouteLineGranularDistances = calculateRouteGranularDistances(coordinates: primaryRoutePoints?.flatList ?? [])
+    }
+    
+    /**
+    Tranform the route data into nested arrays of legs -> steps -> coordinates.
+    The first and last point of adjacent steps overlap and are duplicated.
+    */
+    func parseRoutePoints(route: Route) -> RoutePoints {
```
MaximAlien, line 757: Would it be possible to replace current method with one liner like this?
```
let routeCoordinates = route.legs.compactMap({ $0.steps.compactMap({ $0.shape?.coordinates }).reduce([], +) }).reduce([], +)
```

MaximAlien: We can also implement unit-test for this method to make sure it correctly creates list of coordinates.

MapboxNavigationTests/Fixtures/vanish_point_test.json
ShanMa1991: I'm not sure. But if I remove the file from the  MapboxNavigationTests, the Fixture could not find it.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll fix it.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: No, I'll fix it.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: I see. I'll work on this. Thanks for the suggestion.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion. I'll fix it

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll improve this.

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Yes. I'll add unit test for this function

MapboxNavigation/NavigationMapView.swift
ShanMa1991: I tried to use one line but it may have some problem with the nested array. Yes, I've tested it with playground and will add the unit test to this function.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -251,6 +258,10 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         
         installUserCourseView()
         showsUserLocation = false
+        
+        if routeLineTracksTraversal, let firstRoute = routes?.first {
```
MaximAlien, line 262: I think I've asked about this before, but can you make sure that this check is really needed here at all? I think `commonInit()` will never have valid `routes` since they're assigned only after calling `show(_:legIndex:)` method. This means that we can remove this part completely.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -576,6 +582,10 @@ extension RouteMapViewController: NavigationComponent {
     func navigationService(_ service: NavigationService, didRefresh routeProgress: RouteProgress) {
         mapView.show([routeProgress.route], legIndex: routeProgress.legIndex)
         if routeLineTracksTraversal {
+            mapView.updateUpcomingRoutePointIndex(routeProgress: routeProgress)
+            if let location = router.location {
+                mapView.updateTraveledRouteLine(location.coordinate)
```
MaximAlien, line 587: You can update `updateTraveledRouteLine(_:)` to accept optional and move `if let location = router.location` inside of it to be checked in `guard`.

MapboxNavigationTests/Fixtures/vanish_point_test.json
MaximAlien: There is no need to add `vanish_point_test.json` to MapboxNavigationTests I think. You can remove another copy and make sure it's only added to TestHelper target like on screenshot. Can you also rename it to be: `route-for-vanishing-route-line.json`? 
Screen Shot 2020-12-03 at 10 38 17 AM

MapboxNavigation/NavigationMapView.swift
ShanMa1991: I'll check this part again. Thanks for pointing out this problem!

MapboxNavigation/RouteMapViewController.swift
ShanMa1991: Thanks for the suggestion! I'll work on it.

MapboxNavigationTests/NavigationMapViewTests.swift
```diff
@@ -103,6 +104,57 @@ class NavigationMapViewTests: XCTestCase, MGLMapViewDelegate {
...
+    
+    func testParseRoutePoints() {
+        let route = setUpVanishingRouteLine()
+        let navigationMapView = NavigationMapView(frame: CGRect(origin: .zero, size: .iPhone6Plus), styleURL: Fixture.blankStyle)
+        
+        navigationMapView.initPrimaryRoutePoints(route: route)
+        let nestedList = navigationMapView.primaryRoutePoints!.nestedList
+        let flatList = navigationMapView.primaryRoutePoints!.flatList
+        XCTAssertEqual(nestedList.first!.count, 8)
+        XCTAssertEqual(flatList.count, 33)
```
MaximAlien, line 155: Please try not to use force-unwrapping whenever possible. In case of running tests locally they will fail instead of crashing if there are any errors. E.g.:
```swift
let nestedList = navigationMapView.primaryRoutePoints?.nestedList
let flatList = navigationMapView.primaryRoutePoints?.flatList
XCTAssertEqual(nestedList?.first?.count, 8)
XCTAssertEqual(flatList?.count, 33)
```

Please also add verification check to make sure that `navigationMapView.primaryRouteLineGranularDistances` contains data you really expect.

MapboxNavigationTests/Fixtures/vanish_point_test.json
ShanMa1991: Thanks for the suggestion. I'll work on it

MaximAlien:
You're making good progress. There are still some items I think it'd be good to take care of in scope of this PR if possible:
1. Right after starting navigation puck and route line are not in perfect sync:


2. It seems that there is regression in route refresh (I'm not 100% sure that this PR causes this, but we should attempt to fix it). To increase route refresh updates you can set `RouteControllerProactiveReroutingInterval = 10.0` before starting navigation.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -126,6 +126,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
     var altitude: CLLocationDistance
     var routes: [Route]?
     var isAnimatingToOverheadMode = false
+    var primaryRoutePoints: RoutePoints?
+    var primaryRouteLineGranularDistances: RouteLineGranularDistances?
+    var primaryRouteRemainingDistancesIndex: Int?
```
MaximAlien, line 131: Such variable is not used anywhere, let's remove it.

MapboxNavigation/VanishingRouteLine.swift
```diff
@@ -0,0 +1,372 @@
...
+            return 0
+        }
+    }
+    
+    func calculateRouteGranularDistances(coordinates: [CLLocationCoordinate2D]) -> RouteLineGranularDistances? {
+        if coordinates.isEmpty {
+            return nil
+        } else {
+            return calculateGranularDistances(coordinates)
+        }
```
MaximAlien, line 114: What about more traditional:
```suggestion
        if !coordinates.isEmpty {
            return calculateGranularDistances(coordinates)
        }
        
        return nil
```

MapboxNavigationTests/NavigationMapViewTests.swift
ShanMa1991: Thanks for the suggestion, I'll work on it.

MapboxNavigation/VanishingRouteLine.swift
```diff
@@ -0,0 +1,372 @@
...
+            if timePassedInMilliseconds >= 980 {
+                timer.invalidate()
+                return
+            }
+            let newFractionTraveled = self.preFractionTraveled + traveledDifference * timePassedInMilliseconds.truncatingRemainder(dividingBy: 1000) / 1000
+            guard let mainRouteLayerGradient = self.routeLineGradient(routeProgress.route, fractionTraveled: newFractionTraveled) else { return }
+            let mainCasingLayerStops = self.routeCasingGradient(newFractionTraveled)
+            
+            mainRouteLayer.lineGradient = mainRouteLayerGradient
+            mainRouteCasingLayer.lineGradient = mainCasingLayerStops
```
MaximAlien, line 219: Can you make sure that this part doesn't cause any side effect? E.g. what if active guidance navigation stops suddenly on high speeds? Will it cause any issues to route line gradient?

MapboxNavigation/NavigationMapView.swift
MaximAlien, line 131: Since NavigationMapView and vanishing route line is used only for one route let's rename these variables to:
```suggestion
    var routePoints: RoutePoints?
    var routeLineGranularDistances: RouteLineGranularDistances?
    var remainingDistancesIndex: Int?
```

MapboxNavigation/NavigationMapView.swift
```diff
@@ -126,6 +126,13 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+    var routeLineTracksTraversal: Bool = false
+    var fractionTraveled: Double = 0.0
+    var preFractionTraveled: Double = 0.0
+    var timer: Timer? = nil
```
MaximAlien, line 135: What about describing purpose of this timer? Feel free to add another name if you feel it describes it better.
```suggestion
    var vanishingRouteLineUpdateTimer: Timer? = nil
```

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@
   * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteStyleLayerWithIdentifier:source:)` to style alternative route.
   * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteCasingStyleLayerWithIdentifier:source:)` to style the casing of alternative route.
 
+### Vanishing route line
+
+* Added the ability to synchronize route progress, puck and vanishing route line with a location based distance calculation implementation.  ([#2737](https://github.com/mapbox/mapbox-navigation-ios/pull/2737))
```
MaximAlien, line 18: Nit:
```suggestion
* Added the ability to synchronize route progress, puck and vanishing route line with a location based distance calculation implementation. ([#2737](https://github.com/mapbox/mapbox-navigation-ios/pull/2737))
```

MapboxNavigation/VanishingRouteLine.swift
ShanMa1991: Thanks for the suggestion, I'll fix it

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll fix it

MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll fix it

MapboxNavigation/VanishingRouteLine.swift
ShanMa1991: Thanks for the suggestion. At the start of the navigation, because the `fractionTraveled` hasn't be updated, there's no available lineGradient to update. I think it can be solved by a fixer to use the `routeProgress.fractionTraveled` instead or use speed to update the lineGradient. I'll work on this problem to fix it.

MapboxNavigation/VanishingRouteLine.swift
ShanMa1991: I'll work on this problem and check how it works.

MapboxNavigation/VanishingRouteLine.swift
ShanMa1991: The refresh problem and the start of the navigation problem are both caused by the `navigationMapView.show(_ routes: [Route], legIndex: Int = 0)`. inside the function, it defines the `mainRouteLayer` and `parentLayer` with a `lineGradient` that forms from a 0.0 fractionTraveled. 

So at the start of the navigation, before it receives the routeProgress to call `func updateRoute(_ routeProgress:)`, it shows the vanishing route line with the stop at the start point.  It causes the gap between the puck view and not perfectly synced.

When refresh, the `func navigationService(_ service: NavigationService, didRefresh routeProgress: RouteProgress)` also call the `navigationMapView.show(_ routes: [Route], legIndex: Int = 0)` first to make the vanishing route line stop at the start point. Then it calls the `func updateRoute(_ routeProgress:)` to update the vanishing route line. 

To fix the problem, inside the `navigationMapView.show(_ routes: [Route], legIndex: Int = 0)`, it should replace the default `0.0` value with the `fractionTraveled`. And prepare a resonnable value of the fractionTraveled to be used.

ShanMa1991: I create a new ticket to handle this problem  in #2757

MapboxNavigation/VanishingRouteLine.swift
```diff
@@ -0,0 +1,374 @@
...
+    }
+    
+    func calculateRouteGranularDistances(coordinates: [CLLocationCoordinate2D]) -> RouteLineGranularDistances? {
+        if !coordinates.isEmpty {
+            return calculateGranularDistances(coordinates)
+        }
+        return nil
+    }
+    
+    func calculateGranularDistances(_ coordinates: [CLLocationCoordinate2D]) -> RouteLineGranularDistances {
```
MaximAlien, line 116: Shall we merge `calculateRouteGranularDistances` and `calculateGranularDistances` into one method? Verification which is performed in `calculateRouteGranularDistances` can simply move to `calculateGranularDistances`.

MapboxNavigation/VanishingRouteLine.swift
```diff
@@ -0,0 +1,374 @@
...
+    func getSlicedLinePointsCount(currentLegProgress: RouteLegProgress, currentStepProgress: RouteStepProgress) -> Int {
+        let startDistance = currentStepProgress.distanceTraveled
+        let stopDistance = currentStepProgress.step.distance
+        var slicedLine: Turf.LineString?
+        
+        /**
+         Implement the Turf.lineSliceAlong(lineString, startDistance, stopDistance) to return a sliced lineString.
+         */
+        if let lineString = currentStepProgress.step.shape {
+            if let midPoint = lineString.coordinateFromStart(distance: startDistance) {
```
MaximAlien, line 97: Please merge both conditions into one, it'd be better to omit multiple levels of checks.

MapboxNavigation/VanishingRouteLine.swift
```diff
@@ -0,0 +1,374 @@
...
+                slicedLine = lineString.trimmed(from: midPoint, distance: stopDistance - startDistance)
+            }
+        }
+        
+        if slicedLine != nil {
```
MaximAlien, line 102: Please try to minimize usage of comparisons with `nil`. This check can be replaced with `if let`, force-unwrapping should be removed as well in favor of `slicedLine.coordinates.count`.

MapboxNavigation/VanishingRouteLine.swift
ShanMa1991: Yes. Thanks for the suggestion. I'll work on it.

MapboxNavigation/VanishingRouteLine.swift
ShanMa1991: Yes. Thanks for the suggestion. I'll work on it.

MapboxNavigation/VanishingRouteLine.swift
ShanMa1991: Thanks for the suggestion. I'll fix it.

MapboxNavigation/VanishingRouteLine.swift
ShanMa1991: In the new commit, the refresh problem is solved with using the old `fractionTraveled` to form the two layers for vanishing route line. Then it receive updates from route progress, and slight change from the old `fractionTraveled` stop.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -483,15 +491,18 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
             // In case of main route there is the ability to provide custom `MGLShape` for either route or route casing by implemeting
             // `NavigationMapViewDelegate.navigationMapView(_:shapeFor:)` or `NavigationMapViewDelegate.navigationMapView(_:simplifiedShapeFor:)`.
             if index == 0 {
+                
                 let routeShape = navigationMapViewDelegate?.navigationMapView(self, shapeFor: [route]) ??
                     shape(for: route, legIndex: legIndex, isAlternateRoute: false)
                 
                 let routeSource = addRouteSource(style, identifier: routeSourceIdentifier, shape: routeShape)
                 
+                let fractionTraveledForGradient = routeLineTracksTraversal ? fractionTraveled : 0.0
```
MaximAlien, line 500: Looks like this check will not help to prevent shift at the start of navigation because `fractionTraveled` value is updated only after showing route line itself. I understand that this might be a tricky case to fix, so please create separate ticket for this, we can tackle this as a tail-work.

MapboxNavigationTests/NavigationMapViewTests.swift
```diff
@@ -103,6 +104,63 @@ class NavigationMapViewTests: XCTestCase, MGLMapViewDelegate {
...
+        let testRouteProgress: RouteProgress = RouteProgress(route: route, routeIndex: 0, options: routeOptions, legIndex: 0, spokenInstructionIndex: 0)
+        testRouteProgress.currentLegProgress = RouteLegProgress(leg: route.legs[0], stepIndex: 1, spokenInstructionIndex: 0)
+        testRouteProgress.currentLegProgress.currentStepProgress = RouteStepProgress(step: route.legs[0].steps[1], spokenInstructionIndex: 0)
+        testRouteProgress.currentLegProgress.currentStepProgress.distanceTraveled = stepTraveledDistanceSep
+
+        navigationMapView.initPrimaryRoutePoints(route: route)
+        navigationMapView.updateUpcomingRoutePointIndex(routeProgress: testRouteProgress)
+        navigationMapView.updateTraveledRouteLine(targetPoint)
+
+        //TODO: the expected fraction should be provided
```
MaximAlien, line 145: Is there anything else you'd like to add to this unit-test? If not, worth removing this TODO statement.

MaximAlien (approved):

MapboxNavigation/NavigationMapView.swift
MaximAlien: I can see that you've already created ticket in https://github.com/mapbox/mapbox-navigation-ios/issues/2757. Thanks!

MapboxNavigationTests/NavigationMapViewTests.swift
ShanMa1991: Thanks for the suggestion. I'll fix it.

PR #2741, opened 2020-11-13
Add ability to override congestion levels for specific road classes.

MaximAlien:
### Description

This PR implements the ability to override congestion levels for specific road classes.

### Implementation

Since implementation is using  Mapbox Directions property `RouteStep.segmentIndicesByIntersection` it depends on branch `vk/478-intersection-geometry`. Dependency will be updated after dependent branch is merged to `main`.

Since Android implementation is using road classes from `mapbox_streets_v8` there is slight difference between congestion segments. After iOS is done with implementing same functionality current implementation should re-tested and revised once more.

**Update:** Most recent version was updated and is currently using `MapboxStreetsRoadClass` which matches Android implementation.

### Screenshots or Gifs

Screenshots show congestion consistency comparison between iOS and Android. In this example `.unknown` congestion level for `.motorway` road class is replaced to `.low` (blue color). Currently there is slight shift of color because of gradient which is used on iOS.

Screen Shot 2020-11-13 at 4 09 31 PM

Screen Shot 2020-11-13 at 4 09 14 PM


1ec5:
OfflineRoutingTests is [failing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/2223/workflows/21e96e8c-369f-4547-9818-254b07139fb7/jobs/24153) because of a decoding error that would probably be fixed by mapbox/mapbox-directions-swift#485:

```
/Users/distiller/project/MapboxCoreNavigationTests/OfflineRoutingTests.swift:33: error: -[MapboxCoreNavigationTests.OfflineRoutingTests testOfflineDirections] : failed - Unexpected Failure: unknown(underlying: Swift.DecodingError.keyNotFound(CodingKeys(stringValue: "iso_3166_1_alpha3", intValue: nil), Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "routes", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "legs", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "admins", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0)], debugDescription: "No value associated with key CodingKeys(stringValue: \"iso_3166_1_alpha3\", intValue: nil) (\"iso_3166_1_alpha3\").", underlyingError: nil)))
/Users/distiller/project/MapboxCoreNavigationTests/OfflineRoutingTests.swift:43: error: -[MapboxCoreNavigationTests.OfflineRoutingTests testOfflineDirections] : Asynchronous wait failed: Exceeded timeout of 2 seconds, with unfulfilled expectations: "Calculate route offline".
/Users/distiller/project/MapboxCoreNavigationTests/OfflineRoutingTests.swift:46: error: -[MapboxCoreNavigationTests.OfflineRoutingTests testOfflineDirections] : failed - No route returned
```

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * Added Ukrainian localization. ([#2735](https://github.com/mapbox/mapbox-navigation-ios/pull/2735))
 * Created the `UserHaloCourseView` similar to `UserCourseView` for approximate location on iOS 14 during the navigation to represent user location. Allow the switch between `UserHaloCourseView` and `UserCourseView` when precise mode is changed. ([#2664](https://github.com/mapbox/mapbox-navigation-ios/pull/2664))
 * Added the ability to provide more detailed `Speed limit incorrect` feedback during turn-by-turn navigation. ([#2725](https://github.com/mapbox/mapbox-navigation-ios/pull/2725))
+* Added a new `NavigationMapView.trafficOverrideRoadClasses` property. For any road class in it all route segments with an `.unknown` traffic congestion level and a matching `MapboxDirections.RoadClass` will be replaced with the `.low` congestion level. ([#2741](https://github.com/mapbox/mapbox-navigation-ios/pull/2741)
```
1ec5, line 16: jazzy will only link these enumeration cases if we fully qualify them with the type.

Cartfile
```diff
@@ -2,7 +2,7 @@ binary "https://www.mapbox.com/ios-sdk/MapboxAccounts.json" ~> 2.3.0
 binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.json" ~> 22.0
 binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps/mapbox-ios-sdk-dynamic.json" ~> 6.0
 binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon-ios.json" == 7.1.2
-github "mapbox/mapbox-directions-swift" ~> 1.0
+github "mapbox/mapbox-directions-swift" "vk/478-intersection-geometry"
```
1ec5, line 5: Remember to change this dependency to `~> 1.2` for both Carthage and CocoaPods once mapbox/mapbox-directions-swift#490 lands, or change it back to `~> 1.0` and ensure that the SDK can handle an older version of MapboxDirections gracefully.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -46,6 +46,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      */
     public var tapGestureDistanceThreshold: CGFloat = 50
     
+    /**
+     A collection of road classes for which a congestion level substitution should occur.
+     
+     For any road class included in the `trafficOverrideRoadClasses`, all route segments with an `.unknown` traffic congestion level and a matching `MapboxDirections.RoadClass`
+     will be replaced with the `.low` congestion level.
```
1ec5, line 53: This seems to state that the actual route model would change, but I think what would happen is that the segment would be styled similarly to segments with the `.low` congestion level, right?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -46,6 +46,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+     */
+    public var trafficOverrideRoadClasses: [RoadClasses] = []
```
1ec5, line 55: RoadClasses is an option set, which means you can already specify multiple road classes at a time. If there’s no need to distinguish between, say, tolled motorways and other toll roads, then `RoadClasses` is a sufficient type.

Is there any need to explicitly state the substitution that should occur by making this property a dictionary of RoadClasses to RoadClasses?

1ec5: Or is the intention to preserve the order of the intersections in a parallel data structure?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1136,6 +1144,57 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        leg.steps.forEach {
+            $0.intersections?.forEach {
+                var roadClass = $0.outletRoadClasses
+                if let roadClasses = roadClass, roadClasses.isEmpty {
+                    roadClass = nil
+                }
+                
+                roadClassesInLeg.append(roadClass)
+            }
+        }
```
1ec5, line 1176: We could write this more functionally by flat-mapping legs to steps, flat-mapping steps to intersections, mapping intersections to road classes, and filtering out empty road classes.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1185,20 +1247,50 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+               roadClasses.indices.contains(index),
+               let roadClass = roadClasses[index],
+               congestionLevel == .unknown {
+                
+                for element in trafficOverrideRoadClasses {
+                    if roadClass.contains(element) {
+                        overriddenCongestionLevel = .low
+                        break
+                    }
+                }
```
1ec5, line 1282: This can also be more functional, by calling `contains(where:)` on `trafficOverrideRoadClasses`.

scripts/.gitignore
```diff
@@ -1 +0,0 @@
-Carthage/
```
1ec5, line 1: Was this intentional?

MaximAlien: Yes, this `.gitignore` was added as part of commit which Added script to export/import abbreviations. Since it's no longer here I think `.gitignore` is not needed anymore.

Cartfile
MaximAlien: Yes, will update it after https://github.com/mapbox/mapbox-directions-swift/pull/490 lands.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Yes, for that specific `RoadClasses` congestion level will be replaced from `.unknown` to `.low`. If you feel like this comment in not clear enough I'm open to suggested improvements.

MapboxNavigation/NavigationMapView.swift
MaximAlien: This is actually a possible subject of a change. Since Android implementation is currently using road classes from [mapbox-streets-v8](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/) and iOS is does not, there is a slight difference between road classes in certain road segments. 

After `Road Classification` implementation lands in Mapbox Directions we should revise this public property: e.g. in case if we decide to extend `RoadClasses` with classes from `mapbox-streets-v8` current property can look like you mention:
```suggestion
    public var trafficOverrideRoadClasses: RoadClasses
```

If we plan to add new entity for `mapbox-streets-v8` road classes we'll have to replace current property with it (either it's bitmask, dictionary or array will depend on implementation).

CHANGELOG.md
MaximAlien: Thanks, added fully qualified `CongestionLevel` enumerations.

MaximAlien:
> OfflineRoutingTests is [failing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/2223/workflows/21e96e8c-369f-4547-9818-254b07139fb7/jobs/24153) because of a decoding error that would probably be fixed by [mapbox/mapbox-directions-swift#485](https://github.com/mapbox/mapbox-directions-swift/pull/485):
> 
> ```
> /Users/distiller/project/MapboxCoreNavigationTests/OfflineRoutingTests.swift:33: error: -[MapboxCoreNavigationTests.OfflineRoutingTests testOfflineDirections] : failed - Unexpected Failure: unknown(underlying: Swift.DecodingError.keyNotFound(CodingKeys(stringValue: "iso_3166_1_alpha3", intValue: nil), Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "routes", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "legs", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "admins", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0)], debugDescription: "No value associated with key CodingKeys(stringValue: \"iso_3166_1_alpha3\", intValue: nil) (\"iso_3166_1_alpha3\").", underlyingError: nil)))
> /Users/distiller/project/MapboxCoreNavigationTests/OfflineRoutingTests.swift:43: error: -[MapboxCoreNavigationTests.OfflineRoutingTests testOfflineDirections] : Asynchronous wait failed: Exceeded timeout of 2 seconds, with unfulfilled expectations: "Calculate route offline".
> /Users/distiller/project/MapboxCoreNavigationTests/OfflineRoutingTests.swift:46: error: -[MapboxCoreNavigationTests.OfflineRoutingTests testOfflineDirections] : failed - No route returned
> ```

Yep, I've seen that. Since current PR already depends on `vk/478-intersection-geometry` we'll have to wait till both https://github.com/mapbox/mapbox-directions-swift/pull/490 and https://github.com/mapbox/mapbox-directions-swift/pull/485 land in Mapbox Directions.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Will address it when `Road Classifications` from `mapbox_streets_v8` is available. There might be no need in storing `trafficOverrideRoadClasses` array.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Compact map in second step is not needed as we need to retain `nil`s. In any case thanks for suggestions, I've modified implementation to use higher order functions.

MapboxNavigation/NavigationMapView.swift
MaximAlien: For now I've replaced `[RoadClasses]` array to be RoadClasses option set. Will revise once again when Road Classification feature is implemented.

MapboxNavigation/DayStyle.swift
```diff
@@ -25,7 +25,6 @@ extension UIColor {
     class var trafficModerate: UIColor { get { return #colorLiteral(red: 1, green: 0.5843137255, blue: 0, alpha: 1) } }
     class var trafficHeavy: UIColor { get { return #colorLiteral(red: 1, green: 0.3019607843, blue: 0.3019607843, alpha: 1) } }
     class var trafficSevere: UIColor { get { return #colorLiteral(red: 0.5607843137, green: 0.1411764706, blue: 0.2784313725, alpha: 1) } }
-    class var trafficAlternateLow: UIColor { get { return #colorLiteral(red: 0.4666666687, green: 0.7647058964, blue: 0.2666666806, alpha: 1) } }
```
MaximAlien, line 28: `trafficAlternateLow` is not used anywhere, so probably can be removed.

MaximAlien:
@1ec5, can you please take a look at this PR once more? `RoadClasses` option set was replaced with `MapboxStreetsRoadClass`, so iOS implementation is now in par with Android.

MapboxNavigation/DayStyle.swift
1ec5: Looks like any usage of it was removed in #564. However, removing it would be a backwards-incompatible change, no? Perhaps we should only deprecate it for now.

CHANGELOG.md
```diff
@@ -29,7 +29,8 @@
...
+* Added the `RouteProgress.upcomingRouteAlerts` property to track upcoming points along the route experiencing conditions that may require the user’s attention. The `UpcomingRouteAlertInfo.alert` property contains one of the following types with more details about the alert: `Incident`, `TunnelInfo`, `BorderCrossingInfo`, `TollCollection`, and `RestStop`. ([#2694](https://github.com/mapbox/mapbox-navigation-ios/pull/2694))
+* Added a new `NavigationMapView.trafficOverrideStreetsRoadClasses` property. For any road class in it all route segments with an `CongestionLevel.unknown` traffic congestion level and a matching `MapboxDirections.MapboxStreetsRoadClass` will be replaced with the `CongestionLevel.low` congestion level. ([#2741](https://github.com/mapbox/mapbox-navigation-ios/pull/2741))
```
1ec5, line 33:
```suggestion
* Added a new `NavigationMapView.trafficOverrideStreetsRoadClasses` property. For any road class in it all route segments with an `CongestionLevel.unknown` traffic congestion level and a matching `Intersection.outletMapboxStreetsRoadClass` will be replaced with the `CongestionLevel.low` congestion level. ([#2741](https://github.com/mapbox/mapbox-navigation-ios/pull/2741)) 
```

Example/ViewController.swift
```diff
@@ -408,6 +411,13 @@ class ViewController: UIViewController {
         mapView.showsUserLocation = true
         mapView.userTrackingMode = .follow
         mapView.showsUserHeadingIndicator = true
+        
+        // Use map style which allows to show road traffic.
+        mapView.styleURL = MGLStyle.navigationDayStyleURL
+        
+        // Override traffic congestion level from `.unknown` to `.low`
+        // for `MapboxStreetsRoadClass` passed in `trafficOverrideStreetsRoadClasses`.
+        mapView.trafficOverrideStreetsRoadClasses = [.motorway]
```
1ec5, line 420: Would it be confusing to include this customization by default in the example application? It would mask default behavior that we would otherwise experience during development.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -46,6 +46,14 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
      */
     public var tapGestureDistanceThreshold: CGFloat = 50
     
+    /**
+     A collection of street road classes for which a congestion level substitution should occur.
+     
+     For any street road class included in the `trafficOverrideStreetsRoadClasses`, all route segments with an `CongestionLevel.unknown` traffic congestion level and a matching `MapboxDirections.MapboxStreetsRoadClass`
+     will be replaced with the `CongestionLevel.low` congestion level.
+     */
+    public var trafficOverrideStreetsRoadClasses: Set<MapboxStreetsRoadClass>? = nil
```
1ec5, line 55: This is a difficult concept to describe succinctly in a property name. The current name doesn’t sound like a property, but it doesn’t quite sound like a method either. Based on the documentation comment, `roadClassesWithOverriddenCongestionLevels` would be the most self-explanatory name I could think of.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1035,6 +1043,30 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         return MGLShapeCollectionFeature(shapes: [mainRoute])
     }
     
+    /**
+     Returns an array of `MapboxStreetsRoadClass` objects for specific leg.
+     
+     - parameter leg: Leg of the route.
+     - returns: An array of `MapboxStreetsRoadClass` for specific `RouteLeg`. `MapboxStreetsRoadClass` will be set to `nil` if it's not present in `Intersection`.
+     */
+    func streetsRoadClasses(_ leg: RouteLeg) -> [MapboxStreetsRoadClass?] {
```
1ec5, line 1052: The `_` label isn’t warranted here because `leg` isn’t a road class. Actually, it looks like this method doesn’t use `self` anywhere, so it would be more natural to transform this method into a computed property of an extension of RouteLeg.

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1035,6 +1043,30 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
...
+        // Pick only valid segment indices for specific `Intersection` in `RouteStep`.
+        // Array of segment indexes can look like this: [0, 3, 24, 28, 48, 50, 51, 53].
+        let segmentIndices = leg.steps.compactMap({ $0.segmentIndicesByIntersection?.compactMap({ $0 }) }).reduce([], +)
```
1ec5, line 1055: I think `RouteLeg.segmentRangesByStep` may simplify this computation.

MapboxNavigation/DayStyle.swift
MaximAlien: As I can see it's part of `UIColor` extension, which is not public, so there are no end users who can rely on it. I think there is no need in deprecation and removal is safe in this case.

Example/ViewController.swift
MaximAlien: I guess you're right. To prevent any shift in results due to usage of this property I'll remove this part for sample app.

MapboxNavigation/NavigationMapView.swift
MaximAlien: Seems that `segmentRangesByStep` is not detailed enough, as it stores only lower and upper bounds. `segmentIndicesByIntersection` is more precise in this case as it stores segment indices for `Intersection`.

MapboxCoreNavigation/RouteLeg.swift
```diff
@@ -7,4 +7,25 @@ extension RouteLeg {
             result.coordinates += (step.shape?.coordinates ?? []).dropFirst()
         }
     }
+    
+    /**
+     Returns an array of `MapboxStreetsRoadClass` objects for specific leg. `MapboxStreetsRoadClass` will be set to `nil` if it's not present in `Intersection`.
+     */
+    public var streetsRoadClasses: [MapboxStreetsRoadClass?] {
```
1ec5, line 14: Also mention this new public property in the changelog.

MapboxCoreNavigation/RouteLeg.swift
```diff
@@ -7,4 +7,25 @@ extension RouteLeg {
...
+        
+        // Pick `MapboxStreetsRoadClass` in specific `Intersection` of `RouteStep`.
+        // It is expected that number of `segmentIndices` will be equal to number of `streetsRoadClassesInLeg`.
+        // Array of `MapboxStreetsRoadClass` can look like this: [Optional(motorway), ... , Optional(motorway), nil]
+        let streetsRoadClassesInLeg = steps.compactMap({ $0.intersections?.map({ $0.outletMapboxStreetsRoadClass }) }).reduce([], +)
+        
+        // Map each `MapboxStreetsRoadClass` to the amount of two adjacent `segmentIndices`.
+        // At the end amount of `MapboxStreetsRoadClass` should be equal to the last segment index.
+        let streetsRoadClasses = segmentIndices.enumerated().map({ segmentIndices.indices.contains($0.offset + 1) && streetsRoadClassesInLeg.indices.contains($0.offset) ?
+                                                                    Array(repeating: streetsRoadClassesInLeg[$0.offset], count: segmentIndices[$0.offset + 1] - segmentIndices[$0.offset]) : [] }).reduce([], +)
```
1ec5, line 27: At a glance, it seems like this calculation would be simplified by `zip`ping `segmentIndices`, but it’s hard to be sure. I’m a bit wary of the manual index arithmetic, but I trust you’ve already tested the multi-leg route edge case. I can’t think of any other edge cases at the moment.

1ec5 (approved):

MapboxCoreNavigation/RouteLeg.swift
MaximAlien: Just tried several approaches with zipping and striding, but seems that it doesn't help to make this code more elegant and simple. I propose to move on with landing this PR so that we can start testing and I can follow-up with any better solutions in future PRs.

Issue #2723, opened 2020-11-04
Automatically hide "enable precise location" banner


jill-cardamon:
[#2693 ](https://github.com/mapbox/mapbox-navigation-ios/pull/2693) prompts the user (running iOS 14) to enable precise location to navigate. Currently, the banner hides after 20 seconds. However, it would be better if the accuracy authorization banner persists until precise location is enabled, then automatically hides.

We are already able to observe the notification for a change of accuracy authorization. `StatusView.hide(delay:animated:)` needs to be exposed and used in `NavigationViewController.navigationServiceDidChangeAuthorization(_ service:didChangeAuthorizationFor:)` to hide the authorization accuracy banner at the right time.

truburt:
@jill-cardamon and this one - can close?

jill-cardamon:
yep! will close today for 1.3

1ec5:
Fixed in mapbox/mapbox-navigation-ios#2747.

PR #2747, opened 2020-11-24
Allow the use of multiple status banners

jill-cardamon:
### Description
This PR allows the use of multiple status banners by refactoring `StatusView.swift` to manage the display statuses.
- [x] Create `Status` struct to represent a status, store them in a `statuses` array, and refactor as necessary
- [x] Refactor how statuses are shown and hidden by priority
- [x] How will we deal with special values associated with a `Status`? (e.g.: simulation status banner is interactive, allows for changes in simulation speed)
- [x] Fix [#2723 ](https://github.com/mapbox/mapbox-navigation-ios/issues/2723) (hide "enable precise location" banner when precise location is switched on)
- [x] Test (StatusViewTests.swift)
- [x] Update Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * MapboxCoreNavigation can now be installed using Swift Package Manager. ([#2771](https://github.com/mapbox/mapbox-navigation-ios/pull/2771))
 * The CarPlay guidance panel now shows lane guidance. ([#1885](https://github.com/mapbox/mapbox-navigation-ios/pull/1885))
 * Old versions of routing tiles are automatically deleted from the cache to save storage space. ([#2807](https://github.com/mapbox/mapbox-navigation-ios/pull/2807))
-* Fixed an issue where lane guidance icons would indicate the wrong arrow for certain maneuvers. ([#2796](https://github.com/mapbox/mapbox-navigation-ios/pull/2796), [#2809](https://github.com/mapbox/mapbox-navigation-ios/pull/2809) 
+* Fixed an issue where lane guidance icons would indicate the wrong arrow for certain maneuvers. ([#2796](https://github.com/mapbox/mapbox-navigation-ios/pull/2796), [#2809](https://github.com/mapbox/mapbox-navigation-ios/pull/2809))
 * Fixed a crash showing a junction view. ([#2805](https://github.com/mapbox/mapbox-navigation-ios/pull/2805))
 * Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
 * Fixed swiping for right-to-left languages for the Guidance Card UI to be more intuitive. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
 * Fixed swiping for right-to-left languages for the Guidance Card UI to be more intuitive. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
 * Refactored `StatusView.swift` to allow the to use of multilpe status banners. Statuses can be added with `addNewStatus(status:)` and hidden with `hideStatus(using:)`. ([#2747](https://github.com/mapbox/mapbox-navigation-ios/pull/2747))
-* Fixed an issue where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
+* Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
 
 ## v1.2.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 ## v1.3.0
 
 * MapboxCoreNavigation can now be installed using Swift Package Manager. ([#2771](https://github.com/mapbox/mapbox-navigation-ios/pull/2771))
-* The CarPlay guidance panel now shows lane guidance. ([#2798](https://github.com/mapbox/mapbox-navigation-ios/pull/2798))
+* The CarPlay guidance panel now shows lane guidance. ([#1885](https://github.com/mapbox/mapbox-navigation-ios/pull/1885))
 * Old versions of routing tiles are automatically deleted from the cache to save storage space. ([#2807](https://github.com/mapbox/mapbox-navigation-ios/pull/2807))
-* Fixed an issue where lane guidance icons would indicate the wrong arrow for certain maneuvers. ([#2796](https://github.com/mapbox/mapbox-navigation-ios/pull/2796), [#2809](https://github.com/mapbox/mapbox-navigation-ios/pull/2809))
+* Fixed an issue where lane guidance icons would indicate the wrong arrow for certain maneuvers. ([#2796](https://github.com/mapbox/mapbox-navigation-ios/pull/2796), [#2809](https://github.com/mapbox/mapbox-navigation-ios/pull/2809) 
 * Fixed a crash showing a junction view. ([#2805](https://github.com/mapbox/mapbox-navigation-ios/pull/2805))
 * Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
 * Fixed swiping for right-to-left languages for the Guidance Card UI to be more intuitive. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 * Fixed swiping for right-to-left languages for the Guidance Card UI to be more intuitive. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
 * Refactored `StatusView.swift` to allow the to use of multilpe status banners. Statuses can be added with `addNewStatus(status:)` and hidden with `hideStatus(using:)`. ([#2747](https://github.com/mapbox/mapbox-navigation-ios/pull/2747))
 * Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
-* Refactored `StatusView.swift` to allow the to use of multilpe status banners. Statuses can be added with `addNewStatus(status:)` and hidden with `hideStatus(using:)`. ([#2747](https://github.com/mapbox/mapbox-navigation-ios/pull/2747))
+* `NavigationViewController` can now manage multiple status banners one after another. Renamed the `NavigationViewController.showStatus(title:spinner:duration:animated:interactive:)` method to `NavigationViewController.show(_:)` and added a corresponding `NavigationViewController.hide(_:)` method. Renamed the `NavigationStatusPresenter.showStatus(title:spinner:duration:animated:interactive:)` method to `NavigationStatusPresenter.show(_:)` and added a `NavigationStatusPresenter.hide(_:)` method. ([#2747](https://github.com/mapbox/mapbox-navigation-ios/pull/2747))
 
 ## v1.2.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,6 @@
 * Fixed a crash showing a junction view. ([#2805](https://github.com/mapbox/mapbox-navigation-ios/pull/2805))
 * Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
 * Fixed swiping for right-to-left languages for the Guidance Card UI to be more intuitive. ([#2724](https://github.com/mapbox/mapbox-navigation-ios/pull/2724))
-* Refactored `StatusView.swift` to allow the to use of multilpe status banners. Statuses can be added with `addNewStatus(status:)` and hidden with `hideStatus(using:)`. ([#2747](https://github.com/mapbox/mapbox-navigation-ios/pull/2747))
-* Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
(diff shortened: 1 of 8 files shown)


1ec5:
CocoaPods integration tests are [failing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/2244/workflows/0e0be1ce-8a3c-49e8-be36-3f83472b1623/jobs/24337):

```
[!] CocoaPods could not find compatible versions for pod "Mapbox-iOS-SDK":
  In snapshot (Podfile.lock):
    Mapbox-iOS-SDK (= 6.2.2, ~> 6.0)
```

CocoaPods picked up the edit to v6.3.0 in https://github.com/mapbox/mapbox-gl-native-ios/pull/549#issuecomment-733490715 but apparently not the earlier edit to v6.2.2 in https://github.com/mapbox/mapbox-gl-native-ios/pull/549#issuecomment-733270439:

```
  CDN: trunk Relative path: Specs/a/5/9/Mapbox-iOS-SDK/6.3.0/Mapbox-iOS-SDK.podspec.json modified during this run! Returning local
  CDN: trunk Relative path downloaded: all_pods_versions_3_5_2.txt, save ETag: W/"5fbd8072-29e"
```

This is basically CocoaPods/CocoaPods#10229. We could work around it by forcing the CI job to start with a fresh cache, but at this point, it might be easier to update the integration test fixture by running these commands:

https://github.com/mapbox/mapbox-navigation-ios/blob/72d78457c3e23f854672ce1f37f99cd706bc90c3/scripts/update-version.sh#L35-L37

Rmkasun-8 (approved):

MaximAlien:
Is this PR ready to be reviewed (I can see that it's marked as `Draft`)? Also it'd be good to fix outstanding conflicts with `main` branch.

jill-cardamon:
Yes. Ready to be reviewed. I have two questions if you could answer them in your review:
- Do I delete method calls in other functions and add a deprecated flag to the definition only?
- Advice on timer related issue: used a dispatch queue to make things happen when I want them to, but when I was testing, I realized that hiding (which uses a dispatch queue) takes a while. Not sure if it’s because of the loop I implemented or something else. But it takes a while for it to disappear (longer than the delay duration). I did not notice this issue in the example with the current statuses we use.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -724,21 +743,28 @@ extension NavigationViewController: NavigationServiceDelegate {
     public func navigationServiceShouldDisableBatteryMonitoring(_ service: NavigationService) -> Bool {
         return navigationComponents.allSatisfy { $0.navigationServiceShouldDisableBatteryMonitoring(service) }
     }
-    
+
```
ShanMa1991, line 746: Nit: removing extra whitespaces and cleaning the code would be great.

MapboxNavigation/StatusView.swift
```diff
@@ -117,62 +138,106 @@ public class StatusView: UIControl {
         }
     }
     
+    @available(*, deprecated, message: "Add a status using addNewStatus instead")
     public func showStatus(title: String, spinner spin: Bool = false, duration: TimeInterval, animated: Bool = true, interactive: Bool = false) {
-        show(title, showSpinner: spin, interactive: interactive)
+        // show(title, showSpinner: spin, interactive: interactive)
```
ShanMa1991, line 143: Nit: removing the unnecessary comment would be better.

MapboxNavigation/StatusView.swift
```diff
@@ -117,62 +138,106 @@ public class StatusView: UIControl {
...
         let title = String.localizedStringWithFormat(format, NumberFormatter.localizedString(from: speed as NSNumber, number: .decimal))
-        showStatus(title: title, duration: .infinity, interactive: true)
+        let simulationStatus = Status(id: title, duration: .infinity, interactive: true, priority: StatusView.Priority(rawValue: 2))
+        addNewStatus(status: simulationStatus)
     }
     
     /**
      Shows the status view with an optional spinner.
      */
-    public func show(_ title: String, showSpinner: Bool, interactive: Bool = false) {
```
MaximAlien, line 135: It'd be better to extend existing APIs instead of replacing existing ones. If you feel that newly added API is more concise previous one can be deprecated/or left without any changes.

MapboxNavigation/StatusView.swift
```diff
@@ -46,6 +46,26 @@ public class StatusView: UIControl {
...
+    public struct Status: Identifiable {
+        public var id: String
+        var spinner: Bool = false
+        let duration: TimeInterval
+        var animated: Bool = true
+        var interactive: Bool = false
+        var priority: Priority
+    }
+    
+    struct Priority: RawRepresentable {
```
MaximAlien, line 60: It'd be better to use enum for status priority.

MapboxNavigation/NavigationComponent.swift
```diff
@@ -37,7 +37,11 @@ public protocol CarPlayConnectionObserver: class {
  */
 public protocol NavigationStatusPresenter: class {
     /**
-     Shows the status view for a specified amount of time.
+     Shows a Status for a specified amount of time.
```
MaximAlien, line 40: I think such comment would be accurate for `func addNewStatus(status: StatusView.Status)`. I think it'd better to comment with `- parameter` statements in this case. Please also add docs for `addNewStatus(status:)` and `hideStatus(status:)`.

MapboxNavigation/NavigationComponent.swift
```diff
@@ -37,7 +37,11 @@ public protocol CarPlayConnectionObserver: class {
...
      */
     func showStatus(title: String, spinner: Bool, duration: TimeInterval, animated: Bool, interactive: Bool)
+    
+    func addNewStatus(status: StatusView.Status)
```
MaximAlien, line 44: I'd also rename `addNewStatus(status:)` to `func show(_ status: StatusView.Status)` and `func hideStatus(using: StatusView.Status)` to `hide(_ status: StatusView.Status)`.

MapboxNavigation/TopBannerViewController.swift
```diff
@@ -411,6 +416,7 @@ extension TopBannerViewController: NavigationComponent {
     public func navigationService(_ service: NavigationService, willEndSimulating progress: RouteProgress, becauseOf reason: SimulationIntent) {
         guard reason == .manual else { return }
         statusView.hide(delay: 0, animated: true)
+        // statusView.hide("simulating")
```
MaximAlien, line 419: Is commented part needed? If not, please remove it.

MapboxNavigation/StatusView.swift
1ec5: I think the idea was to allow for arbitrary priorities, similar to constraint priorities. But making the type public would make sense.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * The CarPlay guidance panel now shows lane guidance. ([#1885](https://github.com/mapbox/mapbox-navigation-ios/pull/1885))
 * Fixed an issue where lane guidance icons would indicate the wrong arrow for certain maneuvers. ([#2796](https://github.com/mapbox/mapbox-navigation-ios/pull/2796)) 
 * Fixed a crash showing a junction view. ([#2805](https://github.com/mapbox/mapbox-navigation-ios/pull/2805))
+* Refactored `StatusView.swift` to allow the to use of multilpe status banners. Statuses can be added with `addNewStatus(status:)` and hidden with `hideStatus(using:)`. ([#2747](https://github.com/mapbox/mapbox-navigation-ios/pull/2747))
```
1ec5, line 9: Let’s focus on the API change as it affects developers and remove the implementation details:

```suggestion
* `NavigationViewController` can now manage multiple status banners one after another. Renamed the `NavigationViewController.showStatus(title:spinner:duration:animated:interactive:)` method to `NavigationViewController.show(_:)` and added a corresponding `NavigationViewController.hide(_:)` method. Renamed the `NavigationStatusPresenter.showStatus(title:spinner:duration:animated:interactive:)` method to `NavigationStatusPresenter.show(_:)` and added a `NavigationStatusPresenter.hide(_:)` method. ([#2747](https://github.com/mapbox/mapbox-navigation-ios/pull/2747))
```

Sources/MapboxNavigation/NavigationComponent.swift
```diff
@@ -37,7 +37,9 @@ public protocol CarPlayConnectionObserver: class {
  */
 public protocol NavigationStatusPresenter: class {
     /**
-     Shows the status view for a specified amount of time.
+     Shows a Status for a specified amount of time.
      */
-    func showStatus(title: String, spinner: Bool, duration: TimeInterval, animated: Bool, interactive: Bool)
+    func addNewStatus(status: StatusView.Status)
```
1ec5, line 42: This is technically a backwards-incompatible change, because an application may expect to be able to call `showStatus(title:spinner:duration:animated:interactive:)`. In order for these changes to go into v1._x_, the old method needs to be left in place but marked as deprecated. Its implementation inside StatusView can create a Status and pass it into the new `show(_:)` method.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -462,11 +468,18 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
         UNUserNotificationCenter.current().add(notificationRequest, withCompletionHandler: nil)
     }
     
-    public func showStatus(title: String, spinner: Bool, duration: TimeInterval, animated: Bool, interactive: Bool) {
+    public func addNewStatus(status: StatusView.Status) {
```
1ec5, line 471: Similarly, this is a backwards-incompatible change, so `showStatus(title:spinner:duration:animated:interactive:)` will have to be brought back as a deprecated method.

Tests/MapboxNavigationTests/StatusViewTests.swift
```diff
@@ -0,0 +1,84 @@
...
+        XCTAssertFalse(statusView.isHidden)
+        XCTAssertEqual(self.statusView.statuses.count, 1)
+    }
+}
+
+extension StatusViewTests {
+    
+    // define statuses
+    func firstStatus() -> StatusView.Status {
+        let title1 = NSLocalizedString("FIRST_TEST_STATUS", bundle: .mapboxNavigation, value: "first test status", comment: "the first status banner used for testing")
```
1ec5, line 58: There’s more to [adding a localizable string](https://github.com/mapbox/mapbox-navigation-ios/blob/main/CONTRIBUTING.md#adding-user-facing-text) to this codebase, but for an automated test, we actually want the string to be more predictable, so it would be better to avoid `NSLocalizedString` and hard-code `first test status` as is.

Sources/MapboxNavigation/StatusView.swift
```diff
@@ -46,6 +46,33 @@ public class StatusView: UIControl {
         }
     }
     
+    var statuses: [Status] = []
+
+    public struct Status: Identifiable {
+        public var id: String
+//        let title: String
```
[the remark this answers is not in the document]
1ec5: Per https://github.com/mapbox/mapbox-navigation-ios/pull/2747#discussion_r583781129, it probably is necessary to distinguish between the ID and title, so we’d need to uncomment this property.

1ec5:
Looks like the PR branch got munged up somehow, as if it got merged then rebased. If `git reflog` says merging was the last thing you did, you could try hard-resetting to before that merge then performing just a rebase and finally force-pushing.

jill-cardamon:
The last thing there is a rebase; I didn't use a merge at all.

1ec5:
I think this branch was rebased onto an old copy of the main branch.

CHANGELOG.md
```diff
@@ -3,12 +3,13 @@
 ## v1.3.0
 
 * MapboxCoreNavigation can now be installed using Swift Package Manager. ([#2771](https://github.com/mapbox/mapbox-navigation-ios/pull/2771))
-* The CarPlay guidance panel now shows lane guidance. ([#2798](https://github.com/mapbox/mapbox-navigation-ios/pull/2798))
+* The CarPlay guidance panel now shows lane guidance. ([#1885](https://github.com/mapbox/mapbox-navigation-ios/pull/1885))
```
1ec5, line 6: This change looks like a bad merge; can you revert it? Thanks!

Sources/MapboxNavigation/StatusView.swift
```diff
@@ -46,6 +46,39 @@ public class StatusView: UIControl {
...
+    `Status` is a struct which stores information to be displayed by the `StatusView`
+     */
+    public struct Status {
+        public var identifier: String
+        let title: String
+        var spinner: Bool = false
+        let duration: TimeInterval
+        var animated: Bool = true
+        var interactive: Bool = false
+        var priority: Priority
```
1ec5, line 61: These properties also need to be public (and documented) in order for developers to set them.

Sources/MapboxNavigation/StatusView.swift
```diff
@@ -46,6 +46,39 @@ public class StatusView: UIControl {
...
+        public var rawValue: Int
+
+        public init(rawValue: Int) {
+            self.rawValue = rawValue
+        }
+//        — Highest Priority —
+//            rerouting (rawValue = 0)
+//            enable precise location (rawValue = 1)
+//            simulation banner (rawValue = 2)
+//        — Lowest Priority —
```
1ec5, line 79: The documentation comment for Priority should indicate that lower values are higher priority.

Sources/MapboxNavigation/StatusView.swift
```diff
@@ -46,6 +46,39 @@ public class StatusView: UIControl {
...
+        var animated: Bool = true
+        var interactive: Bool = false
+        var priority: Priority
+    }
+    
+    /**
+     `Priority` is used to display `Status`es by importance
+     */
+    public struct Priority: RawRepresentable {
+        public typealias RawValue = Int
```
1ec5, line 68: Since no specific values are defined, it would be simpler to define a type alias than a raw representable type alias inside a struct:

```
public typealias Priority = Int
```

Sources/MapboxNavigation/StatusView.swift
```diff
@@ -117,62 +151,111 @@ public class StatusView: UIControl {
...
+            hide(with: highestPriorityStatus, delay: highestPriorityStatus.duration)
+        }
+    }
+    
+    /**
+     Hides a given Status without hiding the status view.
+     */
+    func hide(_ status: Status?) {
+        guard let identifier = status?.identifier else { return }
+        guard let row = statuses.firstIndex(where: {$0.identifier.contains(identifier)}) else { return }
```
1ec5, line 195: This should be a simple `==` check now that identifiers are independent of labels.

1ec5 (commented):
This is tantalizingly close – I think the remaining tasks will be pretty small in nature.

Sources/MapboxNavigation/StatusView.swift
```diff
@@ -52,30 +52,57 @@ public class StatusView: UIControl {
     `Status` is a struct which stores information to be displayed by the `StatusView`
      */
     public struct Status {
+        /**
+        `identifier` is a unique string identifier for a `Status`
```
1ec5, line 56: FYI, it isn’t necessary to explicitly name the property in the property’s own documentation comment. For example, this could say “A string that uniquely identifies this status.”

Sources/MapboxNavigation/StatusView.swift
```diff
@@ -52,30 +52,57 @@ public class StatusView: UIControl {
...
     }
     
     /**
      `Priority` is used to display `Status`es by importance
+     Lower values correspond to higher priority.
      */
-    public struct Priority: RawRepresentable {
-        public typealias RawValue = Int
+    public struct Priority {
+        public typealias Priority = Int
```
1ec5, line 95: Replace the struct with the typealias. The `value` property will go away: you’ll be able to use an Int as if it were a Priority.

The previous version with the raw-representable Priority would’ve been appropriate if we wanted to define some preset values like `high` and `low` that were tied to specific values. But only a bare typealias is necessary if we’re treating it as just an ordinary number with special semantics.

1ec5 (approved):
This is good to go once the struct goes away.

PR #2748, opened 2020-11-24
v1.1.0 docs

Udumft:
Publishing `v.1.1.0` docs.


1ec5 (approved):
Good to go once MapboxNavigation v1.1.0 is pushed to CocoaPods trunk.

Udumft:
v1.1.0 was pushed to the trunk.

1ec5:
By the way, the CircleCI failure always happens on PRs targeting publisher-production because there’s no .circleci file. I wonder why CircleCi even shows up, but maybe we can silence it with a blank file or something.

PR #2759, opened 2020-12-07
Update Mapbox Navigation to v1.2.0-alpha.3.

MaximAlien:
Update Mapbox Navigation to `v1.2.0-alpha.3` along with some minor changelog fixes.


CHANGELOG.md
```diff
@@ -5,12 +5,12 @@
 ### User interface
 
 * Fixed an issue which was causing clear map button disappearance in the example app when selecting the route. ([#2718](https://github.com/mapbox/mapbox-navigation-ios/pull/2718))
-* Fixed an issue where maneuver icon was not shown after selecting specific step. ([#2722](https://github.com/mapbox/mapbox-navigation-ios/issues/2722))
+* Fixed an issue where maneuver icon was not shown after selecting specific step. ([#2728](https://github.com/mapbox/mapbox-navigation-ios/pull/2728))
 * Fixed an issue which was preventing the ability to customize the bottom banner height. ([#2705](https://github.com/mapbox/mapbox-navigation-ios/pull/2705))
 * Added method to handle paging for iOS 14 that was causing swipe problems with instructions cards ([#2755](https://github.com/mapbox/mapbox-navigation-ios/pull/2755)):
```
1ec5, line 10: Does this refer to `handlePagingforScrollToItem(indexPath:)`? If the method is intended to be used by an application, then we should document it and mention it here. Otherwise, we should remove the `public` keyword and remove this blurb in favor of the subitems below it. (That removal doesn’t necessarily have to happen ahead of this alpha release, but we might as well adjust the changelog for now.)

/cc @jill-cardamon

1ec5 (approved):

CHANGELOG.md
MaximAlien: For now updated changelog to mention only subitems. @jill-cardamon, feel free to update it if need.

jill-cardamon: Will adjust today!

PR #2763, opened 2020-12-08
Allow customizing the floating buttons

### LINKED ISSUE #2749: Allow the application to curate and customize floating map buttons
We’ve had a number of requests over the years to make the floating map buttons more configurable – to remove a button (#2102), add a button, change the look of a button (#2103), move the buttons to the other side of NavigationViewController, lay them out horizontally (#1139), and so on. It’s already possible to remove the feedback button by setting `NavigationViewController.showsReportFeedback`, but the other buttons are hard-coded.

CarPlayManager already has a hook for customizing the set of map buttons in CarPlay, via the pass-through `carPlayManager(_:mapButtonsCompatibleWith:in:for:)` method. A similar method on NavigationViewControllerDelegate would be beneficial for the main screen. We could also adapt the `MGLOrnamentPosition` approach in mapbox/mapbox-gl-native#13911 to make it easy to move the stack of buttons around.

/cc @mapbox/navigation-ios

ShanMa1991:
### Description

This pr is to add the ability to customize the floating buttons of navigation view, including adding and removing buttons from the default setting, changing the layout and position of the floating buttons.

- [x] fix #2749 
- [x] update change log
- [x] allow the floating buttons to change position between `topLeading` and `topTrailing`
- [x] allow the floating buttons to add and remove button

### Implementation

By creating the property of `floatingButtons`  and the `floatingButtonsPosition` in `NavigationView` and `RouteMapViewController`, it allows the customers to reach and edit the  `floatingButtons` through the `NavigationViewController`.

Create an enum of `MapOrnamentPosition` to control the position of speed limit view and floating buttons. It allows the customer to switch the position of the floating buttons between `topLeading` and `topTrailing`. The `NavigationViewController.floatingButtonsPosition` is default as `MapOrnamentPosition.topTrailing`.

 The `NavigationViewController.floatingButtons` is an array of `UIButton ` that supports to add or remove customized buttons. The default buttons are `FloatingButton`, which is declared with `FloatingButton.rounded(image:selectedImage:size:)` to be consistent.

For example, in the GuidanceCard UI,  by setting the following property through the navigation view controller, we could et the position of the floating buttons to at the top left, keep the middle button from original view setting and add a red button below.
```
        navigationViewController.floatingButtonsPosition = .topLeading
        let tempButton = Button(frame: CGRect(x: 100, y: 100, width: 50, height: 50))
        tempButton.backgroundColor = UIColor.red
        if let tempFloatingButton = navigationViewController.floatingButtons?[1]{
            navigationViewController.floatingButtons = [tempFloatingButton, tempButton]
        }
```
CustomizedFloatingButtons

fix #2749  
/cc @mapbox/navigation-ios


MapboxNavigation/NavigationView.swift
```diff
@@ -127,17 +155,27 @@ open class NavigationView: UIView {
...
+        for floatingButton in oldFloatingButtons {
+            floatingStackView.removeArrangedSubview(floatingButton)
+            floatingButton.removeFromSuperview()
+        }
+    }
+    
     func setupStackViews() {
-        floatingStackView.addArrangedSubviews([overviewButton, muteButton, reportButton])
+        if let buttons = floatingButtons {
+            let newFloatingButtons = Array(buttons.values)
```
1ec5, line 173: `Dictionary.values` isn’t guaranteed to be in any particular order. In the screenshot above, for example, the mute button appears below the feedback button, opposite of the default order on main. Consider using an array to store the buttons instead of a dictionary. If you need to associate each button with an ID of some sort, there are multiple possible approaches, such as making `floatingButtons` a dictionary of tuples or making the IDs properties of the respective button classes.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -200,6 +200,26 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
+    open var floatingButtonsPosition: FloatingButtonsPositions? {
+        get {
+            return mapViewController?.floatingButtonsPosition
+        }
+        set {
+            mapViewController?.floatingButtonsPosition = newValue
+        }
+    }
+    
+    open var floatingButtons: [String:Button]? {
```
1ec5, line 214: This design requires each button to have a unique ID. In general, UIKit and SwiftUI operate under the assumption that views do not have unique IDs. Instead, application code normally holds onto pointers to specific view instances in memory. In other words, if the application wants to specify a mute button, it should refer to `NavigationView.muteButton` (though that property would need to be exposed publicly through NavigationViewController).

MapboxNavigation/NavigationView.swift
ShanMa1991: The sequence of the dictionary could be fixed by replacing the `Array(Dictionary.values)` with `Dictionary.map{$0.value}`. But the problem of this design has some other problem related with the key value assigning as abobe. I would go back to the button arrays instead of the hashMap. Thanks for the suggestion!

MapboxNavigation/NavigationViewController.swift
ShanMa1991: Yes. The dictionary design has some problems related with the pointing. The array could allow the button to change the image and  the other property but still keeps its function. For example, by replacing the image and change the color, but this button still keeps the function of `mute` with pointing to the original `NavigationView.muteButton`  as the gif below:

```
        let tempButton = Button(frame: CGRect(x: 100, y: 100, width: 50, height: 50))
        tempButton.backgroundColor = UIColor.red
        if let tempFloatingButton = navigationViewController.floatingButtons?[1],
           let temp2 = navigationViewController.floatingButtons?[0] {
            tempFloatingButton.backgroundColor = UIColor.green
            tempFloatingButton.setImage(temp2.currentImage, for: .normal)
            navigationViewController.floatingButtons = [tempFloatingButton, tempButton]
        }
```
customizedMute

MapboxNavigation/NavigationViewController.swift
```diff
@@ -200,6 +200,26 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     
     var bottomViewController: ContainerViewController?
     
+    public typealias FloatingButtonsPositions = NavigationView.FloatingButtonsPositions
```
1ec5, line 203: Instead of declaring the enumeration as a nested type and exposing it at the top level via a type alias, declare the enumeration here at the top level, so no type alias is needed.

MapboxNavigation/NavigationView.swift
```diff
@@ -75,6 +75,34 @@ open class NavigationView: UIView {
     lazy var muteButton = FloatingButton.rounded(image: Images.volumeUp, selectedImage: Images.volumeOff)
     lazy var reportButton = FloatingButton.rounded(image: Images.feedback)
     
+    public enum FloatingButtonsPositions {
```
1ec5, line 78: For consistency with the map SDK’s `MGLOrnamentPosition` enumeration, consider calling this enumeration `MapOrnamentPosition` (singular). We’ll probably eventually need to use this same enumeration for other kinds of controls that are positioned over the map. For example, eventually, we may add the ability to place the floating buttons along the bottom or place the speed limit view on the same side as the floating buttons.

MapboxNavigation/NavigationView.swift
```diff
@@ -75,6 +75,34 @@ open class NavigationView: UIView {
...
+        case topRight, topLeft
```
1ec5, line 79: The enumeration and its cases should each be documented to ensure that they appear in the jazzy docset.

MapboxNavigation/NavigationViewLayout.swift
```diff
@@ -25,9 +24,17 @@ extension NavigationView {
...
         speedLimitView.topAnchor.constraint(equalTo: topBannerContainerView.bottomAnchor, constant: 10).isActive = true
-        speedLimitView.leadingAnchor.constraint(equalTo: safeLeadingAnchor, constant: 10).isActive = true
         speedLimitView.widthAnchor.constraint(equalToConstant: FloatingButton.buttonSize.width).isActive = true
         speedLimitView.heightAnchor.constraint(equalToConstant: FloatingButton.buttonSize.height).isActive = true
+        
+        switch floatingButtonsPosition {
+        case .topLeft:
+            floatingStackView.leadingAnchor.constraint(equalTo: safeLeadingAnchor, constant: 10).isActive = true
+            speedLimitView.trailingAnchor.constraint(equalTo: safeTrailingAnchor, constant: -10).isActive = true
+        default:
```
1ec5, line 34: Use `.topRight` here so that, if we ever add more cases to the enumeration, this switch statement will generate a warning about exhaustiveness so we know to update it.

MapboxNavigation/NavigationView.swift
1ec5, line 79: This PR uses `leading` and `trailing` constraints, which means the sides are swapped in right-to-left languages. To avoid confusion, consider calling the cases `topLeading` and `topTrailing`, which would be consistent with `NSRectAlignment`. (Ideally we’d just use `NSRectAlignment`, but that enumeration is only available on iOS 13 and above.)

MapboxNavigation/NavigationViewController.swift
```diff
@@ -200,6 +200,26 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
+    
+    open var floatingButtonsPosition: FloatingButtonsPositions? {
```
1ec5, line 205: Remember to add documentation about these two public properties and also mention them in the changelog.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -15,6 +15,26 @@ class RouteMapViewController: UIViewController {
     var topBannerContainerView: BannerContainerView { return navigationView.topBannerContainerView }
     var bottomBannerContainerView: BannerContainerView { return navigationView.bottomBannerContainerView }
     
+    typealias FloatingButtonsPositions = NavigationView.FloatingButtonsPositions
```
1ec5, line 18: This type alias won’t be necessary if we declare the enumeration at the top level.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -200,6 +200,26 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
+        get {
+            return mapViewController?.floatingButtonsPosition
+        }
+        set {
+            mapViewController?.floatingButtonsPosition = newValue
+        }
+    }
+    
+    open var floatingButtons: [UIButton]? {
```
1ec5, line 214: Should we require these buttons to be of type FloatingButton so that custom buttons will look and behave consistently? If so, FloatingButton and `FloatingButton.rounded(image:selectedImage:size:)` should be documented and made public.

MapboxNavigation/NavigationViewController.swift
ShanMa1991: Thanks for the suggestion, I'll make it public at a top level to avoid typealias and nested type.

MapboxNavigation/NavigationView.swift
ShanMa1991: Thanks for the suggestion, I'll update the name and add it to the jazzy

MapboxNavigation/NavigationViewLayout.swift
ShanMa1991: Thanks for the suggestion, I'll update the case name and remove the default

MapboxNavigation/NavigationViewController.swift
ShanMa1991: Thanks for the suggestion. I'll keep it to be `UIButton` to be more versatile in development, but add the reference of the   `FloatingButton.rounded(image:selectedImage:size:)` in the documentation.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -200,6 +207,32 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
+        }
+        set {
+            if let newPosition = newValue {
+                mapViewController?.floatingButtonsPosition = newPosition
+            }
+        }
+    }
+    
+    /**
+     The  floating buttons in an array of UIButton in navigation view. The default floating buttons include the overview, mute and feedback report button. The default type of the floatingButtons is `FloatingButton`, which is declared with `FloatingButton.rounded(image:selectedImage:size:)` to be consistent.
```
1ec5, line 225: We should make `FloatingButton.rounded(image:selectedImage:size:)` public and document it.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -12,6 +12,13 @@ import Mapbox
  */
 public typealias ContainerViewController = UIViewController & NavigationComponent
 
+/**
+ The position of speed limit view and floating buttons, including the overview, mute and feedback report in a navigation view.
+ */
+public enum MapOrnamentPosition {
+    case topLeading, topTrailing
```
1ec5, line 19: Remember to add a documentation comment for both of these cases. Otherwise, the jazzy docset will show an enumeration with no cases.

1ec5 (approved):
Good to merge once the FloatingButton method is public and the requested documentation is added.

MapboxNavigation/NavigationViewController.swift
ShanMa1991: Thanks for the suggestion! I'll work on this part.

MapboxNavigation/NavigationViewController.swift
ShanMa1991: Thanks for the suggestion, I'll add the two cases to documentation.

MapboxNavigation/Style.swift
```diff
@@ -65,10 +65,16 @@ open class CancelButton: Button { }
 @objc(MBDismissButton)
 open class DismissButton: Button { }
 
-/// :nodoc:
+/**
+ A floating button during the navigation view with images and default size to keep consistent.
```
1ec5, line 69: > A rounded button with an icon that is designed to float above `NavigationMapView`.

MapboxNavigation/Style.swift
```diff
@@ -94,7 +100,14 @@ open class FloatingButton: Button {
         }
     }
     
-    class func rounded<T: FloatingButton>(image: UIImage? = nil, selectedImage: UIImage? = nil, size: CGSize = FloatingButton.buttonSize) -> T {
+    /**
+     Return a `FloatingButton` with given images and size.
+     
+     - parameter image: The `UIImage` of this button.
+     - parameter selectedImage: The `UIImage` of this button when selected.
+     - parameter size: The size of this button, is default as the `FloatingButton.buttonSize` to keep consistent.
```
1ec5, line 108: > The size of this button, or `FloatingButton.buttonSize` if this argument is not specified.

MapboxNavigation/Style.swift
```diff
@@ -65,10 +65,16 @@ open class CancelButton: Button { }
...
 
-/// :nodoc:
+/**
+ A floating button during the navigation view with images and default size to keep consistent.
+ */
 @objc(MBFloatingButton)
 open class FloatingButton: Button {
-    static let buttonSize = CGSize(width: 50, height: 50)
+    /**
+     The default `CGSize` of  a `FloatingButton` in the navigation view.
```
1ec5, line 74: > The default size of a floating button.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -16,7 +16,14 @@ public typealias ContainerViewController = UIViewController & NavigationComponen
  The position of speed limit view and floating buttons, including the overview, mute and feedback report in a navigation view.
  */
 public enum MapOrnamentPosition {
-    case topLeading, topTrailing
+    /**
+     A setting of `.topLeading` will display the view during navigation on the top left in the left-to-right language, and top right in the right-to-left language.
```
1ec5, line 20: > The top-left corner when using a left-to-right language or the top-right corner when using a right-to-left language.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -16,7 +16,14 @@ public typealias ContainerViewController = UIViewController & NavigationComponen
...
+     */
+    case topLeading
+    /**
+     A setting of `.topTrailing` will display the view during navigation on the top right in the left-to-right language, and top left in the right-to-left language.
```
1ec5, line 24: > The top-right corner when using a left-to-right language or the top-left corner when using a right-to-left language.

1ec5 (approved):
Some copyediting nits, in case you have a chance to incorporate them before merging. Thanks for the quick turnaround on this feature!

MapboxNavigation/Style.swift
ShanMa1991: Thanks for the suggestion!  I've edited the documentation as suggested.

PR #2768, opened 2020-12-16
Use custom CocoaPods version for all build jobs.

MaximAlien:
### Description
Fixes build failures related to different CocoaPods versions being used for different environments on CircleCI. E.g.:
- `Xcode 12.2` uses CocoaPods `1.10.0`
- `Xcode 12.0.1` uses CocoaPods `1.10.0.rc.1`
- `Xcode 11.4.1` uses CocoaPods `1.9.1`

I'm not deeply familiar with CircleCI rules/syntax. Feel free to propose better appraches.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
 gem 'fastlane'
 gem 'snapshot'
-gem 'cocoapods', '= 1.10.0'
+gem 'cocoapods', '~> 1.10'
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 gem 'fastlane'
 gem 'snapshot'
+gem 'cocoapods', '= 1.10.0'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 gem 'fastlane'
 gem 'snapshot'
+gem 'cocoapods', '= 1.10.0'
```
1ec5, line 4: It would be useful to test against new minor and patch releases as they come out, since the purpose of the CocoaPods integration test is to verify that a normal workflow would install correctly. v1.10.0 is an important baseline for its XCFramework support, but aside from that we don’t seem to have any other technical constraints.

1ec5 (approved):
Good idea to rely on `bundle` for this fix.

Gemfile
MaximAlien: Makes sense, updated `Gemfile`.

PR #2769, opened 2020-12-16
Make `NavigationAnnotation` public.

MaximAlien:
### Description
Makes `NavigationAnnotation` public to make it easier for external developers to filter-out annotations based on class name.

### DIFF
--- a/MapboxNavigation/Style.swift
+++ b/MapboxNavigation/Style.swift
@@ -581,8 +581,10 @@ open class BottomBannerView: UIView { }
 
 open class BottomPaddingView: BottomBannerView { }
 
-/// :nodoc:
-class NavigationAnnotation: MGLPointAnnotation { }
+/**
+ `NavigationAnnotation` is an annotation, which is used to mark final destination on `NavigationMapView`.
+ */
+public class NavigationAnnotation: MGLPointAnnotation { }
 
 /// :nodoc:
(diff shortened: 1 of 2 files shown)


MapboxNavigation/Style.swift
```diff
@@ -581,8 +581,10 @@ open class BottomBannerView: UIView { }
 
 open class BottomPaddingView: BottomBannerView { }
 
-/// :nodoc:
-class NavigationAnnotation: MGLPointAnnotation { }
+/**
+ `NavigationAnnotation` is an annotation, which is used to mark final destination on `NavigationMapView`.
+ */
+public class NavigationAnnotation: MGLPointAnnotation { }
```
1ec5, line 587: If we only intend to use this annotation type for a waypoint, this would be an opportunity to rename the type to WaypointAnnotation or DestinationAnnotation. If not, we can always do it later by having WaypointAnnotation subclass NavigationAnnotation.

1ec5 (approved):

MapboxNavigation/Style.swift
MaximAlien: There might be users who already depend on `NavigationAnnotation` classname. Since it's not publicly available and we rename it to different name bugs might occur on client side. I propose to rename it later on if such need arises.

Issue #2672, opened 2020-10-01
Binary dependencies are unavailable in XCFramework format


1ec5:
The linker error in mapbox/mapbox-navigation-ios#2665 (specific to CocoaPods and Xcode 12) forced us to implement a workaround in #2667 that sets the `EXCLUDED_ARCHS` build setting not only in the MapboxCoreNavigation and MapboxNavigation framework targets but also in the developer’s application target. That gets us past the immediate issue and allows us to release v1.0.0, but it sets us up for incompatibility down the line. If another pod employs the same workaround, an application would be unable to install both pods at the same time.

The better fix is to rebuild all three binary dependencies with Xcode 12 and distribute them as XCFrameworks. Once that build is available, we can remove the workaround from the podspecs:

* [ ] Mapbox [map SDK]: mapbox/mapbox-gl-native-ios#171 mapbox/mapbox-maps-ios#48
* [x] MapboxAccounts mapbox/mapbox-accounts-ios-internal#36 - released in v2.3.1
* [x] MapboxNavigationNative mapbox/mapbox-navigation-native#2031 - released in v28.1.0

Migrating these dependencies to XCFrameworks will also allow developers to use Carthage directly without a workaround script once Carthage/Carthage#3019 is released.

/cc @mapbox/navigation-ios

1ec5:
This is what it looks like when multiple pods set `EXCLUDED_ARCHS` to different values:

```
[!] Can't merge user_target_xcconfig for pod targets: ["Mapbox-iOS-SDK", "MapboxCoreNavigation"]. Singular build setting EXCLUDED_ARCHS[sdk=iphonesimulator*] has different values.
```

It’s only a warning, but that might be because both pods happen to specify interchangeable values.

1ec5:
> It’s only a warning, but that might be because both pods happen to specify interchangeable values.

It’s a warning when integrating the pod into a real project, but when linting the podspec or pushing it to CocoaPods trunk, it results in an error: #2739.

1ec5:
Now that MapboxAccounts v2.3.1 and MapboxNavigationNative v28.1.0 are available, it should be possible to remove the arm64 workaround from MapboxCoreNavigation.podspec and apply the `user_target_xcconfig` override to just MapboxNavigation.podspec:

[...]

Issue #2629, opened 2020-09-15
Support installation via Swift Package Manager


1ec5:
The navigation SDK is currently available for installation via CocoaPods and Carthage, with some of the lighterweight dependencies (MapboxDirections, Turf, etc.) also available for installation via Swift Package Manager. SPM has a brighter long-term outlook than CocoaPods and Carthage, so we should try to support it as soon as it’s ready.

The navigation SDK can only support SPM once all its dependencies also support SPM, in particular:

* [ ] ~~Mapbox [map SDK]: mapbox/mapbox-gl-native#15241~~
* [x] MapboxMaps [map SDK]: #2794 #2808
* [x] MapboxAccounts: mapbox/mapbox-accounts-ios#1
* [x] MapboxCommon
* [x] MapboxDirections: mapbox/mapbox-directions-swift#362
* [x] MapboxMobileEvents: mapbox/mapbox-events-ios#250
* [x] MapboxNavigationNative
* [x] MapboxSpeech: mapbox/mapbox-speech-swift#25
* [x] Solar: ceeK/Solar#11 ceeK/Solar#50
* [x] Turf: mapbox/turf-swift#30

The following dependencies are also needed for the SPM build to be testable:

* [ ] [Cedar](https://github.com/cedarbdd/cedar/)
* [x] [Quick](https://github.com/Quick/Quick/)
* [x] [Nimble](https://github.com/Quick/Nimble/)
* [x] [OHHTTPStubs](https://github.com/AliSoftware/OHHTTPStubs/)
* [ ] [FBSnapshotTestCase](https://github.com/facebookarchive/ios-snapshot-test-case/)
* [ ] [SnappyShrimp](https://github.com/AndriiDoroshko/SnappyShrimp/)

/cc @mapbox/navigation-ios @frederoni

frederoni:
SPM support in MapboxMobileEvents https://github.com/mapbox/mapbox-events-ios/pull/250

1ec5:
The map SDK, MapboxCommon, and MapboxNavigationNative all require .netrc for downloading, so the changes in apple/swift-package-manager#2833 apple/swift-tools-support-core#88 will help to unblock SPM support in the navigation SDK.

1ec5:
mapbox/mapbox-directions-swift#494 tracks a major gap in code coverage when testing MapboxDirections with Swift Package Manager.

MarcoCarnevali:
@1ec5 are there any news about this?

[...]

PR #2770, opened 2020-12-17
MapboxNavigationNative v28.1

1ec5:
This is a work in progress to upgrade to MapboxNavigationNative v28.1. ~~It doesn’t quite build yet due to a missing type declaration in MapboxNavigationNative.~~

Working towards #2672 and #2629.

/cc @SiarheiFedartsou @MaximAlien

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -1,5 +1,6 @@
 import Foundation
 import CoreLocation
+import MapboxCommon
 import MapboxNavigationNative
 import MapboxMobileEvents
 import MapboxDirections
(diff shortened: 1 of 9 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import CoreLocation
+import MapboxCommon
```
1ec5, line 3: > It doesn’t quite build yet due to a missing type declaration in MapboxNavigationNative.

This import fixed the build. Apparently MapboxCommon defines the MBXExpected class that MapboxNavigationNative now uses (but only forward-declares).

MaximAlien (approved):

PR #2771, opened 2020-12-18
Enable installing Core Navigation using Swift Package Manager

1ec5:
Added a Swift Package Manager manifest for MapboxCoreNavigation.

More to come:

* [x] Make mapbox/mapbox-accounts-ios public
* [x] Release MapboxAccounts v2.3.1: mapbox/mapbox-accounts-ios#1
* [x] Make mapbox/mapbox-common-ios public
* [x] Rerelease MapboxCommon v9.1.1
* [x] Make mapbox/mapbox-navigation-native-ios public
* [x] Add MapboxCoreNavigation to Package.swift
* [x] Add MapboxCoreNavigationTests to Package.swift
* [x] Fix package resolution errors
* [x] Move source and test files to standard SPM directories
* [x] Create separate CMapboxCoreNavigation target for MBXAccounts+CoreNavigationAdditions.h, MBXAccounts+CoreNavigationAdditions.m, and MBXPeerWrapper.h
* [x] Import system libraries where they had been implicit
* [x] Add SPM build job to CircleCI configuration
* [x] Update documentation to mention SPM alongside Carthage and CocoaPods
* [x] Update release script to update SPM installation instructions

Fixes half of #2629. Depends on the creation of public, Package.swift-only repositories for MapboxAccounts, MapboxCommon, and MapboxNavigationNative.

/cc @mapbox/navigation-ios @SiarheiFedartsou @frederoni


Package.swift
```diff
@@ -0,0 +1,37 @@
...
+        .library(
+            name: "MapboxCoreNavigation",
+            targets: ["MapboxCoreNavigation"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+        .package(url: "https://github.com/mapbox/mapbox-accounts-ios/", .branch("1ec5-v2.3.1")),
+        .package(url: "https://github.com/mapbox/mapbox-directions-swift/", from: "1.2.0-alpha.3"),
+        .package(url: "https://github.com/mapbox/mapbox-events-ios/", from: "0.10.6"),
```
1ec5, line 20: Including this dependency seems to stall Xcode when resolving the dependency graph, eventually leading to the following error:

> the Package.resolved file is most likely severely out-of-date and is preventing correct resolution; delete the resolved file and try again

Package.swift
```diff
@@ -0,0 +1,37 @@
...
+        // Targets can depend on other targets in this package, and on products in packages this package depends on.
+        .target(
+            name: "MapboxCoreNavigation",
+            dependencies: [
+                "MapboxAccounts",
+                "MapboxDirections",
+                "MapboxMobileEvents",
+                "MapboxNavigationNative",
+            ],
+            path: "MapboxCoreNavigation"),
```
1ec5, line 34: This is just a workaround for the fact that I haven’t moved the sources into Sources/ and the tests into Tests/.

The presence of the Info.plist file and a few Objective-C files in the MapboxCoreNavigation folder triggers this error:

> found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
> /path/to/mapbox-navigation-ios/MapboxCoreNavigation/Info.plist
> target at '/path/to/mapbox-navigation-ios/MapboxCoreNavigation' contains mixed language source files; feature not supported

Package.swift
```diff
@@ -0,0 +1,48 @@
...
+                "MapboxCoreNavigation",
+            ]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+        .package(name: "MapboxAccounts", url: "https://github.com/mapbox/mapbox-accounts-ios/", .branch("1ec5-v2.3.1")),
+//        .package(name: "MapboxDirections", url: "https://github.com/mapbox/mapbox-directions-swift/", from: "1.2.0-alpha.3"),
+//        .package(name: "MapboxMobileEvents", url: "https://github.com/mapbox/mapbox-events-ios/", from: "0.10.6"),
+//        .package(name: "MapboxNavigationNative", url: "https://github.com/mapbox/mapbox-navigation-native-ios/", from: "28.1.0"),
```
1ec5, line 26: Uncommenting any of these dependencies tends to cause Xcode to crash:

```
Crashed Thread:        10  Dispatch queue: -[IDEExecutionEnvironment initWithWorkspaceArena:] (QOS: UNSPECIFIED)

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [53655]

Application Specific Information:
ProductBuildVersion: 12C33
Fatal error: Duplicate values for key: 'MapboxDirections': file /AppleInternal/BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1103.8.25.8/swift/stdlib/public/core/NativeDictionary.swift, line 747
```

1ec5: When it doesn’t crash, I get this error message instead:

> the Package.resolved file is most likely severely out-of-date and is preventing correct resolution; delete the resolved file and try again

But deleting Package.resolved, .swiftpm, the project’s entire DerivedData including SourcePackages, and the module cache doesn’t resolve the error.

1ec5: When it doesn’t crash or give that error, sometimes I get this error message instead:

> Showing All Messages
repository '/path/to/Library/Developer/Xcode/DerivedData/mapbox-navigation-ios-alubdfsnhjnshqajmpyxkgtmrvbf/SourcePackages/checkouts' has uncommited changes

In fact, that directory is a clean checkout of mapbox-accounts-ios@dbfa4bbd23d31e5d9cf6c68037dde91af1f30dd7 according to Git.

1ec5: The crash still occurs in Xcode 11.4 (which is using the same version of Swift anyhow) when pulling in any two dependencies:

```
Application Specific Information:
ProductBuildVersion: 12D4e
Fatal error: Duplicate values for key: 'MapboxMobileEvents': file /AppleInternal/BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1103.8.25.8/swift/stdlib/public/core/NativeDictionary.swift, line 747
```

1ec5: Per [this StackOverflow post](https://stackoverflow.com/a/63079910/4585461), changing the URLs to .git URLs fixes all these issues. Xcode seems to understand a URL to a repository’s homepage, but not always.

Tests/MapboxCoreNavigationTests/LocationTests.swift
```diff
@@ -1,6 +1,7 @@
 import XCTest
 import CoreLocation
 @testable import MapboxCoreNavigation
+#if !SWIFT_PACKAGE
 @testable import TestHelper
```
1ec5, line 5: I conditionally uncompiled any tests that depend on TestHelper. Unfortunately, that’s just about all the interesting tests in MapboxCoreNavigationTests. MapboxCoreNavigationTests had been depending on TestHelper, which depends on MapboxNavigation. We’ll eventually need to break out the MapboxNavigation-dependent parts of TestHelper into a separate target.

frederoni:
Looks good.

If xcodebuild doesn't find any xcodeproj or xcworkspace, it will fallback to Package.swift
So if you want to test the SPM integration on CI you could e.g:
```
rm -rf MapboxNavigation.xcodeproj
xcodebuild -scheme MapboxNavigation build test -destination "platform=iOS Simulator,name=iPhone 11,OS=latest"
```

1ec5:
That’s clever, thanks!

frederoni (approved):

1ec5:
[Well that’s a new one](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/2443/workflows/bb24a608-b0a1-4d58-88d6-155af86411d5/jobs/26035):

```
xcodebuild: error: Could not resolve package dependencies:
  The server SSH fingerprint failed to verify.


Exited with code exit status 74
```

Unfortunately, MapboxNavigationNative [specifies an SSH URL](https://github.com/mapbox/mapbox-navigation-native-ios/blob/545256162b516fb87eafc91cbf2bb3a55986450a/Package.swift#L21) to the MapboxCommon repository, so I guess we need to set up those credentials on CI.

MaximAlien (approved):

.circleci/config.yml
```diff
@@ -221,6 +221,7 @@ jobs:
       - when:
           condition: << parameters.spm >>
           steps:
+            - run: for ip in $(dig @8.8.8.8 github.com +short); do ssh-keyscan github.com,$ip; ssh-keyscan $ip; done 2>/dev/null >> ~/.ssh/known_hosts
```
1ec5, line 224: We can remove this workaround once rdar://problem/50686014 is fixed in Xcode or MapboxNavigationNative’s Package.swift points to an HTTPS URL for MapboxCommon, whichever comes first.

Tests/MapboxCoreNavigationTests/LocationTests.swift
1ec5: #2792

.circleci/config.yml
1ec5: Removed in #2793, now that mapbox/mapbox-navigation-native-ios#1 has been released.

1ec5: We didn’t end up being able to remove the workaround yet.

zugaldia: @1ec5 is that problematic?

PR #2773, opened 2020-12-18
Improve MapboxNavigation project structure.

MaximAlien:
### Description
This PR aims to slightly improve project structure in MapboxNavigation (UI part), remove unused code and imports.
Previously some entities were not grouped (e.g. feedback, instructions, speech etc), it made it harder to navigate throughout the code. 

### Screenshots or Gifs
Here's how new project structure looks like:
Screen Shot 2020-12-07 at 5 53 41 PM


MapboxNavigation/Cache.swift
```diff
@@ -5,7 +5,7 @@ public typealias CompletionHandler = () -> Void
 /**
  A cache consists of both in-memory and on-disk components, both of which can be reset.
  */
-@objc(MBBimodalCache)        
+@objc(MBBimodalCache)
```
1ec5, line 8: Any idea if we need this class to be exposed to the Objective-C runtime?

1ec5 (approved):
Thanks for tidying up the project!

Finally, that space in “MapboxNavigation/CarPlaySearchController+&nbsp;CPSearchTemplateDelegate.swift” is gone – that has been bugging me for a long time but I keep forgetting to fix it.

MapboxNavigation/Cache.swift
MaximAlien: Not sure, I think it should be safe to remove it.

MaximAlien: Turns out `MBBimodalCache` should be exposed. Will leave it. I've also updated indentation in MapboxNavigation target to match default settings.

PR #2774, opened 2020-12-21
Address build error with Carthage dependencies and Xcode 12.3

avi-c:
Embed Carthage-produced frameworks using copy-frameworks build phase instead of Copy Files build phase. Addresses an issue with Carthage and Xcode 12.3.

### Description
Carthage has a compatibility issue with Xcode 12.3 if frameworks are embedded in a target application using a Copy Files build phase.

### Implementation
Fixes issue by embedding frameworks via a Run Script build phase that calls the `carthage copy-frameworks` command.


1ec5 (approved):
Thanks!

1ec5:
Please cherry-pick this PR to the release-v1.2 branch after merging. Thanks!

1ec5:
#2780 applies the same change to the CarPlay example application and the Bench application.

PR #2793, opened 2021-02-02
MapboxNavigationNative v30.0.0, MapboxCommon v9.2.0

### LINKED ISSUE #2784: App build is rejected by App Store because of "Non-public API usage" in MapboxNavigationNative
**Mapbox Navigation SDK version:** v1.2.0-alpha.3

>Dear Developer,
>
>We identified one or more issues with a recent delivery for your app, "Turnout Navigation & Convoys" 4.4.2 (5). Please correct the following issues, then upload again.
>
>ITMS-90338: Non-public API usage - The app references non-public selectors in Frameworks/MapboxNavigationNative.framework/MapboxNavigationNative: appBuildNumber, getVersionString, osName, userAgentComponents. If method names in your source code match the private Apple APIs listed above, altering your method names will help prevent this app from being flagged in future submissions. In addition, note that one or more of the above APIs may be located in a static library that was included with your app. If so, they must be removed. For further information, visit the Technical Support Information at http://developer.apple.com/support/technical/
>
>Best regards,
>
>The App Store Team

I'm having the same issue as #2720, but in my case I'm using Carthage (0.36.0) so updating CocoaPods won't change anything (I'm already on version 1.10.0 which is supposed to be the version on which this issue is resolved, as suggested by @1ec5)

Any way to solve it for Carthage?

1ec5:
Upgraded to MapboxNavigationNative v30.0.0 and MapboxCommon v9.2.0. MapboxCommon v9.2.0 renames an internal method from `-getVersionString` to something else; the original selector name coincidentally matched a private method in Apple frameworks, nondeterministically triggering ITMS-90338. MapboxCommon never used or redeclared the private method at issue.

Fixes #2784. After merging, cherry-pick these changes (sans the SPM parts) into the release-v1.2 branch.

/cc @mapbox/navigation-ios @mapbox/navnative @mapbox/core-sdk


.circleci/config.yml
```diff
@@ -221,7 +221,6 @@ jobs:
       - when:
           condition: << parameters.spm >>
           steps:
-            - run: for ip in $(dig @8.8.8.8 github.com +short); do ssh-keyscan github.com,$ip; ssh-keyscan $ip; done 2>/dev/null >> ~/.ssh/known_hosts
```
1ec5, line 224: This removes the workaround in https://github.com/mapbox/mapbox-navigation-ios/pull/2771#discussion_r567167979.

Udumft: CI job seems still [failing](https://circleci.com/gh/mapbox/mapbox-navigation-ios/26105?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) on that ssh fingerprint issue. Are there any other ssh'ed dependencies? Maybe it's too early to remove this workaround?

dersim-davaod (approved):
LGTM except @Udumft's comment.

.circleci/config.yml
1ec5: I don’t think there are any other SSH’d dependencies. I wonder if the CI machine is caching something.

1ec5: I’ve restored the workaround. It should be harmless anyways.

chezzdev (approved):

1ec5:
Cherry-picked to the release-v1.2 branch in 501d73ee996c26283830226bba34a7984d626c40.

PR #2807, opened 2021-02-13
MapboxNavigationNative v31.0.1

1ec5:
Upgraded to [MapboxNavigationNative v31.0.1](https://github.com/mapbox/mapbox-navigation-native-ios/releases/tag/31.0.1).

Removed explicit routing tile version discovery now that the navigator can automatically choose the most recent version. Now MapboxNavigationNative automatically deletes old versions of routing tiles from the cache to save storage space. To do so, it needs access to the overall tile cache folder, not just the folder for an individual tile version.

/cc @mapbox/navigation-ios @mskurydin

### DIFF
--- a/Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
+++ b/Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
@@ -69,25 +69,18 @@ open class PassiveLocationDataSource: NSObject {
             return
         }
         
-        directions.fetchAvailableOfflineVersions { [weak self] (versions, error) in
-            guard let self = self, let latestVersion = versions?.first(where: { !$0.isEmpty }), error == nil else {
-                completionHandler?(error)
-                return
-            }
-            
-            do {
-                try self.configureNavigator(withTilesVersion: latestVersion)
-                completionHandler?(nil)
-            } catch {
-                completionHandler?(error)
-            }
+        do {
+            try self.configureNavigator(withTilesVersion: nil)
+            completionHandler?(nil)
+        } catch {
+            completionHandler?(error)
         }
     }
     
     /**
      Creates a cache for tiles of the given version and configures the navigator to use this cache.
      */
-    func configureNavigator(withTilesVersion tilesVersion: String) throws {
+    func configureNavigator(withTilesVersion tilesVersion: String?) throws {
         guard !isConfigured else {
             return
         }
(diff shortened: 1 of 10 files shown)


1ec5:
During turn-by-turn navigation, the navigator isn’t configured to use downloaded routing tiles. We’ve been keeping that disabled because of the potential for two navigators to run at the same time. The successor to #2593 will unify navigator usage to eliminate this risk and always enable tile downloads.

https://github.com/mapbox/mapbox-navigation-ios/blob/de166e315c9f4c4c2e88d3f95f8867b2beeb98d3/Sources/MapboxCoreNavigation/RouteController.swift#L27

The legacy offline navigation functionality continues to accept a path to the folder containing offline tiles. It’s up to the application to remove its version discovery code and pass in an unversioned path. However, this API is now deprecated and undocumented, so we don’t expect applications to need to make changes in this area.

https://github.com/mapbox/mapbox-navigation-ios/blob/de166e315c9f4c4c2e88d3f95f8867b2beeb98d3/Sources/MapboxCoreNavigation/OfflineDirections.swift#L118-L123

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -69,25 +69,18 @@ open class PassiveLocationDataSource: NSObject {
...
-                try self.configureNavigator(withTilesVersion: latestVersion)
-                completionHandler?(nil)
-            } catch {
-                completionHandler?(error)
-            }
+        do {
+            try self.configureNavigator(withTilesVersion: nil)
+            completionHandler?(nil)
+        } catch {
+            completionHandler?(error)
```
1ec5, line 76: If not for backwards compatibility, this would be a good opportunity to make the surrounding method synchronous and throwable. We can make that change in v2.0.0.

/cc @chezzdev @Udumft

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -100,16 +93,16 @@ open class PassiveLocationDataSource: NSObject {
...
-        tilesURL.appendPathComponent(tilesVersion, isDirectory: true)
         // Tiles with different versions shouldn't be mixed, it may cause inappropriate Navigator's behaviour
         try FileManager.default.createDirectory(at: tilesURL, withIntermediateDirectories: true, attributes: nil)
         configureNavigator(withURL: tilesURL, tilesVersion: tilesVersion)
     }
 
-    func configureNavigator(withURL tilesURL: URL, tilesVersion: String) {
-        let endpointConfig = TileEndpointConfiguration(directions: directions, tilesVersion: tilesVersion)
+    func configureNavigator(withURL tilesURL: URL, tilesVersion: String?) {
+        let endpointConfig = TileEndpointConfiguration(directions: directions, tilesVersion: tilesVersion, minimumDaysToPersistVersion: nil)
```
1ec5, line 102: To limit the scope of this PR and derisk the v1.3.0 release, I avoided exposing the new configuration options publicly. But it should be pretty straightforward to do so. The only holdup is understanding the options well enough to document them.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -192,13 +185,17 @@ public protocol PassiveLocationDataSourceDelegate: class {
 extension TileEndpointConfiguration {
     /**
      Initializes an object that configures a navigator to obtain routing tiles of the given version from an endpoint, using credentials that are consistent with the given directions service.
+     
+     - parameter tilesVersion: Routing tile version.
+     - parameter minimumDaysToPersistVersion: The minimum age in days that a tile version much reach before a new version can be requested from the tile endpoint.
+     - parameter removesOldLocalVersions: Whether to automatically remove older versions from the cache. By default, older versions are automatically removed.
      */
-    convenience init(directions: Directions, tilesVersion: String) {
+    convenience init(directions: Directions, tilesVersion: String?, minimumDaysToPersistVersion: Int?, removesOldLocalVersions: Bool = true) {
```
MaximAlien, line 193: I guess docs for `directions` parameter should be added as well.

MaximAlien (approved):

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
1ec5: I couldn’t think of a good way to describe what the Directions object would be used for, other than for its credentials, so I replaced it with its credentials.

Issue #2801, opened 2021-02-08
Highlight lanes based on active direction


1ec5:
Lane indication icons should be highlighted correctly even at non-four-way-intersections, where the lane indications may not precisely match the overall maneuver’s direction.

## Rationale

Beyond the basic fix for #2796 in #2798, we have an opportunity to make the logic for highlighting turn lane icons more robust. At forks and complex intersections, the lane indications may not literally match the direction of the maneuver. For example:

* [This way](https://www.openstreetmap.org/way/754401923#map=19/37.77953/-122.39832) is tagged `turn:lanes=through|through|through;right|right` based on the turn lane signs and road markings, not `left|left|left;right|right` as a routing engine might infer based on the road geometry. The maneuver onto Bryant Street is considered going straight through, whereas a left turn would end up on 4th Street; this left turn maneuver is prohibited.
* At a fork, the left lanes may go straight, in terms of both road geometry and road markings, but the Directions API still sets the maneuver modifier to `left` so that the main maneuver icon and voice instructions will say “keep left”.

After #2798 lands, these mismatches will cause the lanes in question to be completely unhighlighted, even if they can or should be used for the maneuver.

## Implementation notes

`LaneView.draw(_:)` should consider `Intersection.preferredApproachLanes` and `Intersection.usableLaneIndication` when deciding which part of a lane indication icon to highlight. mapbox/mapbox-directions-swift#516 tracks adding these properties. They’ll depend on the `mapbox/driving` profile, so much of the existing logic in `LaneView.draw(_:)` is still required as a fallback for the Valhalla-based cycling profile and the OSRM-based `mapbox/driving-traffic` profile, though I continue to see this spaghetti code as ripe for a thorough rewrite based on a pattern-matching `switch` statement.

/ref mapbox/mapbox-navigation-android#3935
/cc @mapbox/navigation-ios

1ec5:
Now that #2809 has landed, the call to `LaneIndication.ranked(favoring:)` in `LaneView.draw(_:)` should pass in the active direction if it’s available, falling back to the existing maneuver direction. Actually, that fallback might happen further up the stack in `LanesView.update(for:)`.

1ec5:
Like #2882, I don’t think we’ll be able to meaningfully test this feature until we can run MapboxNavigationTests again: #2791.

jill-cardamon:
Fixed in #2942 .

PR #2798, opened 2021-02-04
Add lane guidance to CarPlay. Fix incorrect coloring of dual-use lane icons

avi-c:
### Description

Add support for visual lane guidance instructions in CarPlay. Addresses #1885 
Fix regression that occurred during Swift migration whereby dual use lane icons were not correctly indicating which direction to follow. Fixes #2796  

### Implementation
I used a slightly different approach for adding the Lane views than that which is outlines in #1885. Instead of drawing a snapshot of the `LanesView`, I instead form my own new lane image out of images of each lane.

For the lane coloring, I altered how the lanes are colored to use the Appearance proxy mechanism for the various colors. It no longer hardcodes that `invalid` or unusable lanes are colored with an alpha of 0.4. Instead they use whatever color is specified in the appearance proxy for the LaneView. This gives the client app more control over the visual appearance.

### Screenshots or Gifs
Simulator Screen Shot - iPhone 11 - 2021-02-04 at 15 08 58
Simulator Screen Shot - iPhone 11 - 2021-02-04 at 15 09 00

Simulator Screen Shot - iPhone 11 - 2021-02-04 at 15 14 33
Simulator Screen Shot - iPhone 11 - 2021-02-04 at 15 14 38

PR #2809, opened 2021-02-16
Rewrite lane use diagram selection logic

1ec5:
Rewrote the code in LaneView that selects the LanesStyleKit drawing method. The new implementation separates the business logic from the drawing code. (Some assumptions about the repertoire of available lane configurations remains in the business logic.) In cases where the available drawing methods cannot faithfully depict the lane configuration, the closest match is drawn, biased toward the maneuver direction. The code flow goes from LaneIndication to LaneIndication.Ranking to LaneConfiguration to LanesStyleKit.

This PR comes with a battery of unit tests that’s much more comprehensive and maintainable than the existing snapshot tests, taking advantage of the new code’s testability. Some inaccuracies in the existing snapshot tests have also been fixed. (As of #2798, two test configurations specified maneuver directions that mismatched the test fixtures. The test passed despite these inaccuracies because it didn’t pass the maneuver direction into LaneView. It’s unclear whether the code prior to #2798 was really behaving correctly in these cases.)

This refactoring paves the way for #2801, in which the maneuver direction will be replaced by the active maneuver direction specified by the Directions API. It would be desirable to expose the new LaneConfiguration logic publicly so that applications would have a much more convenient starting point for customizing LaneView’s look and feel. But before we can do that, we’d need to add more lane configurations to navigation-ui-resources so we can remove the assumptions about the available assets: #576.

/cc @mapbox/navigation-ios @avi-c @abhishek1508

### DIFF
--- a/Sources/MapboxNavigation/LaneView.swift
+++ b/Sources/MapboxNavigation/LaneView.swift
@@ -1,9 +1,145 @@
 import UIKit
 import MapboxDirections
 
+extension LaneIndication {
+    static let lefts: LaneIndication = [.sharpLeft, .left, .slightLeft]
+    static let rights: LaneIndication = [.sharpRight, .right, .slightRight]
+    
+    struct Ranking: Equatable {
+        let primary: LaneIndication
+        let secondary: LaneIndication?
+    }
+    
+    /**
+     Separates the indication into primary and secondary indications with a bias toward the given maneuver direction.
+     
+     The return values are influenced by the set of available drawing methods in LanesStyleKit.
+     */
+    func ranked(favoring maneuverDirection: ManeuverDirection?) -> Ranking {
+        var indications = self
+        
+        // There are only assets for the most common configurations, so prioritize the indication that matches the maneuver direction.
+        var primaryIndication: LaneIndication
+        // Prioritize matches with the maneuver direction.
+        if indications.contains(.straightAhead) && maneuverDirection ?? .straightAhead == .straightAhead {
+            primaryIndication = .straightAhead
+        } else if indications.contains(.slightLeft) && maneuverDirection ?? .slightLeft == .slightLeft {
+            primaryIndication = .slightLeft
+        } else if indications.contains(.slightRight) && maneuverDirection ?? .slightRight == .slightRight {
+            primaryIndication = .slightRight
+        } else if (indications.contains(.left) && maneuverDirection?.isLeft ?? true) ||
+                    // No assets for sharp turns; treat them as normal turns.
+                    (indications.contains(.sharpLeft) && maneuverDirection ?? .sharpLeft == .sharpLeft) {
+            primaryIndication = .left
+        } else if (indications.contains(.right) && maneuverDirection?.isRight ?? true) ||
+                    // No assets for sharp turns; treat them as normal turns.
+                    (indications.contains(.sharpRight) && maneuverDirection ?? .sharpRight == .sharpRight) {
+            primaryIndication = .right
+        } else if indications.contains(.uTurn) && maneuverDirection ?? .uTurn == .uTurn {
+            primaryIndication = .uTurn
+        } else {
+            // The lane doesn’t match the maneuver direction, so choose the least extreme indication.
+            // Most likely the lane will appear unhighlighted anyways.
+            if indications.contains(.straightAhead) {
+                primaryIndication = .straightAhead
+            } else if indications.contains(.slightLeft) {
+                primaryIndication = .slightLeft
+            } else if indications.contains(.slightRight) {
+                primaryIndication = .slightRight
+            } else if !indications.isDisjoint(with: .lefts) {
+                primaryIndication = .left
+            } else if !indications.isDisjoint(with: .rights) {
+                primaryIndication = .right
+            } else if indications.contains(.uTurn) {
+                primaryIndication = .uTurn
+            } else {
+                // No indications to draw.
+                return Ranking(primary: [], secondary: nil)
+            }
+        }
+        
+        indications.remove(primaryIndication)
+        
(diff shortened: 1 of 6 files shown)


Tests/MapboxNavigationTests/LaneViewTests.swift
```diff
@@ -0,0 +1,194 @@
...
+                       .init(primary: .straightAhead, secondary: .right))
+        XCTAssertEqual(([.straightAhead, .slightRight] as LaneIndication).ranked(favoring: .sharpRight),
+                       .init(primary: .straightAhead, secondary: .right))
+        XCTAssertEqual(([.straightAhead, .slightRight] as LaneIndication).ranked(favoring: .slightRight),
+                       .init(primary: .right, secondary: .straightAhead))
+        
+        XCTAssertEqual(([.left, .uTurn] as LaneIndication).ranked(favoring: nil),
+                       .init(primary: .left, secondary: nil))
+        XCTAssertEqual(([.left, .uTurn] as LaneIndication).ranked(favoring: .left),
+                       .init(primary: .left, secondary: nil))
```
1ec5, line 132: Read this line as: if the lane has left and U-turn indications, and the maneuver calls for turning left, draw a left turn. (Ideally, we’d draw a U-turn-or-left icon, but we don’t have an asset for that, so we instead prefer the lane use matching the maneuver.)

Sources/MapboxNavigation/LaneView.swift
```diff
@@ -1,9 +1,145 @@
...
+    
+    case straightOrTurn(side: DrivingSide, straight: Bool, turn: Bool)
+    
+    init?(rankedIndications: LaneIndication.Ranking, drivingSide: DrivingSide) {
+        switch (rankedIndications.primary, rankedIndications.secondary) {
+        // Single-use lanes
+        case (.straightAhead, .none):
+            self = .straightOnly
+        case (.slightLeft, .none):
+            self = .slightTurnOnly(side: .left)
```
1ec5, line 103: Then, based on the ranked indications, we choose a generalized “lane configuration” that maps more directly to the set of available drawing methods.

Sources/MapboxNavigation/LaneView.swift
```diff
@@ -98,127 +236,49 @@ open class LaneView: UIView {
...
+        guard let laneConfiguration = LaneConfiguration(rankedIndications: rankedIndications, drivingSide: drivingSide) else {
+            return
+        }
+        
+        switch laneConfiguration {
+        case .straightOnly:
+            LanesStyleKit.drawLaneStraight(frame: bounds, resizing: resizing, primaryColor: appropriateColor)
+        case .slightTurnOnly(side: let side):
+            LanesStyleKit.drawLaneSlightRight(frame: bounds, resizing: resizing, primaryColor: appropriateColor,
+                                              flipHorizontally: side == .left)
```
1ec5, line 262: With the lane configurations in hand, we call into the LanesStyleKit methods, passing in a color based on lane validity (explicitly specified by the route data, not guessed by the logic above) and flipping the paths based on the configuration.

Sources/MapboxNavigation/LaneView.swift
```diff
@@ -98,127 +236,49 @@ open class LaneView: UIView {
...
+        case .turnOnly(side: let side):
+            LanesStyleKit.drawLaneRight(frame: bounds, resizing: resizing, primaryColor: appropriateColor,
+                                        flipHorizontally: side == .left)
+        case .uTurnOnly(side: let side):
+            LanesStyleKit.drawLaneUturn(frame: bounds, resizing: resizing, primaryColor: appropriateColor,
+                                        flipHorizontally: side == .left)
+        case .straightOrTurn(side: let side, straight: let straight, turn: let turn):
+            if isValid && straight && !turn {
```
1ec5, line 270: In some cases, we need to distinguish between the straight and turn parts of a drawn lane use diagram. The ranked primary and secondary indications map directly to whether the straight arrow or turn arrow should be highlighted.

Sources/MapboxNavigation/LaneView.swift
```diff
@@ -98,127 +236,49 @@ open class LaneView: UIView {
...
-                    LanesStyleKit.drawLaneSlightRight(frame: bounds, resizing: resizing, primaryColor: appropriatePrimaryColor)
-                } else {
-                    LanesStyleKit.drawLaneRight(frame: bounds, resizing: resizing, primaryColor: appropriatePrimaryColor)
-                }
-            } else if indications.isSuperset(of: [.sharpLeft]) || indications.isSuperset(of: [.left]) || indications.isSuperset(of: [.slightLeft]) {
-                if indications == .slightLeft {
-                    LanesStyleKit.drawLaneSlightRight(frame: bounds, resizing: resizing, primaryColor: appropriatePrimaryColor, flipHorizontally: true)
-                } else {
-                    LanesStyleKit.drawLaneRight(frame: bounds, resizing: resizing, primaryColor: appropriatePrimaryColor, flipHorizontally: true)
-                }
```
1ec5, line 190: Overall, the logic shouldn’t differ too much from what the old logic was _trying_ to do. But now we’re separating the process into discrete steps and conflating the sharp and slight turns using some shorthand extensions on LaneIndication and ManeuverDirection; these changes make it easier to spot gaps in the logic, of which there were many.

1ec5 (commented):
Here’s a high-level walkthrough of the new process for choosing a lane use diagram to draw:

avi-c (approved):
The logic here looks correct to me and we've discussed it a bunch so I think it is correct. I tested the many variations I have encountered that have caused issues and so far everything appears fixed here. They draw correctly based on the assets we do have today. Will be nice to get a full set of configurations soon so we can call this complete.

Issue #2820, opened 2021-02-20
TestHelper should not depend on MapboxNavigation


1ec5:
TestHelper depends on MapboxNavigation, which has some awkward consequences with SPM:

* MapboxCoreNavigationTests ends up depending on MapboxNavigation, which forces us to disable large swaths of  MapboxCoreNavigationTests: #2792.
* Bench depends on both MapboxNavigation and TestHelper, which also depends on MapboxNavigation. A static framework can’t be linked twice like that.

We should remove TestHelper’s dependency on MapboxNavigation. The only thing that requires MapboxNavigation is Plotter, so perhaps we could change it to plot routes on a static map using MapboxStatic.swift.

/cc @mapbox/navigation-ios

1ec5:
First step is to replace usage of RoutePlotter in RouteControllerSnapshotTests and SimulationLocationManagerTests with a simple check that the route has the right coordinates. Then the only usage of RoutePlotter is in ManeuverArrowTests, which is part of MapboxNavigationTests, so we can move NavigationPlotter.swift from the TestHelper target to the MapboxNavigationTests target and remove TestHelper’s dependency on MapboxNavigation (and the map SDK).

1ec5:
Fixed in mapbox/mapbox-navigation-ios#2861.

Issue #2285, opened 2019-12-11
Remove optional dependency on MapboxGeocoder.swift for search on CarPlay


1ec5:
When installing MapboxNavigation using CocoaPods, any functionality guarded by `canImport(MapboxGeocoder)` is unavailable, namely `RecentItem` and CarPlaySearchController’s conformance to the CPSearchTemplateDelegate protocol: https://github.com/mapbox/mapbox-navigation-ios/issues/1839#issuecomment-438789900. Developers using CocoaPods are forced to copy the entirety of CarPlaySearchController+&nbsp;CPSearchTemplateDelegate.swift and RecentItem.swift into their applications, plus the `carplay_search` asset, in order to implement a search interface powered by MapboxGeocoder.swift in CarPlay.

Until the underlying issue with CocoaPods is resolved or worked around, we can streamline the steps to implement that MapboxGeocoder.swift-powered interface. CarPlaySearchControllerDelegate should be more granular, with hooks for kicking off a geocoding request and receiving the response. Most of the implementation can remain in the navigation SDK, but it should rely on types vended by required libraries, so any GeocodedPlacemark usage should be replaced by Waypoint or perhaps an MGLPointAnnotation subclass akin to MapKit’s MKMapItem class.

The developer would need to write some minimal code to configure the search and convert its responses into waypoints, but these are exactly the hooks a developer would want for customization purposes anyways. We would need to document appropriate default options to avoid surprises like a missing `focalLocation` or a missing `locale`. In the future, once we implement a preview view controller (#808), it may make sense to formally depend on MapboxGeocoder.swift and vend those defaults at that time.

/cc @mapbox/navigation-ios @mapbox/search-sdk

1ec5:
The dependency will be completely removed between #2999 and #3183.

PR #2821, opened 2021-02-20
Migrate application targets to SPM

### LINKED ISSUE #2819: Crash playing reroute sound (SPM only) — RouteVoiceController.init(navigationService:speechSynthesizer:accessToken:host:)
As of #2808, if MapboxNavigation has been installed using SPM, it crashes trying to play the reroute sound when the user goes off-route:

```
#0	0x000000010348ba80 in _swift_runtime_on_report ()
#1	0x00000001035060e3 in _swift_stdlib_reportFatalErrorInFile ()
#2	0x0000000103195f30 in closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x0000000103195a87 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#4	0x0000000103195763 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#5	0x00000001031953e1 in _assertionFailure(_:_:file:line:flags:) ()
#6	0x000000010052e1d5 in RouteVoiceController.init(navigationService:speechSynthesizer:accessToken:host:) at /path/to/mapbox-navigation-ios/Sources/MapboxNavigation/RouteVoiceController.swift:73
#7	0x000000010052dc0c in RouteVoiceController.__allocating_init(navigationService:speechSynthesizer:accessToken:host:) ()
#8	0x00000001004e6032 in NavigationViewController.init(for:routeIndex:routeOptions:navigationOptions:) at /path/to/mapbox-navigation-ios/Sources/MapboxNavigation/NavigationViewController.swift:286
#9	0x00000001004e473a in NavigationViewController.__allocating_init(for:routeIndex:routeOptions:navigationOptions:) ()
#10	0x000000010033332f in ViewController.navigationViewController(navigationService:) at /path/to/mapbox-navigation-ios/Example/ViewController.swift:396
#11	0x000000010032c96f in ViewController.startBasicNavigation() at /path/to/mapbox-navigation-ios/Example/ViewController.swift:216
#12	0x000000010032b53f in closure #1 in ViewController.presentActionsAlertController() at /path/to/mapbox-navigation-ios/Example/ViewController.swift:168
```

https://github.com/mapbox/mapbox-navigation-ios/blob/99e1236b2fd04df1202c86710494e9b2b58bf375/Sources/MapboxNavigation/RouteVoiceController.swift#L73

The root cause seems to be that the Package.swift doesn’t list reroute-sound or any of the other assets in Assets.xcassets as bundle resources:

https://github.com/mapbox/mapbox-navigation-ios/blob/99e1236b2fd04df1202c86710494e9b2b58bf375/Package.swift#L50-L59

/cc @mapbox/navigation-ios

1ec5:
Migrated the Example, Example-CarPlay, and Bench from Carthage-built frameworks to static frameworks built by SPM from the colocated package.

Fixed a bunch of crashes due to bundles not being found. Since SPM builds static frameworks, the usual `Bundle(for:)` trick doesn’t work.

Static linking also breaks `canImport` weak linking, just like with CocoaPods, so even an application that explicitly depends on MapboxGeocoder wouldn’t enable the conditionally linked MapboxGeocoder usage in MapboxNavigation to build. For now, this PR strongly links MapboxGeocoder to fix those build errors, but #2285 tracks removing the dependency altogether. Similarly, Bench doesn’t build because of static linking: #2820.

Fixes #2819. Depends on #2808.

/cc @mapbox/navigation-ios


Sources/MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -33,7 +33,11 @@ extension Bundle {
      The Mapbox Core Navigation framework bundle.
      */
     public class var mapboxCoreNavigation: Bundle {
+        #if SWIFT_PACKAGE
+        return .module
+        #else
         return Bundle(for: RouteController.self)
+        #endif
```
1ec5, line 40: This change is fine, but I’m a little concerned about `mapboxNavigationIfInstalled` below. I don’t think there’s anything like `Bundle.module` for discovering a _different_ module’s bundle.

MaximAlien: I think I'm getting crash in the same place when starting active guidance navigation.

1ec5:
> Since SPM builds static frameworks

I tried setting `type: .dynamic` for MapboxNavigation in Package.swift, but it results in an empty framework with nothing but an Info.plist. 🤷‍♂️ If we could get SPM to generate a dynamic framework for MapboxNavigation, then we’d be much closer to getting Carthage to build MapboxNavigation despite its dependencies coming from SPM.

Sources/MapboxCoreNavigation/BundleAdditions.swift
MaximAlien: Looks like force unwrapping shouldn't be used when accessing Bundle.mapboxCoreNavigation. E.g. this one crashes as well: 
```
let sdkVersion: String = String(describing: Bundle.mapboxCoreNavigation.object(forInfoDictionaryKey: "CFBundleShortVersionString")!)
```

1ec5:
These error messages about the binary dependences are very prevalent; I don’t think I’ve ever opened the project without seeing them and having to clear package caches:

```
artifact of binary target 'MapboxAccounts' failed extraction: The operation couldn’t be completed. (TSCBasic.StringError error 1.)

artifact of binary target 'MapboxMaps' failed extraction: The operation couldn’t be completed. (TSCBasic.StringError error 1.)

artifact of binary target 'MapboxCommon' failed extraction: The operation couldn’t be completed. (TSCBasic.StringError error 1.)

artifact of binary target 'MapboxNavigationNative' failed extraction: The operation couldn’t be completed. (TSCBasic.StringError error 1.)

artifact of binary target 'MapboxCoreMaps' failed extraction: The operation couldn’t be completed. (TSCBasic.StringError error 1.)
```

MaximAlien:
Overall looks good, just got additional crash because of force unwrapping in:
```
mobileEventsManager.initialize(withAccessToken: accessToken, userAgentBase: userAgent, hostSDKVersion: String(describing: Bundle.mapboxCoreNavigation.object(forInfoDictionaryKey: "CFBundleShortVersionString")!))
```

After removing `!` was able to run example successfully.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -2050,17 +1910,16 @@
 				C53F2EE320EBC95600D9798F /* Sources */,
 				C53F2EE920EBC95600D9798F /* Frameworks */,
 				C53F2EF220EBC95600D9798F /* Resources */,
-				C53F2EF920EBC95600D9798F /* Embed Frameworks */,
 				C53F2F0320EBC95600D9798F /* Apply Mapbox Access Token */,
-				DA3D112B25B0AB0100A6AB8C /* Copy Frameworks with Carthage */,
```
MaximAlien, line 2055: Shalle we remove `input.xcfilelist` and `output.xcfilelist` as well?

MaximAlien (approved):

MaximAlien:
`Bench` target cannot be built with errors like:
```
Swift package product 'MapboxGeocoder' is linked as a static library by 'Bench' and 'TestHelper'. This will result in duplication of library code.
```

1ec5:
Yes, unfortunately, Bench will fail until #2820 is fixed.

PR #2823, opened 2021-02-22
Navigator Tiles management

Udumft:
This PR solves `Navigator` instance management by introducing single entry point and sharing existing instances. It also ensures that all Navigators are properly configured when initialized.


Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -65,66 +56,8 @@ open class PassiveLocationDataSource: NSObject {
...
-            } catch {
-                completionHandler?(error)
-            }
-        }
-    }
-    
-    /**
-     Creates a cache for tiles of the given version and configures the navigator to use this cache.
-     */
-    func configureNavigator(withTilesVersion tilesVersion: String) throws {
```
1ec5, line 93: The original idea was to defer the navigator’s creation and configuration until the map SDK starts consuming location updates from PassiveLocationManager. Now that configuration happens earlier when attaching the PassiveLocationManager to the MapView. The nice thing about waiting until `startUpdatingLocation()` is that we could make it completely asynchronous without having to synchronize within a factory method.

Sources/MapboxCoreNavigation/NavigatorProvider.swift
```diff
@@ -0,0 +1,95 @@
...
+    let history: HistoryRecorderHandle
+    
+    init(navigator: Navigator, history: HistoryRecorderHandle) {
+        self.navigator = navigator
+        self.history = history
+    }
+}
+
+/// A fabric, used for obtaining and/or configuring `Navigator`.
+public class NavigatorProvider {
```
1ec5, line 19: Does this class need to be public? It seems to only be used inside private implementations.

MaximAlien: Since `Navigator` is used in `RouteController` and `PassiveLocationDataSource` and is not exposed publicly I've modified it and its properties/methods to be internal.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -21,14 +21,9 @@ open class PassiveLocationDataSource: NSObject {
...
         self.directions = directions
-        
-        let settingsProfile = SettingsProfile(application: ProfileApplication.kMobile, platform: ProfilePlatform.KIOS)
-        self.navigator = try! Navigator(profile: settingsProfile,
-                                        config: NavigatorConfig(),
-                                        customConfig: "",
-                                        tilesConfig: TilesConfig())
 
         self.systemLocationManager = systemLocationManager ?? NavigationLocationManager()
+        self.navigatorWithHistory = NavigatorProvider.sharedWeakNavigator()
```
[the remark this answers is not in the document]
MaximAlien: @1ec5, `Navigator` shared instance logic is now part of `Navigator.swift` extension. `Navigator.shared`, `Navigator.tilesVersion`, `Navigator.tilesURL`, `Navigator.enableHistoryRecorder()`, `Navigator.disableHistoryRecorder()`, `Navigator.history()` is exposed to outer world. Please let me know what you think.

MaximAlien:
@1ec5, can you please give it another pass. After adding proposed changes passive/active navigation seems to be working fine, but it'd be good to add some unit-tests (I think this is currently blocked).

Sources/MapboxCoreNavigation/Navigator.swift
```diff
@@ -1,7 +1,110 @@
...
+        return navigatorWithHistoryRecorder.1
+    }()
+    
+    private static let internalQueue = DispatchQueue(label: "com.mapbox.coreNavigation.navigator.internalQueue")
+    
+    /**
+     Provides a new or an existing one `Navigator` instance along with related `HistoryRecorderHandle`,
+     satisfying provided configuration (`tilesVersion` and `tilesURL`).
+     */
+    private static let navigatorWithHistoryRecorder: (Navigator, HistoryRecorderHandle) = {
```
MaximAlien, line 54: I hope that tuple usage is fine in this case. Not sure it's worth creating another entity for this similar to `NavigatorWithHistory`.

Sources/MapboxCoreNavigation/Navigator.swift
```diff
@@ -1,7 +1,110 @@
...
+                                           runLoopExecutor: runloopExecutor,
+                                           cache: cacheHandle,
+                                           historyRecorder: historyRecorder)
+            }
+        }
+        
+        return (navigator, historyRecorder)
+    }()
+    
     func status(at timestamp: Date) -> NavigationStatus {
```
MaximAlien, line 108: I've removed `navigatorIsActive()` as it wasn't used anywhere.

Sources/MapboxCoreNavigation/Navigator.swift
1ec5: 5ac8c9a5ddebd4f90f49ab391c54ac9ff57707df for #2808 defines a similar struct, NavigatorResources, which is more flexible than a tuple. I think that flexibility will be necessary for #2834.

Sources/MapboxCoreNavigation/Navigator.swift
```diff
@@ -1,7 +1,110 @@
...
+     further changes to this property will have no effect.
+     */
+    static var tilesURL: URL? = nil
+    
+    /**
+     Shared instance on `Navigator`.
+     */
+    static let shared: Navigator = {
+        return navigatorWithHistoryRecorder.0
+    }()
```
1ec5, line 30: I should’ve been clearer in https://github.com/mapbox/mapbox-navigation-ios/pull/2823#discussion_r582147652: the suggestion was to use composition/encapsulation rather than an extension. `MapboxCoreNavigation.Navigator` would be a singleton containing a `MapboxNavigationNative.Navigator`. This way the raw methods defined by MapboxNavigationNative won’t be exposed to the developer. It’ll also be possible to encapsulate other resources at the same time, such as HistoryRecorderHandle – obviating both the `NavigationResources` struct from 5ac8c9a5ddebd4f90f49ab391c54ac9ff57707df and the `(Navigator, HistoryRecorderHandle)` tuple defined in this PR.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -64,66 +53,8 @@ open class PassiveLocationDataSource: NSObject {
     /**
      Starts the generation of location updates with an optional completion handler that gets called when the location data source is ready to receive snapped location updates.
      */
-    public func startUpdatingLocation(completionHandler: ((Error?) -> Void)? = nil) {
+    public func startUpdatingLocation() {
```
1ec5, line 56: This public API change is worth a mention in the changelog.

MaximAlien: Makes sense. Added.

Sources/MapboxCoreNavigation/Navigator.swift
MaximAlien: Tuple is no longer relevant after adding `CoreNavigationNavigator.swift` (btw feel free to propose better naming if you have any ideas). Shall I remove `NavigatorResources` in scope of this PR? As I can see it's not used anywhere at the moment.

Sources/MapboxCoreNavigation/Navigator.swift
MaximAlien: Got it. Please check latest update, `CoreNavigationNavigator.swift` now encapsulates both `HistoryRecorderHandle` and `MapboxNavigationNative.Navigator`. Access point is possible via `Navigator.shared` static computed property.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 * Fixed a potential hang when `PassiveLocationDataSource` fails to download routing tiles. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Fixed an issue where `RouteController` or `PassiveLocationDataSource` sometimes snapped the user’s location assuming a path that violated a turn restriction. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Improved performance and decreased memory usage when downloading routing tiles. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
+* `PassiveLocationManager.startUpdatingLocation(completionHandler:)` no longer requires optional `completionHandler:` property. ([#2823](https://github.com/mapbox/mapbox-navigation-ios/pull/2823))
```
1ec5, line 37:
```suggestion
* Renamed `PassiveLocationManager.startUpdatingLocation(completionHandler:)` to `PassiveLocationManager.startUpdatingLocation()`. This method now runs synchronously like `CLLocationManager.startUpdatingLocation()`. ([#2823](https://github.com/mapbox/mapbox-navigation-ios/pull/2823))
```

1ec5 (approved):

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
ben-xD: The documentation above it was not updated. It still says optional completion handler.

ben-xD: Also, previously an error was passed in the completion handler. Now, the function doesn't throw an error. Is it good to assume no errors will be raised here?

Udumft: Sorry for the inconsistent code comment, we'll update it.
> Is it good to assume no errors will be raised here?

ben-xD: Thanks a lot

Issue #2818, opened 2021-02-19
Congestion levels are incorrectly drawn on route line.


MaximAlien:
Sometimes congestion levels are incorrectly drawn on the route line, this leads to issues like this one:

<img width="249" alt="Screen Shot 2021-02-19 at 11 20 30 AM" src="https://user-images.githubusercontent.com/1496498/108551434-ce85a480-72a4-11eb-96bf-fd8e14ef8711.png">

Root cause of this problem lies in Mapbox Maps SDK for iOS and underlying libs (when line gradient keys have many values after comma, e.g. `19` line gradient will not be rendered at all). To workaround this Navigation SDK provides temporary solution, which rounds precision to `16` digits, this in turn leads to such side effect.

MaximAlien:
Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/2930.

PR #2825, opened 2021-02-22
Add ability to override main and alternative route lines.

MaximAlien:
### Description

Current PR implements functionality, which allows to override main and alternative route lines, along with refactoring which aims to simplify route line gradient generation.

Screen Shot 2021-04-05 at 3 40 57 PM

Also fixes https://github.com/mapbox/mapbox-navigation-ios/issues/2907 and https://github.com/mapbox/mapbox-navigation-ios/issues/2818.

### DIFF
--- a/Sources/MapboxNavigation/Array.swift
+++ b/Sources/MapboxNavigation/Array.swift
@@ -1,4 +1,5 @@
 import Foundation
+import MapboxDirections
 import CoreLocation
 import Turf
 import CoreGraphics
@@ -52,6 +53,51 @@ extension Array where Iterator.Element == [CLLocationCoordinate2D]? {
 }
 
 extension Array where Iterator.Element == CLLocationCoordinate2D {
+
+    /**
+     Returns an array of congestion segments by associating the given congestion levels with the coordinates of the respective line segments that they apply to.
+     
+     This method coalesces consecutive line segments that have the same congestion level.
+     
+     For each item in the `CongestionSegment` collection a `CongestionLevel` substitution will take place that has a streets road class contained in the `roadClassesWithOverriddenCongestionLevels` collection.
+     For each of these items the `CongestionLevel` for `.unknown` traffic congestion will be replaced with the `.low` traffic congestion.
+     
+     - parameter congestionLevels: The congestion levels along a leg. There should be one fewer congestion levels than coordinates.
+     - parameter streetsRoadClasses: A collection of streets road classes for each geometry index in `Intersection`. There should be the same amount of `streetsRoadClasses` and `congestions`.
+     - parameter roadClassesWithOverriddenCongestionLevels: Streets road classes for which a `CongestionLevel` substitution should occur.
+     - returns: A list of `CongestionSegment` tuples with coordinate and congestion level.
+     */
+    func combined(_ congestionLevels: [CongestionLevel],
+                  streetsRoadClasses: [MapboxStreetsRoadClass?]? = nil,
+                  roadClassesWithOverriddenCongestionLevels: Set<MapboxStreetsRoadClass>? = nil) -> [CongestionSegment] {
+        var segments: [CongestionSegment] = []
+        segments.reserveCapacity(congestionLevels.count)
+        
+        var index = 0
+        for (firstSegment, congestionLevel) in zip(zip(self, self.suffix(from: 1)), congestionLevels) {
+            let coordinates = [firstSegment.0, firstSegment.1]
+            
+            var overriddenCongestionLevel = congestionLevel
+            if let streetsRoadClasses = streetsRoadClasses,
+               let roadClassesWithOverriddenCongestionLevels = roadClassesWithOverriddenCongestionLevels,
+               streetsRoadClasses.indices.contains(index),
+               let streetsRoadClass = streetsRoadClasses[index],
+               congestionLevel == .unknown,
+               roadClassesWithOverriddenCongestionLevels.contains(streetsRoadClass) {
+                overriddenCongestionLevel = .low
+            }
+            
+            if segments.last?.1 == overriddenCongestionLevel {
+                segments[segments.count - 1].0 += coordinates
+            } else {
+                segments.append((coordinates, overriddenCongestionLevel))
+            }
+            
+            index += 1
+        }
+        
+        return segments
+    }
     
     func sliced(from: CLLocationCoordinate2D? = nil, to: CLLocationCoordinate2D? = nil) -> [CLLocationCoordinate2D] {
         return LineString(self).sliced(from: from, to: to)?.coordinates ?? []
--- a/Sources/MapboxNavigation/Constants.swift
+++ b/Sources/MapboxNavigation/Constants.swift
@@ -2,12 +2,15 @@ import CoreLocation
 import Foundation
 import MapboxDirections
 
+/**
+ A tuple that pairs an array of coordinates with level of traffic congestion along these coordinates.
+ */
 typealias CongestionSegment = ([CLLocationCoordinate2D], CongestionLevel)
 
 /**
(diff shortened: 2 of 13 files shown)


Sources/MapboxNavigation/Array.swift
```diff
@@ -52,6 +53,51 @@ extension Array where Iterator.Element == [CLLocationCoordinate2D]? {
...
+     
+     For each item in the `CongestionSegment` collection a `CongestionLevel` substitution will take place that has a streets road class contained in the `roadClassesWithOverriddenCongestionLevels` collection.
+     For each of these items the `CongestionLevel` for `.unknown` traffic congestion will be replaced with the `.low` traffic congestion.
+     
+     - parameter congestionLevels: The congestion levels along a leg. There should be one fewer congestion levels than coordinates.
+     - parameter streetsRoadClasses: A collection of streets road classes for each geometry index in `Intersection`. There should be the same amount of `streetsRoadClasses` and `congestions`.
+     - parameter roadClassesWithOverriddenCongestionLevels: Streets road classes for which a `CongestionLevel` substitution should occur.
+     - returns: A list of `CongestionSegment` tuples with coordinate and congestion level.
+     */
+    func combined(_ congestionLevels: [CongestionLevel],
```
1ec5, line 70: Does this method use `self` at all? If not, it should either be static or use `self` instead of `congestionLevels`.

Sources/MapboxNavigation/Constants.swift
```diff
@@ -2,12 +2,15 @@ import CoreLocation
 import Foundation
 import MapboxDirections
 
+/**
+ Tuple, which contains collection of coordinates and level of traffic congestion along these coordinates.
```
1ec5, line 6:
```suggestion
 A tuple that pairs an array of coordinates with level of traffic congestion along these coordinates.
```

Sources/MapboxNavigation/Constants.swift
```diff
@@ -2,12 +2,15 @@ import CoreLocation
...
+ */
 typealias CongestionSegment = ([CLLocationCoordinate2D], CongestionLevel)
```
1ec5, line 8: Maybe at some point we should replace `CongestionSegment` with `Feature` in which the geometry is `LineString` and the `CongestionLevel` is one of the properties. That would be a major change for much later though.

Sources/MapboxNavigation/NavigationMapViewIdentifiers.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+struct IdentifierString {
```
1ec5, line 3: Split this struct into `NavigationMapView.LayerIdentifier`, `NavigationMapView.SourceIdentifier`, and `NavigationMapView.ImageIdentifier`. That’ll simplify the naming of many of these constants and avoid bugs where a layer identifier is used as a source identifier or vice versa.

Sources/MapboxNavigation/RouteLineType.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+enum RouteLineType {
+    
+    case source(isMainRoute: Bool = false, isSourceCasing: Bool = false)
+    
+    case route(isMainRoute: Bool = false)
```
1ec5, line 7: Do these associated values all need default values? Maybe it’s better if client code has to specify whether it’s main route explicitly, to avoid a bug where none of the routes is the main route.

Sources/MapboxNavigation/Route.swift
```diff
@@ -43,6 +43,68 @@ extension Route {
...
+            }
+            
+            features.append(contentsOf: legFeatures)
+        }
+        
+        return features
+    }
+    
+    func identifier(_ routeLineType: RouteLineType) -> String {
+        let identifier = Unmanaged.passUnretained(self).toOpaque()
```
1ec5, line 95: Add a comment explaining why we need to work with unmanaged memory here. I assume it’s to ensure that each route on the map has a unique source and layer identifier. Perhaps it would be sufficient to combine `Route.routeIdentifier` with the route index?

Sources/MapboxNavigation/Route.swift
```diff
@@ -43,6 +43,68 @@ extension Route {
...
+                legFeatures = [feature]
+            }
+            
+            features.append(contentsOf: legFeatures)
+        }
+        
+        return features
+    }
+    
+    func identifier(_ routeLineType: RouteLineType) -> String {
```
1ec5, line 94: This method should become a method of RouteLineType itself that takes a base identifier as an argument.

Sources/MapboxNavigation/Route.swift
MaximAlien: I think it'd be still safer to use address of an object as its identifier. `Route.routeIdentifier`s are equal for single direction calculation, and possibly route indexes depending on situation. To prevent these naming clashes I'd rather use memory address. Will add comment.

Sources/MapboxNavigation/Array.swift
MaximAlien: Not sure it's possible in this case, this method is already applied only on arrays which contain `CLLocationCoordinate2D` objects and merges them with `CongestionLevel` objects.

Sources/MapboxNavigation/RouteLineType.swift
MaximAlien: Agree, default parameters look more error prone.

Sources/MapboxNavigation/Route.swift
MaximAlien: I don't think it'll allow us to make implementation clear and understandable. In my opinion `RouteLineType` should not have identifier, but `Route` object should be responsible for storing its identifier based on specific conditions (passed via `RouteLineType`). If we move identification logic to `RouteLineType` and store base identifier separately it'll not be really clear, in current case both logic and data reside in one entity.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -205,142 +205,118 @@ extension NavigationMapView {
...
+            
             let newFractionTraveled = self.preFractionTraveled + traveledDifference * timePassedInMilliseconds.truncatingRemainder(dividingBy: 1000) / 1000
             
-            if let mainRouteLayerGradient = self.routeLineGradient(routeProgress.route, fractionTraveled: newFractionTraveled) {
-                self.mapView.style.updateLayer(id: mainRouteLayerIdentifier, type: LineLayer.self) { (lineLayer) in
-                    lineLayer.paint?.lineGradient = .expression(Expression.routeLineGradientExpression(mainRouteLayerGradient))
-                }
+            let congestionSegments = routeProgress.route.congestionFeatures(roadClassesWithOverriddenCongestionLevels: self.roadClassesWithOverriddenCongestionLevels)
+            let mainRouteLayerGradient = self.routeLineGradient(congestionSegments,
+                                                                fractionTraveled: newFractionTraveled)
```
MaximAlien, line 213: Calculations here cause significant performance issues on long routes. Improvements should be addressed in scope of another ticket.

ShanMa1991 (approved):

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -205,142 +205,124 @@ extension NavigationMapView {
...
-            filteredGradientStops[0.0] = traversedRouteColor
-            filteredGradientStops[percentTraveled.nextDown] = traversedRouteColor
-            filteredGradientStops[percentTraveled] = minStop.value
-        }
-        
-        var resultGradientStops = [Double: UIColor]()
-
-        filteredGradientStops.filter({ $0.0 >= 0.0 }).forEach {
-            resultGradientStops[Double($0.0)] = $0.1
-        }
```
MaximAlien, line 326: These calculations were removed to slightly improve performance. For example, in new implementation we keep minimum gradient stop is calculated right away, without having additional loop. Whole dictionary copy is not needed as well.

PR #2826, opened 2021-02-22
Navigation camera viewport API.

MaximAlien:
### Description

`NavigationCamera` viewport APIs bring the ability to control camera related behavior, which is specific for navigation (mostly free-drive and active guidance navigation).

`NavigationCamera` API introduces such components:
- `NavigationCamera` class, which works as a manager for all camera related operations (allows to move camera to following, overview and idle states).
- `ViewportDataSource` protocol.
- `NavigationViewportDataSource` class, which conforms to `ViewportDataSource` and listens to location changes in free drive and active guidance modes and prepares `CameraOptions`. This class prepares `CameraOptions` for both iOS and CarPlay, as they're different. Can be overridden by the end user to provide custom data source.
- `CameraView` class, which allows to run `UIViewPropertyAnimator` animations. Will be removed when such feature appears in Maps SDK.
- `CameraStateTransition` protocol.
- `NavigationCameraStateTransition` class, which conforms to `CameraStateTransition` and provides set of specific animated transitions. Can be overridden by the end user.

Current PR also contains an attempt to decouple location related APIs (e.g. `UserCourseView`) and camera related APIs in `NavigationMapView`, this leads to removal/replacement of certain properties (`defaultAltitude`, `zoomedOutMotorwayAltitude` etc).

Example application with custom camera implementation can be found here: https://github.com/mapbox/navigation-ios-examples/pull/92.

### Initial results

https://user-images.githubusercontent.com/1496498/113346228-6df87700-92e8-11eb-843e-4ec37e506194.mov

https://user-images.githubusercontent.com/1496498/113347514-1d821900-92ea-11eb-96ab-c8597c74527e.mov

https://user-images.githubusercontent.com/1496498/113350067-cf6f1480-92ed-11eb-9f10-56078c30735f.mov

https://user-images.githubusercontent.com/1496498/113350690-a4d18b80-92ee-11eb-90a3-dc7cd932def2.mov

https://user-images.githubusercontent.com/1496498/114232298-83951e80-9930-11eb-92bd-2925c9a0c36a.mov

### Known issues

- https://github.com/mapbox/mapbox-maps-ios-internal/issues/575
- https://github.com/mapbox/mapbox-maps-ios-internal/issues/598
- ~~https://github.com/mapbox/mapbox-maps-ios-internal/issues/600~~
- ~~https://github.com/mapbox/mapbox-navigation-ios/issues/2892~~
- ~~https://github.com/mapbox/mapbox-navigation-ios/issues/2891~~
- ~~https://github.com/mapbox/mapbox-navigation-ios/issues/2890~~
- ~~https://github.com/mapbox/mapbox-navigation-ios/issues/2889~~
- https://github.com/mapbox/mapbox-navigation-ios/issues/2912
- https://github.com/mapbox/mapbox-navigation-ios/issues/2921


Sources/MapboxNavigation/UserHaloCourseView.swift
```diff
@@ -7,21 +7,6 @@ import CoreLocation
 public class UserHaloCourseView: UIView, CourseUpdatable {
     private var lastLocationUpdate: Date?
 
-    /**
-     Transforms the location of the user halo.
-     */
-    public func update(location: CLLocation, pitch: CGFloat, direction: CLLocationDegrees, animated: Bool, tracksUserCourse: Bool) {
```
MaximAlien, line 13: This implementation is similar to default one in `CourseUpdatable`, so can be removed.

Sources/MapboxNavigation/NavigationCameraType.swift
```diff
@@ -0,0 +1,16 @@
+/**
+ Possible types of `NavigationCamera`.
+ */
+public enum NavigationCameraType {
```
MaximAlien, line 4: Since camera behaviors are different on iOS and CarPlay there is a need to distinguish `NavigationMapView`s which are created for each platform (because they hold `NavigationCamera` instance). Feel free to propose better naming if you have any better ideas.

Sources/MapboxNavigation/NavigationMapViewDelegate.swift
```diff
@@ -107,36 +99,4 @@ public extension NavigationMapViewDelegate {
...
         return .zero
     }
-}
-
-// MARK: - NavigationMapViewCourseTrackingDelegate methods
-
-/**
- The `NavigationMapViewCourseTrackingDelegate` provides methods for responding to the `NavigationMapView` starting or stopping course tracking.
- */
-public protocol NavigationMapViewCourseTrackingDelegate: class, UnimplementedLogging {
```
MaximAlien, line 117: `NavigationMapViewCourseTrackingDelegate` was removed in favor of `Notification`, which is sent whenever `NavigationCameraState` changes. `NavigationCameraState` also provides more information regarding camera state (`.idle`, `.transitionToFollowing`, `.following`, `.transitionToOverview`, `.overview`).

Example/ViewController.swift
```diff
@@ -258,11 +287,16 @@ class ViewController: UIViewController {
 
         customViewController.userIndexedRoute = (route, 0)
         customViewController.userRouteOptions = routeOptions
-
-        // TODO: Add the ability to show destination annotation.
         customViewController.simulateLocation = simulationButton.isSelected
+        
+        passiveLocationDataSource?.systemLocationManager.stopUpdatingLocation()
```
MaximAlien, line 292: In my opinion this is pretty significant limitation: we have to stop any free-drive related location updates to prevent clash with locations sent during active guidance. At the same time there is no easy way to do this under the hood, because of public nature of `PassiveLocationDataSource`. Any better ideas on how we can improve this?

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -3319,10 +3379,10 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = "com.mapbox.Example-CarPlay";
+				PRODUCT_BUNDLE_IDENTIFIER = "com.Mapbox.CarPlay-Example";
```
MaximAlien, line 3382: Bundle of `Example-CarPlay` target had to be updated to use updated provisioning profile.

Sources/MapboxNavigation/CameraView.swift
```diff
@@ -0,0 +1,162 @@
+import UIKit
+import MapboxMaps
+
+/**
+ A view that represents a camera view port.
+ */
+public class CameraView: UIView {
```
MaximAlien, line 7: `CameraView` was added temporarily to execute animations using `UIViewPropertyAnimator`.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -27,21 +27,6 @@ open class NavigationMapView: UIView {
...
-    
-    /**
-     Returns the altitude the map conditionally zooms out to when user is on a motorway, and the maneuver length is sufficently long.
-     */
-    public var zoomedOutMotorwayAltitude: CLLocationDistance = 2000.0
-    
-    /**
-     Returns the threshold for what the map considers a "long-enough" maneuver distance to trigger a zoom-out when the user enters a motorway.
-     */
-    public var longManeuverDistance: CLLocationDistance = 1000.0
```
MaximAlien, line 43: Since new camera related changes will automatically adapt to road conditions such properties are no longer used.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -332,116 +277,62 @@ open class NavigationMapView: UIView {
...
     /**
      Updates the map view’s preferred frames per second to the appropriate value for the current route progress.
      
-     This method accounts for the proximity to a maneuver and the current power source. It has no effect if `tracksUserCourse` is set to `true`.
+     This method accounts for the proximity to a maneuver and the current power source.
+     It has no effect if `NavigationCameraState` is in `.following` mode.
      */
-    open func updatePreferredFrameRate(for routeProgress: RouteProgress) {
-        guard tracksUserCourse else { return }
+    public func updatePreferredFrameRate(for routeProgress: RouteProgress) {
```
MaximAlien, line 286: I had to slightly rework frame rate changing functionality. After adding new camera animations we have to always update `UserCourseView` frame by frame. At the same time I'm not sure why `pluggedInFramesPerSecond` is always set to `PreferredFPS.lowPower`, I'd expect it to be `PreferredFPS.maximum`.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -332,116 +277,62 @@ open class NavigationMapView: UIView {
...
         } else {
-            mapView.preferredFPS = FrameIntervalOptions.pluggedInFramesPerSecond
+            mapView.preferredFPS = FrameIntervalOptions.defaultFramesPerSecond
         }
     }
     
-    /**
-     Track position on a frame by frame basis. Used for first location update and when resuming tracking mode
-     */
-    public func enableFrameByFrameCourseViewTracking(for duration: TimeInterval) {
```
MaximAlien, line 375: From now on we no longer provide temporary frame by frame updates to `UserCourseView`, it's always updated frame by frame.

Example/NavigationCameraDebugView.swift
```diff
@@ -0,0 +1,68 @@
+import UIKit
+import MapboxMaps
+import MapboxNavigation
+
+class NavigationCameraDebugView: UIView {
```
MaximAlien, line 5: Used for debugging purposes. I intend to keep using it in example application.

ShanMa1991 (commented):
I have a question about the overview. Because before, when we chose the overview mode, the camera is set to fit the route shape, which including the whole route line in the view. But now, the over view mode only included part of the route shape, from the current location to end of the route. Is this change designed to be same as Android, or should we keep the original whole route line?

MaximAlien:
> I have a question about the overview. Because before, when we chose the overview mode, the camera is set to fit the route shape, which including the whole route line in the view. But now, the over view mode only included part of the route shape, from the current location to end of the route. Is this change designed to be same as Android, or should we keep the original whole route line?

Yes, it is intentional to track only remaining portion of the route in overview mode.

Example/ViewController.swift
```diff
@@ -19,12 +19,15 @@ class ViewController: UIViewController {
     var trackStyledFeature: StyledFeature!
     var rawTrackStyledFeature: StyledFeature!
     var speedLimitView: SpeedLimitView!
+    // let passiveLocationDataSource: PassiveLocationDataSource? = nil
```
MaximAlien, line 22: Was used for testing, will be removed.

Sources/MapboxNavigation/NavigationCameraType.swift
avi-c: I was thinking about this. If the only use for now is CarPlay then my sense is to call that out specifically and call this .carPlay (or similar). I don't think our developers would necessarily know what a "head unit" is it may be confusing.
If we are already using the notion of a "head unit" in other parts of the SDK then maybe it will be clear, but as far as I know, the opposite is true: We reference CarPlay in many parts of the SDK and don't refer to a "head unit" anywhere else.

Sources/MapboxNavigation/ViewportDataSourceType.swift
```diff
@@ -0,0 +1,28 @@
+/**
+ Possible types of location related updates `NavigationViewportDataSource` can track.
+ `ViewportDataSourceType` can also be used in custom implementations of classes, which conform to
+ `ViewportDataSource`.
+ */
+public enum ViewportDataSourceType {
+    
+    /**
+     Type, which is used to track raw location updates. In case if such type is used
```
avi-c, line 9: nit: something like: "If 'raw' type is specified, `NavigationViewportDataSource` will..." sounds better to my ear.

avi-c: Also, "continiously" should be "continuously"

avi-c (changes requested):
I think resolving the "head unit" vs. "CarPlay" terminology would be good to do before merging. Otherwise I think this is fine. We should just continue to observe the camera movement to ensure that the Maps SDK doesn't have additional rendering issues crop up that aren't already logged against them.

CHANGELOG.md
```diff
@@ -39,6 +39,13 @@
...
+* Replaced `CourseUpdatable.update(location:pitch:direction:animated:tracksUserCourse:)` with `CourseUpdatable.update(location:pitch:direction:animated:navigationCameraState:)` to provide more agile way of handling `NavigationCameraState`. ([#2826](https://github.com/mapbox/mapbox-navigation-ios/pull/2826))
+* Added `NavigationMapView.init(frame:navigationCameraType:)` to be able to provide type of `NavigationCamera`, which should be used for that specific instance of `NavigationMapView` (either iOS or CarPlay). ([#2826](https://github.com/mapbox/mapbox-navigation-ios/pull/2826))
+* Added `NavigationCamera`, `ViewportDataSourceType`, `ViewportDataSourceDelegate`, `NavigationCameraState` Navigation Viewport Camera APIs. By default Navigation SDK for iOS provides default camera behavior via `NavigationViewportDataSource` and `NavigationCameraStateTransition` classes. If you'd like to override current behavior use `ViewportDataSource` and `CameraStateTransition` protocols for custom behavior. ([#2826](https://github.com/mapbox/mapbox-navigation-ios/pull/2826))
```
1ec5, line 47: Do we want to say something about how snazzy the new camera animations are, not in terms of API but rather the runtime behavior?

Sources/MapboxNavigation/NavigationCameraType.swift
MaximAlien: Yep, I thought that `headUnit` would be more generic and not really dependent on platform branding. At the same time `CarPlay` sounds more exact and clear. I think I'll use `.carPlay` and `.mobile` for `NavigationCameraType`.

Sources/MapboxNavigation/ViewportDataSourceType.swift
MaximAlien: Thanks, updated comments.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -257,30 +169,63 @@ open class NavigationMapView: UIView {
...
     
     public override init(frame: CGRect) {
-        altitude = defaultAltitude
         super.init(frame: frame)
         
         setupMapView(frame)
         commonInit()
     }
     
+    public init(frame: CGRect, navigationCameraType: NavigationCameraType = .mobile) {
```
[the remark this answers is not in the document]
MaximAlien: Yep, added documentation for this and other initializers.

Sources/MapboxNavigation/CLLocation.swift
```diff
@@ -0,0 +1,14 @@
+import CoreLocation
+
+extension CLLocation {
+    
+    /**
+     Returns distance between two coordinates.
+     */
+    class func distance(from: CLLocationCoordinate2D, to: CLLocationCoordinate2D) -> CLLocationDistance {
```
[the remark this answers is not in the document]
MaximAlien: Yes, makes sense. Removed extension.

Sources/MapboxNavigation/NavigationCamera.swift
```diff
@@ -0,0 +1,206 @@
...
+    }
+    
+    // MARK: - NavigationCamera state related methods
+    
+    /**
+     Call to this method executes a transition to `NavigationCameraState.following` state.
+     When started, state will first change to `NavigationCameraState.transitionToFollowing` and then
+     to the final `NavigationCameraState.following` when ended.
+     */
+    public func requestNavigationCameraToFollowing() {
```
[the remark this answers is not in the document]
MaximAlien: Agree, I think `follow()` would be the most suitable naming in this case.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -183,6 +185,12 @@ public class CarPlayNavigationViewController: UIViewController {
         styleManager?.styles = self.styles
     }
     
+    func setInitialCoordinate() {
+        guard let centerCoordinate = navigationService.routeProgress.route.shape?.coordinates.first else { return }
+        navigationMapView?.mapView.cameraManager.setCamera(to: CameraOptions(center: centerCoordinate, zoom: 14.0))
```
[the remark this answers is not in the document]
MaximAlien: This probably can be answered by @d-prukop. Initial zoom is based on following mode in free drive, and there we also set  `followingWithoutRouteZoomLevel` as `14.0`. If I'm not mistaken Dave mentioned that it was the best suited zoom level tested in several locations.

Sources/MapboxNavigation/NavigationCameraConstants.swift
```diff
@@ -0,0 +1,54 @@
...
+        public var rawValue: String
+
+        public init(rawValue: String) {
+            self.rawValue = rawValue
+        }
+        
+        /**
+         A key in the user info dictionary of a `Notification.Name.navigationCameraStateDidChange` notification. The corresponding value is a `NavigationCameraState` object.
+         */
+        public static let stateKey: NotificationUserInfoKey = .init(rawValue: "state")
```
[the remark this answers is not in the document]
MaximAlien: Removed `*key`.

Sources/MapboxNavigation/CameraOptions.swift
```diff
@@ -0,0 +1,26 @@
+import MapboxMaps
+
+extension CameraOptions {
+    
+    /**
+     Returns description of all properties in `CameraOptions`.
+     */
+    public override var description: String {
```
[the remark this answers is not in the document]
MaximAlien: Agree that `debugDescription` is more appropriate here. Since this is mostly aims to simplify debugging I'd rather keep it on Navigation SDK side. But in general having informative `description` for all Bindgen generated interfaces would be good to have.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -0,0 +1,418 @@
...
+              let pitch = cameraOptions.pitch,
+              let padding = cameraOptions.padding,
+              let anchor = cameraOptions.anchor else {
+            completion()
+            return
+        }
+        
+        if transitionDestinationIsOffScreen(location, edgeInsets: padding) {
+            let screenCenterPoint = self.screenCenterPoint(0.0, bounds: mapView.bounds, edgeInsets: padding)
+            let lineString = LineString([mapView.centerCoordinate, location])
```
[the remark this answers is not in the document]
MaximAlien: This part was ported from implementation created by Dave, I'd rather keep it as it is.

Sources/MapboxNavigation/NavigationMapView.swift
MaximAlien: In initial version I'd rather keep public APIs as simple as possible, and then in future iterations add only properties in case if those are really required. Also in case if users would like custom camera behavior it'd be possible to conform to `ViewportDataSource` directly. From the ones listed here `defaultAltitude` (adapted to zoom level) can be exposed.

Sources/MapboxNavigation/NavigationMapView.swift
MaximAlien: Yes, but in current implementation we no longer require `shouldPositionCourseViewFrameByFrame`, as `UserCourseView` is always updated on each frame change. I think in #2643 we incorrectly use `pluggedInFramesPerSecond`, as it's always set to `MGLMapViewPreferredFramesPerSecond.lowPower`, which is not 30 FPS. I think in current PR right thing would be to use `PreferredFPS.maximum` for `pluggedInFramesPerSecond`.

/cc @Udumft

Sources/MapboxNavigation/NavigationMapViewDelegate.swift
MaximAlien: Yes, in this case `NavigationMapViewCourseTrackingDelegate` is not enough though, as we need camera state related changes in more than one place.

Sources/MapboxNavigation/BoundingBox.swift
```diff
@@ -0,0 +1,18 @@
...
+     Returns zoom level inside of specific `CGSize`, in which `BoundingBox` was fit to.
+     */
+    func zoomLevel(fitTo size: CGSize) -> Double {
+        let latitudeFraction = (self.northEast.latitude.radius - self.southWest.latitude.radius) / .pi
+        let longitudeDiff = self.northEast.longitude - self.southWest.longitude
+        let longitudeFraction = ((longitudeDiff < 0) ? (longitudeDiff + 360) : longitudeDiff) / 360
+        let latitudeZoom = log(Double(size.height) / 512.0 / latitudeFraction) / M_LN2
+        let longitudeZoom = log(Double(size.width) / 512.0 / longitudeFraction) / M_LN2
+        
+        return min(latitudeZoom, longitudeZoom, 21.0)
```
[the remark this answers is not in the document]
MaximAlien: It's regardless of the pitch. Pitch coefficient will be taken into account when forming `size` parameter.

CHANGELOG.md
MaximAlien: Yes, makes sense.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -351,12 +346,7 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
         guard let mapView = routeMapViewController.navigationMapView.mapView,
               let centerCoordinate = routeMapViewController.navigationService.routeProgress.route.shape?.coordinates.first else { return }
         
-        let zoom = CGFloat(ZoomLevelForAltitude(routeMapViewController.navigationMapView.defaultAltitude,
-                                                mapView.pitch,
-                                                centerCoordinate.latitude,
-                                                mapView.bounds.size))
-        
-        mapView.cameraManager.setCamera(to: CameraOptions(center: centerCoordinate, zoom: zoom))
+        mapView.cameraManager.setCamera(to: CameraOptions(center: centerCoordinate, zoom: 14.0))
```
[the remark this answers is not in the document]
MaximAlien: For now I'd keep this without any change. At the end there might be no need to set initial coordinate and zoom level right at the start of active guidance navigation (this wasn't the case in 1.x and camera jump started occurring after introducing new Maps SDK camera API). I'll take to folks from Maps SDK regarding this.

avi-c (approved):
Thank you. This looks good.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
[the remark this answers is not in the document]
1ec5: We should get confirmation that these several locations are at different enough latitudes to cover the needs of the wide variety of applications that use or could use this SDK.

Sources/MapboxNavigation/CLLocationDegrees.swift
```diff
@@ -0,0 +1,13 @@
+import CoreLocation
+
+extension CLLocationDegrees {
+    
+    /**
+     Returns Earth's radius for specific latitude.
+     */
+    var radius: Double {
```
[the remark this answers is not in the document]
1ec5: At least the property name and comment should be more specific, given the multiple valid but incompatible ways to calculate the Earth’s radius.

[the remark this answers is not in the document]
MaximAlien: Incorrect naming is used here. This value just represents radian value for specific degree. I've slightly refactored it to use Maps SDK APIs for this. This extension is not used anymore.

Sources/MapboxNavigation/NavigationCameraType.swift
1ec5: `iPhone` would be a fine alternative to `mobile`, since an iPad-optimized camera could conceivably be closer to a CarPlay-optimized camera than an iPhone-optimized one.

1ec5 (commented):
I think we should try to get some more certainty on https://github.com/mapbox/mapbox-navigation-ios/pull/2826#discussion_r610095261 and https://github.com/mapbox/mapbox-navigation-ios/pull/2826#discussion_r609193075 before landing, but otherwise the code looks to be in good shape.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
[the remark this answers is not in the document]
MaximAlien: `NavigationViewportDataSource` now exposes `defaultAltitude`, which is set to `1000` meters by default. It'll be used (along with `latitude` of initial coordinate) to set central coordinate.

Sources/MapboxNavigation/NavigationCameraType.swift
MaximAlien: I was thinking about `.iOS`, but since `CarPlay` is just a part of `iOS` it'd contradict each other. `.iPhone` sounds a bit too restrictive IMO, at this point there is technically no difference between cameras on `iPhone` and `iPad`.

Sources/MapboxNavigation/NavigationMapView.swift
Udumft: TBH, I tried to recall why exactly `defaultFramesPerSecond` and `pluggedInFramesPerSecond` were set as they were, but couldn't...
For implementation in this PR I have a question: we are opting to select `.normal` frame rate unless we encounter a straight area of a route or are going away from maneuver nodes, at which case we are a) ramping up frame rate if device is plugged in or b) slowing down to `.lowPower` otherwise. Why same event causes reverse FPS reaction? Shouldn't we always increase or decrease frame rate at such event?

MaximAlien: I'm not too familiar with specifics/reasoning of underlying logic, but `a` seems to be reasonable. In case of `b` (based on current implementation) we're not necessarily changing frame rate to `PreferredFPS.lowPower`, it's up to user to decide how it can be modified in `minimumFramesPerSecond`.

On the other hand I'm not sure why we change `PreferredFPS` based on such maneuver directions (e.g. one of the cases checks only **slight** [.straightAhead, .slightLeft, .slightRight] directions, I'd expect to increase frame rate when we're approaching a **sharp** maneuver).

In any case this doesn't directly change public APIs so we can tweak it as a tail work if such need arises.

/cc @1ec5

Sources/MapboxNavigation/NavigationCameraType.swift
1ec5: The difference is that an iPad typically has a different form factor, like the way that CarPlay typically has a different form factor. I could picture the camera behaving differently for different form factors, but then again, maybe size classes would be a better way to accomplish that.

Sources/MapboxNavigation/NavigationMapView.swift
1ec5: > On the other hand I'm not sure why we change `PreferredFPS` based on such maneuver directions (e.g. one of the cases checks only **slight** [.straightAhead, .slightLeft, .slightRight] directions, I'd expect to increase frame rate when we're approaching a **sharp** maneuver).

The original intention was to animate as smoothly as possible when making a turn, because any throttling would be perceived as choppy animation. But straight and slight turns don’t cause the camera to swing around, so there’s no need to disable throttling in those cases.

Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -0,0 +1,424 @@
...
+    /**
+     Controls how much the bearing can deviate from the location's bearing, in degrees.
+     
+     In case if set, the `bearing` property of `CameraOptions` during active guidance navigation
+     won't exactly reflect the bearing returned by the location, but will also be affected by the
+     direction to the upcoming framed geometry, to maximize the viewable area.
+     
+     Defaults to `20.0` degrees.
+     */
+    public var maximumBearingSmoothingAngle: CLLocationDirection? = 20.0
```
MaximAlien, line 72: As per our discussion with @d-prukop `distanceToFrameAfterManeuver` and `maximumBearingSmoothingAngle` were made publicly available to provide more control over default camera behavior. There are several other properties, which are exposed in Mapbox Navigation for Android, but haven't made their way into `main` branch yet (and not yet present on iOS). We can add those in scope of tail-work. iOS implementation also doesn't use boolean flags to control usage of such properties (to simplify API), instead certain parameters provided as optionals.

Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -0,0 +1,424 @@
...
+/**
+ Class, which conforms to `ViewportDataSource` protocol and provides default implementation of it.
+ */
+public class NavigationViewportDataSource: ViewportDataSource {
+    
+    /**
+     Delegate, which is used to notify `NavigationCamera` regarding upcoming `CameraOptions`
+     related changes.
+     */
+    public var delegate: ViewportDataSourceDelegate?
```
bamx23, line 15: Should we use `weak` here? I see possible retain loop

MaximAlien: Good catch, will be addressed in another PR.

bamx23: By the way, SwiftLint produces a warning for such issues.

Example/ViewController+FreeDrive.swift
```diff
@@ -116,7 +122,6 @@ extension ViewController {
             let branchNames = branchEdgeIdentifiers.flatMap { edgeNames(identifier: $0) }
             statusString += " at \(branchNames.joined(separator: ", "))"
         }
-        print(statusString)
```
1ec5, line 119: This print statement was intentional: otherwise, all the electronic horizon work above is for naught. Ideally there would be some UI associated with this output, but a print statement was more expedient at the time.

1ec5: Fixed in ca8be54d000308cc280c37a06079bd8b2e7e05a2.

PR #2827, opened 2021-02-23
Add ability to port building highlighting feature

ShanMa1991:
### Description
This Pr is to add the ability to high light choose building after long press in 3D, solving the [#728](https://github.com/mapbox/navigation-sdks/issues/728) The opacity would be varied with zoom level. The building would also appear in certain zoom level. 

- [x] add building highlighting feature
- [x] allow a series of buildings to highlight after long press

### Implementation
Add the calling functions of `mapView.highlightBuildings(at:, in 3D:)` to highlight the chosen waypoint. Do query the feature at the waypoint on map view to find the id of the corresponding building. Then add one fillExtrusionLayer and filter the layer based on the building id. 

It would allow the building from 3D changing to 2D in `DayStyle` and `NightStyle`. The color of building would also change from blue to red when  `NightStyle` is chosen.

### Screenshots or Gifs
As the below .gif, it allow a series of long press to choose multiple buildings to highlight, the color of the building would change under different circumstances.
highlighting

It also allow the end of route the puck located at the waypoint and change 3D to 2D, behavior same as the 1.x.
building

### DIFF
--- a/Sources/MapboxNavigation/Feature.swift
+++ b/Sources/MapboxNavigation/Feature.swift
@@ -1,0 +1,21 @@
+import Foundation
+import Turf
+
+extension Feature {
+
+    var featureIdentifier: Int64? {
+        guard let featureIdentifier = self.identifier else { return nil }
+        
+        switch featureIdentifier {
+        case .string(let identifier):
+            return Int64(identifier)
+        case .number(let identifier):
+            switch identifier {
+            case .double(let identifier):
+                return Int64(identifier)
+            case .int(let identifier):
+                return Int64(identifier)
+            }
+        }
+    }
+}
--- a/Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
+++ b/Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
@@ -1,7 +1,7 @@
 import CoreLocation
 import MapboxMaps
+import Turf
 
-// TODO: Implement the ability to highlight/unhighlight buildings in 2D/3D.
 extension NavigationMapView {
     
     // MARK: - Building Extrusion Highlighting methods
@@ -11,17 +11,97 @@ extension NavigationMapView {
      
      - parameter coordinates: Coordinates which represent building locations.
      - parameter extrudesBuildings: Switch which allows to highlight buildings in either 2D or 3D. Defaults to true.
-     
-     - returns: Bool indicating if number of buildings found equals number of coordinates supplied.
+     - parameter completion: An escaping closure to be executed when the `MapView` feature querying for all `coordinates` ends. A single `Boolean` argument in closure indicates whether or not buildings were found for all coordinates.
      */
-    @discardableResult public func highlightBuildings(at coordinates: [CLLocationCoordinate2D], in3D extrudesBuildings: Bool = true) -> Bool {
-        return false
+    public func highlightBuildings(at coordinates: [CLLocationCoordinate2D],
+                                   in3D extrudesBuildings: Bool = true,
+                                   completion: ((_ foundAllBuildings: Bool) -> Void)? = nil) {
+        var foundBuildingIds = Set<Int64>()
+        let group = DispatchGroup()
+        guard let identifiers = try? mapView.__map.getStyleLayers().compactMap({ $0.id }).filter({ $0.contains("building") }) else {
+            completion?(false)
+            return
+        }
+        
+        coordinates.forEach {
+            group.enter()
+            let screenCoordinate = mapView.point(for: $0, in: self)
+            mapView.visibleFeatures(at: screenCoordinate,
+                                    styleLayers: Set(identifiers),
+                                    completion: { [weak self] result in
+                                        guard let _ = self else { return }
+                                        if case .success(let features) = result {
+                                            if let identifier = features.first?.featureIdentifier {
+                                                foundBuildingIds.insert(identifier)
+                                            }
+                                            group.leave()
+                                        }
+                                    })
+        }
+
+        group.notify(queue: DispatchQueue.main) {
+            self.addBuildingsLayer(with: foundBuildingIds, in3D: extrudesBuildings)
+            completion?(foundBuildingIds.count == coordinates.count)
+        }
     }
     
     /**
      Removes the highlight from all buildings highlighted by `highlightBuildings(at:in3D:)`.
      */
     public func unhighlightBuildings() {
-        
+        guard let _ = try? mapView.style.getLayer(with: IdentifierString.buildingExtrusionLayer, type: FillExtrusionLayer.self).get() else { return }
+        let _ = mapView.style.removeStyleLayer(forLayerId: IdentifierString.buildingExtrusionLayer)
     }
+
+    private func addBuildingsLayer(with identifiers: Set<Int64>, in3D: Bool = false, extrudeAll: Bool = false) {
+        let _ = mapView.style.removeStyleLayer(forLayerId: IdentifierString.buildingExtrusionLayer)
+        if identifiers.isEmpty { return }
+        var highlightedBuildingsLayer = FillExtrusionLayer(id: IdentifierString.buildingExtrusionLayer)
+        highlightedBuildingsLayer.source = "composite"
+        highlightedBuildingsLayer.sourceLayer = "building"
+
+        let extrudeExpression: Expression = Exp(.eq) {
+            Exp(.get) {
+                "extrude"
+            }
+            "true"
+        }
+
+        if extrudeAll {
+            highlightedBuildingsLayer.filter = extrudeExpression
+        } else {
+            highlightedBuildingsLayer.filter = Exp(.all) {
+                extrudeExpression
+                Exp(.inExpression) {
+                    Exp(.id)
+                    Exp(.literal) {
+                        identifiers.map({ Double($0) })
+                    }
+                }
+            }
+        }
+
+        if in3D {
+            highlightedBuildingsLayer.paint?.fillExtrusionHeight = .expression(Expression.buildingExtrusionHeightExpression("height"))
+        } else {
+            highlightedBuildingsLayer.paint?.fillExtrusionHeight = .constant(0.0)
+        }
+
+        highlightedBuildingsLayer.paint?.fillExtrusionBase = .expression(Expression.buildingExtrusionHeightExpression("min_height"))
+
+        highlightedBuildingsLayer.paint?.fillExtrusionOpacity = .expression(
+            Exp(.interpolate) {
+                Exp(.linear)
+                Exp(.zoom)
+                13; 0.5
+                17; 0.8
+            }
+        )
+
+        highlightedBuildingsLayer.paint?.fillExtrusionColor = .constant(.init(color: buildingHighlightColor))
+        highlightedBuildingsLayer.paint?.fillExtrusionHeightTransition = StyleTransition(duration: 0.8, delay: 0)
+        highlightedBuildingsLayer.paint?.fillExtrusionOpacityTransition = StyleTransition(duration: 0.8, delay: 0)
+        mapView.style.addLayer(layer: highlightedBuildingsLayer)
+    }
+
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * Renamed public property `MBCurrentLegAttribute` to `CurrentLegAttribute`. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Renamed public property `MBCongestionAttribute` to `CongestionAttribute`. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * `NavigationViewController.mapView` was renamed to `NavigationViewController.navigationMapView`. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
+* `NavigationMapView.highlightBuildings(at:in3D:)` was renamed to `NavigationMapView.highlightBuildings(at:in3D:completion:)`. ([#2827](https://github.com/mapbox/mapbox-navigation-ios/pull/2827))
 
 ### Location tracking
 
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -23,6 +23,8 @@ class ViewController: UIViewController {
     typealias RouteRequestSuccess = ((RouteResponse) -> Void)
     typealias RouteRequestFailure = ((Error) -> Void)
     
+    private var foundAllBuildings = false
+
     var navigationMapView: NavigationMapView! {
         didSet {
             if let navigationMapView = oldValue {
@@ -132,7 +134,7 @@ class ViewController: UIViewController {
         clearMap.isHidden = true
         longPressHintView.isHidden = false
         
-        // TODO: Unhighlight buildings when clearing map.
+        navigationMapView.unhighlightBuildings()
         navigationMapView.removeRoutes()
         navigationMapView.removeWaypoints()
         waypoints.removeAll()
@@ -324,6 +326,10 @@ class ViewController: UIViewController {
         waypoint.targetCoordinate = destinationCoordinate
         waypoints.append(waypoint)
         
+        // Example of highlighting buildings in 3d and directly using the API on NavigationMapView.
+        let buildingHighlightCoordinates = waypoints.compactMap { $0.targetCoordinate }
+        navigationMapView.highlightBuildings(at: buildingHighlightCoordinates)
+
         requestRoute()
     }
     
--- a/MapboxNavigation.xcodeproj/project.pbxproj
+++ b/MapboxNavigation.xcodeproj/project.pbxproj
@@ -290,6 +290,7 @@
 		AE87207E22CF97B900D7DAB7 /* InstructionsCardCollectionDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE87207D22CF97B900D7DAB7 /* InstructionsCardCollectionDelegate.swift */; };
 		AEC3AC9A2106703100A26F34 /* HighwayShield.swift in Sources */ = {isa = PBXBuildFile; fileRef = AEC3AC992106703100A26F34 /* HighwayShield.swift */; };
 		AED6285622CBE4CE00058A51 /* ViewController+GuidanceCards.swift in Sources */ = {isa = PBXBuildFile; fileRef = AED6285522CBE4CE00058A51 /* ViewController+GuidanceCards.swift */; };
+		B419BFF225F00A9C0086639B /* Feature.swift in Sources */ = {isa = PBXBuildFile; fileRef = B419BFF125F00A9C0086639B /* Feature.swift */; };
 		B430D2FA25534FDC0088CC23 /* UserHaloCourseView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B430D2F925534FDC0088CC23 /* UserHaloCourseView.swift */; };
 		B48CEFC125796FD300696BB3 /* route-for-vanishing-route-line.json in Resources */ = {isa = PBXBuildFile; fileRef = B48CEFAB25796F5A00696BB3 /* route-for-vanishing-route-line.json */; };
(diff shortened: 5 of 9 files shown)


Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -15,13 +15,122 @@ extension NavigationMapView {
...
+    private func buildingIdentifier(at coordinate: CLLocationCoordinate2D, in3D extrudesBuildings: Bool = true) {
+        let screenCoordinate = mapView.point(for: coordinate, in: self)
+        if let identifiers = try? mapView.__map.getStyleLayers().compactMap({$0.id}).filter({ $0.contains("building") }) {
+            mapView.visibleFeatures(at: screenCoordinate,
+                                    styleLayers: Set(identifiers),
+                                    completion: { [weak self] result in
+                                        guard let validSelf = self else { return }
+                                        if case .success(let features) = result {
+                                            if let identifier = self?.featureIdentifierToDouble(features: features) {
+                                                self?.foundBuildingIds.insert(identifier)
```
MaximAlien, line 55: Instead of adding `foundBuildingIds` as member variable this method should provide callback, which will be called whenever `visibleFeatures()` is done searching for building identifiers. Public method `highlightBuildings()` will have to be updated to call that callback as well, instead of returning boolean value synchronously.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -15,13 +15,122 @@ extension NavigationMapView {
...
+                switch numberId {
+                case .double(let doubleId): return Double(doubleId)
+                case .int(let intId): return Double(intId)
+                }
+            }
+        }
+        return nil
+    }
+
+    private func buildingIdentifier(at coordinate: CLLocationCoordinate2D, in3D extrudesBuildings: Bool = true) {
```
MaximAlien, line 46: Do we need to use `extrudesBuildings` here?

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -15,13 +15,122 @@ extension NavigationMapView {
      - returns: Bool indicating if number of buildings found equals number of coordinates supplied.
      */
     @discardableResult public func highlightBuildings(at coordinates: [CLLocationCoordinate2D], in3D extrudesBuildings: Bool = true) -> Bool {
-        return false
+        let _ = coordinates.map({buildingIdentifier(at: $0, in3D: extrudesBuildings) })
+        DispatchQueue.main.asyncAfter(deadline: .now() + handleAsyncTime, execute: {self.addBuildingsLayer(with: self.foundBuildingIds, in3D: extrudesBuildings)})
```
MaximAlien, line 19: Why do you need to run this command asynchronously? Should be similar to previous implementation (besides completion callback).

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -15,13 +15,122 @@ extension NavigationMapView {
...
+                "true"
+            }
+        } else {
+            highlightedBuildingsLayer.filter = Exp(.all) {
+                Exp(.eq) {
+                    Exp(.get) {
+                        "extrude"
+                    }
+                    "true"
+                }
```
MaximAlien, line 83: Shall we store this expression as local variable to be able to re-use it in several places (I can see two usages)?

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -15,13 +15,122 @@ extension NavigationMapView {
...
+                        "height"
+                    }
+                }
+            )
+        } else {
+            highlightedBuildingsLayer.paint?.fillExtrusionHeight = .constant(0.0)
+        }
+
+        highlightedBuildingsLayer.paint?.fillExtrusionBase = .expression(
+            Exp(.interpolate) {
```
MaximAlien, line 111: Can we also re-use this expression syntax (I can see only height/min_height difference)?

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion! I'll remove this part.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Because the `mapView.visibleFeatures(at: styleLayers:,completion:` has an escaping closure. And it only allow the feature querying at one point each time. But  the SDK should allow users to choose multiple waypoints by longPress, and these buildings should be highlighted at the same time after chosen. When dealing with multiple points, we'd better collect all the building IDs from the `mapView.visibleFeatures(at: styleLayers:,completion:` result, and do the extrusion layer filter by these IDs.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: The previous completion callback in 1.x is not working in the Carbon due to the `mapView.visibleFeatures(at: styleLayers:,completion:` having an escaping closure now. The asynchronous completion result would cause the buildingIds delayed with an empty set passed to `func addBuildingsLayer(with)`

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion! I'll extract these expressions as a local variable.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion! I'll extract these expressions as a local variable.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion! I'll extract these expressions as a local variable.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -15,13 +15,99 @@ extension NavigationMapView {
...
      */
     public func unhighlightBuildings() {
-        
+        guard let _ = try? mapView.style.getLayer(with: IdentifierString.buildingExtrusionLayer, type: FillExtrusionLayer.self).get() else { return }
+        foundBuildingIds = Set<Double>()
+        let _ = mapView.style.removeStyleLayer(forLayerId: IdentifierString.buildingExtrusionLayer)
     }
+
+    private func featureIdentifierToDouble(features: [Feature]) -> Double? {
+        if let featureIdentifier = features.first?.identifier {
```
MaximAlien, line 33: Early return with `guard let` would be better here.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -15,13 +15,99 @@ extension NavigationMapView {
...
      Removes the highlight from all buildings highlighted by `highlightBuildings(at:in3D:)`.
      */
     public func unhighlightBuildings() {
-        
+        guard let _ = try? mapView.style.getLayer(with: IdentifierString.buildingExtrusionLayer, type: FillExtrusionLayer.self).get() else { return }
+        foundBuildingIds = Set<Double>()
+        let _ = mapView.style.removeStyleLayer(forLayerId: IdentifierString.buildingExtrusionLayer)
     }
+
+    private func featureIdentifierToDouble(features: [Feature]) -> Double? {
```
MaximAlien, line 32: This function can be moved to be part of `Feature` extension.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
MaximAlien: In case if `buildingIdentifier` runs async operation it should provide callback as well. `highlightBuildings()` will also have to be updated to provide callback instead of returning boolean.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
MaximAlien: So you mean that completion handler in `visibleFeatures()` will be called multiple times (depending on number of `identifiers`)? If so we can provide something similar to `DispatchGroup` to wait till all operations complete, and only after that call our completion callback.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion! I'll create a `Feature` extension to store it.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion! I'll work on that.

ShanMa1991: Thanks for the suggestion! I'll work on that.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Yes. The function of `visibleFeatures()` and `func buildingIdentifier(at:)` will be called multiple times, each time it only handles one waypoint. I'll try to use `DispatchGroup ` to handle this. Thanks for the suggestion!

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion, I've added the completion and call back for the `func highlightBuildings `.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -11,17 +11,93 @@ extension NavigationMapView {
...
+                                        guard let validSelf = self else { return }
+                                        if case .success(let features) = result {
+                                            if let identifier = features.first?.featureIdentifierToDouble() {
+                                                self?.foundBuildingIds.insert(identifier)
+                                            }
+                                            group.leave()
+                                        }})
+        })
+
+        group.notify(queue: DispatchQueue.main) {
```
MaximAlien, line 38: You should wait till all operations complete, instead of notifying right away. After finishing all operations completion block can be called with boolean parameter which will tell whether all identifiers were found.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -11,17 +11,93 @@ extension NavigationMapView {
...
+        }
     }
     
     /**
      Removes the highlight from all buildings highlighted by `highlightBuildings(at:in3D:)`.
      */
     public func unhighlightBuildings() {
-        
+        guard let _ = try? mapView.style.getLayer(with: IdentifierString.buildingExtrusionLayer, type: FillExtrusionLayer.self).get() else { return }
+        foundBuildingIds = Set<Double>()
```
MaximAlien, line 48: Why do you keep exposing `foundBuildingIds` as a member variable? It can be stored locally as client doesn't need it. Instead completion handler should have boolean parameter to check whether all identifiers were found.

ShanMa1991: Thanks for the suggestion, I'll change it to local.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion, I'll update this part.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -11,17 +11,96 @@ extension NavigationMapView {
...
     /**
      Removes the highlight from all buildings highlighted by `highlightBuildings(at:in3D:)`.
      */
     public func unhighlightBuildings() {
-        
+        guard let _ = try? mapView.style.getLayer(with: IdentifierString.buildingExtrusionLayer, type: FillExtrusionLayer.self).get() else { return }
+        mapView.style.removeStyleLayer(forLayerId: IdentifierString.buildingExtrusionLayer)
+    }
+
+    private func addBuildingsLayer(with identifiers: Set<Double>, in3D: Bool = false, extrudeAll: Bool = false) {
```
MaximAlien, line 55: Why identifiers are exposed as `Double`? I think we should keep Int64/UInt64, as before.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -11,17 +11,96 @@ extension NavigationMapView {
...
+        }
     }
     
     /**
      Removes the highlight from all buildings highlighted by `highlightBuildings(at:in3D:)`.
      */
     public func unhighlightBuildings() {
-        
+        guard let _ = try? mapView.style.getLayer(with: IdentifierString.buildingExtrusionLayer, type: FillExtrusionLayer.self).get() else { return }
+        mapView.style.removeStyleLayer(forLayerId: IdentifierString.buildingExtrusionLayer)
```
MaximAlien, line 52: Calls to this method produces warning if result is not handled. Please handle result even if it's not used so that we do not get any warnings.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -11,17 +11,96 @@ extension NavigationMapView {
...
      - parameter coordinates: Coordinates which represent building locations.
      - parameter extrudesBuildings: Switch which allows to highlight buildings in either 2D or 3D. Defaults to true.
-     
-     - returns: Bool indicating if number of buildings found equals number of coordinates supplied.
+     - parameter completion: An escaping closure that operates after the buildings are highted.
+
      */
-    @discardableResult public func highlightBuildings(at coordinates: [CLLocationCoordinate2D], in3D extrudesBuildings: Bool = true) -> Bool {
-        return false
+    public func highlightBuildings(at coordinates: [CLLocationCoordinate2D],
```
MaximAlien, line 17: Can you cover this API with basic unit-tests?

Sources/MapboxNavigation/Feature.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+import Turf
+
+extension Feature {
+
+    func featureIdentifierToDouble() -> Double? {
```
MaximAlien, line 6: Identifier not be exposed as `Double`.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -11,17 +11,96 @@ extension NavigationMapView {
...
+                                   in3D extrudesBuildings: Bool = true,
+                                   completion: @escaping (_ result: Bool) -> Void) {
+        var foundBuildingIds = Set<Double>()
+        let group = DispatchGroup()
+        if let identifiers = try? mapView.__map.getStyleLayers().compactMap({$0.id}).filter({ $0.contains("building") }) {
+            let _ = coordinates.map({
```
MaximAlien, line 23: Can you replace `map` with `foreach`? Please also return early in case if identifiers are not available by using `guard let`.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -11,17 +11,96 @@ extension NavigationMapView {
      
      - parameter coordinates: Coordinates which represent building locations.
      - parameter extrudesBuildings: Switch which allows to highlight buildings in either 2D or 3D. Defaults to true.
-     
-     - returns: Bool indicating if number of buildings found equals number of coordinates supplied.
+     - parameter completion: An escaping closure that operates after the buildings are highted.
```
MaximAlien, line 14:
```suggestion
     - parameter completion: An escaping closure to be executed when the `MapView` feature querying for all `coordinates` ends. A single `Boolean` argument in closure indicates whether or not buildings were found for all coordinates.
```

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
```diff
@@ -11,17 +11,96 @@ extension NavigationMapView {
...
+        var foundBuildingIds = Set<Double>()
+        let group = DispatchGroup()
+        if let identifiers = try? mapView.__map.getStyleLayers().compactMap({$0.id}).filter({ $0.contains("building") }) {
+            let _ = coordinates.map({
+                group.enter()
+                let screenCoordinate = mapView.point(for: $0, in: self)
+                mapView.visibleFeatures(at: screenCoordinate,
+                                        styleLayers: Set(identifiers),
+                                        completion: { [weak self] result in
+                                            guard let validSelf = self else { return }
```
MaximAlien, line 29: `validSelf` usage produces warning here, because it's not used.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: I have tested that with simple map examples. It seems that the filtering of the building ids in `Carbon` only accept `Double` instead of `Int`.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion! I'll add the result part.

Sources/MapboxNavigation/Feature.swift
ShanMa1991: I've run some tests for this part. The identifier in `Carbon` for filtering only accept the `Double` format, instead of the `Int64 ` in 1.x.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion! I'll work on that.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion, I'll work on that.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion, I'll improve this one.

Sources/MapboxNavigation/NavigationMapView+BuildingHighlighting.swift
ShanMa1991: Thanks for the suggestion, I'll remove this part.

MaximAlien (approved):
Overall looks good. Please squash all existing commits before instead of merging all of them.

MaximAlien:
Suggested changelog entry:
```
* `NavigationMapView.highlightBuildings(at:in3D:)` was renamed to `NavigationMapView.highlightBuildings(at:in3D:completion:)`. ([#2827](https://github.com/mapbox/mapbox-navigation-ios/pull/2827))
```

ShanMa1991:
> Suggested changelog entry:
> 
> ```
> * `NavigationMapView.highlightBuildings(at:in3D:)` was renamed to `NavigationMapView.highlightBuildings(at:in3D:completion:)`. ([#2827](https://github.com/mapbox/mapbox-navigation-ios/pull/2827))
> ```

Thanks for the suggestion, I'll update the changelog.

PR #2829, opened 2021-02-23
Simplified CMapboxCoreNavigation

### LINKED ISSUE #2824: Archivation fails for Navigation SDK 2.0 integrated with SPM
**Mapbox Navigation SDK version:** v2.0.0-alpha.1

### Steps to reproduce

 1. Integrate Mapbox Navigation SDK with SPM using `2.0.0-alpha.1` version
 2. Run "Archive" task in Xcode

### Expected behavior

The app builds with no errors.

### Actual behavior

> /Users/bamx23/Library/Developer/Xcode/DerivedData/MyApp-asddjsghfaiuysdtgfoasdgfoaasfjkasdh/SourcePackages/checkouts/mapbox-navigation-ios/Sources/CMapboxCoreNavigation/include/MBXAccounts+CoreNavigationAdditions.h:1:9: 'MapboxAccounts/MapboxAccounts.h' file not found

Screen Shot 2021-02-22 at 17 30 12

1ec5:
Removed the redeclaration of MBXPeerWrapper, now that MapboxNavigationNative imports MapboxCommon in the header that uses it. Removed the dependency on [MapboxAccounts](https://github.com/mapbox/mapbox-accounts-ios/), now that MapboxCommon provides similar functionality.

~~Removed the now-empty CMapboxCoreNavigation target from the SPM build system. Removed the last vestiges of Objective-C from the Carthage and CocoaPods build systems, including the mostly empty Objective-C umbrella headers. Applications that need to check the navigation SDK’s build number or marketing version at runtime can obtain the MapboxNavigation bundle and get its Info.plist’s `CFBundleVersion` or `CFBundleShortVersionString` key, respectively.~~

Fixes #2824 (but only for v2._x_).

/ref mapbox/mapbox-navigation-android#4027
/cc @mapbox/navigation-ios @LukasPaczos @bamx23


1ec5:
> Removed the now-empty CMapboxCoreNavigation target from the SPM build system. Removed the last vestiges of Objective-C from the Carthage and CocoaPods build systems, including the mostly empty Objective-C umbrella headers. Applications that need to check the navigation SDK’s build number or marketing version at runtime can obtain the MapboxNavigation bundle and get its Info.plist’s `CFBundleVersion` or `CFBundleShortVersionString` key, respectively.

I ended up reverting this part of the PR, exciting as it was, because the `+load` method in CMapboxCoreNavigation was still needed as before.

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -339,8 +337,6 @@ public class MapboxNavigationService: NSObject, NavigationService {
     
     public func stop() {
         
-        MBXAccounts.resetSession()
```
bamx23, line 342: Isn't this required when using MapboxCommon?

bamx23 (approved):

Sources/MapboxCoreNavigation/NavigationService.swift
1ec5: No, MapboxCommon doesn’t have an analogue to this method and apparently it isn’t needed for the supported billing methods.

1ec5:
```
    - WARN  | xcodebuild:  /Users/distiller/project/Sources/CMapboxCoreNavigation/include/MBXAccountsLoader.m:10:42: warning: undeclared selector 'serviceSkuToken' [-Wundeclared-selector]
    - ```

chezzdev (approved):

PR #2830, opened 2021-02-24
Predictive Caching Integration

Udumft:
### Description
This PR adds predictive Caching mechanism integration. It provides a configurable setting which enables predictive caching for `Navigator` and/or `Map`.

### Implementation
The PR introduces `PredictiveCacheLocationOptions` as a part of `NavigationOptions` which enables and configures the feature. Creation of predictive cache controllers is encapsulated in `PredictiveCacheManager` which just needs to be retained during the session.

Currently there seems to be some issue with navigation tiles. Tiles seems to be downloading after navigation has began, but it does not handle reroute events. 
Also, subjectively, MapView now takes longer to render at first launch, when there are no tiles downloaded yet. This may be related to providing a custom TileStore path, but I am not sure how and why this would matter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@
 
 * Removed deprecated `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Removed `NavigationViewController.origin` property, which was not used. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
-* Added optional Predictive Caching mechanism which pre-fetches tiles which may become needed during navigation if app goes offline at some point. ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
+* Added `PredictiveCacheOptions` to allow enabling and configuring a Predictive Caching mechanism which pre-fetches tiles which may become needed during navigation if app goes offline at some point. This can be configure either by updating `NavigationOptions` passed to `NavigationViewController` or by manually calling `NavigationMapView.setupPredictiveCaching(_:)` ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
 
 ## v1.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 
 * Removed deprecated `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Removed `NavigationViewController.origin` property, which was not used. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
+* Added optional Predictive Caching mechanism which pre-fetches tiles which may become needed during navigation if app goes offline at some point. ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
 
 ## v1.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@
 
 * Removed deprecated `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Removed `NavigationViewController.origin` property, which was not used. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
-* Added `PredictiveCacheOptions` to allow enabling and configuring a Predictive Caching mechanism which pre-fetches tiles which may become needed during navigation if app goes offline at some point. This can be configure either by updating `NavigationOptions` passed to `NavigationViewController` or by manually calling `NavigationMapView.setupPredictiveCaching(_:)` ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
+* A new predictive cache proactively fetches tiles which may become necessary if the device loses its Internet connection at some point during passive or active turn-by-turn navigation. Pass a `PredictiveCacheOptions` instance into the `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:predictiveCacheOptions:)` initializer as you configure a `NavigationViewController`, or manually call `NavigationMapView.enablePredictiveCaching(options:)`. ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
 
 ## v1.3.0
 
--- a/Sources/MapboxCoreNavigation/PredictiveCacheManager.swift
+++ b/Sources/MapboxCoreNavigation/PredictiveCacheManager.swift
@@ -2,9 +2,9 @@ import Foundation
 import MapboxNavigationNative
 
 /**
- `PredictiveCacheManager` is responsible for creating and retaining `Predictive Caching` related components.
+ Proactively fetches tiles which may become necessary if the device loses its Internet connection at some point during passive or active turn-by-turn navigation.
  
- Typical usage suggests initializing an instance of `PredictiveCacheManager` and retaining it as long as caching is required.
+ Typically, you initialize an instance of this class and retain it as long as caching is required.
  */
 public class PredictiveCacheManager {
     public typealias MapOptions = (tileStore: TileStore, styleSourcePaths: [String])
@@ -12,10 +12,10 @@ public class PredictiveCacheManager {
     private(set) var controllers: [PredictiveCacheController] = []
     
     /**
-     Default initializer
-     
-     - parameter predictiveCacheOptions: `PredictiveCacheOptions` which configures various caching parameters like radiuses of current and destination locations.
-     - parameter mapOptions: A `MapOptions` which contains info about `MapView` tiles like its location and tilesets to be cached. If set to `nil` - predictive caching won't be enbled for map tiles.
+     Initializes a predictive cache.
+    
+     - parameter predictiveCacheOptions: A configuration specifying various caching parameters, such as the radii of current and destination locations.
+     - parameter mapOptions: Information about `MapView` tiles such as the location and tilesets to cache. If this argument is set to `nil`, predictive caching is disabled for map tiles.
      */
     public init(predictiveCacheOptions: PredictiveCacheOptions, mapOptions: MapOptions?) {
         self.controllers.append(initNavigatorController(options: predictiveCacheOptions))
--- a/Sources/MapboxCoreNavigation/PredictiveCacheManager.swift
+++ b/Sources/MapboxCoreNavigation/PredictiveCacheManager.swift
@@ -1,0 +1,57 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ `PredictiveCacheManager` is responsible for creating and retaining `Predictive Caching` related components.
+ 
+ Typical usage suggests initializing an instance of `PredictiveCacheManager` and retaining it as long as caching is required.
+ */
+public class PredictiveCacheManager {
+    public typealias MapOptions = (tileStore: TileStore, styleSourcePaths: [String])
+    
+    private(set) var controllers: [PredictiveCacheController] = []
+    
+    /**
+     Default initializer
+     
+     - parameter predictiveCacheOptions: `PredictiveCacheOptions` which configures various caching parameters like radiuses of current and destination locations.
+     - parameter mapOptions: A `MapOptions` which contains info about `MapView` tiles like its location and tilesets to be cached. If set to `nil` - predictive caching won't be enbled for map tiles.
+     */
+    public init(predictiveCacheOptions: PredictiveCacheOptions, mapOptions: MapOptions?) {
+        self.controllers.append(initNavigatorController(options: predictiveCacheOptions))
+        if let mapOptions = mapOptions {
+            self.controllers.append(contentsOf: initMapControllers(options: predictiveCacheOptions, mapOptions: mapOptions))
+        }
+    }
+    
+    private func initMapControllers(options: PredictiveCacheOptions,
+                                    mapOptions: MapOptions) -> [PredictiveCacheController] {
+        return mapOptions.styleSourcePaths.compactMap {
+            createPredictiveCacheController(options: options,
+                                            tileStore: mapOptions.tileStore,
+                                            dataset: $0)
+        }
+    }
+    
+    private func initNavigatorController(options: PredictiveCacheOptions) -> PredictiveCacheController {
+        return createPredictiveCacheController(options: options)!
+    }
+    
+    private func createPredictiveCacheController(options: PredictiveCacheOptions,
+                                                 tileStore: TileStore? = nil,
+                                                 version: String = "",
+                                                 dataset: String = "mapbox",
+                                                 maxConcurrentRequests: UInt32 = 2) -> PredictiveCacheController? {
+        let cacheOptions = PredictiveCacheControllerOptions(version: version,
+                                                            dataset: dataset,
+                                                            concurrency: maxConcurrentRequests)
+        let predictiveLocationTrackerOptions = PredictiveLocationTrackerOptions(options)
+        if let tileStore = tileStore {
+            return try! Navigator.shared.navigator.createPredictiveCacheController(for: tileStore,
+                                                                                   cacheOptions: cacheOptions,
+                                                                                   locationTrackerOptions: predictiveLocationTrackerOptions)
+        } else {
+            return try! Navigator.shared.navigator.createPredictiveCacheController(for: predictiveLocationTrackerOptions)
+        }
+    }
+}
--- a/Sources/MapboxCoreNavigation/PredictiveCacheOptions.swift
+++ b/Sources/MapboxCoreNavigation/PredictiveCacheOptions.swift
@@ -1,9 +1,11 @@
 import MapboxNavigationNative
 
 /**
- `PredictiveCacheOptions` controls various configurations for a `Predictive Caching` mechanic.
+ Specifies the content that a predictive cache fetches and how it fetches the content.
+ 
+ Pass an instance of this class into the `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:predictiveCacheOptions:)` initializer or `NavigationMapView.enablePredictiveCaching(options:)` method.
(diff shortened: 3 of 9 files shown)


zugaldia:
> Currently there seems to be some issue with navigation tiles. Tiles seems to be downloading after navigation has began, but it does not handle reroute events.

Do we understand the root cause for this issue yet?

/cc: @kkaefer

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@
 * Removed deprecated `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Removed `NavigationViewController.origin` property, which was not used. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Unified usage and configuration of `Navigator` instance under `NavigationProvider`. This also includes single entry point for configuring used tiles version and tile storage location. ([#2823](https://github.com/mapbox/mapbox-navigation-ios/pull/2823)) 
+* Added optional Predictive Caching mechanism which pre-fetches tiles which may become needed during navigation if app goes offline at some point. ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
```
1ec5, line 49: This changelog entry should point out the specific changes in the API, such as renaming `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:)` to `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:predictiveCacheLocationOptions:)`.

Sources/MapboxNavigation/NavigationOptions.swift
```diff
@@ -40,6 +40,14 @@ open class NavigationOptions: NavigationCustomizable {
      */
     open var bottomBanner: ContainerViewController?
     
+    /**
+     Configuration for the Predictive Caching.
+     
+     These parameters control how navigation engine will try to predict and pre-load data related to the route.
+     `nil` value means disabling the feature.
```
1ec5, line 47: Should the default be to not predictively cache, or should the feature be opt-out?

Sources/MapboxNavigation/NavigationOptions.swift
```diff
@@ -53,14 +61,16 @@ open class NavigationOptions: NavigationCustomizable {
      - parameter voiceController: The voice controller that vocalizes spoken instructions along the route at the appropriate times.
      - parameter topBanner: The container view controller that presents the top banner.
      - parameter bottomBanner: The container view controller that presents the bottom banner.
+     - parameter predictiveCacheLocationOptions: Predictive Caching setup.
```
1ec5, line 64: Mention the `nil` behavior here too; this is where developers are more likely to encounter the option.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -324,6 +330,23 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
         }
     }
     
+    func setupPredictiveCaching(_ navigationOptions: NavigationOptions?) {
```
1ec5, line 333: This is code that a developer would have to rewrite if they’re implementing their own turn-by-turn navigation view controller based on NavigationMapView and RouteController. We’ll also have to rewrite it for CarPlayNavigationViewController. Should this code be pushed into NavigationMapView?

Sources/MapboxNavigation/Map.swift
```diff
@@ -0,0 +1,30 @@
+
+import Foundation
+import MapboxCoreMaps
+
+extension Map {
+    func getStyleSourceURLs(_ sourceTypes: [String]) -> [String] {
```
1ec5, line 6: Rename to `styleSourceURLs(types:)`. Even if the map SDK’s public API uses strings to represent URLs, we should use `URL`s in any extensions for added type safety.

Sources/MapboxNavigation/Map.swift
```diff
@@ -0,0 +1,30 @@
...
+                let contents = properties.value as? [String: AnyObject]
+                guard var url = contents?["url"] as? String else {
+                    continue
+                }
+                if url.split(separator: ",").count > 1 || !url.hasPrefix("mapbox://") {
+                    continue // we ignore composite (https://docs.mapbox.com/studio-manual/reference/styles/#source-compositing.) and non-mapbox sources
+                } else {
+                    url.removeFirst("mapbox://".count)
+                    urls.append(url)
+                }
```
1ec5, line 25: `URL` provides more robust URL parsing, for example:

https://github.com/mapbox/mapbox-navigation-ios/blob/a222d2d8bd5d7f9a6b9a0a19746026034314394a/Sources/MapboxNavigation/MGLVectorTileSource.swift#L10-L15

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -262,7 +262,12 @@ open class NavigationMapView: UIView {
             fatalError("Access token was not set.")
         }
         
-        mapView = MapView(with: frame, resourceOptions: ResourceOptions(accessToken: accessToken))
+        let tileStoreURL = FileManager.default.urls(for: .documentDirectory,
```
1ec5, line 265: The document directory is exposed to the user in the Files application and backed up to iCloud by default. It would be more appropriate to stick the cache in Caches:

https://github.com/mapbox/mapbox-navigation-ios/blob/68c9ca655e7e1bf81436a4ce1f6f855f2b0cf221/Sources/MapboxCoreNavigation/BundleAdditions.swift#L60-L67

Caches may be wiped if the device runs out of space; the idea is that caches can be recreated easily. If this is a problem, then we can stick the cache in Application Support, but make sure to [exclude the folder from iCloud](https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/darwin/src/MGLOfflineStorage.mm#L297).

Sources/MapboxCoreNavigation/PredictiveCacheLocationOptions.swift
```diff
@@ -0,0 +1,38 @@
...
+    /// Defaults to 2 km
+    public var currentLocationRadiusInMeters: UInt32
+    /// How far around the active route we're going to cache, _in meters_ (if route is set).
+    ///
+    /// Defaults to 500 m
+    public var routeBufferRadiusInMeters: UInt32
+    /// How far around the destination location we're going to cache, _in meters_ (if route is set).
+    ///
+    /// Defaults to 5 km
+    public var destinationLocationRadiusInMeters: UInt32
```
1ec5, line 19: Use `CLLocationDistance` (which is measured in meters) for any distance measurement in the Swift codebase.

Sources/MapboxCoreNavigation/PredictiveCacheLocationOptions.swift
```diff
@@ -0,0 +1,38 @@
...
+    
+    /// Default initializer
+    public init(currentLocationRadius: UInt32 = 2_000, routeBufferRadius: UInt32 = 500, destinationLocationRadius: UInt32 = 5_000, concurrency: UInt32 = 2) {
+        currentLocationRadiusInMeters = currentLocationRadius
+        routeBufferRadiusInMeters = routeBufferRadius
+        destinationLocationRadiusInMeters = destinationLocationRadius
+        self.concurrency = concurrency
+    }
+    
+    func toPredictiveLocationTrackerOptions() -> PredictiveLocationTrackerOptions {
```
1ec5, line 33: Replace this method with a convenience initializer on an extension to PredictiveLocationTrackerOptions.

Sources/MapboxCoreNavigation/PredictiveCacheLocationOptions.swift
```diff
@@ -0,0 +1,38 @@
+
+import Foundation
+import MapboxNavigationNative
+
+/// `PredictiveCacheLocationOptions` controls various configurations for a `Predictive Caching` mechanic.
+public class PredictiveCacheLocationOptions {
```
1ec5, line 6: This class seems to be about more than just the location of the predictive cache. The “location” in PredictiveLocationTrackerOptions refers to predictive location tracking, so PredictiveLocationOptions or PredictiveCacheOptions would probably be more accurate.

Sources/MapboxCoreNavigation/PredictiveCacheLocationOptions.swift
```diff
@@ -0,0 +1,38 @@
...
+    ///
+    /// Defaults to 5 km
+    public var destinationLocationRadiusInMeters: UInt32
+    /// How many download threads will be used for caching
+    ///
+    /// Defaults to 2
+    public var concurrency: UInt32
+    
+    /// Default initializer
+    public init(currentLocationRadius: UInt32 = 2_000, routeBufferRadius: UInt32 = 500, destinationLocationRadius: UInt32 = 5_000, concurrency: UInt32 = 2) {
```
1ec5, line 26: Move these defaults to the property declarations and remove the arguments from this initializer. If the application wants to customize any of the options, it should set the properties on separate lines after initializing the object.

Sources/MapboxNavigation/NavigationMapView.swift
MaximAlien: Replaced current `tileStoreURL` with `Bundle.mapboxNavigation.suggestedTileURL?.path`.

Sources/MapboxNavigation/Map.swift
MaximAlien: I think that original method naming/signature is not correct as this method returns datasets (e.g. `mapbox.mapbox-streets-v8`), not URLs. I think I'll rename it to something like styleSourceDatasets(_:). It'll keep returning `[String]`.

Sources/MapboxCoreNavigation/PredictiveCacheLocationOptions.swift
MaximAlien: Agree, renamed to `PredictiveCacheOptions`.

Sources/MapboxNavigation/NavigationViewController.swift
MaximAlien: I agree that predictive cache functionality should be part of `NavigationMapView`, this will also allow us to control `PredictiveCacheOptions` via `NavigationMapView` interface, instead of extending `NavigationOptions`.

CHANGELOG.md
Udumft: Updated

Sources/MapboxNavigation/NavigationOptions.swift
Udumft: As I can see from Android implementation - it is up to user to [explicitly enable it](https://github.com/mapbox/mapbox-navigation-android-internal/pull/33/files#diff-c88da1a719e50faaaab70f7dc284ae633128af9325f60a0b7355087bb19e6515R163-R166).

Sources/MapboxNavigation/NavigationViewController.swift
Udumft: I think you are right, it makes sense to have this in `NavigationMapView` instead. But I still think that we should have `PredictiveCacheOptions` inside `NavigationOptions` since this is our default way for customising navigation experience when used together with `NavigationViewController`. At the same time it'll be still possible to configure Predictive Caching from `NavigationMapView` directly.

Sources/MapboxNavigation/NavigationOptions.swift
```diff
@@ -53,14 +61,16 @@ open class NavigationOptions: NavigationCustomizable {
      - parameter voiceController: The voice controller that vocalizes spoken instructions along the route at the appropriate times.
      - parameter topBanner: The container view controller that presents the top banner.
      - parameter bottomBanner: The container view controller that presents the bottom banner.
+     - parameter predictiveCacheOptions: Predictive Caching setup.
```
1ec5, line 64:
```suggestion
     - parameter predictiveCacheOptions: Configuration for predictive caching. These options control how the `PredictiveCacheManager` will try to proactively fetch data related to the route. A `nil` value disables the feature.
```

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 
 * Removed deprecated `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Removed `NavigationViewController.origin` property, which was not used. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
+* Added `PredictiveCacheOptions` to allow enabling and configuring a Predictive Caching mechanism which pre-fetches tiles which may become needed during navigation if app goes offline at some point. This can be configure either by updating `NavigationOptions` passed to `NavigationViewController` or by manually calling `NavigationMapView.setupPredictiveCaching(_:)` ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
```
1ec5, line 50: Now that I’m looking closely at this changelog entry, it makes me wonder if we should’ve called the feature the “proactive cache” for consistency with the longstanding feature known as proactive rerouting. But it’s a bit too late for that change of terminology, I suppose.

```suggestion
* A new predictive cache proactively fetches tiles which may become necessary if the device loses its Internet connection at some point during passive or active turn-by-turn navigation. Pass a `PredictiveCacheOptions` instance into the `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:predictiveCacheOptions:)` initializer as you configure a `NavigationViewController`, or manually call `NavigationMapView.enablePredictiveCaching(options:)`. ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
```

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -286,6 +294,17 @@ open class NavigationMapView: UIView {
         mapView.addGestureRecognizer(mapTapGesture)
     }
     
+    public func setupPredictiveCaching(_ predictiveCacheOptions: PredictiveCacheOptions) {
```
1ec5, line 297: This public method needs a documentation comment. `predictiveCacheOptions` isn’t a “predictive caching”, so there should be a label on this argument. Since the options argument is required, this method will always enable predictive caching, so we can call it `enablePredictiveCaching(options:)`.

Sources/MapboxCoreNavigation/PredictiveCacheOptions.swift
```diff
@@ -0,0 +1,48 @@
+import MapboxNavigationNative
+
+/**
+ `PredictiveCacheOptions` controls various configurations for a `Predictive Caching` mechanic.
+ */
+public class PredictiveCacheOptions {
```
1ec5, line 6: Looks like we aren’t using reference semantics anywhere with this class, so we can make it a struct.

Sources/MapboxCoreNavigation/PredictiveCacheOptions.swift
```diff
@@ -0,0 +1,48 @@
...
+     Defaults to 5000 meters.
+     */
+    public var destinationLocationRadius: CLLocationDistance = 5000
+    
+    /**
+     Maxiumum amount of concurrent requests, which will be used for caching.
+     
+     Defaults to 2 concurrent requests.
+     */
+    public var maxConcurrentRequests: UInt32 = 2
```
1ec5, line 34: Spell out “maximum”.

Sources/MapboxCoreNavigation/PredictiveCacheManager.swift
```diff
@@ -0,0 +1,57 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ `PredictiveCacheManager` is responsible for creating and retaining `Predictive Caching` related components.
+ 
+ Typical usage suggests initializing an instance of `PredictiveCacheManager` and retaining it as long as caching is required.
```
1ec5, line 7:
```suggestion
 Proactively fetches tiles which may become necessary if the device loses its Internet connection at some point during passive or active turn-by-turn navigation.
 
 Typically, you initialize an instance of this class and retain it as long as caching is required.
```

Sources/MapboxCoreNavigation/PredictiveCacheManager.swift
```diff
@@ -0,0 +1,57 @@
...
+public class PredictiveCacheManager {
+    public typealias MapOptions = (tileStore: TileStore, styleSourcePaths: [String])
+    
+    private(set) var controllers: [PredictiveCacheController] = []
+    
+    /**
+     Default initializer
+     
+     - parameter predictiveCacheOptions: `PredictiveCacheOptions` which configures various caching parameters like radiuses of current and destination locations.
+     - parameter mapOptions: A `MapOptions` which contains info about `MapView` tiles like its location and tilesets to be cached. If set to `nil` - predictive caching won't be enbled for map tiles.
```
1ec5, line 18: In general, the type of a parameter doesn’t need to be mentioned explicitly, because jazzy and Quick Help will already display and link the type automatically.

```suggestion
     Initializes a predictive cache.
     
     - parameter predictiveCacheOptions: A configuration specifying various caching parameters, such as the radii of current and destination locations.
     - parameter mapOptions: Information about `MapView` tiles such as the location and tilesets to cache. If this argument is set to `nil`, predictive caching is disabled for map tiles.
```

Sources/MapboxCoreNavigation/PredictiveCacheOptions.swift
```diff
@@ -0,0 +1,48 @@
+import MapboxNavigationNative
+
+/**
+ `PredictiveCacheOptions` controls various configurations for a `Predictive Caching` mechanic.
```
1ec5, line 4:
```suggestion
 Specifies the content that a predictive cache fetches and how it fetches the content.
 
 Pass an instance of this class into the `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:predictiveCacheOptions:)` initializer or `NavigationMapView.enablePredictiveCaching(options:)` method.
```

Sources/MapboxNavigation/NavigationOptions.swift
1ec5, line 47:
```suggestion
     Configuration for predictive caching.
     
     These options control how the `PredictiveCacheManager` will try to proactively fetch data related to the route. A `nil` value disables the feature.
```

1ec5 (approved):
Mostly documentation copyediting, but there are a couple refactoring suggestions that will hopefully be straightforward.

Issue #2078, opened 2019-04-04
Verify jazzy finds USRs on CI


1ec5:
To prevent #1795 from regressing a third time or less widespread USR issues from cropping up on individual symbols, we should add a step to the CircleCI configuration that runs jazzy and looks for any USR in a generated HTML file. Even better, we could parse jazzy’s output to look for USR warnings and fix the few that remain.

/cc @mapbox/navigation-ios

1ec5:
It happened again: #2312.

S2Ler:
Shall we prioritize this? @truburt @1ec5

chezzdev:
I'm working on generating docs on CI, so I'll integrate warning checks in that scope.

PR #2832, opened 2021-02-24
Implement support for CPTemplateApplicationSceneDelegate UIScene-based CarPlay on iOS 13 and above

### LINKED ISSUE #2260: Ensure compatibility with CPTemplateApplicationSceneDelegate
In iOS 13, [CPApplicationDelegate](https://developer.apple.com/documentation/carplay/cpapplicationdelegate) has been deprecated in favor of [CPTemplateApplicationSceneDelegate](https://developer.apple.com/documentation/carplay/cptemplateapplicationscenedelegate). I _think_ the developer can safely call `CarPlayManager.application(_:didConnectCarInterfaceController:to:)` and `application(_:didDisconnectCarInterfaceController:from:)` from within `CPTemplateApplicationSceneDelegate.templateApplicationScene(_:didConnect:to:)` and `templateApplicationScene(_:didDisconnect:from:)`, respectively, but we should add tests to be sure and also update the documentation. (Apple hasn’t even gotten around to updating [their CarPlay getting started documentation](https://developer.apple.com/documentation/carplay/adding_carplay_support_to_your_navigation_app) yet.)

/cc @mapbox/navigation-ios @mapbox/docs

avi-c:
### Description

In iOS 13 Apple added a Template-based mechanism for providing CarPlay support to apps. The Nav SDK needs to include support for these newer APIs while ensuring we don't break the earlier iOS 12 CarPlay delegate pattern until we drop support for iOS 12.

Fixes #2260 

### Implementation
The changes required within the Nav SDK are fairly minimal, mainly adding conformance to the CPTemplateApplicationSceneDelegate protocol. More extensive changes are needed to the Example app in order to adopt UIScene support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Added an optional `StyleManagerDelegate.styleManager(_:viewForApplying:)` method to determine which part of the view hierarchy is affected by a change to a different `Style`. ([#2897](https://github.com/mapbox/mapbox-navigation-ios/pull/2897))
 * Added the `NavigationViewController.styleManager`, `StyleManager.currentStyleType`, and `StyleManager.currentStyle` properties and the `StyleManager.applyStyle(type:)` method to manually change the UI style at any time. ([#2888](https://github.com/mapbox/mapbox-navigation-ios/pull/2888))
 * Fixed crash when switching between day / night modes ([#2896](https://github.com/mapbox/mapbox-navigation-ios/pull/2896))
+* Add support for iOS 13's UIScene based CarPlay API ([#2832](https://github.com/mapbox/mapbox-navigation-ios/pull/2832))
 
 ## v1.3.0
 
--- a/Example/AppDelegate.swift
+++ b/Example/AppDelegate.swift
@@ -54,3 +54,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
     }
 }
+
+@available(iOS 13.0, *)
+extension AppDelegate: UIWindowSceneDelegate {
+
+    func application(_ application: UIApplication, configurationForConnecting connectingSceneSession: UISceneSession, options: UIScene.ConnectionOptions) -> UISceneConfiguration {
+
+        if connectingSceneSession.role == .carTemplateApplication {
+            return UISceneConfiguration(name: "ExampleCarPlayApplicationConfiguration", sessionRole: connectingSceneSession.role)
+        }
+        return UISceneConfiguration(name: "ExampleAppConfiguration", sessionRole: connectingSceneSession.role)
+    }
+
+    func application(_ application: UIApplication, didDiscardSceneSessions sceneSessions: Set<UISceneSession>) {
+
+    }
+}
--- a/Example/AppDelegate.swift
+++ b/Example/AppDelegate.swift
@@ -1,8 +1,6 @@
 import UIKit
 import MapboxNavigation
-#if canImport(CarPlay)
 import CarPlay
-#endif
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
@@ -15,7 +13,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     @available(iOS 12.0, *)
     lazy var carPlaySearchController: CarPlaySearchController = CarPlaySearchController()
 
-    #if canImport(CarPlay)
     @available(iOS 12.0, *)
     lazy var interfaceController: CPInterfaceController? = nil
 
@@ -24,7 +21,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     @available(iOS 12.0, *)
     lazy var sessionConfiguration: CPSessionConfiguration? = nil
-    #endif
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         if isRunningTests() {
--- a/Example/AppDelegate.swift
+++ b/Example/AppDelegate.swift
@@ -1,5 +1,8 @@
 import UIKit
 import MapboxNavigation
+#if canImport(CarPlay)
+import CarPlay
+#endif
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
@@ -11,6 +14,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     
     @available(iOS 12.0, *)
     lazy var carPlaySearchController: CarPlaySearchController = CarPlaySearchController()
+
+    #if canImport(CarPlay)
+    @available(iOS 12.0, *)
+    lazy var interfaceController: CPInterfaceController? = nil
+
+    @available(iOS 12.0, *)
+    lazy var carWindow: CPWindow? = nil
+
+    @available(iOS 12.0, *)
+    lazy var sessionConfiguration: CPSessionConfiguration? = nil
+    #endif
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         if isRunningTests() {
--- a/Example/AppDelegate.swift
+++ b/Example/AppDelegate.swift
@@ -28,6 +28,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         if isRunningTests() {
+            if window == nil {
+                window = UIWindow(frame: UIScreen.main.bounds)
+            }
             window!.rootViewController = UIViewController()
         }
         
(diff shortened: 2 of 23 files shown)


avi-c:
Added UIScene and CPTemplateApplicationSceneDelegate support to the Example app for iOS 13 and above. Allows us to use newer iOS 13 CarPlay SDK.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Fixed an issue where lane guidance icons would indicate the wrong arrow for certain maneuvers. ([#2796](https://github.com/mapbox/mapbox-navigation-ios/pull/2796), [#2809](https://github.com/mapbox/mapbox-navigation-ios/pull/2809) 
 * Fixed a crash showing a junction view. ([#2805](https://github.com/mapbox/mapbox-navigation-ios/pull/2805))
 * Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
+* Add support for iOS 13's UIScene based CarPlay API ([#2832](https://github.com/mapbox/mapbox-navigation-ios/pull/2832))
```
1ec5, line 11: What does this mean in terms of the developer’s migration path? Should we say something like this?

```suggestion
* Added support for the `UISceneDelegateClassName` Info.plist key for integrating `CarPlayManager` with `UIScene` on iOS 13 and above. ([#2832](https://github.com/mapbox/mapbox-navigation-ios/pull/2832))
```

Example/AppDelegate.swift
```diff
@@ -12,8 +15,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     @available(iOS 12.0, *)
     lazy var carPlaySearchController: CarPlaySearchController = CarPlaySearchController()
 
+    #if canImport(CarPlay)
+    @available(iOS 12.0, *)
+    lazy var interfaceController: CPInterfaceController? = nil
```
1ec5, line 20: It’s possible for AppDelegate to refer to CarPlaySearchController even in the normal Example target because CarPlaySearchController.swift defines a stub for CarPlaySearchController if CarPlay is unimportable. That said, I’m not sure the conditional compilation is required any longer, now that we no longer support compiling the SDK using Xcode 10.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -901,7 +901,7 @@
 		C53C19751F38EADE008DB406 /* fr */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = fr; path = fr.lproj/Localizable.strings; sourceTree = "<group>"; };
 		C53C19771F38EAE4008DB406 /* ca */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.plist.strings; name = ca; path = ca.lproj/Localizable.strings; sourceTree = "<group>"; };
 		C53C197A1F38EAEA008DB406 /* Base */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = Base; path = Base.lproj/Localizable.strings; sourceTree = "<group>"; };
-		C53F2F0720EBC95600D9798F /* Example-CarPlay.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = "Example-CarPlay.app"; sourceTree = BUILT_PRODUCTS_DIR; };
+		C53F2F0720EBC95600D9798F /* Example_CarPlay.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = Example_CarPlay.app; sourceTree = BUILT_PRODUCTS_DIR; };
```
1ec5, line 904: Is there a particular reason why the application needed to be renamed from Example-CarPlay to Example_CarPlay? I guess it doesn’t hurt, but it could make it more difficult to merge these changes into the release-v2.0 branch, and we might have other stray references to Example-CarPlay.

avi-c: A couple of things. The UIScene Manifest in the Info.plist file needs an entry for the UISceneDelegate class name. In order for it to match, it needs to contain the product name and the class name, such as "Example.AppDelegate". If the product contains a dash symbol, like Example-CarPlay, the product name in the compiled app will be changed to Example_CarPlay since dashes aren't allowed in the class names. So if I leave our product name as `Example-CarPlay` and try to put an entry in the Info.plist for the UISceneDelegate class name, it will look for `Example_CarPlay.AppDelegate` as the class and fail to find it because the class will actually have a name of `Example-CarPlay.AppDelegate`.

The easiest way to resolve this, short of having a different Info.plist file per target seemed to be to rename the product to Example_CarPlay.

If there are more preferred ways to resolve this I am happy to investigate them.

Example/AppDelegate.swift
avi-c: It got a bit messy because the UIScene API requires the ability to save references to `CPInterfaceController` and `CPWindow` but I can't add the variables in a class extension so I decided to just add them to the AppDelegate base class, but it does make it less clean than the old model of all the CarPlay delegate methods being added as class extensions to `AppDelegate`.

If it's possible to remove the `#if canImport(CarPlay)` conditional then maybe we should just do that and clean up the code a bit. That would make my changes a bit less messy.

MapboxNavigation.xcodeproj/project.pbxproj
1ec5: The Example-CarPlay target’s `PRODUCT_MODULE_NAME` build setting is already set to `$(PRODUCT_NAME:c99extidentifier)`, which resolves to `Example_CarPlay`, so `Example_CarPlay.AppDelegate` would be the class name at runtime even without these changes.

avi-c: I'll try that. I prefer not to make these kinds of product name changes if there is an alternative like you mention.

avi-c: @1ec5 - Check out what I did for this. I had to hardcode the product name to "Example" since I can't change the CarPlay target name to Example because of the regular target. I tried forming a variable expression for it but couldn't figure out if that's possible for the setup we want.

avi-c: @1ec5 - Let me know if this seems sufficient. I'd love to finish this off soon.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -901,7 +901,7 @@
 		C53C19751F38EADE008DB406 /* fr */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = fr; path = fr.lproj/Localizable.strings; sourceTree = "<group>"; };
 		C53C19771F38EAE4008DB406 /* ca */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.plist.strings; name = ca; path = ca.lproj/Localizable.strings; sourceTree = "<group>"; };
 		C53C197A1F38EAEA008DB406 /* Base */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = Base; path = Base.lproj/Localizable.strings; sourceTree = "<group>"; };
-		C53F2F0720EBC95600D9798F /* Example-CarPlay.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = "Example-CarPlay.app"; sourceTree = BUILT_PRODUCTS_DIR; };
+		C53F2F0720EBC95600D9798F /* Example.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = Example.app; sourceTree = BUILT_PRODUCTS_DIR; };
```
1ec5, line 904: This means it isn’t possible to install the normal Example and CarPlay-enabled Example applications simultaneously on a single device or simulator. Not a huge deal, but it may be surprising at first.

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -735,15 +782,3 @@ internal class MapTemplateProvider: NSObject {
         return CPMapTemplate()
     }
 }
-
-#else
-/**
- CarPlay support requires iOS 12.0 or above and the CarPlay framework.
- */
-public class CarPlayManager: NSObject {
```
1ec5, line 743: Finally, jazzy will stop warning about missing USRs for these conditionally compiled symbols. 😌

/ref #2078

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Fixed an issue where lane guidance icons would indicate the wrong arrow for certain maneuvers. ([#2796](https://github.com/mapbox/mapbox-navigation-ios/pull/2796), [#2809](https://github.com/mapbox/mapbox-navigation-ios/pull/2809))
 * Fixed a crash showing a junction view. ([#2805](https://github.com/mapbox/mapbox-navigation-ios/pull/2805))
 * Fixed an issue with CarPlay visual instructions where U-Turn maneuver icons were not being flipped properly based on regional driving side ([#2803](https://github.com/mapbox/mapbox-navigation-ios/pull/2803)) 
+* Add support for iOS 13's UIScene based CarPlay API ([#2832](https://github.com/mapbox/mapbox-navigation-ios/pull/2832))
```
1ec5, line 11: This should go in a new v1.4.0 section.

1ec5 (approved):

avi-c:
Made the change to the CHANGELOG entry as requested and rebased to `main`

PR #2833, opened 2021-02-25
Add the ability to show specific route leg with color overriding

ShanMa1991:
### Description
In 1.x , we're not dealing with the legIndex as in 0.4. This pr is to port the ability to handle specific legIndex for mapView route line. 

The way we specify the leg index and change the route layer color should be like following:
```
navigationMapView.routeCasingColor = UIColor.red
navigationMapView.trafficUnknownColor = UIColor.blue
navigationMapView.show([route], legIndex: 1)
```
The color could be override. If we don't specify the leg Index, the whole route would show congestion color. But if we specify the leg index, only the specified legIndex would show the congestion color, while other legs would show the `routeCasingColor`. 

- [x] allow to show specific route leg
- [x] allow to show current route leg during turn-by-turn navigation with overriding color
- [x] update the changeLog

### Implementation
When generating the congestion features for the whole route, the `CurrentLegAttribute`  of the feature property would represent whether this leg is the specified leg. If the route leg is not specified, or it is the specified one, it would be true.

Then during the generation of the gradient stops for the route layer, if the `CurrentLegAttribute` is true, we have congestion color, if not, we're using the `routeCasingColor`.

### Screenshots or Gifs
The following is how we show specific route leg with  with 
```
navigationMapView.routeCasingColor = UIColor.red
navigationMapView.trafficUnknownColor = UIColor.blue
navigationMapView.show([route], legIndex: 1)
```
showLeg1

The following is that during turn-by turn navigation, the current route leg would be specified with the color overriding. The default specified leg would be the current leg, which is the same as in the 0.4.
```
navigationViewController.navigationMapView.routeCasingColor = UIColor.red
navigationViewController.navigationMapView.trafficUnknownColor = UIColor.blue
```
navigationCurrent


Example/ViewController.swift
```diff
@@ -558,7 +558,7 @@ extension ViewController: NavigationMapViewDelegate {
         return lineLayer
     }
     
-    func navigationMapView(_ navigationMapView: NavigationMapView, shapeFor route: Route) -> LineString? {
+    func navigationMapView(_ navigationMapView: NavigationMapView, shapeFor routes: [Route]) -> LineString? {
```
MaximAlien, line 561: We do not support one shape for all routes since `1.x`. I think this delegate method should be called multiple times per each route, so this change will have to reverted.

Sources/MapboxNavigation/NavigationMapViewDelegate.swift
```diff
@@ -35,10 +35,10 @@ public protocol NavigationMapViewDelegate: class, UnimplementedLogging {
         
      Resulting `LineString` will then be styled using `NavigationMapView.navigationMapView(_:routeLineLayerWithIdentifier:sourceIdentifier:)` provided style or a default congestion style if above delegate method was not implemented.
      - parameter navigationMapView: The NavigationMapView.
-     - parameter route: The route that the sender is asking about.
+     - parameter routes: The routes that the sender is asking about.
      - returns: A `LineString` object that defines the shape of the route, or `nil` in case of default behavior.
      */
-    func navigationMapView(_ navigationMapView: NavigationMapView, shapeFor route: Route) -> LineString?
+    func navigationMapView(_ navigationMapView: NavigationMapView, shapeFor routes: [Route]) -> LineString?
```
MaximAlien, line 41: As I mentioned earlier this change is not needed. It should be possible to modify each route's shape, not all at once.

ShanMa1991: So it should only modify and override one route, right? And it should be default as the mainRoute to be changed?

Example/ViewController.swift
ShanMa1991: I see. It should be similar to `func addRouteLayer`, should be called multiple times, every time form one shape for one route. Thanks for the suggestion, I'll fix this part.

Sources/MapboxNavigation/NavigationMapViewDelegate.swift
MaximAlien: I think so. As I understand `legIndex` was added to control specific leg visibility mostly in active guidance (and there is only one route at a time). I'd run Navigation SDK `0.40` and call `NavigationMapView.show()` method with certain legIndex and check whether all selected legs on those routes are highlighted. 

Maybe @1ec5 can provide more details as well?

ShanMa1991: I see. I created a simple test case as the following for the SDK 0.4:
```
import UIKit
import MapboxDirections
import MapboxCoreNavigation
import MapboxNavigation

    override func viewDidLoad() {
        super.viewDidLoad()
        // Do any additional setup after loading the view.
        let origin = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 38.9131752, longitude: -77.0324047))
        let median1 = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 38.912, longitude: -77.032))
        let median2 = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 38.911, longitude: -77.031))
        let median3 = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 38.9, longitude: -77.03))
        let destination = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 38.89, longitude: -77.029))

        let options = NavigationRouteOptions(waypoints: [origin, median1, median2, destination])

        Directions.shared.calculate(options) { [weak self] (session, result) in
                    switch result {
                    case .failure(let error):
                        print(error.localizedDescription)
                    case .success(let response):
                        guard let route = response.routes?.first, let strongSelf = self else {
                            return
                        }
                        
                        // Since first route is retrieved from response `routeIndex` is set to 0.
                        let navigationService = MapboxNavigationService(route: route, routeOptions: options, simulating: .always)
                        let navigationViewController = NavigationViewController(for: route, routeOptions: options, navigationOptions: navigationOptions)
                        navigationViewController.mapView?.routeCasingColor = #colorLiteral(red: 0.9254902005, green: 0.2352941185, blue: 0.1019607857, alpha: 1)
                        navigationViewController.mapView?.trafficUnknownColor  = #colorLiteral(red: 0.3384324702, green: 0.6636026477, blue: 0.9891712385, alpha: 0.2964898768)
                        navigationViewController.mapView?.show([route], legIndex: 1)
                        

                        strongSelf.present(navigationViewController, animated: true, completion: nil)
                    }
        }
    }
}

```
And the screenshot is as following:
show_test

So the purpose of the ` navigationViewMapView?.show()` in 2.0 should be overriding specific route leg color (that's the one exposed)  of the main route, right? And could act as above screen shoot.

ShanMa1991: And for the `navigationMapView(_ navigationMapView: NavigationMapView, shapeFor route: Route) -> LineString?` it used to be generating a `MGLShape` including the coordinates of the route and corresponding properties, including whether is's main route/ alternative route, whether it's wanted route leg, and the congestion level . But now in 2.0, we separate the `LineString` and `FeatureCollection`, if only returns the main route `LineString`. Should we still keep the delegate of this function? Now the line Layer is mainly controlled by the `FeatureCollection`.

ShanMa1991:
Now the problem is that the layer and source could be added from the `FeatureCollection` with right color and opacity. But when the navigation start, the route lines become invisible. I'll work on this issue.

ShanMa1991:
Fix the opacity expression, now it could allow the overriding route leg color. But it has a problem to display the following vanishing vision during navigation. It seems that there's something wrong with the gradient stops which cause the route line fading away at random place

randomVanishing

Besides, to make sure both of the vanishing route line and route overriding work, the next step is to update the `fractiontravled` method. Because we only need to get the current leg display the vanishing feature, the `fractiontravled` should be calculated on each leg. The other leg's color should only depend on congestion level and opacity.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -721,7 +720,6 @@
 		8A0E0A51257AD9C300C2E924 /* NightStyle.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NightStyle.swift; sourceTree = "<group>"; };
 		8A17635A25CC89D800737520 /* Expression.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Expression.swift; sourceTree = "<group>"; };
 		8A1763CF25CCC38C00737520 /* Double.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Double.swift; sourceTree = "<group>"; };
-		8A3A218F25EEC00200EDA999 /* CoreNavigationNavigator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CoreNavigationNavigator.swift; sourceTree = "<group>"; };
```
MaximAlien, line 724: I think this file should not be removed.

Sources/MapboxNavigation/MapView.swift
```diff
@@ -123,9 +123,12 @@ extension MapView {
         guard let layers = try? __map.getStyleLayers().reversed() else { return nil }
         for layer in layers {
             if !(layer.type == "symbol") && !identifiers.contains(layer.id) {
+                let source = try? __map.getStyleLayerProperty(forLayerId: layer.id, property: "source").value as? String
                 let sourceLayer = try? __map.getStyleLayerProperty(forLayerId: layer.id, property: "source-layer").value as? String
                 
-                if let sourceLayer = sourceLayer,
+                if let source = source,
```
MaximAlien, line 129: This code was also removed in most recent change. Please make sure that rebase/merge was correctly performed.

ShanMa1991: Thanks for the suggestion! I'll check this part and correct it.

MapboxNavigation.xcodeproj/project.pbxproj
ShanMa1991: Thanks for the suggestion! I'll check this part and correct it.

ShanMa1991:
So now the problem is caused by the `func congestionFeatures(legIndex:, roadClassesWithOverriddenCongestionLevels:` is creating multiple feature for one feature collection. And it cause the congestion level and gradient not working as expected. For example, if we do the `let routeFeatures = FeatureCollection(features: [Feature(route.shape!)])` and have a gradient at the median as following:
```
                let gradientStops:[Double: UIColor] =  [
                    0: UIColor.blue,
                    0.5.nextDown: UIColor.blue,
                    0.5: UIColor.red,
                    1: UIColor.red
                ]
```
Even we have 2 legs, but we could still see the color presented for the line gradient  as below:
expected

The next step is to fix the `func congestionFeatures(legIndex:, roadClassesWithOverriddenCongestionLevels:` to make sure the route line source of `FeatureCollection` is generated with one `feature` for each route leg.

Sources/MapboxNavigation/Route.swift
```diff
@@ -44,7 +44,6 @@ extension Route {
     }
     
     func congestionFeatures(legIndex: Int? = nil,
-                            isAlternativeRoute: Bool = false,
```
MaximAlien, line 47: Can you make sure that that `isAlternativeRoute` is really not needed? I guess it was used to filter out alternative route lines in `Navigation SDK 1.x`.

ShanMa1991: Yes, Because now the `isAlternativeRoute` is replaced by `isMainRoute`. Now in the source generation `func shape(forCasingOf)` and `func shape(for:)` is using the `isMainRoute` to generate different feature collection. The layer generation function also use the `isMainRoute` to change the color and opacity. So the `isAlternativeRoute` could be removed from the feature properties.

ShanMa1991:
Due to the `FeatureCollection` generated whole route line layer couldn't support the `line-gradient `, as discussed in https://github.com/mapbox/mapbox-maps-ios-internal/issues/396 and https://github.com/mapbox/mapbox-maps-ios-internal/issues/396. the next step is to generate one `Feature` for each route leg, and a `list` of `Feature` would be used to represent the route line. It would allow the vanishing is happening on the current leg, and the overriding of route line style based on leg index.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -104,6 +104,7 @@ open class NavigationMapView: UIView {
     var fractionTraveled: Double = 0.0
     var preFractionTraveled: Double = 0.0
     var vanishingRouteLineUpdateTimer: Timer? = nil
+    var showSpecificLegIndex: Int?
```
MaximAlien, line 107: What about 
```suggestion
    var currentLegIndex: Int?
```
current naming looks more like a flag.

MaximAlien (approved):
I've tested with `NavigationMapView` and with standalone `NavigationViewController` and it seems to work similarly to what was available in `0.40`. It's great to see that fix turned out to be not really complex one.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for your suggestion! I'll fix this part.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
...
 * Added the `NavigationMapView.navigationMapView(_:didAdd:)` and `NavigationViewController.navigationViewController(_:didAdd:)` delegate methods, which will be called whenever final destination `PointAnnotation` is added to `NavigationMapView` or `NavigationViewController` respectively. ([#2961](https://github.com/mapbox/mapbox-navigation-ios/pull/2961))
+* Add the ability to show specific route leg with color overriding.  When specifying the `legIndex` in `NavigationMapView.show(routes:legIndex:)`, the route line for the specific route leg would show congestion colors, while other route legs would show `NavigationMapView.routeCasingColor` by default. If not specified, all the route legs would show congestion colors. During turn-by-turn navigation, the default specified route leg is the current route leg. The route leg colors could be overriden by `NavigationMapView.routeCasingColor` and the cogestion colors. ([#2833](https://github.com/mapbox/mapbox-navigation-ios/pull/2833))
```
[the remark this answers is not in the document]
ShanMa1991: Thanks for the suggestion, I'll improve this one.

PR #2834, opened 2021-02-25
Electronic horizon

chezzdev:
### Description

This PR targets resolving EHorizon v3 integration.

### Implementation

In the scope of the full integration, we would like to wrap all the generated NavigationNative APIs with Swift classes as well as provide some helper functions like [mpp()](https://github.com/mapbox/mapbox-navigation-android/blob/45a9930965066b59fffddb5840d20e30336cd16b/libnavigation-core/src/main/java/com/mapbox/navigation/core/trip/model/eh/EHorizon.kt#L58).

The initial implementation provides access to the main entry objects (`GraphAccessor`, `RoadObjectsStore`) via generated bindings through `RouteController` and `PassiveLocationDataSource` as well as adds the ability to subscribe to EHorizon update events.


1ec5:
This PR is currently based on the [bamx23/fix-objc-import](https://github.com/mapbox/mapbox-navigation-ios/tree/bamx23/fix-objc-import/) branch, which should be unnecessary once the fixes in #2829 are in place.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -22,16 +22,24 @@ open class PassiveLocationDataSource: NSObject {
...
-        self.navigator = try! Navigator(profile: settingsProfile,
-                                        config: NavigatorConfig(),
-                                        customConfig: "",
-                                        tilesConfig: TilesConfig())
+        let config = try! ConfigFactory.build(for: settingsProfile, config: NavigatorConfig(), customConfig: "")
+        let runLoopExecutor = try! RunLoopExecutorFactory.build()
+        let historyRecorder = try! HistoryRecorderHandle.build(forConfig: config)
+        let cache = try! CacheFactory.build(for: TilesConfig(), config: config, runLoop: runLoopExecutor, historyRecorder: historyRecorder)
+        self.graphAccessor = try! GraphAccessor(cache: cache)
+        self.navigator = try! Navigator(config: config, runLoopExecutor: runLoopExecutor, cache: cache, historyRecorder: historyRecorder)
```
1ec5, line 30: This will look quite different once #2823 lands. We’ll need to account for the need for a GraphAccessor that depends on the cache.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -22,16 +22,24 @@ open class PassiveLocationDataSource: NSObject {
...
 
         self.systemLocationManager = systemLocationManager ?? NavigationLocationManager()
-        
+
         super.init()
-        
+
         self.systemLocationManager.delegate = self
+
+        try! self.navigator.setElectronicHorizonObserverFor(self)
```
1ec5, line 38: Should this happen unconditionally, or is there a use case for applications to benefit from map matching during free driving without the additional overhead of e-horizon updates?

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -85,6 +98,21 @@ open class PassiveLocationDataSource: NSObject {
             }
         }
     }
+
+    /**
+     Sets electronic horizon options. Pass `nil` to reset to defaults.
+     */
+    public func set(electronicHorizonOptions: ElectronicHorizonOptions?) {
```
1ec5, line 105: The public API should use a pure Swift struct instead of ElectronicHorizonOptions. It would be converted to or from an ElectronicHorizonOptions on the fly.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -85,6 +98,21 @@ open class PassiveLocationDataSource: NSObject {
...
+        try! navigator.setElectronicHorizonOptionsFor(electronicHorizonOptions)
+    }
+
+    public var graphAccessor: GraphAccessor
+
+    public var roadObjectsStore: RoadObjectsStore {
```
1ec5, line 111: These classes need Swift wrapper objects, or perhaps their important methods could be exposed by the Navigator class that would be created in #2823 (not to be confused with `MBNNNavigator`).

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonDelegate.swift
```diff
@@ -0,0 +1,26 @@
...
+
+public protocol ElectronicHorizonDelegate: class {
+
+    /**
+     Might be called multiple times when the position changes
+
+     - parameter position: Current electronic horizon position (map matched position + e-horizon tree)
+     - parameter distances: Map road object id -> RoadObjectDistanceInfo for upcoming road objects
+     */
+    func didUpdatePosition(_ position: ElectronicHorizonPosition, distances: [String : RoadObjectDistanceInfo])
```
1ec5, line 12: Is there a simpler struct-like representation that we could convert ElectronicHorizonPosition to on the developer’s behalf?

1ec5, line 12: Create a type alias for strings and use it for the keys of this dictionary. RoadObjectDistanceInfo will need to be converted to a Swift struct; fortunately that seems to be more straightforward than converting ElectronicHorizonPosition to something more usable by the application.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonDelegate.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     Called when entry to line-like (i.e. which has length != null) road object was detected
+     - parameter info: Contains info related to the object
+     */
+
+    func roadObjectDidEnter(_ objectEnterExitInfo: RoadObjectEnterExitInfo)
```
1ec5, line 19: It would be more logical to say that the user entered or exited the road object rather than the other way around.

Tests/MapboxCoreNavigationTests/EHorizonTests.swift
```diff
@@ -0,0 +1,48 @@
+import MapboxNavigationNative
+@testable import MapboxCoreNavigation
+import XCTest
+
+// Happy path repeating EH General Usage Example
+// https://github.com/mapbox/mapbox-navigation-native/blob/0e5ec2b339a129aa67138a4f2cb25a47d909a611/docs/eh_integration.md#general-usage-example
```
1ec5, line 6: Remove this reference to internal documentation.

Tests/MapboxCoreNavigationTests/EHorizonTests.swift
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    func roadObjectDidEnter(_ objectEnterExitInfo: RoadObjectEnterExitInfo) {}
+
+    func roadObjectDidExit(_ objectEnterExitInfo: RoadObjectEnterExitInfo) {}
+}
+
+extension EHorizonTests: RoadObjectsStoreObserver {
+    func onRoadObjectAdded(forId id: String) {
+        _ = try! passiveLocationDataSource.roadObjectsStore.getRoadObjectMetadata(forRoadObjectId: id)
```
1ec5, line 42: Can you add comments explaining what this test is trying to exercise? At a glance, it’s a bit confusing to see a test in this repository exercise only MapboxNavigationNative APIs.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
```diff
@@ -0,0 +1,106 @@
...
+        return try! native.getRoadObjectIdsByEdgeIds(forEdgeIds: edgeIds.map(NSNumber.init))
+    }
+
+    /**
+     Adds road object to be tracked in electronic horizon. In case if object with such id already exists updates it.
+     - parameter roadObjectId unique id of object
+     - parameter location road object location (can be obtained using OpenLRDecoder)
+     */
+    public func addCustomRoadObject(for roadObjectId: RoadObjectId, location: OpenLRLocation) {
+//        native.addCustomRoadObject(forId: roadObjectId, location: <#T##OpenLRLocation#>)
```
chezzdev, line 70: Android has an [implementation](https://github.com/mapbox/mapbox-navigation-android/blob/f589c2fc2cfdee981feb6e1c960891c1aa1d8f58/libnavigation-core/src/main/java/com/mapbox/navigation/core/trip/session/RoadObjectsStore.kt#L83-L96) for this which doesn't involve wrapping all the `OpenLRLocation` structure on the user side. It uses `OpenLRDecoder` assuming the user will know encoded `Location`. Needs to be further investigated.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
chezzdev: Well, at least it makes sense to subscribe to updates when the user sets the delegate. Adjusted the code

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
chezzdev: Agree, adjusted for that.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
chezzdev: Right, wrappers are introduced in the latest commits.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonDelegate.swift
chezzdev: I was aiming at full NavNative functionality support. Not sure if we want to hide some of it or make it accessible less directly. I guess it could complicate the support of such functionality.

chezzdev: Yes, done.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonDelegate.swift
chezzdev: Please take a look at the new naming. Is that better wording?

Tests/MapboxCoreNavigationTests/EHorizonTests.swift
chezzdev: I added it as a reference to the generic EH use-case. The goal was to substitute all the APIs with Swift wrappers and be able to remove NavNative import. That can be done now. 
Eventually, this test isn't testing any logic, so it could be removed if no other application is found.

Sources/MapboxCoreNavigation/EHorizon/EHorizonObjectEnterExitInfo.swift
```diff
@@ -0,0 +1,23 @@
...
+public struct EHorizonObjectEnterExitInfo {
+
+    /** Road object id */
+    public let roadObjectId: String
+
+    /**
+     If object was entered via it's start for `onRoadObjectEnter`
+     or if object was exited via it's end for `onRoadObjectExit`
+     */
+    public let isEnterFromStartOrExitFromEnd: Bool
```
1ec5, line 13: Unlike the other two properties, this property doesn’t appear to be intrinsic to the road object itself, just the context in which the object happens to be used. So I think it would be appropriate to omit this property from the struct and include it alongside the struct as a separate argument to the delegate method.

Sources/MapboxCoreNavigation/EHorizon/EHorizon.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+import MapboxNavigationNative
+
+public class EHorizon {
```
1ec5, line 4: We should probably spell out “electronic” to make these types easier to find in code completion and less ambiguous when reading the code.

Sources/MapboxCoreNavigation/EHorizon/EHorizonEdge.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+import MapboxNavigationNative
+
+public struct EHorizonEdge {
+
+    /** Unique identifier of the directed edge */
+    public let id: UInt
```
1ec5, line 7: In Swift, ID is always spelled out as “identifier”. We should probably also make a type alias for ID-like properties for forward compatibility and a convenient place to document any limitations on the numeric values in the future.

Sources/MapboxCoreNavigation/EHorizon/EHorizonEdge.swift
```diff
@@ -0,0 +1,32 @@
...
+    /** The probability for this edge in percentage */
+    public let probability: Double
+
+    /**
+     The outgoing Edges.
+     NB: MPP can be splitted at some point if some of edges have low probability difference(+/- 0.05),
+     i.e. `out` can contain more than 1 edges with the level 0.
+     Currently we have a limitation for maximum 1 split per electronic horizon.
+     */
+    public let out: [EHorizonEdge]
```
1ec5, line 24: For consistency with `MapboxDirections.Intersection`, let’s call this property `outletEdges`.

Sources/MapboxCoreNavigation/EHorizon/EHorizonEdgeMetadata.swift
```diff
@@ -0,0 +1,72 @@
...
+    public let length: Double
+
+    /** The edge's functional road class */
+    public let functionalRoadClass: RoadClass
+
+    /** The Edge's max speed (m/s) */
+    public let speedLimit: Double?
+
+    /** The Edge's average speed (m/s) */
+    public let speed: Double
```
1ec5, line 19: This speed should be expressed as a `CLLocationSpeed`.

Sources/MapboxCoreNavigation/EHorizon/EHorizonEdgeMetadata.swift
```diff
@@ -0,0 +1,72 @@
...
+    public let heading: Double
+
+    /** The Edge's length in meters. */
+    public let length: Double
+
+    /** The edge's functional road class */
+    public let functionalRoadClass: RoadClass
+
+    /** The Edge's max speed (m/s) */
+    public let speedLimit: Double?
```
1ec5, line 16: Speed limits are typed as `Measurement<UnitSpeed>`.

Sources/MapboxCoreNavigation/EHorizon/RoadClass.swift
```diff
@@ -0,0 +1,84 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ Functional road class
+ See for details: https://wiki.openstreetmap.org/wiki/Key:highway
+ */
+public enum RoadClass {
```
1ec5, line 8: We should reuse `MapboxDirections.MapboxStreetsRoadClass` instead of defining a new enumeration.

Sources/MapboxCoreNavigation/EHorizon/RoadName.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+import MapboxNavigationNative
+
+public struct RoadName {
+
+    public let name: String
+
+    public let isShielded: Bool
```
1ec5, line 8: I suppose this property means the road name is actually a road number that can be displayed in a generic shield? For consistency with MapboxDirections, we should expose either a `name` that’s actually a name or a `code` (aka “ref”) in the event that `MapboxNavigationNative.RoadName.isShielded` is true.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -102,15 +108,15 @@ open class PassiveLocationDataSource: NSObject {
...
+        try! navigator.setElectronicHorizonOptionsFor(electronicHorizonOptions?.nativeOptions)
     }
 
     public var graphAccessor: GraphAccessor
 
-    public var roadObjectsStore: RoadObjectsStore {
-        return try! navigator.roadObjectStore()
-    }
+    public lazy var roadObjectsStore: RoadObjectsStore = {
+        return RoadObjectsStore(try! navigator.roadObjectStore())
```
1ec5, line 118: Is it safe to stash the store here, or should there be a weak backpointer to `MapboxCoreNavigation.Navigator`?

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
1ec5: I’ve rebased this branch atop release-v2.0, which now includes the changes in #2823.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -49,13 +53,43 @@ open class PassiveLocationDataSource: NSObject {
      The location data source’s delegate.
      */
     public weak var delegate: PassiveLocationDataSourceDelegate?
+
+    /**
+     Delegate for Electronic Horizon updates.
+     */
+    public weak var electronicHorizonDelegate: EHorizonDelegate? {
+        didSet {
+            if delegate != nil {
```
Udumft, line 62: Did you mean to check `electronicHorizonDelegate
 != nil` here? Otherwise I don't understand the relation with `delegate`.

Also, it is technically possible to have `PassiveLocationDataSource` and `RouteController` active at the same time and thus they'll compete for being the EH observer. Since both implementations just forward the calls from `NN.Navigator` which is `shared`, maybe it is logical to also move this delegate setting to `Navigator`?

Sources/MapboxCoreNavigation/EHorizon/EHorizonDistanceInfo.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ DistanceInfo is an internal data of OpenLRLineLocation.
+ It's not intended to be read or set from the outside.
+ */
+public struct DistanceInfo {
+
+    public let distance: Double
```
Udumft, line 10: Should we use `CLLocationDistance` for such values here and other EH types?

Sources/MapboxCoreNavigation/EHorizon/EHorizonEdgeMetadata.swift
```diff
@@ -0,0 +1,72 @@
+import Foundation
+import MapboxNavigationNative
+
+public struct EHorizonEdgeMetadata {
+
+    /** The bearing in degrees clockwise at the start of the edge. */
+    public let heading: Double
```
Udumft, line 7: `MapboxDirections` uses `CLLocationDirection` for `headings`

Sources/MapboxCoreNavigation/EHorizon/EHorizonObjectEdgeLocation.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+import MapboxNavigationNative
+
+public struct EHorizonObjectEdgeLocation {
+    /**
+     Offset from the start of edge (0 - 1) pointing to the beginning of road object on this edge
+     will be 0 for all edges in the line-like road object expect the very first one
+     in the case of point-like object percentAlongBegin == percentAlongEnd
```
Udumft, line 8: I understand that this is just the copy-pasted description from `RoadObjectEdgeLocation`, but I've read it several times and still didn't understand what it has to do with line-like road objects. I believe there are few typos in there :)

Sources/MapboxCoreNavigation/EHorizon/EHorizonOptions.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+import MapboxNavigationNative
+
+public struct EHorizonOptions {
+
+    /** The minimum length of the EHorizon ahead of the current position. */
+    public let length: Double
+
+    /** The expansion strategy to be used. */
+    public let expansion: UInt
```
Udumft, line 10: What strategies are available? Could we introduce an enum or is it just a boolean flag?

Sources/MapboxCoreNavigation/EHorizon/EHorizonOptions.swift
```diff
@@ -0,0 +1,38 @@
...
+    public let minTimeDeltaBetweenUpdates: Double?
+
+    public init(length: Double, expansion: UInt, branchLength: Double, minTimeDeltaBetweenUpdates: Double?) {
+        self.length = length
+        self.expansion = expansion
+        self.branchLength = branchLength
+        self.minTimeDeltaBetweenUpdates = minTimeDeltaBetweenUpdates
+    }
+
+    var nativeOptions: ElectronicHorizonOptions {
```
Udumft, line 29: Perhaps it's convenient to have `ElectronicHorizonOptions` extension with proper `init(_:EHorizonOptions)`

Sources/MapboxCoreNavigation/EHorizon/RoadName.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+import MapboxNavigationNative
+
+public struct RoadName {
```
Udumft, line 4: Could we use this struct outside EH context? There are some manipulations with road labeling so maybe we could refactor that later and benefit from this struct in multiple places?

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     Removes road object(i.e. stops tracking it in electronic horizon)
+     - parameter roadObjectId of road object
+     */
+    public func removeCustomRoadObject(for roadObjectId: RoadObjectId) {
+        try! native.removeCustomRoadObject(forId: roadObjectId)
+    }
+
+    public var peer: MBXPeerWrapper?
```
Udumft, line 81: Is it intended to be `public`?

Sources/MapboxCoreNavigation/EHorizon/RoadName.swift
1ec5: Related: https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r585849144. It might even make sense for MapboxDirections to consolidate various RouteStep properties into a public [Road](https://github.com/mapbox/mapbox-directions-swift/blob/cec732768fd85429201e16fe9ece0dc15bc5e443/Sources/MapboxDirections/RouteStep.swift#L307) type that we could reuse here and elsewhere in MapboxCoreNavigation.

Sources/MapboxCoreNavigation/EHorizon/EHorizonDistanceInfo.swift
1ec5: `CLLocationDistance` is measured in meters. What unit is this distance measured in? Is it the same unit as `GraphPath.length`?

/cc @SiarheiFedartsou

Sources/MapboxCoreNavigation/EHorizon/EHorizonEdgeMetadata.swift
1ec5: `CLLocationDirection` is measured starting from 0° at due north. `Measurement<UnitAngle>` is used for relative angles. What is the reference point for this heading?

/cc @SiarheiFedartsou

1ec5:
The new e-horizon integration test is [crashing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/2705/workflows/d9c7db9c-0d54-41b4-90c0-b96b4e7f30e6/jobs/27808) when using SPM, likely because SPM doesn’t build a test bundle that can have an Info.plist to specify an access token in.

```
Test Case '-[MapboxCoreNavigationTests.EHorizonTests testFreeDriveEHorizon]' started.
Precondition failed: A Mapbox access token is required. Go to <https://account.mapbox.com/access-tokens/>. In Info.plist, set the MBXAccessToken key to your access token, or use the Directions(accessToken:host:) initializer.: file MapboxDirections/DirectionsCredentials.swift, line 43
2021-03-06 01:12:37.999939+0000 xctest[3252:68768] Precondition failed: A Mapbox access token is required. Go to <https://account.mapbox.com/access-tokens/>. In Info.plist, set the MBXAccessToken key to your access token, or use the Directions(accessToken:host:) initializer.: file MapboxDirections/DirectionsCredentials.swift, line 43
2021-03-06 01:13:47.872 xcodebuild[960:70393]  IDETestOperationsObserverDebug: Writing diagnostic log for test session to:
/Users/distiller/Library/Developer/Xcode/DerivedData/project-focxxzqvztridmeptmirzegwtlnj/Logs/Test/Run-MapboxNavigation-Package-2021.03.06_01-10-05-+0000.xcresult/Staging/2_Test/Diagnostics/MapboxCoreNavigationTests-651C7BE3-26DC-4228-9040-B369FABCDE4E/MapboxCoreNavigationTests-2346F432-4235-4F2A-A1CB-898FD3EC91D2/Session-MapboxCoreNavigationTests-2021-03-06_011347-yF0pRN.log
2021-03-06 01:13:47.882 xcodebuild[960:30291] [MT] IDETestOperationsObserverDebug: (4CDE6604-E005-42F9-82F1-AB64A54BBE1F) Beginning test session MapboxCoreNavigationTests-4CDE6604-E005-42F9-82F1-AB64A54BBE1F at 2021-03-06 01:13:47.872 with Xcode 12D4e on target <DVTiPhoneSimulator: 0x7fab81db1ce0> {
		SimDevice: iPhone 8 Plus (57411501-EBA1-48D1-BA93-F3BB3E9590E6, iOS 14.4, Booted)
} (14.4 (18D46))
2021-03-06 01:13:47.885 xcodebuild[960:30291] [MT] IDETestOperationsObserverDebug: (4CDE6604-E005-42F9-82F1-AB64A54BBE1F) Finished requesting crash reports. Continuing with testing.
2021-03-06 01:13:51.196300+0000 xctest[3268:70441]  Setting up logger

Restarting after unexpected exit, crash, or test timeout in EHorizonTests/testFreeDriveEHorizon(); summary will include totals from previous launches.
```

```
#0	0x0000000111052a80 in _swift_runtime_on_report ()
#1	0x00000001110cd0e3 in _swift_stdlib_reportFatalErrorInFile ()
#2	0x0000000110d5d6e5 in closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x0000000110d5d327 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#4	0x0000000110d5d8c1 in specialized String.withUTF8<A>(_:) ()
#5	0x0000000110d5bff7 in _assertionFailure(_:_:file:line:flags:) ()
#6	0x000000010f543269 in DirectionsCredentials.init(accessToken:host:) at /Users/mxn/Library/Developer/Xcode/DerivedData/MapboxNavigation-alytgpeiocddhmempehvwkpsbezh/SourcePackages/checkouts/mapbox-directions-swift/Sources/MapboxDirections/DirectionsCredentials.swift:43
#7	0x000000010f523c53 in default argument 0 of Directions.init(credentials:) at /Users/mxn/Library/Developer/Xcode/DerivedData/MapboxNavigation-alytgpeiocddhmempehvwkpsbezh/SourcePackages/checkouts/mapbox-directions-swift/Sources/MapboxDirections/Directions.swift:130
#8	0x000000010f523bcc in globalinit_33_EADDF7C5FE155C304C3482AF65D416BF_func2 at /Users/mxn/Library/Developer/Xcode/DerivedData/MapboxNavigation-alytgpeiocddhmempehvwkpsbezh/SourcePackages/checkouts/mapbox-directions-swift/Sources/MapboxDirections/Directions.swift:115
#9	0x00007fff20106508 in _dispatch_client_callout ()
#10	0x00007fff2010775d in _dispatch_once_callout ()
#11	0x0000000111079e5a in swift_once ()
#12	0x000000010f523cc4 in Directions.shared.unsafeMutableAddressor at /Users/mxn/Library/Developer/Xcode/DerivedData/MapboxNavigation-alytgpeiocddhmempehvwkpsbezh/SourcePackages/checkouts/mapbox-directions-swift/Sources/MapboxDirections/Directions.swift:115
#13	0x000000010f4f555e in default argument 0 of PassiveLocationDataSource.init(directions:systemLocationManager:) ()
#14	0x000000010f4f54dd in EHorizonTests.setUp() at /path/to/mapbox-navigation-ios/Tests/MapboxCoreNavigationTests/EHorizonTests.swift:10
#15	0x000000010f4f55cb in @objc EHorizonTests.setUp() ()
```

Sources/MapboxCoreNavigation/EHorizon/EHorizonObjectEdgeLocation.swift
etl: That's just a typo: expect = except.

Sources/MapboxCoreNavigation/EHorizon/EHorizonDistanceInfo.swift
etl: All distances are in meters.

Sources/MapboxCoreNavigation/EHorizon/EHorizonEdgeMetadata.swift
SiarheiFedartsou: It is the same as for other headings in our codebase - 0° at due north, i.e. `CLLocationDirection ` can be used here.

Sources/MapboxCoreNavigation/EHorizon/EHorizonOptions.swift
SiarheiFedartsou: Yeah, it should be enum I beleive: 0, 1, 2 is available.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
SiarheiFedartsou: Yeah, probably worth not exposing OpenLRDecoder itself yet. Especially since it is becoming not only for OpenLRs...

Sources/MapboxCoreNavigation/EHorizon/EHorizonOptions.swift
SiarheiFedartsou: https://github.com/mapbox/mapbox-navigation-native/pull/2496/files#diff-18df9a7b99e917622e94af64345e38aa0c777cef30aabccaac494d24f79a6babR132

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
1ec5: I’ve removed the custom road object feature for now, until there’s more clarity about how we’d expose it as tail work.

Sources/MapboxCoreNavigation/EHorizon/EHorizonOptions.swift
1ec5: I’m leaving this property as a number because it isn’t technically a closed set of values.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
1ec5: RoadObjectsStoreObserver unfortunately requires this property to be public.

1ec5:
More to do before this PR is in a minimally landable state:

* [ ] Document new public classes and structs
* [ ] Mention new entry point methods and standalone classes in the changelog
* [ ] Disable the e-horizon unit test due to #2851

Some of the review feedback above would also be important to address as tail work immediately after landing:

* [ ] Unify `electronicHorizonDelegate` within `Navigator`: https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r588141211
* [ ] Separate `isEnterFromStartOrExitFromEnd` from its type: https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r585843756
* [ ] Weaken ownership of the the internal RoadObjectsStore: https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r585850336
* [ ] Nest various road-related types under a ~~RoadObject~~ Road struct (which would be empty and unused for now): https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r588160758

1ec5:
If you need to install this branch before it lands:

* For CocoaPods, use the podspecs and Podfile snippet in [this gist](https://gist.github.com/1ec5/4356da6b9d596732d5a5174d74257821).
* For SPM, use the following dependency:
   ```swift
   .package(name: "MapboxNavigation", url: "https://github.com/mapbox/mapbox-navigation-ios.git", .branch("pad-ehorizon"))
   ```

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizon.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ An Electronic Horizon is a probable path (or paths) of a vehicle within the road graph which is
```
1ec5, line 5: “Electronic horizon” is a generic noun so it should be uncapitalized.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonDelegate.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ Electronic horizon delegate. Methods are called in the order specified. `didUpdatePosition` might be called
```
1ec5, line 5: This needs to be spelled out in full in order for jazzy to automatically link it:

```suggestion
 Electronic horizon delegate. Methods are called in the order specified. `didUpdatePosition(_:distances:)` might be called
```

or if the method gets renamed:

```suggestion
 Electronic horizon delegate. Methods are called in the order specified. `didUpdate(_:distances:)` might be called
```

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonOptions.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+import CoreLocation
+import MapboxNavigationNative
+
+/**
+ Defines options for `ElectronicHorizon`.
```
1ec5, line 6:
```suggestion
 Options for configuring how a `RouteController` or `PassiveLocationDataSource` discovers and informs `ElectronicHorizonDelegate` about an electronic horizon while the user navigates.
```

Sources/MapboxCoreNavigation/EHorizon/RoadObjectEdgeLocation.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ Represents location of road object on road graph.
+ For point-like objects will contain single edge with `percentAlongBegin == percentAlongEnd`
```
1ec5, line 6:
```suggestion
 A point object is represented by a single edge whose location has the same `percentAlongBegin` and `percentAlongEnd`.
```

Sources/MapboxCoreNavigation/EHorizon/RoadObjectTransition.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import MapboxNavigationNative
+
+/** Represents information about road object transitions */
```
1ec5, line 4:
```suggestion
/** Details about entering or exiting a road object. */
```

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
1ec5: I ended up removing this virtual delegate in favor of a notification system.

Sources/MapboxCoreNavigation/EHorizon/EHorizonObjectEnterExitInfo.swift
1ec5: I ended up removing this struct as part of transitioning to a notification system – now this information sits alongside the other information in a user info dictionary.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizon.swift
```diff
@@ -0,0 +1,30 @@
...
+ exposed via the same interface as described below.
+
+ We represent the road network ahead of us as a tree of edges. Each intersection has outbound
+ edges and each edge has probability of transition to another edge as well as metadata which
+ can be used to implement sophisticated features on top of it.
+ */
+public class ElectronicHorizon {
+
+    /// The start edge from which EHorizon tree structure can be navigated
+    public let start: Edge
```
1ec5, line 25: Do we really need a separate class to hold the root edge in the subgraph? It’s unclear to me what else this class would ever hold, so maybe we should just inline the root edge in the `electronicHorizonDidUpdatePosition` user info dictionary.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdge.swift
```diff
@@ -0,0 +1,46 @@
...
+         */
+        public let level: UInt
+
+        /** The probability for this edge in percentage */
+        public let probability: Double
+
+        /**
+         The outgoing Edges.
+         NB: MPP can be splitted at some point if some of edges have low probability difference(+/- 0.05),
+         i.e. `out` can contain more than 1 edges with the level 0.
```
bamx23, line 34:
```suggestion
         i.e. `outletEdges` can contain more than 1 edges with the level 0.
```

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
```diff
@@ -0,0 +1,83 @@
...
+        public let names: [RoadName]
+
+        /** The edge's lane counts */
+        public let laneCount: UInt?
+
+        /** The edge's mean elevation */
+        public let meanElevation: CLLocationDistance?
+
+        /** The edge's curvature */
+        public let curvature: UInt
```
1ec5, line 51: What unit is the curvature measured in?

Sources/MapboxCoreNavigation/EHorizon/RoadGraphPosition.swift
```diff
@@ -0,0 +1,20 @@
...
+
+        /** The edge identifier that forms the point object. */
+        public let edgeIdentifier: ElectronicHorizon.Edge.Identifier
+
+        /** The distance from the start of the point object to the user’s location as a fraction of the point object’s length from 0 to 1. */
+        public let fractionFromStart: Double
+
+        init(_ native: GraphPosition) {
+            self.edgeIdentifier = UInt(native.edgeId)
+            self.fractionFromStart = native.percentAlong
```
1ec5, line 17: What does it mean for a point object to have a percentage along the object?

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
```diff
@@ -0,0 +1,97 @@
...
+ Use a `RoadObjectsStore` object to get more information about a road object with a given identifier or get the locations of road objects along `ElectronicHorizon.Edge`s.
+ */
+public typealias RoadObjectIdentifier = String
+
+/**
+ Stores and provides access to metadata about road objects.
+ 
+ You do not create a `RoadObjectsStore` object manually. Instead, use the `RouteController.roadObjectsStore` or `PassiveLocationDataSource.roadObjectsStore` to access the currently active road objects store.
+ */
+public final class RoadObjectsStore {
```
1ec5, line 16: RoadObjectStorage would be a more conventional name in Swift.

Sources/MapboxCoreNavigation/EHorizon/RoadGraphPath.swift
```diff
@@ -0,0 +1,28 @@
...
+    /** A position along a linear object in the road graph. */
+    public struct Path {
+        /** The edge identifiers that form the linear object. */
+        public let edgeIdentifiers: [ElectronicHorizon.Edge.Identifier]
+
+        /** The distance from the start of the linear object to the user’s location as a fraction of the linear object’s length from 0 to 1. */
+        public let fractionFromStart: Double
+
+        /** The distance from the the user’s location to the end of the linear object as a fraction of the linear object’s length from 0 to 1. */
+        public let fractionToEnd: Double
```
Udumft, line 16: Isn't it equal `1 - fractionFromStart`? Do we need both values?

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -27,6 +27,10 @@ open class PassiveLocationDataSource: NSObject {
         
         self.systemLocationManager.delegate = self
     }
+
+    deinit {
+        try! self.navigator.setElectronicHorizonObserverFor(nil)
```
Udumft, line 32: Is it correct to reset EH observer here (and in `RouteController.deinit`)? The only place where it is set up - is `CoreNavigationNavigator` private initializer which is called just once, when instantiating a shared `Navigator`. Thus, running a second active guidance session will result in having no EHObserver set.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
```diff
@@ -0,0 +1,87 @@
...
+
+        init(_ native: EdgeMetadata) {
+            heading = native.heading
+            length = native.length
+            mapboxStreetsRoadClass = MapboxStreetsRoadClass(native.frc, isRamp: native.isRamp)
+            if let speedLimitValue = native.speedLimit as? Double {
+                // TODO: Convert to miles per hour as locally appropriate.
+                speedLimit = Measurement(value: speedLimitValue, unit: UnitSpeed.metersPerSecond).converted(to: .kilometersPerHour)
+            } else {
+                speedLimit = nil
```
Udumft, line 62: NIT: this is redundant. Optional `speedLimit` defaults to `nil`

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -74,27 +83,66 @@ class Navigator {
...
+        try! navigator.setElectronicHorizonObserverFor(nil)
+    }
+    
+    var electronicHorizonOptions: ElectronicHorizonOptions? {
+        didSet {
+            let nativeOptions: MapboxNavigationNative.ElectronicHorizonOptions?
+            if let electronicHorizonOptions = electronicHorizonOptions {
+                nativeOptions = MapboxNavigationNative.ElectronicHorizonOptions(electronicHorizonOptions)
+            } else {
+                nativeOptions = nil
```
Udumft, line 113: NIT: this is redundant. `nativeOptions` will be defaulted to `nil`

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
avi-c: This is a good point. Seems like we should install the observer for the free drive case here in the PassiveLocationDataSource in `startUpdatingLocation()` and in the `RouteController()` so a delegate is set per drive session, active or passive.

avi-c (commented):
On the whole this looks good. I did have the comment on setting the EH delegate that should be reviewed. Otherwise, I think the main EH issue is making sure the docs are pretty through since there are a lot of concepts introduced that would be unfamiliar to most developers who are first trying out the SDK.

Sources/MapboxCoreNavigation/EHorizon/RoadGraphPath.swift
1ec5: Maybe not, and also it would be more intuitive to provide a distance from the start. @SiarheiFedartsou, can we assume that the `length` is always consistent with the percentages? In other words, is there any reason the length might not cover the whole path but the percentages would?

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
1ec5: `nativeOptions` is declared but not initialized as a constant right above here, so the explicit `nil` is required.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
1ec5: Removing this assignment would result in an error because `speedLimit` isn’t explicitly defaulted to `nil`:

```
/path/to/mapbox-navigation-ios/Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift:83:9: Return from initializer without initializing all stored properties
/path/to/mapbox-navigation-ios/Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift:26:20: 'self.speedLimit' not initialized
```

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
SiarheiFedartsou: It is "curvature factor" https://github.com/mapbox/valhalla-internal/blob/c4799e97e0721742101b244c8bf771653b57edb8/valhalla/baldr/directededge.h#L227 . I  guess there are no any units.

Udumft (approved):
Aside of other open questions, I have my complains resolved.
The only possible concern is about `fractionFromStart`/`fractionFromEnd` question, but that is debatable.

1ec5:
@SiarheiFedartsou it’d be great if you or someone else familiar with the navigator’s inner workings could comment on these questions at some point:

* Percentage “along” a point? https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r592883876
* Do the percentages from the start to the location and from the location to the end always add up to the object’s total length? https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r593219838

I’m going to merge this PR in the meantime. There are a number of opportunities to continue refining the API anyways:

* Reduce redundancy between `RoadObjectType`, `RoadObjectMetadata`, and `RouteAlert.Alert` – apparently this will be driven by an upcoming change in MapboxNavigationNative
* Consider eliminating ElectronicHorizon in favor of putting the start edge directly in the notification’s user info dictionary, then rename nested types like `ElectronicHorizon.Edge`: https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r592032215
* Consider consolidating RoadName and other road object–related types under a unified type and returning all the information about it at once: https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r588160758
* Continue to refine and copyedit documentation comments like https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r588151076

Sources/MapboxCoreNavigation/EHorizon/RoadGraphPosition.swift
SiarheiFedartsou: Edges are "long" objects which have length > 0, but point objects are just points on graph. So to define position of some point-like object on graph we use combination of edge id + "percent along". "Percent along" is a float number in 0 - 1 range which defines offset from the beginning of the edge to the exact position of point. Also I'd say that "fraction" is probably better term here than "percent" since it is in 0 - 1 range, but not 0 - 100. We use "percent" to be consistent with Valhalla which historically uses this term.

Sources/MapboxCoreNavigation/EHorizon/RoadGraphPath.swift
SiarheiFedartsou: Not exactly. I made an image with some explanation what actually "graph path" mean:
Screen Shot 2021-03-16 at 10 08 57

SiarheiFedartsou: Also I'd change docstrings for `RoadGraphPath`. It is actually abstract type and it is not necessary [to be connected with current user position](https://github.com/mapbox/mapbox-navigation-ios/blob/78b5de684c512d23065a161251b3f187b1a7597a/Sources/MapboxCoreNavigation/EHorizon/RoadGraphPath.swift#L15).
See our docs https://github.com/mapbox/mapbox-navigation-native/blob/ad622b030121846f55deb1f05edbdd58166ab8e6/bindings/common-types.mbg#L109

1ec5: Fixed in 327b397fa9831c321f37c11fa17b580f1a2cbc90.

Issue #2814, opened 2021-02-19
Route line is drawn over street names layer.


MaximAlien:
Currently route line is drawn over street names layer. For example:

![Screen Shot 2021-02-04 at 3.59.36 PM.png](https://images.zenhubusercontent.com/562e4d2e638e718c995e0957/c9a6d094-70f9-4a73-89d8-e4d3334e40bd)

It is expected that route lines are not drawn over street names layer as it's shown in `v1.x` based Navigation SDK:
![Screen Shot 2021-02-04 at 3.59.39 PM.png](https://images.zenhubusercontent.com/562e4d2e638e718c995e0957/b945c739-fcbd-4464-b965-d8e7286dc748)

PR #2838, opened 2021-02-26
Fix route layer shown over map labels

jill-cardamon:
### Description
This PR is to fix an issue where the route line was drawn over the street layer #2814.

### Implementation
Removed check for empty source during parent layer initialization in `addMainRouteLayer(_:)` that was preventing the `for` loop from going through several of the map layers.

### Screenshots or Gifs
simulator_screenshot_750B6D3B-F728-4D9A-B582-B4C9F25379E8

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -500,12 +500,9 @@ open class NavigationMapView: UIView {
             guard let layers = try? mapView.__map.getStyleLayers().reversed() else { return nil }
             for layer in layers {
                 if !(layer.type == "symbol") && !identifiers.contains(layer.id) {
-                    let source = try? mapView.__map.getStyleLayerProperty(forLayerId: layer.id, property: "source").value as? String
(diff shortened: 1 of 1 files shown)


MaximAlien:
After testing looks good. Improved check seems to be picking up correct layer identifier (similar to Maps SDK `6.x`).

MaximAlien (approved):

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -500,12 +500,9 @@ open class NavigationMapView: UIView {
             guard let layers = try? mapView.__map.getStyleLayers().reversed() else { return nil }
             for layer in layers {
                 if !(layer.type == "symbol") && !identifiers.contains(layer.id) {
-                    let source = try? mapView.__map.getStyleLayerProperty(forLayerId: layer.id, property: "source").value as? String
```
1ec5, line 503: This might have been influenced by how, in v1._x_, we check whether the layer is backed by the Mapbox Streets source, to avoid getting thrown off by developer-defined vector tile sources that happen to have similar source layer names. But as long as the source layer names are reasonably descriptive, it shouldn’t be a major problem to omit the source check.

PR #2848, opened 2021-03-04
Onboard Routing

Udumft:
added Directions extension to utilize Navigatio……nNative.Router for offline routing. Extended CoreNavigationNavigator to also share cacheHandle

### Description
This PR adds usage of Navigation Native `Router` which is used for providing offline (onboard) directions utilising existing cached tile data.

### Implementation
Added an extension for `Directions` with 2 convenience methods: one to attempt usual online routing with automatic fallback to onboard engine, and another one for purely offline calculation.
Currently `Router` does not support Map Matching requests, so implementation covers only `directions`. In the future it may be easily extended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@
 * Removed deprecated `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Removed `NavigationViewController.origin` property, which was not used. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * A new predictive cache proactively fetches tiles which may become necessary if the device loses its Internet connection at some point during passive or active turn-by-turn navigation. Pass a `PredictiveCacheOptions` instance into the `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:predictiveCacheOptions:)` initializer as you configure a `NavigationViewController`, or manually call `NavigationMapView.enablePredictiveCaching(options:)`. ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
-* Added posibility for offline routing and re-routing using `Directions.calculateOffline(_:completionHandler:)` and `Directions.calculateWithCache(_:completionHandler:)`. `Router` default implementation is now configured to use the latter. ([#2848](https://github.com/mapbox/mapbox-navigation-ios/pull/2848)) 
+* Added the `Directions.calculateOffline(options:completionHandler:)` and `Directions.calculateWithCache(options:completionHandler:)` methods, which incorporate routing tiles from the predictive cache when possible to avoid relying on a network connection to calculate the route. `RouteController` now also uses the predictive cache when rerouting. ([#2848](https://github.com/mapbox/mapbox-navigation-ios/pull/2848)) 
 
 ## v1.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 
 * Removed deprecated `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Removed `NavigationViewController.origin` property, which was not used. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
+* Added posibility for offline routing and re-routing using `Directions.calculateOffline(_:completionHandler:)` and `Directions.calculateWithCache(_:completionHandler:)`. `Router` default implementation is now configured to use the latter. ([#2848](https://github.com/mapbox/mapbox-navigation-ios/pull/2848)) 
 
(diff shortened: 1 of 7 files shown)


Sources/MapboxCoreNavigation/Directions.swift
```diff
@@ -0,0 +1,76 @@
...
+     
+     This method retrieves the routes asynchronously from the [Mapbox Directions API](https://www.mapbox.com/api-documentation/navigation/#directions) over a network connection. If a  server error occurs, details about the error are passed into the given completion handler in lieu of the routes. If network error is encountered, onboard routing engine will attempt to provide directions using existing cached tiles.
+     
+     Routes may be displayed atop a [Mapbox map](https://www.mapbox.com/maps/).
+     
+     - parameter options: A `RouteOptions` object specifying the requirements for the resulting routes.
+     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread.
+     - returns: The data task used to perform the HTTP request. If, while waiting for the completion handler to execute, you no longer want the resulting routes, cancel this task.
+     */
+    @discardableResult open func calculateWithCache(_ options: RouteOptions, completionHandler: @escaping RouteCompletionHandler) -> URLSessionDataTask? {
```
1ec5, line 20: The “withCache” means the method no longer reads like a sentence, so the `options` argument should have a label:

```swift
@discardableResult open func calculateWithCache(options: RouteOptions, completionHandler: @escaping RouteCompletionHandler) -> URLSessionDataTask?
```

Alternatively, we could do something similar to the legacy offline routing functionality in v1._x_, in which the method can either use or not use the cache based on an optional argument:

https://github.com/mapbox/mapbox-navigation-ios/blob/65eb7a1eab9f95234cec713bcfaaad1657e3ab10/Sources/MapboxCoreNavigation/OfflineDirections.swift#L191

In that case, the method signature would look like:

```swift
@discardableResult open func calculate(_ options: RouteOptions, usingCache usesCache: Bool = true, completionHandler: @escaping RouteCompletionHandler) -> URLSessionDataTask?
```

Sources/MapboxCoreNavigation/Directions.swift
```diff
@@ -0,0 +1,76 @@
...
+     Begins asynchronously calculating routes using the given options and delivers the results to a closure.
+     
+     This method retrieves the routes asynchronously from onboard routing engine using existing cached tiles.
+     
+     Routes may be displayed atop a [Mapbox map](https://www.mapbox.com/maps/).
+     
+     - parameter options: A `RouteOptions` object specifying the requirements for the resulting routes.
+     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread.
+     */
+    open func calculateOffline(_ options: RouteOptions, completionHandler: @escaping (Result<RouteResponse, DirectionsError>) -> ()) {
```
1ec5, line 49: Similarly, either give the first argument a label or call the method “calculate” but add an optional argument to control whether offline routing is enabled.

1ec5, line 49: Reuse `RouteCompletionHandler` as the completion handler type.

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@
 * Removed deprecated `InstructionsBannerViewDelegate.didDragInstructionsBanner(_:)` method. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Removed `NavigationViewController.origin` property, which was not used. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * A new predictive cache proactively fetches tiles which may become necessary if the device loses its Internet connection at some point during passive or active turn-by-turn navigation. Pass a `PredictiveCacheOptions` instance into the `NavigationOptions(styles:navigationService:voiceController:topBanner:bottomBanner:predictiveCacheOptions:)` initializer as you configure a `NavigationViewController`, or manually call `NavigationMapView.enablePredictiveCaching(options:)`. ([#2830](https://github.com/mapbox/mapbox-navigation-ios/pull/2830))
+* Added posibility for offline routing and re-routing using `Directions.calculateOffline(_:completionHandler:)` and `Directions.calculateWithCache(_:completionHandler:)`. `Router` default implementation is now configured to use the latter. ([#2848](https://github.com/mapbox/mapbox-navigation-ios/pull/2848))
```
1ec5, line 51:
```suggestion
* Added the `Directions.calculateOffline(_:completionHandler:)` and `Directions.calculateWithCache(_:completionHandler:)` methods, which incorporate routing tiles from the predictive cache when possible to avoid relying on a network connection to calculate the route. `RouteController` now uses the predictive cache when rerouting. ([#2848](https://github.com/mapbox/mapbox-navigation-ios/pull/2848)) 
```

Sources/MapboxCoreNavigation/Directions.swift
Udumft: I think the "label" approach is preferable because there are 3 scenarios: use online routing only, use online and fallback to offline, use offline only. Thus controlling it with boolean flags is uncomfortable unless we introduce an enum.

1ec5 (approved):

PR #2855, opened 2021-03-08
Fix partially visible maneuver problem

ShanMa1991:
### Description
This pr is to fix the partially visible maneuver problems during navigation session when multiple waypoints are selected.

### Implementation
By changing the sequence of laying addition to fix the partially visible problem caused by layer overlapping.

### Screenshots or Gifs
maneuver_fix

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -518,9 +518,9 @@ open class NavigationMapView: UIView {
             var parentLayer: String? = nil
             let identifiers = [
                 IdentifierString.arrow,
-                IdentifierString.arrowStroke,
-                IdentifierString.arrowCasingSymbol,
                 IdentifierString.arrowSymbol,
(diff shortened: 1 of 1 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -518,9 +518,9 @@ open class NavigationMapView: UIView {
             var parentLayer: String? = nil
             let identifiers = [
                 IdentifierString.arrow,
-                IdentifierString.arrowSymbol,
```
MaximAlien, line 521: Original sequence of layer was working correctly in Navigation SDK `1.x`, why this change is required in `2.x`?

ShanMa1991: In Carbon if we're using the 1.x , the adding layer sequence would cause the maneuvers not shown at all. The new sequence of the adding identifiers is following the layer addition and overlapping sequence.

MaximAlien (approved):

PR #2858, opened 2021-03-10
Support Driving Side flag

Udumft:
### Description
Exposed NN's EdgeMetadata.isRightHandTraffic value.

### Implementation
There is a typo generated by obj-c wrapper: `isIsRightHandTraffic` which should be corrected later. A task for corresponding fix for a wrapper generation is created.
I didn't update `CHANGELOG` because this is a part of the greater EH implementation which should cover this and other updates.

### DIFF
--- a/Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
+++ b/Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
@@ -51,8 +51,8 @@ extension ElectronicHorizon.Edge {
         /** The ISO 3166-2 code of the country subdivision where this edge is located. */
         public let regionCode: String?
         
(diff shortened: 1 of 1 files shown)


Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
```diff
@@ -50,6 +50,9 @@ extension ElectronicHorizon.Edge {
 
         /** The ISO 3166-2 code of the country subdivision where this edge is located. */
         public let regionCode: String?
+        
+        /** True if in the current place/state uses right-hand traffic, false if left-hand. */
+        public let isRightHandTraffic: Bool
```
1ec5, line 55: Let’s reuse the DrivingSide enumeration here.

1ec5 (approved):

PR #2861, opened 2021-03-12
Remove TestHelper’s dependency on MapboxNavigation

### LINKED ISSUE #2820: TestHelper should not depend on MapboxNavigation

ShanMa1991:
### Description
This pr is to remove TestHelper’s dependency on MapboxNavigation to fix #2820 

### Implementation
- [x] remove the usage of `NavigationPlotter.swift` from `RouteControllerSnapshotTests` and `SimulationLocationManagerTests`,  replace the `FBSnapshotTestCase` with `XCTestCase` for the two tests
- [x] move the target and file path of `RouteControllerTests` (`RouteControllerSnapshotTests`) and `SimulationLocationManagerTests` from `MapboxNavigationTests` to `MapboxCoreNavigationTests`
- [x] move `NavigationPlotter.swift` from the `TestHelper` target to the `MapboxNavigationTests` target
- [x] remove `MapboxNavigation` from link binary of `TestHelper` to remove its dependency on `MapboxNavigation`

### DIFF
--- a/Tests/MapboxCoreNavigationTests/SimulatedLocationManagerTests.swift
+++ b/Tests/MapboxCoreNavigationTests/SimulatedLocationManagerTests.swift
@@ -1,19 +1,20 @@
 import XCTest
-import FBSnapshotTestCase
-import TestHelper
 import MapboxDirections
 @testable import MapboxCoreNavigation
-@testable import MapboxNavigation
+#if !SWIFT_PACKAGE
+import TestHelper
 
-class SimulatedLocationManagerTests: FBSnapshotTestCase {
+class SimulatedLocationManagerTests: XCTestCase {
     override func setUp() {
         super.setUp()
-        recordMode = false
-        agnosticOptions = [.OS, .device]
+    }
+    
+    override func tearDown() {
+        super.tearDown()
     }
 
     func testSimulateRouteDoublesBack() {
-        let route = Fixture.routesFromMatches(at: "sthlm-double-back", options: NavigationMatchOptions(coordinates: [
+        let coordinates:[CLLocationCoordinate2D] = [
             .init(latitude: 59.337928, longitude: 18.076841),
             .init(latitude: 59.337661, longitude: 18.075897),
             .init(latitude: 59.337129, longitude: 18.075478),
@@ -27,7 +28,8 @@ class SimulatedLocationManagerTests: FBSnapshotTestCase {
             .init(latitude: 59.338156, longitude: 18.075723),
             .init(latitude: 59.338311, longitude: 18.074968),
             .init(latitude: 59.33865, longitude: 18.074935),
-        ]))![0]
+        ]
+        let route = Fixture.routesFromMatches(at: "sthlm-double-back", options: NavigationMatchOptions(coordinates: coordinates))![0]
         let locationManager = SimulatedLocationManager(route: route)
         let locationManagerSpy = SimulatedLocationManagerSpy()
         locationManager.delegate = locationManagerSpy
@@ -39,11 +41,9 @@ class SimulatedLocationManagerTests: FBSnapshotTestCase {
         
         locationManager.delegate = nil
         
-        let view = NavigationPlotter(frame: CGRect(origin: .zero, size: CGSize(width: 1000, height: 1000)))
-        view.routePlotters = [RoutePlotter(route: route)]
-        view.locationPlotters = [LocationPlotter(locations: locationManagerSpy.locations, color: #colorLiteral(red: 0.9254902005, green: 0.2352941185, blue: 0.1019607857, alpha: 0.5043463908), drawIndexesAsText: true)]
+        let testCoordinates:[CLLocationCoordinate2D] = locationManagerSpy.locations.map { $0.coordinate }
         
-        verify(view)
+        XCTAssert(testCoordinates == coordinates)
     }
 }
 
@@ -54,3 +54,4 @@ class SimulatedLocationManagerSpy: NSObject, CLLocationManagerDelegate {
         self.locations.append(contentsOf: locations)
     }
 }
+#endif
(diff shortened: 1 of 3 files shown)


Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
```diff
@@ -1,18 +1,15 @@
 import XCTest
-import FBSnapshotTestCase
 import Turf
 import MapboxDirections
 import TestHelper
 @testable import MapboxCoreNavigation
 @testable import MapboxNavigation
 
-class RouteControllerSnapshotTests: FBSnapshotTestCase {
+class RouteControllerSnapshotTests: XCTestCase {
```
1ec5, line 8: These are no longer snapshot tests, so let’s rename the class and file to RouteControllerTests.

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
```diff
@@ -21,7 +18,7 @@ class RouteControllerSnapshotTests: FBSnapshotTestCase {
     }
     
     func testRouteSnappingOvershooting() {
-        let options = NavigationMatchOptions(coordinates: [
+        let coordinates:[Waypoint] = [
```
1ec5, line 21: This is an array of waypoints, not an array of coordinates, so the variable name is inaccurate. But `NavigationMatchOptions(coordinates:profileIdentifier:)` takes an array of coordinates.

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
```diff
@@ -49,19 +47,19 @@ class RouteControllerSnapshotTests: FBSnapshotTestCase {
...
         }
         
-        let view = NavigationPlotter(frame: CGRect(origin: .zero, size: CGSize(width: 1000, height: 1000)))
-        view.routePlotters = [RoutePlotter(route: route)]
-        view.locationPlotters = [LocationPlotter(locations: locations, color: #colorLiteral(red: 0.9254902005, green: 0.2352941185, blue: 0.1019607857, alpha: 0.5043463908), drawIndexesAsText: true),
-                                 LocationPlotter(locations: snappedLocations, color: #colorLiteral(red: 0.4666666687, green: 0.7647058964, blue: 0.2666666806, alpha: 0.3969795335), drawIndexesAsText: true)]
+        let expectedLocations = Set<CLLocation>()
+        coordinates.forEach{
+            expectedLocations.insert(CLLocation(latitude: $0.coordinate.latitude, longitude: $0.coordinate.longitude))
+        }
```
1ec5, line 60: This would be more succinctly written as:

```suggestion
        expectedLocations = coordinates.map {
            CLLocation(latitude: $0.coordinate.latitude, longitude: $0.coordinate.longitude)
        }
```

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
```diff
@@ -49,19 +47,19 @@ class RouteControllerSnapshotTests: FBSnapshotTestCase {
         let routeController = RouteController(along: route, routeIndex: 0, options: equivalentRouteOptions, dataSource: self)
         locationManager.delegate = routeController
         
-        var snappedLocations = [CLLocation]()
+        var snappedLocations = Set<CLLocation>()
```
1ec5, line 50: I suspect we’d still want to verify that the locations are recorded in the correct order, so an array would be more appropriate than a set.

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
```diff
@@ -1,18 +1,15 @@
 import XCTest
-import FBSnapshotTestCase
 import Turf
 import MapboxDirections
 import TestHelper
 @testable import MapboxCoreNavigation
 @testable import MapboxNavigation
```
1ec5, line 6: As far as I can tell, this file no longer uses MapboxNavigation, so we can remove this import statement. The test suite exercises RouteController, a class vended by MapboxCoreNavigation, so we should move this file to the MapboxCoreNavigationTests target.

1ec5:
The build failure in #2862 predates this PR, but it would be useful to address it here so we can demonstrate that the refactor works.

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
ShanMa1991: Thanks for the suggestion, I'll rename the file.

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
ShanMa1991: Thanks for the suggestion, I'll update this part.

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
ShanMa1991: Thanks for the suggestion, I'll replace this part.

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
ShanMa1991: Thanks for the suggestion, I'll change the Set to array to verify the sequence.

Tests/MapboxNavigationTests/RouteControllerSnapshotTests.swift
ShanMa1991: Thanks for the suggestion, I'll change the target of this file.

1ec5:
Now that #2863 has landed, rebasing should fix the xcode-12-examples build. The Xcode_12.2_iOS_14.2 build would still fail due to #2866.

1ec5 (approved):

Tests/MapboxCoreNavigationTests/SimulatedLocationManagerTests.swift
```diff
@@ -1,19 +1,20 @@
 import XCTest
-import FBSnapshotTestCase
-import TestHelper
 import MapboxDirections
 @testable import MapboxCoreNavigation
-@testable import MapboxNavigation
+#if !SWIFT_PACKAGE
```
1ec5, line 4: If we don’t figure out #2866 first, make sure to at least run the tests locally using the Carthage-based target in Xcode. RouteControllerTests and SimulatedLocationManagerTests no longer run in an SPM-based test target, so the Carthage-based MapboxCoreNavigationTests is the only way to know whether these tests still pass.

ShanMa1991: Thanks for the suggestion, I'll test these tests.

PR #2864, opened 2021-03-13
Update Transifex configuration after SPM refactor

1ec5:
Updated the Transifex configuration to reflect file moves in #2808 to support Swift Package Manager. Also synchronized the localizations with Transifex, pulling in some Catalan translations.

/cc @mapbox/navigation-ios


avi-c (approved):
Straightforward and looks good.

avi-c (approved):
I can't perform the build check since I'm not at my work computer but from a textual review this all looks good.

PR #2870, opened 2021-03-16
Port Route Duration Annotations feature to Maps SDK v10 map styling API

avi-c:
Maps v10 replaced the old NSExpression-based style layer API with a new Expressions API.

This necessitates porting over the Route Durations Annotations to the new API.


Example/ViewController.swift
```diff
@@ -131,9 +135,17 @@ class ViewController: UIViewController {
         startButton.isEnabled = false
         clearMap.isHidden = true
         longPressHintView.isHidden = false
+
+        #if false
```
azarovalex, line 139: do we still need these lines?

Sources/MapboxNavigation/Resources/Assets.xcassets/RouteAnnotations/RouteInfoAnnotationLeftHanded.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
+{
+  "images" : [
+    {
+      "filename" : "RouteInfoAnnotationLeftHanded.png",
```
azarovalex, line 4: In my experience .pdf files can take less space, can't we use them here?

Sources/MapboxNavigation/Array.swift
```diff
@@ -1,4 +1,7 @@
 import Foundation
+import MapboxDirections
+import CoreLocation
+import Turf
```
azarovalex, line 4: I'm not sure that this file should know about all Array extensions we have in our project. I would have created Array+RouteStep.swift file, but if this change seems normal for everyone else, we can go with it.

Sources/MapboxNavigation/UIImage.swift
```diff
@@ -43,4 +43,29 @@ extension UIImage {
 
         return scaledImage
     }
+
+    func tint(_ tintColor: UIColor) -> UIImage {
```
azarovalex, line 47: Can't we just use UIImageRenderingModeAlwaysTemplate? I should do something similar.

Sources/MapboxNavigation/Route.swift
```diff
@@ -42,4 +42,62 @@ extension Route {
             return LineString(trimmedPrecedingCoordinates + followingPolyline.trimmed(from: followingPolyline.coordinates[0], distance: distance)!.coordinates.suffix(from: 1))
         }
     }
+
+    var tollIntersections: [Intersection]? {
+        let allSteps = legs.flatMap { return $0.steps }
```
azarovalex, line 47: There are already no return statements in small closures in this file, maybe it's better to remove them here?

Sources/MapboxNavigation/Route.swift
```diff
@@ -42,4 +42,62 @@ extension Route {
...
+
+        let allIntersections = allSteps.compactMap { return $0.intersections }.reduce([], +)
```
azarovalex, line 49:
```suggestion
        let allIntersections = allSteps.flatMap { $0.intersections ?? [] }
```
this will work in one pass, instead of two, a little bit faster🙂

Sources/MapboxNavigation/Route.swift
```diff
@@ -42,4 +42,62 @@ extension Route {
...
+
+        // return the segments as LineStrings
+        return lines.compactMap { coordinateList -> LineString? in
+            return LineString(coordinateList)
+        }
+    }
+}
+
+extension RouteStep {
+    func intersects(_ boundingBox: Turf.BoundingBox) -> Bool {
```
azarovalex, line 93: We can replace this function with one line.
`return shape?.coordinates.contains(where: { boundingBox.contains($0) }) ?? false`
or two lines
```
guard let coordinates = shape?.coordinates else { return false }
return coordinates.contains(where: { boundingBox.contains($0) })
```

Sources/MapboxNavigation/MapboxMapsStyle.swift
```diff
@@ -1,4 +1,5 @@
 import MapboxMaps
+import Turf
```
azarovalex, line 2: it looks like this import isn't used here, or we should uncomment something

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -687,6 +690,51 @@ extension CarPlayManager: MapTemplateProviderDelegate {
...
+        window.rootViewController = nil
+        window.isHidden = true
+        window.removeFromSuperview()
+
+        mainMapTemplate = nil
+        carWindow = nil
+
+        eventsManager.sendCarPlayDisconnectEvent()
+
+        if let shouldDisableIdleTimer = delegate?.carplayManagerShouldDisableIdleTimer(self) {
```
azarovalex, line 730: I think we can replace this if statement with
```swift
UIApplication.shared.isIdleTimerDisabled = delegate?.carplayManagerShouldDisableIdleTimer(self) ?? false
```

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -687,6 +690,51 @@ extension CarPlayManager: MapTemplateProviderDelegate {
...
+            UIApplication.shared.isIdleTimerDisabled = !shouldDisableIdleTimer
```
azarovalex, line 731: do we really need `!` here?
`isIdleTimerDisabled = shouldDisableIdleTimer` seems reasonable

avi-c:
Closing in favor of https://github.com/mapbox/mapbox-navigation-ios/pull/2873

PR #2871, opened 2021-03-17
Don't trigger viewDidLoad in NavigationViewController.init

### LINKED ISSUE #2786: Feedback UI always appears for short routes
**Mapbox Navigation SDK version: 1.1.0 and 1.2.0-rc.1**

_This is similar to https://github.com/mapbox/mapbox-navigation-ios/issues/2056, but I'm filing anew since that one is stale._

### Steps to reproduce

 1. Initialize a `NavigationViewController` with a very short route, one where you're basically on the destination.
 2. Immediately after the initializer, set `showsReportFeedback` to false.
 3. Show the `NavigationViewController`.

```Swift
let route: IndexedRoute = ...
let navigationViewController = NavigationViewController(
    for: route.0,
    routeIndex: route.1,
    routeOptions: routeOptions,
    navigationOptions: navigationOptions
)
navigationViewController.showsReportFeedback = false
...
// embed or show the navigationViewController
```

### Expected behavior

When the `NavigationViewController` is presented, there will be no feedback UI on screen because `showsReportFeedback` was set to false.

### Actual behavior

The [`navigationService(_:, didArriveAt:)`](https://github.com/mapbox/mapbox-navigation-ios/blob/v1.1.0/MapboxNavigation/NavigationViewController.swift#L626) delegate method within `NavigationViewController` is firing during the initializer, which presents the end of route feedback before the assignment to `showsReportFeedback` (or `showsEndOfRouteFeedback`) is applied.

I know this is a bit of an edge case because it only happens for short routes, but any advice on how to avoid this would be greatly appreciated! The feedback UI doesn't match our app's branding and we'd like to avoid presenting it. This wasn't an issue in SDK 0.39.0, which is what I'm upgrading from.

I'm not able to override the `showEndOfRouteFeedback(duration:, completionHandler:)` extension method, nor can I override the `showsEndOfRouteFeedback` or `showsReportFeedback` variables.

azarovalex:
### Description
Fixes #2786.

### Implementation
The main problem I see right now is that the initializer of NavigationViewController uses the `view` property and therefore triggers `viewDidLoad`. This in turn triggers `self.navigationService.start()`. It leads to the situation where a user already created NavigationViewController but hasn't presented it just yet but we've started the navigation session. 
Expected behavior for `viewDidLoad` is to be executed when the actual presentation of the NavigationViewController happens.
In fact, current behavior leads to the bug #2786. If the route is very small we can arrive at the final point during the initialization of NavigationViewController. The call stack will look like this:
`init -> viewDidLoad -> navigationService.start() -> .... -> navigationService(_:,didArrive:)`
[At this point,](https://github.com/mapbox/mapbox-navigation-ios/blob/main/Sources/MapboxNavigation/NavigationViewController.swift#L691) we should notify NavigationViewController.delegate about arrival, but `delegate` will be always nil because we are still inside of initialization and the user has no opportunity to set the delegate yet.
The second bug happens on the next line. We check `showsEndOfRouteFeedback` property and always show a feedback screen since the default value of this property is `true`, this happens even if the user explicitly sets this property to `false`.

The solution that I propose in this PR is to postpone all work with view property until actual viewDidLoad happens.

This is my first PR here, so any comments are welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changes to the Mapbox Navigation SDK for iOS
 
+## v1.4.0
+
+* Fixed an issue when feedback UI was always appearing for short routes ([#2871](https://github.com/mapbox/mapbox-navigation-ios/pull/2871))
+
 ## v1.3.0
 
 * MapboxCoreNavigation can now be installed using Swift Package Manager. ([#2771](https://github.com/mapbox/mapbox-navigation-ios/pull/2771))
--- a/Sources/MapboxNavigation/NavigationViewController.swift
+++ b/Sources/MapboxNavigation/NavigationViewController.swift
@@ -329,17 +329,10 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
         mapViewController.destination = route.legs.last?.destination
         mapViewController.view.translatesAutoresizingMaskIntoConstraints = false
         
-        embed(mapViewController, in: view) { (parent, map) -> [NSLayoutConstraint] in
-            return map.view.constraintsForPinning(to: parent.view)
-        }
-        
         //Manually update the map style since the RMVC missed the "map style change" notification when the style manager was set up.
         if let currentStyle = styleManager.currentStyle {
             updateMapStyle(currentStyle, animated: false)
         }
-        
-        mapViewController.view.pinInSuperview()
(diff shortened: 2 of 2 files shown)


MaximAlien (approved):
Overall looks good. Please update `CHANGELOG.md` to mention current fix.

1ec5 (approved):

PR #2873, opened 2021-03-17
Port Route Duration Annotations feature to Maps SDK v10 map styling API

avi-c:
Maps v10 replaced the old NSExpression-based style layer API with a new Expressions API.

This necessitates porting over the Route Durations Annotations to the new API.


avi-c:
Replaces https://github.com/mapbox/mapbox-navigation-ios/pull/2870.

avi-c:
Incorporated some of the feedback that @azarovalex included in https://github.com/mapbox/mapbox-navigation-ios/pull/2870

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -382,6 +382,7 @@
 		DAD903AF23E3DCC80057CF1F /* DateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAD903AE23E3DCC80057CF1F /* DateTests.swift */; };
 		DADD82802161EC0300B8B47D /* UIViewAnimationOptionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DADD827F2161EC0300B8B47D /* UIViewAnimationOptionsTests.swift */; };
 		DAFA92071F01735000A7FB09 /* DistanceFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 351BEC0B1E5BCC72006FE110 /* DistanceFormatter.swift */; };
+		F46FF187260277F7007CC0E0 /* DateComponentsFormatter+MGLNavigationAdditions.swift in Sources */ = {isa = PBXBuildFile; fileRef = F46FF186260277F7007CC0E0 /* DateComponentsFormatter+MGLNavigationAdditions.swift */; };
```
1ec5, line 385: Nit: The “MGL” class prefix is no longer used in general.

1ec5 (commented):
Looks like these are new commits atop release-v2.0 rather than a merge commit. That’s fine, but eventually we’ll need to perform a merge of main into release-v2.0. When we do that, we’ll be able to ignore any conflicts related to route duration annotations as long as this PR has already landed by then.

MapboxNavigation.xcodeproj/project.pbxproj
avi-c: Renamed the file to remove the MGL prefix

avi-c:
@1ec5 - I think all the feedback is addressed. Would it be possible to approve this so we can merge it in?

avi-c:
@1ec5 - Let me know if there is more to do here or if I can update and merge it in soon.

Sources/MapboxNavigation/DayStyle.swift
```diff
@@ -155,6 +162,11 @@ open class DayStyle: Style {
         NavigationMapView.appearance().trafficUnknownColor = .trafficUnknown
         NavigationMapView.appearance().buildingDefaultColor = .defaultBuildingColor
         NavigationMapView.appearance().buildingHighlightColor = .defaultBuildingHighlightColor
+        NavigationMapView.appearance().routeDurationAnnotationColor = .routeDurationAnnotationColor
+        NavigationMapView.appearance().routeDurationAnnotationSelectedColor = .selectedRouteDurationAnnotationColor
+        NavigationMapView.appearance().routeDurationAnnotationFontName = "DIN Pro Medium"
```
1ec5, line 167: In general, when styling text on the map that we expect to render [server-side](https://docs.mapbox.com/ios/maps/api/6.3.0/customizing-fonts.html#server-side-fonts), we need to specify a list of fonts including a fallback font. Otherwise, the fontstack could be missing glyphs that are required for rendering the street names, even if the same street names render just fine on the base map: mapbox/mapbox-maps-ios#278.

Since font selection for a symbol layer is so specific to the style specification and unlike UIKit/Core Text font selection, the immediate goal should be to match the fonts in the base style.

Consider using the runtime styling API to get the `textFont` property of the symbol layer used to render roads (so the layer would have a source layer of `road`) and copying that expression to the layer for the annotations. Typically, a style will use multiple distinct fontstacks for various road label layers, but perhaps we could blithely pick the topmost matching layer or add heuristics to detect and ignore italics and boldface.

avi-c: These are good points. I changed the code to allow for a list of font names to be used and not just a single one. This lets the developer choose fonts that may be present in their map style. Also changed the default font of `DIN Pro Medium` to also include `Noto Sans CJK JP Medium` and `Arial Unicode MS Regular` as fallbacks in order to support more glyphs "out of the box" with our default set.
I think that the use of UIAppearance overrides for these map rendered fonts is probably the clearest way we can allow for supporting multiple fonts and fallbacks. I think it's preferable to trying to locate the font settings of a specific layer in the style, which feels more hidden and brittle, even if the source layer name for roads, `road` is unlikely to change very often in mapbox-streets.

CHANGELOG.md
```diff
@@ -99,6 +99,9 @@
   * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteCasingStyleLayerWithIdentifier:source:)` to style the casing of alternative route.
 * Fixed an issue where the route line periodically peeked out from behind the user puck even though `NavigationViewController.routeLineTracksTraversal` was enabled. ([#2737](https://github.com/mapbox/mapbox-navigation-ios/pull/2737))
 * Created the `UserHaloCourseView` similar to `UserCourseView` for approximate location on iOS 14 during the navigation to represent user location. Allow the switch between `UserHaloCourseView` and `UserCourseView` when precise mode is changed. ([#2664](https://github.com/mapbox/mapbox-navigation-ios/pull/2664))
+* Added option to show route duration callouts when previewing route alternatives ([#2734](https://github.com/mapbox/mapbox-navigation-ios/pull/2734)):
+  * `NavigationMapView.showRouteDurations(along routes:)` to show duration annotation callouts on the map for the provided routes.
```
1ec5, line 103:
```suggestion
  * `NavigationMapView.showRouteDurations(along:)` to show duration annotation callouts on the map for the provided routes.
```

CHANGELOG.md
```diff
@@ -99,6 +99,9 @@
   * `NavigationMapViewDelegate.navigationMapView(_:alternativeRouteCasingStyleLayerWithIdentifier:source:)` to style the casing of alternative route.
 * Fixed an issue where the route line periodically peeked out from behind the user puck even though `NavigationViewController.routeLineTracksTraversal` was enabled. ([#2737](https://github.com/mapbox/mapbox-navigation-ios/pull/2737))
 * Created the `UserHaloCourseView` similar to `UserCourseView` for approximate location on iOS 14 during the navigation to represent user location. Allow the switch between `UserHaloCourseView` and `UserCourseView` when precise mode is changed. ([#2664](https://github.com/mapbox/mapbox-navigation-ios/pull/2664))
+* Added option to show route duration callouts when previewing route alternatives ([#2734](https://github.com/mapbox/mapbox-navigation-ios/pull/2734)):
```
1ec5, line 102: This is already mentioned in the v1.2.0 release notes. If we’re keeping the feature functional but otherwise leaving it in place, then there’s no need to mention anything in the changelog.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -5,6 +5,12 @@ import MapboxDirections
 import MapboxCoreNavigation
 import Turf
 
+
+private enum RouteDurationAnnotationTailPosition: Int {
```
1ec5, line 9: This probably doesn’t need a raw type. It can also be a closed enumeration.

Sources/MapboxNavigation/DayStyle.swift
```diff
@@ -165,6 +172,11 @@ open class DayStyle: Style {
         NavigationMapView.appearance().alternativeTrafficUnknownColor = .alternativeTrafficUnknown
         NavigationMapView.appearance().buildingDefaultColor = .defaultBuildingColor
         NavigationMapView.appearance().buildingHighlightColor = .defaultBuildingHighlightColor
+        NavigationMapView.appearance().routeDurationAnnotationColor = .routeDurationAnnotationColor
+        NavigationMapView.appearance().routeDurationAnnotationSelectedColor = .selectedRouteDurationAnnotationColor
+        NavigationMapView.appearance().routeDurationAnnotationFontName = ["DIN Pro Medium", "Noto Sans CJK JP Medium", "Arial Unicode MS Regular"]
```
1ec5, line 177: Should this array be extracted into a constant?

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -123,6 +130,13 @@ open class NavigationMapView: UIView {
      Most recent user location, which is used to place `UserCourseView`.
      */
     var mostRecentUserCourseViewLocation: CLLocation?
+
+    @objc dynamic public var routeDurationAnnotationSelectedColor: UIColor = .selectedRouteDurationAnnotationColor
+    @objc dynamic public var routeDurationAnnotationColor: UIColor = .routeDurationAnnotationColor
+    @objc dynamic public var routeDurationAnnotationSelectedTextColor: UIColor = .selectedRouteDurationAnnotationTextColor
+    @objc dynamic public var routeDurationAnnotationTextColor: UIColor = .routeDurationAnnotationTextColor
+    @objc dynamic public var routeDurationAnnotationFontName: [String] = ["DIN Pro Medium", "Noto Sans CJK JP Medium", "Arial Unicode MS Regular"]
```
1ec5, line 138: We should document the font name option somewhere, perhaps here. A key part of the documentation would be linking to the map SDK’s font selection guide.

1ec5, line 138: I still think a better _default_ behavior should be to match the style’s existing road label fontstack, even if this UIAppearance option remains as an override. But `routeDurationAnnotationFontName` is not new, so this can be tail work.

1ec5, line 138: Now that this property is an array of font names, the name of the property should be plural. This change should be documented in the changelog.

avi-c: I'm not totally convinced, but I do think you have a good point in that not picking up the map style's font stack removes the ability of the map designer from changing this without help from engineering.

I wrote up https://github.com/mapbox/mapbox-navigation-ios/issues/2994 to track this later change.

Sources/MapboxNavigation/NavigationMapView.swift
avi-c: We do need the raw type (I believe) so that the raw value can be set in the feature properties dictionary in order to be picked up by the map runtime styling.

Sources/MapboxNavigation/DayStyle.swift
```diff
@@ -56,6 +56,13 @@ extension UIFont {
     // General styling
     fileprivate class var defaultPrimaryText: UIFont { get { return UIFont.systemFont(ofSize: 26) } }
     fileprivate class var defaultSecondaryText: UIFont { get { return UIFont.systemFont(ofSize: 16) } }
+
+    /**
+     List of Mapbox Maps font names to be used for any symbol layers added by the Navigation SDK.
+     These are used for features such as Route Duration Annotations that are optionally added during route preview.
+     See https://docs.mapbox.com/ios/maps/api/6.3.0/customizing-fonts.html for more information about server-side fonts.
+     */
+    fileprivate class var defaultNavigationSymbolLayerFontList: [String] { return ["DIN Pro Medium", "Noto Sans CJK JP Medium", "Arial Unicode MS Regular"] }
```
1ec5, line 65: Nit: Rename to `defaultNavigationSymbolLayerFontNames` for consistency.

Sources/MapboxNavigation/DayStyle.swift
```diff
@@ -56,6 +56,13 @@ extension UIFont {
     // General styling
     fileprivate class var defaultPrimaryText: UIFont { get { return UIFont.systemFont(ofSize: 26) } }
     fileprivate class var defaultSecondaryText: UIFont { get { return UIFont.systemFont(ofSize: 16) } }
+
+    /**
+     List of Mapbox Maps font names to be used for any symbol layers added by the Navigation SDK.
+     These are used for features such as Route Duration Annotations that are optionally added during route preview.
+     See https://docs.mapbox.com/ios/maps/api/6.3.0/customizing-fonts.html for more information about server-side fonts.
```
1ec5, line 63: Where will the (not terribly intuitive) font fallback behavior be documented for map SDK v10? Will it be in the v10 API reference or a guide on docs.mapbox.com?

/cc @mapbox/docs

avi-c: That is a very good question. I looked for the font documentation in the v10 beta documentation and it isn't present. I went with the less desired approach of mentioning the v6.3 documentation until the place in the v10 documentation is sorted out.

Perhaps I should write a ticket to update this comment when the docs for Maps v10 are finalized.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -135,6 +135,12 @@ open class NavigationMapView: UIView {
     @objc dynamic public var routeDurationAnnotationColor: UIColor = .routeDurationAnnotationColor
     @objc dynamic public var routeDurationAnnotationSelectedTextColor: UIColor = .selectedRouteDurationAnnotationTextColor
     @objc dynamic public var routeDurationAnnotationTextColor: UIColor = .routeDurationAnnotationTextColor
+
+    /**
+     List of Mapbox Maps font names to be used for any symbol layers added by the Navigation SDK.
+     These are used for features such as Route Duration Annotations that are optionally added during route preview.
+     See https://docs.mapbox.com/ios/maps/api/6.3.0/customizing-fonts.html for more information about server-side fonts.
```
1ec5, line 142: Use Markdown formatting:

```suggestion
     This property affects the font applied to the route duration annotations that the `showRouteDurations(along:)` method adds to the map view. [The map view selects the font](https://docs.mapbox.com/ios/maps/api/6.3.0/customizing-fonts.html) in the same manner as any other symbol layer that you add to the style.
```

1ec5 (approved):

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -805,6 +825,326 @@ open class NavigationMapView: UIView {
...
+        """
+        ["get", "sortOrder"]
+        """
+
+        if let expressionData = symbolSortKeyString.data(using: .utf8), let expJSONObject = try? JSONSerialization.jsonObject(with: expressionData, options: []) {
+
+            mapView.mapboxMap.__map.setStyleLayerPropertyForLayerId(NavigationMapView.routeDurationAnnotationsLayerIdentifier,
+                                                          property: "symbol-sort-key",
+                                                          value: expJSONObject)
+        }
```
1ec5, line 1064: I overlooked the possibility of using `Expression` here: #2996.

PR #2876, opened 2021-03-18
Allow the Navigator set access token other than in Info.plist

### LINKED ISSUE #2851: Navigator.shared only accepts access token from Info.plist
#2823 introduced unconditional usage of `Directions.shared`, which means there’s no way to set the access token other than in Info.plist. There needs to be a way to preemptively, programmatically set the access token or DirectionsCredentials that the first instance of Navigator would use. Probably this would end up being a static property on the Navigator class, backed by a global static variable, as unfortunate as that would be.

https://github.com/mapbox/mapbox-navigation-ios/blob/df75965f1eb92c03db21b9ad548200b2fe5e0312/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift#L60

Info.plist is supposed to be an optional way of setting the access token; we always officially support a programmatic alternative for developers who can’t put the token in their Info.plist for various reasons, such as dynamically switching between access tokens. Moreover, as seen in https://github.com/mapbox/mapbox-navigation-ios/pull/2834#issuecomment-791829886, automated tests often need to be able to mock an access token.

/cc @mapbox/navigation-ios

ShanMa1991:
### Description
This pr is to fix the integration error and other errors in the MapboxcoreNavigationTest.
- [x] remove deprecated property list key and check.
- [x] add workaround for bundle info check under Carthage
- [x] fix the directions in MapboxNavigationService.
- [x] fix #2851 , allow the `Navigator` accepts DirectionsCredentials from `PassiveLocationDataSource`, `RouteController` and `PredictiveCacheManager`.
- [x] allow all the MapboxCoreNavigationTests run/build (except some exempt from SPM).
- [x] update the changelog.

### Implementation
Creating a property called `credentials` in `Navigator` to be set before getting the shared navigator. It could take value value from `PassiveLocationDataSource` and `RouteController`.
By adding a `credentials` property to `PredictiveCacheOptions`, it also allows the `PredictiveCacheManager` passing the DirectionsCredentials to `Navigator`.

### DIFF
--- a/MapboxNavigation.xcodeproj/project.pbxproj
+++ b/MapboxNavigation.xcodeproj/project.pbxproj
@@ -3307,6 +3307,7 @@
 				PRODUCT_NAME = "$(TARGET_NAME:c99extidentifier)";
 				SKIP_INSTALL = YES;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SYSTEM_FRAMEWORK_SEARCH_PATHS = "";
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Debug;
@@ -3343,6 +3344,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.mapbox.TestHelper;
 				PRODUCT_NAME = "$(TARGET_NAME:c99extidentifier)";
 				SKIP_INSTALL = YES;
+				SYSTEM_FRAMEWORK_SEARCH_PATHS = "";
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Release;
--- a/Sources/MapboxCoreNavigation/BundleAdditions.swift
+++ b/Sources/MapboxCoreNavigation/BundleAdditions.swift
@@ -23,12 +23,6 @@ extension Bundle {
         }
     }
     
-    var locationAlwaysUsageDescription: String? {
-        get {
-            return object(forInfoDictionaryKey: "NSLocationAlwaysUsageDescription") as? String
-        }
-    }
-    
     /**
      The Mapbox Core Navigation framework bundle.
      */
--- a/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
+++ b/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
@@ -46,6 +46,13 @@ class Navigator {
     }()
     
     /**
+     The Authorization & Authentication credentials that are used for this service. If not specified - will be automatically intialized from the token and host from your app's `info.plist`.
+     
+     - precondition: `credentials` should be set before getting the shared navigator for the first time.
+     */
+    static var credentials: DirectionsCredentials? = nil
+    
+    /**
      Provides a new or an existing `MapboxCoreNavigation.Navigator` instance. Upon first initialization will trigger creation of `MapboxNavigationNative.Navigator` and `HistoryRecorderHandle` instances,
      satisfying provided configuration (`tilesVersion` and `tilesURL`).
      */
@@ -68,7 +75,7 @@ class Navigator {
         let settingsProfile = SettingsProfile(application: ProfileApplication.kMobile,
                                               platform: ProfilePlatform.KIOS)
         
-        let endpointConfig = TileEndpointConfiguration(credentials: Directions.shared.credentials,
+        let endpointConfig = TileEndpointConfiguration(credentials:Navigator.credentials ?? Directions.shared.credentials,
                                                        tilesVersion: Self.tilesVersion,
                                                        minimumDaysToPersistVersion: nil)
         
--- a/Sources/MapboxCoreNavigation/NavigationService.swift
+++ b/Sources/MapboxCoreNavigation/NavigationService.swift
@@ -575,7 +575,9 @@ private func checkForLocationUsageDescription() {
     guard let _ = Bundle.main.bundleIdentifier else {
         return
     }
-    if Bundle.main.locationAlwaysUsageDescription == nil && Bundle.main.locationWhenInUseUsageDescription == nil && Bundle.main.locationAlwaysAndWhenInUseUsageDescription == nil {
-        preconditionFailure("This application’s Info.plist file must include a NSLocationWhenInUseUsageDescription. See https://developer.apple.com/documentation/corelocation for more information.")
+    if Bundle.main.locationWhenInUseUsageDescription == nil && Bundle.main.locationAlwaysAndWhenInUseUsageDescription == nil {
+        if UserDefaults.standard.object(forKey: "NSLocationWhenInUseUsageDescription") == nil && UserDefaults.standard.object(forKey: "NSLocationAlwaysAndWhenInUseUsageDescription") == nil {
+                    preconditionFailure("This application’s Info.plist file must include a NSLocationWhenInUseUsageDescription. See https://developer.apple.com/documentation/corelocation for more information.")
+        }
     }
 }
--- a/Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
+++ b/Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
@@ -23,6 +23,17 @@ let waitForInterval: TimeInterval = 5
 class MapboxCoreNavigationTests: XCTestCase {
     var navigation: MapboxNavigationService!
     
+    override func setUp() {
+        super.setUp()
+        UserDefaults.standard.set("Location Usage Description", forKey: "NSLocationWhenInUseUsageDescription")
+        UserDefaults.standard.set("Location Usage Description", forKey: "NSLocationAlwaysAndWhenInUseUsageDescription")
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        UserDefaults.resetStandardUserDefaults()
+    }
+    
     func testNavigationNotificationsInfoDict() {
         navigation = MapboxNavigationService(route: route, routeIndex: 0, routeOptions: routeOptions, directions: directions, simulating: .never)
         let now = Date()
--- a/Tests/MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
+++ b/Tests/MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
@@ -34,7 +34,7 @@ class NavigationEventsManagerTests: XCTestCase {
         let locationManager = NavigationLocationManager()
         let service = MapboxNavigationService(route: firstRoute, routeIndex: 0,
                                               routeOptions: firstRouteOptions,
-                                              directions: nil,
+                                              directions: directions,
                                               locationSource: locationManager,
                                               eventsManagerType: NavigationEventsManagerSpy.self,
                                               simulating: .always)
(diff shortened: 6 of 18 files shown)


Sources/MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -23,12 +23,6 @@ extension Bundle {
         }
     }
     
-    var locationAlwaysUsageDescription: String? {
-        get {
-            return object(forInfoDictionaryKey: "NSLocationAlwaysUsageDescription") as? String
```
1ec5, line 28: 👍 `NSLocationAlwaysUsageDescription` is deprecated in iOS 11 and above.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -45,6 +45,8 @@ class Navigator {
         return RoadObjectsStore(try! navigator.roadObjectStore())
     }()
     
+    static var accessToken: String? = nil
```
1ec5, line 48: Let’s store the DirectionsCredentials instead of the bare access token. That way the application can also customize the API endpoint.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -45,6 +45,8 @@ class Navigator {
...
+    
     /**
      Provides a new or an existing `MapboxCoreNavigation.Navigator` instance. Upon first initialization will trigger creation of `MapboxNavigationNative.Navigator` and `HistoryRecorderHandle` instances,
      satisfying provided configuration (`tilesVersion` and `tilesURL`).
```
1ec5, line 52: Add a `- precondition:` mentioning that the property above should be set before getting the shared navigator for the first time.

1ec5 (commented):
`Directions.calculateOffline(options:completionHandler:)` gets `Navigator.shared` so its credentials should be set beforehand:

https://github.com/mapbox/mapbox-navigation-ios/blob/3b6b2605b723f6563a295a3460b32fdbd29f6d46/Sources/MapboxCoreNavigation/Directions.swift#L50-L51

Similarly, `PredictiveCacheManager.createPredictiveCacheController(options:tileStore:version:dataset:maxConcurrentRequests:)` accesses `Navigator.shared`. It gets called from `PredictiveCacheManager.initMapControllers(options:mapOptions:)` and `PredictiveCacheManager.initNavigatorController(options:)`, which in turn get called from `PredictiveCacheManager(predictiveCacheOptions:mapOptions:)`. This is a public initializer, so it should somehow accept a Directions object like RouteController and PassiveLocationDataSource do. It already takes a PredictiveCacheOptions, so it would be convenient to add a Directions- or DirectionsCredentials-typed property to PredictiveCacheOptions.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
ShanMa1991: Thanks for the suggestion, I'll update this part.

1ec5:
@ShanMa1991 pointed out that the tests are still [crashing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/2839/workflows/ecdf454d-fee0-405c-9fed-a4391dd537f7/jobs/28534) due to other Info.plist keys that are missing from the trampoline application that SPM generates:

```
Fatal error: This application’s Info.plist file must include a NSLocationWhenInUseUsageDescription. See https://developer.apple.com/documentation/corelocation for more information.: file MapboxCoreNavigation/NavigationService.swift, line 579
2021-03-22 21:47:47.436333+0000 xctest[3006:68362] Fatal error: This application’s Info.plist file must include a NSLocationWhenInUseUsageDescription. See https://developer.apple.com/documentation/corelocation for more information.: file MapboxCoreNavigation/NavigationService.swift, line 579
```

We’ll need to programmatically exclude this particular assertion from tests.

Tests/MapboxCoreNavigationTests/NavigationEventsManagerTests.swift
```diff
@@ -78,7 +78,7 @@ class NavigationEventsManagerTests: XCTestCase {
         ])
         let eventTimeout = 0.3
         let route = Fixture.route(from: "DCA-Arboretum", options: routeOptions)
-        let dataSource = MapboxNavigationService(route: route, routeIndex: 0, routeOptions: routeOptions)
+        let dataSource = MapboxNavigationService(route: route, routeIndex: 0, routeOptions: routeOptions, directions: directions, simulating: .never)
```
1ec5, line 81: The default value for `simulating` is `onPoorGPS`; would setting it to `never` affect the test behavior?

ShanMa1991: Thanks for the suggestion, I'll fix this part

ShanMa1991:
@1ec5 
Now when the core tests is calling the `func checkForLocationUsageDescription()` of `LegacyRouteController`  and the `NavigationService` ,  the `Bundle.main` is for the XCTest not the core test, so it cause the Bundle could not find the Info.plist . So we just use the `if SWIFT_PACKAGE` to skip the tests.

But if we turn to using the Bundle.module under `if SWIFT_PACKAGE` for `Bundle.mapboxCoreNavigation` to replace the `Bundle.main` in  `LegacyRouteController`  and the `NavigationService`,  the generated Info.plist in SPM doesn't contain certain keys such as `CFBundleShortVersionString` or `NSLocationAlwaysAndWhenInUseUsageDescription` . It will cause the crush in  the turn-by-turn navigation in `NavigationEventsManager` as well.

ShanMa1991:
@1ec5 
The current failure of the `NavigationServiceTests` and `MapboxCoreNavigationTests` are all related with `RouteController`, including the ` func userIsOnRoute(_ location:)` and `func locationManager(_ manager:, didUpdateLocations:)` . As already filed in #2883 , the two tests failures maybe need to be handled later after the fix of the `RouteController`.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -293,6 +293,7 @@
 		B430D2FA25534FDC0088CC23 /* UserHaloCourseView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B430D2F925534FDC0088CC23 /* UserHaloCourseView.swift */; };
 		B48CEFC125796FD300696BB3 /* route-for-vanishing-route-line.json in Resources */ = {isa = PBXBuildFile; fileRef = B48CEFAB25796F5A00696BB3 /* route-for-vanishing-route-line.json */; };
 		B4F2BF3C25FAF69500D7AE2E /* NavigationPlotter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3559FE4E21C3195700B6613F /* NavigationPlotter.swift */; };
+		B4F9FBDF260A61F8008C3D78 /* MapboxCoreNavigation.h in Headers */ = {isa = PBXBuildFile; fileRef = C5ADFBCC1DDCC7840011824B /* MapboxCoreNavigation.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
1ec5, line 296: It’s surprising that MapboxCoreNavigation.h would be added to the TestHelper target. That’s the umbrella header for the MapboxCoreNavigation target and would only be useful for Carthage-based builds. (An umbrella header shouldn’t part of any target other than its own module.)

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -572,10 +572,14 @@ private func checkForUpdates() {
 }
 
 private func checkForLocationUsageDescription() {
+    #if SWIFT_PACKAGE
```
1ec5, line 575: This disables the check even during production when using SPM. We need to find some way of only disabling the check when running tests built using SPM. For example, the test target could set a value in NSUserDefaults that this code can check for. The code in the test target that sets the default could be conditioned on `SWIFT_PACKAGE` while the code that checks for the default can check it regardless of the package manager. (Same thing in LegacyRouteController.swift.)

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -55,7 +55,7 @@ class NavigationServiceTests: XCTestCase {
     }
 
     func testDefaultUserInterfaceUsage() {
-        XCTAssertTrue(dependencies.navigationService.eventsManager.usesDefaultUserInterface, "MapboxCoreNavigationTests should have an implicit dependency on MapboxNavigation due to running inside the Example application target.")
+        XCTAssertTrue(!dependencies.navigationService.eventsManager.usesDefaultUserInterface, "MapboxCoreNavigationTests shouldn't have an implicit dependency on MapboxNavigation due to removing the Example application target as the test host.")
```
1ec5, line 58: At some point we should add a corresponding assertion that `userDefaultUserInterface` is true within MapboxNavigationTests.

MapboxNavigation.xcodeproj/project.pbxproj
ShanMa1991: Thanks for the suggestion,  I'll remove this from the `TestHelper` target. Because sometimes the `TestHelper` still could not find the header of the `MapboxCoreNavigation`, and it would cause the error in the local building.

Sources/MapboxCoreNavigation/NavigationService.swift
ShanMa1991: Thanks for the suggestion, I'll work on this part

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
ShanMa1991: Thanks for the suggestion, I'll add a `TODO` note in this part to add the check in the MapboxNavigationTests.

Sources/MapboxCoreNavigation/NavigationService.swift
ShanMa1991: I add the `UserDefaults` info for location usage in the `MapboxCoreNavigationTests.swift` cases, and the check of the  `UserDefaults` info check in the `func checkForLocationUsageDescription()` to avoid the` Info.plist` check when running tests built using SPM.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
ShanMa1991: I add a corresponding test of `func testDefaultUserInterfaceUsage()` in the `MapboxNavigationTests.NavigationViewControllerTests.swift` when running inside the Example application target.

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -23,6 +23,17 @@ let waitForInterval: TimeInterval = 5
 class MapboxCoreNavigationTests: XCTestCase {
     var navigation: MapboxNavigationService!
     
+    override func setUp() {
+        super.setUp()
+        UserDefaults.standard.set("Location Usage Description", forKey: "NSLocationWhenInUseUsageDescription")
+        UserDefaults.standard.set("Location Usage Description", forKey: "NSLocationAlwaysAndWhenInUseUsageDescription")
```
1ec5, line 29: Sorry for being unclear: my suggestion in https://github.com/mapbox/mapbox-navigation-ios/pull/2876#discussion_r603571053 was to simply set a generic flag like `MBXIsTestingCoreNavigation` to `true` and checking that it’s true in `checkForLocationUsageDescription()`. However, this approach is even better in a way; we can follow this pattern in the future if there’s any code that cares about the actual value of the key, not just whether it’s present. 👍

1ec5 (approved):

PR #2880, opened 2021-03-23
Fix memory leak when showing shield icon.

MaximAlien:
### Description

PR fixes memory leak when showing shield icon.


1ec5 (approved):
I think mapbox/mapbox-maps-ios#128 should be fixed so developers don’t accidentally introduce bugs like the one fixed here.

azarovalex (approved):

Issue #2791, opened 2021-01-29
Migrate snapshot tests away from unmaintained testing frameworks


1ec5:
MapboxNavigationTests currently relies on a varied assortment of testing frameworks; most of them are redundantly used for snapshot tests:

* [Cedar](https://github.com/cedarbdd/cedar/) (unmaintained)
* [Quick](https://github.com/Quick/Quick/)
* [Nimble](https://github.com/Quick/Nimble/)
* [OHHTTPStubs](https://github.com/AliSoftware/OHHTTPStubs/)
* [FBSnapshotTestCase](https://github.com/facebookarchive/ios-snapshot-test-case/) (unmaintained)
* [SnappyShrimp](https://github.com/AndriiDoroshko/SnappyShrimp/) (unmaintained)

Some of these testing frameworks are unmaintained and probably won’t survive the transition to SPM in #2629. We should migrate any tests using Cedar, FBSnapshotTestCase, or SnappyShrimp to an actively maintained fork or an alternative library. Ideally, we should end up with fewer testing dependencies, because each dependency adds a learning curve and an unnecessary increase in the time it takes to set up a development environment using Carthage or CocoaPods.

/cc @mapbox/navigation-ios

zugaldia:
Are any of these libraries shared across Mapbox SDKs? Wondering if there's an opportunity here to consolidate testing strategies across products.

/cc: @knov

1ec5:
> Are any of these libraries shared across Mapbox SDKs?

* All our Swift service libraries use OHHTTPStubs for mocking network-related tests.
* Apex uses SnappyShrimp.
* 1TAP uses Quick and Nimble.
* MapboxMobileEvents uses Cedar.

1ec5:
The map SDK uses XCUITest directly for UI testing, but there’s a limit to how much we can share approaches between the two SDKs, because testing the internals of a view is different than testing a whole UI.

1ec5:
[iOSSnapshotTestCase](https://github.com/uber/ios-snapshot-test-case) is a popular successor to FBSnapshotTestCase, but [it isn’t maintained either](https://github.com/uber/ios-snapshot-test-case/pull/126#issuecomment-723352614) and doesn’t support SPM: uber/ios-snapshot-test-case#91.

1ec5:
[SnapshotTesting](https://github.com/pointfreeco/swift-snapshot-testing/) looks like a promising replacement for all three unsupported frameworks.

[...]

PR #2882, opened 2021-03-23
Add more turn lane configurations

### LINKED ISSUE #576: Lane icons need to handle more lane configurations

1ec5:
LaneView now supports the additional turn lane icon assets introduced in mapbox/navigation-ui-resources#21, including the combination U-turn–left turn that’s so common in the iOS simulator’s home of Cupertino:

uturn

~~As seen in the screenshot, some of the wider U-turn-laden icons get cropped even though they don’t bleed over the edge of the canvas in PaintCode. It’s probably a layout issue in LaneView or LanesView.~~ (Ignore the fact that the number 2 lane is highlighted as a valid lane; that seems to be a bug in OSRM that isn’t present in a Valhalla-based Directions API endpoint. I’m guessing OSRM isn’t recognizing the divided intersection, so it thinks there are two very closely spaced left turns, the first of which can use either left turn lane.)

Unfortunately, I was unable to meaningfully update LaneViewTests, because it’s part of MapboxNavigationTests, which don’t run as part of SPM builds due to #2791 and don’t run as part of Carthage builds due to a lack of support for Carthage in MapboxNavigation. All I was able to do there was to incorporate some renames that took place internally in LaneView, but there are no test cases to exercise the new assets.

More to come:

* [x] Import a new LanesStyleKit from PaintCode
* [x] Remove limitations on the output of `LaneIndication.ranked(favoring:)`
* [x] Add cases to `LaneConfiguration` that reflect the expanded repertoire of assets
* [x] Add cases to `LaneView.draw(_:)` for calling the new LanesStyleKit methods
* [ ] Consolidate similar cases in `LaneConfiguration(rankedIndications:drivingSide:)`
* [ ] Update LaneViewTests
* [x] Fix cropping of some of the wider icons
* [ ] Update the changelog

Fixes #576.

/cc @mapbox/navigation-ios @abhishek1508 @avi-c


1ec5:
> As seen in the screenshot, some of the wider U-turn-laden icons get cropped even though they don’t bleed over the edge of the canvas in PaintCode. It’s probably a layout issue in LaneView or LanesView.

The problem turned out to be that most of the arrows added in mapbox/navigation-ui-resources#21 don’t respect the `scaleX` variable and thus the `flipHorizontally` variable, and also that the origins of the new arrows aren’t at the center of the drawing.

avi-c:
I have verified that the dual-use icons are now composited correctly and that the "slight" turn variants are shown correctly. I haven't been able to verify all the combinations in the simulator, but I believe we can merge this and deal with any specific issues as they pop up.

avi-c (approved):
Passes my test cases and the code looks good. Hopefully we can simplify this soon using additional context from the Directions API.

PR #2884, opened 2021-03-24
Upgrade to MapboxMaps v10.0.0-beta.16, MapboxNavigationNative v47.0.0

MaximAlien:
### Description

PR updates to latest version of Mapbox Maps SDK.

Release note entries for the next beta:

* Added the `ElectronicHorizon.Edge.Metadata.directionality` property to indicate whether the edge is a one-way or two-way road. (#2884)
* The `ElectronicHorizon.Edge.Metadata.countryCode` property is now set to an ISO 3166-1 alpha-2 code instead of an alpha-3 code. (#2884)
* `PredictiveCacheManager` avoids requesting too many tiles too rapidly. (#2884)
* Fixed an issue where `PredictiveCacheManager` loaded no tiles if `PredictiveCacheOptions.currentLocationRadius` was set to too small a value. (#2884)
* `PredictiveCacheManager` no longer aggressively caches the entirety of a long route. (#2884)
* Fixed an issue where `RouteController` fails to consider a multi-destination route complete. (#2884)
* `RouteController` more reliably detects when location updates jump. (#2884)

As a precaution, an application that uses a previous prerelease of navigation SDK v2.0.0 may need to be reinstalled (clearing the Application Support folder) to avoid a crash on launch.


Package.swift
```diff
@@ -25,7 +25,7 @@ let package = Package(
         .package(name: "MapboxGeocoder", url: "https://github.com/mapbox/MapboxGeocoder.swift.git", from: "0.14.0"),
         .package(name: "MapboxMobileEvents", url: "https://github.com/mapbox/mapbox-events-ios.git", from: "0.10.8"),
         .package(name: "MapboxNavigationNative", url: "https://github.com/mapbox/mapbox-navigation-native-ios.git", from: "45.0.2"),
-        .package(name: "MapboxMaps", url: "https://github.com/mapbox/mapbox-maps-ios.git", .exact("10.0.0-beta.15")),
+        .package(name: "MapboxMaps", url: "https://github.com/mapbox/mapbox-maps-ios.git", .revision("7425b7420fbf91d3666dff8dc229acdb44f2db33")),
```
1ec5, line 28: This is technically based on the latest commit in mapbox/mapbox-maps-ios#205, rather than main, so we’ll need to update it to a branch or tag before landing this PR, to avoid a reference to a commit hash that may go away due to a rebase upstream.

Sources/MapboxNavigation/RouteMapViewController.swift
```diff
@@ -405,7 +405,9 @@ class RouteMapViewController: UIViewController {
     }
     
     @objc func resetFrameRate(_ sender: UIGestureRecognizer) {
-        navigationMapView.mapView.preferredFPS = NavigationMapView.FrameIntervalOptions.defaultFramesPerSecond
+        mapView.update {
```
1ec5, line 408: Will it be a problem to make this call asynchronous? If I’m not mistaken, the purpose of this code is to ensure that pans and other gestures trigger a smooth frame rate to avoid a stuttering effect. Hopefully there’s no delay built into this call that would cause the gesture to stutter for a moment.

MaximAlien: Doesn't seem like there is no issue in this particular case.

Package.swift
MaximAlien: Updated to `main` to get more recent changes.

Sources/MapboxCoreNavigation/PredictiveCacheManager.swift
```diff
@@ -44,7 +44,10 @@ public class PredictiveCacheManager {
                                                  maxConcurrentRequests: UInt32 = 2) -> PredictiveCacheController? {
         let cacheOptions = PredictiveCacheControllerOptions(version: version,
                                                             dataset: dataset,
-                                                            concurrency: maxConcurrentRequests)
+                                                            // TODO: Make sure to use correct domain.
+                                                            dataDomain: .navigation,
```
MaximAlien, line 48: @Udumft, there are two domains: `maps` and `navigation`, it'd be good if you can check which one is correct.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
```diff
@@ -83,7 +83,8 @@ extension ElectronicHorizon.Edge {
             laneCount = native.laneCount as? UInt
             altitude = native.meanElevation as? Double
             curvature = UInt(native.curvature)
-            countryCode = native.countryCode
+            // TODO: Make sure to use correct country code.
+            countryCode = native.countryCodeIso3
```
1ec5, line 87: Apple platforms standardize on ISO 3166-1 alpha-2 codes, so ignore the alpha-3 code.

Sources/MapboxCoreNavigation/PredictiveCacheManager.swift
Udumft: `.maps` is the correct one for this case: `CacheOptions` are configured for handling Maps caches, while `Navigator`'s controller is initialized without additional options and automatically detects navigator settings.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
1ec5: MapboxNavigationNative doesn’t support alpha-2 codes as of v47.0.0, but it’ll come in the next release, so I made the documentation vague enough to allow for either for the time being. When we upgrade MapboxNavigationNative again, we’ll want to switch over to alpha-2 codes.

MapboxNavigation-Documentation.podspec
```diff
@@ -49,7 +49,7 @@ Pod::Spec.new do |s|
   s.dependency "MapboxGeocoder.swift", "~> 0.10.0"
   s.dependency "MapboxMaps", "10.0.0-beta.15"
```
1ec5, line 50: We’ll need to update the podspecs (and CocoaPods installation test fixture) once v10.0.0-beta.16 is released. Until then, the CocoaPods installation test will fail.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdgeMetadata.swift
1ec5: Never mind, I had a stale build. The alpha-2 code is available as well.

Sources/MapboxNavigation/RouteMapViewController.swift
```diff
@@ -155,9 +155,8 @@ class RouteMapViewController: UIViewController {
         self.init()
         self.navigationService = navigationService
         self.delegate = delegate
-        automaticallyAdjustsScrollViewInsets = false
```
1ec5, line 158: Should we ticket out any tail work to ensure that the map view’s camera insets respect the safe area insets?

1ec5 (approved):
The changelog also needs to be updated:

https://github.com/mapbox/mapbox-navigation-ios/blob/46aa80acd59742cf59224c40db54b4f4457db893/CHANGELOG.md#L9

Otherwise I think we’re good to go.

Sources/MapboxNavigation/RouteMapViewController.swift
MaximAlien: This part was removed because since iOS 11 `automaticallyAdjustsScrollViewInsets` was deprecated. I'll check this in scope of navigation camera viewport API work.

PR #2885, opened 2021-03-24
Add a will refresh event

Woildan:
### Description

Hello,

My team needed to be notified before a route refresh occurs in our project so I added a willRefresh event to the navigation service & router protocol.

As this may be useful to more people here is a PR. I'll let you judge if this can be added to the project or not.

I've also included a missing default implementation for the "didRefresh" event to the router delegate.

### Implementation

I've followed the general pattern used in the SDK:
- the "willRefresh" event is triggered before the actual request is initiated (follows "willReroute" behavior)
- a notification is posted at the same time the delegate is called
- code documentation follows general wording used throughout the project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v1.4.0
 
 * Fixed an issue when feedback UI was always appearing for short routes ([#2871](https://github.com/mapbox/mapbox-navigation-ios/pull/2871))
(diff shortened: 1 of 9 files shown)


Woildan:
Hello, any chance this MR will be included in the official codebase at some point or should I just close the request and maintain it in our fork?

S2Ler:
@Woildan Currently we are aiming to include this functionality in v2.1 release. Thanks for the contribution!

Issue #2841, opened 2021-03-01
MapView annotation selection doesn't work


bamx23:
**Mapbox Navigation SDK version:** 2.0.0-alpha.2

### Steps to reproduce

 1. Create NavigationMapView
 2. Set `navigationMapView.mapView.annotationManager.interactionDelegate = self`
 3. Add annotation `navigationMapView.mapView.annotationManager.addAnnotation(PointAnnotation(coordinate: coordinate))`
 4. Tap on the annotation

### Expected behavior

`func didSelectAnnotation(annotation: Annotation)` is called.

### Actual behavior

No delegate method is triggered.

### Suggested solution

I guess a bug is here:
https://github.com/mapbox/mapbox-navigation-ios/blob/bb4cfb6cea1391e88bf2ae6b79b7584fac2621b8/Sources/MapboxNavigation/NavigationMapView.swift#L283

`mapView.gestureRecognizers` should be used instead of `gestureRecognizers`(or a combination of both).

cc: @1ec5 @MaximAlien

bamx23:
Eventually, the suggested solution breaks route selection.

1ec5:
> Set `navigationMapView.mapView.annotationManager.interactionDelegate = self`

This seems like a responsibility of the map SDK, not the navigation SDK. It does seem like there should be only one interaction delegate in the application overall – should it be the responsibility of the application?

This issue also affects the DocsCode target in navigation-ios-examples, which is used for the “First steps” guide.

MaximAlien:
Route line selection is a regression on Navigation SDK side. @jill-cardamon, can you please check what might be wrong here?

jill-cardamon:
Yep. Already working on it.

MaximAlien:
This is not Navigation SDK issue. Ticket was created on Maps SDK side.

jill-cardamon:
Fixed in [#2909](https://github.com/mapbox/mapbox-navigation-ios/pull/2909).

PR #2901, opened 2021-03-30
Fix MapView annotation selection not working

jill-cardamon:
### Description

This PR fixes an issue where mapView annotation selection wasn't working. #2841

### Implementation

`setupGestureRecognizers()` now only requires failures of `mapView`'s `UITapGestureRecognizer` instead of all gesture recognizers.

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -297,9 +297,10 @@ open class NavigationMapView: UIView {
     }
     
     func setupGestureRecognizers() {
-        let gestures = gestureRecognizers ?? []
-        let mapTapGesture = UITapGestureRecognizer(target: self, action: #selector(didRecieveTap(sender:)))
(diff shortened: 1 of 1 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -297,9 +297,10 @@ open class NavigationMapView: UIView {
     }
     
     func setupGestureRecognizers() {
-        let gestures = gestureRecognizers ?? []
-        let mapTapGesture = UITapGestureRecognizer(target: self, action: #selector(didRecieveTap(sender:)))
-        mapTapGesture.requireFailure(of: gestures)
+        let mapTapGesture = UITapGestureRecognizer(target: self, action: #selector(didReceiveTap(sender:)))
+        for recognizer in mapView.gestureRecognizers! where recognizer is UITapGestureRecognizer {
```
MaximAlien, line 301: It'd be safer not to use force unwrapping here.

MaximAlien (approved):

PR #2902, opened 2021-03-30
RouteMapViewController refactor

### LINKED ISSUE #1502: Combine NavigationViewController and RouteMapViewController

Udumft:
### Description
Resolves #1502 
This PR divides RouteMapViewController` functionality into several objects which are used directly by `NavigationViewController`.

### Implementation
Communication between `NavigationViewController` and new components is done through few new protocols like `NavigationViewData` (to provide required resources and access to entities) and `ViewCycleObserver` (to handle ViewController's view events). Both protocols do not cover all interaction between classes, but I still wanted to decouple them as much as possible and potentially allow extending. So, any ideas on that matter are highly appreciated.
I also wanted to reuse new components in `CarPlayNavigationViewController` but unfortunately it's view structure and logic is a bit different.


Sources/MapboxNavigation/NavigationViewData.swift
```diff
@@ -0,0 +1,28 @@
...
+}
+
+
+protocol ViewCycleObserver {
+    func onViewDidLoad(_: UIView)
+    func onViewWillAppear(_: Bool)
+    func onViewDidAppear(_: Bool)
+    func onViewWillDisappear(_: Bool)
+    func onViewDidDisappear(_: Bool)
+    func onDidLayoutSubviews()
```
1ec5, line 18: The “on-” naming convention is atypical of Swift code. Instead, it should be sufficient to name the methods, e.g., `viewDidLoad`. If the problem is a conflict with the usual UIView methods, then we should make the method names more specific.

Overall, it seems like ViewCycleObserver avoids creating a separate protocol for each view for which another class might want to track. Separate protocols would allow a single controller class to track the lifecycle of multiple views without resorting to a series of `if let`s. But if we stick to the ViewCycleObserver approach, then let’s call it a delegate protocol. It looks like each of the classes that currently conform to ViewCycleObserver is a NavigationComponent, so what if we formalize that relationship by calling it NavigationComponentDelegate?

Sources/MapboxNavigation/NavigationViewHelper.swift
```diff
@@ -0,0 +1,170 @@
+
+import MapboxDirections
+import MapboxCoreNavigation
+import CoreLocation
+import UIKit
+
+
+class NavigationViewHelper: NavigationComponent, ViewCycleObserver {
```
1ec5, line 8: With this name, NavigationViewHelper is likely to become a bit of a junk drawer in the future. Let’s see if we can find more appropriate homes for each of its members.

If we whittle this class down to just the methods involved with runtime styling, then it would be more reusable like CameraController and we can potentially make it public. That was the original intent behind RouteMapViewController, before it gained a lot of other features that weren’t as closely tied to the map view.

Sources/MapboxNavigation/NavigationViewHelper.swift
```diff
@@ -0,0 +1,170 @@
...
+    
+    private var currentLegIndexMapped = 0
+    private var currentStepIndexMapped = 0
+    
+    // MARK: - Public methods
+    
+    func populateName(for waypoint: Waypoint, populated: @escaping (Waypoint) -> Void) {
+        guard waypoint.name == nil else { return populated(waypoint) }
+        let location = CLLocation(latitude: waypoint.coordinate.latitude, longitude: waypoint.coordinate.longitude)
+        CLGeocoder().reverseGeocodeLocation(location) { (places, error) in
```
1ec5, line 40: It somehow escaped my attention that we’re still using Core Location to geocode the waypoint name when no name has been set by the developer or filled in by the Directions API. I’m not entirely sure we actually hit this code, since for better or worse, the Directions API normally fills in the street name as a fallback waypoint name. Even if it doesn’t, it should be fine to leave the name unset; the only consequence is that EndOfRouteViewController’s not-too-prominent waypoint name label would wind up empty.

1ec5: And even if we keep this method around, it has no dependency on the surrounding class, so it would be better to put it in an extension of the Waypoint class.

Sources/MapboxNavigation/NavigationViewHelper.swift
```diff
@@ -0,0 +1,170 @@
...
+        for gestureRecognizer in navigationMapView.mapView.gestureRecognizers ?? [] {
+            gestureRecognizer.addTarget(self, action: #selector(resetFrameRate(_:)))
+        }
+    }
+    
+    @objc func resetFrameRate(_ sender: UIGestureRecognizer) {
+        navigationMapView.mapView.update {
+            $0.render.preferredFramesPerSecond = NavigationMapView.FrameIntervalOptions.defaultFramesPerSecond
+        }
+    }
```
1ec5, line 78: Frame rate management appears to be entirely dependent on NavigationMapView; can we move this functionality to that class?

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -314,49 +367,47 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
+        cameraController = CameraController()
+        overlayController = OverlayController()
         
-        embed(mapViewController, in: view) { (parent, map) -> [NSLayoutConstraint] in
-            return map.view.constraintsForPinning(to: parent.view)
-        }
+        arrivalController?.navigationViewData = self
+        navigationViewHelper?.navigationViewData = self
+        cameraController?.navigationViewData = self
+        overlayController?.navigationViewData = self
```
1ec5, line 399: I think we should minimize usage of backpointers if possible: classes that contain backpointers either have to account for the possibility of them going away at any time, mucking up the code, or use implicitly unwrapped optionals, which can be risky. A delegate protocol – or in this case, a data source protocol – would similarly require optional method calls, but at least the communication between the delegator and delegate is very explicit and isn’t as sensitive to object lifecycles, which makes further refactoring more straightforward.

Sources/MapboxNavigation/OverlayController.swift
```diff
@@ -0,0 +1,390 @@
+
+import UIKit
+import MapboxDirections
+import MapboxCoreNavigation
+import MapboxMaps
+import Turf
+import CoreLocation
+
+
+class OverlayController: NavigationComponent, ViewCycleObserver {
```
1ec5, line 10: The term “overlay” often refers to a kind of layer within the map view. The map SDK and this library tend to use the term “ornament” instead.

Sources/MapboxNavigation/OverlayController.swift
```diff
@@ -0,0 +1,390 @@
...
+        child.willMove(toParent: navigationViewData.navigationViewController)
+        navigationViewData.navigationViewController.addChild(child)
+        container.addSubview(child.view)
+        if let childConstraints: [NSLayoutConstraint] = constraints?(navigationViewData.navigationViewController, child) {
+            navigationViewData.navigationViewController.view.addConstraints(childConstraints)
+        }
+        child.didMove(toParent: navigationViewData.navigationViewController)
+    }
+    
+    private func labelCurrentRoadFeature(at location: CLLocation) {
```
1ec5, line 133: #1440 will probably end up slimming down this method considerably.

Sources/MapboxNavigation/OverlayController.swift
```diff
@@ -0,0 +1,390 @@
...
+    @objc func toggleMute(_ sender: UIButton) {
+        sender.isSelected = !sender.isSelected
+        
+        let muted = sender.isSelected
+        NavigationSettings.shared.voiceMuted = muted
+    }
+    
+    @objc func feedback(_ sender: Any) {
+        showFeedback()
+    }
```
1ec5, line 71: I guess these methods are only here because they also pertain to ornaments. I’m not sure they’d provide much value as part of a public OverlayController class, so it may not be worth the extra overhead to keep them separate from NavigationViewController.

Sources/MapboxNavigation/OverlayController.swift
```diff
@@ -0,0 +1,390 @@
...
+                } else {
+                    self.navigationView.wayNameView.isHidden = true
+                }
+            case .failure:
+                NSLog("Failed to find visible features.")
+            }
+        })
+    }
+    
+    private func roadFeature(for line: Feature) -> (roadName: String?, shieldName: NSAttributedString?) {
```
1ec5, line 261: This attributed string building logic is specific to the layout needed by WayNameView, so it should live in that class.

Sources/MapboxNavigation/OverlayController.swift
```diff
@@ -0,0 +1,390 @@
...
+               configurationURL.scheme == "mapbox",
+               let tileSetIdentifiers = configurationURL.host?.components(separatedBy: ",") {
+                return tileSetIdentifiers
+            }
+        } catch {
+            NSLog("Failed to get source properties with error: \(error.localizedDescription).")
+        }
+        
+        return []
+    }
```
1ec5, line 319: These methods have no dependencies on the surrounding class, so they should live in NavigationMapView or an extension of MapView.

Sources/MapboxNavigation/OverlayController.swift
```diff
@@ -0,0 +1,390 @@
...
+    
+    private func roadShieldTextColor(line: Feature) -> UIColor? {
```
1ec5, line 321: This method should also be part of WayNameLabel.

Sources/MapboxNavigation/OverlayController.swift
```diff
@@ -0,0 +1,390 @@
...
+        case "yellow":
+            return .yellow
+        case "orange":
+            return .orange
+        default:
+            return .black
+        }
+    }
+    
+    private func roadShieldAttributedText(for text: String, textColor: UIColor, imageName: String) -> NSAttributedString? {
```
1ec5, line 348: This method should live on NavigationMapView.

Sources/MapboxNavigation/OverlayController.swift
```diff
@@ -0,0 +1,390 @@
...
+                                                                                        scale: UIScreen.main.scale)
+        return NSAttributedString(attachment: attachment)
+    }
+    
+    /**
+     Method, which returns a boolean value indicating whether the tile source is a supported version of the Mapbox Streets source.
+     */
+    private func isMapboxStreets(_ identifiers: [String]) -> Bool {
+        return identifiers.contains("mapbox.mapbox-streets-v8") || identifiers.contains("mapbox.mapbox-streets-v7")
+    }
```
1ec5, line 363: This method should be a static method of an extension of VectorSource. It was originally on an extension of MGLVectorTileSource but got inlined in RouteMapViewController when upgrading to map SDK v10.

Sources/MapboxNavigation/CameraController.swift
```diff
@@ -0,0 +1,302 @@
+import MapboxDirections
+import MapboxCoreNavigation
+import MapboxMaps
+
+
+class CameraController: NavigationComponent, ViewCycleObserver {
```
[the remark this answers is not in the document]
MaximAlien: IMO current PR should be rebased onto `maxim/camera-viewport-api` so that we start using `NavigationCamera` here as well. At this point `CameraController` uses properties, which are no longer present in mentioned PR (e.g. `tiltedCamera`, `contentInsets` are now prepared inside `NavigationViewportDataSource`, transitions are executed using `CameraAnimator` instead of `setCamera` calls).

`maxim/camera-viewport-api` also modifies `RouteMapViewController` a bit changes will be needed again to accomodate those viewport camera changes. @Udumft, let me know if you need more clarifications there.

CHANGELOG.md
```diff
@@ -50,7 +50,8 @@
...
-* Added the `Directions.calculateOffline(options:completionHandler:)` and `Directions.calculateWithCache(options:completionHandler:)` methods, which incorporate routing tiles from the predictive cache when possible to avoid relying on a network connection to calculate the route. `RouteController` now also uses the predictive cache when rerouting. ([#2848](https://github.com/mapbox/mapbox-navigation-ios/pull/2848)) 
+* Added the `Directions.calculateOffline(options:completionHandler:)` and `Directions.calculateWithCache(options:completionHandler:)` methods, which incorporate routing tiles from the predictive cache when possible to avoid relying on a network connection to calculate the route. `RouteController` now also uses the predictive cache when rerouting. ([#2848](https://github.com/mapbox/mapbox-navigation-ios/pull/2848))
+* Decomposed `RouteMapViewController`'s functionality into several components: `NavigationViewHelper`, `CameraController`, `ArrivalController` and `OverlayController` which are handled by `NavigationViewController` ([#2902](https://github.com/mapbox/mapbox-navigation-ios/pull/2902))
```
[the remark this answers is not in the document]
Udumft: yes, these are all internal, but I thought it was worth mentioning for devs who interact and possibly customize it.

Sources/MapboxNavigation/ArrivalController.swift
```diff
@@ -0,0 +1,171 @@
+
+import UIKit
+import MapboxDirections
+import MapboxCoreNavigation
+import MapboxMobileEvents
+
+
+class ArrivalController {
```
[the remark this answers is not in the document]
Udumft: I've documented  new components. Right now this class is just about `EndOfRouteViewController`. If in future it changes, I think it would be easy to update `ArrivalController` too.

Sources/MapboxNavigation/NavigationViewData.swift
Udumft: I can hardly imagine a case where one controller will have to maintain multiple views, so I've decided to stick to this delegate protocol solution.

Sources/MapboxNavigation/NavigationViewHelper.swift
Udumft: In result, it turns out that it only handles events related to Route Line drawing and it's related features, so I've documented and renamed it

Sources/MapboxNavigation/NavigationViewHelper.swift
Udumft: I like the idea of removing this one :)

Sources/MapboxNavigation/NavigationViewHelper.swift
Udumft: Done

Sources/MapboxNavigation/NavigationViewController.swift
Udumft: This is the best way of communication I ended up with. What do you think could work better?

Sources/MapboxNavigation/OverlayController.swift
Udumft: Done

Sources/MapboxNavigation/OverlayController.swift
Udumft: I think it does it's job to lighten the `NavigationViewController` for now. If we decide to expose `OrnamentsController` publicly, we'll need to do more tweaks and moving these methods should not become a problem if it will be needed

Sources/MapboxNavigation/OverlayController.swift
Udumft: Done

Sources/MapboxNavigation/OverlayController.swift
Udumft: Moved to `MapView` extension as there are already few similar items exist.

Sources/MapboxNavigation/OverlayController.swift
Udumft: Done

Sources/MapboxNavigation/OverlayController.swift
Udumft: Having this as part of `NavigationMapView` is not straightforward. This method is used only as a part `WayNameView` setup, and since it's setup is quite specific, having it in `NavigationMapView` seems incorrect. It is a part of a custom setup method, which may not be needed anywhere else, and also doing such would make `WayNameView` dependent on `NavigationMapView`. Instead I've decided to move it to `WayNameView` extension together with other setup methods

Sources/MapboxNavigation/OverlayController.swift
Udumft: Done

Sources/MapboxNavigation/CameraController.swift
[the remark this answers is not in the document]
Udumft: That makes sense. I'll try rebasing and see how it goes

Sources/MapboxNavigation/ArrivalController.swift
```diff
@@ -0,0 +1,172 @@
...
+import MapboxDirections
+import MapboxCoreNavigation
+import MapboxMobileEvents
+
+/// A component to encapsulate `EndOfRouteViewController` presenting logic such as enabling/disabling, handling autolayout, keyboard, positioning camera, etc.
+class ArrivalController: NavigationComponentDelegate {
+    
+    // MARK: - Properties
+    
+    weak private(set) var navigationViewData: NavigationViewData!
```
1ec5, line 12: I’m still concerned about the use of this backpointer to NavigationViewController etc. Can we add methods to NavigationComponentDelegate that return whatever ArrivalController et al. need from those classes?

1ec5 (commented):
The main thing left in my opinion is to eliminate the backpointer usage, if possible, but otherwise this refactor would be good to go. I can change this to an approval if there’s really no way to remove the backpointer, but let’s give it serious consideration first.

Sources/MapboxNavigation/RouteLineController.swift
```diff
@@ -0,0 +1,147 @@
+
+import MapboxDirections
+import MapboxCoreNavigation
+import CoreLocation
+import UIKit
+
+/// A component meant to assist displaying route line and related items like arrows, waypoints, annotations and other.
+class RouteLineController: NavigationComponent, NavigationComponentDelegate {
```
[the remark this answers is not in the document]
Udumft: This class handles items which are related to the route line. IMO it's name describes it's purpose, but adding `Overlay` suffix may sound like it is responsible for something not related to the route line.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -0,0 +1,334 @@
+
+import UIKit
+import MapboxDirections
+import MapboxCoreNavigation
+import MapboxMaps
+import Turf
+import CoreLocation
+
+/// A components, designed to help manage `NavigationMapView` ornaments logic.
+class OrnamentsController: NavigationComponent, NavigationComponentDelegate {
```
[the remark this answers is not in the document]
Udumft: It would be problematic to nest it inside `NavigationMapView` since this class also closely operates with `NavigationView` members.

Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -0,0 +1,77 @@
+import Foundation
+import UIKit
+import Turf
+import MapboxMaps
+
+
+extension WayNameView {
```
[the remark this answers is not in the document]
Udumft: Done

Sources/MapboxNavigation/ArrivalController.swift
Udumft: I tried to minimize dependencies through the protocol as much as I could. Currently it boiled down to using `NavigationView` and `Router`. It would be hard (or even impossible?) to get further than that since `controllers` logic is meant to interact with `NavigationView` and ease the burden for `NavigationViewController` by that.

Sources/MapboxNavigation/ArrivalController.swift
```diff
@@ -7,6 +7,8 @@ import MapboxMobileEvents
 /// A component to encapsulate `EndOfRouteViewController` presenting logic such as enabling/disabling, handling autolayout, keyboard, positioning camera, etc.
 class ArrivalController: NavigationComponentDelegate {
     
+    typealias EndOfRouteDismissHandler = (EndOfRouteFeedback?)->()
```
1ec5, line 10: Nit: Rename to EndOfRouteDismissalHandler.

Sources/MapboxNavigation/ArrivalController.swift
1ec5: I’m pretty sure we’ll need to rethink this approach when we make ArrivalController or any of the other new classes public for use in custom UIs apart from NavigationViewController, since NavigationView is private too. It gets back to the original reason for wanting to get rid of RouteMapViewController, that it encapsulated too much behavior that was only usable as part of NavigationViewController.

Sources/MapboxNavigation/OrnamentsController.swift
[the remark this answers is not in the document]
1ec5: It would just be syntactic sugar: the type would be nested, but that doesn’t change the relationship between NavigationMapView and this class in terms of object pointers. We’re doing something similar with electronic horizon types.

Regardless, I think “map view” should be somewhere in the type name, since there are other kinds of ornaments, and “ornament” should be singular.

```swift
extension NavigationMapView {
    class OrnamentController: NavigationComponent, NavigationComponentDelegate {
        // All this remains unchanged.
    }
}
```

Sources/MapboxNavigation/RouteLineController.swift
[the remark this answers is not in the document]
1ec5: “Overlay” is a term used in runtime styling, similar to “annotations”, but I understand your concern about making it sound too generic. How about `NavigationMapView.RouteOverlayController`?

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -163,23 +160,27 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
         }
     }
 
     /**
      Controls whether or not the FeedbackViewController shows a second level of detail for feedback items.
      */
-    public var detailedFeedbackEnabled: Bool = false {
-        didSet {
-            mapViewController?.detailedFeedbackEnabled = detailedFeedbackEnabled
+    public var detailedFeedbackEnabled: Bool {
```
1ec5, line 175: Since the default value is no longer part of the public type declaration, mention the default value in the documentation comment (also for `annotatesSpokenInstructions`).

1ec5 (approved):
This PR accomplishes the removal of RouteMapViewController in a literal sense, but I still have reservations about how the new classes are just as tightly coupled to NavigationViewController.

Rather than hold up this major refactoring any longer, I’d suggest opening tail work tickets about eliminating the dependence on backpointers in NavigationViewData, with an eye toward eventually making these new classes resuable in custom UIs and CarPlay. (That reuse won’t be possible quite yet because NavigationView is also private, but at least we should prepare for the possibility.)

The suggested renaming and type nesting can also be tail work, as long as we make a decision on whether to go ahead with those changes by the release candidate.

Udumft:
Tailwork captured in #2953

PR #2909, opened 2021-04-02
Fix route selection regression and annotation selection

jill-cardamon:
### Description

This PR is to address a regression introduced in [#2901](https://github.com/mapbox/mapbox-navigation-ios/pull/2901) as well as fix MapView annotation selection.

### Implementation

Added a target to the existing tap gesture recognizer to address route and waypoint selection. Maps team also investigated the gesture recognizer conflict and found that the logic that causes MapView's single tap gesture recognizer to fail is not implemented yet. Tested on both the Nav SDK side and in 1Tap.

cc/ @bamx23

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -309,12 +309,9 @@ open class NavigationMapView: UIView {
     }
     
     func setupGestureRecognizers() {
-        let gestures = mapView.gestureRecognizers ?? []
-        let mapTapGesture = UITapGestureRecognizer(target: self, action: #selector(didReceiveTap(sender:)))
-        for recognizer in gestures where recognizer is UITapGestureRecognizer {
-            mapTapGesture.requireFailure(of: [recognizer])
+        for recognizer in mapView.gestureRecognizers ?? [] where recognizer is UITapGestureRecognizer {
+            recognizer.addTarget(self, action: #selector(didReceiveTap(sender:)))
         }
-        mapView.addGestureRecognizer(mapTapGesture)
     }
     
     /**
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -309,8 +309,9 @@ open class NavigationMapView: UIView {
     }
     
     func setupGestureRecognizers() {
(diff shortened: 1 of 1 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -297,10 +297,9 @@ open class NavigationMapView: UIView {
     }
     
     func setupGestureRecognizers() {
-        let gestures = gestureRecognizers ?? []
-        let mapTapGesture = UITapGestureRecognizer(target: self, action: #selector(didRecieveTap(sender:)))
-        mapTapGesture.requireFailure(of: gestures)
-        mapView.addGestureRecognizer(mapTapGesture)
+        for recognizer in mapView.gestureRecognizers ?? [] where recognizer is UITapGestureRecognizer {
```
bamx23, line 300: Do we need to add a target to all of the `UITapGestureRecognizer` instances?

jill-cardamon: We technically don't, but the gesture recognizers from the Maps SDK aren't named, so there's no way to filter through them effectively. The only way would be through target actions but those are private instance variables. That approach would either be fragile or rejected by Apple's review process. Trying to add to a certain gesture recognizer by order (ex// trying to add target to the last gesture recognizer in the `mapView.gestureRecognizer` array) also isn't very reliable. This fix is potentially temporary for now anyway until the Maps team implements logic on their side.

MaximAlien: @jill-cardamon, can you please create a follow-up ticket for this? If this is temporary workaround we should revise current implementation. If I understand this correctly idea would be to register only one gesture recognizer on Navigation SDK side, which will allow to select routes and waypoints. Such gesture recognizer is supposed to fail when such mechanism is available on Maps SDK side, right? 

Please also squash current PR into one commit, we do not want duplication in history based on previous PRs.

MaximAlien (approved):

Sources/MapboxNavigation/NavigationMapView.swift
MaximAlien: It'd be also good to make sure that route selection logic works on Navigation SDK `1.x` and when custom gestures are added on client side.

PR #2911, opened 2021-04-05
Allow MapboxCoreNavigationTests  running under Swift Package Manager

### LINKED ISSUE #2792: Mocked tests don’t run under Swift Package Manager
#2771 excluded a number of Core Navigation integration tests when implementing Swift Package Manager support. These tests depend on TestHelper, primarily to load test fixtures. We’ll probably have to break up TestHelper into at least two targets: it includes Objective-C headers and also includes some mocks that depend on MapboxNavigation, which does not yet support SPM.

Here are the affected test suites in MapboxCoreNavigationTests:

* LocationTests
* MapboxCoreNavigationTests
* MapboxNavigationServiceSpec
* NavigationEventsManagerTests
* NavigationServiceTests
* OfflineRoutingTests
* PassiveLocationDataSourceTests
* RouteProgressTests (partial)
* TunnelAuthorityTests

There will be other affected test suites in MapboxNavigationTests once MapboxNavigation supports SPM. Some of these test suites may need to be rewritten first due to dependencies on testing frameworks that don’t support SPM: #2791. The gap in test coverage is similar to mapbox/mapbox-directions-swift#494, but this repository’s tests don’t depend on OHHTTPStubs and don’t need to support watchOS.

/cc @mapbox/navigation-ios

ShanMa1991:
### Description
This pr is to fix #2792 to allow Core Navigation integration tests run after implementing Swift Package Manager support.

- [x] create `CTestHelper` as a target in Package.swift, which contains MMEEventsManager+Spy.h/.m 
- [x] add `TestHelper` as target in Package.swift, dependent on `CTestHelper`, move it from Tests folder to Sources folder.
- [x] move `LeakTest.swift` from `MapboxNavigationTests` to `TestHelper`.
- [x] move `CLHeadingPrivate.h` from `MapboxCoreNavigationTests` to `CTestHelper`. 
- [x] remove redundant dependencies from `MapboxCoreNavigationTests`, add resources in `TestHelper`.
- [x] use `NSUserDefaults` to avoid SPM generated Info.plist, which is lacking certain keys including `NSLocationWhenInUseUsageDescription`, `NSLocationAlwaysAndWhenInUseUsageDescription` and `CFBundleShortVersionString`


1ec5:
@ShanMa1991 and I are puzzling over [these errors](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/3016/workflows/b7e2f7e2-4f82-4581-8c9c-18655843de73/jobs/29615) in the SPM build:

```
error: Unable to resolve build file: XCBCore.BuildFile (The workspace has a reference to a missing target with GUID 'PACKAGE-TARGET:CTestHelper') (in target 'MapboxCoreNavigationTests' from project 'MapboxNavigation')
error: Unable to resolve build file: XCBCore.BuildFile (The workspace has a reference to a missing target with GUID 'PACKAGE-TARGET:TestHelper') (in target 'MapboxCoreNavigationTests' from project 'MapboxNavigation')
…
error: Unable to resolve build file: XCBCore.BuildFile (The workspace has a reference to a missing target with GUID 'PACKAGE-TARGET:CTestHelper') (in target 'TestHelper' from project 'MapboxNavigation')
```

ShanMa1991:
The  [core test failure under Carthage](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/3016/workflows/b7e2f7e2-4f82-4581-8c9c-18655843de73/jobs/29617) will be handled in a separate ticket #2913 .

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -95,7 +95,13 @@ struct NavigationEventDetails: EventDetails {
     let simulation: Bool
     let startTimestamp: Date?
     let sdkIdentifier: String
-    let sdkVersion: String = String(describing: Bundle.mapboxCoreNavigation.object(forInfoDictionaryKey: "CFBundleShortVersionString")!)
+    var sdkVersion: String {
+        if let stringForShortVersion = Bundle.mapboxCoreNavigation.object(forInfoDictionaryKey: "CFBundleShortVersionString") {
+            return String(describing: stringForShortVersion)
+        } else {
+            return String(describing: UserDefaults.standard.object(forKey: "CFBundleShortVersionString"))
```
1ec5, line 102: If the Core Navigation bundle can’t be found and the user default isn’t set, wrapping the `nil` user default in a string would result in the string “`nil`”, which would be unexpected. The previous code would’ve crashed in that circumstance, so it would be reasonable to issue a `preconditionFailure(_:)` instead.

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -95,7 +95,13 @@ struct NavigationEventDetails: EventDetails {
     let simulation: Bool
     let startTimestamp: Date?
     let sdkIdentifier: String
-    let sdkVersion: String = String(describing: Bundle.mapboxCoreNavigation.object(forInfoDictionaryKey: "CFBundleShortVersionString")!)
+    var sdkVersion: String {
+        if let stringForShortVersion = Bundle.mapboxCoreNavigation.object(forInfoDictionaryKey: "CFBundleShortVersionString") {
```
1ec5, line 99: I think we need to make `Bundle.mapboxCoreNavigation` work under SPM in general and not just for tests: #2922. Depending on the fix for that issue, the user default fallback may or may not be necessary.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -59,10 +58,6 @@ class NavigationServiceTests: XCTestCase {
         try? Navigator.shared.navigator.resetRideSession()
     }
 
-    func testDefaultUserInterfaceUsage() {
-        XCTAssertTrue(!dependencies.navigationService.eventsManager.usesDefaultUserInterface, "MapboxCoreNavigationTests shouldn't have an implicit dependency on MapboxNavigation due to removing the Example application target as the test host.")
```
1ec5, line 63: There should still be a test that `usesDefaultUserInterface` is false when the developer installs Core Navigation alone.

Sources/MapboxCoreNavigation/EventDetails.swift
ShanMa1991: Thanks for the suggestion, I'll add the check of this one.

Sources/MapboxCoreNavigation/EventDetails.swift
ShanMa1991: Yes. Because now the UserDefaults only added for test incase of the SPM generated Info.plist lacking certain keys.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
ShanMa1991: Now the `TestHelper` has the dependency of both `MapboxNavigation` and `MapboxCoreNavigation`. when it's running under SPM, it would be true. When it's under Carthage, because we removed the Example app as the host, it would be false. After the fix of #2922 , it may be fixed.

Sources/MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -27,18 +27,33 @@ extension Bundle {
...
     
     /**
      The Mapbox Navigation framework bundle, if installed.
      */
     class var mapboxNavigationIfInstalled: Bundle? {
-        // Assumption: MapboxNavigation.framework includes NavigationViewController and exposes it to the Objective-C runtime as MapboxNavigation.NavigationViewController.
-        guard let NavigationViewController = NSClassFromString("MapboxNavigation.NavigationViewController") else {
+        get {
+            #if SWIFT_PACKAGE
+            for bundleIdentifier in Bundle.allBundles.compactMap({ $0.bundleIdentifier }) {
```
1ec5, line 45: 👍 This is a much more elegant approach than the Objective-C runtime tricks I was thinking of. It happens to work when SPM is building static frameworks, because the bundle we’re looking for would be a resource bundle rather than a framework bundle.

Sources/MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -47,6 +62,50 @@ extension Bundle {
...
+    }()
+    
+    /**
+     Returns the value associated with the specific key in the Mapbox Navigation bundle's  information property list, if installed.
+     */
+    public class func object(forMapboxNavigationInfoDictionaryKey key: String) -> String? {
+        if let stringForKey = Bundle.mapboxNavigationIfInstalled?.object(forInfoDictionaryKey: key) {
+            return stringForKey as? String
+        } else if let infoDictionary = Bundle.mapboxNavigationInfoDictionary {
+            return infoDictionary[key] as? String
```
1ec5, line 90: Unlike `Bundle.object(forInfoDictionaryKey:)`, this method is casting the return value as a string, preventing it from returning, say, a number. This is good enough for what we’re using the method for, but in that case we should rename the method to `string(forMapboxNavigationInfoDictionaryKey:)`.

1ec5 (approved):

Sources/MapboxCoreNavigation/BundleAdditions.swift
ShanMa1991: Thanks for the suggestion! I'll update the name of the method.

Issue #2914, opened 2021-04-07
RouteMapViewController is never deallocated.


MaximAlien:
**Mapbox Navigation SDK version:** `release-v2.0`

### Steps to reproduce

 1. Open example application and start active guidance navigation.
 2. Stop active guidance navigation and close `NavigationViewController`.
 3. Repeat similar actions several times.

### Expected behavior

`RouteMapViewController` and dependent objects should be deallocated.

### Actual behavior

`RouteMapViewController`, which is embedded into `NavigationViewController` and all child views like `NavigationMapView` are never deallocated. For example:

<img width="1680" alt="Screen Shot 2021-04-06 at 6 17 42 PM" src="https://user-images.githubusercontent.com/1496498/113932038-06d73880-97a8-11eb-9983-f424b2b2bbe0.png">

/cc @mapbox/navigation-ios

MaximAlien:
It's interesting that Profiler couldn't detect this leak.

PR #2915, opened 2021-04-07
Fix RouteMapViewController deallocation issue.

MaximAlien:
### Description

Closing #2914.

Problem was due to strong reference usage when accessing `self` in closure.

### DIFF
--- a/Sources/MapboxNavigation/RouteMapViewController.swift
+++ b/Sources/MapboxNavigation/RouteMapViewController.swift
@@ -182,17 +182,13 @@ class RouteMapViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        
-        let navigationMapView = self.navigationMapView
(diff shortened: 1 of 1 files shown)


1ec5 (approved):

MaximAlien:
After fix:

Screen Shot 2021-04-07 at 5 11 15 PM

Sources/MapboxNavigation/RouteMapViewController.swift
```diff
@@ -182,17 +182,13 @@ class RouteMapViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        
-        let navigationMapView = self.navigationMapView
-        // TODO: Verify whether content insets should be changed on map view.
-        view.layoutIfNeeded()
 
-        navigationMapView.tracksUserCourse = true
+        self.navigationMapView.tracksUserCourse = true
```
azarovalex, line 186: Maybe without `self`?

PR #2919, opened 2021-04-09
Release v1.4

chezzdev:
Release 1.4 preparation.

- Updated translations,
- Bumped version,
- Edited minor things in the changelog.


.gitattributes
```diff
@@ -0,0 +1 @@
+*.strings diff
```
1ec5, line 1: Run scripts/convert_string_files.sh to convert the files back to plain text format from the binary format that `tx` insists on.

chezzdev: Thanks, it helped with build errors too.

1ec5 (approved):

PR #2920, opened 2021-04-09
v2.0.0-beta.4

1ec5:
Updated the Xcode project, CocoaPods podspecs, readme, and CocoaPods installation test fixture for v2.0.0-beta.4.

Upgraded to MapboxDirections v2.0.0-beta.2.

/cc @mapbox/navigation-ios


MaximAlien (approved):

Sources/MapboxNavigation/LanesView.swift
```diff
@@ -82,7 +82,7 @@ open class LanesView: UIView, NavigationComponent {
         }
         
         let subviews = tertiaryInstruction.components.compactMap { (component) -> LaneView? in
-            if case let .lane(indications: indications, isUsable: isUsable) = component {
+            if case let .lane(indications: indications, isUsable: isUsable, preferredDirection: _) = component {
                 let maneuverDirection = visualInstruction?.primaryInstruction.maneuverDirection ?? ManeuverDirection(rawValue: indications.description)
```
1ec5, line 86: We’ll make use of the new `preferredDirection` associated value when we implement #2801. It would be obvious to stick this value in front of the primary maneuver direction, but `preferredDirection` is a `LaneIndication`, which isn’t as straightforward to convert to a `ManeuverDirection`. mapbox/mapbox-directions-swift#534 would make it a `ManeuverDirection`, which will greatly simplify #2801.

Sources/MapboxNavigation/VisualInstruction.swift
```diff
@@ -51,7 +51,7 @@ extension VisualInstruction {
 
     func lanesImage(side: DrivingSide, direction: ManeuverDirection?, useableColor: UIColor, unuseableColor: UIColor, size: CGSize, scale: CGFloat) -> UIImage? {
         let subimages = components.compactMap { (component) -> UIImage? in
-            if case let .lane(indications: indications, isUsable: isUsable) = component {
+            if case let .lane(indications: indications, isUsable: isUsable, preferredDirection: _) = component {
                 return laneImage(side: side, indication: indications, maneuverDirection: direction, isUsable: isUsable, useableColor: useableColor, unuseableColor: unuseableColor, size: CGSize(width: size.height, height: size.height))
```
1ec5, line 55: Same as above, it’s tempting to use `preferredDirection` here but it’s the wrong type. (This method handles the lane image for CarPlay.)

PR #2924, opened 2021-04-12
Fix MapboxCoreNavigation bundle when building with SPM

### LINKED ISSUE #2922: Application bundle misidentified as MapboxCoreNavigation when building with SPM
`Bundle.mapboxCoreNavigation` and `Bundle.mapboxNavigationIfInstalled` return the host application bundle when MapboxCoreNavigation is built as an SPM target. These computed variables rely on the old canary class trick, which is unreliable when the target produces a static framework, as it does under SPM:

https://github.com/mapbox/mapbox-navigation-ios/blob/70020623e427a32239b344f48d0af6428fcbe565/Sources/MapboxCoreNavigation/BundleAdditions.swift#L26-L42

By contrast, `Bundle.mapboxNavigation` uses `Bundle.module` to reliably return the bundle associated with the MapboxNavigation target, which SPM synthesizes for this purpose:

https://github.com/mapbox/mapbox-navigation-ios/blob/70020623e427a32239b344f48d0af6428fcbe565/Sources/MapboxNavigation/Bundle.swift#L7-L15

The incorrect `Bundle.mapboxCoreNavigation` and `Bundle.mapboxNavigationIfInstalled` behavior causes several bugs:

* Events are misstating the application version as the SDK version.
* User agent strings for junction views and routing tiles state the name and version of the application where it should state that of MapboxCoreNavigation.
* At launch, NavigationService and LegacyRouteController check for updates based on the application version instead of the SDK version.

/cc @mapbox/navigation-ios

ShanMa1991:
### Description
This pr is to fix #2922 .

### Implementation
When running the app generated by SPM, the `Info.plist` is also default to the one generated by SPM, which is missing some certain keys, such as the `CFBundleShortVersionString` . This pr is to create a resource file `MBXInfo.plist` same as `Info.plist`, and copy it to the MapboxCoreNavigation. When the bundle couldn't find certain keys contained in `Info.plist` , it would check the `MBXInfo.plist`.

Update the `update-script.sh` to copy the `Info.plist` to `MBXInfo.plist`.

### DIFF
--- a/Sources/MapboxCoreNavigation/BundleAdditions.swift
+++ b/Sources/MapboxCoreNavigation/BundleAdditions.swift
@@ -54,6 +54,49 @@ extension Bundle {
     }
     
     /**
+     Returns a dictionary of `MBXInfo.plist` in Mapbox Core Navigation.
+     */
+    static var mapboxCoreNavigationMBX: Dictionary<String, Any>? {
+        guard let fileURL = Bundle.mapboxCoreNavigation.url(forResource: "MBXInfo", withExtension: "plist"),
+              let infoDictionary = NSDictionary(contentsOf: fileURL) as? Dictionary<String, Any> else { return nil }
+        return infoDictionary
+    }
+    
+    /**
+     Returns a dictionary of `MBXInfo.plist` in Mapbox Navigation framework bundle, if installed.
+     */
+    static var mapboxNavigationMBX: Dictionary<String, Any>? {
+        guard let fileURL = Bundle.mapboxNavigationIfInstalled?.url(forResource: "MBXInfo", withExtension: "plist"),
+              let infoDictionary = NSDictionary(contentsOf: fileURL) as? Dictionary<String, Any> else { return nil }
+        return infoDictionary
+    }
+    
+    /**
+     Returns the value associated with the specific key in the Mapbox Navigation bundle's  information property list, if installed.
+     */
+    public class func object(forMapboxNavigationInfoDictionaryKey key: String) -> String? {
+        if let stringForKey = Bundle.mapboxNavigationIfInstalled?.object(forInfoDictionaryKey: key) {
+            return stringForKey as? String
+        } else if let infoDictionary = Bundle.mapboxNavigationMBX {
+            return infoDictionary[key] as? String
+        } else {
+            return nil
+        }
+    }
+    /**
+     Returns the value associated with the specific key in the Mapbox Core Navigation bundle's  information property list.
+     */
+    public class func object(forMapboxCoreNavigationInfoDictionaryKey key: String) -> String? {
+        if let stringForKey = Bundle.mapboxCoreNavigation.object(forInfoDictionaryKey: key) {
+            return stringForKey as? String
+        } else if let infoDictionary = Bundle.mapboxCoreNavigationMBX {
+            return infoDictionary[key] as? String
+        } else {
+            return nil
+        }
+    }
+    
+    /**
      A file URL representing a directory in which the application can place downloaded tile files.
      */
     public var suggestedTileURL: URL? {
--- a/Sources/MapboxCoreNavigation/BundleAdditions.swift
+++ b/Sources/MapboxCoreNavigation/BundleAdditions.swift
@@ -40,11 +40,20 @@ extension Bundle {
      The Mapbox Navigation framework bundle, if installed.
      */
     class var mapboxNavigationIfInstalled: Bundle? {
-        // Assumption: MapboxNavigation.framework includes NavigationViewController and exposes it to the Objective-C runtime as MapboxNavigation.NavigationViewController.
-        guard let NavigationViewController = NSClassFromString("MapboxNavigation.NavigationViewController") else {
+        get {
+            #if SWIFT_PACKAGE
+            for bundleIdentifier in Bundle.allBundles.compactMap({ $0.bundleIdentifier }) {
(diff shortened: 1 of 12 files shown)


Sources/MapboxCoreNavigation/MBXInfo.plist
```diff
@@ -0,0 +1,24 @@
...
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>$(EXECUTABLE_NAME)</string>
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
```
1ec5, line 14: There are other places where we’re using `Bundle.mapboxCoreNavigation` and `Bundle.mapboxNavigationIfInstalled` to get Info.plist keys:

https://github.com/mapbox/mapbox-navigation-ios/blob/6cc9728a615ed9b08632a6c2d09f7fca57a5c542/Sources/MapboxCoreNavigation/URLSession.swift#L19

The Xcode build system normally resolves placeholder tokens in Info.plist in one of the last build phases, but it won’t process this property list the same way. As a result, the line above will literally put `$(PRODUCT_NAME)` in the user agent string instead of `MapboxCoreNavigation`, and instead of `MapboxNavigation` it’ll substitute the host application’s name.

Until SPM gains support for Apple Generic Versioning, update-version.sh will need to set `CFBundleName` to the expected name:

```bash
plutil -replace CFBundleName -string 'MapboxCoreNavigation' MBXInfo.plist
```

(likewise for Source/MapboxNavigation/MBXInfo.plist)

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -98,8 +98,12 @@ struct NavigationEventDetails: EventDetails {
     var sdkVersion: String {
         if let stringForShortVersion = Bundle.mapboxCoreNavigation.object(forInfoDictionaryKey: "CFBundleShortVersionString") {
             return String(describing: stringForShortVersion)
+        } else if let fileURL = Bundle.mapboxCoreNavigation.url(forResource: "MBXInfo", withExtension: "plist"),
+                  let infoDictionary = NSDictionary(contentsOf: fileURL) as? Dictionary<String, Any>,
```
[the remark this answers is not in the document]
ShanMa1991: Thanks for the suggestion! I'll add the two in the extension.

Sources/MapboxCoreNavigation/MBXInfo.plist
ShanMa1991: Thanks for the suggestion! I'll fix this part.

Sources/MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -27,7 +27,13 @@ extension Bundle {
      The Mapbox Core Navigation framework bundle.
      */
     public class var mapboxCoreNavigation: Bundle {
-        return Bundle(for: RouteController.self)
+        get {
+            #if SWIFT_PACKAGE
+            return .module
```
1ec5, line 32: Also use `Bundle.module` in the fallback at the bottom of `mapboxNavigationIfInstalled` below.

Sources/MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -47,6 +53,49 @@ extension Bundle {
...
     }
     
+    /**
+     Returns a dictionary of `MBXInfo.plist` in Mapbox Core Navigation.
+     */
+    static var mapboxCoreNavigationMBX: Dictionary<String, Any>? {
+        guard let fileURL = Bundle.mapboxCoreNavigation.url(forResource: "MBXInfo", withExtension: "plist"),
+              let infoDictionary = NSDictionary(contentsOf: fileURL) as? Dictionary<String, Any> else { return nil }
+        return infoDictionary
+    }
```
1ec5, line 63: This still fetches and parses the plist every time it’s called. This would make it compute only once, lazily on the first use:

```suggestion
    static let mapboxCoreNavigationInfoDictionary: [String: Any]? = {
        guard let fileURL = Bundle.mapboxCoreNavigation.url(forResource: "MBXInfo", withExtension: "plist"),
              let infoDictionary = NSDictionary(contentsOf: fileURL) as? Dictionary<String, Any> else { return nil }
        return infoDictionary
    }()
```

Sources/MapboxCoreNavigation/MBXInfo.plist
1ec5: Since we haven’t released yet, we’ll need to manually run the `plutil` commands to replace these tokens.

Sources/MapboxCoreNavigation/BundleAdditions.swift
ShanMa1991: Thanks for the suggestion, I'll fix this part.

ShanMa1991: Because the `NSBundle.value(forKey: "mapboxNavigation")` is not working, and the `Bundle.module` would return the current module, I choose to filter the `bundleIdentifier` for current all bundles and to get the `MapboxNavigation` bundle.

PR #2927, opened 2021-04-13
Add integration test for Swift Package Manager

### LINKED ISSUE #2903: Add `SwiftPackageManagerTest` by analogy with `CocoaPodsTest/PodInstall`
Currently, we have the [CocoaPodsTest](https://github.com/mapbox/mapbox-navigation-ios/tree/main/Tests/CocoaPodsTest/PodInstall), which verifies that a test project can be built with Navigation SDK installed via CocoaPods.
Swift Package Manager gradually becoming the preferable way to install Navigation SDK, so I propose to add a new SwiftPackageManagerTest that will try to build a test project that has our Navigation SDK installed via SPM, similar to the CocoaPods one that we already have.

azarovalex:
### Description
Fixes #2903.

### Implementation
We already build the Example scheme on CI, where our SDKs are included via SPM, however, we added the `mapbox-navigation-ios` package to the Xcode project: 
image

This isn't the way our package will be integrated into the consumer apps. For example, we don't have MapboxNavigation and MapboxCoreNavigation in our [Package.resolved](https://github.com/mapbox/mapbox-navigation-ios/blob/release-v2.0/MapboxNavigation.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved) file.

On the other hand, my sample app contains an external reference to the `MapboxNavigation` package, which is the expected way to integrate our SDK:
image

By integrating the package this way we now have it in [Package.resolved](https://github.com/mapbox/mapbox-navigation-ios/blob/3cf0ef9b36c4ce8e2dc27b4c078098d260c0bc4b/Tests/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved#L50) file.
Also now we have `XCRemoteSwiftPackageReference` section in [.pbxproj file](https://github.com/mapbox/mapbox-navigation-ios/blob/3cf0ef9b36c4ce8e2dc27b4c078098d260c0bc4b/Tests/SPMTest/SPMTest.xcodeproj/project.pbxproj#L354-L363), which we don't have in current development project.
All these will help us to test SPM integration fairer and avoid issues like #2824.

I also added new CircleCI step that builds and archives this sample app.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -255,6 +255,26 @@ jobs:
       - *build-Example
       - *save-cache
 
+  spm-integration-test:
+    parameters:
+      iOS:
+        type: string
+        default: "14.4"
+      xcode:
+        type: string
+        default: "12.4.0"
+    macos:
+      xcode: << parameters.xcode >>
+    environment:
+      HOMEBREW_NO_AUTO_UPDATE: 1
+    steps:
+      - checkout
+      - *prepare-mapbox-file
+      - *prepare-netrc-file
+      - run: sed -i '' -e "s/6cc9728a615ed9b08632a6c2d09f7fca57a5c542/`git rev-parse --verify HEAD`/" Tests/SPMTest/SPMTest.xcodeproj/project.pbxproj # Replace SPM package reference with current revision
+      - run: cd Tests/SPMTest && xcodebuild -scheme SPMTest -destination 'platform=iOS Simulator,OS=<< parameters.iOS >>,name=iPhone 8 Plus' -derivedDataPath derivedData/ clean build | xcpretty
+      - run: cd Tests/SPMTest && xcodebuild -scheme SPMTest -sdk iphoneos<< parameters.iOS >> -destination generic/platform=iOS -derivedDataPath derivedData/ clean archive CODE_SIGNING_ALLOWED="NO" | xcpretty
+
   ios-trigger-metrics:
     parameters:
       xcode:
@@ -294,6 +314,7 @@ workflows:
           iOS: "14.0"
           lint: true
       - xcode-12-examples
+      - spm-integration-test
       - ios-trigger-metrics:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -272,9 +272,7 @@ jobs:
       - *prepare-mapbox-file
       - *prepare-netrc-file
       - *add-github-to-known-hosts
-      - run: sed -i '' -e "s/6cc9728a615ed9b08632a6c2d09f7fca57a5c542/`git rev-parse --verify HEAD`/" Tests/SPMTest/SPMTest.xcodeproj/project.pbxproj # Replace SPM package reference with current revision
-      - run: cd Tests/SPMTest && xcodebuild -scheme SPMTest -destination 'platform=iOS Simulator,OS=<< parameters.iOS >>,name=iPhone 8 Plus' -derivedDataPath derivedData/ clean build | xcpretty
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -255,6 +255,27 @@ jobs:
...
+    macos:
+      xcode: << parameters.xcode >>
+    environment:
+      HOMEBREW_NO_AUTO_UPDATE: 1
+    steps:
+      - checkout
+      - *prepare-mapbox-file
+      - *prepare-netrc-file
+      - *add-github-to-known-hosts
+      - run: sed -i '' -e "s/6cc9728a615ed9b08632a6c2d09f7fca57a5c542/`git rev-parse --verify HEAD`/" Tests/SPMTest/SPMTest.xcodeproj/project.pbxproj # Replace SPM package reference with current revision
```
1ec5, line 275: Should we factor this command into a script to allow verifying locally on the command line?

Tests/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,124 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "MapboxCommon",
+        "repositoryURL": "https://github.com/mapbox/mapbox-common-ios.git",
+        "state": {
+          "branch": null,
+          "revision": "50a87705438ee5b2c890a1fcacb54dfd867d738f",
+          "version": "10.0.2"
```
1ec5, line 10: update-version.sh runs `pod update` to ensure that the CocoaPods integration test’s Podfile.lock reflects the dependency versions that CocoaPods would resolve in downstream projects. What would be the equivalent for SPM, to ensure that we aren’t just testing a hard-coded set of resolved versions?

Tests/SPMTest/SPMTest/AppDelegate.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  AppDelegate.swift
+//  SPMTest
+//
+//  Created by Alexander Azarov on 13.04.21.
+//
```
1ec5, line 6: Nit: This repository relies on a LICENSE.md in the root folder, so a comment block is unnecessary.

Tests/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
azarovalex: According to [this documentation page](https://developer.apple.com/documentation/swift_packages/building_swift_packages_or_apps_that_use_them_in_continuous_integration_workflows), `xcodebuild` will always update all the dependencies to the latest versions compatible with the `Package.swift` files referenced in the `.pbxproj` file. In our case, this is [one file](https://github.com/mapbox/mapbox-navigation-ios/blob/main/Package.swift).
To ensure that `.pbxproj` will always reference the latest version of this `Package.swift` file, I use `sed` command to substitute current commit hash.
image

Tests/SPMTest/SPMTest/AppDelegate.swift
azarovalex: Xcode created this automatically and I forgot to delete, thx for pointing out.

.circleci/config.yml
azarovalex: Sure, I added a new script `scripts/build_spm_integration_test.sh` in this PR.

1ec5 (approved):

PR #2931, opened 2021-04-13
Fix the simulate route options in Example

### LINKED ISSUE #2883: Route simulation no longer works
On the release-v2.0 branch, enabling the Simulate Route option in the Example application no longer seems to have any effect when the iOS Simulator is also set to simulate a route. SimulatedLocationManager is supposed to cause RouteController to ignore any location updates that come in through Core Location in favor of synthesized location updates along the route. But that doesn’t seem to be working anymore, so the only way to simulate a route using SimulatedLocationManager is to switch the iOS simulator to the None preset.

/cc @mapbox/navigation-ios

ShanMa1991:
### Description
This pr is to fix #2883 .  

The `ViewController` defines a `navigationMapView` , but when we request routes and form a `NavigationViewController`, it also contains a `navigationMapView`. So during the turn by turn navigation, the `ViewController` defined `navigationMapView` should be removed.  This pr is to uninstall the before defined `navigationMapView` when the turn by turn navigation starts.

### DIFF
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -150,9 +150,9 @@ class ViewController: UIViewController {
         clearMap.isHidden = true
         longPressHintView.isHidden = false
         
-        navigationMapView.unhighlightBuildings()
-        navigationMapView.removeRoutes()
-        navigationMapView.removeWaypoints()
+        navigationMapView?.unhighlightBuildings()
+        navigationMapView?.removeRoutes()
+        navigationMapView?.removeWaypoints()
         waypoints.removeAll()
     }
     
@@ -184,7 +184,7 @@ class ViewController: UIViewController {
 
         guard activeNavigationViewController == nil else { return }
 
-        present(navigationViewController)
+        presentAndRemoveMapview(navigationViewController)
     }
     
     func beginCarPlayNavigation() {
@@ -241,7 +241,7 @@ class ViewController: UIViewController {
         // Example of building highlighting in 2D.
         navigationViewController.waypointStyle = .building
         
-        present(navigationViewController, completion: beginCarPlayNavigation)
+        presentAndRemoveMapview(navigationViewController, completion: beginCarPlayNavigation)
     }
     
     func startBasicNavigation() {
@@ -262,7 +262,7 @@ class ViewController: UIViewController {
         // Control floating buttons position in a navigation view.
         navigationViewController.floatingButtonsPosition = .topTrailing
         
-        present(navigationViewController)
+        presentAndRemoveMapview(navigationViewController)
     }
     
     func startCustomNavigation() {
@@ -291,7 +291,7 @@ class ViewController: UIViewController {
         let navigationViewController = NavigationViewController(for: route, routeIndex: 0, routeOptions: routeOptions, navigationOptions: options)
         navigationViewController.delegate = self
 
-        present(navigationViewController, completion: beginCarPlayNavigation)
+        presentAndRemoveMapview(navigationViewController, completion: beginCarPlayNavigation)
     }
     
     func startGuidanceCardsNavigation() {
@@ -304,7 +304,7 @@ class ViewController: UIViewController {
         let navigationViewController = NavigationViewController(for: route, routeIndex: 0, routeOptions: routeOptions, navigationOptions: options)
         navigationViewController.delegate = self
         
-        present(navigationViewController, completion: beginCarPlayNavigation)
+        presentAndRemoveMapview(navigationViewController, completion: beginCarPlayNavigation)
     }
     
     // MARK: - UIGestureRecognizer methods
@@ -450,12 +450,13 @@ class ViewController: UIViewController {
         return navigationViewController
     }
     
-    func present(_ navigationViewController: NavigationViewController, completion: CompletionHandler? = nil) {
+    func presentAndRemoveMapview(_ navigationViewController: NavigationViewController, completion: CompletionHandler? = nil) {
         navigationViewController.modalPresentationStyle = .fullScreen
         activeNavigationViewController = navigationViewController
         
         present(navigationViewController, animated: true) {
             completion?()
+            self.navigationMapView = nil
         }
     }
     


Example/ViewController.swift
```diff
@@ -166,6 +166,9 @@ class ViewController: UIViewController {
     
     @IBAction func simulateButtonPressed(_ sender: Any) {
         simulationButton.isSelected = !simulationButton.isSelected
+        navigationMapView.mapView.update{
+            $0.location.puckType = simulationButton.isSelected ? .puck2D() : nil
+        }
```
1ec5, line 171: This isn’t covered by the initial bug report, but it makes sense.

Example/ViewController.swift
```diff
@@ -472,7 +475,7 @@ class ViewController: UIViewController {
     }
 
     func navigationService(route: Route, routeIndex: Int, options: RouteOptions) -> NavigationService {
-        let mode: SimulationMode = simulationButton.isSelected ? .always : .onPoorGPS
+        let mode: SimulationMode = simulationButton.isSelected ? .always : .never
```
1ec5, line 478: The previous code here was correct: when the button was deselected, simulation was supposed to only kick in when the user enters a tunnel or otherwise loses GPS reception. The bug report in #2883 is that `always` has no effect: no simulation ever takes place when the option is enabled.

Example/ViewController.swift
```diff
@@ -166,6 +168,14 @@ class ViewController: UIViewController {
...
     @IBAction func simulateButtonPressed(_ sender: Any) {
         simulationButton.isSelected = !simulationButton.isSelected
+        // When the simulation mode is on and the `LocationProvider` of the `MapView` is set up as `PassiveLocationManager`, the `NavigationMapView` should stop receiving location updates from the `PassiveLocationManager` and simply follow the simulated route during turn-by-turn navigation.
+        if simulationButton.isSelected {
+            navigationMapView.mapView.locationManager.locationProvider.stopUpdatingHeading()
+            navigationMapView.mapView.locationManager.locationProvider.stopUpdatingLocation()
+        } else {
+            navigationMapView.mapView.locationManager.locationProvider.startUpdatingHeading()
+            navigationMapView.mapView.locationManager.locationProvider.startUpdatingLocation()
+        }
```
1ec5, line 178: Pressing the button by itself should have no effect on how the preview map behaves. We should pause location updates when hiding the map in `uninstall(_:)`, since map location updates would be useless at that point, and resume location updates when reconfiguring the map in `configure(_:)`.

Example/ViewController.swift
```diff
@@ -154,6 +154,8 @@ class ViewController: UIViewController {
         navigationMapView.removeRoutes()
         navigationMapView.removeWaypoints()
         waypoints.removeAll()
+        navigationMapView.mapView.locationManager.locationProvider.startUpdatingHeading()
+        navigationMapView.mapView.locationManager.locationProvider.startUpdatingLocation()
```
1ec5, line 158: Similarly, pressing the Clear button shouldn’t affect how the map tracks the user’s location – that button is purely about the route-related overlays that appear on the map.

Example/ViewController.swift
```diff
@@ -473,7 +483,10 @@ class ViewController: UIViewController {
 
     func navigationService(route: Route, routeIndex: Int, options: RouteOptions) -> NavigationService {
         let mode: SimulationMode = simulationButton.isSelected ? .always : .onPoorGPS
-        
+        if simulationButton.isSelected {
+            navigationMapView.mapView.locationManager.locationProvider.stopUpdatingHeading()
+            navigationMapView.mapView.locationManager.locationProvider.stopUpdatingLocation()
+        }
```
1ec5, line 489: It would be surprising if merely getting a navigation service – as the method name implies – has a side effect of pausing location updates going to the map view.

ShanMa1991: I'll remove this part from this scope. Thanks for the suggestion.

Example/ViewController.swift
ShanMa1991: The problem after check is that the `func uninstall(_ navigationMapView: NavigationMapView) ` is not called during the runtime. after remove the `navigationMapView` when turn-by-turn navigation start, it's fixed.

Example/ViewController.swift
```diff
@@ -450,12 +451,13 @@ class ViewController: UIViewController {
...
     }
     
-    func present(_ navigationViewController: NavigationViewController, completion: CompletionHandler? = nil) {
+    func presentAndRemoveMapview(_ navigationViewController: NavigationViewController, completion: CompletionHandler? = nil) {
         navigationViewController.modalPresentationStyle = .fullScreen
         activeNavigationViewController = navigationViewController
         
         present(navigationViewController, animated: true) {
             completion?()
+            self.navigationMapView = nil
```
1ec5, line 460: We only want to remove the map view from the view hierarchy, not from memory. Call `uninstall(_:)` instead of nilling out this property. After all, once the trip ends, the application goes back to this view controller, which needs to show the same map view again.

ShanMa1991: But if we only remove the map view from the view hierarchy without set it to `nil`, the default `locationProvider` overridden by `PassiveLocationManager` will continue sending location updates, which leads to rerouting.

1ec5: Is there no way to stop a location provider while the map is in memory?

1ec5: Looks like this is what we did in v1._x_, so I suppose it’s fine to deallocate the map view then:

https://github.com/mapbox/mapbox-navigation-ios/blob/41525d9487fbd26e6b3cb3817aa1112ec69e73d4/Example/ViewController.swift#L377

ShanMa1991: Yes. The `navigationMapView ` has a property observer to remove itself from its superview by calling the `func uninstall()` . So it would stop the location updating and reduce the memory usage at the same time.

https://github.com/mapbox/mapbox-navigation-ios/blob/92823986228eb1ec7ee8413562b18298d68cb777/Example/ViewController.swift#L32-L36

1ec5 (approved):
Thanks for iterating on the solution; I think this looks good. I’ve opened mapbox/mapbox-navigation-ios-examples#100 to make it clearer to developers that they need to take a similar measure when transitioning from free driving to turn-by-turn navigation.

PR #2933, opened 2021-04-15
Reimplement label localization atop Expression

1ec5:
Restored the `NavigationMapView.localizeLabels()` implementation for automatically localizing map labels into the user’s language. The previous implementation was removed as part of upgrading to map SDK v10 in #2808, because it relied on [an NSExpression-based localization feature](https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L1494-L1637) in map SDK v6._x_, which has been replaced by a new Expression type system. This PR ports map SDK v6._x_’s NSExpression-based localization feature to Expression as part of MapboxNavigation, in anticipation of a more cross-platform solution in future versions of the map SDK. Note that this implementation continues to differ from map SDK v6._x_’s default behavior, in that road labels remain unlocalized by design.

More to come:

* [x] Port NSExpression-based localization to Expression (and from Objective-C to Swift)
* [ ] Add support for localizing formatted expressions
* [x] Port MGLVectorStyleLayer-based localizable layer discovery to VectorLayer
* [ ] Port NSExpression tests to Expression

/cc @mapbox/navigation-ios @mapbox/maps-ios-reviewers

### DIFF
--- a/Example/ViewController+FreeDrive.swift
+++ b/Example/ViewController+FreeDrive.swift
@@ -70,6 +70,7 @@ extension ViewController {
         
         navigationMapView.mapView.on(.styleLoaded, handler: { [weak self] _ in
             guard let self = self else { return }
+            self.navigationMapView.localizeLabels()
             self.addStyledFeature(self.trackStyledFeature)
             self.addStyledFeature(self.rawTrackStyledFeature)
         })
--- a/Sources/MapboxNavigation/Expression.swift
+++ b/Sources/MapboxNavigation/Expression.swift
@@ -1,5 +1,11 @@
 import MapboxMaps
 
+extension VectorSource {
+    static func preferredMapboxStreetsLanguage(for preferences: [String]) -> String? {
+        return nil
+    }
+}
+
 extension Expression {
     
     static func routeLineWidthExpression(_ multiplier: Double = 1.0) -> Expression {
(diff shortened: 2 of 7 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -843,8 +843,39 @@ open class NavigationMapView: UIView {
...
-        // TODO: Implement ability to localize road labels.
+        guard !AccountManager.hasChinaBaseURL,
+              let layers = try? mapView.style.styleManager.getStyleLayers() else {
+            return
+        }
+        
+        let streetsSourceIdentifiers: [String] = mapView.streetsSources().map { $0.id }
+        
+        for layerInfo in layers where layerInfo.type == "symbol" {
+            guard var layer = try? mapView.style.getLayer(with: layerInfo.id, type: SymbolLayer.self).get(),
```
1ec5, line 862: This call fails on styles that set `text-field` to a formatted expression, such as Streets and Navigation Day: mapbox/mapbox-maps-ios#268. Unfortunately, the label localization feature has no effect in these styles.

1ec5: Per https://github.com/mapbox/mapbox-maps-ios/issues/268#issuecomment-820317143, the bug is fixed in v2.0.0-beta.17.

Example/ViewController+FreeDrive.swift
```diff
@@ -70,6 +70,7 @@ extension ViewController {
         
         navigationMapView.mapView.on(.styleLoaded, handler: { [weak self] _ in
             guard let self = self else { return }
+            self.navigationMapView.localizeLabels()
```
MaximAlien, line 73: We were also calling `localizeLabels()` in `RouteMapViewController`, `CarPlayMapViewController` and `CarPlayNavigationViewController` in `1.x`.

Sources/MapboxNavigation/Expression.swift
```diff
@@ -1,5 +1,11 @@
 import MapboxMaps
 
+extension VectorSource {
```
MaximAlien, line 3: Shall we move this `VectorSource` extension into `VectorSource.swift`?

Sources/MapboxNavigation/MapView.swift
```diff
@@ -138,4 +138,47 @@ extension MapView {
         
         return []
     }
+    
+    /**
+     Returns a boolean value indicating whether the tile source is a supported version of the Mapbox Streets source.
+     */
+    func isMapboxStreets(_ identifiers: [String]) -> Bool {
+        return identifiers.contains("mapbox.mapbox-streets-v8") || identifiers.contains("mapbox.mapbox-streets-v7")
```
MaximAlien, line 146: Nit: IMO storing identifiers in array would be more agile. In some places we have too many `or` comparisons.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -843,8 +843,39 @@ open class NavigationMapView: UIView {
...
+            }
+            
+            // Road labels should match road signage.
+            let sourceLayerIdentifier = try? mapView.style.styleManager.getStyleLayerProperty(forLayerId: layer.id, property: "source-layer").value as? String
+            let isLabelLayer = VectorSource.roadLabelLayerIdentifiersByTileSetIdentifier.values.contains(sourceLayerIdentifier ?? "")
+            let locale = isLabelLayer ? Locale(identifier: "mul") : nil
+            
+            let localizedTextField = textField.localized(into: locale)
+            if localizedTextField != textField {
+                layer.layout?.textField = .expression(localizedTextField)
```
MaximAlien, line 876: Does this update work as expected? I thought that preferred way to update layers would be to call `Style.updateLayer(id:type:update:)`.

Sources/MapboxNavigation/Expression.swift
```diff
@@ -30,4 +36,86 @@ extension Expression {
...
+            
+            if propertyName == "name" || propertyName.starts(with: "name_") {
+                var localizedPropertyName = "name"
+                if locale?.identifier != "mul" {
+                    let preferences: [String]
+                    if let identifier = locale?.identifier {
+                        preferences = [identifier]
+                    } else {
+                        preferences = Locale.preferredLanguages
+                    }
```
MaximAlien, line 59: I'd just set default value for `preferences` instead of introducing else statement.

Sources/MapboxNavigation/MapView.swift
1ec5: These methods were moved from RouteMapViewController. We’ll need to keep an eye on #2902, which also moves them elsewhere.

This might be a good opportunity to drop support for Mapbox Streets v7. We supported both versions of the Streets source initially as a migration path, but I don’t think we need to do so long-term.

Sources/MapboxNavigation/Expression.swift
1ec5: It’s a matter of style, but in general, I prefer to minimize usage of local variables as much as possible to avoid mutability surprises further down in the code, even though it’s slightly more verbose.

Example/ViewController+FreeDrive.swift
1ec5: It is still being called in RouteMapViewController, CarPlayMapViewController, and CarPlayNavigationViewController, but it must’ve fallen out of the example application by accident.

Sources/MapboxNavigation/Expression.swift
```diff
@@ -30,4 +36,86 @@ extension Expression {
             }
         }
     }
+    
+    func localized(into locale: Locale?) -> Expression {
+        switch elements.first {
```
1ec5, line 41: This usage is now broken as of MapboxMaps v2.0.0-beta.18: mapbox/mapbox-maps-ios#299.

bamx23:
Is this one still a `release blocker`?

1ec5:
This proof of concept has run its course. The map SDK once again comes with a built-in label localization feature that the navigation SDK can hook into: #3330. The map SDK is missing some functionality that was included in this PR, but there are separate issues to track that work.

PR #2940, opened 2021-04-16
Rename CLLocationDirection.swift to CLLocationDirection+Camera.swift

jill-cardamon:
### Description
This PR renames CLLocationDirection.swift to CLLocationDirection+Camera.swift because docs failed to generate with filename conflict (similar to #2078 and #1952). This was added to the docs for v2.0.0-beta.5.


1ec5:
> similar to #2078 and #1952

@azarovalex, #2078 might be an interesting task to pick up, to ensure that these post-release, pre-publish errors don’t crop up again in the future.

1ec5 (approved):

PR #2942, opened 2021-04-20
Highlight lanes based on active direction

jill-cardamon:
### Description
This PR highlights lanes based on active direction using new properties from Directions API. #2801
- [x] Use active direction in ranked call
- [x] Include fallback to maneuver direction
- [x] update CHANGELOG.md

### Implementation
Use the preferred direction in the call to `LaneIndication.ranked(favoring:)` in `LaneView.draw(_:)` by passing it in when the `LaneView` is initialized by `LanesView.update(for:)`.

### Screenshots or Gifs

Before these changes, we would see issues like this one. The highlighted lanes conflict.

Simulator Screen Shot - iPhone 8 Plus - 2021-04-20 at 10 05 29

After the changes, we see the correct maneuver instruction: 

Simulator Screen Shot - iPhone 8 Plus - 2021-04-20 at 10 08 19

### DIFF
--- a/Sources/MapboxNavigation/LaneView.swift
+++ b/Sources/MapboxNavigation/LaneView.swift
@@ -343,6 +343,12 @@ open class LaneView: UIView {
         }
     }
     
+    var preferredDirection: LaneIndication {
+        didSet {
+            setNeedsDisplay()
+        }
+    }
+    
     var drivingSide: DrivingSide = .right {
         didSet {
             setNeedsDisplay()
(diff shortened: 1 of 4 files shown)


Sources/MapboxNavigation/LaneView.swift
```diff
@@ -401,7 +401,7 @@ open class LaneView: UIView {
         self.init(frame: LaneView.defaultFrame)
         backgroundColor = .clear
         self.indications = indications
-        maneuverDirection = direction ?? ManeuverDirection(rawValue: indications.description)
+        maneuverDirection = direction
```
1ec5, line 404: It’s probably still reasonable to keep this fallback for when the route doesn’t have the new active direction information.

1ec5 (approved):
Suggested changelog entry:

* Fixed an issue where lane guidance icons would sometimes highlight the wrong arrow. ([#2942](https://github.com/mapbox/mapbox-navigation-ios/pull/2942))

PR #2944, opened 2021-04-21
Navigation camera improvements.

MaximAlien:
### Description

PR introduces several Navigation Camera Viewport API improvements:
- Applied new Mapbox Maps camera related API changes.
- Fixed https://github.com/mapbox/mapbox-navigation-ios/issues/2945.
- Implemented https://github.com/mapbox/mapbox-navigation-ios/issues/2946.
- Fixed issue, which was causing incorrect camera course (by updating to Mapbox Maps `10.0.0-beta.18.1`).
- Fixed potential retain cycle when using `ViewportDataSourceDelegate`.
- Updated to Mapbox Maps `10.0.0-beta.18.1` and Mapbox Navigation Native `48.0.5`.

### DIFF
--- a/Sources/MapboxNavigation/FollowingCameraOptions.swift
+++ b/Sources/MapboxNavigation/FollowingCameraOptions.swift
@@ -148,6 +148,13 @@ public struct IntersectionDensity {
 public struct BearingSmoothing {
     
     /**
+     Controls whether bearing smoothing will be performed or not.
+     
+     Defaults to `true`.
+     */
+    public var enabled = true
+    
+    /**
      Controls how much the bearing can deviate from the location's bearing, in degrees.
      
      In case if set, the `bearing` property of `CameraOptions` during active guidance navigation
@@ -156,7 +163,7 @@ public struct BearingSmoothing {
      
      Defaults to `45.0` degrees.
      */
-    public var maximumBearingSmoothingAngle: CLLocationDirection? = 45.0
+    public var maximumBearingSmoothingAngle: CLLocationDirection = 45.0
 }
 
 /**
--- a/Sources/MapboxNavigation/FollowingCameraOptions.swift
+++ b/Sources/MapboxNavigation/FollowingCameraOptions.swift
@@ -120,7 +120,7 @@ public struct IntersectionDensity {
      
      Defaults to `true`.
      */
-    var enabled: Bool = true
+    public var enabled: Bool = true
     
     /**
      Multiplier, which will be used to adjust the size of the portion of the remaining step that's
@@ -128,7 +128,7 @@ public struct IntersectionDensity {
      
      Defaults to `7.0`.
      */
-    var averageDistanceMultiplier: Double = 7.0
+    public var averageDistanceMultiplier: Double = 7.0
     
     /**
      Minimum distance between intersections to count them as two instances.
@@ -138,7 +138,7 @@ public struct IntersectionDensity {
      
      Defaults to `20.0` meters.
      */
-    var minimumDistanceBetweenIntersections: CLLocationDistance = 20.0
+    public var minimumDistanceBetweenIntersections: CLLocationDistance = 20.0
 }
 
 /**
--- a/Sources/MapboxNavigation/FollowingCameraOptions.swift
+++ b/Sources/MapboxNavigation/FollowingCameraOptions.swift
@@ -15,20 +15,20 @@ public struct FollowingCameraOptions {
     public var defaultPitch: Double = 45.0
     
     /**
-     Highest altitude, which will be used when producing camera frame in `NavigationCameraState.following`
+     Minimum zoom, which will be used when producing camera frame in `NavigationCameraState.following`
      state.
      
-     Defaults to `50000.0` meters, or zoom level `10.50` (approximately).
+     Defaults to `10.50`.
      */
-    public var highestAltitude: CLLocationDistance = 50000.0
+    public var minimumZoomLevel: Double = 10.50
     
     /**
-     Lowest altitude, which will be used when producing camera frame in `NavigationCameraState.following`
+     Maximum zoom, which will be used when producing camera frame in `NavigationCameraState.following`
      state. It will be also used as initial value when active guidance navigation starts.
      
-     Defaults to `900.0` meters, or zoom level `16.35` (approximately).
+     Defaults to `16.35`.
      */
-    public var lowestAltitude: CLLocationDistance = 900.0
+    public var maximumZoomLevel: Double = 16.35
     
     /**
      If `true`, `NavigationViewportDataSource` will continuously modify `CameraOptions.center` property
--- a/Sources/MapboxNavigation/FollowingCameraOptions.swift
+++ b/Sources/MapboxNavigation/FollowingCameraOptions.swift
@@ -15,20 +15,20 @@ public struct FollowingCameraOptions {
     public var defaultPitch: Double = 45.0
     
     /**
-     Minimum altitude, which will be used when producing camera frame in `NavigationCameraState.following`
+     Highest altitude, which will be used when producing camera frame in `NavigationCameraState.following`
      state.
      
      Defaults to `50000.0` meters, or zoom level `10.50` (approximately).
      */
-    public var minimumAltitude: CLLocationDistance = 50000.0
+    public var highestAltitude: CLLocationDistance = 50000.0
     
     /**
-     Maximum altitude, which will be used when producing camera frame in `NavigationCameraState.following`
+     Lowest altitude, which will be used when producing camera frame in `NavigationCameraState.following`
      state. It will be also used as initial value when active guidance navigation starts.
      
      Defaults to `900.0` meters, or zoom level `16.35` (approximately).
      */
-    public var maximumAltitude: CLLocationDistance = 900.0
+    public var lowestAltitude: CLLocationDistance = 900.0
     
     /**
      If `true`, `NavigationViewportDataSource` will continuously modify `CameraOptions.center` property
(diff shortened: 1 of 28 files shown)


Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -53,31 +59,17 @@ public class NavigationViewportDataSource: ViewportDataSource {
      */
     public var defaultAltitude: CLLocationDistance = 1000.0
```
bamx23, line 60: Is it used anywhere?

Sources/MapboxNavigation/OverviewCameraOptions.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+
+/**
+ Options, which are used to control what `CameraOptions` parameters will be modified by
+ `NavigationViewportDataSource` in `NavigationCameraState.overview` state.
+ */
+public struct OverviewCameraOptions {
```
bamx23, line 7: It's basically part of `FollowingCameraOptions`. Maybe it's better to extract these parameters to a common structure (something like `CameraUpdateOptions`)? It would allow generalizing decisions on camera changes in `NavigationViewportDataSource`

Sources/MapboxNavigation/FollowingCameraOptions.swift
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     Controls how much the bearing can deviate from the location's bearing, in degrees.
+     
+     In case if set, the `bearing` property of `CameraOptions` during active guidance navigation
+     won't exactly reflect the bearing returned by the location, but will also be affected by the
+     direction to the upcoming framed geometry, to maximize the viewable area.
+     
+     Defaults to `45.0` degrees.
+     */
+    public var maximumBearingSmoothingAngle: CLLocationDirection? = 45.0
```
bamx23, line 87: 20° was too small?

Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -174,106 +166,154 @@ public class NavigationViewportDataSource: ViewportDataSource {
...
+            }
             
             return
         }
         
         if let location = activeLocation, let routeProgress = routeProgress {
+            var compoundManeuvers: [[CLLocationCoordinate2D]] = []
+            let frameGeometryAfterManeuver = followingCameraOptions.frameGeometryAfterManeuver
             let pitchСoefficient = self.pitchСoefficient(routeProgress, currentCoordinate: location.coordinate)
             let pitch = maximumPitch * pitchСoefficient
```
bamx23, line 213: What about extracting `maximumPitch` to `options`? And what do you think about introducing `minimumPitch`(even with 0 as a default value)?

Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -174,106 +166,154 @@ public class NavigationViewportDataSource: ViewportDataSource {
...
+                followingCarPlayCamera.padding = .zero
+            }
             
             return
         }
         
         if let location = activeLocation, let routeProgress = routeProgress {
+            var compoundManeuvers: [[CLLocationCoordinate2D]] = []
+            let frameGeometryAfterManeuver = followingCameraOptions.frameGeometryAfterManeuver
             let pitchСoefficient = self.pitchСoefficient(routeProgress, currentCoordinate: location.coordinate)
```
bamx23, line 212: Don't you want to make hardcoded values inside `pitchСoefficient` configurable?

Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -174,106 +166,154 @@ public class NavigationViewportDataSource: ViewportDataSource {
...
-                    }
-                    
-                    return 0.0
-                }
+            if options.followingCameraOptions.zoomUpdatesAllowed {
+                followingMobileCamera.zoom = CGFloat(self.zoom(coordinatesToManeuver + coordinatesForManeuverFraming,
+                                                               pitch: pitch,
+                                                               edgeInsets: viewportPadding,
+                                                               defaultZoomLevel: 2.0,
+                                                               maxZoomLevel: 16.35))
```
bamx23, line 261: What about making `zoom` max/min configurable too? Especially that zoom level `2` seems too distant for "following" mode. Anyway, I would recommend extracting all the hardcoded values in constants, at least repeating ones.

Sources/MapboxNavigation/FollowingCameraOptions.swift
MaximAlien: Was changed for parity with Android, it's set to `45.0` degrees there.

Sources/MapboxNavigation/NavigationViewportDataSource.swift
MaximAlien: Yep, some of those will be extracted to match Android implementation.

Sources/MapboxNavigation/NavigationViewportDataSource.swift
MaximAlien: Yep, it's needed when active guidance navigation just starts to prevent huge jump from initial location to current one.

Sources/MapboxNavigation/NavigationViewportDataSource.swift
MaximAlien: `maximumPitch` is now exposed as `FollowingCameraOptions.defaultPitch`. Android doesn't include `minimumPitch` for now, so we'll keep it as it is (at least for now).

Sources/MapboxNavigation/NavigationViewportDataSource.swift
MaximAlien: `pitchСoefficient` can now be modified via `PitchNearManeuver` struct.

Sources/MapboxNavigation/FollowingCameraOptions.swift
```diff
@@ -0,0 +1,210 @@
...
+ */
+public struct IntersectionDensity {
+    
+    /**
+     Controls whether additional coordinates after the upcoming maneuver will be framed
+     to provide the view extension.
+     
+     Defaults to `true`.
+     */
+    var enabled: Bool = true
```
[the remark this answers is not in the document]
MaximAlien: Yep, good catch. Fixed.

Sources/MapboxNavigation/FollowingCameraOptions.swift
```diff
@@ -0,0 +1,210 @@
...
+     */
+    public var minimumAltitude: CLLocationDistance = 50000.0
+    
+    /**
+     Maximum altitude, which will be used when producing camera frame in `NavigationCameraState.following`
+     state. It will be also used as initial value when active guidance navigation starts.
+     
+     Defaults to `900.0` meters, or zoom level `16.35` (approximately).
+     */
+    public var maximumAltitude: CLLocationDistance = 900.0
```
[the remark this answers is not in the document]
MaximAlien: Was also thinking about this. Renamed `maximumAltitude` to `lowestAltitude` and `minimumAltitude` to `highestAltitude`. Hopefully such names are more descriptive.

Sources/MapboxNavigation/OverviewCameraOptions.swift
MaximAlien: Even though those properties have similar naming each of them will be used in two different states, so I don't think that common place will give us significant benefits. Also there is a chance that this part will be changed based on end-users feedback. I'll keep in mind your proposal though.

Udumft (approved):
LGTM with few nits

azarovalex (approved):

Sources/MapboxNavigation/ViewportDataSource.swift
```diff
@@ -44,7 +44,7 @@ public protocol ViewportDataSource: class {
  Delegate, which is used to notify `NavigationCamera` regarding upcoming `CameraOptions`
  related changes.
  */
-public protocol ViewportDataSourceDelegate {
+public protocol ViewportDataSourceDelegate: class {
```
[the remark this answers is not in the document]
MaximAlien: Good catch. Replaced to `AnyObject`.

Sources/MapboxNavigation/FollowingCameraOptions.swift
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     Controls how much the bearing can deviate from the location's bearing, in degrees.
+     
+     In case if set, the `bearing` property of `CameraOptions` during active guidance navigation
+     won't exactly reflect the bearing returned by the location, but will also be affected by the
+     direction to the upcoming framed geometry, to maximize the viewable area.
+     
+     Defaults to `45.0` degrees.
+     */
+    public var maximumBearingSmoothingAngle: CLLocationDirection? = 45.0
```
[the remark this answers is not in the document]
MaximAlien: Agree, added `enabled` option.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -126,48 +117,60 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
-            mapView.zoom = zoom
+        let centerTimingParameters = UICubicTimingParameters(controlPoint1: CGPoint(x: 0.0, y: 0.0), controlPoint2: CGPoint(x: 1.0, y: 1.0))
+        animatorCenter = mapView.camera.makeAnimator(duration: duration, timingParameters: centerTimingParameters) { (transition) in
+            transition.center.toValue = center
         }
         
-        animatorBearing.stopAnimation()
-        animatorBearing.addAnimations {
-            mapView.bearing = bearing
+        animatorCenter?.startAnimation()
```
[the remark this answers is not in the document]
MaximAlien: Just tried to wrap these calls into a separate method, but don't see any benefits at this point (method has to be too generic to accomodate logic for both transitions and updates). Will consider this improvement as a tail-work though.

PR #2949, opened 2021-04-23
Remove `ElectronicHorizon` struct

### LINKED ISSUE #2906: Electronic horizon position notification should directly contain pointer to edge tree
As of #2834, the `electronicHorizonDidUpdatePosition` notification’s user info dictionary’s `treeKey` is set to an instance of the `ElectronicHorizon` struct, whose only purpose is to hold the starting `ElectronicHorizon.Edge` of the tree. This is redundant. Assuming we don’t intend to add more properties to `MapboxNavigationNative.ElectronicHorizon` in the future, we should eliminate the struct and place the starting edge directly in the user info dictionary. The extensive struct documentation comment should be split between `ElectronicHorizon.Edge`, `ElectronicHorizonOptions`, and `treeKey`. Nested types like `ElectronicHorizon.Edge` should be renamed to no longer nest underneath `ElectronicHorizon`.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/2834#discussion_r592032215
/cc @mapbox/navigation-ios

azarovalex:
### Description
Fixes #2906.

### Implementation
Removed the `ElectronicHorizon` struct and renamed `ElectronicHorizon.Edge` to `ElectronicHorizonEdge`.


Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdge.swift
```diff
@@ -1,48 +1,47 @@
...
+
+ During active turn-by-turn navigation, the user-selected route and its metadata influence the path of the electronic horizon determined by `RouteController`. During passive navigation (free-driving), no route is actively selected, so `PassiveLocationDataSource` will determine the most probable path from the vehicle’s current location. You can receive notifications about changes in the current state of the electronic horizon by observing the `Notification.Name.electronicHorizonDidUpdatePosition`, `Notification.Name.electronicHorizonDidEnterRoadObject`, and `Notification.Name.electronicHorizonDidExitRoadObject` notifications.
+
+ Use a `RoadGraph` object to get an edge with a given identifier.
+ */
+public struct ElectronicHorizonEdge {
```
bamx23, line 11: Maybe we should call it `RoadGraphEdge`? We have other entities with such semantics and this one seems to fit the same idea.

azarovalex: I think the less breaking our changes are for a user, the better it will be. Renaming from `ElectronicHorizon.Edge` to `ElectronicHorizonEdge ` feels simpler for the end user IMHO.
I also considered not removing `ElectronicHorizon` at all and instead convert it to enum without cases so that it will serve as a namespace, but then decided not to go with this soulution and just remove `ElectronicHorizon` at all.

bamx23: Booth names are "breaking". I don't consider it a huge issue while the SDK is in beta. My concern is about naming itself, but I don't insist on `RoadGraphEdge`.

1ec5: Nesting the type under `RoadGraph` as `RoadGraph.Edge` would be pretty elegant.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdge.swift
```diff
@@ -1,48 +1,47 @@
...
+
+ During active turn-by-turn navigation, the user-selected route and its metadata influence the path of the electronic horizon determined by `RouteController`. During passive navigation (free-driving), no route is actively selected, so `PassiveLocationDataSource` will determine the most probable path from the vehicle’s current location. You can receive notifications about changes in the current state of the electronic horizon by observing the `Notification.Name.electronicHorizonDidUpdatePosition`, `Notification.Name.electronicHorizonDidEnterRoadObject`, and `Notification.Name.electronicHorizonDidExitRoadObject` notifications.
+
+ Use a `RoadGraph` object to get an edge with a given identifier.
+ */
```
1ec5, line 10: Since this is now the documentation comment for the edge type specifically, we shouldn’t bury the lede in terms of describing an edge. Let’s start off with the first paragraph of the old documentation, then follow with the rest.

Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -343,7 +343,8 @@ extension ElectronicHorizon {
         public static let positionKey: NotificationUserInfoKey = .init(rawValue: "position")
         
         /**
-         A key in the user info dictionary of a `Notification.Name.electronicHorizonDidUpdatePosition` notification. The corresponding value is an `ElectronicHorizon` at the root of a tree of edges in the routing graph. */
+         A key in the user info dictionary of a `Notification.Name.electronicHorizonDidUpdatePosition` notification. The corresponding value is an `ElectronicHorizonEdge` at the root of a tree of edges in the routing graph. This graph represents a probable path (or paths) of a vehicle within the routing graph for a certain distance in front of the vehicle, thus extending the user’s perspective beyond the “visible” horizon as the vehicle’s position and trajectory change.
```
1ec5, line 346: 👍 It’s good that we’re putting some more detail here, now that there isn’t a general “electronic horizon” type to look up.

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdge.swift
azarovalex: Renamed `ElectronicHorizonEdge` to `RoadGraph.Edge`

Sources/MapboxCoreNavigation/EHorizon/ElectronicHorizonEdge.swift
azarovalex: Rearranged some lines in this doc.

Example/ViewController+FreeDrive.swift
```diff
@@ -97,17 +97,17 @@ extension ViewController {
     }
     
     @objc func didUpdateElectronicHorizonPosition(_ notification: Notification) {
-        guard let horizon = notification.userInfo?[ElectronicHorizon.NotificationUserInfoKey.treeKey] as? ElectronicHorizon else {
+        guard let horizon = notification.userInfo?[RoadGraph.NotificationUserInfoKey.treeKey] as? RoadGraph.Edge else {
```
bamx23, line 100: Maybe it's better to rename `horizon` into `startEdge` or smth?

Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -308,28 +308,28 @@ public extension Notification.Name {
     /**
      Posted when the user’s position in the electronic horizon changes. This notification may be posted multiple times after `electronicHorizonDidEnterRoadObject` until the user transitions to a new electronic horizon.
      
-     The user info dictionary contains the keys `ElectronicHorizon.NotificationUserInfoKey.positionKey`, `ElectronicHorizon.NotificationUserInfoKey.treeKey`, `ElectronicHorizon.NotificationUserInfoKey.updatesMostProbablePathKey`, and `ElectronicHorizon.NotificationUserInfoKey.distancesByRoadObjectKey`.
+     The user info dictionary contains the keys `RoadGraph.Edge.NotificationUserInfoKey.positionKey`, `RoadGraph.Edge.NotificationUserInfoKey.treeKey`, `RoadGraph.Edge.NotificationUserInfoKey.updatesMostProbablePathKey`, and `RoadGraph.Edge.NotificationUserInfoKey.distancesByRoadObjectKey`.
```
bamx23, line 311: `RoadGraph.NotificationUserInfoKey`? Here and everywhere else

Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -308,28 +308,28 @@ public extension Notification.Name {
...
+     The user info dictionary contains the keys `RoadGraph.Edge.NotificationUserInfoKey.roadObjectIdentifierKey` and `RoadGraph.Edge.NotificationUserInfoKey.transitionKey`.
     */
     static let electronicHorizonDidExitRoadObject: Notification.Name = .init(rawValue: "ElectronicHorizonDidExitRoadObject")
 }
 
-extension ElectronicHorizon {
+extension RoadGraph {
     /**
-     Keys in the user info dictionaries of various notifications posted by instances of `RouteController` or `PassiveLocationDataSource` about `ElectronicHorizon`s.
+     Keys in the user info dictionaries of various notifications posted by instances of `RouteController` or `PassiveLocationDataSource` about `RoadGraph.Edge`s.
```
bamx23, line 332: "about `RoadGraph`"?

Sources/MapboxCoreNavigation/EHorizon/RoadGraphEdge.swift
```diff
@@ -33,16 +37,16 @@ extension ElectronicHorizon {
 
         /**
          The edges to which the user could transition from this edge.
-         
+
          The most probable path may be split at some point if some of edges have a low probability difference (±0.05). For example, `outletEdges` can contain more than one edge with `level` set to 0. Currently, there is a maximum limit of one split per electronic horizon.
          */
-        public let outletEdges: [Edge]
+        public let outletEdges: [RoadGraph.Edge]
```
bamx23, line 43: Is this change needed?

Sources/MapboxCoreNavigation/EHorizon/RoadGraphEdge.swift
```diff
@@ -33,16 +37,16 @@ extension ElectronicHorizon {
...
 
         init(_ native: ElectronicHorizonEdge) {
             self.identifier = UInt(native.id)
             self.level = UInt(native.level)
             self.probability = native.probability
-            self.outletEdges = native.out.map(Edge.init)
+            self.outletEdges = native.out.map(RoadGraph.Edge.init)
```
bamx23, line 49: Is this change needed?

Sources/MapboxCoreNavigation/EHorizon/RoadGraphPath.swift
```diff
@@ -7,7 +7,7 @@ extension RoadGraph {
     /** A position along a linear object in the road graph. */
     public struct Path {
         /** The edge identifiers that fully or partially coincide with the linear object. */
-        public let edgeIdentifiers: [ElectronicHorizon.Edge.Identifier]
+        public let edgeIdentifiers: [RoadGraph.Edge.Identifier]
```
bamx23, line 10: Do we need to explicitly mention `RoadGraph` here(as we are already inside of `RoadGraph`)? Same for `Position`

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
```diff
@@ -26,7 +26,7 @@ public final class RoadGraph {
      
      - returns: A line string corresponding to the given edge identifier, or `nil` if the edge is inaccessible.
      */
-    public func edgeShape(edgeIdentifier: ElectronicHorizon.Edge.Identifier) -> LineString? {
+    public func edgeShape(edgeIdentifier: RoadGraph.Edge.Identifier) -> LineString? {
```
bamx23, line 29: Seems that we don't need `RoadGraph.` prefix here and in the `edgeMetadata` method.

bamx23 (approved):
LGTM

PR #2930, opened 2021-04-13
Upgrade to Mapbox Maps v10.0.0-beta.18, Mapbox Navigation Native v48.0.4 and Mapbox Common v11.0.2.

MaximAlien:
### Description

Upgraded dependencies to Mapbox Maps `v10.0.0-beta.18`, Mapbox Navigation Native `v48.0.4` and Mapbox Common `v11.0.2`.

Upgrade also fixes:
- https://github.com/mapbox/mapbox-navigation-ios/issues/2818

PR #2952, opened 2021-04-24
Address history dumping related feedback.

MaximAlien:
### Description

PR addresses feedback from https://github.com/mapbox/mapbox-navigation-ios/pull/2930.

### DIFF
--- a/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
+++ b/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
@@ -22,19 +22,23 @@ class Navigator {
     static var tilesURL: URL? = nil
     
     /**
-     Path to the file where history could be stored when `Navigator.dumpHistory(_:)` is called.
+     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called.
      */
-    static var historyFilePath: URL? = nil
+    static var historyDirectoryURL: URL? = nil
     
     /**
-     Store history to the file stored in `Navigator.historyFilePath` and asynchronously run a callback
-     when dumping finishes.
+     Store history to the directory stored in `Navigator.historyDirectoryURL` and asynchronously run a callback
+     when writing finishes.
      
-     - parameter completion: A block object to be executed when history dumping ends.
+     - parameter completionHandler: A block object to be executed when history dumping ends.
      */
-    func dumpHistory(_ completion: @escaping (String?) -> Void) {
+    func writeHistory(completionHandler: @escaping (URL?) -> Void) {
         historyRecorder.dumpHistory { (path) in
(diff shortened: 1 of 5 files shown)


Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -22,19 +22,23 @@ class Navigator {
     static var tilesURL: URL? = nil
     
     /**
-     Path to the file where history could be stored when `Navigator.dumpHistory(_:)` is called.
+     Path to the directory where history could be stored when `Navigator.writeHistory(completionHandler:)` is called.
```
MaximAlien: We actually need directory, not file in this case.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -22,19 +22,23 @@ class Navigator {
...
+     when writing finishes.
      
-     - parameter completion: A block object to be executed when history dumping ends.
+     - parameter completionHandler: A block object to be executed when history dumping ends.
      */
-    func dumpHistory(_ completion: @escaping (String?) -> Void) {
+    func writeHistory(completionHandler: @escaping (URL?) -> Void) {
         historyRecorder.dumpHistory { (path) in
-            completion(path)
+            if let path = path, let historyFileURL = URL(string: path) {
```
1ec5, line 37: Does the original string contain a file: URL or a file path?

```suggestion
            if let path = path, let historyFileURL = URL(fileURLWithPath: path) {
```

1ec5 (approved):

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
MaximAlien: `fileURLWithPath` works fine as well.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -107,7 +111,7 @@ class Navigator {
                                                 config: NavigatorConfig(),
                                                 customConfig: customConfig)
         
-        historyRecorder = HistoryRecorderHandle.build(forHistoryFile: Navigator.historyFilePath?.absoluteString ?? "", config: configFactory)
+        historyRecorder = HistoryRecorderHandle.build(forHistoryFile: Navigator.historyDirectoryURL?.absoluteString ?? "", config: configFactory)
```
bamx23, line 114: @MaximAlien It should be `Navigator.historyDirectoryURL?.path`, not `Navigator.historyDirectoryURL?.absoluteString`. `url.absoluteString` returns a string in format `file:///...`

PR #2955, opened 2021-04-27
Exposing TileStores

Udumft:
### Description
Added open properties for getting corresponding TileStore instances from MapView and Navigator.
This is required as a part of `Offline Regions` feature implementation as `TileStore` is the core instance used to manipulate the regions.

### Implementation
Previously both implementation did depend on `Bundle.mapboxNavigation.suggestedTileURL`. Now default location is resolved by Common's `TileStore.getInstance`.
Navigator `TileStore` instance is retained separately to avoid cases when `static tilesURL` got updated after Navigator was instantiated.
Map's `TileStore` may potentially be nil, as Map supports a use-case with no `TileStore` configured. However, currently such option is impossible due to hardcoded MapView setup. This will be modified in future PRs.


Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -54,6 +53,8 @@ class Navigator {
         return RoadObjectsStore(navigator.roadObjectStore())
     }()
     
+    var tileStore: TileStore!
```
azarovalex, line 56: I don't really like using implicitly unwrapped optionals, I think we already had a few crashes because of them. Is there an option not to use them?

Udumft: actually yes. and also it possible to remove other optionals above as all of them are resolved in `init`

azarovalex (approved):

PR #2956, opened 2021-04-27
TileStore config for Navigator and MapView

### LINKED ISSUE #2926: TileStore config for MapView and Navigator
As a part of API updating, MapsView and Navigator configs will now expect a `TileStore` instance to be passed instead of a path string.
We need to provide a setting which will allow to configure this `TileStore` which should be shared by both `MapView` and `Navigator` objects. Default value should be taken from `TileStore.getInstance()`.
Also, there should be a less obvious way to configure `MapView` and `Navigator` independently to support advanced use-cases.

Currently, `MapView` tile store path is [configured](https://github.com/mapbox/mapbox-navigation-ios/blob/release-v2.0/Sources/MapboxNavigation/NavigationMapView.swift#L289) via `ResourceOptions` inside `NavigationMapView`. This is not configurable atm. and should be changed.
`Navigatorr` config is also [encapsulated](https://github.com/mapbox/mapbox-navigation-ios/blob/release-v2.0/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift#L22) and not configurable at runtime.
Existing defaulting to `.caches` directory is obsolete.

Udumft:
### Description
Resolves #2926 
Adds setting to control Navigator and MapView tile store location.

### Implementation
Navigator TileStore can be configured only before it is initialised. There are 2 entry points which instantiate a Navigator instance and it is never deallocated (for now): FreeDrive (aka `PassiveLocationDataSource` and Active Guidance (aka `RouteController). Navigator always has to have a tile storage somewhere.
MapView is handled by `NavigationMapView` which can be used separately or as a part of `NavigationViewController`. MapView can have no on-device tile storage at all.


Sources/MapboxNavigation/NavigationOptions.swift
```diff
@@ -47,6 +47,48 @@ open class NavigationOptions: NavigationCustomizable {
      */
     open var predictiveCacheOptions: PredictiveCacheOptions?
     
+    
+    /**
+     To add docs for cases
+     */
+    public enum TileStoreLocation {
```
Udumft, line 54: I am concerned about this thing. From one perspective, since Map and Navigator have slightly different use-cases (Map may have no TileStore) using `URL?` as a setting does not cover all possible cases. In this case `nil` for Nav == default, while `nil` for map == no tile store.
Introducing this enum construct allows fine control and convenience usage like:
```
options.tileStoreLocation = .custom(url) // to pass user-provided location, or
options.tileStoreLocation = .isolated(.default, noStorage) // to disable Map Tile storage while keeping Nav at default
```
On the other hand - this setting is not reusable while manipulating bare `NavigationView` or `NavigationService`. It is also very sensitive to initialisation order. If User wants to create his custom `NavigationService` to pass it in `NavigationOptions` - he has to also take care of configuring a tile storage parameter.

Udumft: And yes, naming is an open question

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -285,6 +305,8 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     
     var viewObservers: [NavigationComponentDelegate] = []
     
+    var mapTileStore: TileStore?
```
Udumft, line 308: Had to store it here to be able to reach it during `loadView` and configure `NavigationView`. Tile storage is configured at initialisation time for `MapView` which forces piping this as init parameter through `NavigationView` and `NavigationMapView`. Pretty much the same happens for Navigator.
Apart of ravaging the init API, it also makes tile store setup very sensitive to init order. If any component is created before `NavigationViewController.init` - it will ignore `NavigationOptions` setup.

Sources/MapboxNavigation/NavigationOptions.swift
1ec5: An enumeration with associated values seems reasonable to me.

> this setting is not reusable while manipulating bare NavigationView or NavigationService

Not sure I follow – we could modify those classes to accept a TileStoreLocation too, right?

Sources/MapboxNavigation/NavigationOptions.swift
```diff
@@ -47,6 +47,48 @@ open class NavigationOptions: NavigationCustomizable {
...
+        }
+        public enum MapStorage {
+            case `default`
+            case custom(URL)
+            case noStorage
+        }
+        
+        case `default`
+        case custom(URL)
+        case isolated(NavigatorStorage, MapStorage)
```
1ec5, line 67: This case would be clearer with labels for each of the associated values.

Sources/MapboxNavigation/NavigationOptions.swift
```diff
@@ -61,15 +103,17 @@ open class NavigationOptions: NavigationCustomizable {
      - parameter topBanner: The container view controller that presents the top banner.
      - parameter bottomBanner: The container view controller that presents the bottom banner.
      - parameter predictiveCacheOptions: Configuration for predictive caching. These options control how the `PredictiveCacheManager` will try to proactively fetch data related to the route. A `nil` value disables the feature.
+     - parameter tileStoreLocation: TODO
      */
-    public convenience init(styles: [Style]? = nil, navigationService: NavigationService? = nil, voiceController: RouteVoiceController? = nil, topBanner: ContainerViewController? = nil, bottomBanner: ContainerViewController? = nil, predictiveCacheOptions: PredictiveCacheOptions? = nil) {
+    public convenience init(styles: [Style]? = nil, navigationService: NavigationService? = nil, voiceController: RouteVoiceController? = nil, topBanner: ContainerViewController? = nil, bottomBanner: ContainerViewController? = nil, predictiveCacheOptions: PredictiveCacheOptions? = nil, tileStoreLocation: TileStoreLocation = .default) {
```
1ec5, line 108: There are a lot of arguments with default values. If the default is sensible in most cases, we could omit this argument from the initializer and expect developers to set the `tileStoreLocation` property on a separate line.

This goes for the rest of the arguments too. I’m not sure why we really need a convenience initializer, since there’s no inherent order to the arguments other than the order in which we implemented each feature. v2.0 would be a good opportunity to remove this initializer, though maybe in a separate PR.

Sources/MapboxNavigation/NavigationViewController.swift
1ec5: > If any component is created before `NavigationViewController.init` - it will ignore `NavigationOptions` setup.

This suggests that maybe those components should also accept a NavigationOptions in their initializers, or that the tile store should be independent of NavigationOptions.

Sources/MapboxNavigation/NavigationOptions.swift
Udumft: > Not sure I follow – we could modify those classes to accept a TileStoreLocation too, right?

In this case we should frame it outside `NavigationOptions` to avoid ugly`NavigationService.init` definition, and it makes `TileStoreLocation` a standalone type which I'd like to avoid. Also, in terms of bare `NavigationMapView` plain `TileStore?` is a more robust way for picking storage location.

Sources/MapboxNavigation/NavigationViewController.swift
Udumft: >that the tile store should be independent of NavigationOptions

Not sure to follow this part. Did you mean to extract TileStore configuration from `NavigationOptions`? But that ruins concept of `NavigationOptions` as single customization source.
Having components accept `NavigationOptions` seems and overkill since it just needs a TileStore bit. But it brings us back to having `TileStoreLocation` as a standalone type suitable for configuring components. I guess I'll try this case

Udumft:
I've updated the code to utilise `TileStoreLocation` in all components where TileStore setting is required. Hopefully it will be clearer for users what kind of setting is required and help avoid initialisation order issues

Sources/MapboxCoreNavigation/TileStoreLocation.swift
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ To add docs for cases
+ */
+public enum TileStoreLocation {
+    public enum Strict {
+        case `default`
+        case custom(URL)
+        
+        public var tilStoreURL: URL? {
```
azarovalex, line 13: small typo here

Sources/MapboxCoreNavigation/TileStoreLocation.swift
```diff
@@ -0,0 +1,64 @@
...
+            switch self {
+            case .default:
+                return nil
```
azarovalex, line 16: Those enums seem too sophisticated IMHO, especially the fact that `Strict.tileStoreURL` can be nil.
I thought that `Strict` enum is used to explicitly show that the url is always present.
I wonder if there is a way to simplify code here.

Sources/MapboxCoreNavigation/TileStoreLocation.swift
```diff
@@ -0,0 +1,64 @@
+
+import Foundation
+import MapboxCommon
+
+/**
+ To add docs for cases
+ */
+public enum TileStoreLocation {
```
bamx23, line 8: I have an idea for this class:
```swift
public struct TileStoreConfiguration {
    enum Location: RawRepresentable {
        case `default`
        case custom(URL)

        init?(rawValue: URL?) {
            self = rawValue.map { .custom($0) } ?? .default
        }

        var rawValue: URL? {
            switch self {
            case .default:
                return nil
            case .custom(let url):
                return url
            }
        }

        var tileStore: TileStore {
            switch self {
            case .default:
                return TileStore.getInstance()
            case .custom(let url):
                return TileStore.getInstanceForPath(url.path)
            }
        }
    }

    var navigatorLocation: Location
    var mapLocation: Location?

    static var `default`: Self {
        .init(navigatorLocation: .default, mapLocation: .default)
    }
    static func custom(_ url: URL) -> Self {
        .init(navigatorLocation: .custom(url), mapLocation: .custom(url))
    }
    static func isolated(navigationLocation: Location, mapLocation: Location?) -> Self {
        .init(navigatorLocation: navigationLocation, mapLocation: mapLocation)
    }
}
```

Udumft: I like this! I think I'll try this in next iteration

bamx23: I'm not sure if `Location` enum needs `RawRepresentable`, but it looks cooler :) But also this makes some ambiguity for `let x = Location(rawValue: nil)` vs `let x: Location? = nil`.

Sources/MapboxNavigation/NavigationOptions.swift
```diff
@@ -47,6 +47,12 @@ open class NavigationOptions: NavigationCustomizable {
      */
     open var predictiveCacheOptions: PredictiveCacheOptions?
```
bamx23, line 48: And, by the way, we have `...Options` semantic here. Maybe it's better to keep this in this PR and rename `TileStoreConfiguration` into `TileStoreOptions`?

Sources/MapboxCoreNavigation/TileStoreConfiguration.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+    
+    /**
+     Location of Navigator tiles data
+     */
+    public let navigatorLocation: Location
+    /**
+     Location of Map tiles data
+     */
+    public let mapLocation: Location?
```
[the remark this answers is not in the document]
Udumft: Yes, it should have an option to be nil. Map may have no local tile store. So to distinguish `default` and `no tilestore` we need it optional.
But you are also right about the only place it is used and is defaulted - that's a bug :)

Sources/MapboxNavigation/NavigationOptions.swift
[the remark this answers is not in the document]
Udumft: Good catch! Even though `NavigationMapView.enablePredictiveCaching` is covered by constructor, `PredictiveCacheManager` is publicly available and thus can potentially be called before any other setup is done.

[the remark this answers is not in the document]
bamx23: It worth thinking about how to avoid such potential problems in the future. Maybe it can be solved by extracting Navigator pre-initialization config into a separate struct to make a single entry point that at least easier to validate in Xcode by inspecting "show all callers" or something like that.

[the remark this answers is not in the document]
Udumft: I thought about that, but I see there are few issues: either we force this struct to be configured before any interaction w/ SDK (technically we introduce an SDK setup step) and trust user to do so at proper moment, or we encapsulate this setup anywhere where it may be triggered which makes the process transparent for users, but requires us to maintain all such places. What we are doing currently looks more like a latter case.

Udumft: Unfortunately it will cause naming conflict with Common's `MBXTileStoreOptions`. I would like to avoid namespace usage.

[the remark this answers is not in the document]
bamx23: I understand why you've done it this way. What I'm talking about is to make future internal changes safer, to reduce the chance of forgetting to put property initiation into some of the places where a Navigator instance can be created.

Udumft:
I think I've covered all entry points to `Navigator` initialization now. Indeed, that might become complicated to add more settings or maintain them. We might consider introducing some kind of a setup mechanism for such things to ease the support and reduce potential risk of human error.

Sources/MapboxNavigation/NavigationOptions.swift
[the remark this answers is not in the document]
Udumft: I think we should better frame it outside current PR, because this seems to me like a complex solution, covering more than tile store configuration.

Sources/MapboxCoreNavigation/TileStoreConfiguration.swift
```diff
@@ -0,0 +1,82 @@
...
+ Describes `TileStore` setup.
+ 
+ This struct encapsulates logic for handling `default` and `custom` paths as well as providing corresponding `TileStore`s.
+ It also covers differences between tile storages for Map and Navigation data. Tupically, you won't need to configure these and rely on defaults, unless you provide pre-downloaded data withing your app in which case you'll need `custom()` path to point to your data.
+ */
+public struct TileStoreConfiguration {
+    /**
+     Describes filesystem location for tile storage folder
+     */
+    public enum Location{
```
bamx23, line 15: NIT 
```suggestion
    public enum Location {
```

bamx23 (approved):

Sources/MapboxNavigation/NavigationOptions.swift
1ec5: I agree that “options” would be a more consistent naming choice. Internal type conflicts ideally shouldn’t influence what we expose publicly to developers, but this would be an opportunity to prefer the more grammatical “TileStorage” terminology over “TileStore” (which incidentally sounds like [a kind of shop](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dtiles)).

Sources/MapboxCoreNavigation/TileStoreConfiguration.swift
```diff
@@ -0,0 +1,82 @@
+
+import Foundation
+import MapboxCommon
+
+/**
+ Describes `TileStore` setup.
```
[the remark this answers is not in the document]
Udumft: It is. This class is exposed to user and is a main point for Offline Regions feature.

Sources/MapboxCoreNavigation/TileStoreConfiguration.swift
```diff
@@ -0,0 +1,82 @@
...
+        /**
+         Encapsulated default location.
+         
+         `rawValue` for this case will return `nil`
+         */
+        case `default`
+        /**
+         User-provided path to tile storage folder
+         */
+        case custom(URL)
```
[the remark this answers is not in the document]
Udumft: Using `URL` type is problematic because we cannot exactly say that `nil == default` value. Map and Navigator treat `nil` differently: Navigator will switch to a default location while Map may have no tile storage at all.
IMHO in this case an enum with associated value is more descriptive than a struct with static constant.

PR #2958, opened 2021-04-27
Fix WayNameView flashes 

### LINKED ISSUE #2954: Current road name label flashes previous road name on unnamed road
If the user travels along a named road then turns onto an unnamed road, the current road name label flashes the previous road’s name when toggling between overview mode and course tracking mode.

When the current road is unnamed, the current road name label continues to contain the previous road name. Every code path should set the label’s `text`, even if it just sets it to `nil`:

https://github.com/mapbox/mapbox-navigation-ios/blob/1555074946d80cd3c9b86c0f140af0e2c8b33e6b/Sources/MapboxNavigation/RouteMapViewController.swift#L555-L560 https://github.com/mapbox/mapbox-navigation-ios/blob/1555074946d80cd3c9b86c0f140af0e2c8b33e6b/Sources/MapboxNavigation/RouteMapViewController.swift#L688-L697

/cc @mapbox/navigation-ios

ShanMa1991:
### Description
This pr is to fix #2954 

### Implementation
When the road name is empty, set the text of the `wayNameView ` to `nil` and hide it. When the camera is transiting from overview mode to following mode, check where the `wayNameView` has text inside, if not, hide it.


Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -80,9 +80,11 @@ open class WayNameView: UIView {
         }
         
         if let roadName = feature.properties["name"] as? String {
```
azarovalex, line 82: Maybe one `if` with two conditions, instead of two nested `if`s?

Sources/MapboxNavigation/CameraController.swift
```diff
@@ -101,7 +101,9 @@ class CameraController: NavigationComponent, NavigationComponentDelegate {
         case .transitionToFollowing, .following:
             navigationViewData.navigationView.overviewButton.isHidden = false
             navigationViewData.navigationView.resumeButton.isHidden = true
-            navigationViewData.navigationView.wayNameView.isHidden = false
+            if let roadName = navigationViewData.navigationView.wayNameView.text {
```
azarovalex, line 104: `wayNameView` also has `attributedText` property, but looks like `UILabel` always keeps them in sync.

azarovalex (approved):

Sources/MapboxNavigation/WayNameView.swift
ShanMa1991: Thanks for the suggestion, I'll improve this part.

Sources/MapboxNavigation/CameraController.swift
ShanMa1991: Yes. The two properties when change one, the other one will be replaced too.

1ec5, line 104: This check should also account for an empty string. MapboxNavigation has a handy shortcut for getting a non-empty string, to simplify the logic a bit. Also, `roadName` appears to be unused:

```suggestion
            if let _ = navigationViewData.navigationView.wayNameView.text?.nonEmptyString {
```

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -110,7 +110,7 @@ extension NavigationMapView {
             guard navigationView.resumeButton.isHidden else { return }
             
             if let roadName = roadName {
-                navigationView.wayNameView.text = roadName
+                navigationView.wayNameView.text = roadName.isEmpty ? nil : roadName
```
1ec5, line 113:
```suggestion
                navigationView.wayNameView.text = roadName.nonEmptyString
```

1ec5 (approved):
Good to go once the edge case below is addressed.

Sources/MapboxNavigation/CameraController.swift
1ec5: I guess the `isEmpty` check below makes it unlikely that the name would be an empty string here, but being consistent here gives us more confidence that a future refactor won’t reintroduce the blank bubble.

CHANGELOG.md
```diff
@@ -65,6 +65,7 @@
 * Fixed an issue where `PassiveLocationDataSource` and `RouteController` did not use the access token and host specified by `PassiveLocationDataSource.directions` and `RouteController.directions`, respectively. Added the `PredictiveCacheOptions.credentials` property for specifying the access token and host used for prefetching resources. ([#2876](https://github.com/mapbox/mapbox-navigation-ios/pull/2876))
 * The top banner can now show a wider variety of turn lane configurations, such as combination U-turn/left turn lanes and combination through/slight right turn lanes. ([#2882](https://github.com/mapbox/mapbox-navigation-ios/pull/2882))
 * Removed `ElectronicHorizon` struct, now an electronic horizon notification directly contain pointer to a starting edge. `ElectronicHorizon.Edge` was renamed to `RoadGraph.Edge`. `ElectronicHorizon.NotificationUserInfoKey` was renamed to `RoadGraph.NotificationUserInfoKey`. ([#2949](https://github.com/mapbox/mapbox-navigation-ios/pull/2949))
+* Fixed an issue that the current road name label flashes when camera state changes or travels onto an unamed road. ([#2958](https://github.com/mapbox/mapbox-navigation-ios/pull/2958))
```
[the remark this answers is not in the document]
ShanMa1991: Thanks for the suggestion! I'll fix this part.

Sources/MapboxNavigation/CameraController.swift
ShanMa1991: Thanks for the suggestion, I'll improve this one.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion, I'll improve this one.

PR #2968, opened 2021-04-30
Allow to use map SDK location layer during turn-by-turn navigation

### LINKED ISSUE #3033: UserPuckCourseView should be subclassable
UserPuckCourseView should be `open` instead of `public` so that applications can customize some aspect of it without having to reimplement the basic design, integration with the UI style, or the arrow fading feature added in #2359.

https://github.com/mapbox/mapbox-navigation-ios/blob/5e5f4bd7aabbb7b7251b8feb908eafca0b8ebee3/Sources/MapboxNavigation/UserCourseView.swift#L38

/cc @mapbox/navigation-ios

ShanMa1991:
### Description
This pr is to allow developer choose between UserCourseView and map SDK location layer. 

- [x] allow the developer to use map SDK
- [x] allow the map puck receiving location update
- [x] allow the map puck receiving heading update
- [x] update the vanishing route line synced with the map sdk puck.
- [x] allow the customized Puck update orientation under PassiveLocationManager
- [x] Fix #3033. update the `UserHaloCourseView` and `UserPuckCourseView` to open access and subclassable.

### Implementation

Added the ability to allow developers choose between `NavigationMapView.userCourseView` and map SDK location layer through the `NavigationMapView.puckType` . The default puck is the `NavigationMapView.userCourseView`. To configure a customized 2D Puck as following :
```
        var configuration = Puck2DConfiguration()
        if #available(iOS 13.0, *) {
            configuration.topImage = UIImage(systemName: "arrow.up")
            configuration.scale = .constant(2.0)
        }
```
To  configure a customized 3D Puck as following:
```
        let uri = Bundle.main.url(forResource: "someModelFile", withExtension: "gltf")
        let myModel = Model(uri: uri,
                            position: [-122.396152, 37.79129], // Coordinates of the model in `[longitude, latitude]`format
                            orientation: [0, 0, 90])
        let scalingExpression = someExpression // Setting an expression to  scale the model based on camera zoom
        let configuration = Puck3DConfiguration(model: myModel, modelScale: .expression(scalingExpression))
```
To display the puck on NavigationMapView, use `navigationMapView.puckType = .puck3D(configuration: configuration)`.
To display the puck during turn-by-turn navigation, use `navigationViewController.navigationMapView.puckType = .puck3D(configuration: configuration)`

To allow the puck rotate under `PassiveLocationManager`, developers need to receive `Notification.passiveLocationDataSourceDidUpdate` and call the `navigationMapView.updateUserCourseView(location)` when there's a new location to adjust the rotation of the customized puck.

### Screenshots or Gifs
2D customized puck example during turn-by-turn navigation:
mapPuck

3D customized puck example during turn-by-turn navigation:
3DPuckExample

3D customized puck example under PassiveLocationManager:
passiveLocationPuck


ShanMa1991:
When we call the `mapView.location.locationProvider(_:didUpdateLocations:)`, it will automatically assign the `latestLocation.heading` to the puck. And the SimulatedLocationManager doesn't have a function to update a simulated `CLHeading`.

ShanMa1991:
Instead of creating a routeLine as a `LocationConsumer` to receive location update, the vanishing route line in `navigationMapView` also need `routeProgress` to calculate the `fractionTraveled`.  So it keeps synced with the map SDK puck at the same rate. The map SDK puck also avoids receiving location update directly from the default `AppleLocationProvider()`, especially during simulation mode or `PassiveLocationManager`.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -135,6 +135,41 @@ open class NavigationMapView: UIView {
         }
     }
     
+    var customizedPuckType: PuckType? {
```
MaximAlien, line 138: I wasn't following initial discussion regarding new puck functionality, but if our aim is to provide both 2D/3D puck capabilities and retain already existing `UserCourseView` what about exposing such entity:
```swift
public enum PuckType {
    
    case puck2D(configuration: Puck2DConfiguration)
    
    case puck3D(configuration: Puck3DConfiguration)
    
    case `default`(userCourseView: UserCourseView)
}
```

Based on such API user will be able to provide 2D/3D implementation of the puck whereas Navigation SDK will be providing default implementation like this:

```swift
public var puckType: PuckType = .default(userCourseView: UserPuckCourseView(frame: CGRect(origin: .zero, size: 75.0)))
```

/cc @1ec5

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -356,18 +391,40 @@ open class NavigationMapView: UIView {
         
         mostRecentUserCourseViewLocation = location
         
-        // While animating to overview mode, don't animate the puck.
-        let duration: TimeInterval = animated && navigationCamera.state != .transitionToOverview ? 1 : 0
-        UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear]) { [weak self] in
-            guard let screenCoordinate = self?.mapView.screenCoordinate(for: location.coordinate) else { return }
-            self?.userCourseView.center = CGPoint(x: screenCoordinate.x, y: screenCoordinate.y)
+        if customizedPuckType != nil {
```
MaximAlien, line 394: I do not fully understand this logic. In public API you expose `customizedPuckType` and based on it draw custom location puck. At the same time only `customizedPuck3DConfiguration` used. Also not sure why we need to stop updating location in Maps SDK location provider. Is it because of interference with location provided by NavigationNative?

I guess we should aim for such puck behavior:
- When only raw location is available we'll use it for the puck.
- If `PassiveLocationDataSource` is used we'll use snapped location for the puck.
- In case of active guidance puck will always use snapped location.

ShanMa1991: For the `customizedPuckType`, it won't be `nil` if either `customizedPuck3DConfiguration ` or `customizedPuck2DConfiguration ` is used to draw puck.

The problem is that when under simulation mode, the `locationProvider` of the default `mapView.location` will continue sending snapped location update to the puck. If we want the puck only to receive location from the `SimulatedLocationManager`, we could only do it from here (because now we don't provide a wrapped locationProvider for the `SimulatedLocationManager`). And if we didn't stop the location update from the snapped location provider, it would show a puck shifting behavior. 

And if `PassiveLocationDataSource` is used, and used by `mapView.location.overrideLocationProvider(with:)`. The user puck will default receiving location update from the locationProvider, the `PassiveLocationDataSource`, not the snapped one, unless we override it with the `AppleLocationProvider()`.

So maybe we should check the simulation mode, if it's under simulation, it should only receive location update from the `SimulatedLocationManager` through the `public func updateUserCourseView(_ : animated:)` . Otherwise, we just need to allow the user puck receive the location from it's `locationProvider`.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion! I'll improve this one by your advice!

ShanMa1991:
Now the camera is shaking stronger than #2998, when in simulation mode during turn-by-turn navigation, the camera should only receive the `routeProgress` for the location Update

zugaldia:
Per team chat, we've decide that https://github.com/mapbox/mapbox-navigation-ios/issues/2998 is not blocking this PR to land. However, we expect a fix from upstream dependencies ahead of the final v2 release. We will be using https://github.com/mapbox/mapbox-navigation-ios/issues/2998 to track that work.

ShanMa1991:
Because right now the mapSDK doesn't support the puck layer bearing rotation based on `course` but on `heading`. Now when the `CLHeading` is missing from the `Location` information, the mapSDK puck would not rotate and will always point to one side.

To handle this issue, the following cases are considered:
1. During free drive, we're using PassiveLocationManager, and call the `navigationMapView.updateUserCourseView(location)` to receive location update from `NotificationCenter.` instead of from the `mapView.location.locationProvider`.
2. During the turn-by-turn navigation. If it's under `simulation` mode, the mapSDK puck will only be updated from `navigationMapView.updateUserCourseView`, to avoid the influence of the `mapView.location.locationProvider`, which may cause the shifting of puck location and bearing,  we stop the location update from it.
3. During turn-by-turn navigation. If it's not under `simulation` mode, we only receive location update to update the puck from the `mapView.location.locationProvider`, using the snapped location to indicate the true location and true heading.

Sources/MapboxNavigation/PuckType.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import MapboxMaps
+
+public enum PuckType {
```
1ec5, line 4: Historically, we’ve only used the term “puck” informally. In most other contexts, developers encounter terms such as “user location indicator” or “user course view” that are independent of the specific look and feel. Consider renaming this type to `UserLocationStyle` to allow for a look and feel that looks nothing like the traditional hockey puck. Even though the map SDK is using the “puck” terminology more formally these days, a UserCourseView could look like anything, and the term “type” is even more overloaded than “style”.

Sources/MapboxNavigation/PuckType.swift
```diff
@@ -0,0 +1,12 @@
...
+    
+    case puck2D(configuration: Puck2DConfiguration)
+    
+    case puck3D(configuration: Puck3DConfiguration)
+    
+    case `default`
```
1ec5, line 10: The default case should be the first one.

Sources/MapboxNavigation/PuckType.swift
1ec5, line 4: This public type and each of its cases should be documented.

Sources/MapboxNavigation/PuckType.swift
1ec5, line 10: Currently, the `NavigationMapView.userCourseView` property sits alongside `NavigationMapView.puckType` but only applies when `puckType` is set to `default`. Ideally, we should avoid parallel state, because it often gets out of sync and confuses developers. Consider moving `userCourseView` to an associated value of this case. It would be unusual for a `default` case to have an associated value, so we could rename it to `courseView` at the same time.

Sources/MapboxNavigation/UserCourseView.swift
```diff
@@ -35,7 +35,7 @@ public extension CourseUpdatable {
 /**
  A view representing the user’s location on screen.
  */
-public class UserPuckCourseView: UIView, CourseUpdatable {
+open class UserPuckCourseView: UIView, CourseUpdatable {
```
1ec5, line 38: This change to UserPuckCourseView and UserHaloCourseView is worth mentioning in the changelog, small as it is.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
...
 * Added the ability to show specific route leg with color overriding. When specifying the `legIndex` in `NavigationMapView.show(_:legIndex:)`, the route line for the specific route leg would show congestion colors, while other route legs would show `NavigationMapView.routeCasingColor` by default. If not specified, all the route legs would show congestion colors. During turn-by-turn navigation, the default specified route leg is the current route leg. The route leg colors could be overridden by `NavigationMapView.routeCasingColor` and the congestion colors. ([#2833](https://github.com/mapbox/mapbox-navigation-ios/pull/2833))
+* Added the ability to allow developers choose between `NavigationMapView.userCourseView` and map SDK location layer during turn-by-turn navigation. By providing corresponding configuration for `.puck2D` or `.puck3D`, developers could use the location puck from `MapboxMaps` during turn-by-turn navigation. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
```
1ec5, line 34: The developer is the main audience of this changelog, so we can address them directly:

```suggestion
* Added the `NavigationMapView.puckType` property to customize how the user’s current location is displayed on the map. Set this property to `PuckType.puck2D` or `PuckType.puck3D` to use the [location layer](https://docs.mapbox.com/ios/maps/api/10.0.0-beta.20/Enums/LayerType.html#/s:10MapboxMaps9LayerTypeO17locationIndicatoryA2CmF) powered by the Mapbox Maps SDK instead of the default view-backed implementation specified by the `NavigationMapView.userCourseView` property. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
```

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -202,34 +202,54 @@ extension NavigationMapView {
...
-            }
-            
-            let newFractionTraveled = self.preFractionTraveled + traveledDifference * timePassedInMilliseconds.truncatingRemainder(dividingBy: 1000) / 1000
-            
+        switch puckType {
+        case .default:
+            let startDate = Date()
+            vanishingRouteLineUpdateTimer = Timer.scheduledTimer(withTimeInterval: 0.05, repeats: true, block: { timer in
+                let timePassedInMilliseconds = Date().timeIntervalSince(startDate) * 1000
+                if timePassedInMilliseconds >= 980 {
```
1ec5, line 210: I overlooked it before, but this code should probably have more comments explaining what’s significant about 980&nbsp;milliseconds, among other things.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -202,34 +202,54 @@ extension NavigationMapView {
...
+                        lineLayer.lineGradient = .expression(Expression.routeLineGradientExpression(mainRouteLayerGradient))
+                    }
+                    
+                    try self.mapView.mapboxMap.style.updateLayer(withId: mainRouteCasingLayerIdentifier) { (lineLayer: inout LineLayer) throws in
+                        let mainRouteCasingLayerGradient = self.routeLineGradient(fractionTraveled: newFractionTraveled)
+                        
+                        lineLayer.lineGradient = .expression(Expression.routeLineGradientExpression(mainRouteCasingLayerGradient))
+                    }
+                } catch {
+                    NSLog("Failed to update main route line layer.")
```
1ec5, line 232: Should we swallow the error like this, or does an error updating the layer indicate a more serious internal inconsistency in the SDK?

If we should log, use `print(_:)` instead of `NSLog(_:)`.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -161,6 +161,18 @@ open class NavigationMapView: UIView {
         }
     }
     
+    var simulation: Bool = true
+    var passiveLocationUsed: Bool = false
+    
+    /**
+     The `puckType` is used to provide  customized options for the user location puck during turn-by-turn navigation. Defaults to `UserCourseView`. Developers could switch to map SDK location layer by providing configuration to `.puck2D` or `.puck3D`.
+     */
```
1ec5, line 169: Specifics about each option should go in individual documentation comments on the `PuckType` cases.

```suggestion
    /**
     Specifies how the map displays the user’s current location, including the appearance and underlying implementation.
     
     By default, this property is set to `PuckType.default`.
     */
```

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -374,18 +409,60 @@ open class NavigationMapView: UIView {
...
+                    if var orientation = model.orientation,
+                       orientation.count > 2 {
+                        orientation[2] = orientation[2] + location.course
+                        model.orientation = orientation
+                    }
+                    
+                    let modelSourceCode = [NavigationMapView.ModelKeyIdentifier.modelSouce: model]
+                    if let data = try? JSONEncoder().encode(modelSourceCode),
+                       let jsonDictionary = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any] {
+                        try? mapView.mapboxMap.style.setSourceProperty(for: NavigationMapView.SourceIdentifier.puck3DSource, property: "models", value: jsonDictionary)
```
1ec5, line 461: Is it possible to use Model more directly, rather than serializing to JSON?

Sources/MapboxNavigation/PuckType.swift
ShanMa1991: Thanks for the suggestion! I'll change the name to the `UserLocationStyle `.

Sources/MapboxNavigation/PuckType.swift
ShanMa1991: Thanks for the suggestion, I'll improve the sequence and the documentation for this part.

Sources/MapboxNavigation/UserCourseView.swift
ShanMa1991: Thanks for the suggestion, I'll add changeLog list for it.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: thanks for the suggestion, I'll update the log to `print`

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: Because in case if the time was already in the last interval, it would invalidate the timer and wait for the next routeProgress update. I'll add comment explaining this one. Thanks for the suggestion.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Because the source type of the 3D puck layer is`ModelSource`, which is internal and could not be updated directly.

Sources/MapboxNavigation/PuckType.swift
ShanMa1991: Thanks for the suggestion, I've add the sync from the `puckType` to the `userCourseView` and add configuration for it to customize.

Sources/MapboxNavigation/UserLocationStyle.swift
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ A type that represents a `UIView` that is `CourseUpdatable`.
+ */
+public typealias UserCourseView = UIView & CourseUpdatable
+
+/**
+ The style of user location indicator in a navigation map view.
+ */
+public enum UserLocationStyle {
```
MaximAlien, line 12: Please also update `jazzy.yml` with newly added public symbol.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
MaximAlien: BTW this functionality was added when we were not updating location puck frame by frame. Since new camera functionality requires frame by frame updates we can probably re-evaluate current implementation and simplify it by using MapView callbacks instead of such timers.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -202,34 +202,55 @@ extension NavigationMapView {
...
+                    }
+                    
+                    try self.mapView.mapboxMap.style.updateLayer(withId: mainRouteCasingLayerIdentifier) { (lineLayer: inout LineLayer) throws in
+                        let mainRouteCasingLayerGradient = self.routeLineGradient(fractionTraveled: newFractionTraveled)
+                        
+                        lineLayer.lineGradient = .expression(Expression.routeLineGradientExpression(mainRouteCasingLayerGradient))
+                    }
+                } catch {
+                    print("Failed to update main route line layer.")
+                }
```
MaximAlien, line 234: This is basically a duplicate of code in `.default` case (other than `newFractionTraveled` and `fractionTraveled` properties usage). Can it be extracted into separate method?

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -374,18 +413,60 @@ open class NavigationMapView: UIView {
...
+            userCourseView.update(location: location,
+                                  pitch: cameraOptions.pitch!,
+                                  direction: cameraOptions.bearing!,
+                                  animated: animated,
+                                  navigationCameraState: navigationCamera.state)
+        case .puck2D(configuration: _):
+            if simulation {
+                if !passiveLocationUsed {
+                    let locationProvider = mapView.location.locationProvider
+                    mapView.location.locationProvider(locationProvider!, didUpdateLocations: [location])
```
MaximAlien, line 435: Can you add guard check to prevent forced unwrapping of `locationProvider`?

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -268,6 +302,11 @@ open class NavigationMapView: UIView {
             self.updateUserCourseView(location)
         }
         
+        mapView.mapboxMap.onNext(.cameraChanged) { [weak self] _ in
```
MaximAlien, line 305: Maybe `onEvery` is more appropriate here?

Sources/MapboxNavigation/UserLocationStyle.swift
ShanMa1991: Thanks for the suggestion, I'll update the docs for the `UserLocationStyle `

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: Thanks for the suggestion, I'll update the implementation of the vanishing route line.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Yes, you're right. The location provider could be override and changed after the mapView set up. The `onEvery ` could make sure that the `passiveLocationUsed` is keeping on track with the change.

ShanMa1991:
To update the vanishing route line based on frame rate has following problem now:
1. the `mapView.mapboxMap.onEvery(.renderFrameFinished)` should be called multiple time when the `routeProgress` changed to update the route. Otherwise the the route line would keep using the old `routeProgress` in every following frame.
2. Is there a way to stop the `mapView.mapboxMap.onEvery()` event? Otherwise there'll would be too many event listenner going on, especially for long route.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: Due to the [comment](https://github.com/mapbox/mapbox-navigation-ios/pull/2968#issuecomment-857285377) below, the vanishing route line method now is still using the Timer incase of overwhelming calculation.

ShanMa1991:
After test, the another constraint that block using the frame to update vanishing route line is based on the following reason. 
1. For example, we could implement the following way in the `RouteLineController.init(_:)` to replace the Timer:
```
        init(_ navigationViewData: NavigationViewData) {
            self.navigationViewData = navigationViewData
            routeProgress = navigationViewData.router.routeProgress
            let date = Date()
            self.navigationMapView.mapView.mapboxMap.onEvery(.renderFrameFinished) { [weak self] _ in
                print("timeInterval:\(abs(date.timeIntervalSinceNow))")
                guard let self = self,
                      let location = self.navigationMapView.mostRecentUserCourseViewLocation else { return }
                self.navigationMapView.updateUserCourseView(location)
                self.navigationMapView.updateTraveledRouteLine(location.coordinate)
                self.navigationMapView.updateRoute(self.routeProgress)
            }
        }
```
During the turn by turn navigation, if we zoom pretty close, and we have a following log of the time interval of the render frame finished:
```
timeInterval:6.394365072250366
timeInterval:6.417618036270142
timeInterval:6.426851034164429
timeInterval:6.450319051742554
//That's a long time interval where the frame will not be updated
timeInterval:19.978631019592285
timeInterval:19.996222019195557
timeInterval:20.01080596446991
timeInterval:20.027930974960327
timeInterval:20.043519020080566
timeInterval:20.059892058372498
timeInterval:20.076043009757996
```
That means that when we zoom pretty close and the route line would not be updated because the frame will not keep rendering update. There would show a tail of the route line behind the puck.

2. And another way to replace the Timer is to use location Indicator callback similar as in Android(though we don't have it right now). But the current constraint is that the rotation of the mapSDK puck is only based on `heading`, not `course`. When in simulation mode, or other circumstances, the puck would always point to one side. To fix this issue, in this pr, we stop receiving the location update from `mapView.locationProvider` in these cases to update the puck, and use the `navigationMapView.updateUserCourseView(_:animated:)` to update the puck layer and source properties directly as a workaround for the `course` based rotation. Only after mapSDK could allow the puck rotation based on `course`, we could allow the puck receive location update from the locationProvider in these cases. And then we could use the locationConsumer of other way to update the vanishing route line.


So based on the above reasons, the tail work of replacing the Timer of vanishing route line would be resolved and addressed in another separate ticket with the render frame rate improvement, or the rotation improvement.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -374,18 +410,65 @@ open class NavigationMapView: UIView {
...
+            let cameraOptions = CameraOptions(cameraState: mapView.cameraState)
+            userCourseView.update(location: location,
+                                  pitch: cameraOptions.pitch!,
+                                  direction: cameraOptions.bearing!,
+                                  animated: animated,
+                                  navigationCameraState: navigationCamera.state)
+        case .puck2D(configuration: _):
+            if simulation {
+                if !(mapView.location.locationProvider is PassiveLocationManager) {
+                    mapView.location.locationProvider.stopUpdatingLocation()
```
1ec5, line 431: I don’t quite understand why we’d only shut off the map view’s built-in location provider when using SimulatedLocationManager (since that’s the only case when `simulation` is true). Is it because PassiveLocationManager and ReplayLocationManager don’t conform to LocationProvider? If so, perhaps MapboxNavigation could extend these classes to conform to LocationProvider to eliminate this and other special cases.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -161,6 +162,19 @@ open class NavigationMapView: UIView {
         }
     }
     
+    var simulation: Bool = true
```
1ec5, line 165: This property’s name should make clear that it’s a Boolean flag (e.g., “simulatesLocation”). It‘s also only applicable to two of the possible value of `puckType`, so it should be an associated value instead of a parallel property.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -161,6 +162,19 @@ open class NavigationMapView: UIView {
...
+    
+    /**
+     Specifies how the map displays the user’s current location, including the appearance and underlying implementation.
+     
+     By default, this property is set to `UserLocationStyle.courseView`.
+     */
+    public var puckType: UserLocationStyle = .courseView(configuration: UserPuckCourseView(frame: CGRect(origin: .zero, size: 75.0))) {
```
1ec5, line 172: Now that the enumeration has been renamed, let’s rename the property to `userLocationStyle` to match it.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@
...
+* Added the `NavigationMapView.puckType` property to customize how the user’s current location is displayed on the map. Set this property to `UserLocationStyle.puck2D` or `UserLocationStyle.puck3D` to use the [locationlayer](https://docs.mapbox.com/ios/maps/api/10.0.0-beta.20/Enums/LayerType.html#/s:10MapboxMaps9LayerTypeO17locationIndicatoryA2CmF) powered by the Mapbox Maps SDK instead of the default view-backed implementation specified by the `NavigationMapView.userCourseView` property. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
+* Updated the `UserPuckCourseView` and `UserHaloCourseView` to open access, so that the two user course views could be customized and subclassable. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
```
1ec5, line 37: This explains why the developer might want to subclass these classes:

```suggestion
* If you need to customize the appearance of the user location indicator, you can subclass `UserPuckCourseView` and `UserHaloCourseView` as a starting point. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
```

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@
...
 * Added the ability to show specific route leg with color overriding. When specifying the `legIndex` in `NavigationMapView.show(_:legIndex:)`, the route line for the specific route leg would show congestion colors, while other route legs would show `NavigationMapView.routeCasingColor` by default. If not specified, all the route legs would show congestion colors. During turn-by-turn navigation, the default specified route leg is the current route leg. The route leg colors could be overridden by `NavigationMapView.routeCasingColor` and the congestion colors. ([#2833](https://github.com/mapbox/mapbox-navigation-ios/pull/2833))
+* Added the `NavigationMapView.puckType` property to customize how the user’s current location is displayed on the map. Set this property to `UserLocationStyle.puck2D` or `UserLocationStyle.puck3D` to use the [locationlayer](https://docs.mapbox.com/ios/maps/api/10.0.0-beta.20/Enums/LayerType.html#/s:10MapboxMaps9LayerTypeO17locationIndicatoryA2CmF) powered by the Mapbox Maps SDK instead of the default view-backed implementation specified by the `NavigationMapView.userCourseView` property. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
```
1ec5, line 36:
```suggestion
* Added the `NavigationMapView.puckType` property to customize how the user’s current location is displayed on the map. Set this property to `UserLocationStyle.puck2D` or `UserLocationStyle.puck3D` to use the [location indicator layer](https://docs.mapbox.com/ios/maps/api/10.0.0-beta.21/Enums/LayerType.html#/s:10MapboxMaps9LayerTypeO17locationIndicatoryA2CmF) powered by the Mapbox Maps SDK instead of the default view-backed implementation specified by the `NavigationMapView.userCourseView` property. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
```

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -374,18 +410,65 @@ open class NavigationMapView: UIView {
...
+                        location.altitude
+                    ]
+                    do {
+                        try mapView.mapboxMap.style.setLayerProperties(for: NavigationMapView.LayerIdentifier.puck2DLayer,
+                                                                       properties: [
+                                                                        "location": newLocation,
+                                                                        "bearing": location.course
+                                                                       ])
+                    } catch {
+                        NSLog("Failed to perform operation while updating 2D puck bearing arrow with error: \(error.localizedDescription).")
```
1ec5, line 446: Here’s another use of NSLog.

Sources/MapboxNavigation/NavigationMapView.swift
1ec5: ModelSource became internal in mapbox/mapbox-maps-ios#198. I don’t see any rationale associated with it, so it might have been just for lack of external demand. ModelSource would be more type-safe and forward-compatible than raw JSON usage. Since this is how the map SDK currently requires you to use the `models` property, let’s track ModelSource usage as tail work and also request an upstream change.

/cc @macdrevx

Sources/MapboxNavigation/NavigationMapView.swift
1ec5: This also calls into question whether we need separate PassiveLocationDataSource and PassiveLocationManager classes. Maybe there should be just PassiveLocationManager, with the LocationProvider conformance implemented in an extension in MapboxNavigation.

Sources/MapboxNavigation/UserLocationStyle.swift
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     The course view representing the user’s location and course on the map, switching between `UserPuckCourseView` and `UserHaloCourseView` based on the level of location accuracy.
+     */
+    case courseView(configuration: UserCourseView)
+    
+    /**
+     A 2-dimensional puck from `MapboxMaps`. Optionally provide `Puck2DConfiguration` to configure the puck's appearance.
+     */
+    case puck2D(configuration: Puck2DConfiguration)
```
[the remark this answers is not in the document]
ShanMa1991: I think this configuration should be optional, at the same handling as the MapboxMaps.  I'll provide a default value of `Puck2DConfiguration()` for it.

docs/jazzy.yml
```diff
@@ -137,6 +137,7 @@ custom_categories:
       - RouteSnappingMinimumHorizontalAccuracy
       - RouteSnappingMinimumSpeed
       - NavigationSettings
+      - UserLocationStyle
```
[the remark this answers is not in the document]
ShanMa1991: Thanks for the suggestion, I'll improve this part.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -418,6 +418,7 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
                               didPassVisualInstructionPoint: firstInstruction,
                               routeProgress: navigationService.routeProgress)
         }
+        navigationMapView?.simulation = (navigationService.locationManager is SimulatedLocationManager)
```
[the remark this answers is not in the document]
ShanMa1991: I see, I'll add the `ReplayLocationManager ` to this part.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Because during the turn-by-turn navigation simulation, the default map view’s built-in location provider is the AppleLocationProvider.  If we increase the simulation speed, the location from the AppleLocationProvider would be different from the routeProgress of the simulation. And the heading from the  built-in location provider would be `nil`. If we don't shut off the map view’s built-in location provider, the puck would receive both updates from the AppleLocationProvider as a LocationConsumer,  and the SimulatedLocationManager through route progress, which cause it shifting in location and also twinkling in rotation.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll rename it to `simulatesLocation `. This property defaults to `true` to allow the puck receive location update from the routeController. But it doesn't change with the puckType.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll rename this property.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the reminder, I'll remove this one.

ShanMa1991:
The tail work of the vanishing route line implementation will be addressed in #3069 separately.

Example/ViewController+FreeDrive.swift
```diff
@@ -43,6 +43,7 @@ extension ViewController {
         
         if let location = notification.userInfo?[PassiveLocationDataSource.NotificationUserInfoKey.locationKey] as? CLLocation {
             trackStyledFeature.lineString.coordinates.append(contentsOf: [location.coordinate])
+            navigationMapView.updateUserCourseView(location)
```
1ec5, line 46: `updateUserCourseView(_:animated:)` should probably be renamed at this point, since it affects the user location indicator even if it’s a location layer instead of a view. Consider `moveUserLocation(to:animated:)`, which also clarifies the relationship between the method and its first argument.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@
...
+* Added the `NavigationMapView.userLocationStyle` property to customize how the user’s current location is displayed on the map. Set this property to `UserLocationStyle.puck2D` or `UserLocationStyle.puck3D` to use the [location indicator layer](https://docs.mapbox.com/ios/maps/api/10.0.0-beta.21/Enums/LayerType.html#/s:10MapboxMaps9LayerTypeO17locationIndicatoryA2CmF) powered by the Mapbox Maps SDK instead of the default view-backed implementation specified by the `NavigationMapView.userCourseView` property. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
```
1ec5, line 36: I forgot to mention the associated value as part of the fully-qualified case names:

```suggestion
* Added the `NavigationMapView.userLocationStyle` property to customize how the user’s current location is displayed on the map. Set this property to `UserLocationStyle.puck2D(configuration:)` or `UserLocationStyle.puck3D(configuration:)` to use the [location indicator layer](https://docs.mapbox.com/ios/maps/api/10.0.0-beta.21/Enums/LayerType.html#/s:10MapboxMaps9LayerTypeO17locationIndicatoryA2CmF) powered by the Mapbox Maps SDK instead of the default view-backed implementation specified by the `NavigationMapView.userCourseView` property. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
```

Sources/MapboxNavigation/UserLocationStyle.swift
```diff
@@ -0,0 +1,27 @@
...
+public typealias UserCourseView = UIView & CourseUpdatable
+
+/**
+ The style of the user location indicator in a `NavigationMapView`.
+ */
+public enum UserLocationStyle {
+    /**
+     The course view representing the user’s location and course on the map, switching between `UserPuckCourseView` and `UserHaloCourseView` based on the level of location accuracy.
+     */
+    case courseView(configuration: UserCourseView)
```
1ec5, line 16: The associated value is the view to display, not a configuration object. Since the enumeration name already says “view”, there’s no need for a separate label:

```swift
case courseView(_ view: UserCourseView)
```

Sources/MapboxNavigation/UserLocationStyle.swift
```diff
@@ -0,0 +1,27 @@
...
+    
+    /**
+     A 2-dimensional puck from `MapboxMaps`. Optionally provide `Puck2DConfiguration` to configure the puck's appearance.
+     */
+    case puck2D(configuration: Puck2DConfiguration = Puck2DConfiguration())
```
1ec5, line 21: If there’s ever a need to distinguish between a customized 2D puck and an uncustomized one, accepting `nil` here would allow the SDK to make that distinction:

```swift
case puck2D(configuration: Puck2DConfiguration? = nil)
```

It would cause users of this case to need to handle `nil`, which may or may not be beneficial.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -418,6 +418,7 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
                               didPassVisualInstructionPoint: firstInstruction,
                               routeProgress: navigationService.routeProgress)
         }
+        navigationMapView?.simulatesLocation = (navigationService.locationManager is SimulatedLocationManager) || (navigationService.locationManager is ReplayLocationManager)
```
1ec5, line 421: NavigationLocationManager is `open`, so we should avoid special-casing SimulatedLocationManager and ReplayLocationManager if possible, because the developer is free to implement their own subclass that may or may not simulate location.

Sources/MapboxNavigation/NavigationMapView.swift
1ec5: Is there a situation where NavigationMapView should continue to use AppleLocationProvider? If we’re always trying to prevent AppleLocationProvider from being active, then we could shut it off more directly:

```swift
// No need for simulatesLocation flag
if let appleLocationProvider = mapView.location.locationProvider as? AppleLocationProvider {
    appleLocationProvider.stopUpdatingLocation()
}
```

1ec5 (approved):
Overall, I’m concerned about special-casing specific location managers/providers to achieve the normal behavior, which could leave an application-defined location manager/provider with unexpected behavior despite conforming to the right protocols.

I’m approving this PR on the condition that `simulatesLocation` be based on a property of the location manager class (via the NavigationLocationManager class and any necessary overrides), as discussed in https://github.com/mapbox/mapbox-navigation-ios/pull/2968#discussion_r649511124. I think that’s the minimum necessary to ensure that custom subclasses still work correctly. But we should investigate https://github.com/mapbox/mapbox-navigation-ios/pull/2968#discussion_r648690422 as tail work soon while there’s still time: having NavigationLocationManager conform to LocationProvider would be pretty elegant compared to the confusing and limiting PassiveLocationDataSource/PassiveLocationManager split we have now.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: There's a case that the `NavigationMapView` continue using `AppleLocationProvider`, that's during the turn-by-turn navigation with active guidance, not under simulation mode. And in this case, the `navigationMapView.updateUserCourseView` will still be called, to update the `navigationMapView.mostRecentUserCourseViewLocation`, which will be used for future vanishing route line. And the 2D puck or the 3D puck would not be skipped from the routeController updates, but only receiving the true location update from the `AppleLocationProvider`.

Example/ViewController+FreeDrive.swift
ShanMa1991: Thanks for the suggestion, I'll update the name.

Sources/MapboxNavigation/UserLocationStyle.swift
ShanMa1991: Thanks for the suggestion, I'll update this part.

Sources/MapboxNavigation/UserLocationStyle.swift
ShanMa1991: Thanks for the suggestion, I'll improve this part here.

Sources/MapboxNavigation/NavigationViewController.swift
ShanMa1991: Thanks for the suggestion, I add the property of the `simulatesLocation` to `NavigationLocationManager` to control the `navigationMapView.simulatesLocation` through the `navigationService.locationManager`. And also when the simulation mode is `.onPoorGPS`, it would still be `true` to display the motion of `userLocationStyle` based on route controller.

Sources/MapboxCoreNavigation/NavigationLocationManager.swift
```diff
@@ -17,6 +17,8 @@ open class NavigationLocationManager: CLLocationManager {
             allowsBackgroundLocationUpdates = true
         }
     }
+    
+    public var simulatesLocation: Bool = false
```
1ec5, line 21: Remember to document this property. You can link it to the other property you’ve added (and vice versa).

```md
 - seealso: `NavigationMapView.simulatesLocation`
```

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -418,7 +418,8 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
                               didPassVisualInstructionPoint: firstInstruction,
                               routeProgress: navigationService.routeProgress)
         }
-        navigationMapView?.simulatesLocation = (navigationService.locationManager is SimulatedLocationManager) || (navigationService.locationManager is ReplayLocationManager)
+
+        navigationMapView?.simulatesLocation = navigationService.locationManager.simulatesLocation || (navigationService.simulationMode != .never)
```
1ec5, line 422: @ShanMa1991 pointed out that `.onPoorGPS` would require this setting to change dynamically. Does NavigationViewController get notified somehow when simulation mode kicks in due to poor GPS?

ShanMa1991: Yes, I'll add the `navigationViewController.navigationService(_ :willBeginSimulating: becauseOf:)` and  `navigationViewController.navigationService(_ :willEndSimulating: becauseOf:)` to notify the `navigationMapView.simulatesLocation`

PR #2974, opened 2021-05-04
Fix NavigationViewController crash in storyboard.

### LINKED ISSUE #2973: NavigationViewController nonfunctional and can crash when initialized in storyboard
The workflow in the “[Storyboards](https://docs.mapbox.com/ios/navigation/guides/storyboards/)” guide doesn’t work because `NavigationViewController(coder:)` doesn’t call either `NavigationViewController(for:routeIndex:routeOptions:navigationOptions:)` or `NavigationViewController(for:routeIndex:routeOptions:navigationService:)`. Over time, `NavigationViewController(for:routeIndex:routeOptions:navigationOptions:)` has taken on some important responsibilities like setting up the speech controller and navigation service; without calling into it, the view controller would be nonfunctional. Additionally, if the application sets properties like `indexedRoute` and `navigationOptions` when segueing, as recommended by the guide, it would immediately crash because of setters that depend on implicitly unwrapped optionals (under the assumption that the other initializer has already been called).

`NavigationViewController(coder:)` would differ subtly from `NavigationViewController(for:routeIndex:routeOptions:navigationOptions:)` in that it needs to assume `navigationService` has already been initialized and set.

/ref #1063
/cc @mapbox/navigation-ios

MaximAlien:
### Description

Fixes #2973.

### Implementation

Since it's not possible to call other initializers in `NavigationViewController.init(coder:)` new public `NavigationViewController.initialize(for:routeIndex:routeOptions:navigationOptions:)` method was introduced. It's meant to be called right after `NavigationViewController` instantiation in storyboard. I've also brought back `NavigationViewController` to `Navigation.storyboard` to give the ability to create it using [this](https://docs.mapbox.com/ios/navigation/guides/storyboards/) guide.

### DIFF
--- a/Sources/MapboxNavigation/NavigationViewController.swift
+++ b/Sources/MapboxNavigation/NavigationViewController.swift
@@ -385,7 +385,9 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     func setupNavigationService() {
         guard let route = _route,
               let routeIndex = routeIndex,
-              let routeOptions = _routeOptions else { return }
+              let routeOptions = _routeOptions else {
+            fatalError("`route`, `routeIndex` and `routeOptions` must be valid to create an instance of `NavigationViewController`.")
+        }
         
         if !(routeOptions is NavigationRouteOptions) {
             print("`Route` was created using `RouteOptions` and not `NavigationRouteOptions`. Although not required, this may lead to a suboptimal navigation experience. Without `NavigationRouteOptions`, it is not guaranteed you will get congestion along the route line, better ETAs and ETA label color dependent on congestion.")
--- a/Sources/MapboxNavigation/NavigationViewController.swift
+++ b/Sources/MapboxNavigation/NavigationViewController.swift
@@ -287,10 +287,12 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     var cameraController: CameraController?
     var ornamentsController: NavigationMapView.OrnamentsController?
     var routeOverlayController: NavigationMapView.RouteOverlayController?
-    
-    var viewObservers: [NavigationComponentDelegate] = []
-    
+    var viewObservers: [NavigationComponentDelegate?] = []
     var mapTileStore: TileStoreConfiguration.Location? = .default
+    var _route: Route?
+    var _routeIndex: Int?
+    var _routeOptions: RouteOptions?
+    var navigationOptions: NavigationOptions?
     
     // MARK: - NavigationViewData implementation
         
@@ -327,9 +329,12 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
             mapTileStore = options.tileStoreConfiguration.mapLocation
         }
         
-        super.init(nibName: nil, bundle: nil)
+        _route = route
+        _routeIndex = routeIndex
+        _routeOptions = routeOptions
+        self.navigationOptions = navigationOptions
         
-        initialize(for: route, routeIndex: routeIndex, routeOptions: routeOptions, navigationOptions: navigationOptions)
+        super.init(nibName: nil, bundle: nil)
     }
     
     /**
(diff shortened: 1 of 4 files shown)


Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -318,6 +318,35 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     required public init(for route: Route, routeIndex: Int, routeOptions: RouteOptions, navigationOptions: NavigationOptions? = nil) {
         super.init(nibName: nil, bundle: nil)
         
+        initialize(for: route, routeIndex: routeIndex, routeOptions: routeOptions, navigationOptions: navigationOptions)
```
azarovalex, line 321: I expected the fix to be in `required public init?(coder aDecoder: NSCoder)`, does it work now?

MaximAlien: It'll not be possible to provide needed information in `required public init?(coder aDecoder: NSCoder)`. E.g. `route` object should already be available before calling that initializer, which doesn't seem to be possible. Seems that the only way is to initialize all needed objects afterwards and only after that present `NavigationViewController`.

MaximAlien: Example in this [guide](https://github.com/mapbox/ios-sdk/pull/1155/files#diff-15eb416535da909019f8d0f1dd8faaf5dcaf60dfe33e1e44df91bed2965dc113) shows how new usage can look like.

azarovalex: I see your point now, the problem is that if a user forgets to call `initialize` method the app will crash. 😿
But I guess it's the best we can do, hope this problem will be noticeable during the development process.

1ec5: > E.g. `route` object should already be available before calling that initializer, which doesn't seem to be possible.

The separate `initialize` method is really unfortunate. Its name points to a larger problem: most of the stuff that has been happening in the initializer only needs to take place by the time the view controller is presented, right? Ideally we’d move it all into `viewDidLoad()`. As it is, developers using NavigationViewController programmatically have to be careful about prematurely initializing the object. Developers don’t expect initializing a view controller or view to have side effects. I guess the tradeoff is that delaying these steps until `viewDidLoad()` would force us to think about situations where `navigationService` etc. are `nil`, so we’d have to revisit a bunch of implicitly unwrapped optionals.

I guess I’m OK with this approach as a stopgap fix, but in that case we should think about a broader lifecycle cleanup as tail work, with an eye towards deprecating the new `initialize` method.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -306,7 +306,7 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     }
     
     /**
-     Initializes a navigation view controller that presents the user interface for following a predefined route based on the given options.
+     Initializes a `NavigationViewController` that presents the user interface for following a predefined route based on the given options.
```
1ec5, line 309: By the way, it’s OK to refer to the thing the class represents rather than the symbol. There’s no ambiguity because the developer would read this documentation comment in the context of the NavigationViewController class reference already.

1ec5 (approved):
The changelog should mention that developers can once again segue to Navigation.storyboard or add NavigationViewController to a storyboard, and that `prepare(for:)` needs to call `NavigationViewController.initialize(for:routeIndex:routeOptions:navigationOptions:)`.

Sources/MapboxNavigation/NavigationViewController.swift
MaximAlien: I agree that additional initializer is not an optimal solution. I've changed implementation to use optional parameters for `route`, `routeOptions` `routeIndex` etc. User will be able to change it before `NavigationViewController` presentation. Example can be found [here](https://github.com/mapbox/mapbox-navigation-ios-examples/pull/113).

PR #2975, opened 2021-05-04
Fix the Maneuver selection in preview

### LINKED ISSUE #2912: Maneuver selection doesn't work correctly.
**Mapbox Navigation SDK version:** `maxim/camera-viewport-api`

### Steps to reproduce

 1. Start active guidance navigation.
 2. Select maneuver from top banner menu.
 3. Observe behavior.

### Expected behavior

It is expected that after maneuver selection maneuver is centered in the middle of the screen.

### Actual behavior

https://user-images.githubusercontent.com/1496498/113632475-0d3aa880-9620-11eb-907c-baa8bb010206.mov

ShanMa1991:
### Description
This pr is to fix #2912 , to allow the camera on the top of the selected maneuver in preview.

### Implementation
By adding the constraint of padding to the CameraOptions to set the center of `maneuverLocation` to be the center of the camera of `NavigationMapView`

### Screenshots or Gifs
maneuver_fix

### DIFF
--- a/Sources/MapboxNavigation/CameraController.swift
+++ b/Sources/MapboxNavigation/CameraController.swift
@@ -75,7 +75,7 @@ class CameraController: NavigationComponent, NavigationComponentDelegate {
     func center(on step: RouteStep, route: Route, legIndex: Int, stepIndex: Int, animated: Bool = true, completion: CompletionHandler? = nil) {
         navigationMapView.navigationCamera.stop()
 
-        let edgeInsets = navigationMapView.safeArea + UIEdgeInsets(top: 10.0, left: 20.0, bottom: 10.0, right: 20.0)
+        let edgeInsets = navigationMapView.safeArea + UIEdgeInsets.centerEdgeInsets
         let cameraOptions = CameraOptions(center: step.maneuverLocation,
                                           padding: edgeInsets,
                                           zoom: navigationMapView.mapView.zoom,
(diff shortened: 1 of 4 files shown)


Sources/MapboxNavigation/CameraController.swift
```diff
@@ -74,8 +74,10 @@ class CameraController: NavigationComponent, NavigationComponentDelegate {
     
     func center(on step: RouteStep, route: Route, legIndex: Int, stepIndex: Int, animated: Bool = true, completion: CompletionHandler? = nil) {
         navigationMapView.navigationCamera.stop()
-        // TODO: Verify that camera is positioned correctly.
+
+        let edgeInsets = navigationMapView.safeArea + UIEdgeInsets(top: 10.0, left: 20.0, bottom: 10.0, right: 20.0)
```
MaximAlien, line 78: Would it be possible to extract this `UIEdgeInsets` hardcoded value into separate constant? I think there are already four usages of similar value in other classes.

MaximAlien (approved):

Sources/MapboxNavigation/CameraController.swift
ShanMa1991: Thanks for the suggestion! I'll work in this part.

MaximAlien (approved):

Issue #2929, opened 2021-04-13
Arrival frame at destination changed


ShanMa1991:
**Mapbox Navigation SDK version:** release-v2.0

### Steps to reproduce

choose a building as destination and start navigation. 
Because the `func frameDestinationArrival(for location:)` removed from `NavigationViewController` and  the camera  changed from the overview mode to the tracking mode, the arrival frame has changed.

### Expected behavior
After the arrival at the destination, the destination should be at the center of the mapView, and the user location puck should be at the destination.
![expected](https://user-images.githubusercontent.com/48976398/114620307-4c8c7900-9c60-11eb-89a9-c01958b2ede3.png)

### Actual behavior
The destination is not at the center, the user location puck is not synced with the destination as well.
![actual](https://user-images.githubusercontent.com/48976398/114620955-fc61e680-9c60-11eb-8799-77f843e70cb0.png)

PR #2976, opened 2021-05-04
bring the arrival frame back to same as 1.x

ShanMa1991:
### Description
This pr is to fix the #2929 . After the arrival at the destination, the destination should be at the center of the mapView in an overview mode, which is the essential to display the building highlighting feature. And the user location puck should be at the destination.

### Implementation
By adding the center and pitch constraint to the camera, to let the destination at the center of the `mapView` and also update the `userCourseView` to be at the destination point.

### Screenshots or Gifs
arrival_fix

### DIFF
--- a/Sources/MapboxNavigation/NavigationViewController.swift
+++ b/Sources/MapboxNavigation/NavigationViewController.swift
@@ -738,20 +738,12 @@ extension NavigationViewController: NavigationServiceDelegate {
         
         arrivalController?.showEndOfRouteIfNeeded(self,
                                                   advancesToNextLeg: advancesToNextLeg,
-                                                  completion: { [ weak self] _ in
-                                                    self?.frameDestinationArrival(for: service.router.location)
-                                                  },
+                                                  completion: nil,
                                                   onDismiss: { [weak self] in
                                                     self?.navigationService.endNavigation(feedback: $0)
                                                     self?.handleCancelAction()
                                                   })
         return advancesToNextLeg
-    }
-    
-    func frameDestinationArrival(for location: CLLocation?) {
-        if waypointStyle == .annotation { return }
-        guard let location = location else { return }
-        navigationMapView?.updateUserCourseView(location, animated: false)
     }
 
     public func navigationService(_ service: NavigationService, willBeginSimulating progress: RouteProgress, becauseOf reason: SimulationIntent) {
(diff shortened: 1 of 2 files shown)


Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -730,13 +730,21 @@ extension NavigationViewController: NavigationServiceDelegate {
         
         arrivalController?.showEndOfRouteIfNeeded(self,
                                                   advancesToNextLeg: advancesToNextLeg,
-                                                  completion: nil,
+                                                  completion: { [ weak self] _ in
+                                                    self?.frameDestinationArrival(for: service.router.location)
+                                                  },
```
MaximAlien, line 735: Do we really need this part? `frameDestinationArrival(for:)` is basically duplication of already existing logic, which updates user location puck on each `MapView` frame change. I've also tested that there is no difference with/without this part, so probably better to remove it. `CameraOptions` change looks good.

MaximAlien (approved):

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -730,13 +730,21 @@ extension NavigationViewController: NavigationServiceDelegate {
...
                                                   onDismiss: { [weak self] in
                                                     self?.navigationService.endNavigation(feedback: $0)
                                                     self?.handleCancelAction()
                                                   })
         return advancesToNextLeg
     }
+    
+    func frameDestinationArrival(for location: CLLocation?) {
+        if waypointStyle == .annotation { return }
+        guard let location = location else { return }
```
azarovalex, line 745:
```suggestion
        guard waypointStyle != .annotation, let location = location else { return }
```

azarovalex: I guess guard should be used for early return.

azarovalex (approved):

Sources/MapboxNavigation/NavigationViewController.swift
ShanMa1991: Thanks for the suggestion, I'll improve this part.

Sources/MapboxNavigation/NavigationViewController.swift
ShanMa1991: Thanks for the suggestion, I'll remove this part. And should we also keep the zoom level after arrival same as 1.x ?

PR #2977, opened 2021-05-04
Clear ImageRepository cache when style change

### LINKED ISSUE #2938: Banner instruction attributed strings cache should be invalidated upon UI color mode change
When the color mode is flipped from light to dark mode, the SDK will continue to reuse cached attributed strings for things like Exit numbers in banner instructions. this can lead to color mismatches, as seen in the "Exit 50" image in the banner below, even when using dynamic colors:
Simulator Screen Shot - iPhone 11 - 2021-04-15 at 10 32 35

v1.4

### Steps to reproduce

 1. Assign a dynamic color to the PrimaryLabel for the DayStyle
 2. Bring up a banner instruction with an exit number
 3. Switch UI color mode to Dark

### Expected behavior
All text follows the dynamic color

### Actual behavior
Regular text changes but any generated images do not.

This occurs because we cache these exit images in the function `private func exitShield(side:text:dataSource:cacheKey:)` and similar. The `ImageRepository` cache should be cleared when the color mode changes.

ShanMa1991:
### Description
This pr is to fix #2938 by resetting `imageRepository` cache and reload `instruction` when style change.

### DIFF
--- a/Sources/MapboxNavigation/InstructionLabel.swift
+++ b/Sources/MapboxNavigation/InstructionLabel.swift
@@ -33,6 +33,13 @@ open class InstructionLabel: StylableLabel, InstructionPresenterDataSource {
         }
     }
 
+    open override func update() {
+        super.update()
+        imageRepository.resetImageCache(nil)
+        let previousInstruction = instruction
+        instruction = previousInstruction
+    }
+    
     private var instructionPresenter: InstructionPresenter?
 }
 
--- a/Sources/MapboxNavigation/InstructionLabel.swift
+++ b/Sources/MapboxNavigation/InstructionLabel.swift
@@ -40,6 +40,11 @@ open class InstructionLabel: StylableLabel, InstructionPresenterDataSource {
         instruction = previousInstruction
     }
     
+    open override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        update()
+    }
+    
     private var instructionPresenter: InstructionPresenter?
 }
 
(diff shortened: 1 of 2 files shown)


Sources/MapboxNavigation/StyleManager.swift
```diff
@@ -179,6 +179,7 @@ open class StyleManager {
         guard currentStyleType != styleType else { return }
         
         NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(timeOfDayChanged), object: nil)
+        ImageRepository.shared.resetImageCache(nil)
```
MaximAlien, line 182: Not sure that this allows to resolve reported issue. Here's behavior with cache resetting:

https://user-images.githubusercontent.com/1496498/118257487-8c16d280-b463-11eb-8e81-fa78c78afe2c.mov

As you can see icons still remain in color of previously used style. I'd recommend looking into whether cache resetting really removes files, and if so whether removing really correct approach. Maybe there will be need to reload list manually etc.

ShanMa1991: Yes, now it's not working to fix. i'll continue investing this issue to fix.

Sources/MapboxNavigation/InstructionLabel.swift
```diff
@@ -33,6 +33,13 @@ open class InstructionLabel: StylableLabel, InstructionPresenterDataSource {
         }
     }
 
+    open override func update() {
+        super.update()
```
avi-c, line 37: There is still one case of https://github.com/mapbox/mapbox-navigation-ios/issues/2938 where this doesn't work optimally. If you are currently viewing a UIView that that contains an InstructionLabel, such as the Step List, the cache isn't invalidated (and thus the colors are not updating) if you toggle light/dark mode. If you dismiss the UI and bring it back up then the colors will be correct.

You can add support for the `traitCollection` changing and trigger an invalidation there as well. For example, you could add: 

`open override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
        super.traitCollectionDidChange(previousTraitCollection)
        update()
    }`

to `InstructionLabel.swift` as well.

avi-c (changes requested):

Sources/MapboxNavigation/InstructionLabel.swift
ShanMa1991: Thanks for the suggestion! I'll work on this case.

avi-c (approved):

PR #2978, opened 2021-05-05
Update private Carthage dependencies

avi-c:
Update version of Quick/Quick to v3.1.2 and Quick/Nimble to v9.0.1 for Xcode 12.5 support. Remove unnecessary Cedar dependency.


MaximAlien:
Thanks for fixing these warnings in Xcode 12.5. I think we'll also have to update `Package.swift`.

MaximAlien (approved):

Cartfile.private
```diff
@@ -1,6 +1,5 @@
 github "mapbox/MapboxGeocoder.swift" ~> 0.10
-github "Quick/Quick" ~> 2.0
-github "Quick/Nimble" ~> 8.0
-github "CedarBDD/Cedar" ~> 1.0
```
1ec5, line 4: #3034 restores the Cedar dependency to avoid a compiler error, though it does so using the mapbox-events-ios checkout instead of directly depending on the framework via Carthage.

PR #2979, opened 2021-05-05
Update private Carthage Dependencies to support Xcode v12.5

avi-c:
Updates Quick/Quick to v3.1.2 and Nimble to v9.0.1.

Also pick up Nav Native v48.0.5.

Nav v2 version of https://github.com/mapbox/mapbox-navigation-ios/pull/2978


MaximAlien:
I think it'd be better to create separate PR, which is based on top of `main`. Current PR brings all changes which we've made in `release-v2.0`.

1ec5:
The same changes landed in main in #2978. Perhaps we can merge them the normal way as part of #2983 instead of applying the changes manually in this PR?

PR #2980, opened 2021-05-05
Fix Index out of bounds crash in RouteLegProgress.currentStep

ShanMa1991:
### Description
This pr is to add the range check of the `routeProgress.currentLeg.steps` before assign value to `routeProgress.currentLegProgress.stepIndex`, to avoid the out of bounds crush in the `RouteLegProgress.currentStep`

### DIFF
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -318,11 +318,14 @@ open class RouteController: NSObject {
         let newStepIndex = Int(status.stepIndex)
         let newIntersectionIndex = Int(status.intersectionIndex)
         
+        let oldLegIndex = progress.legIndex
         if newLegIndex != progress.legIndex {
             progress.legIndex = newLegIndex
         }
+        
+        if newStepIndex != progress.currentLegProgress.stepIndex {
+            precondition(progress.currentLegProgress.leg.steps.indices.contains(newStepIndex), "The stepIndex: \(newStepIndex) is higher than steps count: \(progress.currentLegProgress.leg.steps.count) of the leg :\(newLegIndex) or not included. The old leg index: \(oldLegIndex) will not be updated in this case.")
 
-        if (newStepIndex != progress.currentLegProgress.stepIndex) && (newStepIndex < progress.currentLeg.steps.endIndex) {
             progress.currentLegProgress.stepIndex = newStepIndex
         }
         
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -321,8 +321,8 @@ open class RouteController: NSObject {
         if newLegIndex != progress.legIndex {
             progress.legIndex = newLegIndex
         }
-        
-        if newStepIndex != progress.currentLegProgress.stepIndex {
+
+        if (newStepIndex != progress.currentLegProgress.stepIndex) && (newStepIndex < progress.currentLeg.steps.endIndex) {
             progress.currentLegProgress.stepIndex = newStepIndex
         }
         
--- a/Sources/MapboxCoreNavigation/RouteLegProgress.swift
+++ b/Sources/MapboxCoreNavigation/RouteLegProgress.swift
@@ -16,7 +16,7 @@ open class RouteLegProgress: Codable {
      */
     public var stepIndex: Int {
         didSet {
-            assert(stepIndex >= 0 && stepIndex < leg.steps.endIndex)
+            precondition(leg.steps.indices.contains(stepIndex), "It's not possible to set the stepIndex: \(stepIndex) when it's higher than steps count \(leg.steps.count) or not included.")
             currentStepProgress = RouteStepProgress(step: currentStep)
         }
     }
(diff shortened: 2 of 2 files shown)


Sources/MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -520,6 +520,7 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
             guard forcedStepIndex < routeProgress.currentLeg.steps.count else { return }
             routeProgress.currentLegProgress.stepIndex = forcedStepIndex
         } else {
+            guard routeProgress.currentLegProgress.stepIndex < routeProgress.currentLeg.steps.endIndex - 1 else { return }
```
MaximAlien, line 523: Is `LegacyRouteController` really used anywhere at this point? As I can see it was deprecated in favor of `RouteController`, so not sure that such preventive check works.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -283,7 +283,7 @@ open class RouteController: NSObject {
         if (newLegIndex != progress.legIndex) {
             progress.legIndex = newLegIndex
         }
-        if (newStepIndex != progress.currentLegProgress.stepIndex) {
+        if (newStepIndex != progress.currentLegProgress.stepIndex) && (newStepIndex < progress.currentLeg.steps.endIndex - 1) {
```
MaximAlien, line 286: Since `RouteLegProgress.currentStep` is a computed property, which accesses array by index, which might be out of bounds would it work if we check if index is valid right before that? It'd be also beneficial to create a unit test for such use case.

Sources/MapboxCoreNavigation/LegacyRouteController.swift
ShanMa1991: Thanks for the suggestion, I'll remove this part. I thought it would be better if we could keep it synced with the changes.

MaximAlien:
@1ec5, since `LegacyRouteController` is not used anymore do you think it'd be acceptable to remove it now?

MaximAlien:
@bamx23, btw is there a chance that you know when such crash occurs? E.g. was it during active guidance navigation during route refresh or something else? Also how often it happens?

MaximAlien:
I guess I'm fine with such fix. On the other hand without knowing actual root cause and without the ability to reliably reproduce it there is a chance that we might get something similar in another place. I'd also put preventive check into `stepIndex` setter to run it for all callers, instead of selective ones (not sure if that'll cause other side effects though in case when `currentStepProgress` is not updated).

MaximAlien (approved):
Please update changelog as well.

Sources/MapboxCoreNavigation/RouteController.swift
S2Ler, line 286: I wonder why it is `(newStepIndex < progress.currentLeg.steps.endIndex - 1)` instead of `(newStepIndex < progress.currentLeg.steps.endIndex)`. If we check `RouteLegProgress.stepIndex` it has the following assert: `assert(stepIndex >= 0 && stepIndex < leg.steps.endIndex)`. Is it intentional to ignore the last index?

bamx23:
> @bamx23, btw is there a chance that you know when such crash occurs? E.g. was it during active guidance navigation during route refresh or something else? Also how often it happens?

According to the crash log in the issue, it happened during a simulated active guidance session. Crashlytics was broken at that time, but according to Organizer, this happened only once.

1ec5:
> since `LegacyRouteController` is not used anymore do you think it'd be acceptable to remove it now?

It comes up occasionally as a workaround for use cases that MapboxNavigationNative doesn’t yet support fully, such as walking, or as a stopgap when there’s a severe bug in the navigator, but we aren’t actively developing it other than keeping it reasonably stable on a “best effort only” basis.

Sources/MapboxCoreNavigation/RouteController.swift
1ec5, line 286: I think it’s premature to apply this defensive fix. We should look into why the navigator thought the new index would be valid even though the route model didn’t have as many steps. This method may turn out to be merely a cascading failure, which we would want to address with an assertion rather than a no-op.

@bamx23, if it would be possible to track down the route request or navigator history for #2980, that would be helpful for this investigation.

1ec5: That’s what @MaximAlien said above in https://github.com/mapbox/mapbox-navigation-ios/pull/2980#issuecomment-841191115; I didn’t see that earlier. I’m just concerned that we’re papering over this crash even further so that the next time, we’ll see a crash or weird behavior when the leg and intersection indices get updated but the step index does not.

ShanMa1991: Thanks for the suggestion, I'll correct this one

ShanMa1991: Yes. I'm worried that if the `stepIndex` didn't get updated but other indexes are updated, there would be some further problems.

Sources/MapboxCoreNavigation/RouteLegProgress.swift
```diff
@@ -144,8 +144,6 @@ open class RouteLegProgress: Codable {
         self.leg = leg
         self.stepIndex = stepIndex
         
-        precondition(leg.steps.indices.contains(stepIndex), "It's not possible to create RouteLegProgress without any steps or when stepIndex is higher than steps count.")
```
bamx23, line 147: Why was this precondition removed? `didSet` for `stepIndex` won't be triggered in init.

bamx23 (approved):

Sources/MapboxCoreNavigation/RouteLegProgress.swift
ShanMa1991: It's removed because it's after we give value for the `stepIndex `. I'll move it back here.

Issue #2766, opened 2020-12-11
Navigation can stop when transitioning between night and day


johnnewman:
**Mapbox Navigation SDK version: 1.1.0 and 1.2.0-alpha.3**
Xcode 12.2
iOS 14

### Steps to reproduce

 1. Start a navigation session at daylight while simulating along the route.
 2. Open settings and change the time to night.
 3. Reopen the navigation session to transition the night style.
 4. _Repeat, flipping from day and night until simulation stops. (It usually only takes one more flip to day)_

### Expected behavior

Navigation continues running.

### Actual behavior

When navigation stops, you can still interact with the map and there is no indicator that it is frozen/paused. Navigation and voice guidance can be restarted by flipping the time back to day or night one more time.

- [Gist containing the example code](https://gist.github.com/johnnewman/8a2d8c5a3cfec66d1af97bf021219576). This was taken from the SDK installation guide [here](https://docs.mapbox.com/ios/navigation/guides/install/#request-a-route) and tweaked to supply `MapboxNavigationService` and `NavigationOptions` instances to the `NavigationViewController`.
- [Video showing the freeze](https://drive.google.com/file/d/1YBeh4ipJLG7GnEmqVvc7eLMJa67UesLO/view?usp=sharing) (with audio). You can see the distance to the next maneuver keeps counting down in the top left corner, but the puck is frozen.

johnnewman:
Updating title and description because I was able to reproduce this while actually navigating in a vehicle, not just simulating a route.

johnnewman:
I have tried testing this with automatic day/night switching, but the Timer instance that handles this switching [in the StyleManager class](https://github.com/mapbox/mapbox-navigation-ios/blob/main/MapboxNavigation/StyleManager.swift#L135) is not scheduled on a run loop. I think this is another bug because that initializer requires that the Timer be added to a run loop. Or it could be initialized with [`Timer.scheduledTimer(withTimeInterval:repeats:block:)`](https://developer.apple.com/documentation/foundation/timer/2091889-scheduledtimer) to automatically be assigned the current run loop.

When I edit that line in StyleManager to `scheduledTimer(...)`, I'm seeing the map style properly update while simulating a drive, without the need to open the Settings app and manually change the system time. Also after making that timer tweak, I haven't seen this same freeze for sunrise or sunset after a few simulated drives. I don't know what causes the freeze when changing the system time.

johnnewman:
[...]

PR #2995, opened 2021-05-12
Catch-up with `main`.

### LINKED ISSUE #2850: Merge main to release-v2.0
main has diverged from the point at which release-v2.0 branched. To avoid regressions from v1.3.0 to v2.0.0, we need to merge main to release-v2.0. The main source of conflicts will be #2734, once it lands.

MaximAlien:
### Description

Catch-up `release-v2.0` with `main`.

Fixes #2850.


MaximAlien:
There is a chance that https://github.com/mapbox/mapbox-navigation-ios/pull/2881 wasn't merged as expected and fix for https://github.com/mapbox/mapbox-navigation-ios/issues/2766 might not fully work. @azarovalex, it'd be great if you could affirm this.

Example/AppDelegate.swift
```diff
@@ -12,8 +13,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
 
+    @available(iOS 12.0, *)
+    lazy var interfaceController: CPInterfaceController? = nil
+
+    @available(iOS 12.0, *)
+    lazy var carWindow: CPWindow? = nil
+
+    @available(iOS 12.0, *)
+    lazy var sessionConfiguration: CPSessionConfiguration? = nil
+
```
MaximAlien, line 24: @avi-c, do we really need these properties in `AppDelegate.swift`?

1ec5: Good catch, these properties appear to be unused, if I’m not mistaken.

1ec5:
For reference, this file resolves conflicts in the files listed in the commit message for 4128beece06d483f16a5e5dfca21b3fa9e3dc414.

1ec5 (approved):

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -42,8 +42,6 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
         }
     }
     
-    var didChangeAuthorizationIsFirstCalled = true
-
```
jill-cardamon: The delegate method `locationManagerDidChangeAuthorization(_:)` is called at startup and whenever the user changes authorization permissions. `didChangeAuthorizationIsFirstCalled` was to deal with the first case. Without this property, the halo view and enable precise location banner would show up when they weren't supposed to.

MaximAlien: Thanks for flagging that. I've tried on several tests and do not see any issues, at the same time to bring back your implementation from `main` we'll have to possibly bring back `MBNavigationAccuracyAuthorization` and maybe other APIs. I'm inclined to merge this PR as is and create another ticket in scope of which we'll perform more through testing and add fixes if needed.

PR #2999, opened 2021-05-13
Remove optional dependency on MapboxGeocoder.swift for search on CarPlay

jill-cardamon:
### Description
This PR removes the optional dependency on MapboxGeocoder.swift for search on CarPlay #2285 while moving the functionality to the app level.

- [x] Remove code powered by MapboxGeocoder.swift from CarPlaySearchController+CPSearchTemplate.swift
- [x] Replace `GeocodedPlacemark` and `RecentItem` usage
- [x] Move functionality to app level
- [x] Test and fix bugs
- [x] Update CHANGELOG.md, add comments for public properties/methods
- [x] Clean code and remove files that are no longer used

### Implementation
Removed all functions in CarPlaySearchController+CPSearchTemplate.swift that are powered by MapboxGeocoder.swift. Replaced them with functions in `CarPlaySearchControllerDelegate` protocol that now conforms to `CPSearchTemplateDelegate`. These methods are then implemented in AppDelegate+CarPlay.swift where AppDelegate extends CarPlaySearchControllerDelegate. Added `NavGeocodedPlacemark` and `Recentitem` structs to replace functionality of `GeocodedPlacemark` and `RecentItem`.

Blackbox tested in the simulator. Known issue: destination name not showing up after push to ViewController.


Sources/MapboxNavigation/CarPlaySearchController+CPSearchTemplateDelegate.swift
```diff
@@ -6,69 +6,39 @@ import MapboxDirections
...
+    }
+    
     public static let CarPlayGeocodedPlacemarkKey: String = "MBGecodedPlacemark"
     
     static var MaximumInitialSearchResults: UInt = 5
     static var MaximumExtendedSearchResults: UInt = 10
     /// A very coarse location manager used for focal location when searching
-    fileprivate static let coarseLocationManager: CLLocationManager = {
+    // TODO: Confirm that this location manager can become public to use in AppDelegate
+    public static let coarseLocationManager: CLLocationManager = {
```
jill-cardamon, line 26: This CLLocationManager was fileprivate before. I made it public to access within AppDelegate+CarPlay.swift because it was a bad idea to initialize a new CLLocationManager in AppDelegate+CarPlay.swift. Should I actually do that instead of making this public?

Sources/MapboxNavigation/CarPlaySearchController+CPSearchTemplateDelegate.swift
```diff
@@ -159,14 +75,7 @@ extension CarPlaySearchController: CPListTemplateDelegate {
...
 extension GeocodedPlacemark {
-    @available(iOS 12.0, *)
-    func listItem() -> CPListItem {
-        let item = CPListItem(text: formattedName, detailText: subtitle, image: nil, showsDisclosureIndicator: true)
-        item.userInfo = [CarPlaySearchController.CarPlayGeocodedPlacemarkKey: self]
-        return item
-    }
-    
-    var subtitle: String? {
+    open var subtitle: String? {
```
jill-cardamon, line 78: This property is necessary for displaying the address since `Geocoder.shared.geocode()` returns the address as separate from the placemarks. Can I move this to MapboxGeocoder.swift?

Example/AppDelegate+CarPlay.swift
```diff
@@ -153,6 +154,44 @@ extension AppDelegate: CarPlayManagerDelegate {
 
 @available(iOS 12.0, *)
 extension AppDelegate: CarPlaySearchControllerDelegate {
+    struct RecentsHolder {
+        static var _recentItems: [RecentItem] = RecentItem.loadDefaults()
+        static var _recentSearchItems: [CPListItem]? = []
+        static var _recentSearchText: String? = ""
+    }
```
1ec5, line 161: It looks like this struct is meant solely to group the three properties. I think it would be clearer to move the properties to AppDelegate in order to declare them right next to their respective computed properties, unless we expect to need to reuse this state as a whole in some way.

Example/AppDelegate+CarPlay.swift
```diff
@@ -153,6 +154,44 @@ extension AppDelegate: CarPlayManagerDelegate {
...
+    
+    var recentItems: [RecentItem] {
+        get {
+            return RecentsHolder._recentItems
+        }
+        set(newValue) {
+            RecentsHolder._recentItems = newValue
+        }
+    }
```
1ec5, line 170: Do we expect these computed properties to get any more complex in the future, or can we turn them into ordinary stored properties, eliminating the need for RecentsHolder and its properties?

Example/AppDelegate+CarPlay.swift
```diff
@@ -171,6 +210,93 @@ extension AppDelegate: CarPlaySearchControllerDelegate {
...
+        
+        // Search for placemarks using MapboxGeocoder.swift
+        let shouldSearch = searchText.count > 2
+        if shouldSearch {
+            let options = forwardGeocodeOptions(searchText)
+            Geocoder.shared.geocode(options, completionHandler: { [weak self] (placemarks, attribution, error) in
+                guard let strongSelf = self else { return }
+                guard let placemarks = placemarks else {
+                    // TODO: FIX INPUT TO COMPLETION HANDLER
+                    completionHandler((self?.resultsOrNoResults(items, limit: MaximumSearchResults.initial))!)
```
1ec5, line 227: Is this to-do item still needed?

Example/AppDelegate+CarPlay.swift
```diff
@@ -171,6 +210,93 @@ extension AppDelegate: CarPlaySearchControllerDelegate {
...
+        recentSearchItems = items
+        
+        if items.count > 0 {
+            if let limit = limit {
+                return Array<CPListItem>(items.prefix(Int(limit)))
+            }
+            
+            return items
+        } else {
+            let title = NSLocalizedString("CARPLAY_SEARCH_NO_RESULTS", bundle: .mapboxNavigation, value: "No results", comment: "Message when search returned zero results in CarPlay")
```
1ec5, line 295: This localized message moved from MapboxNavigation to Example, so the corresponding entries in the .strings files will need to be updated as well.

Alternatively, would it be possible to keep this “No results” item as a built-in fallback when `resultsOrNoResults(_:limit:)` returns `nil` or an empty array?

Sources/MapboxNavigation/CarPlaySearchController.swift
```diff
@@ -1,13 +1,123 @@
...
+    var recentItems: [RecentItem] { get set }
+    var recentSearchItems: [CPListItem]? { get set }
+    var recentSearchText: String? { get set }
+    
+    func searchTemplate(_ searchTemplate: CPSearchTemplate, updatedSearchText searchText: String, completionHandler: @escaping ([CPListItem]) -> Void)
+    func searchTemplate(_ searchTemplate: CPSearchTemplate, selectedResult item: CPListItem, completionHandler: @escaping () -> Void)
+    func forwardGeocodeOptions(_ searchText: String) -> ForwardGeocodeOptions
+    func selectResult(item: CPListItem, completionHandler: @escaping () -> Void)
+    func recentSearches(_ searchText: String) -> [CPListItem]
+    func resultsOrNoResults(_ items: [CPListItem], limit: UInt?) -> [CPListItem]
```
1ec5, line 23: These methods overuse the `_` keyword. A method name should only omit the first argument’s label if the first part of the method name (e.g., “forwardGeocodeOptions”) describes the first argument. Otherwise, use a preposition or the local variable name as the argument label.

Sources/MapboxNavigation/CarPlaySearchController.swift
```diff
@@ -1,13 +1,123 @@
...
+//    func listTemplate(_ listTemplate: CPListTemplate, didSelect item: CPListItem, completionHandler: @escaping () -> Void)
```
1ec5, line 24: Is this method needed, or can we remove the comment?

Sources/MapboxNavigation/CarPlaySearchController.swift
```diff
@@ -1,13 +1,123 @@
...
+}
+
+// TODO: add comments to all public
+public struct NavGeocodedPlacemark: Equatable, Codable {
```
1ec5, line 28: Remember to add these documentation comments.

Sources/MapboxNavigation/CarPlaySearchController.swift
```diff
@@ -1,13 +1,123 @@
...
+        try container.encode(course, forKey: .course)
+        try container.encode(timestamp, forKey: .timestamp)
+    }
+    
+    convenience init(model: CLLocationModel) {
+        self.init(coordinate: CLLocationCoordinate2DMake(model.latitude, model.longitude), altitude: model.altitude, horizontalAccuracy: model.horizontalAccuracy, verticalAccuracy: model.verticalAccuracy, course: model.course, speed: model.speed, timestamp: model.timestamp)
+    }
+}
+
+struct CLLocationModel: Codable {
```
1ec5, line 112: Is this a workaround for the fact that CLLocation isn’t Codable? If so, it may be simpler to extend CLLocation with Codable conformance.

Sources/MapboxNavigation/CarPlaySearchController+CPSearchTemplateDelegate.swift
1ec5: MapboxGeocoder.swift would be a reasonable place to implement this functionality, since the specific logic is peculiar to the Mapbox Geocoding API. However, as a first step, we can move it to the Example application. The logic is nontrivial and not easily guessable, so we may need to include it in examples until it finds its way into MapboxGeocoder.swift. Presumably the Mapbox Search API has something similar that an application can hook into.

Sources/MapboxNavigation/CarPlaySearchController+CPSearchTemplateDelegate.swift
1ec5: It’s only used for setting the `ForwardGeocodeOptions.focalLocation`, which is configured in the AppDelegate as of this PR. So it would be better to move this property to the AppDelegate. Once #2588 lands, it’ll become redundant to PassiveLocationManager-related notifications, so we’ll be able to remove it entirely.

1ec5 (commented):
In addition to the feedback below, remember to mention the renamed/moved/added public symbols in the changelog.

Example/AppDelegate+CarPlay.swift
1ec5: I didn’t notice that these properties are declared in an extension. It would be fine to declare three ordinary stored properties as part of the main AppDelegate declaration and use them within this extension. We’re already quite from from the point where this file would be self-contained anyways.

Sources/MapboxNavigation/CarPlaySearchController.swift
jill-cardamon: I extended CLLocation to conform to Encodable, but non-convenience initializers must be declared directly in the CLLocation class and not within an extension. Personally, the elegant solution was creating a convenience initializer using a model instead of decoding each CLLocation property individually.

Sources/MapboxNavigation/CarPlaySearchController.swift
jill-cardamon: Removed!

Example/AppDelegate+CarPlay.swift
jill-cardamon: Moving recentSearchItems would require AppDelegate to be available for iOS 12 or later since it is of type `[CPListItem]`.

1ec5: As a workaround, the property could be typed as `[AnyObject]` and cast to `[CPListItem]` where necessary.

Sources/MapboxNavigation/RecentItems.swift
```diff
@@ -1,71 +1,68 @@
-#if canImport(MapboxGeocoder)
 import Foundation
-import MapboxGeocoder
 import CarPlay
 
-struct RecentItem: Codable, Equatable {
-    static func ==(lhs: RecentItem, rhs: RecentItem) -> Bool {
+public struct RecentItem: Equatable, Codable {
```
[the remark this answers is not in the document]
MaximAlien: We'll also have to add it to `jazzy.yml` and document.

Sources/MapboxNavigation/CarPlaySearchControllerDelegate.swift
```diff
@@ -41,4 +42,15 @@ public protocol CarPlaySearchControllerDelegate: AnyObject {
      animated or not.
      */
     func popTemplate(animated: Bool)
+    
+    var recentItems: [RecentItem] { get set }
```
MaximAlien, line 46: We must update these newly added protocol symbols.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -765,6 +764,8 @@
 		8A2081C925E07CED00F9B8A6 /* NavigationMapViewIdentifiers.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = NavigationMapViewIdentifiers.swift; sourceTree = "<group>"; };
 		8A2081CA25E07CED00F9B8A6 /* RouteLineType.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteLineType.swift; sourceTree = "<group>"; };
 		8A285E232637859800B11ECD /* OrnamentsController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = OrnamentsController.swift; sourceTree = "<group>"; };
+		8A2081C925E07CED00F9B8A6 /* NavigationMapViewIdentifiers.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = NavigationMapViewIdentifiers.swift; sourceTree = "<group>"; };
+		8A2081CA25E07CED00F9B8A6 /* RouteLineType.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RouteLineType.swift; sourceTree = "<group>"; };
```
MaximAlien, line 768: This addition looks strange, at a first glance these files do not seem to be needed.

Sources/MapboxNavigation/RecentItems.swift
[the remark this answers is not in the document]
jill-cardamon: Do you mean rename to `RecentItem.swift` without the plural?

MapboxNavigation.xcodeproj/project.pbxproj
jill-cardamon: Hm I haven't touched these files. Maybe there was an issue with the rebase.

Sources/MapboxNavigation/CarPlaySearchControllerDelegate.swift
jill-cardamon: With doc comments?

MaximAlien: Yep, it doesn't have to happen in this PR though. It'd greatly improve readability.

MapboxNavigation.xcodeproj/project.pbxproj
MaximAlien: Let's remove these additions to prevent any issues in future.

Sources/MapboxNavigation/CarPlaySearchController.swift
```diff
@@ -1,6 +1,105 @@
 import CarPlay
+import MapboxGeocoder
 import MapboxDirections
 
+public struct NavGeocodedPlacemark: Equatable, Codable {
```
MaximAlien, line 5: This struct will also have to be documented, probably also makes sense to rename it to `NavigationGeocodedPlacemark`.

Sources/MapboxNavigation/RecentItems.swift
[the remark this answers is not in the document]
MaximAlien: Let's keep `RecentItem` without renaming changes.

Sources/MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -64,30 +64,6 @@ public protocol CarPlayManagerDelegate: AnyObject, UnimplementedLogging {
...
-     
-     - parameter carPlayManager: The CarPlay manager instance.
-     - parameter searchTemplate: The search template currently accepting user input.
-     - parameter item: The search result the user has selected.
-     - parameter completionHandler: Called when the delegate is done responding to the selection.
-     
-     - postcondition: You must call `completionHandler` within this method.
-     */
-    func carPlayManager(_ carPlayManager: CarPlayManager, searchTemplate: CPSearchTemplate, selectedResult item: CPListItem, completionHandler: @escaping () -> Void)
-
```
MaximAlien, line 90: These two delegate method are not really used by `CarPlayManager`.

Example/AppDelegate+CarPlay.swift
```diff
@@ -237,6 +243,164 @@ extension AppDelegate: CarPlaySearchControllerDelegate {
...
+        var allScopes: PlacemarkScope = .all
+        allScopes.remove(.postalCode)
+        options.allowedScopes = allScopes
+        options.maximumResultCount = MaximumSearchResults.extended
+        options.includesRoutableLocations = true
+        
+        return options
+    }
+    
+    func recentSearches(with searchText: String) -> [CPListItem] {
```
MaximAlien, line 260: This method was previously exposed as part of `CarPlaySearchControllerDelegate`, since it's not really used on SDK level I've decided to remove it from there, and it's used only on application level.

MaximAlien:
@jill-cardamon, @1ec5 can you please take a look at the most recent changes? I've re-worked public API of `CarPlaySearchControllerDelegate` to make it a bit easier to understand, some code is used purely on application level now. 

It'd be great to also cover functionality in `RecentItem`, `NavigationGeocodedPlacemark` etc with tests, we can do that as a tail work I guess.

MapboxNavigation.podspec
```diff
@@ -47,6 +47,7 @@ Pod::Spec.new do |s|
   s.dependency "MapboxMaps", "10.0.0-rc.3"
   s.dependency "Solar", "~> 2.1"
   s.dependency "MapboxSpeech-pre", "2.0.0-alpha.1"
+  s.dependency "MapboxGeocoder.swift", "~> 0.10.0"
```
1ec5, line 50: Can we can remove the dependency on MapboxGeocoder.swift for CocoaPods too?

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -101,7 +101,7 @@
 		35726EE81F0856E900AFA1B6 /* DayStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 35726EE71F0856E900AFA1B6 /* DayStyle.swift */; };
 		3582A25020EEC46B0029C5DE /* Router.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3582A24F20EEC46B0029C5DE /* Router.swift */; };
 		3582A25220EFA9680029C5DE /* RouterDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3582A25120EFA9680029C5DE /* RouterDelegate.swift */; };
-		358E31D622562698009B3EC2 /* CarPayCompassView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 358E31D522562697009B3EC2 /* CarPayCompassView.swift */; };
+		358E31D622562698009B3EC2 /* CarPlayCompassView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 358E31D522562697009B3EC2 /* CarPlayCompassView.swift */; };
```
1ec5, line 104: Heh, nice catch. CarPay will have to wait. 😛

Sources/MapboxNavigation/NavigationGeocodedPlacemark.swift
```diff
@@ -0,0 +1,161 @@
...
+        try container.encodeIfPresent(routableLocations, forKey: .routableLocations)
+    }
+    
+    public static func ==(lhs: NavigationGeocodedPlacemark, rhs: NavigationGeocodedPlacemark) -> Bool {
+        return lhs.title == rhs.title &&
+            lhs.address == rhs.address
+    }
+    
+    @available(iOS 12.0, *)
+    public func listItem() -> CPListItem {
```
1ec5, line 87: Remember to add some documentation about this public method.

1ec5 (approved):

MapboxNavigation.podspec
MaximAlien: This is actually needed to be able to successfully build Mapbox Navigation. We also link `MapboxGeocoder` for SPM.

PR #2991, opened 2021-05-10
Update NN to 49.0.1, RoadObject overhaul

chezzdev:
### Description

This PR updates NavigationNative to 49.0.1 as long as provides access to reworked APIs for `RoadObject`, `RouteAlert` and related classes.

### Implementation

Public constructors were added to many EH types in an attempt to make SDK user testing easier (per request from @bamx23).

Fixes: #2868 

## Changelog entries for beta-release:
* Added public initializers for `RoadGraphEdge`, `RoadGraphEdgeMetadata`, `RoadGraphPath`, `RoadGraphPosition`, and `RouteObjectEdgeLocation`. ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
* Removed `Alert` enum, and `alert`, `distance`, `length`, `beginCoordinate`, `endCoordinate`, `beginSegmentIndex`, and `endSegmentIndex` properties from `RouteAlerts`. ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
* Added `RoadObject` a property to `RouteAlerts`. ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
* Added `RoadObject` struct which describes the object on the road. ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
* `RoadObjectDistanceInfo` is now an enum which contains information about distance to the road object of a concrete type/shape (gantry, polygon, line, point, subgraph). ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
* Added `RoadObjectPosition` struct which contains information about the position of the object on the graph and its geo-position. ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
* `RoadObjectLocation` is now an enum that represents the location of a road object in the road graph of a concrete type/shape. ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
* Added `OpenLROrientation` struct which describes the relationship between the road object and the referenced line. Used for `OpenLRLineLocation` struct. ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
* Added `OpenLRSideOfRoad` struct which describes the relationship between the road object and the road. Used for `OpenLRPointLocation` struct. ([#2991](https://github.com/mapbox/mapbox-navigation-ios/pull/2991))
[...]

PR #3002, opened 2021-05-13
Add code documentation

jill-cardamon:
### Description
This PR adds code documentation for #2991.

### DIFF
--- a/Sources/MapboxCoreNavigation/Distance/GantryDistanceInfo.swift
+++ b/Sources/MapboxCoreNavigation/Distance/GantryDistanceInfo.swift
@@ -9,6 +9,9 @@ public struct GantryDistanceInfo {
     /** Distance to the gantry object */
     public let distance: CLLocationDistance
 
+    /**
+     Initializes a new `GantryDistanceInfo` object with a given distance to the object.
+     */
     public init(distance: CLLocationDistance) {
         self.distance = distance
     }
--- a/Sources/MapboxCoreNavigation/Distance/GantryDistanceInfo.swift
+++ b/Sources/MapboxCoreNavigation/Distance/GantryDistanceInfo.swift
@@ -10,7 +10,7 @@ public struct GantryDistanceInfo {
     public let distance: CLLocationDistance
 
     /**
-     Initializes a new `GantryDistanceInfo` object with a given distance to the object.
+     Initializes a new `GantryDistanceInfo` object with a given distance to the gantry object.
      */
     public init(distance: CLLocationDistance) {
         self.distance = distance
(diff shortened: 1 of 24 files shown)


Sources/MapboxCoreNavigation/Distance/GantryDistanceInfo.swift
```diff
@@ -9,6 +9,9 @@ public struct GantryDistanceInfo {
     /** Distance to the gantry object */
     public let distance: CLLocationDistance
 
+    /**
+     Initializes a new `GantryDistanceInfo` object with a given distance to the object.
```
1ec5, line 13: I think the second “object” probably refers to a road object, as opposed to a Swift object. This might be less confusing:

```suggestion
     Initializes a new `GantryDistanceInfo` object with a given distance to the gantry.
```

Sources/MapboxCoreNavigation/Distance/LineDistanceInfo.swift
```diff
@@ -12,18 +12,27 @@ public struct LineDistanceInfo {
...
-    /** Distance along the road graph from current position to end of road object */
+    /** Distance along the road graph from current position to end of the most distant exit */
     public let distanceToEnd: CLLocationDistance
 
-    /** If we enter road object from it's start, if already "within" object - always false */
+    /** If we enter road object from its start, if already "within" object - always false */
     public let isEntryFromStart: Bool
 
-    /** Length of "long" objects */
+    /** Length of objects, measured in meters. */
```
1ec5, line 24:
```suggestion
    /** Length of the road object, measured in meters. */
```

(Also below in the initializer’s documentation comment.)

Sources/MapboxCoreNavigation/Distance/PointDistanceInfo.swift
```diff
@@ -6,9 +6,12 @@ import MapboxNavigationNative
  */
 public struct PointDistanceInfo {
 
-    /** Distance distance to the point object */
+    /** Distance to the point object, measure in meters. */
```
1ec5, line 9:
```suggestion
    /** Distance to the point object, measured in meters. */
```

Sources/MapboxCoreNavigation/Distance/RoadObjectDistanceInfo.swift
```diff
@@ -7,10 +7,29 @@ import MapboxNavigationNative
  * type/shape (gantry, polygon, line, point etc.).
  */
 public enum RoadObjectDistanceInfo {
+    /**
+     The information about distance to the road object represented as a point.
+     */
     case point(PointDistanceInfo)
```
1ec5, line 13: Once we inline these “info” types as associated values per https://github.com/mapbox/mapbox-navigation-ios/pull/2991#discussion_r631186845, those structs’ properties’ documentation comments will also be inlined here.

1ec5 (approved):
A few nits but overall this takes care of all the documentation-related feedback in #2991.

Sources/MapboxCoreNavigation/Distance/GantryDistanceInfo.swift
jill-cardamon: I think using `road object` might be less confusing.

PR #3004, opened 2021-05-14
Custom road objects for EH

chezzdev:
### Description

The PR provides APIs to work with custom road objects. 
The flow for custom road objects is the following:
- match a road object using `RoadObjectMatcher` by providing info about a road object in an appropriate format;
- receive matched road object via `RoadObjectMatcherDelegate` or handle the matching error;
- add matched road object to `RoadObjectsStore` to enable its tracking in the electronic horizon;
- remove road object by its identifier from `RoadObjectsStore` when you don't need it anymore.

Fixes https://github.com/mapbox/navigation-sdks/issues/921.


Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
```diff
@@ -0,0 +1,116 @@
...
+            }
+            delegate?.didMatchRoadObject(result: .success(RoadObject(roadObject)))
+        } else if roadObject.isError(), let error = roadObject.error {
+            guard let error = error as? MapboxNavigationNative.RoadObjectMatcherError else {
+                preconditionFailure("Road object matcher error can't be constructed. Unknown error type.")
+            }
+            delegate?.didMatchRoadObject(result: .failure(RoadObjectMatcherError(error)))
+        } else {
+            preconditionFailure("Expected type is neither a value nor an error.")
+        }
```
S2Ler, line 114: Looks like a patter which likely to be used in other places as well. I would consider creating a converter from `MBXExpected` to `Swift.Result`. One way of doing this is creating an init method on `Swift.Result`.

chezzdev: Nice idea. We'll still need to convert from one `Result` to another, but that's a much more readable code. Another catch is that we'll need to make an `Error`-conformance of native type, but that shouldn't be a big hit on usability.

S2Ler: We could implement a more trickery approach by providing "conversion" methods to `Result.init(expected)`, but I don't think it worth it. Your solution is good enough, which can be improved upon if needed.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
```diff
@@ -0,0 +1,113 @@
...
+    public func cancel(identifier: RoadObjectIdentifier) {
+        native.cancel(forId: identifier)
+    }
+
+    init(_ native: MapboxNavigationNative.RoadObjectMatcher) {
+        self.native = native
+    }
+
+    deinit {
+        native.setListenerFor(nil)
```
azarovalex, line 94: Shouldn't NavNative handle this?

azarovalex (approved):
LGMT.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcherDelegate.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ `RoadObjectMatcher` delegate.
+ */
+public protocol RoadObjectMatcherDelegate: class {
```
MaximAlien, line 7: Probably makes sense to update to:
```suggestion
public protocol RoadObjectMatcherDelegate: AnyObject {
```

to prevent build warnings in Xcode `12.5`.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
chezzdev: Do you mean whether listeners are held weakly in NavNative? I'm not counting on that, but let's ask @SiarheiFedartsou

Sources/MapboxCoreNavigation/EHorizon/OpenLRStandard.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+import MapboxNavigationNative
+
+/** Standard of OpenLR */
+public enum OpenLRStandard {
```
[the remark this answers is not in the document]
chezzdev: Good idea. I've created https://github.com/mapbox/mapbox-navigation-ios/issues/3012 to remember about it.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
```diff
@@ -56,6 +56,36 @@ public final class RoadObjectsStore {
...
+        }
+        native.addCustomRoadObject(for: nativeObject)
+    }
+
+    /**
+     Removes road object and stops tracking it in the electronic horizon.
+
+     - parameter identifier: Identifier of the road object that should be removed.
+     */
+    public func removeCustomRoadObject(identifier: RoadObjectIdentifier) {
```
1ec5, line 78: Rename to `removeRoadObject(with:)` or `unregisterRoadObject(with:)`. In this case, the road object itself isn’t being passed in, so it needs to be named explicitly in the method name. The argument type is already an “identifier”, so there’s an opportunity to clarify the relationship between the road object and the identifier, though `identifier` as a label wouldn’t be incorrect.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
```diff
@@ -56,6 +56,36 @@ public final class RoadObjectsStore {
...
+        native.removeCustomRoadObject(forId: identifier)
+    }
+
+    /**
+     Removes all custom road objects from the store and stops tracking them in the electronic horizon.
+     */
+    public func removeAllCustomRoadObjects() {
```
1ec5, line 85: Rename to `removeAllRoadObjects()` or, if we really need to emphasize that only application-provided road objects are removed, then `removeAllUserDefinedRoadObjects()`.

Sources/MapboxCoreNavigation/EHorizon/OpenLRStandard.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     [TomTom OpenLR](http://www.openlr.org/).
+
+     Supported references: line location, point along line, polygon.
+     */
+    case tomTom
```
1ec5, line 12: Consider a vendor-neutral case name: TomTom is one member of the OpenLR Association that defines this standard.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
```diff
@@ -56,6 +56,36 @@ public final class RoadObjectsStore {
...
 
+    /**
+     Adds a road object to be tracked in the electronic horizon. In case if an object with such identifier already exists, updates it.
+     NB: road objects obtained from route alerts cannot be added via this API.
+
+     - parameter roadObject: Custom road object, acquired from `RoadObjectMathcer`.
+     */
+    public func addCustomRoadObject(_ roadObject: RoadObject) {
+        guard let nativeObject = roadObject.native else {
+            assertionFailure("You can only add matched custom road objects, acquired from RoadObjectMatcher.")
```
bamx23, line 67: It might be better to make the method `throwing` and throw an error here. Similar to what Maps do for custom style layers/sources

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
```diff
@@ -56,6 +56,36 @@ public final class RoadObjectsStore {
         return native.getRoadObjectIdsByEdgeIds(forEdgeIds: edgeIdentifiers.map(NSNumber.init))
     }
 
+    /**
+     Adds a road object to be tracked in the electronic horizon. In case if an object with such identifier already exists, updates it.
+     NB: road objects obtained from route alerts cannot be added via this API.
+
+     - parameter roadObject: Custom road object, acquired from `RoadObjectMathcer`.
+     */
+    public func addCustomRoadObject(_ roadObject: RoadObject) {
```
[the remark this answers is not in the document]
chezzdev: Looks like "user-defined" won't be accurate at all times, as these objects may be manipulated by the SDK client but not the end-user. This is something not provided out-of-the-box, not necessarily user-initiated. I see Maps are using the "custom" term for styles and events which is semantically similar to the context of road objects.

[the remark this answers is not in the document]
chezzdev: `add` looks OK without such specification, however both `remove` and `removeAll` will only work for custom objects.

Sources/MapboxCoreNavigation/EHorizon/OpenLRStandard.swift
chezzdev: Agreed to leave as is.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcherError.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+import MapboxNavigationNative
+
+/**
+ An error that occures during road object matching.
+ */
+public struct RoadObjectMatcherError: Error {
+
+    /** Description of the error. */
+    public let description: String
```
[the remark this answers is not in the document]
chezzdev: Do you think it makes sense even if we don't provide localized message?

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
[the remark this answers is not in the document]
1ec5: > Looks like "user-defined" won't be accurate at all times, as these objects may be manipulated by the SDK client but not the end-user.

I should’ve clarified: “user-defined” in the API would refer to the developer, not the end user, as in `userInfo` in notifications.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
1ec5: Also, I think `preconditionFailure(_:)` would be more specific in this case.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
```diff
@@ -56,6 +56,36 @@ public final class RoadObjectsStore {
         return native.getRoadObjectIdsByEdgeIds(forEdgeIds: edgeIdentifiers.map(NSNumber.init))
     }
 
+    /**
+     Adds a road object to be tracked in the electronic horizon. In case if an object with such identifier already exists, updates it.
+     NB: road objects obtained from route alerts cannot be added via this API.
+
+     - parameter roadObject: Custom road object, acquired from `RoadObjectMathcer`.
```
1ec5, line 63:
```suggestion
     - parameter roadObject: Custom road object, acquired from `RoadObjectMatcher`.
```

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     Matches given polyline to the graph.
+     Polyline should define a valid path on the graph,
+     i.e. it should be possible to drive this path according to traffic rules.
+
+     - parameter polyline: Polyline representing the object.
+     - parameter identifier: Unique identifier of the object.
+     */
+    public func matchPolyline(_ polyline: [CLLocation], identifier: RoadObjectIdentifier) {
```
1ec5, line 42: For consistency, this method should be called `match(_:identifier:)` and accept a `Turf.LineString`.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     Matches a given polygon to the graph.
+     "Matching" here means we try to find all intersections of the polygon with the road graph
+     and track distances to those intersections as distance to the polygon.
+
+     - parameter polygon: Polygon representing the object.
+     - parameter identifier: Unique identifier of the object.
+     */
+    public func matchPolygon(_ polygon: [CLLocation], identifier: RoadObjectIdentifier) {
```
1ec5, line 54: For consistency, this method should be called `match(_:identifier:)` and accept a `Turf.Polygon`. I guess MapboxNavigationNative doesn’t yet support holes in polygons, but using Polygon would allow us to add support later on without breaking backwards compatibility.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
```diff
@@ -0,0 +1,113 @@
...
+        native.matchGantry(forGantry: gantry, id: identifier)
+    }
+
+    /**
+     Matches given point to road graph.
+
+     - parameter point: Point representing the object.
+     - parameter identifier: Unique identifier of the object.
+     */
+    public func matchPoint(_ point: CLLocationCoordinate2D, identifier: RoadObjectIdentifier) {
```
1ec5, line 76: For consistency, this method should be called `match(_:identifier:)` and accept a `CLLocationCoordinate2D` as it does now.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     Matches given gantry (i.e. polyline orthogonal to the road) to the graph.
+     "Matching" here means we try to find all intersections of the gantry with the road graph
+     and track distances to those intersections as distance to the gantry.
+
+     - parameter gantry: Gantry representing the object.
+     - parameter identifier: Unique identifier of the object.
+     */
+    public func matchGantry(_ gantry: [CLLocation], identifier: RoadObjectIdentifier) {
```
1ec5, line 66: This method should be called `match(gantry:identifier:)` and accept a `Turf.LineString`.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
chezzdev: I used `assertionFailure` intentionally not to crash in Release configurations and identify that such state is non-fatal.
`preconditionFailure` is about non-recoverable failures.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
chezzdev: `Turf.Polygon` is represented as 2-d array of `CLLLocation`. It's not very clear would flattening to 1-d array always produce expected results.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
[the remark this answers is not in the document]
chezzdev: Yeah, we can interpret user as developer, but Cocoa APIs I've checked mostly imply end-user (user defaults - user settings, UserActivity, UserLocation, etc.). I'd say "custom" is a more descriptive term, but that I guess is a matter of taste.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
1ec5: Ah, OK, does it need to be an array of `CLLocation`s though? That would imply that things like timestamps and headings are respected. I’m pretty sure the navigator just uses CLLocation because that’s what the automated Objective-C header generation tool uses. An array of `CLLocationCoordinate2D` or a `Turf.MultiPoint` would be more intuitive.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
1ec5: I think we should still have this method accept a Polygon with the caveat that only the outer ring is respected at the moment.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcherError.swift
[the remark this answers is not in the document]
chezzdev: I left the field while implementing `errorDescription`.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
chezzdev: Chose `preconditionFailure` as the case is only possible if the user passes `RoadObject` created with public init, which is a programming error, not a state that should or can be handled.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
chezzdev: `Polygon.outerRing` is used here.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
chezzdev: `Turf.MultiPoint` is used here.

1ec5 (approved):

Sources/MapboxCoreNavigation/EHorizon/RoadObjectsStore.swift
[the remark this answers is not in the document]
1ec5: It may be that only “user-defined” is used this way, but not “user” in general. Xcode also has a “user-defined runtime attributes” feature that is actually defined by developers.

PR #3006, opened 2021-05-14
Asynchronous navigation status

### LINKED ISSUE #2846: Migrate to asynchronous navigator status
RouteController exposes the current route progress and snapped location synchronously, but MapboxNavigationNative’s Navigator class’s synchronous status accessor is apparently deprecated in favor of an asynchronous, callback-based interface. We’ll need some way to bridge that expectation, perhaps by polling and storing the latest value.

https://github.com/mapbox/mapbox-navigation-ios/blob/0f45ff22cc0702b9d2246bdb113d14aad7f53a37/Sources/MapboxCoreNavigation/Navigator.swift#L5-L9 https://github.com/mapbox/mapbox-navigation-ios/blob/0f45ff22cc0702b9d2246bdb113d14aad7f53a37/Sources/MapboxCoreNavigation/RouteController.swift#L110-L120

/cc @mapbox/navigation-ios

azarovalex:
### Description
Fixes #2846.

### Implementation
Currently, this branch is forked from `pad-custom-objects`, because I used NavNative API from 49.0.0.

#### PassiveLocationDataSource implementation:
Adds itself as an observer in the initializer for the whole lifetime of the object. Whenever a new location is passed to this object, it passes it to the NavNative and waits for the callback. Since NavNative returns callback on a background thread, we need to dispatch back on the main thread. We also need to store the latest known location to use it when NavNative calls the callback. This potentially can create inconsistency, however, the delay should be negligible.

#### RouteController implementation
Also adds itself as an observer in the initializer, however in this case it can possible lead to problems because corresponding NavigationService can be stopped, but the RouteController will continue to receive status updates from NavNative that were triggered by other objects. However, here I assume that we will only have one active guidance at a time. Another solution is for NavigationService to subscribe and unsubscribe Router from NavNative statuses but in this case, NavNative dependency will leak from RouteController to NavigationService.
The implementation here is similar to what we have in PassiveLocationDataSource, but we also need to rewrite `updateRouteLeg` and `userIsOnRoute` functions to use the latest known NavNative status, instead of fetching a new one.

### Questions to this solution
The PR looks similar to what Android folks have, my implementation also works at a first glance, however I still have some questions: 
1. NavigationService can be paused, should we unsubscribe Router from the NavNative in this case? In ideal world we should, however NavigationService uses Router protocol and they both don’t know anything about NavNative and concrete RouteController implementation. I don’t want to add functions like `start` and `stop` to Router, and I assume we will always have only one active guidance at a time.
2. Function `userIsOnRoute` now works a little bit different, consider this case:
```
let location = some random location
let router = create router
router.locationManager(_:..., didUpdateLocations: [location]) // the very first location update
router.isUserOnRoute(location)
```
The last line now returns false instead of true, because the navigation status update is asynchronous and we haven't received the callback yet.

### Checklist
- [x] Answer all questions
- [x] Wait for UI glitches to be fixed in `release-v.2.0` branch and retest the solution
- [x] Update the corresponding tests (CI builds fail at the moment)


Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -239,8 +238,8 @@ open class RouteController: NSObject {
     /// updateRouteLeg is used to notify nav-native of the developer changing the active route-leg.
     private func updateRouteLeg(to value: Int) {
         let legIndex = UInt32(value)
-        if navigator.changeRouteLeg(forRoute: 0, leg: legIndex), let timestamp = location?.timestamp {
-            updateIndexes(status: navigator.status(at: timestamp), progress: routeProgress)
+        if navigator.changeRouteLeg(forRoute: 0, leg: legIndex), let status = navigator.getStatus() {
```
1ec5, line 241: `Navigator.getStatus()` may be less safe than the preexisting call to `getStatusForMonotonicTimestampNanoseconds(_:)`, which we introduced in #2648 and relied on as a fix for #2583 and #2113. https://github.com/mapbox/mapbox-navigation-ios/issues/2846#issuecomment-824996423 proposed stashing away a copy of the status whenever we get it from NavigatorObserver callback; then we could use it directly and synchronously here.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -29,6 +29,7 @@ open class PassiveLocationDataSource: NSObject {
         super.init()
         
         self.systemLocationManager.delegate = self
+        navigator.addObserver(for: self)
```
1ec5, line 32: This implementation makes both PassiveLocationDataSource and RouteController individually register as observers, which leads to the questions about observation lifetime that you raised above. Since `MapboxCoreNavigation.Navigator` is already mirroring `MapboxNavigationNative.Navigator` to some extent, can we have the former register as an observer for the latter and broadcast updates to PassiveLocationDataSource and RouteController via notifications, similar to the existing progress change notifications?

Sources/MapboxCoreNavigation/RouteController.swift
azarovalex: `Navigator.getStatus()` documentation states that it returns the last passed to NavigatorObserver status, so I don't think we should stash away a copy.

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
azarovalex: I don't see how this can solve the lifetime problem, RouteController anyway will have to subscribe to the updates.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -70,7 +70,14 @@ class Navigator {
      Provides a new or an existing `MapboxCoreNavigation.Navigator` instance. Upon first initialization will trigger creation of `MapboxNavigationNative.Navigator` and `HistoryRecorderHandle` instances,
      satisfying provided configuration (`tilesVersion` and `tilesURL`).
      */
-    static let shared: Navigator = Navigator()
+    static var shared: Navigator {
+        return _navigator
+    }
+    
+    // Used in tests to recreate the navigator
+    static var _navigator: Navigator = .init()
```
azarovalex, line 78: Used only in tests before the bug in NavNative will be fixed:
https://github.com/mapbox/mapbox-navigation-native/pull/3110

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -180,3 +188,41 @@ extension Navigator: ElectronicHorizonObserver {
         NotificationCenter.default.post(name: .electronicHorizonDidPassRoadObject, object: nil, userInfo: userInfo)
     }
 }
+
+extension Navigator: NavigatorObserver {
+    func onStatus(for origin: NavigationStatusOrigin, status: NavigationStatus) {
+        guard origin == .locationUpdate else { return }
```
azarovalex, line 194: Other possible origins are `legChange`, `setRoute` and `unconditional`. 
I filter them here to save the same behavior as we had before.

Sources/MapboxCoreNavigation/Date.swift
```diff
@@ -15,6 +15,6 @@ extension Date {
     
     var nanosecondsSince1970: Double {
         // UnitDuration.nanoseconds requires iOS 13
-        return timeIntervalSince1970 * 1e6
+        return timeIntervalSince1970 * 1e9
```
azarovalex, line 18: This was an old bug that hasn't affect anything, but now this line matters.

Sources/MapboxCoreNavigation/FixLocation.swift
```diff
@@ -10,7 +10,7 @@ extension FixLocation {
         }
         
         self.init(coordinate: location.coordinate,
-                  monotonicTimestampNanoseconds: 0, // use time instead; see also `Navigator.status(at:)`
+                  monotonicTimestampNanoseconds: Int64(location.timestamp.nanosecondsSince1970),
```
azarovalex, line 13: NavNative explicitly requested this change.

Sources/MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -184,7 +184,7 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
         return true
     }
     
-    public func userIsOnRoute(_ location: CLLocation) -> Bool {
+    public func userIsOnRoute(_ location: CLLocation) -> Bool? {
```
azarovalex, line 187: We receive navigation statuses asynchronously, so it's possible that navigation status is still unknown.
Now we return `nil` in this case. This function is only used in the rerouting logic.

Sources/MapboxCoreNavigation/FixLocation.swift
1ec5: Does the navigator expect `monotonicTimestampNanoseconds` to be monotonic? It won’t be in some cases like when moving across time zone boundaries. As I understand it, MapboxNavigationNative added support for monotonic timestamps for Android location providers in particular, which are even more likely to have their normal timestamps go back and forth.

A monotonic timestamp would come from the system, not from a particular location. However, on iOS, that monotonic timestamp would be less informative than a normal location timestamp. It’s much, much more common for Core Location to give us multiple batched location updates with the same timestamp. It’s no problem if the navigator is using `monotonicTimestampNanoseconds` as essentially an alias for `time` (but converted to nanoseconds), but we should make sure MapboxNavigationNative isn’t making any assumptions about `monotonicTimestampNanoseconds` that don’t hold with regard to `CLLocation.timestamp`.

/cc @SiarheiFedartsou @kmadsen

Sources/MapboxCoreNavigation/LegacyRouteController.swift
1ec5: Maybe we should just inline the `true` fallback here instead of scattering it in the call sites, where they could get inconsistent in the future? If we do have to make this change to the public Router protocol, let’s mention it in the changelog.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -254,13 +254,25 @@ open class RouteController: NSObject {
         rawLocation = location
         
         locations.forEach { navigator.updateLocation(for: FixLocation($0)) }
+    }
+    
+    @objc private func handleNavigationStatusNotification(_ notification: NSNotification) {
```
1ec5, line 259: Nit: Avoid “handle” in method names if possible; it’s a sign that the method could be named more descriptively or decomposed further. A common pattern for notification responder methods is e.g. `navigationStatusDidChange(_:)`.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -254,13 +254,25 @@ open class RouteController: NSObject {
...
+        guard let userInfo = notification.userInfo,
+              let status = userInfo[Navigator.NotificationUserInfoKey.statusKey] as? NavigationStatus else { return }
+        DispatchQueue.main.async { [weak self] in
+            self?.handleNavigationStatusUpdate(status)
+        }
+    }
 
-        let status = navigator.status(at: location.timestamp)
-        
+    private func handleNavigationStatusUpdate(_ status: NavigationStatus) {
```
1ec5, line 267: Rename to `update(to:)`.

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -30,8 +30,9 @@ class MapboxCoreNavigationTests: XCTestCase {
     
     override func tearDown() {
         super.tearDown()
+        navigation = nil
         UserDefaults.resetStandardUserDefaults()
-        Navigator.shared.navigator.resetRideSession()
+        Navigator._recreateNavigator()
```
1ec5, line 35: Is it safe to use this underscore method? Is it intended to be used by the SDK or is it an implementation detail subject to change without notice?

Sources/MapboxCoreNavigation/Date.swift
1ec5: How embarrassing! Can’t wait to use Measurement more consistently once we can require newer iOS versions.

1ec5 (approved):
The improvements to the tests are especially welcome. I’m a bit concerned about how we’re calling a non-monotonic timestamp a monotonic timestamp. If it works, that’s great, but longer-term, how do we know MapboxNavigationNative won’t introduce an assumption about monotonicity that breaks on iOS? Perhaps a test of some non-monotonic edge cases could give us some confidence on that front.

Sources/MapboxCoreNavigation/FixLocation.swift
SiarheiFedartsou: > Does the navigator expect monotonicTimestampNanoseconds to be monotonic? 

It is desirable, but not required. The only thing that we shouldn't have significant "jumps" backwards in time.

>  As I understand it, MapboxNavigationNative added support for monotonic timestamps for Android location providers in particular, which are even more likely to have their normal timestamps go back and forth.

On Android they have special field in Location which is based on system uptime and is guaranteed to be monotonic.

> but we should make sure MapboxNavigationNative isn’t making any assumptions about monotonicTimestampNanoseconds that don’t hold with regard to CLLocation.timestamp.

Actually at the moment if `monotonicTimestampNanoseconds` is zero we simply convert `time` field to it and use "as is". This change was required, because some new parts of NN rely on the fact that you fill this `monotonicTimestampNanoseconds` field. Actually we consider `time` as just additional field which could help in debug for example, but `monotonicTimestampNanoseconds` should be used instead.

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
azarovalex: This method was introduced only temporarily, due to the bug in the resetRideSession method.
This bug will be fixed here: https://github.com/mapbox/mapbox-navigation-native/pull/3110

azarovalex:
I do not observe any visual differences in route simulation between this branch and `release-v2.0` branch, all tests pass normally, so I will merge the PR.

PR #3007, opened 2021-05-14
Application allowed to include user metadata in feedback events

jill-cardamon:
### Description
This PR allows the application to include optional user metadata in feedback events.

- [x] add `appMetadata` property to feedback events
- [x] make `appMetadata` optional
- [x] update changelog

### Implementation
Added optional `NavigationEventDetails.appMetadata` and `EventDetails.appMetadata` properties as optional dictionaries that is set on `SessionState`'s `sessionId`, `name`, `version`, and `userId` properties. `NavigationEventsManager.recordFeedback()` now has an optional boolean parameter `withAppMetadata` to include this user metadata.


CHANGELOG.md
```diff
@@ -75,6 +75,7 @@
 * Fixed an issue where lane guidance icons would sometimes highlight the wrong arrow.([#2942](https://github.com/mapbox/mapbox-navigation-ios/pull/2942))
 * The duration annotations added by the `NavigationMapView.showRouteDurations(along:)` method are now set in the fonts you specify using the `NavigationMapView.routeDurationAnnotationFontNames` property. Use this property to specify a list of fallback fonts for better language support. ([#2873](https://github.com/mapbox/mapbox-navigation-ios/pull/2873))
 * Removed deprecated and obsoleted `EventsManager` typealias, `StatusView.delegate`, `StatusView.canChangeValue`, `RouteLegProgress.upComingStep`, `StatusViewDelegate`, `DeprecatedStatusViewDelegate` and `BottomBannerViewController.init(delegate:)`. ([#2993](https://github.com/mapbox/mapbox-navigation-ios/pull/2993))
+* `NavigationEventsManager` now allows the application to include application metadata `appMetadata` such as the application name and version in feedback events by setting `withAppMetadata` parameter of  `NavigationEventsManager.recordFeedback()` to `true`. ([#3007](https://github.com/mapbox/mapbox-navigation-ios/pull/3007)).
```
MaximAlien, line 78: Overall looks good to me. Would it be possible to implement basic unit-test for newly implemented functionality? Please also rebase onto `release-v2.0` to fix outstanding conflicts.

MaximAlien (approved):

S2Ler:
I see that `MapboxNavigationTests.ImageDownloaderTests.testDownloadingAnImage` fails time to time. Would be good to check what's wrong.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -24,6 +24,12 @@ open class NavigationEventsManager {
     
     weak var dataSource: EventsManagerDataSource?
     
+    /**
+     Optional app metadata that can be used to associate app state, specifically app name and version, with events in the telemetry pipeline.
+     Specifically, developers can provide application name and version, user identifier, and session identifier.
```
1ec5, line 29: The developer is the audience, so we can address them directly using the second person. It’s also a bit awkward to talk about the telemetry “pipeline”, since the telemetry backend isn’t accessible to the developer.

```suggestion
     Optional application metadata that that can help Mapbox more reliably diagnose problems that occur in the SDK.
     
     For example, you can provide your application’s name and version, a unique identifier for the end user, and a session identifier.
```

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -300,14 +310,15 @@ open class NavigationEventsManager {
...
+     @param withAppMetadata An optional dictionary used to include application metadata.
+     @return Returns a UUID used to identify the feedback event.
      
      If you provide a custom feedback UI that lets users elaborate on an issue, you should call this before you show the custom UI so the location and timestamp are more accurate.
      
      You can then call `updateFeedback(uuid:type:source:description:)` with the returned feedback UUID to attach any additional metadata to the feedback.
      */
-    public func recordFeedback(type: FeedbackType = .general, description: String? = nil) -> UUID? {
-        return enqueueFeedbackEvent(type: type, description: description)
+    public func recordFeedback(type: FeedbackType = .general, description: String? = nil, withAppMetadata: [String: String?]? = nil) -> UUID? {
```
1ec5, line 320: Similarly, the “with” is unnecessary here.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -147,13 +153,17 @@ open class NavigationEventsManager {
         return eventDictionary
     }
     
-    func navigationFeedbackEvent(type: FeedbackType, description: String?) -> NavigationEventDetails? {
+    func navigationFeedbackEvent(type: FeedbackType, description: String?, withAppMetadata: [String:String?]? = nil) -> NavigationEventDetails? {
         guard let dataSource = dataSource, let sessionState = sessionState else { return nil }
-
-        var event = NavigationEventDetails(dataSource: dataSource, session: sessionState, defaultInterface: usesDefaultUserInterface)
+        if (withAppMetadata != nil) {
+            appMetadata = withAppMetadata
```
1ec5, line 159: Is the intent that passing in metadata would update the cached metadata but omitting it would reuse whatever had been passed in before? If so, I think it would be clearer to move this statement to the call sites.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
jill-cardamon: I'll make this change.

CHANGELOG.md
```diff
@@ -98,6 +98,7 @@
 * Fixed an issue where the current road name label sometimes displayed the name of an intersecting road instead of the current road or blinked in and out. ([#3019](https://github.com/mapbox/mapbox-navigation-ios/pull/3019))
 * Fixed an issue where offline route calculation might hang up. ([#3040](https://github.com/mapbox/mapbox-navigation-ios/pull/3040))
 * Fixed the moment of custom feedback event creation. ([#3049](https://github.com/mapbox/mapbox-navigation-ios/pull/3049))
+* `NavigationEventsManager` now allows the application to include application metadata `NavigationEventsManager.userInfo` such as the application name and version in feedback events by including `userInfo` parameter of  `NavigationEventsManager.recordFeedback()`. ([#3007](https://github.com/mapbox/mapbox-navigation-ios/pull/3007)).
```
1ec5, line 101: Technically this PR renames some symbols. It’s pretty verbose, but jazzy requires us to include the fully qualified method name:

```suggestion 
* Renamed `NavigationEventsManager(dataSource:session:defaultInterface:)` to `NavigationEventsManager(dataSource:session:defaultInterface:userInfo:)` and `NavigationEventsManager.recordFeedback(type:description:)` to `NavigationEventsManager.recordFeedback(type:description:userInfo:)`. The new optional argument contains application metadata, such as the application name and version, that is included in each event to help Mapbox triage and diagnose unexpected behavior. ([#3007](https://github.com/mapbox/mapbox-navigation-ios/pull/3007)).
```

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -300,14 +307,18 @@ open class NavigationEventsManager {
...
      
      If you provide a custom feedback UI that lets users elaborate on an issue, you should call this before you show the custom UI so the location and timestamp are more accurate.
      
      You can then call `updateFeedback(uuid:type:source:description:)` with the returned feedback UUID to attach any additional metadata to the feedback.
      */
-    public func recordFeedback(type: FeedbackType = .general, description: String? = nil) -> UUID? {
-        return enqueueFeedbackEvent(type: type, description: description)
+    public func recordFeedback(type: FeedbackType = .general, description: String? = nil, userInfo: [String: String?]? = nil) -> UUID? {
+        if (userInfo != nil) {
+            self.userInfo = userInfo
```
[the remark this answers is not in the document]
jill-cardamon: That makes sense; we can let the developer set `NavigationEventsManager.userInfo` before sending feedback events, and this info will still be included.

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -116,6 +116,7 @@ struct NavigationEventDetails: EventDetails {
     var rating: Int?
     var comment: String?
     var userId: String?
+    var appMetadata: [String: String?]? = nil
```
azarovalex, line 119: `nil` is the default value already, we don't assign it explicitly in other fields, so I guess it would be better to remove it here.

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -136,6 +137,10 @@ struct NavigationEventDetails: EventDetails {
         sessionIdentifier = session.identifier.uuidString
         originalRequestIdentifier = session.originalRoute.routeIdentifier
         requestIdentifier = dataSource.routeProgress.route.routeIdentifier
+        
+        if (userInfo != nil) {
```
azarovalex, line 141: Do we really need this check? Both `userInfo` and `appMetadata` are Optional, you can just assign it directly.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -147,13 +153,14 @@ open class NavigationEventsManager {
         return eventDictionary
     }
     
-    func navigationFeedbackEvent(type: FeedbackType, description: String?) -> NavigationEventDetails? {
+    func navigationFeedbackEvent(type: FeedbackType, description: String?, userInfo: [String:String?]? = nil) -> NavigationEventDetails? {
         guard let dataSource = dataSource, let sessionState = sessionState else { return nil }
-
-        var event = NavigationEventDetails(dataSource: dataSource, session: sessionState, defaultInterface: usesDefaultUserInterface)
+
```
S2Ler, line 158: Redundant empty spaces.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -24,6 +24,12 @@ open class NavigationEventsManager {
     
     weak var dataSource: EventsManagerDataSource?
     
+    /**
+     Optional application metadata that that can help Mapbox more reliably diagnose problems that occur in the SDK.
+     For example, you can provide your application’s name and version, a unique identifier for the end user, and a session identifier.
+    */
+    var userInfo: [String: String?]? = nil
```
S2Ler, line 31: I don't see this property is read anywhere. Can you please clarify what it is for and how it should be used? 

I could think that this is a set of properties to be included in each event. At the same time, user can provide another set of properties right to the method like `recordFeedback(type:description:userInfo:)` and these two sets will be merged. But right now it seems like it is not used at all.

CHANGELOG.md
S2Ler: Tests are welcomed.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
jill-cardamon: With newest commit, `userInfo` is set by the developer and then set as `appMetadata` for each event.

CHANGELOG.md
jill-cardamon: We can't test against the schema directly, and unit-tests just verify that it outputs some JSON that _looks_ correct. We have to go through a test app to check whether it is working properly. Currently doing so with @bamx23.

S2Ler: We can verify that `userInfo` is propagated to `MMEEventsManager` as intended.

Sources/MapboxCoreNavigation/ActiveNavigationEventDetails.swift
```diff
@@ -65,6 +68,8 @@ struct ActiveNavigationEventDetails: NavigationEventDetails {
         originalRequestIdentifier = session.originalRoute?.routeIdentifier
         requestIdentifier = dataSource.routeProgress.route.routeIdentifier
                 
+        appMetadata = userInfo
```
bamx23, line 71: Maybe It's better to use `appMetadata` for the argument name in the constructor. Or use `useInfo` for the property name.

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -18,13 +18,15 @@ struct PerformanceEventDetails: EventDetails {
     var event: String?
     var counters: [Counter] = []
     var attributes: [Attribute] = []
+    var appMetadata: [String: String?]?
```
bamx23, line 21: Maybe it worth placing this property to the protocol as it's common for every navigation event

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -56,6 +56,12 @@ open class NavigationEventsManager {
         }
     }
     
+    /**
+     Optional application metadata that that can help Mapbox more reliably diagnose problems that occur in the SDK.
+     For example, you can provide your application’s name and version, a unique identifier for the end user, and a session identifier.
```
bamx23, line 61: Should we provide the constants for the keys mentioned in this docs note?

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -129,6 +135,7 @@ open class NavigationEventsManager {
         var event = ActiveNavigationEventDetails(dataSource: dataSource, session: sessionState, defaultInterface: usesDefaultUserInterface)
         event.event = MMEEventTypeNavigationCancel
         event.arrivalTimestamp = sessionState.arrivalTimestamp
+        event.appMetadata = userInfo
```
bamx23, line 138: You can use the initializer above to provide userInfo (and for similar instantiations of `ActiveNavigationEventDetails` and `PassiveNavigationEventDetails`)

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -183,12 +193,14 @@ open class NavigationEventsManager {
     
     func navigationFeedbackEventDetails(type: FeedbackType, description: String?) -> NavigationEventDetails? {
         var event: NavigationEventDetails
-        
+    
         if let activeNavigationDataSource = activeNavigationDataSource {
             event = ActiveNavigationEventDetails(dataSource: activeNavigationDataSource,
-                                                 session: sessionState, defaultInterface: usesDefaultUserInterface)
+    session: sessionState, defaultInterface: usesDefaultUserInterface)
```
bamx23, line 199: This line should be shifted to align with `(`

Sources/MapboxCoreNavigation/ActiveNavigationEventDetails.swift
jill-cardamon: There was some back and forth about this naming. We decided to use userInfo for the user-facing variable name, but the event naming needs to match the event schema which uses appMetadata.

Sources/MapboxCoreNavigation/EventDetails.swift
jill-cardamon: Ah it is in the NavigationEventDetails protocol

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
jill-cardamon: As in the naming? Hm...I don't think it's necessary. We don't want to include this value in the event if it isn't set. Do you think it could cause an issue with ingestion?

Sources/MapboxCoreNavigation/ActiveNavigationEventDetails.swift
```diff
@@ -42,6 +42,9 @@ struct ActiveNavigationEventDetails: NavigationEventDetails {
     var rating: Int?
     var comment: String?
     var userIdentifier: String?
+    var appMetadata: [String: String?]?
+    var name: String?
+    var version: String?
```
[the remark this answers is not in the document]
jill-cardamon: Removed!

Sources/MapboxCoreNavigation/ActiveNavigationEventDetails.swift
bamx23: Here I'm talking about consistency between initializer arguments and properties. At the moment this (and similar) struct is not a part of the public API, so both namings are OK, but it would be better to make it sonsistent.

Sources/MapboxCoreNavigation/EventDetails.swift
bamx23: Is there any implementation of the `EventDetails` protocol without this field?

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
bamx23: I'm talking about `you can provide your application’s name and version, a unique identifier for the end user, and a session identifier`. This line says "I can" but I don't know what key to use. If we want this to be internal, maybe it's better not to document this property? Otherwise, I would expect at least to mention dictionary keys available to developers.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -343,7 +350,8 @@ open class NavigationEventsManager {
      
      @param type A `FeedbackType` used to specify the type of feedback
      @param description A custom string used to describe the problem in detail.
-     @return Returns a UUID used to identify the feedback event
+     @param userInfo An optional dictionary used to include application metadata.
```
bamx23, line 353: There is no such argument now

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -37,10 +39,11 @@ struct PerformanceEventDetails: EventDetails {
         let value: String
     }
     
-    init(event: String, session: SessionState, createdOn created: Date?) {
+    init(event: String, session: SessionState, createdOn created: Date?, appMetadata: [String:String?]? = nil) {
         self.event = event
         sessionIdentifier = session.identifier.uuidString
         self.created = created ?? Date()
+        self.appMetadata = userInfo
```
bamx23, line 46:
```suggestion
        self.appMetadata = appMetadata
```

bamx23 (approved):

PR #3008, opened 2021-05-15
Fix "class" as protocol deprecation warnings

S2Ler:
### Description

It is Swift 5.4 warning:

> Using 'class' keyword for protocol inheritance is deprecated; use 'AnyObject' instead


MaximAlien:
Looks like we'll also have to replace `class` to `AnyObject` in `RouteControllerDataSource` as well.

MaximAlien (approved):

S2Ler:
> Looks like we'll also have to replace `class` to `AnyObject` in `RouteControllerDataSource`.

I don't know how I missed this, thanks for pointing it out. ✅

Issue #2951, opened 2021-04-23
Remove --allow-warnings flag in config.yml.


MaximAlien:
In scope of https://github.com/mapbox/mapbox-navigation-ios/pull/2930 while running `$ pod lib lint MapboxCoreNavigation.podspec` `--allow-warnings` flag was introduced to prevent such build warnings: https://github.com/mapbox/mapbox-navigation-ios/pull/2930#issuecomment-819115091:
```
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXConfigurationServiceObserver+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXServerConfiguration+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXLogWriterBackend+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXConfigurationServiceStopUpdateConfig+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXTokenGenerator+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXAccuracyAuthorizationCallback+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXConfigurationService+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXLiveTrackingState+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXPermissionStatusCallback+Private_Internal.h'
    - WARN  | xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'MapboxCommon' does not include header 'MBXEventsServiceObserver+Private_Internal.h'
[...]

PR #3010, opened 2021-05-17
Update to Mapbox Maps `10.0.0-beta.20`, Navigation Native `50.0.0` and Common `12.0.0`.

MaximAlien:
### Description

PR updates Mapbox Maps to `10.0.0-beta.20` (pointing to `main` for now), Navigation Native `50.0.0` and Common `12.0.0`.

Also fixes:
- https://github.com/mapbox/mapbox-navigation-ios/issues/2951

### DIFF
--- a/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
+++ b/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
@@ -105,7 +105,7 @@ class Navigator {
                                                        minimumDaysToPersistVersion: nil)
 
         let tileStorePath = Self.tilesURL?.path ?? ""
-        tileStore = TileStore.getInstanceForPath(tileStorePath)
+        tileStore = TileStore.__getInstanceForPath(tileStorePath)
         let tilesConfig = TilesConfig(tilesPath: tileStorePath,
                                       tileStore: tileStore,
                                       inMemoryTileCache: nil,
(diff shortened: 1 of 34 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -502,10 +506,19 @@ open class NavigationMapView: UIView {
...
         if let lineLayer = lineLayer {
             do {
-                try mapView.style.addLayer(lineLayer,
-                                           layerPosition: isMainRoute ?
-                                            LayerPosition(above: mapView.mainRouteLineParentLayerIdentifier) :
-                                            LayerPosition(below: parentLayerIndentifier))
+                var layerPosition: MapboxMaps.LayerPosition? = nil
+                
+                if isMainRoute {
+                    if let aboveLayerIdentifier = mapView.mainRouteLineParentLayerIdentifier {
```
1ec5, line 512: If the identifier is nil, this method silently returns without adding a layer or printing to the console. Is that intentional? It’d be easier to keep track of cases like this by making `layerPosition` constant (but still optional).

1ec5 (approved):
Looks pretty reasonable to me. Just a question about control flow in one of the updated methods.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -72,22 +72,13 @@ class Navigator {
      Restrict direct initializer access.
      */
     private init() {
```
S2Ler, line 74: Looks like there will be a lot of conflicts with #3014. cc @Udumft

Sources/MapboxNavigation/NavigationMapView.swift
MaximAlien: I've updated the code to use `let` instead of `var`. Otherwise code should work fine: if there is no `layerPosition` layer will be added at the top of the stack.

PR #3014, opened 2021-05-18
Navigator Tile version switching

### LINKED ISSUE #2986: Navigator Tile version switch mechanism
NN v49.x exposes `FallbackVersionsObserver`. It's notifications are a part of Offline Regions feature: when Navigator needs to change Tiles version - it publishes a list of available tiles versions in current `TileStore`.

This task is about implementing handler for `FallbackVersionsObserver` which will ask user to pick (or not) a version and then restart Navigator instance in most possible transparent way. It also should handle restoring from "fallback" state when such notification arrives. Restarting a Navigator should save current route and progress. Restart itself should be as smooth as possible, minimising the recreation overhead.

`onFallbackVersionsFound(forVersions:)` is sent when there are no tiles with current Navigator version is available, but there are other suitable versions exist in TileStore. User may pick any of the provided options, or not pick anything at all. In the latter case Navigation will continue but map matching will be impossible which results in poor user experience. Selecting a version forces Navigator into `fallback` state.

`onCanReturnToLatest()` is sent when Navigator is currently in a `fallback` state and `latest` tiles became available (for example in case Internet connection is restored). User should be prompted to restart Navigator once again for a switch. Denying to restart leaves old tile version active.

Udumft:
Resolves #2986 

### Description
NN.Navigator reports events when navigation tiles become unavailable and when never version becomes feasible again. Unfortunately Navigator itself is not designed to switch tiles versions on the fly, which requires restarting it. 

### Implementation
PR adds implementation for `FallbackVersionsObverver`, storing state in `CoreNavigation.Navigator`. I chose state variable instead of listeners approach due to possible multiple listening entities. Consumers monitor state changes and transfer delegate calls up to user level. Also introduced implementation for restarting Navigator with custom version.


Udumft:
Started as a draft just to kick off the reviews.
There are still ongoing discussion about versioning logic, and an issue on NavNative side, preventing "switch to online" event.

Example/ViewController.swift
```diff
@@ -540,6 +540,34 @@ extension ViewController: NavigationViewControllerDelegate {
         dismissActiveNavigationViewController()
         clearNavigationMapView()
     }
+    
+    // MARK: - Offline Navigation Tiles switching
```
bamx23, line 544: Can we place it in a separate `extension ViewController: NavigationTilesVersioningDelegate`?

Example/ViewController.swift
```diff
@@ -540,6 +540,34 @@ extension ViewController: NavigationViewControllerDelegate {
...
+    
+    func tilesVersioning(wantsSwitchTilesVersion versions: [String], completion: @escaping (String?) -> ()) {
+        let alertController = UIAlertController(title: "Switch to Offline Regions",
+                                                message: "There are next tiles versions available:",
+                                                preferredStyle: .actionSheet)
+        
+        versions.forEach{ version in
```
bamx23, line 551: NIT
```suggestion
        versions.forEach { version in
```

Example/ViewController.swift
```diff
@@ -540,6 +540,34 @@ extension ViewController: NavigationViewControllerDelegate {
         dismissActiveNavigationViewController()
         clearNavigationMapView()
     }
+    
+    // MARK: - Offline Navigation Tiles switching
+    
+    func tilesVersioning(wantsSwitchTilesVersion versions: [String], completion: @escaping (String?) -> ()) {
+        let alertController = UIAlertController(title: "Switch to Offline Regions",
+                                                message: "There are next tiles versions available:",
+                                                preferredStyle: .actionSheet)
```
bamx23, line 549: For `.actionSheet` you should provide a pivot point so as not to crash on iPads.

Example/ViewController.swift
```diff
@@ -540,6 +540,34 @@ extension ViewController: NavigationViewControllerDelegate {
...
+        
+        versions.forEach{ version in
+            alertController.addAction(UIAlertAction(title: version, style: .default, handler: { _ in
```
bamx23, line 552: You say `These are not designed to be displayed to User directly.` in docs. Can we somehow parse the version and provide a more human-readable string here?

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -41,19 +41,24 @@ class Navigator {
...
-    let navigator: MapboxNavigationNative.Navigator
+    private(set) var navigator: MapboxNavigationNative.Navigator!
     
-    let cacheHandle: CacheHandle
+    private(set) var cacheHandle: CacheHandle!
     
-    let roadGraph: RoadGraph
+    private(set) var roadGraph: RoadGraph!
     
     lazy var roadObjectsStore: RoadObjectsStore = {
```
bamx23, line 52: This `lazy var` becomes invalid after the reset

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -41,19 +41,24 @@ class Navigator {
         }
     }
     
-    let historyRecorder: HistoryRecorderHandle
+    private(set) var historyRecorder: HistoryRecorderHandle!
```
bamx23, line 44: Maybe it's better to move all these fields to a class/struct and recreate them all at once?

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -69,15 +74,24 @@ class Navigator {
...
+    func restartNavigator(forcing version: String? = nil) {
+        unsubscribeNavigator()
+        tileVersionState = .nominal
+        navigator?.shutdown()
+        navigator = nil
+        
         let settingsProfile = SettingsProfile(application: ProfileApplication.kMobile,
                                               platform: ProfilePlatform.KIOS)
         
         let endpointConfig = TileEndpointConfiguration(credentials:Navigator.credentials ?? Directions.shared.credentials,
```
bamx23, line 91: NIT
```suggestion
        let endpointConfig = TileEndpointConfiguration(credentials: Navigator.credentials ?? Directions.shared.credentials,
```

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -69,15 +74,24 @@ class Navigator {
...
     
     /**
-     Restrict direct initializer access.
+     Destroys and creates new instance of Navigator together with other related entities.
+     
+     Typically, this method is used to restart a Navigator with a specific Version during switching to offline or online modes.
+     - parameter version: String representing a tile version name. `nil` value means "latest". Specifying exact version also enables `fallback` mode which will passively monitor newer version available and will notify `tileVersionState` if found.
      */
-    private init() {
+    func restartNavigator(forcing version: String? = nil) {
```
bamx23, line 82: NIT: It's better to move methods below the constructor

bamx23, line 82: Basically, such an approach for recreating a Navigator might be a good candidate for making a static activation method somewhere so that we can stop putting initialization of the static fields of this class everywhere. Or at least we might consider making a struct with these fields and provide both an old approach with putting this activation config to existing places and a new approach with a public static activation method.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -273,6 +273,42 @@ open class RouteController: NSObject {
...
         // Check for faster route proactively (if reroutesProactively is enabled)
         refreshAndCheckForFasterRoute(from: location, routeProgress: routeProgress)
+        
+        verifyNavigationTilesVersion()
+    }
+    
+    private var wantsSwitchTilesVersion = false
+    private var wantsRestoreTilesVersion = false
+    
+    private func verifyNavigationTilesVersion() {
```
bamx23, line 283: Looks like a copy-paste. There should be a way to generalize this code

Sources/MapboxNavigation/NavigationTilesVersioningDelegate.swift
```diff
@@ -0,0 +1,40 @@
...
+     
+     - parameter restoreCallback: A callback to perform or not the switching. You may ignore calling it if you choose to ignore the event.
+     */
+    func tilesVersioning(wantsRestoreTilesVersion restoreCallback: @escaping (Bool) -> ())
+}
+
+
+public extension NavigationTilesVersioningDelegate {
+    func tilesVersioning(wantsSwitchTilesVersion versions: [String], completion: @escaping (String?) -> ()) {
+        logUnimplemented(protocolType: NavigationTilesVersioningDelegate.self,  level: .info)
```
bamx23, line 34: Doesn't the default implementation do anything?

Sources/MapboxNavigation/NavigationViewControllerDelegate.swift
```diff
@@ -11,7 +11,7 @@ import Turf
  
  For convenience, several location-related methods in the `NavigationServiceDelegate` protocol have corresponding methods in this protocol.
  */
-public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
+public protocol NavigationViewControllerDelegate: VisualInstructionDelegate, NavigationTilesVersioningDelegate {
```
bamx23, line 14: Is it a good idea to provide a common point with `NavigationTilesVersioningDelegate` only for highest level delegates/objects? And for `PassiveLocationManager` you've added a delegate property, but for `NavigationViewController` it's done by extending an existing delegate.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -273,6 +273,42 @@ open class RouteController: NSObject {
...
+        switch Navigator.shared.tileVersionState {
+        case .shouldFallback(let versions):
+            if !wantsSwitchTilesVersion {
+                wantsSwitchTilesVersion = true
+                delegate?.router(self, wantsSwitchTilesVersion: versions, completion: { [weak self] version in
```
S2Ler, line 288: It is clear from the code that this completion should be called on the main thread. I would either document this assumption or made sure that the code below is called on the correct thread.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -273,6 +273,42 @@ open class RouteController: NSObject {
         }
         // Check for faster route proactively (if reroutesProactively is enabled)
         refreshAndCheckForFasterRoute(from: location, routeProgress: routeProgress)
+        
+        verifyNavigationTilesVersion()
+    }
+    
+    private var wantsSwitchTilesVersion = false
```
S2Ler, line 280: nit: I know it is not the patter followed in the SDK but it is recommended to specify the type of the variable if it is in a top level declaration/class/struct/etc. It makes compiler do less work.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -41,19 +41,24 @@ class Navigator {
...
         return RoadObjectsStore(navigator.roadObjectStore())
     }()
 
-    let tileStore: TileStore
+    private(set) var tileStore: TileStore!
+    
+    /**
+     Current Navigator status in terms of tile versioning.
+     */
+    fileprivate(set) var tileVersionState: TileVersionState!
```
S2Ler, line 61: Looks like can be `private`.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -41,19 +41,24 @@ class Navigator {
         }
     }
     
-    let historyRecorder: HistoryRecorderHandle
+    private(set) var historyRecorder: HistoryRecorderHandle!
     
-    let navigator: MapboxNavigationNative.Navigator
+    private(set) var navigator: MapboxNavigationNative.Navigator!
```
S2Ler, line 46: Force unwrapping is very dangerous practice is should be avoided if possible.

I join @bamx23 and suggest to extract these variables into class. Then you can have static method which will create these dependencies and then use this static factory method in `init` and `restart`. 

Here is a variant with tuples just to make it clearer what I suggest:

```swift
    static func create(version: String?) -> (tileVersionState: TileVersionState, tileStore: TileStore, historyRecorder: HistoryRecorderHandle, cacheHandle: CacheHandle, roadGraph: RoadGraph, navigator: MapboxNavigationNative.Navigator) {
        let tileVersionState = TileVersionState.nominal

        let settingsProfile = SettingsProfile(application: ProfileApplication.kMobile,
                                              platform: ProfilePlatform.KIOS)

        let endpointConfig = TileEndpointConfiguration(credentials:Navigator.credentials ?? Directions.shared.credentials,
                                                       tilesVersion: version ?? Self.tilesVersion,
                                                       minimumDaysToPersistVersion: nil,
                                                       isFallback: version != nil)

        let tileStorePath = Self.tilesURL?.path ?? ""
        let tileStore = TileStore.getInstanceForPath(tileStorePath)
        let tilesConfig = TilesConfig(tilesPath: tileStorePath,
                                      tileStore: tileStore,
                                      inMemoryTileCache: nil,
                                      onDiskTileCache: nil,
                                      mapMatchingSpatialCache: nil,
                                      threadsCount: nil,
                                      endpointConfig: endpointConfig)

        let historyAutorecordingConfig = [
            "features": [
                "historyAutorecording": true
            ]
        ]

        var customConfig = ""
        if let jsonDataConfig = try? JSONSerialization.data(withJSONObject: historyAutorecordingConfig, options: []),
           let encodedConfig = String(data: jsonDataConfig, encoding: .utf8) {
            customConfig = encodedConfig
        }

        let configFactory = ConfigFactory.build(for: settingsProfile,
                                                config: NavigatorConfig(),
                                                customConfig: customConfig)

        let historyRecorder = HistoryRecorderHandle.build(forHistoryFile: Navigator.historyDirectoryURL?.path ?? "", config: configFactory)

        let runloopExecutor = RunLoopExecutorFactory.build()
        let cacheHandle = CacheFactory.build(for: tilesConfig,
                                         config: configFactory,
                                         runLoop: runloopExecutor,
                                         historyRecorder: historyRecorder)

        let roadGraph = RoadGraph(MapboxNavigationNative.GraphAccessor(cache: cacheHandle))

        let navigator = MapboxNavigationNative.Navigator(config: configFactory,
                                                     runLoopExecutor: runloopExecutor,
                                                     cache: cacheHandle,
                                                     historyRecorder: historyRecorder)

        return (tileVersionState: tileVersionState, tileStore: tileStore, historyRecorder: historyRecorder, cacheHandle: cacheHandle, roadGraph: roadGraph, navigator: navigator)
    }
```

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -139,6 +166,23 @@ class Navigator {
     }
 }
 
+extension Navigator: FallbackVersionsObserver {
+    
+    enum TileVersionState {
+        case nominal
```
S2Ler, line 172: Would be good to have a comment describing what `nominal` `TileVersionState` means.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
Udumft: I am not sure to follow the idea about this activation method. Do you mean to store all settings in a special struct and pass it to constructor/restart method? What are the advantages?

Sources/MapboxCoreNavigation/RouteController.swift
Udumft: Yeah, I tried, but since `RouteController` and `PassiveLocationDataSource` have different delegates to call and different completion logic to handle, generalisation looked way uglier than code duplication.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
Udumft: Force unwrapping is used to allow initializing properties using a method call inside `init()`.
I've been thinking in a similar way in [subsequent PR](https://github.com/mapbox/mapbox-navigation-ios/pull/3015/files#diff-fb53bdfdf70730b6dbbb59d044c859accd4e7709e8857bab1462d493298e265bR76), where this approach is put to real use-case.
I didn't consider introduce a dedicated type instead of a tuple to wrap all the properties because I couldn't come up with a decent naming for the context of such type.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
bamx23: An activation method is just a suggestion to think about.

> Do you mean to store all settings in a special struct and pass it to constructor/restart method? What are the advantages?

No, I basically mean the suggestion in this comment: https://github.com/mapbox/mapbox-navigation-ios/pull/3014#discussion_r635116232. So that we don't have to force-unwrap all fields, but to have a single field with all the parameters of an active navigator.

Sources/MapboxNavigation/NavigationTilesVersioningDelegate.swift
Udumft: I don't think we should do something in default implementation other than ignore the event. Forcing a version switch by default may result in unexpected UI hiccups for user

Sources/MapboxNavigation/NavigationViewControllerDelegate.swift
Udumft: Good question. I thought that `NavigationViewControllerDelegate` covers various aspects of the Navigation process, so  handling switching may fall there too. And `PassiveLocationManager.delegate` is actually a `LocationProviderDelegate` which is designed for a specific purpose and tiles logic does not fit to that context.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
S2Ler: I want to encourage you to remove force unwraps and use the work from this [next PR](https://github.com/mapbox/mapbox-navigation-ios/pull/3015/files#diff-fb53bdfdf70730b6dbbb59d044c859accd4e7709e8857bab1462d493298e265bR76). 

Personally, I ok with using tuples. 

I've seen how force unwraps results in crashes in production, it is effortless to break during modifications, unreliable, and the compiler doesn't help you. In other words, they are very hard to maintain due to relaxed support from the Swift compiler.

Example/ViewController.swift
```diff
@@ -540,21 +540,38 @@ extension ViewController: NavigationViewControllerDelegate {
...
         let alertController = UIAlertController(title: "Switch to Offline Regions",
                                                 message: "There are next tiles versions available:",
                                                 preferredStyle: .actionSheet)
         
-        versions.forEach{ version in
-            alertController.addAction(UIAlertAction(title: version, style: .default, handler: { _ in
+        versions.forEach { version in
+            // converting version name format from "2020_01_02-03_00_00" -> "2020/01/02 03:00:00"
+            let parts = version.split(separator: "-", maxSplits: 1, omittingEmptySubsequences: true)
+            let versionName = parts.first!.replacingOccurrences(of: "_", with: "/") + " " +
```
bamx23, line 557: Maybe it's better to parse this into a `Date` and show a localized string from formatted?

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -49,16 +49,16 @@ class Navigator {
     
     private(set) var roadGraph: RoadGraph!
     
-    lazy var roadObjectsStore: RoadObjectsStore = {
+    func roadObjectsStore() -> RoadObjectsStore {
```
bamx23, line 52: I would leave this as a dynamic `var`

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -163,27 +168,33 @@ open class PassiveLocationDataSource: NSObject {
     private var wantsRestoreTilesVersion = false
     
     private func verifyNavigationTilesVersion() {
+        guard delegate != nil else { return }
```
bamx23, line 171: `guard let`?

S2Ler: Agree, and in other places as well.

Udumft: Corrected

Udumft:
I see that opinions are inclined to encapsulate version naming and notify user only about the event of switching offline/online, so I removed related code.

Example/ViewController.swift
Udumft: I removed the naming piece, so it's not relevant anymore.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
Udumft: Introduced `NativeHandlersFactory` for that functionality.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -273,6 +273,42 @@ open class RouteController: NSObject {
...
+                        Navigator.shared.restartNavigator(forcing: version)
+                        self.updateNavigator(with: self._routeProgress)
+                        self.updateRouteLeg(to: self._routeProgress.legIndex)
+                    }
+                    self.wantsSwitchTilesVersion = false
+                })
+            }
+        case .shouldReturnToLatest:
+            if !wantsRestoreTilesVersion {
+                wantsRestoreTilesVersion = true
```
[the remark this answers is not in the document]
Udumft: Good catch. Corrected

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
Udumft: See `NativeHandlersFactory` if I got it right (:

Example/ViewController.swift
Udumft: Done

Udumft:
@truburt , do you think we are fine to stick with no versions idea for now and mark this one actually "ready for review"?

truburt:
> @truburt , do you think we are fine to stick with no versions idea for now and mark this one actually "ready for review"?

Yes, definitely, let's follow this path.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
bamx23: This indeed looks great. Would you mind considering storing the object `factory` in `Navigator` directly and use dynamic properties for `historyRecorder`, `navigator` etc? In this case `factory` could be renamed to something else and creation for this factory might be moved to a private static method. This would allow us to reduce code duplication in `init` and `restartNavigator` to a minimum to prevent possible inconsistency in those methods.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -72,58 +77,56 @@ class Navigator {
...
+     */
+    func restartNavigator(forcing version: String? = nil) {
+        unsubscribeNavigator()
+        tileVersionState = .nominal
+        navigator.shutdown()
         
-        roadGraph = RoadGraph(MapboxNavigationNative.GraphAccessor(cache: cacheHandle))
+        let factory = NativeHandlersFactory(tileStorePath: Self.tilesURL?.path ?? "",
+                                            credentials: Self.credentials ?? Directions.shared.credentials,
+                                            tilesVersion: Self.tilesVersion,
```
bamx23, line 107: Shouldn't it be like this:
```suggestion
                                            tilesVersion: version ?? Self.tilesVersion,
```

Sources/MapboxCoreNavigation/NativeHandlersFactory.swift
```diff
@@ -0,0 +1,100 @@
...
+        TilesConfig(tilesPath: tileStorePath,
+                    tileStore: TileStore.getInstanceForPath(tileStorePath),
+                    inMemoryTileCache: nil,
+                    onDiskTileCache: nil,
+                    mapMatchingSpatialCache: nil,
+                    threadsCount: nil,
+                    endpointConfig: endpointConfig)
+    }()
+    
+    lazy var configHandle: ConfigHandle = {
```
bamx23, line 79: I really like how it looks now

Sources/MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -159,6 +159,28 @@ public protocol NavigationServiceDelegate: AnyObject, UnimplementedLogging {
...
+    /**
+     Called when navigation service wants to switch Navigation Tiles version
+     
+     This happens when latest Navigation tiles are not available for current location (due to connectivity issues for example), but there is pre-downloaded offline data available which is older version.
+     
+     - parameter service: The navigation service reporting a switch attempt
+     - parameter fallbackCallback: Completion callback which automatically resolves the latest suitable version available. Sender will reconfigure the navigation engine immediately. You may ignore calling the callback at all if you choose to ignore the version switch. Keep in mind that such decision may result in poor user experience.
+     - seealso: `navigationService(_:, wantsRestoreTilesVersion:)`
+     */
+    func navigationService(_ service: NavigationService, wantsFallbackTilesVersion fallbackCallback: @escaping () -> ())
```
bamx23, line 171: `() -> Void`? (here and in similar methods)

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -155,6 +160,43 @@ open class PassiveLocationDataSource: NSObject {
...
+    
+    private var wantsSwitchTilesVersion = false
+    private var wantsRestoreTilesVersion = false
+    
+    private func verifyNavigationTilesVersion() {
+        guard let delegate = delegate else { return }
+        
+        switch Navigator.shared.tileVersionState {
+        case .shouldFallback(let versions):
+            guard let version = versions.first else { return }
```
bamx23, line 175: I heard that `.first` here is the oldest one

Sources/MapboxCoreNavigation/RouteController.swift
bamx23: Thought about this again... It holds some logic(like selecting the latest version) and I would recommend reconsidering moving this to a class. Maybe both delegates (`PassiveLocationDataSourceDelegate` and `RouterDelegate`) can inherit `NavigationTilesVersioningDelegate`?

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
Udumft: I thought it would be better to not store this factory because navigator settings could be potentially updated and if factory remains the same we either have to update it too (which adds more code) or such settings will be ignored on navigator restart.

bamx23: If it's potentially possible that only some of the parameters may change, then OK, the current approach is great

Udumft:
After discussions with native team we came to the solution that Nav SDK should take care about reassigning observers for EH, Fallback, RoadObjectStore/Matcher and recreating PredictiveCacheControllers. The latter are sensitive to `navigator.setRoute` call and should be created before that, but that should be fixed in next NN release.
Until then I leave this PR as a draft, but in general it should be code-ready.

The downside of such solution is that all `RoadObjects` become invalid once we restart Navigator instance. We decided that it is sufficient for now to warn users about it with a notification and doc notes.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -1179,6 +1179,13 @@
 			path = Support;
 			sourceTree = "<group>";
 		};
+		2BBED92D265CFBAB00F90032 /* Recovered References */ = {
```
azarovalex, line 1182: I think it shouldn't be here.

Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -377,3 +377,41 @@ extension RoadGraph {
...
+
+public extension Notification.Name {
+    /**
+     Posted when Navigator has not enough tiles for map matching on current tiles version, but there are suitable older versions inside underlying Offline Regions. Navigator will restart after this notification is issued.
+     
+     Such action invalidates all existing matched `RoadObject`s which should be re-applied manually.
+     
+     The user info dictionary contains the key `Navigator.NotificationUserInfoKey.versionsKey`
+    */
+    static let navigatorWantsFallbackToOfflineVersion: Notification.Name = .init(rawValue: "NavigatorWantsFallbackToOfflineVersion")
```
azarovalex, line 389: maybe rename to `willFallback`?

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -81,21 +81,64 @@ class Navigator {
...
+        navigator.setFallbackVersionsObserverFor(nil)
+    }
+     
+    private func setupElectronicHorizonOptions() {
+        let nativeOptions: MapboxNavigationNative.ElectronicHorizonOptions?
+        if let electronicHorizonOptions = electronicHorizonOptions {
+            nativeOptions = MapboxNavigationNative.ElectronicHorizonOptions(electronicHorizonOptions)
+        } else {
+            nativeOptions = nil
+        }
```
bamx23, line 140:
```suggestion
        let nativeOptions = electronicHorizonOptions.map(MapboxNavigationNative.ElectronicHorizonOptions.init)
```

Sources/MapboxCoreNavigation/PassiveLocationDataSource.swift
```diff
@@ -162,6 +168,35 @@ open class PassiveLocationDataSource: NSObject {
...
+        guard let version = (notification.userInfo?[Navigator.NotificationUserInfoKey.versionsKey] as? [String])?.last else
+        {
+            return
+        }
+        
+        Navigator.shared.restartNavigator(forcing: version)
+    }
+    
+    @objc func restoreToOnline(_ notification: Notification) {
+        Navigator.shared.restartNavigator(forcing: nil)
```
bamx23, line 197: Is it okay that in case there is no `PassiveLocationDataSource` nor `RouteController` we don't call `Navigator.shared.restartNavigator`? Or the case of unbalanced calls of `Navigator.shared.restartNavigator` when we went offline and then unsubscribed from these notifications on destroying all `PassiveLocationDataSource` and `RouteController`.
Shouldn't we call `Navigator.shared.restartNavigator` inside the Navigator instance itself?

Udumft: Technically, switching to offline/online could not happen while there are no active `RouteController` or `PassiveLocationDataSource` since this mechanism relies on location updates. No updates - no switching. 
But I guess we could indeed try to restart navigator in `Navigator` itself. The only thing to keep in mind here is that it should also re-apply route and progress data (as in `RouteController`) minding the initialisation order.

bamx23: ATM both classes make their specific job after Navigator is reset. But in general, we might consider `will`/`did` notifications for both going offline and online notifications so that if the class wants to do something before Navigator is reset it can subscribe to `will` notifications.

Udumft: We discussed notification with @truburt and @korshaknn and came to conclusion that currently there are no reasons to alert `will*` notifications. I'll move the restart logic to `Navigator` in next commit.

bamx23 (approved):
LGTM

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -109,13 +148,53 @@ class Navigator {
...
+        switch tileVersionState {
+        case .nominal, .shouldReturnToLatest:
+            tileVersionState = .shouldFallback(versions)
+            
+            guard let fallbackVersion = versions.last else { return }
+            
+            Navigator.shared.restartNavigator(forcing: fallbackVersion)
+            
+            let userInfo: [Navigator.NotificationUserInfoKey: Any] = [
+                .tilesVersionKey: fallbackVersion // internal use only
```
S2Ler, line 177: It says that is for internal use only, but this key mentioned in public API and the changelog. Didn't follow up all the conversations but this looks suspicious for me.

Udumft: Yeah, the comment might be misleading. This key (and value) used internally for version switching. Even though it is technically accessable for SDK users, we do not encourage them to use or rely on it. Actually, it has next to no practical meaning for them right now, other than curiosity or logging.
In future, we might add more version-related logic for configuring navigation, at which point such data will be handy.

S2Ler: Then I think we can hide this information from the public API and changelog. Any changes to this key in the future will be considered as a breaking change, so it would be good to hide the implementation details. The key will still appear in the `userInfo`, but it will be considered a private API because there is no documentation and public keys. We will have more room for future changes related to this functionality. 

@mapbox/navigation-ios What other think about this?

1ec5: Yes, let’s keep it as an undocumented API by removing any mention of it from the changelog and documentation comments on public symbols. (It’s OK to mention it in documentation comments on internal symbols though.)

CHANGELOG.md
```diff
@@ -93,6 +93,7 @@
...
 * Added methods for convenience checking Navigation tiles in a `TileStore`: `containsLatestNavigationTiles(forCacheLocation:, completion:)`, `tileRegionContainsLatestNavigationTiles(forId:, cacheLocation:, completion:)` and methods for getting Navigation tiles from `TilesetDescriptorFactory`: `getLatest(forCacheLocation:)`, `getSpecificVersion(forCacheLocation:, version:)`. ([#3015](https://github.com/mapbox/mapbox-navigation-ios/pull/3015))
 * Fixed a thread-safety issue in `UnimplementedLogging` protocol implementation. ([#3024](https://github.com/mapbox/mapbox-navigation-ios/pull/3024))
+* Implemented automatic switching to older navigation tiles versions (if they are present in current `TileStore`) and back. You can subscribe to `navigationDidSwitchToFallbackVersion` with selected version at `tilesVersionKey` user info key and `navigationDidSwitchToTargetVersion` notifications to monitor such events. ([#3014](https://github.com/mapbox/mapbox-navigation-ios/pull/3014))
```
1ec5, line 96: Nit: Symbols need to be fully qualified in order for jazzy to automatically link them:

```suggestion
* Implemented automatic switching to older navigation tiles versions (if they are present in current `TileStore`) and back. You can subscribe to `Notification.Name.navigationDidSwitchToFallbackVersion` and `Notification.Name.navigationDidSwitchToTargetVersion` notifications to monitor such events. ([#3014](https://github.com/mapbox/mapbox-navigation-ios/pull/3014))
```

Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -377,3 +377,52 @@ extension RoadGraph {
...
+    
+    /**
+     Posted when Navigator was switched to a fallback offline tiles version, but latest tiles became available again. Navigator has restarted when this notification is issued.
+     
+     Such action invalidates all existing matched `RoadObject`s which should be re-applied manually.
+     */
+    static let navigationDidSwitchToTargetVersion: Notification.Name = .init(rawValue: "NavigatorDidRestoreToOnlineVersion")
+    
+    /**
+     Posted when NavNative sends updated navigation status.
```
S2Ler, line 399: Is "NavNative" is officially exposed name?

Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -377,3 +377,52 @@ extension RoadGraph {
...
+        public init(rawValue: String) {
+            self.rawValue = rawValue
+        }
+        
+        /**
+         A key in the user info dictionary of a `Notification.Name.navigationDidSwitchToFallbackVersion` notification. The corresponding value is a string representation of selected tiles version.
+         
+         For internal use only.
+         */
+        public static let tilesVersionKey: NotificationUserInfoKey = .init(rawValue: "tilesVersion")
```
S2Ler, line 422: Just curious, what's the purpose of exposing an "internal" thing as public?

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
S2Ler: Looks like this is still relevant.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -117,13 +157,53 @@ class Navigator {
...
+    enum TileVersionState {
+        /// No tiles version switch is required. Navigator has enough tiles for map matching.
+        case nominal
+        /// Navigator does not have tiles on current version for map matching, but TileStore contains regions with required tiles of a different version
+        case shouldFallback([String])
+        /// Navigator is in a fallback mode but newer tiles version were successefully downloaded and ready to use.
+        case shouldReturnToLatest
+    }
+    
+    func onFallbackVersionsFound(forVersions versions: [String]) {
```
S2Ler, line 176: What thread this callbacks are called on? Any guarantees from NavNative about the thread?

azarovalex (approved):

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
Udumft: Currently it is called on a worker thread. But I think it's better to do the navigator reset logic on main thread

Sources/MapboxCoreNavigation/CoreConstants.swift
Udumft: Accidentally lost :nodoc: tag while rebasing. Corrected.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -117,13 +157,56 @@ class Navigator {
...
-            navigator.setElectronicHorizonOptionsFor(nativeOptions)
+        case .shouldFallback:
+            break // do nothing
+        }
+    }
+
+    func onCanReturnToLatest() {
+        switch tileVersionState {
+        case .nominal, .shouldFallback:
+            tileVersionState = .shouldReturnToLatest
```
S2Ler, line 201: Probably makes sense to read/mutate `tileVersionState` on the main thread as well. It is very easy to introduce a data race when variables mutates and read from different threads.

S2Ler (approved):
🚀

PR #3015, opened 2021-05-18
TileStore navigation tiles API

### LINKED ISSUE #2987: Nav Tile expiration API
We need an API method for easy checking if specific Offline Region tiles should be updated.

Map tiles are updated automatically, but Navigation tiles require explicit action. It is user's responsibility to update regions whenever needed and appropriate, but we should allow seamless way to verify region expiration. 

Current flow suggests next actions (subject to change as there are some flaws in this design):
* Get all offline regions from storage
* get latest Nav tile descriptor
* verify if offline regions contain that tile descriptor
* if no - such region should be updated
This scenario does not cover cases when there are no Nav tiles supposed to be in offline region as it will always report it as expired.

Udumft:
Resolves #2987 

### Description

The PR adds convenience methods to verify validity of Navigation tiles in Offline Region or entire TileStore, and an accessor method for corresponding TilesetDescriptor which is required for downloading navigation tiles.

### Implementation

TilesetDescriptor methods are a bit awkward as they require a `CacheHandle` which is configured during Navigator instance startup. Getting a descriptor should be independent of Navigation, but share configuration values. That's why there is a chunk of duplicated logic. I was thinking about moving `CacheHandle` initialisation logic into `CoreNavigationNavigator` but could not come up with a solution which looked pretty. The problem is that `CacheHandle` construction is coupled with other entities initialisation and it messes up the Navigator init sequence.

### DIFF
--- a/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
+++ b/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
@@ -41,23 +41,24 @@ class Navigator {
         }
     }
     
-    let historyRecorder: HistoryRecorderHandle
+    private(set) var historyRecorder: HistoryRecorderHandle
     
-    let navigator: MapboxNavigationNative.Navigator
+    private(set) var navigator: MapboxNavigationNative.Navigator
     
-    let cacheHandle: CacheHandle
+    private(set) var cacheHandle: CacheHandle
     
-    let roadGraph: RoadGraph
+    private(set) var roadGraph: RoadGraph
     
     lazy var roadObjectsStore: RoadObjectsStore = {
         return RoadObjectsStore(navigator.roadObjectStore())
     }()
-
-    let tileStore: TileStore
-
+    
     lazy var roadObjectMatcher: RoadObjectMatcher = {
         return RoadObjectMatcher(MapboxNavigationNative.RoadObjectMatcher(cache: cacheHandle))
     }()
+
+
+    private(set) var tileStore: TileStore
     
     /**
      The Authorization & Authentication credentials that are used for this service. If not specified - will be automatically intialized from the token and host from your app's `info.plist`.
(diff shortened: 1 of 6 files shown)


Udumft:
Starting as a draft due to potential issue in TileStore which does not call a completion for tilesetDescriptor query, which hangs the processing.

Udumft:
Updated PR to be merged directly to `release-v2.0` instead of previous base #3014.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -41,23 +41,24 @@ class Navigator {
         }
     }
     
-    let historyRecorder: HistoryRecorderHandle
+    private(set) var historyRecorder: HistoryRecorderHandle
```
Udumft, line 44: Currently these changes are useless, but it's a fallout of it's previous base PR. This will be utilised with Navigator Version switching

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -76,60 +77,31 @@ class Navigator {
      Restrict direct initializer access.
      */
     private init() {
-        let settingsProfile = SettingsProfile(application: ProfileApplication.kMobile,
```
MaximAlien, line 79: Would it make sense to rebase current changes onto https://github.com/mapbox/mapbox-navigation-ios/pull/3010? I think I've updated shared navigator there as well to accomodate new API changes.

Sources/MapboxCoreNavigation/TilesetDescriptorFactory.swift
```diff
@@ -0,0 +1,27 @@
...
+import MapboxNavigationNative
+import MapboxDirections
+
+extension TilesetDescriptorFactory {
+    /**
+     Gets TilesetDescriptor which corresponds to current Navigator dataset and the specified `version`.
+     - parameter cacheLocation: A `Location` where cache data is stored.
+     - parameter version: TilesetDescriptor version.
+     */
+    open class func getSpecificVersion(forCacheLocation cacheLocation: TileStoreConfiguration.Location = .default, version: String) -> TilesetDescriptor {
```
MaximAlien, line 11: Maybe
```suggestion
    open class func getSpecificVersion(for cacheLocation: TileStoreConfiguration.Location = .default, version: String) -> TilesetDescriptor {
```
would be less cumbersome (similarly for second method in this extension)?

Sources/MapboxCoreNavigation/TileStore.swift
```diff
@@ -0,0 +1,105 @@
...
+                                            } else if expected.isError() {
+                                                completion(nil)
+                                            } else {
+                                                assertionFailure("Unexpected value or error: \(expected), expected: \(NSNumber.self)")
+                                                completion(nil)
+                                            }
+                                        })
+    }
+    
+    private func temporaryDirectory() -> URL {
```
MaximAlien, line 97: Is this method used anywhere?

Sources/MapboxCoreNavigation/TilesetDescriptorFactory.swift
Udumft: Even though it's suboptimal, but other `TilesetDescriptorFactory` methods are mapped from Obj-C and look like `getSpecificVersion(forCache:, version:)`. I decided to repeat the pattern to blend the extension methods.

Sources/MapboxCoreNavigation/TileStore.swift
```diff
@@ -0,0 +1,95 @@
+import MapboxNavigationNative
+
+
+extension TileStore {
+    public typealias ContainsCompletion = (Bool?) -> Void
```
bamx23, line 5: Maybe it's better to use `Result<Bool, Error>` here?

Sources/MapboxCoreNavigation/TilesetDescriptorFactory.swift
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    /**
+     Gets TilesetDescriptor which corresponds to the latest availble version of routing tiles.
+     
+     Intended for using when creating off-line tile packs.
+     
+     - parameter location: A `Location` where cache data is stored.
+     */
+    open class func getLatest(forCacheLocation cacheLocation: TileStoreConfiguration.Location = .default) -> TilesetDescriptor {
```
bamx23, line 23: Do we need these methods to be `open`?

bamx23 (approved):

Sources/MapboxCoreNavigation/TileStore.swift
Udumft: That would introduce some inconsistency. All `TileStore` methods swift wrappers are defined in `MapboxMaps` module and have `Result` argument. They also introduce their defined `Error` types. Thus, if I am going to also use `Error`s, that would be another type (since I cannot import maps here for reuse) which may be implicit and confusing for users. So I decided to make it explicitly different by using another completion type.

Sources/MapboxCoreNavigation/TilesetDescriptorFactory.swift
Udumft: You are right. Corrected

PR #3019, opened 2021-05-19
Use  EHorizon to improve road label during turn-by-turn navigation

### LINKED ISSUE #1440: Reimplement current road name label atop navigator status/electronic horizon

ShanMa1991:
### Description
This pr is to fix #1440 , by improving the road label after feature query results from the EHorizon notification.

### Implementation
 Receiving the e-horizon road names through `Notification`. During the turn-by-turn navigation, when there're multiple feature query results, comparing the road name of the feature with the e-horizon road names. And then choosing the most similar feature to label the road if it's not the nearest one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@
 * Developers can create an instance of `NavigationViewController` from `UIStoryboardSegue`, which is located in `Navigation.storyboard`. To successfully create `NavigationViewController` instance developer has to pre-define `route`, `routeIndex`, `routeOptions` and `navigationOptions` properties of `NavigationViewController` in `UIViewController.prepare(for:sender:)`. ([#2974](https://github.com/mapbox/mapbox-navigation-ios/pull/2974))
 * Added methods for convenience checking Navigation tiles in a `TileStore`: `containsLatestNavigationTiles(forCacheLocation:, completion:)`, `tileRegionContainsLatestNavigationTiles(forId:, cacheLocation:, completion:)` and methods for getting Navigation tiles from `TilesetDescriptorFactory`: `getLatest(forCacheLocation:)`, `getSpecificVersion(forCacheLocation:, version:)`. ([#3015](https://github.com/mapbox/mapbox-navigation-ios/pull/3015))
 * Fixed a thread-safety issue in `UnimplementedLogging` protocol implementation. ([#3024](https://github.com/mapbox/mapbox-navigation-ios/pull/3024))
+* Fixed an issue where the current road name label sometimes displayed the name of an intersecting road instead of the current road or blinked in and out. ([#3019](https://github.com/mapbox/mapbox-navigation-ios/pull/3019))
 
 ## v1.4.0
 
--- a/Sources/MapboxNavigation/OrnamentsController.swift
+++ b/Sources/MapboxNavigation/OrnamentsController.swift
@@ -58,6 +58,8 @@ extension NavigationMapView {
         
         var labelRoadNameCompletionHandler: (LabelRoadNameCompletionHandler)?
         
+        var electronicHorizonRoadNames = [String]()
+        
         // MARK: - Lifecycle
         
         init(_ navigationViewData: NavigationViewData, eventsManager: NavigationEventsManager) {
@@ -70,12 +72,41 @@ extension NavigationMapView {
                                                    selector: #selector(orientationDidChange(_:)),
                                                    name: UIDevice.orientationDidChangeNotification,
                                                    object: nil)
+            NotificationCenter.default.addObserver(self,
+                                                   selector: #selector(didUpdateEHorizonRoadName),
+                                                   name: .electronicHorizonDidUpdatePosition,
+                                                   object: nil)
+
         }
         
         private func suspendNotifications() {
             NotificationCenter.default.removeObserver(self,
                                                       name: UIDevice.orientationDidChangeNotification,
                                                       object: nil)
+            NotificationCenter.default.removeObserver(self,
+                                                      name: .electronicHorizonDidUpdatePosition,
+                                                      object: nil)
+        }
+        
+        @objc func didUpdateEHorizonRoadName(_ notification: Notification) {
+            guard let startingEdge = notification.userInfo?[RoadGraph.NotificationUserInfoKey.treeKey] as? RoadGraph.Edge else { return }
+            let edgeIdentifier = startingEdge.identifier
+            
+            guard let routeController = navigationViewData.router as? RouteController,
+                  let metadata = routeController.roadGraph.edgeMetadata(edgeIdentifier: edgeIdentifier) else { return }
+            
+            electronicHorizonRoadNames = metadata.names.map { name -> String in
+                switch name {
+                case .name(let name):
+                    return name
+                case .code(let code):
+                    return "(\(code))"
+                }
+            }
+            
+            if electronicHorizonRoadNames.isEmpty {
+                electronicHorizonRoadNames = ["\(metadata.mapboxStreetsRoadClass.rawValue)"]
+            }
         }
         
         @objc func orientationDidChange(_ notification: Notification) {
@@ -256,7 +287,21 @@ extension NavigationMapView {
                     var smallestLabelDistance = Double.infinity
                     var latestFeature: MBXFeature?
                     
+                    var minimumEditDistance = Int.max
+                    var similarFeature: MBXFeature?
+
                     for queriedFeature in queriedFeatures {
+                        // Calculate the Levenshtein–Damerau edit distance between the EHorizon road name and the feature property road name, and then use the smallest one for the road label.
+                        if let roadName = queriedFeature.feature.properties["name"] as? String,
+                           !roadName.isEmpty,
+                           !self.electronicHorizonRoadNames.isEmpty {
+                            let stringEditDistance = self.electronicHorizonRoadNames.first!.minimumEditDistance(to: roadName)
+                            if stringEditDistance < minimumEditDistance {
+                                minimumEditDistance = stringEditDistance
+                                similarFeature = queriedFeature.feature
+                            }
+                        }
+                        
                         var lineStrings: [LineString] = []
                         
                         if queriedFeature.feature.geometry.geometryType == MBXGeometryType_Line,
@@ -289,7 +334,14 @@ extension NavigationMapView {
                     }
                     
                     var hideWayName = true
-                    if smallestLabelDistance < 5 {
+                    if latestFeature != similarFeature {
+                        let style = self.navigationMapView.mapView.mapboxMap.style
+                        if let similarFeature = similarFeature,
+                           self.navigationView.wayNameView.setupWith(roadFeature: similarFeature,
+                                                                     using: style) {
+                            hideWayName = false
+                        }
+                    } else if smallestLabelDistance < 5 {
                         let style = self.navigationMapView.mapView.mapboxMap.style
                         if let latestFeature = latestFeature,
                            self.navigationView.wayNameView.setupWith(roadFeature: latestFeature,
--- a/Sources/MapboxNavigation/OrnamentsController.swift
+++ b/Sources/MapboxNavigation/OrnamentsController.swift
@@ -100,12 +100,8 @@ extension NavigationMapView {
                 case .name(let name):
                     return name
                 case .code(let code):
-                    return "(\(code))"
+                    return code
                 }
-            }
-            
-            if electronicHorizonRoadNames.isEmpty {
-                electronicHorizonRoadNames = ["\(metadata.mapboxStreetsRoadClass.rawValue)"]
             }
         }
         
@@ -295,7 +291,7 @@ extension NavigationMapView {
                         if let roadName = queriedFeature.feature.properties["name"] as? String,
                            !roadName.isEmpty,
                            !self.electronicHorizonRoadNames.isEmpty {
-                            let stringEditDistance = self.electronicHorizonRoadNames.first!.minimumEditDistance(to: roadName)
+                            let stringEditDistance = self.electronicHorizonRoadNames.map{ $0.minimumEditDistance(to: roadName) }.reduce(0, +)
                             if stringEditDistance < minimumEditDistance {
                                 minimumEditDistance = stringEditDistance
                                 similarFeature = queriedFeature.feature
--- a/Sources/MapboxCoreNavigation/MinimumEditDistance.swift
+++ b/Sources/MapboxCoreNavigation/MinimumEditDistance.swift
@@ -2,7 +2,7 @@ import Foundation
 
 extension String {
     // Adapted from https://github.com/raywenderlich/swift-algorithm-club/blob/master/Minimum%20Edit%20Distance/MinimumEditDistance.playground/Contents.swift
-    func minimumEditDistance(to word: String) -> Int {
+    public func minimumEditDistance(to word: String) -> Int {
         let fromWordCount = count
         let toWordCount = word.count
         


ShanMa1991:
Because the road name provided by ElectronicHorizon may be "Main Street", while the feature query results with the name property may be "Main St" and "Mission St". The next step is to compare the string similarity.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -58,6 +58,8 @@ extension NavigationMapView {
         
         var labelRoadNameCompletionHandler: (LabelRoadNameCompletionHandler)?
         
+        var eHorizonRoadName = [String]()
```
MaximAlien, line 61: Nit: I think that we mostly use `electronicHorizon`, so probably worth renaming to:
```suggestion
        var electronicHorizonRoadNames = [String]()
```

CHANGELOG.md
```diff
@@ -90,6 +90,7 @@
 * Removed deprecated and obsoleted `EventsManager` typealias, `StatusView.delegate`, `StatusView.canChangeValue`, `RouteLegProgress.upComingStep`, `StatusViewDelegate`, `DeprecatedStatusViewDelegate` and `BottomBannerViewController.init(delegate:)`. ([#2993](https://github.com/mapbox/mapbox-navigation-ios/pull/2993))
 * Fixed a potential memory leak when using `MultiplexedSpeechSynthesizer`. ([#3005](https://github.com/mapbox/mapbox-navigation-ios/pull/3005))
 * Fixed an issue where instruction banners could appear in the wrong color after switching between `Style`s. ([#2977](https://github.com/mapbox/mapbox-navigation-ios/pull/2977))
+* Improved the road lable accuracy during turn-by-turn navigation by using the e-horizon notification road name to improve the road feature result. ([#3019](https://github.com/mapbox/mapbox-navigation-ios/pull/3019))
```
MaximAlien, line 93:
```suggestion
* Improved the road name labeling accuracy during turn-by-turn navigation by using `Notification.Name.electronicHorizonDidUpdatePosition`. ([#3019](https://github.com/mapbox/mapbox-navigation-ios/pull/3019))
```

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -252,7 +288,20 @@ extension NavigationMapView {
...
                     for queriedFeature in queriedFeatures {
+                        // Calculate the similarity between the EHorizon road name and the feature property road name, and then use the most similar one for the road label.
+                        if let roadName = queriedFeature.feature.properties["name"] as? String,
+                           !roadName.isEmpty {
+                            let jaccardSimilarity = self.calculateJaccardSimilarity(targetString: roadName, compareSet: eHorizonRoadNameSet, allCounts: eHorizonRoadNameCount)
+                            if jaccardSimilarity > largestJaccardSimilarity {
+                                largestJaccardSimilarity = jaccardSimilarity
+                                similarFeature = queriedFeature.feature
+                            }
+                        }
```
MaximAlien, line 303: Do we really need to implement comparison algorithm and choose the one which is the most accurate? What about choosing the one provided by Electronic Horizon by default?

CHANGELOG.md
MaximAlien: As a second thought: I'm not even sure that we should add this part, since no public APIs are affected.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Because right now we still need the feature which contains the `shield`  and `reflen`  property from the queried result to add into the `wayNameView` . If we only update the name from the EHorizon, the other properties maybe delayed compared with the name.  Now the comparison is to pick up the best match query result for the EHorizon name and then update the `wayNameView` based on previous logics.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion, I'll improve this part.

CHANGELOG.md
ShanMa1991: Thanks for the suggestion, I'll remove this part.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -296,6 +349,21 @@ extension NavigationMapView {
             })
         }
         
+        private func calculateJaccardSimilarity(targetString: String, compareSet: Set<String>, allCounts: Int) -> Double {
```
MaximAlien, line 352: Let's move this method into `Set` extension, there is a chance that it might be useful in future. Instead of passing raw `targetString` I think it'd make sense to use `Set` directly. Based on proposed algorithm below `allCounts` will not be needed.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -284,7 +333,11 @@ extension NavigationMapView {
                     }
                     
                     var hideWayName = true
-                    if smallestLabelDistance < 5 {
+                    if latestFeature != similarFeature {
+                        if self.navigationView.wayNameView.setupWith(roadFeature: similarFeature!, using: self.navigationMapView.mapView.style) {
```
MaximAlien, line 337: Please do additional check to verify that `similarFeature` and `latestFeature` do not contain `nil` to prevent force unwrapping.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -296,6 +349,21 @@ extension NavigationMapView {
...
+            var interSectionSize = 0
+            var unionSize = allCounts
+            for (_, string) in seperateStrings.enumerated() {
+                if compareSet.contains(string) {
+                    interSectionSize += 1
+                } else {
+                    unionSize += 1
+                }
+            }
+            return Double(interSectionSize) / Double(unionSize)
```
MaximAlien, line 364: I think whole algorithm can be simplified to something like:
```swift
let queriedRoadNames = Set<String>()
let electronicHorizonRoadNames = Set<String>()

let jaccardSimilarity =
    Double(queriedRoadNames.intersection(electronicHorizonRoadNames).count) /
    Double(queriedRoadNames.union(electronicHorizonRoadNames).count)
```

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -70,12 +72,41 @@ extension NavigationMapView {
...
         private func suspendNotifications() {
             NotificationCenter.default.removeObserver(self,
                                                       name: UIDevice.orientationDidChangeNotification,
                                                       object: nil)
+            NotificationCenter.default.removeObserver(self,
+                                                      name: .electronicHorizonDidUpdatePosition,
+                                                      object: nil)
+        }
+        
+        @objc func didUpdateEHorizonRoadName(_ notification: Notification) {
```
MaximAlien, line 91: Please also use `e(E)lectoronicHorizon` everywhere instead of `e(E)Horizon`.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion, I'll add the check.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion. The `queriedRoadNames ` should be kept as an array of String instead of a set of String because of the following case. 
```
electronicHorizonRoadNames: ["Tom Tom Street"]
queriedRoadNames: ["Tom Tom St", "Tom Past St"]
```

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion, due to the following comment, the function is moved to the `String` extension.

CHANGELOG.md
```diff
@@ -90,6 +90,7 @@
 * Removed deprecated and obsoleted `EventsManager` typealias, `StatusView.delegate`, `StatusView.canChangeValue`, `RouteLegProgress.upComingStep`, `StatusViewDelegate`, `DeprecatedStatusViewDelegate` and `BottomBannerViewController.init(delegate:)`. ([#2993](https://github.com/mapbox/mapbox-navigation-ios/pull/2993))
 * Fixed a potential memory leak when using `MultiplexedSpeechSynthesizer`. ([#3005](https://github.com/mapbox/mapbox-navigation-ios/pull/3005))
 * Fixed an issue where instruction banners could appear in the wrong color after switching between `Style`s. ([#2977](https://github.com/mapbox/mapbox-navigation-ios/pull/2977))
+* Improved the road name labeling accuracy during turn-by-turn navigation by using `Notification.Name.electronicHorizonDidUpdatePosition`. ([#3019](https://github.com/mapbox/mapbox-navigation-ios/pull/3019))
```
1ec5, line 93: It‘s sufficient to explain implementation details in this PR’s description; the changelog entry only needs to describe the bug that was fixed.

```suggestion
* Fixed an issue where the current road name label sometimes displayed the name of an intersecting road instead of the current road or blinked in and out. ([#3019](https://github.com/mapbox/mapbox-navigation-ios/pull/3019))
```

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -207,7 +238,7 @@ extension NavigationMapView {
                 
                 streetLabelLayer.sourceLayer = sourceLayerIdentifier
                 streetLabelLayer.lineOpacity = .constant(1.0)
-                streetLabelLayer.lineWidth = .constant(20.0)
+                streetLabelLayer.lineWidth = .expression(Expression.routeLineWidthExpression())
```
1ec5, line 241: If I’m not mistaken, #155 intentionally made this line layer a rather wide fixed width to reduce the changes that the query would come back with no results. It is possible for the query to return irrelevant results (cross streets etc.), but we would filter out those irrelevant results by comparing to the electronic horizon road name data.

Sources/MapboxNavigation/String.swift
```diff
@@ -39,4 +39,20 @@ extension String {
         }
         return digest.lazy.map { String(format: "%02x", $0) }.joined()
     }
+    
+    /**
+     Calculate the Jaccard Similarity of current string with a string set and count.
+     */
+    public func calculateJaccardSimilarity(compareSet: Set<String>, allCounts: Int, separators: CharacterSet = CharacterSet(charactersIn: " ")) -> Double {
```
1ec5, line 46: Jaccard similarity seems to be appropriate except that the Mapbox Streets source abbreviates words in English, which can frustrate any word-by-word comparison. We’re using Levenshtein–Damerau edit distance algorithm for finding matching route summary strings; it goes character by character:

https://github.com/mapbox/mapbox-navigation-ios/blob/69370c2788f2e03d88e43699a060d36a415efaa5/Sources/MapboxCoreNavigation/MinimumEditDistance.swift#L5

Both algorithms will disambiguate the most blatant differences:

Electronic horizon | Feature querying | Jacard | Levenshtein–Damerau
----|----|----|----
Main Street | Main St | **0.50** | **4**
Main Street | W Elm St | 0 | 9

Electronic horizon | Feature querying | Jacard | Levenshtein–Damerau
----|----|----|----
14th Street Northwest | 14th St NW | **0.33** | **12**
14th Street Northwest | 8th St NW | 0 | 14

But Jacard similarity won’t detect differences in abbreviated words:

Electronic horizon | Feature querying | Jacard | Levenshtein–Damerau
----|----|----|----
[West Peachtree Street Northwest](https://www.openstreetmap.org/#map=19/33.78079/-84.38759) | W Peachtree St NW | **0.25** | **15**
West Peachtree Street Northwest | Peachtree Pl NE | **0.25** | 19

Electronic horizon | Feature querying | Jacard | Levenshtein–Damerau
----|----|----|----
Franklin Street | Franklin St | **0.5** | **4**
Franklin Street | Franklin Ct | **0.5** | 5

The basic Levenshtein algorithm would be more appropriate than Levenshtein–Damerau, since the latter considers transpositions whereas these street names would only ever be abbreviated.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion, I'll remove the road width change.

CHANGELOG.md
ShanMa1991: Thanks for the suggestion, I'll update the ChangeLog.

Sources/MapboxNavigation/String.swift
ShanMa1991: Thanks for the suggestion, I'll use the `func minimumEditDistance(to word: String) -> Int` for the similarity comparison. And it's also concerning that the names from the Electronic horizon road graph would be an Array of road name strings, instead of a single one. How do we handle this issue?

1ec5: Typically, the road object will only have multiple names because the subsequent names are actually route numbers. It’s possible for a road to have multiple names per se, but that’s relatively very rare. It would be totally reasonable to compare only the road object’s first name (not route number).

ShanMa1991: Thanks for the suggestion, I'll use the first name to compare.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -70,12 +72,41 @@ extension NavigationMapView {
...
+                switch name {
+                case .name(let name):
+                    return name
+                case .code(let code):
+                    return "(\(code))"
+                }
+            }
+            
+            if electronicHorizonRoadNames.isEmpty {
+                electronicHorizonRoadNames = ["\(metadata.mapboxStreetsRoadClass.rawValue)"]
```
1ec5, line 108: This raw value could be something like `trunk`, uncapitalized and unlocalized, which wouldn’t be appropriate to show to the user and would be very unlikely to match a queried feature’s name.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -70,12 +72,41 @@ extension NavigationMapView {
...
             NotificationCenter.default.removeObserver(self,
                                                       name: UIDevice.orientationDidChangeNotification,
                                                       object: nil)
+            NotificationCenter.default.removeObserver(self,
+                                                      name: .electronicHorizonDidUpdatePosition,
+                                                      object: nil)
+        }
+        
+        @objc func didUpdateEHorizonRoadName(_ notification: Notification) {
+            guard let startingEdge = notification.userInfo?[RoadGraph.NotificationUserInfoKey.treeKey] as? RoadGraph.Edge else { return }
```
1ec5, line 92: Seems like this method should always set `electronicHorizonRoadNames`, even if only to set it to `nil`, to keep a stale name from lingering if the user transitions from a named road to an unnamed road.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -70,12 +72,41 @@ extension NavigationMapView {
...
+            
+            guard let routeController = navigationViewData.router as? RouteController,
+                  let metadata = routeController.roadGraph.edgeMetadata(edgeIdentifier: edgeIdentifier) else { return }
+            
+            electronicHorizonRoadNames = metadata.names.map { name -> String in
+                switch name {
+                case .name(let name):
+                    return name
+                case .code(let code):
+                    return "(\(code))"
```
1ec5, line 103: A code is unlikely to match a queried feature’s name, but at least it would be slightly more likely without the parentheses. We would want to show the user a shield rather than a raw route number, but if we had to show a raw route number, it wouldn’t need to be in parentheses.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -289,7 +334,14 @@ extension NavigationMapView {
                     }
                     
                     var hideWayName = true
-                    if smallestLabelDistance < 5 {
+                    if latestFeature != similarFeature {
+                        let style = self.navigationMapView.mapView.mapboxMap.style
+                        if let similarFeature = similarFeature,
+                           self.navigationView.wayNameView.setupWith(roadFeature: similarFeature,
```
1ec5, line 340: While you’re here, I didn’t realize this method was publicly using a non-Swift type:

https://github.com/mapbox/mapbox-navigation-ios/blob/83eaeba5d6421a36193f9caae403685afd1b0bf6/Sources/MapboxNavigation/WayNameView.swift#L70

We should rename it to `update(feature:style:)`, and the `feature` argument should be of type Feature (from Turf), not MBXFeature. Hopefully that’s a small enough change that it can happen as part of this PR, but feel free to break it out separately if it turns into something larger.

1ec5 (approved):
Thanks for taking a look at the string comparison logic. A couple more things can probably be simplified, and we should make sure to correct the public WayNameView method before the final release.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion, I'll remove this part

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: The `electronicHorizonRoadNames ` could be empty after the removal of the  `mapboxStreetsRoadClass `. And then the `similarFeature` would be `nil`, it would lead to the `wayNameView` hidden. So it would not cause the stale name kept when transition to an unnamed road.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion, it would return the code directly.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion. It's included in #3039 .

Sources/MapboxNavigation/String.swift
ShanMa1991: After the tests, the following cases require all the strings inside the `electronicHorizonRoadNames` to be considered:
```
electronicHorizonRoadNames : ["I 280", "Junipero Serra Freeway"]
Feature querying: ["El Monte Rd", "Junipero Serra Fwy"]
Only consider the first one: final label: "El Monte Rd"
```

Sources/MapboxNavigation/OrnamentsController.swift
1ec5: Thanks, I made the same request in https://github.com/mapbox/mapbox-navigation-ios/pull/3039#discussion_r641789971. (`MapboxCommon.Feature` is not the same as `Turf.Feature`.)

PR #1714, opened 2018-09-17
CarPlay documentation

1ec5:
Classes and methods available only in certain iOS versions are now visible in the jazzy-generated docset. This includes all CarPlay-related symbols, which depend on iOS 12, as well as some symbols related to AVSpeechSynthesizer and Measurement that depend on iOS 10. To make this work, I had to strip out `@available` attributes. I haven’t fully investigated the root cause of the problem, but it seems to lie somewhere deep in a jazzy improvisation.

Fleshed out documentation of public CarPlay-related symbols and refined their presentation in the docset’s table of contents and cover page. Deleted an old draft of a CarPlay getting started guide.

Added `@objc` attributes to a number of CarPlay-related symbols to ensure compatibility with Objective-C code.

starting points
custom category

Fixes #1682.

/cc @mapbox/navigation-ios @mapbox/docs

PR #3022, opened 2021-05-20
Restore availability macros to jazzy docset

1ec5:
SourceKit seems to finally generate a proper USR for symbols marked with availability macros. The v2.0.0-beta.9 and v2.0.0-beta.10 documentation was successfully generated without the need for the workaround added in #1714, without the side effects noted in https://github.com/mapbox/mapbox-navigation-ios/issues/1682#issuecomment-421984847. So I think we can simply remove the workaround at this point.

/cc @mapbox/navigation-ios


MaximAlien (approved):

1ec5:
Cherry-picked to the main branch in 31e1605291b5f871638eddaeb1d87f6d49bad9fd, because I needed this change in order to generate the jazzy docset for v1.4.1 using jazzy 0.13.6 and Xcode 12.4.

PR #3023, opened 2021-05-21
Fix "class" as protocol deprecation warnings

S2Ler:
It is Swift 5.4 warning:
`Using "class" keyword for protocol inheritance is deprecated;
use "AnyObject" instead`

This is the same as https://github.com/mapbox/mapbox-navigation-ios/pull/3008 but for `main` branch.

### DIFF
--- a/Sources/MapboxCoreNavigation/LegacyRouteController.swift
+++ b/Sources/MapboxCoreNavigation/LegacyRouteController.swift
@@ -5,7 +5,7 @@ import Polyline
 import MapboxMobileEvents
 import Turf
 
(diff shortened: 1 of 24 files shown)


MaximAlien (approved):
For such changes we could use cherry-pick I guess. At some point we'll have to update build jobs to use `Xcode 12.5` to really see that such change work as expected.

S2Ler:
> For such changes we could use cherry-pick I guess. At some point we'll have to update build jobs to use `Xcode 12.5` to really see that such change work as expected.

Cherry pick won't work due to difference in branches.

PR #3024, opened 2021-05-24
Fix UnimplementedLogging.logUnimplemented method

S2Ler:
The method is part of the public SDK API, and there are no guarantees
that it will be called in a thread-safe manner. The fact current
implementation isn't thread-safe was proofed by running a test with
Thread Sanitizer (TSan) enabled (see the log at the end).

- UnimplementedLogging.logUnimplemented reimplemented so that it is safe
to use from multiple threads.
- Adds a test to verify that the new implementation is safe.
- Adds a test to catch performance regressions in implementation.

Below is a log from TSan when run for old implementation:
```
WARNING: ThreadSanitizer: Swift access race (pid=14419)
  Modifying access of Swift variable at 0x00010a65a7a0 by thread T2:
    #0 UnimplementedLogging.logUnimplemented(protocolType:level:function:) <null>:12416464 (MapboxCoreNavigationTests:arm64+0x3a7464)
    #1 func1() in UnimplementedClass #1 in UnimplementedLoggingTests.testUnimplementedLoggingMultithreaded() UnimplementedLoggingTests.swift:9 (MapboxCoreNavigationTests:arm64+0xb26f4)
    #2 closure #1 in UnimplementedLoggingTests.testUnimplementedLoggingMultithreaded() UnimplementedLoggingTests.swift:24 (MapboxCoreNavigationTests:arm64+0xb1e9c)
    #3 partial apply for closure #1 in UnimplementedLoggingTests.testUnimplementedLoggingMultithreaded() <compiler-generated> (MapboxCoreNavigationTests:arm64+0xb1f34)
    #4 partial apply for thunk for @callee_guaranteed (@unowned Int) -> () <null>:12416464 (libswiftDispatch.dylib:arm64+0x5030)
    #5 _dispatch_client_callout2 <null>:12416464 (libdispatch.dylib:arm64+0x3d80)

  Previous read of size 8 at 0x00010a65a7a0 by thread T4:
    #0 UnimplementedLogging.logUnimplemented(protocolType:level:function:) <null>:12416464 (MapboxCoreNavigationTests:arm64+0x3a6f9c)
    #1 func1() in UnimplementedClass #1 in UnimplementedLoggingTests.testUnimplementedLoggingMultithreaded() UnimplementedLoggingTests.swift:9 (MapboxCoreNavigationTests:arm64+0xb26f4)
    #2 closure #1 in UnimplementedLoggingTests.testUnimplementedLoggingMultithreaded() UnimplementedLoggingTests.swift:24 (MapboxCoreNavigationTests:arm64+0xb1e9c)
    #3 partial apply for closure #1 in UnimplementedLoggingTests.testUnimplementedLoggingMultithreaded() <compiler-generated> (MapboxCoreNavigationTests:arm64+0xb1f34)
    #4 partial apply for thunk for @callee_guaranteed (@unowned Int) -> () <null>:12416464 (libswiftDispatch.dylib:arm64+0x5030)
    #5 _dispatch_client_callout2 <null>:12416464 (libdispatch.dylib:arm64+0x3d80)

  Location is global 'warned' at 0x00010a65a7a0 (MapboxCoreNavigationTests+0x000000e3a7a0)

  Thread T2 (tid=3981887, running) is a GCD worker thread

  Thread T4 (tid=3981890, running) is a GCD worker thread

SUMMARY: ThreadSanitizer: Swift access race (MapboxCoreNavigationTests:arm64+0x3a7464) in UnimplementedLogging.logUnimplemented(protocolType:level:function:)+0x770
```

- [x] Do I need to update the changelog for such fix?

### DIFF
--- a/Sources/MapboxCoreNavigation/UnimplementedLogging.swift
+++ b/Sources/MapboxCoreNavigation/UnimplementedLogging.swift
@@ -18,29 +18,62 @@ public protocol UnimplementedLogging {
 
 public extension UnimplementedLogging {
     func logUnimplemented(protocolType: Any, level: OSLogType, function: String = #function) {
-        
         let protocolDescription = String(describing: protocolType)
         let selfDescription = String(describing: type(of: self))
         
-        let description = (selfDescription, function)
+        let description = UnimplementedLoggingState.Description(typeDescription: selfDescription, function: function)
 
-        let alreadyWarned = warned.contains { elem -> Bool in
-            elem == description
-        }
-        
-        guard !alreadyWarned else {
+        guard _unimplementedLoggingState.markWarned(description) == .marked else {
             return
         }
         
         let log = OSLog(subsystem: "com.mapbox.navigation", category: "delegation.\(selfDescription)")
         let formatted: StaticString = "Unimplemented delegate method in %@: %@.%@. This message will only be logged once."
         os_log(formatted, log: log, type: level, selfDescription, protocolDescription, function)
-        unimplementedTestLogs?.append((selfDescription, function))
-        warned.append(description)
     }
 }
 
-fileprivate var warned: [(String, String)] = []
+/// Contains a list of unimplemented log descriptions so that we won't log the same warnings twice.
+/// Because this state is a global object and part of the public API it has synchronization primitive using a lock.
+/// - note: The type is safe to use from multiple threads.
+final class UnimplementedLoggingState {
+    struct Description: Equatable {
+        let typeDescription: String
+        let function: String
+    }
 
+    enum MarkingResult {
+        case alreadyMarked
+        case marked
+    }
 
-var unimplementedTestLogs: [(String, String)]? = nil
+    private let lock: NSLock = .init()
+    private var warned: [Description] = []
+
+    func markWarned(_ description: Description) -> MarkingResult {
+        lock.lock(); defer {
+            lock.unlock()
+        }
+        guard !warned.contains(description) else {
+            return .alreadyMarked
+        }
+        warned.append(description)
+        return .marked
+    }
+
+    func clear() {
+        lock.lock(); defer {
+            lock.unlock()
+        }
+        warned.removeAll()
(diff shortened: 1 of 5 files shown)


Sources/MapboxCoreNavigation/UnimplementedLogging.swift
```diff
@@ -18,29 +18,62 @@ public protocol UnimplementedLogging {
...
-        unimplementedTestLogs?.append((selfDescription, function))
-        warned.append(description)
     }
 }
 
-fileprivate var warned: [(String, String)] = []
+/// Contains a list of unimplemented log descriptions so that we won't log the same warnings twice.
+/// Because this state is a global object and part of the public API it has synchronization primitive using a lock.
+/// - note: The type is safe to use from multiple threads.
+final class UnimplementedLoggingState {
```
1ec5, line 39: This class could be nested inside UnimplementedLogging as `UnimplementedLogging.State`.

1ec5 (approved):
Thanks for adding the tests. Brings some peace of mind whenever locking is introduced.

> Do I need to update the changelog for such fix?

Yes, UnimplementedLogging is public rather than an implementation detail. We can say something about a fix for a thread-safety issue in UnimplementedLogging.

Sources/MapboxCoreNavigation/UnimplementedLogging.swift
S2Ler: Unfortunately you can't nest types inside a protocol. Swift complains with:
```
Type 'State' cannot be nested in protocol extension of 'UnimplementedLogging'
```
or
```
Type 'State' cannot be nested in protocol 'UnimplementedLogging'
```

S2Ler:
> > Do I need to update the changelog for such fix?
> 
> Yes, UnimplementedLogging is public rather than an implementation detail. We can say something about a fix for a thread-safety issue in UnimplementedLogging.

Changelog updated

PR #3025, opened 2021-05-24
Fix ImageCache image cost estimation logic

S2Ler:
How much an image takes in memory can be different for different images:
- An image created with CGContext will have a different memory size
than the same image but loaded from a jpg file.
Example: https://stackoverflow.com/a/25706554/425838.
- Size depends on the image's colour space (how many bytes needed to
encode one pixel of the image).

The fix uses the data returned by CGImage, which is the most reliable
source of information available to estimate the size of an image in
memory.


1ec5 (approved):
Thanks for thoroughly explaining the rationale for this change!

PR #3026, opened 2021-05-24
Fix versions in a jazzy docs

S2Ler:
MINOR_VERSION equal to 1.4
SHORT_VERSION equal to 1.4.0

SPM doesn't even recognise 1.4 as a valid SEMVER.
Cocoapods recognize 1.4 as a valid SEMVER, but it doesn't hurt to
specify patch version (1.4.0) as well.

The fix works correctly with "release-v2.0" branch as well.

### DIFF
--- a/docs/cover.md
+++ b/docs/cover.md
@@ -24,13 +24,13 @@ To install the MapboxNavigation framework in an application using [Swift Package
 
 1. Enter `https://github.com/mapbox/mapbox-navigation-ios.git` as the package repository and click Next.
 
-1. Set Rules to Version, Up to Next Major, and enter `${MINOR_VERSION}` as the minimum version requirement. Click Next.
+1. Set Rules to Version, Up to Next Major, and enter `${SHORT_VERSION}` as the minimum version requirement. Click Next.
 
 To install the MapboxCoreNavigation framework in another package rather than an application, run `swift package init` to create a Package.swift, then add the following dependency:
 
 ```swift
 // Latest prerelease
-.package(name: "MapboxNavigation", url: "https://github.com/mapbox/mapbox-navigation-ios.git", from: "${MINOR_VERSION}")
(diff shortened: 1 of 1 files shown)


docs/cover.md
```diff
@@ -48,10 +48,10 @@ To install the MapboxNavigation framework using [CocoaPods](https://cocoapods.or
 1. Create a [Podfile](https://guides.cocoapods.org/syntax/podfile.html) with the following specification:
    ```ruby
    # Latest stable release
-   pod 'MapboxNavigation', '~> ${MINOR_VERSION}'
+   pod 'MapboxNavigation', '~> ${SHORT_VERSION}'
```
1ec5, line 51: > Cocoapods recognize 1.4 as a valid SEMVER, but it doesn't hurt to
> specify patch version (1.4.0) as well.

Apparently this causes CocoaPods to accept new patch versions but not new minor versions. CocoaPods.org recommends just specifying the minor version in its installation instructions, so consistency with that would lead to less confusion. Any downside to keeping this line the same but changing the others?

1ec5 (approved):
Great catch!

docs/cover.md
```diff
@@ -24,13 +24,13 @@ To install the MapboxNavigation framework in an application using [Swift Package
 
 1. Enter `https://github.com/mapbox/mapbox-navigation-ios.git` as the package repository and click Next.
 
-1. Set Rules to Version, Up to Next Major, and enter `${MINOR_VERSION}` as the minimum version requirement. Click Next.
+1. Set Rules to Version, Up to Next Major, and enter `${SHORT_VERSION}` as the minimum version requirement. Click Next.
```
1ec5, line 27: document.sh defines `SHORT_VERSION` but doesn’t replace it in the readme:

https://github.com/mapbox/mapbox-navigation-ios/blob/d331990147231d8a57bf953390d63b6b178baf9a/scripts/document.sh#L42

PR #3029, opened 2021-05-25
Partially implement MapboxNavigationTests in SPM

S2Ler:
- All assets moved to TestHelper target for easier maintenance.
- Introduced a new test helper target - "CarPlayTestHelper", which holds
some mocks required for testing CarPlay. It had to be written in ObjC
because Swift forbids specific unsafe patterns. In particular:
CPNavigationSession and CPInterfaceController init methods are marked
unavailable.
- Use https://github.com/alanzeino/ios-snapshot-test-case.git in SPM,
although, it is likely to be replaced with https://github.com/pointfreeco/swift-snapshot-testing
which doesn't have all the features but has first-class Xcode and SPM
support. Or extended with a custom fork to support more configuration
points.
- All snapshot-based tests are disabled because it is not clear how to
provide configuration for https://github.com/alanzeino/ios-snapshot-test-case.git
version with SPM.
- Xcode base MapboxNavigationTests are not tested and should be fixed
separately.

This is a first step. The next ones can be:
- Migrate to https://github.com/pointfreeco/swift-snapshot-testing or fork https://github.com/alanzeino/ios-snapshot-test-case.git with the needed functionality added. 
- Restore Xcode based testing as well.
- Think how to fix some tests.

Anyway, this PR will let you write new tests using SPM.


S2Ler:
@1ec5 I need this PR to be reviewed as it blocks a lot of other PRs and work.

S2Ler:
Follow up: mapbox/mapbox-navigation-ios#3043 builds upon this PR and migrates to a new snapshots testing library.

Package.swift
```diff
@@ -81,5 +92,40 @@ let package = Package(
...
+            exclude: [
+                // Exclude snapshot testing tests because it is not clear how to configure FBSnapshotTestCase in SPM.
+                "Info.plist",
+                "MapboxNavigationTests-Bridging.h",
+                "BottomBannerSnapshotTests.swift",
+                "CarPlayCompassViewSnapshotTests.swift",
+                "GuidanceCardsSnapshotTests.swift",
+                "SnapshotTest+Mapbox.swift",
+                "NavigationViewControllerTests.swift",
+                "CarPlayManagerTests.swift", // There are issues with setting accessToken
```
1ec5, line 114: Is the issue that SPM runs tests without a bundle that includes Info.plist? In #2876, @ShanMa1991 worked around that issue by having Mapbox(Core)Navigation look for a key in UserDefaults as an optional alternative to Info.plist; the user default would only be set inside the test. (#2911 implemented a more robust solution for Info.plist keys that are needed during production, but that probably isn’t necessary here.)

Package.swift
```diff
@@ -73,6 +77,13 @@ let package = Package(
                 .process("tiles"),
             ]
         ),
+        .target(
+            name: "CarPlayTestHelper",
+            exclude: [
+                "Info.plist",
+                "CarPlayTestHelper.h",
```
1ec5, line 84: Good idea to split out this target for the Objective-C-based CarPlay mocking hack.

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -2130,6 +2238,24 @@
...
+				E2D14949265CF97D008135A3 /* Headers */,
+				E2D1494A265CF97D008135A3 /* Sources */,
+				E2D1494B265CF97D008135A3 /* Frameworks */,
+				E2D1494C265CF97D008135A3 /* Resources */,
+			);
+			buildRules = (
+			);
+			dependencies = (
+			);
+			name = CarPlayTestHelper;
```
1ec5, line 2254: If I understand correctly, this target and framework exists for running MapboxNavigationTests under SPM; do we even need it for running the tests under Carthage? The target seems to exist solely for some Objective-C code. It’s a similar story with CMapboxCoreNavigation and CTestHelper, but we simplified things a bit by keeping the targets and products unified under Carthage with an alternative umbrella header that SPM ignores.

Tests/MapboxNavigationTests/SKUTests.swift
```diff
@@ -23,26 +23,27 @@ class SKUTests: XCTestCase {
         
         XCTAssertEqual(speechSkuToken?.skuId, SkuID.navigationUser.rawValue)
     }
-    
-    func testSKUTokensMatch() {
+
+    // NOTE: Disable due to `UIApplication.shared.delegate` is nil during SPM testing.
```
1ec5, line 27: Is there an alternative way to present a view controller in this environment, or would we perhaps split out a separate XCUITest target (as tail work)?

Tests/MapboxNavigationTests/Support/AccessToken.swift
```diff
@@ -0,0 +1,18 @@
...
+}
+
+extension DirectionsCredentials {
+    static var mocked: DirectionsCredentials {
+        return .init(accessToken: .mockedAccessToken, host: nil)
+    }
+}
+
+extension String {
+    static var mockedAccessToken: String { "any" }
```
1ec5, line 17: This works, though it’s possible that lower-level libraries may introduce more rigorous client-side validation of access tokens in the future. MapboxDirectionsTests uses [a cheeky placeholder access token](https://github.com/mapbox/mapbox-directions-swift/blob/60b4a028119cef727648d981d2daccea0641138c/Tests/MapboxDirectionsTests/DirectionsTests.swift#L14-L15) that might be easier to spot while debugging.

1ec5 (approved):
I meant to hit approve. I don’t think my feedback in https://github.com/mapbox/mapbox-navigation-ios/pull/3029#pullrequestreview-673771531 is critical enough to block this PR.

Package.swift
S2Ler: I think I checked the stack trace and there wasn't a way to provide this access token at that time. I will get back to this after the chain of build system PRs will be merged.

Tests/MapboxNavigationTests/SKUTests.swift
S2Ler: We can't run XCUITests in SPM, can we? I think I need to put `#if SWIFT_PACKAGE` instead of disabling. 

Adding to the list of tail work. Will get back to this soon.

Tests/MapboxNavigationTests/Support/AccessToken.swift
S2Ler: Thank you for your suggestion, I will change from "any" to "pk.feedCafeDadeDeadBeef-BadeBede.FadeCafeDadeDeed-BadeBede" and reuse "BogusToken" name convention.

MapboxNavigation.xcodeproj/project.pbxproj
S2Ler: I added this target to the Xcode project so that the imports like https://github.com/mapbox/mapbox-navigation-ios/blob/270bf280af4edc3d5bcec0043d4e5b18f2f740a6/Tests/MapboxNavigationTests/CarPlayNavigationViewControllerTests.swift#L8 would work in both Carthage and SPM. I can wrap them in `#if SWIFT_PACKAGE`, but it seems to bring more burden for future changes. You must not forget to add this ifdef when writing code using the SPM package in Xcode.

Tests/MapboxNavigationTests/Support/AccessToken.swift
S2Ler: Fixed in https://github.com/mapbox/mapbox-navigation-ios/pull/3029/commits/473c67c8dcd37037e0b72ff3311ddcc375a8007e

Package.swift
S2Ler: So here is what happens:

`CarPlayMapViewController.setupPassiveLocationManager()` creates `PassiveLocationDataSource` which in turn accesses `Directions.shared`. `Directions.shared` is initialized with the `DirectionsCredentials()` which looks for access token only in Info.plist.

1ec5: > `CarPlayMapViewController.setupPassiveLocationManager()` creates `PassiveLocationDataSource` which in turn accesses `Directions.shared`.

It’s potentially problematic that `Directions.shared` is used in this case. What if a developer needs to specify an access token programmatically rather than in Info.plist? Most other classes have a way to specify the credentials explicitly on initialization.

S2Ler: I'm happy to address this concern as a new issue.

PR #3031, opened 2021-05-26
Bring back xcframework based Carthage support for MapboxCoreNavigation

S2Ler:
### Description
Our Carthage support is broken currently. This PR fixes it for MapboxCoreNavigation framework.
We can't add support for MapboxNavigation at the moment, because it depends on https://github.com/mapbox/mapbox-maps-ios, which doesn't support Carthage distribution at the moment. Tracking issue: mapbox/mapbox-maps-ios#66.

### Implementation

- Use xcframeworks for MapboxNavigationNative and MapboxCommon.
- MapboxNavigation shared scheme removed.
- SPMTest project removed because breaks Carthage build.
- Carthage updated to 0.38.0 on CI.
- Update CI config.


.circleci/config.yml
```diff
@@ -262,25 +262,6 @@ jobs:
...
-    macos:
-      xcode: << parameters.xcode >>
-    environment:
-      HOMEBREW_NO_AUTO_UPDATE: 1
-    steps:
-      - checkout
-      - *prepare-mapbox-file
-      - *prepare-netrc-file
-      - *add-github-to-known-hosts
-      - run: scripts/build_spm_integration_test.sh << parameters.iOS >>
```
1ec5, line 282: > SPMTest project removed because breaks Carthage build.

I think you mentioned at one point that it breaks because Carthage somehow discovers and tries to build SPMTest.xcodeproj when building mapbox-navigation-ios as a dependency, even though the project is in a subsubfolder. Any idea why it affected SPMTest.xcodeproj but not PodInstall.xcodeproj or PodInstall.xcworkspace, which remains untouched?

1ec5 (approved):

.circleci/config.yml
S2Ler: To be precise, I don't know all the details of Carthage's implementation. But let me try to make it a bit more clear what's happening because my previous explanation wasn't clear enough.

The error from Carthage is:
```
A shell task (/usr/bin/xcrun xcodebuild -project ./CarthageTest/Carthage/Checkouts/mapbox-navigation-ios/Tests/SPMTest/SPMTest.xcodeproj CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES -list) failed with exit code 74:
xcodebuild: error: Could not resolve package dependencies:
 Failed to clone repository ./CarthageTest/Carthage/Checkouts/mapbox-navigation-ios:
 fatal: repository './CarthageTest/Carthage/Checkouts/mapbox-navigation-ios' does not exist
```
even though `./CarthageTest/Carthage/Checkouts/mapbox-navigation-ios` path do exists. I've replaced full paths with "./" for brefity. 

I guess that the SPMTest project references the package in a hacky way (you can't do the same with the latest Xcode from the UI) and that causes the issues when Carthage tries to "open" this project. 

I found that digging deeper into the problem isn't worth it because I added "spm-test-job" jobs that run `swift test` commands. If this command succeeds, it almost guarantees that the SPM package is working. SPMTest project doesn't bring any more confidence. I would even say that it is misleading due to the way it uses the SPM package. Nobody will use our project in the way the SPMTest project uses it. 

So having said that above, I suggest just removing the SPMTest project and keep investing in `swift test`.

.circleci/config.yml
```diff
@@ -4,12 +4,11 @@ step-library:
   - &restore-cache
       restore_cache:
         keys:
-          - nav-cache-v5-{{ .Environment.CIRCLE_JOB }}-{{ checksum "Cartfile.resolved" }}
-          - nav-cache-v5-{{ .Environment.CIRCLE_JOB }} # used if checksum fails
```
S2Ler, line 8: This line broke the cache for me. See https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/3758/workflows/d308c9bd-af8a-4a23-8298-cbd9c24345d8/jobs/34125 run. 

@1ec5 Can you please clarify what it does and if I can remove it?

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 * MapboxCoreNavigation depends on MapboxDirections v2.0.0-beta.3 but no longer depends on MapboxAccounts. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808), [#2829](https://github.com/mapbox/mapbox-navigation-ios/pull/2829), [#2837](https://github.com/mapbox/mapbox-navigation-ios/pull/2837))
 * MapboxNavigation and MapboxCoreNavigation require iOS 11.0 or above to run. iOS 10._x_ is no longer supported. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Xcode 12.4 or above is now required for building this SDK from source.
+* Carthage v0.38 or above is now required for installing this SDK if you use Carthage. ([#3031](https://github.com/mapbox/mapbox-navigation-ios/pull/3031))
+* The Mapbox Navigation SDK supports Macs with Apple silicon.
```
S2Ler, line 16: @1ec5 Please suggest better wording for this change.

README.md
```diff
@@ -69,9 +68,6 @@ To install the MapboxNavigation framework using [CocoaPods](https://cocoapods.or
 
 1. Create a [Podfile](https://guides.cocoapods.org/syntax/podfile.html) with the following specification:
    ```ruby
-   # Latest stable release
```
S2Ler, line 72: Motivation: 6261ce794f802535dc56b5d9b5c071f291db43f9

.circleci/config.yml
1ec5: This step restores the Carthage folder (with all the checkouts and build products) from a cache from the previous build. It’s based on a checksum of the resolved file, so that any time a dependency version changes, there’s a cache miss. It’s also possible to force a cache miss by incrementing the `v5` to a different version.

CHANGELOG.md
1ec5:
```suggestion
* You can now build this SDK on Macs with Apple Silicon. ([#3031](https://github.com/mapbox/mapbox-navigation-ios/pull/3031))
```

README.md
1ec5: Yep, makes sense. I think we had removed it previously but it came back in during a merge from main.

1ec5 (approved):

.circleci/config.yml
S2Ler: As far as I see the cache with a key `- nav-cache-v5-{{ .Environment.CIRCLE_JOB }} # used if checksum fails` isn't saved anywhere.

PR #3034, opened 2021-05-27
Fix Cedar.framework path

S2Ler:
It used to be at Carthage/Build/iOS/Cedar.framework
Correct location is Carthage/Checkouts/mapbox-events-ios/Frameworks/Cedar.framework


MaximAlien (approved):

1ec5:
The Cedar dependency was incorrectly removed in #2978. I’m not sure how that change passed all the tests.

Issue #2921, opened 2021-04-09
Incorrect padding on CarPlay.


MaximAlien:
**Mapbox Navigation SDK version:** `main`, `release-v2.0`

### Steps to reproduce

 1. Run `Example-CarPlay` on both iOS and CarPlay.
 2. Start Guidance Card UI example on iOS.
 3. Observer behavior on CarPlay.

### Expected behavior

It is expected that padding on CarPlay fills only available space, instead it's always set to `.zero`. Issue is also reproducible on `main` branch of `1.x` and seems to be related to incorrect `safeArea`.

### Actual behavior

https://user-images.githubusercontent.com/1496498/114247500-12fafb80-994a-11eb-846e-babe81c081bf.mov

Issue #2893, opened 2021-03-29
Incorrect route is requested on CarPlay.


MaximAlien:
**Mapbox Navigation SDK version:** `maxim/camera-viewport-api`

### Steps to reproduce

 1. Open `Example-CarPlay` application on iOS and CarPlay, request route and start active guidance navigation.
 2. After active guidance navigation starts finish it.
 3. Request different route, but using CarPlay head-unit directly.
 4. Select route, which was found and start active guidance navigation. Observe behavior.

### Expected behavior

It is expected that CarPlay uses route, which was requested using head-unit directly, not the one which was requested on iOS. For more clarity on this follow similar steps on `1.x` version of Navigation SDK.

### Actual behavior

CarPlay uses incorrect route (route which was requested on iOS).

MaximAlien:
Cannot reproduce this issue anymore. Likely because of other fixes to CarPlay, which landed recently. I'll keep this ticket opened to make sure we verify on real CarPlay head-unit.

pavel-4:
Work as it expected on `0.4 (115)`(device: **_iPhone 8, iOS v. 14.6_**): creating a different route via CapPlay returns to customer its geometry on iPhone and not repeats older one, that was created before via iPhone.

MaximAlien:
No longer reproducible as of [v2.0.0-rc.2](https://github.com/mapbox/mapbox-navigation-ios/releases/tag/v2.0.0-rc.2).

PR #3036, opened 2021-05-27
CarPlay refactoring and clean up.

MaximAlien:
### Description

CarPlay refactoring and clean up.

- Fixes https://github.com/mapbox/mapbox-navigation-ios/issues/2921.


zugaldia:
Are there any concrete tickets that this PR addresses?

MaximAlien:
> Are there any concrete tickets that this PR addresses?

@zugaldia, I'm hoping to at least fix https://github.com/mapbox/mapbox-navigation-ios/issues/2921 in scope of this PR. And possibly https://github.com/mapbox/mapbox-navigation-ios/issues/2893. Current PR should also prevent some crashes, which might be caused by force unwrapping.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -411,27 +474,54 @@ public class CarPlayNavigationViewController: UIViewController {
         var buttons: [CPGridButton] = []
         let starImage = UIImage(named: "star", in: .mapboxNavigation, compatibleWith: nil)!
         for i in 1...5 {
-            let button = CPGridButton(titleVariants: ["\(i) star\(i == 1 ? "" : "s")"], image: starImage, handler: buttonHandler)
+            let button = CPGridButton(titleVariants: ["\(i) star\(i == 1 ? "" : "s")"],
```
1ec5, line 477: These title variants should be localized, accounting for the fact that pluralization works differently in each language. Here’s an example to follow:

https://github.com/mapbox/mapbox-navigation-ios/blob/39bcdb29a2386bd6d7fd3eb311b7b9bf0a710149/Sources/MapboxNavigation/RatingControl.swift#L104-L109 https://github.com/mapbox/mapbox-navigation-ios/blob/39bcdb29a2386bd6d7fd3eb311b7b9bf0a710149/Sources/MapboxNavigation/Resources/en.lproj/Localizable.strings#L157-L158 https://github.com/mapbox/mapbox-navigation-ios/blob/39bcdb29a2386bd6d7fd3eb311b7b9bf0a710149/Sources/MapboxNavigation/Resources/en.lproj/Localizable.stringsdict#L21-L36

Sources/MapboxNavigation/RouteVoiceController.swift
```diff
@@ -172,15 +172,16 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     }
         
     @objc open func didPassSpokenInstructionPoint(notification: NSNotification) {
-        let routeProgress = notification.userInfo![RouteController.NotificationUserInfoKey.routeProgressKey] as! RouteProgress
+        guard let routeProgress = notification.userInfo?[RouteController.NotificationUserInfoKey.routeProgressKey] as? RouteProgress else { return }
```
1ec5, line 175: This would be another place to assert in Debug configuration to avoid silently cascading an internal consistency error.

1ec5 (commented):
Thanks for performing this refactoring and reformatting. It makes it difficult to see what changed substantively, but the refactoring and reformatting was long overdue.

Some of the localization issues below are significant, but they aren’t caused by this PR and we can treat them as tail work.

> Fixes #2921.

Can you elaborate on how this issue was fixed? `CarPlayMapViewController.viewSafeAreaInsetsDidChange()` seems to be unchanged. Did something change in terms of when we set the padding?

> Current PR should also prevent some crashes, which might be caused by force unwrapping.

Which of the force-unwraps was a potential crasher but not a sign of an internal consistency error? I may have missed it due to the reformatting noise. There’s no problem with introducing guard statements, but we should be careful not to simply mute crashes so that they result in downstream unexpected behavior even in Debug configuration.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -230,15 +250,18 @@ public class CarPlayNavigationViewController: UIViewController {
...
+        navigationMapView?.addArrow(route: routeProgress.route,
+                                    legIndex: routeProgress.legIndex,
+                                    stepIndex: routeProgress.currentLegProgress.stepIndex + 1)
     }
     
     @objc func progressDidChange(_ notification: NSNotification) {
-        let routeProgress = notification.userInfo![RouteController.NotificationUserInfoKey.routeProgressKey] as! RouteProgress
-        let location = notification.userInfo![RouteController.NotificationUserInfoKey.locationKey] as! CLLocation
+        guard let routeProgress = notification.userInfo?[RouteController.NotificationUserInfoKey.routeProgressKey] as? RouteProgress,
+              let location = notification.userInfo?[RouteController.NotificationUserInfoKey.locationKey] as? CLLocation else { return }
```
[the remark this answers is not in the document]
MaximAlien: Yes, I guess it makes sense.

Example/AppDelegate+CarPlay.swift
```diff
@@ -111,11 +130,32 @@ extension AppDelegate: CarPlayManagerDelegate {
...
+    
+    func carPlayManager(_ carPlayManager: CarPlayManager,
+                        trailingNavigationBarButtonsCompatibleWith traitCollection: UITraitCollection,
+                        in template: CPTemplate,
+                        for activity: CarPlayActivity) -> [CPBarButton]? {
         switch activity {
         case .previewing:
             let disableSimulateText = "Disable Simulation"
-            let enableSimulateText =  "Enable Simulation"
+            let enableSimulateText = "Enable Simulation"
```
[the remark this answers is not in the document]
MaximAlien: Localized these two strings.

Example/UIViewController.swift
```diff
@@ -1,17 +0,0 @@
-import UIKit
-
-extension UIViewController {
-    
-    func presentAlert(_ title: String? = nil, message: String? = nil, handler: ((UIAlertAction) -> Void)? = nil) {
```
MaximAlien, line 5: Not used anywhere, can be removed.

MaximAlien:
> Thanks for performing this refactoring and reformatting. It makes it difficult to see what changed substantively, but the refactoring and reformatting was long overdue.
> 
> Some of the localization issues below are significant, but they aren’t caused by this PR and we can treat them as tail work.

I think I'll address your comments in scope of this PR.

> 
> > Fixes #2921.
> 
> Can you elaborate on how this issue was fixed? `CarPlayMapViewController.viewSafeAreaInsetsDidChange()` seems to be unchanged. Did something change in terms of when we set the padding?

Turned out to be a one-liner fix. 
```swift
```

was added when presenting `CarPlayNavigationViewController`.

> > Current PR should also prevent some crashes, which might be caused by force unwrapping.
> 
> Which of the force-unwraps was a potential crasher but not a sign of an internal consistency error? I may have missed it due to the reformatting noise. There’s no problem with introducing guard statements, but we should be careful not to simply mute crashes so that they result in downstream unexpected behavior even in Debug configuration.

Example/ViewController.swift
```diff
@@ -507,14 +508,34 @@ extension ViewController: NavigationMapViewDelegate {
...
     private func presentWaypointRemovalAlert(completionHandler approve: @escaping ((UIAlertAction) -> Void)) {
-        let title = NSLocalizedString("REMOVE_WAYPOINT_CONFIRM_TITLE", value: "Remove Waypoint?", comment: "Title of alert confirming waypoint removal")
-        let message = NSLocalizedString("REMOVE_WAYPOINT_CONFIRM_MSG", value: "Do you want to remove this waypoint?", comment: "Message of alert confirming waypoint removal")
-        let removeTitle = NSLocalizedString("REMOVE_WAYPOINT_CONFIRM_REMOVE", value: "Remove Waypoint", comment: "Title of alert action for removing a waypoint")
-        let cancelTitle = NSLocalizedString("REMOVE_WAYPOINT_CONFIRM_CANCEL", value: "Cancel", comment: "Title of action for dismissing waypoint removal confirmation sheet")
-        
-        let waypointRemovalAlertController = UIAlertController(title: title, message: message, preferredStyle: .alert)
-        let removeAction = UIAlertAction(title: removeTitle, style: .destructive, handler: approve)
-        let cancelAction = UIAlertAction(title: cancelTitle, style: .cancel, handler: nil)
+        let title = NSLocalizedString("REMOVE_WAYPOINT_CONFIRM_TITLE",
```
MaximAlien, line 511: In my opinion such styling just makes readability better.

1ec5 (approved):

Udumft (approved):
LGTM

Example/ViewController.swift
```diff
@@ -507,34 +508,48 @@ extension ViewController: NavigationMapViewDelegate {
...
+        
         [removeAction, cancelAction].forEach(waypointRemovalAlertController.addAction(_:))
         
         self.present(waypointRemovalAlertController, animated: true, completion: nil)
     }
 }
 
-// MARK: - RouteVoiceControllerDelegate methods
-
-extension ViewController: RouteVoiceControllerDelegate {
```
MaximAlien, line 526: Probably good idea would to create separate example for this.

Example/ViewController.swift
```diff
@@ -546,9 +561,3 @@ extension ViewController: NavigationViewControllerDelegate {
         clearNavigationMapView()
     }
 }
-
-// MARK: - VisualInstructionDelegate methods
-
-extension ViewController: VisualInstructionDelegate {
```
MaximAlien, line 552: Here as well, standalone example would be better I guess.

Example/WaypointConfirmationViewController.swift
```diff
@@ -1,13 +0,0 @@
-import UIKit
-
-protocol WaypointConfirmationViewControllerDelegate: NSObjectProtocol {
```
MaximAlien, line 3: Not used anywhere at this point. We also have standalone example here in https://github.com/mapbox/mapbox-navigation-ios-examples/blob/main/Navigation-Examples/Examples/Waypoint-Arrival-Screen.swift.

PR #3039, opened 2021-05-28
Update Mapbox Maps to `v10.0.0-beta.21` and Navigation Native `52.0.0`.

MaximAlien:
### Description

Update Mapbox Maps to `v10.0.0-beta.21`, Mapbox Common `13.0.0`, Navigation Native `52.0.0` and MapboxMobileEvents `1.0.2`.


Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -67,7 +67,7 @@ open class WayNameView: UIView {
     ///
     /// This method attempts to extract the road name and shield image as well as styling information and tries to display it. Return result shows if it was a success.
     @discardableResult
-    public func setupWith(roadFeature feature: MBXFeature, using style: MapboxMaps.Style?) -> Bool {
+    public func setupWith(roadFeature feature: MapboxCommon.Feature, using style: MapboxMaps.Style?) -> Bool {
```
1ec5, line 70: This is a public method, but `MapboxCommon.Feature` is an Objective-C type rather than a proper Swift type. Rename this method to `update(feature:style:)` and change the first argument’s type to `Turf.Feature`.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/3019#discussion_r641713731

Example/CustomViewController.swift
```diff
@@ -8,7 +8,7 @@ class CustomViewController: UIViewController {
     
     var destinationAnnotation: PointAnnotation! {
         didSet {
-            navigationMapView.mapView.annotations.addAnnotation(destinationAnnotation)
+//            navigationMapView.mapView.annotations.addAnnotation(destinationAnnotation)
```
1ec5, line 11: Looks like we’ll need to use the AnnotationOrchestrator to create a PointAnnotationManager before calling `syncAnnotations(_:)` on it. Interestingly, NavigationMapView will have to (and will have the opportunity to) keep track of the annotations directly.

Sources/MapboxNavigation/WayNameView.swift
MaximAlien: I've changed method signature, on the other hand I think we still need to use `MapboxCommon.Feature` instead of Turf.Feature, because current method is called from `MapboxMap.queryRenderedFeatures(at:options:completion:)`, which provides `MapboxCommon.Feature` as a result. I'm not sure why this method is public though, I think we can make it internal.

1ec5: We should convert `MapboxCommon.Feature`s to `Turf.Feature`s.

Package.swift
```diff
@@ -25,9 +25,9 @@ let package = Package(
     dependencies: [
         .package(name: "MapboxDirections", url: "https://github.com/mapbox/mapbox-directions-swift.git", .exact("2.0.0-beta.3")),
         .package(name: "MapboxGeocoder", url: "https://github.com/mapbox/MapboxGeocoder.swift.git", from: "0.14.0"),
-        .package(name: "MapboxMobileEvents", url: "https://github.com/mapbox/mapbox-events-ios.git", from: "0.10.8"),
-        .package(name: "MapboxNavigationNative", url: "https://github.com/mapbox/mapbox-navigation-native-ios.git", from: "50.0.0"),
-        .package(name: "MapboxMaps", url: "https://github.com/mapbox/mapbox-maps-ios.git", .exact("10.0.0-beta.20")),
+        .package(name: "MapboxMobileEvents", url: "https://github.com/mapbox/mapbox-events-ios.git", from: "1.0.0"),
+        .package(name: "MapboxNavigationNative", url: "https://github.com/mapbox/mapbox-navigation-native-ios.git", from: "52.0.0"),
+        .package(name: "MapboxMaps", url: "https://github.com/mapbox/mapbox-maps-ios.git", .exact("10.0.0-beta.21")),
```
1ec5, line 30: If we do go ahead and upgrade to v10.0.0-rc.1 in this PR, it’ll require us to also upgrade to Turf v2.0.0-beta.1 and MapboxDirections v2.0.0-beta.4.

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -94,23 +117,34 @@ class MapboxCoreNavigationTests: XCTestCase {
...
             
             return routeProgress?.currentLegProgress.stepIndex == 2
         }
         
         navigation.start()
         
-        locations.forEach { navigation.router!.locationManager!(navigation.locationManager, didUpdateLocations: [$0])}
+        locations.forEach {
+            navigation.router?.locationManager?(navigation.locationManager, didUpdateLocations: [$0])
+        }
```
1ec5, line 147: Tests are consistently failing, even though this PR only reformats this method:

```
Test Case '-[MapboxCoreNavigationTests.MapboxCoreNavigationTests testNewStep]' started.
2021-06-10 17:11:55.083155+0000 xctest[3367:66941] This app has attempted to access privacy-sensitive data without a usage description. The app's Info.plist must contain an “NSLocationWhenInUseUsageDescription” key with a string value explaining to the user how the app uses this data
lanes: nil
lanes: nil
lanes: nil
lanes: nil
2021-06-10 17:11:55.085502+0000 xctest[3367:67542] <nav-native> No PersistentConfig found at navigation/config.json
lanes: nil
lanes: nil
2021-06-10 17:11:55.085882+0000 xctest[3367:67542] <nav-native> Unable to find the latest version of routing tiles!
lanes: nil
lanes: nil
lanes: nil
lanes: nil
lanes: nil
2021-06-10 17:11:55.086291+0000 xctest[3367:67542] <nav-native> Async config init - Unable to determine the routing tiles version
lanes: nil
…
lanes: nil
2021-06-10 17:11:55.097732+0000 xctest[3367:66941] Warning: <nav-native> Location at timestamp 1623345115081732096ns was filtered out.
2021-06-10 17:11:55.098354+0000 xctest[3367:67575] Warning: <nav-native> Got location timestamp from the past: 1623345116081732096ns <= 1623345116095142912ns. No status will be produced.
/Users/distiller/project/Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift:149: error: -[MapboxCoreNavigationTests.MapboxCoreNavigationTests testNewStep] : Asynchronous wait failed: Exceeded timeout of 5 seconds, with unfulfilled expectations: "Expect notification 'RouteControllerDidPassSpokenInstructionPoint' from <MapboxCoreNavigation.RouteController: 0x7facf80ce1e0>".
/Users/distiller/project/Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift:150: error: -[MapboxCoreNavigationTests.MapboxCoreNavigationTests testNewStep] : XCTAssertNil failed: "Error Domain=com.apple.XCTestErrorDomain Code=0 "(null)""
2021-06-10 17:12:00.568802+0000 xctest[3367:66941] <nav-native> Path to history directory couldn't be empty
2021-06-10 17:12:00.569621+0000 xctest[3367:66941] <nav-native> PersistentConfig path: navigation/config.json
2021-06-10 17:12:00.569904+0000 xctest[3367:66941] <nav-native> Tiles path = ""
2021-06-10 17:12:00.570175+0000 xctest[3367:66941] <nav-native> TileStore instance was passed
2021-06-10 17:12:00.570438+0000 xctest[3367:66941] <nav-native> Endpoint config = "{"host":"https://example.com","isFallback":false,"minDiffInDaysToConsiderServerVersion":0,"navigatorVersion":"52.0.0","token":"***...*","userAgent":"MapboxCoreNavigation/2.0.0 MapboxDirections/2.0.0 iOS/14.4.0 (x86_64)","version":"","versionBeforeFallback":""}"
2021-06-10 17:12:00.570768+0000 xctest[3367:66941] <nav-native> Creating navigator v.52.0.0
2021-06-10 17:12:00.576749+0000 xctest[3367:67543] <nav-native> No PersistentConfig found at navigation/config.json
2021-06-10 17:12:00.605907+0000 xctest[3367:67543] <nav-native> Unable to find the latest version of routing tiles!
2021-06-10 17:12:00.606070+0000 xctest[3367:67543] <nav-native> Async config init - Unable to determine the routing tiles version
Test Case '-[MapboxCoreNavigationTests.MapboxCoreNavigationTests testNewStep]' failed (5.528 seconds).
```

Maybe the issue is that these lines are feeding the navigator locations in a tight loop without waiting for a callback indicating that the navigator has finished responding to each location. If the navigator expects to get location updates in series, then ideally it would also have a method that accepts an array of these batched locations, so that we don’t have to pump them into the navigator in a for loop.

As it is, it would be inconvenient but possible to refactor this test to wait for a progress change notification before feeding the navigator an additional location. However, that would not accurately exercise what happens in the wild when Core Location gives RouteController a batch of locations all at once. Essentially, MapboxNavigationNative should support the same semantics as `CLLocationManagerDelegate.locationManager(_:didUpdateLocations:)` as opposed to `CLLocationManagerDelegate.locationManager(_:didUpdateTo:from:)`, which was deprecated in iOS 6.0.

/cc @mapbox/navnative

Package.swift
MaximAlien: I think we'll not be able to update just yet because Mapbox Maps [v10.0.0-rc.1](https://github.com/mapbox/mapbox-maps-ios/releases/tag/v10.0.0-rc.1) depends on Mapbox Common `v14.0.1`.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -376,21 +412,22 @@ open class NavigationMapView: UIView {
...
-        
-        var preferredFramesPerSecond = FrameIntervalOptions.defaultFramesPerSecond
-        let maneuverDirections: [ManeuverDirection] = [.straightAhead, .slightLeft, .slightRight]
-        if let maneuverDirection = routeProgress.currentLegProgress.upcomingStep?.maneuverDirection,
-           maneuverDirections.contains(maneuverDirection) ||
-            (durationUntilNextManeuver > FrameIntervalOptions.durationUntilNextManeuver &&
-                durationSincePreviousManeuver > FrameIntervalOptions.durationSincePreviousManeuver) {
-            preferredFramesPerSecond = UIDevice.current.isPluggedIn ? FrameIntervalOptions.pluggedInFramesPerSecond : minimumFramesPerSecond
-        }
+        // FIXME: Mapbox Maps SDK v10.0.0-beta.21 doesn't publicly expose the ability to change `preferredFramesPerSecond`.
```
MaximAlien, line 415: Reported in https://github.com/mapbox/mapbox-navigation-ios/issues/3064.

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -92,43 +116,93 @@ class MapboxCoreNavigationTests: XCTestCase {
...
     
-    func testNewStep() {
-        // Coordinates from beginning of step[1] to end of step[2]
-        let coordinates = route.legs[0].steps[1].shape!.coordinates + route.legs[0].steps[2].shape!.coordinates
-        let locations: [CLLocation]
-        let now = Date()
-        locations = coordinates.enumerated().map { CLLocation(coordinate: $0.element,
-                                                              altitude: -1, horizontalAccuracy: -1, verticalAccuracy: -1, course: -1, speed: 10,
-                                                              timestamp: now + $0.offset) }
+    func disabled_testNewStep() {
```
MaximAlien, line 119: I've decided to disable `testNewStep` for now. With current implementation of Navigator its original implementation contradicts with other tests (e.g. `testOrderOfExecution` will start failing in case if we use `delay` between each location update and remove this [part](https://github.com/mapbox/mapbox-navigation-ios/blob/release-v2.0/Sources/MapboxCoreNavigation/NavigationService.swift#L324-L330)). Location updates without any delays at this point cause problems as well (Navigator skips certain timestamps even if they're always increasing, `testNotReroutingForAllSteps` is failing because of this).

There is a high chance that `testNotReroutingForAllSteps` will be disabled as well. Even though locally in most of the cases test passes in CircleCI it mostly fails.

Most likely failures in these tests unveil broader problems we currently have in `MapboxCoreNavigationTests`: e.g. Navigator is shared `MapboxNavigationService` between tests, in my opinion tests should be isolated from each other and work independently.

S2Ler: Yeah, I've seen other "shared state" kind of issues in other places as well.

Example/CustomViewController.swift
```diff
@@ -78,6 +80,11 @@ class CustomViewController: UIViewController {
                                                                                   right: 10.0)
         
         navigationMapView.navigationCamera.viewportDataSource = navigationViewportDataSource
+        
+        navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { [weak self] _ in
```
S2Ler, line 84: Very unfortunate API :(

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -321,14 +321,6 @@ public class MapboxNavigationService: NSObject, NavigationService {
     }
     
     public func start() {
-        // Jump to the first coordinate on the route if the location source does
```
S2Ler: Can you please comment why this is no longer needed?

Sources/MapboxCoreNavigation/Result+Expected.swift
```diff
@@ -2,7 +2,8 @@ import Foundation
 @_implementationOnly import MapboxCommon_Private
 
 extension Result {
-    init(expected: MBXExpected<AnyObject, AnyObject>) {
+
```
S2Ler, line 5: Is this empty line needed?

Sources/MapboxCoreNavigation/TileStore.swift
```diff
@@ -17,16 +17,11 @@ extension TileStore {
      worker thread; it is the responsibility of the user to dispatch to a
      user-controlled thread.
      */
-    public func containsLatestNavigationTiles(forCacheLocation cacheLocation: TileStoreConfiguration.Location = .default, completion: @escaping ContainsCompletion) {
+    public func containsLatestNavigationTiles(forCacheLocation cacheLocation: TileStoreConfiguration.Location = .default,
```
S2Ler, line 20: Thanks for that, much easier to read than having everything on one line.

S2Ler (approved):

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
azarovalex: I also faced the shared state problem when I was implementing NavNative async API, but I manage to solve it by recreating Navigator before every test. So currently there should be no issues with this shared state anymore.
There is still an issue that NavNative skips some callbacks if we send all the locations in one batch, this can be fixed by using ReplayLocationManager with big `speedMultiplier`.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -132,28 +132,53 @@ class NavigationServiceTests: XCTestCase {
         }
     }
 
-    func testNotReroutingForAllSteps() {
-        let navigation = dependencies.navigationService
-        let route = navigation.route
+    func disabled_testNotReroutingForAllSteps() {
```
azarovalex, line 135: I guess we need a ticket to fix those tests as soon as possible because they help us to spot bugs quite often, I can try replacing the `forEach` loops with the `ReplayLocationManager`, in theory, it will fix these test.

S2Ler:
Would be good to refine commits. 30 commits is too much to understand.

Sources/MapboxCoreNavigation/NavigationService.swift
MaximAlien: Hmm, I think this file is already outdated. This change was reverted back.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
MaximAlien: > Would be good to refine commits. 30 commits is too much to understand.

I usually squash my commits when landing PRs, will do this in this PR as well. If it's easier to review as one commit I can interactively rebase I guess, or softly reset to parent branch.

1ec5 (approved):

MaximAlien:
> Would be good to refine commits. 30 commits is too much to understand.

I usually squash my commits when landing PRs, will do this in this PR as well. If it's easier to review as one commit I can interactively rebase I guess, or softly reset to parent branch.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
MaximAlien: I agree, these tests were disabled just to land current PR, as it brings other important changes. I'll create separate ticket to tackle outstanding problem with tests.

Sources/MapboxCoreNavigation/Result+Expected.swift
MaximAlien: Not too critical, but I try to follow style which adds empty line for all properties, methods etc, which were declared right inside another class or struct. I think we have similar style in most of the other places.

PR #3040, opened 2021-05-28
Offline routing

### LINKED ISSUE #3037: Offline Routing race condition
Offline routing is not always working. Sometimes it does not fire a result callback. It turns out that this is caused by race condition for deallocating Native Router instance before it completes its task. As a resolution suggested by NN team - we should store `MapboxNavigationNative.Router` instead of [de]allocating it for each request.

### Steps to reproduce

 1. Run the Example App
 2. Let Ambient cache download tiles a bit
 3. Disable internet connection
 4. try to request a route around your current location

### Expected behavior
Route is created using offline cache

### Actual behavior
Nothing happens

Udumft:
### Description
Resolves #3037 
Fixed a bug where Router was deallocated before it actually schedules a route calculation, which results in callback not being called.

### Implementation
Moved MBNNRouter instance to be a part of Navigator wrapper. Such instance has low overhead of keeping alive, and is recommended by NN to persist it.


bamx23 (approved):

azarovalex (approved):
LGTM

PR #3047, opened 2021-06-02
Fix offline routing

bamx23:
### Description

`Directions.calculateOffline` uses only `path` of Directions URL and no parameters. Even with simple route options, this leads to conflict in Polyline versions and the response will be parsed with a wrong route geometry.

### Implementation

Full URL should be provided to a router.


Udumft (approved):
Thanks for covering this!

PR #3050, opened 2021-06-03
Fixing Navigation end event loss of cancelling

Udumft:
vk-694-cancel-event-loss: added forcing EventsManager datasource to o……verride nilling weak reference during deallocation.

### Description
Resolves an issue, where `Navigation end` event was not logged if trips was cancelled.

### Implementation
Issue is caused by nilling weak reference to a `NavigationService` during it's deinit stage. To avoid that I've added forcing a strong reference during event logging.

### DIFF
--- a/Sources/MapboxCoreNavigation/NavigationEventsManager.swift
+++ b/Sources/MapboxCoreNavigation/NavigationEventsManager.swift
@@ -22,7 +22,23 @@ open class NavigationEventsManager {
     
     var outstandingFeedbackEvents = [CoreFeedbackEvent]()
     
-    weak var dataSource: EventsManagerDataSource?
+    func withBackupDataSource(_ forcedDataSource: EventsManagerDataSource, action: () -> Swift.Void) {
+        backupDataSource = forcedDataSource
+        action()
+        backupDataSource = nil
+    }
+    
+    private var backupDataSource: EventsManagerDataSource?
+    private weak var _dataSource: EventsManagerDataSource?
+    var dataSource: EventsManagerDataSource?
+    {
+        get {
+            return _dataSource ?? backupDataSource
+        }
+        set {
+            _dataSource = newValue
+        }
+    }
     
     /**
      Indicates whether the application depends on MapboxNavigation in addition to MapboxCoreNavigation.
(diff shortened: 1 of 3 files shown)


Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -266,7 +266,9 @@ public class MapboxNavigationService: NSObject, NavigationService {
     
     deinit {
         suspendNotifications()
-        endNavigation()
+        eventsManager.withBackupDataSource(self) {
```
bamx23, line 269: Looks too hacky, but I can't really come to a better solution. Making `dataSource` strong seems to be even worse.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -22,7 +22,23 @@ open class NavigationEventsManager {
     
     var outstandingFeedbackEvents = [CoreFeedbackEvent]()
     
-    weak var dataSource: EventsManagerDataSource?
+    func withBackupDataSource(_ forcedDataSource: EventsManagerDataSource, action: () -> Swift.Void) {
```
bamx23, line 25:
```suggestion
    func withBackupDataSource(_ forcedDataSource: EventsManagerDataSource, action: () -> Void) {
```

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -266,7 +266,9 @@ public class MapboxNavigationService: NSObject, NavigationService {
...
+            endNavigation()
```
bamx23, line 270: NIT: A potential issue might be if there are async operations inside

bamx23 (approved):

Sources/MapboxCoreNavigation/NavigationService.swift
Udumft: True. But I don't see any better options

Sources/MapboxCoreNavigation/NavigationService.swift
Udumft: Same story...

PR #3055, opened 2021-06-07
Remove some redundant force unwraps

S2Ler:
With this change, the API for `NavigationService` is safer, although it is a breaking change. Would be good to land before RC.

### DIFF
--- a/Sources/MapboxCoreNavigation/NavigationService.swift
+++ b/Sources/MapboxCoreNavigation/NavigationService.swift
@@ -49,12 +49,12 @@ public protocol NavigationService: CLLocationManagerDelegate, RouterDataSource, 
     /**
      The router object that tracks the user’s progress as they travel along a predetermined route.
      */
-    var router: Router! { get }
+    var router: Router { get }
     
     /**
      The events manager, responsible for all telemetry.
      */
-    var eventsManager: NavigationEventsManager! { get }
+    var eventsManager: NavigationEventsManager { get }
     
     /**
      The route along which the user is expected to travel, plus its index in the `RouteResponse`, if applicable.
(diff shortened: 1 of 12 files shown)


ShanMa1991 (approved):

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -148,12 +152,12 @@ public class MapboxNavigationService: NSObject, NavigationService {
     /**
      The active router. By default, a `RouteController`.
      */
-    public var router: Router!
+    public var router: Router { _router! }
```
S2Ler, line 155: `router` is set after `super.init` so it is "safe" to make it non-optional. If it can crash it will crash in init method.

1ec5: I like this approach of keeping the implicitly unwrapped optional but making it an internal implementation detail instead of exposing it to developers.

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -148,12 +152,12 @@ public class MapboxNavigationService: NSObject, NavigationService {
...
     
     /**
      The events manager. Sends telemetry back to the Mapbox platform.
      */
-    public var eventsManager: NavigationEventsManager!
+    public var eventsManager: NavigationEventsManager { _eventsManager! }
```
1ec5, line 160: This and other changes to the public API in this pull request are technically backwards-incompatible, since it’s possible for application code to perform a `nil` check that would compile previously but not anymore. A short blurb in the changelog about these properties would be nice.

1ec5 (approved):

Sources/MapboxCoreNavigation/NavigationService.swift
S2Ler: Can't agree more!

Sources/MapboxCoreNavigation/NavigationService.swift
S2Ler: Done

PR #3058, opened 2021-06-07
Fix #3056 RouteController shouldDiscard location logic

### LINKED ISSUE #3056: [Bug] RouterDelegate.router(_:shouldDiscard:) handled incorrectly
**Mapbox Navigation SDK version:** "release-v2.0 branch"
https://github.com/mapbox/mapbox-navigation-ios/blob/bb6c0c1d3e69b765aa986a18eb8fa98a775802eb/Sources/MapboxCoreNavigation/RouteController.swift#L247-L252

This code seems to have inverted logic. If `shouldDiscard` returns true, then the discard won't happen. And visa versa.

S2Ler:
This is kind of bug with a "breaking change" fix because some users can depend on this bug now as our tests have shown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@
 * Fixed an issue where 'Navigation cancel' event might be not logged. ([#3050](https://github.com/mapbox/mapbox-navigation-ios/pull/3050))
 * Fixed an issue where offline route calculation might hang up. ([#3040](https://github.com/mapbox/mapbox-navigation-ios/pull/3040))
 * Fixed the moment of custom feedback event creation. ([#3049](https://github.com/mapbox/mapbox-navigation-ios/pull/3049))
+* Fixed a bug in `RouterDelegate.router(_:shouldDiscard:)` handling. If you implemented this method, you will need to reverse the value you return. Previously, if you returned `true`, the `Router` wouldn't discard the location. ([#3058](https://github.com/mapbox/mapbox-navigation-ios/pull/3058))
 
 ## v1.4.0
 
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -15,7 +15,7 @@ import Turf
 open class RouteController: NSObject {
     public enum DefaultBehavior {
         public static let shouldRerouteFromLocation: Bool = true
-        public static let shouldDiscardLocation: Bool = true
+        public static let shouldDiscardLocation: Bool = false
         public static let didArriveAtWaypoint: Bool = true
         public static let shouldPreventReroutesWhenArrivingAtWaypoint: Bool = true
         public static let shouldDisableBatteryMonitoring: Bool = true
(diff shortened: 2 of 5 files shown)


Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -247,7 +247,7 @@ open class RouteController: NSObject {
     public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
         guard let location = locations.last else { return }
         
-        guard delegate?.router(self, shouldDiscard: location) ?? DefaultBehavior.shouldDiscardLocation else {
+        guard !(delegate?.router(self, shouldDiscard: location) ?? DefaultBehavior.shouldDiscardLocation) else {
```
Udumft, line 250: Maybe we should also invert `DefaultBehavior.shouldDiscardLocation` to maintain existing 'default' behavior? This will not only also make the fix 'less breaking' since many users rely on defaults but also won't require that many unit tests changes below

Tests/MapboxCoreNavigationTests/RouteControllerTests.swift
```diff
@@ -65,3 +68,13 @@ extension RouteControllerTests: RouterDataSource {
         return NavigationLocationManager.self
     }
 }
+
+extension RouteControllerTests: RouterDelegate {
+    func router(_ router: Router, didRefresh routeProgress: RouteProgress) {
```
Udumft, line 73: Why did you add this implementation too?

zugaldia:
Looks like this is one we want to highlight in the release notes too.

Tests/MapboxCoreNavigationTests/RouteControllerTests.swift
S2Ler: Xcode said that I must implement this method, I will remove and check again, Thanks.

Sources/MapboxCoreNavigation/RouteController.swift
S2Ler: Great idea! I will give it a try.

CHANGELOG.md
```diff
@@ -99,6 +99,7 @@
 * Fixed an issue where 'Navigation cancel' event might be not logged. ([#3050](https://github.com/mapbox/mapbox-navigation-ios/pull/3050))
 * Fixed an issue where offline route calculation might hang up. ([#3040](https://github.com/mapbox/mapbox-navigation-ios/pull/3040))
 * Fixed the moment of custom feedback event creation. ([#3049](https://github.com/mapbox/mapbox-navigation-ios/pull/3049))
+* Fixed `RouterDelegate.router(_:shouldDiscard:)` handling. If you implement this method, you will need to reverse the value you return from this method. Previously, if you returned `true`, the `Router` wouldn't discard the location.
```
Udumft, line 102:
```suggestion
* Fixed a bug in `RouterDelegate.router(_:shouldDiscard:)` handling. If you implemented this method, you will need to reverse the value you return. Previously, if you returned `true`, the `Router` wouldn't discard the location. ([#3058](https://github.com/mapbox/mapbox-navigation-ios/pull/3058))
```

Udumft (approved):
LGTM with NIT correction.

CHANGELOG.md
S2Ler: Thank you for the fix!

PR #3059, opened 2021-06-08
Catch up with `main`.

MaximAlien:
### Description

Catch up `release-v2.0` branch with `main`.


CHANGELOG.md
```diff
@@ -102,6 +102,7 @@
 
 ## v1.4.0
 
+* Fixed the moment of custom feedback event creation. ([#2495](https://github.com/mapbox/mapbox-navigation-ios/pull/2495))
```
MaximAlien, line 105: Not sure why we add this changelog item here, `1.4.0` was released already.

S2Ler: Nice catch, it was an old resurrected PR. So we must move this changelog to an unreleased v.1.4.1 version as far as I see.

1ec5 (approved):

PR #3061, opened 2021-06-08
Fix the force wrapping crash in active guidance

### LINKED ISSUE #3060: Navigator starts before the present of navigationViewController in Example
**Mapbox Navigation SDK version:** `release-v2.0`

Crash sometimes occurs right after starting active guidance navigation in Example app.

Crash log:
[Example_2021-06-08-153112_MacBook-Pro.txt](https://github.com/mapbox/mapbox-navigation-ios/files/6616408/Example_2021-06-08-153112_MacBook-Pro.txt)

ShanMa1991:
### Description
Fix #3060

### DIFF
--- a/Example/ViewController+FreeDrive.swift
+++ b/Example/ViewController+FreeDrive.swift
@@ -138,10 +138,9 @@ extension ViewController {
     }
     
     func edgeNames(identifier: RoadGraph.Edge.Identifier) -> [String] {
-        let passiveLocationDataSource = (navigationMapView.mapView.location.locationProvider as! PassiveLocationManager).dataSource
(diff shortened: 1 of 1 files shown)


MaximAlien (approved):

MaximAlien (commented):
It's beneficial to have such check, at the same time root cause of this crash is a bit different. It occurs because at the time of getting a notification `navigationMapView` is `nil`, here's example:

Screen Shot 2021-06-09 at 12 15 56 PM

We should find the reason why `navigationMapView` is still `nil`, but `Navigator` was already created.

S2Ler:
Where are we at with this PR?

ShanMa1991:
@S2Ler  Jill is working on investigating this issue #3060 , the problem is not from the force wrapping, but should be the navigator and the navigationMapView initialization sequence. But it would be safe to close this pr.

Issue #3062, opened 2021-06-08
Crash previewing a step from the step table — RouteLegProgress.init(leg:stepIndex:spokenInstructionIndex:)


dmiluski:
I'm running into a crasher while running on Nav SDK betas. We should be able to update SDKs to latest in our next sprint, but wanted to share in case this helped find an issue in the SDKs

**Mapbox Navigation SDK version:**

-v2 beta 8

### Steps to reproduce

 1. Begin Navigation
 2. Tap top banner to show steps of the directions
 3. Scroll to the bottom, tap arrive at destination (sometime crash)

### Expected behavior

Not crash

### Actual behavior

Crash

![stepCrash](https://user-images.githubusercontent.com/5083390/121575392-2ecb5e00-c9dc-11eb-87f8-a455636e75ed.gif)

```
Crashed: com.apple.main-thread
0  Straightaway                   0x102e7dc08 specialized RouteLegProgress.init(leg:stepIndex:spokenInstructionIndex:) + 4367932424 (RouteLegProgress.swift:4367932424)
1  Straightaway                   0x102e7c5f8 RouteLegProgress.init(leg:stepIndex:spokenInstructionIndex:) + 4367926776 (<compiler-generated>:4367926776)
2  Straightaway                   0x102d58d44 specialized NavigationViewController.topBanner(_:didSelect:stepIndex:cell:) + 960 (NavigationViewController.swift:960)
3  Straightaway                   0x102d8dbb8 protocol witness for StepsViewControllerDelegate.stepsViewController(_:didSelect:stepIndex:cell:) in conformance TopBannerViewController + 4366949304 (<compiler-generated>:4366949304)
4  Straightaway                   0x102d79af8 StepsViewController.tableView(_:didSelectRowAt:) + 206 (StepsViewController.swift:206)
5  Straightaway                   0x102d79c14 @objc StepsViewController.tableView(_:didSelectRowAt:) + 4366867476 (<compiler-generated>:4366867476)
6  UIKitCore                      0x184ea3f3c -[UITableView _selectRowAtIndexPath:animated:scrollPosition:notifyDelegate:isCellMultiSelect:] + 1104
7  UIKitCore                      0x184ea3ad4 -[UITableView _selectRowAtIndexPath:animated:scrollPosition:notifyDelegate:] + 108
8  UIKitCore                      0x184ea42a0 -[UITableView _userSelectRowAtPendingSelectionIndexPath:] + 268
[...]

PR #3063, opened 2021-06-08
Fix crash related with the stepIndex in RouteStepProgress

### LINKED ISSUE #2997: currentLegProgress.currentStepProgress is `nil` when navigation start
**Mapbox Navigation SDK version:** v2.0

### Steps to reproduce

 1.  choose  `Features/Location/City run` in Smulator
 2. choose a destination
 3.  start navigation in simulation mode

### Expected behavior
Navigation starts
### Actual behavior
The navigation stops related with the `currentStepProgress`

ShanMa1991:
### Description
This pr is to fix #2997, and also the crush reported in the #3062 

### Implementation
The `stepIndex` would trigger the `currentStepProgress`  change, and the `currentStep` with the index limit. It's safer to put it after the  `currentStepProgress` initialization and the `precondition` check.

### DIFF
--- a/Sources/MapboxCoreNavigation/RouteLegProgress.swift
+++ b/Sources/MapboxCoreNavigation/RouteLegProgress.swift
@@ -142,11 +142,11 @@ open class RouteLegProgress: Codable {
      */
     public init(leg: RouteLeg, stepIndex: Int = 0, spokenInstructionIndex: Int = 0) {
         self.leg = leg
-        self.stepIndex = stepIndex
         
         precondition(leg.steps.indices.contains(stepIndex), "It's not possible to create RouteLegProgress without any steps or when stepIndex is higher than steps count.")
         
         currentStepProgress = RouteStepProgress(step: leg.steps[stepIndex], spokenInstructionIndex: spokenInstructionIndex)
+        self.stepIndex = stepIndex
     }
 
     typealias StepIndexDistance = (index: Int, distance: CLLocationDistance)


Sources/MapboxCoreNavigation/RouteLegProgress.swift
```diff
@@ -142,11 +142,11 @@ open class RouteLegProgress: Codable {
      */
     public init(leg: RouteLeg, stepIndex: Int = 0, spokenInstructionIndex: Int = 0) {
         self.leg = leg
-        self.stepIndex = stepIndex
         
         precondition(leg.steps.indices.contains(stepIndex), "It's not possible to create RouteLegProgress without any steps or when stepIndex is higher than steps count.")
         
         currentStepProgress = RouteStepProgress(step: leg.steps[stepIndex], spokenInstructionIndex: spokenInstructionIndex)
+        self.stepIndex = stepIndex
```
1ec5, line 149: Not sure I understand how the previous location of this statement was causing the crash in #3062 or the nil in #2997 (which itself is strange because the property is non-optional). Within this initializer, the `stepIndex` argument shadows the `stepIndex` instance property, and nothing else in this initializer reads from the `RouteLegProgress.stepIndex` property. (Neither `RouteLegProgress.leg` nor `RouteLegProgress.currentStepProgress` has a setter that depends on `RouteLegProgress.stepIndex`.) Since nothing reads from the property, it should’t matter where we set it, as long as it’s after setting `RouteLegProgress.leg`.

ShanMa1991: for the #2997, the break issue comes in the following and several other places with the `RouteLegProgress.init(leg:stepIndex:spokenInstructionIndex:)`, and the `nil` of the `routeProgress.currentLegProgress.currentStepProgress`. 
https://github.com/mapbox/mapbox-navigation-ios/blob/c32cf61f0038dceb05984e56a7913821dda23c29/Sources/MapboxNavigation/NavigationViewController.swift#L1054-L1057
where after setting the initializing of the `RouteLegProgress` and reach the `RouteLegProgress.currentStep`, then the `currentStep` is nil. I remembered that I've checked the index of the `legIndex` and `stepIndex`,and it happened just after the navigation session start, they're not out of the bound.  But after that fix, and the problem could not reproduce anymore.

ShanMa1991:
Close this pr because cannot reproduce this crush

PR #3068, opened 2021-06-10
Add `.pathShape(for:)` and `.position(for:)` methods to RoadGraph.

### LINKED ISSUE #3035: Add new methods to RoadGraph
There are new methods in MBNNGraphAccessor, which are not yet supported in our RoadGraph API.

- [x] `getPathShape(for path: MBNNGraphPath) -> [CLLocation]?`
- [x] `getPositionCoordinate(for position: MBNNGraphPosition) -> CLLocation?`

We should add corresponding public methods in our Navigation SDK.

azarovalex:
### Description
Fixes #3035.

### DIFF
--- a/Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
+++ b/Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
@@ -32,6 +32,35 @@ public final class RoadGraph {
         }
         return LineString(locations.map { $0.coordinate })
     }
+    
+    /**
+     Returns a line string geometry corresponding to the given path.
+     
+     - returns: A line string corresponding to the given path, or `nil` if any of path edges is inaccessible.
+     */
+    public func shape(of path: Path) -> LineString? {
+        let nativePath = GraphPath(edges: path.edgeIdentifiers.map { NSNumber(value: $0) },
+                                   percentAlongBegin: path.fractionFromStart,
+                                   percentAlongEnd: path.fractionToEnd,
+                                   length: path.length)
+        guard let locations = native.getPathShape(for: nativePath) else {
+            return nil
+        }
+        return LineString(locations.map { $0.coordinate })
+    }
+    
+    /**
+     Returns a point corresponding to the given position.
+     
+     - returns: A point corresponding to the given position, or `nil` if the edge is inaccessible.
+     */
+    public func shape(of position: Position) -> Point? {
+        let nativePosition = GraphPosition(edgeId: UInt64(position.edgeIdentifier), percentAlong: position.fractionFromStart)
+        guard let location = native.getPositionCoordinate(for: nativePosition) else {
+            return nil
+        }
+        return Point(location.coordinate)
+    }
 
     init(_ native: GraphAccessor) {
         self.native = native
--- a/Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
+++ b/Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
@@ -12,7 +12,7 @@ public final class RoadGraph {
     /**
      Returns metadata about the edge with the given edge identifier.
      
-     - returns: Metadata about the edge with the given edge identifier, or `nil` if the edge is inaccessible.
+     - returns: Metadata about the edge with the given edge identifier, or `nil` if the edge is not in the cache.
      */
     public func edgeMetadata(edgeIdentifier: Edge.Identifier) -> Edge.Metadata? {
         if let edgeMetadata = native.getEdgeMetadata(forEdgeId: UInt64(edgeIdentifier)) {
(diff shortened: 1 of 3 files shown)


Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
```diff
@@ -32,6 +32,35 @@ public final class RoadGraph {
         }
         return LineString(locations.map { $0.coordinate })
     }
+    
+    /**
+     Returns a line string geometry corresponding to the given path.
+     
+     - returns: A line string corresponding to the given path, or `nil` if any of path edges is inaccessible.
+     */
+    public func pathShape(for path: Path) -> LineString? {
```
1ec5, line 41: “Path” seems redundant in this method signature. Since a `RoadGraph.Path` is already being passed in, it can be just `shape(of:)`.

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
```diff
@@ -32,6 +32,35 @@ public final class RoadGraph {
...
+        }
+        return LineString(locations.map { $0.coordinate })
+    }
+    
+    /**
+     Returns a point corresponding to the given position.
+     
+     - returns: A point corresponding to the given position, or `nil` if the edge is inaccessible.
+     */
+    public func position(for position: Position) -> Point? {
```
1ec5, line 57: “Position” is what’s being passed in, not what’s returned, so rename to `shape(of:)`. That’s the same signature as the method above, which accomplishes the same effect as what I suggested in https://github.com/mapbox/mapbox-navigation-ios/issues/3035#issuecomment-850520609.

azarovalex: Renamed to `shape(of position: Position) -> Point?`. Hope I got your point correctly.

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
```diff
@@ -32,6 +32,35 @@ public final class RoadGraph {
         }
         return LineString(locations.map { $0.coordinate })
     }
+    
+    /**
+     Returns a line string geometry corresponding to the given path.
+     
+     - returns: A line string corresponding to the given path, or `nil` if any of path edges is inaccessible.
```
S2Ler, line 39: Maybe it is just me, but I don't understand what `inaccessible` means in this context.

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
```diff
@@ -32,6 +32,35 @@ public final class RoadGraph {
...
+     */
+    public func shape(of path: Path) -> LineString? {
+        let nativePath = GraphPath(edges: path.edgeIdentifiers.map { NSNumber(value: $0) },
+                                   percentAlongBegin: path.fractionFromStart,
+                                   percentAlongEnd: path.fractionToEnd,
+                                   length: path.length)
```
S2Ler, line 45: Looks like this is a good candidate for convenience init method:
```swift
extension GraphPath {
    convenience init(_ path: RoadGraph.Path) {
        self.init(edges: path.edgeIdentifiers.map { NSNumber(value: $0) },
                  percentAlongBegin: path.fractionFromStart,
                  percentAlongEnd: path.fractionToEnd,
                  length: path.length)
    }
}

```

S2Ler:
Can we add tests for these additions?

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
1ec5: This method allows an arbitrary path to be passed in, so it probably refers to the case where the application passes in a long-irrelevant path that’s no longer connected to the current electronic horizon.

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
1ec5: For consistency, also a convenience initializer for GraphPosition based on Position.

1ec5:
> Can we add tests for these additions?

The electronic horizon–related functionality is quite interconnected, possibly making it necessary to create an integration test rather than something more granular, but it would be helpful to identify the requirements for making something in RoadGraph testable.

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
S2Ler: We can add something along these lines to the documentation comment.

chezzdev:
@azarovalex I guess we should proceed with this PR to land it before API freeze.

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
azarovalex: image
This is a comment from the corresponding NavNative method, the folks say that this situation can happen if the navigation tile not in the cache anymore, I updated the comment accordingly.

Sources/MapboxCoreNavigation/EHorizon/RoadGraph.swift
azarovalex: Thanks for the nit! I updated the code.

S2Ler: Did you miss the push? I don't see the updated code.

chezzdev (approved):

S2Ler (approved):
Looks good!

PR #3072, opened 2021-06-11
Fix issues in ImageDownload and ImageDownloader

### LINKED ISSUE #3067: Crash in ImageDownloader
**Mapbox Navigation SDK version:** 2.0.0-beta.12

### Steps to reproduce

```
Crashed: com.apple.NSURLSession-delegate
0  OneTap                         0x1007e72b0 ImageDownloader.activeOperation(with:) + 72 (ImageDownloader.swift:72)
1  OneTap                         0x1007e8530 specialized ImageDownloader.urlSession(_:dataTask:didReceive:completionHandler:) + 98 (<compiler-generated>:98)
2  OneTap                         0x1007e7408 @objc ImageDownloader.urlSession(_:dataTask:didReceive:completionHandler:) + 4339856392 (<compiler-generated>:4339856392)
3  CFNetwork                      0x199c15038 _CFNetworkHTTPConnectionCacheSetLimit + 155128
```

I have no exact steps to reproduce. The crash happened during a simulated navigation session here:

https://github.com/mapbox/mapbox-navigation-ios/blob/48c312c3fd2830a08b2079a58db535dfcaa8508e/Sources/MapboxNavigation/ImageDownloader.swift#L72

due to no synchronization in this method for `activeOperation` call:

https://github.com/mapbox/mapbox-navigation-ios/blob/48c312c3fd2830a08b2079a58db535dfcaa8508e/Sources/MapboxNavigation/ImageDownloader.swift#L97-L103

All 3 `URLSessionDataDelegate` methods should have synchronization for `operations` dictionary.

cc: @truburt @1ec5 @MaximAlien

S2Ler:
Fixes #3067

- Fix a lot of thread-safety issues.
- Fix incorrect handling of errors:
   - Completions are called two times for HTTP status code >= 400.
   - `ImageDownload` operation wasn't properly transition into finished
      state.
- `ImageDownload` operation didn't have proper `isReady` state.
- Increase test coverage for these types.

#### `ImageDownloader`
- `ImageDownloader` is fixed by properly using Dispatch Queue: perform protected state read/write inside the queue.
-  concurrent `accessQueue` replaced by a serial one, because most accesses to the queue modifies the protected state that require a barrier to be placed. That makes the `accessQueue` merely a serial queue. 
#### `ImageDownload`
`ImageDownload` is a subclass of NSOperation. 
- The implementation is improved with a simpler to use `State` variable. No more multiple flags juggling. 
- A queue replaced with a lock. 
- Replace `objc_sync_enter` with a lock.

### DIFF
--- a/Sources/MapboxNavigation/ImageDownloader.swift
+++ b/Sources/MapboxNavigation/ImageDownloader.swift
@@ -11,36 +11,35 @@ protocol ReentrantImageDownloader {
 }
 
 class ImageDownloader: NSObject, ReentrantImageDownloader, URLSessionDataDelegate {
-    private var sessionConfiguration: URLSessionConfiguration = URLSessionConfiguration.default
+    private let sessionConfiguration: URLSessionConfiguration
 
-    lazy private var urlSession: URLSession = {
-        return URLSession.init(configuration: sessionConfiguration, delegate: self, delegateQueue: nil)
+    private lazy var urlSession: URLSession = {
+        return URLSession(configuration: sessionConfiguration, delegate: self, delegateQueue: nil)
     }()
 
-    private var downloadQueue: OperationQueue
-    private var accessQueue: DispatchQueue
+    private let downloadQueue: OperationQueue
+    private let accessQueue: DispatchQueue
 
-    private var operationType: ImageDownload.Type = ImageDownloadOperation.self
+    private var operationType: ImageDownload.Type
     private var operations: [URL: ImageDownload] = [:]
 
-    private var headers: [String: String] = ["Accept": "image/*;q=0.8"]
+    private let headers: [String: String] = ["Accept": "image/*;q=0.8"]
 
-    override init() {
+    init(sessionConfiguration: URLSessionConfiguration = .default,
+         operationType: ImageDownload.Type = ImageDownloadOperation.self) {
+        self.sessionConfiguration = sessionConfiguration
+        self.operationType = operationType
+
         self.downloadQueue = OperationQueue()
         self.downloadQueue.name = Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloader"
-        self.accessQueue = DispatchQueue(label: Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloaderInternal", attributes: .concurrent)
-    }
+        self.accessQueue = DispatchQueue(label: Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloaderInternal")
+        super.init()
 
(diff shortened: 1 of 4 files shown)


Tests/MapboxNavigationTests/ImageDownloaderTests.swift
```diff
@@ -135,4 +124,134 @@ class ImageDownloaderTests: XCTestCase {
...
+        XCTAssertNotNil(errorReturned)
+    }
+
+    func testDownloadWith400StatusCode() {
+        var imageReturned: UIImage?
+        var dataReturned: Data?
+        var errorReturned: Error?
+        let imageDownloaded = expectation(description: "Image Downloaded")
+
+        let faultyUrl = URL(string: "https://www.mapbox.com/incorrect")!
```
S2Ler, line 154: Not very reliable way of testing 4xx status code. Will extend the url spy with ability to emulate errors

Sources/MapboxNavigation/ImageDownloader.swift
```diff
@@ -11,47 +11,46 @@ protocol ReentrantImageDownloader {
...
-            self.operationType = op
-        }
+        self.accessQueue = DispatchQueue(label: Bundle.mapboxNavigation.bundleIdentifier! + ".ImageDownloaderInternal")
+        super.init()
+
+        // NOTE: `lazy var` isn't thread safe while `init` methods are.
+        //       `lazy var` for `urlSession` is used to specify `self` as a delegate which only available after
+        //       all properties are initialized ans `super.init` is called if appropriate.
+        //       So we trigger `init` here so that `urlSession` is initialized in a thread safe manner.
+        _ = urlSession
```
1ec5, line 42: Does the lazy syntax benefit us here? Maybe it’d be simpler to inline that initialization here.

1ec5 (approved):
Locking internal to a class is always scary, but I think you’ve covered your bases.

Tests/MapboxNavigationTests/ImageDownloaderTests.swift
S2Ler: Done

Sources/MapboxNavigation/ImageDownloader.swift
S2Ler: Good questions, thanks for asking. Without going deep into explaining the rationale why I've kept `lazy var` I think we can change `private lazy var urlSession: URLSession` to `private var urlSession: URLSession!`.

S2Ler:
> Locking internal to a class is always scary, but I think you’ve covered your bases.

Sorry, can you please rephrase?

PR #3074, opened 2021-06-11
Update refinement

Udumft:
Additional logic updates for underlying version bump #3039:
* Refined history recorder logic to always have a default path. 
* Added possibility to restore a fallback navigator to the version provided instead of always picking latest version for restoring.
* Added version info to restore from fallback notification.

### DIFF
--- a/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
+++ b/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
@@ -21,9 +21,13 @@ class Navigator {
     static var tilesURL: URL? = nil
     
     /**
-     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called.
+     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called. Defaults to user support directory.
      */
-    static var historyDirectoryURL: URL? = nil
+    static var historyDirectoryURL: URL = {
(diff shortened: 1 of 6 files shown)


Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -21,9 +21,13 @@ class Navigator {
     static var tilesURL: URL? = nil
     
     /**
-     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called.
+     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called. Defaults to user support directory.
```
1ec5, line 24:
```suggestion
     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called. Defaults to the Application Support directory.
```

1ec5 (approved):

zugaldia:
> Refined history recorder logic to always have a default path.

Just to make sure I got this right. This doesn't mean that history recording is enabled by default, right? (it should be disabled by default).

Udumft:
Actually, it is enabled by default now. Previously `historyRecorder` was not instantiated due to empty storage path.
Speaking of history: is it possible on Android to enable history recording in the middle of Active guidance or Free drive?

Udumft:
I've made a [PR](https://github.com/mapbox/mapbox-navigation-ios/pull/3081) to disable it again.

PR #3079, opened 2021-06-14
Add Xcode 12.5 swift test job

### LINKED ISSUE #3048: Add Xcode 12.5 to our CI Builds
CircleCI added support for Xcode 12.5. This Xcode version is known for breaking some project builds. We can make the following changes:
1. Move jobs that use Xcode 12.0 to run only on "release-v2.0" and "main" branches. (Motivation: I think Xcode 12.0 isn't popular, but at least we will spot the regression after merges). 
2. Add Xcode 12.5 build job as a replacement for Xcode 12.0 jobs. 

@mapbox/navigation-ios

S2Ler:
Fixes #3048

- Move Xcode 12.4 swift test to a new "extended-workflow". 
- Create Xcode 12.5 swift test.
- Move iOS 13.7 Xcode 12.5 swift test to the "extended-workflow".
- "extended-workflow" should run an extended set of tests once a day that will
allow us to spot regressions in more cases while saving the cost on CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -333,6 +333,11 @@ workflows:
           xcode: "12.4.0"
           iOS: "14.4"          
           device: "iPhone 12 Pro Max"
+      - spm-test-job:
+          name: "swift test; Xcode 12.5; iOS 14.5"
+          xcode: "12.5.0"
+          iOS: "14.5"          
+          device: "iPhone 12 Pro Max"          
       - ios-trigger-metrics:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -296,7 +296,27 @@ jobs:
           command: xcodebuild -scheme MapboxNavigation-Package test -destination "platform=iOS Simulator,OS=<< parameters.iOS >>,name=<< parameters.device >>" | xcpretty
 
 workflows:
-  workflow:
+  extended-workflow:
+    jobs:
+      - spm-test-job:
+          name: "swift test; Xcode 12.4; iOS 14.4"
+          xcode: "12.4.0"
+          iOS: "14.4"          
+          device: "iPhone 12 Pro Max"     
+      - spm-test-job:
+          name: "swift test; Xcode 12.4; iOS 13.7"
+          xcode: "12.4.0"
+          iOS: "13.7"          
+          device: "iPhone 11 Pro Max"
+    triggers:
+       - schedule:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -296,7 +296,27 @@ jobs:
...
+          iOS: "14.4"          
+          device: "iPhone 12 Pro Max"     
+      - spm-test-job:
+          name: "swift test; Xcode 12.4; iOS 13.7"
+          xcode: "12.4.0"
+          iOS: "13.7"          
+          device: "iPhone 11 Pro Max"
+    triggers:
+       - schedule:
+           cron: "0 0 * * *"
```
MaximAlien, line 313: Would probably make sense to add informational comment describing how often this job will be executed.

S2Ler: Can't agree more, looks cryptic

Package.swift
```diff
@@ -61,7 +61,10 @@ let package = Package(
             resources: [.copy("MBXInfo.plist")]),
         .target(
             name: "CTestHelper",
-            dependencies: ["MapboxMobileEvents"]),
+            dependencies: [
+                "MapboxMobileEvents",
+                "MapboxCoreNavigation",
```
1ec5, line 66: Interesting that the build was passing without this change, but I agree with adding it as a dependency.

1ec5 (approved):
Looks good. The cron comment would be nice too.

Package.swift
S2Ler: Yeah, no idea why wouldn't it build...

.circleci/config.yml
```diff
@@ -296,7 +296,27 @@ jobs:
...
+          iOS: "14.4"          
+          device: "iPhone 12 Pro Max"     
+      - spm-test-job:
+          name: "swift test; Xcode 12.4; iOS 13.7"
+          xcode: "12.4.0"
+          iOS: "13.7"          
+          device: "iPhone 11 Pro Max"
+    triggers:
+       - schedule:
+           cron: "0 0 * * *" # Once per day at 00:00
```
azarovalex, line 313: Is there any reason to run CI build every day? I don't think we will monitor these builds every day. IMHO CI builds for every PR should be enough.

S2Ler: Yes, the motivation is in the description. 

Visibility issue is a thing which can be solved with a slack bot integration.

PR #3081, opened 2021-06-14
Disabling History recording by default

Udumft:
### Description
[Previous PR](https://github.com/mapbox/mapbox-navigation-ios/pull/3074) accidentally enabled history recording by default. We should not do this.

### Implementation
Basically a reverse commit.

### DIFF
--- a/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
+++ b/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
@@ -21,13 +21,11 @@ class Navigator {
     static var tilesURL: URL? = nil
     
     /**
-     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called. Defaults to user support directory.
+     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called.
+     
+     Setting `nil` disables history recording. Defaults to `nil`.
      */
-    static var historyDirectoryURL: URL = {
-        let supportDir = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first!
-        
-        return supportDir.appendingPathComponent("com.mapbox.navigation.history")
-    }()
+    static var historyDirectoryURL: URL? = nil
     
     /**
      Store history to the directory stored in `Navigator.historyDirectoryURL` and asynchronously run a callback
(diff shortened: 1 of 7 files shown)


S2Ler (approved):

1ec5 (approved):
Looks good on a technical basis, but check with the MapboxNavigationNative team on the intended default behavior before merging.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -21,13 +21,11 @@ class Navigator {
...
+     Path to the directory where history file could be stored when `Navigator.writeHistory(completionHandler:)` is called.
+     
+     Setting `nil` disables history recording. Defaults to `nil`.
      */
-    static var historyDirectoryURL: URL = {
-        let supportDir = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first!
-        
-        return supportDir.appendingPathComponent("com.mapbox.navigation.history")
-    }()
+    static var historyDirectoryURL: URL? = nil
```
MaximAlien, line 28: Would it be possible to write a unit-test for this case? E.g. when `historyDirectoryURL` is `nil` `writeHistory(completionHandler:)` should not dump any data, right?

Udumft: Added tests

Tests/MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
```diff
@@ -97,6 +94,33 @@ class PassiveLocationDataSourceTests: XCTestCase {
...
+    
+    func testNoHistoryRecording() {
+        PassiveLocationDataSource.historyDirectoryURL = nil
+        
+        Navigator.shared.restartNavigator()
+        
+        let noHistoryExpectation = XCTestExpectation(description: "History should not be written on 'nil' path")
+        noHistoryExpectation.isInverted = true
+        PassiveLocationDataSource.writeHistory { _ in
+            noHistoryExpectation.fulfill()
```
MaximAlien, line 106: Not too critical, but I guess to make such test even more reliable we can check whether path to file is available in callback, but overall looks good to me.

Udumft:
Tests are suddenly failed due to unexpected dependency between them. Trying to investigate, but it seems that `RouteControllerTests` are not isolated.

Udumft:
I've created NN ticket to investigate weird test behavior.

Tests/MapboxCoreNavigationTests/PassiveLocationDataSourceTests.swift
Udumft: Current test is expected not to call the callback at all. But I've added additional check for writing history one.

PR #3091, opened 2021-06-16
Create NavigationLocationProvider for NavigationMapView

### LINKED ISSUE #3070: Create NavigationLocationProvider for NavigationMapView
Now the MapboxMaps has the `LocationProvider` protocol for the `MapView` to provide the location and location update.  `NavigationLocationManager` itself conforms to the `CLLocationManager` protocol and has several subclasses such as `SimulatedLocationManager` and . We should provide something similar to `AppleLocationProvider` for the location manager as a location provider during navigation.

ShanMa1991:
### Description
This pr is to fix #3070 , create the `NavigationLocationProvider` based on the `NavigationLocationManager`, to provide location provider for `NavigationMapView`.
- [x] renamed `PassiveLocationManager` to `PassiveLocationProvider`
- [x] renamed `PassiveLocationDataSource` to `PassiveLocationManager`


Sources/MapboxNavigation/NavigationLocationProvider.swift
```diff
@@ -0,0 +1,108 @@
...
+ 
+ If your application displays a `MapView` before starting turn-by-turn navigation, call `LocationManager.overrideLocationProvider(with:)` to override default location provider so that the map view always shows the location snapped to the road network. For example, use this class to show the user’s current location as they wander around town.
+ 
+ This class depends on `NavigationLocationManager` to detect the user’s location as it changes.
+ */
+open class NavigationLocationProvider: NSObject, LocationProvider, CLLocationManagerDelegate {
+    /**
+     The location provider's location manager, which detects the user’s location as it changes.
+     */
+    public var locationManager: NavigationLocationManager
```
1ec5, line 17: NavigationLocationProvider contains NavigationLocationManager, similar to the way PassiveLocationManager contains PassiveLocationDataSource. But it’s a bit confusing that NavigationLocationProvider and PassiveLocationManager are both LocationProviders, considering their names. Should we rename PassiveLocationManager to PassiveLocationProvider, and if so, should we rename PassiveLocationDataSource to PassiveLocationManager for consistency with the other NavigationLocationManager classes?

ShanMa1991: Personally,  because `PassiveLocationManager ` has been used for a long time and works pretty fine so far. I think maybe that the `PassiveLocationManager` should not be subclass of `NavigationLocationProvider `. Because `NavigationLocationProvider `  should be responsible to provide location update during turn-by-turn navigation for camera and puck instead of snapped location (provided by default `AppleLocationProvider`). They should at be the same level, one for free-drive, and one for turn-by-turn active guidance navigation.

1ec5: Sure, PassiveLocationManager doesn’t have to be a subclass of NavigationLocationProvider; the main benefit was the ability to remove some redundant implementations of methods like `requestWhenInUseAuthorization()`, but they aren’t very large implementations anyways. Essentially my feedback is about naming: PassiveLocationManager and NavigationLocationProvider could be siblings and it would still be confusing that only one is called a “provider” even though both are LocationProviders and neither is a CLLocationManager.

ShanMa1991: I see. Thanks for the suggestion! I'll rename the `PassiveLocationManager ` to `PassiveLocationProvider ` .

Sources/MapboxNavigation/PassiveLocationProvider.swift
```diff
@@ -10,7 +10,7 @@ import MapboxMaps
  
  This class depends on `PassiveLocationDataSource` to detect the user’s location as it changes. If you want location updates but do not need to display them on a map and do not want a dependency on the MapboxNavigation module, you can use `PassiveLocationDataSource` instead of this class.
```
1ec5, line 11: Now that PassiveLocationManager has been renamed PassiveLocationProvider, we can rename PassiveLocationDataSource to PassiveLocationManager for consistency with NavigationLocationManager. (Or, if you prefer, we could rename both classes to end with “LocationDataSource”.)

Sources/MapboxNavigation/NavigationLocationProvider.swift
```diff
@@ -0,0 +1,108 @@
...
+        set {
+            locationManager.headingOrientation = newValue
+        }
+    }
+
+    /**
+     Initializes the location provider with the `NavigationLocationManager`.
+     */
+    public override init() {
+        locationManager = NavigationLocationManager()
```
1ec5, line 61: If NavigationLocationProvider is intended to be the way to adapt a ReplayLocationManager or SimulatedLocationManager to a map view, then we’d need a convenience initializer that can set `locationManager` to a passed-in instance of NavigationLocationManager instead of creating a new one automatically.

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@
 * Improved performance and decreased memory usage when downloading routing tiles. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Renamed `PassiveLocationManager.startUpdatingLocation(completionHandler:)` to `PassiveLocationManager.startUpdatingLocation()`. This method now runs synchronously like `CLLocationManager.startUpdatingLocation()`. ([#2823](https://github.com/mapbox/mapbox-navigation-ios/pull/2823))
 * Added the `RouteController.historyDirectoryURL` property and `RouteController.writeHistory(completionHandler:)` method for recording details about a trip for debugging purposes. ([#2930](https://github.com/mapbox/mapbox-navigation-ios/pull/2930))
+* Added the `NavigationLocationProvider` class to conform to `LocationProvider` protocol, which depends on `NavigationLocationManager` to detect the user’s location as it changes during turn-by-turn navigation. ([#3070](https://github.com/mapbox/mapbox-navigation-ios/pull/3070))
```
1ec5, line 50: This might be an opportunity to point out how SimulatedLocationManager and ReplayLocationManager can now be used with a standalone NavigationMapView.

1ec5 (approved):
Some additional iteration would be welcome, but I think the remaining work is straightforward enough that another round of review won’t be necessary.

Sources/MapboxNavigation/PassiveLocationProvider.swift
ShanMa1991: Thanks for the suggestion, I'll rename the `PassiveLocationDataSource` to `PassiveLocationManager` for consistency

Sources/MapboxNavigation/NavigationLocationProvider.swift
ShanMa1991: Thanks for the suggestion, I'll add a `locationManager` as the parameter for the initialization.

PR #3106, opened 2021-06-22
Adjust the layer position to make puck layer at topmost

### LINKED ISSUE #3098: 3D puck layer not at the top most of navigationMapView
**Mapbox Navigation SDK version: v2.0.0-beta.13 to v2.0.0-beta.14

### Steps to reproduce
provide 3D configuration to the `navigationMapView.userLocationStyle`
 1. start navigation
 2. clear map
 3. start again

### Expected behavior
The 3D puck should be at the top most, same as 2D puck

### Actual behavior
It's covered by maneuvers as following:
puck3D

ShanMa1991:
### Description
This pr is to fix #3098 , because the 3D puck from map side doesn't support the default topmost layer position similar as 2D puck. This pr is to check the 3D puck layer and make it above the maneuvers when we call `func navigationMapView.addArrow`.

### Implementation
Because the 3D puck from map side doesn't support the default topmost layer position similar as 2D puck. This pr is to check the 3D puck layer and make it above the maneuvers when we call `func navigationMapView.addArrow`

### Screenshots or Gifs

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -884,6 +884,15 @@ open class NavigationMapView: UIView {
             let shaftLength = max(min(30 * mapView.metersPerPointAtLatitude(latitude: maneuverCoordinate.latitude), 30), 10)
             let shaftPolyline = route.polylineAroundManeuver(legIndex: legIndex, stepIndex: stepIndex, distance: shaftLength)
             
+            var puckLayerIdentifier: String?
+            switch userLocationStyle {
+            case .puck2D(configuration: _):
+                puckLayerIdentifier = NavigationMapView.LayerIdentifier.puck2DLayer
+            case .puck3D(configuration: _):
+                puckLayerIdentifier = NavigationMapView.LayerIdentifier.puck3DLayer
+            default: break
+            }
+            
             if shaftPolyline.coordinates.count > 1 {
                 let allLayerIds = mapView.mapboxMap.style.allLayerIdentifiers.map{ $0.id }
                 let mainRouteLayerIdentifier = route.identifier(.route(isMainRoute: true))
@@ -907,8 +916,8 @@ open class NavigationMapView: UIView {
                     try mapView.mapboxMap.style.addSource(arrowSource, id: NavigationMapView.SourceIdentifier.arrowSource)
                     arrowLayer.source = NavigationMapView.SourceIdentifier.arrowSource
                     
-                    if allLayerIds.contains(NavigationMapView.LayerIdentifier.puck3DLayer) {
-                        try mapView.mapboxMap.style.addLayer(arrowLayer, layerPosition: .below(NavigationMapView.LayerIdentifier.puck3DLayer))
+                    if let puckLayer = puckLayerIdentifier, allLayerIds.contains(puckLayer) {
+                        try mapView.mapboxMap.style.addLayer(arrowLayer, layerPosition: .below(puckLayer))
                     } else if mapView.mapboxMap.style.sourceExists(withId: NavigationMapView.LayerIdentifier.waypointCircleLayer) {
                         try mapView.mapboxMap.style.addLayer(arrowLayer, layerPosition: .below(NavigationMapView.LayerIdentifier.waypointCircleLayer))
                     } else {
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -904,7 +905,9 @@ open class NavigationMapView: UIView {
                     try mapView.mapboxMap.style.addSource(arrowSource, id: NavigationMapView.SourceIdentifier.arrowSource)
                     arrowLayer.source = NavigationMapView.SourceIdentifier.arrowSource
                     
-                    if mapView.mapboxMap.style.sourceExists(withId: NavigationMapView.LayerIdentifier.waypointCircleLayer) {
+                    if allLayerIds.contains(NavigationMapView.LayerIdentifier.puck3DLayer) {
```
MaximAlien, line 908: If layer already exists why're we adding it, instead of updating?

ShanMa1991: It is to check whether the 3D puck layer exists, if the 3D puck layer exists, we add the `arrowLayer` below the puck3D layer. It's the same case for `arrowSymbolLayer`, to add it below the 3D puck layer.

ShanMa1991:
The CarPlay has issue with adding maneuvers based on the following error, which leads to the abrupt stop of adding the `arrowStrokeLayer`, `arrowSymbolLayer` and `arrowSymbolCasingLayer`
```
Failed to perform operation while adding maneuver arrow with error: Cannot find layer with id: 0x000060000070eb00.main.route_line.
```

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: The CarPlay issue with maneuvers should be related with the layerID problem as [below](https://github.com/mapbox/mapbox-navigation-ios/pull/3106#issuecomment-867119146) , add the check of layer existence before set layer position. Similar issue happened for the way point circle layer.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -929,7 +932,11 @@ open class NavigationMapView: UIView {
                     
                     try mapView.mapboxMap.style.addSource(arrowStrokeSource, id: NavigationMapView.SourceIdentifier.arrowStrokeSource)
                     arrowStrokeLayer.source = NavigationMapView.SourceIdentifier.arrowStrokeSource
-                    try mapView.mapboxMap.style.addLayer(arrowStrokeLayer, layerPosition: .above(mainRouteLayerIdentifier))
+                    if allLayerIds.contains(mainRouteLayerIdentifier) {
```
ShanMa1991, line 935: The CarPlay's issue with maneuvers should be related with the layerID problem here. Sometimes the `mapView` could not find the `mainRouteLayerIdentifier`. Thus add the check of layer existence before setting layer position. Similar issue happened for the waypoint circle layer.

S2Ler:
Thanks for writing tests 👍

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -929,7 +932,11 @@ open class NavigationMapView: UIView {
...
+                        try mapView.mapboxMap.style.addLayer(arrowStrokeLayer, layerPosition: .above(mainRouteLayerIdentifier))
+                    } else {
+                        try mapView.mapboxMap.style.addLayer(arrowStrokeLayer, layerPosition: .below(NavigationMapView.LayerIdentifier.arrowLayer))
```
1ec5, line 938: Nit: factor this line out of the conditional so that only the layer position is in the conditional.

Tests/MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -235,6 +235,32 @@ class NavigationViewControllerTests: XCTestCase {
         XCTAssert(newAnnotations.contains { $0.coordinate.distance(to: secondDestination) < 1 }, "New destination annotation does not exist on map")
     }
     
+    func testpuck3DLayerPosition() {
```
1ec5, line 238: Nit:

```suggestion
    func testPuck3DLayerPosition() {
```

1ec5 (approved):

PR #3109, opened 2021-06-22
Fix final annotation modification.

### LINKED ISSUE #3108: Unable to modify `PointAnnotation`, when using `NavigationMapViewDelegate.navigationMapView(_:didAdd:)` delegate method.
**Mapbox Navigation SDK version:** `release-v2.0`

After introducing `PointAnnotationManager` there is no way to change image on final destination `PointAnnotation`.

Example application in https://github.com/mapbox/mapbox-navigation-ios-examples/blob/main/Navigation-Examples/Examples/Custom-Destination-Marker.swift also doesn't behave correctly because of this issue.

One of the solutions would be to update delegate method to something like this: `NavigationMapViewDelegate.navigationMapView(_:didAdd:pointAnnotationManager:)` and expose `PointAnnotationManager` instance publicly, as it's internal now.

MaximAlien:
### Description

Fixes #3108.


S2Ler:
Do we need an entry in the changelog for public `public var pointAnnotationManager: PointAnnotationManager?`?

MaximAlien:
> Do we need an entry in the changelog for public `public var pointAnnotationManager: PointAnnotationManager?`?

Yep, good idea. Updated changelog.

S2Ler (approved):

PR #3110, opened 2021-06-22
Fix `FeedbackViewController` presentation on iPad.

### LINKED ISSUE #3094: `FeedbackViewController` is incorrectly shown on iPad.
**Mapbox Navigation SDK version:** `v2.0.0-beta.13`

### Steps to reproduce

 1. Open https://github.com/mapbox/mapbox-navigation-ios-examples/blob/main/Navigation-Examples/Examples/Advanced.swift.
2. Start active guidance navigation.
3. Observe behavior.

### Expected behavior

`FeebackViewController` should be shown right at the bottom of NavigationViewController.

### Actual behavior

`FeebackViewController` is misplaced. For example:

Screen Shot 2021-06-17 at 10 56 50 AM

MaximAlien:
### Description

Fixes #3094.

Screen Shot 2021-06-22 at 5 15 06 PM

### DIFF
--- a/Sources/MapboxNavigation/Transitioning.swift
+++ b/Sources/MapboxNavigation/Transitioning.swift
@@ -1,29 +1,30 @@
 import UIKit
 
-class Interactor: UIPercentDrivenInteractiveTransition {
-    var hasStarted = false
-    var shouldFinish = false
-}
-
 class DismissAnimator: NSObject { }
+
 extension DismissAnimator: UIViewControllerAnimatedTransitioning {
+    
     func transitionDuration(using transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval {
         return 0.5
     }
     
     func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
-        guard let fromVC = transitionContext.viewController(forKey: .from) else { return }
-        guard let toVC = transitionContext.viewController(forKey: .to) else { return }
-        let containerView = transitionContext.containerView
+        guard let fromViewController = transitionContext.viewController(forKey: .from),
+              let toViewController = transitionContext.viewController(forKey: .to) else { return }
         
-        let point = CGPoint(x: 0, y: UIScreen.main.bounds.size.height)
-        let height = fromVC.view.bounds.height-toVC.view.frame.minY
-        let finalFrame = CGRect(origin: point, size: CGSize(width: fromVC.view.bounds.width, height: height))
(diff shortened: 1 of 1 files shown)


S2Ler:
It is time consuming to split the PR into stylistic changes and the bug fix. Would be good to separate the changes into two commits.

MaximAlien:
> It is time consuming to split the PR into stylistic changes and the bug fix. Would be good to separate the changes into two commits.

Divided PR into two commits, please let me know if it's better to review like this. Even though this one was straightforward (and project was compilable for both commits) in many cases project might become broken when building from intermediate commits.

S2Ler:
> > It is time consuming to split the PR into stylistic changes and the bug fix. Would be good to separate the changes into two commits.
> 
> Divided PR into two commits, please let me know if it's better to review like this. Even though this one was straightforward (and project was compilable for both commits) in many cases project might become broken when building from intermediate commits.

Yeah, it is not always easy to separate the changes. Thanks for doing this!

Sources/MapboxNavigation/Transitioning.swift
```diff
@@ -59,13 +59,29 @@ extension PresentAnimator: UIViewControllerAnimatedTransitioning {
...
-        let finalFrame = CGRect(origin: CGPoint(x: 0, y: fromVC.view.bounds.height - height),
-                                size: CGSize(width: fromVC.view.bounds.width, height: height))
+        // In case of certain presentation styles on iPad `FeedbackViewController` is placed incorrectly.
+        // This check adjusts its size in such case.
+        var y = fromVC.view.bounds.height - height
+        var width = fromVC.view.bounds.width
+        let modalPresentationStyles: [UIModalPresentationStyle] = [.pageSheet, .formSheet]
+        if UIDevice.current.userInterfaceIdiom == .pad,
+           modalPresentationStyles.contains(fromVC.modalPresentationStyle) {
+            y = UIScreen.main.bounds.size.height - height
```
S2Ler, line 76: Just curious, can't we use the same calculation on every platform? Will it break some use case which isn't obvious just by looking at this code?

S2Ler (approved):

Sources/MapboxNavigation/Transitioning.swift
MaximAlien: Yep, made some changes to make this fix more generic. Also fixed some minor animation transitioning issue.

PR #3112, opened 2021-06-23
Fix more lane icon edge cases and tests

### LINKED ISSUE #3107: Reenable LaneViewTests
LaneViewTests was excluded from the MapboxNavigationTests target that #3043 revived via SPM, due to numerous assertion failures. At a glance, it appears to be a mix of copypasta from #2882 (which was merged while MapboxNavigationTests was still kaput) and some potential bugs in the SDK itself.

/cc @mapbox/navigation-ios

1ec5:
This PR fixes several edge cases in lane indication ranking that I missed in #2882 due to disabled unit tests. It also fixes numerous errors in the tests themselves. In some cases, the library or tests weren’t wrong, but the additional assets in #2882 make a better result possible. LaneViewTests has been reenabled, now that it passes.

Fixes #3107.

### DIFF
--- a/Sources/MapboxNavigation/LaneView.swift
+++ b/Sources/MapboxNavigation/LaneView.swift
@@ -28,6 +28,10 @@ extension LaneIndication {
             primaryIndication = .slightLeft
         } else if indications.contains(.slightRight) && maneuverDirection ?? .slightRight == .slightRight {
             primaryIndication = .slightRight
+        } else if indications.isSubset(of: [.left, .straightAhead]) && maneuverDirection ?? .slightLeft == .slightLeft {
+            primaryIndication = .left
+        } else if indications.isSubset(of: [.right, .straightAhead]) && maneuverDirection ?? .slightRight == .slightRight {
+            primaryIndication = .right
         } else if indications.contains(.left) && maneuverDirection ?? .left == .left {
             primaryIndication = .left
         } else if indications.contains(.right) && maneuverDirection ?? .right == .right {
@@ -36,6 +40,14 @@ extension LaneIndication {
             primaryIndication = .sharpLeft
         } else if indications.contains(.sharpRight) && maneuverDirection ?? .sharpRight == .sharpRight {
             primaryIndication = .sharpRight
+        } else if indications.contains(.left) && !indications.contains(.sharpLeft) && !indications.contains(.uTurn) && maneuverDirection ?? .sharpLeft == .sharpLeft {
+            primaryIndication = .left
+        } else if indications.contains(.right) && !indications.contains(.sharpRight) && !indications.contains(.uTurn) && maneuverDirection ?? .sharpRight == .sharpRight {
+            primaryIndication = .right
+        } else if !indications.isDisjoint(with: .lefts) && !indications.isDisjoint(with: .rights) && maneuverDirection?.isLeft ?? false {
+            primaryIndication = .left
+        } else if !indications.isDisjoint(with: .lefts) && !indications.isDisjoint(with: .rights) && maneuverDirection?.isRight ?? false {
+            primaryIndication = .right
         } else if indications.contains(.uTurn) && maneuverDirection ?? .uTurn == .uTurn {
             primaryIndication = .uTurn
         } else {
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/LaneView.swift
```diff
@@ -71,10 +83,10 @@ extension LaneIndication {
             secondaryIndication = .straightAhead
         } else if !primaryIndication.isDisjoint(with: .rights) && indications.contains(.slightLeft) {
             // No assets for slight or sharp opposite turns.
-            secondaryIndication = .slightLeft
+            secondaryIndication = .left
```
1ec5, line 86: We still don’t have assets for slight or sharp opposite turns, but at least we have assets for opposite turns. A user is more likely to distinguish between a left lane and a left-right lane than between a left lane and a slight left lane, even if the maneuver requires going slight left.

Sources/MapboxNavigation/LaneView.swift
```diff
@@ -36,6 +40,14 @@ extension LaneIndication {
...
         } else if indications.contains(.sharpRight) && maneuverDirection ?? .sharpRight == .sharpRight {
             primaryIndication = .sharpRight
+        } else if indications.contains(.left) && !indications.contains(.sharpLeft) && !indications.contains(.uTurn) && maneuverDirection ?? .sharpLeft == .sharpLeft {
+            primaryIndication = .left
+        } else if indications.contains(.right) && !indications.contains(.sharpRight) && !indications.contains(.uTurn) && maneuverDirection ?? .sharpRight == .sharpRight {
+            primaryIndication = .right
+        } else if !indications.isDisjoint(with: .lefts) && !indications.isDisjoint(with: .rights) && maneuverDirection?.isLeft ?? false {
+            primaryIndication = .left
+        } else if !indications.isDisjoint(with: .lefts) && !indications.isDisjoint(with: .rights) && maneuverDirection?.isRight ?? false {
+            primaryIndication = .right
```
1ec5, line 50: These cases handle lanes with opposite turns, such as left-right. Previously, we’d fall into the big else case below that tries to choose the “least extreme” indication. It’d be a tie between left and right, so that logic would choose left because it happens to come first. This more generic logic accounts for slight and sharp turns too.

Sources/MapboxNavigation/LaneView.swift
```diff
@@ -36,6 +40,14 @@ extension LaneIndication {
             primaryIndication = .sharpLeft
         } else if indications.contains(.sharpRight) && maneuverDirection ?? .sharpRight == .sharpRight {
             primaryIndication = .sharpRight
+        } else if indications.contains(.left) && !indications.contains(.sharpLeft) && !indications.contains(.uTurn) && maneuverDirection ?? .sharpLeft == .sharpLeft {
+            primaryIndication = .left
+        } else if indications.contains(.right) && !indications.contains(.sharpRight) && !indications.contains(.uTurn) && maneuverDirection ?? .sharpRight == .sharpRight {
+            primaryIndication = .right
```
1ec5, line 46: This is tricky. A distinction between sharp left, left, and left doesn’t matter much in general – a sharp left maneuver should bias the ranking towards whatever is on the left. However, a user is likely to distinguish between a left and a U-turn to the left.

Tests/MapboxNavigationTests/LaneViewTests.swift
```diff
@@ -116,71 +116,75 @@ class LaneViewTests: XCTestCase {
...
     }
     
     func testLaneConfiguration() {
         XCTAssertEqual(LaneConfiguration(rankedIndications: .init(primary: .uTurn, secondary: nil, tertiary: nil), drivingSide: .right),
                        .uTurn(side: .left))
         XCTAssertEqual(LaneConfiguration(rankedIndications: .init(primary: .uTurn, secondary: nil, tertiary: nil), drivingSide: .left),
                        .uTurn(side: .right))
-        XCTAssertNil(LaneConfiguration(rankedIndications: .init(primary: .uTurn, secondary: .left, tertiary: nil), drivingSide: .right))
+        XCTAssertEqual(LaneConfiguration(rankedIndications: .init(primary: .uTurn, secondary: .left, tertiary: nil), drivingSide: .right),
+                       .turnOrUTurn(side: .left, turn: false, uTurn: true))
```
1ec5, line 168: These cases simply weren’t updated to reflect the assets added in #2882.

1ec5:
As an aside, I wonder if we could streamline a lot of this logic by making the maneuver directions comparable. Then we could map a lane’s indications into a range of maneuver directions and compare against that range instead of individual indications.

S2Ler (approved):
It is hard for me to verify the correctness of this logic, but overall it looks good. Inline comments helped as well.

PR #3118, opened 2021-06-25
Restore some tests

S2Ler:
Restores:
- `CarPlayManagerTests.swift`
- `NavigationViewControllerTests.swift`
- `LeaksSpec.swift`

Some tests in `NavigationViewControllerTests.swift` had to be disabled due to be in not working condition.

### DIFF
--- a/Sources/MapboxCoreNavigation/NativeHandlersFactory.swift
+++ b/Sources/MapboxCoreNavigation/NativeHandlersFactory.swift
@@ -30,7 +30,9 @@ class NativeHandlersFactory {
     // MARK: - Native Handlers
     
     lazy var historyRecorder: HistoryRecorderHandle? = {
-        HistoryRecorderHandle.build(forHistoryDir: historyDirectoryURL?.path ?? "", config: configHandle)
+        historyDirectoryURL.flatMap {
+            HistoryRecorderHandle.build(forHistoryDir: $0.path, config: configHandle)
+        }
     }()
     
     lazy var navigator: MapboxNavigationNative.Navigator = {
--- a/Tests/MapboxNavigationTests/CarPlayManagerTests.swift
+++ b/Tests/MapboxNavigationTests/CarPlayManagerTests.swift
@@ -21,9 +21,10 @@ class CarPlayManagerTests: XCTestCase {
 
     override func setUp() {
         super.setUp()
+        DirectionsCredentials.injectSharedToken(.mockedAccessToken)
         ResourceOptionsManager.default.resourceOptions.accessToken = .mockedAccessToken
         eventsManagerSpy = NavigationEventsManagerSpy()
-        manager = CarPlayManager(directions: .mocked, eventsManager: eventsManagerSpy, navigationViewControllerClass: CarPlayNavigationViewControllerTestable.self)
+        manager = CarPlayManager(directions: .mocked, eventsManager: eventsManagerSpy, carPlayNavigationViewControllerClass: CarPlayNavigationViewControllerTestable.self)
         searchController = CarPlaySearchController()
     }
 
@@ -42,9 +43,6 @@ class CarPlayManagerTests: XCTestCase {
     }
 
     func testEventsEnqueuedAndFlushedWhenCarPlayConnected() {
-        // NOTE: Xcode is going to complain here - ignore. This is a known XCTest bug.
-        guard #available(iOS 12, *) else { return }
-
         simulateCarPlayConnection(manager!)
 
         let expectedEventName = MMEventTypeNavigationCarplayConnect
@@ -55,9 +53,6 @@ class CarPlayManagerTests: XCTestCase {
     }
 
     func testEventsEnqueuedAndFlushedWhenCarPlayDisconnected() {
-        // NOTE: Xcode is going to complain here - ignore. This is a known XCTest bug.
-        guard #available(iOS 12, *) else { return }
-
         simulateCarPlayDisconnection()
 
         let expectedEventName = MMEventTypeNavigationCarplayDisconnect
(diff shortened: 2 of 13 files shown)


Sources/MapboxCoreNavigation/NativeHandlersFactory.swift
```diff
@@ -30,7 +30,9 @@ class NativeHandlersFactory {
     // MARK: - Native Handlers
     
     lazy var historyRecorder: HistoryRecorderHandle? = {
-        HistoryRecorderHandle.build(forHistoryDir: historyDirectoryURL?.path ?? "", config: configHandle)
+        historyDirectoryURL.flatMap {
```
S2Ler, line 33: I found an error in the logs from NavNative. This fixes the issue.

Sources/TestHelper/CoreLocation.swift
```diff
@@ -134,14 +134,29 @@ extension Array where Element == CLLocation {
...
-                                             course: $0.element.course,
-                                             speed: $0.element.speed,
-                                             timestamp: timestamp + $0.offset) }
+    public func shifted(to firstTimestamp: Date) -> [CLLocation] {
+        var previousLocation: CLLocation?
+        var shiftedLocations: [CLLocation] = []
+        for location in self {
+            let timestamp: Date
+            if let previousLocation = previousLocation {
+                timestamp = previousLocation.timestamp + previousLocation.coordinate.distance(to: location.coordinate) / previousLocation.speed
```
1ec5, line 143: If I’m reading this correctly, we were previously adding 1&nbsp;second to each location’s timestamp, assuming the frequency that Core Location normally runs on. This instead changes the offset to be based on interpolation. This change could be problematic, because it assumes that `CLLocation.speed` can help predict the next location’s timestamp, only to be used for informational purposes (like displaying your current speed next to a speed limit). But [`CLLocation.speed`](https://developer.apple.com/documentation/corelocation/cllocation/1423798-speed) is only the instantaneous speed reading from GPS. I don’t see how it would be any more reliable than the 1&nbsp;second assumption, unless we happened to hard-code timestamps and speeds in the fixtures that make that assumption true.

S2Ler: Thanks for the feedback. I've got in a loop trying different approaches in the hope of fixing failing tests and went too far. So I will revisit this code.

Sources/MapboxCoreNavigation/NativeHandlersFactory.swift
S2Ler: oh, I got used to this syntax and didn't think twice before using it. If you think that it is better to use simpler, more obvious syntax, let me know, I will update this code.

Sources/TestHelper/CoreLocation.swift
S2Ler: I removed these changes for now.

S2Ler:
@mapbox/navigation-ios @1ec5 @MaximAlien please review.

Tests/MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -42,9 +43,6 @@ class CarPlayManagerTests: XCTestCase {
     }
 
     func testEventsEnqueuedAndFlushedWhenCarPlayConnected() {
-        // NOTE: Xcode is going to complain here - ignore. This is a known XCTest bug.
-        guard #available(iOS 12, *) else { return }
```
1ec5, line 46: Awesome that we can remove this workaround now. 👍

Tests/MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -273,21 +331,14 @@ class NavigationViewControllerTests: XCTestCase {
         let route = Fixture.route(from: "DCA-Arboretum", options: options)
         let navigationViewController = NavigationViewController(for: route, routeIndex: 0, routeOptions: options)
         
-        viewController.present(navigationViewController, animated: false, completion: nil)
+        _ = navigationViewController.view
```
1ec5, line 334: This is much better than the previous approach of presenting the whole view controller asynchronously.

1ec5 (approved):
Great work!

Tests/MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -204,7 +181,7 @@ class CarPlayManagerTests: XCTestCase {
             CLLocationCoordinate2D(latitude: 34.054081, longitude: -118.243412),
         ])
         choice.userInfo = (Fixture.route(from: "route-with-banner-instructions", options: options), 0, options)
-
+        CarPlayMapViewController.swizzleMethods()
```
MaximAlien, line 184: I wonder if we should un-swizzle at the end of execution to prevent any other side-effects (for `UNUserNotificationCenter` as well)?

Tests/MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -1,64 +1,106 @@
...
 @testable import TestHelper
 @testable import MapboxCoreNavigation
 @testable import MapboxNavigation
 
 let otherResponse = Fixture.JSONFromFileNamed(name: "route-for-lane-testing")
 
+private let mockedUNUserNotificationCenter: MockedUNUserNotificationCenter = .init()
+
+/// `UNUserNotificationCenter.current()` crashes when run from SPM tests.
+/// In order to fix the crash we mock `UNUserNotificationCenter` but swizzling `UNUserNotificationCenter.current()` and
```
MaximAlien, line 14:
```suggestion
/// In order to fix the crash we mock `UNUserNotificationCenter` by swizzling `UNUserNotificationCenter.current()` and
```

MaximAlien:
Overall looks good, thanks. Left some minor comments related to typos and swizzling.

Tests/MapboxNavigationTests/CarPlayManagerTests.swift
S2Ler: Good idea, thank you!

Tests/MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -1,64 +1,106 @@
...
+            _ = navigationViewController.view // trigger view load
+            let navigationService = navigationViewController.navigationService!
+            let router = navigationService.router!
+            let firstCoord      = router.routeProgress.nearbyShape.coordinates.first!
+            let firstLocation   = location(at: firstCoord)
+
+            var poi = [CLLocation]()
+            let taylorStreetIntersection = router.route.legs.first!.steps.first!.intersections!.first!
+            let turkStreetIntersection   = router.route.legs.first!.steps[3].intersections!.first!
+            let fultonStreetIntersection = router.route.legs.first!.steps[5].intersections!.first!
```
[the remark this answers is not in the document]
S2Ler: I don't like force unwraps as well. 
Even though I haven't written this code and just moved it, I've updated this method with preconditions. 

Would be good at some point to go over all the tests and replace force unwraps with XCFail or something.

MaximAlien (approved):

PR #3122, opened 2021-06-28
Allow feedbacks during free drive

azarovalex:
### Description

Fixes https://github.com/mapbox/navigation-sdks/issues/991.

### Implementation

`NavigationEventsManager` was strongly coupled with `EventsManagerDataSource` that requires `Router` and `RouteProgess` which we don't have during FreeDrive. I decided not to change `EventsManagerDataSource`, but instead create a new protocol `EventsManagerFreeDriveDataSource` which will be used during the FreeDrive. I also overrode the logic that creates feedback events in the class `FreeDriveNavigationEventsManager`.

The `NavigationEventDetails` was a structure that contained all telemetry fields and knew too much about active guidance, so I separated it into two different structs `ActiveGuidanceEventDetails` and `FreeDriveEventDetails` with a protocol `NavigationEventDetails` that contains all common fields.


Sources/MapboxCoreNavigation/FreeDriveEventDetails.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+import CoreLocation
+
+struct FreeDriveEventDetails: NavigationEventDetails {
```
1ec5, line 4: We avoid using the term “free drive” in API symbol names, in part because it’s too easy for the term to make something sound like an action method, but also because this jargon is unintuitive to developers who aren’t in the automotive industry. Instead, we always use the term “passive”, as in “passive navigation”. (Even the Android documentation has to quickly turn around and [define free-driving as passive navigation](https://docs.mapbox.com/android/navigation/guides/free-drive/).)

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -4,6 +4,19 @@ import MapboxDirections
 
 let NavigationEventTypeRouteRetrieval = "mobile.performance_trace"
 
+/**
+ The `EventsManagerFreeDriveDataSource` protocol declares values required for recording free drive events.
+ */
+public protocol EventsManagerFreeDriveDataSource: AnyObject {
```
1ec5, line 10: Consider a more generic name for this protocol, like “RawLocatable” or “RawLocationDataSource”.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -14,6 +27,32 @@ public protocol EventsManagerDataSource: AnyObject {
     var locationProvider: NavigationLocationManager.Type { get }
 }
 
+open class FreeDriveNavigationEventsManager: NavigationEventsManager {
```
1ec5, line 30: Would it be possible to genericize NavigationEventsManager to avoid the need for subclassing? I think it could become problematic that NavigationEventsManager is being treated almost as an abstract base class, due to these overrides, even though it normally isn’t abstract. It’s reminiscent of how NavigationMapView used to subclass MGLMapView but had to work around increasingly complex logic specific to non-navigation maps.

If separate types are necessary for active navigation versus passive navigation, then I’d suggest factoring out whatever is in common between both activities into a new base class or protocol with default implementations.

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -63,13 +76,26 @@ struct NavigationEventDetails: EventDetails {
...
+        guard let stringForShortVersion = Bundle.string(forMapboxCoreNavigationInfoDictionaryKey: "CFBundleShortVersionString") else {
+            preconditionFailure("CFBundleShortVersionString must be set in the Info.plist.")
+        }
+        return stringForShortVersion
+    }
+    var screenBrightness: Int { Int(UIScreen.main.brightness * 100) }
+    var volumeLevel: Int { Int(AVAudioSession.sharedInstance().outputVolume * 100) }
+}
+
+struct ActiveGuidanceEventDetails: NavigationEventDetails {
```
1ec5, line 97: This is the first time we’re making an explicit distinction between navigation and “active navigation”, but it does make some sense considering “passive navigation”. (So far, navigation would be assumed to be active unless otherwise specified.) I think “ActiveNavigation” would be preferable, because “guidance” is a term we normally restrict to guidance instructions across the Mapbox navigation stack, despite broader usage elsewhere in the industry.

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -52,9 +52,22 @@ struct PerformanceEventDetails: EventDetails {
     }
 }
 
-struct NavigationEventDetails: EventDetails {
-    let audioType: String = AVAudioSession.sharedInstance().audioType
-    let applicationState: UIApplication.State = {
+protocol NavigationEventDetails: EventDetails {
```
1ec5, line 55: A protocol may be the best solution, but in case it wasn’t considered, another approach would be composition: keep this struct intact but add a property that could be either active-specific or passive-specific.

1ec5 (commented):
Some architectural and naming notes.

Sources/MapboxCoreNavigation/FreeDriveEventDetails.swift
azarovalex: Thanks, renamed all `FreeDrive` classes to the `PassiveNavigation`.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
azarovalex: This protocol contains values that we want to report in passive navigation mode, it can be updated in the future to contain other fields as well. However, I renamed the `FreeDrive` part to the `PassiveNavigation`.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
azarovalex: I agree, removed the inheritance here. NavigationEventsManager now has two optional data sources – one for passive navigation, one for active. The idea of creating a data source protocol doesn't work in this case, the reported values are absolutely different.

Sources/MapboxCoreNavigation/EventDetails.swift
azarovalex: Thx, renamed the classes according to your comment.

Sources/MapboxCoreNavigation/EventDetails.swift
azarovalex: I think the protocol is fine here, I didn't find another approach that seems okish for me.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -202,27 +182,31 @@ open class NavigationEventsManager {
...
+            event = ActiveNavigationEventDetails(dataSource: activeNavigationDataSource,
+                                                 session: sessionState, defaultInterface: usesDefaultUserInterface)
+        } else if let passiveNavigationDataSource = passiveNavigationDataSource {
+            event = PassiveNavigationEventDetails(dataSource: passiveNavigationDataSource)
+        } else {
+            assertionFailure("NavigationEventsManager is unable to create feedbacks without a datasource.")
+        }
         
-        event.screenshot = captureScreen(scaledToFit: 250)?.base64EncodedString()
+        event?.feedbackType = type.description
```
1ec5, line 199: We should try to keep this property strongly typed as an enumeration until the last possible moment before serialization as JSON. The code has been pulling out the description like this, but this is an opportunity to introduce a bit more safety using FeedbackType.

Sources/MapboxCoreNavigation/PassiveNavigationEventDetails.swift
```diff
@@ -0,0 +1,61 @@
+import Foundation
+import CoreLocation
+
+struct PassiveNavigationEventDetails: NavigationEventDetails {
+    let coordinate: CLLocationCoordinate2D?
+    let created = Date()
+    let sessionIdentifier = "session-id" // TODO: create global session
```
1ec5, line 7: Should the session identifier be an extensible enumeration?

Sources/MapboxCoreNavigation/PassiveNavigationEventDetails.swift
```diff
@@ -0,0 +1,61 @@
...
+    let driverMode = "freeDrive"
+    
+    var event: String?
+    var userId: String?
```
1ec5, line 11: Consider sticking to Swift naming conventions, relying on the CodingKeys property below to associate a `userId` property with `userIdentifier` in JSON.

Sources/MapboxCoreNavigation/PassiveNavigationEventDetails.swift
```diff
@@ -0,0 +1,61 @@
...
+    var feedbackType: String?
+    var description: String?
+    var screenshot: String?
+    // TODO: add time in foreground and background
```
1ec5, line 15: Perhaps we can implement this using simply start and end `Date`s, avoiding timers, even if all that gets sent is an elapsed time.

Sources/MapboxCoreNavigation/PassiveNavigationEventDetails.swift
azarovalex: For `sessionIdentifier` field I reused the `SessionState` class that we already use for active guidance, it represents the session identifier with UUID type.

Sources/MapboxCoreNavigation/PassiveNavigationEventDetails.swift
azarovalex: I reused the logic we had in `SessionState` class, it doesn't use any timers, just start `Date` and `NotificationCenter` to keep track of iPhone state changes.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -62,8 +76,12 @@ open class NavigationEventsManager {
         return token
     }()
     
-    public required init(dataSource source: EventsManagerDataSource?, accessToken possibleToken: String? = nil, mobileEventsManager: MMEEventsManager = .shared()) {
-        dataSource = source
+    public required init(activeNavigationDataSource: EventsManagerDataSource? = nil,
+                         passiveNavigationDataSource: PassiveNavigationEventsManagerDataSource? = nil,
+                         accessToken possibleToken: String? = nil,
+                         mobileEventsManager: MMEEventsManager = .shared()) {
+        dataSource = activeNavigationDataSource
```
bamx23, line 83: Maybe it's better to rename the old `dataSource` to `activeNavigationDataSource`?

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -158,32 +174,38 @@ open class NavigationEventsManager {
...
-        event.userId = UIDevice.current.identifierForVendor?.uuidString
-        event.feedbackType = type.description
-        event.description = description
+        if let activeNavigationDataSource = dataSource {
+            event = ActiveNavigationEventDetails(dataSource: activeNavigationDataSource,
+                                                 session: sessionState, defaultInterface: usesDefaultUserInterface)
+        } else if let passiveNavigationDataSource = passiveNavigationDataSource {
+            event = PassiveNavigationEventDetails(dataSource: passiveNavigationDataSource, sessionState: sessionState)
+        } else {
+            assertionFailure("NavigationEventsManager is unable to create feedbacks without a datasource.")
```
bamx23, line 191: NIT: You can make `event` nonnull and return nil here.

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -52,21 +53,76 @@ struct PerformanceEventDetails: EventDetails {
...
+            totalTimeInBackground += abs(session.lastTimeInBackground.timeIntervalSinceNow)
         }
-    }()
-    let batteryLevel: Int = UIDevice.current.batteryLevel >= 0 ? Int(UIDevice.current.batteryLevel * 100) : -1
-    let batteryPluggedIn: Bool = [.charging, .full].contains(UIDevice.current.batteryState)
+        percentTimeInForeground = totalTimeInPortrait + totalTimeInLandscape == 0 ? 100 : Int((totalTimeInPortrait / (totalTimeInPortrait + totalTimeInLandscape) * 100))
+    }
+}
+
+struct ActiveNavigationEventDetails: NavigationEventDetails {
```
bamx23, line 124: Maybe we should move `ActiveNavigationEventDetails` to a separate file?

Sources/MapboxCoreNavigation/PassiveNavigationEventDetails.swift
```diff
@@ -0,0 +1,69 @@
...
+    var totalTimeInBackground: TimeInterval = 0
+    
+    init(dataSource: PassiveNavigationEventsManagerDataSource, sessionState: SessionState) {
+        coordinate = dataSource.rawLocation?.coordinate
+        sessionIdentifier = sessionState.identifier.uuidString
+        startTimestamp = sessionState.departureTimestamp
+        updateApplicationStatePercentages(sessionState)
+    }
+    
+    private enum CodingKeys: String, CodingKey {
```
bamx23, line 29: Is it possible to make this enum common for both structures? Or at least make sure that common cases have the same values?

Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -4,17 +4,18 @@ import Polyline
...
+    var event: String? { get set }
     var created: Date { get }
     var sessionIdentifier: String { get }
 }
 
 struct PerformanceEventDetails: EventDetails {
-    let event: String?
     let created: Date
     let sessionIdentifier: String
+    var event: String?
```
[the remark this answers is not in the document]
azarovalex: This is because of the changes in `EventDetails`, this property is now settable too. `NavigationEventsManager` now works with a protocol instead of a concrete type.

Example/ViewController.swift
```diff
@@ -108,6 +118,13 @@ class ViewController: UIViewController {
         self.speedLimitView = speedLimitView
     }
     
+    func setupFeedbackButton() {
+        view.addSubview(feedbackButton)
+        feedbackButton.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor, constant: 12).isActive = true
+        feedbackButton.rightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.rightAnchor, constant: -12).isActive = true
```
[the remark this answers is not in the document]
azarovalex: I guess it doesn't make a lot of difference in this case, because it's the only button on the map and doesn't even contain text. But you are right, for consistency with `MapOrnamentPosition`, I changed the anchor to the trailing.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
azarovalex: Thanks I will do it since it's not a public property

Sources/MapboxCoreNavigation/PassiveNavigationEventDetails.swift
azarovalex: Unfortunately I don't see a way to do this. We can use only one enum for one Codable conformance and this cannot be the same enum since different events have different values. The raw value of enum cases can only be a string literal, this is a compiler restriction, so there is no way to have constants separately in one place.

CHANGELOG.md
```diff
@@ -123,6 +123,7 @@
 * Fixed a retain cycle in `UserCourseView`. ([#3120](https://github.com/mapbox/mapbox-navigation-ios/issues/3120))
 * The `EventsManagerDataSource.router`, `NavigationService.router`, `NavigationService.eventsManager`, `MapboxNavigationService.router`, `MapboxNavigationService.eventsManager` properties are no longer force unwrapped. ([#3055](https://github.com/mapbox/mapbox-navigation-ios/pull/3055))
 * Fixed an issue where traffic congestion segments along the route line blurred into each other when the map was zoomed in far enough. ([#3153](https://github.com/mapbox/mapbox-navigation-ios/pull/3153))
+* Supported feedbacks during passive navigation. Set `PassiveLocationManager` as a data source for `NavigationEventsManager` and then use this class to create and send user feedbacks. `FeedbackViewController` provides drop-in feedback screen and can be used during both passive and active navigation. ([#3122](https://github.com/mapbox/mapbox-navigation-ios/pull/3122))
```
1ec5, line 126: /ref https://github.com/mapbox/mapbox-navigation-ios/pull/3154#discussion_r672684728

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -4,6 +4,19 @@ import MapboxDirections
 
 let NavigationEventTypeRouteRetrieval = "mobile.performance_trace"
 
+/**
+ The `PassiveNavigationEventsManagerDataSource` protocol declares values required for recording passive location events.
```
1ec5, line 8: Nit: In general, it’s better to start a top-level type’s documentation by saying what it is, before saying what it does. For example:

> A data source that declares values required for recording passive location events.

In the Objective-C days, it was common to start a type’s documentation by saying the type’s name, but that’s no longer necessary. It doesn’t make a huge difference in this case, other than removing some redundancy. But in general it’s a good exercise, ensuring we think hard about what a type represents so we don’t wind up with abstract FactoryManagerCoordinators that are hard to reason about.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -387,17 +413,30 @@ open class NavigationEventsManager {
...
+    func incrementDistanceTraveled(by distance: CLLocationDistance) {
+        sessionState.totalDistanceCompleted += distance
+    }
+    
+    func arriveAtWaypoint() {
+        sessionState.departureTimestamp = nil
+        sessionState.arrivalTimestamp = nil
+    }
+    
+    func record(locations: [CLLocation]) {
```
1ec5, line 439: Hide the argument’s label using `_`, since at the point of use, `record(locations: locations)` would be redundant.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -254,26 +279,27 @@ open class NavigationEventsManager {
     }
     
     func enqueueFeedbackEvent(type: FeedbackType, description: String?) -> UUID? {
-        guard let eventDictionary = (try? navigationFeedbackEvent(type: type, description: description)?.asDictionary()) as [String: Any]?? else { return nil }
+        guard let feedbackEvent = navigationFeedbackEvent(type: type, description: description) else { return nil }
+        let eventDictionary = try? feedbackEvent.asDictionary()
```
1ec5, line 283: It’s a bit awkward that this old code had to convert the NavigationEventDetails to a Dictionary just to create a new FeedbackEvent. Maybe there’s a more direct way of doing that, but at least we should add a convenience initializer to FeedbackEvent that takes NavigationEventDetails directly.

1ec5 (approved):
Just a few minor documentation and API naming nits, but otherwise I don’t see any glaring issues left.

Sources/MapboxCoreNavigation/ActiveNavigationEventDetails.swift
```diff
@@ -0,0 +1,233 @@
...
+        distanceCompleted = round(session.totalDistanceCompleted + dataSource.routeProgress.distanceTraveled)
+        distanceRemaining = round(dataSource.routeProgress.distanceRemaining)
+        durationRemaining = round(dataSource.routeProgress.durationRemaining)
+        
+        rerouteCount = session.numberOfReroutes
+        
+        locationEngine = String(describing: dataSource.locationProvider)
+        locationManagerDesiredAccuracy = dataSource.desiredAccuracy
+        
+
```
bamx23, line 107: To many empty lines here

bamx23 (approved):

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
azarovalex: Thanks, this logic looks strange indeed, hope we will find some time for refactoring after GA.

PR #3123, opened 2021-06-28
Update the mapSDK puck in NavigationMapView with location course for bearing

### LINKED ISSUE #3069: Improve the vanishing route line feature based on frame rate and location update
In the #2968 , we introduced the location indicator layer from `MapboxMaps` but due to several constraints, the vanishing route line is still depending on the Timer and the evenly animation of the UserCourseView.  But when we use the snapped location from the MapView for the user location update. The vanishing route line implementation method should also be updated based on the frame rate and the location update.

### LINKED ISSUE #3105: Update the mapSDK puck in NavigationMapView with location course for bearing
MapSDK now supports the CLLocationDirection/course for bearing in the [pr](https://github.com/mapbox/mapbox-maps-ios/pull/428) in v10.0.0-rc.1. We could remove our manual puck bearing update and have a smoothy puck movement.

ShanMa1991:
### Description
This pr is to fix #3105, to remove the manually puck heading update code in `navigationMapView.moveUserLocation(to:animated:)`  introduced in #2968 
- [x] fix #3105 , default the mapSDK puck use location course for puck bearing
- [x] fix #3069 , update the vanishing route line update with location consumer, synced with the location update from `mapView`

### Implementation
Use the `simulatesLocation` to switch between the `course` and `heading` as the puck bearing source, instead of manually stop and update the puck layer.

Create a `SimulatedLocationManger` based `NavigationLocationProvider` to provide location update during simulated session.

Add the `RouteOverlayController` as the location consumer for the `mapView`, to receive the location update.

Use the `renderFrameFinished` to check the location change under simulation, and push location update to allow the vanishing route line synced with puck.

Stop the location update from the CarPlay before navigation session starts, to avoid the background navigator trigger.

### Screenshots or Gifs
puckUpdate

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -474,7 +474,10 @@ open class NavigationMapView: UIView {
                                   animated: animated,
                                   navigationCameraState: navigationCamera.state)
             
-        default: break
+        default:
+            if simulatesLocation, let locationProvider = mapView.location.locationProvider {
+                mapView.location.locationProvider(locationProvider, didUpdateLocations: [location])
+            }
         }
     }
     
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -477,6 +477,7 @@ open class NavigationMapView: UIView {
         default:
             if simulatesLocation, let locationProvider = mapView.location.locationProvider {
                 mapView.location.locationProvider(locationProvider, didUpdateLocations: [location])
+                mapView.location.locationProvider.stopUpdatingLocation()
             }
         }
     }
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -172,7 +172,16 @@ open class NavigationMapView: UIView {
         }
     }
     
-    var simulatesLocation: Bool = true
+    var simulatesLocation: Bool = true {
+        didSet {
+            if simulatesLocation {
+                mapView.location.options.puckBearingSource = .course
+                
+            } else {
+                mapView.location.options.puckBearingSource = .heading
+            }
+        }
+    }
     
     /**
      Specifies how the map displays the user’s current location, including the appearance and underlying implementation.
@@ -262,6 +271,7 @@ open class NavigationMapView: UIView {
                 self.mapView.location.options.puckType = .puck3D(configuration)
             }
         }
+        mapView.location.options.puckBearingSource = simulatesLocation ? .course : .heading
     }
     
     deinit {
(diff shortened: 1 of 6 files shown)


ShanMa1991:
According to the log below, the sequence when we call `func presentAndRemoveNavigationMapView(_:completion:)` in ViewController is as:
```
1. navigator starts, providing route progress
2. Two location updates source exist, one from the still existing navigationMapView with the `passiveLocationManager`, one from the navigator
3. the `present(navigationViewController, animated: true)` is called, the navigationMapView is later removed from the superView and set to nil, only one location update source exists.
```
So when we try to start the navigation, the navigator starts before the navigationMapView is removed. It caused rerouting behavior under simulation.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -172,7 +172,16 @@ open class NavigationMapView: UIView {
...
+    var simulatesLocation: Bool = true {
+        didSet {
+            if simulatesLocation {
+                mapView.location.options.puckBearingSource = .course
+                
+            } else {
+                mapView.location.options.puckBearingSource = .heading
+            }
+        }
+    }
```
1ec5, line 184: Can you add a comment explaining the logic in this property and its `didSet`? At a glance, I get the impression that the puck would be rotated based on the course during simulation (as in SimulatedLocationManager or ReplayLocationManager) but rotated based on the heading during actual navigation. Intuitively, I’d expect the puck to be rotated based on the course when it’s available, only falling back to heading otherwise.

Is the discrepancy because of tunnel simulation? I suppose in a deep tunnel, there may be no course because there’s no GPS location update, but there would likely be a heading because the magnetometer would still function normally.

(#1942 tracks rotating the camera (and puck) based on heading while walking, regardless of the location manager.)

1ec5 (approved):
There’s one part of the PR that I’m unclear about – probably just an awkward situation that needs comments – but otherwise it’s good to go once that’s resolved.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: So right now, the `simulatesLocation` dose two thing:
1. To decide the puck bearing source. During simulated navigation, we use the course for puck bearing because we have no heading data. When the simulated navigation ends, the turn-by-turn navigation is using CLHeading( maybe we should set the puck bearing from course as default when the navigator starts? ).
2. When set up mapView, to keep the route line synced with location puck. if we're using the default `userCourseView`, the route line in navigation is listening to `mapView` location update while the `userCourseView` is using animation through route controller update. We should keep them synced by sending location update. If we're using map puck, in simulated navigation, the location update is should be triggered and sent to the map by frame.
 https://github.com/mapbox/mapbox-navigation-ios/blob/1b7ebf366baf7773d967c0805a8659b51a840c77/Sources/MapboxNavigation/NavigationMapView.swift#L340-L350

So I think the best way to implement the heading/course, and the simulated location update is to split the `simulatesLocation` into two properties. One is public to control the puck bearing source, which could adjust to tunnel check or walking situation and used by customer on their purposes. One is to control the location update from the location manager, decide whether should we manually sending location update to the location consumers.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -262,6 +271,7 @@ open class NavigationMapView: UIView {
                 self.mapView.location.options.puckType = .puck3D(configuration)
             }
         }
+        mapView.location.options.puckBearingSource = simulatesLocation ? .course : .heading
```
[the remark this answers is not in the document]
ShanMa1991: Thanks for the suggestion, I'll improve this part.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -262,6 +262,7 @@ open class NavigationMapView: UIView {
                 self.mapView.location.options.puckType = .puck3D(configuration)
             }
         }
+        mapView.location.options.puckBearingSource = .course
```
ShanMa1991, line 265: Right now the map puck bearing source defaults to the location course, while the `simulatesLocation` is only used to notify the `navigationMapView` that we're using simulated location provider, and keep the puck synced with vanishing route line synced with user location layer.

PR #3127, opened 2021-06-29
Improve performance of Array.boundingBoxPoints by 5x

S2Ler:
Instead of iterating by all points many times, we can solve the problem
in one iteration.

Screenshot 2021-06-29 at 10 14 14


chezzdev (approved):
LGTM

PR #3131, opened 2021-06-30
Add slack notifications for failed builds on main branch

S2Ler:
Currently, we have no visibility into our failed "nightly" builds (which are run as "extended-workflow" in CircleCI). 
Failed builds on main are reported by email, but it would be good to post notifications in Slack as well. 

So this PR adds elementary Slack integration, which reports failed builds on main in Slack.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,30 @@
 version: 2.1
 
+orbs:
+  slack: circleci/slack@4.4.2
+
+commands:
+  notify-build-finished:
+      parameters:
+        notify_success:
+          type: boolean
+          default: false
+      steps:
+        - when:
+            condition: << parameters.notify_success >>
+            steps:
+              - slack/notify:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,5 +1,18 @@
...
 
+orbs:
+  slack: circleci/slack@4.4.2
+
+commands:
+  notify-build-failed:
+      steps:
+        - slack/notify:
+            channel: mobile-bots
+            event: fail
```
1ec5, line 11: If the build passes, will the Slack channel get any notification at all? It might be nice to get a notification there, without the `@here` mention, just for consistency. I don’t think it would be too much noise if it only runs nightly.

1ec5 (approved):

.circleci/config.yml
S2Ler: Thanks for the suggestion, will take a look how to archive this

PR #3138, opened 2021-07-01
Fix some tests

S2Ler:
Some of the tests fails sporadically. This is an attempt to fix that.

### DIFF
--- a/Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
+++ b/Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
@@ -503,15 +503,22 @@ class NavigationServiceTests: XCTestCase {
         let now = Date()
         let trace = Fixture.generateTrace(for: route).shiftedToPresent()
 
-        trace.forEach { navigation.router.locationManager!(navigation.locationManager, didUpdateLocations: [$0]) }
+        trace.forEach {
+            navigation.router.locationManager!(navigation.locationManager, didUpdateLocations: [$0])
+            RunLoop.main.run(until: Date().addingTimeInterval(0.01))
+        }
         
-        waitForNavNativeCallbacks(timeout: 0.5)
-
         let eventsManagerSpy = navigation.eventsManager as! NavigationEventsManagerSpy
-        XCTAssertTrue(eventsManagerSpy.hasFlushedEvent(with: MMEEventTypeNavigationDepart))
+        expectation(description: "Depart Event Flushed") {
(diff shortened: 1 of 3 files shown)


Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -503,31 +503,45 @@ class NavigationServiceTests: XCTestCase {
         let now = Date()
         let trace = Fixture.generateTrace(for: route).shiftedToPresent()
 
-        trace.forEach { navigation.router.locationManager!(navigation.locationManager, didUpdateLocations: [$0]) }
+        trace.forEach {
+            navigation.router.locationManager!(navigation.locationManager, didUpdateLocations: [$0])
+            RunLoop.main.run(until: Date().addingTimeInterval(0.01))
```
azarovalex, line 508: You can use function `waitForNavNativeCallbacks(timeout: 0.01)` from this file, or just remove this function and change all occurrences to `RunLoop.main.run`, if it will work.

azarovalex (approved):
LGTM, I just hope that we will not end up having tests that run 10 minutes each.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
S2Ler: waitForNavNativeCallbacks doesn't wait for callback though? It is just the same runloop. We can back to this once all tests are working.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -503,31 +503,45 @@ class NavigationServiceTests: XCTestCase {
         let now = Date()
         let trace = Fixture.generateTrace(for: route).shiftedToPresent()
 
-        trace.forEach { navigation.router.locationManager!(navigation.locationManager, didUpdateLocations: [$0]) }
+        trace.forEach {
+            navigation.router.locationManager!(navigation.locationManager, didUpdateLocations: [$0])
```
MaximAlien, line 507: Sorry for being too peeky regarding forced unwrapping :) I guess we can just omit it here?

MaximAlien (approved):

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
S2Ler: Yep, will change

PR #3144, opened 2021-07-02
A quick fix for the location update under non-simulation case

ShanMa1991:
### Description
This pr is to trigger the location update when it's changing from simulated to non simulated location manager during navigation.

### DIFF
--- a/Sources/MapboxNavigation/NavigationViewController.swift
+++ b/Sources/MapboxNavigation/NavigationViewController.swift
@@ -853,6 +853,8 @@ extension NavigationViewController: NavigationServiceDelegate {
             component.navigationService(service, didEndSimulating: progress, becauseOf: reason)
         }
         navigationMapView?.mapView.location.overrideLocationProvider(with: AppleLocationProvider())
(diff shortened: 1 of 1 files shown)


1ec5 (approved):
Let’s keep in mind this need to balance starting and stopping the location provider once we have tests in this area.

ShanMa1991:
@1ec5 . Yes, In the `MapboxMaps`, whenever we override a location provider, the previous location prover would stop location update first and then nilled out:
https://github.com/mapbox/mapbox-maps-ios/blob/8eff2d63cc7ad5517ae797115dd7bde64027d1db/Sources/MapboxMaps/Location/LocationManager.swift#L77-L82

So during navigation, when we switch the location provider through this way, it would avoid the previous location provider sending update to current `mapView`.

And in the #3123 , I stopped the `navigationMapView` location update just before we start navigation in `CarPlayManager`.
https://github.com/mapbox/mapbox-navigation-ios/blob/9e8edb37a3ccebd41e35ffbe6a8e1f92a1b47bd9/Sources/MapboxNavigation/CarPlayManager.swift#L231-L233

PR #3147, opened 2021-07-05
Implement new Billing model

S2Ler:
- [x] Integrate with MapboxCommon
- [x] Address billing methods question in `BillingHandler.getSessionToken()`
- [x] Test that billing events work correctly
- [x] Find out what to do with all the edge cases


Sources/MapboxCoreNavigation/Accounts.swift
```diff
@@ -11,18 +11,6 @@ enum NavigationBillingMethod: String {
 @objc(MBXAccounts)
 public class Accounts: NSObject {
     @objc public static var serviceSkuToken: String? {
-        var billingMethodValue = Bundle.main.object(forInfoDictionaryKey: "MBXNavigationBillingMethod") as? String
```
S2Ler, line 14: How should we handle billing methods now? @1ec5 do you have thoughts on that?

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -569,14 +574,47 @@ extension RouteController: Router {
...
+        }
+
         let routeOptions = routeOptions ?? routeProgress.routeOptions
-        routeProgress = RouteProgress(route: routes[indexedRouteResponse.routeIndex], options: routeOptions)
+        routeProgress = RouteProgress(route: route, options: routeOptions)
         self.indexedRouteResponse = indexedRouteResponse
         updateNavigator(with: routeProgress)
     }
+
+    private func shouldStartNewBillingSession(for newRoute: Route, routeOptions: RouteOptions?) -> Bool {
```
Udumft, line 593: This logic seems to be business-related. Do you have confirmation for the conditions when we do or do not bill for changing active route?

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,379 @@
...
+            lock.unlock(); return
+        }
+        _sessions[uuid] = nil
+        lock.unlock()
+        beginBillingSession(for: session.type, uuid: uuid)
+    }
+
+    private func _sessionTypeForRequests() -> SessionType? {
+        for session in _sessions.values {
+            if session.type == .activeGuidance {
```
Udumft, line 346: Why is `activeGuidance` preferred over free drive? Could we just pick 'any' session from `sessions`?

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -569,14 +574,47 @@ extension RouteController: Router {
...
+
+        let newRouteWaypoints = routeOptions.waypoints.dropFirst()
+        let currentRouteRemaingWaypoints = routeProgress.remainingWaypoints
+
+        guard newRouteWaypoints.count == currentRouteRemaingWaypoints.count else {
+            return true
+        }
+
+        for (newWaypoint, currentWaypoint) in zip(newRouteWaypoints, currentRouteRemaingWaypoints) {
+            if newWaypoint.coordinate.distance(to: currentWaypoint.coordinate) > 500 {
```
bamx23, line 611: Why `500`? It's better to move hardcoded literals to some constants

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,379 @@
...
+                return 43200 /*12h*/
+            case .freeDrive:
+                return 3600 /*2h*/
+            }
+        }
+    }
+
+    /// Shared billing handler instance. There is no other instances of `BillingHandler`.
+    private(set) static var shared: BillingHandler = {
+        let accessToken = Directions.shared.credentials.accessToken
```
bamx23, line 177: Is it okay that we don't allow setting custom credentials here?

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,379 @@
...
+            lock.unlock(); return
+        }
+        _sessions[uuid] = nil
+
+        let triggerBillingServiceEvents = !_hasSession(with: session.type)
+            && billingService.getSessionStatus(for: session.type) != .stopped
+        lock.unlock()
+
+        if triggerBillingServiceEvents {
+            billingService.stopBillingSession(for: session.type)
```
Udumft, line 285: Looks like we should also check if there are not active sessions in `_sessions` with the same `session.type`.
User can instantiate 2 `PassiveLocationManager`'s and then deallocate 1. This will call `stopBillingSession(for: .freeDrive)` but the second location manager will remain alive. Even `sessionState(uuid:)` will report that such session is still alive and updates will continue.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,379 @@
...
+    enum SessionType: Equatable {
+        case freeDrive
+        case activeGuidance
+
+        var maxSessionInterval: TimeInterval {
+            switch self {
+            case .activeGuidance:
+                return 43200 /*12h*/
+            case .freeDrive:
+                return 3600 /*2h*/
```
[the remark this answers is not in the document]
S2Ler: yes, 1h, thanks

Sources/MapboxCoreNavigation/BillingHandler.swift
S2Ler: @1ec5

Sources/MapboxCoreNavigation/BillingHandler.swift
S2Ler: business rules.

Sources/MapboxCoreNavigation/RouteController.swift
S2Ler: Discussed

Sources/MapboxCoreNavigation/BillingHandler.swift
1ec5: This allows the developer to set credentials in Info.plist but not at runtime. I suppose that’s unavoidable with something that could potentially be invoked extremely early in the application lifecycle. If a developer needs to set different credentials for specific services, they can still override those at runtime.

Sources/MapboxCoreNavigation/NSLock++.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+@_spi(MapboxInternal)
+extension NSLock {
```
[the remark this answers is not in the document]
1ec5: How about naming it NSLock+MapboxInternal.swift, since we’re using the MapboxInternal SPI?

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -94,10 +100,25 @@ open class PassiveLocationManager: NSObject {
 
         lastRawLocation = locations.last
     }
+
+    /// Suspends the driving session.
+    ///
+    /// Use this method when you no longer need to receive updates of location status to preserve existing billing session.
+    public func pauseDriveSession() {
```
1ec5, line 107: Consider renaming to `stopTripSession()`: start/stop would be consistent with other PassiveLocationManager methods, and “trip” will be extensible to other modes of transportation that might be billed for.

1ec5 (commented):
I haven’t looked as closely as the others, but here are a few things that stuck out for me. Let me know if you need me to take a closer look.

Sources/MapboxCoreNavigation/NSLock++.swift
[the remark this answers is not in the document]
S2Ler: Done

Sources/MapboxCoreNavigation/BillingHandler.swift
bamx23: But we have `NavigationSettings.hared.directions`. Why not using them as it's also a part of the Core?

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -8,8 +11,12 @@ import MapboxDirections
  Unlike `Router` classes such as `RouteController` and `LegacyRouteController`, this class operates without a predefined route, matching the user’s location to the road network at large. You can use a passive location manager to determine a starting point for a route that you calculate using the `Directions.calculate(_:completionHandler:)` method. If the user happens to be moving while you calculate the route, the passive location manager makes it less likely that the route will begin with a short segment on a side road or driveway and a confusing instruction to turn onto the current road.
  
  To find out when the user’s location changes, implement the `PassiveLocationManagerDelegate` protocol, or observe `Notification.Name.passiveLocationManagerDidUpdate` notifications for more detailed information.
+
+ - important: Creating an instance of this type will start a Free Drive session which can be paused by `PassiveLocationManager.pauseTripSession()` if you don't need navigation updates temporarily. The trip session is stopped when the instance is deallocated. From more info read the [Pricing Guide](https://docs.mapbox.com/ios/beta/navigation/guides/pricing/).
```
1ec5, line 15: Some minor rewording to flow better; the other documentation comments below could use similar tweaks. We should also propagate these changes to the changelog and [migration guide](https://github.com/mapbox/mapbox-navigation-ios/wiki/Upgrading-from-v1.x-to-v2.0.x).

```suggestion
 - important: Creating an instance of this class will start a free-driving session. If the application goes into the background or you temporarily stop needing location updates for any other reason, temporarily pause the trip session using the `PassiveLocationManager.pauseTripSession()` method to avoid unnecessary costs. The trip session also stops when the instance is deinitialized. For more information, see the “[Pricing](https://docs.mapbox.com/ios/beta/navigation/guides/pricing/)” guide.
```

We’ll need to remember to update the link to [this guide](https://docs.mapbox.com/ios/beta/navigation/guides/pricing/) once we move the beta site to the main site.

1ec5 (commented):
Kudos, this implementation is more sophisticated than anything I would’ve come up with in the time you put it together, and I think you’ve done a pretty good job of covering your bases with a variety of edge cases.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,393 @@
+// IMPORTANT: Tampering with any file that contains billing code is a violation of our ToS and will result in enforcement of the
+// penalties stipulated in the ToS.
```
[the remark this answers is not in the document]
S2Ler: Thanks for suggestion it makes sense. Tracker for the url: https://github.com/mapbox/navigation-sdks/discussions/1120#discussioncomment-1264188

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,393 @@
...
+        @unknown default:
+            self = .unknown
+        }
+    }
+}
+
+/// Protocol for `BillingServiceNative` implementation. Inversing the dependency on `BillingServiceNative` allows us
+/// to unit test our implementation.
+protocol BillingService {
+    func getSKUTokenIfValid(for sessionType: BillingHandler.SessionType) -> String
```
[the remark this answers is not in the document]
S2Ler: The protocol won't break. `ProductionBillingService` implementation will break though for obvious reasons. All `BillingService` clients won't break. If `MBXBillingService` will change is a dramatic way, then it would probably require some changes to this protocol as well but we don't expect such changes in foreseeable future.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
S2Ler: Thanks for the edit, reads better!

Sources/MapboxCoreNavigation/Accounts.swift
```diff
@@ -11,18 +11,6 @@ enum NavigationBillingMethod: String {
...
-        
-        switch NavigationBillingMethod(rawValue: billingMethodValue ?? NavigationBillingMethod.user.rawValue) {
-        case .user:
-            return TokenGenerator.getSKUToken(for: .navigationMAUS)
-        case .request:
-            return nil
-        case .none:
-            preconditionFailure("Unrecognized billing method \(String(describing: billingMethodValue)). Valid billing methods are: \(NavigationBillingMethod.allValues.map { $0.rawValue }).")
-        }
+        return BillingHandler.shared.serviceSkuToken
```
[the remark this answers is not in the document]
S2Ler: Migration guide update is in `feature/pricing_v2` [wiki](git@github.com:mapbox/mapbox-navigation-ios.wiki.git) branch.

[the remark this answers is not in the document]
S2Ler: Changelog updated.

S2Ler:
@MaximAlien please take a look on CarPlay related change.

S2Ler:
acc2fec60a1f016d5c39d443d8ad629cdbafe718 fixes https://github.com/mapbox/mapbox-navigation-ios/issues/3329

cc @bamx23 @Udumft

S2Ler:
@1ec5 Please give a last read before we merge.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,400 @@
...
+    /// The token to use for service requests like `Directions` etc. 
+    var serviceSkuToken: String {
+        let sessionTypes: [BillingHandler.SessionType] = [.activeGuidance, .freeDrive]
+
+        for sessionType in sessionTypes {
+            switch billingService.getSessionStatus(for: sessionType) {
+            case .running:
+                return billingService.getSKUTokenIfValid(for: sessionType)
+            case .paused, .stopped:
+                continue
```
bamx23, line 234: NIT: It's a bit confusing to use `continue` in the `switch` statement. But not sure if `break` here is less confusing.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,400 @@
...
+        case stopped
+        /// There is an active paused billing session.
+        case paused
+        /// There is an active running billing session.
+        case running
+    }
+
+    /// Supported session types.
+    enum SessionType: Equatable {
+        case freeDrive
```
bamx23, line 166: Maybe it's better to use `passiveNavigation` here? As we are trying not to use the `freeDrive` term in the code.
Sorry if it was already mentioned here.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,400 @@
...
+        if var existingSession = _sessions[uuid] {
+            existingSession.isPaused = false
+            _sessions[uuid] = existingSession
+        }
+        else {
+            let session = Session(type: sessionType, isPaused: false)
+            _sessions[uuid] = session
+        }
+
+        let triggerBeginSessionEvents = billingService.getSessionStatus(for: sessionType) == .stopped
```
bamx23, line 270: Maybe it's better to use `switch` here to prevent call duplication? Or at least store the result of the call in a local variable.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,400 @@
...
+    func pauseBillingSession(with uuid: UUID) {
+        lock.lock()
+        guard var session = _sessions[uuid] else {
+            assertionFailure("Trying to pause non-existing session.")
+            lock.unlock(); return
+        }
+        session.isPaused = true
+        _sessions[uuid] = session
+
+
```
bamx23, line 334:
```suggestion

```

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -0,0 +1,400 @@
...
+}
+
+// MARK: - Tests support
+
+extension BillingHandler {
+    static func __createMockedHandler(with service: BillingService) -> BillingHandler {
+        BillingHandler(service: service)
+    }
+
+    static func __replaceShareInstance(with handler: BillingHandler) {
```
bamx23, line 397: NIT: `__replaceSharedInstance`?

Sources/MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -93,12 +95,14 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
         UIDevice.current.isBatteryMonitoringEnabled = true
 
         super.init()
-        
+        BillingHandler.shared.beginBillingSession(for: .activeGuidance, uuid: sessionUUID)
```
bamx23, line 98: Do we need a disclaimer at the end of this file about billing related code too?

Sources/MapboxCoreNavigation/BillingHandler.swift
S2Ler: Where is our SwiftFormat? :)

Sources/MapboxCoreNavigation/LegacyRouteController.swift
S2Ler: Good catch!

Sources/MapboxCoreNavigation/BillingHandler.swift
S2Ler: There are a lot of discussions about this term but there is no agreed name as of now. This is internal API and can be changed without any problems. I would like to stick with the term used in public docs for now.

Sources/MapboxCoreNavigation/BillingHandler.swift
S2Ler: `break` will stop the loop which isn't a desired outcome.

Sources/MapboxCoreNavigation/BillingHandler.swift
S2Ler: Thanks, I will try to improve this code.

Sources/MapboxCoreNavigation/BillingHandler.swift
S2Ler: You are right, confused with `if`. 
Anyway, don't see a problem with continue. It is consistent with `return` above.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -24,6 +24,10 @@ public typealias ContainerViewController = UIViewController & NavigationComponen
  To be informed of significant events and decision points as the user progresses along the route, set the `NavigationService.delegate` property of the `NavigationService` that you provide when creating the navigation options.
  
  `CarPlayNavigationViewController` manages the corresponding user interface on a CarPlay screen.
+
+ - important: Creating an instance of this type with parameters that uses `RouteController` with will start an Active
+ Guidance session. The trip session is stopped when the instance is deallocated. From more info read the
```
MaximAlien, line 29:
```suggestion
 Guidance session. The trip session is stopped when the instance is deallocated. For more info read the
```

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -118,10 +128,31 @@ open class PassiveLocationManager: NSObject {
...
     }
-    
+
+    /**
+     Pauses the Free Drive session.
+
+     Use this method to extend the existing Free Drive session if you temporarily don't need navigation updates. For
+     more info, read the [Pricing Guide](https://docs.mapbox.com/ios/beta/navigation/guides/pricing/).
+     */
+    public func pauseTripSession() {
```
MaximAlien, line 138: Nit, but naming might be unclear for the end user. When I look at it, it is unclear why name contains `trip`, basically in free drive there is no trip as there is not final destination.

MaximAlien (commented):
CarPlay related logic seems to look good.

Udumft (approved):
LGTM except current merge conflict and NIT code doc typo above

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
S2Ler: Thanks for feedback! This was a [rename](https://github.com/mapbox/mapbox-navigation-ios/pull/3147#discussion_r689764134) from `pauseDriveSession` to `pauseTripSession`. This method is very billing related and you need to understand what is a trip to use it. That's why there is a link to the pricing guide (it is not up yet). And the fact that you confused by `trip` is a good thing because it will encourage you to read the docs ;)

PR #3150, opened 2021-07-07
Fix issue, which prevents from disabling traffic.

### LINKED ISSUE #3130: Setting `MapView.showsTraffic` to `false` no longer works since update to MapboxMaps `v10.0.0-beta.21`.
**Mapbox Navigation SDK version:** since `v2.0.0-beta.13`

### Steps to reproduce

 1. Open example application.
 2. Start active guidance navigation.
 3. Observe behavior.

### Expected behavior

Traffic should not be shown after calling `NavigationMapView.MapView.showsTraffic = false`.

### Actual behavior

Traffic is always shown in active guidance.

MaximAlien:
### Description

Closes #3130.

Issue was caused by local variable/parameter naming clash.

### DIFF
--- a/Sources/MapboxNavigation/MapView.swift
+++ b/Sources/MapboxNavigation/MapView.swift
@@ -44,7 +44,7 @@ extension MapView {
      - parameter layerIdentifier: Identifier of the layer in the tile set; in other words, a source layer identifier. Not to be confused with a style layer.
      */
     func showsTileSet(withIdentifier tileSetIdentifier: String, layerIdentifier: String) -> Bool {
-        let incidentsSourceIdentifiers = sourceIdentifiers(tileSetIdentifier)
+        let sourceIdentifiers = self.sourceIdentifiers(tileSetIdentifier)
         
         for layerIdentifier in mapboxMap.style.allLayerIdentifiers.map({ $0.id }) {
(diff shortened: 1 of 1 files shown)


S2Ler (approved):
It was fun to debug probably :) Nice catch!

Sources/MapboxNavigation/MapView.swift
```diff
@@ -72,17 +72,17 @@ extension MapView {
     func setShowsTileSet(_ isVisible: Bool, withIdentifier tileSetIdentifier: String, layerIdentifier: String) {
```
S2Ler, line 72: Is it possible to test this method without changes to the SDK?

If not, can you share you vision how would you test this fix if you had to?

azarovalex (approved):

Sources/MapboxNavigation/MapView.swift
MaximAlien: I think that testing flow is straightforward in this case: create `MapView` instance, wait till style loads up, disable traffic by setting `MapView.showsTraffic = false`, verify that traffic was really disabled by checking if `MapView.showsTraffic` is set to `false`. 

On the other hand major nuance in implementation of this test is `accessToken` handling, since we do not expose tokens `MapView` will not be able to work with mocked one.

Manual test is pretty fast: start active guidance and make sure that `MapView` doesn't render congestions.

PR #3151, opened 2021-07-07
Add ability to pass custom config through `UserDefaults`

azarovalex:
### Description
Add ability to pass custom navigation config through UserDefaults. The default config will be deep-merged with the application-provided config and the second one will take preference in case of merge conflicts.

Also added new test approach, any feedback will be appreciated.

### DIFF
--- a/MapboxNavigation.xcodeproj/project.pbxproj
+++ b/MapboxNavigation.xcodeproj/project.pbxproj
@@ -8,7 +8,8 @@
 
 /* Begin PBXBuildFile section */
 		11D1F89E269601150053A93F /* NativeHandlersFactoryTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11D1F89D269601150053A93F /* NativeHandlersFactoryTests.swift */; };
-		11D1F8A02696048D0053A93F /* Dictionary.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11D1F89F2696048D0053A93F /* Dictionary.swift */; };
+		11D1F8A02696048D0053A93F /* Dictionary+DeepMerge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11D1F89F2696048D0053A93F /* Dictionary+DeepMerge.swift */; };
+		11D1F8A22696EBD40053A93F /* Dictionary+Equality.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11D1F8A12696EBD40053A93F /* Dictionary+Equality.swift */; };
 		160D8279205996DA00D278D6 /* DataCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = 160D8278205996DA00D278D6 /* DataCache.swift */; };
 		160D827B2059973C00D278D6 /* DataCacheTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 160D827A2059973C00D278D6 /* DataCacheTests.swift */; };
 		162039CF216C348500875F5C /* NavigationEventsManagerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 162039CE216C348500875F5C /* NavigationEventsManagerTests.swift */; };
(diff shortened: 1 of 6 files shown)


Sources/MapboxCoreNavigation/NativeHandlersFactory.swift
```diff
@@ -1,6 +1,8 @@
 import MapboxNavigationNative
 import MapboxDirections
+import Foundation
 
+private let customConfigKey = "com.mapbox.navigation.custom-config"
```
1ec5, line 5: Good idea to prefix the key with a reverse DNS prefix. We should do the same for the `features` key that the `configHandle` getter expects.

1ec5: Would it be better to make this constant public but undocumented? It would reduce the potential for outdated code across the stack if the key ever changes.

1ec5 (approved):

MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -1515,6 +1519,7 @@
 				C561735A1F182113005954F6 /* RouteStep.swift */,
 				351927351F0FA072003A702D /* ScreenCapture.swift */,
 				3A8187C824BDAE9C00708F19 /* URLSession.swift */,
+				11D1F89F2696048D0053A93F /* Dictionary.swift */,
```
S2Ler, line 1522: This file contains only an extension on Dictionary, but from outside it seems like a new ``Dictionary`` type. On Apple platforms there is a non-formal convention to append something to the file name. For example:
- ``Dictionary++``
- ``Dictionary+Extensions``
- ``Dictionary+[GroupName]``

This approach is already used in our repository.

Sources/MapboxCoreNavigation/NativeHandlersFactory.swift
```diff
@@ -87,10 +92,13 @@ class NativeHandlersFactory {
...
-        if let jsonDataConfig = try? JSONSerialization.data(withJSONObject: historyAutorecordingConfig, options: []),
+        var customConfig = UserDefaults.standard.dictionary(forKey: customConfigKey) ?? [:]
+        customConfig.deepMerge(with: historyAutorecordingConfig)
+        
+        var customConfigJSON = ""
+        if let jsonDataConfig = try? JSONSerialization.data(withJSONObject: customConfig, options: []),
            let encodedConfig = String(data: jsonDataConfig, encoding: .utf8) {
-            customConfig = encodedConfig
+            customConfigJSON = encodedConfig
         }
```
S2Ler, line 102: [Optional] You can keep `customConfigJSON` as constant by:
```swift
let customConfigJSON: String
if let jsonDataConfig = try? JSONSerialization.data(withJSONObject: customConfig, options: []),
  let encodedConfig = String(data: jsonDataConfig, encoding: .utf8) {
  customConfigJSON = encodedConfig
}
else {
  customConfigJSON = ""
}
```

This way your clearly convey that this property won't change underneath anywhere else (helps to read code).

Sources/MapboxCoreNavigation/NativeHandlersFactory.swift
```diff
@@ -87,10 +92,13 @@ class NativeHandlersFactory {
             ]
         ]
         
-        var customConfig = ""
-        if let jsonDataConfig = try? JSONSerialization.data(withJSONObject: historyAutorecordingConfig, options: []),
+        var customConfig = UserDefaults.standard.dictionary(forKey: customConfigKey) ?? [:]
+        customConfig.deepMerge(with: historyAutorecordingConfig)
+        
+        var customConfigJSON = ""
+        if let jsonDataConfig = try? JSONSerialization.data(withJSONObject: customConfig, options: []),
```
S2Ler, line 99: Is this expected that `customConfig` serialisation can fail? Not sure about putting an assert, because they can distract developer flows, but at very least we can put something in logs to help catch that something is wrong.

Tests/MapboxCoreNavigationTests/NativeHandlersFactoryTests.swift
```diff
@@ -0,0 +1,76 @@
...
+    
+    override class func build(for profile: SettingsProfile,
+                              config: NavigatorConfig,
+                              customConfig: String) -> ConfigHandle {
+        passedCustomConfig = customConfig
+        return super.build(for: profile, config: config, customConfig: customConfig)
+    }
+}
+
+class NativeHandlersFactoryTests: XCTestCase {
```
S2Ler, line 19: Props for writing tests!

S2Ler (commented):
I have few stylistic comments and one about logging errors, but good work overall! 

Special thanks for writing tests!

Sources/MapboxCoreNavigation/Dictionary.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+extension Dictionary where Value == Any {
+    mutating func deepMerge(with dictionary: Dictionary) {
```
bamx23, line 4: It isn't clear which value will be used during the merge(a new one or current). Maybe it's better to provide a closure here just like a standard `merge` does?

Tests/MapboxCoreNavigationTests/NativeHandlersFactoryTests.swift
```diff
@@ -0,0 +1,82 @@
...
+    
+    var handlersFactory: NativeHandlersFactory!
+    
+    override func setUp() {
+        UserDefaults.standard.set(nil, forKey: customConfigKey)
+        handlersFactory = NativeHandlersFactory(tileStorePath: "tile store path",
+                                                credentials: DirectionsCredentials(),
```
S2Ler, line 27: You can use top level: `directions` of `DirectionsSpy` type.

azarovalex: Thanks, I ended up using `Fixture.credentials`

bamx23 (approved):

PR #3153, opened 2021-07-07
Allow the hard stop of route line gradient

ShanMa1991:
### Description
This pr is to allow the hard stop and abrupt change of congestion level without the blurring effect on route line layer.

### Implementation
Use `Step` instead of the `interpolate ` in expression.

### Screenshots or Gifs

Stop_Example


S2Ler (approved):
lgtm, but I didn't follow the issue closely.

Small nit: It is common to write comments starting with upper case. For example:
`Update the changelog` 
instead of 
`update the changelog`.

MaximAlien (approved):

ShanMa1991:
@S2Ler  Thanks for the suggestion, I'll improve the commit comments info!

PR #3154, opened 2021-07-08
Two step Feedback flow

azarovalex:
### Description
The idea is to allow the app to postpone the second step of the feedback. Previously the app had only 20 seconds to update the feedback with subtype until it will be sent. Now we want the application to decide itself when they want to show the second part of the feedback flow. One of the use cases is to show the feedback subtype screen when after the user finishes the navigation.

### Implementation
`MapboxCoreNavigation` now has a new feedback API that contains only two methods:
1. `createFeedback() -> FeedbackEvent?` – creates the feedback model that contains all the information we will send to the sever and returns this model to the app. Also creates the screenshot of the map, so should be called before any feedback UI is presented. Thus this method doesn't have any parameters anymore – feedback type and descriptions will be added using the second method.
2. `sendFeedback(_ feedback: FeedbackEvent, type: FeedbackType, description: String? = nil)` – updates the feedback model with feedback type and description, sends it to the server.

`FeedbackModel` confirms the Codable protocol, so the application can store it in persistent storage (allows to complete the feedback next time the app will be open).

**TODO checklist:**
- [x] Update changelog.
- [x] Update tests.


Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -2,8 +2,8 @@ import Foundation
 import MapboxDirections
 import Polyline
 
-class CoreFeedbackEvent: Hashable {
-    var id = UUID()
+public class CoreFeedbackEvent: Hashable, Codable {
```
1ec5, line 5: Remember to add class and method documentation for any public class that’s intended to be used by the application.

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -2,8 +2,8 @@ import Foundation
...
+    public var id = UUID()
```
1ec5, line 6: If we’re making this property public, then we should make sure to align it to platform conventions: `identifier`.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -336,40 +326,36 @@ open class NavigationEventsManager {
...
      
-     @param type A `FeedbackType` used to specify the type of feedback
-     @param description A custom string used to describe the problem in detail.
-     @return Returns a UUID used to identify the feedback event
+     @return Returns a feedback event
      
      If you provide a custom feedback UI that lets users elaborate on an issue, you should call this before you show the custom UI so the location and timestamp are more accurate.
      
-     You can then call `updateFeedback(uuid:type:source:description:)` with the returned feedback UUID to attach any additional metadata to the feedback.
+     You can then call `sendFeedback(_:type:source:description:)` with the returned feedback to attach any additional metadata to the feedback and send it.
```
1ec5, line 337: This requirement should be in a `postcondition` callout so that developers don’t miss it.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -336,40 +326,36 @@ open class NavigationEventsManager {
     }
     
     /**
-     Send feedback about the current road segment/maneuver to the Mapbox data team.
+     Create feedback about the current road segment/maneuver to be sent to the Mapbox data team.
```
1ec5, line 329: Maybe this documentation should mention `NavigationViewController.detailedFeedbackEnabled` and the effect it has (or doesn’t have) on this custom deferred feedback flow.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -430,13 +430,10 @@ public class CarPlayNavigationViewController: UIViewController {
             self?.carInterfaceController.popTemplate(animated: true)
             
             // TODO: Fix this Demeter violation with proper encapsulation
-            guard let uuid = self?.navigationService.eventsManager.recordFeedback() else { return }
+            guard let feedback = self?.navigationService.eventsManager.createFeedback() else { return }
```
1ec5, line 433: Now that directly managing feedback is a valid application use case, this Demeter violation becomes more relevant. Perhaps we can add a pass-through property to NavigationService or even higher up in CarPlayManager so the application doesn’t necessarily have to dig several levels down to get the events manager.

1ec5 (commented):
We’ll want to craft a changelog blurb that not only lists the low-level public API additions and changes but also describes the new workflow that this PR enables.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
azarovalex: I'm not sure that mentioning `MapboxNavigation` classes in the Core module is a good idea. 🤔

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
azarovalex: Added a pass-through property to the `CarPlayManager`.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -77,6 +77,13 @@ public class CarPlayNavigationViewController: UIViewController {
         }
     }
     
+    /**
+     Provides methods for creating and sending user feedbacks.
```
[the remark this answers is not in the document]
azarovalex: The only public functions events manager has are about user feedback, so the comment looks legit. Fix the pluralization, thanks.

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -14,16 +22,38 @@ class CoreFeedbackEvent: Hashable {
...
+        var containter = encoder.container(keyedBy: CodingKeys.self)
+        try containter.encode(identifier, forKey: .id)
+        try containter.encode(timestamp, forKey: .timestamp)
+        let eventDictionaryData = try JSONSerialization.data(withJSONObject: eventDictionary)
+        try containter.encode(eventDictionaryData, forKey: .eventDictionaryData)
     }
 }
 
-class FeedbackEvent: CoreFeedbackEvent {
+public class FeedbackEvent: CoreFeedbackEvent {
```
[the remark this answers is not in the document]
azarovalex: Thanks for noticing, there is no need to have two public classes, I removed the inheritance and made only FeedbackEvent public, also added the comment.

CHANGELOG.md
```diff
@@ -123,7 +123,8 @@
...
+* Supported feedbacks during passive navigation. Set `PassiveLocationManager` as a data source for `NavigationEventsManager` and then use this class to create and send user feedback. `FeedbackViewController` provides a drop-in feedback screen and can be used during both passive and active navigation. ([#3122](https://github.com/mapbox/mapbox-navigation-ios/pull/3122))
+* Added `NavigationEventsManager.createFeedback()` method that now returns an instance of `FeedbackEvent`. Application can update this model with additional information and send to the Mapbox data team using new `NavigationEventsManager.sendFeedback(_:type:description:)` method. Use this API if you have custom feedback UI. If you use drop-in UI from `FeedbackViewController` feedback lifecycle will be managed internally. ([#3154](https://github.com/mapbox/mapbox-navigation-ios/pull/3154))
```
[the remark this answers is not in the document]
azarovalex: Thanks, updated the changelog according to your proposal, just added a new item about deferred feedback. Complete changelog now:

```
* You can now solicit user feedback about `PassiveLocationManager` and `NavigationMapView` outside of active turn-by-turn navigation. Set `PassiveLocationManager` as a data source for a `NavigationEventsManager`, then use `NavigationEventsManager` to create and send user feedback. You can use a `FeedbackViewController` to present the user with the same options as during turn-by-turn navigation. Alternatively, if you present a custom feedback UI, call the `NavigationEventsManager.createFeedback()` method and configure the resulting `FeedbackEvent` with any additional context. ([#3122](https://github.com/mapbox/mapbox-navigation-ios/pull/3122))
* You can now manage the feedback event lifecycle allowing users to submit additional details later. Use `NavigationEventsManager.createFeedback()` to create a `FeedbackEvent` and `NavigationEventsManager.sendFeedback(_:type:description:)` to send it to the Mapbox. `FeedbackEvent` conforms to the `Codable` protocol, so the application can store uncompleted feedbacks persistently. [#3154](https://github.com/mapbox/mapbox-navigation-ios/pull/3154)
```

CHANGELOG.md
```diff
@@ -124,7 +124,8 @@
...
+* You can now solicit user feedback about `PassiveLocationManager` and `NavigationMapView` outside of active turn-by-turn navigation. Set `PassiveLocationManager` as a data source for a `NavigationEventsManager`, then use `NavigationEventsManager` to create and send user feedback. You can use a `FeedbackViewController` to present the user with the same options as during turn-by-turn navigation. Alternatively, if you present a custom feedback UI, call the `NavigationEventsManager.createFeedback()` method and configure the resulting `FeedbackEvent` with any additional context. ([#3122](https://github.com/mapbox/mapbox-navigation-ios/pull/3122))
+* You can now manage the feedback event lifecycle allowing users to submit additional details later. Use `NavigationEventsManager.createFeedback()` to create a `FeedbackEvent` and `NavigationEventsManager.sendFeedback(_:type:description:)` to send it to the Mapbox. `FeedbackEvent` conforms to the `Codable` protocol, so the application can store uncompleted feedbacks persistently. [#3154](https://github.com/mapbox/mapbox-navigation-ios/pull/3154)
```
1ec5, line 128: I like the mental picture of putting feedback in the MapBox. 🙃

```suggestion
* You can now manage the feedback event lifecycle, allowing the user to submit additional details later. Use `NavigationEventsManager.createFeedback()` to create a `FeedbackEvent` and `NavigationEventsManager.sendFeedback(_:type:description:)` to send it to Mapbox. `FeedbackEvent` conforms to the `Codable` protocol, so your application can store incomplete feedback across sessions if necessary. ([#3154](https://github.com/mapbox/mapbox-navigation-ios/pull/3154))
```

1ec5 (approved):

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -2,10 +2,10 @@ import Foundation
 import MapboxDirections
 import Polyline
 
-class CoreFeedbackEvent: Hashable {
-    let id = UUID()
-    let timestamp: Date
-
+class CoreFeedbackEvent: Hashable, Codable {
```
S2Ler, line 5: Curious, why it is a class instead of struct?

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -2,10 +2,10 @@ import Foundation
...
+    var identifier = UUID()
```
S2Ler, line 6: What mutability gives us here?

S2Ler: Why the rename from id to identifier?

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -14,38 +14,33 @@ class CoreFeedbackEvent: Hashable {
...
-            eventDictionary["feedbackSubType"] = subtypeList
-        }
-        eventDictionary["source"] = source.description
-        eventDictionary["description"] = description
+    required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        identifier = try container.decode(UUID.self, forKey: .id)
+        timestamp = try container.decode(Date.self, forKey: .timestamp)
+        let eventDictionaryData = try container.decode(Data.self, forKey: .eventDictionaryData)
+        eventDictionary = try JSONSerialization.jsonObject(with: eventDictionaryData) as? [String: Any] ?? [:]
```
S2Ler, line 35: Is it expected that the result json isn't `[String: Any]`? If not, then this conversion will silently fail and you will never know about this.

Sources/MapboxCoreNavigation/FeedbackEvent.swift
```diff
@@ -0,0 +1,34 @@
...
+/**
+ Feedback event that can be created using `NavigationEventsManager.createFeedback()`.
+ Use `NavigationEventsManager.sendFeedback(_:type:description:)` to send it to the server.
+ Conforms to the `Codable` protocol, so the application can store the event persistently.
+ */
+public class FeedbackEvent: Codable {
+    let coreEvent: CoreFeedbackEvent
+    
+    init(eventDetails: NavigationEventDetails) {
+        let dictionary = try? eventDetails.asDictionary()
```
S2Ler, line 12: Should we really silently fail here?

Sources/MapboxCoreNavigation/FeedbackEvent.swift
```diff
@@ -0,0 +1,34 @@
...
+        coreEvent = CoreFeedbackEvent(timestamp: Date(), eventDictionary: dictionary ?? [:])
+    }
+    
+    func update(type: FeedbackType, source: FeedbackSource, description: String?) {
+        coreEvent.eventDictionary["feedbackType"] = type.description
+
+        // if there is a subtype for this event then append the subtype description to our list for this type of feedback
+        if let subtypeDescription = type.subtypeDescription {
+            var subtypeList = [String]()
+            if let existingSubtypeList = coreEvent.eventDictionary["feedbackSubType"] as? [String] {
```
S2Ler, line 22: Repeated `"feedbackSubType"` string constant, easy to forget to rename the second while renaming the first.

S2Ler, line 22: Should we silently fail here if `feedbackSubType` isn't an array of strings?

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -429,14 +436,10 @@ public class CarPlayNavigationViewController: UIViewController {
         let feedbackButtonHandler: (_ : CPGridButton) -> Void = { [weak self] (button) in
             self?.carInterfaceController.popTemplate(animated: true)
             
-            // TODO: Fix this Demeter violation with proper encapsulation
-            guard let uuid = self?.navigationService.eventsManager.recordFeedback() else { return }
```
S2Ler, line 433: Repeated weak self usage can lead to race conditions. If the logic can't be completed without self it is better to guard to strong self at the beginning of the block:
```swift
guard let self = self else { /* optional logic */; return }
```

S2Ler (commented):
I have some comments about the current implementation.

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
azarovalex: I guess this is because of inheritance, other events inherit from this class.

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
azarovalex: I renamed the field in accordance to our Swift style guide.
Also removed mulability.

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
azarovalex: I guess there is no way for this function to fail because we user doesn't have access to this field and we get this data from the real JSON object in line 42.

Sources/MapboxCoreNavigation/FeedbackEvent.swift
azarovalex: We are not failing in this case, this logic creates an empty array in case we don't have an array already.

Sources/MapboxCoreNavigation/FeedbackEvent.swift
azarovalex: Added an `assertionFailure`

Sources/MapboxCoreNavigation/FeedbackEvent.swift
azarovalex: Fixed, thanks for noticing.

S2Ler: Well, ignoring the value is kind of "silent fail". The code which doesn't do what it was asked for (even wrongly) is just a form of fail.

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
S2Ler: Then having an asserting won't hurt. Today everything is ok, tomorrow, something changes in the surrounding code and will start silently failing.

Sources/MapboxCoreNavigation/CoreFeedbackEvent.swift
```diff
@@ -2,52 +2,48 @@ import Foundation
 import MapboxDirections
 import Polyline
 
-class CoreFeedbackEvent: Hashable {
-    let id = UUID()
-    let timestamp: Date
-
+class CoreFeedbackEvent: Hashable, Codable {
+    let identifier: UUID
+    var timestamp: Date
```
S2Ler, line 7: Is timestamp modifies anywhere?

PR #3155, opened 2021-07-08
Add center distance and bearing change threshold before start the camera animation

### LINKED ISSUE #3152: Avoid animating the camera nowhere while the user is at rest
If the current `NavigationCameraState` is `following` but the user is stationary, the navigation SDK should not attempt to animate the camera in response to the redundant location updates that come in from Core Location. These extraneous animations waste power by requiring the map to continue redrawing continuously without any perceptible change on screen.

`NavigationCameraStateTransition.update(_:)` unconditionally creates and starts various animators. Before doing so, it could check whether the original and target locations and headings are within a certain threshold, perhaps varying the threshold based on the zoom level, and bail if the locations and headings are similar enough:

https://github.com/mapbox/mapbox-navigation-ios/blob/a8ad873cc581c170e2dcd1d203d5f3983cf6e569/Sources/MapboxNavigation/NavigationCameraStateTransition.swift#L133-L138

Here’s the full code path that leads to the extraneous animations, courtesy of @macdrevx:

```
#0	0x0000000106b7e130 in CameraAnimationsManager.makeAnimator(duration:timingParameters:animationOwner:animations:) at /path/to/mapbox-navigation-ios-examples/Pods/MapboxMaps/Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager+CameraAnimatorDelegate.swift:26
#1	0x0000000107359fdc in NavigationCameraStateTransition.update(_:) at /path/to/mapbox-navigation-ios-examples/Pods/MapboxNavigation/Sources/MapboxNavigation/NavigationCameraStateTransition.swift:134
#2	0x0000000107359874 in NavigationCameraStateTransition.updateForFollowing(_:) at /path/to/mapbox-navigation-ios-examples/Pods/MapboxNavigation/Sources/MapboxNavigation/NavigationCameraStateTransition.swift:105
#3	0x000000010735edec in protocol witness for CameraStateTransition.updateForFollowing(_:) in conformance NavigationCameraStateTransition ()
#4	0x000000010734d018 in NavigationCamera.viewportDataSource(_:didUpdate:) at /path/to/mapbox-navigation-ios-examples/Pods/MapboxNavigation/Sources/MapboxNavigation/NavigationCamera.swift:99
#5	0x000000010734e9b0 in protocol witness for ViewportDataSourceDelegate.viewportDataSource(_:didUpdate:) in conformance NavigationCamera ()
#6	0x00000001073eb700 in NavigationViewportDataSource.progressDidChange(_:) at /path/to/mapbox-navigation-ios-examples/Pods/MapboxNavigation/Sources/MapboxNavigation/NavigationViewportDataSource.swift:111
#7	0x00000001073eba58 in @objc NavigationViewportDataSource.progressDidChange(_:) ()
#8	0x0000000180337af8 in __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ ()
#9	0x0000000180337ab8 in ___CFXRegistrationPost_block_invoke ()
#10	0x00000001803370d8 in _CFXRegistrationPost ()
#11	0x0000000180336ad0 in _CFXNotificationPost ()
#12	0x00000001807c18d8 in -[NSNotificationCenter postNotificationName:object:userInfo:] ()
#13	0x0000000106663aa8 in RouteController.update(progress:with:rawLocation:upcomingRouteAlerts:) at /path/to/mapbox-navigation-ios-examples/Pods/MapboxCoreNavigation/Sources/MapboxCoreNavigation/RouteController.swift:402
#14	0x0000000106660abc in RouteController.update(to:) at /path/to/mapbox-navigation-ios-examples/Pods/MapboxCoreNavigation/Sources/MapboxCoreNavigation/RouteController.swift:270
#15	0x0000000106660910 in closure #1 in RouteController.navigationStatusDidChange(_:) at /path/to/mapbox-navigation-ios-examples/Pods/MapboxCoreNavigation/Sources/MapboxCoreNavigation/RouteController.swift:263
```

ShanMa1991:
### Description
This pr is to fix #3152 by providing the center distance and bearing threshold before the camera animation when the current camera state is `following`.

### Implementation
Add the camera bearing change threshold as 5.0, when the difference between the target and original camera state is less than the threshold, the camera would not change.

Add the center location change threshold as 5 pixels under current zoom level, when the distance between the target and original location is less than the threshold, the camera would not change.


Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -14,6 +14,11 @@ public class NavigationCameraStateTransition: CameraStateTransition {
     var animatorBearing: BasicCameraAnimator?
     var animatorPitch: BasicCameraAnimator?
     
+    // The bearing change threshould to start animation of camera.
+    var bearingUpdateThreshold: CLLocationDirection = 5.0
+    // The pixels threshold on the target camera zoom level to start animation of camera.
+    var baseCenterUpdateThreshould: Double = 5.0
```
1ec5, line 20: Typo: threshold.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -102,6 +107,16 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
     
     public func updateForFollowing(_ cameraOptions: CameraOptions) {
+        // Check whether the original and target locations and headings are larger than a certain threshold when current camera state is following.
+        guard let mapView = mapView,
+              let center = cameraOptions.center,
+              let zoom = cameraOptions.zoom,
+              let bearing = cameraOptions.bearing else { return }
+
+        let centerUpdateThreshould = baseCenterUpdateThreshould * pow(2, 22.0 - Double(zoom))
+        guard mapView.cameraState.center.distance(to: center) > centerUpdateThreshould,
```
1ec5, line 117: `LocationCoordinate2D.distance(to:)` returns a physical distance in meters, whereas `centerUpdateThreshold` appears to be expressed in screen points.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -102,6 +107,16 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
+              abs(mapView.cameraState.bearing - bearing) > bearingUpdateThreshold else { return }
```
1ec5, line 118: Does this expression need to account for the possibility that  –1 can signify an unset `CLLocationDirection` value?

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -14,6 +14,11 @@ public class NavigationCameraStateTransition: CameraStateTransition {
     var animatorBearing: BasicCameraAnimator?
     var animatorPitch: BasicCameraAnimator?
     
+    // The bearing change threshould to start animation of camera.
+    var bearingUpdateThreshold: CLLocationDirection = 5.0
```
1ec5, line 18: The default `CLLocationManager.headingFilter` is 1°. It doesn’t look like MapboxNavigation or MapboxMaps configures `CLLocationManager.headingFilter` anywhere, so we should probably expect to align to the default heading filter for consistency between the camera and the puck.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -102,6 +107,16 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
+        // Check whether the original and target locations and headings are larger than a certain threshold when current camera state is following.
+        guard let mapView = mapView,
+              let center = cameraOptions.center,
+              let zoom = cameraOptions.zoom,
+              let bearing = cameraOptions.bearing else { return }
+
+        let centerUpdateThreshould = baseCenterUpdateThreshould * pow(2, 22.0 - Double(zoom))
+        guard mapView.cameraState.center.distance(to: center) > centerUpdateThreshould,
+              abs(mapView.cameraState.bearing - bearing) > bearingUpdateThreshold else { return }
+
```
1ec5, line 119: The way this guard statement is currently expressed, it would ignore any animation request that doesn’t change _both_ the map’s center location and its bearing. But there are situations where only one of the two needs to be animated, since this method serves as a bottleneck for all animations. Instead of returning early, I think we’ll need to conditionalize each section of the code below on whether it would have an effect.

1ec5 (commented):
As far as I can tell, in overview mode, the camera’s bearing would never change, so I think the bearing-related part of this PR could result in a nontrivial improvement to performance, but we’ll need to refine the logic a bit.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
ShanMa1991: I see. Thanks for the suggestion. I'll move it to specific check.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
ShanMa1991: I'll add the `=` for it.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
ShanMa1991: Thanks for the suggestion, I'll add the `getMetersPerPixelAtLatitude(_:_:)` for the physical distance in meters, and the pixel threshold as 5.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
ShanMa1991: I see. I'll use the default 1 for the value.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -144,13 +132,23 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
         }
-
-        let centerTimingParameters = UICubicTimingParameters(controlPoint1: CGPoint(x: 0.0, y: 0.0), controlPoint2: CGPoint(x: 1.0, y: 1.0))
-        animatorCenter = mapView.camera.makeAnimator(duration: duration, timingParameters: centerTimingParameters) { (transition) in
-            transition.center.toValue = center
+        
+        // Check whether the location change is larger than a certain threshold when current camera state is following.
+        var updateCameraCenter: Bool = true
+        if state == .following {
+            let metersPerPxiel = getMetersPerPixelAtLatitude(center.latitude, Double(zoom))
```
1ec5, line 139: Typo: pixel.

The latitude could differ between the original and target cameras, but at the moment I can’t think of a situation where that difference would be so great as to affect whether to avoid animating, yet so small as to cause the method to avoid animating. 👍

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -167,12 +165,20 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
+        if updateCameraBearing {
+            let bearingTimingParameters = UICubicTimingParameters(controlPoint1: CGPoint(x: 0.0, y: 0.0), controlPoint2: CGPoint(x: 1.0, y: 1.0))
+            animatorBearing = mapView.camera.makeAnimator(duration: duration, timingParameters: bearingTimingParameters) { (transition) in
+                transition.bearing.toValue = bearing
+            }
+            
+            animatorBearing?.startAnimation()
+        }
         
         if let animatorPitch = animatorPitch, animatorPitch.isRunning {
```
1ec5, line 183: Should we add a similar check for the pitch? Otherwise, a completely idle map will still be guaranteed to animate. Hopefully a preview/overview map would not want to animate its pitch anyways, so we can probably get away with a literal equality check here.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -167,12 +165,20 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
-        animatorBearing = mapView.camera.makeAnimator(duration: duration, timingParameters: bearingTimingParameters) { (transition) in
-            transition.bearing.toValue = bearing
+        // Check whether the bearing change is larger than a certain threshold when current camera state is following.
+        var updateCameraBearing: Bool = true
+        if state == .following {
+            updateCameraBearing = (abs(mapView.cameraState.bearing - bearing) >= 1.0)
         }
         
-        animatorBearing?.startAnimation()
+        if updateCameraBearing {
```
1ec5, line 174: `updateCameraBearing` only appears to be set once and then read in the following conditional statement. We can collapse this into one line.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -107,30 +105,20 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
-        
-        update(cameraOptions)
+        update(cameraOptions, state: .following)
     }
     
     public func updateForOverview(_ cameraOptions: CameraOptions) {
         var cameraOptions = cameraOptions
         cameraOptions.bearing = 0.0
-        update(cameraOptions)
+        update(cameraOptions, state: .overview)
```
1ec5, line 114: I wonder if we need these helper methods anymore. We could make `update(_:state:)` public and move the `cameraOptions.bearing = 0.0` statement into a conditional in that method.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -107,30 +105,20 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
     }
     
     public func cancelPendingTransition() {
         stopAnimators()
     }
     
-    func update(_ cameraOptions: CameraOptions) {
+    func update(_ cameraOptions: CameraOptions, state: NavigationCameraState) {
```
1ec5, line 121: If we do make this method public, I think `update(to:state:)` or `update(options:state:)` would be more accurate, because the method doesn’t modify `cameraOptions` in place as a side effect.

1ec5 (approved):
Let’s make sure to try out the changes to make sure expected animations don’t get ignored by mistake, since it can sometimes be hard to reason about how the three degrees of freedom in this method interact with each other.

I think this PR is headed in the right direction and rolling towards completion, so I’ve approved it. Feel free to merge once the remaining feedback is resolved.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
ShanMa1991: I see. Thanks for the suggestion, I'll replace the `update(_:)` with the `update(to:state:)`

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
ShanMa1991: I'll add the check for the pitch change as well to be aligned with the location center and bearing.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
ShanMa1991: I replaced the two helper methods with the `update(to:state:)`.

PR #3156, opened 2021-07-08
v2.0.0-beta.17

1ec5:
Updated the Xcode project, CocoaPods podspec, and readmes for v2.0.0-beta.17.

In addition to the usual updates, this PR upgrades to MapboxNavigationNative v56.0.0, which includes the following changes versus previous v2.0.0 betas:

* Fixed offline routing issues related to roundabouts that lead to service roads. (valhalla/valhalla#3129)
* An alternative route calculated offline is more likely to be meaningfully distinct from the preferred route than before. (valhalla/valhalla#3183)

Additionally, I took the opportunity to refactor `OpenLRStandard`, with a view toward future enhancements to the road objects/incidents API:

* Replaced the `OpenLRStandard` enumeration with an `OpenLRIdentifier` enumeration that includes a `RoadObjectIdentifier` as an associated value. `RoadObjectMatcher.matchOpenLR(location:standard:identifier:)` has been renamed to `RoadObjectMatcher.matchOpenLR(location:identifier:)`. (#3156)

And a typo in the jazzy table of contents was corrected.

/cc @mapbox/navigation-ios


ShanMa1991 (approved):

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -35,12 +35,13 @@ class Navigator {
...
      */
     func writeHistory(completionHandler: @escaping (URL?) -> Void) {
-        historyRecorder?.dumpHistory { (path) in
+        historyRecorder?.stopRecording { [weak self] (path) in
             if let path = path {
                 completionHandler(URL(fileURLWithPath: path))
             } else {
                 completionHandler(nil)
             }
+            self?.historyRecorder?.startRecording()
```
1ec5, line 44: According to the `HistoryRecorderHandle.stopRecording(result:)` documentation, we have to explicitly restart history recording after it’s done writing out history, a change from before when `dumpHistory(result:)` would keep history recording going.

1ec5: Unfortunately, this change seems to be causing [various test failures](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/4377/workflows/14112c3c-9914-414b-a086-b3e988d70685/jobs/38325):

MapboxCoreNavigationTests.PassiveLocationManagerTests
  testHistoryRecording, XCTAssertNotNil failed
  /Users/distiller/project/Tests/MapboxCoreNavigationTests/PassiveLocationManagerTests.swift:124
  ```
        PassiveLocationManager.writeHistory { url in
            XCTAssertNotNil(url)
            historyExpectation.fulfill()
  ```

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -142,6 +143,7 @@ class Navigator {
         navigator.setElectronicHorizonObserverFor(self)
         navigator.addObserver(for: self)
         navigator.setFallbackVersionsObserverFor(self)
+        historyRecorder?.startRecording()
```
1ec5, line 146: The relevant change in MapboxNavigationNative clearly expects the history recorder to be started explicitly at some point before dumping (or “stopping”) history, so I’ve added a bit here to start recording basically as soon as possible. In principle, we could start recording more lazily, perhaps based on some application-defined behavior, but we don’t have a convenient place to do that, so for now I’m just preserving the previous behavior.

/cc @mskurydin @DmitryAk

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
1ec5: #3157 moves this `startRecording()` call outside the callback to happen synchronously with `stopRecording(result:)`, as recommended by @DmitryAk.

PR #3157, opened 2021-07-09
Restart history recording before finishing writing history

1ec5:
Introduced explicit methods to start and stop history recording. For performance reasons, the navigation SDK no longer automatically records history upon starting a route or free-driving session as it did in previous v2.0.0 betas. This change also makes it possible for an application to avoid the gap in history recording that was introduced in #3156.

For the next beta’s release notes:

* Renamed the `RouteController.writeHistory(completionHandler:)` method to `RouteController.stopRecordingHistory(completionHandler:)` and the `PassiveLocationManager.writeHistory(completionHandler:)` method to `PassiveLocationManager.stopRecordingHistory(completionHandler:)`. ([#3157](https://github.com/mapbox/mapbox-navigation-ios/pull/3157))
* Added the `RouteController.startRecordingHistory()` and `PassiveLocationManager.startRecordingHistory()` methods. You are now responsible for calling this method before stopping and writing history. ([#3157](https://github.com/mapbox/mapbox-navigation-ios/pull/3157))

Original change: ~~Restored the previous history autorecording behavior by closing a gap while writing history during which no history was being recorded. This is a followup to #3156.~~

/cc @mapbox/navigation-ios @DmitryAk

### DIFF
--- a/Sources/MapboxCoreNavigation/PassiveLocationManager.swift
+++ b/Sources/MapboxCoreNavigation/PassiveLocationManager.swift
@@ -204,20 +204,33 @@ open class PassiveLocationManager: NSObject {
     }
     
     /**
+     Starts recording history for debugging purposes.
+     
+     - postcondition: Use the `stopRecordingHistory(writingFileWith:)` method to stop recording history and write the recorded history to a file.
+     */
+    public static func startRecordingHistory() {
+        Navigator.shared.startRecordingHistory()
+    }
+    
+    /**
      A closure to be called when history writing ends.
      
      - parameter historyFileURL: A path to file, where history was written to.
      */
-    public typealias WriteHistoryCompletionHandler = (_ historyFileURL: URL?) -> Void
+    public typealias HistoryFileWritingCompletionHandler = (_ historyFileURL: URL?) -> Void
     
     /**
-     Store history to the directory stored in `PassiveLocationManager.historyDirectoryURL` and asynchronously run a callback
-     when writing finishes.
+     Stops recording history, asynchronously writing any recorded history to a file.
      
-     - parameter completion: A block object to be executed when history writing ends.
+     Upon completion, the completion handler is called with the URL to a file in the directory specified by `PassiveLocationManager.historyDirectoryURL`. The file contains details about the passive location manager’s activity that may be useful to include when reporting an issue to Mapbox.
+     
+     - precondition: Use the `startRecordingHistory()` method to begin recording history. If the `startRecordingHistory()` method has not been called, this method has no effect.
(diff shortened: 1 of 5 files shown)


DmitryAk (approved):
LGTM!

S2Ler (approved):
lgtm

azarovalex (approved):

zugaldia:
What does autorecording mean in this context? We don't want to collect any history unless the developer explicitly asks the SDK to do so.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -190,6 +161,39 @@ open class PassiveLocationManager: NSObject {
         NotificationCenter.default.removeObserver(self)
     }
     
+    // MARK: Accessing Relevant Routing Data
```
1ec5, line 164: jazzy uses marks to generate named headings in the published API reference, which are very useful for grouping and linking to related functionality. We don’t have nearly enough marks of this sort in the codebase, but there are examples of it in SpeedLimitView for instance. The Title Case is annoying, but it’s the style Apple uses in their documentation. I’m open to changing it to Sentence case for consistency with internal-only marks.

1ec5 (commented):
I made substantial changes to this PR in light of https://github.com/mapbox/mapbox-navigation-ios/pull/3157#issuecomment-878328927, pushing the responsibility of starting and stopping history recording to the application level.

Along the way, I also reorganized a few classes to group history-related functionality together and also group electronic horizon functionality together. I’m not too keen on how much code duplication has resulted in an effort to avoid exposing the Navigator singleton, but at least this grouping keeps things reasonably intuitive for developers.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -199,21 +203,34 @@ open class PassiveLocationManager: NSObject {
...
+     - postcondition: Use the `stopRecordingHistory(writingFileWith:)` method to stop recording history and write the recorded history to a file.
+     */
+    public static func startRecordingHistory() {
+        Navigator.shared.startRecordingHistory()
+    }
+    
     /**
      A closure to be called when history writing ends.
      
      - parameter historyFileURL: A path to file, where history was written to.
```
DmitryAk, line 218: It would be better to describe that the `historyFileURL` could be `nil`, e.g. if history recording hadn't been started before the `stopRecordingHistory` was called.

DmitryAk (commented):
LGTM

bamx23 (approved):

S2Ler (approved):

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
1ec5: #3180 tracks adding more developer-facing mark comments.

PR #3159, opened 2021-07-09
Implement a new method to change route

### LINKED ISSUE #2948: Problems when changing the route in the NavViewController while navigating.
**Mapbox Navigation SDK version:**  v2.0.0-beta.5 (and latest v2.0 branch)

I'm currently having an issue with updating waypoints in the NavViewController after navigating a previous route. Basically, my use case is this:
- Driver navigates a series of waypoints. On completion of the final waypoint the driver is asked if they want to navigate back to the start.
- If the driver chooses to navigate back to the start, then a new route is generated consisting of the current location of the driver as the start point, and a destination waypoint being where they start. This should then replace the previous route in the navigation view controller and the driver would navigate to the new destination, any re-routes etc would be to that destination.

However, I'm having problems with this. I update NavViewController with the new route as follows:
```swift
Directions.shared.calculate(returnRouteOptions) { (session, result) in
   switch result {
    case .failure(let error):
        print("Error getting directions: \(error.localizedDescription)")
    case .success(let response):
        guard let route = response.routes?.first else {
            return
        }
        self.currentMapRoute = route
        // If we've already got a nav service then update it, otherwise create a new one
        self.currentNavigationViewController?.indexedRoute = IndexedRoute(route, 0)
        
    }
}
```

This does then plot the new route, however the map also shows the old destination waypoint from the previous route in addition to the new destination waypoint. The route also seems to take account of both. If the driver subsequently goes off route and a re-route event is triggered the update route always changes to go solely to the old destination rather than the new one. It seems that when updating the route data then some state is retained pointing to the old route.

Of course I could be missing something fundamental here and not updating the route correctly. I am struggling to find any references that would show me how to do this though. It does feel like a bug, however.

The following screenshots attempt to illustrate what I'm talking about:
- Initial route:
IMG_8163

- After stating that the 1st waypoint is complete and I want to navigate back to the beginning (initiated via a popup alert). Note the 2 destination markers:
IMG_8164

Also, weirdly there seems to be another bug around the Mapbox logo.. I end up with multiples in various scenarios... This is probably a different bug.

### Steps to reproduce

 1. Start navigating a route with some waypoints
 2. Create a new route with different waypoints
 3. Change the route used by the navigation controller to be the new route.

### Expected behavior
I would expect that the route that is shown in the navigation view controller along with the waypoints change and that re-routes apply to the new route.

### Actual behavior
The destination markers for both routes are shown. Re-routing applies to the first route.

S2Ler:
Fixes #2948

Previous method of just updating `indexedRoute` wasn't robust enough for
cases when the new route has different waypoints than the one it is
replacing. New method allows to update route from the different set of
waypoints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,7 @@
 * The `EventsManagerDataSource.router`, `NavigationService.router`, `NavigationService.eventsManager`, `MapboxNavigationService.router`, `MapboxNavigationService.eventsManager` properties are no longer force unwrapped. ([#3055](https://github.com/mapbox/mapbox-navigation-ios/pull/3055))
 * Fixed an issue where traffic congestion segments along the route line blurred into each other when the map was zoomed in far enough. ([#3153](https://github.com/mapbox/mapbox-navigation-ios/pull/3153))
 * Supported feedbacks during passive navigation. Set `PassiveLocationManager` as a data source for `NavigationEventsManager` and then use this class to create and send user feedbacks. `FeedbackViewController` provides drop-in feedback screen and can be used during both passive and active navigation. ([#3122](https://github.com/mapbox/mapbox-navigation-ios/pull/3122))
+* `NavigationViewController.indexedRoute`, `NavigationService.indexedRoute` and `Router.indexedRoute` properties are readonly now. Use dedicated `Router.updateRoute(with:routeOptions:)` method to update the route. ([#3159](https://github.com/mapbox/mapbox-navigation-ios/pull/#3159))
 
 ## v1.4.1
 
--- a/Sources/MapboxCoreNavigation/LegacyRouteController.swift
+++ b/Sources/MapboxCoreNavigation/LegacyRouteController.swift
@@ -58,12 +58,12 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
     }
     
     public var indexedRoute: IndexedRoute {
-        get {
-            return routeProgress.indexedRoute
-        }
-        set {
-            routeProgress.indexedRoute = newValue
-        }
+        routeProgress.indexedRoute
+    }
+
+    public func updateRoute(with indexedRoute: IndexedRoute, routeOptions: RouteOptions?) {
+        let routeOptions = routeOptions ?? routeProgress.routeOptions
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -122,6 +122,7 @@
 * Fixed a retain cycle in `UserCourseView`. ([#3120](https://github.com/mapbox/mapbox-navigation-ios/issues/3120))
 * The `EventsManagerDataSource.router`, `NavigationService.router`, `NavigationService.eventsManager`, `MapboxNavigationService.router`, `MapboxNavigationService.eventsManager` properties are no longer force unwrapped. ([#3055](https://github.com/mapbox/mapbox-navigation-ios/pull/3055))
 * Fixed an issue where traffic congestion segments along the route line blurred into each other when the map was zoomed in far enough. ([#3153](https://github.com/mapbox/mapbox-navigation-ios/pull/3153))
+* `NavigationViewController.indexedRouter`, `NavigationService.indexedRouter` and `Router.indexedRouter` properties are readonly now. Use dedicated `Router.updateRoute(with:routeOptions:)` method to update the route. ([#3159](https://github.com/mapbox/mapbox-navigation-ios/pull/#3159))
```
Udumft, line 125: `indexedRoute` misstypes

Sources/MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -58,12 +58,12 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
...
-        }
-        set {
-            routeProgress.indexedRoute = newValue
-        }
+        routeProgress.indexedRoute
+    }
+
+    public func updateRoute(with indexedRoute: IndexedRoute, routeOptions: RouteOptions?) {
+        let routeOptions = routeOptions ?? routeProgress.routeOptions
+        routeProgress = RouteProgress(route: indexedRoute.0, routeIndex: indexedRoute.1, options: routeOptions)
```
Udumft, line 66: shouldn't we also `announce(reroute:)` here as it was previously?

S2Ler: An announce happens in `routeProgress` setter.

CHANGELOG.md
S2Ler: Thanks, fixed

Udumft (approved):
LGTM

PR #3161, opened 2021-07-12
Fix feedback creation moment

### LINKED ISSUE #3134: Feedbacks are always sent with incorrect screenshot
We send every feedback twice and with incorrect screenshots. 

**Technical details:** `FeedbackViewController` takes a screenshot in the initializer and queues this feedback to be updated later on the next screen (this behavior is correct), but then it creates a `FeedbackSubtypeViewController` that is inherited from `FeedbackViewController` so it also creates new feedback with a new screenshot (when the screen is overlapped by first feedback controller and the map isn’t visible), this time it updates it with subtype. Since we don’t pass feedback UUID from the first screen to the second we end up sending two feedbacks (first with correct screenshot and no subtype, second with an incorrect screenshot but correct subtype).

**Mapbox Navigation SDK version:** 2.0.0

Here are examples of a screenshot from the `sdk_events` database:
image

azarovalex:
### Description
Fixes #3134 

### Implementation
`FeedbackSubtypeViewController` is inherited from the `FeedbackViewController` and we also record feedback in the initializer, so it's hard to record feedback only once without removing the inheritance.


Sources/MapboxNavigation/FeedbackSubtypeViewController.swift
```diff
@@ -14,9 +14,10 @@ class FeedbackSubtypeViewController: FeedbackViewController {
     /**
      Initialize a new FeedbackSubtypeViewController from a `NavigationEventsManager`.
      */
-    public init(eventsManager: NavigationEventsManager, feedbackType: FeedbackType) {
+    public init(eventsManager: NavigationEventsManager, feedbackType: FeedbackType, feedback: FeedbackEvent) {
```
1ec5, line 17: Let’s note this renamed public initializer in the changelog as a backwards-incompatible change.

1ec5 (approved):

Sources/MapboxNavigation/FeedbackSubtypeViewController.swift
azarovalex: Actually this VC controller is internal, not sure why this init was marked as public.

S2Ler: You can have a type to be internal in swift, and all methods marked as public will still be internal.
When you change the type from internal to the public, all public methods become public, and all internal methods stay internal. 
It might be non-obvious, as we see here. Maybe this is the result of refactoring a type from public to internal.

Sources/MapboxNavigation/FeedbackSubtypeViewController.swift
```diff
@@ -29,6 +30,10 @@ class FeedbackSubtypeViewController: FeedbackViewController {
 
         updateButtonTitle()
     }
+    
+    override func createFeedback() {
+        // do nothing, feedbackUUID should be passed to the initializer
```
bamx23, line 35: NIT: It's just `feedback`, not `feedbackUUID`

bamx23 (approved):

Sources/MapboxNavigation/FeedbackSubtypeViewController.swift
MaximAlien: If I remember correctly `FeedbackSubtypeViewController` should be exposed publicly, as it can be created from `FeedbackViewController`, which is public.

azarovalex: It's not exposed publicly right now for sure, and I don't feel like marking it as public because anyway it can't be used without `FeedbackViewController`. I will remove these misleading `public` keywords from functions.

Sources/MapboxNavigation/FeedbackSubtypeViewController.swift
azarovalex: Thanks!

Issue #3117, opened 2021-06-25
Final destination is not always visible during active guidance navigation.


MaximAlien:
**Mapbox Navigation SDK version:** `release-v2.0`

Final destination is not always visible during active guidance navigation. For example:

<img width="401" alt="Screen Shot 2021-06-25 at 11 57 48 AM" src="https://user-images.githubusercontent.com/1496498/123398907-e2922900-d558-11eb-81b2-2f2bfd06a192.png">

PR #3163, opened 2021-07-12
Fix issue, which was causing problems with adding final destination `PointAnnotation`.

MaximAlien:
### Description

Closing #3117.

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -126,9 +126,49 @@ open class NavigationMapView: UIView {
     public var pointAnnotationManager: PointAnnotationManager?
     
     /**
+     A `UserCourseView` used to indicate the user’s location and course on the map.
+     
+     The `UserCourseView`'s `UserCourseView.update(location:pitch:direction:animated:)` method is frequently called to ensure that its visual appearance matches the map’s camera.
+     */
+    public var userCourseView: UserCourseView = UserPuckCourseView(frame: CGRect(origin: .zero, size: 75.0)) {
+        didSet {
+            oldValue.removeFromSuperview()
+            installUserCourseView()
+        }
+    }
+    
+    /**
+     Specifies how the map displays the user’s current location, including the appearance and underlying implementation.
+     
+     By default, this property is set to `UserLocationStyle.courseView`, the bearing source is location course.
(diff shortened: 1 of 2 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -348,6 +358,16 @@ open class NavigationMapView: UIView {
         mapView.mapboxMap.onNext(.styleLoaded) { [weak self] _ in
```
S2Ler, line 358: Can the style be loaded twice during the lifetime of the NavigationMapView?

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -348,6 +358,16 @@ open class NavigationMapView: UIView {
...
             guard let self = self else { return }
             self.pointAnnotationManager = self.mapView.annotations.makePointAnnotationManager()
+            
+            if let finalDestinationAnnotation = self.finalDestinationAnnotation,
+               let pointAnnotationManager = self.pointAnnotationManager {
+                pointAnnotationManager.syncAnnotations([finalDestinationAnnotation])
```
S2Ler, line 364: Just curious, is it possible the SDK user adds it's own annotations to `pointAnnotationManager ` and we will rewrite them here? Maybe it makes sense to merge what `pointAnnotationManager.annotations` has and add a final destination annotation?

S2Ler (commented):
Thanks for fixing, noticed this issue as well. I have few comments to better understand how it all works.

Sources/MapboxNavigation/NavigationMapView.swift
MaximAlien: In this case no. As per Mapbox Maps docs for this method:
```
This will observe the next (and only the next) event of the specified type. After observation, the underlying subscriber will unsubscribe from the map or snapshotter.
```

This happens in `NavigationMapView.setupMapView(_:navigationCameraType:tileStoreLocation:)` method, which is called once.

Sources/MapboxNavigation/NavigationMapView.swift
S2Ler: Thanks for clarifications

S2Ler (approved):
lgtm

PR #3164, opened 2021-07-13
Fix getting latest descriptors

S2Ler:
The latest tile descriptor is resolved asynchronously in
`MBNNCacheHandle,` but there is no way to wait until the latest
descriptor is resolved in the cache handle. Until we have such
capabilities, we deploy a quick fix to let `MBNNCacheHandle` resolve
the latest descriptor by waiting for  X seconds until we ask it for
the descriptor. We fallback to this workaround only when `Navigator`
hasn't been created yet, or the navigator isn't in the appropriate state
for requested tile descriptor, otherwise we reuse Navigator's
cacheHandle.

The only change in TileStore.swift is adapting to the new async nature
of the `TilesetDescriptorFactory.getLatest` method.


Sources/MapboxCoreNavigation/TileStore.swift
```diff
@@ -19,42 +19,43 @@ extension TileStore {
...
                                               completion: @escaping ContainsCompletion) {
-        let descriptors = [TilesetDescriptorFactory.getLatest(forCacheLocation: cacheLocation)]
-        
-        __getAllTileRegions { expected in
-            if expected.isValue(), let regions = expected.value as? Array<TileRegion> {
-                let lock = NSLock()
-                var count = regions.count
-                var result: Bool? = true
+        TilesetDescriptorFactory.getLatest(forCacheLocation: cacheLocation,
+                                           completionQueue: .main) { latestTilesetDescriptor in
```
Udumft, line 23: Since now the callback has explicit queue definition we should either document that it will always be called on a main thread or allow customising it as well.

Sources/MapboxCoreNavigation/TileStore.swift
```diff
@@ -19,42 +19,43 @@ extension TileStore {
...
+                    completion(nil)
                 }
-            } else if expected.isError() {
-                completion(nil)
-            } else {
-                assertionFailure("Unexpected value or error: \(expected), expected: \(Array<TileRegion>.self)")
-                completion(nil)
             }
         }
     }
```
Udumft, line 61: `TileStore.tileRegionContainsLatestNavigationTiles` also uses `TilesetDescriptorFactory.getLatest` so I think we should refactor it the same way.

Sources/MapboxCoreNavigation/TilesetDescriptorFactory.swift
```diff
@@ -14,14 +14,40 @@ extension TilesetDescriptorFactory {
...
-     - parameter location: A `Location` where cache data is stored.
+     Gets TilesetDescriptor that corresponds to the latest available version of routing tiles.
+
+     It is intended to be used when creating off-line tile packs.
+
+     - Parameters:
+       - location: A `Location` where cache data is stored.
+       - completionQueue: A DispatchQueue on which the completion will be called.
+       - completion: A completion that will be used to pass the latest tileset descriptor.
+       - latestTilesetDescriptor: A tileset descriptor for the latest available version of routing tiles.
```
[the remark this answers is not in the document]
S2Ler: This was added so that it will show up in the Xcode documentation popup, but it turns out it is very buggy and doesn't work as expected. Removing this.

Sources/MapboxCoreNavigation/TilesetDescriptorFactory.swift
```diff
@@ -14,14 +14,40 @@ extension TilesetDescriptorFactory {
     }
 
     /**
-     Gets TilesetDescriptor which corresponds to the latest availble version of routing tiles.
-     
-     Intended for using when creating off-line tile packs.
-     
-     - parameter location: A `Location` where cache data is stored.
+     Gets TilesetDescriptor that corresponds to the latest available version of routing tiles.
```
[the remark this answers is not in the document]
S2Ler: It is not guaranteed. Given that changing this to async version is a breaking change, I agree that we shall refactor right away.

Sources/MapboxCoreNavigation/TileStore.swift
S2Ler: Well, actually a thread on which containsLatestNavigationTiles completion will be called stays the same as before. Here we are getting just the tile descriptor on the main thread.

Udumft (approved):
LGTM

Issue #3171, opened 2021-07-13
`CameraOptions.anchor` and `CameraOptions.padding` are not correctly updated.


MaximAlien:
**Mapbox Navigation SDK version:** `main`

### Steps to reproduce

 1. Open external CarPlay display in iOS Simulator.
 2. Open example application for CarPlay.
 3. Start active guidance navigation.
 4. Wait till steps dropdown menu opens, close it.
 5. Observe behavior.

### Expected behavior

After closing steps dropdown menu `CameraOptions.anchor` and `CameraOptions.padding` should adapt to the new content insets.

### Actual behavior

`CameraOptions` do not adapt to new content insets, which causes issue shown on screen recording.

Issue is a regression after landing https://github.com/mapbox/mapbox-navigation-ios/pull/3155/.

https://user-images.githubusercontent.com/1496498/125539855-88bca35e-e3b1-4618-b693-8aa7fb330919.mp4

PR #3172, opened 2021-07-13
Fix `CameraOptions` update issues.

MaximAlien:
### Description

Closing #3171.


Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -118,6 +117,11 @@ public class NavigationCameraStateTransition: CameraStateTransition {
               let padding = cameraOptions.padding else { return }
         
         let duration = 1.0
+        let centerCoordinatePixelThreshold = 2.0
+        let pitchThreshold = 1.0
+        let bearingThreshold = 1.0
```
1ec5, line 122: Thanks for factoring out these thresholds. It would be a bit clearer to name the variables more descriptively as “minimum”s and give them more specific types like `CLLocationDirection`.

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -172,26 +173,45 @@ public class NavigationCameraStateTransition: CameraStateTransition {
...
             }
             
             animatorPitch?.startAnimation()
         }
+        
+        if let animatorAnchor = animatorAnchor, animatorAnchor.isRunning {
+            animatorAnchor.stopAnimation()
+        }
+        
+        animatorAnchor = mapView.camera.makeAnimator(duration: duration, timingParameters: timingParameters) { (transition) in
```
1ec5, line 190: We should make this animator conditional on whether the value has actually changed, to avoid extraneous animations that could impact performance. It’s probably less likely to occur for the anchor than for the other parameters, but you never know if the application could trigger extra relayouts somehow.

1ec5 (approved):

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
MaximAlien: Makes sense, changed.

PR #3174, opened 2021-07-14
Fix events loss when closing the application

### LINKED ISSUE #3173: Prevent events loss on app force close
When a user force closes the app, we receive the `applicationWillTerminate` notification. At this moment we need to send all queued events to the server. The SDK dispatches this job on a background thread asynchronously, iOS see that the main thread is free and kills the app before the background job is ready. I conducted a few simple experiments to verify this behavior.

**Doesn't print anything:**
```
func applicationWillTerminate(_ notification: NSNotification) {
    DispatchQueue.global().async {
        Thread.sleep(forTimeInterval: 2)
        print("background job is done")
    }
}
```

**Prints the message:**
```
func applicationWillTerminate(_ notification: NSNotification) {
    DispatchQueue.global().async {
        Thread.sleep(forTimeInterval: 2)
        print("background job is done")
    }
    Thread.sleep(forTimeInterval: 3)
}
```

**TL;DR:** we need to block the main thread for a couple of seconds to make sure the network request will fire. According to the iOS documentation, we will have ~5 seconds, so let's block the thread for this duration.

azarovalex:
### Description
Fixes #3173.

See the description in the linked issue.

### DIFF
--- a/Sources/MapboxCoreNavigation/NavigationEventsManager.swift
+++ b/Sources/MapboxCoreNavigation/NavigationEventsManager.swift
@@ -346,7 +346,7 @@ open class NavigationEventsManager {
         }
     }
     
-    //MARK: - Session State Management
+    // MARK: - Session State Management
     @objc private func didChangeOrientation(_ notification: NSNotification) {
         sessionState?.reportChange(to: UIDevice.current.orientation)
     }
@@ -362,6 +362,7 @@ open class NavigationEventsManager {
         }
         
         sendOutstandingFeedbackEvents(forceAll: true)
+        Thread.sleep(forTimeInterval: 3) // block main thread until queued events are sent to the server
     }
     
     func reportReroute(progress: RouteProgress, proactive: Bool) {


Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -362,6 +362,7 @@ open class NavigationEventsManager {
         }
         
         sendOutstandingFeedbackEvents(forceAll: true)
+        Thread.sleep(forTimeInterval: 3) // block main thread until queued events are sent to the server
```
S2Ler, line 365: Are you sure that events library doesn't use main thread to do its work?

S2Ler: How difficult would it be to test this behaviour?

azarovalex: I've tested it, unfortunately, this fix doesn't seem to work 😿
I will close the PR, we will need some persistence solution from the Events library side.

PR #3176, opened 2021-07-14
Base XCTestCase

### LINKED ISSUE #3175: Simplify injecting shared mocks in tests
# Problem

In many of our `XCTestCase` subclasses we setup some mocked globals in `setUp` method. For example:
```swift
DirectionsCredentials.injectSharedToken(.mockedAccessToken)
```

These setups are required, or the tests won't run or crash.

Scattering the same code in many places is error-prone, time-consuming, cumbersome, and heavily against common architectural approaches. 

# Solution

Introduce an XCTestCase subclass which will be used in all the other test cases. 

This subclass will allow us to centralize common setup/teardown logic without repeating it in many places. 

@mapbox/navigation-ios

S2Ler:
Fixed #3175


Tests/MapboxNavigationTests/XCTestCase.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import XCTest
+import TestHelper
```
azarovalex, line 3: do we really need this import?

azarovalex (approved):
lgtm

Tests/MapboxNavigationTests/XCTestCase.swift
S2Ler: Probably not anymore, thanks for catching, I will check.

Udumft, line 3: Why is this import was added?

Udumft (approved):
LGTM with small question

Tests/MapboxNavigationTests/XCTestCase.swift
S2Ler: It used to be needed, leftover.

PR #3177, opened 2021-07-14
Replace JSON with Expressions in route duration annotation styling.

### LINKED ISSUE #2996: Route duration annotation styling should use Expressions, not JSON
#2873 implemented runtime styling of route duration annotations by hard-coding JSON-formatted strings, deserializing them as JSON-formatted dictionaries, and using a private map SDK API to pass these raw dictionaries into the style. Instead, we should initialize the map SDK’s type-safe `Expression` type, which is analogous to NSExpression in map SDK v6._x_ but more consistent with the style specification.

https://github.com/mapbox/mapbox-navigation-ios/blob/62b1622b8290b7a974ca9a0a7f39b5f986e5ae8d/Sources/MapboxNavigation/NavigationMapView.swift#L1054-L1064

ShanMa1991:
### Description
This pr is to fix #2996 

### Implementation
This pr is to remove the usage of `JSON` in runtime styling of route duration annotation, and replaced it with the map SDK’s type-safe `Expression` type.

### Screenshots or Gifs
With the `Expression` styling:
ExpressionStyling
With the `JSON` styling:
JSon

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -1237,66 +1237,32 @@ open class NavigationMapView: UIView {
         shapeLayer.textJustify = .constant(TextJustify.left)
         shapeLayer.symbolZOrder = .constant(SymbolZOrder.auto)
         shapeLayer.textFont = .constant(self.routeDurationAnnotationFontNames)
-
-        try style.addLayer(shapeLayer)
-
-        let symbolSortKeyString =
-        """
-        ["get", "sortOrder"]
-        """
-
-        if let expressionData = symbolSortKeyString.data(using: .utf8),
-           let expJSONObject = try? JSONSerialization.jsonObject(with: expressionData, options: []) {
-            let property = "symbol-sort-key"
-            try style.setLayerProperty(for: routeDurationAnnotationsLayerIdentifier,
-                                       property: property,
-                                       value: expJSONObject)
-        }
-
-        let expressionString =
-        """
-        [
-          "match",
-          ["get", "tailPosition"],
-          [0],
-          "bottom-left",
-          [1],
-          "bottom-right",
-          "center"
-        ]
-        """
-
-        if let expressionData = expressionString.data(using: .utf8),
-           let expJSONObject = try? JSONSerialization.jsonObject(with: expressionData, options: []) {
-            try style.setLayerProperty(for: routeDurationAnnotationsLayerIdentifier,
-                                       property: "icon-anchor",
-                                       value: expJSONObject)
-            try style.setLayerProperty(for: routeDurationAnnotationsLayerIdentifier,
-                                       property: "text-anchor",
-                                       value: expJSONObject)
-        }
-
-        let offsetExpressionString =
-            """
-        [
-          "match",
-          ["get", "tailPosition"],
-          [0],
-          ["literal", [0.5, -1]],
-          ["literal", [-0.5, -1]]
-        ]
-        """
         
-        if let expressionData = offsetExpressionString.data(using: .utf8),
-           let expJSONObject = try? JSONSerialization.jsonObject(with: expressionData, options: []) {
-            try style.setLayerProperty(for: routeDurationAnnotationsLayerIdentifier,
-                                       property: "icon-offset",
-                                       value: expJSONObject)
-            
-            try style.setLayerProperty(for: routeDurationAnnotationsLayerIdentifier,
-                                       property: "text-offset",
-                                       value: expJSONObject)
+        shapeLayer.symbolSortKey = .expression(Exp(.get) {
+            "sortOrder"
+        })
+
+        let anchorExpression = Exp(.match) {
+            Exp(.get) { "tailPosition" }
+            0
+            "bottom-left"
+            1
+            "bottom-right"
+            "center"
         }
+        shapeLayer.iconAnchor = .expression(anchorExpression)
+        shapeLayer.textAnchor = .expression(anchorExpression)
+        
+        let offsetExpression = Exp(.match) {
+            Exp(.get) { "tailPosition" }
+            0
+            Exp(.literal) { [0.5, -1.0] }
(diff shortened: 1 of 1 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -1239,64 +1239,30 @@ open class NavigationMapView: UIView {
...
-                                       property: "icon-offset",
-                                       value: expJSONObject)
-            
-            try style.setLayerProperty(for: routeDurationAnnotationsLayerIdentifier,
-                                       property: "text-offset",
-                                       value: expJSONObject)
+        shapeLayer.iconAnchor = .expression(anchorExpression)
+        shapeLayer.textAnchor = .expression(anchorExpression)
+        
+        let offsetExpression = Exp(.match) {
```
MaximAlien, line 1258: At a first glance looks good. But since I am (and probably others) are not too familiar with the new expressions syntax I wonder if it'd be possible to create unit-tests which could help to understand the new logic better and verify whether expressions syntax works as expected. For example:
- Create `NavigationMapView` instance.
- Add route duration annotation layer with certain features (e.g. which contains `tailPosition` feature).
- Verify that layer property, which is based on this feature is correct (e.g. `iconAnchor`).
- Change feature value (e.g. change `tailPosition` from `.left` to `.right`).
- Verify that layer property changes appropriately.

We can consider this as a tail work though.

MaximAlien (approved):

MaximAlien:
I think new expressions syntax brings regression. Please see the difference. 

Before:

Simulator Screen Shot - iPhone 12 mini - 2021-07-18 at 19 15 29

After:

Simulator Screen Shot - iPhone 12 mini - 2021-07-18 at 19 13 34

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestions, I've added the route duration tests for the `tailPosition ` feature and layer property.

ShanMa1991:
@MaximAlien Yes, I checked that even though the `Expression` and the `JSON` has the same meaning, but they results in different offset behavior, even though the point location and the `tailPosition ` are both random.  There's a fix in map SDK side,  maybe this pr could wait till the fix was merged.

zugaldia:
> There's a fix in map SDK side, maybe this pr could wait till the fix was merged.

Could you link the ticket or PR from maps here?

ShanMa1991:
> Could you link the ticket or PR from maps here?

There's a [ticket](https://github.com/mapbox/mapbox-maps-ios/issues/381) tracking this problem. I'll revise this pr and see how it works.

ShanMa1991:
> I think new expressions syntax brings regression. Please see the difference.
> 
Yes, I though it's caused by this [issue](https://github.com/mapbox/mapbox-maps-ios/issues/381#issuecomment-883699210). But now it's fixed in our side.

Tests/MapboxNavigationTests/RouteDurationTests.swift
```diff
@@ -0,0 +1,91 @@
...
+    override func setUp() {
+        super.setUp()
+    }
+    
+    override func tearDown() {
+        navigationMapView.removeRouteDurations()
+        super.tearDown()
+    }
+    
+    func setupRouteDurationsLayer(tailPosition: RouteDurationAnnotationTailPosition) throws {
```
MaximAlien, line 25: You should not be using own implementation of `RouteDurationAnnotationTailPosition` and `setupRouteDurationsLayer(tailPosition:)`. Test should be using similar implementations of `NavigationMapView`. If they're private it's fine to make them internal (default access control).

Tests/MapboxNavigationTests/RouteDurationTests.swift
```diff
@@ -0,0 +1,91 @@
...
+            Exp(.literal) { [-0.5, -1.0] }
+        }
+        shapeLayer.textOffset = .expression(offsetExpression)
+        
+        try style.addLayer(shapeLayer)
+    }
+    
+    func testTailPositionLeft() throws {
+        let style = navigationMapView.mapView.mapboxMap.style
+        try setupRouteDurationsLayer(tailPosition: .left)
```
MaximAlien, line 71: Instead of `try` you can use `XCTAssertNoThrow`.

Tests/MapboxNavigationTests/RouteDurationTests.swift
```diff
@@ -0,0 +1,91 @@
...
+            0
+            Exp(.literal) { [0.5, -1.0] }
+            Exp(.literal) { [-0.5, -1.0] }
+        }
+        shapeLayer.textOffset = .expression(offsetExpression)
+        
+        try style.addLayer(shapeLayer)
+    }
+    
+    func testTailPositionLeft() throws {
```
MaximAlien, line 69: Test should also involve verification that any change to the `Feature` property, which is used in source will affect layer outlook. E.g. change to `tailPosition` will change its `textAnchor` or `textOffset`. 

Think of test as an entity, which simplifies your life and does the job you'd do manually (e.g. to verify that certain feature works as expected), but a lot faster.

Tests/MapboxNavigationTests/RouteDurationTests.swift
ShanMa1991: Thanks for the suggestion, I'll replace it.

Tests/MapboxNavigationTests/RouteDurationTests.swift
ShanMa1991: Right now it seems unable to test the `Expression` from our side or map side as in the [ticket](https://github.com/mapbox/mapbox-maps-ios/issues/556).

ShanMa1991 (commented):
Removed the test for the route duration annotation Expression styling. There's a ticket in [mapSDK](https://github.com/mapbox/mapbox-maps-ios/issues/556) tracking this issue separately.

1ec5 (approved):

Issue #3178, opened 2021-07-14
`MapView.sourceIdentifiers(_:)` should use `tileSetIdentifier` parameter.


MaximAlien:
**Mapbox Navigation SDK version:** `main`

At this point `MapView.sourceIdentifiers(_:)` doesn't use `tileSetIdentifier` parameter which is intended to be used for sources filtering.

PR #3179, opened 2021-07-15
Add `tileSetIdentifier` parameter usage.

MaximAlien:
### Description

Closing #3178.

### DIFF
--- a/Sources/MapboxNavigation/MapView.swift
+++ b/Sources/MapboxNavigation/MapView.swift
@@ -69,14 +69,14 @@ extension MapView {
     func showsTileSet(withIdentifier tileSetIdentifier: String, layerIdentifier: String) -> Bool {
         let sourceIdentifiers = self.sourceIdentifiers(tileSetIdentifier)
         
-        for layerIdentifier in mapboxMap.style.allLayerIdentifiers.map({ $0.id }) {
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/MapView.swift
```diff
@@ -37,6 +27,39 @@ extension MapView {
...
+    }
+    
+    /**
+     Returns a set of source identifiers for tilesets that are or include the Mapbox Incidents source.
+     
+     - parameter tileSetIdentifier: Identifier of the tile set in the form `user.tileset`.
+     - returns: Set of source identifiers.
+     */
+    func sourceIdentifiers(_ tileSetIdentifier: String) -> Set<String> {
+        return Set(mapboxMap.style.allSourceIdentifiers.filter {
```
S2Ler, line 54: Do you know how big this array grows? May be it makes sense to optimise this method a little?

S2Ler (approved):
Thanks for writing new tests!

Sources/MapboxNavigation/MapView.swift
MaximAlien: Source count is usually not really high (less than hundred I guess) so pre-mature optimization might not be beneficial.

PR #3182, opened 2021-07-16
RouteResponse usage refactoring

Udumft:
### Description
This PR replaces usage of `Route` + Route index in favor of it's original `RouteResponse`. It also removes usage of `Route.routeIdentifier`, as this is a redundant property should be [removed too](https://github.com/mapbox/mapbox-directions-swift/pull/562).

### Implementation
Most of the changes is just a replacing of `IndexedRoute` with `RouteResponse`. But it also forced updating some related entities like `CPTrip.userInfo`, removing `RouteProgress.routeIndex or updating `SessionState` with dedicated `routeIdentifier`. There are many public API changes.


S2Ler:
This PR will clash with #3159

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -13,9 +13,9 @@ public protocol RouterDataSource: AnyObject {
 }
 
 /**
- A route and its index in a `RouteResponse` that sorts routes from most optimal to least optimal.
+ A `RouteResponse` object that sorts routes from most optimal to least optimal and selected route index in it.
  */
-public typealias IndexedRoute = (Route, Int)
+public typealias IndexedRouteResponse = (routeResponse: RouteResponse, routeIndex: Int)
```
S2Ler, line 18: @mapbox/navigation-ios  Does anybody know the reason for using type alias here instead of a full featured type like struct?

S2Ler: I think it is a good opportunity to refactor this type in a struct unless there is a strong reason not to.

Udumft: Done

Sources/MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -62,7 +62,7 @@ public protocol CarPlayManagerDelegate: AnyObject, UnimplementedLogging {
      - parameter desiredSimulationMode: The desired simulation mode to use.
      - returns: A navigation service that manages location updates along `route`.
      */
-    func carPlayManager(_ carPlayManager: CarPlayManager, navigationServiceAlong route: Route, routeIndex: Int, routeOptions: RouteOptions, desiredSimulationMode: SimulationMode) -> NavigationService
+    func carPlayManager(_ carPlayManager: CarPlayManager, navigationServiceAlong routeResponse: RouteResponse, routeIndex: Int, routeOptions: RouteOptions, desiredSimulationMode: SimulationMode) -> NavigationService
```
S2Ler, line 65: Documentation is outdated.

S2Ler: I have a strange feeling reading `navigationServiceAlong routeResponse`. How is that possible to navigate along response? It might be a consequence of the wrong abstraction or a better naming can be found.

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -57,18 +57,18 @@ public protocol NavigationService: CLLocationManagerDelegate, RouterDataSource,
...
     
     /**
-     The route along which the user is expected to travel, plus its index in the `RouteResponse`, if applicable.
+     The route along which the user is expected to travel.
 
      If you want to update the route, use `Router.updateRoute(with:routeOptions:)` method from `router`.
      */
-    var indexedRoute: IndexedRoute { get }
-
+    var route: Route? { get }
```
S2Ler, line 64: Why this because optional? Seems wrong that navigation service can operation without a route.

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -217,17 +217,17 @@ public class MapboxNavigationService: NSObject, NavigationService {
...
      Intializes a new `NavigationService`. Useful convienence initalizer for OBJ-C users, for when you just want to set up a service without customizing anything.
      
-     - parameter route: The route to follow.
-     - parameter routeIndex: The index of the route within the original `RouteController` object.
+     - parameter routeResponse: `RouteResponse` object, containing selection of routes to follow.
+     - parameter routeIndex: The index of the route within the original `RouteResponse` object.
      */
-    convenience init(route: Route, routeIndex: Int, routeOptions options: RouteOptions) {
-        self.init(route: route, routeIndex: routeIndex, routeOptions: options, directions: nil, locationSource: nil, eventsManagerType: nil)
+    convenience init(routeResponse: RouteResponse, routeIndex: Int, routeOptions options: RouteOptions) {
```
S2Ler, line 223: #2589 mentions:
> A convenience initializer for MapboxNavigationService can wrap a Route in a RouteResponse, and a convenience initializer for NavigationViewController can wrap a Route in a RouteResponse and MapboxNavigationService.

Shall we add these convenience inits in this PR? Or what's the reason of not including them?

Example/CustomViewController.swift
```diff
@@ -16,7 +16,8 @@ class CustomViewController: UIViewController {
     
     var simulateLocation = false
 
-    var userIndexedRoute: IndexedRoute?
+    var userRouteResponse: RouteResponse?
```
S2Ler, line 19: Maybe `var indexedUserRouteResponse: IndexedRouteResponse?` can be used?

Sources/MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -350,8 +345,7 @@ open class RouteProgress: Codable {
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
         
-        try container.encode(indexedRoute.0, forKey: .indexedRoute)
-        try container.encode(indexedRoute.1, forKey: .indexedRouteIndex)
```
S2Ler, line 354: `indexedRouteIndex` still appears in `CodingKeys`.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -35,51 +34,48 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
     /**
      A `Route` object constructed by [MapboxDirections](https://docs.mapbox.com/ios/api/directions/).
      */
     public var route: Route? {
-        get {
-            return navigationService.indexedRoute.0
-        }
-        
-        set {
-            _route = newValue
```
S2Ler, line 60: I think this code was added to support Storyboards. @MaximAlien please review.

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -57,18 +57,18 @@ public protocol NavigationService: CLLocationManagerDelegate, RouterDataSource,
...
      */
-    var indexedRoute: IndexedRoute { get }
-
+    var route: Route? { get }
+    
     /**
-     The route along which the user is expected to travel.
-
+     The `RouteResponse` object containing active route, plus its index in this `RouteResponse`, if applicable.
+
```
S2Ler, line 68: nit: sneaky empty spaces :) I would love swift-format integrated so that we will never see such things ever again.

S2Ler:
@1ec5 Your review is required since you created the original ticket.

Sources/MapboxCoreNavigation/NavigationService.swift
Udumft: My motivation not to add these was that all of the routes are supposed to be a result of API calling, and thus are always wrapped in `RouteResponse`. If, for some uncommon reason, user is creating a `Route` object differently, he should take care about correctly [wrapping it](https://github.com/mapbox/mapbox-directions-swift/blob/main/Sources/MapboxDirections/RouteResponse.swift#L41), providing other required fields like options and identifier.

Example/CustomViewController.swift
Udumft: Good catch. Fixed

Sources/MapboxCoreNavigation/NavigationService.swift
Udumft: That's right, it's a fallout of early implementation version

CHANGELOG.md
```diff
@@ -126,6 +126,7 @@
...
 * Supported feedbacks during passive navigation. Set `PassiveLocationManager` as a data source for `NavigationEventsManager` and then use this class to create and send user feedbacks. `FeedbackViewController` provides drop-in feedback screen and can be used during both passive and active navigation. ([#3122](https://github.com/mapbox/mapbox-navigation-ios/pull/3122))
 * `NavigationViewController.indexedRoute`, `NavigationService.indexedRoute` and `Router.indexedRoute` properties are readonly now. Use dedicated `Router.updateRoute(with:routeOptions:)` method to update the route. ([#3159](https://github.com/mapbox/mapbox-navigation-ios/pull/#3159))
+* Replaced usage of `Route` + `routeIndex` in favor of `RouteResponse`. This affects removing `NavigationViewController.indexedRoute`, `NavigationService.indexedRoute`, `Router.indexedRoute` and replacing it with corresponding `indexedRouteResponse` or `routeResponse` properties. `RouteProgress.indexedRoute` was removed completely ([#3182](https://github.com/mapbox/mapbox-navigation-ios/pull/3182)).
```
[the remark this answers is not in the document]
Udumft: Thanks for the suggestion

S2Ler (commented):
I think the following should be done before merge:
- [ ] Verify that Storyboards integration isn't broken cc @MaximAlien 
- [ ] Review from @1ec5

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -13,9 +13,17 @@ public protocol RouterDataSource: AnyObject {
 }
 
 /**
- A route and its index in a `RouteResponse` that sorts routes from most optimal to least optimal.
+ A `RouteResponse` object that sorts routes from most optimal to least optimal and selected route index in it.
  */
-public typealias IndexedRoute = (Route, Int)
+public struct IndexedRouteResponse {
+    public let routeResponse: RouteResponse
+    public let routeIndex: Int
```
1ec5, line 20: Remember to add documentation comments about these properties and also the initializer below so that jazzy recognizes them.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -36,29 +44,29 @@ public protocol Router: CLLocationManagerDelegate {
...
      
-     - parameter route: The route to follow.
+     - parameter routeResponse: `RouteResponse` object, containing selection of routes to follow.
      - parameter routeIndex: The index of the route within the original `RouteResponse` object.
      - parameter directions: The Directions object that created `route`.
      - parameter source: The data source for the RouteController.
      - parameter tileStoreLocation: Configuration of `TileStore` location, where Navigation tiles are stored.
      */
-    init(along route: Route, routeIndex: Int, options: RouteOptions, directions: Directions, dataSource source: RouterDataSource, tileStoreLocation: TileStoreConfiguration.Location)
+    init(along routeResponse: RouteResponse, routeIndex: Int, options: RouteOptions, directions: Directions, dataSource source: RouterDataSource, tileStoreLocation: TileStoreConfiguration.Location)
```
1ec5, line 53: It made sense to say “`along` `route`” in a sentence, but “`along` `routeResponse`” makes less sense. Maybe “`alongRouteAtIndex`, `in` `routeResponse`”, “`routeResponse`, `routeIndex`”, or simply “`indexedRouteResponse`”.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -36,29 +44,29 @@ public protocol Router: CLLocationManagerDelegate {
...
-
     /// The route along which the user is expected to travel.
     ///
     /// You can update the route using `Router.updateRoute(with:routeOptions:)`.
     var route: Route { get }
     
+    /// The `RouteResponse` containing the route along which the user is expected to travel, plus its index in this `RouteResponse`, if applicable.
+    ///
+    /// If you want to update the route use `Router.updateRoute(with:routeOptions:)` method.
+    var indexedRouteResponse: IndexedRouteResponse { get }
```
1ec5, line 68: If it isn’t valid to set `indexedRouteResponse` directly, let’s make it publicly read-only.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -125,7 +133,7 @@ protocol InternalRouter: AnyObject {
     
     var lastRerouteLocation: CLLocation? { get set }
     
-    func setRoute(route: Route, routeIndex: Int, proactive: Bool)
+    func setRoute(route: Route, proactive: Bool, routeOptions: RouteOptions?)
```
1ec5, line 136: While we’re at it, the first argument doesn’t need a label.

Sources/MapboxNavigation/CPTrip.swift
```diff
@@ -3,8 +3,8 @@ import CarPlay
 
 @available(iOS 12.0, *)
 extension CPTrip {
-    convenience init(routes: [Route], routeOptions: RouteOptions, waypoints: [Waypoint]) {
-        let routeChoices = routes.enumerated().map { (routeIndex, route) -> CPRouteChoice in
+    convenience init(routeResponse: RouteResponse, routeOptions: RouteOptions, waypoints: [Waypoint]) {
```
1ec5, line 6: RouteResponse has an `options` property that should reflect the options in the request. Do we need to keep a separate `routeOptions` argument around, or can we assume it’s always the same options?

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -217,17 +217,17 @@ public class MapboxNavigationService: NSObject, NavigationService {
     /**
      Intializes a new `NavigationService`. Useful convienence initalizer for OBJ-C users, for when you just want to set up a service without customizing anything.
      
-     - parameter route: The route to follow.
-     - parameter routeIndex: The index of the route within the original `RouteController` object.
+     - parameter routeResponse: `RouteResponse` object, containing selection of routes to follow.
+     - parameter routeIndex: The index of the route within the original `RouteResponse` object.
```
1ec5, line 221: `routeOptions` isn’t documented, but I wonder if we still need it, because RouteResponse has an `options` property.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -347,35 +343,35 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
      Initializes a `NavigationViewController` with the given route and navigation service.
      
-     - parameter route: The route to navigate along.
+     - parameter routeResponse: `RouteResponse` object, containing selection of routes to follow.
      - parameter routeIndex: The index of the route within the original `RouteResponse` object.
      - parameter routeOptions: the options object used to generate the route.
      - parameter navigationService: The navigation service that manages navigation along the route.
      */
-    convenience init(route: Route, routeIndex: Int, routeOptions: RouteOptions, navigationService service: NavigationService) {
+    convenience init(routeResponse: RouteResponse, routeIndex: Int, routeOptions: RouteOptions, navigationService service: NavigationService) {
```
1ec5, line 372: Now that NavigationService contains the full `indexedRouteResponse`, as well as the route options, do we need the application to pass in the route response, route index, and route options separately here, or is it sufficient to pass in just the navigation service?

1ec5 (approved):
Meant to attach this approval to https://github.com/mapbox/mapbox-navigation-ios/pull/3182#pullrequestreview-713818693:

I think it would be quite beneficial to remove some redundancy, now that RouteResponse encapsulates more information, but this PR is mergeable as is once the few bits of missing documentation are added.

MaximAlien (commented):
This PR brings quite a lot of public API changes. We'll need to create another PR in https://github.com/mapbox/mapbox-navigation-ios-examples to address compilation issues.

MaximAlien:
We'll have to also revise documentation after landing current PR: e.g. in https://docs.mapbox.com/ios/beta/navigation `NavigationViewController` and related APIs are mentioned quite often.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -35,51 +34,48 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
     
-    var _route: Route?
+    var _routeIndex: Int?
+    var _routeResponse: RouteResponse?
+    var _routeOptions: RouteOptions?
     
     /**
      A `Route` object constructed by [MapboxDirections](https://docs.mapbox.com/ios/api/directions/).
      */
     public var route: Route? {
```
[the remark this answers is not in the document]
Udumft: These properties were made read-only to [introduce single entry-point method for changing the route](https://github.com/mapbox/mapbox-navigation-ios/pull/3159): `updateRoute(with:routeOptions:)`. Not sure what to do with storyboard scenario...

Sources/MapboxNavigation/CPTrip.swift
Udumft: It is expected to be same options.
Previously, we limited this `CPTrip.userInfo` to contain exactly `RouteOptions` object. It is not documented anywhere or, exposed publicly (or at least I could not find any public promise made around it). Technically, we could use `routeResponse.options` and cover both `RouteOptions` and `MatchOptions` cases. 
It seems rather safe change to me, so I'll update it if there's no objections.

Udumft: Looks like same applies to `waypoints` argument. It is extracted from the route response

Tests/CocoaPodsTest/PodInstall/Podfile.lock
```diff
@@ -68,4 +68,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ce6227c2aeeca0d51c521648629d4f06ae38a439
 
-COCOAPODS: 1.10.1
+COCOAPODS: 1.10.0
```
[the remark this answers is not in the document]
Udumft: My bad. Will update CocoaPods

Sources/MapboxCoreNavigation/NavigationService.swift
Udumft: This argument serves as specification that it handles `RouteOptions` instead of possible `MatchOptions`.

Sources/MapboxNavigation/NavigationViewController.swift
Udumft: Do you mean a convenience init which would get these values from the custom navigation service provided? Sounds good to me.
But I think we still need to preserve this initializer. Otherwise we'll force users to explicitly instantiate the navigation service even if they are happy with the default one.

Sources/MapboxNavigation/NavigationViewController.swift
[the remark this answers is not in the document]
1ec5: #3159 was primarily concerned with changing the route during a trip, but the storyboard code path is about setting up the trip in the first place. Can we introduce something specific to storyboards, perhaps even with “forStoryboard” in the name?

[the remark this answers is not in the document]
S2Ler: Would be good to have a special method which setups vc after view controller. This method will set appropriate private variables (could be even grouped in a struct) which will be read in the code which setups vc in storyboards case. 

When I read this code I found it confusing why route variable has such strange get/set.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -387,9 +385,37 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     
     // MARK: - Setting-up methods
     
+    /**
+     Updates key settings before loading the view.
+     
+     This method basically re-runs the setup which takes place in `init`. It could be useful if some of the attribtues have changed before `NavigationViewController` did load it's view, or if you did not have access to initializing logic. For example, as a part of `UIStoryboardSegue` configuration.
```
MaximAlien, line 391:
```suggestion
     This method basically re-runs the setup which takes place in `init`. It could be useful if some of the attributes have changed before `NavigationViewController` did load it's view, or if you did not have access to initializing logic. For example, as a part of `UIStoryboardSegue` configuration.
```

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -387,9 +385,37 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
...
+     
+     - parameter routeResponse: `RouteResponse` object, containing selection of routes to follow.
+     - parameter routeIndex: The index of the route within the original `RouteResponse` object.
+     - parameter routeOptions: The route options used to get the route.
+     - parameter navigationOptions: The navigation options to use for the navigation session.
+     - returns `True` if setup was successful, `False` if `view` is already loaded and settings did not apply.
+     */
+    public func prepareViewLoading(routeResponse: RouteResponse, routeIndex: Int, routeOptions: RouteOptions, navigationOptions: NavigationOptions? = nil) -> Bool {
+        guard !isViewLoaded else {
```
MaximAlien, line 400: Why do we need to perform check to see if view was really loaded? I recall that in previous implementation in https://github.com/mapbox/mapbox-navigation-ios/pull/2974 it wasn't needed. Just want to make sure that this will not cause any side effects.

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -36,7 +36,7 @@ class MapboxCoreNavigationTests: TestCase {
     }
     
     func testNavigationNotificationsInfoDict() {
-        navigation = MapboxNavigationService(route: route,
+        navigation = MapboxNavigationService(routeResponse: response,
```
MaximAlien, line 39: It'd be great to create a follow-up ticket to update documentation everywhere. In scope of this PR files like `README.md` and `cover.md` can be updated as they show how to use `NavigationViewController`.

Sources/MapboxNavigation/NavigationViewController.swift
Udumft: This check is added since updating `routeResponse`, `routeIndex` or `routeOptions` will not have effect on the contents `NavigationViewController` if view is already loaded. In this case `view` - is the Navigation View, and these properties are only stored in `NavigationViewController` to initialize the NavigationView. So, this check lets user know if this update actually happened or not. Previous implementation did silent update of a local property, but that did not affect the NavigationView if it was already loaded.

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
Udumft: I missed that, thanks

Sources/MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -189,8 +185,8 @@ open class RouteProgress: Codable {
      - parameter routeIndex: The index of the route in the `RouteResponse`. By default, the route is the first route in the response.
      - parameter legIndex: Zero-based index indicating the current leg the user is on.
      */
-    public init(route: Route, routeIndex: Int, options: RouteOptions, legIndex: Int = 0, spokenInstructionIndex: Int = 0) {
```
MaximAlien, line 192: We need to update documentation here as well. `routeIndex` is no longer present. `options` and `spokenInstructionIndex` will have to be added as well.

MaximAlien (approved):
This PR changes so many public APIs that it's hard to see what can be wrong. On high level seems to look good to me. We should review all docs and update examples to make sure that there is not major regression and everything is up-to-date.

Udumft:
Ticket to update examples project : mapbox/mapbox-navigation-ios-examples/issues/135

Sources/MapboxCoreNavigation/NavigationService.swift
dmiluski: Hi @Udumft ,

I have a use case in which I would prefer injecting a Route and am wondering if I now need to create a convenience init to support? It seems strange to move from a strong required Route, to a model of optionals?

https://github.com/mapbox/straightaway-ios/blob/main/Straightaway/Modules/Map/ContinuousNavigationViewController.swift#L85

Given our UI/ViewModels are intentionally separate from network models, exposing a network response object seemed heavy. 

For additional context, in an upcoming feature change, we will be displaying multiple direction options. Which we will display, and begin navigation upon selection of a given route. 



cc: @irtemed88

dmiluski: Hi @Udumft ,

I was seeing if I could get your suggestion for how I could approach this? I took a first stab at the upgrade migration.
https://github.com/mapbox/straightaway-ios/pull/1047

PR #3183, opened 2021-07-16
Remove MapboxGeocoder.swift dependency

### LINKED ISSUE #2285: Remove optional dependency on MapboxGeocoder.swift for search on CarPlay

1ec5:
MapboxGeocoder.swift is no longer a dependency of MapboxCoreNavigation or MapboxCoreNavigationTests. It is only required for building the CarPlay example application.

Fixes #2285.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/2999#discussion_r662435612
/cc @mapbox/navigation-ios


MapboxNavigation.podspec
```diff
@@ -47,7 +47,6 @@ Pod::Spec.new do |s|
   s.dependency "MapboxMaps", "10.0.0-rc.3"
   s.dependency "Solar", "~> 2.1"
   s.dependency "MapboxSpeech-pre", "2.0.0-alpha.1"
-  s.dependency "MapboxGeocoder.swift", "~> 0.10.0"
```
MaximAlien, line 50: Ah, thanks for looking into this. Now I realized what you meant by removing `MapboxGeocoder`.

MaximAlien (approved):

jill-cardamon (approved):

PR #3186, opened 2021-07-19
NavigationMapView injection

### LINKED ISSUE #3165: Allow injecting NavigationMapView into NavigationViewController
It should be possible for an application to specify an arbitrary NavigationMapView when initializing a NavigationOptions; NavigationViewController should use this NavigationMapView instead of the built-in one, similar to how top and bottom banners can be customized.

This is a prerequisite for providing a route preview view controller as part of the navigation SDK (#808), since that entry point must remain optional. (Some use cases, such as logistics, don’t necessarily require the user to preview a route.) Exposing map-related components in #2953 would make it possible for the application to get some of the benefits of injecting a map view by creating a custom view controller scaffolding instead of NavigationViewController.

Ideally, NavigationViewController should continue using the injected map view without a jarring transition.

/cc @mapbox/navigation-ios

Udumft:
### Description
Resolves #3165 

### Implementation
Added a property to `NavigationOptions` which forwards input `NavigationMapView` instance to `NavigationView` (and in result in `NavigationViewController`. This instance is used instead of originally created `NavigationMapView` if provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
 * Added the `NavigationMapView.userLocationStyle` property to customize how the user’s current location is displayed on the map. Set this property to `UserLocationStyle.puck2D(configuration:)` or `UserLocationStyle.puck3D(configuration:)` to use the [location indicator layer](https://docs.mapbox.com/ios/maps/api/10.0.0-beta.21/Enums/LayerType.html#/s:10MapboxMaps9LayerTypeO17locationIndicatoryA2CmF) powered by the Mapbox Maps SDK instead of the default view-backed implementation specified by the `NavigationMapView.userCourseView` property. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
 * If you need to customize the appearance of the user location indicator, you can subclass `UserPuckCourseView` and `UserHaloCourseView` as a starting point. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
 * Fixed an issue where route line disappears when changing `MapView` style. ([#3136](https://github.com/mapbox/mapbox-navigation-ios/pull/3136))
-* Added `NavigationOptions.navigationMapView` property to allow customization or reusing options for `NavigationViewController.navigationMapView` ([#3186](https://github.com/mapbox/mapbox-navigation-ios/pull/3186)).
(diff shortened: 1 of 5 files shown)


bamx23 (approved):

S2Ler (approved):

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
...
 * If you need to customize the appearance of the user location indicator, you can subclass `UserPuckCourseView` and `UserHaloCourseView` as a starting point. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
 * Fixed an issue where route line disappears when changing `MapView` style. ([#3136](https://github.com/mapbox/mapbox-navigation-ios/pull/3136))
+* Added `NavigationOptions.navigationMapView` property to allow customization or reusing options for `NavigationViewController.navigationMapView` ([#3186](https://github.com/mapbox/mapbox-navigation-ios/pull/3186)).
```
S2Ler, line 42: may be it is me, but I can't translate this changelog entry.

Sources/MapboxNavigation/NavigationOptions.swift
```diff
@@ -52,6 +52,12 @@ open class NavigationOptions: NavigationCustomizable {
      */
     open var tileStoreConfiguration: TileStoreConfiguration = .default
     
+    /**
+     Custom `NavigationMapView` instance to be embedded in navigation UI.
+     
+     If set to `nil` - new `NavigationMapView` instance will be created. When setting a custom instance, `NavigationView` will update it's delegate and camera's `viewportDatasource` to function correctly. You may want to use this property for customization or optimization purposes.
```
S2Ler, line 58:
```suggestion
     If set to `nil`, a default `NavigationMapView` instance will be created. When a custom instance is set, `NavigationView` will update its delegate and camera's `viewportDatasource` to function correctly. You may want to use this property for customization or optimization purposes.
```

CHANGELOG.md
Udumft: I'll try to make it more clear...

Sources/MapboxNavigation/NavigationOptions.swift
Udumft: Thanks for the correction.

zacharyblank:
giphy

cc @irtemed88

PR #3192, opened 2021-07-20
Copy RouteOptions as its most specific type

### LINKED ISSUE #3191: RouteOptions subclass type is erased when rerouting
We allow and sometimes encourage developers to subclass NavigationRouteOptions to fine-tune the Directions API requests that come from the application. For example, mapbox/mapbox-navigation-ios-examples#91 illustrates how to use beta query parameters that the MapboxDirections library doesn’t formally support yet. When either RouteController or LegacyRouteController reroutes the user, the SDK stops using the developer’s subclass, so any customizations implemented in a `urlQueryItems` override get blown away.

## Diagnosis

`RouteOptions.without(waypoint:)` and `RouteProgress.reroutingOptions(with:)` effectively replace the subclass instance with a RouteOptions instance (not even a NavigationRouteOptions instance), so that it can modify the set of waypoints without affecting the original options object.

https://github.com/mapbox/mapbox-navigation-ios/blob/e308e0ce2db6f7382db08698145256d4b7891f10/Sources/MapboxCoreNavigation/Router.swift#L246-L248 https://github.com/mapbox/mapbox-navigation-ios/blob/e308e0ce2db6f7382db08698145256d4b7891f10/Sources/MapboxCoreNavigation/RouteProgress.swift#L249-L251 https://github.com/mapbox/mapbox-navigation-ios/blob/e308e0ce2db6f7382db08698145256d4b7891f10/Sources/MapboxCoreNavigation/RouteOptions.swift#L32-L40

This is an awkward mix of Objective-C and Swift paradigms for dealing with value types: traditionally, in Objective-C, RouteOptions would be immutable and a separate MutableRouteOptions class would have mutable properties, but that complicates subclassing. In Swift, a struct would be the correct way to implement the pass-by-value semantics we’re going for: mapbox/mapbox-directions-swift#564.

In practice, RouteOptions itself has mutable properties, so as a workaround, 70bd68f38f3c1765715cf1c56139056d67a7366b for #2275 implemented NSCopying on RouteOptions, hard-coding the RouteOptions type in the implementation of `copy(with:)`. This means any subclass of RouteOptions, including NavigationRouteOptions, would need to override `copy(with:)` to preserve any customizations after rerouting. NavigationRouteOptions happens to be unaffected, because all it does is set a few good default values in its initializer – values that RouteOptions’ Codable implementation would naturally preserve when round-tripping through JSON.

## Next steps

While we await the bright future after mapbox/mapbox-directions-swift#564 is fixed, `RouteOptions.copy(with:)` should dynamically decode the current object’s type rather than hard-coding RouteOptions per se. Documentation about RouteOptions and NavigationRouteOptions should include a subclassing note that discusses overriding `init(from:)` and `encode(to:)`. The modified `RouteOptions.copy(with:)` would only work to the extent that RouteOptions subclasses are resilient to being round-tripped through JSON.

/cc @mapbox/navigation-ios

1ec5:
Fixed an issue where copying an instance of a RouteOptions subclass resulted in an instance of RouteOptions itself. Now calling `RouteOptions.copy()` always returns the same specific type as the original object.

Along for the ride, `RouteOptions.without(waypoint:)` has been renamed `RouteOptions.without(_:)`, because the argument label is redundant to the argument type at the point of use.

Fixes #3191.

/cc @mapbox/navigation-ios @OttyLab

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,7 @@
 * Supported feedbacks during passive navigation. Set `PassiveLocationManager` as a data source for `NavigationEventsManager` and then use this class to create and send user feedbacks. `FeedbackViewController` provides drop-in feedback screen and can be used during both passive and active navigation. ([#3122](https://github.com/mapbox/mapbox-navigation-ios/pull/3122))
 * `NavigationViewController.indexedRoute`, `NavigationService.indexedRoute` and `Router.indexedRoute` properties are readonly now. Use dedicated `Router.updateRoute(with:routeOptions:)` method to update the route. ([#3159](https://github.com/mapbox/mapbox-navigation-ios/pull/#3159))
 * Fixed an issue where a subclass of `NavigationRouteOptions` would turn into an ordinary `RouteOptions` when rerouting the user. ([#3192](https://github.com/mapbox/mapbox-navigation-ios/pull/3192))
+* Renamed `RouteOptions.without(waypoint:)` to `RouteOptions.without(_:)`. ([#3192](https://github.com/mapbox/mapbox-navigation-ios/pull/3192))
 
 ## v1.4.1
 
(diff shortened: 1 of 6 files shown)


MaximAlien (approved):

1ec5:
1cf370bff636075b4ef507a270baeef3aea0fa9b (3b55a5affa716e0dacc4d6f90cdc5edbb6be744e) has been cherry-picked to the release-v1.2 branch as 62472d5bbef4bb2c7c530dc2299860d189bda342 and to the release-v1.4 branch as c9ed696fb686df028814704129b9b68e11c3aebd.

PR #3195, opened 2021-07-22
Rerouting after arrival

S2Ler:
I've commented some code changes inline, please follow them.

Haven't added for moving puck, will cut a separate ticket, or move this task to GA? @1ec5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@
 * Renamed `PassiveLocationManager.startUpdatingLocation(completionHandler:)` to `PassiveLocationProvider.startUpdatingLocation()`. This method now runs synchronously like `CLLocationManager.startUpdatingLocation()`. ([#2823](https://github.com/mapbox/mapbox-navigation-ios/pull/2823))
 * Added the `RouteController.startRecordingHistory()`, `RouteController.stopRecordingHistory(completionHandler:)`, `PassiveLocationManager.startRecordingHistory()`, and `PassiveLocationManager.stopRecordingHistory(completionHandler:)` methods for recording details about a trip for debugging purposes. ([#3157](https://github.com/mapbox/mapbox-navigation-ios/pull/3157))
 * Renamed `RouterDataSource.locationProvider` and `EventsManagerDataSource.locationProvider` properties to `RouterDataSource.locationManagerType` and `EventsManagerDataSource.locationManagerType` respectively. ([#3199](https://github.com/mapbox/mapbox-navigation-ios/pull/3199))
+* Adds `NavigationViewControllerDelegate.navigationViewController(_:shouldPreventReroutesWhenArrivingAt:)`, which is called each time the user arrives at a waypoint. By default, this method returns true and prevents rerouting upon arriving. ([#3195](https://github.com/mapbox/mapbox-navigation-ios/pull/3195))
+* Fixes an issue where the user would be rerouted even if `NavigationViewControllerDelegate.navigationViewController(_:shouldRerouteFrom:)` returned `false`. To implement reroute after arrival behavior, return `false` in both this method and `NavigationViewControllerDelegate.navigationViewController(_:shouldPreventReroutesWhenArrivingAt:)`, and set `NavigationViewController.showsEndOfRouteFeedback` to `false`. ([#3195](https://github.com/mapbox/mapbox-navigation-ios/pull/3195))
 
 ### Electronic horizon
 
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -274,8 +274,11 @@ open class RouteController: NSObject {
         if willReroute {
             reroute(from: location, along: routeProgress)
         }
-        // Check for faster route proactively (if reroutesProactively is enabled)
-        refreshAndCheckForFasterRoute(from: location, routeProgress: routeProgress)
+
+        if status.routeState != .complete {
+            // Check for faster route proactively (if reroutesProactively is enabled)
+            refreshAndCheckForFasterRoute(from: location, routeProgress: routeProgress)
+        }
     }
     
     private func subscribeNotifications() {
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -43,13 +43,10 @@ open class RouteController: NSObject {
             resetObservation(for: _routeProgress)
         }
         didSet {
-            movementsAwayFromRoute = 0
             updateNavigator(with: _routeProgress)
             updateObservation(for: _routeProgress)
         }
     }
-    
-    var movementsAwayFromRoute = 0
     
     var routeTask: URLSessionDataTask?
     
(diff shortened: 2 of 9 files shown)


Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -274,8 +271,11 @@ open class RouteController: NSObject {
         if willReroute {
             reroute(from: location, along: routeProgress)
         }
-        // Check for faster route proactively (if reroutesProactively is enabled)
-        refreshAndCheckForFasterRoute(from: location, routeProgress: routeProgress)
+
+        if status.routeState != .complete {
```
S2Ler, line 275: Don't need to check for faster route if we arrived.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -522,8 +522,32 @@ extension RouteController: Router {
         
         // If we still wait for the first status from NavNative, there is no need to reroute
         guard let status = status ?? navigator.getStatus() else { return true }
-        let offRoute = status.routeState == .offRoute || status.routeState == .invalid
-        return !offRoute
+
+        /// NavNative doesn't support reroutes after arrival.
+        /// The code below is a port of logic from LegacyRouteController
+        /// This should be removed once NavNative adds support for reroutes after arrival. 
+        if status.routeState == .complete {
```
S2Ler, line 529: If route state is changed to completed, NavNative won't detect off routes anymore, so we drop to temporary and straightforward logic. If the user moves far enough from the destination point, trigger the reroute if the delegate allows that.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -158,7 +158,15 @@ extension NavigationMapView {
         /**
          Take the remaining distance from the upcoming point on the route and extends it by the exact position of the puck.
          */
-        let remainingDistance = traveledIndex.distanceRemaining + calculateDistance(coordinate1: upcomingPoint, coordinate2: location)
```
S2Ler, line 161: This one I want more feedback on. 

**The problem: ** When the progress proceed after the final point, `traveledIndex.distanceRemaining` become zero, but `calculateDistance(coordinate1: upcomingPoint, coordinate2: location)` starts to grow for obvious reasons: distance is always positive.    

**Solution: ** Check if the progress is at final leg, user arrived at the waypoint and the distance remaining is 0. First two checks are straightforward, but the second one is debatable. I've observed that when currentLegProgress.distanceRemaining reaches 0 it doesn't change anymore. And that's what we want as far as I see. If anybody else knows a better check, I'm more than happy to change this code.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -878,7 +882,9 @@ extension NavigationViewController: NavigationServiceDelegate {
     }
     
     public func navigationService(_ service: NavigationService, shouldPreventReroutesWhenArrivingAt waypoint: Waypoint) -> Bool {
-        return navigationComponents.allSatisfy { $0.navigationService(service, shouldPreventReroutesWhenArrivingAt: waypoint) }
+        let componentsWantPreventReroutes = navigationComponents.allSatisfy { $0.navigationService(service, shouldPreventReroutesWhenArrivingAt: waypoint) }
+
+        return componentsWantPreventReroutes && (delegate?.navigationViewController(self, shouldPreventReroutesWhenArrivingAt: waypoint)) ?? RouteController.DefaultBehavior.shouldRerouteFromLocation
```
S2Ler, line 887: This delegate method wasn't propagated to the `NavigationViewControllerDelegate`. Without returning `false` from here, reroutes after arrival won't work.

S2Ler:
@1ec5 ready for review

S2Ler:
Had to remove vanishing line fix due to recent changes in this code. It is not longer that trivial as before and I'll cut a separate ticket for that.

Sources/MapboxNavigation/NavigationViewControllerDelegate.swift
```diff
@@ -64,7 +64,18 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
      - returns: True to allow the navigation view controller to calculate a new route; false to keep tracking the current route.
      */
     func navigationViewController(_ navigationViewController: NavigationViewController, shouldRerouteFrom location: CLLocation) -> Bool
-    
+
+    /**
+     Called when the user arrives at a waypoint.
+
+     Return false to continue checking if reroute is needed. By default, the user will not be rerouted when arriving at a waypoint.
```
1ec5, line 71: The documentation for this method and `NavigationServiceDelegate.navigationService(_:shouldPreventReroutesWhenArrivingAt:)` should link to each other, so that developers who implement a custom UI don’t think they have to downgrade to NavigationViewController or vice versa to get this functionality.

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@
...
 * Added the `RouteController.startRecordingHistory()`, `RouteController.stopRecordingHistory(completionHandler:)`, `PassiveLocationManager.startRecordingHistory()`, and `PassiveLocationManager.stopRecordingHistory(completionHandler:)` methods for recording details about a trip for debugging purposes. ([#3157](https://github.com/mapbox/mapbox-navigation-ios/pull/3157))
 * Renamed `RouterDataSource.locationProvider` and `EventsManagerDataSource.locationProvider` properties to `RouterDataSource.locationManagerType` and `EventsManagerDataSource.locationManagerType` respectively. ([#3199](https://github.com/mapbox/mapbox-navigation-ios/pull/3199))
+* Adds `NavigationViewControllerDelegate.navigationViewController(_:shouldPreventReroutesWhenArrivingAt:waypoint:)` which is called each time the user arrives at a waypoint. By default, this method returns true and prevents rerouting upon arriving. ([#3195](https://github.com/mapbox/mapbox-navigation-ios/pull/3195))
```
1ec5, line 58:
```suggestion
* Adds `NavigationViewControllerDelegate.navigationViewController(_:shouldPreventReroutesWhenArrivingAt:)`, which is called each time the user arrives at a waypoint. By default, this method returns true and prevents rerouting upon arriving. ([#3195](https://github.com/mapbox/mapbox-navigation-ios/pull/3195))
```

Sources/MapboxNavigation/NavigationViewController.swift
1ec5: As tail work, maybe we should replace DefaultBehavior with default method implementations on the respective protocols, in this case NavigationServiceDelegate and NavigationComponent. That way the default behavior happens automatically without us needing to remember to fall back to it.

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@
...
+* Fixes `NavigationViewControllerDelegate.navigationViewController(_:shouldRerouteFrom:)`. To implement reroute after arrival behavior, you can now return `false` from this method, `false` from `NavigationViewControllerDelegate.navigationViewController(_:shouldPreventReroutesWhenArrivingAt:waypoint:)` and set `NavigationViewController.showsEndOfRouteFeedback` to `false`.
```
1ec5, line 59:
```suggestion
* Fixes an issue where the user would be rerouted even if `NavigationViewControllerDelegate.navigationViewController(_:shouldRerouteFrom:)` returned `false`. To implement reroute after arrival behavior, return `false` in both this method and `NavigationViewControllerDelegate.navigationViewController(_:shouldPreventReroutesWhenArrivingAt:)`, and set `NavigationViewController.showsEndOfRouteFeedback` to `false`. ([#3195](https://github.com/mapbox/mapbox-navigation-ios/pull/3195))
```

1ec5 (approved):

Sources/MapboxNavigation/NavigationViewControllerDelegate.swift
S2Ler: I will make a follow up update after discussion. I'm not sure why we need to link these two methods together.

1ec5: Both are public methods that are named similarly, which means a developer might accidentally find the documentation about one when they need the other one.

Issue #2953, opened 2021-04-26
Exposing NavigationMapView components


Udumft:
Continuing after [RouteMapViewController elimination](https://github.com/mapbox/mapbox-navigation-ios/issues/1502), there are further topics to work on. [This PR](https://github.com/mapbox/mapbox-navigation-ios/pull/2902) shaped several components ([ArrivalController](https://github.com/mapbox/mapbox-navigation-ios/blob/release-v2.0/Sources/MapboxNavigation/ArrivalController.swift), [CameraController](https://github.com/mapbox/mapbox-navigation-ios/blob/release-v2.0/Sources/MapboxNavigation/CameraController.swift), [OrnamentsController](https://github.com/mapbox/mapbox-navigation-ios/blob/release-v2.0/Sources/MapboxNavigation/OrnamentsController.swift) and [RouteLineController](https://github.com/mapbox/mapbox-navigation-ios/blob/release-v2.0/Sources/MapboxNavigation/RouteLineController.swift)) which functionality could be exposed to users and further customisation.

Points to keep in mind for implementation:
* Decouple components from `NavigationMapView` or any other private dependency.
* Design configurable components behavior.
* Allow potential usage outside `NavigationViewController`.
* Re-use components in CarPlay

/cc @1ec5 for more points and prioritisation.

PR #3196, opened 2021-07-22
ArrivalController remix

Udumft:
vk-2953-arrival-controller: reworked ArrivalController into EndOfRout……eFeedbackCollecting VC extension. Removed original ArrivalController, removed end of route feedback bits from NavigationView and NavigationViewController.

### Description
Part of #2953 
This PR does the refactoring of one of the NavigationMapView, to simplify it's usage and allow customization and potential usage outside of NavigationView.

### Implementation
I decided that it would be easier to supply this feature as a UIViewController extension with default implementation. Motivation behind it is that Users may want to embed feedback collection to their custom navigation VCs, but since `EndOfRouteViewController` is not exposed we need to provide some wrapping. This also removed feedback-related code from `NavigationMapView`.


Sources/MapboxNavigation/EndOfRouteFeedbackCollecting.swift
```diff
@@ -0,0 +1,142 @@
+import UIKit
+import MapboxDirections
+import MapboxCoreNavigation
+import MapboxMobileEvents
+
+
+public protocol EndOfRouteFeedbackCollecting: UIViewController {
+    typealias EndOfRouteDismissalHandler = (EndOfRouteFeedback?) -> ()
+    
+//    private var endOfRouteViewController: EndOfRouteViewController // expose via protocol?
```
Udumft, line 10: `EndOfRouteViewController` is not available publicly(by the way, why?). So here we could either expose this class, or introduce protocol stub to allow users to provide their own custom `EndOfRouteViewController`'s versions.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -821,12 +802,33 @@ extension NavigationViewController: NavigationServiceDelegate {
...
+            router.routeProgress.isFinalLeg &&
+            showsEndOfRouteFeedback {
+            let destination = route?.legs.last?.destination
+            let controller = showEndOfRouteFeedbackCollection(destination: destination,
+                                                              onDismiss: { [weak self] in
+                                                                self?.navigationService.endNavigation(feedback: $0)
+                                                                self?.handleCancelAction()
+                                                              })
+            
+            navigationMapView?.navigationCamera.stop()
```
Udumft, line 815: Had to spit out camera logic here, outside `EndOfRouteFeedbackCollecting`, to separate the functionality. Hopefully this will be refined later, while tweaking `CameraController`.

Sources/MapboxNavigation/EndOfRouteFeedbackCollecting.swift
```diff
@@ -0,0 +1,142 @@
+import UIKit
+import MapboxDirections
+import MapboxCoreNavigation
+import MapboxMobileEvents
+
+
+public protocol EndOfRouteFeedbackCollecting: UIViewController {
```
Udumft, line 7: Naming suggestions are welcome! :)

Sources/MapboxNavigation/EndOfRouteFeedbackCollecting.swift
```diff
@@ -0,0 +1,142 @@
...
+    typealias EndOfRouteDismissalHandler = (EndOfRouteFeedback?) -> ()
+    
+//    private var endOfRouteViewController: EndOfRouteViewController // expose via protocol?
+    
+    func showEndOfRouteFeedbackCollection(destination: Waypoint?,
+                                          onDismiss: EndOfRouteDismissalHandler?) -> UIViewController
+}
+
+extension EndOfRouteFeedbackCollecting {
```
Udumft, line 16: I would like to introduce as much customization options as possible, but the majority of things are happening inside `EndOfRouteViewController`. This extension is mostly a bootstrapping code. Anyway I think I might invest more into this part to allow, for example, embedding `EndOfRouteViewController` to the top of the screen or something like that.

Issue #3194, opened 2021-07-21
Add speed limit indicator to CarPlay navigation activity


1ec5:
#2291 added SpeedLimitView to the CarPlay navigation activity, and #2847 add a SpeedLimitView to the example application’s preview map view, but we should also add a SpeedLimitView to the CarPlay navigation activity, maybe also the preview activity, for completeness.

/cc @mapbox/navigation-ios @OttyLab

PR #3197, opened 2021-07-22
Add `SpeedLimitView` to `CarPlayMapViewController`.

MaximAlien:
### Description

Closing #3194.

Screen Shot 2021-07-22 at 2 32 44 PM

### DIFF
--- a/Sources/MapboxNavigation/CarPlayMapViewController.swift
+++ b/Sources/MapboxNavigation/CarPlayMapViewController.swift
@@ -1,6 +1,7 @@
 import Foundation
 import MapboxMaps
 import MapboxCoreNavigation
+import MapboxDirections
 
 #if canImport(CarPlay)
 import CarPlay
@@ -22,6 +23,11 @@ public class CarPlayMapViewController: UIViewController {
      The style can be modified programmatically by using `StyleManager.applyStyle(type:)`.
      */
     public private(set) var styleManager: StyleManager?
+    
+    /**
+     A view that displays the current speed limit.
+     */
+    public weak var speedLimitView: SpeedLimitView!
     
     /**
      The interface styles available to `styleManager` for display.
@@ -114,6 +120,8 @@ public class CarPlayMapViewController: UIViewController {
      */
     internal(set) public var dismissPanningButton: CPMapButton?
     
+    // MARK: - Initialization methods
+    
     /**
      Initializes a new CarPlay map view controller.
      
@@ -140,6 +148,8 @@ public class CarPlayMapViewController: UIViewController {
         aCoder.encode(styles, forKey: "styles")
     }
     
+    // MARK: - UIViewController lifecycle methods
+    
     override public func loadView() {
         setupNavigationMapView()
         setupPassiveLocationProvider()
@@ -149,6 +159,7 @@ public class CarPlayMapViewController: UIViewController {
         super.viewDidLoad()
         
         setupStyleManager()
+        setupSpeedLimitView()
         navigationMapView.navigationCamera.follow()
     }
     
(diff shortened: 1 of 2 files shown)


Sources/MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -190,16 +207,52 @@ public class CarPlayMapViewController: UIViewController {
...
     
+    func setupSpeedLimitView() {
+        let speedLimitView = SpeedLimitView()
+        speedLimitView.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(speedLimitView)
+        
+        speedLimitView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor, constant: 8).isActive = true
+        speedLimitView.rightAnchor.constraint(equalTo: view.rightAnchor, constant: -8).isActive = true
+        speedLimitView.widthAnchor.constraint(equalToConstant: 30).isActive = true
+        speedLimitView.heightAnchor.constraint(equalToConstant: 30).isActive = true
```
S2Ler, line 218: nit: There is a nice convenience method for activating multiple constraints:
```suggestion
        NSLayoutConstraint.active([
            speedLimitView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor, constant: 8),
            speedLimitView.rightAnchor.constraint(equalTo: view.rightAnchor, constant: -8),
            speedLimitView.widthAnchor.constraint(equalToConstant: 30),
            speedLimitView.heightAnchor.constraint(equalToConstant: 30),
        ])
```

Sources/MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -190,16 +207,52 @@ public class CarPlayMapViewController: UIViewController {
         styleManager?.styles = styles
     }
     
+    func setupSpeedLimitView() {
+        let speedLimitView = SpeedLimitView()
+        speedLimitView.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(speedLimitView)
+        
+        speedLimitView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor, constant: 8).isActive = true
+        speedLimitView.rightAnchor.constraint(equalTo: view.rightAnchor, constant: -8).isActive = true
```
S2Ler, line 216: nit: Don't we support RTL languages? It is recommended to use trailing/leading instead of right/left. Although it requires all other views to use t/l as well. Probably we need an epic for RTL support.

S2Ler (approved):
Just few optional nits.

Sources/MapboxNavigation/CarPlayMapViewController.swift
azarovalex: Not sure if RTL support is necessary here, afaik this view doesn't contain any localized text.
I would rather think about supporting right-wheel cars, we need to make sure UI looks good there as well.

Sources/MapboxNavigation/CarPlayMapViewController.swift
azarovalex: I prefer the `isActive = true` approach because it takes fewer lines 😀

azarovalex (approved):

Sources/MapboxNavigation/CarPlayMapViewController.swift
MaximAlien: Both approaches work for me, I don't see major advantages/disadvantages between the two though.

Sources/MapboxNavigation/CarPlayMapViewController.swift
MaximAlien: It's hard to check behavior with right-to-left languages. Even though it can be changed via project scheme CarPlay will not be affected and trip preview view will stay unchanged. Once we have real hardware I'll retest this.

Issue #3126, opened 2021-06-28
Layer adding and removing error in CarPlay


ShanMa1991:
**Mapbox Navigation SDK version: v2.0.0-beta.15

### Steps to reproduce

 1. start navigation, choose default basic type
 2. look at the console

### Expected behavior
maneuvers fully visible, vanishing effect works

### Actual behavior
There're several error messages related with layer adding and removal, including the following:
```
2021-06-28 16:11:46.738476-0700 Example[19295:468299] Failed to add route casing source 0x00006000003d4580.main.source_casing with error: Source 0x00006000003d4580.main.source_casing already exists.
2021-06-28 16:11:46.880410-0700 Example[19295:468299] Failed to remove layer MapboxNavigation-MapboxNavigation-resources_waypointSymbolLayer with error: Layer MapboxNavigation-MapboxNavigation-resources_waypointSymbolLayer does not exist.
2021-06-28 16:11:46.880876-0700 Example[19295:468299] Failed to remove layer MapboxNavigation-MapboxNavigation-resources_waypointCircleLayer with error: Layer MapboxNavigation-MapboxNavigation-resources_waypointCircleLayer does not exist.
2021-06-28 16:11:46.881058-0700 Example[19295:468299] Failed to remove source MapboxNavigation-MapboxNavigation-resources_waypointSource with error: Source MapboxNavigation-MapboxNavigation-resources_waypointSource is in use, cannot remove.
2021-06-28 16:11:46.883699-0700 Example[19295:468299] Failed to remove source 0x00006000003d4580.main.source with error: Source 0x00006000003d4580.main.source is in use, cannot remove.
2021-06-28 16:11:46.894284-0700 Example[19295:468299] Failed to add route casing source 0x0000600000282980.main.source_casing with error: Source 0x0000600000282980.main.source_casing already exists.
```
These errors should be related with CarPlay. It would cause the maneuvers part hidden due to the `waypointCircleLayer` and `mainRouteLayer` existence check [failed](https://github.com/mapbox/mapbox-navigation-ios/pull/3106#discussion_r657435532) , and also would cause the vanishing effect not working.

ShanMa1991:
It's caused by the following reasons.
1. When the route refresh, the route line already exists, we should update the sauce and layer, instead of add or remove
2. The delegate for the casing layer and layer is mixed.

[...]

PR #3201, opened 2021-07-23
Improve the performance of vanishing route line feature

### LINKED ISSUE #3141: Improving the performance of vanishing route line feature
Right now the vanishing route line feature asks for a lot of calculation and leads to high CPU usage. Improved in #2737 , we are using the `fractionTraveled`  to update route line and route casing layer for the vanishing effect. But the `fractionTraveled` is based on location from the start point to the end point. 

Whenever we receive a route progress and location update, we call the `updateUpcomingRoutePointIndex(routeProgress:)` to find and cache the index of the upcoming [`RouteLineDistancesIndex`]. And then we call `updateTraveledRouteLine(_:)` to calculate the `fractionTraveled` based on the current location. And Finally we call the `updateRoute(_:)` to update the route and keep the route line synced with puck.

In #2737, because the default `navigationMapView.userCourseView` is using an evenly changed animation, and receiving the location update per second. We want the route line could be synced with the puck during the second, instead of an abrupt jump, Thus we introduced the Timer to cut 1 second(the puck animation duration) into 20 pieces, and cut the `fractionTraveled` into small pieces with the time slice. In each small time slice, we re-calculate the congestion features of the whole route, the route line gradient of the whole route based on `fractionTraveled` and congestion features, and then update the route line gradient.

The whole process involves a lot of calculation. Especially to keep the route line synced with the animated puck in each second, we do roughly 20 times of calculation to calculate congestion feature, route line gradient and update the route line gradient. When we have a really long route, the calculation would consume a lot of CPU usage. 

With the new map puck introduced, the route line is synced with the puck as a location update consumer from `MapView`. It means that the frequency to calculate congestion feature, route line gradient and update the route line gradient is replaced from 20 times per second to ? times based on how frequently the location update from the `mapView` side. 

We could improve the performance of the vanishing route line feature based on some possible ways as following:
1. The `updateUpcomingRoutePointIndex(routeProgress: )` is called per second. For multi leg route, really long route, the of remaining points of the remaining legs or steps could be stored and avoided.
2. The calculation of congestion features  and route line Gradient are called per location update. For really long route with frequently location update, some steps that are far away may need to be stored, and the calculation should be at these small sections of current leg or current step. 
3. When refresh the route, we do calculation of the whole route with the route progress and location update.

### LINKED ISSUE #3170: Verify whether `isAlternativeRoute` parameter usage is needed.
**Mapbox Navigation SDK version:** `main`

`Route.congestionFeatures(legIndex:isAlternativeRoute:roadClassesWithOverriddenCongestionLevels:)` exposes `isAlternativeRoute` parameter, which is later saved to `Feature.properties` and is never used.

In scope of this ticket it should be verified whether `isAlternativeRoute` parameter is really needed, and if not removed.

ShanMa1991:
### Description
This pr is to fix #3141  to improve the performance, reduce calculation and battery usage as below.

|                                                  | previous                     | now                                                                                |
|--------------------------------------------------|------------------------------|----------------------------------------------------------------------------------------|
| calculate route congestion features frequency    | by frame rate                | when refresh(/30s in Example), route leg change, or route congestion level info change |
| update route line gradient method per frame rate | generate route line gradient | update route line gradient                                                             |
| generate route line gradient frequency           | by frame rate                | when refresh(/30s in Example), route leg change, or route congestion level info change |
| update route line gradient                       |                              | truncate the existing line gradient to a smaller size                                  |

### Implementation
- [x] Calculate the current route leg congestion features only when it's changed and `routeLineTracksTraversal` enabled.
- [x] Remove `Timer` related parameters for vanishing route line implementation.
- [x] Replace the `navigationMapView.updateRoute(_:)` with `navigationMapView.updateVanishingRouteLine(coordinate::)` to keep the vanishing route line better synced with `routeProgress` update and location update, even if the frame rate is low.
- [x]  Fix the [issue](https://github.com/mapbox/mapbox-navigation-ios/issues/3126#issuecomment-887041529) cause by delegate mix of route layer and casing layer shape in adding layers. 
- [x] Update the vanishing route line when routeProgress update and refreshed, to keep the route line better synced under low frame rate.
- [x] Update the line gradient stops instead of re-calculating them during vanishing route line, to reduce the calculation.
- [x]  Remove the usage of the `isAlternativeRoute` parameter, fix #3170 
- [x] Fix the `rerouting` and `index` issue with removing the forcefully feeding `mapView` with location update when in non-simulation mode and default `userCourseView` used, and stop to call the navigation service of leg index update in CarPlay.
- [x] Fix the leg index change in `CarPlay` after arrival of intermediate waypoint, which may lead to navigator confused about the status.
- [x] Fix the issue when leg index change, the `CarPlay` doesn't change the styling of route line and the `waypoint`.
- [x] Update tests and documentation
- [x] Update the changeLog.

### Screenshots or Gifs

routeLine

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@
 * If you need to customize the appearance of the user location indicator, you can subclass `UserPuckCourseView` and `UserHaloCourseView` as a starting point. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
 * Fixed an issue where route line disappears when changing `MapView` style. ([#3136](https://github.com/mapbox/mapbox-navigation-ios/pull/3136))
 * Added `NavigationOptions.navigationMapView` property to allow customization or reusing possibilities for `NavigationViewController.navigationMapView` ([#3186](https://github.com/mapbox/mapbox-navigation-ios/pull/3186)).
-* Improved the performance for vanishing route line by reducing calculation for route line update. Replaced `NavigationMapView.updateRoute(_:)` with `NavigationMapView.updateVanishingRouteLine(coordinate:)` to update the vanishing route line, and keep the route line vanishing point synced with the user location. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
+* Renamed the `NavigationMapView.updateRoute(_:)` method to `NavigationMapView.travelAlongRouteLine(to:)`. Improved the performance of updating the route line to change color at the user’s location as they progress along the route. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
 
 ### Location tracking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 * If you need to customize the appearance of the user location indicator, you can subclass `UserPuckCourseView` and `UserHaloCourseView` as a starting point. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
 * Fixed an issue where route line disappears when changing `MapView` style. ([#3136](https://github.com/mapbox/mapbox-navigation-ios/pull/3136))
 * Added `NavigationOptions.navigationMapView` property to allow customization or reusing possibilities for `NavigationViewController.navigationMapView` ([#3186](https://github.com/mapbox/mapbox-navigation-ios/pull/3186)).
+* Improved the performance for vanishing route line by reducing calculation for route line update. Replaced `NavigationMapView.updateRoute(_:)` with `NavigationMapView.updateVanishingRouteLine(coordinate:)` to update the vanishing route line, and keep the route line vanishing point synced with the user location. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
 
 ### Location tracking
 
--- a/Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
+++ b/Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
@@ -146,10 +146,11 @@ extension NavigationMapView {
     }
     
     /**
-     Updates the route line appearance from the origin point to the indicated point
-     - parameter coordinate: current position of the puck
+     Updates the fractionTraveled along the route line from the origin point to the indicated point.
+     
+     - parameter coordinate: Current position of the user location.
      */
-    func updateTraveledRouteLine(_ coordinate: CLLocationCoordinate2D?) {
+    func updateFractionTraveled(_ coordinate: CLLocationCoordinate2D?) {
         guard let granularDistances = routeLineGranularDistances,let index = routeRemainingDistancesIndex, let location = coordinate else { return }
         guard index < granularDistances.distanceArray.endIndex else { return }
         let traveledIndex = granularDistances.distanceArray[index]
@@ -166,7 +167,6 @@ extension NavigationMapView {
         if granularDistances.distance >= remainingDistance {
             let offSet = (1.0 - remainingDistance / granularDistances.distance)
             if offSet >= 0 {
-                preFractionTraveled = fractionTraveled
                 fractionTraveled = offSet
             }
         }
@@ -175,15 +175,19 @@ extension NavigationMapView {
     /**
      Updates the route style layer and its casing style layer to gradually disappear as the user location puck travels along the displayed route.
      
+     - parameter coordinate: Current position of the user location.
      - parameter routeProgress: Current route progress.
      */
-    public func updateRoute(_ routeProgress: RouteProgress) {
-        let mainRouteLayerIdentifier = routeProgress.route.identifier(.route(isMainRoute: true))
-        let mainRouteCasingLayerIdentifier = routeProgress.route.identifier(.routeCasing(isMainRoute: true))
+    public func updateVanishingRouteLine(coordinate: CLLocationCoordinate2D?) {
+        updateFractionTraveled(coordinate)
+        
+        guard let route = routes?.first else { return }
+        
+        let mainRouteLayerIdentifier = route.identifier(.route(isMainRoute: true))
+        let mainRouteCasingLayerIdentifier = route.identifier(.routeCasing(isMainRoute: true))
         
         if fractionTraveled >= 1.0 {
             // In case if route was fully travelled - remove main route and its casing.
-            
             do {
                 try mapView.mapboxMap.style.removeLayer(withId: mainRouteLayerIdentifier)
                 try mapView.mapboxMap.style.removeLayer(withId: mainRouteCasingLayerIdentifier)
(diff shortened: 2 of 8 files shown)


Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -857,8 +857,6 @@ extension NavigationViewController: NavigationServiceDelegate {
             component.navigationService(service, didEndSimulating: progress, becauseOf: reason)
         }
         navigationMapView?.mapView.location.overrideLocationProvider(with: AppleLocationProvider())
-        navigationMapView?.mapView.location.locationProvider.startUpdatingHeading()
-        navigationMapView?.mapView.location.locationProvider.startUpdatingLocation()
```
ShanMa1991, line 861: In case of the locating background sending update to the navigator and cause rerouting issue.

Sources/MapboxNavigation/Route.swift
```diff
@@ -44,13 +44,13 @@ extension Route {
     }
     
     func congestionFeatures(legIndex: Int? = nil,
-                            isAlternativeRoute: Bool = false,
```
ShanMa1991, line 47: remove the `isAlternativeRoute` parameter, which is not using.

Sources/MapboxNavigation/RouteLineController.swift
```diff
@@ -21,8 +21,6 @@ extension NavigationMapView {
             navigationViewData.router
         }
         
-        fileprivate var routeProgress: RouteProgress?
```
ShanMa1991, line 24: instead of storing the routeProgress to update vanishing route line, when the congestion level/route/current leg index is not changed, we only use the location coordinate to truncate the existing gradient stops.

Tests/MapboxNavigationTests/NavigationMapViewTests.swift
```diff
@@ -200,7 +204,7 @@ class NavigationMapViewTests: TestCase {
     func testOverriddenStreetsRouteClassMotorwayMixedCongestionLevels() {
         let route = loadRoute(from: "route-with-mixed-road-classes")
         var congestions = route.congestionFeatures()
-        XCTAssertEqual(congestions.count, 5)
+        XCTAssertEqual(congestions.count, 1)
```
ShanMa1991, line 207: In one route, each leg has one congestion attribute feature. Each attribute feature contains the UIColor and the associate color change start point fraction in this leg distance.

ShanMa1991:
Right now we have 2 different user location indicator layer. One is the `userCourseView`. When the location change, it's on the animation with the duration of 1 second to evenly move from one location to another. The other one is the map puck, which is the location indicator listener from the `mapView`. When the location change, it's moving directly to the new location.

When we use the `userCourseView` and the location change during each frame, the route line would follow the `mostRecentUserCourseViewLocation` to change the break point at the `mostRecentUserCourseViewLocation`. But the `userCourseView` would behave animation for the transition, it would lead to some unsyncing behavior when the speed is very high or very close.

When we use the map puck and location change, both of the puck and the route line are all location indicator listener, and would both move to the update location instantly. It would lead to a better synced behavior than we use the `userCourseView`.

One way to improve the syncing behavior of the `userCourseView`  and the route line, is to change the route line vanishing effect as an evenly animation, like we did before with the `Timer`. Another way is to update the `userCourseView` of the location animation  similar as the map puck.

MaximAlien:
I think there is a regression when multiple routes are returned (alternative route line should not be highlighted):

Simulator Screen Shot - iPhone SE (2nd generation) - 2021-08-03 at 18 16 35

Sources/MapboxNavigation/Array.swift
```diff
@@ -56,48 +56,48 @@ extension Array where Iterator.Element == [CLLocationCoordinate2D]? {
...
+     For each item in the dictionary,  a `CongestionLevel` substitution will take place that has a streets road class contained in the `roadClassesWithOverriddenCongestionLevels` collection.
      For each of these items the `CongestionLevel` for `.unknown` traffic congestion will be replaced with the `.low` traffic congestion.
      
      - parameter congestionLevels: The congestion levels along a leg. There should be one fewer congestion levels than coordinates.
      - parameter streetsRoadClasses: A collection of streets road classes for each geometry index in `Intersection`. There should be the same amount of `streetsRoadClasses` and `congestions`.
      - parameter roadClassesWithOverriddenCongestionLevels: Streets road classes for which a `CongestionLevel` substitution should occur.
-     - returns: A list of `CongestionSegment` tuples with coordinate and congestion level.
+     - returns: A dictionary of the fraction of the congestion level change start point in route leg, and the congestion level after that.
      */
     func combined(_ congestionLevels: [CongestionLevel],
```
MaximAlien, line 69: Looks like refactored method is the source of performance loss. For example it takes around 10 seconds:
```
Started: 2021-08-03 18:19:31.676258-0700
Finished: 2021-08-03 18:19:41.667548-0700
```
to request the route from San Francisco to Seattle.

ShanMa1991: Yes, you're right, the performance loss in the above is caused by the calculation of the distance. In the previous method, we only calculate the distance of each color slice once in the `func routeLineGradient(_:fractionTraveled:isMain:)` , the `func combined(_:streetsRoadClasses:roadClassesWithOverriddenCongestionLevels:)` only store the coordinates slice for each color.

This pr calculate the distance of each color slice from the start of the leg once in the `func combined`, and also the total leg distance once for the fraction. In the `func routeLineGradient`, it re-calculate the leg distance once. The extra calculation of distance based on the `LineString` should be the cause. Thanks for the suggestion, I'll improve this one.

ShanMa1991: Thanks for the suggestion, I'll removed the line gradients stops generating part. And only keeps the gradient stops updating part, with tests added for empty stops, single unknown stops, and different `fractionTraveled`.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -175,15 +175,19 @@ extension NavigationMapView {
     /**
      Updates the route style layer and its casing style layer to gradually disappear as the user location puck travels along the displayed route.
      
+     - parameter coordinate: Current position of the user location.
      - parameter routeProgress: Current route progress.
      */
-    public func updateRoute(_ routeProgress: RouteProgress) {
-        let mainRouteLayerIdentifier = routeProgress.route.identifier(.route(isMainRoute: true))
-        let mainRouteCasingLayerIdentifier = routeProgress.route.identifier(.routeCasing(isMainRoute: true))
+    public func updateVanishingRouteLine(coordinate: CLLocationCoordinate2D?) {
```
1ec5, line 181: If `updateRoute(_:)` is too ambiguous, consider `travelAlongRouteLine(to:)`, to evoke the same terminology as `updateFractionTraveled(_:)` above.

The term “vanishing route line” is sort of an internal shorthand for the visual effect we have by default, but I think it could be confusing to bake it into a symbol name in the public API. For example, a developer may style the route line so that it changes colors but doesn’t vanish behind the puck. For that reason, the option that enables or disables this feature is called `routeLineTracksTraversal` in an effort to be more descriptive of behavior and less prescriptive about look and feel.

If `routeLineTracksTraversal` is set to `false`, does client code still need to update the route line based on progress, or we sure this functionality is only needed for when the vanishing route line feature is enabled?

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -146,10 +146,11 @@ extension NavigationMapView {
...
     
     /**
-     Updates the route line appearance from the origin point to the indicated point
-     - parameter coordinate: current position of the puck
+     Updates the fractionTraveled along the route line from the origin point to the indicated point.
+     
+     - parameter coordinate: Current position of the user location.
      */
-    func updateTraveledRouteLine(_ coordinate: CLLocationCoordinate2D?) {
+    func updateFractionTraveled(_ coordinate: CLLocationCoordinate2D?) {
```
1ec5, line 153: A coordinate isn’t a fraction, so rename this method to `updateFractionTraveled(coordinate:)`, removing the `_`.

Sources/MapboxNavigation/Route.swift
```diff
@@ -70,17 +70,15 @@ extension Route {
                     var feature = Feature(geometry: .lineString(LineString(congestionSegment.0)))
                     feature.properties = [
                         CongestionAttribute: String(describing: congestionSegment.1),
-                        "isAlternativeRoute": isAlternativeRoute,
```
1ec5, line 73: Note to self: this is no longer necessary because we’re putting the alternative route in a separate source to be rendered by a separate layer, no longer using data-driven styling to consolidate layers and sources.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -378,6 +366,22 @@ open class NavigationMapView: UIView {
...
     
+    func updateUserCourseViewWithAccuracy() {
+        let frame = CGRect(origin: .zero, size: 75.0)
+        let isHidden = userCourseView.isHidden
+        switch userLocationStyle {
+        case .courseView(let courseView):
+            userCourseView = reducedAccuracyActivatedMode
+                ? UserHaloCourseView(frame: frame)
+                : courseView
+        default:
```
1ec5, line 377: Explicitly match the `puck2D` and `puck3D` cases here, so that the compiler reminds us to double-check this code if we ever add another case to the enumeration. There are a couple of similar default cases on `UserLocationStyle` elsewhere in this file.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 * If you need to customize the appearance of the user location indicator, you can subclass `UserPuckCourseView` and `UserHaloCourseView` as a starting point. ([#2968](https://github.com/mapbox/mapbox-navigation-ios/pull/2968))
 * Fixed an issue where route line disappears when changing `MapView` style. ([#3136](https://github.com/mapbox/mapbox-navigation-ios/pull/3136))
 * Added `NavigationOptions.navigationMapView` property to allow customization or reusing possibilities for `NavigationViewController.navigationMapView` ([#3186](https://github.com/mapbox/mapbox-navigation-ios/pull/3186)).
+* Improved the performance for vanishing route line by reducing calculation for route line update. Replaced `NavigationMapView.updateRoute(_:)` with `NavigationMapView.updateVanishingRouteLine(coordinate:)` to update the vanishing route line, and keep the route line vanishing point synced with the user location. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
```
1ec5, line 43:
```suggestion
* Renamed the `NavigationMapView.updateRoute(_:)` method to `NavigationMapView.updateVanishingRouteLine(coordinate:)`. Improved the performance of updating the route line to change color at the user’s location as they progress along the route. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201))
```

1ec5 (approved):

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: No, this method is only called when `routeLineTracksTraversal ` is set to `true`. Because it only works is to providing the vanishing effect for the route line. Thus if the customer want to change colors or the route line, this method would not be used.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll add the explicit cases to avoid future error.

PR #3204, opened 2021-07-26
Migrate ISSUE_TEMPLATE.md to GitHub Forms

S2Ler:
This PR migrates from markdown issue template to GitHub Forms.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,51 @@
...
+        Before continuing, can you check if the answer is available within:
+
+        * Guides and examples - https://www.mapbox.com/ios-sdk/navigation/
+        * API reference - https://www.mapbox.com/mapbox-navigation-ios/navigation/
+        * The example app - https://github.com/mapbox/navigation-ios-examples
+  - type: dropdown
+    attributes:
+      label: "Mapbox Navigation SDK version"
+      options:
+        - 2.0.0-beta.19
```
zugaldia, line 20: Does this mean that we'd need to update this template with every release? (Or at least every major release?)

S2Ler: Suggestions are welcomed :) I would list here max 7 most common versions. Including beta here is temporary as we don't have release yet.

1ec5: A freeform textbox is probably OK for this field. Once v2.0.0 is released, most bug reports will be against a version that’s easier to type. 🙂

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,51 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        Hello and thanks for contributing!
+
+        Before continuing, can you check if the answer is available within:
+
+        * Guides and examples - https://www.mapbox.com/ios-sdk/navigation/
+        * API reference - https://www.mapbox.com/mapbox-navigation-ios/navigation/
+        * The example app - https://github.com/mapbox/navigation-ios-examples
```
1ec5, line 15: Should we mention mapbox/mapbox-maps-ios and mapbox/mapbox-directions-swift as other likely places to file issues in?

S2Ler: I like that

S2Ler: done

Issue #3193, opened 2021-07-21
Route line is sometimes invisible after starting active guidance navigation.


MaximAlien:
**Mapbox Navigation SDK version:** `main`

Right after starting active guidance navigation route line is sometimes invisible. For example:

<img width="363" alt="Screen Shot 2021-07-20 at 6 08 39 PM" src="https://user-images.githubusercontent.com/1496498/126416787-c6726b51-0588-43ae-b67a-832d3dd951af.png">

Issue occurs because of this check:

https://github.com/mapbox/mapbox-navigation-ios/blob/e308e0ce2db6f7382db08698145256d4b7891f10/Sources/MapboxNavigation/RouteLineController.swift#L50-L51

Due to timing, `navigationViewData.containerViewController.view.window` can be `nil` sometimes.

PR #3205, opened 2021-07-26
Fix route line visibility.

MaximAlien:
### Description

Closing #3193.

### DIFF
--- a/Sources/MapboxNavigation/RouteLineController.swift
+++ b/Sources/MapboxNavigation/RouteLineController.swift
@@ -47,8 +47,8 @@ extension NavigationMapView {
         // MARK: - Private methods
         
         private func showRouteIfNeeded() {
-            guard navigationViewData.containerViewController.isViewLoaded &&
-                    navigationViewData.containerViewController.view.window != nil else { return }
+            guard navigationViewData.containerViewController.isViewLoaded else { return }
+            
             guard !navigationMapView.showsRoute else { return }
             navigationMapView.show([router.route], legIndex: router.routeProgress.legIndex)
             navigationMapView.showWaypoints(on: router.route, legIndex: router.routeProgress.legIndex)
@@ -153,8 +153,7 @@ extension NavigationMapView {
             }
         }
         
-        func onViewDidAppear() {
-            showRouteIfNeeded()
+        func navigationViewDidAppear(_ animated: Bool) {
             currentLegIndexMapped = router.routeProgress.legIndex
             currentStepIndexMapped = router.routeProgress.currentLegProgress.stepIndex
         }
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/RouteLineController.swift
```diff
@@ -153,8 +153,7 @@ extension NavigationMapView {
             }
         }
         
-        func onViewDidAppear() {
-            showRouteIfNeeded()
```
MaximAlien, line 157: Previous delegate method name was incorrect, so it was never called, not sure if we really need `navigationViewDidAppear(_:)` anyway. Call to `showRouteIfNeeded()` is not needed either as it'll be called each time `MapView` style changes.

Sources/MapboxNavigation/RouteLineController.swift
```diff
@@ -47,8 +47,8 @@ extension NavigationMapView {
         // MARK: - Private methods
         
         private func showRouteIfNeeded() {
-            guard navigationViewData.containerViewController.isViewLoaded &&
-                    navigationViewData.containerViewController.view.window != nil else { return }
```
MaximAlien, line 51: @Udumft, would you be able to provide more details on why you've added this check: `navigationViewData.containerViewController.view.window != nil`? At this point it prevents from showing route line right at the start of active guidance navigation session, when using `NavigationViewController`. I wonder if I miss anything here.

Tests/MapboxNavigationTests/NavigationViewControllerTests.swift
```diff
@@ -329,20 +329,22 @@ class NavigationViewControllerTests: TestCase {
...
               let indexOfArrowStrokeLayer = allLayerIds.firstIndex(of: NavigationMapView.LayerIdentifier.arrowStrokeLayer),
               let indexOfArrowSymbolLayer = allLayerIds.firstIndex(of: NavigationMapView.LayerIdentifier.arrowSymbolLayer),
               let indexOfPuck3DLayer = allLayerIds.firstIndex(of: NavigationMapView.LayerIdentifier.puck3DLayer) else {
             XCTFail("Failed to find all the layers"); return
         }
         
-        XCTAssertNotNil(indexOfArrowStrokeLayer, "Arrow stroke layer failed to be added")
-        XCTAssert(indexOfArrowStrokeLayer < indexOfArrowLayer, "Arrow layer is below arrow stroke layer")
+        // Since `NavigationViewController` presents route line after its presentation it is expected
+        // that maneuver arrow stroke layer will be added above main route line layer.
```
jill-cardamon, line 346: Thanks for adding comments into the test code! That area of the code base would benefit from more in-line comments.

PR #3207, opened 2021-07-27
Fix MapboxCoreNavigation compilation on Xcode 13b3

S2Ler:
MapboxNavigation cannot be fixed due to MapboxMaps that needs to get the fix first.

This is a regression due to a new rule in Xcode 13b3:

Linking Swift packages from application extension targets or watchOS applications no longer emits unresolvable warnings about linking to libraries not safe for use in application extensions. This means that code referencing APIs annotated as unavailable for use in app extensions must now themselves be annotated as unavailable for use in application extensions, in order to allow that code to be used in both apps and app extensions. (66928265)

https://developer.apple.com/documentation/xcode-release-notes/xcode-13-beta-release-notes


S2Ler:
No longer needed since Xcode 13b5

PR #3208, opened 2021-07-28
Make `NavigationService` optional to be able to skip implementation of delegate method on client side.

MaximAlien:
Since `CarPlayManager` always provides default `MapboxNavigationService` implementation there is no need to force user to provide another instance even if it's similar to the previous one.

This is usability improvement, which was found during implementation of https://github.com/mapbox/mapbox-navigation-ios-examples/pull/134.

### DIFF
--- a/Sources/MapboxNavigation/CarPlayManagerDelegate.swift
+++ b/Sources/MapboxNavigation/CarPlayManagerDelegate.swift
@@ -84,7 +84,7 @@ public protocol CarPlayManagerDelegate: AnyObject, UnimplementedLogging {
      - parameter trip: The trip that will be previewed.
      - returns: The actual trip to be previewed. This can be the same trip or a new/alternate trip if desired.
      */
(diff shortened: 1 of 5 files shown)


bamx23 (approved):

Sources/MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -84,7 +84,7 @@ public protocol CarPlayManagerDelegate: AnyObject, UnimplementedLogging {
      - parameter trip: The trip that will be previewed.
      - returns: The actual trip to be previewed. This can be the same trip or a new/alternate trip if desired.
      */
-    func carPlayManager(_ carPlayManager: CarPlayManager, willPreview trip: CPTrip) -> (CPTrip)
+    func carPlayManager(_ carPlayManager: CarPlayManager, willPreview trip: CPTrip) -> CPTrip
```
MaximAlien, line 87: Not sure why these delegate methods return tuple. Changed signature to return object instead.

PR #3209, opened 2021-07-28
Do not show traffic by default in `CarPlayMapViewController`.

MaximAlien:
### Description

To match `NavigationViewController`, `NavigationMapView` and `CarPlayNavigationViewController` behavior, `CarPlayMapViewController` should not show traffic.


jill-cardamon (approved):
lgtm :)

MaximAlien:
> lgtm :) do we need an entry in the changelog or release notes?

I don't think so. Even though this change is visible to the user, public APIs were not changed.

PR #3214, opened 2021-07-30
Location update source should be from routeController during active navigation.

ShanMa1991:
### Description
The rerouting and wrong location update sending to the nav status could be related with two location update source during active navigation. One is from the `routeController`, when we setup the mapView for navigationMapView, when there's a `mostRecentUserCourseViewLocation` changed by the routeProgress, we force the locationProvider to feed a  location update as below:

https://github.com/mapbox/mapbox-navigation-ios/blob/b4f00ef4026f547756530a65ac318b3a511b6714/Sources/MapboxNavigation/NavigationMapView.swift#L341-L355

But at the same time, when it's in simulation mode or not, the default `AppleLocationProvider` is still sending location updates to the location consumers, which would trigger rerouting issue, and it may be different from our routeController.  Even though we nil out the `navigationMapView` in preview, but the `navigationMapView` during navigation is still affected by the background location update. Due to the singleton of the navigator, it would trigger several issues.

### Implementation
Stop the location and heading update from the locationProvider during active navigation from the background. We only feed the location Provider with the location update from the routeController.

### DIFF
--- a/Sources/MapboxNavigation/CarPlayManager.swift
+++ b/Sources/MapboxNavigation/CarPlayManager.swift
@@ -228,10 +228,6 @@ public class CarPlayManager: NSObject {
         let route = navigationService.route
         let routeOptions = navigationService.routeProgress.routeOptions
         
-        // Stop the background `PassiveLocationProvider` sending location and heading update `mapView` before turn-by-turn navigation session starts.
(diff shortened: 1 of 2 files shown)


ShanMa1991:
It's fixed by #3201 by removing the forcefully feeding of location update when default `userCourseView` is used.

PR #3215, opened 2021-08-02
New properties in NavigationSettings

S2Ler:
PassiveLocationManager, RouteController gives you a false sense that you can have various tile store configs for these controllers, but we have just a singleton Navigator. I found that different application parts depend on tile store config + Directions instances. While this is more or less ok using many instances of Directions, it is not ok to use different tile store configs. You need a heavy-weight cache handler to use tileset descriptors, so we shouldn't have more than one. And cache handlers depend on a tile store config.

So to communicate better to the developers that Navigation SDK should use one instance of tileStoreConfiguration and provide ability to change default Directions instance used across the SDK, I added new parameters to NavigationSettings and a new initialize method which can be used to override these parameters. This change is a breaking change and might require fixing docs/examples.


Sources/MapboxCoreNavigation/NavSDK.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+import MapboxDirections
+
+public struct NavSDK {
```
1ec5, line 4: Consider combining this struct with NavigationSettings.

Sources/MapboxCoreNavigation/NavSDK.swift
```diff
@@ -0,0 +1,41 @@
...
+    private static let lock: NSLock = .init()
+    private static var isInitialized: Bool = false
+    public static var directions: Directions!
+    public static var tileStoreConfiguration: TileStoreConfiguration!
+
+    public static func initialize(directions: Directions = .shared,
+                                  tileStoreConfiguration: TileStoreConfiguration = .default) {
```
1ec5, line 11: Not too keen on an initializer that has to be called before everything else in the application. Developers would be prone to forgetting to call it, and overall it feels like a non-Cocoa way of setting up a library versus something more tightly scoped. If this code really has to run as soon as possible, then consider sticking an Objective-C `+initialize` method in CMapboxCoreNavigation.

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -151,7 +151,9 @@ public class MapboxNavigationService: NSObject, NavigationService {
     /**
      A reference to a MapboxDirections service. Used for rerouting.
      */
-    public var directions: Directions
+    public var directions: Directions {
+        NavSDK.directions
```
bamx23, line 155: There is a case when you want to request a route from one endpoint and also use it for rerouting, but use a different endpoint for navigation tiles(in Navigator)

Sources/MapboxCoreNavigation/NavSDK.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+import MapboxDirections
+
+public struct NavSDK {
+    private static let lock: NSLock = .init()
+    private static var isInitialized: Bool = false
+    public static var directions: Directions!
+    public static var tileStoreConfiguration: TileStoreConfiguration!
```
bamx23, line 8: What if we use default values here and don't require calling `initialize` at the very start of the app?

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -40,7 +37,9 @@ open class PassiveLocationManager: NSObject {
     /**
      The directions service that allows the location manager to access road network data.
      */
-    public let directions: Directions
+    public var directions: Directions {
```
bamx23, line 40: Do we need such variables after these changes? I would remove them or make them private

Sources/MapboxCoreNavigation/NavSDK.swift
S2Ler: As per discussion, this will become an advance method with fallback to defaults.

Sources/MapboxCoreNavigation/NavigationService.swift
S2Ler: I decided to keep previous implementation and just default to the NavSDK.directions instead of simply .shared.

Sources/MapboxCoreNavigation/NavSDK.swift
S2Ler: Yep, will do, thanks for suggestion. It will be the best of two worlds.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
S2Ler: I reverted the change

Sources/MapboxCoreNavigation/NavSDK.swift
S2Ler: done, please review the latest revision.

Sources/MapboxCoreNavigation/NavSDK.swift
S2Ler: This is no longer a required call. Please review the latest revision.

Sources/MapboxCoreNavigation/NavigationService.swift
S2Ler: Done, please review the latest revision.

Sources/MapboxCoreNavigation/NavSDK.swift
S2Ler: Done, please review the latest revision.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
S2Ler: Done, please review the latest revision.

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -23,6 +24,47 @@ public class NavigationSettings {
             }
         }
     }
+
+    private let lock: NSLock = .init()
+    private var _directions: Directions?
+    private var _tileStoreConfiguration: TileStoreConfiguration?
```
bamx23, line 30: I would suggest creating a struct with these two properties and store that struct. This reduces code duplication in accessors.

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -73,17 +115,6 @@ public class NavigationSettings {
      */
     public static let shared = NavigationSettings()
     
-    /// Returns a reflection of this class excluding the `properties` variable.
-    lazy var properties: [Mirror.Child] = {
```
bamx23, line 77: Why has this been removed?

S2Ler: It is not used anywhere. Do I miss something?

Sources/MapboxCoreNavigation/NavigationSettings.swift
S2Ler: Done, please review the latest revision.

S2Ler:
Changelog is coming

1ec5 (approved):
I’m really liking how much simpler the code is becoming this way. I wonder if there’s a way to make the lifecycle clearer to developers through more extensive use of UserDefaults, as an optional alternative to calling `initialize`. But apart from some documentation-related nits, I don’t see any red flags with this approach.

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -23,6 +27,76 @@ public class NavigationSettings {
...
+     Not doing so will lead to undefined behavior.
+
+     - Parameters:
+       - directions: Default `Directions` instance. Some types allow you to customize the directions instance and
+     fall back to the `NavigationSettings.directions` by default.
+       - tileStoreConfiguration: Options for configuring how map and navigation tiles are stored on the device. See
+     `TileStoreConfiguration` for more details.
+     */
+    public func initialize(directions: Directions,
+                           tileStoreConfiguration: TileStoreConfiguration) {
```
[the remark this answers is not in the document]
S2Ler: I will explore this approach, thanks for suggestion, I like it

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -1,11 +1,15 @@
...
 
 /**
- A wrapper for the `UserDefaults` class for navigation-specific settings.
- 
- Properties are prefixed before they are stored in `UserDefaults.standard`.
- 
- To specify criteria when calculating routes, use the `NavigationRouteOptions` class. To customize the user experience during a particular turn-by-turn navigation session, use the `NavigationOptions` class when initializing a `NavigationViewController`.
+ Global settings that are used across the SDK for altering navigation behavior.
+
+ To specify criteria when calculating routes, use the `NavigationRouteOptions` class.
```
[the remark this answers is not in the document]
S2Ler: Done, please review the latest revision.

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -23,6 +27,76 @@ public class NavigationSettings {
...
+        state.tileStoreConfiguration
+    }
+
+    /**
+     Initializes the settings with custom instances of globally used types.
+
+     If you don't provide custom values, they will be initialized with the defaults.
+
+     - important: If you want to use this method, you should call it before you start using Navigation SDK.
+     Not doing so will lead to undefined behavior.
```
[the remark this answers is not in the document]
S2Ler: Please help me with this. I've changed the wording a little bit but not sure how to make it more precise without being too wordy and unnecessary specific.

Sources/MapboxCoreNavigation/NavigationSettings.swift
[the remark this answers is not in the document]
S2Ler: Given that it is an improvement, I will create a tail work ticket for this suggestion.

CHANGELOG.md
```diff
@@ -139,7 +138,15 @@
...
+* Removed `NavigationOptions.tileStoreConfiguration` in favor of `NavigationSettings.tileStoreConfiguration`. ([#3215](https://github.com/mapbox/mapbox-navigation-ios/pull/3215))
+* Removed `tileStoreConfiguration` parameter `PredictiveCacheManager.init(predictiveCacheOptions:styleSourcePaths:)` in favor of `NavigationSettings.tileStoreConfiguration`. ([#3215](https://github.com/mapbox/mapbox-navigation-ios/pull/3215))
+* `NavigationViewController.mapTileStore` is now obtained from `NavigationSettings.tileStoreConfiguration`. ([#3215](https://github.com/mapbox/mapbox-navigation-ios/pull/3215))
+* Removed `tileStoreConfiguration` parameter from `NavigationMapView.enablePredictiveCaching(options:)` in favor of `NavigationSettings.tileStoreConfiguration`. ([#3215](https://github.com/mapbox/mapbox-navigation-ios/pull/3215))
```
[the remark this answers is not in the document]
S2Ler: Where would I put these changelong entries?

Sources/MapboxCoreNavigation/NavigationSettings.swift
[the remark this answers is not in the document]
S2Ler: #3259

CHANGELOG.md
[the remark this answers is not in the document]
1ec5: #3376 removed the blurbs from the running changelog, but in the future, you can add intra-release changes to the PR description and/or a draft release’s release notes.

PR #3218, opened 2021-08-03
CarPlay vanishing route line

### LINKED ISSUE #3200: Vanishing route line is not working in CarPlay
**Mapbox Navigation SDK version:** main

Just found that the vanishing route line is not working in CarPlay, but the `navigationMapView.updateRoute(_:)` is called

### Expected behavior
CarPlay

ShanMa1991:
### Description
- [x] fix #3200 
- [x] in case the `routeLineTracksTraversal` changed during navigation session, allow the route line update its color.

### Implementation
Due to the landing of the #3201, it needs the stored route location points to calculate useful `fractionTraveled`. We use the `navigationMapView.initPrimaryRoutePoints(route:)` in `navigationMapView.show(_:legIndex:)` to prepare for it in mobile. But in CarPlay, it would only be called if the route leg index change or rerouting. Thus the  route location points are not stored for the `fractionTraveled`. 

### Screenshots or Gifs
In the following .gif, it shows the we enable and disable the vanishing feature per 3 seconds during the navigation session.
navSession

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -171,6 +171,8 @@ open class NavigationMapView: UIView {
             if routeLineTracksTraversal, let route = self.routes?.first {
                 initPrimaryRoutePoints(route: route)
                 setUpLineGradientStops(along: route)
+            } else {
+                removeLineGradientStops()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -162,11 +162,18 @@ open class NavigationMapView: UIView {
...
     var routeLineGranularDistances: RouteLineGranularDistances?
     var routeRemainingDistancesIndex: Int?
-    var routeLineTracksTraversal: Bool = false
     var fractionTraveled: Double = 0.0
     var currentLegIndex: Int?
     var currentLegCongestionLevels: [CongestionLevel]?
     var currentLineGradientStops = [Double: UIColor]()
+    var routeLineTracksTraversal: Bool = false {
+        didSet {
+            if routeLineTracksTraversal, let route = self.routes?.first {
```
S2Ler, line 171: - does it mean that if routes is set after `routeLineTracksTraversal` this code won't work? 
- what if `routeLineTracksTraversal` changed when navigation started, will it work correctly?

ShanMa1991: 1. Yes, if routes set after `routeLineTracksTraversal `, it won't be triggered. To store the route coordinates points for vanishing route line, by logic it would call the `navigationMapView.show(_:legIndex:)` show the route line first, and then it would check the `routeLineTracksTraversal` and decide whether to store the route points.
2. Good question. If during navigation session and this parameter changes to `true`, it works. I'll add the function to disable this feature when it sets to `false` during navigation session.

ShanMa1991: The gif above tested the parameter value change during navigation.

S2Ler: Thanks for the gif, looks good to me!

S2Ler (approved):
LGTM

MaximAlien (approved):

PR #3222, opened 2021-08-04
Fix retain cycle causing crash in RouteOverlayController on NavigationMapView reuse

irtemed88:
### Description

This PR fixes a crash due to a retain cycle caused as part of `NavigationMapView` reuse across multiple instances of `NavigationViewController` which was made available with the injection work in #3186.

- [x] Updated Changelog

### Implementation

`RouteOverlayController.navigationViewDidLoad` defines two closures for handling map events. Each closure captures a strong reference to `self` which will result in a retain cycle if the referenced `NavigationMapView` object is used across multiple instances of `NavigationViewController`. The crash occurs on presentation of the _subsequent_ `NavigationViewController` 

Steps to reproduce this bug are:

1. Create an instance of `NavigationMapView`
2. Create and present an instance of `NavigationViewController` and inject the map view created in step 1.
3. Dismiss the instance of `NavigationViewController` ensuring the map view is still retained.
4. Create and present a second instance of `NavigationViewController` and inject the map view created in step 1.
5. Observe crash in `RouteOverlayController.showRouteIfNeeded` when `RouteOverlayController.navigationViewData` (which is implicitly unwrapped) is accessed. The crashing controller instance will be the one created by `NavigationViewController` in step 2.

Using `weak self` references breaks the retain cycle allowing `RouteOverlayController` to deinit when the owning `NavigationViewController` is deinit.

It's unclear to me at this point if the handlers added to `navigationMapView.mapView.mapboxMap` also need to be cleaned, but in the interim I believe this fix should address the more immediate crash.

### DIFF
--- a/Sources/MapboxNavigation/RouteLineController.swift
+++ b/Sources/MapboxNavigation/RouteLineController.swift
@@ -139,17 +139,17 @@ extension NavigationMapView {
         // MARK: - NavigationComponentDelegate implementation
         
         func navigationViewDidLoad(_ view: UIView) {
-            navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { [self] _ in
-                navigationMapView.localizeLabels()
-                navigationMapView.mapView.showsTraffic = false
+            navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { [weak self] _ in
+                self?.navigationMapView.localizeLabels()
+                self?.navigationMapView.mapView.showsTraffic = false
                 
                 // FIXME: In case when building highlighting feature is enabled due to style changes and no info currently being stored
                 // regarding building identification such highlighted building will disappear.
             }
             
             // Route line should be added to `MapView`, when its style changes.
-            navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { [self] _ in
-                showRouteIfNeeded()
+            navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { [weak self] _ in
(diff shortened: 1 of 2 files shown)


MaximAlien (approved):

irtemed88:
Looks like CI failed for some reason? Is this something I need to address before merging?

cc @MaximAlien

Sources/MapboxNavigation/RouteLineController.swift
```diff
@@ -139,17 +139,17 @@ extension NavigationMapView {
         // MARK: - NavigationComponentDelegate implementation
         
         func navigationViewDidLoad(_ view: UIView) {
-            navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { [self] _ in
-                navigationMapView.localizeLabels()
-                navigationMapView.mapView.showsTraffic = false
+            navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { [weak self] _ in
+                self?.navigationMapView.localizeLabels()
+                self?.navigationMapView.mapView.showsTraffic = false
```
azarovalex, line 144: In theory `self` can be deallocated between these lines, so as a rule of thumb I always add `guard let self = self else { return }` if self is used more than once.
Not sure if it's critical in this case.

MaximAlien:
> Looks like CI failed for some reason? Is this something I need to address before merging?
> 
> cc @MaximAlien

Looks like it's temporary glitch on CircleCI. I've restarted failed jobs. Besides this everything looks good.

PR #3235, opened 2021-08-10
Fix buggy default MapboxNavigationService creation if custom one is provided

S2Ler:
A side effect of this is that the shared navigator is updated in
quick succession in Router initialization.

Discovered during #3195 QA.


Udumft (approved):
I thought such change was a part of pricing implementation, to eliminate extra navigation service instantiation. Maybe it got reverted by some merge conflicts...

S2Ler:
> I thought such change was a part of pricing implementation, to eliminate extra navigation service instantiation. Maybe it got reverted by some merge conflicts...

Maybe it was, but I want this to get into release earlier and it is easier to review.

PR #3238, opened 2021-08-10
Fix missing feedback subtype descriptions

bamx23:
### Description

There was a missing feedback subtype description for `.incorrectVisual(subtype: .incorrectSpeedLimit)`.
This PR fixes this and prevent similar issues by removing `default` switch case.


S2Ler (approved):
nice, `default` is overrated.

bamx23:
Should I mention this change in the CHANGELOG?

S2Ler:
> Should I mention this change in the CHANGELOG?

I would say, it won't hurt to mention this fix in the changelog so that it can appear in the beta release notes.

PR #3239, opened 2021-08-10
Fix first sent location in NavigationService.start method

### LINKED ISSUE #3237: Initial autogenerated location when NavigationService starts causes teleports and off route detection
**Mapbox Navigation SDK version:** 5adad71c97cde98a431dce0d400619403a8d5839

### Steps to reproduce

 1. Start Example app with Freeway drive simulation in Simulator.
 2. Start history recording
 2. Select a location ahead
 3. Wait a bit
 4. Start navigation
 5. Stop history recording after destination is reached
 6. Read history file

### Expected behavior

There is no teleports in the history.

### Actual behavior

There are teleports in the history and unexpected off route triggered. 

### The cause

[This](https://github.com/mapbox/mapbox-navigation-ios/blob/5adad71c97cde98a431dce0d400619403a8d5839/Sources/MapboxCoreNavigation/NavigationService.swift#L332-L336) code simulates the first location to Navigator which is the beginning of the route. But at this time the car can already be ahead. 

### Solution

Either:
- Don't feed this bogus first location
- Find a better alternative which won't cause teleports. 

### Related issues

#1790

S2Ler:
Fixes #3237


Udumft (approved):
LGTM

Issue #3240, opened 2021-08-10
Cover functionality provided by `NavigationGeocodedPlacemark` with automated tests.


MaximAlien:
**Mapbox Navigation SDK version:** `v2.0.0-beta.21`

https://github.com/mapbox/mapbox-navigation-ios/pull/2999 added new `NavigationGeocodedPlacemark` struct, which aims to simplify work with `GeocodedPlacemark` on CarPlay.

In scope of this ticket, functionality provided by `NavigationGeocodedPlacemark` should be covered by automated tests.

PR #3242, opened 2021-08-10
Add `NavigationGeocodedPlacemarkTests` tests.

MaximAlien:
### Description

Closing #3240.

### DIFF
--- a/MapboxNavigation.xcodeproj/project.pbxproj
+++ b/MapboxNavigation.xcodeproj/project.pbxproj
@@ -173,6 +173,7 @@
 		8A2DFA8626168A300034A87E /* NavigationCameraDebugView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A2DFA8526168A300034A87E /* NavigationCameraDebugView.swift */; };
 		8A30113C25DDCC8A00CE192A /* NavigationCameraConstants.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A30113B25DDCC8A00CE192A /* NavigationCameraConstants.swift */; };
 		8A3474F5269F8B9D00BCF135 /* MapViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A3474F3269F8B9D00BCF135 /* MapViewTests.swift */; };
+		8A38B42126C5AC130088E168 /* CPMapTemplate+MBTestable.m in Sources */ = {isa = PBXBuildFile; fileRef = 8AB316AE26BB315100C3AC76 /* CPMapTemplate+MBTestable.m */; };
 		8A3A219025EEC00200EDA999 /* CoreNavigationNavigator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A3A218F25EEC00200EDA999 /* CoreNavigationNavigator.swift */; };
 		8A41F5B225BF61AE00BD6FCF /* CarPlayActivity.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A41F5B125BF61AE00BD6FCF /* CarPlayActivity.swift */; };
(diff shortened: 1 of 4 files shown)


MapboxNavigation.xcodeproj/project.pbxproj
```diff
@@ -179,14 +179,15 @@
...
 		8A446645260A7B24008BA55E /* BoundingBox.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A446644260A7B24008BA55E /* BoundingBox.swift */; };
+		8A6C468226C2FA2E00452EBE /* NavigationGeocodedPlacemarkTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A6C468126C2FA2E00452EBE /* NavigationGeocodedPlacemarkTests.swift */; };
 		8A8C3D98260175D20071D274 /* CLLocationDirection+Camera.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A8C3D97260175D20071D274 /* CLLocationDirection+Camera.swift */; };
 		8AA849E924E722410008EE59 /* WaypointStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AA849E824E722410008EE59 /* WaypointStyle.swift */; };
 		8AAE94A126A60ADE00AA1127 /* CarPlayMapViewControllerDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AAE94A026A60ADE00AA1127 /* CarPlayMapViewControllerDelegate.swift */; };
-		8AB316C926BCA56D00C3AC76 /* UIViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AB316C826BCA56D00C3AC76 /* UIViewController.swift */; };
-		8AB316CB26BCA72300C3AC76 /* CGSize.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AB316CA26BCA72300C3AC76 /* CGSize.swift */; };
```
S2Ler, line 186: Looks like this change shouldn't be in this commit and can be just reverted, unless it appears again.

Tests/MapboxNavigationTests/NavigationGeocodedPlacemarkTests.swift
```diff
@@ -0,0 +1,85 @@
+import XCTest
+import CarPlay
+import MapboxNavigation
+@testable import TestHelper
```
S2Ler, line 4: `TestHelper` is supposed to be used in tests only so if any API is  internal but needed in tests we can just make this API public.

Tests/MapboxNavigationTests/NavigationGeocodedPlacemarkTests.swift
```diff
@@ -0,0 +1,85 @@
...
+                                                                      routableLocations: nil)
+        
+        let listItem = navigationGeocodedPlacemark.listItem()
+        XCTAssertEqual(listItem.text, title, "Titles should be equal.")
+        XCTAssertEqual(listItem.detailText, subtitle, "Subtitles should be equal.")
+        XCTAssertEqual(listItem.image, nil, "Image should not be set.")
+    }
+    
+    func testNavigationGeocodedPlacemarkEncodingAndDecoding() {
+        let coordinate = CLLocationCoordinate2D(latitude: 0.0, longitude: 0.0)
```
S2Ler, line 25: I think makes sense to change lat/lon to non zero values. 0 is a default value for primitive double values.

S2Ler (approved):

Tests/MapboxNavigationTests/NavigationGeocodedPlacemarkTests.swift
MaximAlien: Yep, removed.

Tests/MapboxNavigationTests/NavigationGeocodedPlacemarkTests.swift
MaximAlien: Don't think it's too critical at this point. Will leave it as it is.

PR #3245, opened 2021-08-11
Introduce IdleTimerManager to streamline idle timer management and fix bugs

### LINKED ISSUE #3227: Improve idle time management
# Problem 

Currently, `NavigationViewController` and `CarPlayManager` changes `UIApplication.shared.isIdleTimerDisabled` in a way that if there are two instances of `NavigationViewController` or `CarPlayManager` disconnects, the actual state of `UIApplication.shared.isIdleTimerDisabled` might end up in inconsistent state. 

Also, `CarPlayManager`s `isIdleTimerDisabled` usage should be revisit so that it doesn't enable idle timer on connect/disconnect. 

# Solution

Use singleton idle timer manager of some kind which will track how many requests to disable/enable timer of various kinds were send and send appropriate values to `UIApplication.shared.isIdleTimerDisabled`. 

Possible places for this management code: 
- `NavigationSettings`
- New `IdleTimerManager` type: the favourable choice, because the case with CarPlay is more complex. 
- Your option. 

cc @mapbox/navigation-ios @1ec5

S2Ler:
fixes #3227

I drafted a POC of how idle timer can look like. Then naming is temporary, but for some reason I really like to use token here :)

Given that behaviour in CarPlayManager is very confusing and contradicts itself, I've change the logic to the following:
- If `carPlayManagerShouldDisableIdleTimer` returns true: idle timer is disabled while CarPlay is active, otherwise, do nothing. 
- If NavigationViewController is visible: disable idle timer. 

Does this sound reasonable to you? Should the logic be different?

### DIFF
--- a/Sources/MapboxNavigation/IdleTimerManager.swift
+++ b/Sources/MapboxNavigation/IdleTimerManager.swift
@@ -1,0 +1,61 @@
+import Foundation
+import UIKit
+@_spi(MapboxInternal) import MapboxCoreNavigation
+
+/// Manages `UIApplication.shared.isIdleTimerDisabled`. Thread-safe.
+final class IdleTimerManager {
+    /// While a cancellable isn't cancelled, the idle timer is disabled. A cancellable is cancelled on deallocation.
+    final class Cancellable {
+        private let onCancel: () -> Void
+
+        fileprivate init(_ onCancel: @escaping () -> Void) {
+            self.onCancel = onCancel
+        }
+
+        deinit {
+            onCancel()
+        }
+    }
+
+    static let shared: IdleTimerManager = .init()
+
+    private let lock: NSLock = .init()
+    /// Number of currently non cancelled `IdleTimerManager.Cancellable` instances.
+    private var _cancellablesCount: Int = 0 {
+        didSet {
+            _updateIdleTimer()
+        }
+    }
+
+    private init() {}
+
+    /**
+     Disables idle timer `UIApplication.shared.isIdleTimerDisabled`
+     while there is at least one non-cancelled `IdleTimerManager.Cancellable` instance.
+
+     - Returns: An instance of cancellable that you should retain until you want the idle timer to be disabled.
+     */
+    func disableIdleTimer() -> IdleTimerManager.Cancellable {
+        let cancellable = Cancellable {
+            self.changeCancellablesCount(delta: -1)
+        }
+        changeCancellablesCount(delta: 1)
+        return cancellable
+    }
+
+    private func _updateIdleTimer() {
+        let isIdleTimerDisabled = _cancellablesCount > 0
+
+        onMainAsync {
+            UIApplication.shared.isIdleTimerDisabled = isIdleTimerDisabled
+        }
+    }
+
+    private func changeCancellablesCount(delta: Int) {
+        lock.lock(); defer {
+            lock.unlock()
+        }
+        _cancellablesCount += delta
+    }
+}
+
--- a/Sources/MapboxNavigation/IdleTimerManager.swift
+++ b/Sources/MapboxNavigation/IdleTimerManager.swift
@@ -2,6 +2,20 @@ import Foundation
 import UIKit
 @_spi(MapboxInternal) import MapboxCoreNavigation
 
+/// An idle timer which is managed by `IdleTimerManager`.
+protocol IdleTimer {
+    func setDisabled(_ disabled: Bool)
+}
+
+/// UIApplication specific idle timer.
+struct UIApplicationIdleTimer: IdleTimer {
+    func setDisabled(_ disabled: Bool) {
+        onMainAsync {
+            UIApplication.shared.isIdleTimerDisabled = disabled
+        }
+    }
+}
+
 /// Manages `UIApplication.shared.isIdleTimerDisabled`. Thread-safe.
 final class IdleTimerManager {
     /// While a cancellable isn't cancelled, the idle timer is disabled. A cancellable is cancelled on deallocation.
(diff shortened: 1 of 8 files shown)


Udumft (commented):
LGTM, token idea seems fit to me. But maybe we should also couple it with `PassiveLocationProvider` lifecycle to disable idling while free drive is active too?

Sources/MapboxNavigation/IdleTimerManager.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+import UIKit
+
+/// Manages UIApplication.shared.isIdleTimerDisabled
+final class IdleTimerManager {
+    final class Token {
```
bamx23, line 6: Maybe `Handle` is a more natural name for this structure. Similar to `CacheHandle`

Sources/MapboxNavigation/IdleTimerManager.swift
```diff
@@ -0,0 +1,43 @@
...
+        didSet {
+            updateIdleTimer()
+        }
+    }
+
+    private init() {}
+
+    /// Disabled idle timer while there is at least one Token alive.
+    func issueIdleToken() -> Token {
+        dispatchPrecondition(condition: .onQueue(.main))
```
bamx23, line 30: Should this precondition be moved to `updateIdleTimer` so that we check token deallocation too?

Sources/MapboxNavigation/IdleTimerManager.swift
```diff
@@ -0,0 +1,43 @@
...
+        let token = Token {
+            self.idleTokenCount -= 1
+        }
+        idleTokenCount += 1
+        return token
+    }
+
+    private func updateIdleTimer() {
+        UIApplication.shared.isIdleTimerDisabled = idleTokenCount > 0
+        print("$$$ Always On: \(UIApplication.shared.isIdleTimerDisabled)")
```
bamx23, line 40: Don't forget to clean up here

bamx23 (approved):

Sources/MapboxNavigation/IdleTimerManager.swift
```diff
@@ -0,0 +1,49 @@
...
+
+     - Returns: An instance of cancellable that you should retain until you want the idle timer to be disabled.
+     */
+    func disableIdleTimer() -> IdleTimerManager.Cancellable {
+        dispatchPrecondition(condition: .onQueue(.main))
+        let token = Cancellable {
+            self.cancellablesCount -= 1
+        }
+        cancellablesCount += 1
+        return token
```
bamx23, line 42:
```suggestion
        let cancellable = Cancellable {
            self.cancellablesCount -= 1
        }
        cancellablesCount += 1
        return cancellable
```

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -174,7 +174,13 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
     /**
      If `true`, `UIApplication.isIdleTimerDisabled` is set to `true` in `viewWillAppear(_:)` and `false` in `viewWillDisappear(_:)`. If your application manages the idle timer itself, set this property to `false`.
      */
-    public var shouldManageApplicationIdleTimer = true
+    public var shouldManageApplicationIdleTimer = true {
+        didSet {
```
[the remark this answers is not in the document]
bamx23: Should we initialize a new token here if it was set back to `true`?

Sources/MapboxNavigation/IdleTimerManager.swift
S2Ler: I end up with Cancellable similar to Combine.Cancellable concept. But couldn't use Combine due to deployment target limitations.

S2Ler: `Combine.Cancellable`, so be it.

Sources/MapboxNavigation/NavigationViewController.swift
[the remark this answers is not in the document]
S2Ler: Should take into account the case when vc isn't visible.

Sources/MapboxNavigation/IdleTimerManager.swift
```diff
@@ -0,0 +1,43 @@
...
+    private var idleTokenCount: Int = 0 {
+        didSet {
+            updateIdleTimer()
+        }
+    }
+
+    private init() {}
+
+    /// Disabled idle timer while there is at least one Token alive.
+    func issueIdleToken() -> Token {
```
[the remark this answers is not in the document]
S2Ler: yep.

Sources/MapboxNavigation/IdleTimerManager.swift
S2Ler: done, thanks

Sources/MapboxNavigation/IdleTimerManager.swift
S2Ler: I've changed the implementation to be thread-safe.

Sources/MapboxNavigation/NavigationViewController.swift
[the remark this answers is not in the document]
S2Ler: Done.

1ec5:
> LGTM, token idea seems fit to me. But maybe we should also couple it with `PassiveLocationProvider` lifecycle to disable idling while free drive is active too?

Seems to me that it should be the other way around. An application may want to automatically enable or disable PassiveLocationProvider depending on whether the application is in the foreground or background. I don’t think the presence of PassiveLocationProvider alone is enough to prevent the screen from locking, because the basic use case for it is just to clean up puck and camera behavior on a map that the user is using passively (hence the name of that class).

1ec5 (approved):

Issue #3241, opened 2021-08-10
Cover functionality provided by `RecentItem` with automated tests.


MaximAlien:
**Mapbox Navigation SDK version:** `v2.0.0-beta.21`

In scope of this ticket `RecentItem` struct should be covered with automated tests.

PR #3247, opened 2021-08-11
Add automated tests for `RecentItem`.

MaximAlien:
### Description

Closing #3241.

### DIFF
--- a/Sources/MapboxNavigation/RecentItem.swift
+++ b/Sources/MapboxNavigation/RecentItem.swift
@@ -2,21 +2,25 @@ import Foundation
 import CarPlay
 
 /**
- Struct, which represents recently found search item on CarPlay.
+ Struct, which represents recently found search item on CarPlay. When storing recent items in an array
+ use dedicated methods for addition: `[RecentItem].add(_:)`, and removal: `[RecentItem].remove(_:)`.
  */
 public struct RecentItem: Equatable, Codable {
     
     /**
      Property, which contains information regarding geocoder result.
      */
-    public var navigationGeocodedPlacemark: NavigationGeocodedPlacemark
+    public private(set) var navigationGeocodedPlacemark: NavigationGeocodedPlacemark
 
     var timestamp: Date
     
-    static var filePathUrl: URL {
+    static var recentItemsPathUrl: URL? {
         get {
-            let documents = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0]
-            let url = URL(fileURLWithPath: documents)
+            guard let documentsDirectory = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first else {
+                return nil
+            }
+            
+            let url = URL(fileURLWithPath: documentsDirectory)
             return url.appendingPathComponent("RecentItems.data")
         }
     }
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/RecentItem.swift
```diff
@@ -2,21 +2,25 @@ import Foundation
...
     /**
      Property, which contains information regarding geocoder result.
      */
-    public var navigationGeocodedPlacemark: NavigationGeocodedPlacemark
+    public private(set) var navigationGeocodedPlacemark: NavigationGeocodedPlacemark
 
     var timestamp: Date
     
-    static var filePathUrl: URL {
+    static var recentItemsPathUrl: URL? {
```
1ec5, line 17: While you’re at it, “URL” [should be capitalized](https://swift.org/documentation/api-design-guidelines/#follow-case-conventions) in the name when it isn’t the first word in the name.

Sources/MapboxNavigation/RecentItem.swift
```diff
@@ -66,12 +70,29 @@ public struct RecentItem: Equatable, Codable {
 
 extension Array where Element == RecentItem {
     
-    public func save() {
-        let encoder = JSONEncoder()
-        let data = try? encoder.encode(self)
-        (try? data?.write(to: RecentItem.filePathUrl)) as ()??
+    /**
+     Method, which allows to save an array of `RecentItem`s into file stored in `recentItemsPathUrl`.
```
1ec5, line 74:
```suggestion
     Saves an array of `RecentItem`s to a file at the path specified by `recentItemsPathURL`.
```

Sources/MapboxNavigation/RecentItem.swift
```diff
@@ -66,12 +70,29 @@ public struct RecentItem: Equatable, Codable {
...
+        
+        return true
     }
-
+    
+    /**
+     Method, which adds `RecentItem` to the collection. In case if similar `RecentItem` already exists
+     in collection, `timestamp` of its first occurrence will be updated.
+     
+     - parameter recentItem: `RecentItem` instance, which will be added to the collection.
```
1ec5, line 94:
```suggestion
     Adds a recent item to the collection. If a similar recent item is already in the collection, this method updates the `timestamp` of that item instead of adding a redundant item.
     
     - parameter recentItem: A recent item to add to the collection.
```

Sources/MapboxNavigation/RecentItem.swift
```diff
@@ -87,8 +108,13 @@ extension Array where Element == RecentItem {
         remove(validNavigationGeocodedPlacemark)
         add(updatedNavigationGeocodedPlacemark)
     }
-
-    mutating func remove(_ recentItem: RecentItem) {
+    
+    /**
+     Method, which removes from the collection first occurrence of a `RecentItem`.
+     
+     - parameter recentItem: `RecentItem` instance, which will be removed from the collection.
```
1ec5, line 115:
```suggestion
     Removes the first matching recent item from the collection.
     
     - parameter recentItem: A recent item to remove from the collection.
```

1ec5 (approved):
Thanks, just some nits to make the documentation a bit more concise, since Quick Help and jazzy already link types in method signatures.

PR #3221, opened 2021-08-04
v2.0.0-beta.21 

Udumft:
### Description
v2.0.0-beta.21 release changes. CHANGELOG appended a bit.

PR #3248, opened 2021-08-11
Update Mapbox Navigation dependencies.

MaximAlien:
Update dependencies to Turf `v2.0.0-rc.1`, Mapbox Common `v16.2.0`, Mapbox Navigation Native `v62.0.0` and Mapbox Maps `v10.0.0-rc.6`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,12 @@
 
 * The Mapbox Navigation SDK for iOS license has changed from the ISC License to the Mapbox Terms of Service. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * You can now install MapboxNavigation using Swift Package Manager, but you can no longer install it using Carthage. If you previously installed MapboxNavigation using Carthage, use Swift Package Manager instead. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
-* MapboxNavigation now depends on [MapboxMaps v10.0.0-rc.5](https://github.com/mapbox/mapbox-maps-ios/). ([#3162](https://github.com/mapbox/mapbox-navigation-ios/pull/3162))
-* MapboxNavigation now depends on MapboxNavigationNative v60.0.0. ([#3162](https://github.com/mapbox/mapbox-navigation-ios/pull/3162))
-* MapboxNavigation now depends on MapboxCommon v14.0.1. ([#3077](https://github.com/mapbox/mapbox-navigation-ios/pull/3077))
+* MapboxNavigation now depends on [MapboxMaps v10.0.0-rc.6](https://github.com/mapbox/mapbox-maps-ios/). ([#3248](https://github.com/mapbox/mapbox-navigation-ios/pull/3248))
+* MapboxNavigation now depends on MapboxNavigationNative v62.0.0. ([#3248](https://github.com/mapbox/mapbox-navigation-ios/pull/3248))
+* MapboxNavigation now depends on MapboxCommon v16.2.0. ([#3248](https://github.com/mapbox/mapbox-navigation-ios/pull/3248))
 * MapboxNavigation now depends on MapboxMobileEvents v1.0.2. ([#3039](https://github.com/mapbox/mapbox-navigation-ios/pull/3039))
 * MapboxCoreNavigation depends on MapboxDirections v2.0.0-beta.7 but no longer depends on MapboxAccounts. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808), [#2829](https://github.com/mapbox/mapbox-navigation-ios/pull/2829), [#2837](https://github.com/mapbox/mapbox-navigation-ios/pull/2837), [#3077](https://github.com/mapbox/mapbox-navigation-ios/pull/3077), [#3182](https://github.com/mapbox/mapbox-navigation-ios/pull/3182))
-* MapboxNavigation now depends on Turf v2.0.0-beta.1. ([#3077](https://github.com/mapbox/mapbox-navigation-ios/pull/3077))
+* MapboxNavigation now depends on Turf v2.0.0-rc.1. ([#3248](https://github.com/mapbox/mapbox-navigation-ios/pull/3248))
 * MapboxNavigation and MapboxCoreNavigation require iOS 11.0 or above to run. iOS 10._x_ is no longer supported. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Xcode 12.4 or above is now required for building this SDK from source.
 * Carthage v0.38 or above is now required for installing this SDK if you use Carthage. ([#3031](https://github.com/mapbox/mapbox-navigation-ios/pull/3031))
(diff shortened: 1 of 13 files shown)


Cartfile
```diff
@@ -1,5 +1,5 @@
-binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" "16.0.0"
-binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.xcframework.json" ~> 60.0
+binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" ~> 16.2.0
+binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.xcframework.json" ~> 62.0
 github "mapbox/mapbox-directions-swift" "v2.0.0-beta.7"
```
MaximAlien, line 3: During Mapbox Navigation release process Mapbox Directions will have to be released to depend on Turf `v2.0.0-rc.1`.

/cc @azarovalex

Cartfile
```diff
@@ -1,5 +1,5 @@
-binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" "16.0.0"
```
MaximAlien, line 1: I wonder whether it really worked before. Carthage emits such error in case if we use `"` for binary dependencies:

```
Parse error: binary dependencies cannot have a git reference for the version specifier in line: binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" "16.0.0"
```

CHANGELOG.md
```diff
@@ -6,12 +6,12 @@
...
-* MapboxNavigation now depends on MapboxCommon v14.0.1. ([#3077](https://github.com/mapbox/mapbox-navigation-ios/pull/3077))
+* MapboxNavigation now depends on [MapboxMaps v10.0.0-rc.6](https://github.com/mapbox/mapbox-maps-ios/). ([#3248](https://github.com/mapbox/mapbox-navigation-ios/pull/3248))
+* MapboxNavigation now depends on MapboxNavigationNative v62.0.0. ([#3248](https://github.com/mapbox/mapbox-navigation-ios/pull/3248))
+* MapboxNavigation now depends on MapboxCommon v16.2.0. ([#3248](https://github.com/mapbox/mapbox-navigation-ios/pull/3248))
 * MapboxNavigation now depends on MapboxMobileEvents v1.0.2. ([#3039](https://github.com/mapbox/mapbox-navigation-ios/pull/3039))
 * MapboxCoreNavigation depends on MapboxDirections v2.0.0-beta.7 but no longer depends on MapboxAccounts. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808), [#2829](https://github.com/mapbox/mapbox-navigation-ios/pull/2829), [#2837](https://github.com/mapbox/mapbox-navigation-ios/pull/2837), [#3077](https://github.com/mapbox/mapbox-navigation-ios/pull/3077), [#3182](https://github.com/mapbox/mapbox-navigation-ios/pull/3182))
```
1ec5, line 13: It’s a bit awkward that these two dependencies are part of the same blurb. When we upgrade MapboxDirections, it’ll be a good opportunity to split the blurb apart.

Cartfile
1ec5: This was broken in #3221, probably due to hand-editing the Cartfile as if it were a Cartfile.resolved, but we didn’t catch the issue somehow.

Tests/MapboxCoreNavigationTests/MapboxCoreNavigationTests.swift
```diff
@@ -118,11 +118,15 @@ class MapboxCoreNavigationTests: TestCase {
     
     func testNewStep() {
```
MaximAlien, line 119: `testNewStep` started failing once again due to mismatch of received spoken instructions. Had to increase number of coordinates to get missing instruction.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
```diff
@@ -92,10 +92,16 @@ final public class RoadObjectMatcher {
...
      - parameter point: Point representing the object.
      - parameter identifier: Unique identifier of the object.
+     - parameter direction: Direction of the provided point, which is going to be matched.
      */
-    public func match(point: CLLocationCoordinate2D, identifier: RoadObjectIdentifier) {
-        let point = MatchableGeometry(id: identifier, coordinates: [point].map(CLLocation.init))
-        native.matchPoints(forPoints: [point], useOnlyPreloadedTiles: false)
+    public func match(point: CLLocationCoordinate2D, identifier: RoadObjectIdentifier, direction: CLLocationDirection? = nil) {
+        var heading: NSNumber? = nil
+        if let direction = direction {
```
1ec5, line 99: A negative `CLLocationDirection` value is documented to mean an invalid value, which MatchablePoint would probably expect to be represented by `nil` instead, so we should check for this case.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
```diff
@@ -92,10 +92,16 @@ final public class RoadObjectMatcher {
 
      - parameter point: Point representing the object.
      - parameter identifier: Unique identifier of the object.
+     - parameter direction: Direction of the provided point, which is going to be matched.
      */
-    public func match(point: CLLocationCoordinate2D, identifier: RoadObjectIdentifier) {
-        let point = MatchableGeometry(id: identifier, coordinates: [point].map(CLLocation.init))
-        native.matchPoints(forPoints: [point], useOnlyPreloadedTiles: false)
+    public func match(point: CLLocationCoordinate2D, identifier: RoadObjectIdentifier, direction: CLLocationDirection? = nil) {
```
1ec5, line 97: Consider replacing CLLocationDirection with CLHeading for forward compatibility in case we ever need to consider the heading accuracy as part of matching. For that matter, even if we leave it as a CLLocationDirection, `heading` would be a less ambiguous label than `direction`, which can also refer to a map rotation direction or a relative direction. (A CLLocation would probably be a more correct representation of a coordinate plus a heading, even if some of the fields happen to be ignored for the time being.)

1ec5 (approved):
Overall this is fine, but we have some opportunities to tighten up Core Location type usage in this part of the codebase.

Sources/MapboxCoreNavigation/EHorizon/RoadObjectMatcher.swift
MaximAlien: Thanks, even though `CLLocation` usage is tempting it'll require public API change. Since third parameter uses default value it seems more reasonable to me. So I've replaced it to `CLHeading` for now.

Cartfile
MaximAlien: By the looks of it yes. Reverted to previous version and changed to `16.2.0`.

Cartfile
azarovalex: Thanks for taking care of the dependencies, I'll take a look at the Turf release

CHANGELOG.md
azarovalex: Agree, I updated these lines in this PR: https://github.com/mapbox/mapbox-navigation-ios/pull/3251/files
Also removed some PR links here, one link should be enough.

PR #3250, opened 2021-08-12
Add subgraph case to RoadObjectLocation

bamx23:
### Description

The case of subgraph was missing in the `RoadObjectLocation` enum. Without it, an app crashes in the `preconditionFailure`.

### Implementation

I've also added a corresponding struct for subgraph edges. I would appreciate help with providing better in-code documentation for the new struct.

### DIFF
--- a/Sources/MapboxCoreNavigation/EHorizon/RoadSubgraphEdge.swift
+++ b/Sources/MapboxCoreNavigation/EHorizon/RoadSubgraphEdge.swift
@@ -1,0 +1,59 @@
+import Foundation
+import Turf
+import MapboxNavigationNative
+
+extension RoadGraph {
+
+    /**
+     The `SubgraphEdge` represents a edge in the complex object which might be considered as a directed graph. The graph might contain loops. `innerEdgeIds` and `outerEdgeIds` properties contain edge ids, which allows to traverse the graph, obtain geometry and calculate different distances inside it.
+     */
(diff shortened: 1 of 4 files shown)


Sources/MapboxCoreNavigation/EHorizon/RoadSubgraphEdge.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+import Turf
+import MapboxNavigationNative
+
+extension RoadGraph {
+
+    /**
+     The `SubgraphEdge` represents a edge in the complex object which might be considered as a directed graph. The graph might contain loops. `innerEdgeIds` and `outerEdgeIds` properties contain edge ids, which allows to traverse the graph, obtain geometry and calculate different distances inside it.
```
azarovalex, line 8:
```suggestion
     The `SubgraphEdge` represents an edge in the complex object which might be considered as a directed graph. The graph might contain loops. `innerEdgeIds` and `outerEdgeIds` properties contain edge ids, which allows to traverse the graph, obtain geometry and calculate different distances inside it.
```

azarovalex (approved):
LGTM, the comment for `RoadGraph.SubgraphEdge` is more about the subgraph location itself, but I guess it's ok.

bamx23:
Thanks, @azarovalex. I'm going to merge this and if improvements in docs are needed, we can make them later.

PR #3187, opened 2021-07-19
Update translations from Transifex

1ec5:
Force-pulled translations from Transifex in all languages, since it has been a while since translations were last pulled in #2864 and some source files have moved around. Added a Castilian Spanish localization, which is now complete enough to show up when pulling translations.

/cc @mapbox/navigation-ios @Guardiola31337

PR #3253, opened 2021-08-12
Update source localization files

1ec5:
I split off an update to the source localization files from #3187 so we can land that PR next without any of the unused strings that apparently have translation issues anyways.

/cc @mapbox/navigation-ios

### DIFF
--- a/Sources/MapboxCoreNavigation/Resources/en.lproj/Localizable.strings
+++ b/Sources/MapboxCoreNavigation/Resources/en.lproj/Localizable.strings
@@ -1,27 +1,3 @@
-/* Format for speech string after completing a maneuver and starting a new step; 1 = distance */
-"CONTINUE" = "Continue for %@";
-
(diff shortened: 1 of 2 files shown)


Sources/MapboxCoreNavigation/Resources/en.lproj/Localizable.strings
```diff
@@ -1,27 +1,3 @@
-/* Format for speech string after completing a maneuver and starting a new step; 1 = distance */
```
azarovalex, line 1: Are these lines currently unused in the SDK?🤔

1ec5: Yes, it’s currently unused. The Directions API (online) and MapboxNavigationNative (offline) are responsible for providing speech strings.

MaximAlien (approved):

PR #3255, opened 2021-08-12
To improve the vanishing route line when rerouting happens

### LINKED ISSUE #3254: To improve the vanishing route line when rerouting happens
When during the navigation session and the `routeLineTracksTraversal` feature is enabled to hide the traversed trace, if there's a rerouting event, it would trigger the `navigationMapView` to generate a new route line. 

 When the user passed the destination but the rerouting event hasn't been triggered, the user location update will continue updating the vanishing route line, which leads to route line `grows` back.

ShanMa1991:
### Description
This pr is to fix #3254 and #3267

### Implementation
According to https://github.com/mapbox/mapbox-navigation-ios/issues/3254#issuecomment-901269168, the issue happened because when the user passed the destination, the route still exits because `fractionTraveled` not set to `1.0`. 

To avoid the `grow back` effect when user passed the destination and keep moving forward, while the `routeProgress` still sending update, we check whether we has passed the final leg waypoint and then remove the route. Thus if there's a rerouting event happened later, it would not generate a new route instead.

### Screenshots or Gifs
destination

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 * Fixed an issue where route line disappears when changing `MapView` style. ([#3136](https://github.com/mapbox/mapbox-navigation-ios/pull/3136))
 * Added `NavigationOptions.navigationMapView` property to allow customization or reusing possibilities for `NavigationViewController.navigationMapView` ([#3186](https://github.com/mapbox/mapbox-navigation-ios/pull/3186)).
 * Renamed the `NavigationMapView.updateRoute(_:)` method to `NavigationMapView.travelAlongRouteLine(to:)`. Improved the performance of updating the route line to change color at the user’s location as they progress along the route. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
+* Fixed an issue when user passed destination and the route line grows back when `routeLineTracksTraversal` enabled. ([#3255](https://github.com/mapbox/mapbox-navigation-ios/pull/3255))
 
 ### Location tracking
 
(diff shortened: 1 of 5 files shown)


ShanMa1991:
The test as the gif is run on "s2ler/rerouting-debug" branch in the Example.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
 * Fixed an issue where route line disappears when changing `MapView` style. ([#3136](https://github.com/mapbox/mapbox-navigation-ios/pull/3136))
 * Added `NavigationOptions.navigationMapView` property to allow customization or reusing possibilities for `NavigationViewController.navigationMapView` ([#3186](https://github.com/mapbox/mapbox-navigation-ios/pull/3186)).
 * Renamed the `NavigationMapView.updateRoute(_:)` method to `NavigationMapView.travelAlongRouteLine(to:)`. Improved the performance of updating the route line to change color at the user’s location as they progress along the route. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
+* Fixed an issue when user passed destination and the route line grows back when `routeLineTracksTraversal` enabled. ([#3255](https://github.com/mapbox/mapbox-navigation-ios/pull/3255))
```
1ec5, line 46:
```suggestion
* Fixed an issue when user passed destination and the route line grows back when `NavigationViewController.routeLineTracksTraversal` is enabled. ([#3255](https://github.com/mapbox/mapbox-navigation-ios/pull/3255))
```

1ec5 (approved):

PR #3257, opened 2021-08-13
Let current road label named from navigation status instead of Electronic Horizon

ShanMa1991:
### Description
This pr is to let current road name label get road name from navigation status, not electronic horizon

### Implementation
- [x] remove the Electronic Horizon source for road name label
- [x] get road name label from status and use it for road name label.
- [x] update changeLog.


Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -187,6 +187,13 @@ public extension Notification.Name {
      The user info dictionary contains the key `MapboxNavigationService.NotificationUserInfoKey.locationAuthorizationKey`.
     */
     static let locationAuthorizationDidChange: Notification.Name = .init(rawValue: "LocationAuthorizationDidChange")
+    
+    /**
+     Posted when `RouteController` detects the road name.
+     
+     The user info dictionary contains the keys `RouteController.NotificationUserInfoKey.roadNameKey`.
```
MaximAlien, line 194: Nit:

```suggestion
     The user info dictionary contains the key `RouteController.NotificationUserInfoKey.roadNameKey`.
```

MaximAlien (approved):

CHANGELOG.md
```diff
@@ -124,7 +124,7 @@
...
 * Fixed a thread-safety issue in `UnimplementedLogging` protocol implementation. ([#3024](https://github.com/mapbox/mapbox-navigation-ios/pull/3024))
 * Implemented automatic switching to older navigation tiles versions (if they are present in current `TileStore`) and back. You can subscribe to `Notification.Name.navigationDidSwitchToFallbackVersion` and `Notification.Name.navigationDidSwitchToTargetVersion` notifications to monitor such events. ([#3014](https://github.com/mapbox/mapbox-navigation-ios/pull/3014))
-* Fixed an issue where the current road name label sometimes displayed the name of an intersecting road instead of the current road or blinked in and out. ([#3019](https://github.com/mapbox/mapbox-navigation-ios/pull/3019))
+* Fixed an issue where the current road name label sometimes displayed the name of an intersecting road instead of the current road or blinked in and out with `NavigationStatus.roadName`. ([#3257](https://github.com/mapbox/mapbox-navigation-ios/pull/3257))
```
1ec5, line 127: `NavigationStatus.roadName` is an internal implementation detail, so I removed the reference to it in 416744d746381874c63eadde3ef114ee8c6f4296.

Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -187,6 +187,13 @@ public extension Notification.Name {
      The user info dictionary contains the key `MapboxNavigationService.NotificationUserInfoKey.locationAuthorizationKey`.
     */
     static let locationAuthorizationDidChange: Notification.Name = .init(rawValue: "LocationAuthorizationDidChange")
+    
+    /**
+     Posted when `RouteController` detects the road name.
+     
+     The user info dictionary contains the key `RouteController.NotificationUserInfoKey.roadNameKey`.
+     */
+    static let routeControllerRoadName: Notification.Name = .init(rawValue: "RouteControllerRoadName")
```
1ec5, line 196: A notification name should describe something that has happened. In this case, the event is posted whenever the road name _changes_, so the name should be `currentRoadNameDidChange`. (This is also nice because the notification could be posted by PassiveLocationManager instead.)

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -73,42 +73,29 @@ extension NavigationMapView {
...
-                }
-            }
-        }
-        
         @objc func orientationDidChange(_ notification: Notification) {
             updateMapViewOrnaments()
         }
         
+        @objc func didUpdateRoadNameFromStatus(_ notification: Notification) {
+            guard let roadName = notification.userInfo?[RouteController.NotificationUserInfoKey.roadNameKey] as? String else { return }
```
1ec5, line 95: We should always update `roadNameFromStatus` even if it’s null. Otherwise, if the user travels from a named road to an unnamed road, the previous road’s name could incorrectly persist in the label.

1ec5 (commented):
Sorry for the belated review. There are a couple things that would be nice to take care of as tail work.

Sources/MapboxNavigation/OrnamentsController.swift
ShanMa1991: Thanks for the suggestion, the tail work is addressed in #3266 .

PR #3258, opened 2021-08-13
Remove `Bench` and `BenchTests` targets.

MaximAlien:
### Description

PR removes `Bench` and `BenchTests` targets as they're no longer used.

/cc @mapbox/navigation-ios


Bench/BenchViewController.swift
```diff
@@ -1,108 +0,0 @@
-import UIKit
-import MapboxDirections
-import MapboxNavigation
-import MapboxCoreNavigation
-import TestHelper
```
MaximAlien, line 5: Usage of this dependency makes successful compilation impossible at this point. If we'd like to keep using these targets we'll have to find a way to be able to use `TestHelper` not only in `MapboxNavigation.xcodeproj` or implement helper method in `Bench` target itself.

S2Ler (approved):
I'm ok with this. Even if we will need something like that at some point, we will revisit the approach and implement.

PR #3261, opened 2021-08-16
Hybrid Router

Udumft:
### Description
This PR introduces "Hybrid ~Theory~ Router" usage. This feature adds API to allow making explicitly online/offline/hybrid route requests, selecting between Directions or onboard router engine.

### Implementation
Removed previous `Directions` extension with similar functionality. Tried to make feature similar to Android analog. Added `NavigationRouter` (naming suggestions are welcome) to be responsible for making requests. User is free to instantiate a router for a single request or keep the object alive to make several requests.
Current implementation has not finished parsing refreshing response (TODO).


Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -471,7 +471,7 @@ extension CarPlayManager {
     }
     
     internal func calculate(_ options: RouteOptions, completionHandler: @escaping Directions.RouteCompletionHandler) {
-        directions.calculateWithCache(options: options, completionHandler: completionHandler)
+        NavigationRouter().requestRoutes(options: options, completionHandler: completionHandler)
```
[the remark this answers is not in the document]
Udumft: I wanted to support 2 cases: where user need just 1 request, and when he needs to make several requests in a row. Depending on it, he should (not)retain the router instance. According to Native team, it is not too expensive to instantiate or deallocate such routers, so there's minimal sense to store shared instances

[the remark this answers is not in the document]
S2Ler: Creating CacheHandle is expensive object to create, @mskurydin please confirm. 

Any reasons not to retain router in Navigator?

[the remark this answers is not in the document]
Udumft: Typically, router is needed once every 2 minutes for route refreshing. If this is not enabled, router will be called only at the beginning of the active guidance or when user offroutes. It seems like keeping it alive has no benefits since most of the time it does nothing.

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
...
+    private var settings: NavigationSettings
+    
+    public init(_ source: RouterSource = .hybrid, settings: NavigationSettings = .shared) {
+        self.source = source
+        self.settings = settings
+        
+        let factory = NativeHandlersFactory(tileStorePath: settings.tileStoreConfiguration.navigatorLocation.tileStoreURL?.path ?? "",
+                                            credentials: settings.directions.credentials)
+        self.router = MapboxNavigationNative.RouterFactory.build(for: source.nativeSource,
+                                                                 cache: factory.cacheHandle,
```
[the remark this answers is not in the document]
Udumft: After chatting with NN, I tried to introduce basic caching for `CacheHandle` to ease using this object across our SDK. Unless that caching is done on NN's side, we'll have to have some platform solution.

CHANGELOG.md
```diff
@@ -18,7 +18,6 @@
 * Carthage v0.38 or above is now required for installing this SDK if you use Carthage. ([#3031](https://github.com/mapbox/mapbox-navigation-ios/pull/3031))
 * You can fully build this SDK on Macs with Apple Silicon. ([#3031](https://github.com/mapbox/mapbox-navigation-ios/pull/3031))
 * Removed the optional dependency on MapboxGeocoder.swift. ([#2999](https://github.com/mapbox/mapbox-navigation-ios/pull/2999), [#3183](https://github.com/mapbox/mapbox-navigation-ios/issues/3183])
-* Added a Castilian Spanish localization. ([#3186](https://github.com/mapbox/mapbox-navigation-ios/pull/3186))
```
S2Ler, line 21: Probably incorrect rebase. Changelog is broken.

Sources/MapboxCoreNavigation/CacheHandlerFactory.swift
```diff
@@ -0,0 +1,25 @@
+
+import MapboxNavigationNative
+
+struct CacheHandlerFactory {
```
S2Ler, line 4: Using enum will make it clear that this class isn't supposed to be created (swiftc won't allow that).

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -149,6 +149,17 @@ public class NavigationSettings {
      */
     public static let shared: NavigationSettings = .init()
     
+    /// Returns a reflection of this class excluding the `properties` variable.
+    lazy var properties: [Mirror.Child] = {
```
S2Ler, line 153: It was removed in main. Any reasons to bring this back?

Sources/TestHelper/TestCase.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 import XCTest
-import MapboxDirections
+@testable import MapboxDirections
```
S2Ler, line 3: Looks like bad rebase as well.

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
...
+            case .offline:
+                return .onboard
+            case .hybrid:
+                return .hybrid
+            }
+        }
+    }
+    
+    public private(set) var activeRequests: [RequestId : RoutingRequest] = [:]
+    public private(set) var source: RouterSource
```
S2Ler, line 38: Is RouterSource modified anywhere?

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
...
+    private var router: RouterInterface
```
S2Ler, line 39: Is `router` property modified anywhere? Can be just `let` ?

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
...
+    private var settings: NavigationSettings
```
S2Ler, line 40: Is it modified anywhere? Can be just `let`?

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
+import Foundation
+import MapboxDirections
+import MapboxNavigationNative
+
+public class NavigationRouter {
```
S2Ler, line 5: This class isn't thread-safe, please consider making it usable from any thread. Is there any reason to keep this class non-thread-safe?

As far as I see, only `activeRequests` property is modified. Protecting it with a lock would be sufficient to make this class thread safe, hopefully. NavNative implementation should be thread-safe.

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
...
+        
+        let factory = NativeHandlersFactory(tileStorePath: settings.tileStoreConfiguration.navigatorLocation.tileStoreURL?.path ?? "",
+                                            credentials: settings.directions.credentials)
+        self.router = MapboxNavigationNative.RouterFactory.build(for: source.nativeSource,
+                                                                 cache: factory.cacheHandle,
+                                                                 historyRecorder: factory.historyRecorder)
+    }
+    
+    deinit {
+        print(">>> dealloc")
```
S2Ler, line 54: We have `LeakTest` class which can be used to write a test for NavigationRouter leaking.

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
...
+        router.cancelRequest(forToken: id)
+        activeRequests[id] = nil
+        print(">>> finished id: \(id)")
+    }
+    
+    public func doRequest<ResponseType: Codable>(options: DirectionsOptions,
+                                                 completion: @escaping (Result<ResponseType, DirectionsError>) -> Void) -> RequestId {
+        let directionsUri = settings.directions.url(forCalculating: options).absoluteString
+        var requestId: RequestId!
+        requestId = router.getRouteForDirectionsUri(directionsUri) { [weak self] (result, _) in // mind exposing response origin?
```
S2Ler, line 67: What if instead of retaining `NavigationRouter` in `RoutingRequest` we retain it here with strong `self`? Have you tried that? Would it work?

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
...
+                                           route: routeJSONString) { [weak self] result, _ in
+            guard let self = self else { return }
+            // change to route deserialization
+            do {
+                let json = result.value as? String
+                guard let data = json?.data(using: .utf8) else {
+                    DispatchQueue.main.async {
+                        completionHandler(self.settings.directions.credentials, .failure(.noData))
+                        self.finish(request: requestId)
+                    }
```
S2Ler, line 149: Calling this completion can be refactored with simple local function with signature like:
```swift
func finish(self: NavigationRouter, result: Result<RouteRefreshResponse, DirectionsError>)
```

it will reduce duplication and the likehood of introducing a bug during future evolution.
Passing `self` will make it sure that no unexpected retain cycles introduced. 

Alternative would be to have this function at the type level as the same is needed in the route request above.

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,179 @@
...
+    }
+    
+    fileprivate func finish(request id: RequestId) {
+        router.cancelRequest(forToken: id)
+        activeRequests[id] = nil
+        print(">>> finished id: \(id)")
+    }
+    
+    public func doRequest<ResponseType: Codable>(options: DirectionsOptions,
+                                                 completion: @escaping (Result<ResponseType, DirectionsError>) -> Void) -> RequestId {
```
S2Ler, line 64: Controlling the `completionQueue` that defaults to .main, would be a good customization point.

S2Ler (commented):
Adding some tests are very much welcomed. 

The API is easy to use, thumbs up!

Sources/MapboxCoreNavigation/NavigationRouter.swift
Udumft: `NavigationRouter` is an alternative to our regular `Directions` response wrappers which always call their completions on main thread. I assume that will be also expected true for `NavigationNavigator`, and it will also be mentioned in the docs.

Sources/MapboxCoreNavigation/NavigationRouter.swift
Udumft: Right, I didn't bother about thread-safety so far since it's at draft stage. But since There is no feedback against the approach in general, I've added it now. NN methods use their worker threads, so it should be thread-safe by default, but I also wrapped it's calls for extra safety

Sources/MapboxCoreNavigation/NavigationRouter.swift
Udumft: Thanks for suggestion. I'll be addressing test coverage when finalizing draft stage.

Sources/MapboxCoreNavigation/NavigationRouter.swift
Udumft: Retaining it here does not guarantee it'll work. It may happen that scheduling a response to worker thread takes longer then the rest of this method and the NN router is deinited and so will callback. I think we had this issue with tileset descriptors fetching method before.

Udumft: Also, it seems to me that tying it's lifetime to a request handle makes sense and is meant to be.

Sources/MapboxCoreNavigation/NavigationRouter.swift
Udumft: Done. Thanks for suggestion!

CHANGELOG.md
Udumft: Resolved

Sources/MapboxCoreNavigation/CacheHandlerFactory.swift
Udumft: Good catch, done

Sources/MapboxCoreNavigation/NavigationSettings.swift
Udumft: Was restored due to error rebase, fixed now.

Sources/MapboxCoreNavigation/NavigationRouter.swift
Udumft: Right, It should be a constant as well as other properties below.

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,321 @@
...
+}
+
+extension DirectionsProfileIdentifier {
+    var nativeProfile: RoutingProfile {
+        return RoutingProfile(profile: rawValue)
+    }
+}
+
+extension URL {
+    func removingSKU() -> URL {
```
Udumft, line 315: Removing the `SKU` token, as NN automatically appends it on their side, which may result in conflict.

S2Ler: Do you have an integration test for this? I'm worried that we break encapsulation here by using magic constant "sku". Can we move this logic to Directions library?

Udumft: I don't think it should be a publicly available Directions feature, as it is a part of internal logic communicating w/ NN. Agreed on magic constants usage, but don't know how to avoid it in this case as ideally it should be shared with Directions too.

S2Ler:
I put some thoughts, but we need to flesh out breaking changes problem + discuss if we can keep allowing using Directions if devs would like to.

Sources/MapboxCoreNavigation/NavigationRouter.swift
1ec5: Can it be shared through the environment, the same way we used to share the SKU token with MapboxDirections?

Udumft: I think we could. It would require internal changes on Directions side too, so maybe we could do it as a tailwork?

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,337 @@
+@_implementationOnly import MapboxCommon_Private
+import MapboxDirections
+import MapboxNavigationNative
+
+/**
+ Provides alternative access to routing API.
+ 
+ Use this class instead `Directions` requests wrapper to request new routes or refresh an existing one. Depending on `RouterSource`, `NavigationRouter` will use online and/or onboard routing engines. This may be used when designing purely online or offline apps, or when you need to provide best possible service regardless of internet collection.
+ */
+public class NavigationRouter: NavigationProvider {
```
1ec5, line 10: NavigationRouter is unrelated to the Router protocol, so we need to find a different name. I’ve wanted to rename the Router protocol to something less confusing for a long time, but I think further overloading the word in the meantime will only increase confusion.

Sources/MapboxCoreNavigation/NavigationProvider.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+import MapboxDirections
+
+/**
+ Protocol which defines a type which can be used for fetching or refreshing routes.
+ 
+ SDK provides conformance to this protocol for `Directions` and `NavigationRouter`.
+ */
+public protocol NavigationProvider {
```
1ec5, line 9: The name of this protocol says that an implementation of this protocol “provides navigation”. That seems a bit too abstract. I’d suggest either:

* `NavigationServiceDataSource`: The data source pattern is familiar to Cocoa developers.
* `RouteCalculating`: Naming protocols as participles seems to be in vogue these days among Swift developers.

The ideal name would actually be “Routing”, since protocol names shouldn’t be nouns if we can help it, but that’s still too close to Router for comfort.

I’m assuming there’s a gnarly reason we aren’t just folding these methods into the existing NavigationServiceDelegate protocol, but if not, then that would be a lightweight option too.

Sources/MapboxCoreNavigation/Directions+NavigationProviderRequest.swift
```diff
@@ -0,0 +1,73 @@
+import Foundation
+import MapboxDirections
+
+extension URLSessionDataTask: NavigationProviderRequest {
+    public var id: UInt64 {
```
1ec5, line 5: This extends a commonly used Swift standard library type with a commonly used property name (which happens to be part of the Identifiable protocol). Can we name this property something else less prone to conflicts?

Sources/MapboxCoreNavigation/Directions+NavigationProviderRequest.swift
```diff
@@ -0,0 +1,73 @@
...
+        copy.routes = copy.routes?.map { $0.copy(with: options) }
+        return copy
+    }
+}
+
+extension Route {
+    func copy(with options: DirectionsOptions) -> Route {
+        let encoder = JSONEncoder()
+        encoder.userInfo[.options] = options
+        let encoded = try! encoder.encode(self)
```
1ec5, line 67: Avoid these implicitly unwrapped optionals, especially because the DirectionsOptions object could be a subclass of RouteOptions defined by the developer, who may need to catch an encoding error. For consistency with #3484, this method should be throwable.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -25,6 +25,16 @@ public struct IndexedRouteResponse {
      */
     public let routeIndex: Int
     
+    /**
+     Returns a route from the `routeResponse` under given `routeIndex` if possible.
+     */
+    public var selectedRoute: Route? {
```
1ec5, line 31: Just `route` is sufficiently clear, or maybe `currentRoute` for consistency with `RouteProgress.currentLeg`. “Selected” implies that there’s a selection action built into the SDK, but that’s still the responsibility of the developer in general.

Sources/MapboxCoreNavigation/NavigationRouter.swift
```diff
@@ -0,0 +1,337 @@
...
+     Valid only for the same instance of `NavigationRouter` that issued it.
+     */
+    public typealias RequestId = UInt64
+    
+    /**
+     A request handler for the ongoing router action.
+     
+     You can use this instance to cancel ongoing task if needed. Retaining this handler will keep related `NavigationRouter` from deallocating.
+     */
+    public struct RoutingRequest: NavigationProviderRequest {
```
1ec5, line 23: Can we rename this to just Request, since it’s already nested under NavigationRouter?

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -18,9 +18,14 @@ public protocol NavigationService: CLLocationManagerDelegate, RouterDataSource,
...
     /**
-     A reference to a MapboxDirections service. Used for rerouting.
+     `NavigationProvider`, used to create route.
+     */
+    var routingProvider: NavigationProvider { get }
+    
+    /**
+     Credentials data, used to authorize server requests.
      */
-    var directions: Directions { get }
```
1ec5, line 23: The removal of this property spurs a bunch of changes to MapboxNavigation, such as in NavigationViewController and JunctionView, which makes me a bit concerned about backwards compatibility and thus eligibility for v2._x_. What if we retain this property but mark it as deprecated and document that it is ignored for rerouting purposes? We can introduce a parallel property of the new, more general type to use as a stand-in.

Sources/MapboxCoreNavigation/CacheHandlerFactory.swift
```diff
@@ -0,0 +1,73 @@
...
+    var credentials: DirectionsCredentials { get }
+    var tilesVersion: String { get }
+    var historyDirectoryURL: URL? { get }
+    var targetVersion: String? { get }
+    var configFactoryType: ConfigFactory.Type { get }
+}
+
+extension NativeHandlersFactory: CacheHandlerData { }
+
+enum CacheHandlerFactory {
```
1ec5, line 15: Can you explain (or document) why we need to use the factory pattern here? Could we fold this functionality into an extension of CacheHandle and use file-level variables for things like `cachedHandle`?

Sources/MapboxCoreNavigation/Directions+NavigationProviderRequest.swift
```diff
@@ -0,0 +1,73 @@
...
+        return calculate(options, completionHandler: completionHandler)
+    }
+    
+    @discardableResult public func requestRoutes(options: MatchOptions, completionHandler: @escaping MatchCompletionHandler) -> NavigationProviderRequest? {
+        return calculate(options, completionHandler: completionHandler)
+    }
+    
+    @discardableResult public func refreshRoute(indexedRouteResponse: IndexedRouteResponse, fromLegAtIndex: UInt32, completionHandler: @escaping RouteCompletionHandler) -> NavigationProviderRequest? {
+        guard case let .route(routeOptions) = indexedRouteResponse.routeResponse.options else {
+            preconditionFailure("Invalid route data passed for refreshing.")
```
1ec5, line 22: This only happens if the options object is a MatchOptions, right? Let’s say something about that in the error message.

1ec5 (commented):
I’ve only gotten a chance to take a cursory look through the PR, but here’s some high-level feedback that hopefully gives some ideas about how we can maintain backwards compatibility and avoid confusion.

Sources/MapboxCoreNavigation/NavigationService.swift
Udumft: I think leaving such dangling property may be even more confusing. Compiler won't complain about something missing, but any interactions with it will be just silently ignored which will most likely result in bugs in user apps.
On the other hand, replacing it with new property allows straightforward migration. If user stays on default settings for routing, he'll just need to cast the object to `Directions` and continue his previous flow.

Sources/MapboxCoreNavigation/NavigationProvider.swift
Udumft: Yeah, naming is hard :)
My current idea is that such objects "provide routing results" so I renamed the protocol. And for our "suggested" implementation I decided to do the same as we did with [Mapbox]NavigationService...

Sources/MapboxCoreNavigation/CacheHandlerFactory.swift
Udumft: Added code doc. Ideally, this will be eventually removed when NN implements valid caching the CacheHandler on their side.

Udumft:
Removing "Do not merge" labels as 2.1 release branch is cut and this PR targets 'main' ( 2.2 ) version

Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
```diff
@@ -0,0 +1,79 @@
...
+    @discardableResult public func refreshRoute(indexedRouteResponse: IndexedRouteResponse, fromLegAtIndex: UInt32, completionHandler: @escaping RouteCompletionHandler) -> NavigationProviderRequest? {
+        guard case let .route(routeOptions) = indexedRouteResponse.routeResponse.options else {
+            preconditionFailure("Invalid route data passed for refreshing. Expected `RouteResponse` containing `.route` `ResponseOptions` but got `.match`.")
+        }
+        
+        let session = (options: routeOptions as DirectionsOptions,
+                       credentials: self.credentials)
+        
+        guard let responseIdentifier = indexedRouteResponse.routeResponse.identifier else {
+            defer {
```
S2Ler, line 29: Defer looks redundant here

Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
```diff
@@ -0,0 +1,79 @@
...
+                completionHandler(session, .failure(.noData))
+            }
+            return nil
+        }
+        
+        return refreshRoute(responseIdentifier: responseIdentifier,
+                            routeIndex: indexedRouteResponse.routeIndex,
+                     fromLegAtIndex: Int(fromLegAtIndex),
```
S2Ler, line 37: Broken formatting.

S2Ler (commented):
Forgot to submit review previously... Here are some comments.

Sources/MapboxCoreNavigation/CacheHandlerFactory.swift
```diff
@@ -0,0 +1,73 @@
...
+        private static func optionalsAreNotEqual<T: Comparable>(_ lhs: T?, _ rhs: T?) -> Bool{
+            if let firstVal = lhs, let secondVal = rhs {
+                return firstVal != secondVal
+            }
+            else {
+                return lhs != nil || rhs != nil
+           }
+        }
+        
+        static func != (lhs: CacheKey, rhs: CacheHandlerData) -> Bool {
```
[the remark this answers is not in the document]
Udumft: I am not sure why I needed `optionalsAreNotEqual` exactly, probably a leftover from testing [mis]matching keys. `Equatable` protocol is not ideal here because we need to compare different types (`CacheKey` vs. `CacheHandlerData`). Extending `CacheHandlerData` with `Equatable` is not possible because it introduces a dependency to `Self` type and won't help comparing different types.

Sources/MapboxCoreNavigation/CacheHandlerFactory.swift
```diff
@@ -0,0 +1,73 @@
...
+                lhs.credentials != rhs.credentials ||
+                optionalsAreNotEqual(lhs.tilesVersion, rhs.tilesVersion) ||
+                optionalsAreNotEqual(lhs.historyDirectoryURL?.absoluteString, rhs.historyDirectoryURL?.absoluteString) ||
+                optionalsAreNotEqual(lhs.targetVersion, rhs.targetVersion) ||
+                lhs.configFactoryType != rhs.configFactoryType
+        }
+    }
+    
+    private static var key: CacheKey? = nil
+    private static var cachedHandle: CacheHandle!
```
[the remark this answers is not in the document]
Udumft: Force unwrap is required either here or when returning this value from `getHandler(for: config: historyRecorder: cacheData:)`. It cannot be non-optional because we cannot initialise a `default` value, but only create it on demand. This should be a safe force-unwrapping since it's private and always gets initialised before usage

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -219,8 +234,8 @@ extension InternalRouter where Self: Router {
             guard case let .success(response) = result, let self = self else {
                 return
             }
-            
-            self.routeProgress.refreshRoute(with: response.route, at: location)
+            self.indexedRouteResponse = .init(routeResponse: response, routeIndex: self.indexedRouteResponse.routeIndex)
+            self.routeProgress.refreshRoute(with: self.indexedRouteResponse.currentRoute!, at: location)
```
S2Ler, line 238: nit: I think precondition with a clear message would be better here in case something goes wrong.

Sources/MapboxCoreNavigation/RoutingProvider.swift
```diff
@@ -0,0 +1,59 @@
...
+public protocol RoutingProvider {
+    
+    /**
+     Routing caluclation method.
+     
+     - parameter options: A `RouteOptions` object specifying the requirements for the resulting routes.
+     - parameter completionHandler: The closure (block) to call with the resulting routes. This closure is executed on the application’s main thread.
+     - returns: Related request. If, while waiting for the completion handler to execute, you no longer want the resulting routes, cancel corresponding task using this handle.
+     */
+    @discardableResult func calculateRoutes(options: RouteOptions,
```
S2Ler, line 18: Looks like `MapboxRoutingProvider` also has documentation comments. Does it make sense to merge them to the base protocol? Xcode automatically provides documentation from from the protocol if you implement the protocol.

Sources/MapboxCoreNavigation/MapboxRoutingProvider.swift
```diff
@@ -0,0 +1,338 @@
...
+         */
+        public func cancel() {
+            router.router.cancelRequest(forToken: requestIdentifier)
+        }
+    }
+    
+    /**
+     Defines source of routing engine to be used for requests.
+     */
+    public enum Source {
```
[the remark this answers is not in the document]
Udumft: I don't see any new source types in the future. Given the domain of that feature, I believe it will be a significant change if we invents something new here

Sources/MapboxCoreNavigation/MapboxRoutingProvider.swift
```diff
@@ -0,0 +1,338 @@
...
+        
+        private enum CodingKeys: CodingKey {
+            case code, message, error
+        }
+    }
+    
+    fileprivate func parseResponse<ResponseType: Codable>(requestId: RequestId, userInfo: [CodingUserInfoKey : Any], result: Expected<AnyObject, AnyObject>, completion: @escaping (Result<ResponseType, DirectionsError>) -> Void) {
+        do {
+            let json = result.value as? String
+            guard let data = json?.data(using: .utf8) else {
```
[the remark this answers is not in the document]
Udumft: Sorry, I didn't get your suggestion

Sources/MapboxCoreNavigation/RoutingProvider.swift
Udumft: Documentation differs for `RoutingProvider` and `MapboxRoutingProvider`. Protocol description is more abstract, defining the purpose of each method, while the latter one provides more insights on how the method does it's job.

Sources/MapboxCoreNavigation/MapboxRoutingProvider.swift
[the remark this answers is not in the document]
S2Ler:
```swift
guard let json = result.value as? String else {...}
let data = Data(json.utf8)
...
```

Sources/MapboxCoreNavigation/MapboxRoutingProvider.swift
```diff
@@ -0,0 +1,338 @@
...
+        /**
+         Fetch data online only
+         
+         Such `MapboxRoutingProvider` is equivalent of using bare `Directions` wrapper.
+         */
+        case online
+        /**
+         Use online data only
+         
+         In order for such `MapboxRoutingProvider` to function properly, proper navigation data should be available onboard. `.offline` router will not be able to refresh routes.
```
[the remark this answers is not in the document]
Udumft: Code comment corrected. I don't quite understand the issue with tenses. Is it not grammatical?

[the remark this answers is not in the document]
S2Ler: It is correct, but maybe not that good to use it here.

Sources/MapboxCoreNavigation/MapboxRoutingProvider.swift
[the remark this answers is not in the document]
Udumft: Why is it preferable?

Sources/MapboxCoreNavigation/MapboxRoutingProvider.swift
```diff
@@ -0,0 +1,338 @@
...
+    
+    //MARK: - Initialization
+    
+    /**
+     Initializes new `MapboxRoutingProvider`.
+     
+     - parameter source: routing engine source to use.
+     - parameter settings: settings object, used to get credentials and cache configuration.
+     */
+    public init(_ source: Source = .hybrid, settings: NavigationSettings = .shared) {
```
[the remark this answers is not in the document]
Udumft: Settings are expected as a parameter to not introduce implicit dependency on it. Plus it is an easy option to enable more flexibility in case it's needed.

Sources/MapboxCoreNavigation/MapboxRoutingProvider.swift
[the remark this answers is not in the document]
S2Ler: `Data(json.utf8)` isn't optional. 
`data(using: .utf8)` is optional. 

It is more clear that this code can fail because result value is of incorrect type. In your variant it is hidden, but the one who knows that data method returns nullable, it is not. 

Just noticed, that you could use `Result+Expected.swift` for parsing NavNative's Expected type.

Sources/MapboxCoreNavigation/CacheHandlerFactory.swift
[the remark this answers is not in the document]
S2Ler: ok, although I don't agree that force unwrap is needed here, I'm not terribly agains it.

Sources/MapboxCoreNavigation/MapboxRoutingProvider.swift
```diff
@@ -0,0 +1,344 @@
...
+        self.source = source
+        self.settings = settings
+        
+        let factory = NativeHandlersFactory(tileStorePath: settings.tileStoreConfiguration.navigatorLocation.tileStoreURL?.path ?? "",
+                                            credentials: settings.directions.credentials)
+        self.router = RouterFactory.build(for: source.nativeSource,
+                                          cache: factory.cacheHandle,
+                                          historyRecorder: factory.historyRecorder)
```
S2Ler, line 27: I found the following issues with this approach:
- `NativeHandlersFactory` is created without the potentially adjusted history URL. In the example app, it leads to two alive CacheHandles. 
- If the above is fixed, there will still be an issue with the Router using a different instance of the history record handler. 

I don't think we are supposed to have more than one history recorder unless it is an artificial use case. 

So we shall reuse the history recorder from `Navigator` and, in the future, decouple HistoryRecorder from Navigator/PassiveLocationProvider/RouteController.

Udumft: We assume that settings are configured before any call to SDK entities is done. Due to `CacheHandlerFactory` `CacheHandle` instance should be reused between `MapboxRoutingProvider` and `Navigator`. But history recorder will be created twice, that's true. Not sure how critical it is, and not sure what's the best way to mitigate this. Referring to `Navigator` here is not an option since it will trigger `MBNNNavigator` initialization which is not desired.

S2Ler: > We assume that settings are configured before any call to SDK entities is done. Due to CacheHandlerFactory CacheHandle instance should be reused between MapboxRoutingProvider and Navigator. 

Please try Example app, it doesn't seem to work as you describe. 

> Referring to Navigator here is not an option since it will trigger MBNNNavigator initialization which is not desired.

Well, then we need similar approach to CacheHandlerFactory. Or change it completely to not only cache cachehandle, but other relevant instances as well.

Udumft: Indeed, there was a difference in cache key form Routing Provider and Navigator, I think I fixed it by adding more parameters to CacheHandlerFactory initialization. At least it should work if `Navigator`'s parameters are not modified before launch.
I also added same caching mechanism for `HistoryRecorder`. I would prefer to solve this issue in a different way, but I don't know how, and it is outside of this task's scope. Currently, only `CacheHandle` and `HistoryRecorder` instances are used in multiple places, so I've added caching only for these 2. With `HandlerFactory` it should be trivial to add caching to any other required entity though.

S2Ler: Let's resolve this thread, although it is not clear what will happen during switch to offline/online for Navigator.

S2Ler (approved):

Issue #3263, opened 2021-08-16
Crash occurs when stopping animation, which has not started yet.


MaximAlien:
**Mapbox Navigation SDK version:** `2.0.0-beta.22`

Crash occurs on Maps SDK side when stopping animation, which has not started yet. For example:

https://github.com/mapbox/mapbox-navigation-ios/blob/851bcc566088a2eecec84da0aaa4ebe443b5690f/Sources/MapboxNavigation/NavigationCameraStateTransition.swift#L221-L234

doesn't verify whether animation is currently running.

PR #3264, opened 2021-08-16
Add `testRunningAnimators`. Fix crash in `NavigationCameraStateTransition.stopAnimators()`.

MaximAlien:
### Description

Closing #3263.

### DIFF
--- a/Sources/MapboxNavigation/NavigationCameraStateTransition.swift
+++ b/Sources/MapboxNavigation/NavigationCameraStateTransition.swift
@@ -54,6 +54,7 @@ public class NavigationCameraStateTransition: CameraStateTransition {
             camera.pitch = 0.0
             if let midPointZoom = camera.zoom {
                 let cameraOptions = CameraOptions(center: location,
+                                                  padding: padding,
                                                   anchor: screenCenterPoint,
                                                   zoom: CGFloat(midPointZoom),
                                                   bearing: bearing,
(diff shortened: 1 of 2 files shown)


bamx23 (approved):

S2Ler (approved):

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
```diff
@@ -228,8 +230,10 @@ public class NavigationCameraStateTransition: CameraStateTransition {
             animatorPadding
         ]
         
-        animators.forEach {
-            $0?.stopAnimation()
+        animators.compactMap({ $0 }).forEach {
+            if $0.isRunning {
+                $0.stopAnimation()
```
1ec5, line 235: @macdrevx suggests `CameraManager.cancelAnimations()`, which might be more elegant.

1ec5 (approved):

Sources/MapboxNavigation/NavigationCameraStateTransition.swift
MaximAlien: There is a chance that `CameraManager.cancelAnimations()` will stop other animators, which are not related to navigation camera. So for now I'd rather stop only animations, which were created in scope of `NavigationCameraStateTransition.swift`.

PR #3266, opened 2021-08-17
Fix the tail work of the road name label based on status.

ShanMa1991:
### Description
This pr is to address some comments in #3257 

### Implementation
- [x] Replace the notification key name of `routeControllerRoadName` with `currentRoadNameDidChange`.
- [x]  Avoid the unnamed road name issue.

### DIFF
--- a/Sources/MapboxCoreNavigation/CoreConstants.swift
+++ b/Sources/MapboxCoreNavigation/CoreConstants.swift
@@ -193,7 +193,7 @@ public extension Notification.Name {
      
      The user info dictionary contains the key `RouteController.NotificationUserInfoKey.roadNameKey`.
      */
-    static let routeControllerRoadName: Notification.Name = .init(rawValue: "RouteControllerRoadName")
+    static let currentRoadNameDidChange: Notification.Name = .init(rawValue: "CurrentRoadNameDidChange")
  
 }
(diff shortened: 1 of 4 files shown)


Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -193,7 +193,7 @@ public extension Notification.Name {
      
      The user info dictionary contains the key `RouteController.NotificationUserInfoKey.roadNameKey`.
      */
-    static let routeControllerRoadName: Notification.Name = .init(rawValue: "RouteControllerRoadName")
+    static let currentRoadNameDidChange: Notification.Name = .init(rawValue: "CurrentRoadNameDidChange")
```
1ec5, line 196: Thanks, perhaps the documentation comment could also be updated to more explicitly refer to the road name changing.

1ec5 (approved):

Sources/MapboxCoreNavigation/CoreConstants.swift
ShanMa1991: Thanks for the suggestion, I added this part in the changeLog instead, the documentation would be aligned with other `Notification.Name`.

PR #3265, opened 2021-08-17
Disable filtering by navigation status origin

S2Ler:
Fixed #3260

We are not supposed to filter out navigation statuses as per NavNative team. Android doesn't filter them out either. 
If it breaking any functionality, we must fix it without filtering out all the events.

PR #3269, opened 2021-08-18
Remove unnecessary Navigator status pulling from updateRouteLeg method

S2Ler:
After navigator.changeRouteLeg, a new navigator status will be generated with origin equal to changeRouteLeg where we update indexes anyway.

Related comment: https://github.com/mapbox/mapbox-navigation-ios/pull/3265#discussion_r690553574

### DIFF
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -227,9 +227,7 @@ open class RouteController: NSObject {
     /// updateRouteLeg is used to notify nav-native of the developer changing the active route-leg.
     private func updateRouteLeg(to value: Int) {
         let legIndex = UInt32(value)
-        if navigator.changeRouteLeg(forRoute: 0, leg: legIndex), let status = navigator.getStatus() {
-            updateIndexes(status: status, progress: routeProgress)
-        }
+        navigator.changeRouteLeg(forRoute: 0, leg: legIndex)
     }
     
     public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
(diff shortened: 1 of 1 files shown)


Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -225,11 +225,9 @@ open class RouteController: NSObject {
...
     
     /// updateRouteLeg is used to notify nav-native of the developer changing the active route-leg.
-    private func updateRouteLeg(to value: Int) {
-        let legIndex = UInt32(value)
-        if navigator.changeRouteLeg(forRoute: 0, leg: legIndex), let status = navigator.getStatus() {
-            updateIndexes(status: status, progress: routeProgress)
-        }
+    private func updateRouteLeg(to legIndex: Int) {
+        precondition(legIndex >= 0, "Leg index can't be negative")
+        navigator.changeRouteLeg(forRoute: 0, leg: UInt32(legIndex))
```
1ec5, line 230: Even if the navigator responds to the leg change and updates the navigator status very quickly and synchronously, RouteController responds to the status update asynchronously. This could violate the developer’s expectation, because `updateRouteLeg(to:)` is called from the public `Router.advanceLegIndex()` method, which looks like a synchronous method to the developer. The developer may be following up with operations that depend on the status being updated, for example to update UI to reflect the new leg.

To set the right expectations, I think `Router.advanceLegIndex()` will need to be renamed `Router.advanceLegIndex(completionHandler:)`. Alternatively, we could leave the redundant `getStatus()` in place if MapboxNavigationNative considers it harmless.

1ec5:
#3342 removes the synchronous `getStatus()` call. The only other thing in this PR is the assertion that the leg index is nonnegative. However, that’s more of a precaution than an expected failure mode, since the developer no longer has the option to set an arbitrary leg index.

Issue #3272, opened 2021-08-18
Improve navigation camera transition to overview state.


MaximAlien:
**Mapbox Navigation SDK version:** `main`

Currently navigation camera for transition to overview mode uses bearing, which is different from bearing, which is used for camera updates in overview mode. This leads to such transition:

https://user-images.githubusercontent.com/1496498/129955930-7e7a83ab-58cb-42cf-8e25-dfee3e8b2289.mp4

PR #3273, opened 2021-08-18
Do not use bearing when navigation camera transitions to the overview mode.

MaximAlien:
### Description

Closing #3272.

New behavior:

https://user-images.githubusercontent.com/1496498/129960068-31d168c0-db36-4a02-a1b3-935fd7f8e9db.mp4


1ec5 (approved):
This would be consistent with v1._x_ behavior. 👍

PR #3280, opened 2021-08-19
Add python script for triggering metrics

azarovalex:
### Description
This PR adds a separate Python script to trigger the metrics pipeline.

### DIFF
--- a/scripts/trigger-metrics.py
+++ b/scripts/trigger-metrics.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+
+import os
+import json
+import requests
+import sys
+
+def TriggerPipeline(token, commit, ci_ref):
+    url = "https://circleci.com/api/v2/project/github/mapbox/mobile-metrics/pipeline"
+
+    headers = {
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
+
+    data = {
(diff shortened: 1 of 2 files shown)


scripts/trigger-metrics.py
```diff
@@ -0,0 +1,52 @@
...
+          "run_ios_navigation_benchmark": True,
+          "target_branch": commit,  
+          "ci_ref": ci_ref,
+        }
+    }
+
+    # Branch in mobile-metrics repo if you want to trigger a custom pipeline
+    # data["branch"] = "test"
+
+    response = requests.post(url, auth=(token, ""), headers=headers, json=data)
```
Udumft, line 27: I am not too familiar with Python, but will this correctly send `token=${MOBILE_METRICS_TOKEN}` parameter? I see there `""` which confuses me.

Udumft (approved):
LGTM unless my question below is actual

S2Ler (approved):

scripts/trigger-metrics.py
azarovalex: https://circleci.com/docs/api/v2/
> The username should be set as the circle-token value, and the password should be left blank

This code works for sure, just a strange CircleCI API.

PR #3284, opened 2021-08-20
Replace SHORT_VERSION in generated docs.

### LINKED ISSUE #3278: Jazzy docs generator doesn't parse sdk version number
Check the landing page at https://docs.mapbox.com/ios/navigation/api/2.0.0-beta.23/

S2Ler:
Fixes #3278

Thanks @1ec5 for finding [the cause of the issue](https://github.com/mapbox/mapbox-navigation-ios/pull/3026#discussion_r692213730).

Screenshot 2021-08-20 at 09 07 22

### DIFF
--- a/scripts/document.sh
+++ b/scripts/document.sh
@@ -40,6 +40,7 @@ mkdir -p /tmp/mbnavigation/
 README=/tmp/mbnavigation/README.md
 cp docs/cover.md "${README}"
 perl -pi -e "s/\\$\\{MINOR_VERSION\\}/${MINOR_VERSION}/" "${README}"
+perl -pi -e "s/\\$\\{SHORT_VERSION\\}/${SHORT_VERSION}/" "${README}"
 # http://stackoverflow.com/a/4858011/4585461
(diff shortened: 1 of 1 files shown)


scripts/document.sh
```diff
@@ -40,6 +40,7 @@ mkdir -p /tmp/mbnavigation/
 README=/tmp/mbnavigation/README.md
 cp docs/cover.md "${README}"
 perl -pi -e "s/\\$\\{MINOR_VERSION\\}/${MINOR_VERSION}/" "${README}"
+perl -pi -e "s/\\$\\{SHORT_VERSION\\}/${SHORT_VERSION}/" "${README}"
```
azarovalex, line 43: I wonder if jazzy supports this, looks like an obvious use case.

bamx23: It seems that `{{module_version}}` in the templates should work as we provide `${SHORT_VERSION}` to `--module-version`. But not sure if it applies to the README file.
The custom variables are not supported for themes: _(check issue 566 in Jazzy)_

bamx23 (approved):

PR #3291, opened 2021-08-24
Add minimum distance limit for vanishing route line effect

ShanMa1991:
### Description
This pr is to remove redundant code, improve the performance of vanishing route line feature with adding minimum distance limit.

### Implementation
- [x] remove redundant code, because right now we only use refresh to update congestion level.
- [x] avoid distance calculation if the user hasn't arrive at destination.
- [x] add distance limit check before update the route line layer, the minimum is 1 pixel according to current latitude and zoom level.


Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -335,8 +335,10 @@ public class CarPlayNavigationViewController: UIViewController {
         }
         
         if routeLineTracksTraversal {
-            if routeProgress.isFinalLeg && routeProgress.currentLegProgress.distanceRemaining <= 0.0 {
-                navigationMapView?.removeRoutes()
+            if routeProgress.isFinalLeg && routeProgress.currentLegProgress.userHasArrivedAtWaypoint {
+                if routeProgress.currentLegProgress.distanceRemaining <= 0.0 {
```
Udumft, line 339: You could have all these checks in one `if` in correct order.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -163,8 +163,8 @@ open class NavigationMapView: UIView {
     var routeLineGranularDistances: RouteLineGranularDistances?
     var routeRemainingDistancesIndex: Int?
     var fractionTraveled: Double = 0.0
+    var preLocation: CLLocationCoordinate2D?
```
Udumft, line 166: I would suggest a bit more descriptive name. It is hard to tell right away what is `pre` prefix here and what this variable is for

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
Udumft: I see this check repeats multiple times. Could we extract it into `RouteProgress` method or so?

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -178,7 +178,15 @@ extension NavigationMapView {
...
      */
     public func travelAlongRouteLine(to coordinate: CLLocationCoordinate2D?) {
-        updateFractionTraveled(coordinate: coordinate)
+        guard preLocation != coordinate,
+              let preCoordinate = preLocation,
+              let currentCoordinate = coordinate else { return }
+        
+        let distance = preCoordinate.distance(to: currentCoordinate)
+        let meterPerPixel = getMetersPerPixelAtLatitude(currentCoordinate.latitude, Double(mapView.cameraState.zoom))
+        guard distance >= meterPerPixel else { return }
```
Udumft, line 187: Is it possible that due to this `guard` `fractionTraveled` will not be updated on low zoom levels and it can malfunction completing the route (i.e. `fractionTraveled >= 1` check)? What if user has to travel less distance to his destination than fits into 1 pixel?

Issue #3281, opened 2021-08-19
Make `NavigationViewportDataSourceOptions` more restrictive.


MaximAlien:
**Mapbox Navigation SDK version:** `main`

At this point `NavigationViewportDataSourceOptions` contains various options, which allows to control camera behavior. Types of certain options are pretty tolerant to values, which can be set by the end user. For example `FollowingCameraOptions.maximumZoomLevel` uses `Double`, which might lead to some undefined behavior if value is not set within 0 to 22 zoom level range.

In scope of this ticket such APIs should be revised and updated to be more restrictive.

1ec5:
Also make this type and its initializer `open`.

PR #3293, opened 2021-08-24
Make navigation camera options more restrictive.

MaximAlien:
### Description

Change `minimumZoomLevel` and `maximumZoomLevel` with closed range in `FollowingCameraOptions`.

Closing #3281.


Udumft (approved):
Does it make sense to limit various 'angle' based properties to 0..<360 like `FollowingCameraOptions.defaultPitch` or maybe just changing it's type to `CLLocationDirection`?

bamx23 (approved):

bamx23:
@MaximAlien as you are going to close #3281, could you also make `init` of `NavigationViewportDataSourceOptions` public in this PR?

MaximAlien:
> Does it make sense to limit various 'angle' based properties to 0..<360 like `FollowingCameraOptions.defaultPitch` or maybe just changing it's type to `CLLocationDirection`?

I think that `CLLocationDirection` type is not meant to be used for pitch. Based on Maps SDK logic I think we should verify that pitch is in range from `0.0` to `85.0`.

PR #3295, opened 2021-08-25
Support instant puck change without lag.

### LINKED ISSUE #3285: Difficulty toggling between puck types when using NavigationMapView injection
**Mapbox Navigation SDK version:** 2.0.0-beta.22

When using the NavigationMapView injection API, it is difficult to change the puck styling between `UserPuckCourseView` and `Puck2D` when transitioning to and from turn-by-turn navigation. Being able to define puck types for each experience and the swap occurring internally would be ideal. 

### Workaround

For now, I have two functions on an extension of `NavigationMapView` that I've defined (see code below). Initially I tried just setting `userLocationStyle`, but updates to this property will only take on the next `.styleLoaded` event which may not occur during the transition to turn-by-turn navigation if using NavigationMapView injection.

#### Usage

Immediately before I create my instance of NavigationViewController, I will call `  navigationMapView.updatePuckToUserCourseConfiguration()`. After navigation is finished, I call `navigationMapView.updatePuckTo2DConfiguration()`.

#### The Extension

```swift
extension NavigationMapView {
    /// Updates the puck indicator on the map to the style expected during turn-by-turn navigation
    func updatePuckToUserCourseConfiguration() {
        var locationOptions = mapView.location.options
        locationOptions.puckType = .none
        mapView.location.options = locationOptions

        let courseView = UserPuckCourseView(frame: CGRect(x: 0, y: 0, width: 75, height: 75))
        userLocationStyle = .courseView(courseView)
        userCourseView = courseView
        userCourseView.isHidden = false

        if let location = mapView.location.latestLocation?.coordinate {
            courseView.update(
                location: CLLocation(location),
                pitch: 0,
                direction: 0,
                animated: false,
                navigationCameraState: .transitionToFollowing
            )
        }
    }

    /// Updates the puck indicator on the map to the style expected when displaying a user's location in overview/freeform mode
    func updatePuckTo2DConfiguration() {
        var locationOptions = mapView.location.options
        locationOptions.puckType = .puck2D(Puck2DConfiguration())
        mapView.location.options = locationOptions

        userLocationStyle = .puck2D(configuration: Puck2DConfiguration())
        userCourseView.isHidden = true
    }
}

```

### Demo

Here's a video of the workaround in action on the Straightaway app

https://user-images.githubusercontent.com/630204/130249361-3dbfbcfe-ad08-429f-a004-9d9fe9106ecb.mp4

ShanMa1991:
### Description
This pr is to fix #3285 by removing the style loaded lag and to change the user location layer instantly.

### Implementation
Right now the map puck change doesn't need to wait for the style loaded.

### Screenshots or Gifs
This is what I use the resume button to toggle between the puck2D and the userCourseView. 

puckChange

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -253,21 +253,17 @@ open class NavigationMapView: UIView {
     }
     
     func setupUserLocation() {
-        mapView.mapboxMap.onNext(.styleLoaded) { [weak self] _ in
-            guard let self = self else { return }
-            switch self.userLocationStyle {
-            case .courseView((let courseView)):
-                self.mapView.location.options.puckType = nil
-                self.userCourseView = courseView
-                self.userCourseView.isHidden = false
-            case .puck2D(configuration: var configuration):
-                self.userCourseView.isHidden = true
-                self.mapView.location.options.puckType = .puck2D(configuration ?? Puck2DConfiguration())
-                configuration = configuration ?? Puck2DConfiguration()
-            case .puck3D(configuration: let configuration):
-                self.userCourseView.isHidden = true
-                self.mapView.location.options.puckType = .puck3D(configuration)
-            }
+        switch userLocationStyle {
+        case .courseView((let courseView)):
+            mapView.location.options.puckType = nil
+            userCourseView = courseView
+            userCourseView.isHidden = false
+        case .puck2D(configuration: let configuration):
+            userCourseView.isHidden = true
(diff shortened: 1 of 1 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -253,21 +253,17 @@ open class NavigationMapView: UIView {
     }
     
     func setupUserLocation() {
-        mapView.mapboxMap.onNext(.styleLoaded) { [weak self] _ in
```
bamx23, line 256: Is it okay that we might call this before the style is loaded?

PR #3296, opened 2021-08-25
Change tap gesture recognizer assignee

OttyLab:
### Description
`NavigationMapView` [implements tap gesture](https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-beta.23/Sources/MapboxNavigation/NavigationMapView.swift#L396) to support alternative route. This `UITapGestureRecognizer` [implements a delegate](https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-beta.23/Sources/MapboxNavigation/NavigationMapView.swift#L1489) so that user code's tap gesture can be handled.

The problem is when some `MapView` option is changed such as `gestures.options.pitchEnabled`, it overwrites the delegate in the step below and user code's gesture code is not called any more.

1. https://github.com/mapbox/mapbox-maps-ios/blob/v10.0.0-rc.6/Sources/MapboxMaps/Gestures/GestureManager.swift#L15
2. https://github.com/mapbox/mapbox-maps-ios/blob/v10.0.0-rc.6/Sources/MapboxMaps/Gestures/GestureManager.swift#L113
3. https://github.com/mapbox/mapbox-maps-ios/blob/v10.0.0-rc.6/Sources/MapboxMaps/Gestures/GestureManager.swift#L124

Instead, if `NavigationMapView` assign the tap gesture to itself, this issue does not happen.

### Implementation

Before: 

```swift
mapView.addGestureRecognizer(mapViewTapGestureRecognizer)
```

After:

```swift
addGestureRecognizer(mapViewTapGestureRecognizer)
```

### Screenshots or Gifs

No images

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -393,7 +393,7 @@ open class NavigationMapView: UIView {
         // Gesture recognizer, which is used to detect taps on route line and waypoint.
         let mapViewTapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(didReceiveTap(sender:)))
         mapViewTapGestureRecognizer.delegate = self
-        mapView.addGestureRecognizer(mapViewTapGestureRecognizer)
+        addGestureRecognizer(mapViewTapGestureRecognizer)
     }
     
     /**


1ec5 (approved):
This works, so I think we can merge it. However, I think the more robust fix would be for NavigationMapView to publicly expose its tap gesture recognizer, as [GestureManager does](https://github.com/mapbox/mapbox-maps-ios/blob/edb0eb646f9dd73a87798d1bc4812d35945667a5/Sources/MapboxMaps/Gestures/GestureManager.swift#L22-L55) with its built-in gesture recognizers. Then the application could have that built-in tap gesture recognizer require the application’s tap gesture recognizer to fail before responding. [This was the recommended way](https://github.com/mapbox/mapbox-gl-native-ios/blob/de911fde2372e8921e1b624dc5ce83e989fb29f6/platform/ios/src/MGLMapView.h#L176-L182) to deal with conflicting gesture recognizers in map SDK v6._x_ and navigation SDK v2._x_. If we make that change, potentially in a separate PR, we should consider undoing this PR so that GestureManager is the one source of truth for gesture recognition on the map part of the window.

PR #3307, opened 2021-08-30
Suport gradient for congestion color change.

ShanMa1991:
### Description
This pr is to allow the developer to choose the route line gradient color change style between the fading one and the steep change one.

### Implementation
- [x] Added the `NavigationMapView.showFadingCongestionColor` property to show blending color between different congestion levels on route lines, which defaults to `false` to show the steep change effect.
- [X]  Added the global variable `GradientCongestionBlendingDistance` to adjust the distance of blending color segment. The blending distance of two congestion level would between the `GradientCongestionblendingDistance` and the 10% of the congestion level distance.
- [X] Add test to cover the `NavigationMapView.routeLineGradient(_:fractionTraveled:isMain:isSoft)` and the soft gradient route line gradient update.
- [x] Update the ChangeLog.

### Screenshots or Gifs
To show the fading color gradient between different traffic congestion level, set `navigationMapView.crossfadesCongestionSegments =  true` as below
gradient

To show the fading color gradient between different traffic congestion level during active guidance navigation,  set `navigationViewController.navigationMapView.crossfadesCongestionSegments =  true` as below
blending

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,8 +51,9 @@
 * Renamed the `NavigationMapView.updateRoute(_:)` method to `NavigationMapView.travelAlongRouteLine(to:)`. Improved the performance of updating the route line to change color at the user’s location as they progress along the route. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
 * Fixed an issue when user passed destination and the route line grows back when `NavigationViewController.routeLineTracksTraversal` is enabled. ([#3255](https://github.com/mapbox/mapbox-navigation-ios/pull/3255))
 * The `NavigationMapView.userLocationStyle` now supports instant user location indicator change without style loaded lag. ([#3295](https://github.com/mapbox/mapbox-navigation-ios/pull/3295))
+<<<<<<< HEAD
 * Fixed incorrect color-coded traffic congestion along the route line and incorrect speeds in the speed limit view after some time had elapsed after rerouting. ([#3344](https://github.com/mapbox/mapbox-navigation-ios/pull/3344]))
-* Added the `NavigationMapView.showFadingCongestionColor` property to show blending color between different congestion levels on route lines, and the global variable `GradientCongestionblendingDistance` to adjust the distance of blending color segment. ([#3307](https://github.com/mapbox/mapbox-navigation-ios/pull/3307))
+* By default, there is no longer a subtle crossfade between traffic congestion segments along a route line. To reenable this crossfade, set the `NavigationMapView.crossfadesCongestionSegments` property to `true`. You can also adjust the length of this crossfade using the global variable `GradientCongestionFadingDistance`. ([#3307](https://github.com/mapbox/mapbox-navigation-ios/pull/3307))
 
 ### Location tracking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,6 @@
 * Renamed the `NavigationMapView.updateRoute(_:)` method to `NavigationMapView.travelAlongRouteLine(to:)`. Improved the performance of updating the route line to change color at the user’s location as they progress along the route. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
 * Fixed an issue when user passed destination and the route line grows back when `NavigationViewController.routeLineTracksTraversal` is enabled. ([#3255](https://github.com/mapbox/mapbox-navigation-ios/pull/3255))
 * The `NavigationMapView.userLocationStyle` now supports instant user location indicator change without style loaded lag. ([#3295](https://github.com/mapbox/mapbox-navigation-ios/pull/3295))
-<<<<<<< HEAD
 * Fixed incorrect color-coded traffic congestion along the route line and incorrect speeds in the speed limit view after some time had elapsed after rerouting. ([#3344](https://github.com/mapbox/mapbox-navigation-ios/pull/3344]))
 * By default, there is no longer a subtle crossfade between traffic congestion segments along a route line. To reenable this crossfade, set the `NavigationMapView.crossfadesCongestionSegments` property to `true`. You can also adjust the length of this crossfade using the global variable `GradientCongestionFadingDistance`. ([#3307](https://github.com/mapbox/mapbox-navigation-ios/pull/3307))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 * Fixed an issue when user passed destination and the route line grows back when `NavigationViewController.routeLineTracksTraversal` is enabled. ([#3255](https://github.com/mapbox/mapbox-navigation-ios/pull/3255))
 * The `NavigationMapView.userLocationStyle` now supports instant user location indicator change without style loaded lag. ([#3295](https://github.com/mapbox/mapbox-navigation-ios/pull/3295))
 * Fixed incorrect color-coded traffic congestion along the route line and incorrect speeds in the speed limit view after some time had elapsed after rerouting. ([#3344](https://github.com/mapbox/mapbox-navigation-ios/pull/3344]))
+* Added the `NavigationMapView.showFadingCongestionColor` property to show blending color between different congestion levels on route lines, and the global variable `GradientCongestionblendingDistance` to adjust the distance of blending color segment. ([#3307](https://github.com/mapbox/mapbox-navigation-ios/pull/3307))
 
 ### Location tracking
 
--- a/Sources/MapboxNavigation/Constants.swift
+++ b/Sources/MapboxNavigation/Constants.swift
@@ -43,7 +43,7 @@ public let NavigationViewMinimumVolumeForWarning: Float = 0.3
 /**
  The distance of blending color between two different congestion level segment in meters.
  */
-public var GradientCongestionblendingDistance: Double = 30.0
+public var GradientCongestionBlendingDistance: Double = 30.0
 
 extension Notification.Name {
     /**
--- a/Sources/MapboxNavigation/Constants.swift
+++ b/Sources/MapboxNavigation/Constants.swift
@@ -40,6 +40,11 @@ public let CongestionAttribute = "congestion"
  */
 public let NavigationViewMinimumVolumeForWarning: Float = 0.3
 
+/**
+ The distance of blending color between two different congestion level segment in meters.
+ */
+public let gradientCongestionblendingDistance: Double = 30.0
+
 extension Notification.Name {
     /**
      Posted when `StyleManager` applies a style that was triggered by change of time of day, or when entering or exiting a tunnel.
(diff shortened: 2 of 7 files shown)


Sources/MapboxNavigation/Constants.swift
```diff
@@ -40,6 +40,11 @@ public let CongestionAttribute = "congestion"
  */
 public let NavigationViewMinimumVolumeForWarning: Float = 0.3
 
+/**
+ The distance of blending color between two different congestion level segment in meters.
+ */
+public var GradientCongestionblendingDistance: Double = 30.0
```
jill-cardamon, line 46: Think you missed a capital here:
```suggestion
public var GradientCongestionBlendingDistance: Double = 30.0
```

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -219,14 +221,14 @@ extension NavigationMapView {
...
+    func routeLineGradient(_ congestionFeatures: [Turf.Feature]? = nil, fractionTraveled: Double, isMain: Bool = true, isSoft: Bool = false) -> [Double: UIColor] {
         var gradientStops = [Double: UIColor]()
-        var distanceTraveled = fractionTraveled
+        var distanceTraveled = 0.0
         
         if let congestionFeatures = congestionFeatures {
             let routeDistance = congestionFeatures.compactMap({ ($0.geometry.value as? LineString)?.distance() }).reduce(0, +)
             var minimumSegment: (Double, UIColor) = (Double.greatestFiniteMagnitude, .clear)
-            
+
```
jill-cardamon, line 231: Watch out for blank line changes.

Sources/MapboxNavigation/Constants.swift
ShanMa1991: Thanks for the suggestion, I'll update this name.

MaximAlien:
@ShanMa1991, feel free to rebase your changes onto `main`. Hopefully build related issues are gone now.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -236,28 +239,23 @@ extension NavigationMapView {
 
                 let lineString = feature.geometry.value as? LineString
                 guard let distance = lineString?.distance() else { return gradientStops }
+                let stopGap = max(min(GradientCongestionBlendingDistance, distance * 0.1) / routeDistance, 0.0000000000000002)
```
Udumft, line 242: Is it possible to avoid this magic constant and use `Double.leastNonzeroMagnitude` instead?

ShanMa1991: Yes, sure, I'll test to see whether it could be accepted by the style expression. Thanks for the suggestion.

ShanMa1991: `Double.leastNonzeroMagnitude` could not be accepted and recognized with the style expression, and would lead to a blending effect without the distance control. The `0.0000000000000002` here is almost the difference of the `CGFloat.nextDow` different in Double, which would give a short blending effect is the congestion level is pretty short.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -55,6 +55,13 @@ open class NavigationMapView: UIView {
      If `true` and there're multiple routes to choose, the alternative route lines would display the congestion levels at different colors, similar to the main route. To customize the congestion colors that represent different congestion levels, override the `alternativeTrafficUnknownColor`, `alternativeTrafficLowColor`, `alternativeTrafficModerateColor`, `alternativeTrafficHeavyColor`, `alternativeTrafficSevereColor` property for the `NavigationMapView.appearance()`.
      */
     public var showsCongestionForAlternativeRoutes: Bool = false
+    
+    /**
+     Controls whether to show fading gradient congestion color on route lines. Defaults to `false`.
+     
+     If `true`, the congestion level in the route line will be shown as fading gradient color instead of abrupt and steep change.
+     */
+    public var showFadingCongestionColor: Bool = false
```
1ec5, line 64: “showFadingCongestionColor” sounds like a method name; “show**s**FadingCongestionColor” would be more appropriate for a property name. Better yet, “crossfadesCongestionSegments” would be more descriptive and avoid confusion with the vanishing (fading out) route line feature.

Sources/MapboxNavigation/Constants.swift
```diff
@@ -40,6 +40,11 @@ public let CongestionAttribute = "congestion"
  */
 public let NavigationViewMinimumVolumeForWarning: Float = 0.3
 
+/**
+ The distance of blending color between two different congestion level segment in meters.
+ */
+public var GradientCongestionBlendingDistance: Double = 30.0
```
1ec5, line 46: We should use the same term throughout this PR: is it fading or blending?

Since this is a physical distance measured in meters, the type should be `CLLocationDistance`.

Should this value be configurable in a Style?

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -236,28 +239,23 @@ extension NavigationMapView {
...
```
1ec5, line 243: Is there a risk of division by zero if the linestring somehow ends up being zero-length?

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@
 * Renamed the `NavigationMapView.updateRoute(_:)` method to `NavigationMapView.travelAlongRouteLine(to:)`. Improved the performance of updating the route line to change color at the user’s location as they progress along the route. ([#3201](https://github.com/mapbox/mapbox-navigation-ios/pull/3201)).
 * Fixed an issue when user passed destination and the route line grows back when `NavigationViewController.routeLineTracksTraversal` is enabled. ([#3255](https://github.com/mapbox/mapbox-navigation-ios/pull/3255))
 * The `NavigationMapView.userLocationStyle` now supports instant user location indicator change without style loaded lag. ([#3295](https://github.com/mapbox/mapbox-navigation-ios/pull/3295))
+* Added the `NavigationMapView.showFadingCongestionColor` property to show blending color between different congestion levels on route lines, and the global variable `GradientCongestionBlendingDistance` to adjust the distance of blending color segment. ([#3307](https://github.com/mapbox/mapbox-navigation-ios/pull/3307))
```
1ec5, line 48:
```suggestion
* By default, there is no longer a subtle crossfade between traffic congestion segments along a route line. To reenable this crossfade, set the `NavigationMapView.showFadingCongestionColor` property to `true`. You can also adjust the length of this crossfade using the global variable `GradientCongestionBlendingDistance`. ([#3307](https://github.com/mapbox/mapbox-navigation-ios/pull/3307))
```

1ec5 (approved):

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: Yes, your concern is correct, I'll add check for it.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, I'll update the name.

Sources/MapboxNavigation/Constants.swift
ShanMa1991: I make it as a global variable like the `RouteControllerProactiveReroutingInterval `, they seems very similar in changing the refresh rate, or the color fading distance.
I'll set it as `CLLocationDistance ` and improve the description term.

Tests/MapboxNavigationTests/NavigationMapViewTests.swift
```diff
@@ -356,6 +356,68 @@ class NavigationMapViewTests: TestCase {
...
+        var fractionTraveledNextDown = Double(CGFloat(fractionTraveled).nextDown)
+        routeLineGradient = navigationMapView.routeLineGradient(congestionFeatures, fractionTraveled: fractionTraveled, isMain: true, isSoft: false)
+        XCTAssertEqual(routeLineGradient[0.0], navigationMapView.traversedRouteColor)
+        XCTAssertEqual(routeLineGradient[1.0], navigationMapView.trafficLowColor)
+        XCTAssertEqual(routeLineGradient[fractionTraveled], navigationMapView.trafficLowColor)
+        XCTAssertEqual(routeLineGradient[fractionTraveledNextDown], navigationMapView.traversedRouteColor)
+        
+        fractionTraveled = 0.999999999
+        fractionTraveledNextDown = Double(CGFloat(fractionTraveled).nextDown)
+        routeLineGradient = navigationMapView.routeLineGradient(congestionFeatures, fractionTraveled: fractionTraveled, isMain: true, isSoft: true)
```
MaximAlien, line 388: Can you please add verification for cases when `GradientCongestionFadingDistance` was modified and for fraction, which is exactly at the gradient itself, not when fraction is already has plain color.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -204,36 +204,37 @@ extension NavigationMapView {
...
+        filteredGradientStops[fractionTraveled] = minimumSegment.1
 
         return filteredGradientStops
     }
     
-    func routeLineGradient(_ congestionFeatures: [Turf.Feature]? = nil, fractionTraveled: Double, isMain: Bool = true) -> [Double: UIColor] {
+    func routeLineGradient(_ congestionFeatures: [Turf.Feature]? = nil, fractionTraveled: Double, isMain: Bool = true, isSoft: Bool = false) -> [Double: UIColor] {
         var gradientStops = [Double: UIColor]()
-        var distanceTraveled = fractionTraveled
+        var distanceTraveled = 0.0
```
MaximAlien, line 231: Is this correct change? What about re-using `fractionTraveled`?

ShanMa1991: Yes, I think it's a correct change.

Because the `distanceTraveled ` is used to record the distance on the route in meters, and then we use the percentage of the `distanceTraveled / routeDistance` to get the traveled segment start or end percentage to compare with `fractionTraveled`. So it should start from `0.0` meters. 

Because the `fractionTraveled` is the percentage between `0` and `1`, while the `distanceTraveled` is distance in meters, previously the small fraction number doesn't affect much in the distance in meters. But it's not correct, especially if the distance of the whole route or the first step is less than 1 meter.

Tests/MapboxNavigationTests/NavigationMapViewTests.swift
ShanMa1991: Thanks for the suggestion, it would be included and solved in the tail work in this ticket #3365

PR #3314, opened 2021-09-01
inTunnels simulation

### LINKED ISSUE #3216: SimulationMode Improvements
This is a cut from #2365 PR, see original discussion there. 

Action points:
- [ ] Introduction of a new simulation mode: .inTunnels
- [ ] Make simulation mode more discoverable. 

cc @mapbox/navigation-ios @LukasPaczos @Guardiola31337 @Skoti

Udumft:
### Description
Resolves #3216 
Adds new simulation mode, to kick the simulation while loosing GPS signal while traversing tunnels. Also exposes simulation setting though `NavigationOptions`.

### Implementation
PR adds new enum case, updates default settings to use `.inTunnels` instead previous default `.onPoorGPS` and adds additional property to `NavigationOptions`.


azarovalex:
Will there be any difference between `.inTunnels` and `.onPoorGPS` in terms of internal logic? For now, it seems that they are the same.

Udumft:
The difference is that `.inTunnels` [kicks in](https://github.com/mapbox/mapbox-navigation-ios/pull/3314/files#diff-5ea4d86736f84e024d6aed6b82168a95c760e150a3b148b163fa50f0498e8908R312) the simulation when there is not only lost GPS signal, but also current road is labeled as a tunnel.

azarovalex (approved):

PR #3318, opened 2021-09-02
Add `PassiveNavigationFeedbackType`.

azarovalex:
Adding `PassiveNavigationFeedbackType`


Sources/MapboxCoreNavigation/FeedbackEvent.swift
```diff
@@ -14,12 +14,12 @@ public class FeedbackEvent: Codable {
         coreEvent = CoreFeedbackEvent(timestamp: Date(), eventDictionary: dictionary ?? [:])
     }
     
-    func update(type: FeedbackType, source: FeedbackSource, description: String?) {
+    func update(type: String, subtype: String?, source: FeedbackSource = .user, description: String?) {
```
azarovalex, line 17: This method is now working with both `FeedbackType` and `PassiveNavigationFeedbackType`, so I guess it's easier just to pass raw values for type and subtype.

bamx23: We may have a common protocol with `description` and `subtypeDescription` properties (or maybe have better naming). But that's optional

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -361,7 +361,23 @@ open class NavigationEventsManager {
      - parameter description: A custom string used to describe the problem in detail.
      */
     public func sendFeedback(_ feedback: FeedbackEvent, type: FeedbackType, description: String? = nil) {
```
bamx23, line 363: Should we rename this to `sendActiveNavigationFeedback`?

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -361,7 +361,23 @@ open class NavigationEventsManager {
...
-        feedback.update(type: type, source: .user, description: description)
+        feedback.update(type: type.description, subtype: type.description, description: description)
```
bamx23, line 364: You set both `type` and `subtype` to be the same? Here and in the next method too.

Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
```diff
@@ -0,0 +1,105 @@
...
+    case incorrectVisual(subtype: PassiveNavigationIncorrectVisualSubtype?)
+    
+    /// Indicates a road closure was observed.
+    case roadIssue(subtype: PassiveNavigationRoadIssueSubtype?)
+    
+    /// Indicates a wrong traffic.
+    case wrongTraffic(subtype: PassiveNavigationWrongTrafficSubtype?)
+    
+    /// The user’s location is not matching the real position of the vehicle.
+    case badGPS
```
bamx23, line 18: `positioningIssue`?

Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
```diff
@@ -0,0 +1,105 @@
...
+public enum PassiveNavigationFeedbackType: CustomStringConvertible {
+    
+    /// Indicates an incorrect visual.
+    case incorrectVisual(subtype: PassiveNavigationIncorrectVisualSubtype?)
+    
+    /// Indicates a road closure was observed.
+    case roadIssue(subtype: PassiveNavigationRoadIssueSubtype?)
+    
+    /// Indicates a wrong traffic.
+    case wrongTraffic(subtype: PassiveNavigationWrongTrafficSubtype?)
```
bamx23, line 15: `trafficIssue`?

Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
```diff
@@ -0,0 +1,105 @@
...
+    case badGPS
+    
+    /// Indicates a custom feedback type and subtype.
+    case custom(type: String, subtype: String?)
+    
+    /// Indicates other feedback. You should provide a `description` string to `NavigationEventsManager.sendPassiveNavigationFeedback(_:type:description:)`
+    /// to elaborate on the feedback if possible.
+    case other
+    
+    public var description: String {
```
bamx23, line 27: Do we really need `CustomStringConvertible`? As I would rather call these properties something like `typeKey` and `subtypeKey`. We can make an extension for the `description` property to concatenate those two.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
azarovalex: Yeah, this was a typo, thanks for noticing.

Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
azarovalex: There was a proposal to use "bad GPS", because "positioning issue" sounds like this is a problem of our SDK.

Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
azarovalex: Idk, in the design doc the name is "Wrong Traffic" :)

Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
azarovalex: Agree, renamed fields to typeKey and subtypeKey

Sources/MapboxCoreNavigation/PassiveNavigationEventDetails.swift
```diff
@@ -58,7 +58,7 @@ struct PassiveNavigationEventDetails: NavigationEventDetails {
         try container.encodeIfPresent(coordinate?.longitude, forKey: .longitude)
         try container.encodeIfPresent(userIdentifier, forKey: .userIdentifier)
         try container.encodeIfPresent(appMetadata, forKey: .appMetadata)
-        try container.encodeIfPresent(feedbackType?.description, forKey: .feedbackType)
+        try container.encodeIfPresent(feedbackType?.typeKey, forKey: .feedbackType)
```
bamx23, line 61: Do we really need `feedbackType` here? I see that we add type and subtype manually to the resulting dictionary. Or maybe it's better to use these coding keys in that place to prevent hardcoded keys.

Sources/MapboxCoreNavigation/NavigationEventsManager.swift
```diff
@@ -399,16 +399,32 @@ open class NavigationEventsManager {
...
+     You can pair this with a custom feedback UI in your app to flag problems during navigation such as road closures, incorrect instructions, etc.
+     
+     - parameter feedback: A `FeedbackEvent` created with `createFeedback()` method.
+     - parameter type: A `PassiveNavigationFeedbackType` used to specify the type of feedback.
+     - parameter description: A custom string used to describe the problem in detail.
+     */
+    public func sendPassiveNavigationFeedback(_ feedback: FeedbackEvent,
+                                              type: PassiveNavigationFeedbackType,
+                                              description: String? = nil) {
+        feedback.update(type: type.typeKey, subtype: type.subtypeKey, description: description)
```
bamx23, line 427: NIT: We can introduce a common `FeedbackType` protocol with `typeKey` and `subtypeKey` and create a private method to reduce code duplication.

bamx23 (approved):
LGTM

PR #3320, opened 2021-09-03
Update dependencies

bamx23:
Updated Mapbox Navigation Native `65.0.0` and Mapbox Directions `2.0.0-beta.9`


Tests/CocoaPodsTest/PodInstall/PodInstall/ViewController.swift
```diff
@@ -6,7 +6,11 @@ class ViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        // Do any additional setup after loading the view, typically from a nib.
+
+        // Test that classes are accessible
+        // and prevent comiler optimizations for unused imports.
+        _ = PassiveLocationManager()
+        _ = NavigationMapView(frame: .zero)
```
S2Ler, line 13: I'm curious where this change comes from?

S2Ler (approved):

Tests/CocoaPodsTest/PodInstall/PodInstall/ViewController.swift
bamx23: I've added a comment for this. If we want to test integration here, it's better to use some classes from the SDK to prevent compiler optimizations for unused imports.

PR #3322, opened 2021-09-03
PassiveLocationManager Telemetry

bamx23:
I've added start/stop events for passive location manager provided navigation sessions.


Sources/MapboxCoreNavigation/EventDetails.swift
```diff
@@ -76,13 +83,9 @@ protocol NavigationEventDetails: EventDetails {
     var driverMode: String { get }
```
azarovalex, line 83: I guess you can move `driverMode` to `GlobalEventDetails` as well, so we can understand if the departure event is coming from free drive or from active guidance

azarovalex: Also `sessionIdentifier` and `startTimestamp `, because you pass `SessionState` to `FreeDriveEventDetails` anyway.

azarovalex: `userIdentifier` and  `appMetadata` can be moved as well. Anything except for things specific to feedback: `screenshot`, `feedbackType` and `description`.

Sources/MapboxCoreNavigation/FreeDriveEventDetails.swift
```diff
@@ -0,0 +1,78 @@
...
+        try container.encode(percentTimeInForeground, forKey: .percentTimeInForeground)
+        try container.encode(percentTimeInPortrait, forKey: .percentTimeInPortrait)
+        try container.encode(screenBrightness, forKey: .screenBrightness)
+        try container.encode(sessionIdentifier, forKey: .sessionIdentifier)
+        try container.encode(simulation, forKey: .simulation)
+        try container.encode(volumeLevel, forKey: .volumeLevel)
+        try container.encodeIfPresent(appMetadata, forKey: .appMetadata)
+        try container.encodeIfPresent(event, forKey: .event)
+        try container.encodeIfPresent(location?.dictionaryRepresentation, forKey: .location)
+        try container.encodeIfPresent(startTimestamp?.ISO8601, forKey: .startTimestamp)
```
azarovalex, line 76: You can add `try container.encode(NavigationEventsManager.applicationSessionIdentifier, forKey: .navigatorSessionIdentifier)`, so we will be able to group events from the same application session. See `ActiveNavigationEventDetails` for reference.

azarovalex (approved):
LGTM, just add some additional fields.

Sources/MapboxCoreNavigation/FreeDriveEventDetails.swift
bamx23: Thanks, I saw this id. But there is no `navigatorSessionIdentifier` in the schema.

Sources/MapboxCoreNavigation/EventDetails.swift
bamx23: `GlobalEventDetails` means that all the fields are either have default implementation in the extension or being updated via the extension method. That's why I would rather not move `sessionIdentifier` and `startTimestamp`.
`driverMode` isn't needed for free drive events. The event schema is different under the hood, so there is no `userIdentifier` either.
We can continue this discussion later as it's a private API.

Sources/MapboxCoreNavigation/FreeDriveEventDetails.swift
bamx23: Well, I've been looking at the old version of the schema. The id is added, thanks.

Sources/MapboxCoreNavigation/FreeDriveEventDetails.swift
```diff
@@ -0,0 +1,80 @@
+import Foundation
+import CoreLocation
+
+struct FreeDriveEventDetails: GlobalEventDetails {
```
1ec5, line 4: We should avoid the term “free drive” in client-side APIs, even if the server API is using that term. It’s confusing for a few reasons, not least of which is that it isn’t necessarily free. I realize this type isn’t public, but we may eventually need to add a related public type, so we would want to be consistent.

How does this type differ from PassiveNavigationEventDetails? It seems to be something that should be named very similarly.

azarovalex: Agree, this second struct is unnecessary in free drive, we can reuse `PassiveNavigationEventDetails` type both for feedback and departure/arrival events, similar to `ActiveNavigationEventDetails` in active guidance.

PR #3325, opened 2021-09-03
Add support for congestion numeric attribute

chezzdev:
Fixes mapbox/navigation-sdks#892.
Depends on https://github.com/mapbox/mapbox-directions-swift/pull/575.

Based on discussions in the team, it's decided to make congestion numeric usage compatible with regular congestion levels.
For that reason, numeric values are matched to congestion levels using [pre-defined ranges](https://github.com/mapbox/nav-api-intake/issues/510#issuecomment-789822797) and then used as usual.

- [x] Match numeric values to congestion levels
- [x] Provide a way to override default ranges
- [x] Prefer `numericCongestionLevel` attribute over `.congestionLevel` one in `RouteLeg`
- [x] Use `.numericCongestionLevel` instead of `.congestionLevel` in `NavigationRouteOptions`
- [x] Tests
- [x] Docs
- [x] Switch to released Directions with https://github.com/mapbox/mapbox-directions-swift/pull/575 merged


Sources/MapboxCoreNavigation/CongestionLevel.swift
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+}
+
+extension RouteLeg {
+    public var resolvedCongestionLevels: [CongestionLevel]? {
+        let congestionLevels: [CongestionLevel]?
+
+        if let numeric = segmentNumericCongestionLevels {
+            congestionLevels = numeric.map(CongestionLevel.init)
```
bamx23, line 57: I would suggest being explicit here
```suggestion
            congestionLevels = numeric.map(CongestionLevel.init(numericValue:))
```

Sources/MapboxCoreNavigation/CongestionLevel.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+import MapboxDirections
+
+public typealias CongestionRange = Range<NumericCongestionLevel>
+
+extension CongestionRange {
```
bamx23, line 6: We can make the whole extension public as every element is public

Sources/MapboxCoreNavigation/CoreConstants.swift
```diff
@@ -113,6 +113,16 @@ public var RouteControllerIncorrectCourseMultiplier: Int = 4
...
 
+//MARK: - Congestion level default ranges
+
+public let CongestionRangeLow: CongestionRange = 0..<40
+
+public let CongestionRangeModerate: CongestionRange = 40..<60
+
+public let CongestionRangeHeavy: CongestionRange = 60..<80
+
+public let CongestionRangeSevere: CongestionRange = 80..<101
```
bamx23, line 124: Docs?

Tests/MapboxCoreNavigationTests/CongestionLevelTests.swift
```diff
@@ -0,0 +1,9 @@
+import XCTest
+@testable import MapboxCoreNavigation
+
+class CongestionLevelTests: XCTestCase {
+
+    func testCongestionRangeDefaulValues() {
```
bamx23, line 6:
```suggestion
    func testCongestionRangeDefaultValues() {
```

Tests/MapboxCoreNavigationTests/CongestionLevelTests.swift
```diff
@@ -0,0 +1,9 @@
...
+        CongestionRange.resetCongestionRangesToDefault()
```
bamx23, line 7: Should there be some tests?

Sources/MapboxCoreNavigation/CongestionLevel.swift
```diff
@@ -0,0 +1,66 @@
...
+        self.heavy = heavy
+        self.severe = severe
+    }
+
+    public static func resetCongestionRangesToDefault() {
+        setCongestionRanges(low: CongestionRangeLow, moderate: CongestionRangeModerate, heavy: CongestionRangeHeavy, severe: CongestionRangeSevere)
+    }
+}
+
+extension CongestionLevel {
```
bamx23, line 30: This file is missing in the project

bamx23 (approved):
Looks good (with some NITs)

Sources/MapboxCoreNavigation/CongestionLevel.swift
```diff
@@ -0,0 +1,66 @@
...
+    public private(set) static var heavy: CongestionRange = CongestionRangeHeavy
+    public private(set) static var severe: CongestionRange = CongestionRangeSevere
+
+    public static func setCongestionRanges(low: CongestionRange, moderate: CongestionRange, heavy: CongestionRange, severe: CongestionRange) {
+        precondition(low.lowerBound >= 0, "Congestion level ranges can't include negative values.")
+        precondition(low.upperBound <= moderate.lowerBound, "Values from the moderate congestion level range can't intersect with or be lower than ones from the low congestion level range.")
+        precondition(moderate.upperBound <= heavy.lowerBound, "Values from the heavy congestion level range can't intersect with or be lower than ones from the moderate congestion level range.")
+        precondition(heavy.upperBound <= severe.lowerBound, "Values from the severe congestion level range can't intersect with or be lower than ones from the heavy congestion level range.")
+        precondition(severe.upperBound <= CongestionRangeSevere.upperBound, "Congestion level ranges can't include values greater than 100.")
```
Udumft, line 17: Will it be correct if there's a gap between subsequent congestion levels? Should we instead verify if `low.upperBound == moderate.lowerBound`? Otherwise it seems that it will treat those gaps as "unknown" congestion?

chezzdev: Not sure, what would be the use case for the gaps in intervals, but the "unknown" level of congestion for a value that you haven't defined for some reason looks semantically correct.

Udumft (approved):
LGTM

PR #3330, opened 2021-09-07
Reimplement label localization atop Expression (redux)

1ec5:
Restored the `NavigationMapView.localizeLabels()` implementation for automatically localizing map labels into the user’s language. The previous implementation was removed as part of upgrading to map SDK v10 in #2808, because it relied on [an NSExpression-based localization feature](https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L1494-L1637) in map SDK v6._x_. This implementation continues to be based on the map SDK’s built-in label localization functionality rather than reinventing the wheel as in #2933.

/cc @mapbox/navigation-ios @neelmistry94 @tobrun

### DIFF
--- a/Sources/MapboxNavigation/CarPlayMapViewController.swift
+++ b/Sources/MapboxNavigation/CarPlayMapViewController.swift
@@ -189,7 +189,7 @@ public class CarPlayMapViewController: UIViewController {
     func setupNavigationMapView() {
         let navigationMapView = NavigationMapView(frame: UIScreen.main.bounds, navigationCameraType: .carPlay)
         navigationMapView.delegate = self
-        navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { _ in
+        navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { _ in
             navigationMapView.localizeLabels()
         }
         
--- a/Sources/MapboxNavigation/MapView.swift
+++ b/Sources/MapboxNavigation/MapView.swift
@@ -45,7 +45,7 @@ extension MapView {
     }
     
     /**
-     Returns a set of source identifiers for tilesets that are or include the Mapbox Incidents source.
+     Returns a set of source identifiers for tilesets that are or include the given source.
      
      - parameter tileSetIdentifier: Identifier of the tile set in the form `user.tileset`.
      - returns: Set of source identifiers.
@@ -197,5 +197,28 @@ extension MapView {
             let identifiers = tileSetIdentifiers($0.id, sourceType: $0.type.rawValue)
             return VectorSource.isMapboxStreets(identifiers)
         }
+    }
+    
+    /**
+     Attempts to localize road labels into the local language and other labels into the given locale.
+     */
+    func localizeLabels(into locale: Locale) {
+        guard !ResourceOptionsManager.hasChinaBaseURL,
+              let mapboxStreetsSource = streetsSources().first else { return }
+        
+        let streetsSourceTilesetIdentifiers = tileSetIdentifiers(mapboxStreetsSource.id)
+        let roadLabelSourceLayerIdentifier = streetsSourceTilesetIdentifiers.compactMap { VectorSource.roadLabelLayerIdentifiersByTileSetIdentifier[$0]
+        }.first
+        
+        let style = mapboxMap.style
+        let localizableLayerIdentifiers = style.allLayerIdentifiers.lazy
+            .filter { $0.type == .symbol }
+            // We only know how to localize layers backed by the Mapbox Streets source.
+            .filter { style.layerProperty(for: $0.id, property: "source") as? String == mapboxStreetsSource.id }
+            // Road labels should match road signage, so they should not be localized.
+            // TODO: Actively delocalize road labels into the “name” property: https://github.com/mapbox/mapbox-maps-ios/issues/653
+            .filter { style.layerProperty(for: $0.id, property: "source-layer") as? String != roadLabelSourceLayerIdentifier }
+            .map { $0.id }
+        try? style.localizeLabels(into: locale, forLayerIds: Array(localizableLayerIdentifiers))
     }
 }
--- a/Sources/MapboxNavigation/RouteLineController.swift
+++ b/Sources/MapboxNavigation/RouteLineController.swift
@@ -149,17 +149,13 @@ extension NavigationMapView {
         // MARK: - NavigationComponentDelegate implementation
         
         func navigationViewDidLoad(_ view: UIView) {
-            navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { [weak self] _ in
+            navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { [weak self] _ in
                 self?.navigationMapView.localizeLabels()
                 self?.navigationMapView.mapView.showsTraffic = false
+                self?.showRouteIfNeeded()
                 
                 // FIXME: In case when building highlighting feature is enabled due to style changes and no info currently being stored
                 // regarding building identification such highlighted building will disappear.
-            }
-            
-            // Route line should be added to `MapView`, when its style changes.
-            navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { [weak self] _ in
-                self?.showRouteIfNeeded()
             }
         }
         
--- a/Sources/MapboxNavigation/VectorSource.swift
+++ b/Sources/MapboxNavigation/VectorSource.swift
@@ -1,9 +1,14 @@
-
 import Foundation
 import MapboxMaps
 
 
 extension VectorSource {
+    /// A dictionary associating known tile set identifiers with identifiers of source layers that contain road names.
+    static let roadLabelLayerIdentifiersByTileSetIdentifier = [
+        "mapbox.mapbox-streets-v8": "road",
+        "mapbox.mapbox-streets-v7": "road_label",
+    ]
+    
     /**
      Method, which returns a boolean value indicating whether the tile source is a supported version of the Mapbox Streets source.
      */
--- a/Sources/MapboxNavigation/VectorSource.swift
+++ b/Sources/MapboxNavigation/VectorSource.swift
@@ -15,4 +15,56 @@ extension VectorSource {
     static func isMapboxStreets(_ identifiers: [String]) -> Bool {
         return identifiers.contains("mapbox.mapbox-streets-v8") || identifiers.contains("mapbox.mapbox-streets-v7")
     }
+    
+    /**
+     An array of locales for which Mapbox Streets source v8 has a [dedicated name field](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#name-text--name_lang-code-text).
+     */
+    static let mapboxStreetsLocales = ["ar", "de", "en", "es", "fr", "it", "ja", "ko", "pt", "ru", "vi", "zh-Hans", "zh-Hant"].map(Locale.init(identifier:))
+    
+    /**
+     Returns the BCP 47 language tag supported by Mapbox Streets source v8 that is most preferred according to the given preferences.
+     */
+    static func preferredMapboxStreetsLocalization(among preferences: [String]) -> String? {
+        let preferredLocales = preferences.map(Locale.init(identifier:))
+        let acceptsEnglish = preferredLocales.contains { $0.languageCode == "en" }
+        var availableLocales = mapboxStreetsLocales
+        if !acceptsEnglish {
+            availableLocales.removeAll { $0.languageCode == "en" }
+        }
+        
+        let mostSpecificLanguage = Bundle.preferredLocalizations(from: availableLocales.map { $0.identifier },
+                                                                 forPreferences: preferences)
+            .max { $0.count > $1.count }
+        
+        // `Bundle.preferredLocalizations(from:forPreferences:)` is just returning the first localization it could find.
+        if let mostSpecificLanguage = mostSpecificLanguage, !preferredLocales.contains(where: { $0.languageCode == Locale(identifier: mostSpecificLanguage).languageCode }) {
+            return nil
+        }
+        
+        return mostSpecificLanguage
+    }
+    
+    /**
+     Returns the locale supported by Mapbox Streets source v8 that is most preferred for the given locale.
+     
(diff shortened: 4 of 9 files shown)


Sources/MapboxNavigation/RouteLineController.swift
```diff
@@ -149,18 +149,14 @@ extension NavigationMapView {
...
                 self?.navigationMapView.mapView.showsTraffic = false
+                self?.showRouteIfNeeded()
                 
                 // FIXME: In case when building highlighting feature is enabled due to style changes and no info currently being stored
                 // regarding building identification such highlighted building will disappear.
             }
-            
-            // Route line should be added to `MapView`, when its style changes.
-            navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { [weak self] _ in
-                self?.showRouteIfNeeded()
```
1ec5, line 162: I consolidated this operation with the block above. I’m pretty sure all of these style customizations need to be reapplied every time the style changes.

Sources/MapboxNavigation/MapView.swift
```diff
@@ -198,4 +198,27 @@ extension MapView {
...
+        
+        let streetsSourceTilesetIdentifiers = tileSetIdentifiers(mapboxStreetsSource.id)
+        let roadLabelSourceLayerIdentifier = streetsSourceTilesetIdentifiers.compactMap { VectorSource.roadLabelLayerIdentifiersByTileSetIdentifier[$0]
+        }.first
+        
+        let localizableLayerIdentifiers = style.allLayerIdentifiers.filter { $0.type == .symbol }
+            // We only know how to localize layers backed by the Mapbox Streets source.
+            .filter { style.layerProperty(for: $0.id, property: "source") as? String == mapboxStreetsSource.id }
+            // Road labels should match road signage, so they should not be localized.
+            // TODO: Actively delocalize road labels into the “name” property: https://github.com/mapbox/mapbox-maps-ios/issues/653
```
1ec5, line 219: Road labels are untouched and remain in English if the original style had them in English, as does the default Mapbox Navigation Day/Night v1 style. This is due to a missing special case upstream: mapbox/mapbox-maps-ios#653. It’s a regression from v1._x_.

Tests/MapboxNavigationTests/MapViewTests.swift
```diff
@@ -137,4 +138,151 @@ class MapViewTests: XCTestCase {
...
+        XCTAssertEqual(textFieldExpression(layerIdentifier: "road-labels"),
+                       Exp(.format) {
+                        Exp(.coalesce) { Exp(.get) { "name_en" }; Exp(.get) { "name" } }
+                        FormatOptions()
+                       },
+                       "Road labels should remain in English after localization.")
+        // https://github.com/mapbox/mapbox-maps-ios/issues/652
+        XCTAssertNotEqual(textFieldExpression(layerIdentifier: "place-labels"),
+                          Exp(.format) {
+                            Exp(.coalesce) { Exp(.get) { "name_zh-Hans" }; Exp(.get) { "name" } }
```
1ec5, line 280: Localization is currently broken for most Chinese locales, including Simplified Chinese and Hong Kong Traditional Chinese: mapbox/mapbox-maps-ios#652.

Tests/MapboxNavigationTests/MapViewTests.swift
```diff
@@ -137,4 +138,151 @@ class MapViewTests: XCTestCase {
...
+                            FormatOptions()
+                          },
+                          "Place labels should be in Simplified Chinese after localization.")
+        
+        expect { mapView.localizeLabels(into: Locale(identifier: "tlh")) }
+            .to(throwAssertion(), description: "Localization into Klingon should not be supported. 🖖")
```
1ec5, line 286: The map SDK throws a fatal error rather if we pass in an unsupported Locale: mapbox/mapbox-maps-ios#655. This is problematic because the user could have set the system language preference to some language that the Streets source doesn’t support, triggering a crash. If this issue isn’t fixed upstream, we’ll have to hard-code the same array of supported languages in the navigation SDK, making it even more difficult for the server team to add support for new language.

Sources/MapboxNavigation/MapView.swift
```diff
@@ -198,4 +198,27 @@ extension MapView {
...
+        guard !ResourceOptionsManager.hasChinaBaseURL else { return }
+        
+        let style = mapboxMap.style
+        guard let mapboxStreetsSource = streetsSources().first else { return }
+        
+        let streetsSourceTilesetIdentifiers = tileSetIdentifiers(mapboxStreetsSource.id)
+        let roadLabelSourceLayerIdentifier = streetsSourceTilesetIdentifiers.compactMap { VectorSource.roadLabelLayerIdentifiersByTileSetIdentifier[$0]
+        }.first
+        
+        let localizableLayerIdentifiers = style.allLayerIdentifiers.filter { $0.type == .symbol }
```
S2Ler, line 215: nit: I would apply `.lazy` to this transformation sequence.

Sources/MapboxNavigation/RouteLineController.swift
```diff
@@ -149,18 +149,14 @@ extension NavigationMapView {
         // MARK: - NavigationComponentDelegate implementation
         
         func navigationViewDidLoad(_ view: UIView) {
-            navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { [weak self] _ in
+            navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { [weak self] _ in
```
S2Ler, line 152: @mapbox/navigation-ios do we know if it is safe to change to `onEvery`?

Sources/MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -189,7 +189,7 @@ public class CarPlayMapViewController: UIViewController {
     func setupNavigationMapView() {
         let navigationMapView = NavigationMapView(frame: UIScreen.main.bounds, navigationCameraType: .carPlay)
         navigationMapView.delegate = self
-        navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { _ in
+        navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { _ in
```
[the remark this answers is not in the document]
1ec5: Generally speaking, StyleManager is responsible for applying a new style initially and whenever switching between DayStyle and NightStyle. We can’t use StyleManagerDelegate methods directly, because these operations need to wait until the style has finished loading, or else they won’t have any effect.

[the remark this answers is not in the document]
S2Ler: Disabling traffic on Night/Day looks unexpected to me.

Sources/MapboxNavigation/MapView.swift
```diff
@@ -198,4 +198,27 @@ extension MapView {
...
         }
     }
+    
+    /**
+     Attempts to localize road labels into the local language and other labels into the given locale.
+     */
+    func localizeLabels(into locale: Locale) {
+        guard !ResourceOptionsManager.hasChinaBaseURL else { return }
+        
+        let style = mapboxMap.style
```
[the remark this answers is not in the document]
1ec5: Good catch, this was unnecessary.

Sources/MapboxNavigation/MapView.swift
1ec5: Good idea, done.

Sources/MapboxNavigation/CarPlayMapViewController.swift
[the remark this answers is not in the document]
1ec5: Traffic was disabled in #3209. This change just keeps it off if the user happens to be using CarPlay around sunrise or sunset.

@MaximAlien, is there additional context as to why users should not see traffic during browsing and previewing? These are the situations where we used to use the Navigation Preview Day/Night styles. NavigationMapView may not show traffic by default, but that’s because it isn’t necessarily being used as a preview map; the Example application’s ViewController should enable it in that case.

Sources/MapboxNavigation/VectorSource.swift
```diff
@@ -1,13 +1,70 @@
...
+        
+        return mostSpecificLanguage
+    }
+    
+    /**
+     Returns the locale supported by Mapbox Streets source v8 that is most preferred for the given locale.
+     
+     - parameter locale: The locale to match. To use the system’s preferred language, if supported, specify `nil`. To use the local language, specify a locale with the identifier `mul`.
+     */
+    static func preferredMapboxStreetsLocale(for locale: Locale?) -> Locale? {
```
1ec5, line 52: I ended up porting [`+[MGLVectorTileSource preferredMapboxStreetsLanguageForPreferences:]`](https://github.com/mapbox/mapbox-gl-native-ios/blob/57b231776bb3bf95b90b61cee707cdb8bab33e1a/platform/darwin/src/MGLVectorTileSource.mm#L119-L144) from map SDK v6._x_ to work around mapbox/mapbox-maps-ios#652 and mapbox/mapbox-maps-ios#655. I would very much like not to have to hard-code the supported Mapbox Streets source locales in this codebase, but this is the only way to work around the blanking and crashing issues for now.

Tests/MapboxNavigationTests/MapViewTests.swift
1ec5: > If this issue isn’t fixed upstream, we’ll have to hard-code the same array of supported languages in the navigation SDK, making it even more difficult for the server team to add support for new language.

I wound up implementing this workaround for now. We can easily remove the workaround once the map SDK is fixed.

Tests/MapboxNavigationTests/MapViewTests.swift
1ec5: I worked around this issue for now by ensuring that only supported locales make it through to the map SDK.

Sources/MapboxNavigation/RouteLineController.swift
1ec5: As of #3332, the only thing left that was `onNext` will be the call to `localizeLabels()`, which does need to be reapplied every time the style changes.

MaximAlien: I'm also not sure that we should apply this change. Basically now any changes to the `NavigationMapView.mapView.showsTraffic`, which were made by the end user will be ignored.

Tests/MapboxNavigationTests/MapViewTests.swift
```diff
@@ -137,4 +138,181 @@ class MapViewTests: XCTestCase {
...
+                       },
+                       "Place labels should be in Spanish after localization.")
+        
+        mapView.localizeLabels(into: Locale(identifier: "zh-Hant-TW"))
+        
+        XCTAssertEqual(textFieldExpression(layerIdentifier: "road-labels"),
+                       Exp(.format) {
+                        Exp(.coalesce) { Exp(.get) { "name_en" }; Exp(.get) { "name" } }
+                        FormatOptions()
+                       },
```
MaximAlien, line 260: This test looks pretty large, would it be possible to expose this part into reusable component?

Sources/MapboxNavigation/RouteLineController.swift
1ec5: Any time the style changes, their change to `showsTraffic` would get blown away anyways.

Tests/MapboxNavigationTests/MapViewTests.swift
1ec5: I condensed the tests a bit to make them a bit easier to read.

1ec5:
The breaking API detection tool [flagged](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/5419/workflows/0c240812-0d86-410a-97cf-2aa6d0c9a46d/jobs/44429) changes I wasn’t expecting to see:

```
**** BREAKING CHANGES DETECTED ****

Breaking changes in 'roadLabelLayerIdentifiersByTileSetIdentifier'
*new* static var: `roadLabelLayerIdentifiersByTileSetIdentifier`

Breaking changes in 'mapboxStreetsLocales'
*new* static var: `mapboxStreetsLocales`

Breaking changes in 'localizeLabels(into:)'
*new* method: `localizeLabels(into:)`

Breaking changes in 'preferredMapboxStreetsLocale(for:)'
*new* static method: `preferredMapboxStreetsLocale(for:)`

Breaking changes in 'preferredMapboxStreetsLocalization(among:)'
*new* static method: `preferredMapboxStreetsLocalization(among:)`

Breaking changes in 'sourceIdentifiers(_:)'
*modified* method: `sourceIdentifiers(_:)`
```

New methods wouldn’t be breaking changes unless they’re added to a protocol, and new static variables wouldn’t be breaking changes either.

S2Ler:
@1ec5 please update `.base_api` commit hash to the last commit of this PR and create a task for APIDiff fix.

S2Ler:
I've added all these cases to https://github.com/mapbox/mapbox-navigation-ios/issues/3353. Meanwhile we can either disable diff reporter or do a dance with `.base_api`.

MaximAlien:
Shall we also update changelog to mention that `NavigationMapView.localizeLabels()` now works?

MaximAlien (approved):

Sources/MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -189,7 +189,7 @@ public class CarPlayMapViewController: UIViewController {
     func setupNavigationMapView() {
         let navigationMapView = NavigationMapView(frame: UIScreen.main.bounds, navigationCameraType: .carPlay)
         navigationMapView.delegate = self
-        navigationMapView.mapView.mapboxMap.onNext(.styleLoaded) { _ in
```
MaximAlien, line 192: I think `CarPlayNavigationViewController` contains similar logic in `setupNavigationMapView`, which should be updated.

1ec5:
> Shall we also update changelog to mention that `NavigationMapView.localizeLabels()` now works?

We should mention it in the next release candidate’s release notes but not the running changelog for v2.0.0, because it was already working in v1._x_.

Sources/MapboxNavigation/CarPlayMapViewController.swift
1ec5: Thanks, I fixed it to match what changed here.

PR #3332, opened 2021-09-07
Migrate style URLs to StyleURI

### LINKED ISSUE #3331: No traffic on map when browsing in CarPlay

v2.0.0-beta.25

### Steps to reproduce

Run the Example-CarPlay target in CarPlay.

### Expected behavior

Streets in the Mapbox Navigation Day v1 style are color-coded by traffic congestion.

### Actual behavior

No traffic congestion shown on the map.

browsing

repeatable

/cc @mapbox/navigation-ios

1ec5:
Renamed the `MapboxMaps.Style.navigationDayStyleURL` and `MapboxMaps.Style.navigationNightStyleURL` properties to `StyleURI.navigationDay` and `StyleURI.navigationNight`, respectively, for consistency with the changes in mapbox/mapbox-maps-ios#342. Removed the `MapboxMaps.Style.navigationDayStyleURL(version:)` and `MapboxMaps.Style.navigationNightStyleURL(version:)` methods in favor of these renamed properties. This is a backwards-incompatible change.

Restored the traffic layer to the CarPlay browsing and previewing activities, for consistency with NavigationMapView. Default to the Navigation Day style in the Example application to make that default behavior apparent. Previously, that map view showed Streets by default but would switch to Navigation Night and then Navigation Day when using the Toggle Day/Night option in the gear menu.

 

Fixes #3331.

/cc @mapbox/navigation-ios


Sources/MapboxNavigation/Style.swift
```diff
@@ -35,14 +35,14 @@ open class Style: NSObject {
     /**
      URL of the style to display on the map during turn-by-turn navigation.
      */
-    @objc open var mapStyleURL: URL = MapboxMaps.Style.navigationDayStyleURL
+    @objc open var mapStyleURL = URL(string: StyleURI.navigationDay.rawValue)!
```
1ec5, line 38: StyleURI is a Swift type that doesn’t bridge to Objective-C. This property must be available to Objective-C to participate in the UIAppearance system, which relies on Objective-C runtime reflection. The new code is uglier than the old code, but I don’t see a more elegant approach except for migrating the map SDK back to straight-up URLs.

bamx23 (approved):

MaximAlien (approved):

MaximAlien:
Thanks. I think it'd also make sense to walk through the docs and examples to see if we need to make any further updates.

PR #3334, opened 2021-09-07
Partially update the layer property instead of updating the whole layer.

### LINKED ISSUE #3333: Partially update the layer property instead of updating the whole layer.
Right now when there's a new location comes in, we update the whole line layer with the new `lineGradient` property. It involves the retrieval of the whole route line layer and the layer's jsonObject. To avoid the high CPU consuming, we should only update the layer property with the `mapView.mapboxMap.style.setLayerProperty(for:property:value:)`, and only the json format of the new `lineGradient` expression.

https://github.com/mapbox/mapbox-navigation-ios/blob/3c81bdb6df768707073f9d96188ffd92d6a79985/Sources/MapboxNavigation/NavigationMapView%2BVanishingRouteLine.swift#L183-L191

ShanMa1991:
### Description
This pr is to fix #3333 , to allow the route line gradient be updated through the `mapView.mapboxMap.style.setLayerProperty(for:property:value:)` with only the route line gradient expression transformed to JSON, instead of retrieving the whole line layer and the whole route line JSON object.

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
+++ b/Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
@@ -179,35 +179,28 @@ extension NavigationMapView {
             return
         }
         
-        let mainRouteLayerGradient = self.updateRouteLineGradientStops(fractionTraveled: fractionTraveled, gradientStops: currentLineGradientStops)
+        let mainRouteLayerGradient = updateRouteLineGradientStops(fractionTraveled: fractionTraveled, gradientStops: currentLineGradientStops)
         let mainRouteLayerGradientExpression = Expression.routeLineGradientExpression(mainRouteLayerGradient, lineBaseColor: trafficUnknownColor)
-        
-        if let data = try? JSONEncoder().encode(mainRouteLayerGradientExpression.self),
-           let jsonObject = try? JSONSerialization.jsonObject(with: data, options: []) {
-            do {
-                try mapView.mapboxMap.style.setLayerProperty(for: mainRouteLayerIdentifier,
-                                                             property: "line-gradient",
-                                                             value: jsonObject)
-            } catch {
-                print("Failed to update main route line layer.")
-            }
-        }
+        setLayerLineGradient(for: mainRouteLayerIdentifier, exp: mainRouteLayerGradientExpression)
         
         let mainRouteCasingLayerGradient = routeLineGradient(fractionTraveled: fractionTraveled)
         let mainRouteCasingLayerGradientExpression = Expression.routeLineGradientExpression(mainRouteCasingLayerGradient, lineBaseColor: routeCasingColor)
+        setLayerLineGradient(for: mainRouteCasingLayerIdentifier, exp: mainRouteCasingLayerGradientExpression)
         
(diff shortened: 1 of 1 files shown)


Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -179,18 +179,32 @@ extension NavigationMapView {
...
-                lineLayer.lineGradient = .expression(Expression.routeLineGradientExpression(mainRouteLayerGradient, lineBaseColor: trafficUnknownColor))
+        let mainRouteLayerGradient = self.updateRouteLineGradientStops(fractionTraveled: fractionTraveled, gradientStops: currentLineGradientStops)
+        let mainRouteLayerGradientExpression = Expression.routeLineGradientExpression(mainRouteLayerGradient, lineBaseColor: trafficUnknownColor)
+        
+        if let data = try? JSONEncoder().encode(mainRouteLayerGradientExpression.self),
+           let jsonObject = try? JSONSerialization.jsonObject(with: data, options: []) {
+            do {
+                try mapView.mapboxMap.style.setLayerProperty(for: mainRouteLayerIdentifier,
+                                                             property: "line-gradient",
+                                                             value: jsonObject)
```
1ec5, line 190: Working with raw JSON here is really unfortunate. Client code shouldn’t have to know about the implementation details underpinning the `LineLayer.lineGradient` property.

1ec5 (approved):
This is acceptable as a workaround while we wait for an improvement upstream in the map SDK’s runtime styling API that doesn’t require client code to trade off type safety against incremental updates to individual properties.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: There's a [ticket](https://github.com/mapbox/mapbox-maps-ios/issues/659) tracking some other ways to update the `Expression` property in Map side.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -179,18 +179,32 @@ extension NavigationMapView {
...
-            
-            try mapView.mapboxMap.style.updateLayer(withId: mainRouteCasingLayerIdentifier) { (lineLayer: inout LineLayer) throws in
-                let mainRouteCasingLayerGradient = routeLineGradient(fractionTraveled: fractionTraveled)
-                lineLayer.lineGradient = .expression(Expression.routeLineGradientExpression(mainRouteCasingLayerGradient, lineBaseColor: routeCasingColor))
+        }
+        
+        let mainRouteCasingLayerGradient = routeLineGradient(fractionTraveled: fractionTraveled)
+        let mainRouteCasingLayerGradientExpression = Expression.routeLineGradientExpression(mainRouteCasingLayerGradient, lineBaseColor: routeCasingColor)
+        
+        if let data = try? JSONEncoder().encode(mainRouteCasingLayerGradientExpression.self),
```
bamx23, line 199: Can we make a private method to reduce code duplication here?

bamx23 (approved):

Issue #3312, opened 2021-09-01
user-agent doesn't include SDK versions when SPM is used.


S2Ler:
**Mapbox Navigation SDK version:** v2.0.0-beta.24

The implementation from https://github.com/mapbox/mapbox-navigation-ios/blob/d2e6223dceae929c7affb95709494a4a6d968a94/Sources/MapboxCoreNavigation/URLSession.swift#L10-L66 doesn't work with SPM based distribution.

ShanMa1991:
The implementation of the `stringForShortVersion` from the `MBXInfo`, is because the SPM would automatically generate the `Info.plist ` which doesn't contain some certain keys, which include the bundle short version string. I'm not sure how it works right now with the SPM.

PR #3335, opened 2021-09-07
Fix bundle names retrieval in `URLSession.userAgent`.

MaximAlien:
Closing #3312.

Result string previously looked like this when using SPM:
`Example/2.0.0 Example/2.0.0 iOS/14.5.0 (x86_64)`

Now it looks like this:
`Example/2.0.0 MapboxNavigation/2.0.0 MapboxCoreNavigation/2.0.0 Example/2.0.0 iOS/14.5.0 (x86_64)`

### DIFF
--- a/Sources/MapboxCoreNavigation/URLSession.swift
+++ b/Sources/MapboxCoreNavigation/URLSession.swift
@@ -27,10 +27,22 @@ extension URLSession {
         let bundleComponents = bundles.compactMap { (bundle) -> String? in
             guard let name = bundle?.object(forInfoDictionaryKey: "CFBundleName") as? String ?? bundle?.bundleIdentifier else { return nil }
             var stringForShortVersion: String? {
+                #if SWIFT_PACKAGE
+                let mapboxNavigationName = "MapboxNavigation_MapboxNavigation"
+                #else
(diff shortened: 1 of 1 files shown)


S2Ler (approved):

Sources/MapboxCoreNavigation/URLSession.swift
```diff
@@ -8,23 +8,44 @@ extension URLSession {
...
             .mapboxCoreNavigation,
             .init(for: Directions.self),
         ]
+        #endif
+
         let bundleComponents = bundles.compactMap { (bundle) -> String? in
             guard let name = bundle?.object(forInfoDictionaryKey: "CFBundleName") as? String ?? bundle?.bundleIdentifier else { return nil }
             var stringForShortVersion: String? {
+                #if SWIFT_PACKAGE
+                let mapboxNavigationName = "MapboxNavigation_MapboxNavigation"
```
bamx23, line 31: Should we use `"MapboxNavigation"` for the `name` part of the user agent? As for now, we have different user agents for SPM and non-SPM installations.

MaximAlien: Agree, changed to use default bundle names for SPM and CocoaPods.

PR #3323, opened 2021-09-03
Free drive feedback UI

azarovalex:
Updates UI according to the changes in Core part: https://github.com/mapbox/mapbox-navigation-ios/pull/3318

PR #3337, opened 2021-09-08
Copyedit feedback categories

1ec5:
A couple of the feedback categories added in #3323 for passive navigation are misspelled or otherwise inconsistent with analogous feedback categories for active turn-by-turn navigation.

/cc @mapbox/navigation-ios

### DIFF
--- a/Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
+++ b/Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
@@ -49,8 +49,8 @@ public enum PassiveNavigationFeedbackType: FeedbackType {
             return "incorrect_speed_limit"
         case .roadIssue(subtype: .streetPermanentlyBlockedOff):
             return "street_permanently_blocked_off"
-        case .roadIssue(subtype: .streetTemporaryBlockedOff):
-            return "street_temporary_blocked_off"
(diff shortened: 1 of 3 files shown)


Sources/MapboxCoreNavigation/PassiveNavigationFeedbackType.swift
```diff
@@ -49,7 +49,7 @@ public enum PassiveNavigationFeedbackType: FeedbackType {
             return "incorrect_speed_limit"
         case .roadIssue(subtype: .streetPermanentlyBlockedOff):
             return "street_permanently_blocked_off"
-        case .roadIssue(subtype: .streetTemporaryBlockedOff):
+        case .roadIssue(subtype: .streetTemporarilyBlockedOff):
             return "street_temporary_blocked_off"
```
1ec5, line 53: Should we also correct the typoe in the raw value, or was this spelling intentional for matching what the backend expects?

azarovalex: Thanks for noticing! This constant was in the design doc, and Android merged with the same typo too. Server engineers told me that they don't have a hardcoded list of feedback types, so let's change the raw value too.
I asked Android folks to make the changes too.

azarovalex (approved):

PR #3338, opened 2021-09-08
Expose serviceAccessToken from MBXAccounts

S2Ler:
This is need in multi-access-token case. We should make sure that token
and access token matches.


Sources/TestHelper/BillingServiceMock.swift
```diff
@@ -30,6 +30,7 @@ public final class BillingServiceMock: BillingService {
     private var _sessionStates: [BillingHandler.SessionType: BillingHandler.SessionState] = [:]
     private var _events: [Event] = []
 
+    public var accessToken: String = ""
```
bamx23, line 33: Should we use some non-empty string here and cover it with a test?

S2Ler: slightly different though

bamx23 (approved):

PR #3340, opened 2021-09-08
Make `pointAheadUser` calculated once per update

bamx23:
For large route steps we spend too much time in `LineString.sliced(from:to:)` called to find `pointAheadUser` for the code that looks for a road name. And we've been doing this call multiple times _(hundreds of times in some cases)_ for the same location update. This can even provoke crashes due to CPU consumption.

This seems to be a simple yet effective optimization.


S2Ler (approved):
So cool. Please add a changelog entry.

Udumft (approved):
LGTM

Issue #2161, opened 2019-06-24
Exit colors do not update when day/night style changes


johnnewman:
**Mapbox Navigation SDK version: 0.34.0**

When the system time causes a day/night style change while navigating, exits in the instruction banner view don't change color. This is the `ExitView.appearance().foregroundColor` within `DayStyle.swift`.

It happens while transitioning from day into night as well as night into day:

<img src="https://user-images.githubusercontent.com/1591735/60050824-8eb98480-969f-11e9-95de-98898e832852.jpg" width="250" /> <img src="https://user-images.githubusercontent.com/1591735/60050853-a133be00-969f-11e9-8aff-606e3dec00bb.jpg" width="250" />

1ec5:
The wrong-colored generic shields in CarPlay (#2191) are probably related.

avi-c:
I believe the issue here is that the generated ExitView & GenericShieldView images are cached and reused, but the cache is not invalidated when the style changes. When the cached images are used in the new style, they will most likely not be the correct colors.

Probably the easiest fix is to invalidate the cache whenever the style changes, and when the appearance trait collection changes (so that Dark/Light mode changes work too).

avi-c:
Scratch that. No need to invalidate the cache. The issue is that the foreground color is not being set before the views are snapshotted. This means that the same image is retrieved from the cache regardless of what the UIAppearance proxy foregroundColor is set to.

A PR with a fix is coming shortly.

johnnewman:
I am still seeing this issue in Mapbox Navigation SDK 0.39.0.

1ec5:
Looks like the issue was reported again in #2938 and fixed in #2977 by invalidating the cache, as originally proposed in https://github.com/mapbox/mapbox-navigation-ios/issues/2161#issuecomment-560933324.

Issue #3336, opened 2021-09-08
[Bug]: Unable to style `GenericRouteShield`.


MaximAlien:
### Mapbox Navigation SDK version

2.0.0-beta.25

### Steps to reproduce

1. Add such styling to `CustomDayStyle` and `CustomNightStyle`:

```swift
PrimaryLabel.appearance(whenContainedInInstancesOf: [InstructionsBannerView.self]).normalTextColor = UIColor.green
SecondaryLabel.appearance(whenContainedInInstancesOf: [InstructionsBannerView.self]).normalTextColor = UIColor.red
GenericRouteShield.appearance().foregroundColor = UIColor.blue
ExitView.appearance().foregroundColor = UIColor.yellow
```

2. Run Styled UI example with route, which includes route shields.

### Expected behavior

It is expected that route shields are styled like this:

![Simulator Screen Shot - iPod touch (7th generation) - 2021-09-07 at 18 16 42](https://user-images.githubusercontent.com/1496498/132430275-a2e79c06-4fdf-4e21-ba12-39e613a5d8dd.png)

### Actual behavior

![Simulator Screen Shot - iPod touch (7th generation) - 2021-09-07 at 18 18 03](https://user-images.githubusercontent.com/1496498/132430286-8d60018b-b583-4e5b-9744-88ad4b888eb0.png)

### Is this a one-time issue or a repeatable issue?

repeatable

PR #2280, opened 2019-12-03
Set the foregroundColor of the ExitView and GenericShieldView to the datasource textColor before snapshotting them

avi-c:
Fixes https://github.com/mapbox/mapbox-navigation-ios/issues/2161 and fixes https://github.com/mapbox/mapbox-navigation-ios/issues/2191

PR #3341, opened 2021-09-08
Fix `GenericRouteShield` styling.

MaximAlien:
Closing #3336.


Sources/MapboxNavigation/InstructionPresenter.swift
```diff
@@ -179,7 +201,8 @@ class InstructionPresenter {
             attachment.image = image
         } else {
             let view = GenericRouteShield(pointSize: dataSource.font.pointSize, text: text)
-            view.foregroundColor = dataSource.textColor
+            view.foregroundColor = GenericRouteShield.appearance().foregroundColor
+            view.borderWidth = GenericRouteShield.appearance().borderWidth
```
MaximAlien, line 205: @avi-c, it's been a while since this logic was modified, but do you remember why you've decided to use `dataSource.textColor` instead of `GenericRouteShield`s appearance. Maybe I'm missing some edge case here.

Tests/MapboxNavigationTests/ManeuverViewSnapshotTests.swift
```diff
@@ -60,18 +68,19 @@ class ManeuverViewSnapshotTests: TestCase {
...
+                            as: .image(precision: 0.95))
     }
 
     func testRightUTurn() {
-        maneuverView.drivingSide = .left
         maneuverView.visualInstruction = maneuverInstruction(.turn, .uTurn)
-        assertImageSnapshot(matching: maneuverView.layer, as: .image(precision: 0.95))
+        maneuverView.drivingSide = .left
+
+        assertImageSnapshot(matching: UIImageView(image: maneuverView.imageRepresentation),
```
MaximAlien, line 82: Using `UIView.imageRepresentation` makes sure to verify whether flipping transformations were applied. This is relevant for right/left driving side roads.

Sources/MapboxNavigation/InstructionPresenter.swift
1ec5: #2280 was intended to fix #2161, but #2977 ended up actually fixing that issue when switching between day and night. Not sure if the changes in #2280 are still required.

Sources/MapboxNavigation/InstructionPresenter.swift
```diff
@@ -179,7 +206,8 @@ class InstructionPresenter {
             attachment.image = image
         } else {
             let view = GenericRouteShield(pointSize: dataSource.font.pointSize, text: text)
```
1ec5, line 208: Should the font also come from the appearance proxy?

Sources/MapboxNavigation/InstructionPresenter.swift
1ec5: One thing to watch out for is that CarPlay always uses the night style, even during the daytime when the phone uses the day style.

MaximAlien: Behavior was changed: user will have to provide two different appearances for iOS and CarPlay, Navigation SDK by default also provides colors for different interface styles.

avi-c: Your first comment got buried under my email notifications. 

I can't recall exactly why I chose the data source over the straight appearance proxy but it might have been that I was using the shield within another view container and needed it to defer to the datasource to find the right color.

Sources/MapboxNavigation/InstructionPresenter.swift
MaximAlien: Font is applied to `NSTextAttachment` afterwards, so I don't think it's relevant for cache. I'll run some manual tests though and consider adding unit-test.

Sources/MapboxNavigation/GenericRouteShield.swift
```diff
@@ -94,25 +94,26 @@ public class GenericRouteShield: StylableView {
...
+            appearance.foregroundColor,
+            appearance.borderColor,
+            appearance.borderWidth,
+            appearance.cornerRadius,
+            traitCollection.userInterfaceIdiom.rawValue
+        ]
+        
+        if #available(iOS 12.0, *) {
+            criticalProperties.append(traitCollection.userInterfaceStyle.rawValue)
+        }
```
1ec5, line 115: Thanks, this is a lot simpler.

Sources/MapboxNavigation/Style.swift
```diff
@@ -49,6 +49,15 @@ open class Style: NSObject {
...
     @objc open func apply() { }
     
+    /**
+     Applies the style based on provided `UITraitCollection`.
+     
+     - parameter traitCollection: `UITraitCollection` instance, properties of which
+     (e.g. `userInterfaceStyle`, `userInterfaceIdiom` etc), will be used for applying style.
+     */
+    @available(iOS 12.0, *)
+    open func apply(for traitCollection: UITraitCollection?) { }
```
1ec5, line 59: Let’s mention the addition of this method in the changelog.

1ec5 (approved):

Sources/MapboxNavigation/Style.swift
MaximAlien: Decided not to add this public method. All styling configuration for CarPlay will happen in `DayStyle` (in case if CarPlay supports both light and dark styling).

PR #3342, opened 2021-09-08
Update to Mapbox Maps `v10.0.0-rc.8`, Navigation Native `66.0.0` and Common `18.0.0`.

MaximAlien:
Update Mapbox Navigation to Mapbox Maps `v10.0.0-rc.8`, Mapbox Navigation Native `66.0.0` and Mapbox Common `18.0.0`.


Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -252,7 +252,11 @@ open class RouteController: NSObject {
         
         rawLocation = location
         
-        locations.forEach { navigator.updateLocation(for: FixLocation($0)) }
+        locations.forEach {
+            navigator.updateLocation(for: FixLocation($0)) { _ in
```
MaximAlien, line 256: Not sure that we have to do anything here. Additional logging might be helpful, but if there are a lot of locations it might clutter the output. I also think that this callback should be optional.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -105,21 +105,37 @@ open class PassiveLocationManager: NSObject {
     
     var lastRawLocation: CLLocation?
     
+    /**
+     A closure to be called when manual location was successfully set.
+     
+     - parameter success: Boolean value, which is set to `true` in case if manual location was
+     successfully set, otherwise `false`.
+     */
+    public typealias UpdateLocationCompletionHandler = (_ success: Bool) -> Void
```
1ec5, line 114: Would it be useful to have this completion handler take a `Result<Error, CLLocation>` instead of a Boolean?

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -907,8 +907,11 @@ open class NavigationMapView: UIView {
             let maneuverCoordinate = step.maneuverLocation
             guard step.maneuverType != .arrive else { return }
             
+            let metersPerPointAtLatitude = Projection.metersPerPoint(for: maneuverCoordinate.latitude,
```
1ec5, line 910: This variable name no longer reflects the method name; we can just call it `metersPerPoint`.

Sources/MapboxNavigation/OrnamentsController.swift
```diff
@@ -285,12 +285,12 @@ extension NavigationMapView {
...
-                        if queriedFeature.feature.geometry.geometryType == GeometryType_Line,
-                           let coordinates = queriedFeature.feature.geometry.extractLocationsArray() as? [CLLocationCoordinate2D] {
-                            lineStrings.append(LineString(coordinates))
-                        } else if queriedFeature.feature.geometry.geometryType == GeometryType_MultiLine,
-                                  let coordinates = queriedFeature.feature.geometry.extractLocations2DArray() as? [[CLLocationCoordinate2D]] {
-                            for coordinates in coordinates {
+                        if queriedFeature.feature?.geometry.type == .LineString,
+                           let lineString = queriedFeature.feature?.geometry.value as? LineString {
+                            lineStrings.append(lineString)
+                        } else if queriedFeature.feature?.geometry.type == .MultiLineString,
```
1ec5, line 291: Use a switch/case statement for clarity and better type safety.

This change means that the generated Objective-C headers now use enumerations to differentiate types within a variant. We should take advantage of that change by introducing switch/case statements elsewhere, such as the RoadObjectLocation initializer that was patched in #3250. Look for matches of the regular expression `if\s+[^{]+\.is\w+\(\)` in the codebase.

Cartfile.resolved
```diff
@@ -1,5 +1,5 @@
 binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" "17.1.0"
```
1ec5, line 1: Remember to run `carthage update` to reresolve these dependencies.

CHANGELOG.md
```diff
@@ -11,10 +11,10 @@
...
 * You can now install MapboxNavigation using Swift Package Manager, but you can no longer install it using Carthage. If you previously installed MapboxNavigation using Carthage, use Swift Package Manager instead. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
-* MapboxNavigation now depends on [MapboxMaps v10.0.0-rc.7](https://github.com/mapbox/mapbox-maps-ios/). ([#3297](https://github.com/mapbox/mapbox-navigation-ios/pull/3297))
-* MapboxNavigation now depends on MapboxNavigationNative v65.0.1. ([#3320](https://github.com/mapbox/mapbox-navigation-ios/pull/3320))
-* MapboxNavigation now depends on MapboxCommon v17.1.0. ([#3297](https://github.com/mapbox/mapbox-navigation-ios/pull/3297))
 * MapboxNavigation now depends on MapboxMobileEvents v1.0.3. ([#3320](https://github.com/mapbox/mapbox-navigation-ios/pull/3320))
+* MapboxNavigation now depends on [MapboxMaps v10.0.0-rc.8](https://github.com/mapbox/mapbox-maps-ios/). ([#3342](https://github.com/mapbox/mapbox-navigation-ios/pull/3342))
+* MapboxNavigation now depends on MapboxNavigationNative v66.0.0. ([#3342](https://github.com/mapbox/mapbox-navigation-ios/pull/3342))
```
1ec5, line 16: Were there any changes in MapboxNavigationNative v66.0.0 worth mentioning as improvements over the version used by navigation SDK v1._x_?

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -105,21 +105,37 @@ open class PassiveLocationManager: NSObject {
...
+            completion?(success)
+        }
     }
 
-    private func didUpdate(locations: [CLLocation]) {
+    private func didUpdate(locations: [CLLocation], completion: UpdateLocationCompletionHandler? = nil) {
         for location in locations {
-            navigator.updateLocation(for: FixLocation(location))
+            navigator.updateLocation(for: FixLocation(location)) { success in
+                completion?(success)
```
bamx23, line 137: Should we call completion for every location in the array?

1ec5: Good point, probably should just call it once. Should it be considered successful only if all the updates succeed, or only if the one of them succeeds?

MapboxNavigation-SPM.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -51,26 +51,26 @@
...
       },
       {
         "package": "MapboxNavigationNative",
         "repositoryURL": "https://github.com/mapbox/mapbox-navigation-native-ios.git",
         "state": {
           "branch": null,
-          "revision": "9fd85d49b91be345333c352016ea3b7b48db2bc1",
-          "version": "65.0.2"
+          "revision": "7c8ae18aa21bea2b4e801f0b6aab6a2a21d4aec1",
+          "version": "66.0.0"
```
1ec5, line 73: v66.0.1 is now available. No need to make it a strict requirement, but the resolved files can be updated.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -37,6 +37,14 @@ public struct IndexedRouteResponse {
     }
 }
 
+/**
+ A closure to be called when `RouteLeg` was changed.
+ 
+ - parameter success: Boolean value, which is set to `true` in case if `RouteLeg` was
+ successfully changed, otherwise `false`.
+ */
+public typealias AdvanceLegCompletionHandler = (_ success: Bool) -> Void
```
MaximAlien, line 46: This is a breaking change, which affects both `RouteController` and `LegacyRouteController`. I'm also hesitant about introducing it: previously we did not provide any information to the user about success. Now it's available, which might be useful.

CHANGELOG.md
MaximAlien: Yes, there were several deprecations in `v66.0.0`:
- `Navigator.getStatus()` was deprecated in favor of `NavigatorObserver`.

Sources/MapboxNavigation/OrnamentsController.swift
MaximAlien: Agree. Changes, which apply switch/case instead of if/else statements were applied to `RoadObjectLocation.swift`, `OrnamentsController.swift` and `DistancedRoadObject.swift`.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
MaximAlien: To be serious I'm not sure what would be better here. Looks like completion only in case when all locations were processed would be preferable, as user might not really care about every location in particular. BTW this method is only called from two places: first one is `CLLocationDelegate` and we do not use that completion at all there, second one is in `PassiveLocationManager.updateLocation(_:completion:)` and we provide only one location there, instead of collection.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
MaximAlien: Please check my reply [here](https://github.com/mapbox/mapbox-navigation-ios/pull/3342/files#r710530286). I like your proposal, but this completion handler is only used in one method, which uses only one location instead of their collection, so it'll provide excessive information, as user already has similar value as function parameter.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
1ec5: OK, if we ever do need a `PassiveLocationManager.updateLocations(_:completion:)`, then we can revisit this logic at that time.

CHANGELOG.md
1ec5: `MapboxNavigationNative.Navigator` is fully encapsulated by `MapboxCoreNavigation.Navigator`, which itself is internal, so we can leave that out of the changelog. I was wondering if there were any behavioral changes worth mentioning, but I guess this release was relatively uneventful on that front.

Sources/MapboxCoreNavigation/Router.swift
1ec5: Similar to https://github.com/mapbox/mapbox-navigation-ios/pull/3342#discussion_r709664302, it would be helpful to define this completion handler to accept a `Result<Error, RouteProgress>`. Even though the navigator doesn’t provide us with an Error in the failure case, we can create a basic one on the spot. In the future, if the navigator does start providing more information, we can naturally pass it along without having to break the public API. But if we don’t go through that trouble now, we can introduce a parallel method and completion handler type when the time comes.

Remember to mention the change in the changelog.

1ec5 (approved):

1ec5:
[These tests are failing](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/5497/workflows/bd9f291e-89fc-40db-899f-f0d6d782eb6a/jobs/44999) on the SPM build:

```
Failing tests:
	MapboxCoreNavigationTests:
		MapboxCoreNavigationTests.testNewStep()
		NavigationServiceTests.testProactiveRerouting()
		NavigationServiceTests.testReroutingFromALocationSendsEvents()
		RouteControllerTests.testRerouteAfterArrival()
```

Zeroing in on one:

>     ✗ testProactiveRerouting, Asynchronous wait failed: Exceeded timeout of 10 seconds, with unfulfilled expectations: "Expect notification 'RouteControllerDidReroute' from <MapboxCoreNavigation.RouteController: 0x7fdade1af8f0>".

This test works by setting up a route request from 38.853108, -77.043331 (DCA) to 38.910736, -76.966906 (Arboretum) and running a full trace along that fixture. But the fixture looks pretty weird; I wonder if that is causing a reroute at a different time than expected:

DCA-Arboretum

Sources/MapboxCoreNavigation/Router.swift
MaximAlien: `UpdateLocationCompletionHandler` and `AdvanceLegCompletionHandler` were modified to use `Result` instead of `Boolean`. Error enums are internal, which gives us an opportunity to change them in future.

CHANGELOG.md
```diff
@@ -178,6 +178,7 @@
 * Feedback categories and subcategories for active navigation were changed. "Incorrect visual" category was renamed to "Looks Incorrect". For the full updated list of supported categories see enum `ActiveNavigationFeedbackType`. ([#3339]((https://github.com/mapbox/mapbox-navigation-ios/pull/3339))
 * Fixed bundle names retrieval in `URLSession.userAgent`. ([#3335]((https://github.com/mapbox/mapbox-navigation-ios/pull/3335))
 * Fixed IndexedRouteResponse handling after rerouting. ([#3344]((https://github.com/mapbox/mapbox-navigation-ios/pull/3344))
+* Renamed `Router.advanceLegIndex()` to `Router.advanceLegIndex(completionHandler:)`. Added `AdvanceLegCompletionHandler` and `UpdateLocationCompletionHandler`, which allow to report result of advancing route leg in `Router` and updating location in `PassiveLocationManager` respectively. ([#3344]((https://github.com/mapbox/mapbox-navigation-ios/pull/3344))
```
1ec5, line 181:
```suggestion
* Renamed the `Router.advanceLegIndex()` method to `Router.advanceLegIndex(completionHandler:)` and the `PassiveLocationDataSource.updateLocation(_:)` method to `PassiveLocationManager.updateLocation(_:completionHandler:)`. These methods are now asynchronous, and their completion handlers indicate whether the operation succeeded. ([#3344](https://github.com/mapbox/mapbox-navigation-ios/pull/3344))
```

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -213,24 +214,37 @@ open class RouteController: NSObject {
...
     }
     
     /// updateRouteLeg is used to notify nav-native of the developer changing the active route-leg.
-    private func updateRouteLeg(to value: Int) {
+    private func updateRouteLeg(to value: Int, completionHandler: AdvanceLegCompletionHandler? = nil) {
         let legIndex = UInt32(value)
-        if navigator.changeRouteLeg(forRoute: 0, leg: legIndex), let status = navigator.getStatus() {
-            updateIndexes(status: status, progress: routeProgress)
+        
+        navigator.changeRouteLeg(forRoute: 0, leg: legIndex) { [weak self] success in
```
1ec5, line 232: The removal of this synchronous call to `MapboxNavigationNative.Navigator.getStatus()` completes #2846 and supersedes #3269.

1ec5 (approved):

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -630,7 +630,7 @@ class NavigationServiceTests: TestCase {
         XCTAssertTrue(delegate.recentMessages.contains("navigationService(_:didArriveAt:)"))
     }
 
-    func testProactiveRerouting() {
+    func disabled_testProactiveRerouting() {
```
MaximAlien, line 633: Revival of several disabled tests will be tackled as a tail-work in this ticket: https://github.com/mapbox/mapbox-navigation-ios/issues/3375.

PR #3344, opened 2021-09-09
Fix IndexedRouteResponse handling after rerouting

OttyLab:
## Summary
Reroute does not update `indexedRouteResponse` and refresh uses old route's identifier to update route annotations. This provides unexpected speed limit, congestion information display.

## Detail

Reroute does not update [`indexedRouteResponse`](https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-beta.25/Sources/MapboxCoreNavigation/RouteController.swift#L39). `indexedRouteResponse.routeResponse.identifier` represents route's identifier and used [when refreshing](https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-beta.25/Sources/MapboxCoreNavigation/Router.swift#L173). 

Therefore, `indexedRouteResponse` needs to be updated [here](https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-beta.25/Sources/MapboxCoreNavigation/RouteController.swift#L584-L589).

## Screenshot
After the reroute, speed limit shows 90km/h but 1 min later it becomes 70km/h. This road is 90km/h and 70km/h is wrong.

https://user-images.githubusercontent.com/13183117/132643775-4662960c-4ce0-4a7d-9bc9-6f4158101a54.mp4

## Expected behavior
https://user-images.githubusercontent.com/13183117/132643913-755627bd-211f-4c36-9e5e-d882737c32c1.mp4

### DIFF
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -597,6 +597,7 @@ extension RouteController: Router {
                 guard case let .route(routeOptions) = response.options else { return } //TODO: Can a match hit this codepoint?
                 strongSelf._routeProgress = RouteProgress(route: route, options: routeOptions, legIndex: 0)
                 strongSelf._routeProgress.currentLegProgress.stepIndex = 0
+                strongSelf.indexedRouteResponse = .init(routeResponse: response, routeIndex: 0)
                 strongSelf.announce(reroute: route, at: location, proactive: false)
                 
             case let .failure(error):


1ec5:
Thanks for contributing this fix! This PR includes some additional debugging code that would need to be cleaned up, but I went ahead and rolled your commit into #3345, which also includes more substantial changes in this area. Please have a look at that PR. Hopefully we can land the whole thing soon, but if not, then we can clean up this PR and merge it instead.

Example/ViewController.swift
```diff
@@ -437,6 +437,7 @@ class ViewController: UIViewController {
         self?.navigationMapView.removeWaypoints()
         self?.response = response
         
+        print("!! Identifier1=\(response.identifier)")//@DEBUG
```
1ec5, line 440: Remove this debug statement.

Example/ViewController.swift
```diff
@@ -500,7 +501,8 @@ class ViewController: UIViewController {
     }
 
     func navigationService(response: RouteResponse, routeIndex: Int, options: RouteOptions) -> NavigationService {
-        let mode: SimulationMode = simulationButton.isSelected ? .always : .onPoorGPS
+        //let mode: SimulationMode = simulationButton.isSelected ? .always : .onPoorGPS
+        let mode: SimulationMode = simulationButton.isSelected ? .always : .never //@DEBUG
```
1ec5, line 505: Revert this debugging-only change.

MapboxNavigation-SPM.xcodeproj/project.pbxproj
```diff
@@ -73,6 +73,7 @@
 		C53F2F0720EBC95600D9798F /* Example.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = Example.app; sourceTree = BUILT_PRODUCTS_DIR; };
 		C57AD98920EAA39A0087B24B /* Entitlements.plist */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.plist.xml; path = Entitlements.plist; sourceTree = "<group>"; };
 		C5D9800C1EFA8BA9006DBF2E /* CustomViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = CustomViewController.swift; path = Example/CustomViewController.swift; sourceTree = "<group>"; };
+		D27E7C5F26E9E72300CA5926 /* 212_3.gpx */ = {isa = PBXFileReference; lastKnownFileType = text.xml; path = 212_3.gpx; sourceTree = "<group>"; };
```
1ec5, line 76: This file isn’t included in the PR, so we’ll need to leave it out of the project.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -598,6 +598,7 @@ extension RouteController: Router {
                 strongSelf._routeProgress = RouteProgress(route: route, options: routeOptions, legIndex: 0)
                 strongSelf._routeProgress.currentLegProgress.stepIndex = 0
                 strongSelf.announce(reroute: route, at: location, proactive: false)
+                strongSelf.indexedRouteResponse = .init(routeResponse: response, routeIndex: 0)
```
1ec5, line 601: This is correct for now. #3345 will change the behavior of the method that this closure is passed into, so that the closest matching route will no longer be guaranteed to be at index 0; instead, the closure will receive the index of the closest matching route.

1ec5 (commented):
On second thought, #3345 has amassed so many changes that we’re a bit wary of landing it literally right before v2.0.0-rc.1, so we’ll clean up this more targeted fix, land it for v2.0.0-rc.1, and land #3345 right after to squeeze in additional testing before v2.0.0-rc.2.

1ec5 (approved):

1ec5:
Updated the changelog in e68d748c9f295d3cc67dee725e3c68b33ce8fedc.

Issue #3276, opened 2021-08-18
Camera shaking and migrating with CarPlay in simulation mode.


ShanMa1991:
**Mapbox Navigation SDK version:** main

### Steps to reproduce

 1. open CarPlay
 2. choose multiple waypoints in mobile
 3. start navigation in simulation mode.
 4. start several times as above.

### Expected behavior
The carPlay and mobile the camera direction is not changing.

### Actual behavior
Large shaking of camera in both mobile and CarPlay even on straight road. Sometimes the camera is shifting to a different and remote location and then migrating back.
![camera](https://user-images.githubusercontent.com/48976398/129976923-684cc743-1ef9-43f3-bab4-44d56410c047.gif)

ShanMa1991:
It may also be related with the navigator status issue, when repeated couple of times of multi leg routes, it would lead to the crash as following:
```
#0	0x0000000103e09b50 in _swift_runtime_on_report ()
#1	0x0000000103e83e8c in _swift_stdlib_reportFatalErrorInFile ()
#2	0x0000000103b1b8aa in _assertionFailure(_:_:file:line:flags:) ()
#3	0x0000000100ceeb32 in RouteProgress.legIndex.didset at /Users/shanma/Documents/Code/mapbox-navigation-ios/Sources/MapboxCoreNavigation/RouteProgress.swift:24
#4	0x0000000100ceedba in RouteProgress.legIndex.setter ()
#5	0x0000000100cdc696 in RouteController.updateIndexes(status:progress:) at /Users/shanma/Documents/Code/mapbox-navigation-ios/Sources/MapboxCoreNavigation/RouteController.swift:315
#6	0x0000000100cda6e2 in RouteController.update(to:) at /Users/shanma/Documents/Code/mapbox-navigation-ios/Sources/MapboxCoreNavigation/RouteController.swift:263
#7	0x0000000100cda225 in closure #1 in RouteController.navigationStatusDidChange(_:) at /Users/shanma/Documents/Code/mapbox-navigation-ios/Sources/MapboxCoreNavigation/RouteController.swift:251
#8	0x0000000100c29230 in thunk for @escaping @callee_guaranteed () -> () ()
#9	0x000000010496e578 in _dispatch_call_block_and_release ()
#10	0x000000010496f74e in _dispatch_client_callout ()
#11	0x000000010497db3f in _dispatch_main_queue_callback_4CF ()
[...]

PR #3345, opened 2021-09-09
Fix inconsistent route progress and index after rerouting

### LINKED ISSUE #3343: Parallel copies of route in Router can get out of sync
Router implements store both an IndexedRouteResponse (that is, a RouteResponse plus a route index) and a RouteProgress (that is, a Route plus a leg index etc.) side by side. As a result, they effectively hold two parallel copies of the current Route that can get out of sync. For example, after a reroute, RouteController sets the `routeProgress` property but neglects to update the `indexedRouteResponse` property:

https://github.com/mapbox/mapbox-navigation-ios/blob/3c81bdb6df768707073f9d96188ffd92d6a79985/Sources/MapboxCoreNavigation/RouteController.swift#L587-L589

This contrasts to LegacyRouteController, which does happen to set both properties:

https://github.com/mapbox/mapbox-navigation-ios/blob/5e9954da1a5b4523dfe032a491d1beb6020abd80/Sources/MapboxCoreNavigation/LegacyRouteController.swift#L62-L64

In both RouteController and LegacyRouteController, `routeProgress` is publicly settable but `indexedRouteResponse` is not. This makes it entirely possible for client code to have inconsistent behavior depending on how it gets the route from the Router.

/cc @mapbox/navigation-ios

1ec5:
After rerouting or proactive rerouting, keep the array of routes in the same order as in the Directions API response. Instead of reshuffling the array to put the most similar route first, update the route index to point to the most similar route (except when proactively rerouting to the most optimal route). This change makes the post-rerouting state more consistent with the pre-rerouting state as of #3182.

`RouteController.routeProgress` and `LegacyRouteController.routeProgress` are no longer writable. Synchronize `RouteController.indexedRouteResponse` after rerouting to ensure consistency with `routeProgress`.

Consolidated calls to `NavigationServiceDelegate.navigationService(_:willRerouteFrom:)` and `NavigationServiceDelegate.navigationService(_:didRerouteAlong:at:proactive:)` and the posting of `Notification.Name.routeControllerWillReroute` and `Notification.Name.routeControllerDidReroute` to ensure they occur consistently around reroutes. When posting these notifications, pass in a proactive rerouting flag directly instead of relying on a parallel `didFindFasterRoute` flag that was difficult to reason about.

Fixes #3343 and potentially other bugs.

/cc @mapbox/navigation-ios


Sources/MapboxCoreNavigation/LegacyRouteController.swift
```diff
@@ -375,11 +357,12 @@ open class LegacyRouteController: NSObject, Router, InternalRouter, CLLocationMa
...
                  ])
                  return
-            case let .success(response):
-                guard case let .route(options) = response.options, !(response.routes?.isEmpty ?? true) else {
-                    return
-                }
-                strongSelf.updateRoute(with: .init(routeResponse: response, routeIndex: 0), routeOptions: options) // unconditionally getting the first route above
+            case let .success(indexedResponse):
+                let response = indexedResponse.routeResponse
+                guard let route = response.routes?[indexedResponse.routeIndex] else { return }
```
azarovalex, line 362: Should we check array bounds here and in `RouteController`?

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -245,72 +241,89 @@ extension InternalRouter where Self: Router {
...
-            self?.isRerouting = false
-            
-            guard case let .success(response) = result else {
-                return
-            }
-            guard let route = response.routes?.first else { return }
-            
-            self?.lastProactiveRerouteDate = nil
+        calculateRoutes(from: location, along: routeProgress) { [weak self] (session, result) in
+            guard let strongSelf = self else { return }
```
S2Ler, line 245: nit: `guard let self = self else { return }` but we use strongSelf as well. so just For you information.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -245,72 +241,89 @@ extension InternalRouter where Self: Router {
...
+                guard let firstLeg = route.legs.first, let firstStep = firstLeg.steps.first else {
+                    return
                 }
                 
-                self?.indexedRouteResponse = .init(routeResponse: response, routeIndex: 0)
-                self?.setRoute(route, proactive: true, routeOptions: routeOptions)
+                let routeIsFaster = firstStep.expectedTravelTime >= RouteControllerMediumAlertInterval &&
+                    currentUpcomingManeuver == firstLeg.steps[1] && route.expectedTravelTime <= 0.9 * durationRemaining
+                
+                if routeIsFaster {
```
azarovalex, line 264: maybe use `guard` here? the level of nesting is too deep imho
also, I want to add a linter on the length of the line in the near future

S2Ler:
One of the tests are failing.

azarovalex (approved):
lgtm, but if we add changes in the latest moment before the RC, we need to test them at least manually

Sources/MapboxCoreNavigation/Router.swift
1ec5: Thanks, fixed.

Sources/MapboxCoreNavigation/Router.swift
1ec5: I changed it back to a guard statement.

Sources/MapboxCoreNavigation/LegacyRouteController.swift
1ec5: If `indexedResponse`’s index is out of bounds with respect to its own response, that would be a bug in whatever initialized the IndexedRouteResponse. In this case, it would be `Router.calculateRoutes(from:along:completion:)`, but even ignoring rerouting, the application could’ve passed in an inconsistent IndexedRouteResponse in the first place. It would be reasonable to add a check, but we should do that earlier and more systematically, probably as a separate PR.

1ec5: I added a couple strategic assertions, not as early as I think we probably should to catch developer error, but at least enough to catch internal inconsistency before it ricochets across the SDK. This forced me to revisit one spot where we had been explicitly creating an out-of-bounds IndexedRouteResponse, in the case where the rerouting request came back with no routes. Now that’s an error; everything that kicks off the reroute will abort the reroute.

1ec5:
> One of the tests are failing.

This is strange; the test doesn’t seem like it should exercise anything that I touched, but it does fail on two different build machines:

```
MapboxNavigationTests.NavigationViewControllerTests
  testDestinationAnnotationUpdatesUponReroute, XCTAssertTrue failed - New destination annotation does not exist on map
  /Users/distiller/project/Tests/MapboxNavigationTests/NavigationViewControllerTests.swift:317
  
        // Verify that there is a destination on the second route.
        XCTAssert(newAnnotations
                    .compactMap { $0.geometry.value as? Turf.Point }
  
```

```
/Users/distiller/project/Tests/MapboxNavigationTests/NavigationViewControllerTests.swift:317: error: -[MapboxNavigationTests.NavigationViewControllerTests testDestinationAnnotationUpdatesUponReroute] : XCTAssertTrue failed - New destination annotation does not exist on map
2021-09-09 08:32:57.941844+0000 xctest[3517:69986] <nav-native> PersistentConfig path: /Users/distiller/Library/Developer/CoreSimulator/Devices/DED5E213-8043-44B3-9F26-6D4CB4FED885/data/Library/Application Support/.mapbox/tile_store/navigation/config.json
2021/09/09 08:32:57.942135 [WARN] No metadata found for tiles
2021/09/09 08:32:57.942284 [WARN] No metadata found for tiles
2021-09-09 08:32:57.942432+0000 xctest[3517:69986] <nav-native> Tiles path = "/Users/distiller/Library/Developer/CoreSimulator/Devices/DED5E213-8043-44B3-9F26-6D4CB4FED885/data/Library/Application Support/.mapbox/tile_store"
2021-09-09 08:32:57.942602+0000 xctest[3517:69986] <nav-native> TileStore instance was passed
2021-09-09 08:32:57.942827+0000 xctest[3517:69986] <nav-native> Endpoint config = "{"host":"https://api.mapbox.com","isFallback":false,"minDiffInDaysToConsiderServerVersion":0,"navigatorVersion":"65.0.1","token":"***...*","userAgent":"MapboxNavigation/2.0.0 MapboxCoreNavigation/2.0.0 iOS/14.4.0 (x86_64)","version":"","versionBeforeFallback":""}"
2021-09-09 08:32:57.943033+0000 xctest[3517:69986] <nav-native> Creating navigator v.65.0.1
2021/09/09 08:32:57.943218 [WARN] No metadata found for tiles
2021/09/09 08:32:57.943636 [WARN] No metadata found for tiles
2021/09/09 08:32:57.944111 [WARN] No metadata found for tiles
2021/09/09 08:32:57.944225 [WARN] No metadata found for tiles
2021/09/09 08:32:57.944343 [WARN] No metadata found for tiles
Test Case '-[MapboxNavigationTests.NavigationViewControllerTests testDestinationAnnotationUpdatesUponReroute]' failed (4.353 seconds).
```

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -573,30 +564,28 @@ extension RouteController: Router {
...
-            case let .success(response):
-                guard let route = response.routes?.first else { return }
+            case let .success(indexedResponse):
+                let response = indexedResponse.routeResponse
+                guard let route = response.routes?[indexedResponse.routeIndex] else { return }
                 guard case let .route(routeOptions) = response.options else { return } //TODO: Can a match hit this codepoint?
-                strongSelf._routeProgress = RouteProgress(route: route, options: routeOptions, legIndex: 0)
-                strongSelf._routeProgress.currentLegProgress.stepIndex = 0
+                strongSelf.routeProgress = RouteProgress(route: route, options: routeOptions)
+                strongSelf.indexedRouteResponse = indexedResponse
```
1ec5, line 588: The route index and step index default to 0, so I removed the explicit call to change the step index to 0.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -622,6 +611,7 @@ extension RouteController: Router {
 
         let routeOptions = routeOptions ?? routeProgress.routeOptions
         routeProgress = RouteProgress(route: route, options: routeOptions)
+        announce(reroute: route, at: location, proactive: false)
```
1ec5, line 614: This was the one other place where the did-reroute notification needed to get inlined. It’s weird that an application explicitly setting the route will get a notification back telling it what it just did, but that’s the only way that various parts of MapboxNavigation will be able to respond to the new route being set. As a case in point, the addition of this line fixes the integration test failure.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -268,47 +266,54 @@ extension InternalRouter where Self: Router {
...
                 
-                self?.indexedRouteResponse = .init(routeResponse: response, routeIndex: 0)
-                self?.setRoute(route, proactive: true, routeOptions: routeOptions)
+                // Prefer the most optimal route (the first one) over the route that matched the original choice.
+                self.indexedRouteResponse = .init(routeResponse: response, routeIndex: 0)
+                
+                // If the upcoming maneuver in the new route is the same as the current upcoming maneuver, don’t announce it.
+                // FIXME: There must be a better way to skip a redundant initial instruction than to assume the old route’s spoken instruction index is appliable to the new route.
+                let spokenInstructionIndex = self.routeProgress.currentLegProgress.currentStepProgress.spokenInstructionIndex
+                self.routeProgress = RouteProgress(route: route, options: routeOptions ?? self.routeProgress.routeOptions, legIndex: 0, spokenInstructionIndex: spokenInstructionIndex)
```
1ec5, line 275: While inlining `setRoute(_:proactive:routeOptions:)` here, I restored an old comment from before #1986 that explained why we try to persist the spoken instruction index in the new route progress. Apparently we consider it confusing to hear the “Head northwest on Main Street” instruction out of the blue without the rerouting chime.

This bit of gymnastics dates back to when we originally implemented proactive rerouting in #346. At the time, there was no concept of an instruction index, only an “alert level” that increased as the user neared the maneuver point. Maintaining the alert level made sense as a way to suppress that initial instruction, since the alert level would’ve been guaranteed to be nonzero at any time beyond the beginning of the old step.

We migrated to server-generated instructions in #614, eliminating the concept of alert levels in favor of instruction indices. 903e0207722f889abffe191dde52fdc8799331e8 migrated this code to persist the instruction index, but that introduced a potential problem: unlike alert levels, there’s theoretically an arbitrary number of instruction indices that could vary from one step to the next, or between two steps belonging to two different routes.

I don’t know for sure that this persistence is causing any problems, so I’ve retained the behavior for now, but it needs a closer look. There’s got to be a better way to suppress the initial departure instruction, if we still think that’s necessary.

1ec5:
[Another failure](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/5361/workflows/c9bbc6f3-035c-4929-8373-126e3ab61a28/jobs/44029):

```
MapboxCoreNavigationTests.BillingHandlerUnitTests
  testFailedMauBillingDoNotStopSession, Asynchronous wait failed: Exceeded timeout of 1 seconds, with unfulfilled expectations: "Billing event triggered".
  /Users/distiller/project/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift:182
  
        handler.beginBillingSession(for: .activeGuidance, uuid: sessionUUID)
        waitForExpectations(timeout: 1, handler: nil)
        XCTAssertEqual(handler.sessionState(uuid: sessionUUID), .running)
  
```

S2Ler:
> [Another failure](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/5361/workflows/c9bbc6f3-035c-4929-8373-126e3ab61a28/jobs/44029), which might be spurious:
> 
> ```
> MapboxCoreNavigationTests.BillingHandlerUnitTests
>   testFailedMauBillingDoNotStopSession, Asynchronous wait failed: Exceeded timeout of 1 seconds, with unfulfilled expectations: "Billing event triggered".
>   /Users/distiller/project/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift:182
>   
>         handler.beginBillingSession(for: .activeGuidance, uuid: sessionUUID)
>         waitForExpectations(timeout: 1, handler: nil)
>         XCTAssertEqual(handler.sessionState(uuid: sessionUUID), .running)
>   
> ```

Can you please increase timeout. It might be that machine became slow.

1ec5:
> lgtm, but if we add changes in the latest moment before the RC, we need to test them at least manually

Since we’re literally on the cusp of releasing v2.0.0-rc.1, I’ve cleaned up #3344 so we can land just that targeted fix for #3343 and the related bugs. We’ll land this PR right after v2.0.0-rc.2 to give us some additional time to test the more comprehensive fixes before releasing v2.0.0-rc.2.

1ec5:
[This build](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/5420/workflows/06c964af-083c-476b-8fa6-04e9f4802451/jobs/44438) flags some breaking changes:

```
**** BREAKING CHANGES DETECTED ****

Breaking changes in 'calculateRoutes(from:along:completion:)'
*new* method: `calculateRoutes(from:along:completion:)`

Breaking changes in 'RouteController'
*modified* var: `routeProgress` in `RouteController`
```

The new method is internal, and the modified variable is intentional (we release-noted it in v2.0.0-rc.1).

S2Ler:
I've added all these cases to https://github.com/mapbox/mapbox-navigation-ios/issues/3353. Meanwhile we can either disable diff reporter or do a dance with `.base_api`.

ShanMa1991 (commented):
When I test this pr on `Example-CarPlay` with multiple route legs under simulation mode and repeated several times. There's a failure occurring as following:

https://github.com/mapbox/mapbox-navigation-ios/blob/f462297c0c68c21dcc4c9e4d73bedf3dce0d013f/Sources/MapboxCoreNavigation/RouteController.swift#L310-L318

It happened in a two legs route which would advance from the first leg to the next leg. The `routeProgress.legIndex` as `1`.

But the route in the `routeProgress` only has one leg , and the `oldLegIndex = 0`. Then it triggers the assertion failure of  `legIndex < route.legs.endIndex` after the `progress.legIndex = newLegIndex`.

So my concern is that why the routeProgress contains the route only has 1 leg which in fact should have 2 on this branch. Is it related with the `indexedRouteResponse`?

1ec5:
> When I test this pr on Example-CarPlay with multiple route legs under simulation mode and repeated several times. There's a failure occurring as following

I’m unable to reproduce this issue when simulating a route with two legs in Example-CarPlay on this branch.

destination 3

Did you use the navigation SDK’s built-in route simulation feature or the iOS or Xcode route simulation feature?

ShanMa1991:
> Did you use the navigation SDK’s built-in route simulation feature or the iOS or Xcode route simulation feature? Can you share the coordinates of the waypoints you used?

Yes, I'm using the navigation SDK’s built-in route simulation feature with the waypoints inside the SF downtown area. This issue could be reproduced with multiple times of 2-leg routes navigation star and cancel, along with the issue #3276 .

ShanMa1991:
The inconsistent route progress and index issue in multi-leg rout navigation under simulation mode will be tracked in #3276 . Thanks for the testing !

ShanMa1991 (approved):

1ec5:
After squashing and merging, [this build](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/5489/workflows/45dada3d-679c-4c2b-85aa-49d93bb25142/jobs/44952) failed:

```
Test Case '-[MapboxCoreNavigationTests.BillingHandlerUnitTests testSessionBeginFailed]' started.
Warning: Using NavigationSettings.initialize(directions:tileStoreConfiguration:) after corresponding variables was initialized. Possible reasons: Initialize called more than once, or the following properties was accessed before initialization: `tileStoreConfiguration`, `directions`. This might result in an undefined behaviour. 
2021-09-16 23:49:38.511471+0000 xctest[1677:62393] [Billing] Trip session isn't started. Please check that you have the correct Mapboox Access Token
2021-09-16 23:49:38.521475+0000 xctest[1677:62689] <nav-native> No PersistentConfig found at /Users/distiller/Library/Developer/CoreSimulator/Devices/DED5E213-8043-44B3-9F26-6D4CB4FED885/data/Library/Application Support/.mapbox/tile_store/navigation/config.json
2021-09-16 23:49:38.521783+0000 xctest[1677:62689] <nav-native> Unable to find the latest version of routing tiles!
2021-09-16 23:49:38.522037+0000 xctest[1677:62689] <nav-native> Async config init - Unable to determine the routing tiles version
/Users/distiller/project/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift:159: error: -[MapboxCoreNavigationTests.BillingHandlerUnitTests testSessionBeginFailed] : XCTAssertEqual failed: ("stopped") is not equal to ("running")
Test Case '-[MapboxCoreNavigationTests.BillingHandlerUnitTests testSessionBeginFailed]' failed (0.276 seconds).
```

@S2Ler, should we increase the timeout for this test too, or is there a deeper issue?

https://github.com/mapbox/mapbox-navigation-ios/blob/1e3598766ef2f6656f9e896d4f2a4a25896bd5df/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift#L160

S2Ler:
> After squashing and merging, [this build](https://app.circleci.com/pipelines/github/mapbox/mapbox-navigation-ios/5489/workflows/45dada3d-679c-4c2b-85aa-49d93bb25142/jobs/44952) failed. Rerunning the build succeeded.
> 
> ```
> Test Case '-[MapboxCoreNavigationTests.BillingHandlerUnitTests testSessionBeginFailed]' started.
> Warning: Using NavigationSettings.initialize(directions:tileStoreConfiguration:) after corresponding variables was initialized. Possible reasons: Initialize called more than once, or the following properties was accessed before initialization: `tileStoreConfiguration`, `directions`. This might result in an undefined behaviour. 
> 2021-09-16 23:49:38.511471+0000 xctest[1677:62393] [Billing] Trip session isn't started. Please check that you have the correct Mapboox Access Token
> 2021-09-16 23:49:38.521475+0000 xctest[1677:62689] <nav-native> No PersistentConfig found at /Users/distiller/Library/Developer/CoreSimulator/Devices/DED5E213-8043-44B3-9F26-6D4CB4FED885/data/Library/Application Support/.mapbox/tile_store/navigation/config.json
> 2021-09-16 23:49:38.521783+0000 xctest[1677:62689] <nav-native> Unable to find the latest version of routing tiles!
> 2021-09-16 23:49:38.522037+0000 xctest[1677:62689] <nav-native> Async config init - Unable to determine the routing tiles version
> /Users/distiller/project/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift:159: error: -[MapboxCoreNavigationTests.BillingHandlerUnitTests testSessionBeginFailed] : XCTAssertEqual failed: ("stopped") is not equal to ("running")
> Test Case '-[MapboxCoreNavigationTests.BillingHandlerUnitTests testSessionBeginFailed]' failed (0.276 seconds).
> ```
> 
> @S2Ler, should we increase the timeout for this test too, or is there a deeper issue?
> 
> https://github.com/mapbox/mapbox-navigation-ios/blob/1e3598766ef2f6656f9e896d4f2a4a25896bd5df/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift#L160

yes, increasing to 3 or 5 should fix it.

PR #3349, opened 2021-09-09
Fix unit tests for M1 chip

S2Ler:
Instead of preconditions I used to default to save range of values so that the test don't crash.

I think the fix is an acceptable change.

### DIFF
--- a/Sources/MapboxNavigation/FollowingCameraOptions.swift
+++ b/Sources/MapboxNavigation/FollowingCameraOptions.swift
@@ -5,7 +5,7 @@ import CoreLocation
  `NavigationViewportDataSource` in `NavigationCameraState.following` state.
  */
 public struct FollowingCameraOptions: Equatable {
-    
+
     /**
      Pitch, which will be taken into account when preparing `CameraOptions` during active guidance
      navigation.
@@ -62,7 +62,7 @@ public struct FollowingCameraOptions: Equatable {
      
      Defaults to `true`.
      */
-    public var zoomUpdatesAllowed = true
+    public var zoomUpdatesAllowed: Bool = true
     
     /**
      If `true`, `NavigationViewportDataSource` will continuously modify `CameraOptions.bearing` property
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/FollowingCameraOptions.swift
```diff
@@ -5,21 +5,23 @@ import CoreLocation
...
      
      Defaults to `45.0` degrees.
+
+     - Invariant: Acceptable range of values is `0...85`.
      */
     public var defaultPitch: Double = 45.0 {
         didSet {
             if defaultPitch < 0.0 {
-                preconditionFailure("Lower bound of the pitch should not be lower than 0.0")
+                defaultPitch = 0
```
bamx23, line 20: Can we use `assertionFailure` here?

S2Ler: It will break tests as well for me.

bamx23: Can we disable asserts in tests? Like `-assert-config Release` in "Other Swift Flags".

MaximAlien: I'd really like to keep assertions or preconditions. Right now we're basically silently modifying value in case if it's not valid, this breaks the contract.

S2Ler: > Can we disable asserts in tests? Like -assert-config Release in "Other Swift Flags".

Can you explain how to use this with SPM?

bamx23: > Can you explain how to use this with SPM?

Not really( 

Maybe we should use wrapped asserts that can be disabled in tests in runtime? I agree with @MaximAlien that it should be possible to use asserts/preconditions in the SDK.

S2Ler: > this breaks the contract.

the contact says that acceptable values are `0...85`, so, there is no a violation. 

> that it should be possible to use asserts/preconditions in the SDK.

ok, I will remove tests which cause a crash because it blocks my development and we can fix this separately.

S2Ler: #3354

PR #3359, opened 2021-09-13
Add ability to record NavNative history in Example

S2Ler:
It is beneficial for sharing history files with NavNative team.


azarovalex (approved):
lgtm, should we also record the history in active navigation?

S2Ler:
> lgtm, should we also record the history in active navigation?

I think it will, it just that you can't start/stop it from Active Guidance screen which is an acceptable limitation to me.

PR #3361, opened 2021-09-13
Use snapped location for reroutes

azarovalex:
In the case of off-route, we have to send the initial coordinate to Directions API. Currently we use the raw location for it (i.e. we get bearing and coordinate from the system) what can lead to awful use experience (route built in the wrong direction, problems on parallel roads, etc). For directions API requests we should always use coordinate, bearing, and other parameters provided by map matcher(i.e. by NN).

### DIFF
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -278,7 +278,7 @@ open class RouteController: NSObject {
         updateRoadName(status: status)
         
         if willReroute {
-            reroute(from: location, along: routeProgress)
+            reroute(from: CLLocation(status.location), along: routeProgress)
         }
 
         if status.routeState != .complete {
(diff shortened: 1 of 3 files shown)


S2Ler:
Looks like we have more work to do with tests now. Hopefully we just need to update the tests are it is not a regression.

azarovalex:
> Looks like we have more work to do with tests now. Hopefully, we just need to update the tests are it is not a regression.

Yeah, the problem with the test is that we don't know in advance the location of the reroute, because we get it from the NN.
I solved this by verifying that the distance between the locations is less than 10 meters.

Udumft (approved):

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -278,7 +278,7 @@ open class RouteController: NSObject {
         updateRoadName(status: status)
         
         if willReroute {
-            reroute(from: location, along: routeProgress)
+            reroute(from: CLLocation(status.location), along: routeProgress)
```
MaximAlien, line 281: Is it enough to use snapped location only for this case? I can see that there are also a couple of places in this method, where it'd be preferable to use snapped location: e.g. we check whether reroute will be performed using raw location, but to make a request we use snapped one.

PR #3362, opened 2021-09-13
Don't update Navigator with locations after pause

### LINKED ISSUE #3348: [Bug]: Paused billing trip sessions might result in requests made inside NavNative billed per request

v2.0.0-rc.1

### Steps to reproduce

1. Start passive navigation with PassiveLocationManager
2. Pause trip session with `pauseTripSession`
3. Continue driving in passive navigation

### Expected behavior

NavNative's Navigator isn't updated with locations so that it won't try to download tiles. 

### Actual behaviour

NavNative's Navigator continues to work and download tiles.

repeatable

# Sub tasks
- [ ] Integrate pause/resume from NavNative.

S2Ler:
Fixes #3348

### DIFF
--- a/Sources/MapboxCoreNavigation/PassiveLocationManager.swift
+++ b/Sources/MapboxCoreNavigation/PassiveLocationManager.swift
@@ -132,6 +132,12 @@ open class PassiveLocationManager: NSObject {
     }
 
     private func didUpdate(locations: [CLLocation], completion: UpdateLocationCompletionHandler? = nil) {
+        // NOTE: We should stop updating `navigator` with locations if billing session isn't running.
+        //       To be replaced by `Navigator` pause/resume functionality once available.
+        guard BillingHandler.shared.sessionState(uuid: sessionUUID) == .running else {
+            completion?(.failure(PassiveLocationManagerError.sessionIsNotRunning)); return
+        }
+        
         for location in locations {
             navigator.updateLocation(for: FixLocation(location)) { success in
                 let result: Result<CLLocation, Error>
@@ -383,4 +389,6 @@ extension TileEndpointConfiguration {
 
 enum PassiveLocationManagerError: Error {
     case failedToChangeLocation
+    /// Location updates are not possible when session is paused.
+    case sessionIsNotRunning
 }
--- a/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift
+++ b/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift
@@ -264,7 +264,6 @@ final class BillingHandlerUnitTests: TestCase {
         class UpdatesSpy: PassiveLocationManagerDelegate {
             var onProgressUpdate: (() -> Void)?
 
-
             func passiveLocationManager(_ manager: PassiveLocationManager,
                                         didUpdateLocation location: CLLocation,
                                         rawLocation: CLLocation) {
@@ -299,6 +298,12 @@ final class BillingHandlerUnitTests: TestCase {
             .beginBillingSession(.freeDrive),
             .pauseBillingSession(.freeDrive)
         ])
+
+        XCTAssertNil(passiveLocationManager.rawLocation, "Location updates should be blocked")
+
+        passiveLocationManager.resumeTripSession()
+        locationManager.tick()
+        XCTAssertNotNil(passiveLocationManager.rawLocation)
     }
 
     func testTokens() {


S2Ler:
Probably will implement https://github.com/mapbox/mapbox-navigation-ios/issues/3348#issuecomment-918620148 as well.

S2Ler:
> Probably will implement [#3348 (comment)](https://github.com/mapbox/mapbox-navigation-ios/issues/3348#issuecomment-918620148) as well.

implemented

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -97,18 +97,22 @@ open class PassiveLocationManager: NSObject {
...
     
     /**
-     Starts the generation of location updates. 
+     Starts the generation of location updates and resumes trip session.
+
+     To learn more read `PassiveLocationManager.resumeTripSession()` method documentation.
      */
+    @available(*, deprecated, message: "Use resumeTripSession() instead.")
     public func startUpdatingLocation() {
-        systemLocationManager.startUpdatingLocation()
```
bamx23, line 103: What if I call this method directly?

S2Ler: You are on you own. You can still use `pauseTripSession`.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -97,18 +97,22 @@ open class PassiveLocationManager: NSObject {
     public weak var delegate: PassiveLocationManagerDelegate?
     
     /**
-     Starts the generation of location updates. 
+     Starts the generation of location updates and resumes trip session.
+
+     To learn more read `PassiveLocationManager.resumeTripSession()` method documentation.
      */
+    @available(*, deprecated, message: "Use resumeTripSession() instead.")
```
Udumft, line 104: Why not remove this method completely? There will be breaking changes anyway. Also, please update `PassiveLocationManager.init` code doc postcondition as it recommends this method.

Less related question: while we are cleaning up the API, why not leave this method name but replace it's contents with `resumeTripSession`? This might be less declarative that this action is related to billing conditions, but then we could avoid introducing pair of new methods and won't end up with `startUpdatingLocation/resumeTripSession/pauseTripSession` but will have just `start/pauseUpdatingLocation`

S2Ler: @Udumft Thanks for these questions:

> Why not remove this method completely

Because of no breaking change policy mostly. 

> There will be breaking changes anyway.

Yeah, it seems that RC still can contain breaking changes, although I've been told otherwise. 

> Also, please update PassiveLocationManager.init code doc postcondition as it recommends this method.

Thanks

> Less related question: While cleaning up the API, why not leave this method name but replace its contents with resumeTripSession? It might be less declarative that this action is related to billing conditions, but then we could avoid introducing pair of new methods and won't end up with startUpdatingLocation/resumeTripSession/pauseTripSession but will have start/pauseUpdatingLocation

From my point of view, it won't be evident that pausing location updates have more side effects. I expect us to have more sessions like API in the future, like pausing navigator, sessions for telemetry, billing sessions won't go anywhere, and we want to provide a means of controlling it. So my idea is that there will be a generic name that won't embed the implementation detail, which is a good thing for library evolution. 

@mapbox/navigation-ios We can debate the rename of "pause/resume trip session" to something else. For example: "start/stop".

S2Ler: I renamed pause/resume trip session to `startTripSession` and `stopTripSession`. Along with this, trip session doesn't start with `PassiveLocationManager` instantiation (this was weird behaviour in the first place). This is a safe change because:
- `startUpdatingLocation` (`startTripSession` now) was required to be called after creating `PassiveLocationManager` and Maps calls this method through `PassiveLocationProvider`. 
- We mention the requirement to call this method in `init` `postcondition`. 

I think this simplifies this class usage and adoption.

Also removed startUpdatingLocation instead of deprecating it.

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@
 * Fixed an issue where `RouteController` or `PassiveLocationManager` sometimes snapped the user’s location assuming a path that violated a turn restriction. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Improved performance and decreased memory usage when downloading routing tiles. ([#2808](https://github.com/mapbox/mapbox-navigation-ios/pull/2808))
 * Renamed `PassiveLocationManager.startUpdatingLocation(completionHandler:)` to `PassiveLocationProvider.startUpdatingLocation()`. This method now runs synchronously like `CLLocationManager.startUpdatingLocation()`. ([#2823](https://github.com/mapbox/mapbox-navigation-ios/pull/2823))
+* Remove `PassiveLocationManager.startUpdatingLocation` in favour of `PassiveLocationManager.startTripSession()` which works better with new pricing option. ([#3362](https://github.com/mapbox/mapbox-navigation-ios/pull/3362)).
```
chezzdev, line 67: We should remove the previous point as this one rewrites it from the perspective of 1.x.
Also please mention the addition of `stopTripSession`.

S2Ler: Can't remove it completely though I should merge two entries.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -96,19 +98,13 @@ open class PassiveLocationManager: NSObject {
...
-    }
-    
     var lastRawLocation: CLLocation?
     
     /**
      Manually sets the current location.
      
-     This method stops any automatic location updates.
+     This method stops any automatic location updates
+     without pausing the Free Drive session (`PassiveLocationManager.pauseTripSession()`).
```
S2Ler, line 107: Method changed.

Sources/MapboxNavigation/PassiveLocationProvider.swift
```diff
@@ -76,11 +76,11 @@ open class PassiveLocationProvider: NSObject, LocationProvider {
...
 
     public func startUpdatingLocation() {
-        locationManager.startUpdatingLocation()
+        locationManager.startTripSession()
     }
 
     public func stopUpdatingLocation() {
-        locationManager.systemLocationManager.stopUpdatingLocation()
+        locationManager.stopTripSession()
     }
```
1ec5, line 84: I appreciate that renaming these methods clarifies the lifespan of a billing session, but losing the consistency between PassiveLocationManager method names versus PassiveLocationProvider and CLLocationManager method names would be an unfortunate tradeoff. A developer is already used to thinking in terms of starting and stopping location updates in a location manager, since that’s the terminology that Core Location uses. A trip-based billing session beginning and ending would be only one of multiple consequences of starting and stopping location updates.

The map SDK automatically calls `stopUpdatingLocation()` and `startUpdatingLocation()` when the user sends the application to the background and foreground, respectively. Is it clear to the developer that using the application switcher or lock screen always resets the trip session, as these lines imply? It’s surprising that these methods seems like they do something entirely different than `stopUpdatingHeading()` and `startUpdatingHeading()`, which get called at the same time.

Sources/MapboxCoreNavigation/PassiveLocationManager.swift
```diff
@@ -130,23 +130,37 @@ open class PassiveLocationManager: NSObject {
...
+
+     Stopped trip sessions don't generate location and navigation status updates.
 
      Use this method to extend the existing Free Drive session if you temporarily don't need navigation updates. For
-     more info, read the [Pricing Guide](https://docs.mapbox.com/ios/beta/navigation/guides/pricing/).
+     For more info, read the [Pricing Guide](https://docs.mapbox.com/ios/beta/navigation/guides/pricing/).
      */
-    public func pauseTripSession() {
-        BillingHandler.shared.pauseBillingSession(with: sessionUUID)
+    public func stopTripSession() {
```
1ec5, line 140: `startUpdatingLocation()`, `startUpdatingHeading()`, `stopUpdatingLocation()`, and `stopUpdatingHeading()` use the verbs “start” and “stop” because they start and stop a series of updates and there’s no meaningful difference between starting and resuming that stream of updates. By contrast, a “session” is a meaningful mode of operation; a developer would hesitate to call `startTripSession()` multiple times because they would be billed for multiple sessions. However, these methods try to reuse an existing session, so “pause” and “resume” would be more accurate verbs to use in these method names.

Sources/MapboxNavigation/PassiveLocationProvider.swift
S2Ler: Thank you for feedback. After reading this multiple times including [this](https://github.com/mapbox/mapbox-navigation-ios/pull/3362#discussion_r709622432) comment I'm inclining to:
- I revert all the changes I've made to API surface level and just keep the code that doesn't send location updates to Navigator when trip session is paused. This will give total control to devs except that they can't control when trip session first starts.

S2Ler: Done. The PR simplified to the bare minimum.

1ec5 (approved):

PR #3366, opened 2021-09-14
Update `RoadObjectType` documentation.

MaximAlien:
Update `RoadObjectType` documentation.

### DIFF
--- a/Sources/MapboxCoreNavigation/EHorizon/RoadObjectType.swift
+++ b/Sources/MapboxCoreNavigation/EHorizon/RoadObjectType.swift
@@ -2,31 +2,60 @@ import Foundation
 import MapboxDirections
 import MapboxNavigationNative
 
-/** Type of the road object */
+/**
+ Type of the road object.
+ */
 public enum RoadObjectType {
 
-    /** Road object represents some road incident */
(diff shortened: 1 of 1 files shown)


Sources/MapboxCoreNavigation/EHorizon/RoadObjectType.swift
```diff
@@ -2,31 +2,60 @@ import Foundation
...
-/** Type of the road object */
+/**
+ Type of the road object.
+ */
 public enum RoadObjectType {
 
-    /** Road object represents some road incident */
+    /**
+     An alert providing information about incidents on a route. Incidents can include *congestion*,
+     *massTransit*, and more (see `Kind` for the full list of incident types).
```
1ec5, line 12: Technically, we could use backticks to link to the individual cases of `Incident.Kind`, but I think it’s sufficient to link to `Incident.Kind` and refer to these example kinds in plain English.

```suggestion
     An alert providing information about incidents on a route. Incidents can include congestion,
     mass Transit, and more (see `Incident.Kind` for the full list of incident types).
```

1ec5 (approved):

Issue #3363, opened 2021-09-13
Rename `RoadObjectType` to `RoadObject.Type`


1ec5:
For consistency with other type types, `RoadObjectType` should be renamed to `RoadObject.Type` – that is, `Type` in an extension on `RoadObject`. If we’re unable to make this change ahead of v2.0.0, then we should add a typealias from `RoadObjectType` to `RoadObject.Type` to preserve backwards compatibility.

/cc @mapbox/navigation-ios

PR #3045, opened 2021-06-01
Rename RoadObjectsStore to RoadObjectStore

ShanMa1991:
### Description
This pr is to fix #3041 to rename `RoadObjectsStore` and `RoadObjectsStoreDelegate` to `RoadObjectStore` and `RoadObjectStoreDelegate`.

PR #3367, opened 2021-09-14
Rename `RoadObjectType` to `RoadObject.Kind` (and other related symbols).

MaximAlien:
Closing #3363.


Sources/MapboxCoreNavigation/EHorizon/ObjectType.swift
```diff
@@ -0,0 +1,109 @@
+import Foundation
+import MapboxDirections
+import MapboxNavigationNative
+
+extension RoadObject {
+    
+    /**
+     Type of the road object.
+     */
+    public enum ObjectType {
```
MaximAlien, line 10: Even though it's possible to use enum with `Type` name, it'll be impossible to use initializers of that enum. That's the reason I've used `ObjectType` name here.

Sources/MapboxCoreNavigation/EHorizon/DistancedRoadObject.swift
```diff
@@ -13,15 +13,15 @@ public enum DistancedRoadObject {
      - parameter type: Road object type.
      - parameter distance: Distance to the point object, measured in meters.
      */
-    case point(identifier: RoadObjectIdentifier, type: RoadObjectType, distance: CLLocationDistance)
+    case point(identifier: RoadObjectIdentifier, type: RoadObject.ObjectType, distance: CLLocationDistance)
```
1ec5, line 16: I didn’t realize it earlier, but `RoadObjectIdentifier` could also be renamed to `RoadObject.Identifier`. `RoadObjectType` was the main type that would’ve been awkward, so it isn’t a huge deal. But if you’re up to it, it would be even nicer to go through all the `RoadObject*` types and nest them under `RoadObject`. If not, we can treat it as tail work: it isn’t so bad to keep a deprecated type alias around with the original, unnested name.

Sources/MapboxCoreNavigation/EHorizon/ObjectType.swift
1ec5: Oh, good point. That’s why we’ve always called it “Kind” instead. There are some related “Kind” types, so we should try to be consistent with `RoadObject.Kind`. Otherwise, `RoadObject.ObjectType` is more redundant than what we started out with.

Sources/MapboxCoreNavigation/EHorizon/RoadObject.swift
```diff
@@ -21,7 +21,7 @@ public struct RoadObject {
     public let location: RoadObjectLocation
 
     /** Type of the road object with metadata. */
-    public let type: RoadObjectType
+    public let type: RoadObject.ObjectType
```
1ec5, line 24: If we end up renaming the enumeration to `Kind`, then this property should also be renamed to `kind`.

docs/jazzy.yml
```diff
@@ -181,7 +181,7 @@ custom_categories:
       - RoadObjectMatcher
       - RoadObjectsStore
       - RoadObjectLocation
-      - RoadObjectType
+      - ObjectType
```
1ec5, line 184: This needs to be fully qualified in order for jazzy to know what it refers to.

```suggestion
      - RoadObject.Kind
```

Sources/MapboxCoreNavigation/EHorizon/ObjectType.swift
MaximAlien: Renamed `RoadObject.ObjectType` to `RoadObject.Kind`.

docs/jazzy.yml
```diff
@@ -182,10 +182,10 @@ custom_categories:
       - RoadObjectMatcher
       - RoadObjectsStore
```
1ec5, line 183: #3045 renamed `RoadObjectsStore` and `RoadObjectsStoreDelegate` to `RoadObjectStore` and `RoadObjectStoreDelegate`, respectively. I guess those names should stay as is, because `Store` and `StoreDelegate` would be misleading in any code that relies on type inference. (It would only be slightly less weird if we called it `Storage`.) In any case, these entries in the table of contents need to be updated to match the declared type names.

docs/jazzy.yml
```diff
@@ -181,11 +181,11 @@ custom_categories:
...
-      - RoadObjectLocation
-      - RoadObjectType
+      - RoadObject.Location
+      - RoadObject.Kind
       - RoadObjectMatcherDelegate
       - RoadObjectsStoreDelegate
-      - RoadObjectEdgeLocation
+      - RoadObject.EdgeLocation
       - RoadObjectMatcherError
       - RoadObjectPosition
```
1ec5, line 190: This is the only other type that would make sense to rename to a nested type, for consistency with `RoadObject.Location`.

1ec5 (approved):
In addition to the one remaining type below, the changelog has separate blurbs for `RoadObjectPosition`, `RoadObjectLocation`, `RoadObjectLocation.subgraph`, `RoadGraph.SubgraphEdge`, and `RoadObjectLocation.subgraph.shape`, all of which seem redundant since `RoadObject` itself is new. I think these blurbs can be removed in their entirety.

docs/jazzy.yml
MaximAlien: Yep, was updated right after your comment.

PR #3378, opened 2021-09-20
New billing trip starts on each waypoint

S2Ler:
This is tail work for new pricing.

### DIFF
--- a/Sources/MapboxCoreNavigation/BillingHandler.swift
+++ b/Sources/MapboxCoreNavigation/BillingHandler.swift
@@ -295,7 +295,9 @@ final class BillingHandler {
                 switch error {
                 case .invalidSkuId:
                     preconditionFailure("Invalid sku_id: \(error)")
-                case .tokenValidationFailed, .resumeFailed, .unknown:
+                case .tokenValidationFailed:
+                    assertionFailure("Token validation failed. Please check that you have the correct Mapboox Access Token.")
+                case .resumeFailed, .unknown:
                     break
                 }
                 self?.failedToBeginBillingSession(with: uuid, with: error)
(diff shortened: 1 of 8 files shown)


Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -240,6 +240,10 @@ open class RouteController: NSObject {
         let legIndex = UInt32(value)
         if navigator.changeRouteLeg(forRoute: 0, leg: legIndex), let status = navigator.getStatus() {
             updateIndexes(status: status, progress: routeProgress)
+            /** NOTE:
+             `navigator.changeRouteLeg(forRoute:leg:)` will return true if the leg actually changed.
+             */
+            BillingHandler.shared.beginNewBillingSessionIfRunning(with: sessionUUID)
```
S2Ler, line 246: This is the most reliable way I found to identify a new Waypoint start. @1ec5 This is the most critical section of the PR.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -307,6 +307,34 @@ final class BillingHandler {
...
+        lock.lock()
+
+        guard let session = _sessions[uuid], !session.isPaused else {
+            return
+        }
+
+        lock.unlock()
+
+        billingService.beginBillingSession(for: session.type) { [logger] error in
+            os_log("New trip session isn't started. Please check that you have the correct Mapboox Access Token",
```
1ec5, line 325: If we’re pretty sure this is misconfiguration on the part of the developer, rather than internal inconsistency out of the developer’s control, then this should be a precondition failure to get the developer’s attention, similar to what we do when no access token has been set:

https://github.com/mapbox/mapbox-navigation-ios/blob/6b827eeb5009f0fe6ddf6db7fc61494232180d82/Sources/MapboxCoreNavigation/BillingHandler.swift#L184

Sources/MapboxCoreNavigation/RouteController.swift
1ec5: Yes, I agree that this is the most natural bottleneck for the start of a new leg.

1ec5 (approved):

Sources/MapboxCoreNavigation/BillingHandler.swift
S2Ler: `tokenValidationFailed` can be called when server doesn't validate token or token is empty (we already have check for that). If we put precondition here we might crash developers' apps due to bug on server. It is far from desirable. I can try to put assert here though.

PR #3379, opened 2021-09-20
Add public getter for the feedback event contents

azarovalex:
This will be an undocumented API because we don't want customers to rely on the content. However, we want to share the content in case our customers what to audit it.


Sources/MapboxCoreNavigation/FeedbackEvent.swift
```diff
@@ -33,4 +33,9 @@ public class FeedbackEvent: Codable {
         coreEvent.eventDictionary["source"] = source.description
         coreEvent.eventDictionary["description"] = description
     }
+    
+    /// :nodoc:
+    public var contents: [String: Any] {
+        coreEvent.eventDictionary
```
bamx23, line 39: Should we filter anything internal from this?

azarovalex: No, I checked that there is nothing internal. All these things can be obtained from public methods, except for some session identifiers, but I don't think we should remove them because 
1. they are not billing connected and cannot alter the SDK in any way
2. our customers want to audit the full data if possible

bamx23 (approved):

PR #3382, opened 2021-09-21
Generate prerelease podspecs

### LINKED ISSUE #3357: Planning to release v2-beta or v2-rc MapboxNavigation in CocoaPods?
The latest version in [CocoaPods/Specs/MapboxNavigation](https://github.com/CocoaPods/Specs/tree/master/Specs/5/8/b/MapboxNavigation) is 1.4.2 which does not support arm64. MapboxNavigation does not have any v2 version in cocoapods. Currently, Podspec does not have a method to install dependency directly from git.

If there is no critical bug in the SDK, could you please release the v2 version as a beta or rc and make them available at least? If not, what is the estimated date you will release?

Thanks.

1ec5:
With this change, the release script now generates a pair of prerelease podspecs if a prerelease version is being published. These podspecs can then be pushed to CocoaPods trunk under the `MapboxCoreNavigation-pre` and `MapboxNavigation-pre` names.

Fixes #3357.

/cc @mapbox/navigation-ios

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ Carthage/Checkouts
 .swiftpm/
 
 Examples/Swift/Navigation_Example.mobileprovision
+
+*-pre.podspec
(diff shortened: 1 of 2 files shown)


MaximAlien (approved):

bamx23:
Should we add the pre podspec to the .gitignore file?

bamx23 (approved):

PR #3383, opened 2021-09-21
Allow guidance cards to receive the instruction update

### LINKED ISSUE #3243: Guidance card shouldn't default choose the last instruction of the step
In the guidance car view, we default choose to update the last instruction of the step. But if the step is pretty long in distance and has several instructions, and it would cause a problem, that the card is showing the final one while the driver is still far away from it.

https://github.com/mapbox/mapbox-navigation-ios/blob/4f5015aeb2f4f200148301488d253395cd1f128c/Sources/MapboxNavigation/InstructionsCardView.swift#L35-L37

### LINKED ISSUE #3381: [Bug]: Guidance Card fails to update the distance.

2.0.0

### Steps to reproduce

1. Choose `Example`, location as `Free Drive`
2. Choose a destination on the `I280` highway, to form a single waypoint route `without turn`.
3. Start navigation

The issue is caused by the following that the `InstructionsCardViewController.instructionContainerView(at:)` doesn't return the valid `InstructionsCardContainerView` to be updated with the distance.

https://github.com/mapbox/mapbox-navigation-ios/blob/8361af6769f3656d7d95adf59f412e6b40265ba7/Sources/MapboxNavigation/InstructionsCardViewController.swift#L207-L214

### Expected behavior

The distance on the guidance card will be updated with the `routeProgress`.

### Actual behavior

The distance in the guidance card fails to be updated as below, we could see the bottom banner has the right distance while the guidance card not.

guidanceCardDistance

repeatable

ShanMa1991:
### Description
This pr is to fix #3243 and #3381 to allow the guidance cards receiving the `VisualInstructionBanner` update and distance update from `navigationService` .

### Implementation
To fix #3381 , the `InstructionsCardViewController.updateCurrentVisibleInstructionCard(for:)` is added to pass the `VisualInstructionBanner` to the current active guidance card. In case the `VisualInstructionBanner` notification arrives before the guidance card configuration and set up, it's stored to prepare for the configuration of the first step.

The #3243 is caused by the sequence change of subviews of the `instructionCollectionView` during the active navigation. Thus we loop through its subviews for the casting instead of the fixed position.

### Screenshots or Gifs
guidanceCards

### DIFF
--- a/Sources/MapboxNavigation/InstructionsCardViewController.swift
+++ b/Sources/MapboxNavigation/InstructionsCardViewController.swift
@@ -36,6 +36,8 @@ open class InstructionsCardViewController: UIViewController {
     public private(set) var isInPreview = false
     public var currentStepIndex: Int?
     
+    var currentInstruction: VisualInstructionBanner?
+    
     public var steps: [RouteStep]? {
         guard let stepIndex = routeProgress?.currentLegProgress.stepIndex, let steps = routeProgress?.currentLeg.steps else { return nil }
         var mutatedSteps = steps
@@ -169,6 +171,14 @@ open class InstructionsCardViewController: UIViewController {
         }
     }
     
+    func updateCurrentVisibleInstructionCard(for instruction: VisualInstructionBanner) {
+        guard let remainingStepsCount = routeProgress?.currentLegProgress.remainingSteps.endIndex else { return }
+        let indexPath = IndexPath(row: 0, section: 0)
+        if let container = instructionContainerView(at: indexPath), indexPath.row < remainingStepsCount {
+            container.updateInstruction(instruction)
+        }
+    }
+    
     open func updateVisibleInstructionCards(at indexPaths: [IndexPath]) {
         guard let legProgress = routeProgress?.currentLegProgress else { return }
         let remainingSteps = legProgress.remainingSteps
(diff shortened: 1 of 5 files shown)


Sources/MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -289,9 +299,12 @@ extension InstructionsCardViewController: UICollectionViewDataSource {
...
         
         let step = steps[indexPath.row]
-        let firstStep = indexPath.row == 0
-        let distance = firstStep ? distanceRemaining : step.distance
-        cell.configure(for: step, distance: distance)
+        if indexPath.row == 0 {
+            cell.configure(for: step, distance: distanceRemaining, instruction: self.currentInstruction)
+        } else {
+            cell.configure(for: step, distance: step.distance)
+
```
bamx23, line 306: NIT
```suggestion
```

Sources/MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -169,6 +171,14 @@ open class InstructionsCardViewController: UIViewController {
         }
     }
     
+    open func updateCurrentVisibleInstructionCard(for instruction: VisualInstructionBanner) {
+        guard let remainingStepsCount = routeProgress?.currentLegProgress.remainingSteps.endIndex else { return }
+        let indexPath = IndexPath(row: 0, section: 0)
+        if let container = instructionContainerView(at: indexPath), indexPath.row < remainingStepsCount {
```
bamx23, line 177: Should we reorder conditions like this?
```suggestion
        if indexPath.row < remainingStepsCount, let container = instructionContainerView(at: indexPath) {
```

bamx23 (approved):

Sources/MapboxNavigation/InstructionsCardViewController.swift
ShanMa1991: Thanks for the suggestion, I'll improve this part.

Issue #3377, opened 2021-09-18
Unable to hide user location indicator after showing it


1ec5:
#2808 and #2968 removed the `NavigationMapView.showsUserLocation` property in favor of `userLocationStyle`, but  there’s no obvious way to hide the user location indicator after showing it. Either `UserLocationStyle` should have a `none` case or, better yet, `userLocationStyle` should be optional. As a workaround, you can set `userCourseView.isHidden` to `true`, but that feels imbalanced when you just set `userLocationStyle` to a value to show the user location indicator, and it doesn’t work if you set `userLocationStyle` to anything other than `courseView(_:)`.

/cc @mapbox/navigation-ios

ShanMa1991:
The issue is that during free drive or a single `navigationMapView` without the active guidance, if the user choose the default `userCourseView`,  the `mapView` will stop providing location update.  Or if the developer want to hide the user location layer, we could only to set `userCourseView`, because map puck couldn't be hidden right now, but it will lead to the  `mapView` stop providing location update either.

So we may want to allow the mapView provide a way to hide the mapPuck when we set it to `nil` but still sending location update in the background.

ShanMa1991:
There's a ticket opened in the Maps repo about providing the ability to hide the map puck but keeps the location update. https://github.com/mapbox/mapbox-maps-ios/issues/691

Issue #3369, opened 2021-09-15
[Bug]: CourseView userLocationStyle doesn't sync view with location


S2Ler:
### Mapbox Navigation SDK version

v2.0.0-rc.1

### Steps to reproduce

Remove `navigationMapView.userLocationStyle = .puck2D()` from Example app.

### Expected behavior

Course view syncs with location

### Actual behavior

Course view stays at top left corner. 

https://user-images.githubusercontent.com/413986/133429874-e3636a18-06fa-4623-b806-afb2810bb43a.mp4

### Is this a one-time issue or a repeatable issue?

repeatable

ShanMa1991:
Right now `userCourseView` only listens to the `routeController` during the active guidance navigation. Thus when there's a free driving or just a `navigationMapView`, the `navigationMapView` only receives location update information from the `mapView`, which also provide location puck layer to indicate the user location. The user should use the map puck as the location indicate layer.

MaximAlien:
This issue is reproducible with plain `NavigationMapView` as well. For example:

![Simulator Screen Shot - iPod touch (7th generation) - 2021-09-07 at 18 24 14](https://user-images.githubusercontent.com/1496498/133514887-49bcc603-1861-4c7e-9c26-3e295c138275.png)

This looks like a regression in 2.0, we already had similar ticket for 1.x https://github.com/mapbox/mapbox-navigation-ios/issues/2513, which was resolved.

ShanMa1991:
Right now if the `puckType` is `nil`, the `mapView` would stop the location update in the background, not only displaying the `puck`.

https://github.com/mapbox/mapbox-maps-ios/blob/bedc49744970602ebc10c6de3f8a2b4cc6f084a6/Sources/MapboxMaps/Location/LocationManager.swift#L203-L205

PR #3384, opened 2021-09-21
Stop publicly using `NavigationMapView.userCourseView` in favor of `NavigationMapView.userLocationStyle`.

MaximAlien:
Closing #3377 and #3369.


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -125,19 +135,23 @@ open class NavigationMapView: UIView {
      
      The `UserCourseView`'s `UserCourseView.update(location:pitch:direction:animated:)` method is frequently called to ensure that its visual appearance matches the map’s camera.
      */
-    public var userCourseView: UserCourseView = UserPuckCourseView(frame: CGRect(origin: .zero, size: 75.0)) {
```
MaximAlien, line 128: `NavigationMapView.userCourseView` is now used only internally to prevent issues when user can modify both `NavigationMapView.userCourseView` and `NavigationMapView.userLocationStyle`.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -125,19 +135,23 @@ open class NavigationMapView: UIView {
...
     }
     
     /**
      Specifies how the map displays the user’s current location, including the appearance and underlying implementation.
      
-     By default, this property is set to `UserLocationStyle.courseView`, the bearing source is location course.
+     By default, this property is set to `UserLocationStyle.puck2D(configuration:)`, the bearing source is location course.
      */
-    public var userLocationStyle: UserLocationStyle = .courseView(UserPuckCourseView(frame: CGRect(origin: .zero, size: 75.0))) {
+    public var userLocationStyle: UserLocationStyle? = .puck2D(configuration: nil) {
```
MaximAlien, line 154: `NavigationMapView.userLocationStyle` is now optional and in case of `nil` will not show current location indicator.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -382,20 +396,6 @@ open class NavigationMapView: UIView {
...
-        let frame = CGRect(origin: .zero, size: 75.0)
-        let isHidden = userCourseView.isHidden
-        switch userLocationStyle {
-        case .courseView(let courseView):
-            userCourseView = reducedAccuracyActivatedMode ? UserHaloCourseView(frame: frame) : courseView
-        case .puck2D(_):
-            userCourseView = reducedAccuracyActivatedMode ? UserHaloCourseView(frame: frame) : UserPuckCourseView(frame: frame)
-        case .puck3D(_):
-            userCourseView = reducedAccuracyActivatedMode ? UserHaloCourseView(frame: frame) : UserPuckCourseView(frame: frame)
-        }
```
MaximAlien, line 395: I think in current implementation we need to tackle case when user switches `reducedAccuracyActivatedMode` to `true` and then back to `false` while using either 2D or 3D puck. In current implementation we'll fallback to `UserPuckCourseView`, which is not correct.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -258,23 +272,23 @@ open class NavigationMapView: UIView {
...
     }
     
     func setupUserLocation() {
         switch userLocationStyle {
-        case .courseView((let courseView)):
+        case .courseView(let courseView):
             mapView.location.options.puckType = nil
-            userCourseView = courseView
-            userCourseView.isHidden = false
+            userCourseView = courseView ?? UserPuckCourseView(frame: CGRect(origin: .zero, size: 75.0))
```
bamx23, line 283: Should we handle `reducedAccuracyActivatedMode` here?

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -258,23 +272,23 @@ open class NavigationMapView: UIView {
...
         case .puck2D(configuration: let configuration):
-            userCourseView.isHidden = true
+            userCourseView?.removeFromSuperview()
```
bamx23, line 285: Can we just call `userCourseView = nil`? (here and in the next cases)

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -258,23 +272,23 @@ open class NavigationMapView: UIView {
...
             mapView.location.options.puckType = .puck2D(configuration ?? Puck2DConfiguration())
         case .puck3D(configuration: let configuration):
-            userCourseView.isHidden = true
+            userCourseView?.removeFromSuperview()
             mapView.location.options.puckType = .puck3D(configuration)
+        case .none:
+            userCourseView?.removeFromSuperview()
```
bamx23, line 291: Should we change `mapView.location.options.puckType` here?

Sources/MapboxNavigation/NavigationMapView.swift
1ec5: As long as this is a stored property, it has the potential to get out of sync with the associated value of `UserLocationStyle.courseView(_:)` when that’s the value of the `userLocationStyle` property. Can we make this a computed property or eliminate it somehow?

Whether we remove `userCourseView` or make it private, it’s effectively gone to the developer, so we should note the removal in the changelog.

PR #3385, opened 2021-09-21
Add distance check to prevent vanishing route line update when users go off the route.

### LINKED ISSUE #3319: [Bug]: When user passed the previous route in a new direction before rerouting, the vanishing route line grows back

2.0.0

### Steps to reproduce

Trigger a rerouting event. Before generating a new route line, there're some new routeProgress updates come in.

### Expected behavior

Before the new route line generates, we should see the vanishing route line growing shorter or being stale.

### Actual behavior

The vanishing route line grows back due to the distance from the user is getting further away from the original waypoint and route.

repeatable

ShanMa1991:
### Description

This pr is to fix #3319 by adding the distance check of user location to pre-defined route. When the distance from user location to pre-defined route larger than `OffRouteDistanceUpdateThreshold`, the vanishing route line would stop updating until the rerouting event triggered to generate a new route line.

`OffRouteDistanceUpdateThreshold` is public and defaults to 15m. The `OffRouteDistanceUpdateThreshold` should be larger than the driving distance per second according to the speed. Otherwise, in some edge cases, such as in big turn or driving too fast, it would see the normal driving as off line behavior and stop updating the vanishing effect. Because the normal driving speed is 60km/h, it's safe to set the `OffRouteDistanceUpdateThreshold` between `15m ~ 20m`.

### Implementation
Add the `navigationMapView.findDistanceToNearestPointOnCurrentLine(coordinate:granularDistances:upcomingIndex)` to calculate the distance from the coordinate to the stored route line coordinates with the upcoming index. 

Because during the route refresh and rerouting event, the route line coordinates and upcoming index may be changed and result in a large distance calculated. The `navigationMapView.offRouteDistanceCheckDisabled` is introduced to control whether to disable the distance check. So during the refresh and rerouting event the distance check is disabled. After the new route line is displayed and coordinates stored, before the next location update comes in, the distance check would be enabled again.

### Screenshots or Gifs
As following, when `OffRouteDistanceUpdateThreshold` set to `15m`, when the distance from user location is larger than the pre-defined route, the vanishing effect would stop, which fix #3319 . But when the rerouting event triggered and a new route line generated, the vanishing effect would start again.
reroute_15_1

When `OffRouteDistanceUpdateThreshold` set to `15m`, and user is away from the pre-defined route, but distance is less than 15m, the vanishing effect would continue, which shows as the following that the route line grows back for `1 second`. Because the threshold is 15m, while the distance calculated is `12m`.
reroute_15_2

Then we decrease the threshold of `OffRouteDistanceUpdateThreshold` to `10m`, we could see that under the same situation, the off route driving is detected. So there's no more `1 second` growing back route line behavior.
reroute_10_2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 * Fixed an issue where `UserPuckCourseView` is trimmed when using custom frame for `UserLocationStyle.courseView(_:)`. ([#3601](https://github.com/mapbox/mapbox-navigation-ios/pull/3601))
 * Fixed an issue where route line blinks when style is changed during active navigation. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
 * Fixed an issue where route line missing traffic colors after refresh or rerouting. ([#3622](https://github.com/mapbox/mapbox-navigation-ios/pull/3622))
-* Fixed an issue when user goes offline and the route line grows back when `NavigationViewController.routeLineTracksTraversal` was enabled. Added global variable `OffRouteDistanceUpdateThreshold` as the distance threshold. When the distance of user location to the route is larger than the `OffRouteDistanceUpdateThreshold`, the vanishing effect of route line would stop until the new route line gets generated. ([#3385](https://github.com/mapbox/mapbox-navigation-ios/pull/3385))
+* Fixed an issue when user goes offline and the route line grows back when `NavigationViewController.routeLineTracksTraversal` enabled. When the distance of user location to the route is larger than certain distance threshold, the vanishing effect of route line would stop until the new route line gets generated. ([#3385](https://github.com/mapbox/mapbox-navigation-ios/pull/3385))
 
 ### Banners and guidance instructions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Fixed an issue where `UserPuckCourseView` is trimmed when using custom frame for `UserLocationStyle.courseView(_:)`. ([#3601](https://github.com/mapbox/mapbox-navigation-ios/pull/3601))
 * Fixed an issue where route line blinks when style is changed during active navigation. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
 * Fixed an issue where route line missing traffic colors after refresh or rerouting. ([#3622](https://github.com/mapbox/mapbox-navigation-ios/pull/3622))
+* Fixed an issue when user goes offline and the route line grows back when `NavigationViewController.routeLineTracksTraversal` was enabled. Added global variable `OffRouteDistanceUpdateThreshold` as the distance threshold. When the distance of user location to the route is larger than the `OffRouteDistanceUpdateThreshold`, the vanishing effect of route line would stop until the new route line gets generated. ([#3385](https://github.com/mapbox/mapbox-navigation-ios/pull/3385))
 
 ### Banners and guidance instructions
 
--- a/Sources/MapboxNavigation/Constants.swift
+++ b/Sources/MapboxNavigation/Constants.swift
@@ -46,6 +46,11 @@ public let NavigationViewMinimumVolumeForWarning: Float = 0.3
  */
 public var GradientCongestionFadingDistance: CLLocationDistance = 30.0
 
+/**
+ The maximum distance threshold of vanishing route line update. When the user's location to the route line is larger than the threshold, the user is off the route and the route line won't be updated.
+ */
+public var OffRouteDistanceUpdateThreshold: CLLocationDistance = 15.0
+
 extension Notification.Name {
     /**
      Posted when `StyleManager` applies a style that was triggered by change of time of day, or when entering or exiting a tunnel.
--- a/Sources/MapboxNavigation/Constants.swift
+++ b/Sources/MapboxNavigation/Constants.swift
@@ -46,11 +46,6 @@ public let NavigationViewMinimumVolumeForWarning: Float = 0.3
  */
 public var GradientCongestionFadingDistance: CLLocationDistance = 30.0
 
-/**
- The maximum distance threshold of vanishing route line update. When the user's location to the route line is larger than the threshold, the user is off the route and the route line won't be updated.
- */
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 * Fixed an issue where `ReplayLocationManager` didn't update location timestamps when a new loop started. ([#3550](https://github.com/mapbox/mapbox-navigation-ios/pull/3550))
 * Fixed the background location update issue during active navigation when using default `.courseView` for `NavigationMapView.userLocationStyle`. ([#3533](https://github.com/mapbox/mapbox-navigation-ios/pull/3533))
 * Fixed an issue where `UserPuckCourseView` is trimmed when using custom frame for `UserLocationStyle.courseView(_:)`. ([#3601](https://github.com/mapbox/mapbox-navigation-ios/pull/3601))
+* Fixed an issue when user goes offline and the route line grows back when `NavigationViewController.routeLineTracksTraversal` was enabled. Added global variable `OffRouteDistanceUpdateThreshold` as the distance threshold. When the distance of user location to the route is larger than the `OffRouteDistanceUpdateThreshold`, the vanishing effect of route line would stop untile the new route line generated. ([#3385](https://github.com/mapbox/mapbox-navigation-ios/pull/3385))
```
MaximAlien, line 25:
```suggestion
* Fixed an issue when user goes offline and the route line grows back when `NavigationViewController.routeLineTracksTraversal` was enabled. Added global variable `OffRouteDistanceUpdateThreshold` as the distance threshold. When the distance of user location to the route is larger than the `OffRouteDistanceUpdateThreshold`, the vanishing effect of route line would stop until the new route line gets generated. ([#3385](https://github.com/mapbox/mapbox-navigation-ios/pull/3385))
```

MaximAlien: But I'd leave it to @1ec5 to propose better wording :)

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -1199,6 +1199,7 @@ open class NavigationMapView: UIView {
     var routeRemainingDistancesIndex: Int?
     var fractionTraveled: Double = 0.0
     var currentLegIndex: Int?
+    var offRouteDistanceCheckDisabled: Bool = false
```
MaximAlien, line 1202: What about using `offRouteDistanceCheckEnabled`? I think it'll be easier to understand, as you're checking it with unary operator anyway.

ShanMa1991: Thanks for the suggestion, I'll change it to `offRouteDistanceCheckEnabled`.

Sources/MapboxNavigation/Constants.swift
```diff
@@ -46,6 +46,11 @@ public let NavigationViewMinimumVolumeForWarning: Float = 0.3
  */
 public var GradientCongestionFadingDistance: CLLocationDistance = 30.0
 
+/**
+ The maximum distance threshold of vanishing route line update. When the user's location to the route line is larger than the threshold, the user is off the route and the route line won't be updated.
+ */
+public var OffRouteDistanceUpdateThreshold: CLLocationDistance = 15.0
```
MaximAlien, line 52: To play it safe I'd rather keep this value internal for now. We'll have a chance to do a couple rounds of testing first and then see whether this is really something we should expose as public variable.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -122,6 +124,24 @@ extension NavigationMapView {
         return RouteLineGranularDistances(distance: distance, distanceArray: indexArray.compactMap{ $0 })
     }
     
+    func findDistanceToNearestPointOnCurrentLine(coordinate: CLLocationCoordinate2D, granularDistances: RouteLineGranularDistances, upcomingIndex: Int) -> CLLocationDistance {
+        guard granularDistances.distanceArray.indices.contains(upcomingIndex) else { return 0.0 }
+
+        var coordinates = [CLLocationCoordinate2D]()
+        for index in max(0, upcomingIndex - 10)...upcomingIndex  {
```
MaximAlien, line 131: Why `10` specifically in this case? Maybe also makes sense to provide a comment, which describes your motivation.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
```diff
@@ -292,4 +292,35 @@ class VanishingRouteLineTests: TestCase {
         let currentLineGradientStops = navigationMapView.routeLineGradient(congestionFeatures, fractionTraveled: 0.0)
         XCTAssertEqual(currentLineGradientStops[0.0], navigationMapView.trafficUnknownColor, "Failed to use trafficUnknownColor for route line when no congestion level found.")
     }
+    func testFindDistanceToNearestPointOnCurrentLine() {
+        // https://github.com/mapbox/mapbox-navigation-android/blob/0478c43781fdf7489f10f22c0055fadf181970f6/libnavui-maps/src/test/java/com/mapbox/navigation/ui/maps/internal/route/line/MapboxRouteLineUtilsTest.kt#L814
+        
+        let route = getMultilegRoute()
+        let coordinate = route.shape!.coordinates[15]
```
MaximAlien, line 299: I think that in such cases it's better to add assert check whether coordinate is really available.

MaximAlien (approved):

Sources/MapboxNavigation/Constants.swift
ShanMa1991: Thanks for the suggestion, I'll make this value internal now.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: We could also take other numbers for the passed points, but to keep the parity with Android. It's to add extend to sliced route line a little longer incase the curved shape of the route.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
ShanMa1991: Thanks for the suggestion, I'll add check for the coordinate.

Issue #3386, opened 2021-09-22
[Bug]: Prevent crashing when `RouteProgress.distance=0`


tsuz:
### Mapbox Navigation SDK version

v2.0.0-rc.2

### Steps to reproduce

1. Choose two locations that snap to the same location (for example, `139.70504646232456,35.675949089898836;139.7050228620614,35.67594510574615`)
2. Notice the response from the API has `routes[].legs[].steps[].distance=0`
3. The SDK crashes because a number is not divisible by zero [source](https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-rc.2/Sources/MapboxCoreNavigation/RouteProgress.swift#L79-L81)

### Expected behavior

The SDK should not crash if `RouteProgress.distance=0`

### Actual behavior

The SDK crashes if `RouteProgress.distance=0`

### Is this a one-time issue or a repeatable issue?

repeatable

PR #3387, opened 2021-09-22
handle route with no distance

tsuz:
Fixes https://github.com/mapbox/mapbox-navigation-ios/issues/3386

### Rationale for the change

If `distance=0`, one can make a case that `fractionTraveled` should be either `0`, `1`, or `nil` value. (probably `nil` is the most correct way). For now, to be consistent with [`RouteStepProgress's implementation`](https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-rc.2/Sources/MapboxCoreNavigation/RouteStepProgress.swift#L35), returning `1` should suffice.

In the long term, `fractionTraveled` could become an optional parameter (`Double` or `nil`) but this would require handling this value used in other parts of the code ([example1](https://github.com/mapbox/mapbox-navigation-ios/blob/85f1e2217ed5d4ebb4bd8957cf6effda975c5bc5/Sources/MapboxNavigation/NavigationMapView.swift#L620-L645), [example2](https://github.com/mapbox/mapbox-navigation-ios/blob/007d5629abd2cc65075eeb4aa7e5224a11f04f9b/Sources/MapboxCoreNavigation/RouteProgress.swift#L202))


azarovalex (approved):
lgtm, special thanks for the test

1ec5 (approved):
Thanks, it makes sense that we’d need to account for division by zero due to the two waypoints being coincidental. Are there any other places that need to avoid division by zero?

Suggested changelog blurb:

* Fixed a crash when navigating along a route 0&nbsp;meters long (for example, because two waypoints snap to the same location). ([#3387](https://github.com/mapbox/mapbox-navigation-ios/pull/3387))

tsuz:
@1ec5 

Thanks for the review!

>  Are there any other places that need to avoid division by zero?

From just searching the code, these are a few that could have a division by zero. (there are many more tho)

https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-rc.2/Sources/MapboxCoreNavigation/SimulatedLocationManager.swift


https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-rc.2/Sources/MapboxCoreNavigation/SimulatedLocationManager.swift#L213

https://github.com/mapbox/mapbox-navigation-ios/blob/267a18380d7d78f8e136d92ef2359e036915d2a5/Sources/MapboxNavigation/Feedback/FeedbackViewController.swift#L312

CHANGELOG.md
```diff
@@ -73,6 +73,7 @@
 * By default, there is no longer a subtle crossfade between traffic congestion segments along a route line. To reenable this crossfade, set the `NavigationMapView.crossfadesCongestionSegments` property to `true`. You can also adjust the length of this crossfade using the global variable `GradientCongestionFadingDistance`. ([#3153](https://github.com/mapbox/mapbox-navigation-ios/pull/3153), [#3307](https://github.com/mapbox/mapbox-navigation-ios/pull/3307))
 * The duration annotations added by the `NavigationMapView.showRouteDurations(along:)` method are now set in the fonts you specify using the `NavigationMapView.routeDurationAnnotationFontNames` property. Use this property to specify a list of fallback fonts for better language support. ([#2873](https://github.com/mapbox/mapbox-navigation-ios/pull/2873))
 * Fixed an issue when route line was sometimes invisible after starting turn-by-turn navigation. ([#3205](https://github.com/mapbox/mapbox-navigation-ios/pull/3205))
+* Fixed a crash when navigating along a route 0 meters long (for example, because two waypoints snap to the same location). ([#3387](https://github.com/mapbox/mapbox-navigation-ios/pull/3387))
```
1ec5, line 76: This should go in the “Location tracking” section instead, since it isn’t necessarily related to the UI.

tsuz: Updated!

tsuz:
@1ec5 

I checked again if there are any `distance` or `duration` parameters that are used as division denominators; fortunately, I could not find any. However, there are [other parts of the code](https://github.com/mapbox/mapbox-navigation-ios/pull/3387#issuecomment-924734519) not related to the API response that could pose a risk.

1ec5:
6b7b142c9fe142eb6889db7c4d6f46a7abd3b1b9 added the test fixture to the other Xcode project that’s used by Carthage. Builds started failing due to the missing file reference, but only sometimes, for some reason.

PR #3388, opened 2021-09-22
Adds a script to use debug version of MapboxNavigationNative

S2Ler:
This is MVP and by no means perfect solution. Possible improvements:
- [x] Automatic destination selection
- [ ] Support device builds
- [ ] Make these builds on NavNative CI instead of building locally

### DIFF
--- a/scripts/debug_navigation_native.sh
+++ b/scripts/debug_navigation_native.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+# This script replaces SPM checked out MapboxNavigationNative framework for simulator with a debug version.
+
+if [[ -z $1 ]]; then 
+    echo "Usage $0 <path to MapboxNavigationNative git repository>"
+    exit 1
+fi
+
+set -x
+set -e
+
+NATIVE_PATH=$1
+DESTINATION="generic/platform=iOS Simulator"
+NATIVE_VERSION=$(python3 << END
+import json
+import sys
+
+packageResolved = json.loads(open("Package.resolved", "r").read())
+for pin in packageResolved["object"]["pins"]:
+    if pin["package"] == "MapboxNavigationNative":
+        sys.stdout.write(pin["state"]["version"])
+        exit(0)
+
+print("Haven't found Package", file=sys.stderr)
+exit(1)
+END 
+)
+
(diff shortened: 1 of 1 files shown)


scripts/debug_navigation_native.sh
```diff
@@ -0,0 +1,64 @@
...
+
+set -x
+set -e
+
+# Install a tool which parses JSONs
+brew list jq || brew install jq
+
+NATIVE_PATH=$1
+DESTINATION=$2
+NATIVE_VERSION=$(jq '.object.pins[] | select(.package=="MapboxNavigationNative") | .state.version' ./Package.resolved | sed 's/^"\(.*\)"$/\1/')
```
bamx23, line 18: I guess you can inline a python code here and not install `jq`

scripts/debug_navigation_native.sh
```diff
@@ -0,0 +1,64 @@
...
+fi
+
+set -x
+set -e
+
+# Install a tool which parses JSONs
+brew list jq || brew install jq
+
+NATIVE_PATH=$1
+DESTINATION=$2
```
bamx23, line 17: Can we use "any simulator" here?

scripts/debug_navigation_native.sh
S2Ler: thank you for suggestion, implemented

scripts/debug_navigation_native.sh
S2Ler: "generic/platform=iOS Simulator" works. Thank you

scripts/debug_navigation_native.sh
```diff
@@ -0,0 +1,74 @@
...
+NATIVE_PATH=$1
+DESTINATION="generic/platform=iOS Simulator"
+NATIVE_VERSION=$(python3 << END
+import json
+import sys
+
+packageResolved = json.loads(open("Package.resolved", "r").read())
+for pin in packageResolved["object"]["pins"]:
+    if pin["package"] == "MapboxNavigationNative":
+        sys.stdout.write(pin["state"]["version"])
```
bamx23, line 22: NIT: You can use `revision` here and get the commit hash

bamx23 (approved):
Great work, thanks!

PR #3393, opened 2021-09-22
Use the pre-existing `mapView` location provider when simulation mode turned off.

### LINKED ISSUE #3392: Avoid using the `AppleLocationProvider()` as the default when switch off the simulation mode during active navigation.
When we turn off the simulation mode and the simulated location provider during active navigation, we right now choose the default  `AppleLocationProvider()` as following:

https://github.com/mapbox/mapbox-navigation-ios/blob/8361af6769f3656d7d95adf59f412e6b40265ba7/Sources/MapboxNavigation/NavigationViewController.swift#L888-L892

But we should switch to the pre-existing location provider of the `mapView`, incase the user using their own location provider, or the injected `mapView`.

ShanMa1991:
This pr is to fix #3392 by storing the pre-existing `locationProvider` of the `mapView` during active navigation. When the simulation mode turned on or the `mapView` is set up, we store the `locationProvider` of the `mapView` in `navigationMapView`. When the simulation mode is off, we use it for location update.


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -341,6 +351,7 @@ open class NavigationMapView: UIView {
         mapView = MapView(frame: frame, mapInitOptions: mapInitOptions)
         mapView.translatesAutoresizingMaskIntoConstraints = false
         mapView.ornaments.options.scaleBar.visibility = .hidden
+        storeLocationProvider()
```
bamx23, line 354: Does it affect the free-driving? It's when you override the location provider with `PassiveLocationProvider` after a `NavigationMapView` is created.

ShanMa1991: Right now the `locationProvider` is only responsible to store the `locationProvider`. The only case that it would be used, is that when the `simulatesLocation` sets to `false` and it happens after the active navigation starts or the simulation mode of navigation service turned off. Because it's related with the `NavigationLocationManager` during active driving. 

During free drive, there's no change of the `simulatesLocation` of the `navigationMapView`. And if the user decides to use their own location provider for the `mapView` during active navigation, it works after tests.

bamx23 (approved):

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -857,6 +856,8 @@ extension CarPlayManager: CarPlayNavigationViewControllerDelegate {
         if let passiveLocationProvider = navigationMapView?.mapView.location.locationProvider as? PassiveLocationProvider {
             passiveLocationProvider.locationManager.resumeTripSession()
         }
+        
+        self.carPlayNavigationViewController = nil
```
ShanMa1991, line 860: It's to nil out the `CarPlayNavigationViewController` in `CarPlayManager` when the `CarPlayNavigationViewController` dismissed and exit navigation

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -182,26 +208,45 @@ public class MapboxNavigationService: NSObject, NavigationService {
     private var _simulationSpeedMultiplier: Double = 1.0
     
     private func simulate(intent: SimulationIntent = .manual) {
-        guard !isSimulating else { return }
         let progress = router.routeProgress
         delegate?.navigationService(self, willBeginSimulating: progress, becauseOf: intent)
+        announceSimulatingDidChange(simulatingUpdate: .willBeginSimulating)
+        guard !isSimulating else {
```
ShanMa1991, line 214: Because when CarPlay is used and under simulated mode, the `navigationService.simulate(intent:)` would be called twice. When we choose to start navigation from mobile, then the previous `guard !isSimulating else { return }` before the notification would block sending notification to the CarPlay.

Issue #3396, opened 2021-09-23
[Bug]: `CPMapTemplate.tripEstimateStyle` has incorrect style at the beginning of active guidance navigation on CarPlay.


MaximAlien:
### Mapbox Navigation SDK version

v2.0.0-rc.2

### Steps to reproduce

1. Enable Automatic appearance on CarPlay (it should enable light appearance).
1. Open Example application on CarPlay.
1. Start active guidance navigation and observe behavior.

### Expected behavior

It is expected that `CPMapTemplate.tripEstimateStyle` is white at the beginning of active guidance navigation.

### Actual behavior

`CPMapTemplate.tripEstimateStyle` is always black, even if light appearance is used.

![IMG_8068](https://user-images.githubusercontent.com/1496498/134597678-7cfdf0ed-7911-4e60-bf17-19b0b7bf44fa.jpg)

### Is this a one-time issue or a repeatable issue?

repeatable

PR #3397, opened 2021-09-23
Fix `CPMapTemplate.tripEstimateStyle` update issues.

MaximAlien:
Closing #3396.

### DIFF
--- a/Sources/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/Sources/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -168,10 +168,6 @@ public class CarPlayNavigationViewController: UIViewController {
         switch traitCollection.userInterfaceStyle {
         case .dark:
             mapTemplate.tripEstimateStyle = .dark
-        case .unspecified:
-            fallthrough
-        case .light:
-            fallthrough
         default:
             mapTemplate.tripEstimateStyle = .light
         }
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/Feedback/FeedbackSubtypeCollectionViewCell.swift
```diff
@@ -107,10 +107,10 @@ class FeedbackSubtypeCollectionViewCell: UICollectionViewCell {
 
     override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
-        if #available(iOS 12.0, *) {
-            if previousTraitCollection?.userInterfaceStyle != self.traitCollection.userInterfaceStyle {
-                circleView.layer.borderColor = circleOutlineColor.cgColor
-            }
+
+        if #available(iOS 12.0, *),
```
MaximAlien, line 111: Just minor styling improvement in unrelated piece of our codebase.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -142,13 +142,41 @@ public class CarPlayNavigationViewController: UIViewController {
...
+     */
+    func updateTripEstimateStyle(_ userInterfaceStyle: UIUserInterfaceStyle) {
+        switch traitCollection.userInterfaceStyle {
+        case .dark:
+            mapTemplate.tripEstimateStyle = .dark
+        case .unspecified:
+            fallthrough
+        case .light:
+            fallthrough
+        default:
```
bamx23, line 175: Do we need this `default` case here? Or do we need the previous two?

bamx23 (approved):

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
MaximAlien: Using all cases makes it more explicit, but I guess we can simplify it to use only `default` statement.

Issue #3390, opened 2021-09-22
Improve viewport parameters on CarPlay.


MaximAlien:
At this point Navigation SDK doesn't provide an optimal viewport options on CarPlay during active guidance, this results in a limited observability of the road. For example:

<img width="400" alt="Screen Shot 2021-09-22 at 9 40 37 AM" src="https://user-images.githubusercontent.com/1496498/134391870-11aee4ec-002b-48f0-a082-afa2b5ca482a.png">

On screenshot posted above user puck is located in the center of the viewport, but it should be shifted down. Pitch should be also increased.

PR #3405, opened 2021-09-24
Improve viewport parameters on CarPlay.

MaximAlien:
Closing #3390.

Before:
Screen Shot 2021-09-24 at 3 50 36 PM

Screen Shot 2021-09-24 at 3 51 09 PM

After:
Screen Shot 2021-09-24 at 3 49 03 PM

Screen Shot 2021-09-24 at 3 49 33 PM

/cc @johnk1 @danpat

### DIFF
--- a/Sources/MapboxNavigation/NavigationViewportDataSource.swift
+++ b/Sources/MapboxNavigation/NavigationViewportDataSource.swift
@@ -306,8 +306,6 @@ public class NavigationViewportDataSource: ViewportDataSource {
                 } else {
                     followingCarPlayCamera.padding = carPlayCameraPadding
                 }
-                
-                followingCarPlayCamera.padding = carPlayCameraPadding
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -192,7 +192,8 @@ public class NavigationViewportDataSource: ViewportDataSource {
             let geometryFramingAfterManeuver = followingCameraOptions.geometryFramingAfterManeuver
             let pitchСoefficient = self.pitchСoefficient(routeProgress, currentCoordinate: location.coordinate)
             let pitch = followingCameraOptions.defaultPitch * pitchСoefficient
-            let carPlayCameraPadding = mapView.safeArea + UIEdgeInsets.centerEdgeInsets
+            var carPlayCameraPadding = mapView.safeArea + UIEdgeInsets.centerEdgeInsets
+            carPlayCameraPadding.bottom += 25.0
```
bamx23, line 196: Should we move this hardcoded number to a constant? (or maybe even to a config option)

bamx23 (approved):

Sources/MapboxNavigation/NavigationViewportDataSource.swift
MaximAlien: `NavigationViewportDataSource` doesn't provide the ability to change specific values of viewport padding for iOS and CarPlay, but completely different padding value can be provided after setting `FollowingCameraOptions.paddingUpdatesAllowed` to false. So I think that it'd be better to leave this part unmodified. I've provided comment, which give more info about the reason why we need to increase bottom padding though.

PR #3407, opened 2021-09-27
Fix the tilt of the `UserHaloCourseView`.

### LINKED ISSUE #3406: [Bug]: Approximate location circle pitches wrong

2.0.0

### Steps to reproduce

1. Choose `Example` and start navigation
2. In `Privacy-Location Service-Example`, switch the `Precise Location` to `off`
3. tilt the map of the `Example` app

### Expected behavior

The circle representing the `UserHaloCourseView` should be tilted at the same angle of the `mapView` camera.

### Actual behavior

The circle is tilted to 45 degree or a strange degree like the following (contributed by @bamx23 ):

https://github.com/mapbox/mapbox-navigation-ios/blob/8257c395745e18255e391978e387034f00086290/Sources/MapboxNavigation/UserCourseView.swift#L16-L33

userHaloCourseView

repeatable

ShanMa1991:
### Description
This pr is to fix #3406 to allow the `UserHaloCourseView` of the approximate location circle could tilt at the same angle of the camera and `mapView` during the active guidance navigation.

### Implementation
This issue is fixed by allowing the `haloView` of the `UserHaloCourseView` to receive the pitch transform.

### Screenshots or Gifs
fixedHalo

### DIFF
--- a/Sources/MapboxNavigation/UserHaloCourseView.swift
+++ b/Sources/MapboxNavigation/UserHaloCourseView.swift
@@ -61,24 +61,6 @@ open class UserHaloCourseView: UIView, CourseUpdatable {
         haloView.backgroundColor = .clear
         addSubview(haloView)
     }
-    
-    /**
-     Transforms the location of the user halo course view.
-     */
-    public func update(location: CLLocation, pitch: CGFloat, direction: CLLocationDegrees, animated: Bool, navigationCameraState: NavigationCameraState) {
-        let duration: TimeInterval = animated ? 1 : 0
-        UIView.animate(withDuration: duration, delay: 0, options: [.beginFromCurrentState, .curveLinear], animations: {
-            // `UserHaloCourseView` pitch is changed only during transition to the overview mode.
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/UserHaloCourseView.swift
```diff
@@ -61,6 +61,24 @@ open class UserHaloCourseView: UIView, CourseUpdatable {
         haloView.backgroundColor = .clear
         addSubview(haloView)
     }
+    
+    /**
+     Transforms the location of the user halo course view.
+     */
+    public func update(location: CLLocation, pitch: CGFloat, direction: CLLocationDegrees, animated: Bool, navigationCameraState: NavigationCameraState) {
```
MaximAlien, line 68: Ah, it was probably me, who removed this part some time ago. At the same time I think in default case to omit code duplication we can use something like this in `CourseUpdatable` protocol extension:
```swift
if !(self is UserHaloCourseView) {
    let angle = CGFloat(CLLocationDegrees(direction - location.course).toRadians())
    self.layer.setAffineTransform(CGAffineTransform.identity.rotated(by: -angle))
}
```

MaximAlien (approved):

PR #3408, opened 2021-09-27
Fix issue, when user puck goes offscreen in some cases.

MaximAlien:
PR fixes an issue, which resulted in user puck going offscreen when there are sharp maneuvers.

Issue was caused by https://github.com/mapbox/mapbox-navigation-ios/pull/3275 (center coordinate of the bounding box should be used, instead of two coordinates of the bounding box itself).

/cc @bamx23


S2Ler (approved):
👍

bamx23 (approved):

PR #3409, opened 2021-09-28
Migrate Turf GeoJSON usage to GeoJSONObject and JSONValue

1ec5:
Migrated Turf GeoJSON usage to GeoJSONObject and JSONValue. Factored out EventLocation to begin migrating some ad-hoc event dictionaries to Codable structs.

Depends on mapbox/turf-swift#154, mapbox/mapbox-maps-ios#715, and #3413. There are no additional public-facing changes on top of the Turf changes.

/cc @mapbox/navigation-ios

### DIFF
--- a/Sources/MapboxCoreNavigation/CLLocation.swift
+++ b/Sources/MapboxCoreNavigation/CLLocation.swift
@@ -4,25 +4,6 @@ import MapboxNavigationNative
 import Turf
 
 extension CLLocation {
-    
-    /// Returns a dictionary representation of the location.
-    public var dictionaryRepresentation: [String: Any] {
-        var locationDictionary: [String: Any] = [:]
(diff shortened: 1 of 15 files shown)


Sources/MapboxCoreNavigation/CLLocation.swift
```diff
@@ -3,26 +3,93 @@ import MapboxDirections
 import MapboxNavigationNative
 import Turf
 
-extension CLLocation {
+struct EventLocation {
```
Udumft, line 6: I'd suggest extracting this part into separate file.

MaximAlien:
Overall looks good to me. I guess build failures will be resolved after finalizing update to the latest version of Mapbox Maps.

MaximAlien (approved):

PR #3413, opened 2021-09-29
MapboxMaps v10.0.0-rc.10

1ec5:
Upgraded the MapboxMaps dependency to mapbox/mapbox-maps-ios@312a5f9a746dd12e7bfed417110e1ff0382d45e4 in anticipation of v10.0.0-rc.10.

This upgrade introduces two `@_spi(Restricted)` imports to account for mapbox/mapbox-maps-ios#708. The map SDK’s default marker image has been vendored into the MapboxNavigation asset catalog, now that mapbox/mapbox-maps-ios#707 has removed the default annotation image option from the map SDK.

Hold until v10.0.0-rc.10 is released.

/cc @mapbox/navigation-ios


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -902,7 +903,7 @@ open class NavigationMapView: UIView {
               let triangleImage = Bundle.mapboxNavigation.image(named: "triangle")?.withRenderingMode(.alwaysTemplate) else { return }
         
         do {
-            try mapView.mapboxMap.style.addImage(triangleImage, id: NavigationMapView.ImageIdentifier.arrowImage)
+            try mapView.mapboxMap.style.addImage(triangleImage, id: NavigationMapView.ImageIdentifier.arrowImage, stretchX: [], stretchY: [])
```
1ec5, line 906: mapbox/mapbox-maps-ios#720 proposes giving these parameters default arguments so they can be omitted again.

1ec5:
#3409 incorporates changes needed for mapbox/mapbox-maps-ios#715.

Sources/MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -208,8 +208,8 @@ open class CarPlayMapViewController: UIViewController {
         
         navigationMapView.userLocationStyle = .puck2D()
         
-        navigationMapView.mapView.ornaments.options.logo._visibility = .hidden
-        navigationMapView.mapView.ornaments.options.attributionButton._visibility = .hidden
+        navigationMapView.mapView.ornaments.options.logo.visibility = .hidden
+        navigationMapView.mapView.ornaments.options.attributionButton.visibility = .hidden
```
1ec5, line 212: We’re going through some lengths to keep hiding attribution on CarPlay, even though the `visibility` property is now `Restricted` SPI, not intended for customer use. Unfortunately, we don’t really have a choice, because the map SDK only provides an attribution control that requires interactivity. mapbox/mapbox-maps-ios#745 proposes an option to add an unobtrusive textual attribution label to the corner of the map.

In the meantime, the user can still access attribution through the phone screen, assuming the application configures CarPlay navigation in the usual way.

S2Ler:
> Looks like `MapboxCoreNavigationTests.testOrderOfExecution()`, `NavigationServiceTests.testMultiLegRoute()` and `NavigationServiceTests.testLocationCourseShouldNotChange()` have similar root cause (navigator reports incorrect status, which contains unexpected `statusIndex`).

I merged this branch into my branch with fixes: https://github.com/mapbox/mapbox-navigation-ios/tree/feature/async-updates-maps-10

I see that completeRoute still fails, but other tests seem to be fine.

Issue #3414, opened 2021-09-29
[Bug]: `URLSession.userAgent` doesn't contain information regarding `MapboxNavigation` while using SPM.


MaximAlien:
### Mapbox Navigation SDK version

v2.0.0-rc.3

### Steps to reproduce

Call `URLSession.userAgent` and check its contents.

### Expected behavior

At this point `MapboxNavigation` is missing:
```
Example/2.0.0 MapboxCoreNavigation/2.0.0 iOS/14.5.0 (x86_64)
```

### Actual behavior

`URLSession.userAgent` should look like this:

```
Example/2.0.0 MapboxNavigation/2.0.0 MapboxCoreNavigation/2.0.0 iOS/14.5.0 (x86_64)
```

### Is this a one-time issue or a repeatable issue?

repeatable

PR #3415, opened 2021-09-29
Improve `MapboxNavigation` bundle detection.

MaximAlien:
Closing #3414.

### DIFF
--- a/Sources/MapboxCoreNavigation/BundleAdditions.swift
+++ b/Sources/MapboxCoreNavigation/BundleAdditions.swift
@@ -39,21 +39,47 @@ extension Bundle {
     }
     
     /**
+     Provides `Bundle` instance, based on provided bundle name and class inside of it.
+     
+     - parameter `bundleName`: Name of the bundle.
+     - parameter `class`: Class, which is located inside of the bundle.
+     - returns: Instance of the bundle if it was found, otherwise `nil`.
+     */
+    static func bundle(for bundleName: String, class: AnyClass) -> Bundle? {
+        let candidates = [
+            // Bundle should be present here when the package is linked into an App.
+            Bundle.main.resourceURL,
+            
+            // Bundle should be present here when the package is linked into a framework.
+            Bundle(for: `class`).resourceURL
+        ]
+        
+        for candidate in candidates {
+            let bundlePath = candidate?.appendingPathComponent(bundleName + ".bundle")
+            if let bundle = bundlePath.flatMap(Bundle.init(url:)) {
+                return bundle
+            }
+        }
+        
+        return nil
+    }
+    
+    /**
      The Mapbox Navigation framework bundle, if installed.
      */
     class var mapboxNavigationIfInstalled: Bundle? {
         get {
-            #if SWIFT_PACKAGE
-            for bundleIdentifier in Bundle.allBundles.compactMap({ $0.bundleIdentifier }) {
-                if bundleIdentifier.contains("MapboxNavigation-MapboxNavigation") {
-                    return Bundle(identifier: bundleIdentifier)
-                }
+            // Assumption: MapboxNavigation.framework includes NavigationViewController and exposes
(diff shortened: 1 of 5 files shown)


Sources/MapboxNavigation/Bundle.swift
```diff
@@ -17,15 +17,13 @@ extension Bundle {
...
+     Returns `UIImage` by searching for it in the current `Bundle` instance.
+     
+     - parameter named: Name of the image.
+     - returns: `UIImage` instance if image was found, `nil` otherwise.
+     */
     func image(named: String) -> UIImage? {
         return UIImage(named: named, in: self, compatibleWith: nil)
     }
-    
-    var microphoneUsageDescription: String? {
```
MaximAlien, line 24: Doesn't seem to be used anywhere. Removing it.

Tests/MapboxCoreNavigationTests/NavigationServiceTests.swift
```diff
@@ -64,7 +64,11 @@ class NavigationServiceTests: TestCase {
     }
 
     func testDefaultUserInterfaceUsage() {
-        XCTAssertTrue(!dependencies.navigationService.eventsManager.usesDefaultUserInterface, "MapboxCoreNavigationTests shouldn't have an implicit dependency on MapboxNavigation due to removing the Example application target as the test host.")
+        let usesDefaultUserInterface = dependencies.navigationService.eventsManager.usesDefaultUserInterface
+        // Even though `MapboxCoreNavigationTests` does not directly link `MapboxNavigation`, it uses
+        // `TestHelper`, which in turn uses `MapboxNavigation`. This means that its bundle will be present
+        // in `MapboxCoreNavigationTests.xctest`.
+        XCTAssertTrue(usesDefaultUserInterface, "Due to indirect linkage, `MapboxNavigation` will be linked to `MapboxCoreNavigationTests`.")
```
MaximAlien, line 71: This looks like an important change. Previously we incorrectly reported `mapboxNavigationIfInstalled`, because `Bundle.allBundles` did not return all used bundles. As I can see `usesDefaultUserInterface` is used in multiple places while sending navigation events, so it'd be good to understand consequences. 

/cc @mapbox/navigation-ios

Sources/MapboxCoreNavigation/BundleAdditions.swift
```diff
@@ -101,38 +127,4 @@ extension Bundle {
...
-        
-        return url.appendingPathComponent("tiles")
-    }
-    
-    /**
-     A file URL at which the application can place a downloaded tile file with the given version identifier.
-     */
-    public func suggestedTileURL(version: String) -> URL? {
-        return suggestedTileURL?.appendingPathComponent(version)
-    }
```
MaximAlien, line 137: These symbols are not used anywhere as I can see, so I've removed them. Please let me know if you think that we should keep them.

Sources/MapboxNavigation/Bundle.swift
1ec5: 👍 This was used for the audio feedback feature that was added in #719 but removed in #870.

1ec5 (approved):

Sources/MapboxCoreNavigation/BundleAdditions.swift
S2Ler: This is public API though. It is a breaking change to remove it. I would keep them.

S2Ler (changes requested):

Sources/MapboxCoreNavigation/BundleAdditions.swift
S2Ler: I will bring back these symbols and merge the PR to proceed with the release. @1ec5 asked to included this PR in the release.

S2Ler: @MaximAlien Please open separate PR if you think we shall remove these methods.

S2Ler (approved):

Issue #3234, opened 2021-08-10
Crash: RouteLegProgress.init


dmiluski:
**Mapbox Navigation SDK version:**

```
      {
        "package": "MapboxCommon",
        "repositoryURL": "https://github.com/mapbox/mapbox-common-ios.git",
        "state": {
          "branch": null,
          "revision": "b8e510896bf064cf14d12624e071839d823ce013",
          "version": "16.0.0"
        }
      },
      {
        "package": "MapboxCoreMaps",
        "repositoryURL": "https://github.com/mapbox/mapbox-core-maps-ios.git",
        "state": {
          "branch": null,
          "revision": "19276a02acb21e76fe6e030e2fd9d8691424f92c",
          "version": "10.0.0-rc.5"
        }
      },
      {
        "package": "MapboxDirections",
        "repositoryURL": "https://github.com/mapbox/mapbox-directions-swift.git",
        "state": {
          "branch": null,
          "revision": "f55782cc1d8db87a35043c20722af9a0fb23fc0e",
          "version": "2.0.0-beta.7"
        }
      },
      {
        "package": "MapboxMobileEvents",
        "repositoryURL": "https://github.com/mapbox/mapbox-events-ios.git",
        "state": {
          "branch": null,
          "revision": "8e0c725271dee0a0c957b9b9738aef81b601f12b",
          "version": "1.0.2"
        }
      },
      {
        "package": "MapboxMaps",
        "repositoryURL": "https://github.com/mapbox/mapbox-maps-ios.git",
        "state": {
          "branch": null,
          "revision": "f809dd694b789a56ad2b99956cf124e1f633ab43",
          "version": "10.0.0-rc.5"
        }
      },
      {
        "package": "MapboxNavigation",
        "repositoryURL": "https://github.com/mapbox/mapbox-navigation-ios",
        "state": {
          "branch": null,
          "revision": "40b17184f7b67f0019121cdedad190df55b9ea00",
          "version": "2.0.0-beta.21"
        }
      },
      {
        "package": "MapboxNavigationNative",
        "repositoryURL": "https://github.com/mapbox/mapbox-navigation-native-ios.git",
        "state": {
          "branch": null,
          "revision": "545bad915c41cadd808c43baf62b51a2349ccb68",
          "version": "60.0.0"
        }
      },
```

### Steps to reproduce

 1. Found in Crashlytics Report

```
Crashed: com.apple.main-thread
[...]

PR #3416, opened 2021-09-29
Fix the preview crash resulted from the table view scrolling quickly.

### LINKED ISSUE #3062: Crash previewing a step from the step table — RouteLegProgress.init(leg:stepIndex:spokenInstructionIndex:)

ShanMa1991:
### Description
This pr is to fix #3062 and #3234 by adding the check when we select the banner instruction for preview.

### Implementation
The crash reported from #3062 , #3234 and a recent crash from user are all resulted from the call of the following:
https://github.com/mapbox/mapbox-navigation-ios/blob/8257c395745e18255e391978e387034f00086290/Sources/MapboxNavigation/StepsViewController.swift#L192-L206
And then we need to form a `RouetLegProgress` to preview the step as:
https://github.com/mapbox/mapbox-navigation-ios/blob/8257c395745e18255e391978e387034f00086290/Sources/MapboxNavigation/NavigationViewController.swift#L1104-L1106

But the issue is that if we scroll the table quickly, during when the `routeProgress` changed and reload the table view data, or the `routeProgress` received in the table view is different from the `routeProgress` in the `navigationService`, there's a chance that the `stepIndex` would be out of range of the current chosen leg steps count. 

So fat the safest way is to add check for the `stepIndex` and the steps count in both the table view select and before the initialization `RouetLegProgress` for preview.

### DIFF
--- a/Sources/MapboxNavigation/NavigationViewController.swift
+++ b/Sources/MapboxNavigation/NavigationViewController.swift
@@ -1104,7 +1104,7 @@ extension NavigationViewController: TopBannerViewControllerDelegate {
     
     public func topBanner(_ banner: TopBannerViewController, didSelect legIndex: Int, stepIndex: Int, cell: StepTableViewCell) {
         let progress = navigationService.routeProgress
-        guard progress.route.checkSelectedIndex(legIndex: legIndex, stepIndex: stepIndex) else { return }
+        guard progress.route.containsStep(at: legIndex, stepIndex: stepIndex) else { return }
         let legProgress = RouteLegProgress(leg: progress.route.legs[legIndex], stepIndex: stepIndex)
         let step = legProgress.currentStep
         self.preview(step: step, in: banner, remaining: progress.remainingSteps, route: progress.route)
--- a/Sources/MapboxNavigation/NavigationViewController.swift
+++ b/Sources/MapboxNavigation/NavigationViewController.swift
@@ -1104,6 +1104,7 @@ extension NavigationViewController: TopBannerViewControllerDelegate {
     
     public func topBanner(_ banner: TopBannerViewController, didSelect legIndex: Int, stepIndex: Int, cell: StepTableViewCell) {
         let progress = navigationService.routeProgress
+        guard stepIndex < progress.route.legs[legIndex].steps.count else { return }
         let legProgress = RouteLegProgress(leg: progress.route.legs[legIndex], stepIndex: stepIndex)
         let step = legProgress.currentStep
         self.preview(step: step, in: banner, remaining: progress.remainingSteps, route: progress.route)
--- a/Sources/MapboxNavigation/NavigationViewController.swift
+++ b/Sources/MapboxNavigation/NavigationViewController.swift
@@ -1104,7 +1104,7 @@ extension NavigationViewController: TopBannerViewControllerDelegate {
     
     public func topBanner(_ banner: TopBannerViewController, didSelect legIndex: Int, stepIndex: Int, cell: StepTableViewCell) {
         let progress = navigationService.routeProgress
-        guard stepIndex < progress.route.legs[legIndex].steps.count else { return }
+        guard progress.route.checkSelectedIndex(legIndex: legIndex, stepIndex: stepIndex) else { return }
         let legProgress = RouteLegProgress(leg: progress.route.legs[legIndex], stepIndex: stepIndex)
         let step = legProgress.currentStep
         self.preview(step: step, in: banner, remaining: progress.remainingSteps, route: progress.route)
--- a/Sources/MapboxNavigation/Route.swift
+++ b/Sources/MapboxNavigation/Route.swift
@@ -155,6 +155,16 @@ extension Route {
             return LineString(coordinateList)
         }
     }
+    
+    /**
+     Returns true if both the legIndex and stepIndex are valid in the route.
+     */
+    func checkSelectedIndex(legIndex: Int, stepIndex: Int) -> Bool {
+        guard let stepCount = legs[safe: legIndex]?.steps.count,
+              stepIndex < stepCount,
+              stepIndex >= 0 else { return false }
+        return true
+    }
 }
 
 extension RouteStep {
--- a/Sources/MapboxNavigation/Route.swift
+++ b/Sources/MapboxNavigation/Route.swift
@@ -159,11 +159,8 @@ extension Route {
     /**
      Returns true if both the legIndex and stepIndex are valid in the route.
      */
-    func checkSelectedIndex(legIndex: Int, stepIndex: Int) -> Bool {
-        guard let stepCount = legs[safe: legIndex]?.steps.count,
-              stepIndex < stepCount,
-              stepIndex >= 0 else { return false }
-        return true
+    func containsStep(at legIndex: Int, stepIndex: Int) -> Bool {
+        return legs[safe: legIndex]?.steps.indices.contains(stepIndex) ?? false
     }
 }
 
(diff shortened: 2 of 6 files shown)


Sources/MapboxNavigation/StepsViewController.swift
```diff
@@ -192,18 +192,24 @@ public class StepsViewController: UIViewController {
 extension StepsViewController: UITableViewDelegate {
     public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         tableView.deselectRow(at: indexPath, animated: true)
+        let legIndex = indexPath.section
         let cell = tableView.cellForRow(at: indexPath) as! StepTableViewCell
         // Since as we progress, steps are removed from the list, we need to map the row the user tapped to the actual step on the leg.
         // If the user selects a step on future leg, all steps are going to be there.
         var stepIndex: Int
-        if indexPath.section > 0 {
```
MaximAlien, line 199: Do you think that it'd be possible to extract this logic into separate function and add unit-test for it? With unit-test we can easily provide negative test case scenario, which will help in future.

S2Ler:
> the routeProgress received in the table view is different from the routeProgress in the navigationService, there's a chance that the stepIndex would be out of range of the current chosen leg steps count.

Why are they getting out of sync? While checking for out of bound access will fix the crash it can result in instructions to be incorrect. I'm currently working on [this](https://github.com/mapbox/mapbox-navigation-ios/tree/feature/restore-tests) branch that will fix some of the crashes related to local route progress isn't in sync with what NavNative reports. We shall investigate why top banner and navigation service are not in sync.

ShanMa1991:
> While checking for out of bound access will fix the crash it can result in instructions to be incorrect.

No, this check is to make sure that the preview is not working, it would not result in the wrong instruction preview or the wrong top banner behavior. I know the root fix is to fix the out of sync behavior of `routeProgress` in the navigation components with the Nav `status`, but this situation may still happen when we scroll or swipe too fast or hard while the new `routeProgress` is forming. 

I think we could right now fix it for stability reason and to avoid the crashes. There're repeated crash reports from the customers caused by this issue. When we fix the out of sync issue, we could remove this safe check. Because these checks only work to prevent the instruction preview, it would not lead to wrong instruction banner presenting or any other weird behavior.

Sources/MapboxNavigation/StepsViewController.swift
ShanMa1991: Thanks for the suggestion! I'll add test for this part.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -1103,6 +1103,7 @@ extension NavigationViewController: TopBannerViewControllerDelegate {
     
     public func topBanner(_ banner: TopBannerViewController, didSelect legIndex: Int, stepIndex: Int, cell: StepTableViewCell) {
         let progress = navigationService.routeProgress
+        guard stepIndex < progress.route.legs[legIndex].steps.count else { return }
```
bamx23, line 1106: Should we also check `legIndex` here? Or to use `safe:` subscript?

ShanMa1991: Thanks for the suggestion, I'll add check to this part as well.

Sources/MapboxNavigation/StepsViewController.swift
```diff
@@ -192,18 +192,21 @@ public class StepsViewController: UIViewController {
...
+        if legIndex > 0 {
             stepIndex = indexPath.row
         } else {
             stepIndex = indexPath.row + routeProgress.currentLegProgress.stepIndex
             // For the current leg, we need to know the upcoming step.
-            stepIndex += indexPath.row + 1 > sections[indexPath.section].count ? 0 : 1
+            stepIndex += indexPath.row + 1 > sections[legIndex].count ? 0 : 1
```
1ec5, line 205:
```suggestion
            if sections[legIndex].indices.contains(indexPath.row + 1) {
                stepIndex += 1
            }
```

1ec5 (approved):

Sources/MapboxNavigation/Route.swift
```diff
@@ -155,6 +155,16 @@ extension Route {
...
     }
+    
+    /**
+     Returns true if both the legIndex and stepIndex are valid in the route.
+     */
+    func checkSelectedIndex(legIndex: Int, stepIndex: Int) -> Bool {
+        guard let stepCount = legs[safe: legIndex]?.steps.count,
+              stepIndex < stepCount,
+              stepIndex >= 0 else { return false }
+        return true
```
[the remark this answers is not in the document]
ShanMa1991: Thanks for the suggestion, I'll improve this part.

PR #3417, opened 2021-09-30
Extend ReplayLocationManager with new callbacks

S2Ler:
`onTick` allows to get each replayed location with its index

`onReplayLoopCompleted` allows to stop replay after all locations
replayed.

Prevent crashes if incoming locations array is empty.

### DIFF
--- a/Sources/MapboxCoreNavigation/ReplayLocationManager.swift
+++ b/Sources/MapboxCoreNavigation/ReplayLocationManager.swift
@@ -18,9 +18,10 @@ open class ReplayLocationManager: NavigationLocationManager {
     /**
      `locations` to be replayed.
      */
-    public var locations: [CLLocation]! {
+    public var locations: [CLLocation] {
         didSet {
             currentIndex = 0
+            verifyParameters()
         }
     }
     
@@ -47,10 +48,22 @@ open class ReplayLocationManager: NavigationLocationManager {
     var startDate: Date?
     
     private var synthesizedLocation: CLLocation?
+
+    /**
+     A handler that is called when `ReplayLocationManager` finished replaying `locations` and about to start from the
+     beginning. Return false to stop replaying `locations`.
+     */
+    var onReplayLoopCompleted: ((ReplayLocationManager) -> Bool)?
+
+    /**
+     A handler that is called on each replayed location along with the location index in `locations` array.
+     */
+    var onTick: ((_ index: Int, CLLocation) -> Void)?
     
     public init(locations: [CLLocation]) {
         self.locations = locations.sorted { $0.timestamp < $1.timestamp }
         super.init()
+        verifyParameters()
     }
     
     deinit {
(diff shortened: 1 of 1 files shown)


Sources/MapboxCoreNavigation/ReplayLocationManager.swift
```diff
@@ -68,24 +80,35 @@ open class ReplayLocationManager: NavigationLocationManager {
     }
     
     @objc internal func tick() {
-        guard let startDate = startDate else { return }
+        guard let startDate = startDate,
+              !locations.isEmpty else {
+                  return
```
Udumft, line 85: Should we message somehow that replaying is stopped due to empty locations? For example we could use delegate to report an error or 'pause' state.

Sources/MapboxCoreNavigation/ReplayLocationManager.swift
```diff
@@ -47,6 +47,17 @@ open class ReplayLocationManager: NavigationLocationManager {
...
+    /**
+     A handler that is called when `ReplayLocationManager` finished replaying `locations` and about to start from the
+     beginning. Return false to stop replaying `locations`.
+     */
+    public var onReplayLoopCompleted: ((ReplayLocationManager) -> Bool)?
+
+    /**
+     A handler that is called on each replayed location along with the location index in `locations` array.
+     */
+    public var onTick: ((_ index: Int, CLLocation) -> Void)?
```
Udumft, line 60: What would be the use-cases for these? For example why `onReplayLoopCompleted` is a callback instead of a flag? What would `onTick` be used for what cannot be covered by `delegate.locationManager(_, didUpdateLocations:)`?

S2Ler: This is mainly for tests. I made this vars internal at first and then thought it can be useful for devs. 

The use case is to replay locations with ReplayLocationManager. The delegate for `ReplayLocationManager` is `RouteController` and tests needs a way to get callbacks when replay is finished to continue assertions (`onReplayLoopCompleted` callback) and perform some logic on each location updates (`onTick` callback). 

wdyt?

S2Ler: This will require having a lot of boilerplate in tests. I would rather make it internal for now and if a request comes, reconsider decision.

Sources/MapboxCoreNavigation/ReplayLocationManager.swift
S2Ler: Too much hassle for such a simple case just for tests. I will add precondition.

S2Ler:
Tests will be fixed in subsequent PR. This one doesn't go to main yet.

Udumft (approved):

Sources/MapboxCoreNavigation/ReplayLocationManager.swift
```diff
@@ -18,9 +18,10 @@ open class ReplayLocationManager: NavigationLocationManager {
     /**
      `locations` to be replayed.
      */
-    public var locations: [CLLocation]! {
+    public var locations: [CLLocation] {
```
azarovalex, line 21: Technically it's a breaking change in public API 🙂
It was possible to write:
```
if manager.locations != nil {
```

S2Ler: Thanks, will mention this in the changelog 👍

PR #3418, opened 2021-09-30
Handling Example's RouteIndex

### LINKED ISSUE #3403: Example application assumes selected route is first route in response
The Example application sets up NavigationViewController with incorrect parameters, leading to inconsistent UI and internal inconsistencies.

## Problem

When selecting an alternative route in the Example application, the ETA and various attributes can be very inaccurate after route refreshing.

## Diagnosis

Each time the user taps to select a new route, ViewController sends that route to the front of the array:

https://github.com/mapbox/mapbox-navigation-ios/blob/b3f19116000945769f1957e83e30f64308be46b8/Example/ViewController.swift#L558-L562

When teeing up NavigationService and NavigationViewController, it therefore assumes the selected route is at index 0:

https://github.com/mapbox/mapbox-navigation-ios/blob/b3f19116000945769f1957e83e30f64308be46b8/Example/ViewController.swift#L263-L264

But NavigationService and NavigationViewController don’t want to know the index of the route in this particular array; these classes need to know the index of the route in the route response that was passed in. In fact, it needs to be the index of the route in the original, unmodified route response. Otherwise, the Route Refresh API will refresh the wrong route, because it only accepts the route index and original response UUID. It knows nothing of the swapping that took place earlier on the client side.

Because of this mismatch, the ETA and various attributes can be very inaccurate after route refreshing, even potentially with the wrong number of legs, exhibiting all the symptoms of #3343. This may also lead to other inconsistencies and undefined behavior throughout Core Navigation.

Thanks to @OttyLab for helping to figure out the root cause.

## Proposed solution

[The “Route Lines Styling” example](https://github.com/mapbox/mapbox-navigation-ios-examples/blob/5dff4d5ae8da64d9ee3b7a3acded7f02ec91459f/Navigation-Examples/Examples/Route-Lines-Styling.swift#L193-L195) only keeps track of a dynamic route index as the user selects routes. Then it [carefully builds a temporary array](https://github.com/mapbox/mapbox-navigation-ios-examples/blob/5dff4d5ae8da64d9ee3b7a3acded7f02ec91459f/Navigation-Examples/Examples/Route-Lines-Styling.swift#L43-L47) in priority order to pass into `NavigationMapView.show(_:legIndex:)`. ViewController should do the same thing.

/cc @mapbox/navigation-ios

Udumft:
Resolves #3403 

This PR refines handling list of routes and passing a valid `routeIndex` to the `[CarPlay]NavigationViewController`.
Example does not longer swap data in the original `RouteResponse` but instead stores user selection separately.

### DIFF
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -54,20 +54,44 @@ class ViewController: UIViewController {
             }
         }
     }
+    
+    var currentRouteIndex = 0 {
+        didSet {
+            showCurrentRoute()
+        }
+    }
 
+    func showCurrentRoute() {
+        guard let currentRoute = currentRoute else { return }
+        
+        var routes = [currentRoute]
+        routes.append(contentsOf: self.routes!.filter {
+            $0 != currentRoute
+        })
+        // Show congestion levels on alternative route lines if there're multiple routes in the response.
+        navigationMapView.showsCongestionForAlternativeRoutes = true
+        navigationMapView.show(routes)
+        navigationMapView.showWaypoints(on: currentRoute)
+        navigationMapView.showRouteDurations(along: routes)
+    }
+    
+    var currentRoute: Route? {
+        return routes?[currentRouteIndex]
+    }
+    
+    var routes: [Route]? {
+        return response?.routes
+    }
+    
     var response: RouteResponse? {
         didSet {
-            guard let routes = response?.routes, let currentRoute = routes.first else {
+            guard let routes = response?.routes, !routes.isEmpty else {
                 clearNavigationMapView()
                 return
             }
             
             startButton.isEnabled = true
-            // Show congestion levels on alternative route lines if there're multiple routes in the response.
-            navigationMapView.showsCongestionForAlternativeRoutes = true
-            navigationMapView.show(routes)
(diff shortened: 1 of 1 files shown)


Example/ViewController.swift
```diff
@@ -54,20 +54,44 @@ class ViewController: UIViewController {
             }
         }
     }
+    
+    var currentRouteIndex = 0 {
```
Udumft, line 58: I intentionally did not use `IndexedRouteResponse` to store this as it is convenient to have custom `didSet` for index update.

Example/ViewController.swift
```diff
@@ -54,20 +54,44 @@ class ViewController: UIViewController {
...
+        }
+    }
 
+    func showCurrentRoute() {
+        guard let currentRoute = currentRoute else { return }
+        
+        var routes = [currentRoute]
+        routes.append(contentsOf: self.routes!.filter {
+            $0 != currentRoute
+        })
```
1ec5, line 70: This would be a bit easier to understand:

```swift
guard var prioritizedRoutes = routes else { return }
prioritizedRoutes.insert(prioritizedRoutes.remove(at: currentRouteIndex), at: 0)
```

1ec5 (approved):

PR #3426, opened 2021-09-30
Add `CPInterfaceController.safePopTemplate(animated:)` to be able to safely pop templates.

MaximAlien:
Add `CPInterfaceController.safePopTemplate(animated:)` to be able to safely pop templates.


Sources/MapboxNavigation/CPInterfaceController.swift
```diff
@@ -0,0 +1,20 @@
...
+     Allows to safely pop existing `CPTemplate`.
+     
+     In case if there is only one `CPTemplate` left in the stack of templates, popping operation
+     will not be performed.
+     
+     - parameter animated: Boolean flag which determines whether `CPTemplate` popping will be
+     animated or not.
+     */
+    func safePopTemplate(animated: Bool) {
+        if templates.count == 1 { return }
```
bamx23, line 16: Do you have any ideas why this happens?

MaximAlien: By the looks of it this happens when user selects search result. At the same time I was not able to reproduce it manually. Only by calling `popTemplate(animated:)` several times.

ShanMa1991 (approved):

Issue #3375, opened 2021-09-17
Fix and enable `NavigationServiceTests`. 


MaximAlien:
In scope of https://github.com/mapbox/mapbox-navigation-ios/pull/3342/ `testProactiveRerouting` and `testReroutingFromALocationSendsEvents` were disabled due to failures because of different Navigation Native behavior.

In scope of this ticket root cause of failures should be investigated and disabled tests fixed.

Issue #3298, opened 2021-08-25
Crash: RouteController.updateIndexes(status:progress:) 


dmiluski:
Experiencing Crash using Nav SDK
Currently our top crasher
Device States: 21% occurs in background

```
Crashed: com.apple.main-thread
0  Straightaway                   0x10109a910 RouteController.updateIndexes(status:progress:) + 4347160848
1  Straightaway                   0x101099f54 RouteController.update(to:) + 4347158356
2  Straightaway                   0x1010a0104 partial apply for closure #1 in RouteController.navigationStatusDidChange(_:) + 4347183364
3  Straightaway                   0x1010571d8 thunk for @escaping @callee_guaranteed () -> () + 4346884568
4  libdispatch.dylib              0x1804e1a84 _dispatch_call_block_and_release + 32
5  libdispatch.dylib              0x1804e381c _dispatch_client_callout + 20
6  libdispatch.dylib              0x1804f1c70 _dispatch_main_queue_callback_4CF + 884
7  CoreFoundation                 0x180870340 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 16
8  CoreFoundation                 0x18086a218 __CFRunLoopRun + 2524
9  CoreFoundation                 0x180869308 CFRunLoopRunSpecific + 600
10 GraphicsServices               0x197eec734 GSEventRunModal + 164
11 UIKitCore                      0x1832e775c -[UIApplication _run] + 1072
12 UIKitCore                      0x1832ecfcc UIApplicationMain + 168
13 Straightaway                   0x100b71f1c main + 12 (AppDelegate.swift:12)
14 libdyld.dylib                  0x180525cf8 start + 4
```

**Mapbox Navigation SDK version:**

        "package": "MapboxNavigation",
        "repositoryURL": "https://github.com/mapbox/mapbox-navigation-ios",
        "state": {
          "branch": null,
          "revision": "40b17184f7b67f0019121cdedad190df55b9ea00",
          "version": "2.0.0-beta.21"
        }

### Steps to reproduce

Found in Crashlytics

### Expected behavior

Not Crash

### Actual behavior

Crashes

1ec5:
Are you still seeing this crash after upgrading to v2.0.0-rc.1 or above? We’ve made some changes to route and route progress management that could resolve it.

dmiluski:
Hi @1ec5 ,

[...]

PR #3432, opened 2021-10-04
Fix routeProgress get out of sync with NavNative status

### LINKED ISSUE #3343: Parallel copies of route in Router can get out of sync

S2Ler:
Fixes #3343 #3375 #3298


1ec5 (commented):
Wow, this is quite a PR. Can you say a bit more about what fixed #3343 and #3298?

S2Ler:
> Wow, this is quite a PR. Can you say a bit more about what fixed #3343 and #3298?

#3343 I read this issue once again and it won't fix this particular issue, but probably fixes some symptoms.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -231,8 +265,8 @@ open class RouteController: NSObject {
     @objc private func navigationStatusDidChange(_ notification: NSNotification) {
         guard let userInfo = notification.userInfo,
               let status = userInfo[Navigator.NotificationUserInfoKey.statusKey] as? NavigationStatus else { return }
-        DispatchQueue.main.async { [weak self] in
-            self?.update(to: status)
+        onMainQueueSync {
```
S2Ler, line 268: `navigationStatusDidChange` is called on the thread that create NavNative navigator which is most likely to be the main thread, but not guaranteed at the moment. Solutions:
- `onMainQueueSync`.
- Put preconditions that navigator must be created on the main thread.
- Enforce the creation of NavNative navigator on the main thread. 

I think the best solution would be " Enforce the creation of NavNative navigator on the main thread".

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -261,13 +295,11 @@ open class RouteController: NSObject {
     }
     
     @objc func fallbackToOffline(_ notification: Notification) {
-        self.updateNavigator(with: self.routeProgress)
-        self.updateRouteLeg(to: self.routeProgress.legIndex)
+        updateNavigator(with: self.routeProgress, completion: { _ in })
```
S2Ler, line 298: `updateNavigator` already updates routeLeg. I don't see a need to do this twice using two separate methods, given that both are async.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -435,30 +467,14 @@ open class RouteController: NSObject {
         BillingHandler.shared.beginBillingSession(for: .activeGuidance, uuid: sessionUUID)
 
         subscribeNotifications()
-        updateNavigator(with: routeProgress)
-        updateObservation(for: routeProgress)
+        updateNavigator(with: routeProgress, completion: {_ in })
```
S2Ler, line 470: Technically speaking active guidance doesn't start until the completion is called. We can think how to improve this part.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -435,30 +467,14 @@ open class RouteController: NSObject {
...
-    func resetObservation(for progress: RouteProgress) {
-        progress.legIndexHandler = nil
-    }
-    
-    func updateObservation(for progress: RouteProgress) {
-        progress.legIndexHandler = { [weak self] (oldValue, newValue) in
-            guard newValue != oldValue else {
-                return
-            }
-            self?.updateRouteLeg(to: newValue)
```
S2Ler, line 458: You are not supposed to change legIndex manually in progress. It should only change in response to Navigator status update.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -597,7 +613,7 @@ extension RouteController: Router {
             return userIsWithinRadiusOfDestination(location: location)
         }
         else {
-            let offRoute = status.routeState == .offRoute || status.routeState == .invalid
+            let offRoute = status.routeState == .offRoute
```
S2Ler, line 616: `status.routeState == .invalid` doesn't mean we are offroute. It means that we are still in passive navigation. See https://github.com/mapbox/mapbox-navigation-ios/pull/3432/files#r722128724

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -251,39 +255,40 @@ extension InternalRouter where Self: Router {
...
-                var routeOptions: RouteOptions?
-                if case let .route(options) = response.options {
-                    routeOptions = options
-                }
-                
-                // Prefer the most optimal route (the first one) over the route that matched the original choice.
-                self.indexedRouteResponse = .init(routeResponse: response, routeIndex: 0)
-                
-                // If the upcoming maneuver in the new route is the same as the current upcoming maneuver, don’t announce it.
-                // FIXME: There must be a better way to skip a redundant initial instruction than to assume the old route’s spoken instruction index is appliable to the new route.
```
S2Ler, line 281: It doesn't look like this code works with new NavNative implementation. We need to find another way of skipping these instructions on the NavNative side.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -251,39 +255,40 @@ extension InternalRouter where Self: Router {
...
+                self.isRerouting = false; return
             }
+            var routeOptions: RouteOptions?
+            if case let .route(options) = response.options {
+                routeOptions = options
+            }
+
+            // Prefer the most optimal route (the first one) over the route that matched the original choice.
+            let indexedRouteResponse = IndexedRouteResponse(routeResponse: response, routeIndex: 0)
+            (self as! RouteController)
```
S2Ler, line 286: leftover, to be fixed.

Sources/TestHelper/DirectionsSpy.swift
```diff
@@ -28,14 +29,14 @@ public class DirectionsSpy: Directions {
         
         let session: Directions.Session = (options: options, credentials: credentials)
         guard let lastCalculateOptionsCompletion = lastCalculateOptionsCompletion else {
-            assert(false, "Can't fire a completion handler which doesn't exist!")
+            XCTFail("Can't fire a completion handler which doesn't exist!")
```
S2Ler, line 32: Tests shouldn't stop if we hit this code-path. It is ok to just fail.

Sources/TestHelper/DirectionsSpy.swift
```diff
@@ -28,14 +29,14 @@ public class DirectionsSpy: Directions {
...
             return
         }
         
         if let error = error {
             lastCalculateOptionsCompletion(session, .failure(error))
         } else {
-            let response = RouteResponse(httpResponse: nil, routes: routes, waypoints: waypoints, options: .route(options), credentials: credentials)
+            let response = RouteResponse(httpResponse: nil, identifier: "mock", routes: routes, waypoints: waypoints, options: .route(options), credentials: credentials)
```
S2Ler, line 39: Identifier is required for NavNative.

PR #3439, opened 2021-10-05
Docs generation automation

### LINKED ISSUE #2078: Verify jazzy finds USRs on CI

### LINKED ISSUE #3184: Update jazzy table of contents
The jazzy table of contents needs to be updated to reflect #3045 and other changes:

```
WARNING: No documented top-level declarations match name "AccountManager" specified in categories file
WARNING: No documented top-level declarations match name "PointAnnotation" specified in categories file
WARNING: No documented top-level declarations match name "Style" specified in categories file
WARNING: No documented top-level declarations match name "RoadObjectsStore" specified in categories file
WARNING: No documented top-level declarations match name "RoadObjectsStoreDelegate" specified in categories file
```

#2078 or something like it would help to catch issues like this sooner.

/cc @mapbox/navigation-ios

chezzdev:
This PR adds docs generation on CI, specifically:
- builds docs for each commit
- checks for warnings and USR messages in Jazzy output (resolves #2078)
- pushes docs back to the repo for release tags (see note below)
- fixes current warnings from Jazzy (resolves #3184)

### Note on publishing docs
I tested the publishing setup as close to a real environment as possible, except I haven't used any tags to avoid possible confusion on the customer side. Thus this PR only pushes generated docs to the new branch `docs-v...` from `publisher-production` and not to it directly. We'll need to verify that the branch has the expected content and merge it manually. Once it's done, the script can be altered to push directly to `publisher-production` (issue pending).


docs/jazzy.yml
```diff
@@ -63,9 +63,6 @@ custom_categories:
       - NavigationMapView
       - NavigationMapViewDelegate
       - MapView
-      - AccountManager
-      - PointAnnotation
```
chezzdev, line 67: These symbols don't exist in Nav SDK.

docs/jazzy.yml
```diff
@@ -63,9 +63,6 @@ custom_categories:
...
-      - Style
```
chezzdev, line 68: This is already mentioned in the "Styling" section.

docs/jazzy.yml
1ec5: Yep, these were previously extensions in the navigation SDK, but we’ve since removed the extensions.

.circleci/config.yml
```diff
@@ -579,6 +586,53 @@ jobs:
...
+            matches:
+              pattern: "^v.*\\..*\\..*$"
+              value: << pipeline.git.tag >>
+          steps:
+            - run:
+                name: Push docs to publisher-production
+                command: |
+                  git remote set-url origin https://x-access-token:$(mbx-ci github writer token)@github.com/mapbox/mapbox-navigation-ios.git
+                  git config user.email "release-bot@mapbox.com"
+                  git config user.name "Release SDK bot"
```
1ec5, line 625: Should this be something like “Mapbox” or “MapboxCI” for consistency with other release bots around Mapbox? Or just “Mapbox Releases”, since the e-mail address doesn’t seem to be specific to a particular SDK?

Gemfile
```diff
@@ -2,3 +2,4 @@ source 'https://rubygems.org'
 gem 'fastlane'
 gem 'snapshot'
 gem 'cocoapods', '~> 1.10'
+gem 'jazzy'
```
1ec5, line 5: Consider specifying a version range so we don’t find out about a new major version when it breaks and interrupts a release process.

scripts/document.sh
```diff
@@ -53,7 +43,7 @@ find Sources/Mapbox{Core,}Navigation/ -name '*.swift' -exec \
 find Sources/Mapbox{Core,}Navigation/ -name '*.[hm]' -exec \
     perl -pi -e 's/([<"])MapboxCoreNavigation\b/$1MapboxNavigation/' {} \;
 
-jazzy \
+bundle exec jazzy \
```
1ec5, line 46: Nice, I guess we could even simplify this documentation (which is for developers who want to preview the generated documentation):

https://github.com/mapbox/mapbox-navigation-ios/blob/5382dc06ef592805dc17689ace4be423aa6d8ee9/docs/README.md#L3

.circleci/config.yml
```diff
@@ -579,6 +586,53 @@ jobs:
...
+                  git checkout -- Sources
+                  git checkout publisher-production
+                  BRANCH="docs-${OUTPUT}" # push directly to publisher-production when verified on a release tag
+                  git checkout "${BRANCH}" || git checkout -b "${BRANCH}"
+                  git add "${OUTPUT}"
+                  git commit -m "v${OUTPUT}"
+                  git push origin "${BRANCH}"
```
1ec5, line 632: Break out these steps into a separate script, as in #2668, so that we can easily recover from any failed automation manually.

.circleci/config.yml
```diff
@@ -658,6 +712,13 @@ workflows:
           context: Slack Orb
       - spm-core-integration-test-job:
           name: "Xcode 13; iOS 15.0; SPM Core test"
+      - generate-docs-job:
+          name: Generate docs and push it for release tag
```
1ec5, line 716: This is a bit verbose compared to other job names.

```suggestion
          name: Generate/Publish Documentation
```

.circleci/config.yml
```diff
@@ -579,6 +586,53 @@ jobs:
...
+      - *install-mbx-ci
+      - *restore-cache-podmaster
+      - *restore-cache-cocoapods-installation
+      - run:
+          name: Generate docs
+          command: |
+            OUTPUT=$( echo << pipeline.git.tag >> | sed 's/^v//' )
+            OUTPUT="${OUTPUT:-documentation}"
+            echo "export OUTPUT=$OUTPUT" >> $BASH_ENV
+            OUTPUT="${OUTPUT}" scripts/document.sh 2>&1 | tee docs.output && ! egrep -e "(WARNING)|(USR)" docs.output
```
1ec5, line 611: #2078 requests checking for missing USR warnings. Some of the warnings are expected due to `#if` usage, but others are not. How should we distinguish the two cases?

chezzdev: Is there a pattern for the expected warnings? There are no warnings at this time.

.circleci/config.yml
```diff
@@ -658,6 +712,13 @@ workflows:
           context: Slack Orb
       - spm-core-integration-test-job:
           name: "Xcode 13; iOS 15.0; SPM Core test"
+      - generate-docs-job:
+          name: Generate docs and push it for release tag
+          context: Slack Orb
+          filters:
+            tags:
+              only: /.*/
```
bamx23, line 720: Do you want to filter out this build in regular branches? As it these lines:
```yaml
branches:
  ignore: /.*/
```

docs/jazzy.yml
```diff
@@ -181,11 +178,7 @@ custom_categories:
...
       - RoadObjectStore
-      - RoadObject.Location
-      - RoadObject.Kind
       - RoadObjectMatcherDelegate
       - RoadObjectStoreDelegate
-      - RoadObject.EdgeLocation
       - RoadObjectMatcherError
-      - RoadObject.Position
-      - RoadObject.Identifier
+
```
bamx23, line 184: White spaces
```suggestion

```

scripts/document.sh
bamx23, line 46: Do you have `bundle install` somewhere? I would even recommend running `bundle check || bundle install --path vendor/bundle` and cache gems:
```yaml
step-library:
    - &restore-gems-cache
        restore_cache:
            keys:
                - cache-v1-gems-{{ checksum "Gemfile.lock" }}

    - &save-gems-cache
        save_cache:
            key: cache-v1-gems-{{ checksum "Gemfile.lock" }}
            paths:
                - vendor/bundle
```

.circleci/config.yml
chezzdev: Changed it to Mapbox Releases.

Gemfile
chezzdev: Good catch, pinned it to `~> 0.14`.

scripts/document.sh
chezzdev: Yeah, now we need to require Bundler in this instruction.

.circleci/config.yml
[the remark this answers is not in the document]
chezzdev: Good catch!

scripts/document.sh
chezzdev: I've added `bundle check || bundle install` to the script. As for caching, it's already implemented in job steps. I've renamed them to be more clear and universal. PTAL

.circleci/config.yml
chezzdev: This was a way to trigger the job for all the tags. Reading docs closely, I understood that we can put semver filter here, it will still trigger job for all the branch pushes.

.circleci/config.yml
```diff
@@ -579,6 +586,53 @@ jobs:
...
+            OUTPUT=$( echo << pipeline.git.tag >> | sed 's/^v//' )
+            OUTPUT="${OUTPUT:-documentation}"
+            echo "export OUTPUT=$OUTPUT" >> $BASH_ENV
+            OUTPUT="${OUTPUT}" scripts/document.sh 2>&1 | tee docs.output && ! egrep -e "(WARNING)|(USR)" docs.output
+      - *save-cache-podmaster
+      - *save-cache-cocoapods-installation
+      - when:
+          condition:
+            matches:
+              pattern: "^v.*\\..*\\..*$"
```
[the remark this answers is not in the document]
chezzdev: Yeah, we can add a more strict pattern. To allow prerelease versions to match I've added `(-.+)?`.
https://github.com/mapbox/mapbox-navigation-ios/blob/9a200f6c075b239bd6eaff0af340782d78551d0e/.circleci/config.yml#L711-L713
I've also moved it to a workflow level.

PR #3442, opened 2021-10-05
Only allow the current guidance card be highlighted.

### LINKED ISSUE #3441: Incorrect Guidance Cards highlighted
### Version: 
2.0.0

### Steps to reproduce
1. choose a complex route with multiple turn and short distance between each turn
2. start navigation with `Guidance Cards` in `Example`

### Expected behavior
Only the current guidance card (the left most one) would be highlighted when the distance to end of this route is less than 500 ft

### Actual behavior
1. The next guidance card, if its distance is less than 500ft, will also be highlighted.
2. And the other guidance card, in the preview section, if their distances are less than 500 ft, would also be highlighted.

### Cause

This issue is caused by the  calling of `InstructionsCardContainerView.updateInstructionCard(distance:)`. It's called by 2 places.
1. by the card cell generation. Because when the card is formed, we only need to highlight the current one.
2. by the `InstructionsCardViewController.updateVisibleInstructionCards(at:)`. Because it's called when there's new routeProgress, and all the visible cards (both the current one and the next one) would be updated, their distances would be compared with the threshold to consider whether to highlight.

To fix this issue, we should only allow the current card be updated with the background color.

ShanMa1991:
### Description
This pr is to fix #3441 by only allowing the current guidance card step highlighted.

### Implementation
Adding the check of the card step before the distance check, to see whether it's the current step. If false, don't allow the background color change.

### Screenshots or Gifs
fxiedCard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@
 * Fixed an issue where some banner instructions were occasionally skipped. ([#3265](https://github.com/mapbox/mapbox-navigation-ios/pull/3265))
 * Improved the current road name label’s performance and fixed a potential crash when updating it. ([#3340](https://github.com/mapbox/mapbox-navigation-ios/pull/3340))
 * Fixed an issue where arrival guidance card appears too early. ([#3383](https://github.com/mapbox/mapbox-navigation-ios/pull/3383))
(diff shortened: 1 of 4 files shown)


S2Ler (approved):
👍

bamx23 (approved):

PR #3451, opened 2021-10-07
Reload guidance cards on leg change

S2Ler:
Currently, only visible guidance cards will be reloaded. If the number of steps on first leg is 1, and on the following leg more than 1, guidance cards on the following leg will have only one instruction until second step is reached.


S2Ler:
> Do we need to mention this in the changelog?

I will add

S2Ler:
I investigated the behavior and the fix doesn't make things worse, only better.

PR #3454, opened 2021-10-08
Rerouting avoiding Dangerous Maneuvers

Udumft:
Added `RouterDelegate/NavigationServiceDelegate/NavigationViewControllerDelegate` method to allow overriding  `RouteOptions.avoidManeuversInOriginRadius` for rerouting. This setting is useful for adjusting rerouting logic according to current user's velocity to not force him into unexpected turns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,7 +91,7 @@
 ### Other changes
 
 * Added the `RouterDelegate.router(_:initialManeuverBufferWhenReroutingFrom:)` method and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to configure rerouting buffering. ([#3454](https://github.com/mapbox/mapbox-navigation-ios/pull/3454))
-* Added `RouterDelegate.router(_:, requestSourceForReroutingWith:)` and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to provide customization for rerouting mechanism by providing a user-defined route as a reroute. ([#3472](https://github.com/mapbox/mapbox-navigation-ios/pull/3472))
+* Added `RouterDelegate.router(_:requestBehaviorForReroutingWith:)` and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to provide customization for rerouting mechanism by providing a user-defined route as a reroute. ([#3472](https://github.com/mapbox/mapbox-navigation-ios/pull/3472))
 
 ## v2.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@
 ### Other changes
 
 * Added the `RouterDelegate.router(_:initialManeuverBufferWhenReroutingFrom:)` method and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to configure rerouting buffering. ([#3454](https://github.com/mapbox/mapbox-navigation-ios/pull/3454))
+* Added `RouterDelegate.router(_:, requestSourceForReroutingWith:)` and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to provide customization for rerouting mechanism by providing a user-defined route as a reroute. ([#3472](https://github.com/mapbox/mapbox-navigation-ios/pull/3472))
 
 ## v2.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,7 +90,7 @@
 
 ### Other changes
 
-* Added `RouterDelegate.router(_:, maneuverOffsetWhenReroutingFrom:)` and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to expose configuration of safe distance marging when reroute occures.  ([#3454](https://github.com/mapbox/mapbox-navigation-ios/pull/3454))
+* Added the `RouterDelegate.router(_:initialManeuverBufferWhenReroutingFrom:)` method and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to configure rerouting buffering. ([#3454](https://github.com/mapbox/mapbox-navigation-ios/pull/3454))
 
 ## v2.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,10 @@
 * Fixed an issue where the user interface did not necessarily display distances in the same units as the route by default. `NavigationRouteOptions` and `NavigationMatchOptions` now set `DirectionsOptions.distanceMeasurementSystem` to a default value matching the `NavigationSettings.distanceUnit` property. ([#3541](https://github.com/mapbox/mapbox-navigation-ios/pull/3541))
 * Introduced `RoutingProvider` to parameterize routing fetching and refreshing during active guidance sessions.  `Directions.calculateWithCache(options:completionHandler:)` and `Directions.calculateOffline(options:completionHandler)` functionality is deprecated by `MapboxRoutingProvider`. It is now recommended to use `MapboxRoutingProvider` to request or refresh routes instead of `Directions` object but you may also provide your own `RoutingProvider` implementation to `NavigationService`, `RouteController` or `LegacyRouteController`. Using `directions` property of listed above entities is discouraged, you should use corresponding `routingProvider` instead, albeit `Directions` also implements the protocol. ([#3261](https://github.com/mapbox/mapbox-navigation-ios/pull/3261))
 
+### Other changes
+
+* Added `RouterDelegate.router(_:, maneuverOffsetWhenReroutingFrom:)` and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to expose configuration of safe distance marging when reroute occures.  ([#3454](https://github.com/mapbox/mapbox-navigation-ios/pull/3454))
+
 ## v2.0.1
 
 * Added the `Notification.Name.didArriveAtWaypoint` constant for notifications posted when the user arrives at a waypoint. ([#3514](https://github.com/mapbox/mapbox-navigation-ios/pull/3514))
--- a/Sources/MapboxCoreNavigation/NavigationServiceDelegate.swift
+++ b/Sources/MapboxCoreNavigation/NavigationServiceDelegate.swift
@@ -43,7 +43,7 @@ public protocol NavigationServiceDelegate: AnyObject, UnimplementedLogging {
     /**
      Configures distance (in meters) before the first maneuver in requested reroute.
      
-     If implemented, this method allows to override set `RouteOptions.avoidManeuversInOriginRadius` value which is useful when adjusting reroute according to current user velocity in order to avoid dangerous maneuvers in the beginning of the route.
+     If implemented, this method allows to override set `RouteOptions.initialManeuverAvoidanceRadius` value which is useful when adjusting reroute according to current user velocity in order to avoid dangerous maneuvers in the beginning of the route.
      
      This method is called after `navigationService(_:willRerouteFrom:)` is called, and before `navigationService(_:didRerouteAlong:)` is called.
      
--- a/Sources/MapboxCoreNavigation/NavigationServiceDelegate.swift
+++ b/Sources/MapboxCoreNavigation/NavigationServiceDelegate.swift
@@ -33,7 +33,7 @@ public protocol NavigationServiceDelegate: AnyObject, UnimplementedLogging {
     /**
      Called immediately before the navigation service calculates a new route.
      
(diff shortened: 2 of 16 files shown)


Sources/MapboxCoreNavigation/RouterDelegate.swift
```diff
@@ -1,6 +1,27 @@
 import Foundation
 import CoreLocation
 import MapboxDirections
+import Turf
+
+/**
+ Configuration for offsetting first route maneuver during rerouting.
+ */
+public enum ReroutingManeuverOffset {
```
1ec5, line 9: “Offset” implies a simple magnitude in a certain direction. “Buffer” would be a more accurate way of describing this behavior if we can’t simplify it down to a single number. If we can simplify it, “buffer” would also be less ambiguous than “offset”, which is usually a lateral offset in map contexts.

Sources/MapboxCoreNavigation/RouterDelegate.swift
```diff
@@ -1,6 +1,27 @@
...
+    /**
+     Leaves original `RouteOptions.initialManeuverAvoidanceRadius` to be used for rerouting attempt.
+     */
+    case `default`
+    /**
+     Disables offsetting the maneuver.
+     */
+    case disabled
```
1ec5, line 17: What’s the difference between `disabled` and `radius(0)`? Could we simplify this type to `LocationDistance?`, where `nil` is the default behavior and a value of 0 disables the behavior?

CHANGELOG.md
```diff
@@ -32,6 +33,10 @@
 * Fixed a retain cycle in CarPlay implementation of a navigation map view that prevented `NavigationMapView` instances from being deallocated after CarPlay is stopped. ([#3552](https://github.com/mapbox/mapbox-navigation-ios/pull/3552))
 * Fixed an issue where the entire route line was colored as `NavigationMapView.routeCasingColor` instead of `NavigationMapView.trafficUnknownColor` when traffic congestion data was missing. ([#3577](https://github.com/mapbox/mapbox-navigation-ios/pull/3577))
 
+### Other changes
+
+* Added `RouterDelegate.router(_:, maneuverOffsetWhenReroutingFrom:)` and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to expose configuration of safe distance marging when reroute occures.  ([#3454](https://github.com/mapbox/mapbox-navigation-ios/pull/3454))
```
1ec5, line 38:
```suggestion
* Added the `RouterDelegate.router(_:maneuverOffsetWhenReroutingFrom:)` method and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to configure rerouting buffering. ([#3454](https://github.com/mapbox/mapbox-navigation-ios/pull/3454))
```

Sources/MapboxCoreNavigation/RouterDelegate.swift
```diff
@@ -28,13 +49,26 @@ public protocol RouterDelegate: AnyObject, UnimplementedLogging {
...
+     
+     If implemented, this method allows to override set `RouteOptions.initialManeuverAvoidanceRadius` value which is useful when adjusting reroute according to current user velocity in order to avoid dangerous maneuvers in the beginning of the route.
+     
+     This method is called after `router(_:willRerouteFrom:)` is called, and before `router(_:didRerouteAlong:)` is called.
+     
+     - parameter router: The router that has detected the need to calculate a new route.
+     - parameter location: The user’s current location.
+     - returns: `ReroutingManeuverOffset` value which overrides or leaves maneuvers offset as it was originally set.
+     */
+    func router(_ router: Router, maneuverOffsetWhenReroutingFrom location: CLLocation) -> ReroutingManeuverOffset
```
1ec5, line 70: Similar to `initialManeuverAvoidanceRadius`, add “initial” so that the developer doesn’t conclude that it affects every maneuver.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -310,6 +310,18 @@ extension InternalRouter where Self: Router {
         routeTask?.cancel()
         let options = progress.reroutingOptions(with: origin)
         
+        if isRerouting {
+            switch (delegate?.router(self, maneuverOffsetWhenReroutingFrom: origin) ?? RouteController.DefaultBehavior.reroutingManeuverRadius) {
+            case .disabled:
+                options.initialManeuverAvoidanceRadius = nil
```
1ec5, line 316: This is `nil` when talking to the Directions API because the Directions API’s default behavior is not to avoid a maneuver at all. But since the default behavior in the navigation SDK is different, `nil` can have a different meaning.

Sources/MapboxCoreNavigation/RouterDelegate.swift
Udumft: Directions will report an error if user tries to send `radius(0)`. `disabled` is a proper way to disable the feature. I decided to use an enum for this to a) make it explicit when feature is disabled or enabled and b) provide a clean explanation what measure is used for calculation. It will be much easier to introduce new cases like `distanceAhead` or `numUpcomingIntersections` or anything else if needed.

Sources/MapboxCoreNavigation/Router.swift
Udumft: Default behavior is configured by `RouteOptions.initialManeuverAvoidanceRadius`. `ReroutingManeuverOffset` is used then to avoid using `nil`s and thus remove the ambiguity of it's meaning by asking directly if user wants the `default` is he configured previously or to override the value. If user didn't configure `initialManeuverAvoidanceRadius` and left it too as a default - this is the same default as Directions expect.

Sources/MapboxCoreNavigation/RouterDelegate.swift
chezzdev: I can see how with this enum, we want to present 3 states of overriding `RouteOptions.initialManeuverAvoidanceRadius`:
- don't override (use original value, that was set during Directions request), (1)
- override:
  - disable buffer whatsoever, (2)
  - override buffer with a concrete radius value. (3)

And it looks kind of like `Optional(Optional(LocationDistance))` type, where external `nil` would represent NavigationSDK-default behavior (reuse the original value) and internal `nil` would represent directions-swift-default behavior (don't account for any radius).

While this is a justified approach, I would find it more simple and clear, if for the state (2) we could use value `0`.

This applies to directions-swift too, as `0` is semantically an equivalent to `nil` in `RouteOptions.initialManeuverAvoidanceRadius`, and they both mean "don't take any buffer into account".
However, if we set `.initialManeuverAvoidanceRadius` to `0`, we will actually [take a buffer of radius `1` into account](https://github.com/mapbox/mapbox-directions-swift/pull/609/files#diff-b186eea493906feb8b8810201920ee0e5d06a8cf5ff1085728c52cfd70715267R209-R212), while building a route.

In short, as [suggested in Minh's comment](https://github.com/mapbox/mapbox-navigation-ios/pull/3454/files#r747157165), I think we could use `LocationDistance?` in place of this type, but for that to work, we should allow for `0` value in `RouteOptions.initialManeuverAvoidanceRadius` and make it equivalent to `nil`.

Even if we stay with this better-described type, I think the change to directions-swift will help avoid ambiguity in using this property.

MaximAlien:
Moved to `v2.1-rc` for now as `v2.1-beta` is no longer relevant. @Udumft please also rebase onto `main`.

Sources/MapboxCoreNavigation/RouterDelegate.swift
Udumft: Using `LocationDistance?` instead of `ReroutingManeuverBuffer` creates ambiguity in next way: `nil` is used in `RouteOptions` to disable the feature, while `nil` for rerouting actually means "do not modify RouteOptions value". Even more - it may look like a bug if user compares 2 cases:
1. `RouteOptions.initialManeuverAvoidanceRadius = nil` + `initialManeuverBufferWhenReroutingFrom = nil` -> feature is disabled on a reroute;
2. `RouteOptions.initialManeuverAvoidanceRadius = 100` + `initialManeuverBufferWhenReroutingFrom = nil` -> radius of 100m will be calculated.

Thus, same delegate return value "seem" to have different result.

Using `0` as a value to disable the feature also does not align with current implementation of `RouteOption`'s equivalent.
Given these mismatches I see it is confusing to try to use bare optional. IMHO an enum with explicit and documented options is a good choice to clarify user's intent.

1ec5: > nil is used in RouteOptions to disable the feature

Technically, RouteOptions uses `nil` for the Directions API’s default behavior, which happens to be disabled: https://github.com/mapbox/mapbox-navigation-ios/pull/3454#discussion_r747159518. It’s akin to the difference between RouteOptions defaulting `includesSteps` to `false` but NavigationRouteOptions defaulting it to `true`, because _navigation_ route options are more nuanced. Doesn’t the “when rerouting” in `initialManeuverBufferWhenReroutingFrom` give us enough leeway to introduce different semantics?

Udumft: I've uploaded implementation variant with using bare optional. Note [delegate method comment](https://github.com/mapbox/mapbox-navigation-ios/pull/3454/files#diff-6d9228a7ef618b724af839a609c163b32a2f6e9ec90a797ce55d881ac60b9540R48) is now responsible for explaining it's return value.
I still believe that `enum` would work much better to cover that use case.

Udumft:
PR updated, @mapbox/navigation-ios , please give it another review round

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -351,6 +364,20 @@ extension InternalRouter where Self: Router {
...
         }
+        
+        switch delegate?.router(self, requestSourceForReroutingWith: options) {
+        case .default, .none:
+            routeTask = routingProvider.calculateRoutes(options: options) {(session, result) in
+                parseResult(session, result)
+                self.routeTask = nil
+            }
+        case .custom(let reroutingResult):
+            let (options, result) = reroutingResult()
```
chezzdev, line 375: Are we expecting that the user will alter the options for his request?

Sources/MapboxCoreNavigation/RouterDelegate.swift
```diff
@@ -1,6 +1,46 @@
 import Foundation
 import CoreLocation
 import MapboxDirections
+import Turf
+
+/**
+ Configuration for offsetting first route maneuver during rerouting.
+ */
+public enum ReroutingManeuverBuffer {
```
chezzdev, line 9: Is this enum is still used?

Sources/MapboxCoreNavigation/NavigationServiceDelegate.swift
```diff
@@ -33,13 +34,42 @@ public protocol NavigationServiceDelegate: AnyObject, UnimplementedLogging {
...
+     
+     - note: Multiple method calls will not interrupt the first ongoing request.
+     
+     This method is called after `navigationService(_:initialManeuverBufferWhenReroutingFrom:)` is called, and before `navigationService(_:didRerouteAlong:)` is called.
+     
+     - parameter service: The navigation service that will calculate a new route.
+     - parameter location: The user’s current location.
+     - returns: `.default` to let the SDK handle retrieving new `Route`, or `.custom` to provide your own reroute.
+     */
+    func navigationService(_ service: NavigationService, requestSourceForReroutingWith options: RouteOptions) -> ReroutingRequest
```
chezzdev, line 71: I'm not really getting from the signature, what "source" is.

1ec5: That must be a reference to `RouteLeg.source` etc. MapboxDirections long ago copied the MapKit terminology for waypoints, but “origin” probably would’ve been more intuitive at the time. The alternative would be to call it the “location for rerouting”.

Sources/MapboxCoreNavigation/RouterDelegate.swift
Udumft: Forgot to remove it, thanks

Sources/MapboxCoreNavigation/RouterDelegate.swift
```diff
@@ -1,6 +1,46 @@
...
+     
+     Equivalent to setting `RouteOptions.initialManeuverAvoidanceRadius` to the same radius at the beginning.
+     */
+    case radius(LocationDistance)
+}
+
+/**
+ Configuration for a source to get new `Route` when rerouting occurs.
+ */
+public enum ReroutingRequest {
```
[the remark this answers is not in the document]
Udumft: Well indeed, naming it like "behavior" makes sense and also helps naming corresponding delegate method. Thanks for suggestion.

Sources/MapboxCoreNavigation/Router.swift
Udumft: Since user is able to provide his own response, it seems logical to give him control over altering options as well.

chezzdev (approved):
Thanks for swiftly addressing feedback.

PR #3466, opened 2021-10-11
Improve vanishing route line test.

### LINKED ISSUE #3365: Cover the route line styling with automated test.
This ticket should provide the automated test for route line styling effect, including the congestion level displaying in alternative route, the step/fading gradient color on route lines.

### LINKED ISSUE #3607: [Bug]: CarPlay couldn't have soft color gradient

2.0.0-

### Steps to reproduce

1. `carPlayManager.carPlayNavigationViewController?.navigationMapView?.crossfadesCongestionSegments = true`
2. Build `Example-CarPlay` and start navigation

### Expected behavior

The route line gradient shows soft change.

### Actual behavior

The route line gradient transition is sharp.

repeatable

ShanMa1991:
### Description
This pr is to fix #3365 and #3607, and also to combine the segments with the different congestion level but same color, to avoid extra gradient stops.

### Implementation
This pr is adding tests to cover the following functions:
- [x] the basic projected distance calculation, 
- [x] the one coordinates of last step of each route leg leading index issue
- [x] the turn on/off of the `routeLineTracksTraversal` during navigation, 
- [x] the turn on/off of the `crossfadesCongestionSegments` during navigation, 
- [x] the distance threshold to update the route line, 
- [x] the initializing of stored properties of route line and routeProgress update. 
- [x] when there's no congestion level information in the route leg, the `trafficUnknownColor` would be applied to the route line.

This pr also fixes #3607 by allowing the navigationMapView regenerate the gradient stops or route line when `navigationMapView.crossfadesCongestionSegments` turns on/off during active navigation. When `routeLineTracksTraversal` enabled, the change of `navigationMapView.crossfadesCongestionSegments` will generate a new line gradient stops. When the next location update comes in, the line gradient will be updated. When `routeLineTracksTraversal` disabled, the change of `navigationMapView.crossfadesCongestionSegments` will generate a new route line.

### Screenshots or Gifs
After the fix, when enable `carPlayManager.carPlayNavigationViewController?.navigationMapView?.crossfadesCongestionSegments` as `true`:

carPlaySoft

During the active navigation, uses the resume button to switch on and off `navigationMapView?.crossfadesCongestionSegments`  , could see the color gradient transition change
switchResume


ShanMa1991:
Right now the calculation of `routeRemainingDistancesIndex` is blocking the #3319.
To make sure the `routeRemainingDistancesIndex` is correctly calculated and could pass the parity check with Android, the implementation of the `LineString.ineSliceAlong(startDist: stopDist:)` in [Turf-swift Pr](https://github.com/mapbox/turf-swift/pull/166) is needed.

ShanMa1991:
Added the `LineString.ineSliceAlong(startDist: stopDist:)` in the extension of `MapboxCoreNavigation` to support the calculation of `NavigationMapView. routeRemainingDistancesIndex`.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -238,15 +232,20 @@ extension NavigationMapView {
...
             var minimumSegment: (Double, UIColor) = isMain ? (0.0, .trafficUnknown) : (0.0, .alternativeTrafficUnknown)
 
             for (index, feature) in congestionFeatures.enumerated() {
                 var associatedFeatureColor = routeCasingColor
-                if case let .string(congestionLevel) = feature.properties?[CongestionAttribute],
-                   case let .boolean(isCurrentLeg) = feature.properties?[CurrentLegAttribute],
+                if case let .boolean(isCurrentLeg) = feature.properties?[CurrentLegAttribute],
                    isCurrentLeg {
-                    associatedFeatureColor = congestionColor(for: congestionLevel, isMain: isMain)
+                    if case let .string(congestionLevel) = feature.properties?[CongestionAttribute] {
```
ShanMa1991, line 244: When there's no congestion level information in the feature, we should default apply the `trafficUnknownColor` to the route line.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -238,15 +232,20 @@ extension NavigationMapView {
                     return nil
                 }
             }.reduce(0, +)
+            // lastRecordSegment records the last segmentEndPercentTraveled and associated congestion color added to the gradientStops.
+            var lastRecordSegment: (Double, UIColor) = (0.0, traversedRouteColor)
```
ShanMa1991, line 236: `lastRecordSegment` is used to record the last congestion color added to the `gradientStops`. If the current congestion color is the same as previous one, they will be combined.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -35,15 +35,20 @@ extension NavigationMapView {
...
+            let endIndex = routeLeg.steps.count - 1
+            for (index, routeStep) in routeLeg.steps.enumerated() {
                 if let routeShape = routeStep.shape {
-                    if !routeShape.coordinates.isEmpty {
-                        return routeShape.coordinates
-                    } else { return [] }
+                    if index != endIndex {
+                        legCoordinates.append(routeShape.coordinates)
+                    } else if let finalStep = routeShape.coordinates.first {
+                        legCoordinates.append([finalStep])
```
1ec5, line 45: It would be clearer to map the steps to coordinates arrays, as before, but then drop the last coordinate if it’s a duplicate, or perhaps only do that for an arrival step.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
```diff
@@ -0,0 +1,249 @@
...
+    func getRouteProgress() -> RouteProgress {
+        let route = getRoute()
+        let routeProgress = RouteProgress(route: route, options: routeOptions, legIndex: 0, spokenInstructionIndex: 0)
+        routeProgress.currentLegProgress = RouteLegProgress(leg: route.legs[0], stepIndex: 2, spokenInstructionIndex: 0)
+        routeProgress.currentLegProgress.currentStepProgress = RouteStepProgress(step: route.legs[0].steps[2], spokenInstructionIndex: 0)
+        return routeProgress
+    }
+    
+    func linaGradientToString(lineGradient: Value<StyleColor>?) -> String {
+        return lineGradient.debugDescription.components(separatedBy: "(").last!.components(separatedBy: ")").first!
```
MaximAlien, line 66: Nit: even tough using force unwrapping in tests can be sometimes useful to see errors early, while running in Xcode it leads to undesired stops while running all tests. I'd recommend to remove force unwrapping and add assert, in case of string is invalid.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
```diff
@@ -0,0 +1,249 @@
...
+    func testRouteLineGradientWithCombinedColor() {
+        let route = getRoute()
+        
+        navigationMapView.trafficModerateColor = navigationMapView.trafficUnknownColor
+        navigationMapView.routes = [route]
+        navigationMapView.routeLineTracksTraversal = true
+        navigationMapView.show([route], legIndex: 0)
+        
+        let expectedGradientStops = [0.0 : navigationMapView.trafficUnknownColor]
+        XCTAssertEqual(expectedGradientStops, navigationMapView.currentLineGradientStops, "Failed to combine the same color of congestion sgement.")
```
MaximAlien, line 186:
```suggestion
        XCTAssertEqual(expectedGradientStops, navigationMapView.currentLineGradientStops, "Failed to combine the same color of congestion segment.")
```

Tests/MapboxCoreNavigationTests/Extensions/LineStringTest.swift
```diff
@@ -0,0 +1,66 @@
+import XCTest
+import Turf
+import CoreLocation
+import TestHelper
+@testable import MapboxCoreNavigation
+
+class LineStringTests: TestCase {
```
MaximAlien, line 7: Should `LineStringTests` be part of Turf instead?

Sources/MapboxCoreNavigation/LineString.swift
```diff
@@ -0,0 +1,60 @@
+import Turf
+
+extension LineString {
+    /**
+     Returns the portion of the line string that begins at the given start distance and extends the given stop distance along the line string.
+     
+     This method is roughly equivalent to the [turf-line-slice-along](https://turfjs.org/docs/#lineSliceAlong) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/)).
+     */
+    public func lineSliceAlong(startDistance: LocationDistance, stopDistance: LocationDistance) -> LineString? {
```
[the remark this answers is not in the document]
ShanMa1991: Yes, thanks for the suggestion, I'll rebase it and remove this part after we adopted the latest release of [turf-swift](https://github.com/mapbox/turf-swift) that include this method.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -59,25 +64,29 @@ extension NavigationMapView {
...
         
         /**
          Add to the count of remaining points all of the remaining legs.
          */
-        for index in stride(from: routeProgress.legIndex + 1, to: completeRoutePoints.nestedList.count, by: 1) {
+        let startIndex = routeProgress.legIndex + 1
+        let endIndex = completeRoutePoints.nestedList.endIndex
+        for index in startIndex..<endIndex {
             allRemainingPoints += completeRoutePoints.nestedList[index].flatMap{ $0 }.count
         }
```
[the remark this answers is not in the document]
ShanMa1991: Thanks for the suggestion, I'll update this part to remove the for loop.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: Thanks for the suggestion, I'll update this part.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
ShanMa1991: Thanks for the suggestion, I'll add the assertion for helping methods in tests.

Tests/MapboxCoreNavigationTests/Extensions/LineStringTest.swift
ShanMa1991: Yes, I'll rebase the pr and remove this part after we adopted the latest release of turf-swift that include this method.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -37,9 +37,7 @@ extension NavigationMapView {
         let nestedList = route.legs.map { (routeLeg: RouteLeg) -> [[CLLocationCoordinate2D]] in
             return routeLeg.steps.map { (routeStep: RouteStep) -> [CLLocationCoordinate2D] in
                 if let routeShape = routeStep.shape {
-                    if !routeShape.coordinates.isEmpty {
-                        return routeShape.coordinates
-                    } else { return [] }
+                    return routeShape.coordinates
```
ShanMa1991, line 40: Even `mapbox-directions-swift` parses the route with one more duplicate coordinate in the last step of each route leg, we don't want to drop it here. Because in the `navigationMapView.updateUpcomingRoutePointIndex(routeProgress:)`, we still have to deal with the coordinate of the route step. In this case, we still store them all to be synced with the `routeProgress` that comes in later.

MaximAlien:
I've noticed that whenever we drive through the tunnel and when style changes route line will blink without taking into account vanishing point and current progress, for example:

https://user-images.githubusercontent.com/1496498/142293758-e1254001-a0af-4e1f-bd81-0dc759d480e5.mov

We should fix this issue as a tail-work.

MaximAlien:
Can you make sure that `crossfadesCongestionSegments` option works as expected on CarPlay as well? It doesn't seem to work now when we set:
```swift
carPlayManager.carPlayNavigationViewController?.navigationMapView?.crossfadesCongestionSegments = true
```

For example

Screen Shot 2021-11-17 at 2 41 17 PM
:

ShanMa1991:
> Can you make sure that `crossfadesCongestionSegments` option works as expected on CarPlay as well? It doesn't seem to work now when we set:

Yes, sure, in fact we should also re-generate the gradient stops when this value change and display it.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -59,26 +57,30 @@ extension NavigationMapView {
...
+        let currentLegSteps = completeRoutePoints.nestedList[routeProgress.legIndex]
+        var allRemainingPoints = 0
         /**
          Find the count of remaining points in the current step.
          */
-        var allRemainingPoints = getSlicedLinePointsCount(currentLegProgress: currentLegProgress, currentStepProgress: currentStepProgress)
+        let lineString = currentStepProgress.step.shape ?? LineString([])
+        if currentStepProgress.distanceTraveled < 0 {
+            allRemainingPoints += currentLegSteps[currentLegProgress.stepIndex].count
+        } else if let slicedLineString = lineString.trimmed(from: currentStepProgress.distanceTraveled, to: currentStepProgress.step.distance) {
```
MaximAlien, line 68: If I understand correctly `LineString.trimmed(from:to:)` was added in Turf `v2.1.0`, but we did not update Mapbox Directions to it. Should we create new version of Mapbox Directions `v2.1.0-rc.2`, which will use Turf `v2.1.0`? It'll allow to drop explicit Turf usage in Mapbox Navigation.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
```diff
@@ -0,0 +1,262 @@
...
+    
+    func testUpdateUpcomingRoutePointIndex() {
+        // https://github.com/mapbox/mapbox-navigation-android/blob/0ca183f7cb7bec930521ea9bcd59d0e8e2bef165/libnavui-maps/src/test/java/com/mapbox/navigation/ui/maps/route/line/api/MapboxRouteLineApiTest.kt#L802
+        let route = getRoute()
+        
+        navigationMapView.initPrimaryRoutePoints(route: route)
+        navigationMapView.routeLineGranularDistances = nil
+        XCTAssertEqual(navigationMapView.fractionTraveled, 0.0)
+        
+        let routeProgress = getRouteProgress()
```
MaximAlien, line 106: It is unclear what this method returns, can you please improve its signature or provide documentation comment? E.g. why stepIndex is set to `2` specifically etc.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
```diff
@@ -0,0 +1,262 @@
...
+    func testUpdateRouteLineWithDifferentDistance() {
+        let route = getRoute()
+        let routeProgress = getRouteProgress()
+        let coordinate = route.shape!.coordinates[1]
+        
+        navigationMapView.routes = [route]
+        navigationMapView.routeLineTracksTraversal = true
+        navigationMapView.show([route], legIndex: 0)
+        navigationMapView.updateUpcomingRoutePointIndex(routeProgress: routeProgress)
+        setUpCameraZoom(at: 5.0)
```
MaximAlien, line 146: `testUpdateRouteLineWithDifferentDistance` passes even without setting zoom level, do we need it here? I'd also add brief comment on what exactly you do in each test. At this point it's hard to follow some of them.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
```diff
@@ -0,0 +1,262 @@
...
+        setUpCameraZoom(at: 16.0)
+        
+        navigationMapView.routes = [route]
+        navigationMapView.routeLineTracksTraversal = true
+        navigationMapView.show([route], legIndex: 0)
+        navigationMapView.updateUpcomingRoutePointIndex(routeProgress: routeProgress)
+        navigationMapView.travelAlongRouteLine(to: coordinate)
+
+        let fractionTrvaled = navigationMapView.fractionTraveled
+        let fractionTrvaledNextDown = Double(CGFloat(fractionTrvaled).nextDown)
```
MaximAlien, line 218:
```suggestion
        let fractionTraveled = navigationMapView.fractionTraveled
        let fractionTraveledNextDown = Double(CGFloat(fractionTraveled).nextDown)
```

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
```diff
@@ -0,0 +1,262 @@
+import XCTest
+import MapboxDirections
+import TestHelper
+import Turf
+import MapboxMaps
+@testable import MapboxNavigation
+@testable import MapboxCoreNavigation
```
MaximAlien, line 7: There is no need to directly use Turf:
```suggestion
import XCTest
import MapboxDirections
import TestHelper
import MapboxMaps
@testable import MapboxNavigation
@testable import MapboxCoreNavigation
```

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: Because right now `MapboxDirections` is not using the new API of `LineString.trimmed(from:to:)` yet of the Turf `v2.1.0`, I'm not sure if we need to upgrade `MapboxDirections` to include this one.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
ShanMa1991: Thanks for the suggestion, I'll add the comment and documentation for each test. The default zoom level is smaller than `1` which could pass the test.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
ShanMa1991: Thanks for the suggestion, the extra directly importing is removed.

Tests/MapboxNavigationTests/VanishingRouteLineTests.swift
ShanMa1991: Thanks for the suggestion, comments are added in each unit test for documentation.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
1ec5: It’s OK for MapboxNavigation to explicitly depend on Turf if it has a more specific need than MapboxDirections.

1ec5 (approved):
Remember to describe the changes that fixed #3607 in the changelog.

PR #3472, opened 2021-10-13
Custom Rerouting

### LINKED ISSUE #3464: Allow NavigationServiceDelegate to supply new route when rerouting
NavigationServiceDelegate should have an optional method `navigationService(_:willRerouteFrom:completion:)` that accepts a callback that the method implementation should call once a new route has been calculated asynchronously. This will enable applications that use the Map Matching API or a custom routing engine (“Bring Your Own Route”) to more seamlessly reroute the user when the user goes off-route. Currently, the application is expected to implement `navigationService(_:shouldRerouteFrom:)` by returning false and asynchronously calling `Router.updateNavigator(with:)`, but this part of the workflow is unintuitive and requires reaching into the Router.

/cc @mapbox/navigation-ios

Udumft:
Resolves #3464 

This PR modifies `RouterDelegate.router(_:willRerouteFrom:)` method to allow user to select if he want's to customize the output route.

### DIFF
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -639,7 +639,7 @@ extension RouteController: Router {
             }
         }
         
-        announceImpendingReroute(at: location)
+        let rerouteRequest = announceImpendingReroute(at: location)
         
         self.lastRerouteLocation = location
         
@@ -647,7 +647,7 @@ extension RouteController: Router {
         if isRerouting { return }
         isRerouting = true
         
-        calculateRoutes(from: location, along: progress) { [weak self] (session, result) in
+        let calculationCompletion: IndexedRouteCompletionHandler = { [weak self] (session, result) in
             guard let self = self else { return }
 
             switch result {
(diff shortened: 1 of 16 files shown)


Udumft:
Originally it was labeled for 2.1, but since there are some breaking changes, could we move it to 2.2 maybe?

MaximAlien:
@Udumft, could you please rebase this PR onto `main` as there are few conflicts now? Dependencies you're using here were updated as well in main directly in https://github.com/mapbox/mapbox-navigation-ios/pull/3590.

MaximAlien:
Moved to `v2.1-rc` for now as `v2.1-beta` is no longer relevant.

Udumft:
Since this PR introduces a breaking change while we are at RC stage, I switch it to target 2.2 version

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -668,6 +668,30 @@ extension RouteController: Router {
...
+                let (options, result) = reroutingResult()
+                let session = Directions.Session(options: options,
+                                                 credentials: NavigationSettings.shared.directions.credentials)
+                
+                DispatchQueue.main.async {
+                    switch result {
+                    case .failure(let error):
+                        return calculationCompletion(session, .failure(error))
+                    case .success(let response):
+                        guard let mostSimilarIndex = response.routes?.index(mostSimilarTo: progress.route) else {
```
azarovalex, line 686: NCS pointed out that this algorithm can sometimes select the route with the biggest ETA, should we somehow allow them to select one of the routes using their own algorithm?
Currently, it's a little bit misleading that we call `willReroute` callback not before the reroute itself, but before we start to fetch possible reroutes.
Maybe we should rename `willReroute` to `willFetchReroutes` and add a new `willReroute` callback where an application can select the best route.🤔

Udumft: A bit ugly, but user can achieve this if they just remove "slower" routes from `reroutingResult`.

We could (and actually will) revisit rerouting logic once we incorporate NavNative's rerouting implementation.

azarovalex (approved):
lgtm

PR #3477, opened 2021-10-13
Use TestCase super class and correctly call super.setup/teardown

S2Ler:


S2Ler:
> Did you notice any issues related to this misuse?

I tried to run test cases one by one and observed some crashes, so I went and fixed them + make it consistent in all Test Cases.

PR #3496, opened 2021-10-18
Update default attributes per profile

### LINKED ISSUE #3495: NavigationRouteOptions causes Directions API error in cycling profile
When requesting a route from the Directions API using `DirectionsProfileIdentifier.cycling`, the Directions API returns an error about an unrecognized `attributes` value of `congestion_numeric`.

#3325 replaced usage of `AttributeOptions.congestionLevel` with `AttributeOptions.numericCongestionLevel` in NavigationRouteOptions. However, the Directions API’s `cycling` profile still uses `congestionLevel` and returns an error if `numericCongestionLevel` is passed in. Until the `cycling` profile adds support for numeric congestion levels, we need to special case it to request `congestionLevel` instead of `numericCongestionLevel`.

https://github.com/mapbox/mapbox-navigation-ios/blob/060ab11e372519d37357281a8e51c3f3f8be5ab9/Sources/MapboxCoreNavigation/NavigationRouteOptions.swift#L26-L30

/cc @mapbox/navigation-ios @LukasPaczos

1ec5:
Reverted to non-numeric congestion levels for cycling profile while the profile is powered by OSRM. Request speed limits for the walking profile too, because it no longer triggers an error. (Speed limits along walkways are rare, but if the route happens to traverse the roadway itself, then it might be interesting to know how fast motor vehicles can whiz by the user.)

Fixes #3495. Cherry-pick to release-v2.0 after merging.

/cc @mapbox/navigation-ios


1ec5:
Cherry-picked to the release-v2.0 branch in 6cd54765924c24f921bb86a71c5adc6a50c552db.

PR #3503, opened 2021-10-21
Fix guidance cards

S2Ler:
- 20166bb106cad4dec286ae96407d71e8bb68a9a8 Adjust Night Theme appearance overrides for the views that are used in Guidance Cards so that they will use appropriate color instead of defaulting to Light Theme.
- 9834d5bb36d510bc85969b6ea0875d814cebda7d Improve Instructions cards updates: Without the fix InstructionsCardContainerView.commonInit is called on each progress update due to `InstructionsCardViewController.updateVisibleInstructionCards` method. It results in duplicated constraints being created. This commit separates the code that should be called only once from the code that can be called on each update.
- 172a505bd790d4bdff636e1a89179afd19ff2732 Remove redundant forced layout: It is generally not recommended to force layout unless it 100% needed like in case when you want to animate constraints change and even in this case animation might be not perfect.

Day Style Night Style

Night Style - Highlighted Day Style - Highlighted


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 * MapboxCoreNavigation now depends on [MapboxNavigationNative v78._x_](https://github.com/mapbox/mapbox-navigation-native-ios/releases/tag/78.0.0). ([#3488](https://github.com/mapbox/mapbox-navigation-ios/pull/3488))
 
+### Banners and guidance instructions
+
+* Fixed Night Style for guidance instructions. ([#3503](https://github.com/mapbox/mapbox-navigation-ios/pull/3503))
```
1ec5, line 11: Can you explain in a sentence or two what got fixed, both here and in the PR description? Thanks!

1ec5: The second point might be worth calling out if we think the redundant constraints were causing a performance issue:

```suggestion
* `InstructionsCardViewController` now adapts to `NightStyle`. ([#3503](https://github.com/mapbox/mapbox-navigation-ios/pull/3503))
* Fixed an issue where `InstructionsCardViewController` installed duplicate Auto Layout constraints on each location update. ([#3503](https://github.com/mapbox/mapbox-navigation-ios/pull/3503))
```

1ec5 (approved):

Issue #3480, opened 2021-10-14
Show current road name label on CarPlay


1ec5:
The current road name and route shield should be displayed in a label toward the bottom of the map during the `browsing` and `navigating` activities on CarPlay, similar to the display in NavigationViewController.

The NavigationMapView inside NavigationViewController embeds a WayNameView displaying the current road name and route shield. CarPlayMapViewController and CarPlayNavigationViewController should do the same. SpeedLimitView is similarly embedded inside these view controllers’ root views before the view hierarchy gets flattened for CarPlay. We‘ll need to make sure the map viewport has enough bottom padding to keep the puck from underlapping the current road name label.

This would be a nice improvement to make in tandem with https://github.com/mapbox/mapbox-navigation-ios/issues/2548#issuecomment-942955616, though it isn’t a prerequisite for this enhancement.

/cc @mapbox/navigation-ios

1ec5:
During the `navigating` activity, we could position the current road name label in the usual position, but that would require it to avoid underlapping the trip summary panel, which might be difficult to do reliably. Alternatively, we could center it directly above the trip summary panel, assuming the maneuver panel can’t grow enough to touch the trip summary panel in any of the standard CarPlay screen aspect ratios.

Issue #2548, opened 2020-08-18
Expose current way name during active navigation


avi-c:
The Nav SDK Example app has client side code in RouteMapViewController that utilizes map feature querying to deduce the current road name that the user is driving on. At present any Nav SDK customer needs to duplicate this code or invent another approach in order to calculate the current road name.

This should be something the SDK provides through the delegate pattern used for other nav updates.

Providing this as an abstraction through the SDK also allows us to swap out the implementation if we want to use a new feature of Nav Native or find a more performant way to calculate the road name.

avi-c:
The newly implemented PassiveLocationManager provides this feature during free drive but we still don't have it during active nav.

1ec5:
> The Nav SDK Example app has client side code in RouteMapViewController that utilizes map feature querying to deduce the current road name that the user is driving on.

Here’s the implementation. It’s pretty involved.

https://github.com/mapbox/mapbox-navigation-ios/blob/8687f332ab21c7d6c7471a635f11800838c0bb8b/MapboxNavigation/RouteMapViewController.swift#L641-L751

Because it’s part of RouteMapViewController, an application that implements a custom UI is unable to take advantage of this implementation.

This implementation is basically a giant workaround for the lack of reliable road name information in the Directions API response. (The response tells us the name of the road at the beginning of each step but doesn’t reflect any name changes between steps.) Even so, we should be able to expose a generic enough interface that we could swap in a more reliable source of road names in the future without breaking backwards compatibility.

> This should be something the SDK provides through the delegate pattern used for other nav updates.

There is already a `NavigationViewControllerDelegate.navigationViewController(_:roadNameAt:)` method for the application to _provide_ a road name to label, but not one to tell the application what we think is the current road name. In any case, I think this piece of information would fit better in RouteStepProgress. PassiveLocationDataSource exposes it as part of a notification because there isn’t an analogous structure as part of passive navigation.

> The newly implemented PassiveLocationManager provides this feature during free drive but we still don't have it during active nav.

This reflects the fact that, unfortunately, MapboxNavigationNative only provides this information during free driving. If the navigator were to provide this information when a route is set, then we could stash the road name in RouteStepProgress and there would be no need for feature querying.

/cc @mapbox/navnative

1ec5:
>> The newly implemented PassiveLocationManager provides this feature during free drive but we still don't have it during active nav.
> 
[...]

PR #3527, opened 2021-10-27
Road name labeling functionality for CarPlay.

MaximAlien:
Closing #3480.

Road name labeling on CarPlay during free-drive:

Screen Shot 2021-10-27 at 3 06 07 PM

Screen Shot 2021-10-27 at 3 06 14 PM

Road name labeling on CarPlay during active-guidance:

Screen Shot 2021-10-27 at 5 32 27 PM

Screen Shot 2021-10-27 at 5 32 37 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
 * Fixed an issue where `VisualInstructionDelegate.label(_:willPresent:as:)` was never called. Your `NavigationViewControllerDelegate` class can now implement this method to customize the contents of a visual instruction during turn-by-turn navigation. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
 * Fixed an issue where certain dual- or triple-use lanes were blank in the tertiary instruction banner. ([#3569](https://github.com/mapbox/mapbox-navigation-ios/pull/3569), [mapbox/navigation-ui-resources#26](https://github.com/mapbox/navigation-ui-resources/pull/26))
 * Fixed an issue where dual-use slight turn lanes were sometimes depicted as normal turn lanes in the tertiary instruction banner. ([#3569](https://github.com/mapbox/mapbox-navigation-ios/pull/3569), [mapbox/navigation-ui-resources#26](https://github.com/mapbox/navigation-ui-resources/pull/26))
+* Setting the `WayNameView.isHidden` property to `true` now keeps the view hidden even after the user goes onto a named road. ([#3534](https://github.com/mapbox/mapbox-navigation-ios/pull/3534))
+* Added the `WayNameView.backgroundColor` and `WayNameView.borderWidth` properties for customizing how the current road name is labeled. ([#3534](https://github.com/mapbox/mapbox-navigation-ios/pull/3534))
 
 ### Map
 
@@ -43,6 +45,7 @@
 * Added the `CarPlayActivity.panningInNavigationMode` case, which allows to track a state when user is panning a map view while actively navigating. ([#3545](https://github.com/mapbox/mapbox-navigation-ios/pull/3545))
 * Fixed an issue that caused the panning dismissal button to stop working on CarPlay. ([#3543](https://github.com/mapbox/mapbox-navigation-ios/pull/3543))
 * Fixed an issue which caused the inability to see `SpeedLimitView` and `CarPlayCompassView` when left-hand traffic mode is used on CarPlay. ([#3583](https://github.com/mapbox/mapbox-navigation-ios/pull/3583))
+* Added the `CarPlayMapViewController.wayNameView` and `CarPlayNavigationViewController.wayNameView` properties to show the current road name on CarPlay. `CarPlayNavigationViewController.compassView`, `CarPlayNavigationViewController.speedLimitView` and `CarPlayMapViewController.speedLimitView` are kept as strong references, thus available throughout the lifetime of a parent object. ([#3534](https://github.com/mapbox/mapbox-navigation-ios/pull/3534))
 
 ### Other changes
 
--- a/Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
+++ b/Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
@@ -52,7 +52,7 @@ extension NavigationMapView {
             streetLabelLayer.lineWidth = .constant(20.0)
             streetLabelLayer.lineColor = .constant(.init(.white))
             
-            if ![ProfileIdentifier.walking, DirectionsProfileIdentifier.cycling].contains(router.routeProgress.routeOptions.profileIdentifier) {
+            if ![ProfileIdentifier.walking, ProfileIdentifier.cycling].contains(router.routeProgress.routeOptions.profileIdentifier) {
                 // Filter out to road classes valid only for motor transport.
                 let filter = Exp(.inExpression) {
                     "class"
--- a/Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
+++ b/Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
@@ -1,0 +1,171 @@
+import MapboxMaps
+import MapboxDirections
+import MapboxCoreNavigation
+
+extension NavigationMapView {
+    
+    func labelCurrentRoadFeature(at location: CLLocation,
+                                 router: Router,
+                                 wayNameView: WayNameView,
+                                 roadNameFromStatus: String?) {
+        guard let stepShape = router.routeProgress.currentLegProgress.currentStep.shape,
+              !stepShape.coordinates.isEmpty else {
+                  return
+              }
+        
+        // Add Mapbox Streets if the map does not already have it
+        if mapView.streetsSources().isEmpty {
+            var streetsSource = VectorSource()
+            streetsSource.url = "mapbox://mapbox.mapbox-streets-v8"
+            
+            let sourceIdentifier = "com.mapbox.MapboxStreets"
+            
+            do {
+                try mapView.mapboxMap.style.addSource(streetsSource, id: sourceIdentifier)
+            } catch {
+                NSLog("Failed to add \(sourceIdentifier) with error: \(error.localizedDescription).")
+            }
+        }
+        
+        guard let mapboxStreetsSource = mapView.streetsSources().first else { return }
+        
+        let identifierNamespace = Bundle.mapboxNavigation.bundleIdentifier ?? ""
+        let roadLabelStyleLayerIdentifier = "\(identifierNamespace).roadLabels"
+        let roadLabelLayer = try? mapView.mapboxMap.style.layer(withId: roadLabelStyleLayerIdentifier) as? LineLayer
+        
+        if roadLabelLayer == nil {
+            var streetLabelLayer = LineLayer(id: roadLabelStyleLayerIdentifier)
+            streetLabelLayer.source = mapboxStreetsSource.id
+            
+            var sourceLayerIdentifier: String? {
+                let identifiers = mapView.tileSetIdentifiers(mapboxStreetsSource.id,
+                                                             sourceType: mapboxStreetsSource.type.rawValue)
+                if VectorSource.isMapboxStreets(identifiers) {
+                    return identifiers.compactMap({ VectorSource.roadLabelLayerIdentifiersByTileSetIdentifier[$0] }).first
+                }
+                
+                return nil
+            }
+            
+            streetLabelLayer.sourceLayer = sourceLayerIdentifier
+            streetLabelLayer.lineOpacity = .constant(1.0)
+            streetLabelLayer.lineWidth = .constant(20.0)
+            streetLabelLayer.lineColor = .constant(.init(.white))
+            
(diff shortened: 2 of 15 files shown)


Sources/MapboxNavigation/UIImage.swift
```diff
@@ -15,22 +15,35 @@ extension UIImage {
...
-    func withCenteredText(_ text: String, color: UIColor, font: UIFont, scale: CGFloat) -> UIImage {
-        let maxHeight = font.lineHeight * 2
-        var constrainedSize = size
-        constrainedSize.width = min(constrainedSize.width, constrainedSize.width * maxHeight / constrainedSize.height)
-        constrainedSize.height = min(constrainedSize.height, maxHeight)
+    func withCenteredText(_ text: String,
+                          color: UIColor,
+                          font: UIFont,
+                          size: CGSize? = nil) -> UIImage {
+        let maxHeight = size?.height ?? font.lineHeight * 2
```
MaximAlien, line 22: In case if size is set we'll use it to provide maximum height for the image. In this case this height is height of `UILabel` container.

CHANGELOG.md
```diff
@@ -21,11 +21,13 @@
...
 
 * Fixed the crash when scrolling the guidance cards while the orientaion changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
+* A `WayNameView` now can be fully hidden by using `WayNameView.isAlwaysHidden` property. ([#3534](https://github.com/mapbox/mapbox-navigation-ios/pull/3534))
 
 ### CarPlay
 
 * Added the `CarPlayActivity.panningInNavigationMode` case, which allows to track a state when user is panning a map view while actively navigating. ([#3545](https://github.com/mapbox/mapbox-navigation-ios/pull/3545))
 * Fixed an issue that caused the panning dismissal button to stop working on CarPlay. ([#3543](https://github.com/mapbox/mapbox-navigation-ios/pull/3543))
+* Added the `CarPlayMapViewController.wayNameView` and `CarPlayNavigationViewController.wayNameView` properties to show the current road name on CarPlay. `CarPlayNavigationViewController.compassView`, `CarPlayNavigationViewController.speedLimitView` and `CarPlayMapViewController.speedLimitView` are kept as strong references, thus available throughout the lifetime of a parent object. ([#3534](https://github.com/mapbox/mapbox-navigation-ios/pull/3534))
```
1ec5, line 30: The parent object in this case would be CarPlayManager, right?

Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -10,10 +10,43 @@ open class WayNameLabel: StylableLabel {}
 /// :nodoc:
```
1ec5, line 10: If we expect the developer to use this class or any of its members, we should document it. Until now, these classes have largely been used in Style subclasses in a particular way, so they didn’t require much beyond an example in practice, though documentation is long overdue.

Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -10,10 +10,43 @@ open class WayNameLabel: StylableLabel {}
...
+        }
+    }
+    
+    var _isHidden: Bool = false
+    
+    /**
+     Allows to hide or show `WayNameView`. In case if `WayNameView.isAlwaysHidden` is set to `true`
+     modification of this value will have no effect and `WayNameView` will remain hidden.
+     */
+    open override var isHidden: Bool {
```
1ec5, line 33: The separate `isHidden` and `isAlwaysHidden` properties are a bit unintuitive. `isHidden` is normally a way for a developer to hide the view without it automatically coming back. We can accomplish that by embedding another view within WayNameView that we automatically hide when there’s no road name to display. Alternatively, we should introduce a property `visibility` of type `OrnamentVisibility` (from the map SDK) that allows for an `adaptive` visibility by default.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/3534#discussion_r739372134

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
```diff
@@ -0,0 +1,171 @@
+import MapboxMaps
+import MapboxDirections
+import MapboxCoreNavigation
+
+extension NavigationMapView {
+    
+    func labelCurrentRoadFeature(at location: CLLocation,
+                                 router: Router,
+                                 wayNameView: WayNameView,
+                                 roadNameFromStatus: String?) {
```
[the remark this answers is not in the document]
1ec5: Querying on both “screens” should be fine in principle, even simultaneously, but there could be a scaling issue going on. Maybe the map SDK’s querying functionality is confused about the fact that the CarPlay map has a different scale and aspect ratio.

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
```diff
@@ -0,0 +1,171 @@
...
+            
+            do {
+                var layerPosition: MapboxMaps.LayerPosition? = nil
+                if let firstLayerIdentifier = mapView.mapboxMap.style.allLayerIdentifiers.first?.id {
+                    layerPosition = .below(firstLayerIdentifier)
+                }
+                
+                try mapView.mapboxMap.style.addLayer(streetLabelLayer, layerPosition: layerPosition)
+            } catch {
+                NSLog("Failed to add \(roadLabelStyleLayerIdentifier) with error: \(error.localizedDescription).")
```
1ec5, line 85: Replace `NSLog(_:)` with `print(_:)` I guess.

S2Ler: Or even better with OSLog.

CHANGELOG.md
MaximAlien: By parent object I mean either `CarPlayNavigationViewController` or `CarPlayMapViewController`.

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
MaximAlien: I do not think that replacing this particular part will change anything. It'd be better to handle such logs in scope of another ticket (I think we already have at least two, which should revise logging on both iOS and CarPlay).

Sources/MapboxNavigation/WayNameView.swift
MaximAlien: Created ticket for this in https://github.com/mapbox/mapbox-navigation-ios/issues/3584.

Sources/MapboxNavigation/WayNameView.swift
MaximAlien: `isAlwaysHidden` was introduced to have similarity with public property of `SpeedLimitView.isAlwaysHidden`. Not sure that usage of `OrnamentVisibility` will help us here, we'll basically use only `.adaptive` and `.hidden` cases, `.visible` will not be relevant. 

Wrapping `WayNameView` in another view, which can be ultimately hidden leaves us with breaking API changes as we currently expose `WayNameView` on iOS as well, and it'd make sense to be able to hide it as well there. I'll do some more tests though to see if we can use such approach.

MaximAlien: In the end I've decided to go with the wrapping approach. Please check the latest changes. Hopefully I did not bring any regression, but now to completely hide `WayNameView` we can use `isHidden` property.

Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -3,17 +3,32 @@ import UIKit
 import Turf
 import MapboxMaps
 
-/// :nodoc:
+/**
+ Label, which is used to show a road name and a shield icon.
```
S2Ler, line 7:
```suggestion
 A label that is used to show a road name and a shield icon.
```

S2Ler (commented):
My main concern is about duplicated code. The rest is just small comments about clarity.

Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -32,40 +47,95 @@ open class WayNameView: UIView {
...
+        set {
+            layer.borderWidth = newValue
+        }
+    }
+    
+    /**
+     Initializes a newly allocated `WayNameView` object with the specified frame rectangle.
+     
+     - parameter frame: The frame rectangle for the `WayNameView`.
+     */
```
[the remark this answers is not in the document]
1ec5: Quick Help will pick up the comment from `UIView(frame:)`, but jazzy won’t because it doesn’t have access to the Apple documentation. It won’t even link to the Apple documentation: realm/jazzy#13. That said, I don’t think this comment is adding anything that developers would need to know beyond what UIView provides.

Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -32,40 +47,95 @@ open class WayNameView: UIView {
...
         commonInit()
     }
     
-    public required init?(coder aDecoder: NSCoder) {
-        super.init(coder: aDecoder)
+    /**
+     Returns a `WayNameView` object initialized from data in a given unarchiver.
+     
+     - parameter decoder: An unarchiver object.
+     */
```
1ec5, line 109: This comment is also redundant.

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
```diff
@@ -0,0 +1,171 @@
+import MapboxMaps
+import MapboxDirections
+import MapboxCoreNavigation
+
+extension NavigationMapView {
+    
+    func labelCurrentRoadFeature(at location: CLLocation,
```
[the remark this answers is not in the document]
1ec5: #2548 tracks consolidating and exposing the logic for labeling the current road name.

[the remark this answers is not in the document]
MaximAlien: @S2Ler, `NavigationMapView.labelCurrentRoadFeature(at:router:wayNameView:roadNameFromStatus:)` is already reused by iOS and CarPlay. Can you elaborate regarding: 

> The original code already has a bug that incorrectly hides road labels.

`WayNameView` is hidden whenever road name info is missing and we were not able to query it. It'll be shown back whenever is such info is available once again.

Sources/MapboxNavigation/WayNameView.swift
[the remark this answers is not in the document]
MaximAlien: Makes sense, removed docs from default methods.

Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -204,7 +204,7 @@ public class NavigationViewportDataSource: ViewportDataSource {
             
             // Bottom of the viewport on CarPlay should be placed at the same level with
             // trip estimate view.
-            carPlayCameraPadding.bottom += 25.0
+            carPlayCameraPadding.bottom += 65.0
```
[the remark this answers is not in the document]
MaximAlien: Other properties depend on this specific property (e.g `WayNameView` position). I plan to check possible improvements here in scope if tail-work.

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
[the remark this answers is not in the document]
S2Ler: > NavigationMapView.labelCurrentRoadFeature(at:router:wayNameView:roadNameFromStatus:) is already reused by iOS and CarPlay.

I'm sorry, missed that part, that's cool that there is no duplication 👍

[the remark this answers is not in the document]
S2Ler: > `WayNameView` is hidden whenever road name info is missing and we were not able to query it. It'll be shown back whenever is such info is available once again.

I haven't got a chance to create a ticket, but on the freeways, I've noticed that road name disappears even though it shouldn't.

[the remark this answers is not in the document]
S2Ler: @MaximAlien  #3594 created a bug for this.

Sources/MapboxNavigation/WayNameView.swift
```diff
@@ -32,40 +47,82 @@ open class WayNameView: UIView {
...
     }
     
-    public required init?(coder aDecoder: NSCoder) {
-        super.init(coder: aDecoder)
+    public required init?(coder decoder: NSCoder) {
+        super.init(coder: decoder)
         commonInit()
     }
     
     func commonInit() {
```
S2Ler, line 104: not relevant to this PR, but makes sense to change access level for commmonInit to private as far as I see.

S2Ler (approved):

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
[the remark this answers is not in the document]
MaximAlien: Thanks. Since querying depends on the zoom level we'll have to find more reliable ways of getting road name information. Hopefully in scope of https://github.com/mapbox/mapbox-navigation-ios/issues/2548 we'll be able to fix that.

PR #3533, opened 2021-10-28
Stop the location update in the background during active navigation using default course view.

ShanMa1991:
### Description

This pr is to stop the location update in the background of the `mapView` during the active navigation when users using the default course view for the user location indicator layer.

Right now we have 2 types of user indicator layers, the map puck and the `.courseView`. During the active navigation, if we choose the map puck, we rely on the location update from `mapView` to update the location of user location indicator and the vanishing route line, so they could keep in sync.

But when we use the default `.courseView`, the user indicator layer and the vanishing route line are both listen to the router. If there're location updates in the background of the `mapView` during the active navigation, it could trigger the navigator to gives different location and `routeProgress`, which lead to a jumping location of user location indicator and the flickering of vanishing route line effect due to the `fractionTraveled` is calculated based on the location.

And because the map side decides to open or shut down location update based on the `puckType` , we right now have an empty `puck2D` to make sure we are able to receive location when using a standalone `navigationMapView`. But in active navigation when we choose the default `.courseView` we should stop the location update in the background like before to avoid the drifting location information.

https://github.com/mapbox/mapbox-maps-ios/blob/75f38ef5c15a7ff97bd41f4243ad343befcd3e6a/Sources/MapboxMaps/Location/LocationManager.swift#L183-L205

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * Added the `ReplayLocationManager.replayCompletionHandler` property that allows you to loop location. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528), [3550](https://github.com/mapbox/mapbox-navigation-ios/pull/3550))
 * Changed the behavior of `ReplayLocationManager` so that it doesn't loop locations by default. ([#3550](https://github.com/mapbox/mapbox-navigation-ios/pull/3550))
 * Fixed an issue where `ReplayLocationManager` didn't update location timestamps when a new loop started. ([#3550](https://github.com/mapbox/mapbox-navigation-ios/pull/3550))
+* Fixed the background location update issue during active navigation when using default `.courseView` for `NavigationMapView.userLocationStyle`. ([#3533](https://github.com/mapbox/mapbox-navigation-ios/pull/3533))
 
 ### Banners and guidance instructions
 
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -573,6 +573,11 @@ open class NavigationMapView: UIView {
             setupUserLocation()
         }
     }
+    var inActiveNavigation: Bool = false {
+        didSet {
+            setupUserLocation()
+        }
+    }
     
     /**
      Most recent user location, which is used to place `UserCourseView`.
@@ -604,7 +609,7 @@ open class NavigationMapView: UIView {
         } else {
             switch userLocationStyle {
             case .courseView(let courseView):
-                mapView.location.options.puckType = .puck2D(emptyPuckConfiguration)
+                mapView.location.options.puckType = inActiveNavigation ? nil : .puck2D(emptyPuckConfiguration)
                 
                 courseView.tag = NavigationMapView.userCourseViewTag
                 mapView.addSubview(courseView)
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Added the the `PassiveLocationManager.rawLocation` and `PassiveLocationManager.location` properties to get the latest raw and idealized locations, respectively. ([#3474](https://github.com/mapbox/mapbox-navigation-ios/pull/3474))
 * Fixed an issue where `ReplayLocationManager` would crash if initialized with just one location. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528))
 * Added the `ReplayLocationManager.onReplayLoopCompleted` property that allows you to stop the location manager from looping back to the beginning of the route. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528))
+* Fixed the route line flickering issue during active navigation when using default `.courseView` for `NavigationMapView.userLocationStyle`. ([#3533](https://github.com/mapbox/mapbox-navigation-ios/pull/3533))
```
MaximAlien, line 14: Is there an easy way to reproduce this issue?

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -233,6 +233,7 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
                               routeProgress: navigationService.routeProgress)
         }
 
+        navigationMapView?.inActiveNavigation = true
```
MaximAlien, line 236: Do we also need to add similar change for `CarPlayNavigationViewController`?

ShanMa1991: Yes, it's added in the `CarPlayNavigationViewController ` after the active navigation starts.

CHANGELOG.md
ShanMa1991: Not quite sure whether the https://github.com/mapbox/mapbox-navigation-ios/issues/3580 of the  course view drifting is related with it. But the route line flickering issue is observed by the customer during real navigation with the course view.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -1399,7 +1404,7 @@ open class NavigationMapView: UIView {
                     self.travelAlongRouteLine(to: location.coordinate)
                 }
             default:
-                if self.simulatesLocation, let locationProvider = self.mapView.location.locationProvider {
+                if self.simulatesLocation && self.inActiveNavigation, let locationProvider = self.mapView.location.locationProvider {
```
ShanMa1991, line 1407: During the free drive, we don't send the location to trigger the `locationManager` update because the map puck is already listening to the location update.

MaximAlien, line 1407: Nit, more _Swifty_ way I guess:
```suggestion
                if self.simulatesLocation, self.inActiveNavigation, let locationProvider = self.mapView.location.locationProvider {
```

MaximAlien (approved):

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion! I'll improve this part.

PR #3541, opened 2021-11-01
Fix distance measurement system

### LINKED ISSUE #2578: Navigation distance unit should default to route option's `distanceMeasurementSystem`
**Mapbox Navigation SDK version:** 0.40

# Introduction
I want to have English as language and Metric as the measurement unit: to achieve that I set route options locale to `en_US` and `distanceMeasurementSystem` to `.metric`.

Without any other modification, I get the navigation instruction in English/Metric while the UI alerts are still in Imperial.

I expect the route options to be the single source of truth (as described in #1055).

# Fix
I would expect that the `NavigationViewController` uses the route options `distanceMeasurementSystem` to set the NavigationService's distanceUnit, while it still read use locale: from `NavigationViewController.swift`

```swift
NavigationSettings.shared.distanceUnit = routeOptions.locale.usesMetric ? .kilometer : .mile
```

Shouldn't this line be something like this, or am I doing something wrong?

```swift
NavigationSettings.shared.distanceUnit = routeOptions. distanceMeasurementSystem == .metric ? .kilometer : .mile
```

I'm willing to open a PR in this direction if agreed.

# Workaround
As a workaround, I explicitly override the `NavigationSettings`'s `distanceUnit` after the creation of the NavigationViewController.

S2Ler:
# deea28cf6b6889f3794f94056207bfcad48296b5

Fixes #2578

Prefers distance measurement system from RouteOptions instead of just current locale. It allows having US locale and metric system distance. 

NavigationRouteOptions now reads global NavigationSettings for preferred distance measurement system. This fixes CarPlay implementation for route previews. 

# a6ee65d9f30f211e4b83cb90f275026187791e37

Deletes misleading property from NavigationSettings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@
 ### Other changes
 
 * Renamed the `Locale.usesMetric` property to `Locale.measuresDistancesInMetricUnits`. `Locale.usesMetric` is still available but deprecated. ([#3547](https://github.com/mapbox/mapbox-navigation-ios/pull/3547))
-* Fixed an issue where setting `DirectionsOptions.distanceMeasurementSystem` did not affect the units displayed in the user interface. ([#3541](https://github.com/mapbox/mapbox-navigation-ios/pull/3541))
+* Fixed an issue where the user interface did not necessarily display distances in the same units as the route by default. `NavigationRouteOptions` and `NavigationMatchOptions` now set `DirectionsOptions.distanceMeasurementSystem` to a default value matching the `NavigationSettings.distanceUnit` property. ([#3541](https://github.com/mapbox/mapbox-navigation-ios/pull/3541))
 
 ## v2.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 ### Other changes
 
 * Renamed the `Locale.usesMetric` property to `Locale.measuresDistancesInMetricUnits`. `Locale.usesMetric` is still available but deprecated. ([#3547](https://github.com/mapbox/mapbox-navigation-ios/pull/3547))
+* Fixed an issue where setting `DirectionsOptions.distanceMeasurementSystem` did not affect the units displayed in the user interface. ([#3541](https://github.com/mapbox/mapbox-navigation-ios/pull/3541))
 
 ## v2.0.1
 
--- a/Sources/MapboxCoreNavigation/NavigationService.swift
+++ b/Sources/MapboxCoreNavigation/NavigationService.swift
@@ -308,8 +308,8 @@ public class MapboxNavigationService: NSObject, NavigationService {
         
         let routerType = routerType ?? DefaultRouter.self
         _router = routerType.init(alongRouteAtIndex: routeIndex, in: routeResponse, options: routeOptions, directions: self.directions, dataSource: self)
-        NavigationSettings.shared.distanceUnit = routeOptions.locale.measuresDistancesInMetricUnits ? .kilometer : .mile
-        
+        NavigationSettings.shared.distanceUnit = .init(routeOptions.distanceMeasurementSystem)
+
         let eventType = eventsManagerType ?? NavigationEventsManager.self
         _eventsManager = eventType.init(activeNavigationDataSource: self,
                                         accessToken: self.directions.credentials.accessToken)
--- a/Sources/MapboxCoreNavigation/NavigationSettings.swift
+++ b/Sources/MapboxCoreNavigation/NavigationSettings.swift
@@ -122,25 +122,11 @@ public class NavigationSettings {
     
     /**
      Specifies the preferred distance measurement unit.
-     - note: Anything but `kilometer` and `mile` will fall back to the default measurement for the current locale.
-        Meters and feets will be used when the presented distances are small enough. See `DistanceFormatter` for more information.
+     - Meters and feet will be used when the presented distances are small enough. See `DistanceFormatter` for more information.
      */
     public dynamic var distanceUnit : LengthFormatter.Unit = Locale.current.measuresDistancesInMetricUnits ? .kilometer : .mile {
         didSet {
             notifyChanged(property: .distanceUnit, value: distanceUnit.rawValue)
-        }
-    }
-    
-    var usesMetric: Bool {
-        get {
-            switch distanceUnit {
-            case .kilometer:
-                return true
-            case .mile:
-                return false
-            default:
-                return Locale.current.measuresDistancesInMetricUnits
-            }
         }
     }
     
--- a/Sources/MapboxCoreNavigation/NavigationSettings.swift
+++ b/Sources/MapboxCoreNavigation/NavigationSettings.swift
@@ -122,7 +122,7 @@ public class NavigationSettings {
     
     /**
      Specifies the preferred distance measurement unit.
-     - Meters and feet will be used when the presented distances are small enough. See `DistanceFormatter` for more information.
+     Meters and feet will be used when the presented distances are small enough. See `DistanceFormatter` for more information.
      */
     public dynamic var distanceUnit : LengthFormatter.Unit = Locale.current.measuresDistancesInMetricUnits ? .kilometer : .mile {
         didSet {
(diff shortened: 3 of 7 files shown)


Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -187,3 +187,18 @@ extension String {
         return "MB" + self
     }
 }
+
+extension MeasurementSystem {
+    /// Converts `MapboxDirections.MeasurementSystem` into `LengthFormatter.Unit`.
+    public var distanceUnit: LengthFormatter.Unit {
```
S2Ler, line 193: Not sure if we need to expose this to public or hide under "spi".

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -308,8 +308,8 @@ public class MapboxNavigationService: NSObject, NavigationService {
         
         let routerType = routerType ?? DefaultRouter.self
         _router = routerType.init(alongRouteAtIndex: routeIndex, in: routeResponse, options: routeOptions, directions: self.directions, dataSource: self)
-        NavigationSettings.shared.distanceUnit = routeOptions.locale.usesMetric ? .kilometer : .mile
-        
+        NavigationSettings.shared.distanceUnit = routeOptions.distanceMeasurementSystem.distanceUnit
```
1ec5, line 311: It was possible before to display metric units with an American English or British English locale, but it required setting `NavigationSettings.distanceUnit` explicitly. This makes that explicit setting unnecessary.

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -187,3 +187,18 @@ extension String {
...
+        return self == .metric ? .kilometer : .mile
+    }
+}
+
+extension LengthFormatter.Unit {
+    /// Converts `LengthFormatter.Unit` into `MapboxDirections.MeasurementSystem`.
+    public var measurementSystem: MeasurementSystem {
```
1ec5, line 200: Replace this with a `MeasurementSystem` initializer that takes a `LengthFormatter.Unit` as an unlabeled argument (and vice versa).

Sources/MapboxCoreNavigation/NavigationSettings.swift
1ec5: `MeasurementSystem` is defined by MapboxDirections; these conversions should ultimately live there too, so I’d suggest hiding them for now using `:nodoc:`.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 * Fixed an issue where `ReplayLocationManager` would crash if initialized with just one location. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528))
 * Added the `ReplayLocationManager.onReplayLoopCompleted` property that allows you to stop the location manager from looping back to the beginning of the route. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528))
 
+### Other changes
+
+* Fixed an issue where `NavigationRouteOptions.distanceMeasurementSystem` and `NavigationMatchOptions.distanceMeasurementSystem` wasn't taken into account for formatting distance strings. ([#3541](https://github.com/mapbox/mapbox-navigation-ios/pull/3541))
```
1ec5, line 17:
```suggestion
* Fixed an issue where setting `DirectionsOptions.distanceMeasurementSystem` did not affect the units displayed in the user interface. ([#3541](https://github.com/mapbox/mapbox-navigation-ios/pull/3541))
```

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -505,6 +505,7 @@ extension CarPlayManager {
      */
     public func previewRoutes(between waypoints: [Waypoint], completionHandler: @escaping CompletionHandler) {
         let options = NavigationRouteOptions(waypoints: waypoints)
+        options.distanceMeasurementSystem = NavigationSettings.shared.distanceUnit.measurementSystem
```
1ec5, line 508: It’s actually backwards that MapboxNavigationService and NavigationViewController tell NavigationSettings what locale settings to use. NavigationSettings is a singleton intended to apply consistently across the whole application, but allowing individual instances of other classes to set this one global value risks race conditions.

The order here is more sensible. What if we go one step further and push this line into NavigationRouteOptions itself?

S2Ler: Makes total sense!

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -122,28 +122,14 @@ public class NavigationSettings {
     
     /**
      Specifies the preferred distance measurement unit.
-     - note: Anything but `kilometer` and `mile` will fall back to the default measurement for the current locale.
-        Meters and feets will be used when the presented distances are small enough. See `DistanceFormatter` for more information.
+     - Meters and feet will be used when the presented distances are small enough. See `DistanceFormatter` for more information.
```
1ec5, line 125:
```suggestion
     Meters and feet will be used when the presented distances are small enough. See `DistanceFormatter` for more information.
```

Sources/MapboxCoreNavigation/NavigationService.swift
```diff
@@ -308,8 +308,8 @@ public class MapboxNavigationService: NSObject, NavigationService {
         
         let routerType = routerType ?? DefaultRouter.self
         _router = routerType.init(alongRouteAtIndex: routeIndex, in: routeResponse, options: routeOptions, directions: self.directions, dataSource: self)
-        NavigationSettings.shared.distanceUnit = routeOptions.locale.usesMetric ? .kilometer : .mile
-        
+        NavigationSettings.shared.distanceUnit = .init(routeOptions.distanceMeasurementSystem)
```
1ec5, line 311: Is this still necessary, now that `NavigationRouteOptions.distanceMeasurementSystem` is set to match `NavigationSettings.distanceUnit` by default? If the developer goes out of their way to change `distanceMeasurementSystem` away from the default, should we keep the UI in sync, or should we allow the developer to configure voice instructions in a different unit than the UI?

1ec5 (approved):

Sources/MapboxCoreNavigation/NavigationService.swift
S2Ler: I will check the docs after this merged.

Sources/MapboxCoreNavigation/NavigationSettings.swift
```diff
@@ -187,3 +173,18 @@ extension String {
         return "MB" + self
     }
 }
+
+extension MeasurementSystem {
+    /// :nodoc: Converts `LengthFormatter.Unit` into `MapboxDirections.MeasurementSystem`.
+    public init(_ lengthUnit: LengthFormatter.Unit) {
+        let metricUnits: [LengthFormatter.Unit] = [.kilometer, .centimeter, .meter, .millimeter]
```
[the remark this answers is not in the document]
S2Ler: Always looking ahead ⏩

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -220,8 +220,8 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
                                        simulating: navigationOptions?.simulationMode)
         navigationService.delegate = self
         
-        NavigationSettings.shared.distanceUnit = routeOptions.locale.usesMetric ? .kilometer : .mile
-        
+        NavigationSettings.shared.distanceUnit = .init(routeOptions.distanceMeasurementSystem)
```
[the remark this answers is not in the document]
S2Ler: yep.

Sources/MapboxCoreNavigation/NavigationService.swift
S2Ler: https://github.com/mapbox/mapbox-navigation-ios/issues/3587

Sources/MapboxCoreNavigation/NavigationService.swift
S2Ler: https://github.com/mapbox/mapbox-navigation-ios/issues/3588

Issue #3539, opened 2021-10-29
[Bug]: Panning interface dismissal button stops working after one active-guidance navigation session.


MaximAlien:
### Mapbox Navigation SDK version

`v2.0.0`

### Steps to reproduce

1. Start active-guidance navigation session on CarPlay.
2. Stop navigation session.
3. Open panning interface.
4. Try to close panning interface.

### Expected behavior

Panning interface should be dismissed.

### Actual behavior

Panning interface cannot be dismissed. For example:

https://user-images.githubusercontent.com/1496498/139509410-03849e20-54bc-440a-a0be-eff4a6874b39.mov

### Is this a one-time issue or a repeatable issue?

repeatable

PR #3543, opened 2021-11-01
Fix issue, which was causing panning dismissal button to stop working on CarPlay.

MaximAlien:
PR fixes an issue, which was causing close panning button to stop working after several active-guidance navigation sessions. Root cause of this was that `closeButton` was always using the same `CPMapTemplate` even after several dismissals and presentations.

Closing #3539.

### DIFF
--- a/Sources/MapboxNavigation/CarPlayManager.swift
+++ b/Sources/MapboxNavigation/CarPlayManager.swift
@@ -399,15 +399,16 @@ extension CarPlayManager: CPApplicationDelegate {
         guard let carPlayMapViewController = carPlayMapViewController else {
             return nil
         }
-        var mapButtons = [
+        
+        let panMapButton = carPlayMapViewController.panningInterfaceDisplayButton(for: mapTemplate)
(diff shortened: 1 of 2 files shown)


Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -399,15 +399,16 @@ extension CarPlayManager: CPApplicationDelegate {
...
+        carPlayMapViewController.panMapButton = panMapButton
+        
+        let mapButtons = [
             carPlayMapViewController.recenterButton,
+            panMapButton,
             carPlayMapViewController.zoomInButton,
             carPlayMapViewController.zoomOutButton
         ]
-        let panMapButton = carPlayMapViewController.panMapButton ??
-            carPlayMapViewController.panningInterfaceDisplayButton(for: mapTemplate)
```
MaximAlien, line 408: Not sure why before we were using this check, doesn't seem to significantly help with performance, and is the root case of this issue.

S2Ler (approved):

PR #3544, opened 2021-11-02
Fix the scrolling guidance cards during orientation change crash

### LINKED ISSUE #3502: [Bug]: InstructionsCardViewController on orientation change

2.0.0
iOS version: 15

### Steps to reproduce

change the orientation when during the navigation using active guidance cards

### Expected behavior

Work normally

### Actual behavior

Crash with the following information:
```
Attempted to scroll the collection view to an out-of-bounds item (1) when there are only 1 items in section 0. 
```
in the following place:
https://github.com/mapbox/mapbox-navigation-ios/blob/5382dc06ef592805dc17689ace4be423aa6d8ee9/Sources/MapboxNavigation/InstructionsCardViewController.swift#L242

repeatable

ShanMa1991:
### Description
This pr is to fix #3502 

### Implementation
By adding the items count check before the` InstructionsCardViewController.handlePagingforScrollToItem(indexPath:)` would prevent the crash, which is happening of scrolling the guidance card hard during the orientation change animation.

For example, there're some cases that when the orientation change and the swiping the guidance cards hard, the remanning steps is decreasing, it would lead to the `steps.count` as `7`, while the `indexBeforeSwipe.row` as `7`, it would lead to the crash.  The items count check would prevent this rare edge case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### Banners and guidance instructions
 
-* Fixed the crash when scrolling the guidance cards while the orienttaion change during active navigation. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
+* Fixed the crash when scrolling the guidance cards while the orientaion changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
 
 ## v2.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 * Fixed an issue where `ReplayLocationManager` would crash if initialized with just one location. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528))
 * Added the `ReplayLocationManager.onReplayLoopCompleted` property that allows you to stop the location manager from looping back to the beginning of the route. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528))
 
+### Banners and guidance instructions
+
+* Fixed the crash when scrolling the guidance cards while the orienttaion change during active navigation. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
+
 ## v2.0.1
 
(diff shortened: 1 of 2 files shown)


Sources/MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -246,6 +246,7 @@ open class InstructionsCardViewController: UIViewController {
     
     @objc func orientationDidChange(_ notification: Notification) {
         instructionsCardLayout.invalidateLayout()
+        guard let itemCount = steps?.count, itemCount >= 0 && indexBeforeSwipe.row < itemCount else { return }
```
S2Ler, line 249: nit: itemCount cannot be less that zero, just a redundant check.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 * Fixed an issue where `ReplayLocationManager` would crash if initialized with just one location. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528))
 * Added the `ReplayLocationManager.onReplayLoopCompleted` property that allows you to stop the location manager from looping back to the beginning of the route. ([#3528](https://github.com/mapbox/mapbox-navigation-ios/pull/3528))
 
+### Banners and guidance instructions
+
+* Fixed the crash when scrolling the guidance cards while the orienttaion change during active navigation. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
```
S2Ler, line 17: Typo in "orienttaion". 

nit: This changelog entry is hard to read, I would suggest to simplify it.

S2Ler (approved):

Sources/MapboxNavigation/InstructionsCardViewController.swift
bamx23, line 249: There is the same check in the `snapToIndexPath` method that is the only other method that uses `handlePagingforScrollToItem` so it should be better to move this validation to `handlePagingforScrollToItem`.

ShanMa1991: Thanks for the suggestion, I'll move it to the `handlePagingforScrollToItem`.

PR #383, opened 2017-07-13
Use fractions for voice only

frederoni:
Fixes #375 - Use fractions for voice

@1ec5 @bsudekum 👀

PR #3547, opened 2021-11-02
Deprecate custom usesMetric in favor of iOS defined usesMetricSystem

S2Ler:
iOS has standard `usesMetricSystem` variable that [works differently](https://github.com/mapbox/mapbox-navigation-ios/pull/3547#discussion_r741311298) for UK locale

### DIFF
--- a/Sources/MapboxCoreNavigation/Locale.swift
+++ b/Sources/MapboxCoreNavigation/Locale.swift
@@ -24,15 +24,17 @@ extension Locale {
      */
     public static var nationalizedCurrent = Locale(identifier: preferredLocalLanguageCountryCode)
     
+    @available(*, deprecated, message: "Use Locale.current.measuresDistancesInMetricUnits instead")
     public static var usesMetric: Bool {
-        let locale = self.current as NSLocale
-        guard let measurementSystem = locale.object(forKey: .measurementSystem) as? String else {
-            return false
-        }
-        return measurementSystem == "Metric"
+        Locale.current.measuresDistancesInMetricUnits
     }
-    
+
+    @available(*, deprecated, renamed: "measuresDistancesInMetricUnits")
     public var usesMetric: Bool {
+        measuresDistancesInMetricUnits
+    }
+
+    public var measuresDistancesInMetricUnits: Bool {
         let locale = self as NSLocale
         guard let measurementSystem = locale.object(forKey: .measurementSystem) as? String else {
             return false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@
 * Fixed an issue where the entire route line was colored as `NavigationMapView.routeCasingColor` instead of `NavigationMapView.trafficUnknownColor` when traffic congestion data was missing. ([#3577](https://github.com/mapbox/mapbox-navigation-ios/pull/3577))
 * Fixed an issue where `RouteStepProgress.currentIntersection` was always returning invalid value, which in turn caused inability to correctly detect whether specific location along the route is in tunnel, or not. ([#3559](https://github.com/mapbox/mapbox-navigation-ios/pull/3559))
 
+### Other changes
+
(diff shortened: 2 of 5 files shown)


Sources/MapboxCoreNavigation/Locale.swift
```diff
@@ -23,20 +23,13 @@ extension Locale {
      Returns a `Locale` from `preferredLocalLanguageCountryCode`.
      */
     public static var nationalizedCurrent = Locale(identifier: preferredLocalLanguageCountryCode)
-    
+
     public static var usesMetric: Bool {
```
1ec5, line 27: Might as well deprecate this method if it’s as simple as writing `Locale.current.usesMetric`.

Sources/MapboxCoreNavigation/Locale.swift
```diff
@@ -23,20 +23,13 @@ extension Locale {
...
-        let locale = self.current as NSLocale
-        guard let measurementSystem = locale.object(forKey: .measurementSystem) as? String else {
-            return false
-        }
-        return measurementSystem == "Metric"
+        return current.usesMetricSystem
     }
     
+    @available(*, deprecated, renamed: "usesMetricSystem")
     public var usesMetric: Bool {
```
1ec5, line 32: `usesMetricSystem` behaves differently for `en-GB` locale, because the UK uses miles and yards for road distances and feet and inches for width and height restrictions, but they use the metric system for most other purposes: https://github.com/mapbox/mapbox-navigation-ios/pull/383#discussion_r127935522.

```swift
Locale(identifier: "en-GB").usesMetricSystem // true
NSLocale(localeIdentifier: "en-GB").usesMetricSystem // true
NSLocale(localeIdentifier: "en-GB").object(forKey: .measurementSystem) // "U.K."
```

This method is specifically used to decide the units for distances:

https://github.com/mapbox/mapbox-navigation-ios/blob/db95f8b68b528027036a92a83eb425274b88a905/Sources/MapboxNavigation/NavigationViewController.swift#L218

Instead of deprecating this method, we should rename it to `measuresDistancesInMeters`. Maybe there’s something in `MeasurementFormatter` that can make this decision more data-driven and avoid the use of NSLocale objects.

S2Ler: That makes sense, thanks, I will revisit this PR

S2Ler: Renamed

Sources/MapboxCoreNavigation/Locale.swift
```diff
@@ -24,15 +24,17 @@ extension Locale {
...
+        Locale.current.usesMetricDistanceUnits
     }
-    
+
+    @available(*, deprecated, renamed: "usesMetricDistanceUnits")
     public var usesMetric: Bool {
+        usesMetricDistanceUnits
+    }
+
+    public var usesMetricDistanceUnits: Bool {
```
1ec5, line 37: Nit: “Uses” isn’t a very meaningful word. `prefersMetricDistances`? `measuresDistancesInMetricUnits`?

1ec5 (approved):
Suggested changelog entry:

* Renamed the `Locale.usesMetric` property to `Locale.usesMetricDistanceUnits`. ([#3547](https://github.com/mapbox/mapbox-navigation-ios/pull/3547))

Sources/MapboxCoreNavigation/Locale.swift
S2Ler: Renamed to `measuresDistancesInMetricUnits` 👍

S2Ler:
> Suggested changelog entry:
> 
> * Renamed the `Locale.usesMetric` property to `Locale.usesMetricDistanceUnits`. ([#3547](https://github.com/mapbox/mapbox-navigation-ios/pull/3547))

Added, also, mentioning that the old method is still there.

bamx23 (approved):

PR #3550, opened 2021-11-03
Improve tests stability

S2Ler:
1. Do async status update during a free drive
    It is not needed as this method is already called on the main thread.
    The same approach was implemented in RouteController.
1. Fix ReplayLocationManager
    1. Fixes retain cycles introduced by "perform(_:with:afterDelay)".
    1. Don't loop by default. Looping most likely won't work as you would expect.
    1. Fix locations looping by correctly incrementing location timestamps.
1. An attempt to stabilize tests
    1. Stopping ReplayLocationManager at the end of the tests.
    1. Don't use ReplayLocationManager.tick manually as it starts the replay loop unconditionally.


S2Ler:
Changing default loop behaviour for ReplayLocationManager is technically a breaking change, but this type unlikely to be used in production and the loop behaviour is unlikely a desired behaviour.

Sources/MapboxCoreNavigation/ReplayLocationManager.swift
```diff
@@ -88,50 +90,73 @@ open class ReplayLocationManager: NavigationLocationManager {
...
+
+    /// Shift `locations` so that sent locations always have increasing timestamps, taking into account location deltas.
+    private func advanceLocationsForNextLoop() {
+        /// Previous location that is used to calculate deltas between locations.
+        var previousOldLocation: CLLocation = locations.last!
+        /// Previous timestamp that is used to advance timestamps.
+        var previousNewLocationTimestamp = previousOldLocation.timestamp
+
+        for (idx, location) in locations.enumerated() {
+            let delta: TimeInterval = idx == 0 ? 1 : location.timestamp.timeIntervalSince(previousOldLocation.timestamp)
```
S2Ler, line 155: should take speed into account

Sources/MapboxCoreNavigation/ReplayLocationManager.swift
```diff
@@ -88,50 +90,73 @@ open class ReplayLocationManager: NavigationLocationManager {
             synthesizedLocation = location
             delegate?.locationManager?(self, didUpdateLocations: [location])
             onTick?(currentIndex, location)
-            NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(tick), object: nil)
+            nextTickWorkItem?.cancel()
         }
 
         func scheduleNextTick(afterDelay delay: TimeInterval) {
-            perform(#selector(tick), with: nil, afterDelay: delay)
```
S2Ler, line 95: Retains self. So, in case of looping, there is a retain cycle that can only be broken by explicitly stopping updates. GCD doesn't have this problem.

Sources/MapboxCoreNavigation/ReplayLocationManager.swift
S2Ler: no it shouldn't.

Sources/MapboxCoreNavigation/ReplayLocationManager.swift
```diff
@@ -88,50 +90,73 @@ open class ReplayLocationManager: NavigationLocationManager {
...
+            DispatchQueue.main.asyncAfter(deadline: .now() + delay, execute: nextTickWorkItem)
         }
 
         guard locations.count > 1 else {
             sendTick(with: locations[0])
-            let startFromBeginning = onReplayLoopCompleted?(self) ?? true
+            let startFromBeginning = replayCompletionHandler?(self) ?? false
             if startFromBeginning {
+                advanceLocationsForNextLoop()
                 // We can't calculate the delay from the next location as we have only one, so we fallback to 1s.
```
azarovalex, line 109: Guess this comment isn't true anymore

S2Ler: Nice catch

azarovalex (approved):

Issue #3548, opened 2021-11-02
Provide an option to fit all routes instead of only one in preview mode on CarPlay.


MaximAlien:
https://github.com/mapbox/mapbox-navigation-ios/pull/3511 brought the ability to show all requested routes instead of only one in preview mode on CarPlay. At the same time geometry of only currently selected route is taken into account. 

In scope of this ticket new parameter should be introduced in the API. It'll allow to decide whether only one route or all routes should be presented. In case if all routes are presented there should also be the ability to control whether geometry of all routes is taken into account or only current one.

PR #3556, opened 2021-11-03
Implement the ability to show routes based on style provided in `RoutesPresentationStyle`.

MaximAlien:
Closing #3548.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@
 
 * Added the `NavigationViewController.usesNightStyleWhileInTunnel` and `CarPlayNavigationViewController.usesNightStyleWhileInTunnel` properties, which allow to disable dark style usage, while traversing the tunnels. ([#3559](https://github.com/mapbox/mapbox-navigation-ios/pull/3559))
 
+### Route overlay
+
+* Added the `RoutesPresentationStyle` enumeration, which allows to control routes presentation style while calling `NavigationMapView.showcase(_:routesPresentationStyle:animated:)`. ([#3556](https://github.com/mapbox/mapbox-navigation-ios/pull/3556))
+
 ### CarPlay
 
 * Added the `CarPlayActivity.panningInNavigationMode` case, which allows to track a state when user is panning a map view while actively navigating. ([#3545](https://github.com/mapbox/mapbox-navigation-ios/pull/3545))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@
 
 ### Route overlay
 
-* Renamed the `NavigationMapView.showcase(_:animated:)` method to `NavigationMapView.showcase(_:routesPresentationStyle:animated:)`, with an optional parameter to control whether the camera fits to unselected routes in addition to the selected route. ([#3556](https://github.com/mapbox/mapbox-navigation-ios/pull/3556))
+* Fixed an issue where `NavigationMapView.showcase(_:animated:)` was clipping unselected routes by renaming it to the `NavigationMapView.showcase(_:routesPresentationStyle:animated:)`, with an optional parameter to control whether the camera fits to unselected routes in addition to the selected route. ([#3556](https://github.com/mapbox/mapbox-navigation-ios/pull/3556))
 
 ### CarPlay
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@
 
 ### Route overlay
 
-* Added the `RoutesPresentationStyle` enumeration, which allows to control routes presentation style while calling `NavigationMapView.showcase(_:routesPresentationStyle:animated:)`. ([#3556](https://github.com/mapbox/mapbox-navigation-ios/pull/3556))
+* Renamed the `NavigationMapView.showcase(_:animated:)` method to `NavigationMapView.showcase(_:routesPresentationStyle:animated:)`, with an optional parameter to control whether the camera fits to unselected routes in addition to the selected route. ([#3556](https://github.com/mapbox/mapbox-navigation-ios/pull/3556))
 
 ### CarPlay
 
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -124,11 +124,13 @@ open class NavigationMapView: UIView {
      are currently deselected or inactive. The order of routes in this array may differ from
      the order in the original `RouteResponse`, for example in response to a user selecting a preferred
      route.
+     - parameter routesPresentationStyle: Route lines presentation style. By default the map will be
+     updated to fit all routes.
      - parameter animated: `true` to asynchronously animate the camera, or `false` to instantaneously
      zoom and pan the map.
      */
     public func showcase(_ routes: [Route],
-                         routesPresentationStyle: RoutesPresentationStyle = .single,
+                         routesPresentationStyle: RoutesPresentationStyle = .all(shouldFit: true),
                          animated: Bool = false) {
         guard let activeRoute = routes.first,
               let coordinates = activeRoute.shape?.coordinates,
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -1210,7 +1210,7 @@ open class NavigationMapView: UIView {
      Attempts to localize labels into the system’s preferred language.
      
      This method automatically modifies the `SymbolLayer.textField` property of any symbol style
-     layer whose source is the <a href="https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#overview">Mapbox Streets source</a>.
+     layer whose source is the [Mapbox Streets source](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#overview).
      The user can set the system’s preferred language in Settings, General Settings, Language & Region.
      
      This method avoids localizing road labels into the system’s preferred language, in an effort
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -736,8 +736,8 @@ open class NavigationMapView: UIView {
         
         // Pick a random tail direction to keep things varied.
         guard let randomTailPosition = [
-            RouteDurationAnnotationTailPosition.left,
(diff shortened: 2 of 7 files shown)


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -118,22 +138,38 @@ open class NavigationMapView: UIView {
...
      
-     - parameter routes: The routes to visualize in order of priority. The first route is displayed as if it is currently selected or active, while the remaining routes are displayed as if they are currently deselected or inactive. The order of routes in this array may differ from the order in the original `RouteResponse`, for example in response to a user selecting a preferred route.
-     - parameter legIndex: The zero-based index of the currently active leg along the active route. The active leg is highlighted more prominently than inactive legs.
+     - parameter routes: The routes to visualize in order of priority. The first route is displayed
+     as if it is currently selected or active, while the remaining routes are displayed as if they
+     are currently deselected or inactive. The order of routes in this array may differ from the
+     order in the original `RouteResponse`, for example in response to a user selecting a preferred
+     route.
+     - parameter legIndex: The zero-based index of the currently active leg along the active route.
+     The active leg is highlighted more prominently than inactive legs.
```
MaximAlien, line 172: I took an opportunity to reformat comments in `NavigationMapView` to be a bit shorter. It's pretty inconvenient to use horizontal scrolling while reading them.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -102,14 +107,29 @@ open class NavigationMapView: UIView {
...
+     as if it is currently selected or active, while the remaining routes are displayed as if they
+     are currently deselected or inactive. The order of routes in this array may differ from
+     the order in the original `RouteResponse`, for example in response to a user selecting a preferred
+     route.
+     - parameter animated: `true` to asynchronously animate the camera, or `false` to instantaneously
+     zoom and pan the map.
      */
-    public func showcase(_ routes: [Route], animated: Bool = false) {
+    public func showcase(_ routes: [Route],
+                         routesPresentationStyle: RoutesPresentationStyle = .single,
```
1ec5, line 131: Does this need to be a preference? It seems pretty reasonable to fit the camera to all the provided routes, unless we’re planning to use this method when displaying alternative routes during turn-by-turn navigation. If a caller doesn’t want to show any alternative routes, it wouldn’t pass in any alternative routes.

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 
 * Fixed the crash when scrolling the guidance cards while the orientaion changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
 
+### Route overlay
+
+* Added the `RoutesPresentationStyle` enumeration, which allows to control routes presentation style while calling `NavigationMapView.showcase(_:routesPresentationStyle:animated:)`. ([#3556](https://github.com/mapbox/mapbox-navigation-ios/pull/3556))
```
1ec5, line 27:
```suggestion
* Renamed the `NavigationMapView.showcase(_:animated:)` method to `NavigationMapView.showcase(_:routesPresentationStyle:animated:)`, with an optional parameter to control whether the camera fits to unselected routes in addition to the selected route. ([#3556](https://github.com/mapbox/mapbox-navigation-ios/pull/3556))
```

Sources/MapboxNavigation/NavigationMapView.swift
S2Ler: I'm with @MaximAlien here. It much easier to read documentation that is wrapped in a reasonable column width (80-120 is most commonly used). Besides this there are other advantages:
- Easier in read on GitHub for Web.
- Works without wrapping setting in the editor.
- Human can wrap text better. 
- Anything else? I've seen many other repos doing the same thing (including the ones from Apple).

S2Ler:
> # Breaking API Changes
> ## `NavigationMapView`
> * _removed_ method: `showcase(_:animated:)` in `NavigationMapView`

At this point we will have to bump base version. Otherwise this will appear in every PR after this lands. 

It is possible to not introduce such breaking change though. Adding a new method instead of modifying the old one. Even adding a new default parameter IS a breaking change for some code. An example:

```swift
import UIKit

// MARK: - v1 API

//func breakingMethod(param1: Bool) {
//
//}

// MARK: - v2 API

func breakingMethod(param1: Bool, param2: Bool = false) {

}

breakingMethod(param1: true)
let methodReference = breakingMethod(param1:) `#1`
```

`#1` line works in v1 version of the API but doesn't work in v2.

Sources/MapboxNavigation/RouteDurationAnnotationTailPosition.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+enum RouteDurationAnnotationTailPosition: Int {
+    
+    case left
+    case right
```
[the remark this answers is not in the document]
MaximAlien: Agree, renamed.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -1152,9 +1209,17 @@ open class NavigationMapView: UIView {
     /**
      Attempts to localize labels into the system’s preferred language.
      
-     This method automatically modifies the `SymbolLayer.textField` property of any symbol style layer whose source is the <a href="https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#overview">Mapbox Streets source</a>. The user can set the system’s preferred language in Settings, General Settings, Language & Region.
+     This method automatically modifies the `SymbolLayer.textField` property of any symbol style
+     layer whose source is the <a href="https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#overview">Mapbox Streets source</a>.
```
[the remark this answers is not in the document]
MaximAlien: Good catch. I've walked through all existing occurrences and changed them to use markdown style. Seems to be working with Xcode `13.1`.

Sources/MapboxNavigation/NavigationMapView.swift
MaximAlien: Since this method might be already used by the users, to play it safe I'd rather keep using presentation of only single route by default. On CarPlay, on the other hand, we'll be presenting all routes, and fit camera to all of them. 

Unfortunately this is a drawback of extending APIs in minor releases: we can try to keep this method but introduce another one. At the same time we might end up with multiple methods, which essentially do the same work, but with only slight difference in signatures.

MaximAlien:
> > # Breaking Changes in MapboxNavigation
> > # Breaking API Changes
> > ## `NavigationMapView`
> > 
> > * _removed_ method: `showcase(_:animated:)` in `NavigationMapView`
> 
> At this point we will have to bump base version. Otherwise this will appear in every PR after this lands.
> 
> It is possible to not introduce such breaking change though. Adding a new method instead of modifying the old one. Even adding a new default parameter IS a breaking change for some code. An example:
> 
> ```swift
> import UIKit
> 
> // MARK: - v1 API
> 
> //func breakingMethod(param1: Bool) {
> //
> //}
> 
> // MARK: - v2 API
> 
> func breakingMethod(param1: Bool, param2: Bool = false) {
> 
> }
> 
> breakingMethod(param1: true)
> let methodReference = breakingMethod(param1:) // 1
> ```
> 
> `// 1` line works in v1 version of the API but doesn't work in v2.

@S2Ler, what do you mean by bumping base version? Changing to the most recent commit from `main` after landing this PR?

I think that this change to new default parameter should not bring any issues, as new parameter is completely different from what we had before. Adding new method for each change like this will not scale in long term, I guess.

Sources/MapboxNavigation/NavigationMapView.swift
1ec5: I’m sure this method is already used in the wild, but that doesn’t necessarily stop us from fixing what was essentially a bug in the method’s behavior (clipping the alternative routes). My inclination would be to frame it as a bug fix in the changelog, which would also skirt the questions around API stability below.

Sources/MapboxNavigation/NavigationMapView.swift
1ec5: This isn’t something I feel strongly about; I just pointed out the downsides for awareness. In my opinion, the fact that GitHub’s file viewing mode doesn’t wrap lines is a sufficient argument in favor of hard wrapping. I’m just less convinced about the other points, since generally, we can assume most developers will use Xcode’s default (which is to soft-wrap).

Sources/MapboxNavigation/NavigationMapView.swift
MaximAlien: Alright, updated method signature to use default argument to fit all routes to map. Changelog entry was updated as well.

Sources/MapboxNavigation/RoutesPresentationStyle.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+/**
+ Style, which will be used when presenting routes on top of a map view by calling
```
S2Ler, line 4:
```suggestion
 A style that will be used when presenting routes on top of a map view by calling
```

Sources/MapboxNavigation/RoutesPresentationStyle.swift
```diff
@@ -0,0 +1,21 @@
...
+ */
+public enum RoutesPresentationStyle {
+    
+    /**
+     Only first route will be presented on a map view.
+     */
+    case single
+    
+    /**
+     All routes will be used presented on a map view.
```
S2Ler, line 15:
```suggestion
     All routes will be presented on a map view.
```

Sources/MapboxNavigation/NavigationMapView.swift
S2Ler: Even with wraps, the downside is that the text becomes hard to read because the text is too long[^1].

There are downsides as well as @1ec5 mentioned. But I think we should prioritize the decisions that make the lives of the devs who use our SDK easier/better. 

[^1]: [Readability of the text](https://ux.stackexchange.com/questions/108801/what-is-the-best-number-of-paragraph-width-for-readability)

avi-c (approved):
Other than the small syntax changes in the comments that Aliaksandr requested this looks good to me.

I'm approving pending their added approval.

MaximAlien:
I think that all review comments were addressed, so I'm moving forward with merge. Feel free to comment if you'd like to add any improvements. I'll try to land them in subsequent updates.

Issue #3470, opened 2021-10-12
Provide an option, which allows to disable style changing when driving in tunnels.


MaximAlien:
In scope of this ticket new option should be exposed to provide the ability to disable style changing when user is driving in the tunnel.

PR #3559, opened 2021-11-04
Add ability to disable dark style, while traversing through the tunnel.

MaximAlien:
Closing #3470.

### DIFF
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -310,6 +310,7 @@ open class RouteController: NSObject {
         updateSpokenInstructionProgress(status: status, willReRoute: willReroute)
         updateVisualInstructionProgress(status: status)
         updateRoadName(status: status)
+        updateDistanceToIntersection(from: CLLocation(status.location))
         
         if willReroute {
             reroute(from: CLLocation(status.location), along: routeProgress)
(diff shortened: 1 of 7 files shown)


Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -513,6 +523,23 @@ open class CarPlayNavigationViewController: UIViewController {
...
+            styleManager?.applyStyle(type: .night)
+        }
+        
+        // Exiting tunnel
+        if traversingTunnel, !inTunnel {
+            traversingTunnel = false
+            styleManager?.timeOfDayChanged()
+        }
+    }
+
```
MaximAlien, line 542: The more features we port to CarPlay, the more duplication we'll have. This method is basically copy of a similar method in `NavigationViewController`. Not sure what's the best way to extract it into different entity, as it uses member variables and there is a chance that we might want to tweak logic for iOS and CarPlay separately.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -805,7 +811,7 @@ extension NavigationViewController: NavigationServiceDelegate {
         }
     }
     
-    private func attemptToHighlightBuildings(_ progress: RouteProgress, with location: CLLocation) {
+    private func attemptToHighlightBuildings(_ progress: RouteProgress) {
```
MaximAlien, line 814: Removed unused parameter.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -783,13 +788,14 @@ extension NavigationViewController: NavigationServiceDelegate {
...
 
-        attemptToHighlightBuildings(progress, with: location)
+        attemptToHighlightBuildings(progress)
         
         // Finally, pass the message onto the `NavigationViewControllerDelegate`.
         delegate?.navigationViewController(self, didUpdate: progress, with: location, rawLocation: rawLocation)
     }
     
     private func checkTunnelState(at location: CLLocation, along progress: RouteProgress) {
+        guard usesDarkStyleWhileInTunnel else { return }
```
ShanMa1991, line 798: If the check of `usesDarkStyleWhileInTunnel` before the judgment of the `traversingTunnel` and `inTunnel`, there's a case that the `traversingTunnel` wouldn't be updated. Would it affect the following and later tunnel check if the `usesDarkStyleWhileInTunnel` is changed during navigation?

ShanMa1991 (approved):

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
1ec5: Long term, we should align the two screens so that CarPlayNavigationViewController and NavigationViewController can share a more generic version of NavigationView, probably protocol-based.

Sources/MapboxNavigation/NavigationViewController.swift
```diff
@@ -110,6 +110,11 @@ open class NavigationViewController: UIViewController, NavigationStatusPresenter
         }
     }
     
+    /**
+     Controls whether dark style will be used whenever traversing through a tunnel. Defaults to `true`.
+     */
+    public var usesDarkStyleWhileInTunnel: Bool = true
```
1ec5, line 116: `usesNightStyleWhileInTunnel` would be consistent with the naming of NightStyle. The style’s name is counterintuitive, but it was based on the map style’s name and was implemented before we generalized the feature to tunnels. At least developers wouldn’t wonder if there’s a difference between a dark style and a night style.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -66,6 +71,8 @@ open class CarPlayNavigationViewController: UIViewController {
     var mapTemplate: CPMapTemplate
     var carInterfaceController: CPInterfaceController
     
+    private var traversingTunnel = false
```
1ec5, line 74: Rename to `isTraversingTunnel`.

Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -321,6 +322,37 @@ open class RouteController: NSObject {
...
+    func updateDistanceToIntersection(from location: CLLocation) {
+        guard var intersections = routeProgress.currentLegProgress.currentStepProgress.step.intersections else { return }
+        
+        // The intersections array does not include the upcoming maneuver intersection.
+        if let upcomingStep = routeProgress.currentLegProgress.upcomingStep,
+           let upcomingIntersection = upcomingStep.intersections?.first {
+            intersections += [upcomingIntersection]
+        }
+        
+        routeProgress.currentLegProgress.currentStepProgress.intersectionsIncludingUpcomingManeuverIntersection = intersections
```
1ec5, line 334: This bug fix is worth calling out in the changelog.

Sources/MapboxNavigation/NavigationViewController.swift
MaximAlien: Even though `isTraversingTunnel` used only in this method, there is a chance that we might use it in other places of `NavigationViewController` in future. So yes, I agree that we should assign correct value to it each time. Modified the code.

Issue #3561, opened 2021-11-04
Add the ability to highlight buildings on CarPlay during active-guidance.


MaximAlien:
At this point there is no ability to highlight buildings during active guidance in `CarPlayNavigationViewController`. In scope of this ticket such ability should be added.

PR #3564, opened 2021-11-05
Implement the ability to highlight buildings during active-guidance on CarPlay.

MaximAlien:
Closing #3561.

### DIFF
--- a/Sources/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/Sources/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -12,7 +12,7 @@ import CarPlay
  - seealso: `NavigationViewController`
  */
 @available(iOS 12.0, *)
-open class CarPlayNavigationViewController: UIViewController {
+open class CarPlayNavigationViewController: UIViewController, BuildingHighlighting {
     
     // MARK: Child Views and Styling Configuration
     
@@ -72,6 +72,16 @@ open class CarPlayNavigationViewController: UIViewController {
      The style can be modified programmatically by using `StyleManager.applyStyle(type:)`.
      */
     public private(set) var styleManager: StyleManager?
+    
+    /**
+     Allows to control highlighting of the destination building on arrival. By default destination buildings will not be highlighted.
+     */
+    public var waypointStyle: WaypointStyle = .annotation
+    
+    var approachingDestinationThreshold: CLLocationDistance = DefaultApproachingDestinationThresholdDistance
+    var passedApproachingDestinationThreshold: Bool = false
+    var currentLeg: RouteLeg?
+    var buildingWasFound: Bool = false
     
     var mapTemplate: CPMapTemplate
     var carInterfaceController: CPInterfaceController
@@ -452,9 +462,10 @@ open class CarPlayNavigationViewController: UIViewController {
         
         // Reapply runtime styling changes each time the style changes.
         navigationMapView.mapView.mapboxMap.onEvery(.styleLoaded) { [weak self] _ in
-            self?.navigationMapView?.localizeLabels()
-            self?.navigationMapView?.mapView.showsTraffic = false
-            self?.updateRouteOnMap()
+            guard let self = self else { return }
+            self.navigationMapView?.localizeLabels()
+            self.navigationMapView?.mapView.showsTraffic = false
+            self.updateRouteOnMap()
         }
         
         navigationMapView.mapView.ornaments.options.compass.visibility = .hidden
@@ -555,6 +566,8 @@ open class CarPlayNavigationViewController: UIViewController {
         
         // Check to see if we're in a tunnel.
         checkTunnelState(at: location, along: routeProgress)
+        
+        attemptToHighlightBuildings(routeProgress, navigationMapView: navigationMapView)
         
         let legIndex = routeProgress.legIndex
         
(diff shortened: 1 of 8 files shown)


Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -68,10 +68,19 @@ open class CarPlayNavigationViewController: UIViewController {
      */
     public private(set) var styleManager: StyleManager?
     
+    /**
+     Allows to control highlighting of the destination building on arrival. By default destination buildings will not be highlighted.
```
bamx23, line 72: The documentation looks like it is for a different property.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -542,6 +553,33 @@ open class CarPlayNavigationViewController: UIViewController {
         }
     }
     
+    private func attemptToHighlightBuildings(_ progress: RouteProgress) {
+        // In case if distance was fully covered - do nothing.
+        // FIXME: This check prevents issue which leads to highlighting random buildings after arrival to final destination.
+        // At the same time this check will prevent building highlighting in case of arrival in overview mode/high altitude.
+        if progress.fractionTraveled >= 1.0 { return }
+        if waypointStyle == .annotation { return }
```
bamx23, line 561: Not: May it be a better fit for guard?

bamx23 (approved):
Also, do we need an update in changelog?

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -68,10 +68,19 @@ open class CarPlayNavigationViewController: UIViewController {
...
+    public var waypointStyle: WaypointStyle = .annotation
+    
     var mapTemplate: CPMapTemplate
     var carInterfaceController: CPInterfaceController
     
     private var isTraversingTunnel = false
+    private var approachingDestinationThreshold: CLLocationDistance = 250.0
+    private var passedApproachingDestinationThreshold: Bool = false
+    private var currentLeg: RouteLeg?
+    private var foundAllBuildings = false
```
S2Ler, line 83: Let's use consistent style:
```suggestion
    private var foundAllBuildings: Bool = false
```

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -68,10 +68,19 @@ open class CarPlayNavigationViewController: UIViewController {
...
+    /**
+     Allows to control highlighting of the destination building on arrival. By default destination buildings will not be highlighted.
+     */
+    public var waypointStyle: WaypointStyle = .annotation
+    
     var mapTemplate: CPMapTemplate
     var carInterfaceController: CPInterfaceController
     
     private var isTraversingTunnel = false
+    private var approachingDestinationThreshold: CLLocationDistance = 250.0
```
S2Ler, line 80: I think this is a copy from `NavigationViewController` and we shall think how to restructure code so that there won't be so much duplicated code.

At very least, 250.0 is a shared constant that can be extracted.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -542,6 +553,33 @@ open class CarPlayNavigationViewController: UIViewController {
         }
     }
     
+    private func attemptToHighlightBuildings(_ progress: RouteProgress) {
```
S2Ler, line 556: This is a copy from `NavigationViewController`. Even, fixme is copied. We need to extract this logic in a reusable algorithm.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
MaximAlien: This is original name of the property we came up with to control building highlighting.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
MaximAlien: Agree, added guard statement.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
MaximAlien: Was also renamed to `buildingWasFound` to be more precise.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
MaximAlien: Agree, since there are also several class member properties involved in the building highlighting process I've created internal `BuildingHighlighting` protocol. Please check latest changes.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
MaximAlien: Some refactoring was made and now building highlighting logic is the same for `CarPlayNavigationViewController` and `NavigationViewController`.

Sources/MapboxNavigation/Constants.swift
```diff
@@ -21,8 +21,9 @@ public var RouteLineWidthByZoomLevel: [Double: Double] = [
 ]
 
 /**
- The minium distance remaining on a route before overhead zooming is stopped.
+ The minimum distance remaining on a route before overhead zooming is stopped.
  */
+@available(*, deprecated, message: "This value is no longer used.")
 public var NavigationMapViewMinimumDistanceForOverheadZooming: CLLocationDistance = 200
```
MaximAlien, line 27: This property is no longer used as it seems, so I've deprecated it.

Issue #3563, opened 2021-11-05
[Bug]: Route duration annotation symbol layer adding error


ShanMa1991:
### Mapbox Navigation SDK version

2.1.0

### Steps to reproduce

This issue doesn't only exist in iOS 15 but also in iOS 14. 
1. Choose Example
2. Select a waypoint to show the route and the route duration

### Expected behavior

Route duration annotation should show up

### Actual behavior

No route annotation symbol layer shows up, with the following error message:

```
Error occured while adding route annotation symbol layer: Layer MapboxNavigation-MapboxNavigation-resources_routeDurationAnnotationsLayer is not in style.
```

It seems that the source adding is correct, but it always failed to add the symbol layer.

### Is this a one-time issue or a repeatable issue?

repeatable

MaximAlien:
Looks like it's a matter of checking whether layer is currently present or not (not only source). Something like this should help in `NavigationMapView.addRouteAnnotationSymbolLayer(features:)`:
```swift
var shapeLayer: SymbolLayer
if style.layerExists(withId: routeDurationAnnotationsLayerIdentifier),
   let symbolLayer = try style.layer(withId: routeDurationAnnotationsLayerIdentifier) as? SymbolLayer {
    shapeLayer = symbolLayer
} else {
    shapeLayer = SymbolLayer(id: routeDurationAnnotationsLayerIdentifier)
}
```

PR #3566, opened 2021-11-05
Fix issue, which was preventing from adding route duration layer to the map view.

MaximAlien:
Closing #3563.


ShanMa1991 (approved):

ShanMa1991:
Maybe we could also add this feature in CarPlay as well for parity with mobile ?

MaximAlien:
> Maybe we could also add this feature in CarPlay as well for parity with mobile ?

Yep, I think it'd make sense. Created ticket in: https://github.com/mapbox/mapbox-navigation-ios/issues/3574. Technically this is public API, so only example might be updated. At the same time there might be some CarPlay-related issues along the way.

PR #3568, opened 2021-11-08
Glue docsets with SourceKitten instead of CocoaPods

### LINKED ISSUE #3452: Migrate docset generation from CocoaPods to SPM
Since #285, the documentation generation script has combined the MapboxCoreNavigation and MapboxNavigation docsets into a single docset using a dedicated CocoaPods podspec. This hack is fragile, requiring us to artificially avoid reusing the same file name across both modules. Whenever we accidentally introduce a file name collision, jazzy fails completely with an obtuse error, such as in #2363.

It turns out jazzy has an officially supported technique for glomming together two modules, by generating their docsets manually using SourceKitten then passing the SourceKitten output into jazzy: https://github.com/realm/jazzy/issues/1194#issuecomment-633012820 https://github.com/realm/jazzy/issues/564#issuecomment-265289632. It appears this workflow requires SPM; CocoaPods would be unnecessary.

This refactor would allow us to rename some extension files back to more natural names, like CoreConstants.swift to Constants.swift. Migrating to SPM would also enable us to remove the hack for CocoaPods compatibility that modifies the source code before running jazzy:

https://github.com/mapbox/mapbox-navigation-ios/blob/55545880ae0610599eb1924e4278caf17fb0da73/scripts/document.sh#L40-L46

Thanks for @jmkiley for pointing out the SPM-based workflow.

/cc @mapbox/navigation-ios

chezzdev:
Fixes #3452.

Interesting parts here are:
- I wasn't able to utilize [`--spm`](https://github.com/realm/jazzy/issues/1194#issuecomment-633012820)-mode of sourcekitten, as there's no known way to specify a platform to SPM, hence errors from trying to resolve dependencies for macOS platform. Instead, I used MapboxNavigation-SPM.xcodeproj and `xcodebuild`-mode of sourcekitten with project, destination and scheme specification. This way, package dependencies are resolved just fine.
- turns out there's sourcekitten binary [bundled](https://github.com/realm/jazzy/blob/7728656ba0e2714c05bbf22c83e17d567d289f8a/bin/sourcekitten) with jazzy, so we don't need to install it separately.
- bye-bye MapboxNavigation-Documentation.podspec


scripts/document.sh
```diff
@@ -36,33 +43,36 @@ perl -pi -e "s/\\$\\{SHORT_VERSION\\}/${SHORT_VERSION}/" "${README}"
...
+    --module_version ${RELEASE_VERSION} \
+    --sourcekitten-sourcefile core.json,ui.json \
+    2>&1 | tee docs.output
+    
+if egrep -e "(WARNING)|(USR)" docs.output; then
+    echo "Please eliminate Jazzy warnings"
+    exit 1
+fi
+    
+rm core.json ui.json
```
bamx23, line 70: Is it worth moving this line before the `if` block above?

scripts/document.sh
```diff
@@ -36,33 +43,36 @@ perl -pi -e "s/\\$\\{SHORT_VERSION\\}/${SHORT_VERSION}/" "${README}"
...
     --root-url "${BASE_URL}/navigation/api/${RELEASE_VERSION}/" \
     --theme ${THEME} \
     --output ${OUTPUT} \
-    --module_version ${RELEASE_VERSION}
+    --module_version ${RELEASE_VERSION} \
+    --sourcekitten-sourcefile core.json,ui.json \
+    2>&1 | tee docs.output
+    
+if egrep -e "(WARNING)|(USR)" docs.output; then
+    echo "Please eliminate Jazzy warnings"
```
bamx23, line 66: Could you add `docs.output` to artifacts or print it here?

bamx23 (approved):
Great work! Have you checked the diff between the old and this new way of docs generation?

scripts/document.sh
chezzdev: The output that's captured in this file will also be printed to the console, thanks to `tee`.
This output will also be the last one in the script, so no need to reprint it or save otherwise.

scripts/document.sh
bamx23: And should we remove the `docs.output`? _(this one should be here, after the `if` for sure)_

chezzdev: Yeah, we should! Updated.

PR #3569, opened 2021-11-09
Rewrite lane use diagram drawing and selection logic

### LINKED ISSUE #3374: Streamline lane use diagram selection logic
#2809 rewrote LaneView to select lane use diagram drawing methods more comprehensively. Unfortunately, the resulting code is still far too complex and difficult to reason about, much less port to Kotlin for Android. mapbox/navigation-ui-resources#26 refactored the assets and introduced a new algorithm for straightforwardly selecting an asset. We should rewrite LaneView according to that algorithm, not only for simplicity but also to handle the lane configurations added in that PR and mapbox/navigation-ui-resources#25. The test cases in #2809 will be valuable, but we can expect a few of them to fail because we’re intentionally omitting certain rare fallbacks from the new algorithm to keep the algorithm elegant.

mapbox/mapbox-navigation-android#3934 tracks the same integration work, but it’s more urgent on Android because even common dual-use lane configurations aren’t handled correctly yet.

/cc @mapbox/navigation-ios @abhishek1508 @LukasPaczos @kmadsen

### LINKED ISSUE #3530: [Bug]: SlightLeft-SlightRight lanes doesn't render

v2.0.0

### Steps to reproduce

Set route from 1.3054251688045753;103.88307997278685 to 1.309264205529388;103.87483300071494 and start guidance cards Example navigation. One of the guidance cards will have a non-rendered lane. 

### Expected behavior

`LaneIndication.Ranking`
```
▿ Ranking
  ▿ primary : slight left
    - rawValue : 32
  ▿ secondary : Optional<LaneIndication>
    ▿ some : right
      - rawValue : 4
  - tertiary : nil
```

can be rendered

### Actual behavior

Empty lane view

repeatable

1ec5:
Imported the expanded repertoire of lane StyleKit methods from mapbox/navigation-ui-resources#26. Rewrote the lane use diagram selection logic to follow [the algorithm described in navigation-ui-resources](https://github.com/mapbox/navigation-ui-resources/blob/4a287b92ddeeec502bca9da849e505dcdf73e1ef/docs/lanes.md).

For type safety and confidence about coverage, the new logic still relies heavily on enumerations and switch statements, so it doesn’t actually reduce the line count by so much. It would be a fraction of this length by relying on string processing, but we can revisit if we need to make this logic extensible by developers. Regardless, the new logic is considerably more straightforward than before, with fallbacks but no edge cases. The new StyleKit assets and tests vastly outweigh the selection logic.

De Anza Boulevard[^deanza] Nicoll Highway[^nicoll] Mountbatten Road[^mountbatten] Dixie Highway[^dixie]

[^deanza]: [De Anza Boulevard](https://www.openstreetmap.org/way/393494129)
[^nicoll]: [Nicoll Highway](https://www.openstreetmap.org/way/476758104)
[^mountbatten]: [Mountbatten Road](https://www.openstreetmap.org/way/746988839) (recently [edited](https://www.openstreetmap.org/changeset/113534817) to have a different turn lane configuration, not shown here)
[^dixie]: [Dixie Highway](https://www.openstreetmap.org/way/760892734)

Fixes #3374 and fixes #3530.

/cc @mapbox/navigation-ios @avi-c @abhishek1508


Sources/MapboxNavigation/LaneView.swift
```diff
@@ -1,342 +1,337 @@
...
-                tertiaryIndication = nil
-            }
-            return Ranking(primary: primaryIndication, secondary: secondaryIndication, tertiary: tertiaryIndication)
-        } else {
-            // No secondary indication to draw.
-            return Ranking(primary: primaryIndication, secondary: nil, tertiary: nil)
+        guard let method = styleKitMethod(turnClassifications: turnClassifications, favoredTurnClassification: favoredTurnClassification) ??
+                favoredTurnClassification.map({ styleKitMethod(turnClassifications: [$0], favoredTurnClassification: $0) }) ??
+                styleKitMethod(turnClassifications: [.straightAhead], favoredTurnClassification: nil) else {
+            preconditionFailure("No StyleKit method for straight ahead.")
```
1ec5, line 158: Rest assured, we have a StyleKit method for straight ahead, so this precondition is just for completeness’ sake.

Sources/MapboxNavigation/LaneView.swift
```diff
@@ -1,342 +1,337 @@
...
         }
+        return method
```
1ec5, line 160: This is the crux of the change, the top level of the algorithm that was implemented in Swift.

MaximAlien (approved):
With so many changes it's hard to find anything that might look out of order. In general looks good to me. Thanks.

S2Ler (approved):
Agree with Maxim.

Issue #3562, opened 2021-11-05
MapboxSpeechSynthesizer `willSpeak` callback is not called


azarovalex:
Reproducible on the version 2.0.0

**Steps to reproduce:**
1. Open the development app
2. Start an active navigation to any point
3. willSpeak won't be called for every instruction

Expected behavior: `willSpeak` should be called

Example of the current incorrect behavior for a random trip:
<img width="291" alt="image" src="https://user-images.githubusercontent.com/1976216/140527837-4407d62c-b6f9-4aab-ae3a-27b18686d978.png">

azarovalex:
The PR with the fix was merged

PR #3572, opened 2021-11-09
Add public `managesAudioSession` flag and fix `willSpeak` logic

azarovalex:
This PR fixes two issues in `MapboxSpeechSynthesizer`:
1. `managesAudioSession` is a new flag in `MapboxSpeechSynthesizer ` that defines who controls the shared AVAudioSession – SDK or Application. Default value is `true` that means that SDK manages the session, as it used to be before. If application has background music they want to manage the audio session themselves, so they should set this flag to `false`.
Without setting this flag to `false` we will be stopping the `AVAudioSession` after every voice instruction and this will stop application music as well.

2. We used to call `willSpeak` before downloading and caching the instruction, not before playing it. `MapboxSpeechSynthesizer` pre-caches first three voice instruction on initialization and this leads to bug https://github.com/mapbox/mapbox-navigation-ios/issues/3562. I fixed it by moving `willSpeak` call right before playing the instruction.

TODO:
- [x] update changelog


Sources/MapboxNavigation/MultiplexedSpeechSynthesizer.swift
```diff
@@ -40,6 +40,11 @@ open class MultiplexedSpeechSynthesizer: SpeechSynthesizing {
         speechSynthesizers.forEach { $0.locale = locale }
     }
     
+    /// Controls if this speech synthesizer is allowed to manage the shared `AVAudioSession`.
+    /// Set this field to `false` if you want to manage the session yourself, for example if your app has background music.
+    /// Default value is`true`.
+    public var managesAudioSession = true
```
Udumft, line 46: This has to be a computed property which will forward `newValue` to underlying speech synthesizers.

azarovalex: Good idea, thanks

S2Ler:
> # Breaking API Changes
> ## `InstructionsCardCell`
> * _removed_ method: `layoutSubviews()` in `InstructionsCardCell`
> * _removed_ var: `container` in `InstructionsCardCell`

I'm fixing this.

Sources/MapboxNavigation/MapboxSpeechSynthesizer.swift
```diff
@@ -35,6 +35,11 @@ open class MapboxSpeechSynthesizer: NSObject, SpeechSynthesizing {
      */
     public var audioPlayer: AVAudioPlayer?
     
+    /// Controls if this speech synthesizer is allowed to manage the shared `AVAudioSession`.
+    /// Set this field to `false` if you want to manage the session yourself, for example if your app has background music.
+    /// Default value is`true`.
```
bamx23, line 40:
```suggestion
    /// Default value is `true`.
```

Sources/MapboxNavigation/MultiplexedSpeechSynthesizer.swift
```diff
@@ -40,6 +40,18 @@ open class MultiplexedSpeechSynthesizer: SpeechSynthesizing {
         speechSynthesizers.forEach { $0.locale = locale }
     }
     
+    /// Controls if this speech synthesizer is allowed to manage the shared `AVAudioSession`.
+    /// Set this field to `false` if you want to manage the session yourself, for example if your app has background music.
+    /// Default value is`true`.
+    public var managesAudioSession: Bool {
+        get {
+            speechSynthesizers.reduce(true, { $0 && $1.managesAudioSession })
```
bamx23, line 48: NIT: What do you think about such approach?
```suggestion
            !speechSynthesizers.contains(where: { !$0.managesAudioSession })
```

Sources/MapboxNavigation/MultiplexedSpeechSynthesizer.swift
```diff
@@ -40,6 +40,18 @@ open class MultiplexedSpeechSynthesizer: SpeechSynthesizing {
         speechSynthesizers.forEach { $0.locale = locale }
     }
     
+    /// Controls if this speech synthesizer is allowed to manage the shared `AVAudioSession`.
+    /// Set this field to `false` if you want to manage the session yourself, for example if your app has background music.
+    /// Default value is`true`.
```
bamx23, line 45:
```suggestion
    /// Default value is `true`.
```

Sources/MapboxNavigation/SpeechSynthesizing.swift
```diff
@@ -18,6 +18,10 @@ public protocol SpeechSynthesizing: AnyObject {
     var isSpeaking: Bool { get }
     /// Locale setting to vocalization. This locale will be used as 'default' if no specific locale is passed for vocalizing each individual instruction.
     var locale: Locale? { get set }
+    /// Controls if this speech synthesizer is allowed to manage the shared `AVAudioSession`.
+    /// Set this field to `false` if you want to manage the session yourself, for example if your app has background music.
+    /// Default value is`true`.
```
bamx23, line 23:
```suggestion
    /// Default value is `true`.
```

Sources/MapboxNavigation/SystemSpeechSynthesizer.swift
```diff
@@ -30,6 +30,11 @@ open class SystemSpeechSynthesizer: NSObject, SpeechSynthesizing {
     
     public var locale: Locale? = Locale.autoupdatingCurrent
     
+    /// Controls if this speech synthesizer is allowed to manage the shared `AVAudioSession`.
+    /// Set this field to `false` if you want to manage the session yourself, for example if your app has background music.
+    /// Default value is`true`.
```
bamx23, line 35:
```suggestion
    /// Default value is `true`.
```

Tests/MapboxNavigationTests/NavigationViewControllerTestDoubles.swift
```diff
@@ -17,6 +17,7 @@ class SpeechSynthesizerStub: SpeechSynthesizing {
     var volume: Float = 1.0
     var isSpeaking: Bool = false
     var locale: Locale? = Locale.autoupdatingCurrent
+    var managesAudioSession: Bool = false
```
bamx23, line 20: Why is it false in tests?

bamx23 (approved):

Sources/MapboxNavigation/MultiplexedSpeechSynthesizer.swift
azarovalex: Agree, might be easier to understand

PR #3575, opened 2021-11-09
Allow the customization of TopBannerViewController.

### LINKED ISSUE #3567: NavigationViewControllerDelegate unable to override instruction label
As of #1530, the recommended approach to override the contents of the maneuver instruction label is to implement `VisualInstructionDelegate.label(_:willPresent:as:)` within a `NavigationViewControllerDelegate`-conforming class, because `NavigationViewControllerDelegate` conforms to `VisualInstructionDelegate`. However, this method is never called on a `NavigationViewControllerDelegate`.

Here’s where the override would take place:

https://github.com/mapbox/mapbox-navigation-ios/blob/d2d031317a7fdb888265d643993a4e41d3e8de15/Sources/MapboxNavigation/InstructionLabel.swift#L36

However, nothing ultimately sets `instructionDelegate` beyond the containing banner’s delegate. By code inspection, this issue seems to apply to both traditional banners and guidance cards.

This issue was reported in v1.4.2 but likely affects v2._x_ too.

/cc @mapbox/navigation-ios

ShanMa1991:
### Description
This pr is to allow the customization of `TopBannerViewController` and fix #3567 

### Implementation
1. `TopBannerViewController`'s `lanesView` , `nextBannerView` , `statusView` , `junctionView` are exposed publicly to allow access and modification of it's contents.
2.  `NavigationViewControllerDelegate ` right now extends `TopBannerViewControllerDelegate`,both  `TopBannerViewControllerDelegate` and `InstructionsBannerViewDelegate` extend `VisualInstructionDelegate`. So right now through the `NavigationViewControllerDelegate`, developers could call `fNavigationViewControllerDelegate.label(_:willPresent:as:)` to customize the `instructionsBannerView` of the `TopBannerViewController` during navigation.

### Screenshots or Gifs
For example, in `Example`, we allow the `ViewController` extends `NavigationViewControllerDelegate` as following to customize the `instructionsBannerView` of `TopBannerViewController` by presenting a reversed instruction.

```
extension ViewController: NavigationViewControllerDelegate {
    func label(_ label: InstructionLabel, willPresent instruction: VisualInstruction, as presented: NSAttributedString) -> NSAttributedString? {
        let forwards = Array(presented.string)
        let reverse = String(forwards.reversed())
        var range = NSRange(location: 0, length: presented.string.count)
        let attributes = presented.attributes(at: 0, effectiveRange: &range)
        return NSAttributedString(string: reverse, attributes: attributes)
    }
}
```
reversedInstruction

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,12 @@
 ### Banners and guidance instructions
 
 * Fixed the crash when scrolling the guidance cards while the orientation changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
+* Added the ability to customize the `TopBannerViewController` through the publicly exposed `lanesView`, `nextBannerView`, `statusView` and `junctionView`. The `TopBannerViewController.instructionsBannerView` could also be customized through `NavigationViewControllerDelegate.label(_:willPresent:as:)` during active navigation. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
 
 ### Map
 
 * Added the `NavigationViewController.usesNightStyleWhileInTunnel` and `CarPlayNavigationViewController.usesNightStyleWhileInTunnel` properties, which allow to disable dark style usage, while traversing the tunnels. ([#3559](https://github.com/mapbox/mapbox-navigation-ios/pull/3559))
+
 
 ### CarPlay
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,10 @@
 
 ### Banners and guidance instructions
 
-* Fixed the crash when scrolling the guidance cards while the orientation changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
-* Added the ability to customize the `TopBannerViewController` through the publicly exposed `lanesView`, `nextBannerView`, `statusView` and `junctionView`. The `TopBannerViewController.instructionsBannerView` could also be customized through `NavigationViewControllerDelegate.label(_:willPresent:as:)` during active navigation. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
+* Fixed the crash when scrolling the guidance cards while the orientaion changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
+* Added the `TopBannerViewController.lanesView`, `TopBannerViewController.nextBannerView`, `TopBannerViewController.statusView` and `TopBannerViewController.junctionView` properties. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
+* The `InstructionsBannerViewDelegate` and `TopBannerViewControllerDelegate` protocols now conform to the `VisualInstructionDelegate` protocol. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
+* Fixed an issue where `VisualInstructionDelegate.label(_:willPresent:as:)` was never called. Your `NavigationViewControllerDelegate` class can now implement this method to customize the contents of a visual instruction during turn-by-turn navigation. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
 
 ### Map
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 
 ### Banners and guidance instructions
 
-* Fixed the crash when scrolling the guidance cards while the orientaion changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
+* Fixed the crash when scrolling the guidance cards while the orientation changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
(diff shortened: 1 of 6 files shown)


Sources/MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -6,7 +6,7 @@ import MapboxDirections
 /**
  `InstructionsBannerViewDelegate` provides methods for reacting to user interactions in `InstructionsBannerView`.
  */
-public protocol InstructionsBannerViewDelegate: AnyObject, UnimplementedLogging {
+public protocol InstructionsBannerViewDelegate: VisualInstructionDelegate {
```
1ec5, line 9: This protocol should still conform to `AnyObject` and `UnimplementedLogging` for backwards compatibility. Also, `UnimplementedLogging` is useful for highlighting missing delegate method implementations.

Sources/MapboxNavigation/TopBannerViewController.swift
```diff
@@ -4,7 +4,7 @@ import UIKit
 import MapboxCoreNavigation
 import MapboxDirections
 
-public protocol TopBannerViewControllerDelegate: AnyObject, UnimplementedLogging {
+public protocol TopBannerViewControllerDelegate: VisualInstructionDelegate {
```
1ec5, line 7: Keep `AnyObject` and `UnimplementedLogging` here too.

Sources/MapboxNavigation/TopBannerViewController.swift
```diff
@@ -71,19 +71,10 @@ open class TopBannerViewController: UIViewController {
...
-    
-    lazy var junctionView: JunctionView = {
-        let view: JunctionView = .forAutoLayout()
-        view.isHidden = true
-        return view
-    }()
+    public var lanesView: LanesView = .forAutoLayout(hidden: true)
+    public var nextBannerView: NextBannerView = .forAutoLayout(hidden: true)
+    public var statusView: StatusView = .forAutoLayout(hidden: true)
+    public var junctionView: JunctionView = .forAutoLayout(hidden: true)
```
1ec5, line 77: These public properties should be documented.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 ### Banners and guidance instructions
 
 * Fixed the crash when scrolling the guidance cards while the orientaion changes. ([#3544](https://github.com/mapbox/mapbox-navigation-ios/pull/3544))
+* Added the ability to customize the `TopBannerViewController` through the publicly exposed `lanesView`, `nextBannerView`, `statusView` and `junctionView`. The `TopBannerViewController.instructionsBannerView` could also be customized through `NavigationViewControllerDelegate.label(_:willPresent:as:)` during active navigation. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
```
1ec5, line 24:
```suggestion
* Added the `TopBannerViewController.lanesView`, `TopBannerViewController.nextBannerView`, `TopBannerViewController.statusView` and `TopBannerViewController.junctionView` properties. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
* The `InstructionsBannerViewDelegate` and `TopBannerViewControllerDelegate` protocols now conform to the `VisualInstructionDelegate` protocol. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
* Fixed an issue where `VisualInstructionDelegate.label(_:willPresent:as:)` was never called. Your `NavigationViewControllerDelegate` class can now implement this method to customize the contents of a visual instruction during turn-by-turn navigation. ([#3575](https://github.com/mapbox/mapbox-navigation-ios/pull/3575))
```

1ec5 (commented):
This PR doesn’t address `InstructionsCardContainerViewDelegate`. We should either leave #3567 open or split out a separate issue for the same issue when using guidance cards.

Sources/MapboxNavigation/InstructionsBannerView.swift
ShanMa1991: But because `VisualInstructionDelegate` already conforms to `AnyObject` and `UnimplementedLogging`, there's a redundant declaration of the two protocols.

Sources/MapboxNavigation/TopBannerViewController.swift
ShanMa1991: Thanks for the suggestion, I'll add the documentation for the public properties

Sources/MapboxNavigation/TopBannerViewController.swift
ShanMa1991: Same as the `InstructionsBannerViewDelegate`. There's a redundant declaration of  `AnyObject` and `UnimplementedLogging` because `VisualInstructionDelegate` already conforms to them.

S2Ler:
> # Breaking API Changes
> ## `InstructionsBannerViewDelegate`
> * _modified_ protocol: `InstructionsBannerViewDelegate`
>   
>   
>   
>   Type of change:
>   Inherited Types
>   
>   
>   
>   
>   From:
>   `[["key.name": AnyObject], ["key.name": UnimplementedLogging]]`
>   
>   
>   To:
>   `[["key.name": VisualInstructionDelegate]]`
> 
> ## `TopBannerViewControllerDelegate`
> * _modified_ protocol: `TopBannerViewControllerDelegate`
>   
>   
>   
>   Type of change:
>   Inherited Types
>   
>   
>   
>   
>   From:
>   `[["key.name": AnyObject], ["key.name": UnimplementedLogging]]`
>   
>   
>   To:
>   `[["key.name": VisualInstructionDelegate]]`

Remember to bump `.base_api` file with an explanation of why.

Issue #3542, opened 2021-11-01
Route line shows `routeCasingColor` instead of associated congestion color.


ShanMa1991:
From `v2.0.0-rc.6`, we have noticed that the route line would show `routeCasingColor` instead of associated congestion color during the active navigation as the following. A customer also found this issue since `v2.0.0-rc.6`. And this bug leads to the route line disappearing during the active navigation when they set the `routeCasingColor` to `.clear`, which is tracked in [slack](https://mapbox.slack.com/archives/C8LQNA2R1/p1635342782119900).

We could see the route line color from `trafficLowColor` the lighter blue turns to `routeCasingColor` as darker blue, gif contributed by @OttyLab in [slack](https://mapbox.slack.com/archives/C8LQNA2R1/p1635752824193700?thread_ts=1635342782.119900&cid=C8LQNA2R1)
![rc6_routeCasingColor](https://user-images.githubusercontent.com/48976398/139735284-b12a7fec-c931-4f68-872c-330f37932d36.gif)

It's caused by the following:
https://github.com/mapbox/mapbox-navigation-ios/pull/3413/commits/3722a0ee1ab47b98144cb65479c083302cfb0f6b#diff-2b1a7feb6a6afb29ecf66effdb70c04c63a6fc30da5499c64d081f9f6bfef42aR246 

The `associatedFeatureColor` set as `routeCasingColor` would have 2 causes:
1. The compiling issue leads to `nil` `CongestionLevel` info. We have observed in an internal dogfooding application and notified by a customer that there're jam / congestion band disappearing suddenly during active navigation. Before `v2.0.0-rc.6`, if we have `nil` for congestion level information, we would assign the `trafficUnknownColor` to it. 
https://github.com/mapbox/mapbox-navigation-ios/blob/3d4db0c89c16721a0631c4ba6f0c76bdbb0fd973/Sources/MapboxNavigation/NavigationMapView%2BVanishingRouteLine.swift#L240-L243
https://github.com/mapbox/mapbox-navigation-ios/blob/3d4db0c89c16721a0631c4ba6f0c76bdbb0fd973/Sources/MapboxNavigation/NavigationMapView%2BVanishingRouteLine.swift#L337
[...]

PR #3577, opened 2021-11-09
Use `trafficUnknownColor` for route line when no congestion level found.

ShanMa1991:
### Description
According to #3542 , the regression cause that when there's no congestion level found in the `RouteLeg`, the route leg would display a default darker blue of `routeCasingColor ` instead of `trafficUnknownColor`. This pr is to fix this issue.

### Implementation
When there's no congestion level found, default show the `trafficUnknownColor` or `alternativeTrafficUnknownColor` on the current leg.

### Screenshots or Gifs
In `Example`, By setting the ` routeLeg.segmentNumericCongestionLevels` and `routeLeg.segmentCongestionLevels` as `nil` before we generate congestion features in `Route.congestionFeatures(legIndex:roadClassesWithOverriddenCongestionLevels:)`, we could get the situation when no congestion level found in the current route. 

Then we have the following before the fix, which would show the darker blue as the `routeCasingColor` on the current active leg during active navigation.
beforeFix

With the fix, we would assign the `trafficUnknownColor` for current active leg during navigation:
fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
 ### Other changes
 
 * Fixed a retain cycle in CarPlay implementation of a navigation map view that prevented `NavigationMapView` instances from being deallocated after CarPlay is stopped. ([#3552](https://github.com/mapbox/mapbox-navigation-ios/pull/3552))
-* Fixed a regression when no congestion level information found, the current active route leg shows `NavigationMapView.routeCasingColor` instead of `NavigationMapView.trafficUnknownColor`. ([#3577](https://github.com/mapbox/mapbox-navigation-ios/pull/3577))
+* Fixed an issue where the entire route line was colored as `NavigationMapView.routeCasingColor` instead of `NavigationMapView.trafficUnknownColor` when traffic congestion data was missing. ([#3577](https://github.com/mapbox/mapbox-navigation-ios/pull/3577))
 
 ## v2.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 ### Other changes
 
 * Fixed a retain cycle in CarPlay implementation of a navigation map view that prevented `NavigationMapView` instances from being deallocated after CarPlay is stopped. ([#3552](https://github.com/mapbox/mapbox-navigation-ios/pull/3552))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 ### Other changes
 
 * Fixed a retain cycle in CarPlay implementation of a navigation map view that prevented `NavigationMapView` instances from being deallocated after CarPlay is stopped. ([#3552](https://github.com/mapbox/mapbox-navigation-ios/pull/3552))
+* Fixed a regression when no congestion level information found, the current active route leg shows `NavigationMapView.routeCasingColor` instead of `NavigationMapView.trafficUnknownColor`. ([#3577](https://github.com/mapbox/mapbox-navigation-ios/pull/3577))
```
1ec5, line 33:
```suggestion
* Fixed an issue where the entire route line was colored as `NavigationMapView.routeCasingColor` instead of `NavigationMapView.trafficUnknownColor` when traffic congestion data was missing. ([#3577](https://github.com/mapbox/mapbox-navigation-ios/pull/3577))
```

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -332,7 +335,7 @@ extension NavigationMapView {
     /**
      Given a congestion level, return its associated color.
      */
-    func congestionColor(for congestionLevel: String?, isMain: Bool) -> UIColor {
+    func congestionColor(for congestionLevel: String? = nil, isMain: Bool) -> UIColor {
```
1ec5, line 338: Change the call site to pass in `nil` instead of providing a default value here. According to the Swift API design guidelines, parameters that have default argument values go at the end of the parameter list, but that change doesn’t seem to be necessary here.

1ec5 (approved):

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
ShanMa1991: Thanks for the suggestion and the reference to the Swift API design guidelines, I've removed the default value here.

PR #3581, opened 2021-11-10
Add ability to use custom rounding mechanism

jill-cardamon:
This PR adds the ability to use a custom rounding mechanism for the remaining distance during CarPlay active guidance using delegate methods.

### DIFF
--- a/Sources/MapboxCoreNavigation/DistanceFormatter.swift
+++ b/Sources/MapboxCoreNavigation/DistanceFormatter.swift
@@ -1,12 +1,38 @@
 import CoreLocation
 import MapboxDirections
 
+/**
+ :nodoc:
+ An object responsible for the rounding behavior of distances according to locale.
+ */
 public struct RoundingTable {
+    /**
+     :nodoc:
+     `Threshold` supplies rounding behavior for a given maximum distance.
+     */
     public struct Threshold {
+        /**
+         :nodoc:
+         The maximum distance that the `Threshold` is applicable.
+         */
         public let maximumDistance: Measurement<UnitLength>
+        
+        /**
+         :nodoc:
+         The increment that a given distance with be rounded to.
+         */
         public let roundingIncrement: Double
+        
+        /**
+         :nodoc:
+         The maximum number of digits following the decimal point.
+         */
         public let maximumFractionDigits: Int
         
+        /**
+         :nodoc:
+         Returns a rounded `Measurement<UnitLength>` for a given distance.
+         */
         public func measurement(of distance: CLLocationDistance) -> Measurement<UnitLength> {
             var measurement = Measurement(value: distance, unit: .meters).converted(to: maximumDistance.unit)
             measurement.value.round(roundingIncrement: roundingIncrement)
@@ -15,6 +41,10 @@ public struct RoundingTable {
         }
     }
     
+    /**
+     :nodoc:
+     An array of `Threshold`s that detail the rounding behavior.
+     */
     public let thresholds: [Threshold]
     
     /**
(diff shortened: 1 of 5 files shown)


S2Ler:
@jill-cardamon Do we have linked issues for this PR?

Sources/MapboxNavigation/CarPlayManagerDelegate.swift
```diff
@@ -169,6 +169,15 @@ public protocol CarPlayManagerDelegate: AnyObject, UnimplementedLogging {
                         to parentViewController: UIViewController,
                         pointAnnotationManager: PointAnnotationManager)
     
+    /**
+     Offers the delegate an opportunity to use a customized rounding mechanism for the remaining distance.
+     
+     - parameter carPlayManager: The CarPlay manager instance.
+     - parameter remainingDistance: The distance remaining to be rounded.
+     - returns: An optional value representing the distance remaining.
+     */
```
S2Ler, line 178: Probably it should be:
> The remaining distance to be rounded
> An optional value representing the remaining distance

jill-cardamon: ah thank you. i will update this comment.

S2Ler (approved):

1ec5:
It turns out that the customer request that led to this PR went a little further in expecting CarPlay to display a decimal distance unconditionally, whereas Apple has made a design decision to truncate trailing zeros. This is a matter for the MeasurementFormatter built into CarPlay, which we have no access to. We employ custom rounding to reduce the number of decimal places, which works reliably, but adding decimal places doesn’t always work for this reason.

Sources/MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -1,26 +1,26 @@
 import CoreLocation
 import MapboxDirections
 
-struct RoundingTable {
-    struct Threshold {
-        let maximumDistance: Measurement<UnitLength>
-        let roundingIncrement: Double
-        let maximumFractionDigits: Int
+public struct RoundingTable {
```
1ec5, line 4: We’re choosing to leave this API undocumented for now because most developers shouldn’t adjust it; usability would likely be impaired with any other values. Still, we could document it for our internal purposes to avoid confusion, hiding the documentation from the published docset using `:nodoc:`.

Sources/MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -1,26 +1,26 @@
...
+        public func measurement(of distance: CLLocationDistance) -> Measurement<UnitLength> {
             var measurement = Measurement(value: distance, unit: .meters).converted(to: maximumDistance.unit)
             measurement.value.round(roundingIncrement: roundingIncrement)
             measurement.value.round(precision: pow(10, Double(maximumFractionDigits)))
             return measurement
         }
     }
     
-    let thresholds: [Threshold]
+    public let thresholds: [Threshold]
```
1ec5, line 18: In order for the developer to have any control over the rounding behavior, the `metric`, `uk`, and `us` static properties below need to be public as well.

jill-cardamon: Couldn't they create extensions like so?

```
extension Measurement where UnitType == UnitLength {
    public func localized() -> Measurement<UnitLength> {
        let threshold: RoundingTable = .custom
        return threshold.threshold(for: distance).measurement(of: distance)
    }
}

extension RoundingTable {
    static var custom: RoundingTable = RoundingTable(thresholds: [
         // thresholds go here
    ])
}
```

1ec5: I think the idea is for developers to be able to change the behavior of existing UI elements, right? If so, they need to hook into whatever is already being called. It isn’t possible to reimplement a method in quite that way; you’d need to swizzle the method, something we wouldn’t want to encourage.

jill-cardamon: Ah I see. Thank you for clarifying.

Issue #3582, opened 2021-11-11
[Bug]: `SpeedLimitView` and `CarPlayCompassView` are not visible on CarPlay if right-to-left mode is used.


MaximAlien:
### Mapbox Navigation SDK version

`v2.1.0-beta.1`

### Steps to reproduce

1. Enable right-to-left mode on CarPlay head-unit.
2. Start active-guidance navigation on CarPlay and observe behavior (most reliably issue can be reproduced in free-way).

### Expected behavior

`SpeedLimitView` and `CarPlayCompassView` should be visible.

### Actual behavior

`SpeedLimitView` and `CarPlayCompassView` are not visible.

### Is this a one-time issue or a repeatable issue?

repeatable

PR #3583, opened 2021-11-11
Fix issue, which was causing inability to see `SpeedLimitView` and `CarPlayCompassView` when right-to-left mode is used on CarPlay.

MaximAlien:
Closing #3582.


bamx23 (approved):
Do we need to add a changelog line do this?

MaximAlien:
> Do we need to add a changelog line do this?

Yep, added.

PR #3599, opened 2021-11-16
Fix rendering issue with route preview stretchable route duration callouts

avi-c:
Maps v10 changed the API for how stretchable image assets are defined for map symbol style layers. This caused some rendering issues with the offsets that worked in Maps v6.3.

The fix is to update the offsets used to declare the stretchable regions and content area of the route duration annotation callouts.

Screenshots of the before and after:

Annotations Error

Annotations Working

### DIFF
--- a/Sources/MapboxNavigation/NavigationMapView.swift
+++ b/Sources/MapboxNavigation/NavigationMapView.swift
@@ -1149,8 +1149,11 @@ open class NavigationMapView: UIView {
         // Right-hand pin
         if let image = Bundle.mapboxNavigation.image(named: "RouteInfoAnnotationRightHanded") {
             // define the "stretchable" areas in the image that will be fitted to the text label
+            // These numbers are the pixel offsets into the PDF image asset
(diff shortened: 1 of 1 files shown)


S2Ler:
@avi-c Thanks for the fix 👍 Can you please share how you found the correct magic numbers for this fix? Can we make a comment in the code to help future us to understand these numbers?

avi-c:
@S2Ler  - Very good questions. The numbers come from finding the offsets into the image asset that contain the stretchable area. The same technique as shown here: https://docs.mapbox.com/mapbox-gl-js/example/add-image-stretchable/

To get this working for our case just too a bit of trial and error of finding the correct values for our specific PDF assets: 
RouteInfoAnnotationLeftHanded.pdf and RouteInfoAnnotationRightHanded.pdf

I added a couple of comments to help for the future.

S2Ler (approved):
🚀

MaximAlien (approved):

PR #3603, opened 2021-11-17
Route Access Restrictions visualisation

Udumft:
This PR adds possibility to draw portions of a route which pass restricted roads.
The overlay is implemented as a separate customizable route line layer

Simulator Screen Shot - iPhone 12 Pro Max - 2021-11-17 at 17 30 44
Simulator Screen Shot - iPhone 12 Pro Max - 2021-11-17 at 17 31 02
.


Tests/MapboxNavigationTests/NavigationMapViewTests.swift
```diff
@@ -47,7 +47,7 @@ class NavigationMapViewTests: TestCase {
         ])
         
         XCTAssertEqual(congestionSegments.count, 1)
-        XCTAssertEqual(congestionSegments[0].0.count, 10)
+        XCTAssertEqual(congestionSegments[0].0.count, 6)
```
Udumft, line 50: This is caused by [the change in merging method](https://github.com/mapbox/mapbox-navigation-ios/pull/3603/files#diff-16778b2708d85c9a45c23688c69727bff4f569fb3f240cf93618380dbe262663R92). Previously it would duplicate bordering points. For example when merging segments with coords `1 -> 2 -> 3` it would result in `[1, 2, 2, 3]`. With change above it would result in `[1,2,3]`.
I didn't notice changes in resulting route lines drawing, but TBH I didn't test it too much. Does anybody knows if such merging was intended.

Sources/MapboxNavigation/Constants.swift
```diff
@@ -25,6 +30,11 @@ public var RouteLineWidthByZoomLevel: [Double: Double] = [
  */
 public var NavigationMapViewMinimumDistanceForOverheadZooming: CLLocationDistance = 200
 
+/**
+ Attribute name for the route line that is used for identifying restricted areas along the route.
+ */
+public let RestrictedRoadClassAttribute = "isRestrictedRoad"
```
Udumft, line 36: There are also few convenience constants exposed to user, like table of route line widths depending on zoom level (`RouteLineWidthByZoomLevel`). The only value for restricted areas display I could think of is `dash array attributes` which may be potentially useful, but I doubt it since if user is implementing his own layers he won't need to align dashing parameters to anything mapbox-provided.

Tests/MapboxNavigationTests/NavigationMapViewTests.swift
```diff
@@ -390,34 +455,78 @@ class NavigationMapViewTests: TestCase {
...
+            composeFeature(0, true),
+            composeFeature(1, false),
+            composeFeature(2, true),
+            composeFeature(3, false)
+        ]
+        
+        var fractionTraveled = 0.0 // not traversed at all
+        var routeLineGradient = navigationMapView.routeLineRestrictionStops(features, fractionTraveled: fractionTraveled)
+        XCTAssertEqual(routeLineGradient[0.0], navigationMapView.routeRestrictedAreaColor)
+        XCTAssertEqual(routeLineGradient[1.0], navigationMapView.traversedRouteColor)
```
Udumft, line 511: After modifications of route line gradients calculation in [this commit](https://github.com/mapbox/mapbox-navigation-ios/commit/c949f78baabf84779b6a1bcbd99cc25fbe9e1db6) this test starts failing. I see that it has changed how ending segment is handled, but it still looks odd to me: In this particular case `routeLineGradient[0.75] == routeRestrictedAreaColor` (0.75 is the border between 2nd restricted segment and last non-restricted segment). Since it is "not soft" gradient, I would expect that `routeLineGradient[0.75.nextUp] == traversedColor`, but that is not the case and in fact it's `routeLineGradient[0.75.nextUp.nextUp] == traversedColor`. I am not sure why there is this little gap and how it appears. 
@ShanMa1991 maybe you could help me with it?

ShanMa1991: Hi, yes, this gap exits. Because right now when we choose `soft`, we expected that the gradient stop of one feature is the `[segmentStartPercentTraveled + stopGap, segmentEndPercentTraveled - stopGap]`. Because the `stopGap` is based on the length of current congestion segment length, which would avoid the out of bound issue between a very long segment and a very short segment.

and when we choose `sharp` gradient, we actually still use the constraints of the current segment as `[segmentStartPercentTraveled.nextUp, segmentEndPercentTraveled.nextDown]`. So for example, for the 3rd segment in this test, the actual end percentage is `0.7500000000000001`, but we record the `0.75` for its end. And for the 4th segment, the actual start percentage is `0.7500000000000001`, because it's different than the 3rd one, we record it with `0.7500000000000002` as the start. Thus there's a `0.0000000000000002` gap between the 3rd segment and the 4th segment in the line gradient stops. It's larger than the previous `0.0000000000000001` but could avoid the check of next segment and the segment length issue.

And for the `0.0000000000000002` gap, because right now when we choose `sharp` gradient, we're using the following Expression:
```
        lineLayer.lineGradient = .expression(
            Exp(.step) {
                Exp(.lineProgress)
                UIColor.blue
                0.75
                UIColor.blue
                0.7500000000000002
                UIColor.red
            }
        )
```
And we have a following result, which still shows a sharp transition in the lowest zoom level.
double

And the edge case is that what if the `fractionTraveled` is in this gap, for example, `0.7500000000000001`.  In this case, the `minimumSegment` would record `(0.75, associatedFeatureColor)`, and then apply to the `gradientStops`. There's still a sharp change between the two segments.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -339,4 +367,19 @@ extension NavigationMapView {
             return isMain ? trafficUnknownColor : alternativeTrafficUnknownColor
         }
     }
+    
+    func routeLineRestrictionStops(_ restrictionFeatures: [Turf.Feature], fractionTraveled: Double) -> [Double: UIColor] {
```
ShanMa1991, line 371: Should it be named similar to the `routeLineCongestionGradient(_:fractionTraveled:isMain:isSoft:)`?

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -34,6 +34,13 @@ open class NavigationMapView: UIView {
      */
     public var showsCongestionForAlternativeRoutes: Bool = false
     
+    /**
+     Controls wheter to show restricted portions of a route line.
+     
+     Restricted areas are drawn using `routeRestrictedAreaColor` which is customizable.
+     */
+    public var showsRestrictedAreasOnRoute: Bool = false
```
ShanMa1991, line 42: Right now in CarPlay, we usually set the route line properties in the `CarPlayManagerDelegate.carPlayManager(_:didPresent:)` through `carPlayManager.carPlayNavigationViewController`. If we set the `showsRestrictedAreasOnRoute` here as `true`, and choose to start navigation from mobile, then the the route line maybe still don't show the restricted area. We may need to allow a setter of `showsRestrictedAreasOnRoute` to trigger the new route line generation and display when it enabled, otherwise it may not work in this certain case.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -392,9 +406,11 @@ open class NavigationMapView: UIView {
         if let legIndex = currentLegIndex {
             let congestionFeatures = route.congestionFeatures(legIndex: legIndex,
                                                               roadClassesWithOverriddenCongestionLevels: roadClassesWithOverriddenCongestionLevels)
-            currentLineGradientStops = routeLineGradient(congestionFeatures,
-                                                         fractionTraveled: fractionTraveled,
-                                                         isSoft: crossfadesCongestionSegments)
+            currentLineGradientStops = routeLineCongestionGradient(congestionFeatures,
+                                                                   fractionTraveled: fractionTraveled,
+                                                                   isSoft: crossfadesCongestionSegments)
+            currentRestrictedAreasStops = routeLineRestrictionStops(route.restrictedRoadsFeatures(),
```
ShanMa1991, line 412: Because we only need the `currentRestrictedAreasStops` when `showsRestrictedAreasOnRoute` enabled, but the `navigationMapView.setUpLineGradientStops(along:)` is called when `routeLineTracksTraversal` enabled. It would be better if we could move it under the check of `showsRestrictedAreasOnRoute` in `navigationMapView.show(_:legIndex:)` and the setter of `showsRestrictedAreasOnRoute`.

Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
Udumft: Renamed

Sources/MapboxNavigation/NavigationMapView.swift
Udumft: Good catch! Corrected

Sources/MapboxNavigation/NavigationMapView.swift
Udumft: You are right, restricted areas were not showing when connecting to CarPlay. I missed it because for tests I just changed the default value for this flag. Thanks for flagging!

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -34,6 +34,19 @@ open class NavigationMapView: UIView {
...
     
+    /**
+     Controls wheter to show restricted portions of a route line.
+     
+     Restricted areas are drawn using `routeRestrictedAreaColor` which is customizable.
+     */
+    public var showsRestrictedAreasOnRoute: Bool = false {
+        didSet {
+            if let routes = self.routes {
+                show(routes, legIndex: self.currentLegIndex)
```
ShanMa1991, line 45: So if we make resume button to switch on/off of `navigationMapView.showsRestrictedAreasOnRoute` during active navigation, we have the following result, because if we call the `navigationMapView.show(_:legIndex)` directly, the previous stored route points and upcoming route point index are cleared. There'll be a 1 second gap of showing the whole route line when `routeLineTracksTraversal` enabled. Until the next `routeProgress` comes in and re-store the upcoming route point index and `fractionTraveled`, could we get the right route line display.
switch

But we could make this as a tail work because right now the CarPlay start with the `showsRestrictedAreasOnRoute` enabled or disabled, The gap is almost non. There's not much case that it would be turned on/off during active navigation.

ShanMa1991 (approved):

Sources/MapboxNavigation/NavigationMapView.swift
Udumft: Added a [ticket](https://github.com/mapbox/mapbox-navigation-ios/issues/3636) for tracking this.

Issue #3604, opened 2021-11-17
[Bug]: Incorrect ornaments padding on CarPlay.


MaximAlien:
### Mapbox Navigation SDK version

`v2.1.0-rc.1`

### Steps to reproduce

1. Open example application on CarPlay.
2. Simulate location updates on the road, so that compass and speed limit views are visible.
3. Observe behavior.

### Expected behavior

Speed limit and compass views have excessive padding. For example:

<img width="500" alt="Screen Shot 2021-11-17 at 11 56 00 AM" src="https://user-images.githubusercontent.com/1496498/142273221-fbe4f061-a3ac-49f8-abec-71265fc08791.png">

### Actual behavior

Speed limit and compass views should have similar padding as `CPMapTemplate` button at the bottom of the screen.

### Is this a one-time issue or a repeatable issue?

repeatable

PR #3605, opened 2021-11-17
Fix excessive padding on CarPlay during active navigation.

MaximAlien:
Closing #3604.

Since there is no ability to detect whether CarPlay head-unit is in either left of right driving side mode (at least I was not able to find it, feel free to provide any suggestions if you have any) I had to use `safeAreaInsets` value as a criteria.

### DIFF
--- a/Sources/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/Sources/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -408,7 +408,15 @@ open class CarPlayNavigationViewController: UIViewController {
         }
     }
     
-    open override func updateViewConstraints() {
+    public override func viewSafeAreaInsetsDidChange() {
+        super.viewSafeAreaInsetsDidChange()
+        
+        // Trigger update of view constraints to correctly position views like `SpeedLimitView` and
+        // `CarPlayCompassView`.
+        view.setNeedsUpdateConstraints()
+    }
+    
+    public override func updateViewConstraints() {
         // Since there is no ability to detect current driving side mode of the CarPlay head-unit,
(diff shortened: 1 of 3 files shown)


ShanMa1991 (approved):

ShanMa1991:
It's great if we could add this to `changeLog`.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -395,6 +408,40 @@ open class CarPlayNavigationViewController: UIViewController {
...
+    public override func viewSafeAreaInsetsDidChange() {
+        super.viewSafeAreaInsetsDidChange()
+        
+        // Trigger update of view constraints to correctly position views like `SpeedLimitView` and
+        // `CarPlayCompassView`.
+        view.setNeedsUpdateConstraints()
+    }
+    
+    public override func updateViewConstraints() {
+        // Since there is no ability to detect current driving side mode of the CarPlay head-unit,
```
1ec5, line 420: Does `UIApplication.shared.userInterfaceLayoutDirection` reflect the driving side? CarPlay ostensibly implements the driving side adaptation feature using RTL language support.

MaximAlien: No matter what driving side I choose in CarPlay head-unit options I always get `.leftToRight` in `UIUserInterfaceLayoutDirection`. It's interesting that `Sony XAV-AX1000` doesn't even provide changing driving side. With `Intellidash Pro` it's possible, so I suspect that head-unit support this option only by drawing certain menus on the right.

1ec5: (For posterity, it does not.)

PR #3610, opened 2021-11-18
Remove the usage of lineString.trimmed(from:to:) due to high calculation

ShanMa1991:
This pr is to remove the usage of the `LineString.trimmed(from:to:)` in the calculation of the upcoming route point index, because it's too large in the long step.

Previously we were using the `currentStepProgress.step.distance` as the `stopDistance` of `LineString.trimmed(from:to:)`, which is sometimes larger than the total distance of the current step lineString.  In this case. We only need to find the start index, and then calculate the remaining coordinates count in the line string from the start index, instead of iterating over the lineString till the end.


Sources/MapboxNavigation/NavigationMapView+VanishingRouteLine.swift
```diff
@@ -63,10 +62,11 @@ extension NavigationMapView {
          Find the count of remaining points in the current step.
          */
         let lineString = currentStepProgress.step.shape ?? LineString([])
-        if currentStepProgress.distanceTraveled < 0 {
-            allRemainingPoints += currentLegSteps[currentLegProgress.stepIndex].count
-        } else if let slicedLineString = lineString.trimmed(from: currentStepProgress.distanceTraveled, to: currentStepProgress.step.distance) {
-            allRemainingPoints += slicedLineString.coordinates.dropLast().count
+        if currentStepProgress.distanceTraveled <= 0 {
```
MaximAlien, line 65: Since we're not using `LineString.trimmed(from:to:)` anymore does it make sense to remove Turf from explicit dependencies?

ShanMa1991: Sure, I'll remove the Turf from explicit dependencies.

Package.swift
```diff
@@ -32,7 +32,6 @@ let package = Package(
         .package(name: "Nimble", url: "https://github.com/Quick/Nimble.git", from: "9.0.1"),
         .package(name: "SnapshotTesting", url: "https://github.com/pointfreeco/swift-snapshot-testing.git", from: "1.8.1"),
         .package(name: "OHHTTPStubs", url: "https://github.com/AliSoftware/OHHTTPStubs.git", from: "9.1.0"),
-        .package(name: "Turf", url: "https://github.com/mapbox/turf-swift.git", from: "2.1.0"),
```
MaximAlien, line 35: Please also update `Package.resolved` by running $ swift package update.

MaximAlien (approved):

Package.swift
ShanMa1991: Because `MapboxDirections` could link to the `Turf` 2.1.0, the `swift package update` won't bring down the version of `Turf` to 2.0.0

PR #3613, opened 2021-11-22
Fix style change caused route line update delay.

### LINKED ISSUE #3612: [Bug]: CarPlay shows the route line blinking when drive through tunnel
According to the [comment](https://github.com/mapbox/mapbox-navigation-ios/pull/3466#issuecomment-972176049), when user drives in and off the tunnel, the route line blinks.

It's because when the style changes in CarPlay, it would need to update the route line.
https://github.com/mapbox/mapbox-navigation-ios/blob/992b4b5f4d63cb1a3623bef95e4bc6dd0e7d7158/Sources/MapboxNavigation/CarPlayNavigationViewController.swift#L453-L458

But when the vanishing effect enabled, it doesn't update the upcoming route point index. So until the next routeProgress comes in, it would update the `fractionTraveled`.
https://github.com/mapbox/mapbox-navigation-ios/blob/992b4b5f4d63cb1a3623bef95e4bc6dd0e7d7158/Sources/MapboxNavigation/CarPlayNavigationViewController.swift#L676-L684

ShanMa1991:
### Description
This pr is to fix #3612 by packing the route line update methods into the `func updateRouteLine(routeProgress:coordinate:)` through `CarPlayNavigationViewController` and `RouteLineController`, to reduce code duplication and avoid the route line update delay.

### Implementation
Previously when style changes or when reroute event happens in CarPlay, the `CarPlayNavigationViewController` only shows the route line without updating the upcoming route point index, which cause a 1 second delay and the route line blinking. This pr is to pack the route line update methods into `CarPlayNavigationViewController.updateRouteLine(routeProgress:coordinate:)` and `RouteLineController.updateRouteLinee(routeProgress:coordinate:)`. So when the style changes, rerouting and refresh event happens, `mobile` and `CarPlay` will it directly to update the route line without delay. 

### Screenshots or Gifs
Choose the location as following to start the route with driving through tunnels:
```
latitude: 37.789812
longitude: -122.470759
```
styleChange

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * Fixed an issue where `ReplayLocationManager` didn't update location timestamps when a new loop started. ([#3550](https://github.com/mapbox/mapbox-navigation-ios/pull/3550))
 * Fixed the background location update issue during active navigation when using default `.courseView` for `NavigationMapView.userLocationStyle`. ([#3533](https://github.com/mapbox/mapbox-navigation-ios/pull/3533))
 * Fixed an issue where `UserPuckCourseView` is trimmed when using custom frame for `UserLocationStyle.courseView(_:)`. ([#3601](https://github.com/mapbox/mapbox-navigation-ios/pull/3601))
+* Fixed an issue where route line blinks when users drive through tunnel. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
 
 ### Banners and guidance instructions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 * Fixed an issue where `ReplayLocationManager` didn't update location timestamps when a new loop started. ([#3550](https://github.com/mapbox/mapbox-navigation-ios/pull/3550))
 * Fixed the background location update issue during active navigation when using default `.courseView` for `NavigationMapView.userLocationStyle`. ([#3533](https://github.com/mapbox/mapbox-navigation-ios/pull/3533))
 * Fixed an issue where `UserPuckCourseView` is trimmed when using custom frame for `UserLocationStyle.courseView(_:)`. ([#3601](https://github.com/mapbox/mapbox-navigation-ios/pull/3601))
-* Fixed an issue where route line blinks when users drive through tunnel. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
+* Fixed an issue where route line blinks when style is changed during active navigation. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
 
 ### Banners and guidance instructions
 
--- a/Sources/MapboxNavigation/CarPlayNavigationViewController.swift
+++ b/Sources/MapboxNavigation/CarPlayNavigationViewController.swift
@@ -640,7 +640,7 @@ open class CarPlayNavigationViewController: UIViewController, BuildingHighlighti
     }
     
     @objc func refresh(_ notification: NSNotification) {
-        updateRouteLine(routeProgress: navigationService.routeProgress, coordinate: navigationService.router.location?.coordinate)
+        navigationMapView?.updateRouteLine(routeProgress: navigationService.routeProgress, coordinate: navigationService.router.location?.coordinate)
     }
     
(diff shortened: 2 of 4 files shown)


Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -692,10 +684,21 @@ open class CarPlayNavigationViewController: UIViewController, BuildingHighlighti
         let nextStep = progress.currentLegProgress.stepIndex + 1
         
         navigationMapView?.addArrow(route: progress.route, legIndex: legIndex, stepIndex: nextStep)
-        navigationMapView?.show([progress.route], legIndex: legIndex)
+        updateRouteLine(routeProgress: progress, coordinate: navigationService.router.location?.coordinate)
```
ShanMa1991, line 687: Avoid the usage of `navigationMapView.show(::legIndex:)` during active navigation in `mobile` and `CarPlay`, because we may have the `routeLineTracksTraversal` turned on. Instead, packing the route line update to `updateRouteLine(routeProgress:coordinate:)` to avoid this issue.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * Fixed an issue where `ReplayLocationManager` didn't update location timestamps when a new loop started. ([#3550](https://github.com/mapbox/mapbox-navigation-ios/pull/3550))
 * Fixed the background location update issue during active navigation when using default `.courseView` for `NavigationMapView.userLocationStyle`. ([#3533](https://github.com/mapbox/mapbox-navigation-ios/pull/3533))
 * Fixed an issue where `UserPuckCourseView` is trimmed when using custom frame for `UserLocationStyle.courseView(_:)`. ([#3601](https://github.com/mapbox/mapbox-navigation-ios/pull/3601))
+* Fixed an issue where route line blinks when users drive through tunnel. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
```
MaximAlien, line 24:
```suggestion
* Fixed an issue where route line blinks when style is changed during active navigation. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
```

Sources/MapboxNavigation/RouteLineController.swift
```diff
@@ -153,12 +148,19 @@ extension NavigationMapView {
             }
         }
         
-        func navigationService(_ service: NavigationService, didRefresh routeProgress: RouteProgress) {
+        private func updateRouteLine(routeProgress: RouteProgress, coordinate: CLLocationCoordinate2D?) {
```
MaximAlien, line 151: `updateRouteLine(routeProgress:coordinate:)` is duplicated for iOS and CarPlay, do you think it'll be possible to move it to common place, so that it can be re-used? For example `NavigationMapView`.

MaximAlien: Please also add basic documentation for this method to describe its purpose.

Sources/MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -692,10 +684,21 @@ open class CarPlayNavigationViewController: UIViewController, BuildingHighlighti
...
         navigationMapView?.showWaypoints(on: progress.route, legIndex: legIndex)
     }
     
+    func updateRouteLine(routeProgress: RouteProgress, coordinate: CLLocationCoordinate2D?) {
+        navigationMapView?.show([routeProgress.route], legIndex: routeProgress.legIndex)
+        if routeLineTracksTraversal {
```
MaximAlien, line 693: It's a matter of style, but it's easier to read code when we're using early return, up to you though:
```suggestion
    if !routeLineTracksTraversal {
        return
    }
```

MaximAlien (approved):

Sources/MapboxNavigation/RouteLineController.swift
ShanMa1991: Thanks for the suggestion, I'll add documentation and move this method to `NavigationMapView` to avoid duplication.

CHANGELOG.md
ShanMa1991: Thanks for the suggestion, I'll improve the changeLog entry.

Issue #3479, opened 2021-10-13
[Bug]: Navigation camera transition between following and overview modes is not smooth.


MaximAlien:
### Mapbox Navigation SDK version

main

### Steps to reproduce

1. Start active-guidance navigation.
2. Several times switch between overview and following mode.
3. Observe behavior.

### Expected behavior

Transitions should be smooth (without abrupt cancellation, reverted animation etc).

### Actual behavior

Transitions are not smooth, for example:

https://user-images.githubusercontent.com/1496498/137217647-e03b6a95-87e1-4b08-96e2-ecc60625439e.mp4

### Is this a one-time issue or a repeatable issue?

repeatable

MaximAlien:
Fix was added on Maps SDK side in https://github.com/mapbox/mapbox-maps-ios/pull/789, which will be part of `v10.2.0` release. Navigation SDK will pick it up in `v2.2`, hence updating the milestone.

PR #3614, opened 2021-11-23
Update dependencies.

MaximAlien:
Update Mapbox Maps to `v10.2.0-rc.1`, Navigation Native to `v81.0.0` and Common to `v21.0.0-rc.1`.

Also closing https://github.com/mapbox/mapbox-navigation-ios/issues/3479.

At this point there is a regression related to showing location indicator. It's being tracked here: https://github.com/mapbox/mapbox-maps-ios/issues/886.

Currently failing tests:
- ~~testRerouteAfterArrival~~
- testPausedFreeDrivePausesNavNativeNavigator
- testRouteSnappingOvershooting
- ~~testArrive~~
- ~~testPuck3DLayerPosition~~
- ~~testRunningAnimators~~
- ~~testTripPerWaypoint~~


MaximAlien:
Update is not possible at this time due to:
```
error: Dependencies could not be resolved because root depends on 'mapbox-navigation-native-ios' 80.0.0..<81.0.0 and root depends on 'mapbox-maps-ios' 10.2.0-beta.1.
'mapbox-maps-ios' is incompatible with 'mapbox-navigation-native-ios' because 'mapbox-maps-ios' >= 10.2.0-beta.1 depends on 'mapbox-common-ios' 21.0.0-rc.1 and 'mapbox-navigation-native-ios' >= 80.0.0 depends on 'mapbox-common-ios' 20.1.0..<21.0.0.
```

Waiting for the latest version of Mapbox Navigation Native to proceed further.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -7,26 +7,21 @@ import os.log
 @_implementationOnly import MapboxCommon_Private
 
 /// BillingService from MapboxCommon
-private typealias BillingServiceNative = MapboxCommon_Private.BillingService
+private typealias BillingServiceNative = MapboxCommon_Private.BillingServiceFactory
```
MaximAlien, line 10: /cc @S2Ler, there are a couple of billing related changes.

S2Ler: Looks good.

Sources/MapboxCoreNavigation/BillingHandler.swift
```diff
@@ -76,48 +71,48 @@ private final class ProductionBillingService: BillingService {
     }
 
     func getSKUTokenIfValid(for sessionType: BillingHandler.SessionType) -> String {
-        MapboxCommon_Private.BillingService.getSessionSKUTokenIfValid(for: tripSku(for: sessionType))
+        BillingServiceNative.getInstance().getSessionSKUTokenIfValid(for: tripSku(for: sessionType))
```
S2Ler, line 74: nit: I would extract `BillingServiceNative.getInstance()` into a computed property. 

```swift
var nativeBillingService: <WHAT TYPE HERE?> {
    BillingServiceNative.getInstance()
}
```

azarovalex (approved):
deps bump is ok

Cartfile
```diff
@@ -1,4 +1,4 @@
-binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" ~> 20.1.0
-binary "https://api.mapbox.com/downloads/v2/carthage/mobile-navigation-native/MapboxNavigationNative.xcframework.json" ~> 80.0
+binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" == 21.0.0-rc.1
```
1ec5, line 1: Pinning this version is unfortunate but necessary for consistency [with SPM](https://github.com/mapbox/mapbox-navigation-native-ios/blob/beb14b30bdb3bcde32544756f1867d82b6d758bf/Package.swift#L8) and CocoaPods. But we’ll need to make sure to go back to a tadpole operator once MapboxCommon v21.0.0 is released and MapboxNavigationNative depends on it.

Sources/MapboxCoreNavigation/BillingHandler.swift
1ec5: Nit: name it `shared`. 🙂

MaximAlien: Introduced `NativeBillingService.shared` for this purpose.

MaximAlien: I did some tests and see that `BillingServiceNative.getInstance()` somehow caches its instances (they're sometimes different, but still reused). Hopefully using only one shared instance won't cause any problems.

PR #3616, opened 2021-11-24
Fix issues in zoom calculations for cases when maneuver is further away from the current location.

MaximAlien:
Here are a few visual examples:

`IntersectionDensity.averageDistanceMultiplier = 100.0` (original behavior, before applying fix and regardless of value, which was set)

Screen Shot 2021-11-24 at 11 51 23 AM

`IntersectionDensity.averageDistanceMultiplier = 20.0`

Screen Shot 2021-11-24 at 2 54 51 PM

`IntersectionDensity.averageDistanceMultiplier = 7.0` (default value)

Screen Shot 2021-11-24 at 3 11 29 PM

### DIFF
--- a/Sources/MapboxNavigation/NavigationViewportDataSource.swift
+++ b/Sources/MapboxNavigation/NavigationViewportDataSource.swift
@@ -229,7 +229,7 @@ public class NavigationViewportDataSource: ViewportDataSource {
             let coordinatesToManeuver = routeProgress.currentLegProgress.currentStep.shape?.coordinates.sliced(from: location.coordinate) ?? []
             
             if options.followingCameraOptions.centerUpdatesAllowed {
-                var center: CLLocationCoordinate2D = location.coordinate
+                var center = location.coordinate
                 if let boundingBox = BoundingBox(from: coordinatesToManeuver + coordinatesForManeuverFraming) {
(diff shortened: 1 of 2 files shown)


Sources/MapboxNavigation/NavigationViewportDataSource.swift
```diff
@@ -476,8 +493,8 @@ public class NavigationViewportDataSource: ViewportDataSource {
                         return intersection.distance(to: intersectionLocations[index])
                     })
                     let filteredIntersectionDistances = intersectionDensity.enabled
-                        ? intersectionDistances.filter { $0 > intersectionDensity.minimumDistanceBetweenIntersections }
-                        : intersectionDistances
+                    ? intersectionDistances.filter { $0 > intersectionDensity.minimumDistanceBetweenIntersections }
```
S2Ler, line 496: nit: I find the former variant easier to read.

S2Ler (approved):

Sources/MapboxNavigation/NavigationViewportDataSource.swift
MaximAlien: Agree, but Xcode considers this as a correct indentation.

PR #3619, opened 2021-11-29
Cocoapods CI Job

Udumft:
Bumped Xcode version for CocoaPods CI jobs as it used a deprecated one.


S2Ler:
With this change we don't have a single job that checks that project builds on Xcode 12.4. We shall consider dropping support for Xcode 12.4 or add a nightly build with it.

S2Ler (approved):

PR #3620, opened 2021-11-29
Respect heading while walking

### LINKED ISSUE #1942: Rotate map based on heading while walking

jill-cardamon:
This PR fixes #1942 by reworking #2215.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
 * Fixed an issue where route line blinks when style is changed during active navigation. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
 * Fixed an issue where route line missing traffic colors after refresh or rerouting. ([#3622](https://github.com/mapbox/mapbox-navigation-ios/pull/3622))
 * Fixed an issue when user goes offline and the route line grows back when `NavigationViewController.routeLineTracksTraversal` enabled. When the distance of user location to the route is larger than certain distance threshold, the vanishing effect of route line would stop until the new route line gets generated. ([#3385](https://github.com/mapbox/mapbox-navigation-ios/pull/3385))
+* If the user walks away from the route, they may be rerouted onto a route that initially travels in the opposite direction. This is only the case along steps that require walking on foot. ([#3620](https://github.com/mapbox/mapbox-navigation-ios/pull/3620))
+* Added `RouteControllerNotificationUserInfoKey.headingKey` to the user info dictionary of `Notification.Name.routeControllerWillReroute`, `Notification.Name.routeControllerDidReroute`, and `Notification.Name.routeControllerProgressDidChange` notifications. ([#3620](https://github.com/mapbox/mapbox-navigation-ios/pull/3620))
+* Added a `Router.heading` property that may contain a heading from the location manager. ([#3620](https://github.com/mapbox/mapbox-navigation-ios/pull/3620))
 
 ### Banners and guidance instructions
 
@@ -58,6 +61,7 @@
 ### Camera
 
 * Fixed an issue where on routes with large distances between current location and next manuever camera zoom level was too low. To control navigation camera zoom level use `IntersectionDensity.averageDistanceMultiplier` coefficient. ([#3616](https://github.com/mapbox/mapbox-navigation-ios/pull/3616))
+* While the user is walking, the map rotates according to the user’s heading instead of their course. ([#3620](https://github.com/mapbox/mapbox-navigation-ios/pull/3620))
 
 ### Other changes
 
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -223,6 +223,10 @@ open class RouteController: NSObject {
         }
     }
     
+    public func locationManager(_ manager: CLLocationManager, didUpdateHeading newHeading: CLHeading) {
+        heading = newHeading
+    }
+    
     /**
      Asynchronously updates NavNative navigator with the new `RouteProgress`.
 
@@ -474,9 +478,6 @@ open class RouteController: NSObject {
             .rawLocationKey: rawLocation, // raw
         ]
         userInfo[.headingKey] = heading
-        
-        // to do: DETERMINE IF THIS IS NEEDED
-        userInfo[.rawHeadingKey] = heading // heading snapping not yet implemented (an unnecessary?)
         
         NotificationCenter.default.post(name: .routeControllerProgressDidChange, object: self, userInfo: userInfo)
     }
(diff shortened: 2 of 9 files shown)


Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -464,11 +471,15 @@ open class RouteController: NSObject {
...
-            NotificationUserInfoKey.locationKey: location, //guaranteed value
-            NotificationUserInfoKey.rawLocationKey: rawLocation, //raw
-        ])
+        
+        var userInfo: [RouteController.NotificationUserInfoKey: Any] = [
+            .routeProgressKey: progress,
+            .locationKey: location, // guaranteed value
+            .rawLocationKey: rawLocation, // raw
+        ]
+        userInfo[.headingKey] = heading
```
1ec5, line 480: Above, there’s a `rawHeading` constant too, but it isn’t set here. Does this code need to distinguish between a raw heading set by Core Location and one that’s somehow cleaned up by MapboxNavigationNative? My understanding is that MapboxNavigationNative only cleans up courses, not headings. However, if we were to introduce that functionality at some point, changing the behavior of `headingKey` could be considered backwards-incompatible.

/ref https://github.com/mapbox/mapbox-navigation-ios/pull/2215#discussion_r328379605

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -350,16 +355,20 @@ extension InternalRouter where Self: Router {
...
+        NotificationCenter.default.post(name: .routeControllerWillReroute, object: self, userInfo: userInfo)
     }
     
     func announce(reroute newRoute: Route, at location: CLLocation?, proactive: Bool) {
         var userInfo = [RouteController.NotificationUserInfoKey: Any]()
-        if let location = location {
-            userInfo[.locationKey] = location
-        }
+        // to do: WAS IT OKAY TO REMOVE THIS IF LET?
+        userInfo[.locationKey] = location
```
1ec5, line 370: Yes, that’s fine. If `location` is `nil`, it removes the `locationKey` entry from the dictionary. Sometimes that Swift behavior is surprising, but here the dictionary was empty to begin with.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -663,9 +663,10 @@ open class NavigationMapView: UIView {
      Updates `UserLocationStyle` to provided location.
      
      - parameter location: Location, where `UserLocationStyle` should be shown.
+     - parameter heading: Optional heading, which allows camera to move using heading instead of bearing for walking directions.
      - parameter animated: Property, which determines whether `UserLocationStyle` transition to new location will be animated.
      */
-    public func moveUserLocation(to location: CLLocation, animated: Bool = false) {
+    public func moveUserLocation(to location: CLLocation, with heading: CLHeading? = nil, animated: Bool = false) {
```
1ec5, line 669: Change the `with` label to a more descriptive `facing` or just leave it as the default `heading`.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -663,9 +663,10 @@ open class NavigationMapView: UIView {
      Updates `UserLocationStyle` to provided location.
      
      - parameter location: Location, where `UserLocationStyle` should be shown.
+     - parameter heading: Optional heading, which allows camera to move using heading instead of bearing for walking directions.
```
1ec5, line 666: “Bearing” is an Android-centric term we have to use internally because of MapboxNavigationNative and to some extent the map SDK, but it’s confusing to anyone who works directly with Core Location.

```suggestion
     - parameter heading: Optional heading, which allows camera to move using heading instead of course for walking directions.
```

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -690,6 +691,16 @@ open class NavigationMapView: UIView {
         }
         
         let cameraOptions = CameraOptions(cameraState: mapView.cameraState)
+        
+        if let heading = heading {
+            userCourseView.update(location: location,
+                                  pitch: cameraOptions.pitch!,
+                                  direction: heading.trueHeading,
```
1ec5, line 698: If `trueHeading` is negative, fall back to `magneticHeading`, and if that’s also negative, fall back to `course`.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -690,6 +691,16 @@ open class NavigationMapView: UIView {
...
+                                  animated: animated,
+                                  navigationCameraState: navigationCamera.state)
+            return
+        }
+        
         userCourseView.update(location: location,
                               pitch: cameraOptions.pitch!,
                               direction: cameraOptions.bearing!,
```
1ec5, line 706: By this point, `cameraOptions.bearing` is already set to something match the course. Where does that happen? Should that code account for heading while walking instead of the conditional above?

1ec5 (commented):
Remember to update the changelog. You can crib from the blurbs that were in #2215.

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@
 * Fixed an issue where route line blinks when style is changed during active navigation. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
 * Fixed an issue where route line missing traffic colors after refresh or rerouting. ([#3622](https://github.com/mapbox/mapbox-navigation-ios/pull/3622))
 * Fixed an issue when user goes offline and the route line grows back when `NavigationViewController.routeLineTracksTraversal` enabled. When the distance of user location to the route is larger than certain distance threshold, the vanishing effect of route line would stop until the new route line gets generated. ([#3385](https://github.com/mapbox/mapbox-navigation-ios/pull/3385))
+* If the user walks away from the route, they may be rerouted onto a route that initially travels in the opposite direction. This is only the case along steps that require walking on foot. ([#3620](https://github.com/mapbox/mapbox-navigation-ios/pull/3620))
```
1ec5, line 27: As tail work, we should consider not suppressing the initial “Head north on” instruction after a reroute if it starts out going in the opposite direction.

1ec5 (approved):

CHANGELOG.md
jill-cardamon: Agreed. Will cut ticket now :)

S2Ler (approved):

CHANGELOG.md
jill-cardamon: Ticket is here: #3627.

Issue #3621, opened 2021-11-29
[Bug]: Route line blinks when refresh and reroute


ShanMa1991:
### Mapbox Navigation SDK version

v2.1.0-rc.2

### Steps to reproduce

1. Set the `RouteControllerProactiveReroutingInterval` to `5` or smaller `3` in Example
2. Start navigation in `basic` mode.

### Expected behavior

The route line doesn't blink

### Actual behavior

The route line blinks.
![refresh](https://user-images.githubusercontent.com/48976398/143946311-97d0b367-f6a1-475c-945f-2eb7a6629893.gif)

This issue is observed after the merge of #3261

### Is this a one-time issue or a repeatable issue?

repeatable

ShanMa1991:
Hi @Udumft would you take a look at this issue ?

Udumft:
Looks like this blink is caused by [overwriting route](https://github.com/mapbox/mapbox-navigation-ios/blob/main/Sources/MapboxCoreNavigation/RouteProgress.swift#L137) during refreshing. Before introducing HybridRouter, it just [updated leg's annotations](https://github.com/mapbox/mapbox-navigation-ios/blob/main/Sources/MapboxCoreNavigation/RouteProgress.swift#L154) but HybridRouter does not return `RefreshedRoute` anymore, so we'll have to allow another way to update it. This will require [such API](https://github.com/mapbox/mapbox-directions-swift/pull/634) on Directions.swift.

PR #3622, opened 2021-11-29
Fix the congestion color missing issue of route line when refresh.

### LINKED ISSUE #3542: Route line shows `routeCasingColor` instead of associated congestion color.

ShanMa1991:
### Description
This or is to fix #3542 by changing the coordinates count issue brought by https://github.com/mapbox/mapbox-directions-swift/pull/587 in generating congesting features of route line.

### Implementation
https://github.com/mapbox/mapbox-directions-swift/pull/587 added one more coordinate in the final step of each route leg, it would lead to the `resolvedCongestionLevels` count larger than the coordinates count as in https://github.com/mapbox/mapbox-navigation-ios/issues/3542#issuecomment-972380046 . This pr is to fix this issue by changing the count constraint and to allow the congestion feature generation.

Much of the work is from #3603 

### Screenshots or Gifs
The refresh rate is set to 5, and we could see that the route line doesn't lost its traffic color after refresh as following.
The route line blinks is #3621 .

refresh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Fixed the background location update issue during active navigation when using default `.courseView` for `NavigationMapView.userLocationStyle`. ([#3533](https://github.com/mapbox/mapbox-navigation-ios/pull/3533))
 * Fixed an issue where `UserPuckCourseView` is trimmed when using custom frame for `UserLocationStyle.courseView(_:)`. ([#3601](https://github.com/mapbox/mapbox-navigation-ios/pull/3601))
 * Fixed an issue where route line blinks when style is changed during active navigation. ([#3613](https://github.com/mapbox/mapbox-navigation-ios/pull/3613))
+* Fixed an issue where route line missing traffic colors after refresh or rerouting. ([#3622](https://github.com/mapbox/mapbox-navigation-ios/pull/3622))
 
 ### Banners and guidance instructions
 
--- a/Sources/MapboxNavigation/Array.swift
+++ b/Sources/MapboxNavigation/Array.swift
@@ -89,7 +89,7 @@ extension Array where Iterator.Element == CLLocationCoordinate2D {
             }
             
             if segments.last?.1 == overriddenCongestionLevel {
-                segments[segments.count - 1].0 += coordinates
+                segments[segments.count - 1].0 += [firstSegment.1]
             } else {
                 segments.append((coordinates, overriddenCongestionLevel))
             }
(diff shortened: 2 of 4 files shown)


Udumft (approved):
LGTM

PR #3632, opened 2021-12-03
Updating `main` branch with `release-v2.2` changes

azarovalex:
Updating main branch with changes after the v2.2.0-alpha.1 release


MaximAlien:
I did not notice that base branch in https://github.com/mapbox/mapbox-navigation-ios/pull/3631 was `release-v2.2`. If I'm not mistaken releases for alpha/beta versions can go directly to `main`, we do a branch cut only after first release candidate. 

I guess current approach also works, but it'll involve a bit more work, as `release-v2.2` will have to be kept in sync with `main`. I'd rather submit all changes to `main` directly for now.

azarovalex:
Yeah, you are right @MaximAlien, I forgot that we cut this branch only for rc, as described here:
https://docs.google.com/document/d/1cxPp_ZlmNSgNECyGO6i-Bb87oTEAVRqcYi9o_5o1lBE/edit#heading=h.6g2smdxyzv1i

I will delete the branch, but we still need to merge this PR

MaximAlien (approved):

Issue #3641, opened 2021-12-08
[Bug]: Re-routing sound playback stops music playback in other applications. 


MaximAlien:
### Mapbox Navigation SDK version

`v2.1.0-rc.2`

### Steps to reproduce

1. Start music playback in application of choice (e.g. Shazam, Apple Music).
2. Try to request route, which is likely to be not optimal and re-route will occur.
3. Observe behavior.

To be able to easily reproduce this issue it makes sense to implement `NavigationViewControllerDelegate.navigationViewController(_:shouldRerouteFrom:))` and modify code internally in `update(to:)` to always re-route:
```swift
let willReroute = delegate?.router(self, shouldRerouteFrom: location) ?? false
```

To continuously playback voice instruction, it's possible to modify `RouteController.updateSpokenInstructionProgress(status:willReroute:)` to always use the same voice instruction. For example:

```swift
func updateSpokenInstructionProgress(status: NavigationStatus, willReRoute: Bool) {
    announcePassage(of: routeProgress.currentLegProgress.currentStepProgress.step.instructionsSpokenAlongStep![0], routeProgress: routeProgress)
}
```

### Expected behavior

It is expected that volume of music playback in different application goes down when voice instruction and re-route sounds are played and returns back to normal when Navigation SDK voice instruction/re-route sound finishes.

### Actual behavior

Music playback in another application completely stops.

### Is this a one-time issue or a repeatable issue?

repeatable

PR #3642, opened 2021-12-08
Fix playback issues, while playing music, re-routing sound and voice instruction at the same time.

MaximAlien:
Remove tryMixAudio() in favor of tryDuckAudio() to prevent voice instructions and music playback issues.

Closing #3641.


Sources/MapboxNavigation/RouteVoiceController.swift
```diff
@@ -62,8 +62,12 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
      */
     public init(navigationService: NavigationService, speechSynthesizer: SpeechSynthesizing? = nil, accessToken: String? = nil, host: String? = nil) {
         self.speechSynthesizer = speechSynthesizer ?? MultiplexedSpeechSynthesizer(accessToken: accessToken, host: host)
+        rerouteSoundPlayer = try! AVAudioPlayer(data: NSDataAsset(name: "reroute-sound", bundle: .mapboxNavigation)!.data,
```
bamx23, line 65: Why does it double the existing initialization below?

MaximAlien: It's not needed actually, I'd like to have it in initializer only. Updated code.

Sources/MapboxNavigation/RouteVoiceController.swift
```diff
@@ -107,23 +111,6 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
-            try tryMixAudio()
-        } catch {
-            let wrapped = SpeechError.unableToControlAudio(instruction: instruction, action: .mix, underlying: error)
-            failure(wrapped)
-            return
-        }
-    }
-    
-    func tryMixAudio() throws {
-        guard speechSynthesizer.managesAudioSession else { return }
```
S2Ler, line 121: This should be taken into account or new just break apps audio session without anyway to interfere.

1ec5 (approved):
Remember to add a changelog entry describing the scenarios where the ducking/mixing issues are expected to improve.

PR #3643, opened 2021-12-09
Improve landscape layout during active navigation.

### LINKED ISSUE #1139: Improve layout in landscape orientation

MaximAlien:
PR improves layout in the landscape mode by minimizing space top and bottom banner use. Examples:

Simulator Screen Shot - iPod touch (7th generation) - 2022-01-19 at 21 46 02

Simulator Screen Shot - iPod touch (7th generation) - 2022-01-19 at 21 45 45

Simulator Screen Shot - iPhone 12 mini - 2022-01-19 at 21 43 21

Fixes #1139.

/cc @lawsonkight @abhishek1508

### DIFF
--- a/Sources/MapboxNavigation/CameraController.swift
+++ b/Sources/MapboxNavigation/CameraController.swift
@@ -134,8 +134,23 @@ class CameraController: NavigationComponent, NavigationComponentDelegate {
         let courseViewMinimumInsets = UIEdgeInsets(top: 75.0, left: 75.0, bottom: 75.0, right: 75.0)
         var insets = navigationMapView.mapView.safeArea
         insets += courseViewMinimumInsets
-        insets.top += topBannerContainerView.bounds.height
-        insets.bottom += bottomBannerContainerView.bounds.height + 10.0
+        
+        switch navigationViewData.navigationView.traitCollection.verticalSizeClass {
+        case .unspecified:
+            fallthrough
+        case .regular:
+            insets.top += topBannerContainerView.bounds.height
+            insets.bottom += bottomBannerContainerView.bounds.height + 10.0
+        case .compact:
+            let inset = navigationViewData.navigationView.topBannerContainerView.frame.width + 10.0
+            if UIApplication.shared.userInterfaceLayoutDirection == .rightToLeft {
+                insets.right += inset
+            } else {
+                insets.left += inset
+            }
+        @unknown default:
+            break
+        }
     
         return insets
     }
--- a/Sources/MapboxNavigation/NavigationView.swift
+++ b/Sources/MapboxNavigation/NavigationView.swift
@@ -51,6 +51,9 @@ open class NavigationView: UIView {
         static let feedback = UIImage(named: "feedback", in: .mapboxNavigation, compatibleWith: nil)!.withRenderingMode(.alwaysTemplate)
     }
     
+    var compactConstraints = [NSLayoutConstraint]()
+    var regularConstraints = [NSLayoutConstraint]()
+    
     // MARK: Data Configuration
     
     weak var delegate: NavigationViewDelegate? {
@@ -66,8 +69,8 @@ open class NavigationView: UIView {
     var tileStoreLocation: TileStoreConfiguration.Location? = .default
     private var _navigationMapView: NavigationMapView? = nil
     lazy var navigationMapView: NavigationMapView = {
-        _navigationMapView?.frame = self.bounds
-        let navigationMapView = _navigationMapView ?? NavigationMapView(frame: self.bounds, tileStoreLocation: tileStoreLocation)
+        _navigationMapView?.frame = bounds
+        let navigationMapView = _navigationMapView ?? NavigationMapView(frame: bounds, tileStoreLocation: tileStoreLocation)
         navigationMapView.isHidden = false
         navigationMapView.translatesAutoresizingMaskIntoConstraints = false
         
@@ -80,11 +83,11 @@ open class NavigationView: UIView {
     
     // MARK: End of Route UI
     
-    lazy var endOfRouteShowConstraint: NSLayoutConstraint? = self.endOfRouteView?.bottomAnchor.constraint(equalTo: self.safeBottomAnchor)
+    lazy var endOfRouteShowConstraint: NSLayoutConstraint? = endOfRouteView?.bottomAnchor.constraint(equalTo: safeBottomAnchor)
     
-    lazy var endOfRouteHideConstraint: NSLayoutConstraint? = self.endOfRouteView?.topAnchor.constraint(equalTo: self.bottomAnchor)
+    lazy var endOfRouteHideConstraint: NSLayoutConstraint? = endOfRouteView?.topAnchor.constraint(equalTo: bottomAnchor)
     
-    lazy var endOfRouteHeightConstraint: NSLayoutConstraint? = self.endOfRouteView?.heightAnchor.constraint(equalToConstant: Constants.endOfRouteHeight)
+    lazy var endOfRouteHeightConstraint: NSLayoutConstraint? = endOfRouteView?.heightAnchor.constraint(equalToConstant: Constants.endOfRouteHeight)
     
     var endOfRouteView: UIView? {
         didSet {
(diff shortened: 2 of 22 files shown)


Sources/MapboxNavigation/NavigationView.swift
```diff
@@ -229,34 +236,194 @@ extension NavigationView {
...
 
-        floatingStackView.topAnchor.constraint(equalTo: topBannerContainerView.bottomAnchor, constant: 10).isActive = true
+        let regularBottomBannerConstraints = [
+            regularBottomBannerContainerViewTrailingConstraint,
+            regularBottomBannerContainerViewLeadingConstraint,
+            regularBottomBannerContainerViewBottomConstraint
+        ]
+        
+        let compactBottomBannerContainerViewTrailingConstraint = bottomBannerContainerView.trailingAnchor.constraint(equalTo: trailingAnchor,
+                                                                                                                     constant: -(screenWidth * 0.6))
```
1ec5, line 294: In the iPhone SE screenshot you provided, the distance and time labels collide into each other. CarPlay removes the possibility of collisions by wrapping units onto a second line, but that would be a more complex change. Perhaps we could hide the distance label when there isn’t enough room.

It’s a little unusual that the ✖️ close button hangs out on the right end, somewhere in the middle of the screen. Moving it out of the bottom banner would help make things fit better. Would it be straightforward to replace it with a floating button in the bottom-right corner of the map? That way the close button is in a predictable location on screen, and the ℹ️ attribution button doesn’t seem like it’s floating out in the middle of nowhere.

Sources/MapboxNavigation/NavigationView.swift
```diff
@@ -229,34 +236,194 @@ extension NavigationView {
...
+        let regularTopBannerConstraints = [
+            regularTopBannerContainerViewLeadingConstraint,
+            regularTopBannerContainerViewTrailingConstraint,
+            regularTopBannerContainerViewTopConstraint
+        ]
+        
+        let screenWidth = UIScreen.main.bounds.width
+        let compactTopBannerContainerViewLeadingConstraint = topBannerContainerView.leadingAnchor.constraint(equalTo: leadingAnchor)
+        let compactTopBannerContainerViewTrailingConstraint = topBannerContainerView.trailingAnchor.constraint(equalTo: trailingAnchor,
+                                                                                                               constant: -(screenWidth * 0.6))
```
1ec5, line 252: The top banner and bottom banner each have a `SeparatorView` to better delineate the top and bottom banners from the map when the banner background color has low contrast with the map style’s background color. Can we add a similar divider to the trailing side of the banner views?

Sources/MapboxNavigation/CameraController.swift
```diff
@@ -134,8 +134,18 @@ class CameraController: NavigationComponent, NavigationComponentDelegate {
...
-        insets.bottom += bottomBannerContainerView.bounds.height + 10.0
+        
+        switch navigationViewData.navigationView.traitCollection.verticalSizeClass {
+        case .unspecified:
+            fallthrough
+        case .regular:
+            insets.top += topBannerContainerView.bounds.height
+            insets.bottom += bottomBannerContainerView.bounds.height + 10.0
+        case .compact:
+            insets.left += navigationViewData.navigationView.topBannerContainerView.frame.width + 10.0
```
1ec5, line 145: The leading and trailing constraints below are direction-independent, so the flip in a right-to-left language environment. Does that affect the insets required here?

Sources/MapboxNavigation/NavigationView.swift
1ec5, line 252: What does this look like when using guidance cards? That design has the upcoming card peek out from the trailing side, but now would it be clipped or float in the middle of the map somewhere?

1ec5 (commented):
This is an elegant solution to the map-crowding problem identified in #1139. A few things based on visual inspection of the screenshot in https://github.com/mapbox/mapbox-navigation-ios/pull/3643#issue-798289135.

Sources/MapboxNavigation/NavigationView.swift
MaximAlien: Yep, it's currently clipped. I'll have to improve layout there, but in general it also seem to look better (guidance card no longer fills the whole parent view width). For example:

Simulator Screen Shot - iPhone 12 Pro Max - 2022-01-07 at 15 06 57

Sources/MapboxNavigation/CameraController.swift
MaximAlien: Good catch, improved current insets for the right-to-left languages. Interestingly enough Google Maps for iOS doesn't flip top and bottom banner views, whereas Apple Maps flip.

Sources/MapboxNavigation/NavigationView.swift
MaximAlien: Regarding close button: it's a bit tricky to movie it at this point without breaking several APIs. I can also see that most of the navigation apps on the market also show close button in bottom bar menu. I've changed position of attribution button to take into account orientation and notch presence.

I'll play around with a possibility of hiding secondary label in bottom menu, this is indeed an issue on devices with smaller screen sizes.

Sources/MapboxNavigation/NavigationView.swift
MaximAlien: Yep, will add.

1ec5: > guidance card no longer fills the whole parent view width

It’s certainly an improvement on an iPhone screen. Does this change affect the layout on iPad in either orientation? If I remember correctly, we had gotten it so that the subsequent cards would remain visible beside the current card, so that you’d see three up at a time in a sort of marquee layout. But maybe I’m confusing that with a custom integration with guidance cards at the application level.

Sources/MapboxNavigation/StepsViewControllerDelegate.swift
```diff
@@ -0,0 +1,24 @@
+import UIKit
+
+/**
+ `StepsViewControllerDelegate` provides methods for user interactions in a `StepsViewController`.
+ */
+public protocol StepsViewControllerDelegate: AnyObject {
```
1ec5, line 6: Speaking of the step table, what does it look like when you tap or swipe down on the top banner? Does it still get slotted into the space between the top and bottom banners as it does now? If so, we can tackle improving that layout as tail work. There isn’t a very natural place for the step table in this landscape orientation, but maybe presenting it as a modal view controller would be an improvement.

1ec5 (approved):
Suggested changelog entry:

* In landscape orientation, `NavigationViewController`’s top and bottom banners take up less space, leaving more room for the map. ([#3643](https://github.com/mapbox/mapbox-navigation-ios/pull/3643))

Before merging, double-check that the behavior on iPad is reasonable in both orientations: https://github.com/mapbox/mapbox-navigation-ios/pull/3643#discussion_r788444615.

Sources/MapboxNavigation/NavigationView.swift
MaximAlien: I've made a change to only show new layout in landscape mode for devices with regular and compact size classes, this means that iPads are not affected for now. I'm not sure that we should integrate new layout on iPads, as there is enough space in portrait/landscape mode anyway. Guidance cards width is also not affected, so only one guidance card is currently shown, next one is only partly visible. Current look:

Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-01-20 at 11 18 19

I'm thinking about doing some more tests to see whether we should perform any changes on iPad. Guidance cards improvements (which will be done in a separate ticket) might help with making a decision.

Sources/MapboxNavigation/StepsViewControllerDelegate.swift
MaximAlien: Yep, it's currently fit between leading and trailing anchors of top and bottom banners:

Simulator Screen Shot - iPod touch (7th generation) - 2022-01-20 at 11 22 31

It doesn't look too bad, but more space would improve things, I'll think about possible improvements.

PR #3647, opened 2021-12-10
Route linke blinking on refresh

### LINKED ISSUE #3621: [Bug]: Route line blinks when refresh and reroute

Udumft:
Resolves #3621 
Temporarily set Directions to track [custom commit](https://github.com/mapbox/mapbox-directions-swift/pull/634) to test new Directions API before it's release.

PR reverts route refreshing methods on `RouteProgress` to utilize new refreshing protocol.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## main
 
+* Fixed route line blinking when route is refreshed. ([#3647](https://github.com/mapbox/mapbox-navigation-ios/pull/3647))
 * Fixed a crash when approaching an intersection in which one of the lanes is a merge lane. ([#3699](https://github.com/mapbox/mapbox-navigation-ios/pull/3699))
 
 ## v2.2.0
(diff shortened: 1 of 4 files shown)


Udumft:
Bot mentions that public method `refreshRoute(with:at:)` was removed, but in fact it was replaced with backwards-compatible version which uses `RouteRefreshing` protocol, which is adopted by `Route`.

ShanMa1991 (approved):
We may need a ChangeLog item to address the issue #3621 has been fixed.

chezzdev:
Why do we need to undeprecate `refreshRoute(with:at:)` method?

Udumft:
It was deprecated previously because another method for route refreshing was introduced, so old one should've become unused, while new one was not meant to be public. Since approach changed, new method is not required, and old one may stay usable as before.
Also, there's a route refreshing overhaul coming, where we'll revisit this mechanism, so there will be a chance to deprecate anything we won't need anymore

PR #3649, opened 2021-12-10
Clean working directory before switching branches

chezzdev:
The script fails to publish docs because previous actions could pollute the working directory. In my case, it was `Package.resolved` from `Navigation-SPM.xcodeproj` that was updated during docs generation.

### DIFF
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -14,7 +14,7 @@ if ! [[ "${VERSION}" =~ ^([0-9]+\.){2}[0-9]+(-.+)?$ ]]; then
   exit 1
 fi
 
-
+git checkout -- .
 git checkout publisher-production
 git add "${VERSION}"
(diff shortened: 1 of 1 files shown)


bamx23 (approved):

scripts/publish-docs.sh
```diff
@@ -14,7 +14,7 @@ if ! [[ "${VERSION}" =~ ^([0-9]+\.){2}[0-9]+(-.+)?$ ]]; then
   exit 1
 fi
 
-
+git checkout -- .
```
MaximAlien, line 17: Would it make sense to remove untracked files well? E.g. something like `$ git clean -fdx`?

chezzdev: In this flow, the folder with generated documentation is considered untracked. We're only cleaning modified not-staged files here.

PR #3651, opened 2021-12-13
Add documentation for `InstructionsBannerViewDelegate`.

ShanMa1991:
This pr is to add documentation for `InstructionsBannerView`, `BaseInstructionsBannerView` and the publicly exposed properties.`InstructionsBannerViewDelegate` is moved to a separate file as well.

### DIFF
--- a/Sources/MapboxNavigation/InstructionsBannerView.swift
+++ b/Sources/MapboxNavigation/InstructionsBannerView.swift
@@ -4,58 +4,75 @@ import MapboxCoreNavigation
 import MapboxDirections
 
 /**
- `InstructionsBannerViewDelegate` provides methods for reacting to user interactions in `InstructionsBannerView`.
- */
-public protocol InstructionsBannerViewDelegate: VisualInstructionDelegate {
-    /**
-     Called when the user taps the `InstructionsBannerView`.
-     */
-    func didTapInstructionsBanner(_ sender: BaseInstructionsBannerView)
-    
-    /**
-     Called when the user swipes either left, right, or down on the `InstructionsBannerView`
-     */
-    func didSwipeInstructionsBanner(_ sender: BaseInstructionsBannerView, swipeDirection direction: UISwipeGestureRecognizer.Direction)
-}
-
-public extension InstructionsBannerViewDelegate {
-    /**
-     `UnimplementedLogging` prints a warning to standard output the first time this method is called.
-     */
-    func didTapInstructionsBanner(_ sender: BaseInstructionsBannerView) {
-        logUnimplemented(protocolType: InstructionsBannerViewDelegate.self, level: .debug)
-    }
-    
-    /**
-     `UnimplementedLogging` prints a warning to standard output the first time this method is called.
-     */
-    func didSwipeInstructionsBanner(_ sender: BaseInstructionsBannerView, swipeDirection direction: UISwipeGestureRecognizer.Direction) {
-        logUnimplemented(protocolType: InstructionsBannerViewDelegate.self, level: .debug)
-    }
-}
-
-/// :nodoc:
+ A banner view that contains the current step instruction and responds to tap and swipe gestures.
+ 
+ This class responds and gets updated as the user progresses along a route according to the `NavigationComponent` and `BaseInstructionsBannerView` protocol.
+*/
 @IBDesignable
 open class InstructionsBannerView: BaseInstructionsBannerView, NavigationComponent {
+    
+    /**
+     Updates the instructions banner info as the user progresses along a route.
+     
+     - parameter service: The `NavigationService` instance that passes the instruction.
+     - parameter instruction: The `VisualInstructionBanner` instance to be presented.
+     - parameter routeProgress: The`RouteProgress` instance that the instruction banner view is updating.
+     */
     public func navigationService(_ service: NavigationService, didPassVisualInstructionPoint instruction: VisualInstructionBanner, routeProgress: RouteProgress) {
         update(for: instruction)
     }
 }
 
-/// :nodoc:
+/**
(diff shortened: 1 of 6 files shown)


Sources/MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -4,65 +4,85 @@ import MapboxCoreNavigation
...
     public weak var dividerView: UIView!
     weak var _separatorView: UIView!
+    
+    /**
+     A invisible helper view for visualizing the result of the constraints.
+     */
     public weak var separatorView: SeparatorView!
+    
+    /**
+     A drag view indicates whether there're more steps in the remaining route.
```
MaximAlien, line 63: Not sure I follow naming here. What is `drag` view?

Sources/MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -4,65 +4,85 @@ import MapboxCoreNavigation
...
+     */
     public weak var primaryLabel: PrimaryLabel!
+    
+    /**
+     A secondary instruction label below the `primaryLabel` to provide detail information of current step.
+     */
     public weak var secondaryLabel: SecondaryLabel!
+    
+    /**
+     A styled lable indicates the remaining distance along the current step.
```
MaximAlien, line 47:
```suggestion
     A styled label indicates the remaining distance along the current step.
```

Sources/MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -4,65 +4,85 @@ import MapboxCoreNavigation
...
+     */
     public weak var maneuverView: ManeuverView!
+    
+    /**
+     A primary instruction label indicates the current step.
+     */
     public weak var primaryLabel: PrimaryLabel!
+    
+    /**
+     A secondary instruction label below the `primaryLabel` to provide detail information of current step.
```
MaximAlien, line 42:
```suggestion
     A secondary instruction label below the `PrimaryLabel`, which provides detailed information about the current step.
```

Sources/MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -4,65 +4,85 @@ import MapboxCoreNavigation
...
 }
 
-/// :nodoc:
+/**
+ A banner view that contains the current step instruction along a route and responds to tap and swipe gestures, as the base of `InstructionsCardView` and `InstructionsBannerView`.
+*/
 open class BaseInstructionsBannerView: UIControl {
+    
+    /**
+     A view that contains a simple image indicating a type of maneuver.
```
MaximAlien, line 32:
```suggestion
     A view that contains an image indicating a type of maneuver.
```

I think it'd be better to have a neutral phrasing. Nit: maybe it'd be worth noting what component will generate these images.

Sources/MapboxNavigation/InstructionsBannerView.swift
ShanMa1991: Because we could have a drag gesture of that to show more step instructions? As described in the following:
https://github.com/mapbox/mapbox-navigation-ios/blob/68b2c5711d3a52cc792010de90a1a087a5644826/Sources/MapboxNavigation/InstructionsBannerView.swift#L318-L319

MaximAlien:
Please also add documentation for parameters in `updateDistance(for:)` and `update(for:)`.

Sources/MapboxNavigation/InstructionsBannerView.swift
MaximAlien: Ah, I see. Let's maybe mention that: `A view, which indicates that there're more steps in the current route. This view can be swiped to the bottom to see all of these remaining steps.`

Sources/MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -4,65 +4,85 @@ import MapboxCoreNavigation
...
     public weak var distanceLabel: DistanceLabel!
+    
+    /**
+     A vertical view divides between `maneuverView` and `distanceLabel` to `primaryLabel` and `secondaryLabel`.
+     */
     public weak var dividerView: UIView!
     weak var _separatorView: UIView!
+    
+    /**
+     A invisible helper view for visualizing the result of the constraints.
```
MaximAlien, line 58:
```suggestion
     An invisible helper view for visualizing the result of the constraints.
```

Not sure why we ended up making it public. As it doesn't seem to be really useful for the end user.

Sources/MapboxNavigation/InstructionsBannerView.swift
ShanMa1991: Thanks for the suggestion, I've updated this part to provide the use case.

Sources/MapboxNavigation/InstructionsBannerView.swift
ShanMa1991: Thanks for the suggestion, the spelling error fix and improvements are made.

Sources/MapboxNavigation/InstructionsBannerView.swift
```diff
@@ -4,65 +4,87 @@ import MapboxCoreNavigation
...
+     */
     public weak var secondaryLabel: SecondaryLabel!
+    
+    /**
+     A styled label indicates the remaining distance along the current step.
+     */
     public weak var distanceLabel: DistanceLabel!
+    
+    /**
+     A vertical view divides between `ManeuverView` and `DistanceLabel` to `PrimaryLabel` and `SecondaryLabel`.
```
MaximAlien, line 52:
```suggestion
     A vertical view, which is used as a divider between `ManeuverView`/`DistanceLabel` views to the left and `PrimaryLabel`/`SecondaryLabel` views to the right.
```

MaximAlien (approved):

Issue #3628, opened 2021-12-03
Fix issues, which cause test failures in `testPausedFreeDrivePausesNavNativeNavigator` and `testRouteSnappingOvershooting`.


MaximAlien:
In scope of https://github.com/mapbox/mapbox-navigation-ios/pull/3614 `testPausedFreeDrivePausesNavNativeNavigator` and `testRouteSnappingOvershooting` were disabled as cause of their failures was unclear.

In scope of this ticket cause of these issues should be found and tests re-written as per new Navigation Native behavior. See `FIXME` tags for these tests to check additional details.

S2Ler:
I suspect that one of the reason we have issues in tests is that there can be more than one RouteController. I'm currently investigating this in scope of https://github.com/mapbox/mapbox-navigation-ios/pull/3652

MaximAlien:
Thanks @S2Ler. I've added [PR](https://github.com/mapbox/mapbox-navigation-ios/pull/3763) that removes FIXME statements for tests you've fixed.

PR #3652, opened 2021-12-14
Remove route from NavNative Navigator when transition from out of Active Guidance

### LINKED ISSUE #3298: Crash: RouteController.updateIndexes(status:progress:)

S2Ler:
When we start active guidance navigation, we update NavNative Navigator with the route for this session. When we end active guidance navigation, though, the route isn't reset, and NavNative continues to think that it is in the Active Guidance. It is incorrect behavior. 

This change is likely to fix #3298.

After the change, some tests started to fail more because the tests were setup in a way that more than one RouteController existed and they interfered with each other. So, I had to fix the tests as well. This work also addressed #3628. 

Even after these changes, there are still some flaky tests. Can it be related to many PassiveLocationProviders at the same time? Will investigate separately. 

_The follow-up will add a debug assertion to telegraph developers that they are not supposed to create more than one RouteController to be alive simultaneously._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@
 
 * Extracted `MapboxNavigationNative_Private` usage into a type alias to fix a compilation in Xcode 12.4. ([#3662](https://github.com/mapbox/mapbox-navigation-ios/pull/3662))
 
+### Location tracking
+
+* Fixed an issue where dismissing `NavigationViewController` could cause `RouteController` to crash or `PassiveLocationProvider` to behave like active turn-by-turn navigation. It is a programmer error to have more than one alive `NavigationViewController`, `NavigationService` or `RouteController` simultaneously. ([#3652](https://github.com/mapbox/mapbox-navigation-ios/pull/3652))
+
 ## v2.1.0
 
 ### Pricing
--- a/Sources/MapboxCoreNavigation/RouteController.swift
+++ b/Sources/MapboxCoreNavigation/RouteController.swift
@@ -168,12 +168,15 @@ open class RouteController: NSObject {
 
     func changeRouteProgress(_ routeProgress: RouteProgress,
                              completion: @escaping (Bool) -> Void) {
-        updateNavigator(with: routeProgress) { [weak self] isSuccessful in
+        updateNavigator(with: routeProgress) { [weak self] result in
             guard let self = self else { return }
-            if isSuccessful {
+            switch result {
+            case .success:
                 self.routeProgress = routeProgress
+                completion(true)
+            case .failure:
+                completion(false)
             }
-            completion(isSuccessful)
         }
     }
     
@@ -235,43 +238,23 @@ open class RouteController: NSObject {
      whether the change was successful.
      */
     private func updateNavigator(with progress: RouteProgress,
-                                 completion: ((Bool) -> Void)?) {
+                                 completion: ((Result<RouteInfo, Error>) -> Void)?) {
         let encoder = JSONEncoder()
         encoder.userInfo[.options] = progress.routeOptions
         guard let routeData = try? encoder.encode(progress.route),
               let routeJSONString = String(data: routeData, encoding: .utf8) else {
-                  completion?(false)
+                  completion?(.failure(RouteControllerError.failedToSerializeRoute))
                   return
         }
 
         let routeRequest = Directions().url(forCalculating: progress.routeOptions).absoluteString
-        
-        navigator.setRoutesFor(Routes(routesResponse: routeJSONString,
-                                      routeIndex: 0,
-                                      legIndex: UInt32(progress.legIndex),
-                                      routesRequest: routeRequest)) { result in
-            if result.isValue() {
-                if let routeInfo = result.value as? RouteInfo {
-                    os_log("Navigator updated to routeSequenceNumber = %{public}d",
-                           log: RouteController.log,
-                           type: .debug,
-                           routeInfo.routeSequenceNumber)
-                }
-                completion?(true)
-            }
-            else if result.isError() {
-                if let reason = result.error as? String {
-                    os_log("Failed to update navigator with reason: %{public}@",
-                           log: RouteController.log,
-                           type: .error,
-                           reason)
-                }
-                completion?(false)
-            }
-            else {
-                assertionFailure("Invalid Expected value: \(result)")
-                completion?(false)
-            }
+        let routes = Routes(routesResponse: routeJSONString,
+                            routeIndex: 0,
+                            legIndex: UInt32(progress.legIndex),
+                            routesRequest: routeRequest)
+
+        Navigator.shared.setRoutes(routes) { result in
+            completion?(result)
         }
     }
     
@@ -575,6 +558,7 @@ open class RouteController: NSObject {
     }
     
     deinit {
+        removeRoutes(completion: nil)
         BillingHandler.shared.stopBillingSession(with: sessionUUID)
         unsubscribeNotifications()
         routeTask?.cancel()
@@ -739,6 +723,17 @@ extension RouteController: Router {
         }
     }
 
+    private func removeRoutes(completion: ((Error?) -> Void)?) {
+        Navigator.shared.setRoutes(nil) { result in
+            switch result {
+            case .success:
+                completion?(nil)
+            case .failure(let error):
+                completion?(error)
+            }
+        }
+    }
+
     private func shouldStartNewBillingSession(for newRoute: Route, routeOptions: RouteOptions?) -> Bool {
         guard let routeOptions = routeOptions else {
             // Waypoints are read from routeOptions.
@@ -771,4 +766,5 @@ extension RouteController: InternalRouter { }
 enum RouteControllerError: Error {
     case internalError
     case failedToChangeRouteLeg
+    case failedToSerializeRoute
 }
--- a/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
+++ b/Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
@@ -1,9 +1,11 @@
 import MapboxNavigationNative
 import MapboxDirections
+import os.log
 @_implementationOnly import MapboxCommon_Private
 
 class Navigator {
-    
+    static let log: OSLog = .init(subsystem: "com.mapbox.navigation", category: "Navigator")
+
     /**
      Tiles version string. If not specified explicitly - will be automatically resolved
      to the latest version.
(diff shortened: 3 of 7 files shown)


Sources/MapboxCoreNavigation/RouteController.swift
```diff
@@ -570,6 +570,11 @@ open class RouteController: NSObject {
         BillingHandler.shared.stopBillingSession(with: sessionUUID)
         unsubscribeNotifications()
         routeTask?.cancel()
+        Navigator.shared.navigator.setRoutesFor(nil) { _ in
```
bamx23, line 573: For drop-in UI, this `deinit` happens when the view controller is deallocated, which might happen after passive navigation is started. Is there a way to call this earlier?

Sources/MapboxNavigation/TopBannerViewController.swift
```diff
@@ -353,9 +353,7 @@ extension TopBannerViewController: NavigationComponent {
         instructionsBannerView.update(for: instruction)
         lanesView.update(for: instruction)
         nextBannerView.navigationService(service, didPassVisualInstructionPoint: instruction, routeProgress: routeProgress)
-        DispatchQueue.main.async {
```
bamx23, line 356: Is it guaranteed to be called on the main queue?

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@
 * Added the `RouterDelegate.router(_:initialManeuverBufferWhenReroutingFrom:)` method and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to configure rerouting buffering. ([#3454](https://github.com/mapbox/mapbox-navigation-ios/pull/3454))
 * Added `RouterDelegate.router(_:requestBehaviorForReroutingWith:)` and related `NavigationServiceDelegate` and `NavigationViewControllerDelegate` methods to provide customization for rerouting mechanism by providing a user-defined route as a reroute. ([#3472](https://github.com/mapbox/mapbox-navigation-ios/pull/3472))
 
+### Location tracking
+
+* Fixed an issue where the transition from Active to Passive navigation didn't update underlying Navigator properly. It is a programmer error to have more than one active RouteController simultaneously. ([#3652](https://github.com/mapbox/mapbox-navigation-ios/pull/3652))
```
bamx23, line 22: Should we have this disclaimer here about simultaneous instances? 
And it sounds a bit strange to me. Maybe this is better?
```suggestion
* Fixed an issue where the transition from Active to Passive navigation didn't update underlying Navigator properly. It is an integration issue to have more than one active RouteController simultaneously. ([#3652](https://github.com/mapbox/mapbox-navigation-ios/pull/3652))
```

bamx23 (approved):

Sources/MapboxNavigation/TopBannerViewController.swift
S2Ler: Other views are modified here. So I would say yes. We will have a main thread sanitizer yelling at us if it won't be.

CHANGELOG.md
S2Ler: > And it sounds a bit strange to me. Maybe this is better?

This is a commonly used phrase. For example, take a look on this Apple documentation: https://developer.apple.com/documentation/watchconnectivity/wcsessionactivationstate/wcsessionactivationstateinactive?changes=_2_6&language=objc

S2Ler: > Should we have this disclaimer here about simultaneous instances?

Isn't "It is a programmer error to have more than one active RouteController simultaneously." enough?

Sources/MapboxCoreNavigation/RouteController.swift
S2Ler: I don't know how to call it earlier unless we create an explicit method. 

IMHO, We shall improve FD/AG transition similar to the Android way in v3. These modes should be tightly coupled as our underlying library is.

S2Ler: @1ec5 please review this code, maybe you have some suggestions.

CHANGELOG.md
bamx23: Nevermind, I'm okay with this text, just didn't get the reason why dow we mention it here.

S2Ler: I have a follow up PR that will add a documentation entry related to this requirement. If this change is accepted at all.

Sources/MapboxCoreNavigation/RouteController.swift
1ec5: Would it be possible to do some of this teardown when canceling or arriving? If there needs to be an explicit method, some kind of counterpart to `updateRoute(with:routeOptions:completion:)`, that wouldn’t be so bad as long as NavigationViewController would normally take care of calling it for the developer. Changing the behavior in this way wouldn’t necessarily be a breaking change: if an application uses RouteController directly but doesn’t call the explicit teardown method, it would just mean that #3298 would go unfixed for the application, right?

By the way, `MapboxCoreNavigation.Navigator` should encapsulate `MapboxNavigationNative.Navigator` well enough that RouteController shouldn’t have to root around `MapboxCoreNavigation.Navigator.navigator`.

> IMHO, We shall improve FD/AG transition similar to the Android way in v3. These modes should be tightly coupled as our underlying library is.

Broadly speaking, I would normally say the reverse: maybe MapboxNavigationNative needs to be less tightly coupled in the long run rather than forcing MapboxCoreNavigation to be more tightly coupled.

A developer might reasonably expect to be able to initialize a view controller and hang onto it for a while before presenting it. The more we tie underlying behavior to the lifecycle of the view controller object, the more we open ourselves up to side effects when the view controller leaks, either due to a bug in the SDK or unanticipated application behavior.

S2Ler: > Broadly speaking, I would normally say the reverse: maybe MapboxNavigationNative needs to be less tightly coupled in the long run rather than forcing MapboxCoreNavigation to be more tightly coupled.

> A developer might reasonably expect to be able to initialize a view controller and hang onto it for a while before presenting it. The more we tie underlying behavior to the lifecycle of the view controller object, the more we open ourselves up to side effects when the view controller leaks, either due to a bug in the SDK or unanticipated application behavior.

For you information: @mapbox/navnative

S2Ler: > Would it be possible to do some of this teardown when canceling or arriving? If there needs to be an explicit method, some kind of counterpart to updateRoute(with:routeOptions:completion:), that wouldn’t be so bad as long as NavigationViewController would normally take care of calling it for the developer. 

I'll try. It should be possible, yes. 

> Changing the behavior in this way wouldn’t necessarily be a breaking change: if an application uses RouteController directly but doesn’t call the explicit teardown method, it would just mean that #3298 would go unfixed for the application, right?

Yes, I think so. 

> By the way, MapboxCoreNavigation.Navigator should encapsulate MapboxNavigationNative.Navigator well enough that RouteController shouldn’t have to root around MapboxCoreNavigation.Navigator.navigator.

Agree.

S2Ler: Updated.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -167,6 +167,15 @@ public protocol Router: CLLocationManagerDelegate {
     func updateRoute(with indexedRouteResponse: IndexedRouteResponse,
                      routeOptions: RouteOptions?,
                      completion: ((Bool) -> Void)?)
+
+    /**
+     Asynchronously removes currently active route.
+
+     This method is called automatically by `NavigationViewController` at the end of the navigation.
+     If you use `Router` without `NavigationViewController`, you are supposed to call this method when you are done
```
bamx23, line 175: It would be great to add this to the CHANGELOG. As it was said, it's not a breaking change, but something we should proactively point developers to.

bamx23 (approved):

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@
 
 * Extracted `MapboxNavigationNative_Private` usage into a type alias to fix a compilation in Xcode 12.4. ([#3662](https://github.com/mapbox/mapbox-navigation-ios/pull/3662))
 
+### Location tracking
+
+* Fixed an issue where dismissing `NavigationViewController` didn't update underlying Navigator properly. If you use `RouteController` or `NavigationService` outside of `NavigationViewController` you should call `RouteController.removeRoute(completion:)` at the end of the navigation. ([#3652](https://github.com/mapbox/mapbox-navigation-ios/pull/3652))
```
1ec5, line 34: Navigator is an implementation detail:

```suggestion
* Fixed an issue where dismissing `NavigationViewController` could cause `RouteController` to crash or `PassiveLocationProvider` to behave like active turn-by-turn navigation. If you use `RouteController` or `NavigationService` outside of `NavigationViewController` you should call `RouteController.removeRoute(completion:)` at the end of the turn-by-turn navigation session. ([#3652](https://github.com/mapbox/mapbox-navigation-ios/pull/3652))
```

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -167,6 +167,15 @@ public protocol Router: CLLocationManagerDelegate {
...
+     with `Router`.
+     */
+    func removeRoute(completion: (() -> Void)?)
```
1ec5, line 178: Should the completion handler accept a Boolean-typed success argument, similar to `updateRoute(with:options:completion:)`? It may not be necessary now, but it would be more difficult to introduce later.

1ec5 (approved):

Sources/MapboxCoreNavigation/Router.swift
S2Ler: I agree, thanks.

S2Ler: Although it shouldn't fail I would say.

CHANGELOG.md
S2Ler: Thanks

Sources/MapboxCoreNavigation/Router.swift
S2Ler: Done

1ec5 (approved):

PR #3653, opened 2021-12-14
Logging functionality for CarPlay

jill-cardamon:
This PR adds logging functionality to CarPlay to make it easier for us to debug locally.

Still needs testing on a real CarPlay head unit.
Additional ideas:
- output to local server

### DIFF
--- a/Sources/MapboxNavigation/CarPlayManager.swift
+++ b/Sources/MapboxNavigation/CarPlayManager.swift
@@ -2,6 +2,7 @@ import CarPlay
 import MapboxCoreNavigation
 import MapboxDirections
 import MapboxMaps
+import os.log
 
 /**
  `CarPlayManager` is the main object responsible for orchestrating interactions with a Mapbox map on CarPlay.
@@ -77,6 +78,7 @@ public class CarPlayManager: NSObject {
 
     private weak var navigationService: NavigationService?
     private var idleTimerCancellable: IdleTimerManager.Cancellable?
+    private let logger: OSLog = .init(subsystem: "com.mapbox.navigation", category: "CarPlay")
     
     /**
      Programatically begins a CarPlay turn-by-turn navigation session.
@@ -88,11 +90,13 @@ public class CarPlayManager: NSObject {
     public func beginNavigationWithCarPlay(using currentLocation: CLLocationCoordinate2D,
                                            navigationService: NavigationService) {
         // Stop the background `PassiveLocationProvider` sending location and heading update `mapView` before turn-by-turn navigation session starts.
+        os_log("Begin navigation with CarPlay", log: logger, type: .info)
         if let locationProvider = navigationMapView?.mapView.location.locationProvider {
             locationProvider.stopUpdatingLocation()
             locationProvider.stopUpdatingHeading()
             if let passiveLocationProvider = locationProvider as? PassiveLocationProvider {
                 passiveLocationProvider.locationManager.pauseTripSession()
+                os_log("Trip session paused", log: logger, type: .info)
             }
         }
         
@@ -143,6 +147,7 @@ public class CarPlayManager: NSObject {
          routingProvider: RoutingProvider,
          eventsManager: NavigationEventsManager? = nil,
          carPlayNavigationViewControllerClass: CarPlayNavigationViewController.Type? = nil) {
+        os_log("CarPlayManager initialized", log: logger, type: .info)
         self.styles = styles ?? [DayStyle(), NightStyle()]
         self.routingProvider = routingProvider
         self.eventsManager = eventsManager ?? .init(activeNavigationDataSource: nil,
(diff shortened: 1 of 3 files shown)


Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -77,6 +78,7 @@ public class CarPlayManager: NSObject {
 
     private weak var navigationService: NavigationService?
     private var idleTimerCancellable: IdleTimerManager.Cancellable?
+    private let logger: OSLog = .init(subsystem: "com.mapbox.navigation", category: "CarPlayManager")
```
bamx23, line 81: Should we add a factory for loggers? So that we don't duplicate `subsystem` value everywhere.

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
```diff
@@ -23,7 +26,7 @@ extension NavigationMapView {
             do {
                 try mapView.mapboxMap.style.addSource(streetsSource, id: sourceIdentifier)
             } catch {
-                NSLog("Failed to add \(sourceIdentifier) with error: \(error.localizedDescription).")
+                os_log("Failed to add %s with error: %s.", log: logger, type: .error, sourceIdentifier as CVarArg, error.localizedDescription)
```
bamx23, line 29: The `%s` will be redacted out in the logs. Should it be `%{public}s`?

bamx23, line 29: Why do we need casting to `CVarArg` here?

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
```diff
@@ -1,6 +1,9 @@
 import MapboxMaps
 import MapboxDirections
 import MapboxCoreNavigation
+import os.log
+
+private let logger: OSLog = .init(subsystem: "com.mapbox.navigation", category: "RoadNameLabeling")
```
bamx23, line 6: Should we use private global constants in other classes too?

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -160,22 +164,39 @@ public class CarPlayManager: NSObject {
...
     
     func unsubscribeFromNotifications() {
         NotificationCenter.default.removeObserver(self,
                                                   name: .navigationCameraStateDidChange,
                                                   object: carPlayMapViewController?.navigationMapView.navigationCamera)
+        os_log("CarPlayManager suspended notifications.", log: logger, type: .info)
+    }
+    
+    @available(iOS 15.0, *)
+    func getLogEntries() throws -> [OSLogEntryLog] {
```
bamx23, line 178: Is it for debugging? If so, don't forget to remove this

Sources/MapboxNavigation/CarPlayManager.swift
bamx23: Or we can go with a simpler way: creating an internal constant for the log subsystem.

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
jill-cardamon: Cast was needed since os_log uses `StaticString`

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
jill-cardamon: We needed one here since you can't create stored properties in extensions. For consistency, we should either move constants to be global or use a manager for these OSLog constants

Sources/MapboxNavigation/NavigationMapView+RoadNameLabeling.swift
jill-cardamon: Ah yes. I will try this. Thank you

Sources/MapboxNavigation/CarPlayManager.swift
jill-cardamon: We could potentially use `OSLogManager` but would probably want to consider using this manager for all of the SDK.

Sources/MapboxNavigation/CarPlayManager.swift
jill-cardamon: This is for retrieving log entries if you want to see them printed to a file instead of looking at the log via the Console. This functionality is only available via iOS 15.

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -333,6 +354,7 @@ extension CarPlayManager: CPApplicationDelegate {
         interfaceController.setRootTemplate(mapTemplate, animated: false)
             
         eventsManager.sendCarPlayConnectEvent()
+        os_log("CarInterfaceController did connect to window.", log: logger, type: .info)
```
MaximAlien, line 357: I think that `CPApplicationDelegate` delegate methods are deprecated. We'll need similar logging for `CPTemplateApplicationSceneDelegate` as well. I think that it'd be also beneficial to log information regarding `CPInterfaceController` (e.g. its address, main template etc).

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -2,6 +2,7 @@ import CarPlay
 import MapboxCoreNavigation
 import MapboxDirections
 import MapboxMaps
+import os.log
```
MaximAlien, line 5: We'll need to add logging for `CPTemplate` lifecycle callbacks in `CarPlayManagerDelegate`:
- carPlayManager(_:templateWillAppear:animated:)
- carPlayManager(_:templateDidAppear:animated:)
- carPlayManager(_:templateWillDisappear:animated:)
- carPlayManager(_:templateDidDisappear:animated:)

Sources/MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 @_spi(Restricted) import MapboxMaps
 import MapboxCoreNavigation
 import MapboxDirections
+import os.log
```
MaximAlien, line 5: It'd be useful to have logging information when `CarPlayActivity` value changes.

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -561,6 +582,7 @@ extension CarPlayManager {
         
         switch result {
         case let .failure(error):
+            os_log("Failed to calculate routes.", log: logger, type: .debug)
```
MaximAlien, line 585: This looks like an error log type.

Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -93,6 +95,7 @@ public class CarPlayManager: NSObject {
             locationProvider.stopUpdatingHeading()
             if let passiveLocationProvider = locationProvider as? PassiveLocationProvider {
                 passiveLocationProvider.locationManager.pauseTripSession()
+                os_log("Trip session paused", log: logger, type: .debug)
```
MaximAlien, line 98: It'd be good to go through all methods in `CarPlayManager` and see what CarPlay related methods it uses. I think that we should log all callbacks we receive from it.

MaximAlien:
We should add logging for `CarPlaySearchController`/`CPSearchTemplateDelegate` as well.

Sources/MapboxNavigation/CarPlayManager.swift
MaximAlien: I think there are several build errors because of `OSLogEntryLog`. For example getting: error: cannot find type 'OSLogEntryLog' in scope.

muz444 (approved):

PR #3659, opened 2021-12-15
v2.1.1

1ec5:
Updated the Xcode project, CocoaPods podspecs, and changelog for v2.1.1 (which supersedes v2.1.0). This is an out-of-band release behind the latest release train, so the PR targets the release-v2.1 branch instead of main.

/cc @mapbox/navigation-ios


jill-cardamon (approved):
what about the MapboxNavigation-Documentation.podspec, README.md, and custom-navigation.md?

1ec5:
README.md and custom-navigation.md already refer to v2.1.0 because we never cherry-picked 30a319ea91fccf68504f14e7a8839275ef95ebf5 into the release-v2.1 branch. The script avoids bumping anything in those documents for patch releases.

#3568 deleted MapboxNavigation-Documentation.podspec. We no longer use CocoaPods to generate docsets.

PR #3662, opened 2021-12-15
Extract MapboxNavigationNative_Private usage into type alias

1ec5:
For whatever reason, Swift 5.3 (Xcode 12.4) requires us to use a type alias if we refer to a module as a namespace and the module had been imported as `@_implementationOnly` and CocoaPods is doing its magic behind the scenes. This small change fixes building with CocoaPods in Xcode 12.4 and unblocks v2.1.1.

Cherry-pick to the release-v2.1 branch after merging.

/cc @mapbox/navigation-ios


MaximAlien (approved):

jill-cardamon (approved):

1ec5:
Cherry-picked to the release-v2.1 branch in 56aa290a762785ab270ea7f2fb8558ffaefb26ba.

Issue #3597, opened 2021-11-15
Rerouting might not always select the best route


azarovalex:
If Directions API returned multiple routes during rerouting NavsDK uses this code to select the best route:
https://github.com/mapbox/mapbox-navigation-ios/blob/4a2c6904b02d8433af38db5771d405887906350f/Sources/MapboxCoreNavigation/Router.swift#L347-L356

It selects the route which name has the minimal Levenshtein distance with the name of the original route. The route name is essentially a concatenation of all the street names on the route. 
This logic might not always select the best route, we should take ETA into consideration as well.

1ec5:
>  This logic might not always select the best route, we should take ETA into consideration as well.

This is a feature; see the explanation at https://github.com/mapbox/mapbox-navigation-ios/issues/818#issue-272406754.

> It selects the route which name has the minimal Levenshtein distance with the name of the original route. The route name is essentially a concatenation of all the street names on the route.

Not all the street names, but rather the names of the streets that appear for the longest distances along the route. That way, it ends up being a very cheap proxy for the route geometry, but also more nuanced than the raw geometry because it can distinguish between two alternative routes that have similar geometry.

1ec5:
A user reported that `MCE, PIE` was unexpectedly matched to `PIE, Braddell Road` when given the following three route leg summaries:

1. `PIE, Upper Bukit Timah Road`
2. `PIE, Bukit Timah Road`
3. `PIE, Braddell Road`

A human would’ve considered the three routes to be a tie based on the summaries alone, then preferred the first route because of its optimality. They would’ve primarily compared the whole names and secondarily compared individual words, but they wouldn’t have compared individual characters as this application of the Levenshtein algorithm appears to. We could generalize the extension method to operate on a sequence instead of a string, then compare by name and by word.

At a certain point, the summary similarity becomes less relevant than the position in the response, allowing us to ignore the user selection and go with the most optimal route. In this case, the winning candidate was only 17% similar. We should only consider best matches above 50% – or even higher.

1ec5:
> We could generalize the extension method to operate on a sequence instead of a string, then compare by name and by word.

[...]

PR #3664, opened 2021-12-17
Improve rerouting logic if all candidates are very different from the initial route

azarovalex:
Fixes https://github.com/mapbox/mapbox-navigation-ios/issues/3597


Sources/TestHelper/DummyRoute.swift
```diff
@@ -0,0 +1,20 @@
...
+public class DummyRoute: Route {
+
+    public init(description: String) {
+        let leg = RouteLeg(steps: [],
+                           name: description,
+                           distance: 0,
+                           expectedTravelTime: 0,
+                           typicalTravelTime: nil,
+                           profileIdentifier: .automobile)
+        super.init(legs: [leg], shape: nil, distance: 0, expectedTravelTime: 0, typicalTravelTime: nil)
```
azarovalex, line 14: Little hack, because `description` property cannot be overriden

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -404,10 +404,18 @@ extension InternalRouter where Self: Router {
...
-            let leftDistance = left.element.description.minimumEditDistance(to: target)
-            let rightDistance = right.element.description.minimumEditDistance(to: target)
-            return leftDistance < rightDistance
-        }?.offset
+        
+        guard let bestCandidate = map({
+            (route: $0, editDistance: $0.description.minimumEditDistance(to: target))
+        }).enumerated().min(by: { $0.element.editDistance < $1.element.editDistance }) else { return nil }
+
+
```
bamx23, line 412:
```suggestion

```

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -404,10 +404,18 @@ extension InternalRouter where Self: Router {
...
+        // If the most similar route is still more than 50% different from the original route,
+        // we fallback to the fastest route which index is 0.
+        guard bestCandidate.element.route.description.count > 0 else { return 0 }
+        let similarityScore = Double(bestCandidate.element.editDistance) / Double(bestCandidate.element.route.description.count)
```
bamx23, line 416: `bestCandidate.element.route.description` or `target`?

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -404,10 +404,18 @@ extension InternalRouter where Self: Router {
...
-            let rightDistance = right.element.description.minimumEditDistance(to: target)
-            return leftDistance < rightDistance
-        }?.offset
+        
+        guard let bestCandidate = map({
+            (route: $0, editDistance: $0.description.minimumEditDistance(to: target))
+        }).enumerated().min(by: { $0.element.editDistance < $1.element.editDistance }) else { return nil }
+
+
+        // If the most similar route is still more than 50% different from the original route,
```
bamx23, line 413: Why 50%?

1ec5: This number came from https://github.com/mapbox/mapbox-navigation-ios/issues/3597#issuecomment-982362158. There are only ever two road names in the route description per leg, so a 50% match is roughly one of the two names matching (modulo any differences in name length or minor things like “South”). We could probably lower the threshold without much problem, but this becomes a bugfarm, then the per-step approach in https://github.com/mapbox/mapbox-navigation-android/issues/3116#issuecomment-655710099 would be a decent plan B.

Sources/MapboxCoreNavigation/Router.swift
```diff
@@ -404,10 +404,18 @@ extension InternalRouter where Self: Router {
...
+        // we fallback to the fastest route which index is 0.
+        guard bestCandidate.element.route.description.count > 0 else { return 0 }
+        let similarityScore = Double(bestCandidate.element.editDistance) / Double(bestCandidate.element.route.description.count)
+        guard similarityScore < 0.5 else { return 0 }
```
1ec5, line 417: Thanks, I inverted this logic in https://github.com/mapbox/mapbox-navigation-ios/issues/3597#issuecomment-982362158 because I mistook `score` for a measure of similarity, but in fact it’s still based on the edit distance, so it’s really a relative difference.

Sources/MapboxCoreNavigation/Router.swift
1ec5: Good question, I don’t have a good intuition for this. Can we try out a few scenarios to see which would be more correct?

Sources/TestHelper/DummyRoute.swift
1ec5: This is correct, since `Route.description` is based on `RouteLeg.name`:

https://github.com/mapbox/mapbox-directions-swift/blob/28a0bc98ca13ac7cbfa08d06f273dcd94503d696/Sources/MapboxDirections/DirectionsResult.swift#L191
https://github.com/mapbox/mapbox-directions-swift/blob/28a0bc98ca13ac7cbfa08d06f273dcd94503d696/Sources/MapboxDirections/RouteLeg.swift#L353

Tests/MapboxCoreNavigationTests/RerouteTests.swift
```diff
@@ -0,0 +1,34 @@
...
+
+        let mostCommonIndex = candidates.index(mostSimilarTo: originalRoute)
+        XCTAssertEqual(mostCommonIndex, 2)
+    }
+
+    func testSelectsFastestRouteIfAllCandidatesAreVeryDifferent() {
+        let originalRoute = DummyRoute(description: "AAAAA")
+        let candidates = [
+            DummyRoute(description: "ABBBB"),
+            DummyRoute(description: "AABBB"),
```
1ec5, line 24: Let’s add tests where the descriptions are of different lengths.

Tests/MapboxCoreNavigationTests/RerouteTests.swift
```diff
@@ -0,0 +1,34 @@
+import MapboxDirections
+@testable import MapboxCoreNavigation
+import XCTest
+import TestHelper
+
+class RerouteTests: XCTestCase {
+
+    func testSelectsMostSililarRouteForRerouting() {
```
1ec5, line 8:
```suggestion
    func testSelectsMostSimilarRouteForRerouting() {
```

1ec5 (approved):
Suggested changelog entry:

* When rerouting the user, if none of the new routes is very similar to the original route selection, the `Router` now follows the most optimal route, not a route that is only marginally similar. ([#3664](https://github.com/mapbox/mapbox-navigation-ios/pull/3664))

Sources/MapboxCoreNavigation/Router.swift
bamx23: If the best candidate is when we remove letters from the `target`, then we might get more than 1 here. I.e.
```
target = "AAAAAA"
bestCandidate = "AA"
editDistance = 4 // remove any 4 letters
score = 4 / len("AA") = 4 / 2 = 2
```
At the same time, if the best candidate is when you add letters, then we can get the same "larger than 1" if compared to the `target` length. 

We can divide by the sum of the sizes to guarantee that we get the value within `0...1`, but it's a bit more difficult to interpret such a score, i.e. 
```
"AAAAAA" -> "AAAAAA" = 0 / 12 = 0.00
"AAAAAA" -> "AAAAA"  = 1 / 11 = 0.09
"AAAAAA" -> "AAAA"   = 2 / 10 = 0.20
"AAAAAA" -> "AAA"    = 3 /  9 = 0.33
"AAAAAA" -> "AA"     = 4 /  8 = 0.50
"AAAAAA" -> "A"      = 5 /  7 = 0.71
"AAAAAA" -> ""       = 6 /  6 = 1.00

"AAAAAA" -> "AAAABB" = 2 / 12 = 0.17
"AAAAAA" -> "AAABBB" = 3 / 12 = 0.25
"AAAAAA" -> "AABBBB" = 4 / 12 = 0.33
"AAAAAA" -> "BBBBBB" = 6 / 12 = 0.50

"AAAAAA" -> "BBB"    = 6 /  9 = 0.6

"AAAAAA" -> "AAAAAAC"     = 1 / 13 = 0.08
"AAAAAA" -> "AAAAAACCCC"  = 4 / 16 = 0.25
"AAAAAA" -> "AAAAAACCCCC" = 5 / 17 = 0.29
```
Maybe we can compare the score to `0.25` as for "replacing the half of the string" or similar.

azarovalex: I added some tests for different candidate lengths and it seems like @bamx23's algorithm might show better results for some edge cases.
It's now a little bit harder to understand, but I added a comment with a link to this explanation.

Even if we are unable to select the most common route we will fallback to the fastest route which not as bad.

762caliber:
Hi, 


checking in for a update for NCS [#335](https://github.com/mapbox-collab/ncs-collab/issues/335).

Could we confirm that the PR have completed / fixed in which version of code please.


cc @bentruscello , @venkat-b-1983 , @OttyLab

762caliber:
Hi,

checking in for a update for NCS #335.

Could we confirm that the PR have completed / fixed in which version of code please.

cc @bentruscello , @venkat-b-1983 , @OttyLab

PR #3672, opened 2022-01-04
Move location calculation to a background queue

### LINKED ISSUE #3129: SimulatedLocationManager on long routes consumes a lot of CPU on the main thread
**Mapbox Navigation SDK version:** release-v2.0 aa442b9191e8efe79cf35dd790526c8e3a45615b

### Steps to reproduce

 1. Take a 10000 km long route
 2. Start simulated navigation
 3. Check CPU usage on the main thread

### Expected behavior
SimulatedLocationManager doesn't consume a lot of CPU on the main thread.

### Actual behavior
SimulatedLocationManager consumes a lot of CPU on the main thread. 

### Proposed solution 

Move time consuming work in SimulatedLocationManager to a non-main DispatchQueue.

chezzdev:
Fixes #3129.

### DIFF
--- a/Sources/MapboxCoreNavigation/SimulatedLocationManager.swift
+++ b/Sources/MapboxCoreNavigation/SimulatedLocationManager.swift
@@ -58,7 +58,7 @@ open class SimulatedLocationManager: NavigationLocationManager {
         self.currentDistance = currentDistance
         self.route = route
 
-        self.timer = DispatchTimer(countdown: .milliseconds(0), repeating: updateInterval, accuracy: accuracy, executingOn: .main) { [weak self] in
+        self.timer = DispatchTimer(countdown: .milliseconds(0), repeating: updateInterval, accuracy: accuracy, executingOn: .global()) { [weak self] in
             self?.tick()
         }
         
@@ -142,6 +142,7 @@ open class SimulatedLocationManager: NavigationLocationManager {
         let coordinatesNearby = polyline.trimmed(from: newCoordinate, distance: 100)!.coordinates
         
         // Simulate speed based on expected segment travel time
+        let currentSpeed: CLLocationSpeed
         if let expectedSegmentTravelTimes = routeProgress?.currentLeg.expectedSegmentTravelTimes,
             let shape = routeProgress?.route.shape,
             let closestCoordinateOnRoute = shape.closestCoordinate(to: newCoordinate),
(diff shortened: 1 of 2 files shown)


Sources/MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -160,9 +162,9 @@ open class SimulatedLocationManager: NavigationLocationManager {
                                   course: newCoordinate.direction(to: lookAheadCoordinate).wrap(min: 0, max: 360),
                                   speed: currentSpeed,
                                   timestamp: Date())
-        
+
         self.simulatedLocation = location
-        
+
         delegate?.locationManager?(self, didUpdateLocations: [location])
```
Udumft, line 168: Since `tick` is now executed on another thread, should this delegate call be explicitly done on `main`?

Default `CLLocationManager` captures the thread it was instantiated on for location update calls, but our implementation forced a main thread

bamx23, line 168: Should this call me made on the main queue?

bamx23 (commented):
What about a changelog line?

Udumft (commented):
Did you manage to measure CPU usage change after these changes?

1ec5 (commented):
If the performance improvement is non-negligible for long routes, consider adding a changelog entry to that effect. Thanks!

Sources/MapboxCoreNavigation/SimulatedLocationManager.swift
chezzdev: It may make sense to get back a delegate call to the main thread to minimize potential side effects in execution.
However, I doubt that anything downstream should rely on this call being executed on the main thread.

chezzdev: Probably yes.
https://github.com/mapbox/mapbox-navigation-ios/pull/3672#discussion_r833390975

bamx23 (approved):

Sources/MapboxCoreNavigation/SimulatedLocationManager.swift
chezzdev: Now I remember why I left the delegate call on the same queue as `tick` - out tests heavily rely on synchronous delegate call inside `tick`.

chezzdev: @bamx23 multiple tests and test fixtures rely on a synchronous delegate call inside the `tick` method. Since we're not guaranteeing execution on a specific queue, I'm proposing leaving it on a custom background queue.

bamx23: IMO it's a risky change. What if we add a nullable parameter to the manager with the queue on which the callback should be called and keep it `nil` in tests to make calls synchronous while having the default value of `main` queue?

Udumft: But I believe the point of this PR is to make it non-main thread by default... As Alex said, our tests are failing if we introduce explicit main thread call, so there's a good chance the same will happen with user code. We could instead document that delegate will be called on a background instead, so user will be able to manage threading himself if he needs it.

Udumft (approved):
LGTM, but please also document delegate calling thread change.

PR #3676, opened 2022-01-07
Force update map buttons on CarPlay during active navigation.

MaximAlien:
PR fixes a problem, which prevents users from updating map buttons (map, leading and trailing navigation buttons) whenever panning interface is shown.

### DIFF
--- a/Sources/MapboxNavigation/CarPlayManager.swift
+++ b/Sources/MapboxNavigation/CarPlayManager.swift
@@ -685,34 +685,7 @@ extension CarPlayManager: CPMapTemplateDelegate {
             CarPlayManager.currentActivityKey: currentActivity
         ]
 
-        guard let carPlayMapViewController = carPlayMapViewController else { return mapTemplate }
-        
-        if let mapButtons = delegate?.carPlayManager(self,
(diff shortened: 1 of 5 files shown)


Sources/MapboxNavigation/CarPlayManager.swift
```diff
@@ -783,15 +752,33 @@ extension CarPlayManager: CPMapTemplateDelegate {
...
-            carPlayMapViewController.dismissPanningButton = closeButton
-            mapTemplate.mapButtons = [closeButton]
+            if let carPlayMapViewController = carPlayMapViewController {
+                let closeButton = carPlayMapViewController.panningInterfaceDismissalButton(for: mapTemplate)
+                carPlayMapViewController.dismissPanningButton = closeButton
+                mapTemplate.mapButtons = [closeButton]
+            }
+        }
+        
+        if let leadingButtons = delegate?.carPlayManager(self,
```
MaximAlien, line 770: Whenever panning interface is shown we're only checking whether delegate methods provide any changes for `CPMapTemplate` buttons. If so - we use them, if not, originally set buttons will be used.

ShanMa1991 (approved):

PR #3678, opened 2022-01-10
Extend RoutesPresentationStyle to present routes with custom camera options

### LINKED ISSUE #3675: Extend `RoutesPresentationStyle` to support the ability to present routes based on custom `CameraOptions`.
At this point `RoutesPresentationStyle` enumeration only allows to showcase routes on a map view based on their geometry (either single or multiple routes).

In scope of this ticket this enumeration should be extended to support presentation based on custom `CameraOptions`. This will give the opportunity to handle more various cases when application UI is non-standard.

ShanMa1991:
### Description
This pr is to fix #3675 by adding `RoutesPresentationStyle.custom(cameraOptions:)` to support the ability to present routes based on custom camera options.


Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -171,6 +171,8 @@ open class NavigationMapView: UIView {
             show([activeRoute])
         case .all(_):
             show(routes)
+        case .custom(cameraOptions: _):
+            show(routes)
```
MaximAlien, line 175: I think these two cases can be merged into one? I think argument name in `.custom` one can be also skipped as it's not used anyway.

MaximAlien:
Feel free to change base branch to `main` once again (sorry for the back and forth with this) before merge.

Sources/MapboxNavigation/RoutesPresentationStyle.swift
```diff
@@ -18,4 +19,11 @@ public enum RoutesPresentationStyle {
      If `false` geometry of only first route will be used. Defaults to `true`.
      */
     case all(shouldFit: Bool = true)
+    
+    /**
+     All routes will be presented on a map view with the custom camera options.
+     
+     - parameter cameraOptions: The custom `CameraOptions` that the map view ill be presented with.
+     */
+    case custom(cameraOptions: CameraOptions)
```
MaximAlien, line 28: Do you think that it'd make sense to also pass a lit of routes here so that user has the ability to choose? Since it's public API I'd rather make this decision right now.

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Yes. Because the `RoutesPresentationStyle` is used in `navigationMapView.fitCamera(to:routesPresentationStyle:animated:)`, the custom routes or route options could be decided by the `routes` property instead. The `RoutesPresentationStyle` should works for custom `cameraOptions` only.

Sources/MapboxNavigation/RoutesPresentationStyle.swift
ShanMa1991: Because the `RoutesPresentationStyle` is used in `navigationMapView.fitCamera(to:routesPresentationStyle:animated:)`, the custom routes or route options could be decided by the `routes` property instead. The `RoutesPresentationStyle` should works for custom `cameraOptions` only.

Sources/MapboxNavigation/RoutesPresentationStyle.swift
```diff
@@ -8,14 +10,17 @@ public enum RoutesPresentationStyle {
...
     /**
      All routes will be presented on a map view.
      
-     - parameter shouldFit: If `true` geometry of all routes will be used for camera transition.
-     If `false` geometry of only first route will be used. Defaults to `true`.
+     - parameter cameraOptions: The custom `CameraOptions` that the map view ill be presented with.
+     If no value provided, the map view will present all routes from above.
      */
-    case all(shouldFit: Bool = true)
+    case all(cameraOptions: CameraOptions? = nil)
```
MaximAlien, line 25: I think that `shouldFit` property should stay, as it's used to create geometry to either fit all routes/or fit only first route. `CameraOptions` should stay, as it's going to be used by application level to define padding, bearing or pitch where this geometry will be shown.

Please use this API in application, it'll give you better understanding about its usage.

Sources/MapboxNavigation/RoutesPresentationStyle.swift
```diff
@@ -1,4 +1,6 @@
 import Foundation
+import MapboxMaps
+import MapboxDirections
```
MaximAlien, line 3: Do we really need `MapboxDirections` here?

Sources/MapboxNavigation/RoutesPresentationStyle.swift
```diff
@@ -8,14 +10,17 @@ public enum RoutesPresentationStyle {
     
     /**
      Only first route will be presented on a map view.
+     
+     - parameter cameraOptions: The custom `CameraOptions` that the map view ill be presented with.
```
MaximAlien, line 14:
```suggestion
     - parameter cameraOptions: The custom `CameraOptions` that the map view will use while presenting a single route.
```

Sources/MapboxNavigation/RoutesPresentationStyle.swift
```diff
@@ -8,14 +10,17 @@ public enum RoutesPresentationStyle {
...
+     If no value provided, the map view will present the whole first route from above.
```
MaximAlien, line 15:
```suggestion
     If no value was provided, the map view will use default `CameraOptions` value, which is based on current safe area.
```

Sources/MapboxNavigation/RoutesPresentationStyle.swift
ShanMa1991: I see. So the `shouldFit` is provided for the geometry information of the `CameraOptions`. So what if users allow the `shouldFit` be true and also provide the custom `cameraOptions`? Because the geometry of the routes/route would limit the zoom and center. Or in this case, should we only adopt the custom `padding, bearing, pitch` from the `cameraOptions` and the geometry from the routes/route?

Sources/MapboxNavigation/RoutesPresentationStyle.swift
ShanMa1991: No, I'll remove it.

Sources/MapboxNavigation/RoutesPresentationStyle.swift
MaximAlien: I think second one, we should use geometry of the route(s), but `CameraOptions` from the another property, if it's not nil. `shouldFit` only allows to make a decision how geometry will look, not `CameraOptions`.

ShanMa1991: Got it. So right now the `geometry` is determined by the `single/all` and the `shouldFit`, while the `padding, bearing, pitch` is determined by the `CameraOptions`, with the default padding be the safe area.

Sources/MapboxNavigation/RoutesPresentationStyle.swift
ShanMa1991: Thanks for the suggestion, the documentation is improved.

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -1793,19 +1793,24 @@ open class NavigationMapView: UIView {
...
         
-        if case let RoutesPresentationStyle.all(shouldFit) = routesPresentationStyle, shouldFit {
-            geometry = .multiLineString(MultiLineString(routes.map({ $0.shape?.coordinates ?? [] })))
-        } else {
+        switch routesPresentationStyle {
+        case .single(cameraOptions: let cameraOptions):
             geometry = .lineString(LineString(routes.first?.shape?.coordinates ?? []))
+            customCameraOptions = cameraOptions
+        case .all(shouldFit: let shouldFit, cameraOptions: let cameraOptions):
+            geometry = shouldFit ? .multiLineString(MultiLineString(routes.map({ $0.shape?.coordinates ?? [] }))) : .lineString(LineString(routes.first?.shape?.coordinates ?? []))
```
MaximAlien, line 1803: Small nit: Ternary operators are a bit hard to read here, so I'd probably just use usual if/else.

MaximAlien:
Looks good to me, thanks.

MaximAlien (approved):

Sources/MapboxNavigation/NavigationMapView.swift
bamx23, line 1803: NIT: `compactMap` seems to fit here
```suggestion
            geometry = shouldFit ? .multiLineString(MultiLineString(routes.compactMap({ $0.shape?.coordinates }))) : .lineString(LineString(routes.first?.shape?.coordinates ?? []))
```

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -1793,19 +1793,24 @@ open class NavigationMapView: UIView {
...
+        let bearing = (customCameraOptions?.bearing == nil) ? nil : CGFloat(customCameraOptions!.bearing!)
         if let cameraOptions = mapView?.mapboxMap.camera(for: geometry,
-                                                            padding: edgeInsets,
-                                                            bearing: nil,
-                                                            pitch: nil) {
-            mapView?.camera.ease(to: cameraOptions, duration: animated ? 1.0 : 0.0)
+                                                            padding: customCameraOptions?.padding ?? edgeInsets,
+                                                            bearing: bearing,
+                                                            pitch: customCameraOptions?.pitch) {
+         mapView?.camera.ease(to: cameraOptions, duration: animated ? 1.0 : 0.0)
```
bamx23, line 1813:
```suggestion
            mapView?.camera.ease(to: cameraOptions, duration: animated ? 1.0 : 0.0)
```

Sources/MapboxNavigation/NavigationMapView.swift
```diff
@@ -1793,19 +1793,24 @@ open class NavigationMapView: UIView {
...
+        case .single(cameraOptions: let cameraOptions):
             geometry = .lineString(LineString(routes.first?.shape?.coordinates ?? []))
+            customCameraOptions = cameraOptions
+        case .all(shouldFit: let shouldFit, cameraOptions: let cameraOptions):
+            geometry = shouldFit ? .multiLineString(MultiLineString(routes.map({ $0.shape?.coordinates ?? [] }))) : .lineString(LineString(routes.first?.shape?.coordinates ?? []))
+            customCameraOptions = cameraOptions
         }
         
         let edgeInsets = safeArea + UIEdgeInsets.centerEdgeInsets
+        let bearing = (customCameraOptions?.bearing == nil) ? nil : CGFloat(customCameraOptions!.bearing!)
```
bamx23, line 1808: NIT: Not sure if it's easier to read, but no force unwraps
```suggestion
        let bearing = customCameraOptions.flatMap({ $0.bearing }).map({ CGFloat($0) })
```

bamx23 (approved):

Sources/MapboxNavigation/NavigationMapView.swift
ShanMa1991: Thanks for the suggestion, it has been updated.

PR #3683, opened 2022-01-12
Update dependencies: NavNative SDK 83.0.0 and Directions SDK 2.2.0-rc.1

bamx23:
In preparation for the release `v2.2.0-rc.1`.


Sources/MapboxCoreNavigation/NavigationStatus.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import MapboxNavigationNative
+
+extension NavigationStatus {
+    /// Legacy `roadName` property that returns first road name based on the `roads` array.
+    var roadName: String {
+        roads.map({ $0.text }).prefix(while: { $0 != "/" }).joined(separator: " ")
```
bamx23, line 7: This implementation is what NN docs say in case we want the removed `roadName` back.
But do we want to have a different implementation here?

S2Ler (approved):

Sources/MapboxCoreNavigation/NavigationStatus.swift
bamx23: I guess we can work on this later. I've tested that this function works and we get the same road name.

PR #3685, opened 2022-01-12
Fix tap gesture to stop camera transitions

Udumft:
Fixes and issue, where tapping `NavigationMapView` while it transitions the camera to or from `following/overview` states would leave it in `transitioning` state, and thus blocking switching to either mode.

### DIFF
--- a/Sources/MapboxNavigation/NavigationCamera.swift
+++ b/Sources/MapboxNavigation/NavigationCamera.swift
@@ -34,6 +34,7 @@ public class NavigationCamera: NSObject, ViewportDataSourceDelegate {
     
     func setupGestureRecognizers() {
         makeGestureRecognizersDisableCameraFollowing()
+        makeTapGestureRecognizerStopAnimatedTransitions()
     }
     
     // MARK: Reacting to ViewportDataSourceDelegate Updates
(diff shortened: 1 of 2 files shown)


Sources/MapboxNavigation/NavigationCamera.swift
```diff
@@ -34,6 +34,7 @@ public class NavigationCamera: NSObject, ViewportDataSourceDelegate {
     
     func setupGestureRecognizers() {
         makeGestureRecognizersDisableCameraFollowing()
+        makeTapGestureRecognizerStopAnimatedTransitions()
```
ShanMa1991, line 37: Because `func setupGestureRecognizers()` already provides the gestureRecognizer to stop the `cameraStateTransition`. Could we just add the `UITapGestureRecognizer` to it and add the `.following` state to the `func stop()`?

Udumft: This would create a misleading behavior. `makeGestureRecognizersDisableCameraFollowing` is meant to 'stop following', which is not the case for tap gesture. Conditioning logic inside this method to distinguish gesture recognizers, current states, etc. just complicates it. Also, users rely on public `stop()` method to cancel any transaction, so it would create a bug if this method would not do that under some conditions.

ShanMa1991 (approved):

PR #3696, opened 2022-01-21
Restore and fix disabled tests

S2Ler:
Fixes disabled before tests:
- Better Navigator reset method
- Hard retain in new reroute code
- An attempted fix to 'tripPerWaypoint' test

### DIFF
--- a/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift
+++ b/Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift
@@ -590,23 +590,25 @@ final class BillingHandlerUnitTests: TestCase {
         ])
     }
 
-    func disabled_testTripPerWaypoint() {
-        var routeOptions: NavigationRouteOptions {
-            let coordinates = [
-                CLLocationCoordinate2D(latitude: 37.751748, longitude: -122.387589),
-                CLLocationCoordinate2D(latitude: 37.752397, longitude: -122.387631),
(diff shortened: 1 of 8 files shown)


Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift
```diff
@@ -590,34 +590,36 @@ final class BillingHandlerUnitTests: TestCase {
...
 
-            let waypoints = coordinates.map({ Waypoint(coordinate: $0) })
 
             return NavigationRouteOptions(waypoints: waypoints)
         }
         
         let route = Fixture.route(from: "route-with-10-legs", options: routeOptions)
         
         autoreleasepool {
-            let replayLocations = Fixture.generateTrace(for: route).shiftedToPresent().qualified()
```
jill-cardamon, line 617: How can we drop `qualified()` here? Don't we need the route to be qualified for nav native?

jill-cardamon (approved):
lgtm otherwise :)

Tests/MapboxCoreNavigationTests/BillingHandlerTests.swift
S2Ler: It works without this `qualified` function. Maybe it was needed at some point in the past or needed for some usecases?

S2Ler: I will get back to this if we continue to see issues with this test.

PR #3700, opened 2022-01-24
Support to show and hide traffic data from V2 beta version tile set.

ShanMa1991:
### Description
This pr is to allow the mapView to show and hide traffic data from traffic v2 tile set.

### Implementation
Right now the `MapView.showsTileSet(withIdentifiers:layerIdentifier:)` and `MapView.setShowsTileSet(_:withIdentifiers:layerIdentifier:)` are both public to allow developers to show and hide custom  TileSet IDs for the map.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### Map
 
-* Added `MapView.showsTileSet(withIdentifiers:layerIdentifier:)` and `MapView.setShowsTileSet(_:withIdentifiers:layerIdentifier:)`to public to support developers show and hide custom TileSet IDs for the map view. ([#3700](https://github.com/mapbox/mapbox-navigation-ios/pull/3700))
+* Added `MapView.showsTileSet(with:layerIdentifier:)` and `MapView.setShowsTileSet(_:with:layerIdentifier:)` to provide the ability to show and hide custom tile set identifiers on the map view. ([#3700](https://github.com/mapbox/mapbox-navigation-ios/pull/3700))
 
 ### Offline routing
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 * Fixed cases in which the user location indicator floated around when the user was stopped at an intersection in an urban canyon. ([#3705](https://github.com/mapbox/mapbox-navigation-ios/pull/3705))
 * Fixed poor location snapping while the user is inside a tunnel. ([#3705](https://github.com/mapbox/mapbox-navigation-ios/pull/3705))
 
+### Map
+
+* Added `MapView.showsTileSet(withIdentifiers:layerIdentifier:)` and `MapView.setShowsTileSet(_:withIdentifiers:layerIdentifier:)`to public to support developers show and hide custom TileSet IDs for the map view. ([#3700](https://github.com/mapbox/mapbox-navigation-ios/pull/3700))
+
 ### Offline routing
 
 * If routing tiles in local storage are corrupted, the tiles are now redownloaded. ([#3705](https://github.com/mapbox/mapbox-navigation-ios/pull/3705))
--- a/Sources/MapboxNavigation/MapView.swift
+++ b/Sources/MapboxNavigation/MapView.swift
@@ -4,27 +4,38 @@ import MapboxMaps
 import MapboxDirections
 import MapboxCoreNavigation
 
+
+private let trafficTileSetIdentifiers = Set([
+    "mapbox.mapbox-traffic-v1",
+    "mapbox.mapbox-traffic-v2-beta"
+])
+
+private let incidentsTileSetIdentifiers = Set([
+    "mapbox.mapbox-incidents-v1",
+    "mapbox.mapbox-incidents-v2-beta"
+])
+
 /**
  An extension on `MapView` that allows for toggling traffic on a map style that contains a [Mapbox Traffic source](https://docs.mapbox.com/vector-tiles/mapbox-traffic-v1/).
  */
 extension MapView {
     
     /**
-     Returns a list of tile set identifiers for specific `sourceIdentifier`.
+     Returns a set of tile set identifiers for specific `sourceIdentifier`.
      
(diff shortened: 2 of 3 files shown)


Sources/MapboxNavigation/MapView.swift
```diff
@@ -50,24 +61,24 @@ extension MapView {
...
-     Returns a Boolean value indicating whether data from the given tile set layer is currently visible in the map view’s style.
+     Returns a Boolean value indicating whether data from the given tile set layers is currently visible in the map view’s style.
      
-     - parameter tileSetIdentifier: Identifier of the tile set in the form `user.tileset`.
+     - parameter tileSetIdentifiers: Identifiers of the tile sets in the form `user.tileset`.
      - parameter layerIdentifier: Identifier of the layer in the tile set; in other words, a source layer identifier. Not to be confused with a style layer.
      */
-    func showsTileSet(withIdentifier tileSetIdentifier: String, layerIdentifier: String) -> Bool {
-        let sourceIdentifiers = self.sourceIdentifiers(tileSetIdentifier)
+    public func showsTileSet(withIdentifiers tileSetIdentifiers: Set<String>, layerIdentifier: String) -> Bool {
```
MaximAlien, line 80: Does this function work as expected? Do we want it to return `true` only in case if one of the tile set identifiers is currently shown or all of them? Based on documentation it looks unclear.

ShanMa1991: Yes. Because when we have a set of tile sets, we may want to check whether all of them are currently shown. Thanks for the suggestion, I'll improve this part.

Sources/MapboxNavigation/MapView.swift
```diff
@@ -50,24 +61,25 @@ extension MapView {
...
-     Returns a Boolean value indicating whether data from the given tile set layer is currently visible in the map view’s style.
+     Returns a Boolean value indicating whether data from the given tile set layers is currently all visible in the map view’s style.
      
-     - parameter tileSetIdentifier: Identifier of the tile set in the form `user.tileset`.
+     - parameter tileSetIdentifiers: Identifiers of the tile sets in the form `user.tileset`.
      - parameter layerIdentifier: Identifier of the layer in the tile set; in other words, a source layer identifier. Not to be confused with a style layer.
      */
-    func showsTileSet(withIdentifier tileSetIdentifier: String, layerIdentifier: String) -> Bool {
-        let sourceIdentifiers = self.sourceIdentifiers(tileSetIdentifier)
+    public func showsTileSet(withIdentifiers tileSetIdentifiers: Set<String>, layerIdentifier: String) -> Bool {
```
MaximAlien, line 80:
```suggestion
    public func showsTileSet(with tileSetIdentifiers: Set<String>, layerIdentifier: String) -> Bool {
```

Sources/MapboxNavigation/MapView.swift
```diff
@@ -76,24 +88,26 @@ extension MapView {
...
+     Shows or hides data from the given tile set layers.
      
      - parameter isVisible: Parameter, which controls whether layer should be visible or not.
-     - parameter tileSetIdentifier: Identifier of the tile set in the form `user.tileset`.
+     - parameter tileSetIdentifiers: Identifiers of the tile sets in the form `user.tileset`.
      - parameter layerIdentifier: Identifier of the layer in the tile set; in other words, a source layer identifier. Not to be confused with a style layer.
      */
-    func setShowsTileSet(_ isVisible: Bool, withIdentifier tileSetIdentifier: String, layerIdentifier: String) {
-        let sourceIdentifiers = self.sourceIdentifiers(tileSetIdentifier)
+    public func setShowsTileSet(_ isVisible: Bool, withIdentifiers tileSetIdentifiers: Set<String>, layerIdentifier: String) {
```
MaximAlien, line 109:
```suggestion
    public func setShowsTileSet(_ isVisible: Bool, with tileSetIdentifiers: Set<String>, layerIdentifier: String) {
```

Sources/MapboxNavigation/MapView.swift
```diff
@@ -4,27 +4,38 @@ import MapboxMaps
...
     /**
-     Returns a list of tile set identifiers for specific `sourceIdentifier`.
+     Returns a set of tile set identifiers for specific `sourceIdentifier`.
      
      - parameter sourceIdentifier: Identifier of the source, which will be searched for in current style of the `MapView`.
-     - returns: List of tile set identifiers.
+     - returns: Set of tile set identifiers.
      */
-    func tileSetIdentifiers(_ sourceIdentifier: String) -> [String] {
+    func tileSetIdentifiersSet(_ sourceIdentifier: String) -> Set<String> {
```
MaximAlien, line 29: I do not think that we usually mention returned type in method signature. Shall we rename it to `tileSetIdentifiers` even though if we have two methods, which return array and set?

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 * Fixed cases in which the user location indicator floated around when the user was stopped at an intersection in an urban canyon. ([#3705](https://github.com/mapbox/mapbox-navigation-ios/pull/3705))
 * Fixed poor location snapping while the user is inside a tunnel. ([#3705](https://github.com/mapbox/mapbox-navigation-ios/pull/3705))
 
+### Map
+
+* Added `MapView.showsTileSet(withIdentifiers:layerIdentifier:)` and `MapView.setShowsTileSet(_:withIdentifiers:layerIdentifier:)`to public to support developers show and hide custom TileSet IDs for the map view. ([#3700](https://github.com/mapbox/mapbox-navigation-ios/pull/3700))
```
MaximAlien, line 17:
```suggestion
* Added `MapView.showsTileSet(with:layerIdentifier:)` and `MapView.setShowsTileSet(_:with:layerIdentifier:)` to to provide the ability to show and hide custom tile set identifiers on the map view. ([#3700](https://github.com/mapbox/mapbox-navigation-ios/pull/3700))
```

Sources/MapboxNavigation/MapView.swift
ShanMa1991: Yes, thanks for the suggestion, I'll rename it to `tileSetIdentifiers`

PR #3701, opened 2022-01-24
Add initializers to RoundingTable and Threshold

jill-cardamon:
This PR adds initializers to the now public structs `RoundingTable` and `Threshold`.

### DIFF
--- a/Sources/MapboxCoreNavigation/DistanceFormatter.swift
+++ b/Sources/MapboxCoreNavigation/DistanceFormatter.swift
@@ -67,7 +67,7 @@ public struct RoundingTable {
     }
     
     /**
-     :no-doc:
+     :nodoc:
      Initializes a `RoundingTable` with the given thresholds.
      - parameter thresholds: An array of `Threshold`s that dictate rounding behavior.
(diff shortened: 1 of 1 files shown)


Sources/MapboxCoreNavigation/DistanceFormatter.swift
```diff
@@ -56,6 +66,15 @@ public struct RoundingTable {
         } ?? thresholds.last!
     }
     
+    /**
+     :no-doc:
```
1ec5, line 70: FYI, jazzy only recognizes this spelling:

```suggestion
     :nodoc:
```

In practice, though, the documentation gets hidden anyways because of the `:nodoc:`s on `RoundingTable` and `RoundingTable.Threshold`.

1ec5 (approved):

Sources/MapboxCoreNavigation/DistanceFormatter.swift
jill-cardamon: ah oops. thanks for the catch.

PR #3702, opened 2022-01-25
[bugfix] Fix Direction.calculateOffline using hybrid router

Udumft:
This PR fixes `Directions.calculateOffline` method to use strictly offline routing provider.


bamx23 (approved):
Do we need a changelog line for this?

PR #3704, opened 2022-01-26
Instructions card improvements.

MaximAlien:
PR improves such areas of instructions card functionality:
- Public APIs are documented and improved by providing additional parameters, which allow to present `LanesView` and `NextBannerView` in both animated/non-animated ways. This is needed for default steps presentation via `StepsViewController` (animated) and `InstructionsCardViewController` (not animated).
- Instruction cards presentation is simplified by removing complex logic with gradient layers handling.
- Instruction cards layout is improved for several edge cases.

Examples:

Simulator Screen Shot - iPhone 8 Plus iOS 14 1 - 2022-01-26 at 15 23 58

Simulator Screen Shot - iPhone 13 - 2022-01-26 at 15 26 44

Simulator Screen Shot - iPhone 13 - 2022-01-26 at 15 27 11


Sources/MapboxNavigation/InstructionsCardView.swift
```diff
@@ -7,16 +7,14 @@ public class InstructionsCardView: BaseInstructionsBannerView {
     
     var step: RouteStep!
     var distanceFromCurrentLocation: CLLocationDistance!
-    var gradientLayer: CAGradientLayer!
```
1ec5, line 10: The original design of the guidance card layout emphasized the gradients, but it would make sense to change the design at this point to be flatter. My only concern is that, without a gradient, the card can often lack contrast with the surrounding map. In the default day style, a white, non-immediate card is barely discernible from the map with a white background. We could mitigate this issue in the same way as we’ve done with the banners: by applying a subtle outline around the view.

MaximAlien: Agree, added stylable border in latest commit. Some example:

Simulator Screen Shot - iPhone 12 - 2022-01-28 at 15 00 18

Simulator Screen Shot - iPhone 12 - 2022-01-28 at 15 02 46

Sources/MapboxNavigation/Constants+InstructionsCard.swift
```diff
@@ -1,19 +0,0 @@
-import UIKit
-import CoreLocation
-
-/// :nodoc:
-struct InstructionsCardConstants {
```
MaximAlien, line 5: These properties are no longer used anywhere (except `highlightDistance`, which was moved to `Constants.swift` and renamed to `InstructionCardHighlightDistance`).

MaximAlien:
> * _removed_ method: `update(for:)` in `NextBannerView`
> * _removed_ method: `update(for:)` in `LanesView`

Even though these two public methods were modified, I do not consider it as a breaking API change, as every new parameter has default value.

azarovalex (approved):

Sources/MapboxNavigation/InstructionsCardCell.swift
```diff
@@ -24,23 +25,22 @@ public class InstructionsCardCell: UICollectionViewCell {
...
+    
+    public func configure(for step: RouteStep,
+                          distance: CLLocationDistance,
+                          instruction: VisualInstructionBanner? = nil,
+                          isCurrentCardStep: Bool = false) {
         container.updateInstruction(for: step,
-                                    distance: distance,
-                                    instruction: instruction,
-                                    isCurrentCardStep: isCurrentCardStep)
+                                       distance: distance,
```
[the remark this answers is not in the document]
MaximAlien: I think such indentation is correct for methods, which have non-empty first argument.

Sources/MapboxNavigation/InstructionsCardViewController.swift
```diff
@@ -173,18 +173,49 @@ open class InstructionsCardViewController: UIViewController {
...
+                                                           constant: 10.0),
+            instructionCollectionView.trailingAnchor.constraint(equalTo: view.safeTrailingAnchor,
+                                                                constant: -10.0),
             instructionCollectionView.heightAnchor.constraint(equalToConstant: cardSize.height),
-            instructionCollectionView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            instructionCollectionView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
         ]
         
+        // Device is in landscape mode and notch (if present) is located on the left side.
+        if UIApplication.shared.statusBarOrientation == .landscapeRight {
```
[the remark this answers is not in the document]
MaximAlien: No, instructions card functionality is only available on iOS. CarPlay managers instructions on its own.

Issue #3707, opened 2022-01-27
[Bug]: Navigation while offline fails to re-route


dmiluski:
### Mapbox Navigation SDK version

2.2.0

### Steps to reproduce

Able to be demonstrated using Straightaway-iOS v2.37.2

If using Straightaway

1. Download Offline Region
2. Setup Route (> 3 items) contained within region
3. Start Route (just means we show our next screen)
4. Choose preferred route (of alternates list) to begin navigation
5. Turn off wifi/carrier
6. Miss a turn

Navigation ViewController identifies a re-routable event, but is unable to complete re-route

```
▿ DirectionsError
  ▿ network : URLError
    - _nsError : Error Domain=NSURLErrorDomain Code=-1009 "The Internet connection appears to be offline." UserInfo={_kCFStreamErrorCodeKey=50, NSUnderlyingError=0x600004b6a340 {Error Domain=kCFErrorDomainCFNetwork Code=-1009 "(null)" UserInfo={_NSURLErrorNWPathKey=unsatisfied (No network route), _kCFStreamErrorCodeKey=50, _kCFStreamErrorDomainKey=1}}, _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <81557272-8141-4C30-A9EC-5B1BC945264A>.<37>, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalDataTask <81557272-8141-4C30-A9EC-5B1BC945264A>.<37>"
[...]

PR #3708, opened 2022-01-27
[Issue-3707] Support Offline Navigation Rerouting

dmiluski:
## Issue:

Fixes https://github.com/mapbox/mapbox-navigation-ios/issues/3707

## Problem:

If offline/flakey network, a user is unable to reroute during navigation.

## Why:

Even if app/sdk has offline tilesets loaded, this data isn't leveraged for reroute given the default use of `Directions.shared.calculate(::)`. vs method that takes offline tiles into consideration.

## Change:

- Swap Directions method implementation call to leverage offline data

## Alternatives

- Provide own RoutingController in our app. This seemed like overkill given the existing functionality.
- Injectable options? Allows for injection of options, but I"m not sure how useful that is just yet? Would defer to the team in case there are better approaches to support?

### DIFF
--- a/Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
+++ b/Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
@@ -8,30 +8,30 @@ extension URLSessionDataTask: NavigationProviderRequest {
 }
 
 extension Directions: RoutingProvider {
-    
+
     @discardableResult public func calculateRoutes(options: RouteOptions, completionHandler: @escaping RouteCompletionHandler) -> NavigationProviderRequest? {
-        return calculate(options, completionHandler: completionHandler)
+        return calculateWithCache(options, completionHandler: completionHandler)
     }
-    
+
     @discardableResult public func calculateRoutes(options: MatchOptions, completionHandler: @escaping MatchCompletionHandler) -> NavigationProviderRequest? {
-        return calculate(options, completionHandler: completionHandler)
+        return calculateWithCache(options, completionHandler: completionHandler)
     }
-    
+
     @discardableResult public func refreshRoute(indexedRouteResponse: IndexedRouteResponse, fromLegAtIndex: UInt32, completionHandler: @escaping RouteCompletionHandler) -> NavigationProviderRequest? {
         guard case let .route(routeOptions) = indexedRouteResponse.routeResponse.options else {
             preconditionFailure("Invalid route data passed for refreshing. Expected `RouteResponse` containing `.route` `ResponseOptions` but got `.match`.")
         }
-        
+
         let session = (options: routeOptions as DirectionsOptions,
                        credentials: self.credentials)
-        
+
         guard let responseIdentifier = indexedRouteResponse.routeResponse.identifier else {
             DispatchQueue.main.async {
                 completionHandler(session, .failure(.noData))
             }
             return nil
         }
-        
+
         return refreshRoute(responseIdentifier: responseIdentifier,
                             routeIndex: indexedRouteResponse.routeIndex,
                             fromLegAtIndex: Int(fromLegAtIndex),
@@ -73,7 +73,7 @@ extension Route {
         let encoder = JSONEncoder()
         encoder.userInfo[.options] = options
         let encoded = try encoder.encode(self)
-        
+
         let decoder = JSONDecoder()
         decoder.userInfo[.options] = options
         return try decoder.decode(Self.self, from: encoded)


Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
```diff
@@ -8,30 +8,30 @@ extension URLSessionDataTask: NavigationProviderRequest {
...
+
     @discardableResult public func calculateRoutes(options: RouteOptions, completionHandler: @escaping RouteCompletionHandler) -> NavigationProviderRequest? {
-        return calculate(options, completionHandler: completionHandler)
+        return calculateWithCache(options, completionHandler: completionHandler)
     }
-    
+
     @discardableResult public func calculateRoutes(options: MatchOptions, completionHandler: @escaping MatchCompletionHandler) -> NavigationProviderRequest? {
-        return calculate(options, completionHandler: completionHandler)
+        return calculateWithCache(options, completionHandler: completionHandler)
```
dmiluski, line 17: Given this Release Doc deprecation of calculateWithCache, perhaps there is a better way/method/injection to provide this service.
https://github.com/mapbox/mapbox-navigation-ios/blob/main/CHANGELOG.md#location-tracking-2

Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
```diff
@@ -8,30 +8,30 @@ extension URLSessionDataTask: NavigationProviderRequest {
 }
 
 extension Directions: RoutingProvider {
-    
+
     @discardableResult public func calculateRoutes(options: RouteOptions, completionHandler: @escaping RouteCompletionHandler) -> NavigationProviderRequest? {
-        return calculate(options, completionHandler: completionHandler)
+        return calculateWithCache(options, completionHandler: completionHandler)
```
1ec5, line 13: @Udumft, was it intentional that this method calls the more generic calculation method as of #3261, versus what MapboxRoutingProvider does?

Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
```diff
@@ -8,30 +8,30 @@ extension URLSessionDataTask: NavigationProviderRequest {
 }
 
 extension Directions: RoutingProvider {
-    
+
     @discardableResult public func calculateRoutes(options: RouteOptions, completionHandler: @escaping RouteCompletionHandler) -> NavigationProviderRequest? {
-        return calculate(options, completionHandler: completionHandler)
```
dmiluski, line 13: Source of the offline route not functioning. If we instead call calculateWithCache, the Nav UI should also be able to leverage offline routing functionality.

Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
Udumft: Yes, I intentionally did not use offline capabilities here, to maintain usual Directions behavior. This protocol conformance just wraps Directions features, but since Directions on its own does only online calculations - I didn't use `calculateWithCache`.

Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
Udumft: Did you try to use `MapboxRoutingProvider` instead? It is meant to be offline-ready, while Directions is online-only provider. I would suggest not to change these lines since it effectively makes `Directions` object a wrapper for `MapboxRoutingProvider` which is implicit and not desired.

dmiluski: The issue isn't out app's code base. The issue is that NavSDK isn't leveraging an offline friendly approach.

Are you suggesting changing the method call here? Or somewhere else?

dmiluski: Hi @Udumft  ,

I’m looking to address navigation being unable to reroute if offline.

While this approach may not be ideal, if you have a preferred approach i would be happy following it.

The problem I'm running into is that we’ve advertised offline navigation and although it works in other parts of our app, **specifically during navigation** it's failing due to not using a method capable of falling back to offline calculations

dmiluski:
Marking this closed given the option to inject a NavigationService.

Given NavigationSDK uses Directions.shared by default, offline re-routing isn't implicitly supported. To get offline routing to function, we needto inject an offline routing provider MapboxRoutingProvider. 

To do so, we can construct our own [NavigationService](https://github.com/mapbox/mapbox-navigation-ios/blob/main/Sources/MapboxNavigation/NavigationViewController.swift#L217) modeled after MapboxNavigationSDK's initializer/setup which provides the service and provide it to our NavigationViewController for use which we implemented in [Straightaway](https://github.com/mapbox/straightaway-ios/pull/1475)

Sources/MapboxCoreNavigation/Directions+RoutingProvider.swift
dmiluski: MapboxRoutingProvider does function, I just wasn't sure how best to force the NavSDK to use it. Given your slack suggestion, I went with the injected NavigationService which allows configuration of a RoutingProvider.

Issue #3435, opened 2021-10-05
[Enhancement] Original CPRouteChoice.userInfo data should not be overridden by a developer


tsuz:
**Why**

As per [this code][1], `CPRouteChoice.userInfo` can be overwritten by a developer unknowingly and this can cause some code to not execute. 

**Proposals**

- Additional `userInfo` can be added to CPRouteChoice without modifying the data that navigation needs
- Create a detailed documentation on what parameters cannot be overwritten in `userInfo`

**Priority**

Low ~ Medium: There is no immediate impact at this moment but this is a known risk for current use.

[1]: https://github.com/mapbox/mapbox-navigation-ios/blob/v2.0.0-rc.3/Sources/MapboxNavigation/CarPlayManager.swift#L587

MaximAlien:
Based on priority moved to `v2.2`.

PR #3709, opened 2022-01-28
Give users the ability to modify `userInfo` dictionary in `CPRouteChoice` and `CPListItem`.

MaximAlien:
Closing #3435.

In scope of this PR new `CarPlayUserInfo` dictionary is publicly exposed to give the ability to modify data, which is passed by user while performing search or requesting a route on CarPlay.

CarPlay tests were refactored to make them as independent as possible. For example individual tests use their own `CarPlayManagerDelegate` and test only small set of delegate method. This replaces one large general purpose delegate, which was hard to debug and read.

There is is still some tail-work to be done, such as adding more tests for `CarPlaySearchController` etc. It'll be tackled in scope of subsequent PRs in scope of https://github.com/mapbox/mapbox-navigation-ios/issues/3726.

/cc @ChristianSteffens


Sources/MapboxNavigation/CPRouteChoice.swift
```diff
@@ -0,0 +1,34 @@
...
+        
+        /**
+         Index of the route.
+         */
+        let routeIndex: Int
+        
+        /**
+         Options, which were used for calculating results from the Mapbox Directions service.
+         */
+        let options: DirectionsOptions
```
MaximAlien, line 24: Not sure why we exposed `options` separately in tuple, as it's already available in `response`. Will probably remove it, so that `RouteResponseUserInfo` only contains `routeResponse` and `routeIndex` properties.

Tests/MapboxNavigationTests/CarPlayManagerSpec.swift
```diff
@@ -0,0 +1,206 @@
...
+        }
+        
+        describe("Starting an invalid trip.") {
+            context("Precondition should be triggered if CPRouteChoice is invalid.") {
+                let routeChoice = createInvalidRouteChoice()
+                let trip = createTrip(routeChoice)
+                let mapTemplate = CPMapTemplate()
+                
+                expect {
+                    carPlayManager.mapTemplate(mapTemplate, startedTrip: trip, using: routeChoice)
```
MaximAlien, line 184: In case if uses passes `CPRouteChoice`, which has `userInfo` set to type different than `CarPlayUserInfo` - assert.

Tests/MapboxNavigationTests/CarPlayManagerTests.swift
```diff
@@ -1,441 +1,276 @@
...
-            return customTripPreviewTextConfiguration ?? previewTextConfiguration
-        }
-
-        func carPlayManager(_ carPlayManager: CarPlayManager, didBeginNavigationWith service: NavigationService) {
-            //no-op
-        }
-
-        func carPlayManagerDidEndNavigation(_ carPlayManager: CarPlayManager) {
-            //no-op
-        }
```
MaximAlien, line 415: No-op delegate methods are not needed, removed them in `CarPlayManagerSpec.swift`.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
 
 * Added `MapView.showsTileSet(with:layerIdentifier:)` and `MapView.setShowsTileSet(_:with:layerIdentifier:)` to provide the ability to show and hide custom tile set identifiers on the map view. ([#3700](https://github.com/mapbox/mapbox-navigation-ios/pull/3700))
 
+### CarPlay
+
+* Added `CarPlayUserInfo` type alias for storing CarPlay-related user information. ([#3709](https://github.com/mapbox/mapbox-navigation-ios/pull/3709))
```
1ec5, line 21: When introducing a new type, the changelog entry should ideally focus on where the developer will encounter this type. In this case, we can say that, whenever a `CarPlayDelegate` method is called, the `userInfo` property of any passed-in CarPlay object is of type `CarPlayUserInfo`.

Sources/MapboxNavigation/CPRouteChoice.swift
```diff
@@ -0,0 +1,34 @@
+import CarPlay
+import MapboxDirections
+
+@available(iOS 12.0, *)
+extension CPRouteChoice {
+    
+    struct RouteResponseUserInfo {
+        
+        static let key = "com.mapbox.navigation.cpRouteChoice.routeResponse"
```
1ec5, line 9: Let’s use `Bundle.mapboxNavigation.bundleIdentifier` to compose this key for consistency with other keys in the codebase:

https://github.com/mapbox/mapbox-navigation-ios/blob/5018a7f4b32e9aeebf20a32f2cb6169c43b89723/Sources/MapboxNavigation/NavigationMapViewIdentifiers.swift#L5

Sources/MapboxNavigation/CPRouteChoice.swift
```diff
@@ -0,0 +1,34 @@
+import CarPlay
+import MapboxDirections
+
+@available(iOS 12.0, *)
+extension CPRouteChoice {
+    
+    struct RouteResponseUserInfo {
```
1ec5, line 7: Do we expect these structs in the user info dictionary to be accessible to application code? If so, we’ll need to apply access control keywords and potentially add an initializer.

Sources/MapboxNavigation/CPRouteChoice.swift
```diff
@@ -0,0 +1,34 @@
...
+         */
+        let routeIndex: Int
+        
+        /**
+         Options, which were used for calculating results from the Mapbox Directions service.
+         */
+        let options: DirectionsOptions
+    }
+    
+    var routeResponseFromUserInfo: RouteResponseUserInfo? {
```
1ec5, line 27: Is it important to put this data into a formal struct, or should it remain a loosely typed dictionary with defined keys for forward-compatibility?

1ec5 (approved):
No red flags, but some things to think about for forward compatibility.

Sources/MapboxNavigation/CPRouteChoice.swift
MaximAlien: There is no intention to make them public at this point, so I guess we can omit any modifications to make them more accessible.

Sources/MapboxNavigation/CPRouteChoice.swift
MaximAlien: I do not think that it's critical to store values in separate keys of the dictionary. Storing data in internal struct also helps with readability.

Sources/MapboxNavigation/CPTrip.swift
```diff
@@ -23,17 +26,30 @@ extension CPTrip {
...
+              }
+        
+        let originMapItem = MKMapItem(placemark: MKPlacemark(coordinate: originCoordinate))
+        originMapItem.name = waypoints.first?.name
+        
+        let destinationMapItem = MKMapItem(placemark: MKPlacemark(coordinate: destinationCoordinate))
+        destinationMapItem.name = waypoints.last?.name
         
-        self.init(origin: origin, destination: destination, routeChoices: routeChoices)
-        userInfo = options!
```
MaximAlien, line 37: This `userInfo` is not used anywhere at this point, so I'm removing it.

PR #3713, opened 2022-01-28
Add the Sprite and Mapbox designed shield caching and retrieval mechanism.

ShanMa1991:
This pr is to add the `SpriteMetaDataCache` and `SpriteRepository` to support the Sprite caching and Mapbox designed shield retrieval from the memory caching.

By adding the `SpriteRepository` to `NavigationMapView` or `NavigationViewController` or `CarPlayNavigationViewController`, when the style change, provide the new `StyleURI` to the `SpriteRepository.updateRepository(updateStyleURI:updateBaseURL:completion:)` to update the caching of Sprite image and Sprite metadata.

When receiving the `Shield` information, compares the `Shield.baseURL`  with `SpriteRepository.baseURL`. If same,  `SpriteRepository.getShield(displayRef:name:)` with the `Shield.displayRef.` and `Shield.name` to retrieve the single shield icon as `UIImage` format. if different, call `SpriteRepository.updateRepository(updateStyleURI:updateBaseURL:completion:)` to update the repository and then `getShield(displayRef:name:)` in the completion.

If the `Shield` is not available, call `SpriteRepository.SpriteRepository(imageBaseUrl:completion:)` to get the legacy shield icon image instead.


Tests/MapboxNavigationTests/ImageLoadingURLProtocolSpy.swift
```diff
@@ -72,16 +72,13 @@ final class ImageLoadingURLProtocolSpy: URLProtocol {
 
                 switch result {
                 case .success(let imageData):
-                    guard let image: UIImage = UIImage(data: imageData) else {
-                        XCTFail("Invalid image data found"); return
-                    }
                     // send an NSHTTPURLResponse to the client
```
ShanMa1991, line 75: To remove the check of the `UIImage(data: imageData)` is because the `SpriteRepository` is using the `ImageDownloader` to download the data for `SpriteMetaDataCache` as well. The data is not for UIImage, and it blocks the test of `testDownLoadingSpriteMetaData`.

Sources/MapboxNavigation/SpriteRepository.swift
```diff
@@ -0,0 +1,136 @@
...
+    func updateRepository(updateStyleURI: StyleURI? = nil, updateBaseURL: String? = nil, completion: @escaping CompletionHandler) {
+        let styleURI = updateStyleURI ?? styleURI
+        let baseURL = updateBaseURL ?? baseURL
+        
+        guard let styleID = styleURI.rawValue.components(separatedBy: "styles")[safe: 1],
+              let accessToken = NavigationSettings.shared.directions.credentials.accessToken,
+              let spriteRequestURL = URL(string: baseURL + styleID + "/sprite@2x.png?access_token=" + accessToken),
+              let metadataRequestURL = URL(string: baseURL + styleID + "/sprite@2x?access_token=" + accessToken) else {
+                  return
+              }
```
MaximAlien, line 34: Let's use [URLComponents](https://developer.apple.com/documentation/foundation/urlcomponents) for URL generation.

ShanMa1991: Thanks for the suggestion, I'll replace it.

Sources/MapboxNavigation/SpriteMetadataCache.swift
```diff
@@ -0,0 +1,84 @@
+import UIKit
+
+public struct SpriteMetaData: Codable, Equatable {
```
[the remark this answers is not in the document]
1ec5: Add a comment mentioning [this API endpoint documentation](https://docs.mapbox.com/api/maps/styles/#retrieve-a-sprite-image-or-json) for future reference. We can mark the comment as `:nodoc:` if we still want to hide this struct from the public documentation for now.

Sources/MapboxNavigation/SpriteRepository.swift
```diff
@@ -0,0 +1,147 @@
...
+                completion(data)
+                return
+            }
+
+            strongSelf.metadataCache.store(data)
+            completion(data)
+        })
+    }
+    
+    func downLoadSprite(_ spriteURL: URL, completion: @escaping (UIImage?) -> Void) {
```
1ec5, line 85: Nit: “Download” is one word, so `downloadSprite`.

Sources/MapboxNavigation/SpriteRepository.swift
```diff
@@ -0,0 +1,147 @@
...
+        let shieldRect = CGRect(x: spriteMetaData.x, y: spriteMetaData.y, width: spriteMetaData.width, height: spriteMetaData.height)
+        if let croppedCGIImage = spriteImage.cgImage?.cropping(to: shieldRect) {
+            return UIImage(cgImage: croppedCGIImage)
+        }
+        
+        return nil
+    }
+    
+    func getLegacyShield(imageBaseUrl: String, completion: @escaping (UIImage?) -> Void) {
+        guard let requestURL = URL(string: imageBaseUrl + "@2x.png") else {
```
1ec5, line 121: Do we need to hardcode 2× resolution here, or can we make it match the screen resolution like before? Also, is there any chance of reusing the existing URL calculation code from `VisualInstruction.Component.ImageRepresentation` in MapboxDirections?

Sources/MapboxNavigation/SpriteRepository.swift
```diff
@@ -0,0 +1,147 @@
...
+              let accessToken = NavigationSettings.shared.directions.credentials.accessToken else { return nil }
+        
+        let requestTyep = isImage ? "/sprite@2x.png" : "/sprite@2x"
+        urlComponent.path += styleID
+        urlComponent.path += requestTyep
+        urlComponent.queryItems = [URLQueryItem(name: "access_token", value: accessToken)]
+        return urlComponent.url
+    }
+    
+    func downloadMetadata(_ metadataURL: URL, completion: @escaping (Data?) -> Void) {
```
1ec5, line 66: SpriteRepository is functioning similarly to the Directions class of MapboxDirections or the [Snapshot](https://github.com/mapbox/MapboxStatic.swift/blob/083e3daaa60ecdc142ca2e0360288cd233d4c851/Sources/MapboxStatic/Snapshot.swift#L192) class of MapboxStatic, so as tail work, it would be great to align the method signatures for consistency and future extensibility. For example, this method could return a URLSessionDataTask to enable canceling redundant requests.

Sources/MapboxNavigation/SpriteRepository.swift
```diff
@@ -0,0 +1,147 @@
...
+            self.baseURL = baseURL
+            completion()
+        }
+    }
+    
+    func spriteURL(isImage: Bool, baseURL: URL, styleID: String) -> URL? {
+        guard var urlComponent = URLComponents(url: baseURL, resolvingAgainstBaseURL: false),
+              let accessToken = NavigationSettings.shared.directions.credentials.accessToken else { return nil }
+        
+        let requestTyep = isImage ? "/sprite@2x.png" : "/sprite@2x"
```
1ec5, line 59: Typo: `requestType`.

1ec5: Should we be hardcoding 2× resolution here?

1ec5 (approved):
Some minor fit-and-finish comments, but the functionality looks great, and I appreciate the attention given to the legacy fallback. In the long term, we might want to factor out this and other server API request and image response code into MapboxCoreNavigation or even MapboxStatic, but keeping it here in MapboxNavigation will enable us to iterate on the shield work faster for now.

Sources/MapboxNavigation/SpriteRepository.swift
ShanMa1991: Yes. In fact, it's a good idea to use the `ImageRepresentation` for the `SpriteRepository`, it combines the `shield` and  `imageBaseUrl` , and could be used to download both the shield and legacy shield at the same time and store it. Thanks for the suggestion, I'll update this part.

Sources/MapboxNavigation/SpriteRepository.swift
ShanMa1991: I'll use the` VisualInstruction.Component.scale` instead.

Sources/MapboxNavigation/SpriteRepository.swift
ShanMa1991: Thanks for the suggestion, I'll make it a tail work.

PR #3717, opened 2022-02-01
Tileset profile configuration

Udumft:
This PR adds functionality to customize tiles dataset profile when creating `TileEndpointConfiguration`. Configuration is done through defaulted properties, and takes into account `RouteOptions.profileIdentifier` to match user-selected profile.


Sources/MapboxCoreNavigation/HandlerFactory.swift
```diff
@@ -8,6 +8,7 @@ protocol HandlerData {
     var historyDirectoryURL: URL? { get }
     var targetVersion: String? { get }
     var configFactoryType: ConfigFactory.Type { get }
+    var datasetProfile: ProfileIdentifier { get }
```
Udumft, line 11: Having this parameter as a part of caching key has a nuance: since tiles endpoint currently returns the same tiles regardless of our `TilesEndpointConfiguration.dataset` - we are technically don't need to track this parameter. Currently this  may produce "unexpected" re-caching, when user does modify `ProfileIdentifier` in his `RouteOptions` but forgets to do so in `MapboxRoutingProvider` (which may be used even before the actual `NavigationViewController` to request the initial route) or in `TilesetDescriptorFactory`.
In general, this setting now looks like something `NavigationSettings` should take care of, but since it is part of `DirectionsOptions` we have to expose it as an argument. It would help to make such arguments with no default value to highlight the change and force user to setup correct profile, but it is not feasible since we are not modifying public API in patch releases.
Maybe we should consider pushing this feature to the next major release then? /cc @1ec5

1ec5: Tricky situation. I think setting a default value of `automobile` in `MapboxCoreNavigation.Navigator.datasetProfile` seems reasonable for now. If you have ideas about how to simplify this feature across the two libraries, let’s ticket it out and put it on the v3.0 milestone for later consideration.

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -44,6 +44,14 @@ class Navigator {
     
     static func _recreateNavigator() { _navigator = .init() }
     
+    /**
+     Profile setting, used for selecting tiles type for navigation.
+     
+     This property can only be modified before creating `Navigator` shared instance, all
+     further changes to this property will have no effect. Defaults to `automobile`.
+     */
+    static var datasetProfile: ProfileIdentifier = .automobile
```
bamx23, line 53: Could it be placed into `NavigationSettings`? So that we don't have to provide it in constructors of `PassiveLocationManager` and `RouteController`.

Udumft: This setting should mirror `NavigationOptions.datasetProfile`, so having a parallel value in `NavigationSettings` will be confusing. And we cannot remove it from "options" too. See [this thread](https://github.com/mapbox/mapbox-navigation-ios/pull/3717#discussion_r796439137) for more details about exposing it through constructors.

Sources/MapboxCoreNavigation/HandlerFactory.swift
Udumft: I've captured my thoughts in corresponding discussion about our future Nav Settings with reference to this ticket as an example.

Udumft:
The breaking changes detected above are levelled out by default values of new arguments, so there should be no changes in user code unless he wants to use new feature.

Sources/MapboxCoreNavigation/TilesetDescriptorFactory.swift
```diff
@@ -9,13 +9,16 @@ extension TilesetDescriptorFactory {
      - Parameters:
        - version: TilesetDescriptor version.
        - completionQueue: A DispatchQueue on which the completion will be called.
+       - datasetProfile: Profile setting, used for selecting tiles type for navigation.
        - completion: A completion that will be used to pass the tileset descriptor.
      */
     public class func getSpecificVersion(version: String,
```
1ec5, line 15: Does the fact that TilesetDescriptorFactory is defined in MapboxNavigationNative prevent these methods from showing up in the docset?

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
...
 * Fixed an issue where `SpeechSynthesizing.managesAudioSession` was ignored by `RouteVoiceController`. ([#3572](https://github.com/mapbox/mapbox-navigation-ios/pull/3572))
 * Location tracking and routing resources are now freed if no `RouteController` or `PassiveLocationManager` instance is actively being used. ([#3724](https://github.com/mapbox/mapbox-navigation-ios/pull/3724))
+* Added tileset profile configuration for Free Drive, Active Guidance, `MapboRoutingProvider` and Tileset Descriptors API. Now, `RouteOptions.profileIdentifier` is used to determine which type of tiles should be used for navigation. Please make sure to pay attention to `MapboxRoutingProfider` and `TilesetDescriptorFactory` arguments when configuring `ProfileIdentifier`. ([#3717](https://github.com/mapbox/mapbox-navigation-ios/pull/3717))
```
1ec5, line 41: Free Drive, Active Guidance and Tileset Descriptors API aren’t terms we use anywhere else in the documentation, at least not as formal, capitalized names of APIs that look like server APIs. Let’s refer to specific classes or methods that the developer can use. Something like this (but please fact-check the specific symbols I named):

```suggestion
* Added an optional `datasetProfile` argument to the `MapboxRoutingProvider(_:settings:datasetProfile:)`, `PassiveLocationManager(directions:systemLocationManager:eventsManagerType:userInfo:datasetProfile:),` `TilesetDescriptorFactory.getSpecificVersion(version:completionQueue:datasetProfile:completion:)`, and `TilesetDescriptorFactory.getLatest(completionQueue:datasetProfile:completion:)` methods for obtaining routing tiles optimized for a particular mode of transportation. Make sure to configure `MapboxRoutingProvider` and `TilesetDescriptorFactory` with the correct dataset profile if you customize `Directions.profileIdentifier`. ([#3717](https://github.com/mapbox/mapbox-navigation-ios/pull/3717))
```

Sources/MapboxCoreNavigation/CoreNavigationNavigator.swift
```diff
@@ -54,6 +54,14 @@ class Navigator {
     // Used in tests to recreate the navigator
     static func _recreateNavigator() { _navigator = nil }
     
+    /**
+     Profile setting, used for selecting tiles type for navigation.
+     
+     This property can only be modified before creating `Navigator` shared instance, all
+     further changes to this property will have no effect. Defaults to `automobile`.
+     */
+    static var datasetProfile: ProfileIdentifier = .automobile
```
1ec5, line 63: Rename to `datasetProfileIdentifier` for consistency with the actual type, which is `ProfileIdentifier`, not `DatasetProfile`.

1ec5 (approved):
This looks good functionally; here are some documentation-related comments.

Sources/MapboxCoreNavigation/TilesetDescriptorFactory.swift
Udumft: What do you mean? I see there are [references in the doc](https://docs.mapbox.com/ios/navigation/api/2.2.0/Extensions/TilesetDescriptorFactory.html#/s:So28MBNNTilesetDescriptorFactoryC20MapboxCoreNavigationE18getSpecificVersion7version15completionQueue0K0ySS_So17OS_dispatch_queueCySo010MBXTilesetB0CctFZ).

PR #3722, opened 2022-02-02
v2.3.0-beta.2

S2Ler:


Cartfile
```diff
@@ -1,4 +1,4 @@
-binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" == 21.0.1
+binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon.json" ~> 21.1.0-rc.1
```
MaximAlien, line 1: IIRC we still pin to exact versions of non-GA releases of our dependencies.

MapboxNavigation.podspec
```diff
@@ -44,7 +44,7 @@ Pod::Spec.new do |s|
   s.module_name = "MapboxNavigation"
 
   s.dependency "MapboxCoreNavigation", "#{s.version.to_s}"
-  s.dependency "MapboxMaps", "10.3.0-beta.1"
+  s.dependency "MapboxMaps", "10.3.0-rc.1"
```
bamx23, line 47: We should either use `~>` here or even `>=` and `<` to match `from:` in SPM or use `exact` in SPM.

Package.swift
```diff
@@ -22,10 +22,10 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(name: "MapboxDirections", url: "https://github.com/mapbox/mapbox-directions-swift.git", from: "2.2.0"),
+        .package(name: "MapboxDirections", url: "https://github.com/mapbox/mapbox-directions-swift.git", from: "2.3.0-beta.2"),
```
bamx23, line 25: `exact`? As it's exact in Cathage and SPM.

bamx23 (approved):

MapboxNavigation.podspec
S2Ler: let's go with exact in spm

Package.swift
S2Ler: makes sense, let's do it

bamx23:
LGTM _(I've also tested for compatibility with the latest Search SDK, everything works fine)_

